PR #4119, opened 2024-11-12
chore: Reduce memory consumption when migrating huge values

### LINKED ISSUE #4100: Big value serialization in migration

chakaz:
Before this PR:

We serialized a `RESTORE` command for each entry into a string, and then push that string to the wire.
This means that, when serializing an entry of size X, we consume 2X memory during the migration.

This PR:

Instead of serializing into a string, we serialize into the wire directly. Luckily, only a small modification is needed in the way we interact with `crc64`, which works well even in chunks.

Fixes #4100

### DIFF
--- a/src/server/rdb_save.cc
+++ b/src/server/rdb_save.cc
@@ -933,14 +933,13 @@ VersionBuffer MakeRdbVersion() {
   return buf;
 }
 
-CrcBuffer MakeCheckSum(std::string_view dump_res) {
-  uint64_t chksum = crc64(0, reinterpret_cast<const uint8_t*>(dump_res.data()), dump_res.size());
+CrcBuffer WrapCheckSum(uint64_t crc) {
   CrcBuffer buf;
-  absl::little_endian::Store64(buf.data(), chksum);
+  absl::little_endian::Store64(buf.data(), crc);
   return buf;
 }
 
-void AppendFooter(io::StringSink* dump_res) {
+void AppendFooter(io::Sink* dump_res, uint64_t crc) {
   auto to_bytes = [](const auto& buf) {
     return io::Bytes(reinterpret_cast<const uint8_t*>(buf.data()), buf.size());
   };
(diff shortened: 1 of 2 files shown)


src/server/rdb_save.cc
```diff
@@ -952,18 +951,45 @@ void AppendFooter(io::StringSink* dump_res) {
...
+    crc = crc64(crc, reinterpret_cast<const uint8_t*>(sv.data()), sv.size());
+
+    return out->Write(io::Buffer(sv));
+  };
+
+  RdbSerializer serializer(compression_mode, [&](size_t size, SerializerBase::FlushState) {
+    auto size_before = total_size;
+    flush(&serializer);
+    DCHECK_EQ(size, total_size - size_before);
+  });
```
kostasrim, line 992: Do we want to flush unconditionally on large values ? (even when big value serialization is off ?)

src/server/rdb_save.cc
```diff
@@ -952,18 +951,45 @@ void AppendFooter(io::StringSink* dump_res) {
...
+    auto ec = srz->FlushToSink(&s, SerializerBase::FlushState::kFlushMidEntry);
+    RETURN_ON_ERR(ec);
+
+    string_view sv = s.str();
+    if (sv.empty()) {
+      return ec;  // Nothing to do
+    }
+
+    total_size += sv.size();
+    crc = crc64(crc, reinterpret_cast<const uint8_t*>(sv.data()), sv.size());
```
kostasrim, line 983: Nice you can accumulate crc based on the previous value + the new serialized

src/server/rdb_save.cc
```diff
@@ -952,18 +951,45 @@ void AppendFooter(io::StringSink* dump_res) {
...
   CompressionMode compression_mode = GetDefaultCompressionMode();
   if (compression_mode != CompressionMode::NONE) {
     compression_mode = CompressionMode::SINGLE_ENTRY;
   }
-  RdbSerializer serializer(compression_mode);
+  uint64_t crc = 0;
+  uint64_t total_size = 0;
+  auto flush = [&](RdbSerializer* srz) {
+    io::StringSink s;
+
```
kostasrim, line 972: Do want to flush directly to a string and then to the sink ? Is there a way to do it directly ?

kostasrim (commented):
Do you think it would worth adding a small test?

kostasrim:
Anddd welcome back :partying_face:

src/server/rdb_save.cc
chakaz: the reason I do it as such is to be able to compute the CRC value for the additional data, so we can flush to the sink directly (otherwise we can't compute the crc)

src/server/rdb_save.cc
chakaz: :nerd_face:

src/server/rdb_save.cc
chakaz: Nice catch! I was under the (false) assumption that this is called only above `serialization_max_chunk_size` threshold. I'll modify this area.

src/server/rdb_save.cc
kostasrim: > (otherwise we can't compute the crc)

Yep that answers it :smile:

BorysTheDev:
@chakaz I think we need the same for journalStreamer/RestoreStreamer

PR #4125, opened 2024-11-13
fix: Use `MOVED` error type for moved replies

### LINKED ISSUE #4118: Too many `MOVED` error in `# Errorstats` section of `INFO ALL`
Dragonfly saves count per error here:

https://github.com/dragonflydb/dragonfly/blob/7815c8d3d7e41aaf84491a7177c48fab8156e284/src/facade/reply_builder.cc#L321-L336

The issue is that, we `type` is empty (which it is for `MOVED` replies), then we save an entry per `str`, which could be 16k in a regular situation, or more if we do slot migrations.

Then, all `INFO` replies fetch this information (from all threads), even if the `ERRORSTATS` section is not requested.

This is a general problem in the way we do error handling, but hopefully most errors contain constant strings instead of slot id + host.

To fix this only for `MOVED` (like I propose) we could simply add some special handling in the above code (if starts with `MOVED` then...)

chakaz:
**The problem:**

When in cluster mode, `MOVED` replies (which are arguably not even errors) are aggregated per slot-id + remote host, and displayed in `# Errorstats` as such. For example, in a server that does _not_ own 8k slots, we will aggregate 8k different errors, and their counts (in memory).

This slows down all `INFO` replies, takes a lot of memory, and also makes `INFO` replies very long.

**The fix:**

Use `type` `MOVED` for moved replies, making them all the same under `# Errorstats`

Fixes #4118

### DIFF
--- a/src/server/zset_family.cc
+++ b/src/server/zset_family.cc
@@ -1281,7 +1281,7 @@ void BZPopMinMax(CmdArgList args, Transaction* tx, SinkReplyBuilder* builder,
     case OpStatus::KEY_MOVED: {
       auto error = cluster::SlotOwnershipErrorStr(*tx->GetUniqueSlotId());
       CHECK(error.has_value());
-      return builder->SendError(std::move(*error));
+      return builder->SendError(std::move(*error), dfly::cluster::kMovedErrorType);
     }
     default:
       LOG(ERROR) << "Unexpected error " << popped_key.status();
--- a/src/server/zset_family.cc
+++ b/src/server/zset_family.cc
@@ -1279,9 +1279,9 @@ void BZPopMinMax(CmdArgList args, Transaction* tx, SinkReplyBuilder* builder,
     case OpStatus::TIMED_OUT:
       return rb->SendNullArray();
     case OpStatus::KEY_MOVED: {
-      auto error = cluster::SlotOwnershipErrorStr(*tx->GetUniqueSlotId());
+      auto error = cluster::SlotOwnershipError(*tx->GetUniqueSlotId());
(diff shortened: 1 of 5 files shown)


src/server/zset_family.cc
```diff
@@ -1279,7 +1279,7 @@ void BZPopMinMax(CmdArgList args, Transaction* tx, SinkReplyBuilder* builder,
     case OpStatus::TIMED_OUT:
       return rb->SendNullArray();
     case OpStatus::KEY_MOVED: {
-      auto error = cluster::SlotOwnershipErrorStr(*tx->GetUniqueSlotId());
+      auto error = cluster::SlotOwnershipError(*tx->GetUniqueSlotId());
```
romange, line 1282: (unrelated to this change)
must say I do not understand the optional return type here. Why do we check for cluster config inside SlotOwnershipError and return optional if KEY_MOVED can only be due to cluster config. better CHECK for cluster config there and remove optional and remove  CHECKs like `CHECK(error.has_value());`

chakaz: I _think_ that the reason is to have the if-cluster-enabled and also if-slot-owned in the same place, and reuse it many times (currently 3 times)

BorysTheDev: When I wrote this logic I didn't want to include additional headers into cluster_defs file, because it is included quite often, so created a string instead of ErrorReply. But if you want to return Reply we can remove the optional, because  ErrorReply contains OK status. Also I think Roman is right and we can remove checking for cluster_config

BorysTheDev (approved):

PR #4137, opened 2024-11-16
chore: optimize info command

romange:
Info command may have a large latency when returning all the sections.
    But often a single section is required. Specifically,
    SERVER and REPLICATION sections are often fetched by clients
    or management components.

    This PR:
    1. Removes any hops for "INFO SERVER" command.
    2. Removes some redundant stats.
    3. Prints latency stats around GetMetrics command if it took to much.

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -2088,7 +2088,7 @@ static void MergeDbSliceStats(const DbSlice::Stats& src, Metrics* dest) {
 
 void ServerFamily::ResetStat(Namespace* ns) {
   shard_set->pool()->AwaitBrief(
-      [registry = service_.mutable_registry(), this, ns](unsigned index, auto*) {
+      [registry = service_.mutable_registry(), ns](unsigned index, auto*) {
         registry->ResetCallStats(index);
         ns->GetCurrentDbSlice().ResetEvents();
         facade::ResetStats();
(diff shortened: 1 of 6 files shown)


src/server/server_family.cc
```diff
@@ -2187,6 +2196,12 @@ Metrics ServerFamily::GetMetrics(Namespace* ns) const {
 
   result.peak_stats = peak_stats_;
 
+  uint64_t delta_ms = (absl::GetCurrentTimeNanos() - start) / 1000'000;
```
chakaz, line 2199:
```suggestion
  uint64_t delta_ms = (absl::GetCurrentTimeNanos() - start) / 1'000'000;
```

super nit

src/server/server_family.cc
```diff
@@ -2187,6 +2196,12 @@ Metrics ServerFamily::GetMetrics(Namespace* ns) const {
...
+  if (delta_ms > 30) {
+    uint64_t cb_dur = (after_cb - start) / 1000'000;
```
chakaz, line 2201: ditto

src/server/server_family.cc
```diff
@@ -2472,15 +2488,17 @@ void ServerFamily::Info(CmdArgList args, Transaction* tx, SinkReplyBuilder* buil
...
-    util::fb2::LockGuard lk(replicaof_mu_);
     // Thread local var is_master is updated under mutex replicaof_mu_ together with replica_,
     // ensuring eventual consistency of is_master. When determining if the server is a replica and
     // accessing the replica_ object, we must lock replicaof_mu_. Using is_master alone is
     // insufficient in this scenario.
-    if (!replica_) {
-      append("role", "master");
-      append("connected_slaves", m.facade_stats.conn_stats.num_replicas);
+    util::fb2::LockGuard lk(replicaof_mu_);
+    bool is_master = bool(!replica_);
```
chakaz, line 2496: why is the `bool` cast needed here? I'd expect the negative (`!`) to convert to boolean?

src/server/server_family.cc
```diff
@@ -2238,11 +2244,21 @@ void ServerFamily::Info(CmdArgList args, Transaction* tx, SinkReplyBuilder* buil
...
-    size_t uptime = m.uptime;
+
+    uint64_t uptime = time(NULL) - start_time_;
     append("uptime_in_seconds", uptime);
     append("uptime_in_days", uptime / (3600 * 24));
   }
 
+  Metrics m;
+  if (section != "SERVER") {
+    m = GetMetrics(cntx->ns);
```
chakaz, line 2255: I'd add a comment saying something like "Save some time on common path which doesn't require metrics"

chakaz (approved):
Generally LGTM, see some nits though

src/server/server_family.cc
romange: probably not needed.

chakaz (dismissed):

PR #4140, opened 2024-11-17
fix: migration ACK response processing

BorysTheDev:
problem: the migration process can get stuck because of the pause between migration finalization attempts for  https://github.com/dragonflydb/dragonfly/pull/4081

fix: try reading the correct ACK response several times until timeout happens

### DIFF
--- a/src/server/cluster/outgoing_slot_migration.cc
+++ b/src/server/cluster/outgoing_slot_migration.cc
@@ -317,21 +317,33 @@ bool OutgoingMigration::FinalizeMigration(long attempt) {
     return false;
   }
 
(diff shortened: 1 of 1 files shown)


chakaz (commented):
I don't quite understand this fix.
With the fix, will attempt to read multiple times. But will we ever send multiple ACKs? I get that if it times out, we may want to try and read again (why not wait longer though?), but in which circumstance will we see more than 1 ACK? If previous ACKs are waiting? If that's the case, there could be other data waiting in line, no? (like regular journal)

BorysTheDev:
@chakaz When get time out we don't know if the data is lost or delayed we try to resend ack but can get the previous response in this case we try to read one more time. In the main connection can't be other data, All other data is sent via flow connections

chakaz (approved):
Makes sense!

PR #4142, opened 2024-11-18
fix(eviction): Tune eviction threshold in cache mode

BagritsevichStepan:
fixes dragonflydb#4139

### DIFF
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -1214,8 +1214,8 @@ auto DbSlice::DeleteExpiredStep(const Context& cntx, unsigned count) -> DeleteEx
     if (ttl <= 0) {
       auto prime_it = db.prime.Find(it->first);
       CHECK(!prime_it.is_done());
-      ExpireIfNeeded(cntx, prime_it, false);
       result.deleted_bytes += prime_it->second.MallocUsed();
+      ExpireIfNeeded(cntx, prime_it, false);
       ++result.deleted;
     } else {
       result.survivor_ttl_sum += ttl;
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -1290,9 +1290,6 @@ pair<uint64_t, size_t> DbSlice::FreeMemWithEvictionStep(DbIndex db_ind, size_t s
 
     auto time_finish = absl::GetCurrentTimeNanos();
     events_.evicted_keys += evicted_items;
-    DVLOG(2) << "Evicted: " << evicted_bytes;
-    DVLOG(2) << "Number of keys evicted / max eviction per hb: " << evicted_items << "/"
-             << max_eviction_per_hb;
     DVLOG(2) << "Eviction time (us): " << (time_finish - time_start) / 1000;
     return pair<uint64_t, size_t>{evicted_items, evicted_bytes};
   };
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -1214,7 +1214,7 @@ auto DbSlice::DeleteExpiredStep(const Context& cntx, unsigned count) -> DeleteEx
     if (ttl <= 0) {
       auto prime_it = db.prime.Find(it->first);
       CHECK(!prime_it.is_done());
-      result.deleted_bytes += prime_it->second.MallocUsed();
+      result.deleted_bytes += prime_it->first.MallocUsed() + prime_it->second.MallocUsed();
       ExpireIfNeeded(cntx, prime_it, false);
       ++result.deleted;
     } else {
@@ -1324,7 +1324,7 @@ pair<uint64_t, size_t> DbSlice::FreeMemWithEvictionStep(DbIndex db_ind, size_t s
           if (record_keys)
             keys_to_journal.emplace_back(key);
 
-          evicted_bytes += evict_it->second.MallocUsed();
+          evicted_bytes += evict_it->first.MallocUsed() + evict_it->second.MallocUsed();
           ++evicted_items;
           PerformDeletion(Iterator(evict_it, StringOrView::FromView(key)), db_table.get());
 
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -1215,6 +1215,7 @@ auto DbSlice::DeleteExpiredStep(const Context& cntx, unsigned count) -> DeleteEx
       auto prime_it = db.prime.Find(it->first);
       CHECK(!prime_it.is_done());
       ExpireIfNeeded(cntx, prime_it, false);
+      result.deleted_bytes += prime_it->second.MallocUsed();
       ++result.deleted;
     } else {
       result.survivor_ttl_sum += ttl;
--- a/src/server/dragonfly_test.cc
+++ b/src/server/dragonfly_test.cc
@@ -455,19 +455,14 @@ TEST_F(DflyEngineTest, OOM) {
 /// Reproduces the case where items with expiry data were evicted,
 /// and then written with the same key.
 TEST_F(DflyEngineTest, Bug207) {
-  max_memory_limit = 5000000;  // 5mb
-
+  max_memory_limit = 1200000;  // 1.2mb = 300000 * 4
   shard_set->TEST_EnableCacheMode();
 
   ssize_t i = 0;
   RespExpr resp;
-  std::string value(1024, 'b');
   for (; i < 10000; ++i) {
-    resp = Run({"setex", StrCat("key", i), "30", value});
+    resp = Run({"setex", StrCat("key", i), "30", "bar"});
     // we evict some items because 5000 is too much when max_memory_limit is 300000.
-    if (resp != "OK") {
-      continue;
-    }
     ASSERT_EQ(resp, "OK");
   }
 
@@ -482,29 +477,28 @@ TEST_F(DflyEngineTest, Bug207) {
   EXPECT_GT(evicted_count(resp.GetString()), 0);
 
   for (; i > 0; --i) {
-    resp = Run({"setex", StrCat("key", i), "30", value});
+    resp = Run({"setex", StrCat("key", i), "30", "bar"});
     ASSERT_EQ(resp, "OK");
   }
 }
 
 TEST_F(DflyEngineTest, StickyEviction) {
-  max_memory_limit = 5000000;  // 5mb
-
+  max_memory_limit = 600000;  // 0.6mb
   shard_set->TEST_EnableCacheMode();
 
-  std::string value(1024, 'b');
+  string tmp_val(100, '.');
 
   ssize_t failed = -1;
-  for (ssize_t i = 0; i < 2000; ++i) {
+  for (ssize_t i = 0; i < 5000; ++i) {
     string key = StrCat("volatile", i);
-    ASSERT_EQ("OK", Run({"set", key, value}));
+    ASSERT_EQ("OK", Run({"set", key, tmp_val}));
   }
 
   bool done = false;
-  for (ssize_t i = 0; !done && i < 7000; ++i) {
+  for (ssize_t i = 0; !done && i < 5000; ++i) {
     string key = StrCat("key", i);
     while (true) {
-      if (Run({"set", key, value}) != "OK") {
+      if (Run({"set", key, tmp_val}) != "OK") {
         failed = i;
         done = true;
         break;
--- a/src/server/dragonfly_test.cc
+++ b/src/server/dragonfly_test.cc
@@ -456,7 +456,6 @@ TEST_F(DflyEngineTest, OOM) {
 /// and then written with the same key.
 TEST_F(DflyEngineTest, Bug207) {
   max_memory_limit = 5000000;  // 5mb
-  ResetService();
 
   shard_set->TEST_EnableCacheMode();
 
@@ -490,7 +489,6 @@ TEST_F(DflyEngineTest, Bug207) {
 
 TEST_F(DflyEngineTest, StickyEviction) {
   max_memory_limit = 5000000;  // 5mb
-  ResetService();
 
   shard_set->TEST_EnableCacheMode();
 
--- a/src/server/dragonfly_test.cc
+++ b/src/server/dragonfly_test.cc
@@ -26,7 +26,7 @@ ABSL_DECLARE_FLAG(float, mem_defrag_threshold);
 ABSL_DECLARE_FLAG(float, mem_defrag_waste_threshold);
 ABSL_DECLARE_FLAG(uint32_t, mem_defrag_check_sec_interval);
 ABSL_DECLARE_FLAG(std::vector<std::string>, rename_command);
-ABSL_DECLARE_FLAG(double, oom_deny_ratio);
+ABSL_DECLARE_FLAG(double, rss_oom_deny_ratio);
 ABSL_DECLARE_FLAG(bool, lua_resp2_legacy_float);
 
 namespace dfly {
@@ -456,19 +456,20 @@ TEST_F(DflyEngineTest, OOM) {
 /// Reproduces the case where items with expiry data were evicted,
 /// and then written with the same key.
 TEST_F(DflyEngineTest, Bug207) {
-  max_memory_limit = 300000;
-
-  absl::FlagSaver fs;
-  absl::SetFlag(&FLAGS_oom_deny_ratio, 4);
+  max_memory_limit = 5000000;  // 5mb
   ResetService();
 
   shard_set->TEST_EnableCacheMode();
 
   ssize_t i = 0;
   RespExpr resp;
+  std::string value(1024, 'b');
   for (; i < 10000; ++i) {
-    resp = Run({"setex", StrCat("key", i), "30", "bar"});
+    resp = Run({"setex", StrCat("key", i), "30", value});
     // we evict some items because 5000 is too much when max_memory_limit is 300000.
+    if (resp != "OK") {
+      continue;
+    }
     ASSERT_EQ(resp, "OK");
   }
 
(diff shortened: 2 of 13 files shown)


tests/dragonfly/generic_test.py
```diff
@@ -149,7 +149,7 @@ async def test_reply_guard_oom(df_factory, df_seeder_factory):
 @pytest.mark.asyncio
 async def test_denyoom_commands(df_factory):
     df_server = df_factory.create(
-        proactor_threads=1, maxmemory="256mb", oom_deny_commands="get", oom_deny_ratio=0.7
+        proactor_threads=1, maxmemory="256mb", oom_deny_commands="get", rss_oom_deny_ratio=0.7
```
adiholden, line 152: you should not replace oom_deny_ratio with  rss_oom_deny_ratio in this test
This test checks that we return oom for denyoom commands
so now we dont have the ration anymore we will return oom when we reach maxmemory set by user

BagritsevichStepan: Yes, will fix

src/server/dragonfly_test.cc
```diff
@@ -26,7 +26,7 @@ ABSL_DECLARE_FLAG(float, mem_defrag_threshold);
 ABSL_DECLARE_FLAG(float, mem_defrag_waste_threshold);
 ABSL_DECLARE_FLAG(uint32_t, mem_defrag_check_sec_interval);
 ABSL_DECLARE_FLAG(std::vector<std::string>, rename_command);
-ABSL_DECLARE_FLAG(double, oom_deny_ratio);
+ABSL_DECLARE_FLAG(double, rss_oom_deny_ratio);
```
adiholden, line 29: nit:
i dont see you use this flag in this file

src/server/engine_shard.cc
```diff
@@ -717,8 +723,17 @@ void EngineShard::RetireExpiredAndEvict() {
     ttl_delete_target = kTtlDeleteLimit * double(deleted) / (double(traversed) + 10);
   }
 
-  ssize_t eviction_redline = size_t(max_memory_limit * kRedLimitFactor) / shard_set->size();
+  const size_t max_rss_memory =
+      max_memory_limit * std::max(ServerState::tlocal()->rss_oom_deny_ratio, 0.0);
+  const ssize_t rss_memory =
+      ConvertNumericOrRound<uint64_t, ssize_t>(rss_mem_current.load(memory_order_relaxed));
```
adiholden, line 729: why do you need this ConvertNumericOrRound?

src/server/engine_shard.cc
```diff
@@ -704,6 +709,7 @@ void EngineShard::RetireExpiredAndEvict() {
   { std::unique_lock lk(db_slice.GetSerializationMutex()); }
   constexpr double kTtlDeleteLimit = 200;
   constexpr double kRedLimitFactor = 0.1;
```
adiholden, line 711: rename to kEvictionFreeMemoryThreshold
add a comment when shard free memory goes below max_memory*threshold we start evicting

src/server/engine_shard.cc
```diff
@@ -717,8 +723,17 @@ void EngineShard::RetireExpiredAndEvict() {
     ttl_delete_target = kTtlDeleteLimit * double(deleted) / (double(traversed) + 10);
   }
 
-  ssize_t eviction_redline = size_t(max_memory_limit * kRedLimitFactor) / shard_set->size();
+  const size_t max_rss_memory =
+      max_memory_limit * std::max(ServerState::tlocal()->rss_oom_deny_ratio, 0.0);
```
adiholden, line 727: if rss_oom_deny_ratio = -1 than max_rss_memory is 0?
I believe we should write
rss_limit = ServerState::tlocal()->rss_oom_deny_ratio > 0? max_memory_limit * ServerState::tlocal()->rss_oom_deny_ratio :  std::numeric_limits<size_t>::max()

src/server/engine_shard.cc
```diff
@@ -736,10 +751,12 @@ void EngineShard::RetireExpiredAndEvict() {
     }
 
     // if our budget is below the limit
-    if (db_slice.memory_budget() < eviction_redline && GetFlag(FLAGS_enable_heartbeat_eviction)) {
+    if ((db_slice.memory_budget() < memory_budget_threshold || rss_memory > rss_memory_threshold) &&
+        GetFlag(FLAGS_enable_heartbeat_eviction)) {
       uint32_t starting_segment_id = rand() % pt->GetSegmentCount();
-      db_slice.FreeMemWithEvictionStep(i, starting_segment_id,
-                                       eviction_redline - db_slice.memory_budget());
+      size_t goal_bytes = std::max(memory_budget_threshold - db_slice.memory_budget(),
```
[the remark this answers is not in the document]
BagritsevichStepan: ssize_t*

BagritsevichStepan:
@adiholden I believe I've identified the issue: I'm calculating rss_threshold for each shard individually, but the current rss_memory value is a global metric for all shards

src/server/engine_shard.cc
```diff
@@ -716,7 +742,12 @@ void EngineShard::RetireExpiredAndEvict() {
     ttl_delete_target = kTtlDeleteLimit * double(deleted) / (double(traversed) + 10);
   }
 
-  ssize_t eviction_redline = size_t(max_memory_limit * kRedLimitFactor) / shard_set->size();
+  // For memory we are using threshold for free memory
+  const ssize_t shard_memory_budget_threshold =
+      ssize_t(max_memory_limit * kEvictionFreeMemoryThreshold) / shard_set->size();
+
+  // Calculate how much rss memory we need to free per shard
+  ssize_t rss_memory_to_free = CalculateRssMemoryToFreePerShard();
```
adiholden, line 750: in the for loop below we call DeleteExpiredStep which can also remove entries, so calculating the rss memory to free here and updating it with released_bytes is not enough.

src/server/engine_shard.cc
```diff
@@ -735,10 +766,17 @@ void EngineShard::RetireExpiredAndEvict() {
...
 
     // if our budget is below the limit
-    if (db_slice.memory_budget() < eviction_redline && GetFlag(FLAGS_enable_heartbeat_eviction)) {
+    if ((db_slice.memory_budget() < shard_memory_budget_threshold || rss_memory_to_free > 0) &&
+        GetFlag(FLAGS_enable_heartbeat_eviction)) {
       uint32_t starting_segment_id = rand() % pt->GetSegmentCount();
-      db_slice.FreeMemWithEvictionStep(i, starting_segment_id,
-                                       eviction_redline - db_slice.memory_budget());
+      size_t goal_bytes =
+          std::max(rss_memory_to_free, shard_memory_budget_threshold - db_slice.memory_budget());
```
adiholden, line 773: I see now that also the old logic for multiple dbs is broken because memory budget is update only when CacheStats is called (once in heartbeat)

src/server/engine_shard.cc
```diff
@@ -198,6 +198,33 @@ optional<uint32_t> GetPeriodicCycleMs() {
   return clock_cycle_ms;
 }
 
+/* Eviction begins when the shard's free memory falls below (max_memory_limit *
+ * kEvictionFreeMemoryThreshold) / <number of shards> */
+constexpr double kEvictionFreeMemoryThreshold = 0.1;
+/* or eviction starts if the used RSS memory exceeds max_memory_limit * kEvictionRssThreshold */
+constexpr double kEvictionRssThreshold = 1.0 - kEvictionFreeMemoryThreshold;
```
adiholden, line 205: I think this constant is kind of redundant
you can do the same for rss as we do for used memory
rss_limit = rss_oom_deny_ratio * max_memory_limit
eviction_rss_threshold = rss_limit * kEvictionFreeMemoryThreshold
curr_free_rss = rss_limit - rss_mem_current
if (curr_free_rss < eviction_rss_threshold) {
 return eviction_rss_threshold - curr_free_rss
}
return 0

src/server/engine_shard.cc
BagritsevichStepan: Yes, I understand it

tests/dragonfly/memory_test.py
```diff
@@ -107,3 +107,74 @@ async def test_rss_oom_ratio(df_factory: DflyInstanceFactory, admin_port):
...
+    key_size = 1024  # 1 mb
+    num_keys = data_fill_size // key_size
+
+    # Fill data to 25% of max memory
+    await async_client.execute_command("DEBUG", "POPULATE", num_keys, "key", key_size)
+
+    await asyncio.sleep(1)  # Wait for RSS heartbeat update
+
+    # First test that eviction is not triggered without connection creation
+    stats_info = await async_client.info("stats")
```
adiholden, line 142: lets also add a check here that used memory is below maxmemory * 0.9 and rss curr memory is below 
maxmemory * rss_oom_deny_ratio * 0.9

tests/dragonfly/memory_test.py
```diff
@@ -107,3 +107,74 @@ async def test_rss_oom_ratio(df_factory: DflyInstanceFactory, admin_port):
...
+    assert stats_info["evicted_keys"] == 0, "No eviction should start yet."
+
+    # Get RSS memory before creating new connections
+    memory_info = await async_client.info("memory")
+    rss_before_connections = memory_info["used_memory_rss"]
+
+    # Increase RSS memory by 30% of max memory
+    # We can simulate RSS increase by creating new connections
+    # Estimate memory per connection
+    estimated_connection_memory = 15 * 1024  # 15 KB per connection
```
adiholden, line 152: To make this test supper stable and more controlled
I would run
CONFIG SET enable_heartbeat_eviction false
before creating the connections
after that check the rss make sure its above the threshold
then run 
CONFIG SET enable_heartbeat_eviction true
and after that check that rss memory dropped below the threshold and check the stats for evicted keys as you did below

src/server/dragonfly_test.cc
```diff
@@ -456,19 +455,19 @@ TEST_F(DflyEngineTest, OOM) {
...
   shard_set->TEST_EnableCacheMode();
 
   ssize_t i = 0;
   RespExpr resp;
+  std::string value(1024, 'b');
   for (; i < 10000; ++i) {
-    resp = Run({"setex", StrCat("key", i), "30", "bar"});
+    resp = Run({"setex", StrCat("key", i), "30", value});
     // we evict some items because 5000 is too much when max_memory_limit is 300000.
+    if (resp != "OK") {
```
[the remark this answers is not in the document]
BagritsevichStepan: No more needed. Just increased memory here. Because sometimes it fails. Its can be faster than heartbeat

src/server/engine_shard.cc
```diff
@@ -198,6 +202,30 @@ optional<uint32_t> GetPeriodicCycleMs() {
   return clock_cycle_ms;
 }
 
+ssize_t CalculateMemoryBudget(size_t max_memory, size_t used_memory) {
+  if (max_memory >= used_memory) {
+    return max_memory - used_memory;
+  }
+  // negative value indicates memory overuse
+  return -1 * ssize_t(used_memory - max_memory);
```
BorysTheDev, line 210: UB because size_t - size_t = size_t

BagritsevichStepan:
```
ssize_t CalculateMemoryBudget(size_t max_memory, size_t used_memory) {
  if (max_memory >= used_memory) {
    return max_memory - used_memory;
  }
  // negative value indicates memory overuse
  return -1 * ssize_t(used_memory - max_memory);
}
```

BorysTheDev: Ok I got, the result is correct because you use overflow of ssize_t type, but it can be UB I'm not sure. please rewrite

src/server/engine_shard.cc
```diff
@@ -198,6 +202,30 @@ optional<uint32_t> GetPeriodicCycleMs() {
...
+}
+
+ssize_t CalculateFreeMemoryOnShard() {
+  // Calculate how much memory is used by all shards
+  const size_t current_global_rss_memory = used_mem_current.load(memory_order_relaxed);
+  return CalculateMemoryBudget(max_memory_limit, current_global_rss_memory) / shard_set->size();
```
BorysTheDev, line 216: you again use ssize_t overflow that can be UB

src/server/engine_shard.cc
BagritsevichStepan: Do you mean to round size_t to max ssize_t? Because I already implemented this previously
https://github.com/dragonflydb/dragonfly/pull/4142#discussion_r1858484944

BorysTheDev: just update ssize_t CalculateMemoryBudget(ssize_t max_memory, ssize_t used_memory)

BagritsevichStepan: But we are still converting size_t to ssize_t before calling the function. This is even more dangerous because we did not perform the subtraction.

BorysTheDev: in our case, such converting is absolutely safe, because we will never get such a memory size  in the near future

BorysTheDev (dismissed):

BagritsevichStepan:
Upd.:
1. Fixed a bug where using ssize_t still caused overflow. I refactored the code and completely removed the use of ssize_t in this context.
2. Identified why `DflyEngineTest.Bug207` was failing. The issue was that we were only evicting items during `AddOrFindInternal`. During each heartbeat, we expected to evict a significant amount of data but successfully evicted 0 bytes. This happened because the strings were small, inlined, and HasAllocated returned false for them (this issue also exists in the main branch). To fix this, I increased the length of the values in the test.
3. Added additional logs for better debugging and monitoring.

BorysTheDev (dismissed):

src/server/db_slice.cc
```diff
@@ -1291,9 +1291,6 @@ pair<uint64_t, size_t> DbSlice::FreeMemWithEvictionStep(DbIndex db_ind, size_t s
 
     auto time_finish = absl::GetCurrentTimeNanos();
     events_.evicted_keys += evicted_items;
-    DVLOG(2) << "Evicted: " << evicted_bytes;
```
adiholden, line 1294: why did you remove this print?

BagritsevichStepan: Moved this to the [Heartbeat](https://github.com/dragonflydb/dragonfly/pull/4142/files#diff-89b55bcf480e65db9741131907708968288c37fd0993ac1680ba2eb7fb28e873R804)

BagritsevichStepan: Because this callback is not always called

src/server/engine_shard.cc
```diff
@@ -716,7 +732,33 @@ void EngineShard::RetireExpiredAndEvict() {
     ttl_delete_target = kTtlDeleteLimit * double(deleted) / (double(traversed) + 10);
   }
 
-  ssize_t eviction_redline = size_t(max_memory_limit * kRedLimitFactor) / shard_set->size();
+  const size_t shards_count = shard_set->size();
+  const double eviction_memory_budget_threshold = GetFlag(FLAGS_eviction_memory_budget_threshold);
+  const double rss_oom_deny_ratio = ServerState::tlocal()->rss_oom_deny_ratio;
+
+  const bool should_evict =
+      eviction_memory_budget_threshold > 0.0 && GetFlag(FLAGS_enable_heartbeat_eviction);
```
adiholden, line 740: why do you check eviction_memory_budget_threshold > 0.0 ?
nit: maybe rename this var should_evict -> eviction_enabled

src/server/engine_shard.cc
```diff
@@ -734,11 +776,35 @@ void EngineShard::RetireExpiredAndEvict() {
...
+
+    const size_t global_used_memory = used_mem_current.load(memory_order_relaxed);
+
+    // Calculate how many bytes we need to evict on this shard
+    size_t goal_bytes = CalculateHowManyBytesToEvictOnshard(max_memory_limit, global_used_memory,
+                                                            shard_memory_budget_threshold);
+
+    if (should_evict_depending_on_rss_memory) {
+      // Calculate how much rss memory is used by all shards
+      io::Result<io::StatusData> sdata_res = io::ReadStatusInfo();
```
adiholden, line 790: this RetireExpiredAndEvict function runs in the heart beat flow that runs every 10ms
today UpdateMemoryGlobalStats runs every 100ms, and I believe there is a reason for this, not to read the rss data so many time. In this logic now you run this read several times in this heartbeat flow (as the count of dbs) and I dont think we want this.
Further more you use above the used_mem_current that is updated in the UpdateMemoryGlobalStats and will not get updated in this for loop

BagritsevichStepan: But used_mem_current can be updated by other threads

src/server/db_slice.cc
adiholden: oh I see you moved the prints to engine shard. ok :+1:

src/server/engine_shard.cc
adiholden: I understand. The flow is indead very confusing!
I suggest to make this simpler by calculating the bytes goal to evict before the forloop on the dbs and once you enter the for loop you have the goal and you update it with the returned evicted byts value from FreeMemWithEvictionStep
In a later stage (not this PR) we can improve to update the rss fetch in this flow so it will be updated in case of eviction step if needed

src/server/engine_shard.cc
```diff
@@ -716,7 +732,33 @@ void EngineShard::RetireExpiredAndEvict() {
...
+  const size_t shards_count = shard_set->size();
+  const double eviction_memory_budget_threshold = GetFlag(FLAGS_eviction_memory_budget_threshold);
+  const double rss_oom_deny_ratio = ServerState::tlocal()->rss_oom_deny_ratio;
+
+  const bool should_evict =
+      eviction_memory_budget_threshold > 0.0 && GetFlag(FLAGS_enable_heartbeat_eviction);
+  // If rss_oom_deny_ratio is set, we should evict depending on rss memory too
+  const bool should_evict_depending_on_rss_memory = should_evict && rss_oom_deny_ratio > 0.0;
+
+  size_t shard_memory_budget_threshold;
```
adiholden, line 744: In addition I believe that it will be more readable if you write a function that will return the byte goal for eviction in this heartbeat invocation. It will have inside all this new logic that you introduced and call this function only if eviction is enabled

src/server/engine_shard.cc
BagritsevichStepan: What does it mean when `eviction_memory_budget_threshold` is less than 0? I believe it indicates that this flag is disabled

src/server/db_slice.cc
```diff
@@ -1217,6 +1217,7 @@ auto DbSlice::DeleteExpiredStep(const Context& cntx, unsigned count) -> DeleteEx
       auto prime_it = db.prime.Find(it->first);
       CHECK(!prime_it.is_done());
       ExpireIfNeeded(cntx, prime_it, false);
+      result.deleted_bytes += prime_it->second.MallocUsed();
```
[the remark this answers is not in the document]
BagritsevichStepan: yes, already fixed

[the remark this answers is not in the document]
adiholden: also you can add the key malloc used to deleted bytes

src/server/db_slice.cc
```diff
@@ -1216,6 +1216,7 @@ auto DbSlice::DeleteExpiredStep(const Context& cntx, unsigned count) -> DeleteEx
     if (ttl <= 0) {
       auto prime_it = db.prime.Find(it->first);
       CHECK(!prime_it.is_done());
+      result.deleted_bytes += prime_it->second.MallocUsed();
```
adiholden, line 1219: also you can add the key malloc used to deleted bytes

src/server/engine_shard.cc
```diff
@@ -716,11 +771,11 @@ void EngineShard::RetireExpiredAndEvict() {
     ttl_delete_target = kTtlDeleteLimit * double(deleted) / (double(traversed) + 10);
   }
 
-  ssize_t eviction_redline = size_t(max_memory_limit * kRedLimitFactor) / shard_set->size();
-
   DbContext db_cntx;
   db_cntx.time_now_ms = GetCurrentTimeMs();
 
+  size_t goal_bytes = GetGoalBytesToEvict();
```
adiholden, line 777: nit:
goal_bytes -> eviction_bytes / eviction_goal
GetGoalBytesToEvict -> CalculateEvictionByets

BagritsevichStepan:
:face_with_head_bandage:

src/server/engine_shard.cc
adiholden: I dont think we need this logic to disable the flag as we already have enable_heartbeat_eviction flag that disables/enables the heartbeat , and this logic with eviction_memory_budget_threshold > 0 is redundant.
In general if you create such logic to a flag than document it in the flag string description

BagritsevichStepan:
Can not reproduce it

adiholden (approved):

PR #4160, opened 2024-11-20
chore: allow slow and precise memory measurement of an object

### LINKED ISSUE #3776: test failed replication_test.py::test_take_over_counters

romange:
Specifically fixes "MEMORY USAGE" for lists.

The  `mi_collect` call may actually fix #3776  At least in my tests it stopped crashing.

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -46,9 +46,15 @@ constexpr XXH64_hash_t kHashSeed = 24061983;
 constexpr size_t kAlignSize = 8u;
 
 // Approximation since does not account for listpacks.
-size_t QlMAllocSize(quicklist* ql) {
-  size_t res = ql->len * sizeof(quicklistNode) + znallocx(sizeof(quicklist));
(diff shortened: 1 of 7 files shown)


src/core/compact_object.cc
```diff
@@ -291,7 +297,7 @@ static_assert(sizeof(CompactObj) == 18);
 
 namespace detail {
 
-size_t RobjWrapper::MallocUsed() const {
+size_t RobjWrapper::MallocUsed(bool slow) const {
```
chakaz, line 300: This works, but isn't very elegant. Why don't you want to cache this size on the object like we do elsewhere?

romange: I am planning to, once we switch to QList. Also with streams it may take some time as well.
Meanwhile I would like to have at least the "slow" variants because performance is not an issue for "memory usage".

chakaz (approved):

PR #4163, opened 2024-11-21
fix(release/helm): allow empty commits for rerun

Pothulapati:


.github/workflows/docker-release.yml
```diff
@@ -90,7 +90,7 @@ jobs:
 
         go test ./contrib/charts/dragonfly/... -update
 
-        git commit \
+        git commit --allow-empty \
```
romange, line 93: maybe instead of comitting an empty commit we can skip the commit? will `|| true` work for this situation?

Pothulapati: Yep, should work. updated.

PR #4170, opened 2024-11-22
fix: remove DenseSet::IteratorBase::TraverseApply

kostasrim:
reverts #4019 because:

1. `DefragStrMap2` calls `ReallocIfNeeded()` on each `StringMap::iterator` and the traversal also includes all of the elements of a all `links`.  This TraverseApply just adds extra iterations, if the `iterator` passed is `link`.
2. Suspect to cause the crash/regrssion in #4167

I am still not sure if `TraverseApply` actually causes the crash described in (2)

### DIFF
--- a/src/core/dense_set.cc
+++ b/src/core/dense_set.cc
@@ -840,22 +840,4 @@ size_t DenseSet::SizeSlow() {
   return size_;
 }
 
-size_t DenseSet::IteratorBase::TraverseApply(DensePtr* ptr, std::function<void(DensePtr*)> fun) {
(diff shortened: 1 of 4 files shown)


kostasrim:
@romange I can also disable the whole defrag logic for string map until we are sure what the root cause is. It's your call

romange:
what version does this revert correspond to?

kostasrim:
> what version does this revert correspond to?

The PR that introduced this change is part of 1.25. Therefore, by removing it we get the behavior of 1.24

romange (approved):

Issue #4173, opened 2024-11-24
Managed mode operation


chakaz:
Some of our responses reveal information that, in a managed setup, should stay hidden.

Do not return replicas in `CLUSTER` sub-cmds (because replicas aren't reachable by clients):
- [x] `CLUSTER NODES` (emulated cluster)
- [x] `CLUSTER NODES` (real cluster)
- [x] `CLUSTER SHARDS` (emulated cluster)
- [x] `CLUSTER SHARDS` (real cluster)
- [x] `CLUSTER SLOTS` (emulated cluster)
- [x] `CLUSTER SLOTS` (real cluster)
- [ ] `INFO REPLICATION` should return the real number of `connected_slaves`, but should not return the list of them

Unrelated to replicas:
- [ ] `INFO SERVER` should not return `thread_count`
- [ ] `INFO SERVER` should also not reveal which architecture (arm/x86_64) it uses

All of the above should be only changed when some flag, say `--managed_service_info` is specified. Also, it should only apply for the regular port, while keeping admin port as is (i.e. reveal all info).

PR #4174, opened 2024-11-24
chore: Hide replicas from `CLUSTER` subcmds in managed mode

chakaz:
Part of #4173 (see for context)

This fixes:

* `CLUSTER NODES`
* `CLUSTER SHARDS`
* `CLUSTER SLOTS`

But only for emulated cluster mode. @romange - should we also handle real cluster (in a future PR)?

### DIFF
--- a/tests/dragonfly/cluster_test.py
+++ b/tests/dragonfly/cluster_test.py
@@ -252,14 +252,17 @@ def verify_slots_result(port: int, answer: list, replicas) -> bool:
     return True
 
 
-@dfly_args({"proactor_threads": 4, "cluster_mode": "emulated"})
+# --managed_service_info means that Dragonfly is running in a managed service, so some details
+# are hidden from users, see https://github.com/dragonflydb/dragonfly/issues/4173
+@dfly_args({"proactor_threads": 4, "cluster_mode": "emulated", "managed_service_info": "true"})
 async def test_emulated_cluster_with_replicas(df_factory):
-    master = df_factory.create(port=BASE_PORT)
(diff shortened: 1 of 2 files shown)


tests/dragonfly/cluster_test.py
```diff
@@ -252,14 +252,17 @@ def is_local_host(ip: str) -> bool:
...
+# are hidden from users, see https://github.com/dragonflydb/dragonfly/issues/4173
+@dfly_args({"proactor_threads": 4, "cluster_mode": "emulated", "managed_service_info": "true"})
 async def test_emulated_cluster_with_replicas(df_factory):
-    master = df_factory.create(port=BASE_PORT)
+    master = df_factory.create(port=BASE_PORT, admin_port=BASE_PORT + 1000)
     replicas = [df_factory.create(port=BASE_PORT + i, logtostdout=True) for i in range(1, 3)]
 
     df_factory.start_all([master, *replicas])
 
     c_master = aioredis.Redis(port=master.port)
```
romange, line 264: please use `master.client()` and `master.admin_client()` instead of explicit `aioredis.Redis`. the latter handles the closing of the connections automatically.

romange:
@chakaz yes, we should do the same for real cluster in future PRs.

romange:
LGTM

tests/dragonfly/cluster_test.py
romange: yeah, you get it built in.

romange (approved):
Thanks for helping out with this!

PR #4185, opened 2024-11-25
chore: produce core files in regtests

romange:
Should work only for self-hosted runners.
The core files will be kept in /var/crash/
We also copy automatically dragonfly binary into /var/crash to be able to debug later.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -18,15 +18,19 @@ jobs:
     container:
       image: ghcr.io/romange/${{ matrix.container }}
       options: --security-opt seccomp=unconfined --sysctl "net.ipv6.conf.all.disable_ipv6=0"
+      volumes:
+        - /var/crash:/var/crash
 
     steps:
       - uses: actions/checkout@v4
         with:
           submodules: true
 
-      - name: Print cpu info
-        run: cat /proc/cpuinfo
-
+      - name: Print environment info
+        run: |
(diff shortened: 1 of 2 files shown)


tests/dragonfly/requirements.txt
```diff
@@ -15,7 +15,7 @@ pytest-repeat==0.9.3
 pymemcache==4.0.0
 prometheus_client==0.17.0
 aiohttp==3.10.2
-numpy==1.24.3
+numpy
```
kostasrim, line 18: Why this change?

.github/workflows/regression-tests.yml
```diff
@@ -56,6 +60,16 @@ jobs:
           name: logs
           path: /tmp/failed/*
 
+      - name: Copy binary on a self hosted runner
+        if: failure()
+        run: |
```
kostasrim, line 65: Not sure if you "tested" this. I can inject a failure to make sure that it works if you can wait until tomorrow.

kostasrim (approved):

tests/dragonfly/requirements.txt
romange: because numpy is better to install via apt and its version depends on each linux distribution.
these tests do not rely on a specific version

.github/workflows/regression-tests.yml
romange: i tested this.

PR #4188, opened 2024-11-26
fix: debug object encoding names

romange:

### DIFF
--- a/src/server/debugcmd.cc
+++ b/src/server/debugcmd.cc
@@ -57,6 +57,7 @@ struct PopulateBatch {
 };
 
 struct ObjInfo {
+  unsigned type = 0;
   unsigned encoding;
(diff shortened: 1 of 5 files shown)


src/server/debugcmd.cc
```diff
@@ -365,6 +367,50 @@ OpResult<ValueCompressInfo> EstimateCompression(ConnectionContext* cntx, string_
...
+      switch (encoding) {
+        case kEncodingJsonCons:
+          return "jsoncons";
+        case kEncodingJsonFlat:
+          return "jsonflat";
+      }
+      break;
+    case OBJ_STREAM:
+      return "stream";
+    default:;
```
BorysTheDev, line 409: I'd remove default

BorysTheDev (approved):

PR #4194, opened 2024-11-26
chore: qlist improvements

romange:
1. Reduce code duplication.
2. Expose qlist node count via "debug object"
3. Add more tests to qlist_test
4. Fix a bug in QList::Erase

### DIFF
--- a/src/core/qlist_test.cc
+++ b/src/core/qlist_test.cc
@@ -21,6 +21,7 @@ namespace dfly {
 
 using namespace std;
 using namespace testing;
+using absl::StrCat;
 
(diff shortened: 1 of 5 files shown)


src/core/qlist_test.cc
```diff
@@ -382,4 +385,132 @@ TEST_P(OptionsTest, DelRangeD) {
   ASSERT_EQ(30, ql_.Size());
 }
 
+TEST_P(OptionsTest, DelRangeNode) {
```
romange, line 388: the tests are adaptation of quicklist.c tests in valkey.

BorysTheDev (approved):

PR #4196, opened 2024-11-26
Feat(Helm chart): Add env vars into pods

andreibe24:
The current change is required to allow use of [S3 bucket backups](https://www.dragonflydb.io/docs/managing-dragonfly/backups#authentication).


chakaz:
@Pothulapati would it be at all possible for you to take a look? :pray:

Pothulapati (changes requested):
Mostly loooks right. 

Can you add a example/test-case in a new values file under the /contrib/charts/dragonfly/ci', run go test ./... --update` and make sure the output is what we expect and won't break with other changes in the future?

Thanks

andreibe24:
Hi @Pothulapati,
I've posted the CI helm tests.

Yours,
Andrei B

romange (approved):

Pothulapati (approved):
LGTM

PR #4222, opened 2024-11-28
feat: add support for big values in SeederV2

### LINKED ISSUE #4157: Improve the seeder to generate different distributions of big values
* remove the serialization_max_chunk_size=1 from regression tests
* add seeder support for generating different distributions of big values
* use the seeder to add tests for big values

Once these are added also revamp the test `test_network_disconnect_during_migration` to use the new seeder. For more info: https://github.com/dragonflydb/dragonfly/pull/4144#discussion_r1850130930

kostasrim:
* add support for big values in SeederV2

### DIFF
--- a/tests/dragonfly/replication_test.py
+++ b/tests/dragonfly/replication_test.py
@@ -81,6 +81,8 @@ async def test_replication_all(
     if big_value:
         args["serialization_max_chunk_size"] = 4096
 
+    args["allocation_tracker"] = "1000000000:40000000000"
+
     master = df_factory.create(admin_port=ADMIN_PORT, proactor_threads=t_master, **args)
     replicas = [
         df_factory.create(admin_port=ADMIN_PORT + i + 1, proactor_threads=t)
--- a/tests/dragonfly/replication_test.py
+++ b/tests/dragonfly/replication_test.py
@@ -62,34 +62,6 @@ Test full replication pipeline. Test full sync with streaming changes and stable
         pytest.param(
             8, [8, 8], dict(key_target=1_000_000, units=16), 50_000, False, marks=M_STRESS
         ),
-        # Quick general test that replication is working
-        (
-            1,
-            3 * [1],
-            dict(
-                key_target=1_000,
-                huge_value_percentage=10,
-                huge_value_size=4096 * 2,
-                collection_size=10,
-                types=SeederBase.BIG_VALUE_TYPES,
-            ),
-            500,
-            True,
-        ),
-        # Big value
-        (
-            4,
-            [4, 4],
-            dict(
-                key_target=10_000,
-                huge_value_percentage=5,
-                huge_value_size=4096 * 4,
-                collection_size=50,
-                types=SeederBase.BIG_VALUE_TYPES,
-            ),
-            500,
-            True,
-        ),
     ],
 )
 @pytest.mark.parametrize("mode", [({}), ({"cache_mode": "true"})])
--- a/tests/dragonfly/replication_test.py
+++ b/tests/dragonfly/replication_test.py
@@ -18,7 +18,6 @@ from .seeder import Seeder as SeederV2
 from . import dfly_args
 from .proxy import Proxy
 from .seeder import StaticSeeder
-from .seeder import SeederBase
 
 ADMIN_PORT = 1211
 
@@ -132,12 +131,6 @@ async def test_replication_all(
 
     # Check data after stable state stream
     await check()
-
-    if big_value:
-        info = await c_master.info()
-        preemptions = info["big_value_preemptions"]
-        logging.info(f"Preemptions {preemptions}")
-        assert preemptions > 0
 
 
 async def check_replica_finished_exec(c_replica: aioredis.Redis, m_offset):
--- a/tests/dragonfly/replication_test.py
+++ b/tests/dragonfly/replication_test.py
@@ -48,19 +48,19 @@ Test full replication pipeline. Test full sync with streaming changes and stable
     [
         # Quick general test that replication is working
         (1, 3 * [1], dict(key_target=1_000), 500, False),
-        (4, [4, 4], dict(key_target=10_000), 1_000, False),
-        pytest.param(6, [6, 6, 6], dict(key_target=100_000), 20_000, False, marks=M_OPT),
-        # Skewed tests with different thread ratio
-        pytest.param(8, 6 * [1], dict(key_target=5_000), 2_000, False, marks=M_SLOW),
-        pytest.param(2, [8, 8], dict(key_target=10_000), 2_000, False, marks=M_SLOW),
-        # Test with big value size
-        pytest.param(2, [2], dict(key_target=1_000, data_size=10_000), 100, False, marks=M_SLOW),
-        # Test with big value and big value serialization
-        pytest.param(2, [2], dict(key_target=1_000, data_size=10_000), 100, True, marks=M_SLOW),
-        # Stress test
-        pytest.param(
-            8, [8, 8], dict(key_target=1_000_000, units=16), 50_000, False, marks=M_STRESS
-        ),
+        #        (4, [4, 4], dict(key_target=10_000), 1_000, False),
+        #        pytest.param(6, [6, 6, 6], dict(key_target=100_000), 20_000, False, marks=M_OPT),
+        #        # Skewed tests with different thread ratio
+        #        pytest.param(8, 6 * [1], dict(key_target=5_000), 2_000, False, marks=M_SLOW),
+        #        pytest.param(2, [8, 8], dict(key_target=10_000), 2_000, False, marks=M_SLOW),
+        #        # Test with big value size
+        #        pytest.param(2, [2], dict(key_target=1_000, data_size=10_000), 100, False, marks=M_SLOW),
+        #        # Test with big value and big value serialization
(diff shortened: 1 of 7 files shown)


tests/dragonfly/seeder/__init__.py
```diff
@@ -137,6 +138,8 @@ def __init__(
         data_size=100,
         collection_size=None,
         types: typing.Optional[typing.List[str]] = None,
+        huge_value_percentage=0,
```
kostasrim, line 141: For now I will keep a flat probability for each key in the key target to contain a `huge value`

tests/dragonfly/seeder/script-genlib.lua
```diff
@@ -62,7 +92,15 @@ function LG_funcs.add_set(key, keys)
         end
         redis.apcall('SDIFFSTORE', key, keys[i1], keys[i2])
     else
-        local elements = dragonfly.randstr(LG_funcs.esize, LG_funcs.csize)
+        local elements
+        if huge_entry() then
+            -- Hard coded 10 here, meaning up to 10 huge entries per set
```
kostasrim, line 97: //TODO so I don't forget to fix it. replace `10` with `LG_funcs.csize()`

tests/dragonfly/seeder/script-generate.lua
```diff
@@ -85,6 +88,7 @@ while true do
     if counter % 10 == 0 then
         -- calculate intensity (not normalized probabilities)
         -- please see attached plots in PR to undertand convergence
+        -- https://github.com/dragonflydb/dragonfly/pull/2556
```
kostasrim, line 91: took me a few seconds to find this, now we can jump straight to the pr if needed (an it will for anyone who plays around this code)

tests/dragonfly/seeder/script-genlib.lua
```diff
@@ -1,9 +1,19 @@
...
-function LG_funcs.init(dsize, csize)
+function LG_funcs.init(dsize, csize, large_val_perc, large_val_sz)
     LG_funcs.dsize = dsize
     LG_funcs.csize = csize
     LG_funcs.esize = math.ceil(dsize / csize)
+    LG_funcs.huge_value_percentage = large_val_perc
+    LG_funcs.huge_value_size = large_val_sz
+end
+
+local function huge_entry()
```
kostasrim, line 11: I would like to expose this as a metric such that once the seeder finishes it will preempt how many big values it created. However, since this code is a script I don't see a "smart way". 

Maybe a seeder can create a key in dragonfly (`set big_values number_of_big_values`) which can then the poll ? 

@chakaz any ideas/thoughts?

tests/dragonfly/seeder/script-genlib.lua
chakaz: please fix :)

chakaz (commented):
Generally LG
I wish we could reduce repetitiveness around the lua script. Perhaps we could add some `get_size()` function that returns `esize` or `large_val_sz` depending on `huge_entry()`'s value, and then it'll use that value instead of all `if`-`else`?

tests/dragonfly/replication_test.py
```diff
@@ -132,6 +161,12 @@ async def check():
     # Check data after stable state stream
     await check()
 
+    if big_value:
+        info = await c_master.info()
+        preemptions = info["big_value_preemptions"]
```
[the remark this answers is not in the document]
kostasrim: > Where is this computed? I couldn't find

It's a new metric I introduced in my other PR. I will remove this for now and we will add it after it;s merged

tests/dragonfly/seeder/script-genlib.lua
kostasrim: I thought about this and we don't really need `scan`. In fact I baked this metric in the lua script which we return -- works perfectly

tests/dragonfly/seeder/script-genlib.lua
```diff
@@ -1,9 +1,43 @@
...
     LG_funcs.dsize = dsize
     LG_funcs.csize = csize
     LG_funcs.esize = math.ceil(dsize / csize)
+    LG_funcs.huge_value_percentage = large_val_perc
+    LG_funcs.huge_value_size = large_val_sz
+end
+
+local huge_entries = 0
+
+local function huge_entry()
```
chakaz, line 13: Don't you want `huge_entry()` to depend on the key? Such that some keys are huge, while others aren't, based on (say) their hash?
The reason I say this is because the seeder uses many operations to generate the values, like many `lpush`, `hset`, etc. If we do 100 operations per key (just throwing numbers here), doing 5% huge will make them all roughly of the same size...

tests/dragonfly/seeder/script-genlib.lua
```diff
@@ -27,23 +53,24 @@ end
 -- lists
 -- store list of random blobs of default container/element sizes
 
-function LG_funcs.add_list(key)
-    local elements = dragonfly.randstr(LG_funcs.esize, LG_funcs.csize)
-    redis.apcall('LPUSH', key, unpack(elements))
+function LG_funcs.add_list(key, keys)
```
chakaz, line 56:
```suggestion
function LG_funcs.add_list(key, huge_keys)
```

tests/dragonfly/seeder/script-generate.lua
```diff
@@ -18,21 +18,45 @@ local total_ops = tonumber(ARGV[6])
...
 
-LG_funcs.init(data_size, collection_size)
+LG_funcs.init(data_size, collection_size, huge_value_percentage, huge_value_size)
 local addfunc = LG_funcs['add_' .. string.lower(type)]
 local modfunc = LG_funcs['mod_' .. string.lower(type)]
+local huge_entries = LG_funcs["get_huge_entries"]
+
+local huge_keys = 0
+
+local function huge_entry()
```
chakaz, line 36: why is this still needed?

tests/dragonfly/seeder/script-genlib.lua
kostasrim: it's not only the huge_keys but all the keys. To see if it's huge we need `keys[key]` which is `true` if it's a huge entry

tests/dragonfly/seeder/script-generate.lua
kostasrim: because we use that to decide if a key is a big value or not and then we store (true|false) on our `keys` index. It';s the same as `hash(key)` but without the hash (since lua does not expose a hash function)

tests/dragonfly/replication_test.py
```diff
@@ -48,19 +48,19 @@ async def wait_for_replicas_state(*clients, state="online", node_role="slave", t
...
+        #        pytest.param(8, 6 * [1], dict(key_target=5_000), 2_000, False, marks=M_SLOW),
+        #        pytest.param(2, [8, 8], dict(key_target=10_000), 2_000, False, marks=M_SLOW),
+        #        # Test with big value size
+        #        pytest.param(2, [2], dict(key_target=1_000, data_size=10_000), 100, False, marks=M_SLOW),
+        #        # Test with big value and big value serialization
+        #        pytest.param(2, [2], dict(key_target=1_000, data_size=10_000), 100, True, marks=M_SLOW),
+        #        # Stress test
+        #        pytest.param(
+        #            8, [8, 8], dict(key_target=1_000_000, units=16), 50_000, False, marks=M_STRESS
+        #        ),
```
chakaz, line 63: Is this intentionally commented out?

tests/dragonfly/seeder/script-generate.lua
```diff
@@ -18,21 +18,46 @@ local total_ops = tonumber(ARGV[6])
...
+    return huge_entry
+end
 
 local function action_add()
     local key = prefix .. tostring(key_counter)
+    local op_type = string.lower(type)
+    local is_huge = false
+    if op_type ~= "string" and op_type ~= "json" then
+      is_huge = huge_entry()
+    end
```
chakaz, line 51: Please add a comment explaining that only `string` and `json` are handled here, because other types are handled below (and where)

tests/dragonfly/seeder_test.py
```diff
@@ -29,7 +29,15 @@ async def check_list():
...
 
-    s = Seeder(units=1, key_target=10, data_size=10_000, collection_size=1, types=["LIST"])
+    s = Seeder(
+        units=1,
+        key_target=10,
+        data_size=10_000,
+        collection_size=1,
+        types=["LIST"],
+        huge_value_percentage=0,
+        huge_value_size=0,
```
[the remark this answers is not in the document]
kostasrim: Because we get `really big containers` which causes the memory to grow really fast. That's why I rather have two specific parameters. One for the size of each element on the container and one for the total elements per container

tests/dragonfly/seeder/script-generate.lua
kostasrim: We don't handle json or string here, we `just roll a dice` to decide if it should be huge value or not. There are no huge values for strings or json so that's why we skip the roll

chakaz: ok, my bad, can you please add that as a comment?

kostasrim: of course!

chakaz (approved):

PR #4225, opened 2024-11-28
refactor: remove redundant allocations for streamer

BorysTheDev:
fixes: #4158

### DIFF
--- a/src/server/journal/journal_test.cc
+++ b/src/server/journal/journal_test.cc
@@ -1,3 +1,4 @@
+#include <random>
 #include <string>
 
 #include "base/gtest.h"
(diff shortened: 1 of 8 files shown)


src/server/journal/streamer.h
```diff
@@ -62,7 +63,9 @@ class JournalStreamer {
   bool IsStalled() const;
 
   journal::Journal* journal_;
-  std::vector<uint8_t> pending_buf_;
+
+  size_t pending_buf_mem_size = 0;
+  std::vector<std::string> pending_buf_;
```
BorysTheDev, line 68: @chakaz Do you have any ideas why I can not use absl::InlineVector<std::string, 4> here? I've tried but everything got broken

BorysTheDev: I've written the question and understand because of SSO

chakaz (commented):
Nice improvement!
Please see some nits, but that's a good change!

src/server/journal/streamer.cc
```diff
@@ -83,50 +83,49 @@ void JournalStreamer::Cancel() {
...
   }
 
-  // If we do not have any in flight requests we send the string right a way.
-  // We can not aggregate it since we do not know when the next update will follow.
-  // because of potential SOO with strings, we allocate explicitly on heap.
-  uint8_t* buf(new uint8_t[str.size()]);
+  const auto len = pending_buf_mem_size;
+  pending_buf_mem_size = 0;
+  auto tmp_pbuf = std::move(pending_buf_);
+  pending_buf_ = {};
```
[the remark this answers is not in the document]
BorysTheDev: I'd prefer to leave it as is because pending_buf_ was moved and I want to call the default constructor

src/server/journal/streamer.cc
```diff
@@ -83,50 +83,49 @@ void JournalStreamer::Cancel() {
 }
 
 size_t JournalStreamer::GetTotalBufferCapacities() const {
```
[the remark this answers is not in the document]
chakaz: `GetBufferSize()`? `InTransitBytes()`? `UsedBytes()`?

src/server/journal/streamer.cc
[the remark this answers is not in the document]
chakaz: ok :shrug:

src/server/journal/streamer.h
```diff
@@ -62,9 +65,59 @@ class JournalStreamer {
...
+        bufs_.emplace_back();
+      }
+      auto& fron_buf = bufs_.back();
+
+      fron_buf.mem_size += str.size();
+      fron_buf.buf.push_back(std::move(str));
+    }
+
+    // should be called to get the next buffer for sending
+    const Buf& PrepareNext() {
```
chakaz, line 97: How about `Peek()` like `std::queue`?

src/server/journal/streamer.h
```diff
@@ -62,9 +65,59 @@ class JournalStreamer {
...
+      if (bufs_.size() == 1) {
+        bufs_.emplace_back();
+      }
```
chakaz, line 100: I feel like it would be more robust to handle this case in `Pop()` below rather than here.
I.e. do in `Pop()` something like: `bufs_.pop_front(); if empty { bufs.emplace_back(); }`

src/server/journal/streamer.cc
```diff
@@ -86,67 +86,58 @@ void JournalStreamer::Cancel() {
...
-    pending_buf_.resize(pending_buf_.size() + str.size());
-    memcpy(pending_buf_.data() + tail, str.data(), str.size());
     return;
   }
 
-  // If we do not have any in flight requests we send the string right a way.
-  // We can not aggregate it since we do not know when the next update will follow.
-  // because of potential SOO with strings, we allocate explicitly on heap.
-  uint8_t* buf(new uint8_t[str.size()]);
+  const auto& cur_buf = pending_buf_.PrepareNext();
```
chakaz, line 102: We could also change the interface slightly, to have `ConsumeNext()` instead of prepare and then pop. This has the added security that we can't accidentally forget to pop, pop twice, nor to modify the prepared buffer (which I think is the most likely to happen)

chakaz: I imagine `ConsumeNext()` to return the buffer by value and remove it from the internal deque.

src/server/journal/streamer.h
```diff
@@ -62,9 +65,59 @@ class JournalStreamer {
...
+      auto& fron_buf = bufs_.back();
+
+      fron_buf.mem_size += str.size();
+      fron_buf.buf.push_back(std::move(str));
+    }
+
+    // should be called to get the next buffer for sending
+    const Buf& PrepareNext() {
+      if (bufs_.size() == 1) {
+        bufs_.emplace_back();
```
chakaz, line 99: Don't we have to call `emplace_back()` *always* if we return the front? Otherwise future `push()`es will modify an outgoing buffer.. no?
But - I feel like my above proposal (see `ConsumeNext()`) is safer and removes this need...

chakaz: And an extra bonus - if you go with the `ConsumeNext()` approach you no longer have to assume that there's always an entry within `bufs_` - it can safely be empty, and `ConsumeNext()` will just return an empty buffer

src/server/journal/streamer.cc
```diff
@@ -86,67 +86,58 @@ void JournalStreamer::Cancel() {
   }
 }
 
-size_t JournalStreamer::GetTotalBufferCapacities() const {
-  return in_flight_bytes_ + pending_buf_.capacity();
+size_t JournalStreamer::UsedBytes() const {
+  return pending_buf_.size();
```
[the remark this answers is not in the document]
BorysTheDev: in_flight_bytes_ is pending_buf_.front.mem_size, how do you want to include it into calculation

src/server/journal/streamer.h
```diff
@@ -62,9 +65,59 @@ class JournalStreamer {
...
-  std::vector<uint8_t> pending_buf_;
-  size_t in_flight_bytes_ = 0, total_sent_ = 0;
 
+  class PendingBuf {
+   public:
+    struct Buf {
+      size_t mem_size = 0;
+      absl::InlinedVector<std::string, 8> buf;
+
+      static constexpr size_t max_buf_size = 1024;  // depends on UIO_MAXIOV
```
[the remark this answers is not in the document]
BorysTheDev: because we do not always have it

src/server/journal/streamer.h
BorysTheDev: because it's not only peek but creates the next element

src/server/journal/streamer.h
BorysTheDev: we can add new items before pop, but we shouldn't update the one that is already sent. So I can do it.

src/server/journal/streamer.cc
BorysTheDev: in this case we need to move into temporary storage and in this case we can not use SSO and inlinedVector

src/server/journal/streamer.h
BorysTheDev: unfortunately it will not work, because you need to move and copy objects so references are invalidated

src/server/journal/journal_test.cc
```diff
@@ -125,5 +126,74 @@ TEST(Journal, WriteRead) {
   }
 }
 
+TEST(Journal, PendingBuf) {
```
chakaz, line 129: Since you have `pending_buf.h`, please create `pending_buf_test.cc` for its tests :nerd_face:

src/server/journal/pending_buf.h
```diff
@@ -0,0 +1,68 @@
...
+  };
+
+  PendingBuf() : bufs_(1) {
+  }
+
+  bool empty() const {
+    return std::all_of(bufs_.begin(), bufs_.end(), [](const auto& b) { return b.buf.empty(); });
+  }
+
+  void push(std::string str) {
```
chakaz, line 34:
```suggestion
  void Push(std::string str) {
```
I get lowercase `empty()` and `size()` to be stl-like, but let's use upper case for not clearly STL container-like methods

src/server/journal/pending_buf.h
```diff
@@ -0,0 +1,68 @@
...
+    }
+    auto& fron_buf = bufs_.back();
+
+    fron_buf.mem_size += str.size();
+    fron_buf.buf.push_back(std::move(str));
+  }
+
+  // should be called to get the next buffer for sending
+  const Buf& PrepareSendingBuf() {
+    if (bufs_.size() == 1) {
```
chakaz, line 47:
```suggestion
    // Adding to the buffer ensures that future `Push()`es will not modify the in-flight buffer
    if (bufs_.size() == 1) {
```

src/server/journal/journal_test.cc
```diff
@@ -125,5 +126,74 @@ TEST(Journal, WriteRead) {
...
+    ASSERT_EQ(sending_buf.mem_size, 14);
+
+    ASSERT_EQ(sending_buf.buf[0], "one");
+    ASSERT_EQ(sending_buf.buf[1], " small");
+    ASSERT_EQ(sending_buf.buf[2], " test");
+  }
+
+  const size_t string_num = 2000;
+  for (size_t i = 0; i < string_num; ++i) {
+    pbuf.push("big_test");
```
chakaz, line 154: Let's have some of the strings larger than SSO
Also above - please replace one of "one", " small", " test" with a larger than SSO string

src/server/journal/pending_buf.h
```diff
@@ -0,0 +1,68 @@
...
+namespace dfly {
+
+class PendingBuf {
+ public:
+  struct Buf {
+    size_t mem_size = 0;
+    absl::InlinedVector<std::string, 8> buf;
+
+#ifdef UIO_MAXIOV
+    static constexpr size_t max_buf_size = UIO_MAXIOV;
```
chakaz, line 21:
```suggestion
    static constexpr size_t kMaxBufSize = UIO_MAXIOV;
```

src/server/journal/pending_buf.h
```diff
@@ -0,0 +1,68 @@
...
+#elif
+    static constexpr size_t max_buf_size = 1024;
```
chakaz, line 23: Also here

src/server/journal/journal_test.cc
```diff
@@ -125,5 +126,74 @@ TEST(Journal, WriteRead) {
...
+    auto& sending_buf = pbuf.PrepareSendingBuf();
+    ASSERT_EQ(sending_buf.buf.size(), 3);
+    ASSERT_EQ(sending_buf.mem_size, 14);
+
+    ASSERT_EQ(sending_buf.buf[0], "one");
+    ASSERT_EQ(sending_buf.buf[1], " small");
+    ASSERT_EQ(sending_buf.buf[2], " test");
+  }
+
+  const size_t string_num = 2000;
```
chakaz, line 152: Please add
`EXPECT_LT(string_num, PendingBuf::Buf::max_buf_size);`
Then below use `string_num` or `PendingBuf::Buf::max_buf_size` directly, instead of for example `std::min()`

src/server/journal/journal_test.cc
```diff
@@ -125,5 +126,74 @@ TEST(Journal, WriteRead) {
...
+    ASSERT_EQ(sending_buf.mem_size, next_buf_size * 8);
+
+    for (const auto& s : sending_buf.buf) {
+      ASSERT_EQ(s, "big_test");
+    }
+  }
+
+  pbuf.Pop();
+
+  if (next_buf_size < string_num) {
```
chakaz, line 178: Why do we have this `if`?
Tests should be straightforward, and easily explain what they cover
Please remove it (and keep the inside of it)

src/server/journal/journal_test.cc
BorysTheDev: UIO_MAXIOV can be more than 1024 so I would prefer to leave it as is, and we will test different scenarios on the different platforms

BorysTheDev: I've read your next comment and I think I know how to rewrite it

src/server/journal/journal_test.cc
```diff
@@ -125,5 +127,107 @@ TEST(Journal, WriteRead) {
   }
 }
 
+std::string rand_str() {
```
chakaz, line 130: fwiw, we already have `GetRandomHex()`

src/server/journal/journal_test.cc
```diff
@@ -125,5 +127,107 @@ TEST(Journal, WriteRead) {
...
+      std::accumulate(test_data.begin(), test_data.end(), 0,
+                      [](size_t size, const auto& s) { return s.size() + size; });
+
+  ASSERT_FALSE(pbuf.Empty());
+  ASSERT_EQ(pbuf.Size(), last_buf_size);
+
+  {
+    auto& sending_buf = pbuf.PrepareSendingBuf();
+
+    ASSERT_EQ(sending_buf.buf.size(), string_num - PendingBuf::Buf::kMaxBufSize);
```
chakaz, line 218:
```suggestion
    ASSERT_EQ(sending_buf.buf.size(), 1000);
```
right?

chakaz (commented):
LGTM!

src/server/journal/journal_test.cc
BorysTheDev: I would prefer to leave as is

chakaz (approved):

chakaz (approved):

PR #4228, opened 2024-11-29
fix(rax_tree): Fix crash caused by element destruction in RaxTreeMap

BagritsevichStepan:
fixes dragonflydb#4172

This is the simplest solution. However, I propose a different solution that will invoke `raxFreeWithCallback`. The problem is that this will require a capturing lambda, as we need to pass `alloc_`. Therefore, we need to copy and update the `raxFreeWithCallback` method for capturing lambdas.

Additionally, I believe that the `SeekIterator` should be removed, as it seems the issue was caused by it


BagritsevichStepan:
https://github.com/dragonflydb/dragonfly/actions/runs/12082557873

romange:
Can you please explain what the problem is in the PR description?
i.e. `the cause of crash is ..., this solution solves the problem by .... , but i suggest the following solution ...`

src/core/search/rax_tree.h
```diff
@@ -93,10 +93,12 @@ template <typename V> struct RaxTreeMap {
   }
 
   ~RaxTreeMap() {
-    for (auto it = begin(); it != end(); ++it) {
-      V* ptr = &(*it).second;
-      std::allocator_traits<decltype(alloc_)>::destroy(alloc_, ptr);
-      alloc_.deallocate(ptr, 1);
+    while (true) {
```
kostasrim, line 96: There are two things that concern me here:

1. Looking at how `raxRecursiveFree` is implemented it calls `free_callback(raxGetData(n)` and what I introduced before seems to be on par in that respect because `V* ptr = &(*it).second;` is actually the result of `raxGetData(n)`.
2. Since we can reproduce this we should also include an isolated test to assert the root cause of the issue (my mistake not to suggest this earlier)

I would like to understand why deleting the result of `raxGetData(n)` result in segfault

BagritsevichStepan: raxFree passes NULL to raxResursiveFree

BagritsevichStepan: So it's not called there

BagritsevichStepan: I'm still not sure what causes problem. But for sure it's not correct that we are just changing iterator data without any redis method and then passing it to the raxNext

romange: @BagritsevichStepan do you know how to reproduce this?

BagritsevichStepan: Test was added

BagritsevichStepan:
> Can you please explain what the problem is in the PR description? i.e. `the cause of crash is ..., this solution solves the problem by .... , but i suggest the following solution ...`

Done

BagritsevichStepan:
@kostasrim  Run it again: https://github.com/dragonflydb/dragonfly/actions/runs/12140115454

romange (changes requested):

romange:
@BagritsevichStepan I agree that the interface of `raxFreeWithCallback` should be changed by accepting an additional `void* arg` argument that is passed by `raxFreeWithCallback` to each invocation of `free_callback`

BagritsevichStepan:
I decided to introduce Callback calling in the another PR #4255. I don't sure if we want to merge it before next release

src/core/search/rax_tree.h
```diff
@@ -24,30 +24,37 @@ template <typename V> struct RaxTreeMap {
...
+    SeekIterator& operator=(const SeekIterator&) = delete;
+
+    ~SeekIterator() {
+      if (IsValid()) {
+        InvalidateIterator();
+      }
+    }
+
     bool operator==(const SeekIterator& rhs) const {
+      if (!IsValid() || !rhs.IsValid())
```
romange, line 56: why do you prefer this notation?
you do not find `it_.rt == other.rt &&  it_.node == rhs.it_.node;` more readable?

romange (commented):
LGTM

src/core/search/rax_tree.h
BagritsevichStepan: But this is not true. If both iterators are invalid, `it_.rt == rhs.it_.rt,` but `it_.node == rhs.it_.node` does not always hold true. We rely on this behavior when comparing a `FindIterator` with `RaxTreeMap::end()`, where `it_.node` is not set. And also we don't know where raxFind was stopped. In other words, we don't know the node that `it_.node` points to after `raxFind`.

BagritsevichStepan:
> have you run on the traces you have to see it's working well?

Yes, https://github.com/dragonflydb/dragonfly/actions/runs/12162671806

src/core/search/rax_tree.h
BagritsevichStepan: So it should be `(!it_.rt && !other.rt) || (it_.rt && other.rt && it_.node == rhs.it_.node)` . I don't think it is more readable

romange (approved):

Issue #3631, opened 2024-09-03
FT.AGGREGATE does not fully support the SORTBY option


BagritsevichStepan:
Dragonfly does not parse the following command:
```
FT.AGGREGATE index "*" GROUPBY 0 REDUCE COUNT 0 AS count SORTBY 2 @count DESC MAX 100
```

rcl-viveksharma:
same issue

rcl-viveksharma:
any update

BagritsevichStepan:
> any update

Hi Vivek, I will push the changes in one/two weeks. The fix should be released in the next version.

rcl-viveksharma:
thanks mate waiting for it

rcl-viveksharma:
any update mate ?

romange:
Hello @rcl-viveksharma .  no need to ask us for status updates every few days.

rcl-viveksharma:
we are waiting for this aggregate, for a month, thats why

rcl-viveksharma:
hai @BagritsevichStepan was this merged

PR #2413, opened 2024-01-13
Basic FT.AGGREGATE

dranikpg:
Basic aggregate
* GROUPBY
* SORTBY
* LIMIT

PR #4231, opened 2024-11-29
fix(search_family): Remove the output of extra fields in the FT.AGGREGATE command

BagritsevichStepan:
fixes dragonflydb#4230

First PR for complete support of the SORTBY option (#3631)

To see examples please check the issue.
Add code that checks, during aggregate steps, what fields should be printed.

### DIFF
--- a/src/server/search/aggregator.cc
+++ b/src/server/search/aggregator.cc
@@ -11,10 +11,10 @@ namespace dfly::aggregate {
 namespace {
 
 struct GroupStep {
-  PipelineResult operator()(std::vector<DocValues> values) {
+  PipelineResult operator()(PipelineResult result) {
     // Separate items into groups
     absl::flat_hash_map<absl::FixedArray<Value>, std::vector<DocValues>> groups;
-    for (auto& value : values) {
+    for (auto& value : result.values) {
       groups[Extract(value)].push_back(std::move(value));
     }
 
@@ -28,7 +28,18 @@ struct GroupStep {
       }
       out.push_back(std::move(doc));
     }
-    return out;
+
+    absl::flat_hash_set<std::string> fields_to_print;
+    fields_to_print.reserve(fields_.size() + reducers_.size());
+
+    for (auto& field : fields_) {
+      fields_to_print.insert(std::move(field));
+    }
+    for (auto& reducer : reducers_) {
+      fields_to_print.insert(std::move(reducer.result_field));
+    }
+
+    return {std::move(out), std::move(fields_to_print)};
   }
 
   absl::FixedArray<Value> Extract(const DocValues& dv) {
@@ -104,34 +115,42 @@ PipelineStep MakeGroupStep(absl::Span<const std::string_view> fields,
 }
 
 PipelineStep MakeSortStep(std::string_view field, bool descending) {
-  return [field = std::string(field), descending](std::vector<DocValues> values) -> PipelineResult {
+  return [field = std::string(field), descending](PipelineResult result) -> PipelineResult {
+    auto& values = result.values;
+
     std::sort(values.begin(), values.end(), [field](const DocValues& l, const DocValues& r) {
       auto it1 = l.find(field);
       auto it2 = r.find(field);
       return it1 == l.end() || (it2 != r.end() && it1->second < it2->second);
     });
-    if (descending)
+
+    if (descending) {
       std::reverse(values.begin(), values.end());
-    return values;
+    }
+
+    result.fields_to_print.insert(field);
+    return result;
   };
 }
 
 PipelineStep MakeLimitStep(size_t offset, size_t num) {
-  return [offset, num](std::vector<DocValues> values) -> PipelineResult {
+  return [offset, num](PipelineResult result) {
+    auto& values = result.values;
     values.erase(values.begin(), values.begin() + std::min(offset, values.size()));
     values.resize(std::min(num, values.size()));
-    return values;
+    return result;
   };
 }
 
-PipelineResult Process(std::vector<DocValues> values, absl::Span<const PipelineStep> steps) {
+PipelineResult Process(std::vector<DocValues> values,
+                       absl::Span<const std::string_view> fields_to_print,
+                       absl::Span<const PipelineStep> steps) {
+  PipelineResult result{std::move(values), {fields_to_print.begin(), fields_to_print.end()}};
   for (auto& step : steps) {
-    auto result = step(std::move(values));
-    if (!result.has_value())
-      return result;
-    values = std::move(result.value());
+    PipelineResult step_result = step(std::move(result));
+    result = std::move(step_result);
   }
-  return values;
+  return result;
 }
 
 }  // namespace dfly::aggregate
--- a/src/server/search/aggregator.h
+++ b/src/server/search/aggregator.h
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <absl/container/flat_hash_map.h>
+#include <absl/container/flat_hash_set.h>
 #include <absl/types/span.h>
 
 #include <string>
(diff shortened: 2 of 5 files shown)


src/server/search/aggregator.h
```diff
@@ -19,10 +20,16 @@ namespace dfly::aggregate {
 using Value = ::dfly::search::SortableValue;
 using DocValues = absl::flat_hash_map<std::string, Value>;  // documents sent through the pipeline
 
-// TODO: Replace DocValues with compact linear search map instead of hash map
+struct PipelineResult {
+  // Values to be passed to the next step
+  // TODO: Replace DocValues with compact linear search map instead of hash map
```
romange, line 25: is this TODO relevant? I do not see a hash map here

src/server/search/aggregator.cc
```diff
@@ -104,34 +115,42 @@ PipelineStep MakeGroupStep(absl::Span<const std::string_view> fields,
...
+    auto& values = result.values;
     values.erase(values.begin(), values.begin() + std::min(offset, values.size()));
     values.resize(std::min(num, values.size()));
-    return values;
+    return result;
   };
 }
 
-PipelineResult Process(std::vector<DocValues> values, absl::Span<const PipelineStep> steps) {
+PipelineResult Process(std::vector<DocValues> values,
```
romange, line 145: Please put your prod-duty hat, and think about all the VLOG statements you would like to have, if we have problems in production with this code.

src/server/search/aggregator.cc
```diff
@@ -104,34 +115,42 @@ PipelineStep MakeGroupStep(absl::Span<const std::string_view> fields,
...
       auto it1 = l.find(field);
       auto it2 = r.find(field);
       return it1 == l.end() || (it2 != r.end() && it1->second < it2->second);
     });
-    if (descending)
+
+    if (descending) {
       std::reverse(values.begin(), values.end());
-    return values;
+    }
```
BorysTheDev, line 129: instead of reverse, you can do correct sorting

src/server/search/aggregator.cc
```diff
@@ -104,34 +115,42 @@ PipelineStep MakeGroupStep(absl::Span<const std::string_view> fields,
...
     values.erase(values.begin(), values.begin() + std::min(offset, values.size()));
     values.resize(std::min(num, values.size()));
-    return values;
+    return result;
   };
 }
 
-PipelineResult Process(std::vector<DocValues> values, absl::Span<const PipelineStep> steps) {
+PipelineResult Process(std::vector<DocValues> values,
+                       absl::Span<const std::string_view> fields_to_print,
```
BorysTheDev, line 146: const string_view doesn't make any sense

src/server/search/aggregator.cc
```diff
@@ -104,34 +115,42 @@ PipelineStep MakeGroupStep(absl::Span<const std::string_view> fields,
...
+                       absl::Span<const PipelineStep> steps) {
```
BorysTheDev, line 147: the same for const PipelineStep

src/server/search/aggregator.cc
```diff
@@ -104,34 +115,42 @@ PipelineStep MakeGroupStep(absl::Span<const std::string_view> fields,
...
+  PipelineResult result{std::move(values), {fields_to_print.begin(), fields_to_print.end()}};
   for (auto& step : steps) {
-    auto result = step(std::move(values));
-    if (!result.has_value())
-      return result;
-    values = std::move(result.value());
+    PipelineResult step_result = step(std::move(result));
+    result = std::move(step_result);
```
BorysTheDev, line 151: I would prefer to see step method modifying the result instead of all these moves

src/server/search/aggregator.h
```diff
@@ -19,10 +20,16 @@ namespace dfly::aggregate {
...
+  std::vector<DocValues> values;
 
-using PipelineResult = io::Result<std::vector<DocValues>, facade::ErrorReply>;
-using PipelineStep = std::function<PipelineResult(std::vector<DocValues>)>;  // Group, Sort, etc.
+  // Fields from values to be printed
+  absl::flat_hash_set<std::string> fields_to_print;
```
BorysTheDev, line 29: why flat_hash_set

src/server/search/search_family.cc
```diff
@@ -980,22 +980,35 @@ void SearchFamily::FtAggregate(CmdArgList args, Transaction* tx, SinkReplyBuilde
...
-      rb->SendBulkString(k);
-      std::visit(sortable_value_sender, v);
+  const size_t field_count = agg_results.fields_to_print.size();
+  for (const auto& value : agg_results.values) {
+    rb->StartArray(field_count * 2);
+    for (const auto& field : agg_results.fields_to_print) {
+      rb->SendBulkString(field);
+
+      auto it = value.find(field);
+      if (it != value.end()) {
```
BorysTheDev, line 1007: nit: if (auto it = value.find(field); it != value.end())

src/server/search/aggregator.cc
BagritsevichStepan: It was added by Vlad and is fixed in the [second PR](https://github.com/dragonflydb/dragonfly/pull/4232/files#diff-aa06be8c9cd1ea3b09119db79df35e7f31b52e92aa6a6a3fe1c5f83f4b9e0269R133). So I don't think we need to change it here

src/server/search/aggregator.cc
BagritsevichStepan: Do you mean that we need to refactor this logic and have something like `Aggregator` that stores current result and has methods like `DoSort`, `DoReduce`? Or use pointer here

src/server/search/aggregator.h
BagritsevichStepan: This container mainly holds a small number of elements, which makes `flat_hash_set` a good fit here. Do you have any other alternatives?

src/server/search/aggregator.h
BagritsevichStepan: Is relevant but I moved it to the wrong place. It relates to the line `using DocValues = absl::flat_hash_map<std::string, Value>;`. Was added [here](https://github.com/dragonflydb/dragonfly/pull/2413/files#diff-446ca7a8607d8aef6dd49edb8246b10c8ced16685508ad313c7b069d2b5024f1R22)

src/server/search/aggregator.cc
BorysTheDev: yes

src/server/search/aggregator.h
BorysTheDev: why not simple vector or even InlinedVector

BagritsevichStepan: In the second PR, we are adding multiple fields to the `fields_to_print` in the SORTBY option, which may already exist there

src/server/search/aggregator.cc
BorysTheDev: What about this question? to refactor steps and modify the result instead of moving it every time

BagritsevichStepan: Yes, I spoke about it today. I'm finalizing it

BorysTheDev: Are you going to create another PR?

BagritsevichStepan: I think that it makes sense to create another PR

BorysTheDev (approved):

BagritsevichStepan:
Just fixed the nit

BorysTheDev:
> Just fixed the nit

If you address comments please do an additional commit instead of amend and force push, because it's easier to review.

BorysTheDev (approved):

src/server/search/aggregator.cc
BagritsevichStepan: #4290

src/server/search/aggregator.cc
BagritsevichStepan: #4290

PR #4232, opened 2024-11-29
fix(search_family): Support multiple fields in SORTBY option in the FT.AGGREGATE command. SECOND PR

### LINKED ISSUE #3631: FT.AGGREGATE does not fully support the SORTBY option

BagritsevichStepan:
fixes #3631

Second PR for complete support of the SORTBY option. Please review first #4231

### DIFF
--- a/src/server/search/aggregator.cc
+++ b/src/server/search/aggregator.cc
@@ -65,9 +65,9 @@ void Aggregator::DoGroup(absl::Span<const std::string> fields, absl::Span<const 
   }
 }
 
-void Aggregator::DoSort(std::string_view field, bool descending) {
+void Aggregator::DoSort(const SortParams& sort_params) {
   /*
-    Comparator for sorting DocValues by field.
+    Comparator for sorting DocValues by fields.
     If some of the fields is not present in the DocValues, comparator returns:
     1. l_it == l.end() && r_it != r.end()
       asc -> false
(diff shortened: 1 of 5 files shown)


src/server/search/aggregator.cc
```diff
@@ -11,10 +11,10 @@ namespace dfly::aggregate {
 namespace {
 
 struct GroupStep {
-  PipelineResult operator()(std::vector<DocValues> values) {
+  PipelineResult operator()(PipelineResult result) {
```
BorysTheDev, line 14: PipelineResult contains more than result.values but you use only this field, maybe you don't need to change the type of parameter

src/server/search/aggregator.h
```diff
@@ -19,10 +20,16 @@ namespace dfly::aggregate {
...
+  // TODO: Replace DocValues with compact linear search map instead of hash map
+  std::vector<DocValues> values;
 
-using PipelineResult = io::Result<std::vector<DocValues>, facade::ErrorReply>;
-using PipelineStep = std::function<PipelineResult(std::vector<DocValues>)>;  // Group, Sort, etc.
+  // Fields from values to be printed
+  absl::flat_hash_set<std::string> fields_to_print;
+};
+
+using PipelineStep = std::function<PipelineResult(PipelineResult)>;  // Group, Sort, etc.
```
BorysTheDev, line 32: Consider modifying current values instead of doing move

src/server/search/aggregator.h
```diff
@@ -66,6 +73,25 @@ struct Reducer {
...
+  constexpr static int64_t kSortAll = -1;
+
+  bool SortAll() const {
+    return max == kSortAll;
+  }
+
+  /* Fields to sort by. If multiple fields are provided, sorting works hierarchically:
+     - First, the i-th field is compared.
+     - If the i-th field values are equal, the (i + 1)-th field is compared, and so on. */
+  absl::InlinedVector<std::pair<std::string, SortOrder>, 1> fields;
```
BorysTheDev, line 88: I would suggest using a bigger default capacity

src/server/search/aggregator.cc
```diff
@@ -80,22 +80,41 @@ void Aggregator::DoSort(std::string_view field, bool descending) {
...
+  } else {
+    DCHECK_GE(sort_params.max, 0);
+    const size_t limit = std::min(values.size(), size_t(sort_params.max));
+    std::partial_sort(values.begin(), values.begin() + limit, values.end(), comparator);
+    values.resize(limit);
+  }
 
-  result.fields_to_print.insert(field);
+  for (auto& field : sort_params.fields) {
+    result.fields_to_print.insert(field.first);
```
BorysTheDev, line 116: I believe instead of a loop you should use insert(begin, end)

BorysTheDev (approved):

src/server/search/aggregator.cc
BagritsevichStepan: The problem is that` sort_params.fields` is vector of `pair<string, SortOrder>`, `result.fields_to_print` is set of `string_view`. So we need to split `sort_params` to two vectors. I don't think it worth to do it so

src/server/search/aggregator.cc
```diff
@@ -80,22 +80,41 @@ void Aggregator::DoSort(std::string_view field, bool descending) {
...
-    // If some of the values is not present
-    if (l_it == l.end() || r_it == r.end()) {
-      return l_it != l.end();
+    for (const auto& [field, order] : sort_params.fields) {
+      auto l_it = l.find(field);
+      auto r_it = r.find(field);
+
+      // If some of the values is not present
+      if (l_it == l.end() || r_it == r.end()) {
+        if (l_it == l.end() && r_it == r.end()) {
```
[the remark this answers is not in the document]
BagritsevichStepan: But this is two different containers. l and r

src/server/search/aggregator.cc
BorysTheDev: you are right

PR #4235, opened 2024-12-01
chore: update command interface for bitops, bloom, zset families

romange:

### DIFF
--- a/src/server/bitops_family.cc
+++ b/src/server/bitops_family.cc
@@ -39,13 +39,13 @@ using BitsStrVec = vector<string>;
 
 // The following is the list of the functions that would handle the
 // commands that handle the bit operations
-void BitPos(CmdArgList args, Transaction* tx, SinkReplyBuilder* builder);
-void BitCount(CmdArgList args, Transaction* tx, SinkReplyBuilder* builder);
-void BitField(CmdArgList args, Transaction* tx, SinkReplyBuilder* builder);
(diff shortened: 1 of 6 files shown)


src/server/bitops_family.cc
```diff
@@ -39,13 +39,13 @@ using BitsStrVec = vector<string>;
...
-void BitOp(CmdArgList args, Transaction* tx, SinkReplyBuilder* builder);
-void GetBit(CmdArgList args, Transaction* tx, SinkReplyBuilder* builder);
-void SetBit(CmdArgList args, Transaction* tx, SinkReplyBuilder* builder);
+void BitPos(CmdArgList args, const CommandContext& cmd_cntx);
+void BitCount(CmdArgList args, const CommandContext& cmd_cntx);
+void BitField(CmdArgList args, const CommandContext& cmd_cntx);
+void BitFieldRo(CmdArgList args, const CommandContext& cmd_cntx);
+void BitOp(CmdArgList args, const CommandContext& cmd_cntx);
+void GetBit(CmdArgList args, const CommandContext& cmd_cntx);
+void SetBit(CmdArgList args, const CommandContext& cmd_cntx);
```
BorysTheDev, line 48: I don't think that const for CommandContext is ok, because we update it

BorysTheDev (approved):

src/server/bitops_family.cc
romange: we never change any of the fields of the struct, otherwise it would not compile. but I understand what you are saying.

PR #4240, opened 2024-12-02
fix: stream bugs

romange:
This PR syncs some of the improvements that were introduced in streams in Redis 7.2.3 OSS.

1. verify xsetid against max deleted id in the stream
2. Implement precise memory measurement of streams for "memory usage" command.
3. Always compact nodes in stream listpacks after creating new nodes.


src/core/compact_object.cc
```diff
@@ -125,6 +125,95 @@ size_t MallocUsedZSet(unsigned encoding, void* ptr) {
...
+size_t streamRadixTreeMemoryUsage(rax* rax) {
+  size_t size = sizeof(*rax);
+  size = rax->numele * sizeof(streamID);
+  size += rax->numnodes * sizeof(raxNode);
+  /* Add a fixed overhead due to the aux data pointer, children, ... */
+  size += rax->numnodes * sizeof(long) * 30;
+  return size;
+}
+
+size_t MallocUsedStream(stream* s) {
```
kostasrim, line 151: I did introduce the exact same function from Valkey and then removed it to replace it with the MemotryTracker :smile:

kostasrim (approved):

chakaz (approved):

src/core/compact_object.cc
romange: this way we have a verification mechanism that MemotryTracker captures all the cases - we can run an external script that scans all the stream keys and calls "memory usage" to aggregate the sum

PR #4241, opened 2024-12-02
chore: Hide replica info in real cluster if `--managed_service_info`

chakaz:
So far we only handled emulated cluster. This PR adds real cluster support.

Related to #4173

### DIFF
--- a/src/server/main_service.cc
+++ b/src/server/main_service.cc
@@ -815,6 +815,7 @@ void Service::Init(util::AcceptServer* acceptor, std::vector<facade::Listener*> 
   config_registry.RegisterMutable("replication_timeout");
   config_registry.RegisterMutable("table_growth_margin");
   config_registry.RegisterMutable("tcp_keepalive");
+  config_registry.RegisterMutable("managed_service_info");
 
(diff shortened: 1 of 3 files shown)


src/server/main_service.cc
```diff
@@ -815,6 +815,7 @@ void Service::Init(util::AcceptServer* acceptor, std::vector<facade::Listener*>
   config_registry.RegisterMutable("replication_timeout");
   config_registry.RegisterMutable("table_growth_margin");
   config_registry.RegisterMutable("tcp_keepalive");
+  config_registry.RegisterMutable("managed_service_info");
```
romange, line 818: I do not think we should add all new flags here.

romange (approved):

src/server/main_service.cc
chakaz: It helps with the test, but I guess that's not a good excuse to put here :)
On the other hand, I don't think there's a real risk / down side to it either

chakaz: I'll merge now, but if you'd like me to remove this please lmk

Issue #4215, opened 2024-11-28
rdb_test unit test LoadStream2 failed


kostasrim:
https://github.com/dragonflydb/dragonfly/actions/runs/12065815330/job/33645375727#step:10:22705

```
2024-11-28T09:36:48.1561541Z 63: E20241128 09:36:48.155498 26092 server_family.cc:1151] Rdb load failed: Out of memory, or used memory is too high
2024-11-28T09:36:48.1672290Z 63: [       OK ] RdbTest.SnapshotTooBig (65 ms)
2024-11-28T09:36:48.1672873Z 63: [ RUN      ] RdbTest.LoadStream2
2024-11-28T09:36:48.1945081Z 63: E20241128 09:36:48.194010 26100 rdb_load.cc:2798] OOM failed to add key 'mystream' in DB 0
2024-11-28T09:36:48.2026251Z 63: unknown file: Failure
2024-11-28T09:36:48.2027292Z 63: C++ exception with description "Unexpected index" thrown in the test body.
2024-11-28T09:36:48.2028230Z 63: 
2024-11-28T09:36:48.2136158Z 63: [  FAILED  ] RdbTest.LoadStream2 (46 ms)
```

kostasrim:
https://github.com/dragonflydb/dragonfly/actions/runs/12117287310/job/33779419970?pr=4073

BorysTheDev:
Try to reproduce https://github.com/dragonflydb/dragonfly/actions/runs/12141120779

BorysTheDev:
Can't reproduce even with previous tests order

BorysTheDev:
it looks like the problem with the RdbTest.SnapshotTooBig because now I see OOM for another test
https://github.com/dragonflydb/dragonfly/actions/runs/12182992643/job/33983338061?pr=4267

BorysTheDev:
I haven't seen this issue for a long time so I close it for now

PR #4242, opened 2024-12-02
fix: RdbTest.LoadStream3 incorrect file usage

BorysTheDev:
problem: RdbTest.LoadStream3 used incorrect file
fix: change file name to correct

Also, I've moved RdbTest.SnapshotTooBig to the end of the file to check the reproducibility of #4215, because it tests OOM error and I don't have other ideas for now

### DIFF
--- a/src/server/rdb_test.cc
+++ b/src/server/rdb_test.cc
@@ -646,20 +646,11 @@ TEST_F(RdbTest, LoadHugeStream) {
   ASSERT_EQ(2000, CheckedInt({"xlen", "test:0"}));
 }
 
-TEST_F(RdbTest, SnapshotTooBig) {
-  // Run({"debug", "populate", "10000", "foo", "1000"});
-  //  usleep(5000);  // let the stats to sync
-  max_memory_limit = 100000;
-  used_mem_current = 1000000;
-  auto resp = Run({"debug", "reload"});
-  ASSERT_THAT(resp, ErrArg("Out of memory"));
-}
-
 TEST_F(RdbTest, LoadStream2) {
   auto ec = LoadRdb("RDB_TYPE_STREAM_LISTPACKS_2.rdb");
   ASSERT_FALSE(ec) << ec.message();
   auto res = Run({"XINFO", "STREAM", "mystream"});
-  EXPECT_THAT(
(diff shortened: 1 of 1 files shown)


src/server/rdb_test.cc
```diff
@@ -668,15 +659,24 @@ TEST_F(RdbTest, LoadStream2) {
 }
 
 TEST_F(RdbTest, LoadStream3) {
-  auto ec = LoadRdb("RDB_TYPE_STREAM_LISTPACKS_2.rdb");
+  auto ec = LoadRdb("RDB_TYPE_STREAM_LISTPACKS_3.rdb");
```
kostasrim, line 662: Nice catch!

src/server/rdb_test.cc
```diff
@@ -646,20 +646,11 @@ TEST_F(RdbTest, LoadHugeStream) {
...
-  auto resp = Run({"debug", "reload"});
-  ASSERT_THAT(resp, ErrArg("Out of memory"));
-}
-
 TEST_F(RdbTest, LoadStream2) {
   auto ec = LoadRdb("RDB_TYPE_STREAM_LISTPACKS_2.rdb");
   ASSERT_FALSE(ec) << ec.message();
   auto res = Run({"XINFO", "STREAM", "mystream"});
-  EXPECT_THAT(
+  ASSERT_THAT(
```
kostasrim, line 653: I am still not sure why this fails and I don't see how moving `SnapshotTooBig` below will have any effect or it will somehow stop reproducing since on each test case we `initialize everything from the start` so i don't see how changing `maxmemory` of one test case can affect another :shrug: 

Out of curiosity did you try running the test in a loop in a similar machine as the one it failed ?

src/server/rdb_test.cc
BorysTheDev: It was my bug, It looks I haven't saved it before commit when I wrote it

src/server/rdb_test.cc
BorysTheDev: Good idea. But let's merge this fix and I will run the test in a loop.

kostasrim (approved):

PR #4248, opened 2024-12-03
chore: Hide managed service info in `INFO`

### LINKED ISSUE #4173: Managed mode operation

chakaz:
Specifically:
* `INFO REPLICATION` does not list the replicas, but does still show `connected_slaves`
* `INFO SERVER` does not show `thread_count` and `os`

Fixes #4173

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -127,6 +127,9 @@ ABSL_FLAG(bool, s3_sign_payload, true,
 ABSL_FLAG(bool, info_replication_valkey_compatible, true,
           "when true - output valkey compatible values for info-replication");
 
+ABSL_FLAG(bool, managed_service_info, false,
+          "Hides some implementation details from users when true (i.e. in managed service env)");
+
 ABSL_DECLARE_FLAG(int32_t, port);
 ABSL_DECLARE_FLAG(bool, cache_mode);
 ABSL_DECLARE_FLAG(uint32_t, hz);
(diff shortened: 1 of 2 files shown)


src/server/server_family.cc
```diff
@@ -2222,10 +2225,12 @@ void ServerFamily::Info(CmdArgList args, const CommandContext& cmd_cntx) {
     append("dragonfly_version", GetVersion());
     append("redis_mode", GetRedisMode());
     append("arch_bits", 64);
-    append("os", GetOSString());
+    if (!absl::GetFlag(FLAGS_managed_service_info)) {
```
kostasrim, line 2228: We could cache the result of GetFlag :)

kostasrim (approved):

src/server/server_family.cc
chakaz: We shouldn't optimize for `INFO` command in that level, let alone this value can change via `CONFIG SET`.

kostasrim: If it can change by `CONFIG SET` then an `INFO ALL` might contain `some` of the changes depending on the `GetFlag` at that point in time. In other words this is a race (not a data race but a race on what the thread will see when it polls GetFlag twice)

chakaz: We can surely live our lives with this race :rofl:

romange (approved):

PR #4255, opened 2024-12-04
chore(rax_tree): Introduce raxFreeWithCallback call in RaxTreeMap destructor

BagritsevichStepan:
related to [this comment](https://github.com/dragonflydb/dragonfly/pull/4228#issuecomment-2514645979)

### DIFF
--- a/src/redis/rax.c
+++ b/src/redis/rax.c
@@ -1221,29 +1221,44 @@ int raxRemove(rax *rax, unsigned char *s, size_t len, void **old) {
 
 /* This is the core of raxFree(): performs a depth-first scan of the
  * tree and releases all the nodes found. */
-void raxRecursiveFree(rax *rax, raxNode *n, void (*free_callback)(void*)) {
+void raxRecursiveFree(rax *rax, raxNode *n, void (*free_callback)(void*, void*), void* argument) {
     debugnode("free traversing",n);
     int numchildren = n->iscompr ? 1 : n->size;
     raxNode **cp = raxNodeLastChildPtr(n);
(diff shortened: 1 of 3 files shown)


BagritsevichStepan:
https://github.com/dragonflydb/dragonfly/actions/runs/12233835593

src/redis/rax.c
```diff
@@ -1221,29 +1221,44 @@ int raxRemove(rax *rax, unsigned char *s, size_t len, void **old) {
...
     if (free_callback && n->iskey && !n->isnull)
-        free_callback(raxGetData(n));
+        free_callback(raxGetData(n),argument);
     rax_free(n);
     rax->numnodes--;
 }
 
+/* Free the entire radix tree, invoking a free_callback function for each key's data. 
+ * An additional argument is passed to the free_callback function.*/
+ void raxFreeWithCallbackAndArgument(rax *rax, void (*free_callback)(void*, void*), void* argument) {
```
chakaz, line 1243: Should we call this:

```suggestion
 void raxRecursiveFreeWithCallbackAndArgument(rax *rax, void (*free_callback)(void*, void*), void* argument) {
```

?

BagritsevichStepan: I don't think so, because Redis provides `raxFreeWithCallback` as a public API. And `raxRecursiveFree` is an internal method and not part of the public API. So for consistency, I think `raxFreeWithCallbackAndArgument` would be a better choice.

chakaz: ok :shrug:

chakaz (approved):

PR #4258, opened 2024-12-04
chore: update command interface for cluster and search families

romange:

### DIFF
--- a/src/server/cluster/cluster_family.cc
+++ b/src/server/cluster/cluster_family.cc
@@ -124,6 +124,17 @@ void ClusterFamily::Shutdown() {
   });
 }
 
+std::optional<ClusterShardInfos> ClusterFamily::GetShardInfos(ConnectionContext* cntx) const {
+  if (IsClusterEmulated()) {
+    return {GetEmulatedShardInfo(cntx)};
+  }
+
+  if (tl_cluster_config != nullptr) {
+    return GetConfigForStats(cntx);
+  }
+  return nullopt;
+}
+
 ClusterShardInfo ClusterFamily::GetEmulatedShardInfo(ConnectionContext* cntx) const {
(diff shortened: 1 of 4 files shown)


src/server/cluster/cluster_family.cc
```diff
@@ -227,13 +238,11 @@ void ClusterShardsImpl(const ClusterShardInfos& config, SinkReplyBuilder* builde
...
 
 void ClusterFamily::ClusterShards(SinkReplyBuilder* builder, ConnectionContext* cntx) {
-  if (IsClusterEmulated()) {
-    return ClusterShardsImpl({GetEmulatedShardInfo(cntx)}, builder);
-  } else if (tl_cluster_config != nullptr) {
-    return ClusterShardsImpl(GetConfigForStats(cntx), builder);
-  } else {
-    return builder->SendError(kClusterNotConfigured);
+  auto shard_infos = GetShardInfos(cntx);
+  if (shard_infos) {
```
chakaz, line 242: Super nit, but I think it'd be slightly more elegant to first return on errors (i.e. `if (!shard_infos) { SendError(); }`)

src/server/cluster/cluster_family.cc
```diff
@@ -124,6 +124,17 @@ void ClusterFamily::Shutdown() {
   });
 }
 
+std::optional<ClusterShardInfos> ClusterFamily::GetShardInfos(ConnectionContext* cntx) const {
```
chakaz, line 127: Nice! :)

chakaz (approved):

src/server/cluster/cluster_family.cc
romange: i agree, i just automatically followed the previous flow.

PR #4264, opened 2024-12-05
fix(geo): support case insensitive units

andydunstall:
Most `GEO` commands don't seem to support case insensitive units, so seemed simplest to move `AsciiStrToUpper` into `ExtractUnit`

Such as `GEOSEARCH Sicily FROMLONLAT 15 37 BYRADIUS 200 km ASC` (copied from redis docs) would return an error `(error) ERR unsupported unit provided. please use M, KM, FT, MI`


kostasrim (approved):
Nice!

romange:
Is it based on an issue?

kostasrim:
> Is it based on an issue?
> […](#)


no no, Andy found it and he raised a PR

PR #4265, opened 2024-12-05
chore: update command interface for main_service commands

romange:
Clean up command_registry interface as well.

### DIFF
--- a/src/server/command_registry.cc
+++ b/src/server/command_registry.cc
@@ -96,10 +96,9 @@ bool CommandId::IsMultiTransactional() const {
   return CO::IsTransKind(name()) || CO::IsEvalKind(name());
 }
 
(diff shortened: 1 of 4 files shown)


kostasrim (approved):

src/server/rdb_test.cc
```diff
@@ -54,6 +54,7 @@ class RdbTest : public BaseFamilyTest {
 
 void RdbTest::SetUp() {
   InitWithDbFilename();
+  used_mem_current = 0;
```
kostasrim, line 57: This is because of the rdb failures right ?

chakaz: this indeed seems unrelated to this PR :)

chakaz (approved):

chakaz (approved):

Issue #4269, opened 2024-12-06
hard crash on test_migration_timeout_on_sync


kostasrim:
https://github.com/dragonflydb/dragonfly/actions/runs/12193906058/job/34016892594#step:6:935

```
30001➜ @           0xc29b74  (unknown)  absl::lts_20240116::inlined_vector_internal::Storage<>::GetSize()
30001➜    @          0x13ea90a         64  absl::lts_20240116::WriteFailureInfo()
30001➜    @          0x13eab7c        112  absl::lts_20240116::AbslFailureSignalHandler()
30001➜    @     0x7f0a74916420       1504  (unknown)
30001➜    @           0xc28f6a         32  absl::lts_20240116::InlinedVector<>::size()
30001➜    @           0xc2cb6e         32  absl::lts_20240116::InlinedVector<>::empty()
30001➜    @           0xc2c47c         32  dfly::PendingBuf::Empty()::{lambda()#1}::operator()<>()
30001➜    @           0xc2c4ae         48  __gnu_cxx::__ops::_Iter_negate<>::operator()<>()
30001➜    @           0xc2ba19        112  std::__find_if<>()
30001➜    @           0xc2ab38        128  std::__find_if_not<>()
30001➜    @           0xc29a72        128  std::find_if_not<>()
30001➜    @           0xc28e96        160  std::all_of<>()
30001➜    @           0xc28848        112  dfly::PendingBuf::Empty()
30001➜    @           0xc248e7        256  dfly::JournalStreamer::OnCompletion()
30001➜    @           0xc23fc2         64  dfly::JournalStreamer::AsyncWrite()::{lambda()#1}::operator()()
30001➜    @           0xc27381         48  std::_Function_handler<>::_M_invoke()
30001➜    @          0x132e669         64  std::function<>::operator()()
30001➜    @          0x132cc52        224  io::(anonymous namespace)::AsyncWriteState::OnCb()
30001➜    @          0x132c9f1         80  io::(anonymous namespace)::AsyncWriteState::OnCb()::{lambda()#1}::operator()()
30001➜    @          0x132de48         80  std::_Function_handler<>::_M_invoke()
30001➜    @          0x1270a3b         48  std::function<>::operator()()
```

kostasrim:
related to https://github.com/dragonflydb/dragonfly/pull/4225/files

PR #4270, opened 2024-12-06
fix: close socket to prevent onCompletion call after the journal stre…

BorysTheDev:
fixes: #4269

problem: OnCompletion is called after the JournalStreamer has been removed

### DIFF
--- a/src/server/cluster/outgoing_slot_migration.cc
+++ b/src/server/cluster/outgoing_slot_migration.cc
@@ -81,7 +81,8 @@ class OutgoingMigration::SliceSlotMigration : private ProtocolClient {
   }
 
   void Cancel() {
-    cntx_.Cancel();
+    // Close socket for clean disconnect.
+    CloseSocket();
     streamer_.Cancel();
   }
 


kostasrim (dismissed):

src/server/cluster/outgoing_slot_migration.cc
```diff
@@ -81,6 +81,10 @@ class OutgoingMigration::SliceSlotMigration : private ProtocolClient {
   }
 
   void Cancel() {
+    // Close socket for clean disconnect.
+    if (Sock()->IsOpen()) {
+      std::ignore = Sock()->Shutdown(SHUT_RDWR);
```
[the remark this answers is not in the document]
BorysTheDev: yes this stack trace shows that object was removed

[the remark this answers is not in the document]
BorysTheDev: JournalStreamer is removed before Socket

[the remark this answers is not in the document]
romange: we call `WaitForInflightToComplete` to wait for all the callbacks to return from `JournalStreamer::Cancel()`. 
If we have OnCompletion being called, it means WaitForInflightToComplete was not called, or something else is wrong.

src/server/cluster/outgoing_slot_migration.cc
```diff
@@ -81,7 +81,8 @@ class OutgoingMigration::SliceSlotMigration : private ProtocolClient {
   }
 
   void Cancel() {
-    cntx_.Cancel();
+    // Close socket for clean disconnect.
+    CloseSocket();
```
romange, line 85: I think `streamer_.Cancel();` should work as is.

BorysTheDev: I found that we do the same for replica. And streamer_.Cancel(); doesn't work without close socket because we need to cancel OnCompletion callback

src/server/cluster/outgoing_slot_migration.cc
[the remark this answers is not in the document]
BorysTheDev: WaitForInflightToComplete isn't called if the context was canceled.

src/server/cluster/outgoing_slot_migration.cc
adiholden: Any idea why we are not WaitForInflightToComplete inside JournalStreamer::Cancel if cntx is cancelled?

`void JournalStreamer::Cancel() {
    VLOG(1) << "JournalStreamer::Cancel";
    waker_.notifyAll();
    journal_->UnregisterOnChange(journal_cb_id_);
    if (!cntx_->IsCancelled()) {
      WaitForInflightToComplete();
    }
}`

adiholden (approved):

src/server/cluster/outgoing_slot_migration.cc
BorysTheDev: Because we can have different situations
1) we have an error and cancel by error
2) we cancel for our reason
3) everything works good
4) was canceled by an error in another fiber

These different scenarios have different behavior and  WaitForInflightToComplete can wait up to 2 minutes (connection timeout) without any result so if we cancel the connection by our reason there is no sense to wait so long period of time

BorysTheDev: From my point of view we should refactor JournalStreamer, but such refactoring affect replication and I don't think that we should do it right now

PR #4273, opened 2024-12-07
chore: let resp parser provide more useful logs

romange:
1. More warning logs around bad BAD_ARRAYLEN messages
2. Lift the restriction around big bulk strings and log a warning instead.
3. Pull helio


chakaz (dismissed):
LG, but I'd add a unit test to parse such big inputs..

chakaz (approved):

PR #4285, opened 2024-12-10
chore: add active time to stream consumers

romange:
Adjust XINFO command to output active-time property. Store active-time and switch to RDB_TYPE_STREAM_LISTPACKS_3 if FLAGS_stream_rdb_encode_v2 is enabled.

### DIFF
--- a/src/server/rdb_save.cc
+++ b/src/server/rdb_save.cc
@@ -215,8 +215,8 @@ uint8_t RdbObjectType(const PrimeValue& pv) {
       }
       break;
     case OBJ_STREAM:
-      return absl::GetFlag(FLAGS_stream_rdb_encode_v2) ? RDB_TYPE_STREAM_LISTPACKS
-                                                       : RDB_TYPE_STREAM_LISTPACKS_2;
+      return absl::GetFlag(FLAGS_stream_rdb_encode_v2) ? RDB_TYPE_STREAM_LISTPACKS_3
+                                                       : RDB_TYPE_STREAM_LISTPACKS;
     case OBJ_MODULE:
       return RDB_TYPE_MODULE_2;
     case OBJ_JSON:
@@ -713,8 +713,7 @@ error_code RdbSerializer::SaveStreamObject(const PrimeValue& pv) {
       RETURN_ON_ERR(SaveStreamPEL(cg->pel, true));
 
       /* Save the consumers of this group. */
-
-      RETURN_ON_ERR(SaveStreamConsumers(cg));
+      RETURN_ON_ERR(SaveStreamConsumers(rdb_type >= RDB_TYPE_STREAM_LISTPACKS_3, cg));
     }
   }
 
@@ -845,7 +844,7 @@ error_code RdbSerializer::SaveStreamPEL(rax* pel, bool nacks) {
   return error_code{};
 }
 
-error_code RdbSerializer::SaveStreamConsumers(streamCG* cg) {
+error_code RdbSerializer::SaveStreamConsumers(bool save_active, streamCG* cg) {
   /* Number of consumers in this consumer group. */
 
   RETURN_ON_ERR(SaveLen(raxSize(cg->consumers)));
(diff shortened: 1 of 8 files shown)


src/server/rdb_save.cc
```diff
@@ -215,8 +215,8 @@ uint8_t RdbObjectType(const PrimeValue& pv) {
       }
       break;
     case OBJ_STREAM:
-      return absl::GetFlag(FLAGS_stream_rdb_encode_v2) ? RDB_TYPE_STREAM_LISTPACKS
-                                                       : RDB_TYPE_STREAM_LISTPACKS_2;
+      return absl::GetFlag(FLAGS_stream_rdb_encode_v2) ? RDB_TYPE_STREAM_LISTPACKS_3
```
BorysTheDev, line 218: looks strange to me

src/server/stream_family.cc
```diff
@@ -82,7 +82,8 @@ struct NACKInfo {
 
 struct ConsumerInfo {
   string name;
-  size_t seen_time;
+  uint64_t seen_time;
+  int64_t active_time;
```
BorysTheDev, line 86: can we have the same type for time?

src/server/stream_family.cc
```diff
@@ -1254,6 +1257,14 @@ OpResult<ClaimInfo> OpClaim(const OpArgs& op_args, string_view key, const ClaimO
 
   StreamMemTracker tracker;
 
+  // Try to get the consumer. If not found, create a new one.
+  auto cname = WrapSds(opts.consumer);
+  streamConsumer* consumer = streamLookupConsumer(cgr_res->cg, cname);
+  if (consumer == nullptr)
+    consumer = StreamCreateConsumer(cgr_res->cg, opts.consumer, now_ms, SCC_DEFAULT);
+  else
+    consumer->seen_time = now_ms;
```
BorysTheDev, line 1266: I want to suggest making a separate function UpdateOrCreateConsumer

src/server/rdb_save.cc
romange: I do not understand the comment

src/server/stream_family.cc
romange: sure

src/server/stream_family.cc
```diff
@@ -3099,26 +3113,17 @@ variant<bool, facade::ErrorReply> HasEntries2(const OpArgs& op_args, string_view
...
-      consumer = StreamCreateConsumer(group, opts->consumer_name, op_args.db_cntx.time_now_ms,
-                                      SCC_NO_NOTIFY | SCC_NO_DIRTIFY);
-    }
+    sds cname = WrapSds(opts->consumer_name);
+    consumer = streamLookupConsumer(group, cname);
+    uint64_t now_ms = op_args.db_cntx.time_now_ms;
+    if (consumer)
+      consumer->seen_time = now_ms;
+    else
+      consumer = StreamCreateConsumer(group, opts->consumer_name, now_ms, SCC_DEFAULT);
```
BorysTheDev, line 3122: if you create a function UpdateOrCreateConsumer() you can use it here

src/server/stream_family.cc
```diff
@@ -3130,6 +3135,14 @@ variant<bool, facade::ErrorReply> HasEntries2(const OpArgs& op_args, string_view
       requested_sitem.id.val = requested_sitem.group->last_id;
       streamIncrID(&requested_sitem.id.val);
     }
+  } else {
+    // Resolve $ to the last ID in the stream.
+    if (requested_sitem.id.resolve_last_id) {
+      requested_sitem.id.val = last_id;
+      streamIncrID(&requested_sitem.id.val);  // include id's strictly greater
```
BorysTheDev, line 3142: should we check the result of streamIncrID().

BorysTheDev (approved):

src/server/rdb_save.cc
BorysTheDev: You write it in the description that we will use RDB_TYPE_STREAM_LISTPACKS_3 for FLAGS_stream_rdb_encode_v2,
but it looks strange to me

src/server/stream_family.cc
romange: I backported the changes from valkey. They do not check it.

BorysTheDev: we check it sometimes and sometimes no. We can rewrite this method in this case and never check for example

BorysTheDev (approved):

PR #4290, opened 2024-12-11
refactor(search_family): Add Aggregator class

BagritsevichStepan:
related to this [comment](https://github.com/dragonflydb/dragonfly/pull/4231#discussion_r1864173422)

1. Add `Aggregator` class.
2. Remove redundant `std::string` copies entirely during the aggregation steps.

TODO:
1. Add logs
2. Remove the usage of `std::function`

### DIFF
--- a/src/server/search/aggregator.cc
+++ b/src/server/search/aggregator.cc
@@ -77,10 +77,7 @@ void Aggregator::DoSort(std::string_view field, bool descending) {
     if (l_it->second < r_it->second) {
       return !descending;
     }
-    if (l_it->second > r_it->second) {
-      return descending;
-    }
-    return true;  // Elements are equal
+    return descending;  // l_it->second >= r_it->second
   };
 
   std::sort(result.values.begin(), result.values.end(), std::move(comparator));
--- a/src/server/search/aggregator.cc
+++ b/src/server/search/aggregator.cc
@@ -66,18 +66,31 @@ void Aggregator::DoGroup(absl::Span<const std::string> fields, absl::Span<const 
 }
 
 void Aggregator::DoSort(std::string_view field, bool descending) {
+  /*
+    Comparator for sorting DocValues by field.
+    If some of the fields is not present in the DocValues, comparator returns:
+    1. l_it == l.end() && r_it != r.end()
+      asc -> false
+      desc -> false
+    2. l_it != l.end() && r_it == r.end()
+      asc -> true
+      desc -> true
+    3. l_it == l.end() && r_it == r.end()
+      asc -> false
+      desc -> false
+  */
   auto comparator = [&](const DocValues& l, const DocValues& r) {
     auto l_it = l.find(field);
     auto r_it = r.find(field);
 
-    // Handle cases where one of the fields is missing
+    // If some of the values is not present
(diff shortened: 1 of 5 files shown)


src/server/search/aggregator.cc
```diff
@@ -10,51 +10,81 @@ namespace dfly::aggregate {
...
+            [field](const DocValues& l, const DocValues& r) {
+              auto it1 = l.find(field);
+              auto it2 = r.find(field);
+              return it1 == l.end() || (it2 != r.end() && it1->second < it2->second);
+            });
 
-const Value kEmptyValue = Value{};
+  if (descending) {
+    std::reverse(result.values.begin(), result.values.end());
+  }
```
BorysTheDev, line 78: move it into comparator logic and do the correct order during the sort

BorysTheDev (dismissed):

src/server/search/aggregator.cc
```diff
@@ -10,62 +10,89 @@ namespace dfly::aggregate {
...
+      return l_it != l.end() || r_it == r.end();
+    }
+    if (l_it->second < r_it->second) {
+      return !descending;
+    }
+    if (l_it->second > r_it->second) {
+      return descending;
+    }
+    return true;  // Elements are equal
+  };
```
BorysTheDev, line 84: why do we return true if elements are equal for less operator? As I understand you could take previous implementation and return something like "res != descending"

BagritsevichStepan: Yes, it makes sense to return `descending` here. But just to note, we are not guaranteeing any strict ordering in the result when elements are equal. This is because the final order also depends on the initial distribution of keys across shards (results from shards are joined into one vector, and then a sort is performed on this array).

So, it is not a "stable" sort, meaning that the initial order of the keys is influenced by the key distribution across shards, which may give the user the impression of an unstable sort

src/server/search/aggregator.cc
```diff
@@ -10,62 +10,86 @@ namespace dfly::aggregate {
...
+void Aggregator::DoSort(std::string_view field, bool descending) {
+  auto comparator = [&](const DocValues& l, const DocValues& r) {
+    auto l_it = l.find(field);
+    auto r_it = r.find(field);
 
-const Value kEmptyValue = Value{};
+    // Handle cases where one of the fields is missing
+    if (l_it == l.end() || r_it == r.end()) {
+      return l_it != l.end() || r_it == r.end();
+    }
```
BorysTheDev, line 76: why descending doesn't affect this part of the code

BagritsevichStepan: This what redis does:
```
1. l_it == l.end() && r_it != r.end()
   asc -> false
   desc -> false
2. l_it !=  l.end() && r_it == r.end()
  asc -> true
  desc -> true
3. l_it ==  l.end() && r_it == r.end()
   asc -> false
  desc -> false
```

BagritsevichStepan: So descending doesn't affect this part. But I need to fix it to `l_it != l.end() && r_it == r.end()`

BagritsevichStepan:
I simplified the comparator logic, so I don't think it’s necessary to use `std::less `here as we previously discussed.

PR #4307, opened 2024-12-13
fix: Protect `BumpUp()` from running in parallel to serialization

### LINKED ISSUE #4306: Missed replication items in `cache_mode` full sync
We've seen 2 regtests failures:

https://github.com/dragonflydb/dragonfly/actions/runs/12229971008/job/34110555900
https://github.com/dragonflydb/dragonfly/actions/runs/12224081214/job/34096392568

This (very rarely) happens when `cache_mode` is enabled due to unprotected `BumpUp()`

chakaz:
Fixes #4306

### DIFF
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -468,6 +468,7 @@ OpResult<DbSlice::PrimeItAndExp> DbSlice::FindInternal(const Context& cntx, std:
       db.prime.CVCUponBump(change_cb_.back().first, res.it, bump_cb);
     }
 
+    block_counter_.Wait();  // We must not change the bucket's internal order during serialization
     auto bump_it = db.prime.BumpUp(res.it, PrimeBumpPolicy{&fetched_items_});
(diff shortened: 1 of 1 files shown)


src/server/db_slice.cc
```diff
@@ -468,6 +468,7 @@ OpResult<DbSlice::PrimeItAndExp> DbSlice::FindInternal(const Context& cntx, std:
       db.prime.CVCUponBump(change_cb_.back().first, res.it, bump_cb);
     }
 
+    block_counter_.Wait();  // We must not change the bucket's internal order during serialization
```
romange, line 471: not related to this PR but `block_counter_` is not a good name because it does not say much about the semantics of the variable.

kostasrim, line 471: I would have to moved it to line 463 such that the whole block (including the capture version change) is atomic but it should be fine this way as well

kostasrim (approved):

Issue #4289, opened 2024-12-11
FAILED 😰  dragonfly/cluster_test.py::test_cluster_fuzzymigration


BorysTheDev:
https://github.com/dragonflydb/dragonfly/actions/runs/12273058221/job/34243082123

PR #4309, opened 2024-12-13
test: update logs and test for debug purpose

BorysTheDev:
added logs to investigate https://github.com/dragonflydb/dragonfly/issues/4289


kostasrim (commented):
There is no point to merge this as the test fails on x86 and it should be easy to reproduce locally or here by running the test in the loop

BorysTheDev:
> There is no point to merge this as the test fails on x86 and it should be easy to reproduce locally or here by running the test in the loop

But I can't

chakaz (approved):

PR #4314, opened 2024-12-15
fix(search_family): Support boolean and nullable types in indexes

BagritsevichStepan:
fixes dragonflydb#4107, dragonflydb#4129

### DIFF
--- a/src/core/search/sort_indices.cc
+++ b/src/core/search/sort_indices.cc
@@ -19,10 +19,37 @@ using namespace std;
 namespace {}  // namespace
 
 template <typename T>
+SimpleValueSortIndex<T>::ParsedSortValue::ParsedSortValue() : no_value_was_found_{true} {
+}
+
+template <typename T> SimpleValueSortIndex<T>::ParsedSortValue::ParsedSortValue(std::nullopt_t) {
+}
+
+template <typename T>
+SimpleValueSortIndex<T>::ParsedSortValue::ParsedSortValue(T value) : value_(std::move(value)) {
+}
+
+template <typename T> bool SimpleValueSortIndex<T>::ParsedSortValue::HasValue() const {
+  return !no_value_was_found_;
+}
+
+template <typename T> bool SimpleValueSortIndex<T>::ParsedSortValue::IsNullValue() const {
+  return HasValue() && !value_.has_value();
+}
+
+template <typename T> T&& SimpleValueSortIndex<T>::ParsedSortValue::Value() && {
+  return std::move(value_).value();
+}
+
+template <typename T>
 SimpleValueSortIndex<T>::SimpleValueSortIndex(PMR_NS::memory_resource* mr) : values_{mr} {
 }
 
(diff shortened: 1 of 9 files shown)


src/core/search/sort_indices.cc
```diff
@@ -48,21 +75,31 @@ std::vector<ResultScore> SimpleValueSortIndex<T>::Sort(std::vector<DocId>* ids,
...
-  values_[id] = field_value.value();
+  values_[id] = std::move(field_value).Value();
   return true;
 }
 
 template <typename T>
 void SimpleValueSortIndex<T>::Remove(DocId id, const DocumentAccessor& doc,
                                      std::string_view field) {
+  auto it = null_values_.find(id);
+  if (it != null_values_.end()) {
```
BorysTheDev, line 98: nit:
if (auto it = null_values_.find(id); it != null_values_.end()) 

it reduces variable scope

src/core/search/sort_indices.h
```diff
@@ -18,7 +18,24 @@
...
+    explicit ParsedSortValue(std::nullopt_t);
+    explicit ParsedSortValue(T value);
+
+    bool HasValue() const;
+    bool IsNullValue() const;
+    T&& Value() &&;
+
+   private:
+    bool no_value_was_found_ = false;
+    std::optional<T> value_;
```
BorysTheDev, line 35: nit:I would prefer to see std::variant of T and other 2 options

src/core/search/sort_indices.h
```diff
@@ -18,7 +18,24 @@
...
+ protected:
+  class ParsedSortValue {
+   public:
+    ParsedSortValue();
+    explicit ParsedSortValue(std::nullopt_t);
+    explicit ParsedSortValue(T value);
+
+    bool HasValue() const;
+    bool IsNullValue() const;
+    T&& Value() &&;
```
BorysTheDev, line 31: with variant your interface can be much smaller

src/server/search/doc_accessors.cc
```diff
@@ -194,17 +219,32 @@ struct JsonAccessor::JsonPathContainer {
...
   if (!path)
     return search::EmptyAccessResult<StringList>();
 
   auto path_res = path->Evaluate(json_);
   if (path_res.empty())
     return search::EmptyAccessResult<StringList>();
 
+  auto is_convertible_to_string = [&accept_boolean_values](const JsonType& json) -> bool {
+    return json.is_string() || (accept_boolean_values && json.is_bool());
+  };
```
BorysTheDev, line 242: depends on accept_boolean_values use different lambdas

```
[&accept_boolean_values]()  {
    if (accept_boolean_values) {
      return [](const JsonType& json) {
        return json.is_string() || json.is_bool();
      };
    } else {
      return [](const JsonType& json) {
        return json.is_string();
      };
  }();
```

BorysTheDev (approved):

BorysTheDev (approved):

PR #4315, opened 2024-12-16
chore: factor out rdb_load utilities into separate files

romange:
rdb_load.cc is huge and contains many auxillary classes. This PR moves DecompressImpl and ErrorRdb code into detail/

It also fixes minor bugs around error conditions with de-compression: a. Do not check-fail on invalid opcode and return error_code instead. b. Print correctly LZ4 errors.

### DIFF
--- a/src/server/detail/decompress.h
+++ b/src/server/detail/decompress.h
@@ -1,0 +1,30 @@
+// Copyright 2024, DragonflyDB authors.  All rights reserved.
+// See LICENSE for licensing terms.
+//
+#pragma once
+
+#include "io/io.h"
+#include "io/io_buf.h"
+
+namespace dfly {
+
+namespace detail {
+
+class DecompressImpl {
+ public:
+  DecompressImpl() : uncompressed_mem_buf_{1U << 14} {
+  }
+  virtual ~DecompressImpl() {
+  }
+
+  virtual io::Result<io::IoBuf*> Decompress(std::string_view str) = 0;
+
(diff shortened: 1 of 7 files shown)


src/server/detail/decompress.h
```diff
@@ -0,0 +1,30 @@
...
+// See LICENSE for licensing terms.
+//
+#pragma once
+
+#include "io/io.h"
+#include "io/io_buf.h"
+
+namespace dfly {
+
+namespace detail {
```
chakaz, line 11: I'd even move this to `core/`

src/server/detail/decompress.h
```diff
@@ -0,0 +1,30 @@
...
+ public:
+  DecompressImpl() : uncompressed_mem_buf_{1U << 14} {
+  }
+  virtual ~DecompressImpl() {
+  }
+
+  virtual io::Result<io::IoBuf*> Decompress(std::string_view str) = 0;
+
+  static DecompressImpl* CreateLZ4();
+  static DecompressImpl* CreateZstd();
```
chakaz, line 23: Why not return `unique_ptr<DecompressImpl>`?

chakaz: Also, I'd move these to the beginning of the `public` section

chakaz (commented):
Nice improvement! Please see some nits

src/server/detail/decompress.h
romange: we could, but I think at this time it does not really matter.

PR #4328, opened 2024-12-17
fix: crash during getting info about replication

BorysTheDev:
problem: we try to get the memory info during replication start, but the snapshot object doesn't exist yet

### DIFF
--- a/src/server/rdb_save.cc
+++ b/src/server/rdb_save.cc
@@ -1333,6 +1333,10 @@ size_t RdbSaver::Impl::GetTotalBuffersSize() const {
 
   auto cb = [this, &channel_bytes, &serializer_bytes](ShardId sid) {
     auto& snapshot = shard_snapshots_[sid];
+    // before create a snapshot we save header so shard_snapshots_ are vector of nullptr until we
+    // start snapshots saving
+    if (!snapshot)
+      return;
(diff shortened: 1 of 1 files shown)


adiholden (approved):

src/server/rdb_save.cc
```diff
@@ -1333,6 +1333,8 @@ size_t RdbSaver::Impl::GetTotalBuffersSize() const {
 
   auto cb = [this, &channel_bytes, &serializer_bytes](ShardId sid) {
     auto& snapshot = shard_snapshots_[sid];
+    if (!snapshot)
```
romange, line 1336: @BorysTheDev  please add a comment explaining how come RdbSave::Impl with shard_snapshots_ exist but they are not initialized yed.

src/server/rdb_save.cc
```diff
@@ -1355,6 +1357,8 @@ RdbSaver::SnapshotStats RdbSaver::Impl::GetCurrentSnapshotProgress() const {
 
   auto cb = [this, &results](ShardId sid) {
     auto& snapshot = shard_snapshots_[sid];
+    if (!snapshot)
```
romange, line 1360: same

romange (approved):

PR #4330, opened 2024-12-17
chore: add ability to track connections stuck at send

romange:
Add send_delay_seconds/send_delay_ms metrics.

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -638,6 +638,15 @@ optional<ReplicaOfArgs> ReplicaOfArgs::FromCmdArgs(CmdArgList args, SinkReplyBui
   return replicaof_args;
 }
 
+uint64_t GetDelayMs(uint64_t ts) {
+  uint64_t now_ns = fb2::ProactorBase::GetMonotonicTimeNs();
+  uint64_t delay_ns = 0;
+  if (ts < now_ns - 1000000) {  // handle states where ts is larger than now
+    delay_ns = (now_ns - ts) / 1000000;
+  }
+  return delay_ns;
+}
+
 }  // namespace
 
 void SlowLogGet(dfly::CmdArgList args, std::string_view sub_cmd, util::ProactorPool* pp,
@@ -1294,6 +1303,10 @@ void PrintPrometheusMetrics(uint64_t uptime, const Metrics& m, DflyCmd* dfly_cmd
                             MetricType::GAUGE, &resp->body());
   AppendMetricWithoutLabels("pipeline_queue_length", "", conn_stats.dispatch_queue_entries,
                             MetricType::GAUGE, &resp->body());
+  AppendMetricWithoutLabels("send_delay_seconds", "",
+                            double(GetDelayMs(m.oldest_pending_send_ts)) / 1000.0,
+                            MetricType::GAUGE, &resp->body());
+
   AppendMetricWithoutLabels("pipeline_throttle_total", "", conn_stats.pipeline_throttle_count,
                             MetricType::COUNTER, &resp->body());
   AppendMetricWithoutLabels("pipeline_cmd_cache_bytes", "", conn_stats.pipeline_cmd_cache_bytes,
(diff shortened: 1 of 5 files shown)


src/server/server_family.cc
```diff
@@ -638,6 +638,15 @@ optional<ReplicaOfArgs> ReplicaOfArgs::FromCmdArgs(CmdArgList args, SinkReplyBui
   return replicaof_args;
 }
 
+uint64_t GetDelayMs(uint64_t ts) {
+  uint64_t now_ns = fb2::ProactorBase::GetMonotonicTimeNs();
+  uint64_t delay_ns = 0;
+  if (ts < now_ns - 1000000) {  // handle states where ts is larger than now
+    delay_ns = (now_ns - ts) / 1000000;
+  }
```
chakaz, line 646: Why would `ts` ever be larger than `now_ns`, and even if it is, why would you compare that to now - 1s? and why do you set the result to be in units of seconds? and if now is smaller, wouldn't `now_ns - ts` be a negative (i.e. huge unsigned) number?

src/server/server_family.cc
```diff
@@ -2145,6 +2158,13 @@ Metrics ServerFamily::GetMetrics(Namespace* ns) const {
       result.connections_lib_name_ver_map[k] += v;
     }
 
+    auto& send_list = facade::SinkReplyBuilder::pending_list;
+    if (!send_list.empty()) {
+      auto& oldest_member = send_list.front();
+      if (oldest_member.timestamp_ns < result.oldest_pending_send_ts) {
+        result.oldest_pending_send_ts = oldest_member.timestamp_ns;
+      }
```
chakaz, line 2166:
```suggestion
      result.oldest_pending_send_ts = std::min(result.oldest_pending_send_ts, send_list.front());
```

src/server/server_family.cc
```diff
@@ -638,6 +638,15 @@ optional<ReplicaOfArgs> ReplicaOfArgs::FromCmdArgs(CmdArgList args, SinkReplyBui
   return replicaof_args;
 }
 
+uint64_t GetDelayMs(uint64_t ts) {
+  uint64_t now_ns = fb2::ProactorBase::GetMonotonicTimeNs();
+  uint64_t delay_ns = 0;
+  if (ts < now_ns - 1000000) {  // handle states where ts is larger than now
```
chakaz, line 644:
```suggestion
  if (ts < now_ns - 1000000) {  // if more than 1ms has passed between ts and now_ns
```

src/server/server_family.cc
chakaz: Consider also adding:

```cpp
DCHECK(std::is_sorted(send_list.begin(), send_list.end()));
```

chakaz (commented):
Generally LG, see a few nits

src/server/server_family.cc
romange: because oldest_pending_send_ts can be INT64_MAX (its default value). I compare with `now-1ms` just because it's my resolution scale - we do not really care about send being stuck for less than 1ms. now can not be smaller than 1ms assuming that linux monotonic clock did not reset.

src/server/server_family.cc
```diff
@@ -1294,6 +1303,10 @@ void PrintPrometheusMetrics(uint64_t uptime, const Metrics& m, DflyCmd* dfly_cmd
                             MetricType::GAUGE, &resp->body());
   AppendMetricWithoutLabels("pipeline_queue_length", "", conn_stats.dispatch_queue_entries,
                             MetricType::GAUGE, &resp->body());
+  AppendMetricWithoutLabels("send_delay_seconds", "",
```
romange, line 1306: i expose `send_delay_seconds` and not `ms` because prometheus spec requires exporting time metrics in seconds

src/server/server_family.cc
chakaz: Sorry, I meant to remove my comment and only keep the modified suggestions. I understood it when I made the suggestion :(

chakaz (approved):

PR #4338, opened 2024-12-18
fix: cluster tests stability

BorysTheDev:
make ports dynamic for cluster tests
close task for Proxy

### DIFF
--- a/tests/dragonfly/cluster_test.py
+++ b/tests/dragonfly/cluster_test.py
@@ -920,7 +920,7 @@ async def test_cluster_flush_slots_after_config_change(df_factory: DflyInstanceF
     assert await c_replica.execute_command("dbsize") == (100_000 - slot_0_size)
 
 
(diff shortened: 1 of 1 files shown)


tests/dragonfly/cluster_test.py
```diff
@@ -1433,7 +1433,6 @@ async def test_migration_with_key_ttl(df_factory):
     assert await nodes[1].client.execute_command("stick k_sticky") == 0
 
 
-@pytest.mark.skip("test is flaky")
```
chakaz, line 1436: Is this change intentional?
(I assume that changing the ports does not fix the test?)

adiholden (dismissed):

chakaz (commented):
While here, can you also change the hard-coded default port in line 924?

tests/dragonfly/cluster_test.py
BorysTheDev: Thx

chakaz (approved):

PR #4342, opened 2024-12-19
fix: Stack overflow in `DFLYCLUSTER CONFIG`

chakaz:
It's fine to use the heap in such cases, latency doesn't matter.


romange (approved):

romange:
PR title is misleading though. different command.

chakaz:
To verify this 100% I ran regression tests (500 times the flaky test), and it passed
https://github.com/dragonflydb/dragonfly/actions/runs/12409496909

Issue #4344, opened 2024-12-19
test_network_disconnect_during_migration is flaky


BorysTheDev:

BorysTheDev:
https://github.com/dragonflydb/dragonfly/actions/runs/12462175833/job/34782737591
https://github.com/dragonflydb/dragonfly/actions/runs/12466567493/job/34794343571

PR #4345, opened 2024-12-20
fix: test_network_disconnect_during_migration

BorysTheDev:
fixes: https://github.com/dragonflydb/dragonfly/issues/4344
small changes that close resources correctly and help debug the test if it fails again


chakaz (approved):
Approved, but does it really fix the issue (like the description suggests), or does it only clean up and add logs?

BorysTheDev:
@chakaz I can not reproduce the issue even on CI, so I hope if it still exists these changes help me

PR #4362, opened 2024-12-24
feat: add support for meta memcache commands

romange:
This is a stripped down version of supporting the memcache meta requests.

a. Not all meta flags are supported, but TTL, flags, arithmetics are supported. b. does not include reply support.
c. does not include new semantics that are not part of the older, ascii protocol.

The parser interface has not changed significantly, and the meta commands are emulated using the old, high level commands like ADD,REPLACE, INCR etc.

See https://raw.githubusercontent.com/memcached/memcached/refs/heads/master/doc/protocol.txt for more details regarding the meta commands spec.

### DIFF
--- a/src/facade/memcache_parser.cc
+++ b/src/facade/memcache_parser.cc
@@ -6,12 +6,14 @@
 #include <absl/container/flat_hash_map.h>
 #include <absl/container/inlined_vector.h>
 #include <absl/strings/ascii.h>
+#include <absl/strings/escaping.h>
 #include <absl/strings/numbers.h>
 #include <absl/strings/str_split.h>
 #include <absl/types/span.h>
 
 #include "base/logging.h"
 #include "base/stl_util.h"
+#include "facade/facade_types.h"
 
 namespace facade {
 using namespace std;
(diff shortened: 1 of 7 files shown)


src/facade/memcache_parser.cc
```diff
@@ -116,13 +139,159 @@ MP::Result ParseValueless(TokensView tokens, MP::Command* res) {
...
+
 }  // namespace
 
 auto MP::Parse(string_view str, uint32_t* consumed, Command* cmd) -> Result {
   cmd->no_reply = false;  // re-initialize
   auto pos = str.find("\r\n");
   *consumed = 0;
   if (pos == string_view::npos) {
+    // We need more data to parse the command. For get/gets commands this line can be very long.
+    // we limit maxmimum buffer capacity  in the higher levels using max_client_iobuf_len.
```
chakaz, line 294:
```suggestion
    // we limit maximum buffer capacity in the higher levels using max_client_iobuf_len.
```

chakaz (approved):
APPROVED!!! :partying_face:

Issue #3071, opened 2024-05-22
Connecting to Dragonfly via memcache client doesn't work


rocktavious:
**Describe the bug**
We are trying to use the memcached compliant API to connect a service via a Ruby client and we are getting errors.  This same client works with an AWS Elasticache in memcache mode.

The dragonfly configuration

```
kind: Dragonfly
metadata:
  labels:
    app.kubernetes.io/name: dragonfly-memcache
    app.kubernetes.io/instance: dragonfly-opslevel
    app.kubernetes.io/managed-by: kustomize
  name: memcache
  namespace: staging
spec:
  priorityClassName: critical-priority
  replicas: 1
  memcachedPort: 11211
  resources:
    requests:
      cpu: 500m
      memory: 2Gi
    limits:
      cpu: 600m
      memory: 3Gi
 ```
 
 The Ruby app in the same kubernetes cluster showing that we can resolve the DNS
 
 ```
 irb(main):001> require 'resolv'
=> false
irb(main):002> Resolv.getname "10.100.251.152"
=> "memcache.staging.svc.cluster.local"
irb(main):009> Resolv.getaddress "memcache.staging.svc.cluster.local"
=> "10.100.251.152"
```

Also from the shell inside the Ruby pod a sample showing we can connect

```
# telnet memcache.staging.svc.cluster.local 11211
Trying 10.100.251.152...
Connected to memcache.staging.svc.cluster.local.
Escape character is '^]'.
stats
STAT pid 1
STAT uptime 3465
STAT time 1716310286
STAT version v1.16.0
STAT libevent iouring
STAT pointer_size 8
STAT rusage_user 30.763
STAT rusage_system 19.0581
STAT max_connections -1
STAT curr_connections 1
STAT total_connections -1
STAT rejected_connections -1
STAT bytes_read 5284
STAT bytes_written 5046
STAT limit_maxbytes -1
END
```

But when we go to use the client library we get this

```
irb(main):013> require "dalli"
=> false
irb(main):014> dc = Dalli::Client.new("memcache.staging.svc.cluster.local:11211")
=> #<Dalli::Client:0x00007f1cb51ef108 @key_manager=#<Dalli::KeyManager:0x00007f1cc0335b88 @key_options={:digest_class=>Digest::MD5}, @namespace=nil>, @normalized_servers=["memcache.staging.svc.cluster.local:11211"], @options={}, @ring=nil>
irb(main):015> dc.set('abc', 123)
[...]

PR #4366, opened 2024-12-24
feat: support memcache meta responses

### LINKED ISSUE #4348: Potential incompatibility with ruby memcached client
Using dali client set commands times out

**To Reproduce**
install ruby and dali
sudo apt update
sudo apt install ruby-full
sudo gem install dalli

create test.rb with the following content 
```
require 'dalli'

begin
  # Create Memcached client with minimal configuration
  memcached = Dalli::Client.new('localhost:11211', {
    socket_timeout: 30,  # Increased timeout
    compress: false,     # Disable compression
    serializer: nil      # Disable serialization to match telnet behavior
  })

  puts "Testing connection..."

  # Try the same operation that worked in telnet
  result = memcached.set('my_key', 'hello', 900)
  puts "Set result: #{result}"

  value = memcached.get('my_key')
  puts "Retrieved value: #{value}"

rescue => e
  puts "Error: #{e.class} - #{e.message}"
  puts e.backtrace
end

```

Run dragonfly 
`docker run --pull always --network=host --ulimit memlock=-1 docker.dragonflydb.io/dragonflydb/dragonfly:latest --memcached_port 11211
`

Run ruby script 
`ruby test.rb`

**Expected behavior**
Expect script to print response but it times out instead

The following script succeeds

romange:
Fixes #4348 and #3071

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -421,8 +421,8 @@ struct Connection::AsyncOperations {
   }
 
   void operator()(const PubMessage& msg);
-  void operator()(Connection::PipelineMessage& msg);
-  void operator()(const Connection::MCPipelineMessage& msg);
+  void operator()(PipelineMessage& msg);
+  void operator()(const MCPipelineMessage& msg);
   void operator()(const MonitorMessage& msg);
   void operator()(const AclUpdateMessage& msg);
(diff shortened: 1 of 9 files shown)


chakaz (approved):

chakaz (approved):
We really should start using the term memcache**d** in our codebase :rofl:

PR #4380, opened 2024-12-28
Fix test hypothesis

cunla:
# test_connection

Pass

# test_server

- `FLUSHDB ASYNC` not supported (Assuming `SYNC` also not supported)
- `FLUSHALL ASYNC` not supported (Assuming `SYNC` also not supported)
  ==> Removing `ASYNC` parameter
  ==> PASS

# test_set

- `One or more scores can't be converted into double not raised`

# test_zset

- dragonfly raised: `One or more scores can't be converted into double` but fakeredis did not.

# test_hash

- `HEXPIRE` do not support `nx`/`xx`/`gt`/`lt`.
- `HPEXPIRE` not supported.
  ==> Removed from test.

# test_string

- Set str-length range to `min_value=-2147483647, max_value=2147483648` ==> Exception
  `string exceeds maximum allowed size` still occurs for `2_005_975_152`.

# test_list

# test_transaction

# test_joint


romange:
Thank @cunla . Let's:
1. TestServer - agree these are not important for Dragonfly (its flush operation is always asynchronous but the api call is waiting till the operation ends).
2. HEXPIRE - agree, we do not strive for Redis 8 compatibility at this point.
3.  TestString =- yeah, we are more restrictive. have not heard much complains about it.  See https://github.com/dragonflydb/dragonfly/blob/main/docs/differences.md#string-lengths-indices  for more details.
4. TestList - if I remember correctly, it happens during SORT and it's consistent with Redis/Valkey. I am looking at it again now.

romange:
@cunla  I am confirming again that the sequence:
```
lpush foo '\00'
sort foo
```
fails in sort for both Dragonfly and Redis. Seems that the problem in test?

romange:
@cunla  I fixed the crash. Please rebase 🙏🏼

romange (approved):

romange:
Thank you!

Issue #4387, opened 2024-12-30
crash in tiered code


romange:
configuration:
```yml
redis:
  image: 'docker.dragonflydb.io/dragonflydb/dragonfly'
  ulimits:
    memlock: -1
  volumes:
    - dragonflydata:/data
  command: [
    "--port", "6379",
    "--requirepass", "password",
    "--maxmemory", "8G",
    "--cache_mode=true",
    "--tiered_prefix=/data/tiered",
  ]
```

`2024-12-29 06:44:40 F20241229 05:44:40.774821    11 externalalloc.cc:372] Check failed: idx < segments.size() (1 vs. 1)`

also 

[message (14).txt](https://github.com/user-attachments/files/18276060/message.14.txt)
[message (13).txt](https://github.com/user-attachments/files/18276061/message.13.txt)

PR #4388, opened 2024-12-30
fix: ExternalAllocator::Free with large sizes

romange:
ExternalAllocator allocates large sizes directly from the extent tree bypassing segment data structure. Unfortunately, we forgot to align Free() the same way. This PR:

1. Make sure that we add back the allocated range to the extent tree in Free.
2. Rewrite and simplify ExtentTree::Add implementation.


BorysTheDev (approved):

romange:
Addresses one of the crashes in #4387

PR #4399, opened 2025-01-03
feat: allow querying of json objects stored as strings

romange:

### DIFF
--- a/src/server/json_family.cc
+++ b/src/server/json_family.cc
@@ -657,11 +657,29 @@ bool LegacyModeIsEnabled(const std::vector<std::pair<std::string_view, WrappedJs
 
 OpResult<std::string> OpJsonGet(const OpArgs& op_args, string_view key,
                                 const JsonGetParams& params) {
-  OpResult<JsonType*> result = GetJson(op_args, key);
-  RETURN_ON_BAD_STATUS(result);
+  auto it = op_args.GetDbSlice().FindReadOnly(op_args.db_cntx, key).it;
+  if (!IsValid(it))
+    return OpStatus::KEY_NOTFOUND;
+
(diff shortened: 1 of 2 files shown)


chakaz (commented):
Maybe add a unit test?

BagritsevichStepan (commented):
What are we trying to achieve here? As I understand from the issue submitted by the user, they will store JSON just as blob strings. In this case, the basic `GET` command will be used to get the JSON blob.

Also, we are converting the blob string into `JsonType` here. Can you confirm that the reordering of elements does not occur during after this conversion?

Redis does not support this behavior:
```
127.0.0.1:6379> SET mykey '{"name": "John", "age": 30}'
OK
127.0.0.1:6379> JSON.GET mykey
(error) Existing key has wrong Redis type
```

And still other JSON commands are not supported.

romange:
Yes, it's an extra behavior that allows some runtime flexibility - a user can query strings that hold json objects. The main thing - it extends Redis behavior, so I do not see a problem here.
The ordering issue stays with `JSON.GET` of course, but now a user can fetch parts of the object, not the whole string, or use `GET` to fetch the string as is.

BagritsevichStepan:
> Yes, it's an extra behavior that allows some runtime flexibility - a user can query strings that hold json objects. The main thing - it extends Redis behavior, so I do not see a problem here. The ordering issue stays with `JSON.GET` of course, but now a user can fetch parts of the object, not the whole string, or use `GET` to fetch the string as is.

Are we going to support all other JSON.* commands for blob strings?

BagritsevichStepan (approved):

romange:
I think we may support additional read only commands but for now I would not rush to do so. Lets wait if community finds it useful.

PR #4400, opened 2025-01-03
fix: provide resp3 option to CapturingReplyBuilder

BorysTheDev:
fixes: https://github.com/dragonflydb/dragonfly/issues/4389

problem: we didn't provide the resp version into CapturingReplyBuilder

### DIFF
--- a/src/facade/reply_capture.h
+++ b/src/facade/reply_capture.h
@@ -43,8 +43,9 @@ class CapturingReplyBuilder : public RedisReplyBuilder {
   struct SimpleString : public std::string {};  // SendSimpleString
   struct BulkString : public std::string {};    // SendBulkString
 
-  CapturingReplyBuilder(ReplyMode mode = ReplyMode::FULL)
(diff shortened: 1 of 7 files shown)


src/facade/reply_capture.h
```diff
@@ -43,8 +43,9 @@ class CapturingReplyBuilder : public RedisReplyBuilder {
   struct SimpleString : public std::string {};  // SendSimpleString
   struct BulkString : public std::string {};    // SendBulkString
 
-  CapturingReplyBuilder(ReplyMode mode = ReplyMode::FULL)
+  CapturingReplyBuilder(bool is_resp_3, ReplyMode mode = ReplyMode::FULL)
```
chakaz, line 46: I'd like it better if we had an `enum class RespVersion { kResp2, kResp3 };` definition somewhere, and we didn't pass around booleans

chakaz: This enum could also be used in reply builder

chakaz (approved):
Thanks!

PR #4414, opened 2025-01-07
fix: invalid iterator case in Qlist::Erase

romange:
The problem was with reverse iterator that was not set properly when the last node is deleted.

Also, move PushSentinel code into Push.

### DIFF
--- a/src/core/qlist.cc
+++ b/src/core/qlist.cc
@@ -12,6 +12,7 @@ extern "C" {
 
 #include <absl/base/macros.h>
 #include <absl/base/optimization.h>
+#include <absl/strings/escaping.h>
 #include <absl/strings/str_cat.h>
 
 #include "base/logging.h"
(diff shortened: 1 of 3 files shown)


src/core/qlist.cc
```diff
@@ -354,12 +355,43 @@ void QList::Clear() {
...
+  DVLOG(2) << "Push " << absl::CHexEscape(value) << " " << (where == HEAD ? "HEAD" : "TAIL");
+
   /* The head and tail should never be compressed (we don't attempt to decompress them) */
   if (head_) {
     DCHECK(head_->encoding != QUICKLIST_NODE_ENCODING_LZF);
     DCHECK(head_->prev->encoding != QUICKLIST_NODE_ENCODING_LZF);
   }
-  PushSentinel(value, where);
+
+  quicklistNode* orig = head_;
```
romange, line 366: it's all moved from another function. no changes here.

src/core/qlist.cc
```diff
@@ -981,7 +980,7 @@ auto QList::Erase(Iterator it) -> Iterator {
       it.current_ = next;
       it.offset_ = 0;
     } else if (it.direction_ == REV) {
-      it.current_ = prev;
+      it.current_ = len_ ? prev : nullptr;
```
romange, line 983: that's the fix

BorysTheDev (approved):

PR #4416, opened 2025-01-07
fix: ExtentTree use of invalidated iterator

kostasrim:
The changes introduced in https://github.com/dragonflydb/dragonfly/blob/main/src/core/extent_tree.cc#L41 do not considerate that the cached iterator might be invalidated by the erase() above. As such, it causes an illegal access and it was caught by the sanitizers: https://github.com/dragonflydb/dragonfly/actions/runs/12644997917/job/35233466191#step:8:8607

### DIFF
--- a/src/core/extent_tree.cc
+++ b/src/core/extent_tree.cc
@@ -16,7 +16,7 @@ void ExtentTree::Add(size_t start, size_t len) {
   DCHECK_EQ(len_extents_.size(), extents_.size());
 
   auto it = extents_.lower_bound(start);
-  optional<absl::btree_map<size_t, size_t>::iterator> prev_extent;
+  optional<size_t> prev_extent_key;
 
   if (it != extents_.begin()) {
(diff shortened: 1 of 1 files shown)


src/core/extent_tree.cc
```diff
@@ -29,7 +29,7 @@ void ExtentTree::Add(size_t start, size_t len) {
       prev->second += len;
       start = prev->first;
       len += prev_len;
-      prev_extent = prev;
+      prev_extent_key = prev->first;
```
kostasrim, line 32: absl also issued a warning but the test did not fail. Maybe we should somehow be aware of these warnings (because they usually point to the root cause of the problem)

romange:
I thought it's fully compatible with `map`. Now I see it's not:
https://github.com/abseil/abseil-cpp/blob/master/absl/container/btree_map.h#L42C1-L48C41

romange (approved):

Issue #3839, opened 2024-10-01
transaction library shouldn't depend on the cluster files


BorysTheDev:
dfly_transaction lib shouldn't depend on cluster/cluster_utility.cc

dranikpg:
I'd suggest splitting everything into as much separate targets, as possible

PR #4417, opened 2025-01-07
refactor: remove transaction lib on cluster code dependency

BorysTheDev:
fixes: #3839

- no logic was changed
- remove transaction lib dependency on cluster
- remove extra headers

### DIFF
--- a/src/server/cluster/cluster_defs.cc
+++ b/src/server/cluster/cluster_defs.cc
@@ -1,26 +1,19 @@
+// Copyright 2024, DragonflyDB authors.  All rights reserved.
+// See LICENSE for licensing terms.
+//
 
-extern "C" {
-#include "redis/crc16.h"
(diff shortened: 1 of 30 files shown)


src/server/cluster/cluster_defs.cc
```diff
@@ -1,27 +1,17 @@
 
-extern "C" {
-#include "redis/crc16.h"
-}
+#include "cluster_defs.h"
```
romange, line 2: please add the file header

src/server/cluster_support.cc
```diff
@@ -0,0 +1,95 @@
+extern "C" {
```
romange, line 1: file header

romange (commented):
lgtm

romange (approved):

PR #4419, opened 2025-01-07
chore: reduce number of logs in cluster tests

BorysTheDev:
problem: we have timeout for fuzzy_migration_test https://github.com/dragonflydb/dragonfly/actions/runs/12648427037/job/35242743189 because of the number of the logs

fix: reduce number of logs in cluster tests


adiholden (approved):

kostasrim:
I was about to fix this and I noticed your PR :tada:

PR #4421, opened 2025-01-07
chore: split geo and zset families

kostasrim:
I was longing for this change for some time now and I grabbed the opportunity since I already worked on  #4420 

* Split geo family from zset family

I am not sure how I could reduce the size of the PR.

**There are no functional changes** only a refactor.

TODO for separate PR

* Split zset family test to reflect the changes here
* Maybe move the common functions from `ZSetFamily::` to a separate file

### DIFF
--- a/src/server/geo_family.h
+++ b/src/server/geo_family.h
@@ -1,0 +1,34 @@
+// Copyright 2025, DragonflyDB authors.  All rights reserved.
+// See LICENSE for licensing terms.
+//
+
+#pragma once
+
+#include <string_view>
+
+#include "server/common.h"
+
+namespace facade {
+class SinkReplyBuilder;
+}  // namespace facade
+
+namespace dfly {
+
+class CommandRegistry;
+struct CommandContext;
+
+class GeoFamily {
+ public:
+  static void Register(CommandRegistry* registry);
+  using SinkReplyBuilder = facade::SinkReplyBuilder;
+
+ private:
+  static void GeoAdd(CmdArgList args, const CommandContext& cmd_cntx);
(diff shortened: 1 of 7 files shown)


src/server/geo_family.h
```diff
@@ -0,0 +1,34 @@
...
+
+namespace dfly {
+
+class CommandRegistry;
+struct CommandContext;
+
+class GeoFamily {
+ public:
+  static void Register(CommandRegistry* registry);
+  using SinkReplyBuilder = facade::SinkReplyBuilder;
```
romange, line 23: I do not think you need it here, nor its declaration at line 12.

src/server/zset_family.h
```diff
@@ -57,10 +60,45 @@ class ZSetFamily {
...
+
+  using ScoredMember = std::pair<std::string, double>;
+  using ScoredArray = std::vector<ScoredMember>;
+  using ScoredMemberView = std::pair<double, std::string_view>;
+  using ScoredMemberSpan = absl::Span<const ScoredMemberView>;
+
   using SinkReplyBuilder = facade::SinkReplyBuilder;
+  template <typename T> using OpResult = facade::OpResult<T>;
+
+  static void ZAddGeneric(std::string_view key, const ZParams& zparams, ScoredMemberSpan memb_sp,
```
romange, line 77: what's the reason for exposing ZAddGeneric here?

romange: ah, i see.
please add a comment that it's being used in geo_family.cc code.

romange (approved):

PR #4428, opened 2025-01-08
fix: improve error propagation with RESTORE commands

romange:
Also, provide better logs if AddOrNew function fails adding a new entry

### DIFF
--- a/src/server/generic_family_test.cc
+++ b/src/server/generic_family_test.cc
@@ -698,6 +698,13 @@ TEST_F(GenericFamilyTest, Restore) {
   EXPECT_EQ(resp.GetString(), "OK");
   resp = Run({"zrange", "my-zset", "0", "-1"});
   EXPECT_EQ("elon", resp.GetString());
+
+  // corrupt the dump file but keep the crc correct.
+  ZSET_LISTPACK_DUMP[0] = 0x12;
+  uint8_t crc64[8] = {0x4e, 0xa3, 0x4c, 0x89, 0xc4, 0x8b, 0xd9, 0xe4};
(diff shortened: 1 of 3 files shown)


src/server/dragonfly_test.cc
```diff
@@ -673,7 +673,7 @@ TEST_F(DefragDflyEngineTest, TestDefragOption) {
 
   //  Fill data into dragonfly and then check if we have
   //  any location in memory to defrag. See issue #448 for details about this.
-  constexpr size_t kMaxMemoryForTest = 1'100'000;
+  constexpr size_t k4emoryForTest = 1'100'000;
```
kostasrim, line 676: typo?

romange:
I do not see much reason to use a templated Callback in JsonSet. @BagritsevichStepan will follow up with a PR at some point

src/server/generic_family_test.cc
```diff
@@ -845,4 +852,21 @@ TEST_F(GenericFamilyTest, ExpireTime) {
   EXPECT_EQ(expire_time_in_ms, CheckedInt({"PEXPIRETIME", "foo"}));
 }
 
+TEST_F(GenericFamilyTest, RestoreOOM) {
+  max_memory_limit = 20000000;
+  Run({"set", "src", string(5000, 'x')});
+  auto resp = Run({"dump", "src"});
+
+  string dump = resp.GetString();
+  usleep(10000);
```
chakaz, line 861: why sleep? (add comment)

chakaz (commented):
Generally looks good, a few minor questions

chakaz (approved):

PR #4368, opened 2024-12-25
chore: update jsoncons version to 0.178

romange:
jsoncons 0.178 has some pmr related fixes that should contribute to better stability in our codebase.

PR #4431, opened 2025-01-08
chore: Remove template Callbacks

BagritsevichStepan:


romange:
@BagritsevichStepan  please see https://github.com/dragonflydb/dragonfly/pull/4368
I do not understand why callback is needed at all

BagritsevichStepan:
> @BagritsevichStepan please see #4368 I do not understand why callback is needed at all

It was used just to reduce boilerplate code. And we are now performing the same check twice:
```
if (!parsed_json) {
    VLOG(1) << "got invalid JSON string '" << json_str << "' cannot be saved";
    return OpStatus::INVALID_VALUE;
  }
```

romange:
I think this is not justified to change interfaces of function to save a few lines. Moreover, it actually had a bug of copying a json object instead of moving it. With memory allocators is becomes a correctness bug where we probably loose the correct resource. All this is not worth it.

src/server/search/doc_accessors.cc
```diff
@@ -81,8 +81,8 @@ FieldValue ExtractSortableValueFromJson(const search::Schema& schema, string_vie
 }
 
 /* Returns true if json elements were successfully processed. */
-template <typename Callback>
-bool ProcessJsonElements(const std::vector<JsonType>& json_elements, Callback&& cb) {
+bool ProcessJsonElements(const std::vector<JsonType>& json_elements,
+                         absl::FunctionRef<bool(const JsonType&)> cb) {
```
kostasrim, line 85: plz see: https://github.com/dragonflydb/dragonfly/pull/4431#issuecomment-2577686761

Not sure what you agreed with Roman

BagritsevichStepan: It was addressed to another part of code

kostasrim: Yes yes I am aware. Sorry I was not very clear with my question, so: do we want to pass the `cb` here in the first place ? (I thought Roman did not want this in general -- maybe I am missing something).

Furthermore, IMO I don't see a reason to substitute here with `absl::FunctionRef` ?

kostasrim: > So, it's just another code where the template callback needs to be removed

But `absl::FunctionRef` is just a callback (similar to the templated F with small differences beyond the scope of this discussion). Either way, the cb here is fine, all you do is `convert the json to a string and assigning it to the internal buffer`

kostasrim: and IMO, `ProcessJsonElements` is only called in one place, so why don't we remove the callback alltogether and copy paste the body of the lambda there ?

BagritsevichStepan: I see that it is called from two places

kostasrim: You are right, I was looking for the wrong thing

kostasrim (approved):
Ask Roman if he wants to eliminate the callbacks alltogether before you merge this

BagritsevichStepan:
@romange

romange (approved):

PR #4433, opened 2025-01-09
fix: show managed stuff on privileged port

romange:

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -2255,7 +2255,8 @@ void ServerFamily::Info(CmdArgList args, const CommandContext& cmd_cntx) {
     append("dragonfly_version", GetVersion());
     append("redis_mode", GetRedisMode());
     append("arch_bits", 64);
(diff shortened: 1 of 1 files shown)


src/server/server_family.cc
```diff
@@ -2255,7 +2255,8 @@ void ServerFamily::Info(CmdArgList args, const CommandContext& cmd_cntx) {
     append("dragonfly_version", GetVersion());
     append("redis_mode", GetRedisMode());
     append("arch_bits", 64);
-    if (!absl::GetFlag(FLAGS_managed_service_info)) {
+
+    if (!absl::GetFlag(FLAGS_managed_service_info) || cmd_cntx.conn_cntx->conn()->IsPrivileged()) {
```
chakaz, line 2259: I'd wrap this in a lambda, but your call

chakaz (approved):

src/server/server_family.cc
romange: i did not notice it's the same function. Added a variable.

chakaz (approved):

PR #4434, opened 2025-01-09
chore: dragonfly connection refactorings

romange:
1. Move socket read code into a dedicated function. Remove std:: prefix in the code.
2. Add an optional iouring bufring registration. Currently not being used and is disabled by default.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -87,6 +87,7 @@ ABSL_FLAG(bool, migrate_connections, true,
           "happen at most once per connection.");
 
 using namespace util;
+using namespace std;
 using absl::GetFlag;
 using nonstd::make_unexpected;
 
@@ -239,7 +240,7 @@ void LogTraffic(uint32_t id, bool has_more, absl::Span<RespExpr> resp,
   }
 
   // Write the data itself.
-  std::array<iovec, 16> blobs;
+  array<iovec, 16> blobs;
   unsigned index = 0;
   if (next != stack_buf) {
     blobs[index++] = iovec{.iov_base = stack_buf, .iov_len = size_t(next - stack_buf)};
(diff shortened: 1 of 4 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -1241,6 +1243,29 @@ void Connection::HandleMigrateRequest() {
   }
 }
 
+error_code Connection::HandleRecvSocket() {
```
romange, line 1246: the only change in this file. the code is moved to HandleRecvSocket

kostasrim: + the boilerplate in RegisterBufRings

kostasrim (approved):
lgtm

src/facade/dragonfly_connection.cc
```diff
@@ -87,6 +87,7 @@ ABSL_FLAG(bool, migrate_connections, true,
           "happen at most once per connection.");
 
 using namespace util;
+using namespace std;
```
kostasrim, line 90: I am allergic to `unqualified lookup` unless it's used in generic code :smile: 

*Ignore me* :smile:

romange: i do not know what it means but I do not think our code has lots of ambiguous names. as long as we enable it locally in cc file I think it's fine.

kostasrim: Long story short there are different taxonomy rules for looking up names qualified (e.g, `std::some_algo` etc) vs unqualified (e.g, `some_algo`). This can lead to many surprises that are amplified in the presence of generic code (templates and their two phase lookup upon parsing and instantiation). 

A simple example without templates:

```
// Omitting include<algorithm> on purpose
std::accumulate(c.begin(), c.end(), adder());
```

This is easy, the compiler looks `within the std` namespace and it finds `std::accumulate`. 

Now if you do the same `unqualified`
```
accumulate(...)
``

The compiler will look in `namespaces` that are `associated` with the arguments passed in the function call. In our case it's `Vector::Iterator`. Now here is a little bit of madness: **this is NOT platform independend code***. It might work on some machines and it might fail to compile for others. Why? 

Because the `cpp standard does not mandate if the iterator is just a pointer or if it's wrapper struct`. If it's a typedef to an `int` then guess what! The program won't compile. If the implementation is a `struct Iterator` it will because the rules mandate that the namespaces for the compiler to search incldue the namespace of the class. However if it was an int, these namespaces are ignored so the compiler `searches in a different way`. 

This is just without the templates. You get templates and customization points and now you are dancing with `tag_dispatch` and the infamous `two-step dance` (using declaration + followed by an unqualified lookup)

PR #4435, opened 2025-01-09
test: add test_cache_cluster_data_migration

BorysTheDev:
fixes: #4354

### DIFF
--- a/tests/dragonfly/cluster_test.py
+++ b/tests/dragonfly/cluster_test.py
@@ -2612,3 +2612,103 @@ async def test_migration_timeout_on_sync(df_factory: DflyInstanceFactory, df_see
     await push_config(json.dumps(generate_config(nodes)), [node.admin_client for node in nodes])
 
     assert (await StaticSeeder.capture(nodes[1].client)) == start_capture
+
+
+@pytest.mark.slow
+@dfly_args({"proactor_threads": 4, "cluster_mode": "yes", "cache_mode": "true"})
+async def test_cache_cluster_data_migration(df_factory: DflyInstanceFactory, df_seeder_factory):
+    # Check data migration from one node to another with eviction and expiration
+    # not cluster clients are used to avoid processing -MOVE error and have ability compare results
+    instances = [
+        df_factory.create(
+            port=next(next_port),
+            admin_port=next(next_port),
+            vmodule="outgoing_slot_migration=2,cluster_family=2,incoming_slot_migration=2,streamer=2",
+            maxmemory="1G" if i == 0 else "3G",
+        )
+        for i in range(2)
+    ]
+
+    df_factory.start_all(instances)
+
+    nodes = [(await create_node_info(instance)) for instance in instances]
+    nodes[0].slots = [(0, 16383)]
+    nodes[1].slots = []
+
+    await push_config(json.dumps(generate_config(nodes)), [node.admin_client for node in nodes])
+
+    logging.debug("Start seeder")
+    await nodes[0].client.execute_command("DEBUG POPULATE 700000 test 1500 RAND SLOTS 0 16383")
+    # cluster_mode=False, to avoid processing MOVE error and prevent eviction on the target node after migration is finished
+    # stop_on_failure=False, the test shouldn't failed because of unprocessed MOVE error
+    # multi_transaction_probability=0 and max_multikey=1, because cluster can't work with them
+    seeder = df_seeder_factory.create(
+        val_size=1500,
+        stop_on_failure=False,
+        port=nodes[0].instance.port,
+        cluster_mode=False,
+        multi_transaction_probability=0,
+        max_multikey=1,
+    )
+
+    fill_evict_task = asyncio.create_task(seeder.run())
+
+    gen_stop = False
+
+    async def expiration_gen():
+        i = 0
+        try:
+            while not gen_stop:
+                await nodes[0].client.execute_command(f"SET exp_key_{i} val_{i} EX 20")
+                i += 1
+        except redis.exceptions.ResponseError:
+            return
+
+    fill_exp_task = asyncio.create_task(expiration_gen())
+
+    # some time to fill data
+    await asyncio.sleep(20)
+
+    stats = await nodes[0].client.info("STATS")
+    evicted_before_migration = stats["evicted_keys"]
+
+    logging.debug("Start migration")
+    nodes[0].migrations.append(
+        MigrationInfo("127.0.0.1", nodes[1].instance.admin_port, [(0, 16383)], nodes[1].id)
+    )
+    await push_config(json.dumps(generate_config(nodes)), [node.admin_client for node in nodes])
+
+    await wait_for_status(nodes[1].admin_client, nodes[0].id, "FINISHED", 200)
+
+    logging.debug("Stop seeders")
+    seeder.stop()
+    await fill_evict_task
+    gen_stop = True
+    await fill_exp_task
+
+    # different configs are provided for target and source nodes
+    # to get data from the source node, the migration is canceled
+    # for the same reason, the migration is finished on the target node
+    # in such a way every node should contain and provide the same data
+    logging.debug("drop migration for 0th node")
+    nodes[0].migrations = []
+    await push_config(json.dumps(generate_config(nodes)), [nodes[0].admin_client])
+
+    logging.debug("finish migration for 1st")
+    nodes[0].slots = []
+    nodes[1].slots = [(0, 16383)]
+    await push_config(json.dumps(generate_config(nodes)), [nodes[1].admin_client])
+
+    # wait to expire all keys
+    await asyncio.sleep(20)
+
+    source_capture_task = asyncio.create_task(StaticSeeder.capture(nodes[0].client))
+    assert await StaticSeeder.capture(nodes[1].client) == await source_capture_task
+
+    stats = await nodes[0].client.info("STATS")
+    assert stats["evicted_keys"] > evicted_before_migration
+    assert stats["expired_keys"] > 0
+
+    stats = await nodes[1].client.info("STATS")
+    assert stats["evicted_keys"] == 0
+    assert stats["expired_keys"] > 0


tests/dragonfly/cluster_test.py
```diff
@@ -2580,3 +2580,91 @@ async def test_migration_timeout_on_sync(df_factory: DflyInstanceFactory, df_see
...
+    await push_config(json.dumps(generate_config(nodes)), [nodes[0].admin_client])
+
+    logging.debug("finish migration")
+    nodes[0].slots = []
+    nodes[1].slots = [(0, 16383)]
+    await push_config(json.dumps(generate_config(nodes)), [nodes[1].admin_client])
+
+    await asyncio.sleep(1)
+
+    source_capture = await StaticSeeder.capture(nodes[0].client)
```
adiholden, line 2669: could it be that node 0 continues evicting items after the migrations is done? therefore comparing node0 data to node1 data is failing

BorysTheDev: It looks the issue in the data modification

src/server/cluster_support.h
```diff
@@ -26,8 +22,15 @@ class UniqueSlotChecker {
...
   std::optional<SlotId> GetUniqueSlotId() const;
 
+  void Reset() {
+    slot_id_ = kEmptySlotId;
+  }
+
  private:
-  std::optional<SlotId> slot_id_;
+  static constexpr SlotId kEmptySlotId = kMaxSlotNum + 1;
+  static constexpr SlotId kInvalidSlotId = kEmptySlotId + 1;
```
chakaz, line 31: Please add a comment explaining what's the difference between these 2 and why we need 2 distinct values

src/server/transaction.cc
```diff
@@ -325,6 +325,7 @@ void Transaction::InitByKeys(const KeyIndex& key_index) {
   // Stub transactions always operate only on single shard.
   bool is_stub = multi_ && multi_->role == SQUASHED_STUB;
 
+  unique_slot_checker_.Reset();
```
chakaz, line 328: Does this fix a real bug? Because the PR title talks about a test. Please update the title..

src/server/cluster_support.h
BorysTheDev: I think I provide not the best name:
is kNoSlotID will be better? I think in this case we don't need a comment

chakaz: `kNoSlotId` and `kInvalidSlotId`?
Still, I find them confusing.. I wish we could only use one, but if it simplifies I think it merits a comment

BorysTheDev: ok

src/server/cluster_support.h
```diff
@@ -26,8 +22,17 @@ class UniqueSlotChecker {
...
 
+  void Reset() {
+    slot_id_ = kNoSlotId;
+  }
+
  private:
-  std::optional<SlotId> slot_id_;
+  // kNoSlotId - if slot wasn't set at all
+  static constexpr SlotId kNoSlotId = kMaxSlotNum + 1;
+  // kInvalidSlotId - if several different slots was set
```
chakaz, line 32:
```suggestion
  // kInvalidSlotId - if several different slots were set
```

src/server/cluster_support.h
```diff
@@ -26,8 +22,17 @@ class UniqueSlotChecker {
...
+  static constexpr SlotId kInvalidSlotId = kNoSlotId + 1;
```
chakaz, line 33: I just noticed that this is private & internal to this class. Can we move these declarations to the `.cc` file? It'll be cleaner and less confusing (I thought that this was exposed)

BorysTheDev: in this case, kNoSlotId should be public, or moved to and I need to write a constructor. So I would prefer as is

src/server/multi_command_squasher.cc
```diff
@@ -77,15 +77,14 @@ MultiCommandSquasher::MultiCommandSquasher(absl::Span<StoredCmd> cmds, Connectio
...
-    ShardId sid, optional<SlotId> slot_id) {
+MultiCommandSquasher::ShardExecInfo& MultiCommandSquasher::PrepareShardInfo(ShardId sid) {
   if (sharded_.empty())
     sharded_.resize(shard_set->size());
 
   auto& sinfo = sharded_[sid];
   if (!sinfo.local_tx) {
     if (IsAtomic()) {
-      sinfo.local_tx = new Transaction{cntx_->transaction, sid, slot_id};
+      sinfo.local_tx = new Transaction{cntx_->transaction, sid, nullopt};
```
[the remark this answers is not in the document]
BorysTheDev: because we reuse the transaction, and there is no sense to add it here.

tests/dragonfly/cluster_test.py
```diff
@@ -2580,3 +2580,68 @@ async def test_migration_timeout_on_sync(df_factory: DflyInstanceFactory, df_see
...
+        stop_on_failure=False,
+        port=nodes[0].instance.port,
+        cluster_mode=False,
+        multi_transaction_probability=0,
+        max_multikey=1,
+    )
+
+    fill_task = asyncio.create_task(seeder.run())
+    # some time to evict data
+    await asyncio.sleep(10)
```
adiholden, line 2621: how do you know data is evicted in this test?
I dont see stats check, right?
How do you know data is evicted while migrations is in progress?

tests/dragonfly/cluster_test.py
```diff
@@ -2580,3 +2580,68 @@ async def test_migration_timeout_on_sync(df_factory: DflyInstanceFactory, df_see
     await push_config(json.dumps(generate_config(nodes)), [node.admin_client for node in nodes])
 
     assert (await StaticSeeder.capture(nodes[1].client)) == start_capture
+
+
+@dfly_args({"proactor_threads": 4, "cluster_mode": "yes", "cache_mode": "true"})
+async def test_cache_cluster_data_migration(df_factory: DflyInstanceFactory, df_seeder_factory):
+    # Check data migration from one node to another with expiration and eviction
```
adiholden, line 2587: you write with expiration but I dont see anything for expire keys here

adiholden: please add more description on this test. how you check the eviction flow. you took some decisions to use the non cluster client, this needs to be documented to explain why this decision was taken

tests/dragonfly/cluster_test.py
```diff
@@ -2580,3 +2580,68 @@ async def test_migration_timeout_on_sync(df_factory: DflyInstanceFactory, df_see
     await push_config(json.dumps(generate_config(nodes)), [node.admin_client for node in nodes])
 
     assert (await StaticSeeder.capture(nodes[1].client)) == start_capture
+
+
+@dfly_args({"proactor_threads": 4, "cluster_mode": "yes", "cache_mode": "true"})
```
adiholden, line 2585: do you want to seprate the test from the fix PR?
I approve the fixes for dfly code but the test needs a little more work

tests/dragonfly/cluster_test.py
```diff
@@ -2580,3 +2580,68 @@ async def test_migration_timeout_on_sync(df_factory: DflyInstanceFactory, df_see
...
+    logging.debug("Start seeder")
+    await nodes[0].client.execute_command("DEBUG POPULATE 80000 test 10000 RAND SLOTS 0 16383")
+    # cluster_mode set to off to prevent eviction on the target node after migration is finished
+    seeder = df_seeder_factory.create(
+        keys=1000,
+        val_size=10000,
+        stop_on_failure=False,
+        port=nodes[0].instance.port,
+        cluster_mode=False,
+        multi_transaction_probability=0,
```
adiholden, line 2615: why do you set this parameters? this needs documentaion

tests/dragonfly/cluster_test.py
```diff
@@ -2580,3 +2580,68 @@ async def test_migration_timeout_on_sync(df_factory: DflyInstanceFactory, df_see
...
+    )
+    await push_config(json.dumps(generate_config(nodes)), [node.admin_client for node in nodes])
+
+    await wait_for_status(nodes[1].admin_client, nodes[0].id, "FINISHED", 30)
+
+    logging.debug("Stop seeder")
+    seeder.stop()
+    await fill_task
+
+    logging.debug("drop migration for 0th node")
```
adiholden, line 2635: this needs to be documented why we need to do this to be able to test the eviction

tests/dragonfly/cluster_test.py
BorysTheDev: didn't get. 
stop_on_failure=False, because I don't want to get error when the slots will be moved
cluster_mode=False, because I don't want to process MOVE error
multi_transaction_probability is 0 when you set cluster flag for seeder, because cluster can't work with multi transactions

tests/dragonfly/cluster_test.py
BorysTheDev: ok I got it now. I will add comments, I thought it was clear after our discussion regarding cluster migration testing

adiholden: To me it is clear, but for someone who was not on this meeting it is not clear. And this needs documentation

tests/dragonfly/cluster_test.py
```diff
@@ -2612,3 +2612,103 @@ async def test_migration_timeout_on_sync(df_factory: DflyInstanceFactory, df_see
...
+        cluster_mode=False,
+        multi_transaction_probability=0,
+        max_multikey=1,
+    )
+
+    fill_evict_task = asyncio.create_task(seeder.run())
+
+    gen_stop = False
+
+    async def expiration_gen():
```
adiholden, line 2658: you can use ExpirySeeder

tests/dragonfly/cluster_test.py
```diff
@@ -2612,3 +2612,103 @@ async def test_migration_timeout_on_sync(df_factory: DflyInstanceFactory, df_see
...
+        i = 0
+        try:
+            while not gen_stop:
+                await nodes[0].client.execute_command(f"SET exp_key_{i} val_{i} EX 20")
```
adiholden, line 2662: When I run your test the first call here gets OOM exeptions therefor no command here is executed

tests/dragonfly/cluster_test.py
```diff
@@ -2612,3 +2612,103 @@ async def test_migration_timeout_on_sync(df_factory: DflyInstanceFactory, df_see
...
+                i += 1
+        except redis.exceptions.ResponseError:
+            return
+
+    fill_exp_task = asyncio.create_task(expiration_gen())
+
+    # some time to fill data
+    await asyncio.sleep(20)
```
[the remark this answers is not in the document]
BorysTheDev: to fill enough data and start some keys expire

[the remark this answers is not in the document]
adiholden: the first debug populate fills enough data for evictions as far as I understand.
also you can use a smaller expiration time to check expiry f.e 2 seconds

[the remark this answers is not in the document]
BorysTheDev: 2 seconds isn't enough because it's random

tests/dragonfly/cluster_test.py
BorysTheDev: I think it's ok if test is finished successfully

tests/dragonfly/cluster_test.py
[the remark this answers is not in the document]
adiholden: what is random?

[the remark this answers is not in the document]
BorysTheDev: we don't know what the slot of the next key and if the interval is small we can get already migrated slots keys or they will expire before migration

tests/dragonfly/cluster_test.py
adiholden: last assertion assert stats["expired_keys"] > 0

BorysTheDev:
Eviction can proceed after the migration is finished so current approach is flaky

PR #4444, opened 2025-01-13
feat(json_family): Add json_nesting_depth_limit flag

BagritsevichStepan:
This will slightly reduce processing time and memory allocation on the heap. Redis has its own limit set to [128](https://docs.aws.amazon.com/AmazonElastiCache/latest/dg/json-document-overview.html#json-nesting-depth-limit).

[Inspired by this discussion](https://github.com/danielaparker/jsoncons/issues/531#issuecomment-2293758437)

### DIFF
--- a/src/core/json/json_object.cc
+++ b/src/core/json/json_object.cc
@@ -4,16 +4,10 @@
 
 #include "core/json/json_object.h"
 
-#include "base/flags.h"
 #include "base/logging.h"
 
 using namespace jsoncons;
 using namespace std;
-
-ABSL_FLAG(uint32_t, json_nesting_depth_limit, 128,
-          "The maximum allowed nesting depth for JSON documents. "
-          "Any attempt to create a document with a nesting depth greater than the specified limit "
-          "will result in an error.");
 
 namespace dfly {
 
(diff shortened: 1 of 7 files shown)


src/core/json/json_object.cc
```diff
@@ -4,11 +4,17 @@
 
 #include "core/json/json_object.h"
 
+#include "base/flags.h"
 #include "base/logging.h"
 
 using namespace jsoncons;
 using namespace std;
 
+ABSL_FLAG(uint32_t, json_nesting_depth_limit, 128,
```
kostasrim, line 13: @romange

romange:
@BagritsevichStepan  what happens if this limit is crossed?

src/core/json/json_object.cc
```diff
@@ -18,8 +24,14 @@ optional<JsonType> JsonFromString(string_view input, PMR_NS::memory_resource* mr
     return false;
   };
 
+  /* The maximum possible JSON nesting depth is either the specified depth_limit or half of the
+     input size. Since nesting a JSON object requires at least 2 characters. */
+  const uint32_t depth_limit = absl::GetFlag(FLAGS_json_nesting_depth_limit);
```
romange, line 29: I think a reasonable constant is better, no need to expose a flag for this.
Also, please add a test that checks it.

BagritsevichStepan:
Any attempt to create a document with a nesting depth greater than the specified limit will result in an error

src/core/json/json_object.cc
BagritsevichStepan: Since it will reject a document, I thought it would be worth adding it as a flag, e.g., if the user wants to specify it for their case and is confident that slower processing is acceptable for them

romange: but bigger value flag will require more stack? or it is about heap allocations?

romange: Having 256 is reasonable constant. If someone uses a json document with more depth - they should seek professional help.

BagritsevichStepan: Yes, it is about heap allocations

romange (approved):

PR #4467, opened 2025-01-17
feat(dev): allow monitoring a valkey server on localhost

romange:


kostasrim (dismissed):
I trust you that it works :smile: 

P.s. These are taken as is right ? (I mean it's not our code correct?)

kostasrim (approved):

PR #4474, opened 2025-01-17
fix: hiredis requires df to report version >7.4

### LINKED ISSUE #4473: dragonfly reports redis_version:7.2.0 which is incompatible with hiredis
hiredis rejects `ping` command because we report version `7.2` instead of `>7.4`.

See also https://github.com/dragonflydb/dragonfly/issues/2454#issuecomment-2587021353

kostasrim:
hiredis checks for redis version and requires to be greater than 7.4. Dragonfly reports 7.2 and the library throws an error. 

* bump up redis version in info command
* add compatibility test
* bump up py dependencies
* fix warnings and deprecated functions

Resolves #4473

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -230,7 +230,7 @@ using strings::HumanReadableNumBytes;
 
 namespace {
 
-const auto kRedisVersion = "7.2.0";
+const auto kRedisVersion = "7.4.0";
 
 using EngineFunc = void (ServerFamily::*)(CmdArgList args, const CommandContext&);
 
--- a/tests/dragonfly/requirements.txt
+++ b/tests/dragonfly/requirements.txt
@@ -1,14 +1,14 @@
-async-timeout==4.0.2
+async-timeout==4.0.3
 attrs==22.1.0
 Deprecated==1.2.13
 iniconfig==1.1.1
-packaging==21.3
+packaging==23.1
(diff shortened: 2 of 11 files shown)


tests/dragonfly/requirements.txt
```diff
@@ -26,3 +26,4 @@ pytest-icdiff==0.8
 pytest-timeout==2.2.0
 asyncio==3.4.3
 fakeredis[json]==2.26.2
+hiredis==2.4.0
```
kostasrim, line 29: Throwing this here, hiredis 3.0 is incompatible with redis 5.2 (latest release). 

TODO bump up the versions once they are fixed

kostasrim: More on this https://github.com/redis/hiredis-py/pull/189 
and 
https://github.com/redis/redis-py/issues/3358#issuecomment-2553434973

src/server/server_family.cc
```diff
@@ -230,7 +230,7 @@ using strings::HumanReadableNumBytes;
 
 namespace {
 
-const auto kRedisVersion = "7.2.0";
+const auto kRedisVersion = "7.4.0";
```
romange, line 233: consider exposing it in common.h as a constant (or other header file) and then reference it in tests reduce amount of manual changes

romange (approved):
LGTM

PR #4476, opened 2025-01-19
chore: add Dash::Prefetch function

romange:
It's not being used at this time.

### DIFF
--- a/src/core/dash_internal.h
+++ b/src/core/dash_internal.h
@@ -1193,8 +1193,6 @@ template <typename Key, typename Value, typename Policy>
 void Segment<Key, Value, Policy>::Prefetch(Hash_t key_hash) const {
   uint8_t bidx = BucketIndex(key_hash);
   const Bucket& target = bucket_[bidx];
-  uint8_t nid = NextBid(bidx);
-  const Bucket& probe = bucket_[nid];
 
   // Prefetch the home bucket that might hold the key with high probability.
   __builtin_prefetch(&target, 0, 1);
--- a/src/core/dash_internal.h
+++ b/src/core/dash_internal.h
@@ -502,6 +502,7 @@ template <typename _Key, typename _Value, typename Policy = DefaultSegmentPolicy
 
   // Find item with given key hash and truthy predicate
   template <typename Pred> Iterator FindIt(Hash_t key_hash, Pred&& pred) const;
+  void Prefetch(Hash_t key_hash) const;
 
   // Returns valid iterator if succeeded or invalid if not (it's full).
   // Requires: key should be not present in the segment.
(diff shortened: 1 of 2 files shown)


src/core/dash_internal.h
```diff
@@ -1188,6 +1189,17 @@ auto Segment<Key, Value, Policy>::FindIt(Hash_t key_hash, Pred&& pred) const ->
   return Iterator{};
 }
 
+template <typename Key, typename Value, typename Policy>
+void Segment<Key, Value, Policy>::Prefetch(Hash_t key_hash) const {
+  uint8_t bidx = BucketIndex(key_hash);
+  const Bucket& target = bucket_[bidx];
+  uint8_t nid = NextBid(bidx);
+  const Bucket& probe = bucket_[nid];
```
BorysTheDev, line 1197: what the purpose of these 2 lines

src/core/dash_internal.h
```diff
@@ -1188,6 +1189,15 @@ auto Segment<Key, Value, Policy>::FindIt(Hash_t key_hash, Pred&& pred) const ->
   return Iterator{};
 }
 
+template <typename Key, typename Value, typename Policy>
+void Segment<Key, Value, Policy>::Prefetch(Hash_t key_hash) const {
+  uint8_t bidx = BucketIndex(key_hash);
+  const Bucket& target = bucket_[bidx];
+
+  // Prefetch the home bucket that might hold the key with high probability.
+  __builtin_prefetch(&target, 0, 1);
```
BorysTheDev, line 1198: what is the bucket size? if the size of the bucket is more than the cache line maybe it will be better to load the whole bucket

romange: the bucket size is dependent on the key/value size but the bucket metadata is located at the beginning. we use the metadata/fingerprints to do the initial scan.

BorysTheDev: sizeof(Bucket) > 128. I think we can load it whole

romange:
```
SlotBitmap<NUM_SLOTS> slotb_;  // allocation bitmap + pointer bitmap + counter

  /*only use the first 14 bytes, can be accelerated by
    SSE instruction,0-13 for finger, 14-17 for overflowed*/
  FpArray finger_arr_;
  StashFpArray stash_arr_;

  uint8_t stash_busy_ = 0;  // kStashFpLen+1 bits are used
  uint8_t stash_pos_ = 0;   // 4x2 bits for pointing to stash bucket.
```
BucketBase<NUM_SLOTS, NUM_OVR>::CompareFP  is the function that usually hits a cold memory

BorysTheDev (approved):

Issue #4204, opened 2024-11-27
Add tests for the SEARCH commands


BagritsevichStepan:
We should not crash during parsing different options in `FT.*` commands. For example: (empty separator in FT.CREATE)
```
127.0.0.1:6379> ft.create index on json schema $.sometag AS sometag TAG SEPARATOR
Error: Server closed the connection
```

So we should add parsing tests for all `FT.*` commands, especially for the `FT.CREATE`, `FT.SEARCH` and `FT.AGGREGATE` commands.

- [x] #4208
- [x] Add tests for different options in the `FT.CREATE` command. Fixed by #4678
- [x] Add tests for different options in the `FT.SEARCH` command. Fixed by #4478
- [x] Add tests for different options in the `FT.AGGREGATE` command. Fixed by #4479
- [ ] Test all other FT.* commands

PR #4479, opened 2025-01-20
fix(search_family): Add options test for the FT.AGGREGATE command

BagritsevichStepan:
related to the #4204

1. Fix some small issues during parsing of the several SEARCH methods
2. Remove using builder in Parse* methods.
    It was a bad decision to use it because, in some cases, we were returning an error with `builder->SendError(...)`. However, in those cases, we were not handling parser errors with `parser->Error()`. As a result, debug builds were crashing because of this check:
```cpp
CmdArgParser::~CmdArgParser() {
  DCHECK(!error_.has_value()) << "Parsing error occured but not checked";
}
```
Now, we ensure that parsing errors are processed in all cases.
3. Add options test for the `FT.AGGREGATE` command

### DIFF
--- a/src/facade/cmd_arg_parser.h
+++ b/src/facade/cmd_arg_parser.h
@@ -119,7 +119,7 @@ struct CmdArgParser {
 
   // Check if the next value is equal to a specific tag. If equal, its consumed.
   template <class... Args> bool Check(std::string_view tag, Args*... args) {
-    if (cur_i_ + sizeof...(Args) >= args_.size())
+    if (cur_i_ + sizeof...(Args) >= args_.size() || error_)
       return false;
 
     std::string_view arg = SafeSV(cur_i_);
--- a/src/facade/cmd_arg_parser.h
+++ b/src/facade/cmd_arg_parser.h
@@ -119,7 +119,7 @@ struct CmdArgParser {
 
   // Check if the next value is equal to a specific tag. If equal, its consumed.
   template <class... Args> bool Check(std::string_view tag, Args*... args) {
-    if (cur_i_ + sizeof...(Args) >= args_.size() || error_)
+    if (cur_i_ + sizeof...(Args) >= args_.size())
       return false;
 
     std::string_view arg = SafeSV(cur_i_);
--- a/src/server/search/search_family.cc
+++ b/src/server/search/search_family.cc
@@ -128,15 +128,76 @@ ParseResult<search::SchemaField::TagParams> ParseTagParams(CmdArgParser* parser)
 #pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
 #endif
 
-ParseResult<search::Schema> ParseSchema(DocIndex::DataType type, CmdArgParser* parser) {
-  search::Schema schema;
+using ParsedSchemaField =
+    ParseResult<std::pair<search::SchemaField::FieldType, search::SchemaField::ParamsVariant>>;
+
+// Tag fields include: [separator char] [casesensitive]
+ParsedSchemaField ParseTag(CmdArgParser* parser) {
+  auto tag_params = ParseTagParams(parser);
+  if (!tag_params) {
+    return make_unexpected(tag_params.error());
+  }
+  return std::make_pair(search::SchemaField::TAG, std::move(tag_params).value());
+}
+
+ParsedSchemaField ParseText(CmdArgParser* parser) {
+  return std::make_pair(search::SchemaField::TEXT, std::monostate{});
+}
+
+ParsedSchemaField ParseNumeric(CmdArgParser* parser) {
+  return std::make_pair(search::SchemaField::NUMERIC, std::monostate{});
+}
+
+// Vector fields include: {algorithm} num_args args...
+ParsedSchemaField ParseVector(CmdArgParser* parser) {
+  auto vector_params = ParseVectorParams(parser);
+
+  if (parser->HasError()) {
+    auto err = *parser->Error();
+    VLOG(1) << "Could not parse vector param " << err.index;
+    return CreateSyntaxError("Parse error of vector parameters"sv);
+  }
+
+  if (vector_params.dim == 0) {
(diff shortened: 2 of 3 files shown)


src/facade/cmd_arg_parser.h
```diff
@@ -119,7 +119,7 @@ struct CmdArgParser {
 
   // Check if the next value is equal to a specific tag. If equal, its consumed.
   template <class... Args> bool Check(std::string_view tag, Args*... args) {
-    if (cur_i_ + sizeof...(Args) >= args_.size())
+    if (cur_i_ + sizeof...(Args) >= args_.size() || error_)
```
BorysTheDev, line 122: Why?

src/server/search/search_family.cc
```diff
@@ -36,6 +36,31 @@ using namespace facade;
...
+
+template <typename T> using ParseResult = io::Result<T, ErrorReply>;
+
+nonstd::unexpected_type<ErrorReply> CreateSyntaxError(std::string message) {
+  return make_unexpected(ErrorReply{std::move(message), kSyntaxErrType});
+}
+
+nonstd::unexpected_type<ErrorReply> CreateSyntaxError(std::string_view message) {
+  return make_unexpected(ErrorReply{message, kSyntaxErrType});
+}
```
BorysTheDev, line 49: implementation of these functions is shorter than their names, I don't think that you need them

src/server/search/search_family.cc
```diff
@@ -104,66 +128,59 @@ std::optional<search::SchemaField::TagParams> ParseTagParams(CmdArgParser* parse
...
       }
       params = vector_params;
     }
 
     // Flags: check for SORTABLE and NOINDEX
     uint8_t flags = 0;
-    while (parser.HasNext()) {
-      if (parser.Check("NOINDEX")) {
+    while (parser->HasNext()) {
+      if (parser->Check("NOINDEX")) {
```
BorysTheDev, line 178: maybe we can use here TryMapNext

src/server/search/search_family.cc
```diff
@@ -104,66 +128,59 @@ std::optional<search::SchemaField::TagParams> ParseTagParams(CmdArgParser* parse
...
 
     // Determine type
     using search::SchemaField;
-    auto type = parser.MapNext("TAG", SchemaField::TAG, "TEXT", SchemaField::TEXT, "NUMERIC",
-                               SchemaField::NUMERIC, "VECTOR", SchemaField::VECTOR);
-    if (auto err = parser.Error(); err) {
-      builder->SendError(err->MakeReply());
-      return nullopt;
-    }
+    auto type = parser->MapNext("TAG"sv, SchemaField::TAG, "TEXT"sv, SchemaField::TEXT, "NUMERIC"sv,
```
BorysTheDev, line 148: Consider instead of SchemaField type use function pointer 
`std::pair<SchemaField, search::SchemaField::ParamsVariant> (*f)(CmdArgParser*)`

src/server/search/search_family.cc
```diff
@@ -172,28 +189,65 @@ optional<search::Schema> ParseSchemaOrReply(DocIndex::DataType type, CmdArgParse
...
+      auto schema = ParseSchema(index.type, parser);
+      if (!schema)
+        return make_unexpected(schema.error());
+      index.schema = std::move(*schema);
+      break;  // SCHEMA always comes last
+    }
+
+    // Unsupported parameters are ignored for now
+    parser->Skip(1);
+  }
```
BorysTheDev, line 246: Consider using tryMapNext and return function that can parse the next args

src/server/search/search_family.cc
BagritsevichStepan: Why it is shorter?
```
make_unexpected(ErrorReply{...}, kSyntaxErrType})
vs
CreateSyntaxError(...)
```
I think it is worth to keep this methods

BorysTheDev: return CreateSyntaxError("message);
return {{"message", kSyntaxErrType}};

but it's not critical

BorysTheDev: By the way yo don't need 2 CreateSyntaxError functions. Just use with std::string one if you want it

BagritsevichStepan: We have explicit constructors for ErrorReply

BagritsevichStepan: and if you want to return error for io::Result you also can not use {} as I remeber

src/facade/cmd_arg_parser.h
BagritsevichStepan: I think that if an error occurred previously during parsing, why are we still returning true for `parser->Check("TAG")`, for example? It should simply skip all checks

BorysTheDev: We already have an error so further actions don't matter to us if it's true or false. So we don't need to do extra checks

src/server/search/search_family.cc
```diff
@@ -104,96 +128,148 @@ std::optional<search::SchemaField::TagParams> ParseTagParams(CmdArgParser* parse
...
+      continue;
+    }
+
+    auto parse_result = option_parser.value()(parser, &index);
+    if (!parse_result) {
+      return make_unexpected(parse_result.error());
+    }
+    if (!parse_result.value()) {
+      break;
+    }
```
BorysTheDev, line 267: I think you don't need it

BorysTheDev (approved):

src/server/search/search_family.cc
BagritsevichStepan: Why?
For `ParseSchema`, we return false. It is the last option that should be parsed.

BorysTheDev: you already have a check if (!parse_result) so parse_result.value() is always true

BagritsevichStepan: parse_result is `io::Error<bool>`, so when I prove `if (!parse_result)`, I prove that no error occurred. When I check `if (!parse_result.value())`, I prove that the returned result is false.

PR #4485, opened 2025-01-20
feat: add support for noreply and great options

romange:
noreply option is applicable for memcached.
great option ensure the tcp connection is indeed established before proceeding with loadtest.

### DIFF
--- a/src/server/dfly_bench.cc
+++ b/src/server/dfly_bench.cc
@@ -49,8 +49,8 @@ ABSL_FLAG(string, command, "",
 ABSL_FLAG(string, P, "", "protocol can be empty (for RESP) or memcache_text");
 ABSL_FLAG(bool, tcp_nodelay, false, "If true, set nodelay option on tcp socket");
 ABSL_FLAG(bool, noreply, false, "If true, does not wait for replies. Relevant only for memcached.");
-ABSL_FLAG(bool, great, true,
-          "If true, sends a greating command on each connection, "
+ABSL_FLAG(bool, greet, true,
+          "If true, sends a greeting command on each connection, "
           "to make sure the connection succeeded");
 
 using namespace std;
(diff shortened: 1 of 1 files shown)


chakaz (dismissed):

src/server/dfly_bench.cc
```diff
@@ -48,6 +48,10 @@ ABSL_FLAG(string, command, "",
           "__data__ for values, __score__ for doubles");
 ABSL_FLAG(string, P, "", "protocol can be empty (for RESP) or memcache_text");
 ABSL_FLAG(bool, tcp_nodelay, false, "If true, set nodelay option on tcp socket");
+ABSL_FLAG(bool, noreply, false, "If true, does not wait for replies. Relevant only for memcached.");
+ABSL_FLAG(bool, great, true,
```
chakaz, line 52:
```suggestion
ABSL_FLAG(bool, greet, true,
```

src/server/dfly_bench.cc
```diff
@@ -48,6 +48,10 @@ ABSL_FLAG(string, command, "",
...
+          "If true, sends a greating command on each connection, "
```
chakaz, line 53:
```suggestion
          "If true, sends a greeting command on each connection, "
```

src/server/dfly_bench.cc
```diff
@@ -384,18 +398,19 @@ void Driver::Connect(unsigned index, const tcp::endpoint& ep) {
...
-  uint8_t buf[128];
-  auto res_sz = socket_->Recv(io::MutableBytes(buf));
-  CHECK(res_sz) << res_sz.error().message();
-  string_view resp = io::View(io::Bytes(buf, *res_sz));
-  CHECK(absl::EndsWith(resp, "\r\n")) << resp;
+  if (absl::GetFlag(FLAGS_great)) {
+    // TCP Connect does not ensure that the connection was indeed accepted by the server.
+    // if server backlog is too short the connection will get stuck in the accept queue.
+    // Therefore, we send a ping command to ensure that every connection got connected.
+    ec = socket_->Write(io::Buffer("ping\r\n"));
```
chakaz, line 405: You could use `HELLO` instead of `PING`, but I don't know if it really matters

romange: it's easier to read PING's response than that of HELLO. I do not use a parser here

chakaz (approved):

PR #4487, opened 2025-01-20
feat: add 'testing_time' limit option to dfly_bench

romange:

### DIFF
--- a/src/server/dfly_bench.cc
+++ b/src/server/dfly_bench.cc
@@ -779,7 +779,7 @@ int main(int argc, char* argv[]) {
 
   uint32_t thread_key_step = 0;
   const uint32_t qps = GetFlag(FLAGS_qps);
-  const int64_t interval = qps ? 1000'000'000LL / qps : 0;
+  const int64_t interval = qps ? 1'000'000'000LL / qps : 0;
   uint64_t num_reqs = GetFlag(FLAGS_n);
   uint64_t total_conn_num = GetFlag(FLAGS_c) * pp->size();
   uint64_t total_requests = num_reqs * total_conn_num;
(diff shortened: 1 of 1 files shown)


src/server/dfly_bench.cc
```diff
@@ -767,10 +779,11 @@ int main(int argc, char* argv[]) {
 
   uint32_t thread_key_step = 0;
   const uint32_t qps = GetFlag(FLAGS_qps);
-  const int64_t interval = qps ? 1000000000LL / qps : 0;
+  const int64_t interval = qps ? 1000'000'000LL / qps : 0;
```
chakaz, line 782:
```suggestion
  const int64_t interval = qps ? 1'000'000'000LL / qps : 0;
```
Since you already touched it :rofl:

src/server/dfly_bench.cc
```diff
@@ -767,10 +779,11 @@ int main(int argc, char* argv[]) {
...
   uint64_t num_reqs = GetFlag(FLAGS_n);
   uint64_t total_conn_num = GetFlag(FLAGS_c) * pp->size();
   uint64_t total_requests = num_reqs * total_conn_num;
+  uint32_t time_limit = GetFlag(FLAGS_test_time);
```
chakaz, line 786: Since you ignore `num_reqs` in case `time_limit` is specified, I'd `CHECK()` here that at most one of them is set, but not both

chakaz (dismissed):

src/server/dfly_bench.cc
romange: i thought about this but it hurts usability, I am afraid, because n is set to 1000 by default and if we set time_limit to something we would need to explicitly reset `n`. similarly, if we set `n` to 0 by default, then we will have to provide any of them explicitly.

chakaz: makes sense

chakaz (approved):

PR #4500, opened 2025-01-24
fix: several bugs

### LINKED ISSUE #4466: Special Redis commands can cause the Dragonfly service to crash
## Summary
Authenticated users can construct specific commands that cause the dragonfly service to crash.

## Details
Dragonfly version: df-v1.26.1
Payload:
`SCAN 9223372036854775808 MATCH *11*`
Executing the above payload will cause the dragonfly service to crash immediately.
Poof:

Image

Image

## Reporter
The problem was reported by ankki-zsyang, Shenzhen Ankki Technologies Co.Ltd.

romange:
1. Fixes #4466 (we did not check validity of the scan cursor) We fix it by quietly returning "end of scan" result, i.e. no error are returned.
2. Add 34b1048274c8e50a0cc587a3ed9c383a82bb78c5 sha to the lua_undeclared_keys_shas list.
3. Pull helio that fixes bugs with WaitQueue and IdleTasks: https://github.com/romange/helio/commit/2ab4412b78afbf7d3102b472fb0e6f70b194a535

### DIFF
--- a/src/server/script_mgr.cc
+++ b/src/server/script_mgr.cc
@@ -43,6 +43,7 @@ ABSL_FLAG(
     std::vector<std::string>({
         "351130589c64523cb98978dc32c64173a31244f3",  // Sidekiq, see #2442
         "6ae15ef4678593dc61f991c9953722d67d822776",  // Sidekiq, see #2442
+        "34b1048274c8e50a0cc587a3ed9c383a82bb78c5"   // Sidekiq
     }),
     "Comma-separated list of Lua script SHAs which are allowed to access undeclared keys. SHAs are "
     "only looked at when loading the script, and new values do not affect already-loaded script.");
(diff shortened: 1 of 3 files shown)


src/server/script_mgr.cc
```diff
@@ -43,6 +43,7 @@ ABSL_FLAG(
     std::vector<std::string>({
         "351130589c64523cb98978dc32c64173a31244f3",  // Sidekiq, see #2442
         "6ae15ef4678593dc61f991c9953722d67d822776",  // Sidekiq, see #2442
+        "34b1048274c8e50a0cc587a3ed9c383a82bb78c5"   // Sidekiq
```
kostasrim, line 46: is this sidekiq internal ? new scheduler or?

kostasrim:
@romange There are two `alpine builds` missing from the `required checks`. There are `4` here but actually we have `6`. See https://github.com/dragonflydb/dragonfly/actions/runs/12946405416/job/36110868326 

We need to modify the filter in the settings of the project (I don't have access to this)

kostasrim:
> @romange There are two `alpine builds` missing from the `required checks`. There are `4` here but actually we have `6`. See https://github.com/dragonflydb/dragonfly/actions/runs/12946405416/job/36110868326
> 
> We need to modify the filter in the settings of the project (I don't have access to this)

ignore me, the UI has changed and once a test succeeds it does not show up. Everything is fine.

src/server/script_mgr.cc
romange: it's sidekiq pro script

BorysTheDev (approved):

PR #4503, opened 2025-01-25
chore: new docker build pipeline

romange:
Our previous weekly pipeline used qemu, was very slow and over-complicated.
This one uses matrix-build with proper parallelisation and the latest arm64 github runners.

now it takes less than 30 minutes to build everything.
lets make it daily.

### DIFF
--- a/.github/workflows/docker-dev-release.yml
+++ b/.github/workflows/docker-dev-release.yml
@@ -1,0 +1,131 @@
+name: Development Docker Build
+
+on:
+  schedule:
+    - cron: '15 0 * * *'
+  workflow_dispatch:
+
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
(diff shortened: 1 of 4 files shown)


romange:
Successful run: https://github.com/dragonflydb/dragonfly/actions/runs/12966893501/job/36168158524

.github/workflows/docker-dev-release.yml
```diff
@@ -0,0 +1,131 @@
+name: Development Docker Build
+
+on:
+  schedule:
+    - cron: '15 0 * * *'
```
kostasrim, line 5: Why do we start all our cron jobs at the same time ? :smile: (I could be wrong but I think others have the exact same timestamp)

kostasrim (approved):
Run it one time before you merge :)

.github/workflows/docker-dev-release.yml
romange: they do not actually.

PR #4504, opened 2025-01-25
fix: LMPOP implementation was added

vyavdoshenko:
fix: LMPOP implementation was added


src/server/list_family.cc
```diff
@@ -1173,8 +1173,138 @@ void BPopGeneric(ListDir dir, CmdArgList args, Transaction* tx, SinkReplyBuilder
...
+  ShardArgs keys = t->GetShardArgs(shard->shard_id());
+  if (keys.Empty()) {
+    return {};
+  }
+
+  auto& db_slice = t->GetDbSlice(shard->shard_id());
+  std::vector<std::string_view> result;
+
+  for (string_view key : keys) {
+    auto it = db_slice.FindReadOnly(t->GetDbContext(), key, OBJ_LIST);
```
kostasrim, line 1233: We should include the `key` if it's not `list` and if it's first in the command argument list we should return an error. 

Example: 

```
lpush l1 e1
hset foo k1 v1
lmpop 2 foo l1 left
>│(error) WRONGTYPE Operation against a key holding the wrong kind of value
```

src/server/list_family.cc
```diff
@@ -1173,8 +1173,138 @@ void BPopGeneric(ListDir dir, CmdArgList args, Transaction* tx, SinkReplyBuilder
...
+    return;
+  }
+  auto* response_builder = static_cast<RedisReplyBuilder*>(cmd_cntx.rb);
+
+  absl::flat_hash_set<std::string_view> found_keys;
+  found_keys.reserve(std::min(shard_set->size(), num_keys));
+  auto cb = [&](Transaction* t, EngineShard* shard) {
+    std::vector<std::string_view> result = GetNonEmptyKeysFound(shard, t);
+    if (!result.empty()) {
+      found_keys.insert(result.begin(), result.end());
```
kostasrim, line 1260: unfortunately found_keys.insert below is not thread safe. The pattern we use is that we store a vector of results and each shard writes its own result in that vector. Then you can squash the results here after hop

src/server/list_family.cc
```diff
@@ -1173,8 +1173,138 @@ void BPopGeneric(ListDir dir, CmdArgList args, Transaction* tx, SinkReplyBuilder
...
+
+  return result;
+}
+
 }  // namespace
 
+void ListFamily::LMPop(CmdArgList args, const CommandContext& cmd_cntx) {
+  uint32 num_keys;
+  bool is_left;
+  int pop_count;
```
kostasrim, line 1249: you can group them under one struct :)

src/server/list_family.cc
```diff
@@ -1173,8 +1173,138 @@ void BPopGeneric(ListDir dir, CmdArgList args, Transaction* tx, SinkReplyBuilder
...
+      found_keys.insert(result.begin(), result.end());
+    }
+
+    return OpStatus::OK;
+  };
+
+  cmd_cntx.tx->Execute(std::move(cb), false /* possibly another hop */);
+
+  // Find the first key from command arguments that exists in any shard
+  std::optional<std::string_view> key_to_pop = std::nullopt;
```
kostasrim, line 1269: no need for optiona, string_view can be empty and as such it will denote a non found item

src/server/list_family.cc
```diff
@@ -1173,8 +1173,138 @@ void BPopGeneric(ListDir dir, CmdArgList args, Transaction* tx, SinkReplyBuilder
...
+    response_builder->SendNull();
+    return;
+  }
+
+  std::optional<ShardId> key_shard = Shard(*key_to_pop, shard_set->size());
+  OpResult<StringVec> result;
+
+  auto cb_pop = [is_left, pop_count, key_shard, &result, key = *key_to_pop](Transaction* t,
+                                                                            EngineShard* shard) {
+    if (!key_shard.has_value() || *key_shard == shard->shard_id()) {
```
kostasrim, line 1289: At this point `key_shard` should have a value, if it doesn't why do we dispatch another hop?

src/server/list_family.cc
```diff
@@ -1173,8 +1173,138 @@ void BPopGeneric(ListDir dir, CmdArgList args, Transaction* tx, SinkReplyBuilder
...
+
+  return true;
+}
+
+// Returns all non-empty keys found in the list of shard arguments
+std::vector<std::string_view> GetNonEmptyKeysFound(EngineShard* shard, Transaction* t) {
+  ShardArgs keys = t->GetShardArgs(shard->shard_id());
+  if (keys.Empty()) {
+    return {};
+  }
```
kostasrim, line 1227: I think keys won't be ever empty :)

kostasrim (commented):
Hi @vyavdoshenko, thank you for your contribution!! A few preliminary comments

src/server/list_family.cc
vyavdoshenko: Fixed.

src/server/list_family.cc
vyavdoshenko: Fixed.

src/server/list_family.cc
vyavdoshenko: Fixed.

src/server/list_family.cc
vyavdoshenko: Fixed.

src/server/list_family.cc
vyavdoshenko: Fixed.

src/server/list_family.cc
vyavdoshenko: Fixed.

src/server/list_family.cc
```diff
@@ -1173,8 +1174,155 @@ void BPopGeneric(ListDir dir, CmdArgList args, Transaction* tx, SinkReplyBuilder
...
 
+struct LMPopParams {
+  uint32_t num_keys;
+  bool is_left;
+  int pop_count;
+};
+
+// Validates the LMPop command arguments and extracts the values to the output params.
+// If the arguments are invalid sends the appropriate error to builder and returns false.
+bool ValidateLMPopCommand(CmdArgList args, LMPopParams& params, SinkReplyBuilder* builder) {
```
romange, line 1185: our style guide prohibits passing mutable references. please make it `params` a pointer.

src/server/list_family.cc
```diff
@@ -1173,8 +1174,155 @@ void BPopGeneric(ListDir dir, CmdArgList args, Transaction* tx, SinkReplyBuilder
   return rb->SendNullArray();
 }
 
+struct LMPopParams {
+  uint32_t num_keys;
+  bool is_left;
```
romange, line 1179: we introduced the `ListDir` enum, can you please use it?

src/server/list_family.cc
```diff
@@ -1173,8 +1174,155 @@ void BPopGeneric(ListDir dir, CmdArgList args, Transaction* tx, SinkReplyBuilder
...
+  if (!parser.Finalize()) {
+    builder->SendError(parser.Error()->MakeReply());
+    return false;
+  }
+
+  return true;
+}
+
+// Returns all non-empty keys found in the list of shard arguments
+std::vector<std::pair<std::string_view, bool>> GetNonEmptyKeysFound(EngineShard* shard,
```
romange, line 1228: curious, why did you decide to return all the found keys if you need only the first one valid?

src/facade/reply_builder.cc
```diff
@@ -453,4 +453,15 @@ void RedisReplyBuilder::SendEmptyArray() {
   StartArray(0);
 }
 
+void RedisReplyBuilder::SendKeyValueArray(std::string_view key, const StringVec& values) {
```
romange, line 456: why this code is correct, I do not thing it should be added to ReplyBuilder class because it's specific to LMPOP.. You can just use this code inside LMPop or introduce a helper function in list_family.cc

romange (commented):
looks good, some more comments.

src/server/list_family.cc
vyavdoshenko: Fixed.

src/server/list_family.cc
vyavdoshenko: Fixed.

src/facade/reply_builder.cc
vyavdoshenko: Fixed.

src/server/list_family.cc
```diff
@@ -1173,8 +1174,157 @@ void BPopGeneric(ListDir dir, CmdArgList args, Transaction* tx, SinkReplyBuilder
...
+                                                                          Transaction* t) {
+  ShardArgs keys = t->GetShardArgs(shard->shard_id());
+  DCHECK(!keys.Empty());
+
+  auto& db_slice = t->GetDbSlice(shard->shard_id());
+  std::optional<std::pair<std::string_view, bool>> result;
+
+  for (string_view key : keys) {
+    auto res = db_slice.FindReadOnly(t->GetDbContext(), key, OBJ_LIST);
+    if (!res) {
```
romange, line 1238: nit: if you reverse the condition, and the rearrange the code, it will look nice, imho

src/server/list_family.cc
vyavdoshenko: Fixed.
Because it was my wrong assumption, now I understand that's enough to save the first found element per shard because it will be processed first, that's why all further keys in this particular shard can be skipped.

src/server/list_family.cc
```diff
@@ -1173,8 +1174,157 @@ void BPopGeneric(ListDir dir, CmdArgList args, Transaction* tx, SinkReplyBuilder
...
+      if (res.status() == OpStatus::WRONG_TYPE) {
+        result = {key, false};
```
romange, line 1240: i wonder if this is correct, in other words if we pop from "a", "b",  and "a" is not a list, but "b" is list whether we stop and return the error

src/server/list_family.cc
```diff
@@ -1173,8 +1174,157 @@ void BPopGeneric(ListDir dir, CmdArgList args, Transaction* tx, SinkReplyBuilder
...
+    return;
+  }
+  auto* response_builder = static_cast<RedisReplyBuilder*>(cmd_cntx.rb);
+
+  // Create a vector to store results from each shard
+  std::vector<std::optional<std::pair<std::string_view, bool>>> found_keys_per_shard(
+      shard_set->size());
+
+  auto cb = [&](Transaction* t, EngineShard* shard) {
+    // Each shard writes results to its own vector
```
romange, line 1267: `to its own vector` - it's not vector anymore.

romange, line 1267:
```suggestion
    // Each shard writes results to its own space
```

src/server/list_family.cc
```diff
@@ -1173,8 +1174,157 @@ void BPopGeneric(ListDir dir, CmdArgList args, Transaction* tx, SinkReplyBuilder
...
+    found_keys_per_shard[shard->shard_id()] = GetFirstNonEmptyKeyFound(shard, t);
+    return OpStatus::OK;
+  };
+
+  cmd_cntx.tx->Execute(std::move(cb), false /* possibly another hop */);
```
romange, line 1272:
```suggestion
  cmd_cntx.tx->Execute(std::move(cb), false /* followed by another hop */);
```

src/server/list_family.cc
vyavdoshenko: I guess the answer can be found here: ListFamilyTest::LMPopWrongType.
Long story short: we get a list of keys regardless of the type we can process all of them one by one.
If we get the first non-empty key with the wrong type, the code returns the error.
If we get the first non-empty key with the correct type, the code processes this key further to return the result.

src/server/list_family.cc
vyavdoshenko: Fixed.

src/server/list_family.cc
vyavdoshenko: Fixed.

src/server/list_family.cc
```diff
@@ -1173,8 +1174,161 @@ void BPopGeneric(ListDir dir, CmdArgList args, Transaction* tx, SinkReplyBuilder
...
+
+  cmd_cntx.tx->Execute(std::move(cb), false /* followed by another hop */);
+
+  // Find the first existing key from command arguments
+  std::string_view key_to_pop;
+  bool found_wrong_type = false;
+  ArgRange arg_keys(args.subspan(1, params.num_keys));
+
+  // Check each key from arguments
+  for (std::string_view key : arg_keys) {
```
romange, line 1283: what is the complexity of lines 1283-1294 on your opinion?

romange: It is possible to make it linear O(N) + O(M). I will describe briefly how. You can either leave it as is but then please add a `TODO: to fix the complexity`, or alternatively try to fix it.

The approach is similar to `StringFamily::MGet` which needs to send values in the same order as the arguments it gets.

So basically you iterate over each shard via `found_keys_per_shard`. if a shard has a key, you need to check if it's a key with smallest index. You can use ` ShardArgs shard_args = cmnd_cntx.tx->GetShardArgs(sid);` to get all the shard arguments for that shard.  Then by iterating these arguments with  ` for (auto it = shard_args.begin(); it != shard_args.end();` ` you can check their index by `it.index()`. 
of course only the matched argument with that in found_keys_per_shard should be considered.

hope it was clear enough.

vyavdoshenko: Please take a look.
The current implementation is similar to StringFamily::MGet.

src/server/list_family.cc
```diff
@@ -1173,8 +1174,165 @@ void BPopGeneric(ListDir dir, CmdArgList args, Transaction* tx, SinkReplyBuilder
...
+
+  for (string_view key : keys) {
+    auto res = db_slice.FindReadOnly(t->GetDbContext(), key, OBJ_LIST);
+    if (res) {
+      result = {key, true};
+      break;
+    }
+
+    // If the key is not found, check if it's a wrong type error
+    if (!res && res.status() == OpStatus::WRONG_TYPE) {
```
romange, line 1247: `!res` is always true

src/server/list_family.cc
```diff
@@ -1173,8 +1174,165 @@ void BPopGeneric(ListDir dir, CmdArgList args, Transaction* tx, SinkReplyBuilder
...
+        min_index = it.index();
+        key_to_pop = found_key;
+        found_wrong_type = !is_valid_type;
+        break;
+      }
+    }
+  }
+
+  // Handle errors and empty cases first
+  if (key_to_pop.empty() || found_wrong_type) {
```
romange, line 1301: there is a subtle bug here: we allow using an empty key in the datastore.

src/server/list_family.cc
```diff
@@ -1173,8 +1174,165 @@ void BPopGeneric(ListDir dir, CmdArgList args, Transaction* tx, SinkReplyBuilder
...
 
+void ListFamily::LMPop(CmdArgList args, const CommandContext& cmd_cntx) {
+  LMPopParams params;
+  if (!ValidateLMPopCommand(args, &params, cmd_cntx.rb)) {
+    return;
+  }
+  auto* response_builder = static_cast<RedisReplyBuilder*>(cmd_cntx.rb);
+
+  // Create a vector to store first found key for each shard
+  std::vector<std::optional<std::pair<std::string_view, bool>>> found_keys_per_shard(
```
romange, line 1266: fyi, you do not have to write `std::` everywhere, unless you prefer to.

src/server/list_family.cc
```diff
@@ -1173,8 +1174,165 @@ void BPopGeneric(ListDir dir, CmdArgList args, Transaction* tx, SinkReplyBuilder
...
 
+struct LMPopParams {
+  uint32_t num_keys;
+  ListDir dir;
+  int pop_count;
+};
+
+// Validates the LMPop command arguments and extracts the values to the output params.
+// If the arguments are invalid sends the appropriate error to builder and returns false.
+bool ValidateLMPopCommand(CmdArgList args, LMPopParams* params, SinkReplyBuilder* builder) {
```
romange, line 1185: nit: I think `ParseLMPop` is a better name (parse also implies validating).

Also, we have a type  `facade::ErrorReply` that can be returned instead of bool. 
it may hold all types of errors you return here (we had an evolution of types for historical reasons).
This way you do not need to pass `SinkReplyBuilder* builder` to this function and instead check if `ErrorReply::status == kOk` in the caller to see if parsing succeeded.

romange (commented):
We are almost there. I had a few more comments. @adiholden  ptal if you have time.

src/server/list_family.cc
vyavdoshenko: Fixed.

src/server/list_family.cc
vyavdoshenko: Fixed.

src/server/list_family.cc
```diff
@@ -1173,8 +1174,159 @@ void BPopGeneric(ListDir dir, CmdArgList args, Transaction* tx, SinkReplyBuilder
...
 }  // namespace
 
+void ListFamily::LMPop(CmdArgList args, const CommandContext& cmd_cntx) {
+  auto* response_builder = static_cast<RedisReplyBuilder*>(cmd_cntx.rb);
+
+  LMPopParams params;
+  facade::ErrorReply parse_result = ParseLMPop(args, &params);
+
+  if (parse_result.status != OpStatus::OK) {
+    response_builder->SendError(parse_result.ToSv());
```
romange, line 1255: it's better to call `SendError(parse_result);`  - it's more correct

romange (commented):
LGTM!

src/server/list_family.cc
vyavdoshenko: Fixed.

adiholden (approved):

PR #4506, opened 2025-01-26
fix: rpm package setup

### LINKED ISSUE #4505: Problems with dragonfly service on Oracle Linux
**Describe the bug**

1. After installing Dragonfly from dragonfly.x86_64.rpm on OracleLinux, the service does not start. The reason for this issue is the dragonfly.service lists /usr/bin/dragonfly binary, but the correct path is /usr/local/bin/dragonfly.

2. After fixing (1) and starting the service, any attempt to stop the service hangs indefinitely.  I believe this is due to the fact that the process does not react to SIGTERM, (only SIGKILL). Adding TimeoutStopSec=1 to the dragonfly.service resolves the issue, because systemctl uses SIGKILL after 1 sec. 

3. When I try to reboot the host, the problem (2) prevents the host from rebooting. I mean, at all. I have to power cycle the host. Adding TimeoutStopSec=1 to the dragonfly.service resolves this issue as well.

**To Reproduce**

dnf install dragonfly.x86_64.rpm
systemctl enable dragonfly
systemctl start dragonfly
systemctl stop dragonfly
reboot

**Environment**

Oracle Linux Server 9.5
Dragonfly Version: [1.26.1]

romange:
Also, fix the deadlock problem on shutdown on Oracle Linux 5.15 Fixes #4505

### DIFF
--- a/src/server/rdb_test.cc
+++ b/src/server/rdb_test.cc
@@ -458,8 +458,8 @@ TEST_F(RdbTest, JsonTest) {
 class HllRdbTest : public RdbTest, public testing::WithParamInterface<string> {};
 
 TEST_P(HllRdbTest, Hll) {
-  LOG(INFO) << " max memory: " << max_memory_limit
-            << " used_mem_current: " << used_mem_current.load();
+  LOG(ERROR) << " max memory: " << max_memory_limit
+             << " used_mem_current: " << used_mem_current.load();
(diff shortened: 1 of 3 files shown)


chakaz (dismissed):
:shrug:

chakaz (approved):

PR #4513, opened 2025-01-27
fix: crash for delconsumer during stream reading

BorysTheDev:
fixes: https://github.com/dragonflydb/dataplane-private/issues/27

### DIFF
--- a/src/server/stream_family.cc
+++ b/src/server/stream_family.cc
@@ -2335,10 +2335,11 @@ void XReadBlock(ReadOpts* opts, Transaction* tx, SinkReplyBuilder* builder,
       if (sitem.group) {
         range_opts.consumer =
             FindOrAddConsumer(opts->consumer_name, sitem.group, GetCurrentTimeMs());
-      }
+        sitem.consumer = range_opts.consumer;
+        if (!sitem.consumer) {
+          return OpStatus::OUT_OF_MEMORY;
+        }
 
-      range_opts.noack = opts->noack;
-      if (sitem.consumer) {
         if (sitem.consumer->pel->numnodes == 0) {
           LOG(DFATAL) << "Internal error when accessing consumer data, seen_time "
                       << sitem.consumer->seen_time;
@@ -2346,6 +2347,9 @@ void XReadBlock(ReadOpts* opts, Transaction* tx, SinkReplyBuilder* builder,
           return OpStatus::OK;
         }
       }
+
+      range_opts.noack = opts->noack;
+
       result = OpRange(t->GetOpArgs(shard), *wake_key, range_opts);
       key = *wake_key;
     }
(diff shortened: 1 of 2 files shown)


src/server/stream_family.cc
```diff
@@ -2335,17 +2335,21 @@ void XReadBlock(ReadOpts* opts, Transaction* tx, SinkReplyBuilder* builder,
...
-      if (sitem.consumer) {
         if (sitem.consumer->pel->numnodes == 0) {
           LOG(DFATAL) << "Internal error when accessing consumer data, seen_time "
                       << sitem.consumer->seen_time;
           result = OpStatus::CANCELLED;
           return OpStatus::OK;
         }
       }
+
+      range_opts.noack = opts->noack;
```
romange, line 2351: what's this? does the test cover this fix as well?

BorysTheDev: it was 2340 line I didn't changed it.

romange (approved):

src/server/stream_family.cc
```diff
@@ -2335,17 +2335,21 @@ void XReadBlock(ReadOpts* opts, Transaction* tx, SinkReplyBuilder* builder,
       if (sitem.group) {
         range_opts.consumer =
             FindOrAddConsumer(opts->consumer_name, sitem.group, GetCurrentTimeMs());
-      }
+        sitem.consumer = range_opts.consumer;
+        if (!sitem.consumer) {
+          return OpStatus::OUT_OF_MEMORY;
```
kostasrim, line 2340: Slightly misleading in this context, because `FindOrAddConsumer` did not fail because we are out of memory but because the consumer group (`alice` in the test) is deleted and the pointer is `nullptr`

romange: @BorysTheDev  the first comment is worth responding.
@kostasrim  I believe the fix to the crash is this line:
`sitem.consumer = range_opts.consumer;`

BorysTheDev: @romange You are absolutely right `sitem.consumer = range_opts.consumer; ` is a fix, because sitem.consumer contains already invalidated pointer

I've added ```
if (!sitem.consumer) {
          return OpStatus::OUT_OF_MEMORY;
```
for the case if the memory for the consumer can not be allocated, but it isn't a fix of the crash

kostasrim: Well, for (1) and from my understanding `FindOrAddConsumer` shall return null if `cg` is empty or if we try to insert the consumer into the rax data structure and we fail. Not sure if the second can fail under out of memory scenario but either way it would worth a comment. 

For (2), yes I got it and checked it myself. I thought the fix was the early return but that didn't make sense and that's why I left the comment above yesterday.

kostasrim: @BorysTheDev but we don't return `NULL` within `FindOrAddConsumer` only when `OUT_OF_MEMORY` right ?
From my understanding, `cg` can be `null`, if the consumer does not exist then it must be interted with `raxInsert` which can return NULL as well (not sure if this is only for OUT OF MEMORY though)

The rest make sense :)

kostasrim: yep we got checks for `cg` + raxInsert will return null only if it fails to insert. all good :+1:

Issue #4517, opened 2025-01-28
Evict subscribed sharded pub/sub on cluster migrations


kostasrim:
Upon slot migration, the node that owns the slots for channels in sharded pub/sub commands continues to process normally incoming pub/sub commands. Upon completion, the node evicts the connected clients to that given slot, returns a MOVED error for the client to redirect the connection.

There is no state associated with pub/sub that needs to be transferred during slot migration. After the migration completes, clients need to redirect their pub/sub commands to the newly migrated node.

PR #4518, opened 2025-01-28
chore: implement sharded pub/sub

### LINKED ISSUE #3001: Add support for SPUBLISH
Rueidis tests use SPUBLISH. We should implement this.

kostasrim:
Resolves #3001 by adding the basic functionality for sharded pub/sub for Dragonfly cluster.

* add sharded pub/sub
* add slot check logic for pub sub messages
* add test

What's left for sharded pub sub is described in  #4517  but the gist is that we need to evict subscribed connections after a slot migration. This will be added in a follow up PR.


src/server/main_service.cc
```diff
@@ -919,18 +919,33 @@ void Service::Shutdown() {
...
 }
 
+OpResult<KeyIndex> DetermineClusterKeys(const CommandId* cid, CmdArgList args) {
+  if (!cid->IsShardedPSub()) {
+    return DetermineKeys(cid, args);
+  }
+
+  // Sharded pub sub
+  // Command form: SPUBLISH shardchannel message
+  if (cid->name() == "SPUBLISH") {
```
kostasrim, line 929: We need to bypass `DetermineKeys` only for this specific case.

.github/workflows/ci.yml
```diff
@@ -172,7 +172,10 @@ jobs:
           cd ${GITHUB_WORKSPACE}/build
           echo Run ctest -V -L DFLY
           GLOG_alsologtostderr=1 GLOG_vmodule=rdb_load=1,rdb_save=1,snapshot=1 \
-          FLAGS_fiber_safety_margin=4096 FLAGS_list_experimental_v2=true timeout 20m ctest -V -L DFLY
+          FLAGS_fiber_safety_margin=4096 FLAGS_list_experimental_v2=true timeout 20m ctest -V -L DFLY -E allocation_tracker
+
+          # Run allocation tracker test separately. It generates a TON of logs
```
kostasrim, line 177: Allocation tracker test generates a gazillion number of logs. I silenced it that way by excluding the test from running with alsologtostderr flag.

kostasrim: Example https://github.com/dragonflydb/dragonfly/actions/runs/13011160524/job/36288993157 :scream:

src/facade/command_id.h
```diff
@@ -118,7 +123,8 @@ class CommandId {
   bool restricted_ = false;
 
   bool is_pub_sub_ = false;
-  bool is_p_sub_ = false;
+  bool is_sharded_pub_sub_ = false;
+  bool is_p_pub_sub_ = false;
```
BorysTheDev, line 127: I think we need enum

src/facade/facade.cc
```diff
@@ -137,7 +137,9 @@ CommandId::CommandId(const char* name, uint32_t mask, int8_t arity, int8_t first
   if (name_ == "PUBLISH" || name_ == "SUBSCRIBE" || name_ == "UNSUBSCRIBE") {
     is_pub_sub_ = true;
   } else if (name_ == "PSUBSCRIBE" || name_ == "PUNSUBSCRIBE") {
-    is_p_sub_ = true;
+    is_p_pub_sub_ = true;
+  } else if (name_ == "SPUBLISH" || name_ == "SSUBSCRIBE" || name_ == "SUNSUBSCRIBE") {
+    is_sharded_pub_sub_ = true;
```
BorysTheDev, line 142: As I remember we can rename commands. is it applicable to this case?

src/server/main_service.cc
BorysTheDev, line 929: Can you move this code inside DetermineKeys. Because the name isn\t really correct

.github/workflows/ci.yml
```diff
@@ -172,7 +172,10 @@ jobs:
           cd ${GITHUB_WORKSPACE}/build
           echo Run ctest -V -L DFLY
           GLOG_alsologtostderr=1 GLOG_vmodule=rdb_load=1,rdb_save=1,snapshot=1 \
-          FLAGS_fiber_safety_margin=4096 FLAGS_list_experimental_v2=true timeout 20m ctest -V -L DFLY
+          FLAGS_fiber_safety_margin=4096 FLAGS_list_experimental_v2=true timeout 20m ctest -V -L DFLY -E allocation_tracker_test
+
+          # Run allocation tracker test separately. It generates a TON of logs
+          FLAGS_fiber_safety_margin=4096 FLAGS_force_epoll=true GLOG_vmodule=rdb_load=1,rdb_save=1,snapshot=1 timeout 5m ./allocation_tracker_test
```
BorysTheDev, line 178: why do you need GLOG_vmodule=rdb_load=1,rdb_save=1,snapshot=1

kostasrim: +1 we don't!

src/facade/command_id.h
kostasrim: Can you explain how would an enum help here? A variation of these booleans can be true/false depending on the command so not sure how an enum would help here.

IMO these should be a `c-style` bitfield but CommandRegistry is kinda fixed size so the memory footprint of this is insignificant anyways

src/server/main_service.cc
kostasrim: The thing is we need to bypass `DetermineKeys` behaviour only for this particular case so we can't really move it there. 

I agree the naming is not great. Any ideas of a better name ? Otherwise I can think of something :shrug:

src/facade/facade.cc
kostasrim: Good question! Not really, it does not affect us. 

CommandRegistry has a `hash map` for commands and a `hash map` for renamed commands. When we do a lookup, we do so via `Service::FindCmd` which looks on both of those hash maps.

Furthermore, when we register a command in the registry via `>>` operator we handle renamed commands. As CommandId is constructed before we emplace it in the registry this part shall have no issues with renamed commands

kostasrim: However this might have https://github.com/dragonflydb/dragonfly/pull/4518/files#r1932116546!

kostasrim: I will fix it :)

src/facade/command_id.h
BorysTheDev: As I understand we can't have is_p_pub_sub_ and is_sharded_pub_sub_ simultaneously maybe is_pub_sub_ too. So we can have enum PubSubT {NONE, P_PUB_SUB, SHARDED_PUB_SUB, PUB_SUB}

BorysTheDev (approved):

PR #4519, opened 2025-01-28
chore: update release build for arm

BorysTheDev:
use arm64 platform to build arm64 instead of QEMU

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,60 +24,25 @@ jobs:
           draft: true
           token: ${{ secrets.GITHUB_TOKEN }}
 
-  build-qemu:
-    runs-on: ubuntu-latest
-    name: Build aarch64 on ubuntu20.04
+  build-arm:
+    runs-on: ubuntu-24.04-arm
+    name: Build arm64 on ubuntu-24.04-arm
     needs: create-release
+    container:
+      image: ghcr.io/romange/ubuntu-dev:20
+      options: --security-opt seccomp=unconfined --sysctl "net.ipv6.conf.all.disable_ipv6=0"
     steps:
     - uses: actions/checkout@v4
       with:
         submodules: true
-    - uses: uraimo/run-on-arch-action@v2
-      name: Run commands
-      id: runcmd
-      with:
-        arch: aarch64
-        distro: ubuntu20.04
-        githubToken: ${{ github.token }}
-        # Create an artifacts directory
-        setup: |
-          mkdir -p "${PWD}/artifacts"
-
-        # Mount the artifacts directory as /artifacts in the container
-        dockerRunArgs: |
-          --volume "${{ github.workspace }}:/src"
-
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -36,8 +36,6 @@ jobs:
       name: Run commands
       id: runcmd
       with:
-        arch: aarch64
```
romange, line 39: `uraimo/run-on-arch-action` is the action that runs a job on qemu. we do not need it now, so you should make it a regular `run` job, imho.

.github/workflows/release.yml
```diff
@@ -24,60 +24,29 @@ jobs:
...
+              zip libzstd-dev debhelper moreutils bison zlib1g-dev build-essential
+          sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-9 40  \
                               --slave /usr/bin/g++ g++ /usr/bin/g++-9
-        run: |
+    - name: Build artifacts
+      run: |
             # Work around https://github.com/actions/checkout/issues/766
-            git config --global --add safe.directory /src
-            cd /src
+            git config --global --add safe.directory "$GITHUB_WORKSPACE"
```
romange, line 45: I think this is a redundant line now since we do not use containers anymore

kostasrim:
For reference:

build time with QEMU -> roughly 2 hours+
native -> job finishes in 8 minutes 

:scream: :scream: 

Great work :smile:

.github/workflows/release.yml
romange: why did you resolve?

tools/packaging/generate_changelog.sh
```diff
@@ -32,7 +33,7 @@ fi
 prevtag=v0.2.0
 pkgname=`cat ${PACKGE_DIR}/control | grep '^Package: ' | sed 's/^Package: //'`
 git tag -l v* | sort -V | while read tag; do
-    (echo "$pkgname (${tag#v}) unstable; urgency=low"; git log --pretty=format:'  * %s' $prevtag..$tag; git log --pretty='format:%n%n -- %aN <%aE>  %aD%n%n' $tag^..$tag) | cat - ${CHANGE_LOG} | sponge ${CHANGE_LOG}
+    (echo "$pkgname (${tag#v}) unstable; urgency=low"; git log -3 --pretty=format:'  * %s'; git log -3 --pretty='format:%n%n -- %aN <%aE>  %aD%n%n') | cat - ${CHANGE_LOG} | sponge ${CHANGE_LOG}
```
romange, line 36: please revert

BorysTheDev: yes I will do it, as I wrote you I need these changes to be sure that everything works fine. I will revert both sh scripts

.github/workflows/release.yml
BorysTheDev: because we use container now

PR #4520, opened 2025-01-28
chore: add reflex matcher to the benchmarks

romange:
Add a test covering stringmatchlen.

### DIFF
--- a/src/core/dfly_core_test.cc
+++ b/src/core/dfly_core_test.cc
@@ -6,7 +6,13 @@
 #include "core/intent_lock.h"
 #include "core/tx_queue.h"
 
+extern "C" {
+#include "redis/util.h"
+}
+
 namespace dfly {
+
+using namespace std;
 
 class TxQueueTest : public ::testing::Test {
  protected:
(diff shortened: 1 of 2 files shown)


src/core/dfly_core_test.cc
```diff
@@ -65,4 +71,42 @@ TEST_F(IntentLockTest, Basic) {
   ASSERT_TRUE(lk_.Check(IntentLock::EXCLUSIVE));
 }
 
+class StringMatchTest : public ::testing::Test {
+ protected:
+  // wrapper around stringmatchlen with stringview arguments
+  int MatchLen(string_view pattern, string_view str, bool nocase) {
```
kostasrim, line 77: We have a similar function called Matches :smile: Also not sure why you added tests for `stringmatchlen` since it's heavily tested in Valkey + it's used in ACL by both Valkey and us.

kostasrim (approved):

src/core/dfly_core_test.cc
romange: I did not find any tests in valkey besides one very simple "fuzzy" test.

I covered it by tests so we could replace the implementation without breaking the behaviour.

kostasrim: sounds good!

PR #4523, opened 2025-01-29
chore: Add Lua force atomicity flag

### LINKED ISSUE #4522: add workaround for incorrect labeling of sidekiq-pro script
Following an incorrect request here: https://github.com/dragonflydb/dragonfly/issues/2442#issuecomment-2290725198

sidekiq pro 7.3.1 or later is broken on Dragonfly.

1. We should provide a hard-coded workaround to ignore `disable-atomicity` flag for this sha.
2. It should be patched into the 1.26 branch
3. We should communicate to @mperham about this, so he could also fix it there.

chakaz:
We accidentally instructed Sidekiq to add `disable-atomicity` to their script, despite them needing to run atomically.

This hack-ish PR adds a `--lua_force_atomicity_shas` flag to allow specifying which SHAs are forced to run in an atomic fashion, even if they are marked as non-atomic.

Fixes #4522

### DIFF
--- a/src/server/script_mgr.cc
+++ b/src/server/script_mgr.cc
@@ -48,6 +48,13 @@ ABSL_FLAG(
     "Comma-separated list of Lua script SHAs which are allowed to access undeclared keys. SHAs are "
     "only looked at when loading the script, and new values do not affect already-loaded script.");
 
+ABSL_FLAG(std::vector<std::string>, lua_force_atomicity_shas,
+          std::vector<std::string>({
+              "f8133be7f04abd9dfefa83c3b29a9d837cfbda86",  // Sidekiq, see #4522
+          }),
+          "Comma-separated list of Lua script SHAs which are forced to run in atomic mode, even if "
+          "the script specifies disable-atomicity.");
+
 namespace dfly {
 using namespace std;
 using namespace facade;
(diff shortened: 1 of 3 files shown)


src/server/script_mgr.cc
```diff
@@ -265,6 +272,13 @@ io::Result<string, GenericError> ScriptMgr::Insert(string_view body, Interpreter
     return params_opt.get_unexpected();
   auto params = params_opt->value_or(default_params_);
 
+  if (!params.atomic) {
+    auto force_atomic_shas = absl::GetFlag(FLAGS_lua_force_atomicity_shas);
+    if (find(force_atomic_shas.begin(), force_atomic_shas.end(), sha) != force_atomic_shas.end()) {
```
kostasrim, line 277: At some point I will introduce wrappers for the stl algorithms that we use most frequently because I really hate to pass the whole range into the algorithm (begin and end) when 99% it's the full range (and some subrange) Yes we don't want std::ranges but at least we can get a nice function of the form `find(cont, lambda) -> calls std::find(cont.begin(), cont.end(), lambda)`

adiholden (dismissed):

romange (approved):

src/server/script_mgr.cc
chakaz: @kostasrim this is already mostly implemented in absl:
https://github.com/abseil/abseil-cpp/blob/master/absl/algorithm/container.h

However I think that the general thinking there is that it was a mistake (despite being extremely widely used), and that ranges should be the better solution going forward.

kostasrim: Nice I did not know this about absl! There are wrappers in boost as well called boost range!

> and that ranges should be the better solution going forward. 

I agree but ranges is C++20 so I don't think we will use that anytime soon. ranges-v3 or csmstl are also outside of question so until we actually use those I would like to have some syntactice sugar in my life :smile:

romange (approved):

src/server/script_mgr.cc
romange: sugar is bad for your health.

PR #4527, opened 2025-01-29
chore: add more logs to the op_manager code

romange:


romange:
Following another random failure: https://github.com/dragonflydb/dragonfly/actions/runs/13029061313/job/36344088192#step:11:14787

i understand that we Opmanager calls StashNotify more than the expected 100 times but I do not understand yet why.

kostasrim (approved):

PR #4529, opened 2025-01-29
chore: unsubscribe pub/sub connections after cluster migration

### LINKED ISSUE #4517: Evict subscribed sharded pub/sub on cluster migrations

kostasrim:
After a slot migration we should unsubscribe all connections from affected channels. This PR adds this support.

* unsubscribe pub/sub connections for migrated slots
* add test

resolves #4517


kostasrim:
Needs a little bit of clean up which I will handle tomorrow.

src/server/conn_context.cc
```diff
@@ -223,6 +223,16 @@ size_t ConnectionContext::UsedMemory() const {
   return facade::ConnectionContext::UsedMemory() + dfly::HeapSize(conn_state);
 }
 
+void ConnectionContext::Unsubscribe(std::string_view channel) {
+  auto* sinfo = conn_state.subscribe_info.get();
+  sinfo->channels.erase(channel);
```
BorysTheDev, line 228: add DCHECK result > 0

src/server/channel_store.cc
```diff
@@ -302,4 +342,53 @@ void ChannelStoreUpdater::Apply() {
...
+    target->erase(it);
+    continue;
+  }
+
+  // Prepare replacement.
+  auto* replacement = new ChannelStore{target, store->patterns_};
+
+  // Update control block and unlock it.
+  cb.most_recent.store(replacement, memory_order_relaxed);
+  cb.update_mu.unlock();
```
BorysTheDev, line 374: use lock_guard

src/server/channel_store.cc
```diff
@@ -302,4 +342,53 @@ void ChannelStoreUpdater::Apply() {
...
+
+  // Update thread local references. Readers fetch subscribers via FetchSubscribers,
+  // which runs without preemption, and store references to them in self container Subscriber
+  // structs. This means that any point on the other thread is safe to update the channel store.
+  // Regardless of whether we need to replace, we dispatch to make sure all
+  // queued SubscribeMaps in the freelist are no longer in use.
+  shard_set->pool()->AwaitFiberOnAll([this](unsigned idx, util::ProactorBase*) {
+    ServerState::tlocal()->UnsubscribeSlotsAndUpdateChannelStore(
+        ops_, ChannelStore::control_block.most_recent.load(memory_order_relaxed));
```
BorysTheDev, line 383: why do you need to read ChannelStore::control_block.most_recent if you already have it a couple of lines above

BorysTheDev: maybe this code should also be executed under lock?

src/server/cluster/cluster_family.cc
```diff
@@ -625,6 +626,9 @@ void ClusterFamily::DflyClusterConfig(CmdArgList args, SinkReplyBuilder* builder
       auto deleted_slots = (before.GetRemovedSlots(after)).ToSlotRanges();
       deleted_slots.Merge(outgoing_migrations.slot_ranges);
       DeleteSlots(deleted_slots);
+      auto* channel_store = ServerState::tlocal()->channel_store();
+      auto deleted = SlotSet(deleted_slots);
+      channel_store->UnsubscribeAfterClusterSlotMigration(deleted);
```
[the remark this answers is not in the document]
kostasrim: Sure!

PR #4543, opened 2025-02-02
chore: Add benchmarks of qlist compression paths

romange:
1. Add testdata, move the existing test data to zstd-compressed format.
2. Add benchmarks in qlist_test that add lots of records, forcing compression of inner noded, or read compressed list checking the decompression.

Currently qlist uses valkey compression algorithm.

On my machine:
1. (valkey) Compression is 5 times slower than no compression.
2. Decompression is 37 times slower.


src/core/dash_test.cc
```diff
@@ -882,12 +882,11 @@ struct BlankPolicy : public BasicDashPolicy {
 // into a new segment, not every item finds a place.
 TEST_F(DashTest, SplitBug) {
   DashTable<uint64_t, uint64_t, BlankPolicy> table;
+  string path = base::ProgramRunfile("testdata/ids.txt.zst");
```
romange, line 885: unrelated to the PR - replaced test file with the compressed file to save disk space

chakaz (approved):

chakaz (approved):

PR #4547, opened 2025-02-02
chore: move QList::Node definition into dragonfly codebase

romange:

### DIFF
--- a/src/core/qlist.h
+++ b/src/core/qlist.h
@@ -20,6 +20,32 @@ class QList {
  public:
   enum Where { TAIL, HEAD };
 
+  /* Node is a 32 byte struct describing a listpack for a quicklist.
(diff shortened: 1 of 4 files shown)


src/core/qlist.h
```diff
@@ -20,6 +20,32 @@ class QList {
  public:
   enum Where { TAIL, HEAD };
 
+  /* Node is a 32 byte struct describing a listpack for a quicklist.
```
BorysTheDev, line 23: 40 bytes I think
Could you add static_assert

BorysTheDev (approved):

src/core/qlist.h
romange: You are right, I will fix everything in the next PR

Issue #4532, opened 2025-01-30
test_snapshoting_during_migration


kostasrim:
https://github.com/dragonflydb/dragonfly/actions/runs/13044880291/job/36393710585#step:6:1052

```
>               raise error
E               redis.exceptions.BusyLoadingError: Dragonfly is loading the dataset in memory

```

PR #4555, opened 2025-02-04
fix: test_snapshoting_during_migration

BorysTheDev:
fixes: #4532

problem: because of the same name of the snapshot, we load it on the second test iteration.
fix: create a random snapshot name for every iteration


kostasrim (approved):
:partying_face:

PR #4556, opened 2025-02-04
fix: consume parser error on dfly load

kostasrim:
`DFLY LOAD` triggers the parser's destructor `DCHECK` which checks if the parsed error was consumed. 

Furthermore, fixes `dfly load path append nonsense` succeeding with `ok` .

* fix dcheck crash on dfly load
* fix dfly load with more than required arguments

IMO do we really need that `DCHECK` in the destructor ? Sometimes, there is no really reason to even consume it (see this PR as an example).

### DIFF
--- a/src/server/dflycmd.cc
+++ b/src/server/dflycmd.cc
@@ -528,13 +528,7 @@ void DflyCmd::Load(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext* cn
     existing_keys = ServerFamily::LoadExistingKeys::kOverride;
   }
 
-  if (parser.HasNext()) {
-    parser.Error();
-  }
-
-  if (parser.HasError()) {
-    // consume error
-    parser.Error();
+  if (parser.Error() || parser.HasNext()) {
     return rb->SendError(kSyntaxErr);
   }
 
--- a/src/server/dflycmd.cc
+++ b/src/server/dflycmd.cc
@@ -533,6 +533,8 @@ void DflyCmd::Load(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext* cn
   }
 
   if (parser.HasError()) {
+    // consume error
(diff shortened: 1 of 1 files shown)


src/server/dflycmd.cc
```diff
@@ -533,6 +533,8 @@ void DflyCmd::Load(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext* cn
   }
 
   if (parser.HasError()) {
+    // consume error
+    parser.Error();
```
kostasrim, line 537: I think this is redundant. Maybe we should remove the `DCHECK` from the destructor of the parser ?

BorysTheDev: No DCHECK gives us an understanding of whether we check the error or not. Maybe instead of consuming the error we can return a more appropriate error that should be sent via reply builder

BorysTheDev (approved):

src/server/dflycmd.cc
kostasrim: Yes I understand what the DCHECK does. All I am saying is that for certain code paths (like the one found here), there is no really a reason to consume the error. On the other hand you can do that in one go `if (parse.Error())` to both check for error && consume it which I guess is fine also :shrug:

src/server/dflycmd.cc
```diff
@@ -528,11 +528,7 @@ void DflyCmd::Load(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext* cn
     existing_keys = ServerFamily::LoadExistingKeys::kOverride;
   }
 
-  if (parser.HasNext()) {
-    parser.Error();
```
kostasrim, line 532: Yeah, don't consume the error if you have extra arguments. We should reject `dfly load path append nonsense` :)

BorysTheDev (approved):

PR #4566, opened 2025-02-05
chore: reproduce a dcheck fail

romange:


romange:
@adiholden  see here:
https://github.com/dragonflydb/dragonfly/actions/runs/13166759636/job/36748647601?pr=4566#step:10:16263

PR #4576, opened 2025-02-07
fix: migration finalization process

BorysTheDev:


src/server/cluster/outgoing_slot_migration.cc
```diff
@@ -285,14 +285,12 @@ void OutgoingMigration::SyncFb() {
 bool OutgoingMigration::FinalizeMigration(long attempt) {
   // if it's not the 1st attempt and flows are work correctly we try to
   // reconnect and ACK one more time
-  VLOG(1) << "FinalizeMigration for " << cf_->MyID() << " : " << migration_info_.node_info.id
-          << " attempt " << attempt;
+  LOG(INFO) << "Finalize migration for " << cf_->MyID() << " : " << migration_info_.node_info.id
```
kostasrim, line 288: I am hesitant about the `LOG(INFO)` on migrations because this can easily pollute the logs under heavy migration scenarios. Same applies for all other `LOG(INFO)`

src/server/cluster/incoming_slot_migration.cc
```diff
@@ -192,12 +192,21 @@ bool IncomingSlotMigration::Join(long attempt) {
       return false;
     }
 
-    if ((bc_->WaitFor(absl::ToInt64Milliseconds(timeout - passed) * 1ms)) &&
-        (std::all_of(shard_flows_.begin(), shard_flows_.end(),
-                     [&](const auto& flow) { return flow->GetLastAttempt() == attempt; }))) {
+    // if data was sent after LSN, WaitFor() always returns false so to reduce wait time
+    // we check current state and if WaitFor false but GetLastAttempt() == attempt
+    // the Join is faield and we can return false
```
kostasrim, line 197:
```suggestion
    // the Join is failed and we can return false
```

src/server/cluster/outgoing_slot_migration.cc
BorysTheDev: no these logs are usefull we have another type of logs pollution.

src/server/cluster/outgoing_slot_migration.cc
```diff
@@ -316,20 +314,19 @@ bool OutgoingMigration::FinalizeMigration(long attempt) {
   }
 
   absl::Cleanup cleanup([&is_block_active, &pause_fb_opt]() {
-    is_block_active = false;
-    pause_fb_opt->JoinIfNeeded();
+    if (pause_fb_opt) {
```
kostasrim, line 317: Nice catch! We would dereference a `nullopt` and crash!

src/server/cluster/incoming_slot_migration.cc
```diff
@@ -192,12 +192,21 @@ bool IncomingSlotMigration::Join(long attempt) {
       return false;
     }
 
-    if ((bc_->WaitFor(absl::ToInt64Milliseconds(timeout - passed) * 1ms)) &&
-        (std::all_of(shard_flows_.begin(), shard_flows_.end(),
-                     [&](const auto& flow) { return flow->GetLastAttempt() == attempt; }))) {
+    // if data was sent after LSN, WaitFor() always returns false so to reduce wait time
+    // we check current state and if WaitFor false but GetLastAttempt() == attempt
+    // the Join is failed and we can return false
```
kostasrim, line 197: Is this really true? I enter the body of the `while(true)` for the first time and I call 

`wait_res = bc_->WaitFor(wait_time)` and it returns false because it got notified. Now I check with `std::_all_of` and it returns `true`. Why did the migration fail here ? Isn't that the happy path ?

src/server/cluster/incoming_slot_migration.cc
```diff
@@ -192,12 +192,21 @@ bool IncomingSlotMigration::Join(long attempt) {
...
+    const auto is_attempt_correct =
+        std::all_of(shard_flows_.begin(), shard_flows_.end(),
+                    [attempt](const auto& flow) { return flow->GetLastAttempt() == attempt; });
+
+    auto wait_res = bc_->WaitFor(wait_time);
+    if (is_attempt_correct) {
       state_.store(MigrationState::C_FINISHED);
       keys_number_ = cluster::GetKeyCount(slots_);
-      return true;
+      return wait_res;
```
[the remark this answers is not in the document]
BorysTheDev: and this is correct and it was the reason of timeout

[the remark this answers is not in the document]
BorysTheDev: Outgoing slot migration waits for the response from the target node, and that's why we are stuck. it's not deadlock. we have a timeout, this code prevents timeout happens if we see that migration finalization phase fails

src/server/cluster/incoming_slot_migration.cc
```diff
@@ -192,12 +192,21 @@ bool IncomingSlotMigration::Join(long attempt) {
       return false;
     }
 
-    if ((bc_->WaitFor(absl::ToInt64Milliseconds(timeout - passed) * 1ms)) &&
-        (std::all_of(shard_flows_.begin(), shard_flows_.end(),
-                     [&](const auto& flow) { return flow->GetLastAttempt() == attempt; }))) {
+    // if data was sent after LSN, WaitFor() always returns false so to reduce wait time
```
adiholden, line 195: Maybe instead of
bc_->Add();  // the flow isn't finished so we lock it again
in line 95 you should do bc_->Cancle() ? wouldnt this fix the issue?

BorysTheDev: In this case, we need to change our migration finalize algorithm. The current fix should work OK. let's merge it and then we can discuss changes, because we want to add the force finish if Pause failed

src/server/cluster/incoming_slot_migration.cc
```diff
@@ -192,12 +192,23 @@ bool IncomingSlotMigration::Join(long attempt) {
...
+        std::all_of(shard_flows_.begin(), shard_flows_.end(),
+                    [attempt](const auto& flow) { return flow->GetLastAttempt() == attempt; });
+
+    auto wait_res = bc_->WaitFor(wait_time);
+    if (is_attempt_correct) {
+      if (wait_res) {
+        state_.store(MigrationState::C_FINISHED);
+        keys_number_ = cluster::GetKeyCount(slots_);
+      }
+      return wait_res;
```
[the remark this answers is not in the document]
BorysTheDev: I see your point, to my mind it's not really needed because we send the error to the source node in anycase. But it can help a little to debug so I will add them

adiholden (approved):

PR #4580, opened 2025-02-07
chore: minor simplification around WaitWatch

romange:
1. Rename FinalizeWatched to RemovedWatched.
2. Simplify passing of WatchedKeys - instead of passing a callback we pass either a single value or signal that all the shard keys should be used. Less generic but more explicit.


src/server/engine_shard.cc
```diff
@@ -633,9 +629,7 @@ void EngineShard::PollExecution(const char* context, Transaction* trans) {
 
     if ((is_self && disarmed) || continuation_trans_->DisarmInShard(sid)) {
       auto bc = continuation_trans_->GetNamespace().GetBlockingController(shard_id_);
-      if (bool keep = run(continuation_trans_, false); !keep) {
-        // if this holds, we can remove this check altogether.
-        DCHECK(continuation_trans_ == nullptr);
```
kostasrim, line 638: Kinda confusing if this is the case  (cont == nullptr) cause then we don't need this if statement and we can just call run. And the assignment is also not need :thinking:

src/server/transaction.h
```diff
@@ -132,9 +132,9 @@ class Transaction {
   // Callacks should return `OpStatus` which is implicitly converitble to `RunnableResult`!
   using RunnableType = absl::FunctionRef<RunnableResult(Transaction* t, EngineShard*)>;
 
-  // Provides keys to block on for specific shard.
-  using WaitKeysProvider =
-      std::function<std::variant<ShardArgs, ArgSlice>(Transaction*, EngineShard* shard)>;
+  static constexpr std::nullopt_t kShardArgs{std::nullopt};
```
kostasrim, line 135: `nullopt` to denote the type passed via a variable. Not bad :nerd_face:

src/server/transaction.h
```diff
@@ -207,15 +207,14 @@ class Transaction {
 
   // Called by engine shard to execute a transaction hop.
   // txq_ooo is set to true if the transaction is running out of order
-  // not as the tx queue head.
-  // Returns true if the transaction continues running in the thread
+  // not as the tx queue head. Returns true if the transaction concludes.
```
kostasrim, line 210: Had to check every `RunInShard` calls just to be sure the logic was inverted. Imo it's nicer now via the variable `concludes`

kostasrim (dismissed):

src/server/engine_shard.cc
romange: yeah, I will fix it in the following up PRs. I did not want to change the logic.

kostasrim (approved):

PR #4589, opened 2025-02-11
fix: huge TTL in RESTORE cmd is rounded to kMaxExpireDeadlineMs

BorysTheDev:
`redis-cli -x RESTORE x 5430186761345 < foo.dump` returns errors with index out of range
problem: max value for TTL is kMaxExpireDeadlineMs
fix: if TTL > kMaxExpireDeadlineMs; TTL = kMaxExpireDeadlineMs

### DIFF
--- a/src/server/generic_family.cc
+++ b/src/server/generic_family.cc
@@ -106,9 +106,9 @@ class InMemSource : public ::io::Source {
 
 class RestoreArgs {
  private:
-  static constexpr time_t NO_EXPIRATION = 0;
+  static constexpr int64_t NO_EXPIRATION = 0;
 
-  time_t expiration_ = NO_EXPIRATION;
+  int64_t expiration_ = NO_EXPIRATION;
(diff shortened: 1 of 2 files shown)


src/server/generic_family.cc
```diff
@@ -106,17 +106,17 @@ ::io::Result<size_t> InMemSource::ReadSome(const iovec* v, uint32_t len) {
 
 class RestoreArgs {
  private:
-  static constexpr time_t NO_EXPIRATION = 0;
```
romange, line 109: time_t is int64_t so not need to change everywhere

BorysTheDev: time_t isn't int64_t it depends on the implementation

romange: on our implementation it's int64_t :) it can not be 32 bit

romange (approved):

PR #4595, opened 2025-02-12
[do not review] fix: remove explicit CXX from reflex third party

kostasrim:
It was impossible to compile dragonfly on my mac when `CXX` flag was set. The issue is that building the reflex lib explicitly set the `CXX` to `CMAKE_CXX_COMPILER`. However when the first is set in bash and the second is empty, cmake will try to resolve the value of `CMAKE_CXX_COMPILER` which would be different than the one set via the variable `CXX`.

### DIFF
--- a/.github/workflows/daily-builds.yml
+++ b/.github/workflows/daily-builds.yml
@@ -106,17 +106,18 @@ jobs:
           # Furthermore, keep in mind that on macos-15, /Applications/Xcode.app/... have been moved to /Library/Developer
           # so if we upgrade this flow to use a newer version this might fail.
           export CPPFLAGS="-isysroot $(xcrun --sdk macosx --show-sdk-path)"
-          ls -lah /Library/Developer/CommandLineTools/usr/bin
+          export CCFLAGS="-isysroot $(xcrun --sdk macosx --show-sdk-path)"
+          #ls -lah /Library/Developer/CommandLineTools/usr/bin/clang
 
-          echo "next ls"
+          #echo "next ls"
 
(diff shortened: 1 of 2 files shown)


kostasrim:
https://github.com/dragonflydb/dragonfly/actions/runs/13283723509

BorysTheDev (approved):

kostasrim:
I am closing as the changes are not needed. 

Removing `CMAKE_CXX_COMPILER` from reflex causes a linker failure with an error `std::__1::basic_string`. This is a name mangle for `libc++` and not `libstdc++` which we would expect with `gcc-12`.

The interesting thing is that the value of `CMAKE_CXX_COMPILER` when not explicitly set is deduced differently on our github runner(macos-13) and on my machine (macos-14). 

The former, respects the value of `CXX` env variable(when set) such that `CMAKE_CXX_COMPILER` takes the value of `CXX`. However, this is not true for my machine. My `CXX` is set to `/usr/bin/clang++` and its value is *ignored* by `CMAKE_CXX_COMPILER` and instead is deduced to `/Library/Developer/CommandLineTools/usr/bin/clang++`.

This is problematic because `/Library/Dev/...` *does not* set the include and library paths of the c++ standard library as opposed to `/usr/bin/clang++` which does that for us. 

Therefore I am forced to use `CMAKE_CXX_COMPILER` explicitly when invoking cmake even if `CXX` env is set if I want to compile with the llvm supplied by macos. This is rather unfortunate and error prone (as you need to remember to set CMAKE_CXX or you won't compile) but I don't see a good way to resolve.

PR #4601, opened 2025-02-12
fix: add more logs around blocked transaction notifications

### LINKED ISSUE #4600: BLMOVE should not return unexpected status
We should fix the symptom by returning NIL in case of "key not found" error, and provide enough logs to fix the root cause.

romange:
Fixes #4600


romange:
I think https://github.com/dragonflydb/dragonfly/pull/4586 fixes the whole issue. checking it now.

romange:
yeah, it's fixed in 1.26.4 or later

PR #4610, opened 2025-02-13
feat: support lz4 compression method

romange:
The feature is enabled via set_method call but since no code calls it the lz4 compression is still disabled in dragonfly.

### DIFF
--- a/src/core/qlist.cc
+++ b/src/core/qlist.cc
@@ -14,6 +14,7 @@ extern "C" {
 #include <absl/base/optimization.h>
 #include <absl/strings/escaping.h>
 #include <absl/strings/str_cat.h>
+#include <lz4frame.h>
 
 #include "base/logging.h"
 
@@ -33,12 +34,12 @@ using namespace std;
 #define SIZE_ESTIMATE_OVERHEAD 8
 
 /* Minimum listpack size in bytes for attempting compression. */
-#define MIN_COMPRESS_BYTES 48
+#define MIN_COMPRESS_BYTES 256
 
 /* Minimum size reduction in bytes to store compressed quicklistNode data.
  * This also prevents us from storing compression if the compression
  * resulted in a larger size than the original data. */
-#define MIN_COMPRESS_IMPROVE 8
+#define MIN_COMPRESS_IMPROVE 32
 
 /* This macro is used to compress a node.
  *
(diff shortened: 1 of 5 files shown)


src/core/qlist.cc
```diff
@@ -299,11 +379,14 @@ QList::Node* SplitNode(QList::Node* node, int offset, bool after, ssize_t* diff)
 
 }  // namespace
 
+__thread QList::Stats QList::stats;
```
BorysTheDev, line 382: maybe thread_local is better?

src/core/qlist.h
```diff
@@ -19,9 +19,10 @@ namespace dfly {
 class QList {
  public:
   enum Where { TAIL, HEAD };
+  enum COMPR_METHOD { LZF = 0, LZ4 = 1 };
 
-  /* Node is a 32 byte struct describing a listpack for a quicklist.
-   * We use bit fields keep the Node at 32 bytes.
+  /* Node is a 40 byte struct describing a listpack for a quicklist.
```
BorysTheDev, line 24: could you add static_assert(sizeof(Node) == 40)

src/core/qlist.h
```diff
@@ -208,8 +209,30 @@ class QList {
...
+    uint64_t decompression_calls = 0;
+
+    // How many bytes we currently keep compressed.
+    size_t compressed_bytes = 0;
+
+    // how many bytes we compressed from.
+    // Compressed savings are calculated as raw_compressed_bytes - compressed_bytes.
+    size_t raw_compressed_bytes = 0;
+  };
+  static __thread Stats stats;
```
BorysTheDev, line 234: also maybe thread_local is better

BorysTheDev (approved):

src/core/qlist.h
romange: i did https://github.com/dragonflydb/dragonfly/pull/4610/files#diff-54d2b4173b5bde6d33fb2fda2f465e63f6d0168ebb4b8821357be5eccfe00ba4R68

src/core/qlist.cc
romange: thread_local has a higher access price than __thread

BorysTheDev: why

romange:
```
// __thread is more efficient than thread_local when a variable is referenced
// in cc file that does not define it.
```
because of the c++ constructors and lazy initialization rules of thread local variables. you can use godbolt to see it yourself.

PR #4612, opened 2025-02-14
fix: epoll reg tests running on iouring

kostasrim:
I accidentally introduced this bug which switched to iouring instead of epoll on the epoll regression tests.

* properly run epoll instead of iouring
* fix print message on failure

### DIFF
--- a/.github/workflows/epoll-regression-tests.yml
+++ b/.github/workflows/epoll-regression-tests.yml
@@ -57,7 +57,7 @@ jobs:
           s3-bucket: ${{ secrets.S3_REGTEST_BUCKET }}
           # Chain ternary oprator of the form (which can be nested)
           # (expression == condition && <true expression> || <false expression>)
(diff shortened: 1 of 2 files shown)


.github/workflows/epoll-regression-tests.yml
```diff
@@ -57,7 +57,7 @@ jobs:
           s3-bucket: ${{ secrets.S3_REGTEST_BUCKET }}
           # Chain ternary oprator of the form (which can be nested)
           # (expression == condition && <true expression> || <false expression>)
-          epoll: ${{ matrix.proactor == 'Epoll' && 'true' || '' }}
+          epoll: ${{ matrix.proactor == 'Epoll' && 'epoll' || 'iouring' }}
```
kostasrim, line 60: :man_facepalming:

kostasrim:
It now properly uses epoll https://github.com/dragonflydb/dragonfly/actions/runs/13325713851/job/37218402113#step:6:22

adiholden (approved):

Issue #4611, opened 2025-02-14
test_keys_expiration_during_migration


kostasrim:
https://github.com/dragonflydb/dragonfly/actions/runs/13319793690/job/37202103193

```
E               AssertionError: [['in', 'f8276088a519a80edcc02bc03806a5ef74f2e26e', 'SYNC', 101229, '0']]
E               assert (1 != 0 and False)
E                +  where 1 = len([['in', 'f8276088a519a80edcc02bc03806a5ef74f2e26e', 'SYNC', 101229, '0']])
E                +  and   False = all(<generator object wait_for_status.<locals>.<genexpr> at 0xfc63e29cd900>)
```

PR #4613, opened 2025-02-14
fix: test_keys_expiration_during_migration

BorysTheDev:
fixes: #4611
problem: too long migration process

### DIFF
--- a/tests/dragonfly/cluster_test.py
+++ b/tests/dragonfly/cluster_test.py
@@ -1892,13 +1892,13 @@ async def test_keys_expiration_during_migration(df_factory: DflyInstanceFactory)
     await push_config(json.dumps(generate_config(nodes)), [node.admin_client for node in nodes])
 
     logging.debug("Start seeder")
-    await nodes[0].client.execute_command("debug", "populate", "100000", "foo", "100", "RAND")
+    await nodes[0].client.execute_command("debug", "populate", "100", "foo", "100", "RAND")
 
     capture_before = await StaticSeeder.capture(nodes[0].client)
 
     seeder = ExpirySeeder(timeout=4)
     seeder_task = asyncio.create_task(seeder.run(nodes[0].client))
-    await seeder.wait_until_n_inserts(1000)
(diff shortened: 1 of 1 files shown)


tests/dragonfly/cluster_test.py
```diff
@@ -1892,13 +1892,13 @@ async def test_keys_expiration_during_migration(df_factory: DflyInstanceFactory)
     await push_config(json.dumps(generate_config(nodes)), [node.admin_client for node in nodes])
 
     logging.debug("Start seeder")
-    await nodes[0].client.execute_command("debug", "populate", "100000", "foo", "100", "RAND")
+    await nodes[0].client.execute_command("debug", "populate", "100", "foo", "100", "RAND")
```
kostasrim, line 1895: I am confused. You are saying that "migration too long" but I don't see why ?

`DEBUG POPULATE 100000 foo 100 RAND` creates 100k keys, of size 100 with 1 element each. So total size is is roughly `1mb`. I don't see any reason why would this be "too slow" when  it takes <1s to fill this data ?

Furthermore, we fail in

>       await wait_for_status(nodes[1].admin_client, nodes[0].id, "FINISHED")


Why would this fail with only `1mb` of data? This looks like an actual regression than incorrect parameterizon

BorysTheDev: 100 000 * 100 = 10MB + expiry_generator

kostasrim: As I explained internally the whole payload is roughly 50MB. I am hesitant that we reach a timeout of 10 seconds for that even on Debug build. 

However, I don't want to block you and I am not that much familiar with cluster to see what the low hanging fruits here are.

I am accepting this and just to be on the safe side, I will run this test in a loop with an increased timeout just to make sure everything works correctly. Feel free to merge in the meantime

kostasrim: https://github.com/dragonflydb/dragonfly/pull/4615

PR #4634, opened 2025-02-20
fix: performance degradation with Reflex matching

romange:
1. cherry-pick valkey security fixes with stringmatchlen with exponential complexity of stars See https://nvd.nist.gov/vuln/detail/cve-2022-36021
2. Fallback to stringmatchlen for short lengths.
3. Add another backend to GlobMatcher - PCRE2, though do not enable at the moment. While this backend is the fastest one - it requires an additional shared lib dependency.

### DIFF
--- a/src/core/glob_matcher.cc
+++ b/src/core/glob_matcher.cc
@@ -12,8 +12,8 @@ namespace dfly {
 using namespace std;
 
 /* Glob-style pattern matching taken from Redis. */
-int stringmatchlen(const char* pattern, int patternLen, const char* string, int stringLen,
-                   int nocase) {
+static int stringmatchlen_impl(const char* pattern, int patternLen, const char* string,
+                               int stringLen, int nocase, int* skipLongerMatches) {
   while (patternLen && stringLen) {
     switch (pattern[0]) {
       case '*':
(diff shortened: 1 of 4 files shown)


kostasrim:
Is PCR Perl Compatible Regular Expressions ? Never heard of it so I wanted to check it out

src/core/glob_matcher.cc
```diff
@@ -24,11 +24,26 @@ int stringmatchlen(const char* pattern, int patternLen, const char* string, int
...
+         * from anywhere in the rest of the string. If there were
+         * any '*' earlier in the pattern, we can terminate the
+         * search early without trying to match them to longer
+         * substrings. This is because a longer match for the
+         * earlier part of the pattern would require the rest of the
+         * pattern to match starting later in the string, and we
+         * have just determined that there is no match for the rest
+         * of the pattern starting from anywhere in the current
+         * string. */
+        *skipLongerMatches = 1;
```
kostasrim, line 46: Why is this used as a `bool` but is marked as `int` and passed as `int *` ?

src/core/glob_matcher.h
```diff
@@ -8,6 +8,16 @@
 #include <string>
 #include <string_view>
 
+#define FIX_PERFORMANCE_MATCHING
+
+#ifndef FIX_PERFORMANCE_MATCHING
```
kostasrim, line 13: You first define `FIX_PERFORMANCE`  and then you `ifndef` it. Isn't that a bug ?

src/core/glob_matcher.cc
romange: it's taken from valkey codebase.

src/core/glob_matcher.h
romange: `FIX_PERFORMANCE` is a bad name.

it should be REFLEX_PERFORMANCE

adiholden (approved):

Issue #4572, opened 2025-02-06
Add metric for migration errors


andydunstall:
Discussed with @BorysTheDev earlier, `SLOT-MIGRATION-STATUS` may return a state of `ERROR`, but that doesn't indicate the migration failed, only that an error occurred and the migration will be retried internally by Dragonfly.

So from the control planes perspective, the only state that it should take action on in `FINISHED`? And the only reason for returning state `ERROR` is for the control plane to monitor migration errors to fire a warning alert?

In which case can Dragonfly just expose a `dragonfly_migration_errors_total` metric that we can alert on directly, rather than the control plane monitoring state `ERROR`

PR #4643, opened 2025-02-21
feat: add migration error statistic

BorysTheDev:
fixes: #4572

added migration_errors_total statistic
fixed data race

### DIFF
--- a/src/server/cluster/cluster_family.cc
+++ b/src/server/cluster/cluster_family.cc
@@ -1038,6 +1038,22 @@ void ClusterFamily::PauseAllIncomingMigrations(bool pause) {
   }
 }
 
+size_t ClusterFamily::MigrationsErrorNum() const {
+  util::fb2::LockGuard lk(migration_mu_);
+
+  size_t error_num = 0;
+
+  for (const auto& mj : incoming_migrations_jobs_) {
+    error_num += mj->GetErrorsNum();
+  }
+
+  for (const auto& mj : outgoing_migration_jobs_) {
+    error_num += mj->GetErrorsNum();
+  }
+
+  return error_num;
+}
+
 using EngineFunc = void (ClusterFamily::*)(CmdArgList args, const CommandContext& cmd_cntx);
 
 inline CommandId::Handler3 HandlerFunc(ClusterFamily* se, EngineFunc f) {
--- a/src/server/cluster/cluster_family.cc
+++ b/src/server/cluster/cluster_family.cc
@@ -1038,17 +1038,17 @@ void ClusterFamily::PauseAllIncomingMigrations(bool pause) {
   }
 }
 
-size_t ClusterFamily::MigrationsErrorNum() const {
+size_t ClusterFamily::MigrationsErrorsCount() const {
   util::fb2::LockGuard lk(migration_mu_);
 
   size_t error_num = 0;
 
   for (const auto& mj : incoming_migrations_jobs_) {
-    error_num += mj->GetErrorsNum();
+    error_num += mj->GetErrorsCount();
   }
 
   for (const auto& mj : outgoing_migration_jobs_) {
-    error_num += mj->GetErrorsNum();
+    error_num += mj->GetErrorsCount();
   }
 
   return error_num;
--- a/src/server/cluster/incoming_slot_migration.h
+++ b/src/server/cluster/incoming_slot_migration.h
@@ -48,6 +48,7 @@ class IncomingSlotMigration {
   }
 
   void ReportError(dfly::GenericError err) ABSL_LOCKS_EXCLUDED(error_mu_) {
+    error_num_.fetch_add(1, std::memory_order_relaxed);
     util::fb2::LockGuard lk(error_mu_);
     last_error_ = std::move(err);
   }
@@ -55,6 +56,10 @@ class IncomingSlotMigration {
   std::string GetErrorStr() const ABSL_LOCKS_EXCLUDED(error_mu_) {
     util::fb2::LockGuard lk(error_mu_);
     return last_error_.Format();
+  }
+
+  size_t GetErrorsNum() const {
+    return error_num_.load(std::memory_order_relaxed);
   }
 
   size_t GetKeyCount() const;
@@ -70,6 +75,7 @@ class IncomingSlotMigration {
   ExecutionState cntx_;
   mutable util::fb2::Mutex error_mu_;
   dfly::GenericError last_error_ ABSL_GUARDED_BY(error_mu_);
+  std::atomic<size_t> error_num_ = 0;
 
   // when migration is finished we need to store number of migrated keys
   // because new request can add or remove keys and we get incorrect statistic
--- a/src/server/cluster/incoming_slot_migration.h
+++ b/src/server/cluster/incoming_slot_migration.h
@@ -48,7 +48,7 @@ class IncomingSlotMigration {
   }
 
   void ReportError(dfly::GenericError err) ABSL_LOCKS_EXCLUDED(error_mu_) {
-    error_num_.fetch_add(1, std::memory_order_relaxed);
+    errors_count_.fetch_add(1, std::memory_order_relaxed);
     util::fb2::LockGuard lk(error_mu_);
     last_error_ = std::move(err);
   }
@@ -58,8 +58,8 @@ class IncomingSlotMigration {
     return last_error_.Format();
   }
 
-  size_t GetErrorsNum() const {
-    return error_num_.load(std::memory_order_relaxed);
+  size_t GetErrorsCount() const {
+    return errors_count_.load(std::memory_order_relaxed);
   }
 
   size_t GetKeyCount() const;
@@ -75,7 +75,7 @@ class IncomingSlotMigration {
   ExecutionState cntx_;
   mutable util::fb2::Mutex error_mu_;
   dfly::GenericError last_error_ ABSL_GUARDED_BY(error_mu_);
-  std::atomic<size_t> error_num_ = 0;
+  std::atomic<size_t> errors_count_ = 0;
 
   // when migration is finished we need to store number of migrated keys
   // because new request can add or remove keys and we get incorrect statistic
--- a/src/server/cluster/outgoing_slot_migration.cc
+++ b/src/server/cluster/outgoing_slot_migration.cc
@@ -226,11 +226,11 @@ void OutgoingMigration::SyncFb() {
     if (!CheckRespIsSimpleReply("OK")) {
       if (CheckRespIsSimpleReply(kUnknownMigration)) {
         LOG(WARNING) << "Target node does not recognize migration; retrying";
-        cntx_.ReportError(GenericError(std::string(ToSV(LastResponseArgs().front().GetBuf()))));
+        cntx_.ReportError(GenericError(LastResponseArgs().front().GetString()));
         ThisFiber::SleepFor(1000ms);
       } else {
         LOG(WARNING) << "Unable to initialize migration";
-        cntx_.ReportError(GenericError(std::string(ToSV(LastResponseArgs().front().GetBuf()))));
+        cntx_.ReportError(GenericError(LastResponseArgs().front().GetString()));
       }
       continue;
     }
--- a/src/server/cluster/outgoing_slot_migration.cc
+++ b/src/server/cluster/outgoing_slot_migration.cc
@@ -66,9 +66,8 @@ class OutgoingMigration::SliceSlotMigration : private ProtocolClient {
     }
 
     if (!CheckRespIsSimpleReply("OK")) {
-      LOG(WARNING) << "Incorrect response for FLOW cmd: "
-                   << ToSV(LastResponseArgs().front().GetBuf());
-      cntx_.ReportError("Incorrect response for FLOW cmd");
+      cntx_.ReportError(absl::StrCat("Incorrect response for FLOW cmd: ",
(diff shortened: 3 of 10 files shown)


src/server/cluster/outgoing_slot_migration.cc
```diff
@@ -195,11 +195,12 @@ void OutgoingMigration::SyncFb() {
     }
 
     if (cntx_.IsError()) {
-      last_error_ = cntx_.GetError();
-      LOG(ERROR) << last_error_;
+      auto err = cntx_.GetError();
+      LOG(ERROR) << err.Format();
+      ReportError(std::move(err));
+      cntx_.Reset(nullptr);
```
kostasrim, line 201: This will preempt the current fiber because it will need to Join on the error handler fiber `ReportError`. So if this is the case, do we really need to sleep below ?

src/server/cluster/outgoing_slot_migration.cc
```diff
@@ -224,7 +225,8 @@ void OutgoingMigration::SyncFb() {
 
     if (!CheckRespIsSimpleReply("OK")) {
       if (CheckRespIsSimpleReply(kUnknownMigration)) {
-        VLOG(2) << "Target node does not recognize migration; retrying";
+        LOG(WARNING) << "Target node does not recognize migration; retrying";
+        cntx_.ReportError(GenericError(std::string(ToSV(LastResponseArgs().front().GetBuf()))));
```
kostasrim, line 229:
```suggestion
        cntx_.ReportError(GenericError(LastResponseArgs().front().GetString()));
```

kostasrim: And there the same expression below if you are not bored to change that as well

tests/dragonfly/cluster_test.py
```diff
@@ -2943,40 +2943,50 @@ async def do_migration(index):
 
 
 @dfly_args({"proactor_threads": 2, "cluster_mode": "yes"})
-async def test_cluster_sharded_pub_sub(df_factory: DflyInstanceFactory):
```
kostasrim, line 2946: I think you accidentally removed/replaced my test :smile: (test_cluster_sharded_pub_sub)

src/server/cluster/cluster_family.cc
```diff
@@ -1038,6 +1038,22 @@ void ClusterFamily::PauseAllIncomingMigrations(bool pause) {
   }
 }
 
+size_t ClusterFamily::MigrationsErrorNum() const {
+  util::fb2::LockGuard lk(migration_mu_);
```
kostasrim, line 1042: Why do we need this mutex ? `GetErrorNum()` is an `atomic operation` so what's the reason for the extra synchronization here ?

src/server/cluster/incoming_slot_migration.h
```diff
@@ -48,6 +48,7 @@ class IncomingSlotMigration {
   }
 
   void ReportError(dfly::GenericError err) ABSL_LOCKS_EXCLUDED(error_mu_) {
+    error_num_.fetch_add(1, std::memory_order_relaxed);
```
kostasrim, line 51: That's per migration flow right ?

src/server/cluster/cluster_family.cc
BorysTheDev: incoming_migrations_jobs_ and can be removed

src/server/cluster/incoming_slot_migration.h
BorysTheDev: per every migration, not migration flow

kostasrim: yes yes that's what I meant. Lit makes sense

src/server/cluster/outgoing_slot_migration.cc
BorysTheDev: it reduces a load on the instance. I don't see any issues with it and prefer to leave as is

src/server/cluster/cluster_family.cc
kostasrim: oh now I understand. And then the atomic is to do the increment without the lock

tests/dragonfly/cluster_test.py
BorysTheDev: thx

src/server/cluster/outgoing_slot_migration.h
```diff
@@ -53,10 +53,21 @@ class OutgoingMigration : private ProtocolClient {
     return migration_info_;
   }
 
-  std::string GetErrorStr() const {
+  void ReportError(dfly::GenericError err) ABSL_LOCKS_EXCLUDED(error_mu_) {
```
adiholden, line 56: I find the naming here a bit confusing, you already reported the error to the execution state , now you just save last error .  I would call it OnErrorReset or similar and will do the 
cntx_.GetError();
and the cntx_.Reset(nullptr); 
in this function

src/server/cluster/outgoing_slot_migration.cc
```diff
@@ -195,11 +195,12 @@ void OutgoingMigration::SyncFb() {
     }
 
     if (cntx_.IsError()) {
-      last_error_ = cntx_.GetError();
-      LOG(ERROR) << last_error_;
-      ThisFiber::SleepFor(1000ms);  // wait some time before next retry
+      auto err = cntx_.GetError();
+      LOG(ERROR) << err.Format();
```
adiholden, line 199: why do we print here? if we want to print the error I think it should be inside execution state error reporting

src/server/cluster/outgoing_slot_migration.cc
```diff
@@ -195,11 +195,12 @@ void OutgoingMigration::SyncFb() {
...
+      ReportError(std::move(err));
+      cntx_.Reset(nullptr);
+      ThisFiber::SleepFor(500ms);  // wait some time before next retry
```
adiholden, line 202: why did you change the sleep time?

src/server/cluster/cluster_family.cc
```diff
@@ -1038,6 +1038,22 @@ void ClusterFamily::PauseAllIncomingMigrations(bool pause) {
   }
 }
 
+size_t ClusterFamily::MigrationsErrorNum() const {
```
adiholden, line 1041: nit
MigrationsErrorNum->MigrationsErrorsCount

src/server/cluster/cluster_family.cc
```diff
@@ -1038,6 +1038,22 @@ void ClusterFamily::PauseAllIncomingMigrations(bool pause) {
...
+  util::fb2::LockGuard lk(migration_mu_);
+
+  size_t error_num = 0;
+
+  for (const auto& mj : incoming_migrations_jobs_) {
+    error_num += mj->GetErrorsNum();
```
adiholden, line 1047: nit 
also here error count is better I think

src/server/cluster/outgoing_slot_migration.h
```diff
@@ -82,7 +93,9 @@ class OutgoingMigration : private ProtocolClient {
   std::vector<std::unique_ptr<SliceSlotMigration>> slot_migrations_;
   ServerFamily* server_family_;
   ClusterFamily* cf_;
-  dfly::GenericError last_error_;
+  mutable util::fb2::Mutex error_mu_;
+  dfly::GenericError last_error_ ABSL_GUARDED_BY(error_mu_);
+  std::atomic<size_t> error_num_ = 0;
```
adiholden, line 98: error count is better I think 
but could be that reconnect count is even better as the fact that we have error does not say that we did restarted , but in our flow I think that each error results in restarting the migration

src/server/cluster/outgoing_slot_migration.cc
```diff
@@ -223,12 +224,11 @@ void OutgoingMigration::SyncFb() {
     }
 
     if (!CheckRespIsSimpleReply("OK")) {
+      cntx_.ReportError(GenericError(LastResponseArgs().front().GetString()));
       if (CheckRespIsSimpleReply(kUnknownMigration)) {
-        VLOG(2) << "Target node does not recognize migration; retrying";
-        ThisFiber::SleepFor(1000ms);
+        LOG(WARNING) << "Target node does not recognize migration; retrying";
```
adiholden, line 229: does this different prints really help?
if ReportError will print the relevant error to log this will be the best

src/server/cluster/outgoing_slot_migration.h
BorysTheDev: I would prefer to leave the errors count, it is more flexible, because we can improve our logic in the future and for example, restart only one flow if it fails.

src/server/cluster/outgoing_slot_migration.cc
adiholden: not sure why comment is resolved. you can move this message to cntx_.ReportError

tests/dragonfly/cluster_test.py
```diff
@@ -2980,3 +2980,53 @@ async def test_cluster_sharded_pub_sub(df_factory: DflyInstanceFactory):
...
+
+    await wait_for_errors_num(c_nodes[0], 0)
+
+    nodes_info[0].migrations.append(
+        MigrationInfo("127.0.0.1", nodes_info[1].instance.admin_port, [(0, 100)], nodes_info[1].id)
+    )
+
+    await push_config(json.dumps(generate_config(nodes_info)), [c_nodes[0]])
+
+    await wait_for_errors_num(c_nodes[0], 1)
```
adiholden, line 3024: you dont have another push_config after you edit the config so what are we checking ?

src/server/cluster/outgoing_slot_migration.cc
BorysTheDev: marked wrong comment

tests/dragonfly/cluster_test.py
BorysTheDev: I check the missing config on the target node error

adiholden: sorry, I dont understand.
in line 3026 you edit the config but if you dont push it, it has no affect

BorysTheDev: you are right I will check

BorysTheDev: the tests works correctly because we restart the migration process and get the error one more time.

src/server/cluster/outgoing_slot_migration.cc
```diff
@@ -223,13 +220,7 @@ void OutgoingMigration::SyncFb() {
     }
 
     if (!CheckRespIsSimpleReply("OK")) {
-      if (CheckRespIsSimpleReply(kUnknownMigration)) {
-        VLOG(2) << "Target node does not recognize migration; retrying";
```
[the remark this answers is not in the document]
BorysTheDev: no, it wasn't a bug. As I remember @andydunstall removed this error because they parsed the status to understand the migration error status. Now I've added an error statistic so the control plane team can ignore this error, so I returned this error.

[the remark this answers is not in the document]
adiholden: but we will get alerted by that from graffana once they move to track this migration error count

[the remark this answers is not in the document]
BorysTheDev: No, we have discussed with Andy that we generate alerts only if a number of errors is more than some threshold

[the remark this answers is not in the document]
adiholden: @andydunstall I want to make sure control plane test will not fail due to this change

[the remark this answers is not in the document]
andydunstall: in the control plane we see `unknown migration` a lot, given there can be up to 10s between shards configuration being updated

as i understand it, if you remove this we'll end up flooding the logs with redundant warnings which doesn't sound right? it also backs of 1 second before retrying, so if you remove will it just sit in a busy loop retrying?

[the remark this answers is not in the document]
BorysTheDev: I would prefer do not remove it because it's error, we don't know do you forget to send the config or it's only a delay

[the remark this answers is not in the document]
BorysTheDev: @andydunstall Is it possible to send a config to the target node first?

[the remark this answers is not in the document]
andydunstall: > we don't know do you forget to send the config

The control plane monitors that every node is configured, so from the data planes perspective it's not an error

> Is it possible to send a config to the target node first?

It would add a lot of complexity which doesn't seem worth it?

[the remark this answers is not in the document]
BorysTheDev: @andydunstall What if I add timeout or 10 attempts before report the error?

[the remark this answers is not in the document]
andydunstall: sure - as discussed in chat i don't really see the point given the control plane monitors whether configuration is propagated - but a 30s timeout sounds fine

adiholden (approved):

src/server/cluster/outgoing_slot_migration.cc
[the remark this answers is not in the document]
andydunstall: > it also backs of 1 second before retrying, so if you remove will it just sit in a busy loop retrying?

we're seeing a busy loop retrying... can we add the 1 second back off again?

PR #4644, opened 2025-02-22
chore: get rid of lispack stats

romange:
They were wrong, we usually do not need them,
and they complicated code. The right way to do it - to add them to OBJHIST statistics.

In addition, did some other code improvements without functional changes.

### DIFF
--- a/src/core/string_set.cc
+++ b/src/core/string_set.cc
@@ -51,6 +51,30 @@ bool StringSet::Add(string_view src, uint32_t ttl_sec) {
   return true;
 }
 
+unsigned StringSet::AddMany(absl::Span<std::string_view> span, uint32_t ttl_sec) {
+  std::string_view views[kMaxBatchLen];
+  unsigned res = 0;
+  if (BucketCount() < span.size()) {
+    Reserve(span.size());
+  }
+
+  while (span.size() >= kMaxBatchLen) {
+    for (size_t i = 0; i < kMaxBatchLen; i++)
+      views[i] = span[i];
+
+    span.remove_prefix(kMaxBatchLen);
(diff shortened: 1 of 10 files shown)


src/core/string_set.cc
```diff
@@ -51,6 +51,30 @@ bool StringSet::Add(string_view src, uint32_t ttl_sec) {
   return true;
 }
 
+unsigned StringSet::AddMany(absl::Span<std::string_view> span, uint32_t ttl_sec) {
```
romange, line 54: moved from the header file

src/core/string_set.h
```diff
@@ -29,7 +29,7 @@ class StringSet : public DenseSet {
   // Returns true if elem was added.
   bool Add(std::string_view s1, uint32_t ttl_sec = UINT32_MAX);
 
-  template <typename T> unsigned AddMany(absl::Span<T> span, uint32_t ttl_sec);
+  unsigned AddMany(absl::Span<std::string_view> span, uint32_t ttl_sec);
```
romange, line 32: removed template

src/server/rdb_load.cc
```diff
@@ -2171,11 +2171,7 @@ error_code RdbLoader::Load(io::Source* src) {
     }
 
     ++keys_loaded;
-    int64_t start = absl::GetCurrentTimeNanos();
     RETURN_ON_ERR(LoadKeyValPair(type, &settings));
-    int delta_ms = (absl::GetCurrentTimeNanos() - start) / 1000'000;
```
romange, line 2176: moved inside `LoadKeyValPair` function

kostasrim (approved):
LGTM

PR #4649, opened 2025-02-23
chore: allow sampling of topk hottest keys.

romange:
The core data structure code was added long time ago, now we allow using it via `DEBUG SAMPLE` subcommand.


src/server/table.h
```diff
@@ -130,7 +131,7 @@ struct DbTable : boost::intrusive_ref_counter<DbTable, boost::thread_unsafe_coun
   std::vector<SlotStats> slots_stats;
   ExpireTable::Cursor expire_cursor;
 
-  TopKeys top_keys;
+  TopKeys* top_keys = nullptr;
```
kostasrim, line 134: Why not make it a `unique_ptr`? That way you don't need to handle destruction explicitly in `~DbTable`

src/server/db_slice.cc
```diff
@@ -537,6 +538,10 @@ OpResult<DbSlice::PrimeItAndExp> DbSlice::FindInternal(const Context& cntx, std:
     return OpStatus::KEY_NOTFOUND;
   }
 
+  if (db.top_keys) {
```
kostasrim, line 541: nit: consider a function since you use the same code below as well

src/server/debugcmd.cc
```diff
@@ -584,6 +587,10 @@ void DebugCmd::Run(CmdArgList args, facade::SinkReplyBuilder* builder) {
     return RecvSize(ArgS(args, 1), builder);
   }
 
+  if (subcmd == "TOPK" && args.size() >= 2) {
```
kostasrim, line 590: One of the things I don't really like about how we implement subcommands, is that we miss basic arity checks and they are also orthogonal to our ACL's (subcommands of ACL"s are actual commands stored in the registry as opposed to what we have here). Long time now I wanted to check if we can simplify this, but never got a chance to actually check this. 

Not important just thinking out loud.

kostasrim (commented):
I haven't checked the TopKeys changes since I am unfamiliar with the class. Happy to jump in though if it's needed

src/core/top_keys.h
```diff
@@ -29,29 +30,27 @@ namespace dfly {
...
 //   are too few buckets, or when min_key_count_to_record is too high, depending on actual usage.
 class TopKeys {
+  TopKeys(const TopKeys&) = delete;
+  TopKeys& operator=(const TopKeys&) = delete;
  public:
   struct Options {
     // HeavyKeeper options
-    uint64_t buckets = 1 << 16;
-    uint64_t arrays = 4;
+    uint32_t buckets = 1 << 16;
```
BorysTheDev, line 38: Why do you use sometimes  uint64 for buckets and sometimes uint32

src/server/table.h
BorysTheDev: to my mind unique_ptr is preferable because it's safer.

src/server/table.cc
```diff
@@ -94,6 +94,9 @@ DbTable::DbTable(PMR_NS::memory_resource* mr, DbIndex db_index)
 
 DbTable::~DbTable() {
   DCHECK_EQ(thread_index, ServerState::tlocal()->thread_index());
+  if (top_keys) {
+    delete top_keys;
+  }
```
BorysTheDev, line 99: nit you don't need if(top_keys) delete nullptr is correct statement

src/server/debugcmd.cc
```diff
@@ -1125,4 +1132,59 @@ void DebugCmd::RecvSize(string_view param, facade::SinkReplyBuilder* builder) {
   rb->SendVerbatimString(hist);
 }
 
+void DebugCmd::Topk(CmdArgList args, facade::SinkReplyBuilder* builder) {
+  auto* rb = static_cast<RedisReplyBuilder*>(builder);
+  DCHECK_GE(args.size(), 1u);
+
+  string_view subcmd = ArgS(args, 0);
```
BorysTheDev, line 1139: what about cmdArgParser

BorysTheDev (approved):

src/core/top_keys.h
romange: fixed

src/server/debugcmd.cc
BorysTheDev: I thought maybe we can provide parse function for every command and split the code of parsing and execution of the command

BorysTheDev (approved):

PR #4653, opened 2025-02-24
fix: race condition in ReplicaOfInternal

kostasrim:
#4636 introduced a small race . The problem is that `Start()` might have failed (because a subsequent ReplicaOf command cancelled it). `ReplicaOfInternal` would ignore the `ec` and start the replication fiber via `replica_->StartMainReplicationFiber` but the second ReplicaOf command that followed updated `replica_` so we tried to start replication twice on the same replica instance triggering a check failed that the fiber is running.

We did not have that problem before because we called `new_replica_->Start()` which would start replication. Even if the replication got cancelled because of a second `ReplicaOf` command, it would be fine because the `MainReplicationFiber` would start twice on two separate Replica instances and one of them would cancel right away because`R_CANCELLED` was not set (since `Stop()` explicitly set it by cancelling the context)

### DIFF
--- a/src/server/error.h
+++ b/src/server/error.h
@@ -22,7 +22,7 @@ using facade::kWrongTypeErr;
 
 #define RETURN_ON_ERR_T(T, x)                                          \
   do {                                                                 \
-    T __ec = (x);                                                      \
+    std::error_code __ec = (x);                                        \
     if (__ec) {                                                        \
       DLOG(ERROR) << "Error while calling " #x ": " << __ec.message(); \
       return (T)(__ec);                                                \
@@ -31,7 +31,13 @@ using facade::kWrongTypeErr;
 
 #define RETURN_ON_ERR(x) RETURN_ON_ERR_T(std::error_code, x)
 
-#define RETURN_ON_GENERIC_ERR(x) RETURN_ON_ERR_T(GenericError, x)
+#define RETURN_ON_GENERIC_ERR(x)                                   \
+  do {                                                             \
+    if (x) {                                                       \
+      DLOG(ERROR) << "Error while calling " #x ": " << x.Format(); \
+      return x;                                                    \
+    }                                                              \
+  } while (0)
 
 #endif  // RETURN_ON_ERR
 
--- a/src/server/error.h
+++ b/src/server/error.h
@@ -22,7 +22,7 @@ using facade::kWrongTypeErr;
 
 #define RETURN_ON_ERR_T(T, x)                                          \
   do {                                                                 \
-    std::error_code __ec = (x);                                        \
+    T __ec = (x);                                                      \
     if (__ec) {                                                        \
       DLOG(ERROR) << "Error while calling " #x ": " << __ec.message(); \
       return (T)(__ec);                                                \
@@ -30,6 +30,8 @@ using facade::kWrongTypeErr;
   } while (0)
 
 #define RETURN_ON_ERR(x) RETURN_ON_ERR_T(std::error_code, x)
+
+#define RETURN_ON_GENERIC_ERR(x) RETURN_ON_ERR_T(GenericError, x)
 
 #endif  // RETURN_ON_ERR
 
--- a/src/server/replica.cc
+++ b/src/server/replica.cc
@@ -86,19 +86,18 @@ Replica::~Replica() {
 
 static const char kConnErr[] = "could not connect to master: ";
 
-error_code Replica::Start(facade::SinkReplyBuilder* builder) {
+GenericError Replica::Start() {
   VLOG(1) << "Starting replication";
   ProactorBase* mythread = ProactorBase::me();
   CHECK(mythread);
 
-  auto check_connection_error = [this, builder](error_code ec, const char* msg) -> error_code {
+  auto check_connection_error = [this](error_code ec, const char* msg) -> GenericError {
     if (!cntx_.IsRunning()) {
-      builder->SendError("replication cancelled");
-      return std::make_error_code(errc::operation_canceled);
+      return {std::make_error_code(errc::operation_canceled), "replication cancelled"};
     }
     if (ec) {
-      builder->SendError(absl::StrCat(msg, ec.message()));
       cntx_.ReportCancelError();
+      return {{}, absl::StrCat(msg, ec.message())};
     }
     return ec;
   };
@@ -127,9 +126,8 @@ error_code Replica::Start(facade::SinkReplyBuilder* builder) {
   return {};
 }
 
-void Replica::StartMainReplicationFiber(facade::SinkReplyBuilder* builder) {
+void Replica::StartMainReplicationFiber() {
   sync_fb_ = fb2::Fiber("main_replication", &Replica::MainReplicationFb, this);
-  builder->SendOk();
 }
 
 void Replica::EnableReplication(facade::SinkReplyBuilder* builder) {
@@ -137,8 +135,6 @@ void Replica::EnableReplication(facade::SinkReplyBuilder* builder) {
 
   state_mask_.store(R_ENABLED);                             // set replica state to enabled
   sync_fb_ = MakeFiber(&Replica::MainReplicationFb, this);  // call replication fiber
-
-  builder->SendOk();
 }
 
 void Replica::Stop() {
--- a/src/server/replica.cc
+++ b/src/server/replica.cc
@@ -121,7 +121,7 @@ GenericError Replica::Start() {
   VLOG(1) << "Greeting";
   state_mask_.store(R_ENABLED | R_TCP_CONNECTED);
   ec = Greet();
-  RETURN_ON_GENERIC_ERR(check_connection_error(ec, "could not greet master "));
+  RETURN_ON_ERR(check_connection_error(ec, "could not greet master "));
 
   return {};
 }
--- a/src/server/replica.cc
+++ b/src/server/replica.cc
@@ -97,7 +97,7 @@ GenericError Replica::Start() {
     }
     if (ec) {
       cntx_.ReportCancelError();
-      return {{}, absl::StrCat(msg, ec.message())};
+      return {absl::StrCat(msg, ec.message())};
     }
     return ec;
   };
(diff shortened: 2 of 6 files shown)


src/server/server_family.cc
```diff
@@ -2845,7 +2851,7 @@ void ServerFamily::ReplicaOfInternal(CmdArgList args, Transaction* tx, SinkReply
   }
 
   if (on_err == ActionOnConnectionFail::kReturnOnError) {
-    replica_->StartMainReplicationFiber(builder);
+    new_replica->StartMainReplicationFiber(builder);
```
kostasrim, line 2854: This is the FIX. Same `replica_` instance was used by two fibers and we ended up starting the main replication fiber twice. The rest of the changes above are an early exit to avoid creating the fiber all together in case of an error or if the context got cancelled.

src/server/server_family.cc
```diff
@@ -2831,10 +2831,16 @@ void ServerFamily::ReplicaOfInternal(CmdArgList args, Transaction* tx, SinkReply
...
   util::fb2::LockGuard lk(replicaof_mu_);  // Only one REPLICAOF command can run at a time
-  if (ec && replica_ == new_replica) {
-    service_.SwitchState(GlobalState::LOADING, GlobalState::ACTIVE);
-    SetMasterFlagOnAllThreads(true);
-    replica_.reset();
+  // If there was an error above during Start we must not start the main replication fiber.
+  // However, it could be the case that Start() above connected succefully and by the time
+  // we acquire the lock, the context got cancelled because another ReplicaOf command
+  // executed and acquired the replicaof_mu_ before us.
+  if (ec || new_replica->IsContextCancelled()) {
```
kostasrim, line 2838: Small optimization. We do not need to start the replication fiber if we got an error or if the context is cancelled. 

This would be handled anyway by `new_replica->StartMainReplicationFiber` which would spawn the MainReplicationFiber and then exit because it got cancelled. But yeah, no reason to do that

kostasrim:
https://github.com/dragonflydb/dragonfly/actions/workflows/regression-tests.yml

src/server/replica.h
```diff
@@ -72,6 +72,10 @@ class Replica : ProtocolClient {
 
   std::error_code TakeOver(std::string_view timeout, bool save_flag);
 
+  bool IsContextCancelled() const {
+    return !cntx_.IsRunning();
```
BorysTheDev, line 76: Do you really need such a function

src/server/server_family.cc
```diff
@@ -2831,10 +2831,16 @@ void ServerFamily::ReplicaOfInternal(CmdArgList args, Transaction* tx, SinkReply
...
+    if (replica_ == new_replica) {
```
BorysTheDev, line 2839: what if replica != new_replica for the first command and the second command was cancelled. In this case we will not have replica at all and botjh commands are failed. Maybe lets discuss it with @adiholden

kostasrim: Then we should not have a replica! The second command cancelled the previously then it cancelled itself ? Isn't that what you are describing or am I missing something ?

src/server/replica.h
kostasrim: We can just start the replication fiber and let it cancel. I wanted to avoid that alltogether so I added this. As I explained, this is not part of the fix but an "early optimization" (which is kinda funny because fiber cost is insignificant and this path is non critical)

src/server/server_family.cc
adiholden, line 2839: how can replica_ == new_replica if new_replica->IsContextCancelled()?
when we cancel replication (from replica of command)
we take replicaof mutex
call replica_->Stop
and set replica_ var
So from what I understadn we can not get to the condition replica_ == new_replica and new_replica->IsContextCancelled()

src/server/server_family.cc
```diff
@@ -2845,7 +2851,7 @@ void ServerFamily::ReplicaOfInternal(CmdArgList args, Transaction* tx, SinkReply
   }
```
adiholden, line 2851: It will be more clear I think if you enter the Drakarys call under if (on_err == ActionOnConnectionFail::kReturnOnError)instead of if (tx)

src/server/server_family.cc
adiholden: Another comment here.
I think that if replica was canceld and ec is not error you do not return response to client.

src/server/server_family.cc
kostasrim: I did not want to change the semantics. I will update on this

kostasrim: oh Drakarys needs, nvm

src/server/server_family.cc
```diff
@@ -2831,21 +2831,30 @@ void ServerFamily::ReplicaOfInternal(CmdArgList args, Transaction* tx, SinkReply
...
+  // we acquire the lock, the context got cancelled because another ReplicaOf command
+  // executed and acquired the replicaof_mu_ before us.
+  const bool cancelled = new_replica->IsContextCancelled();
+  if (ec || cancelled) {
+    if (replica_ == new_replica) {
+      service_.SwitchState(GlobalState::LOADING, GlobalState::ACTIVE);
+      SetMasterFlagOnAllThreads(true);
+      replica_.reset();
+    }
+    // SendError if we got cancelled and Start() above succeeded. Otherwise, if ec,
```
adiholden, line 2845: This is confusing. Please send the error from one place here. lets not send it in start but only here.

src/server/server_family.cc
```diff
@@ -2831,21 +2831,30 @@ void ServerFamily::ReplicaOfInternal(CmdArgList args, Transaction* tx, SinkReply
...
   // If we are called by "Replicate", tx will be null but we do not need
   // to flush anything.
-  if (tx) {
-    Drakarys(tx, DbSlice::kDbAll);
-  }
-
   if (on_err == ActionOnConnectionFail::kReturnOnError) {
-    replica_->StartMainReplicationFiber(builder);
+    Drakarys(tx, DbSlice::kDbAll);
+    new_replica->StartMainReplicationFiber(builder);
```
adiholden, line 2857: I suggest to send the ok also here and not inside StartMainReplicationFiber

src/server/replica.cc
```diff
@@ -86,19 +86,18 @@ Replica::~Replica() {
...
 static const char kConnErr[] = "could not connect to master: ";
 
-error_code Replica::Start(facade::SinkReplyBuilder* builder) {
+GenericError Replica::Start() {
   VLOG(1) << "Starting replication";
   ProactorBase* mythread = ProactorBase::me();
   CHECK(mythread);
 
-  auto check_connection_error = [this, builder](error_code ec, const char* msg) -> error_code {
+  auto check_connection_error = [this](error_code ec, const char* msg) -> GenericError {
```
adiholden, line 94: can RETURN_ON_ERR work on GenericError?

src/server/replica.cc
```diff
@@ -86,19 +86,18 @@ Replica::~Replica() {
...
     if (!cntx_.IsRunning()) {
-      builder->SendError("replication cancelled");
-      return std::make_error_code(errc::operation_canceled);
+      return {std::make_error_code(errc::operation_canceled), "replication cancelled"};
     }
     if (ec) {
-      builder->SendError(absl::StrCat(msg, ec.message()));
       cntx_.ReportCancelError();
+      return {{}, absl::StrCat(msg, ec.message())};
```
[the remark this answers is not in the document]
kostasrim: Look how `GenericError::Format` is implemented. The above expression is turned to `absl::StrCat(msg, ec.message()` when formatted.

src/server/replica.cc
kostasrim: Generic error is implicitly convertible to error code. However RETURN_ON_ERR does this:

`GenericError -> ErrorCode` then RETURN_ON_ERR returns that ErrorCode, which is then converted back to GenericError. 

This looses the `details` part of the GenericError. I will fix this. 

I am SAVING my rant around this.

src/server/replica.cc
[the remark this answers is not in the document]
kostasrim: Also we can construct `GenericError` directly from string. We can `omit` the first argument `{}` because we have that constructor. I will take care of this as well

src/server/replica.cc
```diff
@@ -86,59 +86,55 @@ Replica::~Replica() {
...
+  auto check_connection_error = [this](error_code ec, const char* msg) -> GenericError {
     if (!cntx_.IsRunning()) {
-      builder->SendError("replication cancelled");
-      return std::make_error_code(errc::operation_canceled);
+      return {std::make_error_code(errc::operation_canceled), "replication cancelled"};
     }
     if (ec) {
-      builder->SendError(absl::StrCat(msg, ec.message()));
       cntx_.ReportCancelError();
+      return {absl::StrCat(msg, ec.message())};
```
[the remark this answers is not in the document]
kostasrim: It does not do the same concatenation. 
```
330 std::string GenericError::Format() const {
331   if (!ec_ && details_.empty())
332     return "";
333 
334   if (details_.empty())
335     return ec_.message();
336   else if (!ec_)
337     return details_;
338   else
339     return absl::StrCat(ec_.message(), ": ", details_);
340 }
```

ec is `true` so `absl::StrCat(ec_.message(), ": ", details_);` is not identical to {absl::StrCat(msg, ec.message())};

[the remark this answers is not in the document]
adiholden: do we care about the order here? does this realy needs a different implementaion?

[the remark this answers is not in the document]
kostasrim: I would need to check but it's irrelevant. At the end of the day we are discussing here `return {ec, msg}` vs `return {msg}`. Either way is fine

src/server/error.h
```diff
@@ -31,6 +31,14 @@ using facade::kWrongTypeErr;
 
 #define RETURN_ON_ERR(x) RETURN_ON_ERR_T(std::error_code, x)
 
+#define RETURN_ON_GENERIC_ERR(x)                                   \
```
kostasrim, line 34: @adiholden 

I had to, there is no really good solution otherwise. Doing something `RETURN_ON_ERR_T(GenericError, x)` won't work. Yes I tried modifying the micro above by converting the statement: 

`std::error_code __ec = (x); ` ----> `T __ex = (x)` but a) this is not properly being parsed and we get compiler errors, we also do an extra copy in that expression which is fine for `error` code because it's `trivially copyable`. Generic error is not.

PR #4661, opened 2025-02-25
chore: snapshot/tiering improvements

romange:
Mainly comments and refactorings.

There are two functional differrences:
1. flush serialized entries in case we gathered at least K delayed
entries coming from tiered entities.
2. allow loading snapshots larger than memory for tiered enabled datastores.

Also pull latest helio with some bug fixes in gcs code.

### DIFF
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -1915,7 +1915,7 @@ struct RdbLoader::ObjSettings {
   bool has_mc_flags = false;
 
   void Reset() {
-    expiretime = 0;
+    mc_flags = expiretime = 0;
     has_expired = false;
     is_sticky = false;
     has_mc_flags = false;
@@ -1958,6 +1958,9 @@ RdbLoader::~RdbLoader() {
 
 error_code RdbLoader::Load(io::Source* src) {
   CHECK(!src_ && src);
+
+  is_tiered_enabled_ =
+      shard_set->Await(0, [] { return EngineShard::tlocal()->tiered_storage() != nullptr; });
 
   absl::Time start = absl::Now();
   src_ = src;
@@ -2245,7 +2248,7 @@ error_code RdbLoaderBase::EnsureReadInternal(size_t min_to_read) {
   }
   if (*res < min_sz)
     return RdbError(errc::rdb_file_corrupted);
-
+  DVLOG(2) << "EnsureRead " << *res << " bytes";
   bytes_read_ += *res;
 
   DCHECK_LE(bytes_read_, source_limit_);
@@ -2341,6 +2344,7 @@ error_code RdbLoaderBase::SkipModuleData() {
 }
 
 error_code RdbLoaderBase::HandleCompressedBlob(int op_type) {
+  DVLOG(2) << "HandleCompressedBlob: " << op_type;
   RETURN_ON_ERR(AllocateDecompressOnce(op_type));
 
   // Fetch uncompress blob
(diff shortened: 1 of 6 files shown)


src/server/snapshot.h
```diff
@@ -89,6 +89,20 @@ class SliceSnapshot {
     snapshot_fb_.JoinIfNeeded();
   }
 
+  uint64_t snapshot_version() const {
```
romange, line 92: just moved here, no changes

src/server/snapshot.cc
```diff
@@ -309,18 +309,11 @@ void SliceSnapshot::SerializeEntry(DbIndex db_indx, const PrimeKey& pk, const Pr
...
-    // We can't block, so we just schedule a tiered read and append it to the delayed entries
-    util::fb2::Future<PrimeValue> future;
-    EngineShard::tlocal()->tiered_storage()->Read(
-        db_indx, pk.ToString(), pv,
-        [future](const std::string& v) mutable { future.Resolve(PrimeValue(v)); });
-    delayed_entries_.push_back(
-        {db_indx, PrimeKey(pk.ToString()), std::move(future), expire_time, mc_flags});
-    ++type_freq_map_[RDB_TYPE_STRING];
+    // TODO: we loose the stickiness attribute bly cloning like this PrimeKey.
+    SerializeExternal(db_indx, PrimeKey{pk.ToString()}, pv, expire_time, mc_flags);
```
romange, line 316: code moved into `SerializeExternal`

src/server/rdb_load.cc
```diff
@@ -1957,6 +1957,9 @@ RdbLoader::~RdbLoader() {
 error_code RdbLoader::Load(io::Source* src) {
   CHECK(!src_ && src);
 
+  is_tiered_enabled_ =
+      shard_set->Await(0, [] { return EngineShard::tlocal()->tiered_storage() != nullptr; });
```
kostasrim, line 1961: Don't we have a flag for tiering, why not use that instead? Furthermore, I know we can start df with less shards than proactors. What I don't remember is if the shards start at index 0 or no for that case (where we have more proactors than shards). If they don't, then this is a bug here.

src/server/snapshot.cc
```diff
@@ -309,18 +309,11 @@ void SliceSnapshot::SerializeEntry(DbIndex db_indx, const PrimeKey& pk, const Pr
...
   if (pv.IsExternal()) {
-    // We can't block, so we just schedule a tiered read and append it to the delayed entries
-    util::fb2::Future<PrimeValue> future;
-    EngineShard::tlocal()->tiered_storage()->Read(
-        db_indx, pk.ToString(), pv,
-        [future](const std::string& v) mutable { future.Resolve(PrimeValue(v)); });
-    delayed_entries_.push_back(
-        {db_indx, PrimeKey(pk.ToString()), std::move(future), expire_time, mc_flags});
-    ++type_freq_map_[RDB_TYPE_STRING];
+    // TODO: we loose the stickiness attribute bly cloning like this PrimeKey.
```
kostasrim, line 315:
```suggestion
    // TODO: we loose the stickiness attribute by cloning like this PrimeKey.
```

src/server/snapshot.cc
```diff
@@ -372,7 +365,16 @@ bool SliceSnapshot::PushSerialized(bool force) {
...
     do {
       auto& entry = delayed_entries_.back();
-      serializer_->SaveEntry(entry.key, entry.value.Get(), entry.expire, entry.dbid,
+
+      // TODO: https://github.com/dragonflydb/dragonfly/issues/4654
+      // there are a few problems with how we serialize external values.
+      // 1. We may block here too frequently, slowing down the process.
+      // 2. For small bin values, we issue multiple reads for the same page, creating
+      //    read factor amplification that can reach factor of ~60.
+      PrimeValue pv{entry.value.Get()};  // Migh block until the future resolves.
```
kostasrim, line 374:
```suggestion
      PrimeValue pv{entry.value.Get()};  // Might block until the future resolves.
```

kostasrim (approved):
lgtm

PR #4664, opened 2025-02-26
Support Cluster configuration for dfly_bench

romange:
The load tester client opens a connection to each shard, in total `num_shards * FLAGS_c * FLAGS_proactor_threads` connections.

There is no support for MOVED responses at the moment so only static clusters are supported.

### DIFF
--- a/src/server/dfly_bench.cc
+++ b/src/server/dfly_bench.cc
@@ -2,6 +2,11 @@
 // See LICENSE for licensing terms.
 //
 
+extern "C" {
+#include "redis/crc16.h"
+}
+
+#include <absl/container/flat_hash_set.h>
 #include <absl/random/random.h>
 #include <absl/strings/match.h>
 #include <absl/strings/str_cat.h>
@@ -9,6 +14,7 @@
 #include <absl/strings/str_split.h>
 
 #include <queue>
+#include <tuple>
 
 #include "absl/time/clock.h"
 #include "absl/time/time.h"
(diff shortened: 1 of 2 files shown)


src/server/dfly_bench.cc
```diff
@@ -391,12 +417,35 @@ string KeyGenerator::operator()() {
...
 
-  return StrCat(prefix_, key_suffix);
+void KeyGenerator::EnableClusterMode() {
+  hash_slots_.resize(kNumSlots);
+  uint32_t i = 0;
+  uint32_t num_slots_filled = 0;
+
+  // Precompute the hash slots for each of the slot ids so given the slot id
+  // we could generate a key that belongs to that slot.
+  while (num_slots_filled < kNumSlots) {
```
kostasrim, line 435: I thought we would leave some empty cells but then I saw that we increment the num_slots_filled only when hash slot is non empty so at the end we end up with `precomputed keys` for all hash slots. Noice!

src/server/dfly_bench.cc
```diff
@@ -724,6 +802,82 @@ void WatchFiber(atomic_bool* finish_signal, ProactorPool* pp) {
   }
 }
 
+ClusterSpec FetchCluster(const tcp::endpoint& ep, ProactorBase* proactor) {
```
kostasrim, line 805: I went over this function briefly but I *did not* do a thorough review so I can't really guarantee the correctness of it. Moreover, for changes like this it's nice to have tests. And speaking of that do we have tests for `dfly_bench` ?

kostasrim (approved):
I do have some faith that the function `FetchCluster` does the right thing. I can thoroughly review it but I will need to pull the branch locally and play around -- I am not familiar with dfly_bench that much. 

I would also like to see dfly_bench tests. Maybe it's an overkill though :shrug: 

Lastly, happy to pull your branch and play around if you want me to. I did not because dfly_bench is not "core".

src/server/dfly_bench.cc
romange: we do not - it's not part of a production stack

PR #4692, opened 2025-03-03
chore: rename functions/classes in db_slice code

romange:
1. Rename LocalBlockingCounter to LocalLatch because it's a variation of latch (see std::latch for example).
2. Rename PreUpdate to PreUpdateBlocking to emphasize it can block.
3. Fix mutations counting: consider either insertions or changing the existing entry. Before that we incremented this counter for misses as well.


src/server/common.cc
```diff
@@ -481,15 +481,15 @@ BorrowedInterpreter::~BorrowedInterpreter() {
     ServerState::tlocal()->ReturnInterpreter(interpreter_);
 }
 
-void LocalBlockingCounter::unlock() {
-  DCHECK(mutating_ > 0);
+void LocalLatch::unlock() {
```
kostasrim, line 484: I like `LocalLatch` :)

src/server/db_slice.cc
```diff
@@ -1449,13 +1427,29 @@ pair<uint64_t, size_t> DbSlice::FreeMemWithEvictionStep(DbIndex db_ind, size_t s
...
           // returns when whichever condition is met first
           if ((evicted_items == max_eviction_per_hb) || (evicted_bytes >= increase_goal_bytes))
-            return return_cb();
+            goto finish;
         }
       }
     }
+  }  // FiberAtomicGuard
+
+finish:
```
kostasrim, line 1436: I replaced this `goto` in the past :face_with_diagonal_mouth: I wonder if we could somehow still get rid of it but with this double for-loop seems like inevitable

kostasrim (approved):
LGTM

src/server/db_slice.cc
romange: There is nothing wrong in goto statements, it's not a taboo. But we should not use them if there are clearer tools.

adiholden (approved):

PR #4697, opened 2025-03-04
fix: ZMSCORE return value if key does not exist

### LINKED ISSUE #4670: Incorrect reply for ZMSCORE if sorted set does not exist
**Description**
When using `ZMSCORE`, the command should behave the same whether the value can't be found because the member does not exist in the sorted set or *the sorted set does not exist at all*.

But, if the sorted set does not exist, the `ZMSCORE` command returns an empty response.

**To Reproduce**
Execute the the `ZMSCORE abc x` (single member) or `ZMSCORE abc x y z` (multiple members) command. Where `abc` is any non-existent key.

**Expected behavior**
* For a single member: A nil response.
* For multiple members: An array of nil responses with one entry per member.

**Actual behavior**
An empty response is returned.

**Environment**
 - Containerized?: Docker (Simply using `docker.dragonflydb.io/dragonflydb/dragonfly:latest`)
 - Dragonfly Version: 1.27.1

vyavdoshenko:
fix: ZMSCORE return value if key does not exist #4670

### DIFF
--- a/src/server/zset_family.cc
+++ b/src/server/zset_family.cc
@@ -1363,11 +1363,7 @@ OpResult<MScoreResponse> OpMScore(const OpArgs& op_args, string_view key,
 
   // If the key doesn't exist and there's only one member, return KEY_NOTFOUND
   if (res_it.status() == OpStatus::KEY_NOTFOUND) {
-    if (members.Size() == 1) {
-      return OpStatus::KEY_NOTFOUND;
-    }
-
-    // If the key doesn't exist and there are multiple members, return an array of NIL values
+    // If the key doesn't exist return an array of NIL values
     MScoreResponse result(members.Size(), std::nullopt);
     return result;
   }
@@ -2663,10 +2659,6 @@ void ZSetFamily::ZMScore(CmdArgList args, const CommandContext& cmd_cntx) {
   auto* rb = static_cast<RedisReplyBuilder*>(cmd_cntx.rb);
 
   OpResult<MScoreResponse> result = ZGetMembers(args, cmd_cntx.tx, rb);
-
-  if (result.status() == OpStatus::KEY_NOTFOUND) {
-    return rb->SendNull();
-  }
 
   if (result.status() == OpStatus::WRONG_TYPE) {
     return rb->SendError(kWrongTypeErr);
--- a/src/server/zset_family.cc
+++ b/src/server/zset_family.cc
@@ -1360,6 +1360,18 @@ OpResult<unsigned> OpRem(const OpArgs& op_args, string_view key, facade::ArgRang
 OpResult<MScoreResponse> OpMScore(const OpArgs& op_args, string_view key,
                                   facade::ArgRange members) {
   auto res_it = op_args.GetDbSlice().FindReadOnly(op_args.db_cntx, key, OBJ_ZSET);
+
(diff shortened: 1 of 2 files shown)


src/server/zset_family.cc
```diff
@@ -1359,6 +1359,18 @@ OpResult<unsigned> OpRem(const OpArgs& op_args, string_view key, facade::ArgRang
 OpResult<MScoreResponse> OpMScore(const OpArgs& op_args, string_view key,
                                   facade::ArgRange members) {
   auto res_it = op_args.GetDbSlice().FindReadOnly(op_args.db_cntx, key, OBJ_ZSET);
+
+  // If the key doesn't exist and there's only one member, return KEY_NOTFOUND
+  if (res_it.status() == OpStatus::KEY_NOTFOUND) {
+    if (members.Size() == 1) {
+      return OpStatus::KEY_NOTFOUND;
+    }
+
```
kostasrim, line 1368: The OP in the ISSUE is WRONG. We always return an array even for a single element.

src/server/zset_family_test.cc
```diff
@@ -287,6 +287,28 @@ TEST_F(ZSetFamilyTest, ZMScore) {
   EXPECT_THAT(resp.GetVec(), ElementsAre("42", "3.14", ArgType(RespExpr::NIL)));
 }
 
+// Test for ZMSCORE with a single member on a non-existent key
+TEST_F(ZSetFamilyTest, ZMScoreSingleMemberNonExistentKey) {
```
kostasrim, line 291: You can squash the two tests now. We always return an array

src/server/zset_family_test.cc
```diff
@@ -287,6 +287,28 @@ TEST_F(ZSetFamilyTest, ZMScore) {
...
+  EXPECT_THAT(resp, ArgType(RespExpr::NIL))
+      << "Expected NIL response for non-existent key with single member";
+}
+
+// Test for ZMSCORE with multiple members on a non-existent key
+TEST_F(ZSetFamilyTest, ZMScoreMultipleMembersNonExistentKey) {
+  // Case 2: Multiple members with non-existent key (ZMSCORE abc x y z)
+  auto resp = Run({"zmscore", "abc", "x", "y", "z"});
+
+  ASSERT_EQ(RespExpr::ARRAY, resp.type)
```
adiholden, line 304: you can write RespArray(ElementsAre(ArgType(RespExpr::NIL), ArgType(RespExpr::NIL),ArgType(RespExpr::NIL))
instead of the 3 checks below

src/server/zset_family.cc
kostasrim: Screenshot from 2025-03-05 08-41-20

Screenshot from 2025-03-05 08-41-30

kostasrim: First pic is df with your changes, second is valkey/redis

adiholden:
Nit: I would make it explicit in the PR title that the fix is in the return value
i.e : fix: ZMSCORE return value if key does not exist

src/server/zset_family_test.cc
vyavdoshenko: Fixed.

src/server/zset_family.cc
vyavdoshenko: Fixed.

kostasrim:
:ship: :it: :man_cook:

PR #4702, opened 2025-03-05
feat(server): Add PUBSUB SHARDCHANNELS/SHARDNUMSUB

mkaruza:


tests/dragonfly/cluster_test.py
```diff
@@ -2982,6 +2982,64 @@ async def test_cluster_sharded_pub_sub(df_factory: DflyInstanceFactory):
...
+        await c_nodes[0].execute_command("PUBSUB CHANNELS")
+    assert str(moved_error.value) == f"PUBSUB CHANNELS is not supported in cluster mode yet"
+
+    with pytest.raises(redis.exceptions.ResponseError) as moved_error:
+        await c_nodes[0].execute_command("PUBSUB NUMSUB")
+    assert str(moved_error.value) == f"PUBSUB NUMSUB is not supported in cluster mode yet"
+
+    with pytest.raises(redis.exceptions.ResponseError) as moved_error:
+        await c_nodes[0].execute_command("PUBSUB NUMPAT")
+    assert str(moved_error.value) == f"PUBSUB NUMPAT is not supported in cluster mode yet"
```
kostasrim, line 3040: These are somewhat redundant. The unit test your introduced above suffice

tests/dragonfly/cluster_test.py
```diff
@@ -2982,6 +2982,64 @@ async def test_cluster_sharded_pub_sub(df_factory: DflyInstanceFactory):
...
+    # Consume subscription message result from above
+    message = consumer.get_sharded_message(target_node=node_a)
+    assert message == {
+        "type": "subscribe",
+        "pattern": None,
+        "channel": b"pubsub-shard-channel",
+        "data": 1,
+    }
+
+    message = await c_nodes[0].execute_command("PUBSUB SHARDCHANNELS")
```
kostasrim, line 3016: nit: consider adding a few more channels, then check that they are printed here. Use a filter below, to check that we do actually filter and apply pattern correctly (which existed before and it should be correct but why not :smile: )

tests/dragonfly/cluster_test.py
```diff
@@ -2982,6 +2982,64 @@ async def test_cluster_sharded_pub_sub(df_factory: DflyInstanceFactory):
...
+    await c_nodes[0].execute_command("SPUBLISH pubsub-shard-channel 1")
+
+    # Consume subscription message result from above
+    message = consumer.get_sharded_message(target_node=node_a)
+    assert message == {
+        "type": "subscribe",
+        "pattern": None,
+        "channel": b"pubsub-shard-channel",
+        "data": 1,
+    }
```
kostasrim, line 3014: you can remove this as well, the commands are only affecting `subscribers` so the publish part is not "interesting for the test"

kostasrim (commented):
First PR! Congratz :man_cook: 

Left some minor comments.

Also, plz write a short description in the PR and reference the issue that you are closing by this PR. 

For reference see https://github.com/dragonflydb/dragonfly/pull/4696#issue-2894527929

src/server/main_service.cc
```diff
@@ -2362,9 +2367,6 @@ void Service::Monitor(CmdArgList args, const CommandContext& cmd_cntx) {
 void Service::Pubsub(CmdArgList args, const CommandContext& cmd_cntx) {
   auto* rb = static_cast<RedisReplyBuilder*>(cmd_cntx.rb);
 
-  if (IsClusterEnabled()) {
```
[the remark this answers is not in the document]
kostasrim: No not all, cause `PUBSUB` is a family (it has subcommands). So for example https://redis.io/docs/latest/commands/pubsub-shardchannels/#:~:text=An%20active%20shard%20channel%20is,glob%2Dstyle%20pattern%20are%20listed. 

`PUBSUB SHARDCHANNELS` is fine, since shard channels are a notion of `cluster`.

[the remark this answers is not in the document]
kostasrim: However, `PUBSUB CHANNELS` should be rejected (until we have global pub sub available)

src/server/cluster/cluster_family_test.cc
```diff
@@ -625,7 +625,12 @@ TEST_F(ClusterFamilyTest, ClusterModePubSubNotAllowed) {
               ErrArg("PSUBSCRIBE is not supported in cluster mode yet"));
   EXPECT_THAT(Run({"PUNSUBSCRIBE", "ch?"}),
               ErrArg("PUNSUBSCRIBE is not supported in cluster mode yet"));
-  EXPECT_THAT(Run({"PUBSUB", "CHANNELS"}), ErrArg("PUBSUB is not supported in cluster mode yet"));
+  EXPECT_THAT(Run({"PUBSUB", "CHANNELS"}),
```
kostasrim, line 628: Add the opposite checks for non cluster mode. That is `PUBSUB SHARDCHANNELS` gets rejected in `non cluster mode`

tests/dragonfly/cluster_test.py
kostasrim: Oh I miss read this. Keep these, they are usefull

kostasrim: They check that `CHANNELS, NUMBSUB, NUMPAT` is not available in cluster. 

Great!

src/server/cluster/cluster_family_test.cc
kostasrim: Ignore me, I missed that you included those. :100:

src/server/main_service.cc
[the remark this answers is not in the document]
adiholden: Ok, I understand. And the implementaion is just the same..
So if on cluster we allow all commands that subcommand start with "SHARD",
then I think it will be more readable to write the condition
if IsClusterEnabled() && !absl::StartsWith(subcmd, "SHARD")  in this function
and remove the bool reject_cluster

src/server/cluster/cluster_family_test.cc
mkaruza: Using SHARDCHANNELS and SHARDNUMSUB in non cluster mode will work, i'll add this check and update tests

src/server/main_service.cc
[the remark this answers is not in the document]
mkaruza: @adiholden thank you for suggestion, but this is problematic because it just checks SHARD prefix of subcommand and if that command is not valid it will still report error instead of returning warning about wrong subcommand (also, per @kostasrim review we need to cover situation when SHARD subcommands is used in non cluster environment).

kostasrim (dismissed):

src/server/main_service.cc
```diff
@@ -2382,23 +2377,45 @@ void Service::Pubsub(CmdArgList args, const CommandContext& cmd_cntx) {
...
         "\tPrints this help."};
 
     rb->SendSimpleStrArr(help_arr);
     return;
   }
 
-  if (subcmd == "CHANNELS") {
+  auto cmd_err = [&]() {
+    auto err = absl::StrCat("PUBSUB ", subcmd, " is not supported",
+                            IsClusterEnabled() ? " in cluster mode yet" : " in non cluster mode");
```
adiholden, line 2395: @kostasrim I think that the check should be IsClusterEnabledOrEmulated() right?
cause we should enable cluster commands in emulated mode

adiholden: @mkaruza Dragonfly has emulated cluster mode which runs a single shard and in this state cluster and non cluster client are allowed.
This means that we should allow global pubsub command to run both on non cluster and emulated mode and sharded pubsub commands to run on both cluster and emulated mode, hence the check to send the error below should be fixed

src/server/server_family_test.cc
```diff
@@ -547,4 +547,18 @@ TEST_F(ServerFamilyTest, CommandDocsOk) {
   EXPECT_THAT(Run({"command", "docs"}), "OK");
 }
 
+TEST_F(ServerFamilyTest, PubSubCommandErr) {
+  auto* flag = absl::FindCommandLineFlag("cluster_mode");
```
adiholden, line 551: if (auto cluster_mode = absl::GetFlag(FLAGS_cluster_mode); cluster_mode == "")

adiholden (approved):

adiholden:
@mkaruza I can see that  HllFamilyTest.Promote fail on the CI but its not related to your changes.
Please open a ticket with the link to this CI run so we  will be able to follow up on this fail and after this you can rerun the action so you will be able to merge the PR

Issue #4625, opened 2025-02-19
Hash tables improvements


BorysTheDev:
We have at least 2 pointers in our hash tables to key and value. Even if we ignore possible allocator overhead it's 16 bytes overhead per pair. Also when we read them it's 2 accesses to cold memory.

Improvement:
If the key is less than K bytes or a value is fixed size we can store a key and a value together in one chunk of memory. If a key is greater K bytes we store a key with a pointer to a value.
In most cases, a key is quite small, and copying a key is cheaper than cold memory access, also overhead will be at least 8 bytes less per pair.

1) Further improvement: open addressing hash table with linear probing for the next N elements.
2) Further improvement: extension point every Nth element to avoid cold memory access
3) Further improvement: use ptr_vector instead of list (store the size of the vector in the pointer)
4) Further improvement: branchless execution for the fast path.
5) Further improvement: interchanging calculations to utilize superscalar architecture with a branchless approach and SIMD (find empty bucket, checking expired elements, find already exited key etc) 
6) Further improvement: huge operations optimizations (prefetching, bunch processing).

BPTree can also be optimized in the same way.

kostasrim:
Do you mean the DashTable or ?

BorysTheDev:
I mean all hash tables and DashTable too

romange:
I do not think that DashTable is relevant here, it has a completely differrent design.

BorysTheDev:
I think the simplest option to test this approach is StringSet

PR #4705, opened 2025-03-05
feat: add intrusive string

BorysTheDev:
https://github.com/dragonflydb/dragonfly/issues/4625


src/core/oah_entry.h
```diff
@@ -0,0 +1,433 @@
...
+static uint32_t BucketId(uint64_t hash, uint32_t capacity_log) {
+  assert(capacity_log > 0);
+  return hash >> (64 - capacity_log);
+}
+// doesn't possess memory, it should be created and release manually
+class OAHEntry {
+  // we can assume that high 12 bits of user address space
+  // can be used for tagging. At most 52 bits of address are reserved for
+  // some configurations, and usually it's 48 bits.
+  // https://docs.kernel.org/arch/arm64/memory.html
```
vyavdoshenko, line 141: Is that true for other hardware platforms? x86_64?
I guess this approach is useful (or even possible) for Linux only? I guess yes, so this is the performance update for Linux only. We have to have a general approach always to support cross-platform builds.

I think the build system should be updated to integrate this approach, to use it for Linux only.

src/core/oah_entry.h
```diff
@@ -0,0 +1,433 @@
...
+  // TODO add initializer list constructor
+  OAHEntry(size_t vector_size) {
+    // TODO rewrite to simple array
+    data_ = reinterpret_cast<char*>(new std::vector<OAHEntry>(vector_size));
+    SetVectorBit();
+  }
+
+  OAHEntry(const OAHEntry& e) = delete;
+  OAHEntry(OAHEntry&& e) {
+    data_ = e.data_;
```
vyavdoshenko, line 168: This is a memory leak because all members are created before the ctor, even this one. This member initiates with zmalloc. The memory will be lost each time.

src/core/oah_entry.h
```diff
@@ -0,0 +1,433 @@
...
+    DCHECK_EQ(BucketId(Hash(Key()), new_capacity_log), new_bucket_id);
+
+    return new_bucket_id;
+  }
+
+  void SetExpiry(uint32_t at_sec) {
+    DCHECK(!IsVector());
+    if (HasExpiry()) {
+      auto* expiry_pos = Raw();
+      std::memcpy(expiry_pos, &at_sec, sizeof(at_sec));
```
vyavdoshenko: uint32_t* expiry_pos = Raw();
*expiry_pos = at_sec;

src/core/oah_entry.h
```diff
@@ -0,0 +1,433 @@
...
+  inline void Clear() {
+    // TODO add optimization to avoid destructor calls during vector allocator
+    if (!data_)
+      return;
+
+    if (IsVector()) {
+      delete &AsVector();
+    } else {
+      zfree(Raw());
+    }
```
vyavdoshenko: Correct me, if I am wrong, but data_ always initializes with zmalloc, but not always frees with zfree.
This code looks dangerous.

src/core/oah_entry.h
BorysTheDev: I didn't get your point. It's move ctor. We have only one member
 `char* data_ = nullptr;` 
we move data from another object and we don't need to remove anything

src/core/oah_entry.h
BorysTheDev: with reinterpret_cast it will work. thx

src/core/oah_entry.h
BorysTheDev: data_ can be vector whisch initializes with new or pointer to string which initializes with zmalloc. It looks dangerous, but it is ok

src/core/oah_entry.h
vyavdoshenko: This member is initialized with memory. Please take a look at row 61.
data_ = (char*)zmalloc(size);

vyavdoshenko: Ahh, this is another ctor. Please don't mind

src/core/oah_entry.h
BorysTheDev: It's a crossplatform approach

src/core/oah_entry.h
```diff
@@ -0,0 +1,528 @@
...
+    return new_bucket_id;
+  }
+
+  void SetExpiry(uint32_t at_sec) {
+    DCHECK(!IsVector());
+    if (HasExpiry()) {
+      auto* expiry_pos = Raw();
+      std::memcpy(expiry_pos, &at_sec, sizeof(at_sec));
+    } else {
+      *this = OAHEntry(Key(), at_sec);
```
vyavdoshenko: Memory leak?
```
    OAHEntry tmp(Key(), at_sec);
    std::swap(*this, tmp);
```

src/core/oah_entry.h
```diff
@@ -0,0 +1,528 @@
...
+  // some configurations, and usually it's 48 bits.
+  // https://docs.kernel.org/arch/arm64/memory.html
+  // first 3 bits aren't used by pointer
+  static constexpr size_t kVectorBit = 1ULL << 0;
+  static constexpr size_t kExpiryBit = 1ULL << 1;
+  // if bit is set the string length field is 1 byte instead of 4
+  static constexpr size_t kSsoBit = 1ULL << 2;
+
+  // extended hash allows us to reduce keys comparisons
+  static constexpr size_t kExtHashShift = 52;
```
vyavdoshenko, line 149: Your assumption that a pointer is 52 bits on 64-bit platforms. However, there are several mentions that modern 64-bit platforms can use up to 57 bits.
For example: https://www.amd.com/content/dam/amd/en/documents/products/epyc/4th-gen-epyc-processor-architecture-white-paper.pdf
```
LARGER ADDRESSABLE MEMORY
With the potential to expand memory through CXL controllers,
virtual memory is now addressable through 57 bits, and a fifth level
of nested page tables has been implemented to support this.
```
It can be used especially in a virtual machine environment.

BorysTheDev: In this case, we should fix our DenseSet too, because I took this from it.

BorysTheDev: As I understand now CXL is used for ASICs,

src/core/oah_entry.h
BorysTheDev: I don't think so, I do absolutely the same 
```
 OAHEntry& operator=(OAHEntry&& e) {
    std::swap(data_, e.data_);
    return *this;
  }
```

romange:
how memory usage is compared? say you add K 8-byte members to both implementations for K=1000000,1500000,2000000,2500000,3000000?

romange:
On average i see 10-15% improvement in speed,right?

BorysTheDev:
> On average i see 10-15% improvement in speed,right?

For now yes, but we can still do a lot of optimizations.
Also we can tune it to significantly improve some operations (reducing some others or insignificantly increasing memory consumption), for example, Grow can be two times faster if we increase N or increase memory consumption.

BorysTheDev:
> How is memory usage compared? Say you add K 8-byte members to both implementations for K=1000000,1500000,2000000,2500000,3000000?

I don't know if I got the question. I didn't measure memory consumption. Theoretically, if we compare with StringSet, the current implementation is quite similar in memory consumption (I expect 1 byte better per entry, but I didn't check).

src/core/oah_entry.h
```diff
@@ -0,0 +1,528 @@
...
+  }
+
+  bool IsEntry() const {
+    return (data_ != 0) && !(data_ & kVectorBit);
+  }
+
+  PtrVector<OAHEntry>& AsVector() {
+    static_assert(sizeof(PtrVector<OAHEntry>) == sizeof(uint64_t));
+    return *reinterpret_cast<PtrVector<OAHEntry>*>(&data_);
+  }
```
dranikpg, line 202: It seems like you can just create a PtrVector<OAHEntry> out of the data value

src/core/oah_entry.h
```diff
@@ -0,0 +1,528 @@
...
+
+  uint32_t GetKeySize() const {
+    if (HasSso()) {
+      uint8_t size = 0;
+      std::memcpy(&size, Raw() + GetExpirySize(), sizeof(size));
+      return size;
+    }
+    uint32_t size = 0;
+    std::memcpy(&size, Raw() + GetExpirySize(), sizeof(size));
```
dranikpg: memcpy because you are not sure about alignment? (you chose the only order that makes it good)

dranikpg: And it's the preferred way, especially when c++20 bitcast does the same 🤷🏻

src/core/oah_entry.h
```diff
@@ -0,0 +1,528 @@
...
+  }
+
+  uint64_t Release() {
+    uint64_t res = uptr_;
+    uptr_ = 0;
+    return res;
+  }
+
+  void ResizeLog(uint64_t new_log_size) {
+    auto new_ptr = reinterpret_cast<T*>(zmalloc(sizeof(T) << new_log_size));
```
dranikpg, line 77: zmalloc -> i.e. our mimalloc always guarantees align of 16?

src/core/oah_entry.h
```diff
@@ -0,0 +1,528 @@
...
+#include "redis/zmalloc.h"
+}
+
+namespace dfly {
+
+#define PREFETCH_READ(x) __builtin_prefetch(x, 0, 1)
+#define FORCE_INLINE __attribute__((always_inline))
+
+// TODO add allocator support
+template <class T> class PtrVector {
```
dranikpg, line 32: it's templated only to be used from OAHentry with  itself, right?

BorysTheDev:
I've run benchmarks on AWS c7g.16xlarge (arm64), and the results are even better than on my local CPU

| Benchmark | CPU OAHSet | CPU StringSet |
|--------|--------|--------|
| BM_Clone/elements:32000                  |    1330884 ns  |      2685591 ns     |  
| BM_Fill/elements:32000                   |    1638128 ns  |      2026346 ns     |  
| BM_Clear/elements:32000                  |     611710 ns  |      1152499 ns     |  
| BM_Add/elements:1000/Key Size:10         |      35535 ns  |        61020 ns     |  
| BM_Add/elements:10000/Key Size:10        |     370244 ns  |       543129 ns     |  
| BM_Add/elements:100000/Key Size:10       |    4995692 ns  |      7673177 ns     |  
| BM_Add/elements:1000/Key Size:100        |      54810 ns  |        78149 ns     |  
| BM_Add/elements:10000/Key Size:100       |     537805 ns  |       745326 ns     |  
| BM_Add/elements:100000/Key Size:100      |    6709688 ns  |     10060040 ns     |  
| BM_Add/elements:1000/Key Size:1000       |     130500 ns  |       168247 ns     |  
| BM_Add/elements:10000/Key Size:1000      |    1313826 ns  |      1683678 ns     |  
| BM_Add/elements:100000/Key Size:1000     |   18742027 ns  |     26859559 ns     |  
| BM_AddMany/elements:1000/Key Size:10     |      46076 ns  |        59239 ns     |  
| BM_AddMany/elements:10000/Key Size:10    |     451744 ns  |       531307 ns     |  
| BM_AddMany/elements:100000/Key Size:10   |    5792388 ns  |      7514034 ns     |  
| BM_AddMany/elements:1000/Key Size:100    |      62141 ns  |        74747 ns     |  
| BM_AddMany/elements:10000/Key Size:100   |     613032 ns  |       721393 ns     |  
| BM_AddMany/elements:100000/Key Size:100  |    7491017 ns  |      9748602 ns     |  
| BM_AddMany/elements:1000/Key Size:1000   |     135705 ns  |       167385 ns     |  
| BM_AddMany/elements:10000/Key Size:1000  |    1392936 ns  |      1615755 ns     |  
| BM_AddMany/elements:100000/Key Size:1000 |   19511305 ns  |     27544565 ns     |  
| BM_Erase/elements:1000/Key Size:10       |      38065 ns  |        47403 ns     |  
| BM_Erase/elements:10000/Key Size:10      |     594035 ns  |       427876 ns     |  
| BM_Erase/elements:100000/Key Size:10     |    6732755 ns  |      6151252 ns     |  
| BM_Erase/elements:1000/Key Size:100      |      52533 ns  |        66609 ns     |  
| BM_Erase/elements:10000/Key Size:100     |     810278 ns  |       643084 ns     |  
| BM_Erase/elements:100000/Key Size:100    |    9370130 ns  |      8724927 ns     |  
| BM_Erase/elements:1000/Key Size:1000     |     147916 ns  |       149872 ns     |  
| BM_Erase/elements:10000/Key Size:1000    |    1532298 ns  |      1496420 ns     |  
| BM_Erase/elements:100000/Key Size:1000   |   19927403 ns  |     19269993 ns     |  
| BM_Get/elements:1000/Key Size:10         |      14795 ns  |        26821 ns     |  
| BM_Get/elements:10000/Key Size:10        |     183910 ns  |       256741 ns     |  
| BM_Get/elements:100000/Key Size:10       |    3065545 ns  |      4401379 ns     |  
| BM_Get/elements:1000/Key Size:100        |      35506 ns  |        47468 ns     |  
| BM_Get/elements:10000/Key Size:100       |     381048 ns  |       498173 ns     |  
| BM_Get/elements:100000/Key Size:100      |    5584236 ns  |      7877216 ns     |  
| BM_Get/elements:1000/Key Size:1000       |     103973 ns  |       119229 ns     |  
| BM_Get/elements:10000/Key Size:1000      |    1045459 ns  |      1276052 ns     |  
| BM_Get/elements:100000/Key Size:1000     |   12887455 ns  |     19505085 ns     |  
| BM_Grow                                  |   10429151 ns  |     14235076 ns     |

src/core/oah_entry.h
BorysTheDev: for now yes, but I think of reusing it. Also for now it is only StringSet support, after we exchange StringSet to OAHSet I plan to implement OAHEntry for Hashmap

src/core/oah_entry.h
BorysTheDev: All allocators should guarantee 16 for x64, but I need only 8

src/core/oah_entry.h
BorysTheDev: but I need to return the reference. I know how to refactor it, just don;t have time

src/core/oah_entry.h
BorysTheDev: Honestly we don't need memcpy here and in some other places. I didn't have time to optimize the code, so I tried to write the code at least without the problems

romange:
How memory usage looks like?, btw, can you provide a third column for benchmarks showing improvement in percents as it's hard to see by eye which one improved and by how much.

BorysTheDev:
> How memory usage looks like?, btw, can you provide a third column for benchmarks showing improvement in percents as it's hard to see by eye which one improved and by how much.

I didn't compare memory usage, but it should be less than or equal to before (I assumed 3 bits less per entry). Regarding percents, there is no sense in them until we discuss what is our main priority in hash table. For now, I have implemented it to be faster across all operations, but it could be tuned to consume 1-2 bytes less memory per entry than StringSet, or we can improve Grow operation performance by 2x or more. So there is no sense in measurement until the way is seen.

src/core/oah_entry.h
romange: it's incorrect.
see https://github.com/jemalloc/jemalloc/issues/1533#issuecomment-507915829
 for example.

src/core/oah_entry.h
```diff
@@ -0,0 +1,533 @@
...
+//
+
+#pragma once
+
+#include <cassert>
+#include <cstring>
+#include <string_view>
+
+#include "base/hash.h"
+#include "base/logging.h"
```
romange: please do not include `base/logging.h` in header files.
use `cassert` instead.

src/core/oah_entry.h
```diff
@@ -0,0 +1,533 @@
...
+  // TODO consider another option to implement iterator
+  OAHEntry* operator->() {
+    return this;
+  }
+
+  uint64_t GetHash() const {
+    return (data_ & kExtHashShiftedMask) >> kExtHashShift;
+  }
+
+  bool CheckBucketAffiliation(uint32_t bucket_id, uint32_t capacity_log, uint32_t shift_log) {
```
romange: can we have cc file for the implementation?

BorysTheDev: Honestly, I don't want .cc files for the best performance; this is why I use templates instead of inheritance like in DenseSet. But we have quite a limited number of types, so I can move code into .cc and make explicit instantiation, but still it will be slower, because of the lack of inlining.

romange: I think inlining makes sense in some cases but this umbrella decision won't scale and the code becomes less manageable and bloated. functions with several lines should be in cc filer unless it is possible to show they are on a hotspot.

kostasrim: > Honestly, I don't want .cc files for the best performance; this is why I use templates instead of inheritance like in DenseSet.

In most cases it doesn't matter. Compilers are smart creatures that can devirtualize very well -- John Lakos made a great talk about allocators in cppcon (i think 2017) that he talked in great deal why devirtualization works well in modern compilers and why C++ uses PMR instead. It's a misconception to think that CRPT is faster than plain inheritance.

> I think inlining makes sense in some cases but this umbrella decision won't scale and the code becomes less

Modern compilers can perform inlining across translation units. It rarely makes a difference and to be honest -- you simply can't control or manually force inlining unless you use explicit compiler extensions like ` __attribute__((always_inline))`

And don't get me started ranting about `inline` keywoard. It has nothing to do with inlining but you already aware of that..

src/core/oah_entry.h
dranikpg: Just make PtrVector store a reference to the pointer instead and make it a non owning wrapper

dranikpg (approved):

PR #4706, opened 2025-03-05
fix: Remove the dfly logo print from pytest run #4682

vyavdoshenko:
fix: Remove the dfly logo print from pytest run #4682

### DIFF
--- a/src/server/dfly_main.cc
+++ b/src/server/dfly_main.cc
@@ -85,7 +85,7 @@ ABSL_FLAG(uint16_t, uring_recv_buffer_cnt, 0,
           "How many socket recv buffers of size 256 to allocate per thread."
           "Relevant only for modern kernels with io_uring enabled");
 
-ABSL_FLAG(bool, skip_basic_usage, false, "Skip printing basic usage info.");
+ABSL_FLAG(bool, omit_basic_usage, false, "Omit printing basic usage info.");
 
 using namespace util;
 using namespace facade;
@@ -742,7 +742,7 @@ Usage: dragonfly [FLAGS]
 
   ParseFlagsFromEnv();
 
-  if (!GetFlag(FLAGS_skip_basic_usage)) {
+  if (!GetFlag(FLAGS_omit_basic_usage)) {
     PrintBasicUsageInfo();
   }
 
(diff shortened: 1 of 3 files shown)


BorysTheDev (dismissed):

src/server/dfly_main.cc
```diff
@@ -85,6 +85,8 @@ ABSL_FLAG(uint16_t, uring_recv_buffer_cnt, 0,
           "How many socket recv buffers of size 256 to allocate per thread."
           "Relevant only for modern kernels with io_uring enabled");
 
+ABSL_FLAG(bool, skip_basic_usage, false, "Skip printing basic usage info.");
```
romange, line 88: nit: omit_basic_usage

tests/dragonfly/instance.py
```diff
@@ -411,7 +411,9 @@ def __init__(self, params: DflyParams, args):
         self.params = params
         self.instances = []
 
-    def create(self, existing_port=None, path=None, version=100, **kwargs) -> DflyInstance:
+    def create(
+        self, existing_port=None, path=None, version=100, use_skip_basic_usage_flag=True, **kwargs
```
romange, line 415: I think all this is not needed and we can hardcode the flag below

romange: ah, I see, you have a problem when starting the older instance?

vyavdoshenko: Yes, correct.

romange: it is possible to do it based on version.
just add the flag inside the block:
```py
  if version >= 1.26:
 args.setdefault("fiber_safety_margin=4096")
```

src/server/dfly_main.cc
vyavdoshenko: Fixed

tests/dragonfly/instance.py
vyavdoshenko: Fixed

tests/dragonfly/instance.py
```diff
@@ -420,6 +420,8 @@ def create(self, existing_port=None, path=None, version=100, **kwargs) -> DflyIn
         vmod = "dragonfly_connection=1,accept_server=1,listener_interface=1,main_service=1,rdb_save=1,replica=1,cluster_family=1,proactor_pool=1,dflycmd=1,snapshot=1,streamer=1"
         args.setdefault("vmodule", vmod)
         args.setdefault("jsonpathv2")
+        if version > 1.27:
```
BorysTheDev, line 423: why do you need version

vyavdoshenko: @BorysTheDev because I have three failed tests without this check. They use the older versions of Dragonfly. These versions reject starting with unknown flag. 1.27 version is the latest released version.

adiholden (approved):

PR #4714, opened 2025-03-06
refactor: move tl_cluster_config into cluster_config.cc

BorysTheDev:
if you need cluster_config you don't need cluster_family now

### DIFF
--- a/src/server/cluster/cluster_config.cc
+++ b/src/server/cluster/cluster_config.cc
@@ -17,6 +17,9 @@ using namespace std;
 namespace dfly::cluster {
 
 namespace {
+
+thread_local shared_ptr<ClusterConfig> tl_cluster_config;
+
 bool HasValidNodeIds(const ClusterShardInfos& new_config) {
   absl::flat_hash_set<string_view> nodes;
 
@@ -388,6 +391,14 @@ std::vector<MigrationInfo> ClusterConfig::GetFinishedIncomingMigrations(
     const std::shared_ptr<ClusterConfig>& prev) const {
   return prev ? GetMissingMigrations(prev->my_incoming_migrations_, my_incoming_migrations_)
               : std::vector<MigrationInfo>();
+}
+
+std::shared_ptr<ClusterConfig> ClusterConfig::Current() {
+  return tl_cluster_config;
+}
+
+void ClusterConfig::SetCurrent(std::shared_ptr<ClusterConfig> config) {
+  tl_cluster_config = std::move(config);
 }
 
 }  // namespace dfly::cluster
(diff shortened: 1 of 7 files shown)


src/server/cluster/cluster_config.cc
```diff
@@ -17,6 +17,9 @@ using namespace std;
 namespace dfly::cluster {
 
 namespace {
+
+thread_local shared_ptr<ClusterConfig> tl_cluster_config;
```
kostasrim, line 21: I don't understand. Why is this a shared_ptr when it's a thread local ? Why do you need atomic ref count on a non concurrent (thread local) storage ?

kostasrim: Oh I just realized this was moved nvm. General remark, I still think shared pointer for a single thread is an antipattern + you do atomic operations on data that is not shared among threads.

src/server/cluster/cluster_family.cc
```diff
@@ -63,13 +63,11 @@ constexpr char kIdNotFound[] = "syncid not found";
...
 
-thread_local shared_ptr<ClusterConfig> tl_cluster_config;
-
 ClusterShardInfos GetConfigForStats(ConnectionContext* cntx) {
   CHECK(!IsClusterEmulated());
-  CHECK(tl_cluster_config != nullptr);
+  CHECK(ClusterConfig::Current() != nullptr);
 
-  auto config = tl_cluster_config->GetConfig();
+  auto config = ClusterConfig::Current()->GetConfig();
```
kostasrim, line 70: Before, you used the thread local directly. Now you do instead: copy of the shared_ptr which increments atomically the control block/ref count and then call GetConfig(). Is there a reason for this ?

kostasrim: If you don't need ref count and can gurantee the lifetime, why not simply return a `plain pointer` ?

kostasrim (commented):
Every call to `Current()` is a copy of the shared pointer. Is this something we are ok with ?

src/server/cluster/cluster_family.cc
BorysTheDev: We don't know when the life of the config will be finished. So it is the reason of shared_ptr.

src/server/cluster/cluster_config.cc
BorysTheDev: we share the same config between all threads

kostasrim: Then why is it a thread local ?

src/server/cluster/cluster_family.cc
kostasrim: That's fine but why then is it a thread local ?

src/server/cluster/cluster_config.cc
kostasrim: And also, each call to `current` now is a copy and an atomic increment. If this is used concurrently we might stall the processors (especially on NUMA). We should really use the threadlocal when we are in scope and we know that the object won't be destructed (we did that previously)

kostasrim (approved):

src/server/cluster/cluster_config.cc
BorysTheDev: discussed offline

PR #4715, opened 2025-03-06
fix(dfly_bench): support dns resolution for cluster hosts

romange:
also support multiple slot ranges per node.

Initial parsing of MOVED response is done but slot migration is not supported yet.

### DIFF
--- a/src/server/dfly_bench.cc
+++ b/src/server/dfly_bench.cc
@@ -104,9 +104,10 @@ static string GetRandomHex(size_t len) {
   return res;
 }
 
+using SlotRange = pair<uint16_t, uint16_t>;
+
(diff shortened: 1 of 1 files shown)


src/server/dfly_bench.cc
```diff
@@ -104,9 +104,10 @@ static string GetRandomHex(size_t len) {
   return res;
 }
 
+using SlotRange = pair<uint16_t, uint16_t>;
```
BorysTheDev, line 107: maybe we can include cluster_defs.h ?

romange: i prefer to keep dfly_bench independent from the server.

PR #4729, opened 2025-03-07
feat(server): Support dynamic SAVE path

### LINKED ISSUE #4660: Support dynamic s3 path to save df command
Dragonfly has some flags to configure s3 path for backups. However the configuration is static in nature i.e. we only get to configure the path when starting the dragonfly instance. The downside is backups always end up in the same s3 storage path.

I had configured s3 flags and executed the command `save df s3://path`. But it takes this as a file path.

I need dragonfly to support s3 path argument in `save df` command so that we can store different backups in different s3 paths. Something like `save df s3://path/to/s3/storage`.

mkaruza:
Add functionality to use SAVE and BGSAVE commands with dynamic CLOUD storage path. 

New syntax should be (as discussed with @romange):

`SAVE [RDB|DF] [CLOUD_URI] [BASENAME]` where `CLOUD_URI` should start with S3 or GCS prefix.

For example, now it should work to have working directory pointing to some local folder and executing `SAVE DF s3://bucket/snapshots my_snapshot`  would save snapshots to `s3://bucket/snapshots` with basename `my_snapshot`. 

Resolves #4660

### DIFF
--- a/tests/dragonfly/snapshot_test.py
+++ b/tests/dragonfly/snapshot_test.py
@@ -311,6 +311,21 @@ async def test_info_persistence_field(async_client):
     assert "loading:0" in (await async_client.execute_command("INFO PERSISTENCE"))
 
 
+def delete_s3_objects(bucket, prefix):
+    client = boto3.client("s3")
+    resp = client.list_objects_v2(
+        Bucket=bucket,
+        Prefix=prefix,
+    )
+    keys = []
+    for obj in resp["Contents"]:
+        keys.append({"Key": obj["Key"]})
+    client.delete_objects(
+        Bucket=bucket,
+        Delete={"Objects": keys},
+    )
+
+
 # If DRAGONFLY_S3_BUCKET is configured, AWS credentials must also be
 # configured.
 @pytest.mark.skipif(
@@ -338,22 +353,7 @@ async def test_s3_snapshot(async_client, tmp_dir):
         assert await StaticSeeder.capture(async_client) == start_capture
 
     finally:
-
-        def delete_objects(bucket, prefix):
-            client = boto3.client("s3")
-            resp = client.list_objects_v2(
-                Bucket=bucket,
-                Prefix=prefix,
-            )
-            keys = []
-            for obj in resp["Contents"]:
-                keys.append({"Key": obj["Key"]})
-            client.delete_objects(
-                Bucket=bucket,
-                Delete={"Objects": keys},
-            )
-
-        delete_objects(
+        delete_s3_objects(
             os.environ["DRAGONFLY_S3_BUCKET"],
             str(tmp_dir)[1:],
         )
@@ -366,7 +366,7 @@ async def test_s3_snapshot(async_client, tmp_dir):
     reason="AWS S3 snapshots bucket is not configured",
 )
 @dfly_args({**BASIC_ARGS})
-async def test_s3_save(async_client):
+async def test_s3_save_local_dir(async_client):
     seeder = StaticSeeder(key_target=10_000)
     await seeder.run(async_client)
 
@@ -377,22 +377,7 @@ async def test_s3_save(async_client):
         )
 
     finally:
-
-        def delete_objects(bucket, prefix):
-            client = boto3.client("s3")
-            resp = client.list_objects_v2(
-                Bucket=bucket,
-                Prefix=prefix,
-            )
-            keys = []
-            for obj in resp["Contents"]:
-                keys.append({"Key": obj["Key"]})
-            client.delete_objects(
-                Bucket=bucket,
-                Delete={"Objects": keys},
-            )
-
-        delete_objects(
+        delete_s3_objects(
             os.environ["DRAGONFLY_S3_BUCKET"],
             "s3_dump",
         )


src/server/detail/save_stages_controller.cc
```diff
@@ -378,7 +378,7 @@ GenericError SaveStagesController::FinalizeFileMovement() {
 
 // Build full path: get dir, try creating dirs, get filename with placeholder
 GenericError SaveStagesController::BuildFullPath() {
-  fs::path dir_path = GetFlag(FLAGS_dir);
+  fs::path dir_path = fs::path(GetFlag(FLAGS_dir)) / sub_path_;
```
kostasrim, line 381: I have a small concern here. So how it works now, is that `FLAGS_dir` acts as the `base` directory to save on s3. So for example if `dir == s3://myfolder` . Then `SAVE df myname` will save it on s3 at `s3://myfolder/myname.dfs` 

Now with your changes if I do: `SAVE df s3://kostas/myname` it will save to `s3://myfolder/s3://kostas/myname.dfs` which I am think a) it's invalid path and b) is wrong because I want the path to be `s3://kostas/` and not under the base dir `s3://FLAGS_dir/kostas/`. 

@Abhra303 Can you confirm the behavior you want ? So let's say if:

`dir = s3://myfolder` and you call `SAVE df s3://kostas/mysnap` it will save to `s3://kostas/` ignoring the `base dir` correct ?

src/server/server_family.cc
```diff
@@ -2079,25 +2083,29 @@ std::optional<ServerFamily::VersionBasename> ServerFamily::GetVersionAndBasename
...
   }
 
-  string_view basename;
+  string sub_path;
+  string basename;
   if (args.size() == 2) {
-    basename = ArgS(args, 1);
+    fs::path path(ArgS(args, 1));
+    sub_path = path.parent_path().relative_path().string();
+    basename = path.filename().string();
```
kostasrim, line 2091: nit you probably mean `filename` and not `basename` which implies a dir

src/server/server_family.h
```diff
@@ -195,8 +195,8 @@ class ServerFamily {
 
   // if new_version is true, saves DF specific, non redis compatible snapshot.
   // if basename is not empty it will override dbfilename flag.
-  GenericError DoSave(bool new_version, std::string_view basename, Transaction* transaction,
-                      bool ignore_state = false);
+  GenericError DoSave(bool new_version, std::string_view sub_path, std::string_view basename,
```
kostasrim, line 198: Since the arguments of this function seem to grow, would it be worth to group them under a struct to clean up the API ?

kostasrim: Also maybe even worth avoiding the split in `GetVersionSubPathAndBasename`. you can do that later when needed as long as you parse it and you know the path does not have an error. That avoid all the changes here including the tupple below :)

src/server/server_family.cc
mkaruza: Basename was used before, where you specified prefix for all saved files. There was no change in logic there, so i kept basename.

src/server/server_family.h
mkaruza: Make sense, i'll do change interface to accept `GetVersionSubPathAndBasename` struct.

src/server/detail/save_stages_controller.cc
mkaruza: We don't have specification how this should be done, so current implementation is done in a way that you specify only sub paths / basename for save command (this should work for all cloud providers and filesystem in same way). Of course, if we want to have feature that works in a way that you are expecting - i would modify it with next commits.

Also, fyi your example is valid path (on S3/Filesystem).

romange:
I think there is a misunderstanding about the feature, so I would like to clarify it.

We need to support functionality that when dragonfly is configured with `dragonfly --dir /disk/local/snapshots` we can still be able to save it into an arbitrary cloud path using `save df s3://mybucket/path` (by ignoring dir). Regular `SAVE` will work as usual and will use `dir` to write local snapshots. 

The challenge here is that the file system controller is created upon server initialization based on `dir` and it chooses its filesystem type then. once it was defined to say local fs, you can not use it for saving into cloud. so today with `dragonfly --dir /disk/local/snapshots`  you can not save to `s3://mybucket/path` but with `dragonfly --dir s3://bucket1/` you can save into `dragonfly --dir s3://bucket2/anotherdir` but you can not save to `gs://googlebucket/path`. so the goal is to create a temporary controller during `save df` configurations. Similarly we should do it with `DFLY LOAD` as well in the future. maybe even remove `save_controller_` altogether and always use a temporary one.

src/server/detail/save_stages_controller.cc
kostasrim: I understand, basically I was trying to say what Roman wrote below :smile: 

p.s. I did not even my example was valid path -- my knowledge of s3 is fairly limited

src/server/server_family.cc
```diff
@@ -2066,46 +2080,60 @@ void ServerFamily::BgSaveFb(boost::intrusive_ptr<Transaction> trans) {
...
+#ifdef WITH_AWS
+      save_cmd_opts.cloud_uri = ArgS(args, 1);
+#else
+      LOG(ERROR) << "Compiled without AWS support";
+      exit(1);
+#endif
+    } else if (IsGCSPath(ArgS(args, 1))) {
+      save_cmd_opts.cloud_uri = ArgS(args, 1);
+    } else {
+      // Probably basename than
```
romange, line 2116:
```suggestion
      // Probably basename then.
```

romange (commented):
lgtm, what's missing is a test in snapshot_test.py.
we already have s3 snapshot tests lets add one that uses a local dir but saves into cloud.

src/server/server_family.h
```diff
@@ -193,9 +202,7 @@ class ServerFamily {
 
   void StatsMC(std::string_view section, SinkReplyBuilder* builder);
 
-  // if new_version is true, saves DF specific, non redis compatible snapshot.
-  // if basename is not empty it will override dbfilename flag.
-  GenericError DoSave(bool new_version, std::string_view basename, Transaction* transaction,
+  GenericError DoSave(SaveCmdOptions save_cmd_opts, Transaction* transaction,
```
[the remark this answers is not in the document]
kostasrim: It's trivially copyable, IMO by value is fine. I guess the problem is that if we add a member that makes the struct non trivially copyable and then we need to remember to change that (or enforce it via static_assert so if it ever changes we are aware). Sooooo `const &` it is :rofl:

tests/dragonfly/snapshot_test.py
```diff
@@ -359,6 +359,45 @@ def delete_objects(bucket, prefix):
...
 
 
+# If DRAGONFLY_S3_BUCKET is configured, AWS credentials must also be
+# configured.
+@pytest.mark.skipif(
+    "DRAGONFLY_S3_BUCKET" not in os.environ or os.environ["DRAGONFLY_S3_BUCKET"] == "",
+    reason="AWS S3 snapshots bucket is not configured",
+)
+@dfly_args({**BASIC_ARGS})
+async def test_s3_save(async_client):
```
romange, line 369: maybe name it `test_s3_save_local_dir`?

tests/dragonfly/snapshot_test.py
```diff
@@ -359,6 +359,45 @@ def delete_objects(bucket, prefix):
...
+
+    try:
+        # SAVE to S3 bucket with `s3_dump` as filename prefix
+        await async_client.execute_command(
+            "SAVE", "DF", "s3://" + os.environ["DRAGONFLY_S3_BUCKET"], "s3_dump"
+        )
+
+    finally:
+
+        def delete_objects(bucket, prefix):
```
romange, line 381: Please avoid code duplication even if it's a python code.

mkaruza:
Documentation update will follow this PR.

romange (approved):

PR #4731, opened 2025-03-08
fix: improve stack margin for s3 related operations.

romange:
our S3 code relies on aws sdk client, which is extremely stack hungry.
this PR moves some of s3 calls to one-off fibers with increased stacks,
which reduces stack usage for connection fibers executing snapshot save/load operations.


romange:
the run with `fiber_safety_margin=8000` is here: https://github.com/dragonflydb/dragonfly/actions/runs/13739554151

before this PR it failed in multiple places around the s3 code.

adiholden (approved):

PR #4733, opened 2025-03-09
chore: exclude local files changes from git staging

vyavdoshenko:
chore: added script to ignore local file changes in files mentioned in .gitignore but still shown as changed
docs: CONTRIBUTING.md has been updated 
docs: tools/local/SETUP_GIT_IGNORE_LOCAL.md was added


romange:
Curious, how it is better than .gitignore? in any case, we put utilities that help local development under `tools/local`

CONTRIBUTING.md
```diff
@@ -104,6 +104,71 @@ The Dragonfly DB team has agreed to systematically use several pre-commit hooks
...
 when you do your commits.
 
+## Git workflow
+
+### Repository Setup
+
+After cloning the repository, run the script to set up Git to ignore local changes in specific files:
+
+```bash
+chmod +x setup-git-ignore-local.sh
```
romange, line 114: why not submit an executable script?

vyavdoshenko:
> Curious, how it is better than .gitignore? in any case, we put utilities that help local development under `tools/local`

@romange 
These files are in .gitignore added before me. That doesn't help, I verified that on two dev environments. It worked via .gitignore in the past. I used that by simply adding files to the .gitignore. It doesn't work any longer. I guess modern git versions don't support this approach.
That's why I suggest this solution.

romange:
interesting, what does not work with .gitignore? It works on my machine - i just changed `.vscode/settings.json` and it does not show locally.

vyavdoshenko:
> interesting, what does not work with .gitignore? It works on my machine - i just changed `.vscode/settings.json` and it does not show locally.

Screenshot 2025-03-09 at 16 19 33

CONTRIBUTING.md
vyavdoshenko: Fixed.

Issue #4639, opened 2025-02-20
get rid of sds arguments in SortedMap


romange:
(good for starter tasks)

SortedMap still has some methods with sds because zset_family code uses sds and that in turn due to reliance on zzl utlilites like `zzlFind`

1. We should reimplement zzl utilities using string_view instead of sds, (only those that we need of course), 
and use where they are needed in zset_family, similarly to how we implemented `LpFind` vs `lpFind`
2. We should get rid of `SdsWrap` calls in zset_family and pass string_view directly.
3. We should get rid of functions accepting sds arguments inside SortedMap.

An example PR that does such cleanup: https://github.com/dragonflydb/dragonfly/pull/4638

abhijat:
@romange I have a WIP pr to do this cleanup https://github.com/dragonflydb/dragonfly/pull/4738, there is an issue I see there which is seen when inserting empty strings in sorted map:

1. during argument parsing empty strings are passed around as default initialized `string_view` - these return `nullptr` when `data()` is called on them
2. when storing field with score we do an `lpAppend` call which inserts before end of list pack. We pass the result of calling `data()` on default initialized string view here https://github.com/dragonflydb/dragonfly/blob/24b3723241bb564d49d1c87657c4d8e916a36e1b/src/core/sorted_map.cc#L193
3. When a null pointer is inserted to list pack, it is treated as a replacement operation which is incorrect in our case and also triggers assertion failures https://github.com/dragonflydb/dragonfly/blob/a3cabd10d396f637edcde851f10edade3017dead/src/redis/listpack.c#L836

To avoid this in the PR, we handle the corner case by detecting a null pointing string view before `lpAppend` and convert it to `string_view{""}`. 

This seems to work fine for this use case, but just wanted to call this out in case maybe there are other ways to approach this or even keeping `WrapSds` around for this specific code path.

PR #4738, opened 2025-03-10
chore(zset_family/score_map): Replace sds arguments with string_view

abhijat:
The usages of WrapSds are removed from zset_family calls.

In most cases it is a straightforward change to use string_view or string_view.data(), string_view.size(). 

In one case a wrapper had to be introduced.

fixes https://github.com/dragonflydb/dragonfly/issues/4639

### DIFF
--- a/src/core/sorted_map.cc
+++ b/src/core/sorted_map.cc
@@ -108,7 +108,8 @@ constexpr unsigned kConvFlags = DoubleToStringConverter::UNIQUE_ZERO;
 DoubleToStringConverter score_conv(kConvFlags, "inf", "nan", 'e', -6, 21, 6, 0);
 
 // Copied from redis code but uses double_conversion to encode double values.
-unsigned char* zzlInsertAt(unsigned char* zl, unsigned char* eptr, sds ele, double score) {
+unsigned char* zzlInsertAt(unsigned char* zl, unsigned char* eptr, const unsigned char* ele,
+                           const uint32_t ele_size, double score) {
   unsigned char* sptr;
   char scorebuf[128];
   unsigned scorelen = 0;
@@ -124,14 +125,14 @@ unsigned char* zzlInsertAt(unsigned char* zl, unsigned char* eptr, sds ele, doub
   }
 
   if (eptr == NULL) {
-    zl = lpAppend(zl, (unsigned char*)ele, sdslen(ele));
+    zl = lpAppend(zl, ele, ele_size);
     if (score_is_long)
       zl = lpAppendInteger(zl, lscore);
     else
       zl = lpAppend(zl, (unsigned char*)scorebuf, scorelen);
   } else {
     /* Insert member before the element 'eptr'. */
-    zl = lpInsertString(zl, (unsigned char*)ele, sdslen(ele), eptr, LP_BEFORE, &sptr);
+    zl = lpInsertString(zl, ele, ele_size, eptr, LP_BEFORE, &sptr);
 
     /* Insert score after the member. */
     if (score_is_long)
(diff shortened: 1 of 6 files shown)


abhijat:
test failure seems related to this change:

```
src/redis/listpack.c:886]: assert((p) >= (lp)+LP_HDR_SIZE && (p)+(replaced_len) < (lp)+lpGetTotalBytes((lp))) failed!
```

passing in an empty string view ends up being treated as nullptr of zero size, which is different from the behavior for sds

abhijat:
> passing in an empty string view ends up being treated as nullptr of zero size, which is different from the behavior for sds

reproducible by adding an empty field in unit test for zadd.

src/core/sorted_map.cc
```diff
@@ -108,7 +108,8 @@ constexpr unsigned kConvFlags = DoubleToStringConverter::UNIQUE_ZERO;
 DoubleToStringConverter score_conv(kConvFlags, "inf", "nan", 'e', -6, 21, 6, 0);
 
 // Copied from redis code but uses double_conversion to encode double values.
-unsigned char* zzlInsertAt(unsigned char* zl, unsigned char* eptr, sds ele, double score) {
+unsigned char* zzlInsertAt(unsigned char* zl, unsigned char* eptr, const unsigned char* ele,
```
romange, line 111: nit: I would use `const char* ele` to avoid multiple castings on the caller side.
Maybe even better: it's now in our c++ code base, so you could even keep `string_view  ele` here and make the safe conversion to `""sv` in this function.

src/core/sorted_map.cc
```diff
@@ -181,7 +190,26 @@ unsigned char* ZzlInsert(unsigned char* zl, sds ele, double score) {
...
-  return zzlInsertAt(zl, NULL, ele, score);
+  return zzlInsertAt(zl, NULL, (unsigned char*)ele.data(), ele.size(), score);
+}
+
+unsigned char* ZzlFind(unsigned char* lp, std::string_view ele, double* score) {
+  unsigned char *eptr, *sptr;
+
+  if ((eptr = lpFirst(lp)) == nullptr)
+    return nullptr;
+  eptr = lpFind(lp, eptr, (unsigned char*)ele.data(), ele.size(), 1);
```
romange, line 201: can ele be `{null, 0}` ? will it work?

src/core/sorted_map.cc
romange: btw, to signify it's ours - please rename it to `ZzlInsertAt`

src/core/sorted_map.cc
abhijat: Yes, in `lpFind` there are two operations on `ele.data()`, both are checked by the size first, `memcmp` and `lpStringToInt64` both handle nullptr correctly when the size is 0.

src/core/sorted_map.cc
abhijat: changed name and signature

src/core/sorted_map.cc
```diff
@@ -123,15 +124,24 @@ unsigned char* zzlInsertAt(unsigned char* zl, unsigned char* eptr, sds ele, doub
     DCHECK_EQ(scorelen, strlen(scorebuf));
   }
 
+  // Argument parsing converts empty strings to default initialized string views.
+  // Such string views have a null data field, which if passed into lpAppend (via zzlInsertAt)
+  // results in the replace operation being applied on the listpack. In addition to being wrong, it
+  // also causes assertion failures. To circumvent this corner case we pass here a string view
+  // pointing to an empty string on the stack, which has a non-null data field.
+  if (ele.data() == nullptr) {
```
kostasrim, line 132: Why don't we patch `lpAppend` ? Or if we want to enforce a contract, why not simply introduce a wrapper that does this check for us here and use it. That way, we save this "hack"? 

Not that it matters that much, it's a few lines

kostasrim (commented):
Small comment, LGTM

src/core/sorted_map.cc
romange: we do not change code under redis directory and this function is the wrapper

kostasrim: oh right you want to keep the diffs clean

Issue #4708, opened 2025-03-05
some traffic metrics should be labeled as main/other


romange:
A feedback from control plane - they would like to add labels to the following metrics:
1. dragonfly_commands_processed_total
2. dragonfly_connected_clients

the labels should be "main" for connections coming from the main listener, and "other" from non-main listeners.

1. `facade::Listener` has Role variable and `bool IsMainInterface()` accessort.
2. `facade::Connection` has `IsMain` accessor that accesses its listener but I would add a flag `is_main_` similarly to `is_tls_`, `is_http_` - it can be set inside `Connection::OnConnectionStart`.
3. `RecordCmd(bool)` increases `dragonfly_commands_processed_total` , so it we should pass `bool` to say whether that comman came from a "main" connection or not.
4. Similarly handle dragonfly_connected_clients. Each metric should have two separate `uint64_t` variables internally and represented as differrent label in `PrintPrometheusMetrics`.

abhijat:
@romange what should the label string be? The values are `main` and `other`, so should the label be something like `source` or `listener`?

PR #4739, opened 2025-03-10
feat(metrics): Add label for main and other listeners

abhijat:
The stats collected per connection are divided according to main or other listener.

Metrics are decorated with labels listener= main or other.

fixes https://github.com/dragonflydb/dragonfly/issues/4708

The behavior of `INFO` command is unchanged, it still displays the sum of connections and commands instead of separate main and other counts.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -925,16 +925,13 @@ bool Connection::IsMain() const {
   return is_main_;
 }
 
-bool Connection::HasMainListener() const {
+bool Connection::IsMainOrMemcache() const {
   if (is_main_) {
     return true;
   }
 
-  if (const Listener* lsnr = static_cast<Listener*>(listener()); lsnr) {
-    return lsnr->protocol() == Protocol::MEMCACHE;
-  }
-
-  return false;
+  const Listener* lsnr = static_cast<Listener*>(listener());
+  return lsnr && lsnr->protocol() == Protocol::MEMCACHE;
 }
 
 void Connection::SetName(string name) {
@@ -2011,14 +2008,14 @@ void Connection::MarkReadBufferConsumed() {
 }
 
 void Connection::IncrNumConns() {
-  if (HasMainListener())
+  if (IsMainOrMemcache())
     ++stats_->num_conns_main;
   else
     ++stats_->num_conns_other;
 }
 
 void Connection::DecrNumConns() {
-  if (HasMainListener())
+  if (IsMainOrMemcache())
     --stats_->num_conns_main;
   else
     --stats_->num_conns_other;
--- a/src/server/main_service.cc
+++ b/src/server/main_service.cc
@@ -1318,7 +1318,7 @@ bool Service::InvokeCmd(const CommandId* cid, CmdArgList tail_args, SinkReplyBui
     DispatchMonitor(cntx, cid, tail_args);
   }
 
-  ServerState::tlocal()->RecordCmd(cntx->has_main_listener);
+  ServerState::tlocal()->RecordCmd(cntx->has_main_or_memcache_listener);
   Transaction* tx = cntx->transaction;
   auto& info = cntx->conn_state.tracking_info_;
   const bool is_read_only = cid->opt_mask() & CO::READONLY;
--- a/src/facade/dragonfly_connection.h
+++ b/src/facade/dragonfly_connection.h
@@ -263,7 +263,7 @@ class Connection : public util::Connection {
 
   // In addition to the listener role being main, also returns true if the protocol is Memcached.
   // This method returns true for customer facing listeners.
-  bool HasMainListener() const;
+  bool IsMainOrMemcache() const;
 
   void SetName(std::string name);
 
--- a/src/server/conn_context.cc
+++ b/src/server/conn_context.cc
@@ -104,7 +104,7 @@ ConnectionContext::ConnectionContext(facade::Connection* owner, acl::UserCredent
     : facade::ConnectionContext(owner) {
   if (owner) {
     skip_acl_validation = owner->IsPrivileged();
-    has_main_listener = owner->HasMainListener();
+    has_main_or_memcache_listener = owner->IsMainOrMemcache();
   }
 
   keys = std::move(cred.keys);
@@ -127,7 +127,7 @@ ConnectionContext::ConnectionContext(const ConnectionContext* owner, Transaction
     acl_db_idx = owner->acl_db_idx;
     ns = owner->ns;
     if (owner->conn()) {
-      has_main_listener = owner->conn()->HasMainListener();
+      has_main_or_memcache_listener = owner->conn()->IsMainOrMemcache();
     }
   } else {
     acl_commands = std::vector<uint64_t>(acl::NumberOfFamilies(), acl::NONE_COMMANDS);
--- a/src/server/conn_context.h
+++ b/src/server/conn_context.h
@@ -318,7 +318,7 @@ class ConnectionContext : public facade::ConnectionContext {
   FlowInfo* replication_flow = nullptr;
 
   // The related connection is bound to main listener or serves the memcached protocol
-  bool has_main_listener = false;
+  bool has_main_or_memcache_listener = false;
 
  private:
   void EnableMonitoring(bool enable) {


src/facade/dragonfly_connection.cc
```diff
@@ -700,14 +700,18 @@ void Connection::OnPostMigrateThread() {
   }
 
   stats_ = &tl_facade_stats->conn_stats;
-  ++stats_->num_conns;
+  IsMain() ? ++stats_->num_conns_main : ++stats_->num_conns_other;
```
BagritsevichStepan, line 703: Since it repeated several times I suggest return something like uint32_t& GetNumConns() and do `++GetNumConns()` or ``--GetNumConns()`

abhijat: sounds good, added

tests/dragonfly/server_family_test.py
```diff
@@ -172,3 +173,38 @@ async def test_blocking_commands_should_not_show_up_in_slow_log(
 
     # blpop does not show up, only the previous reset
     assert reply[0]["command"] == "SLOWLOG RESET"
+
+
+@dfly_args({"memcached_port": 11211})
+async def test_metric_labels(
+    df_server: DflyInstance, async_client: aioredis.Redis, memcached_client: Client
+):
```
kostasrim, line 181: We do not need a test for the prometheous metrics but I guess it won't hurt ? You can just open your browser and check via `http://localhost:6379/metrics` or just rely on the unit test. Not strongly opinionated about this tbh

kostasrim: Also you don't need `memcached_client` or `memcached_port`

src/server/server_family.cc
```diff
@@ -1281,8 +1281,11 @@ void PrintPrometheusMetrics(uint64_t uptime, const Metrics& m, DflyCmd* dfly_cmd
   const auto& conn_stats = m.facade_stats.conn_stats;
   AppendMetricWithoutLabels("max_clients", "Maximal number of clients", GetFlag(FLAGS_maxclients),
                             MetricType::GAUGE, &resp->body());
-  AppendMetricWithoutLabels("connected_clients", "", conn_stats.num_conns, MetricType::GAUGE,
-                            &resp->body());
+  AppendMetricHeader("connected_clients", "", MetricType::GAUGE, &resp->body());
+  AppendMetricValue("connected_clients", conn_stats.num_conns_main, {"listener"}, {"main"},
+                    &resp->body());
+  AppendMetricValue("connected_clients", conn_stats.num_conns_other, {"listener"}, {"other"},
+                    &resp->body());
```
kostasrim, line 1288: @adiholden Won't this break our dashboards ? Since now we need to use the label as well on our `queries` ? 

Double checking here that this is ok for us.

src/facade/dragonfly_connection.cc
```diff
@@ -1995,6 +1998,10 @@ void Connection::MarkReadBufferConsumed() {
   }
 }
 
+uint32_t& Connection::NumConns() {
+  return IsMain() ? stats_->num_conns_main : stats_->num_conns_other;
+}
```
kostasrim, line 2003: I really do not like to return references especially when we only want to increment what is being referenced. I know that Stefan suggested this but I would also suggest a slightly change: Rename `NumConns()` to `IncrNumConns()` and do the increment here:

```
void Connection::IncrNumConnStats() {
  IsMain() ? ++stats_->blah blah....
}
```

src/facade/dragonfly_connection.cc
```diff
@@ -700,14 +700,18 @@ void Connection::OnPostMigrateThread() {
...
 }
 
 void Connection::OnConnectionStart() {
   ThisFiber::SetName("DflyConnection");
 
   stats_ = &tl_facade_stats->conn_stats;
+
+  if (const Listener* ls = dynamic_cast<Listener*>(listener()); ls) {
+    is_main_ = ls->IsMainInterface();
+  }
```
kostasrim, line 714: I know that the `static_cast` you replaced looks weird and wrong but it isn't.

1. We do not have other `Listener` types so if you `static_cast` here is perfectly fine and the `downcast` is safe to do so at compile time.
2. When you use `dynamic_cast` you are actually doing a `runtime cast` which queries the `RTTI` information of the object. There is an associated overhead with this cast. It's needed when "you are not sure" about the object type. Here, in this context and for our data types, we only have `Listener` so the `static_cast` will always succeed and be correct.

IMO this is our bad design. C++ has solved this problem and it's called CRTP (curiously recurring template pattern) which is a way to express `static polymoprphism` in a `safe way` (avoiding the explicit `static_cast` we used above by delegating that to the base class via the compiler). I never had a change to refactor our class hierarchies and that's why we got `static_cast` here. It's not wrong but it could be expressed better. However, I don't have the time to push such a refactor (it's not hard to do so but it does require time because you need to substitute virtual functions with static casts done by the base class). 

p.s. you probably have encounter CRTP in seastar. All of the `WRITE/READ DMA` functions in there should use `CRTP` internally to decouple interface/impl requirements

src/server/main_service.cc
```diff
@@ -1318,7 +1318,8 @@ bool Service::InvokeCmd(const CommandId* cid, CmdArgList tail_args, SinkReplyBui
     DispatchMonitor(cntx, cid, tail_args);
   }
 
-  ServerState::tlocal()->RecordCmd();
+  const bool is_main = cntx->conn() ? cntx->conn()->IsMain() : false;
```
kostasrim, line 1321: Cn `cntx->conn()` be null here ? If so maybe we can store the `IsMain` in `ConnectionContext` instead ?

src/server/server_family.cc
kostasrim: double checked with @Pothulapati we should be ok with the label. It won't break our dashboards.

tests/dragonfly/server_family_test.py
abhijat: I used memcached client to get a count of `other` connection and command as I see it goes via a listener other than main. Do you recommend to remove this test?

src/facade/dragonfly_connection.cc
abhijat: Do you recommend switching back to `static_cast` here to avoid paying the type query cost?

src/server/main_service.cc
abhijat: I saw this was null during a unit test which runs the `populate` debug command, because it called `InvokeCmd` with a locally created context which had owner set to nullptr https://github.com/dragonflydb/dragonfly/blob/main/src/server/debugcmd.cc#L196

src/facade/dragonfly_connection.cc
romange: @abhijat  we usually do not rely on rtti in our code. Matter of taste and we do not use it.

tests/dragonfly/server_family_test.py
romange: it's actually an interesting point. We want to count the memcache listener as "main" as well. it is customer facing as well.

src/facade/dragonfly_connection.cc
kostasrim: @romange I wrote a paragraph :rofl: But yeah rtti has a small overhead as well :)

src/server/main_service.cc
kostasrim: yep this is a `stub context` and we have a bug. So, why don't we use the context to check if `it's main` or not ?

src/facade/dragonfly_connection.cc
abhijat: changed

src/facade/dragonfly_connection.cc
abhijat: moved back to `static_cast`

tests/dragonfly/server_family_test.py
abhijat: I added a new method which returns true if the role is `MAIN` _or_ the protocol is `MEMCACHED` and adjusted the metrics to use this. The test is also changed to account for this and use the admin port for counting `other` metrics.

This will metric will be incremented for any memcached listener (or main role listener),  I hope this is correct behavior.

src/server/dragonfly_test.cc
```diff
@@ -834,4 +834,16 @@ TEST_F(DflyEngineTest, ReplicaofRejectOnLoad) {
...
+TEST_F(DflyEngineTest, CommandMetricLabels) {
+  EXPECT_EQ(Run({"SET", "foo", "bar"}), "OK");
+  EXPECT_EQ(Run({"GET", "foo"}), "bar");
+  const Metrics metrics = GetMetrics();
+
+  // The test connection counts as other
+  EXPECT_EQ(metrics.facade_stats.conn_stats.command_cnt_other, 2);
+  EXPECT_EQ(metrics.facade_stats.conn_stats.command_cnt_main, 0);
+  EXPECT_EQ(metrics.facade_stats.conn_stats.num_conns_main, 0);
+  EXPECT_EQ(metrics.facade_stats.conn_stats.num_conns_other, 0);
```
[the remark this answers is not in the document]
abhijat: yeah, in `BaseFamilyTest::Run` we seem to directly call `DispatchCommand` on the test connection wrapper, the `HandleRequests` method isn't called which would increment or decrement these.

src/facade/dragonfly_connection.cc
```diff
@@ -918,7 +922,19 @@ bool Connection::IsPrivileged() const {
...
+bool Connection::HasMainListener() const {
+  if (is_main_) {
+    return true;
+  }
+
+  if (const Listener* lsnr = static_cast<Listener*>(listener()); lsnr) {
+    return lsnr->protocol() == Protocol::MEMCACHE;
+  }
+
+  return false;
```
kostasrim, line 937:
```suggestion
  const Listener* lsnr = static_cast<Listener*>(listener());
  return is_main_ ||  lsnr->protocol() == Protocol::MEMCACHE;
```

kostasrim: and maybe rename because `HasMainListener` is kinda a lie if protocol is MEMCACHE. Maybe something like `IsMainOrMemcached` ?

kostasrim (commented):
minor comment LGTM

src/facade/dragonfly_connection.cc
abhijat: The problem with accessing `lsnr->protocol()` without a null check is that I found it to be null during tests when we end up here with `TestConnection`, and it caused segfault. From what I could see the test connection doesn't have a listener pointer.

abhijat: I made both the changes, I ordered it so the `static_cast` is only after checking for `is_main_`, but I can change it to make everything in one composite statement, in that case we always have to do the static cast.

kostasrim (approved):
LGTM

PR #4748, opened 2025-03-11
chore: replace chrono with absl in PipelineCacheSizeTracker

kostasrim:
Use absl Time instead of Chrono in PipelineCacheSizeTracker

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -7,9 +7,9 @@
 #include <absl/container/flat_hash_map.h>
 #include <absl/strings/match.h>
 #include <absl/strings/str_cat.h>
+#include <absl/time/time.h>
 #include <mimalloc.h>
 
-#include <chrono>
 #include <numeric>
 #include <variant>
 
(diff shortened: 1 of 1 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -323,26 +323,24 @@ thread_local vector<Connection::PipelineMessagePtr> Connection::pipeline_req_poo
...
     return watermark_reached;
   }
 
  private:
-  using Tp = std::chrono::time_point<std::chrono::system_clock>;
-  using Clock = std::chrono::system_clock;
   using Limits = std::numeric_limits<size_t>;
 
-  Tp last_check_ = Clock::now();
+  absl::Time last_check_ = absl::Now();
```
kostasrim, line 343: I am not convinced about any performance impacts of this (as promised I will benchmark / perf and see if chrono has any measurable overhead in our hot paths) but FOR SURE it has `SHORTER` naming and requires less typing :rofl: 

I promise once I get some free bandwidth I will make a case for this :smile:

adiholden (approved):

PR #4751, opened 2025-03-11
feat: allow non-ascii values in dfly_bench

romange:
binary strings are supported by RESP protocol so we now generate commands using this protocol, so that we could pass binary strings.

In addition, fixed "done" metric which did not account for number of shards cluster mode.

### DIFF
--- a/src/server/dfly_bench.cc
+++ b/src/server/dfly_bench.cc
@@ -61,6 +61,7 @@ ABSL_FLAG(bool, noreply, false, "If true, does not wait for replies. Relevant on
 ABSL_FLAG(bool, greet, true,
           "If true, sends a greeting command on each connection, "
           "to make sure the connection succeeded");
+ABSL_FLAG(bool, ascii, true, "If true, use ascii characters for values");
 
 using namespace std;
 using namespace util;
@@ -82,9 +83,7 @@ enum Protocol { RESP, MC_TEXT } protocol;
 enum DistType { UNIFORM, NORMAL, ZIPFIAN, SEQUENTIAL } dist_type{UNIFORM};
 constexpr uint16_t kNumSlots = 16384;
 
-string_view kTmplPatterns[] = {"__key__"sv, "__data__"sv, "__score__"sv};
-
-static string GetRandomHex(size_t len) {
+static string GetRandomHex(size_t len, bool ascii) {
   std::string res(len, '\0');
   size_t indx = 0;
 
@@ -103,6 +102,11 @@ static string GetRandomHex(size_t len) {
     }
   }
 
+  if (!ascii) {
+    for (size_t i = 0; i < len; i++) {
+      res[i] += 80;
+    }
+  }
   return res;
 }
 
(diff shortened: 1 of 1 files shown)


src/server/dfly_bench.cc
```diff
@@ -177,12 +184,24 @@ CommandGenerator::CommandGenerator(KeyGenerator* keygen) : keygen_(keygen) {
...
+      cmd_parts_.emplace_back(SCORE);
+    } else {
+      cmd_parts_.emplace_back(p);
+    }
+  }
+
+  if (!cmd_parts_.empty()) {
+    const string_view* cmd = get_if<string_view>(&cmd_parts_.front());
+    if (cmd) {
+      might_hit_ = absl::EqualsIgnoreCase(*cmd, "get") || absl::StartsWithIgnoreCase(*cmd, "mget");
```
kostasrim, line 203:
```suggestion
      might_hit_ = absl::EqualsIgnoreCase(*cmd, "get") || absl::EqualsIgnoreCase(*cmd, "mget");
```

src/server/dfly_bench.cc
```diff
@@ -103,6 +102,11 @@ static string GetRandomHex(size_t len) {
     }
   }
 
+  if (!ascii) {
+    for (size_t i = 0; i < len; i++) {
+      res[i] += 80;
```
kostasrim, line 107: Why +80? Plz explain :smile:

src/server/dfly_bench.cc
```diff
@@ -154,21 +158,24 @@ class CommandGenerator {
...
   bool might_hit_ = false;
   bool noreply_ = false;
+  bool is_ascii_ = true;
 };
 
 CommandGenerator::CommandGenerator(KeyGenerator* keygen) : keygen_(keygen) {
   command_ = GetFlag(FLAGS_command);
-  value_ = string(GetFlag(FLAGS_d), 'a');
+  is_ascii_ = GetFlag(FLAGS_ascii);
+  value_ = string(GetFlag(FLAGS_d), is_ascii_ ? 'a' : char(130));
```
kostasrim, line 178: 130 ?

src/server/dfly_bench.cc
```diff
@@ -204,58 +223,54 @@ string CommandGenerator::Next(SlotRange range) {
...
   return gen_cmd;
 }
 
 string CommandGenerator::FillSet(string_view key) {
+  string res;
+
   if (protocol == RESP) {
-    return absl::StrCat("set ", key, " ", value_, "\r\n");
-  }
+    absl::StrAppend(&res, "*3\r\n$3\r\nset\r\n$", key.size(), "\r\n", key);
```
kostasrim, line 255: I blindly trust you with this :rofl:

kostasrim (approved):

src/server/dfly_bench.cc
romange: hexadecimal characters with `0` padding start at ascii `0` which is 48. +80 moves the whole string to non-ascii range.

src/server/dfly_bench.cc
romange: nonascii

PR #4754, opened 2025-03-12
chore: Handle hsetex options in any order

abhijat:
NX and KEEPTTL can be specified in any order, but only one of each or both must be specified.


kostasrim (approved):
LGTM

PR #4759, opened 2025-03-12
fix: optional brackets were removed for ipv6 host

vyavdoshenko:
fix: optional brackets were removed for IPv6 host
fix: ipv6 support fix in the helio framework: https://github.com/romange/helio/pull/391

### DIFF
--- a/src/server/protocol_client.cc
+++ b/src/server/protocol_client.cc
@@ -135,7 +135,13 @@ ProtocolClient::~ProtocolClient() {
 
 error_code ProtocolClient::ResolveHostDns() {
   char ip_addr[INET6_ADDRSTRLEN];
-  auto ec = util::fb2::DnsResolve(server_context_.host, 0, ip_addr, ProactorBase::me());
+
+  std::string host = server_context_.host;
+  if (!host.empty() && host.front() == '[' && host.back() == ']') {
+    host = host.substr(1, host.size() - 2);
+  }
+
+  auto ec = util::fb2::DnsResolve(host, 0, ip_addr, ProactorBase::me());
   if (ec) {
     LOG(ERROR) << "Dns error " << ec << ", host: " << server_context_.host;
     return make_error_code(errc::host_unreachable);
--- a/src/server/protocol_client.cc
+++ b/src/server/protocol_client.cc
@@ -136,6 +136,10 @@ ProtocolClient::~ProtocolClient() {
 error_code ProtocolClient::ResolveHostDns() {
   char ip_addr[INET6_ADDRSTRLEN];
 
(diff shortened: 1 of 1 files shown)


src/server/protocol_client.cc
```diff
@@ -135,7 +135,13 @@ ProtocolClient::~ProtocolClient() {
 
 error_code ProtocolClient::ResolveHostDns() {
   char ip_addr[INET6_ADDRSTRLEN];
-  auto ec = util::fb2::DnsResolve(server_context_.host, 0, ip_addr, ProactorBase::me());
+
+  std::string host = server_context_.host;
```
kostasrim, line 139: No reason to deep copy here. Probably it's SSO but still better use `const string &`. I would have suggested `string_view` but it won't work with `DnsResolve()`. And I see that we need `substr` below. 

Imo I don't see a point why `DnsResolve` does not accept `std::string_view` so since you opened: https://github.com/romange/helio/pull/391
maybe worth adding that prototype?

vyavdoshenko: @kostasrim 
Sorry, I didn't get it. Which prototype do you mean?

kostasrim: sorry I meant, that we change `DnsResolve()` to accept a `string_view` instead of `const string&`. That way we can use `string_view::substr` to avoid the second copy

vyavdoshenko: @kostasrim 
I suggest leaving this change as is. DnsResolve() is in the helio repo. We have to refactor this method first in the helio to get support for it here.

kostasrim: @vyavdoshenko that's what I am saying. You have already an open PR in helio so maybe you can do that ? I don't mind, we can leave it for the future -- no pressure. I just get OCD from deep copies :smile:

romange: DnsResolve is called so infrequently so it's not worth discassing these things, imho

kostasrim: sounds good

src/server/protocol_client.cc
```diff
@@ -135,7 +135,13 @@ ProtocolClient::~ProtocolClient() {
...
+  if (!host.empty() && host.front() == '[' && host.back() == ']') {
```
romange, line 140: please add a comment explaining what kind of address format is being stripped.

vyavdoshenko: The comment has been added.

romange:
Thanks, btw - you can add the latest helio to this pr

romange (approved):

PR #4761, opened 2025-03-13
feat(dfly_bench): Handle moved slots in cluster

### LINKED ISSUE #4493: support cluster configuration with dfly_bench
`dfly_bench` covers lots of load-testing use-cases but it still does not support the cluster mode.

The goal of  this task is to provide cluster support by sending cluster discovery request if the protocol flag `FLAGS_P` is `RESP`.

In that case we should create `threads` x `c` connections per shard and drive all of them to completion

mkaruza:
When cluster slot reconfiguration happens during dfly_bench execution we will update slot information for each shard connection where this error is received.

Added `--pool_size` argument that can be used to set number of threads that are going to be running.

Resolves #4493


src/server/dfly_bench.cc
```diff
@@ -119,6 +120,91 @@ struct ShardInfo {
 
 using ClusterSpec = vector<ShardInfo>;
 
+template <typename D>
+ClusterSpec FetchCluster(const unique_ptr<FiberSocketBase, D>& socket,
```
kostasrim, line 124: You don't have to make this a template and passing `unique_ptr` as `const &` is an `anti-pattern`. I suggest to use a plain ptr here. `FibserSocketBase*`

src/server/dfly_bench.cc
```diff
@@ -966,7 +976,12 @@ int main(int argc, char* argv[]) {
 
   ClusterSpec shards;
   if (protocol == RESP) {
-    shards = proactor->Await([&] { return FetchCluster(ep, proactor); });
+    shards = proactor->Await([&] {
+      unique_ptr<FiberSocketBase, FiberSocketClose> socket(proactor->CreateSocket());
```
kostasrim, line 980: +1 for `FiberSocketClose`. I am leaving this here for info, we also have `absl::Cleanup` which will perform a cleanup upon destruction. So `absl::Cleanup raii([&socket] { socket->close(); })` and because the order of destructors is reversed and the construction of cleanup comes after the construction of the uring socket we close at the end without passing this FiberSocketClose functor. 

The only reason I comment here, is so you can be aware that there is this utlity function which can provide this dirty RAII. 

p.s. FiberSocketClose is perfectly valid here so zero issues with that

mkaruza: Thanks for info about `absl::Cleanup` i'll switch to cleanup with it since it is used through the code.

kostasrim: I personally don't have any preference or care. It's quick and it works. Use whatever you like :smile:

src/server/dfly_bench.cc
```diff
@@ -35,6 +36,7 @@ extern "C" {
 using std::string;
 
 ABSL_FLAG(uint16_t, p, 6379, "Server port");
+ABSL_FLAG(uint16_t, pool_size, 0, "Number of threads");
```
romange, line 39: you do not need this flag. any helio-powered backend has `proactor_threads` flag that does the same.
you can run `--help` to see all flags.

src/server/dfly_bench.cc
```diff
@@ -647,9 +731,10 @@ void Driver::ParseRESP() {
           CHECK(absl::SimpleAtoi(parts[0], &slot_id));
           vector<string_view> host_port = absl::StrSplit(parts[1], ':');
           CHECK_EQ(host_port.size(), 2u);
-
-          // TODO: to support slot migration.
           LOG_EVERY_T(INFO, 1) << "Moved: " << slot_id << " to " << parts[1];
+          // Slot is moved so update slots information for connecting shard only
+          ClusterSpec shard = FetchCluster(socket_.get(), true);
+          slots_ = shard[0].slots;
```
romange, line 737: why shard 0? for differrent shards it will be different slots.
this driver got `moved` response due to slot not being owned by this shard but where do you handle the case where a shard (driver) needs to know about slots being added to it.

romange: i see it's draft, nm

mkaruza: I have modified FetchCluster function, seting second parameter true will filter cluster information and return only shard node (one that is identified with myself) - so we'll only have one vector element

src/server/dfly_bench.cc
mkaruza: Don't see this `proactor_threads` argument when running `dfly_bench --help` so i have added it

romange: sorry, try `--helpfull`. `--help` shows only the flags from the main module by default.

src/server/dfly_bench.cc
romange: ok, it's not possible to see diffs now for FetchCluster. In any case please add a comment explaining why `0` is fine here.

src/server/dfly_bench.cc
mkaruza: Removed flag

src/server/dfly_bench.cc
```diff
@@ -119,6 +120,88 @@ struct ShardInfo {
 
 using ClusterSpec = vector<ShardInfo>;
 
+ClusterSpec FetchCluster(FiberSocketBase* socket, bool connection_shard_only = false) {
```
romange, line 126: not important now, but please take into account - we almost never use optional arguments, 
I consider them confusing when reading the code (do not quote me but I agree with Rust folks on this).

romange (dismissed):
lgtm

romange:
actually, I still do not understand something. how a driver that has new slots adds them to his slot ranges? I only see that we update those drivers that stumbled upon MOVED but others may not get it because they just got additional slots

mkaruza:
> actually, I still do not understand something. how a driver that has new slots adds them to his slot ranges? I only see that we update those drivers that stumbled upon MOVED but others may not get it because they just got additional slots

Each connection driver have copy of slots that he is requesting. When `MOVED` is received he will update only his local slots. Other drivers that are connected to same node will (probably at some point) get MOVED error and than update their own slot ranges.

This is initial implementation that keeps most of existing execution logic in place. Another approach would be that drivers connecting on same node share slot range vector - but this will require, i expect, additional synchronization between them.

romange:
Well, this does not cover all the cases. Consider slots are moved from shard A to shard B:
`A` will indeed send MOVED but `B` will never send MOVED and these slots won't be covered by the load test.

mkaruza:
> Well, this does not cover all the cases. Consider slots are moved from shard A to shard B: `A` will indeed send MOVED but `B` will never send MOVED and these slots won't be covered by the load test.

Yes, that is not covered. My thought about this was it is not needed simply because these slots, if moved to node A, will not be selected and you still pick random slot from existing ranges. I'll look to handle this case also.

src/server/dfly_bench.cc
```diff
@@ -645,11 +715,20 @@ void Driver::ParseRESP() {
...
+          char ip_addr[INET6_ADDRSTRLEN];
+          std::error_code ec = fb2::DnsResolve(host, ip_addr);
+          CHECK(!ec) << "Could not resolve " << host << " " << ec;
+          auto address = ::boost::asio::ip::make_address(ip_addr);
+
+          uint32_t port;
+          CHECK(absl::SimpleAtoi(addr_parts[1], &port));
+          CHECK_LT(port, 65536u);
+
+          shard_slots_.MoveSlot(ep_, tcp::endpoint(address, port), slot_id);
```
romange, line 731: the way I see it, Dragonfly/Redis should respond with a consistent language, so given the fact that our shards are the same during the test, you need to resolve once when the test started (resolution itself is an implementation detail of the Connect function).

romange:
LGTM

romange (approved):

PR #4763, opened 2025-03-14
fix: Sleep after subscribed to channels

mkaruza:
Introduce sleep after subscribed to channels to give time that channel is propagated.


tests/dragonfly/cluster_test.py
```diff
@@ -3009,6 +3009,10 @@ async def test_cluster_sharded_pubsub_shard_commands(df_factory: DflyInstanceFac
     consumer.ssubscribe("pubsub-shard-channel")
     consumer.ssubscribe("shard-channel")
 
+    # We need to sleep cause we use DispatchBrief internally. Otherwise we can't really gurantee
+    # that the client received the message
+    await asyncio.sleep(1)
```
kostasrim, line 3014: I am unsure about this here. The problem on the other test was that we `unsubscribed` and then the client received a published message after it unsubscribed. This was because of the channel store being updated with memory order relaxed which was fixed. 

Furthermore, I added this sleep here https://github.com/dragonflydb/dragonfly/pull/4740/files#diff-84f62368c6e1619d79a05fda3fd46406c8cee4cd215f6469eb00ebfa091a6c9dR2984 and I am `99.9%` sure it's redundant and it's not needed. Why ? Because, the problem there is that when we tried to `publish a message after we unsubscribed` we somehow still saw as subscribed the "unsubscribed" client. Why? Because the channel store was loaded with the wrong memory order. 

However here, we fail when we call `PUBSUB SHARDCHANNELS` and after we call two times `SSUBSCRIBE` above. There is not `DispatchBrief` in this path I am afraid (it's only used when we Publish a message and we don't do something like this here) so by the time `ssubscribe` returns to the client, the `channel store should contain the two channels`. However, for some reason, when we poll the channel store via `PUBSUB SHARDCHANNELS` it only shows us once. 

There is some consistency issue here of the channel store that we need to understand and fix

tests/dragonfly/cluster_test.py
```diff
@@ -3000,14 +3000,10 @@ async def test_cluster_sharded_pubsub_shard_commands(df_factory: DflyInstanceFac
...
-
-    consumer_client = RedisCluster(startup_nodes=[node_a, node_b])
-    consumer = consumer_client.pubsub()
-
-    consumer.ssubscribe("pubsub-shard-channel")
-    consumer.ssubscribe("shard-channel")
+    # We are executing SSUBSCRIBE commands and wait for them to be sure that
+    # channels are created
+    message = await c_nodes[0].execute_command("SSUBSCRIBE pubsub-shard-channel")
+    message = await c_nodes[0].execute_command("SSUBSCRIBE shard-channel")
```
kostasrim, line 3006: When I get a chance I will check why `consumer.ssubscribe` is async (it might be a bug on redis-py worth reporting)

kostasrim (approved):
LGTM :man_cook:

Issue #4537, opened 2025-01-30
allow scanning based on keys attributes


romange:
Add another filter to the SCAN command:

`ATTR <mask>` 

to be able to scan keys based on additional attributes:
mask may contain the following attributes:

1. `v` - volatile, keys that have ttl
2. `p` - permanent, keys that do not have ttl
3. `a` - accessed, the key has been accessed since the last load/flush event, or the last time `a` flag was reset.
4. `u` - untouched, the key has not been accessed/touched.

in the future:
* `d` the value resides on disk
* `m` the value resides in memory

for (3), (4) we need to introduce an additional flag in CompactObject to track it on keys (note that usually the flags keys are set on values).
Also we will need a `DEBUG ACCESS RESET` to reset `a` attributes for the current database.

Note that we already provide undocumented "BUCKET <id>" filter for SCAN.

lichuang:
Is there anyone taken this issue, i would like to have a try

kostasrim:
Sure @lichuang 🎉

PR #4766, opened 2025-03-14
feat(server): SCAN command add ATTR options

lichuang:
feat(server): SCAN command add ATTR options

fix https://github.com/dragonflydb/dragonfly/issues/4537

### DIFF
--- a/src/server/generic_family.cc
+++ b/src/server/generic_family.cc
@@ -587,7 +587,17 @@ bool ScanCb(const OpArgs& op_args, PrimeIterator prime_it, const ScanOpts& opts,
   }
 
   bool matches = !opts.type_filter || it->second.ObjType() == opts.type_filter;
-
+  if (opts.mask.has_value()) {
+    if (opts.mask == ScanOpts::Mask::Volatile) {
+      matches &= it->second.HasExpire();
+    } else if (opts.mask == ScanOpts::Mask::Permanent) {
+      matches &= !it->second.HasExpire();
+    } else if (opts.mask == ScanOpts::Mask::Accessed) {
+      matches &= it->first.WasTouched();
+    } else if (opts.mask == ScanOpts::Mask::Untouched) {
+      matches &= !it->first.WasTouched();
+    }
+  }
   if (!matches)
     return false;
 
@@ -1735,6 +1745,7 @@ void GenericFamily::Echo(CmdArgList args, const CommandContext& cmd_cntx) {
 }
 
 // SCAN cursor [MATCH <glob>] [TYPE <type>] [COUNT <count>] [BUCKET <bucket_id>]
+// [ATTR <mask>]
 void GenericFamily::Scan(CmdArgList args, const CommandContext& cmd_cntx) {
   string_view token = ArgS(args, 0);
   uint64_t cursor = 0;
--- a/src/server/generic_family_test.cc
+++ b/src/server/generic_family_test.cc
@@ -412,6 +412,49 @@ TEST_F(GenericFamilyTest, Scan) {
   EXPECT_EQ(resp, "");
 }
 
+TEST_F(GenericFamilyTest, ScanWithAttr) {
+  Run({"flushdb"});
+
+  Run({"set", "hello", "world"});
+  Run({"set", "foo", "bar"});
+
+  Run({"expire", "hello", "1000"});
+
+  auto resp = Run({"scan", "0", "attr", "v"});
+  auto vec = StrArray(resp.GetVec()[1]);
+  ASSERT_EQ(1, vec.size());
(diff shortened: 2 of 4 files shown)


lichuang:
@romange 

> for (3), (4) we need to introduce an additional flag in CompactObject to track it on keys (note that usually the flags keys are set on values).

There is `TOUCHED` in `CompactObject`, is it can be used for (3) (4)?

Another question, i want to add some python test cases, but redis.Redis scan function has only these options:
> SCAN cursor [MATCH pattern] [COUNT count] [TYPE type]
(see: https://redis.io/docs/latest/commands/scan/)

romange:
Yes, good point. We mark the keys and do it irrespective of tiering, so it should work well for this issue. 
We will also need unit tests to test this behaviour.

Regarding the testing - I believe this falls well into the unit test territory, I would add the tests to `generic_family_test` (it is possible to use pytest redis client but I think it's just an overkill)

src/server/generic_family_test.cc
```diff
@@ -412,6 +412,49 @@ TEST_F(GenericFamilyTest, Scan) {
...
+  resp = Run({"scan", "0", "attr", "a"});
+  vec = StrArray(resp.GetVec()[1]);
+  EXPECT_EQ(0, vec.size());
+
+  // before run get "foo", scan with a attr should return "foo" and "hello"
+  resp = Run({"scan", "0", "attr", "u"});
+  vec = StrArray(resp.GetVec()[1]);
+  EXPECT_EQ(2, vec.size());
+
+  ASSERT_THAT(Run({"get", "foo"}), "bar");
```
lichuang, line 443: @romange  In this test case, i found that running "get" of some key, it wiil not make the key `touched`, so this test case failed, can you have a look what is wrong.

src/server/generic_family_test.cc
```diff
@@ -412,6 +412,49 @@ TEST_F(GenericFamilyTest, Scan) {
...
+
+  // after run get "foo", scan with a attr should return "foo"
+  resp = Run({"scan", "0", "attr", "a"});
+  vec = StrArray(resp.GetVec()[1]);
+  EXPECT_EQ(1, vec.size());
```
romange, line 448: please use ASSERT_EQ variations on comparing sizes to prevent invalid access further

src/server/generic_family_test.cc
romange: it makes the key touched and you can verify it by adding debug printings for example at line db_slice.cc:624

src/server/generic_family.cc
```diff
@@ -587,7 +587,17 @@ bool ScanCb(const OpArgs& op_args, PrimeIterator prime_it, const ScanOpts& opts,
...
 
   bool matches = !opts.type_filter || it->second.ObjType() == opts.type_filter;
-
+  if (opts.mask.has_value()) {
+    if (opts.mask == ScanOpts::Mask::Volatile) {
+      matches = it->second.HasExpire();
+    } else if (opts.mask == ScanOpts::Mask::Permanent) {
+      matches = !it->second.HasExpire();
+    } else if (opts.mask == ScanOpts::Mask::Accessed) {
+      matches = it->second.WasTouched();
```
romange, line 596: if you check db_slice.cc:624 you will see that Touched attribute refers to the key, i.e. it->first

src/server/generic_family.cc
```diff
@@ -587,7 +587,17 @@ bool ScanCb(const OpArgs& op_args, PrimeIterator prime_it, const ScanOpts& opts,
...
+    } else if (opts.mask == ScanOpts::Mask::Untouched) {
+      matches = !it->second.WasTouched();
```
romange, line 598: same here

romange:
@lichuang a friendly ping :)
you almost finished, just checking if you are planning to address the comments :)

lichuang:
> @lichuang a friendly ping :) you almost finished, just checking if you are planning to address the comments :)

sorry, i am a little busy these days, i will finish it ASAP.

src/server/generic_family_test.cc
lichuang: > it makes the key touched and you can verify it by adding debug printings for example at line db_slice.cc:624

i see what is wrong, HasExpire() works on it->second object(which is value), while HasTouched() works on it->first object(which is key), is that expected?

romange: yes, this by design though I agree it's confusing.
the reason is that if you call "SADD foo bar"   and then `SET foo bar`  then all the attributes on the value (second) will be reset. if we want to preserve the attributes we should use keys (first)

lichuang: now this pr is ready for review :)

romange (approved):

PR #4777, opened 2025-03-16
fix(json_family): Fix memory tracking for JSON

BagritsevichStepan:
fixes dragonflydb#4725

Bugs that were found:
1. The first bug was related to conversions between `size_t` and `int64_t`. We performed `size_t` += `int64_t` for negative values of `int64_t`. In this case, `int64_t` is first converted to a large `size_t` value before being added to `bytes_used`.
Additionally, this check doesn't make sense:
```cpp
if (size < 0) {
    DCHECK(static_cast<int64_t>(u_.json_obj.cons.bytes_used) >= size);
}
```

- Because `size_t` is always ≥ 0, while `size` here is < 0.
- I suggest avoiding operations that can potentially cause an overflow and adding utility functions to handle such cases safely.

2. The second bug was that `db_slice.AddOrFind` could deallocate some memory, at least due to `owner_->tiered_storage()->ReclaimMemory`. And `JsonMemTracker` was created before calling the `db_slice.AddOrFind` method. As a result, we initialized `start_size_` in `JsonMemTracker` with the currently used memory. However, `AddOrFind` might deallocate some memory, and later, when we call `SetJsonSize` after inserting a new JSON and calculating the current memory usage, the reported usage may decrease because `AddOrFind` has freed some memory. Here is the relevant snippet:
```
E20250318 16:19:37.185142 107026 json_family.cc:61] JsonMemTracker start size: 41418008
E20250318 16:19:37.185150 107026 json_family.cc:85] JsonMemTracker size: 41418008
E20250318 16:19:37.185191 107026 json_family.cc:85] JsonMemTracker size: 41417000
E20250318 16:19:37.185196 107026 json_family.cc:85] JsonMemTracker size: 41417000
E20250318 16:19:37.185209 107026 json_family.cc:85] JsonMemTracker size: 41417544
E20250318 16:19:37.185214 107026 json_family.cc:66] JsonMemTracker current size in SetJsonSize: 41417544
F20250318 16:19:37.185221 107026 compact_object.cc:64] Check failed: result >= 0 (-448 vs. 0) Can't decrease 0 from 448
```
The initial size is 41,418,008. We print the current size before calling the `AddOrFind` method (which is also 41,418,008), and then we print it again after the method call. As we can see, the size decreases to 41417000 .

I also think that I found another bugs in json memory tracking. So, I'm suggesting to add an `json_mem_tracking` tests.

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -43,6 +43,13 @@ namespace {
 constexpr XXH64_hash_t kHashSeed = 24061983;
 constexpr size_t kAlignSize = 8u;
 
+// This is needed for LOG_EVERY_N to work.
+#ifdef NDEBUG
+#define FATAL_IF_DEBUG ERROR
+#else
+#define FATAL_IF_DEBUG FATAL
+#endif
+
 // Approximation since does not account for listpacks.
 size_t QlMAllocSize(quicklist* ql, bool slow) {
   size_t node_size = ql->len * sizeof(quicklistNode) + znallocx(sizeof(quicklist));
@@ -61,7 +68,8 @@ size_t UpdateSize(size_t size, int64_t update) {
   }
 
   int64_t result = static_cast<int64_t>(size) + update;
-  DCHECK_GE(result, 0) << "Can't decrease " << size << " from " << -update;
+  LOG_IF_EVERY_N(FATAL_IF_DEBUG, result < 0, 30)
+      << "Can't decrease " << size << " from " << -update;
   return result;
 }
 
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -43,13 +43,6 @@ namespace {
 constexpr XXH64_hash_t kHashSeed = 24061983;
 constexpr size_t kAlignSize = 8u;
 
-// This is needed for LOG_EVERY_N to work.
-#ifdef NDEBUG
-#define FATAL_IF_DEBUG ERROR
-#else
-#define FATAL_IF_DEBUG FATAL
-#endif
-
 // Approximation since does not account for listpacks.
 size_t QlMAllocSize(quicklist* ql, bool slow) {
   size_t node_size = ql->len * sizeof(quicklistNode) + znallocx(sizeof(quicklist));
@@ -68,8 +61,10 @@ size_t UpdateSize(size_t size, int64_t update) {
   }
 
   int64_t result = static_cast<int64_t>(size) + update;
-  LOG_IF_EVERY_N(FATAL_IF_DEBUG, result < 0, 30)
-      << "Can't decrease " << size << " from " << -update;
+  if (result < 0) {
+    DCHECK(false) << "Can't decrease " << size << " from " << -update;
+    LOG_EVERY_N(ERROR, 30) << "Can't decrease " << size << " from " << -update;
+  }
   return result;
 }
 
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -55,6 +55,16 @@ size_t QlMAllocSize(quicklist* ql, bool slow) {
   return node_size + ql->count * 16;  // we account for each member 16 bytes.
 }
 
+size_t UpdateSize(size_t size, int64_t update) {
+  if (update >= 0) {
+    return size + update;
+  }
+
+  int64_t result = static_cast<int64_t>(size) + update;
+  DCHECK_GE(result, 0) << "Can't decrease " << size << " from " << -update;
+  return result;
+}
+
 inline void FreeObjSet(unsigned encoding, void* ptr, MemoryResource* mr) {
   switch (encoding) {
     case kEncodingStrMap2: {
@@ -927,10 +937,7 @@ void CompactObj::SetJson(JsonType&& j) {
 void CompactObj::SetJsonSize(int64_t size) {
   if (taglen_ == JSON_TAG && JsonEnconding() == kEncodingJsonCons) {
     // JSON.SET or if mem hasn't changed from a JSON op then we just update.
-    if (size < 0) {
-      DCHECK(static_cast<int64_t>(u_.json_obj.cons.bytes_used) >= size);
-    }
-    u_.json_obj.cons.bytes_used += size;
+    u_.json_obj.cons.bytes_used = UpdateSize(u_.json_obj.cons.bytes_used, size);
   }
 }
 
--- a/src/server/json_family.cc
+++ b/src/server/json_family.cc
@@ -339,6 +339,8 @@ OpResult<DbSlice::ItAndUpdater> SetJson(const OpArgs& op_args, string_view key,
 
   op_args.shard->search_indices()->RemoveDoc(key, op_args.db_cntx, res.it->second);
 
+  JsonMemTracker tracker;
+
   std::optional<JsonType> parsed_json = ShardJsonFromString(json_str);
   if (!parsed_json) {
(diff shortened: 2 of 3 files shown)


src/server/json_family.cc
```diff
@@ -354,6 +355,9 @@ OpResult<DbSlice::ItAndUpdater> SetJson(const OpArgs& op_args, string_view key,
   } else {
     res.it->second.SetJson(std::move(*parsed_json));
   }
+
+  tracker.SetJsonSize(res.it->second, res.is_new);
+
```
BagritsevichStepan, line 360: I think we should call `res.it->post_updater.Run()` here, but I'm not entirely sure why it wasn't added before

src/core/compact_object.cc
```diff
@@ -55,6 +55,29 @@ size_t QlMAllocSize(quicklist* ql, bool slow) {
...
+    unsigned_update = kMaxUpdate + 1;
+  }
+
+  if (unsigned_update > std::numeric_limits<size_t>::max()) {
+    LOG(DFATAL) << "Overflow. Update size is too large. Size: " << size << ", update: " << update;
+    return 0;
+  }
+
+  DCHECK(static_cast<uint64_t>(size) >= unsigned_update);
+  return size - unsigned_update;
```
kostasrim, line 78: size_t has a range of 2^64 and int 2^(64-1) signed/unsigned. It's guranteed that we won't see a size (in bytes) as big as 2^64 so we can just do `static_cast<int64_t>` instead of all of this. In fact we do this in other places as well (see AccountObjectMemory in db_slice). 


```
    DCHECK_GE(static_cast<int64_t>(stats.obj_memory_usage) + size, 0)
        << "Can't decrease " << size << " from " << stats.obj_memory_usage;

```

kostasrim: Good catch for the implicit conversion of the signed operand to unsigned. That was mistake.

src/server/json_family.cc
kostasrim: plz explain

src/server/json_family.cc
```diff
@@ -1329,10 +1333,8 @@ OpResult<bool> OpSet(const OpArgs& op_args, string_view key, string_view path,
       }
     }
 
-    JsonMemTracker mem_tracker;
```
kostasrim, line 1332: Can you explain what kind of problems you see here ?

src/core/compact_object.cc
BagritsevichStepan: Yes, you are right

src/server/json_family.cc
BagritsevichStepan: SetJson method removes old json (if it presents) and inserts the new one, so after updating the data value we should to call post_updater. Because the value have been changed and at least memory usage is changed, that's why we are calling here `SetJsonSize`

As example I'm checking the function JsonMutateOperation, which calls it [here](https://github.com/dragonflydb/dragonfly/blob/main/src/server/json_family.cc#L651).

I also see what that AddOrFind returns AutoUpdater that calls post update on the DbSlice, which updates memory stats and checked if this key is watched or not. In theory we can reproduce it with pub sub.

So after SetJsonSize we should call res.it->post_updater.Run() I think.

src/server/json_family.cc
BagritsevichStepan: I was getting negative values for `diff` here:
```
void SetJsonSize(PrimeValue& pv, bool is_op_set) {
    const size_t current = static_cast<MiMemoryResource*>(CompactObj::memory_resource())->used();
    int64_t diff = static_cast<int64_t>(current) - static_cast<int64_t>(start_size_);
```
But before, we didn't have this key, and we added it, so the current memory usage should only increase. However, I observed a different behavior here — the `current` memory usage was actually lower than `start_size_`.

src/server/json_family.cc
kostasrim: hmm maybe there is something that I am missing but don't we call the `AutoUpdater::Run()` also as part of the `destructor` ? See:

```
   454 DbSlice::AutoUpdater::~AutoUpdater() {
   455   Run();
   456 }
```

So unless I am missing something we run the `AutoUpdater` anyway. My question is why we manually run it before we add to the vector of the search indices. Maybe there is a reason maybe there isn't -- either way we should check on that :smile:

src/server/json_family.cc
kostasrim: But that makes sense right ? So you did not have an object at first. You create one, and let's say its size now is 500bytes. Now we run json set again with an object that requires 300 bytes. What's the delta ? `-200`

src/server/json_family.cc
BagritsevichStepan: Ah, yeah, I missed this. This should be ok

src/server/json_family.cc
BagritsevichStepan: I didn't understand you. The code here is correct, but for some reason, I'm getting negative values for the newly added key. The size was 0, and the difference was somehow -790. Then we call `pv.SetJsonSize(diff);` with a negative value, and the `DCHECK(static_cast<uint64_t>(size) >= unsigned_update);` fails because we're trying to decrease the size from 0 by 790 bytes.

BagritsevichStepan: After debugging we know the exact issue. I updated the PR description

src/server/json_family.cc
```diff
@@ -953,6 +957,7 @@ OpResult<long> OpDel(const OpArgs& op_args, string_view key, string_view path,
   PrimeValue& pv = it_res->it->second;
   JsonType* json_val = pv.GetJson();
 
+  JsonMemTracker tracker;
```
BagritsevichStepan, line 960: I also decided to move `JsonMemTracker` slightly in the `OpDel` command to initialize it after the `FindMutable` method. I'm not sure if `FindMutable` can `deallocate` anything, but I did it just for safety.

src/core/compact_object.cc
```diff
@@ -55,6 +55,16 @@ size_t QlMAllocSize(quicklist* ql, bool slow) {
   return node_size + ql->count * 16;  // we account for each member 16 bytes.
 }
 
+size_t UpdateSize(size_t size, int64_t update) {
+  if (update >= 0) {
+    return size + update;
+  }
+
+  int64_t result = static_cast<int64_t>(size) + update;
+  DCHECK_GE(result, 0) << "Can't decrease " << size << " from " << -update;
```
kostasrim, line 64: I would also consider logging it ever `n` seconds so we are aware

kostasrim (dismissed):

src/core/compact_object.cc
```diff
@@ -55,6 +55,19 @@ size_t QlMAllocSize(quicklist* ql, bool slow) {
   return node_size + ql->count * 16;  // we account for each member 16 bytes.
 }
 
+size_t UpdateSize(size_t size, int64_t update) {
+  if (update >= 0) {
+    return size + update;
+  }
+
```
kostasrim, line 62:
```suggestion
```

kostasrim: It's fine to do the arithmetic with the cast :)

kostasrim (approved):

src/core/compact_object.cc
kostasrim: But you can also remove it on your other json PR :)

BagritsevichStepan: Ok

PR #4778, opened 2025-03-16
chore: prefetch keys in opmget

romange:
Also, stop deduplicating keys by default, but keep it as a run-time flag.
All in all, this PR improves MGET (100% miss) throughput by at least 7%:
from 1.23M to 1.32M qps.
(100% miss traffic allows measuring more effectively the impact of this code)

Also, finally fix TieredStorageTest.FlushAll test.

### DIFF
--- a/src/server/string_family.cc
+++ b/src/server/string_family.cc
@@ -35,6 +35,10 @@
 #include "server/transaction.h"
 #include "util/fibers/future.h"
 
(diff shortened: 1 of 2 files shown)


BorysTheDev (approved):

dranikpg (commented):
Factor in the cost of computing key hashes twice 🙄

PR #4779, opened 2025-03-16
feat(search_family): Add basic support for the FT.CONFIG

BagritsevichStepan:
fixes dragonflydb#4352

Add basic support of the MAXSEARCHRESULTS and MAXAGGREGATERESULTS


src/server/search/search_family.cc
```diff
@@ -599,6 +599,111 @@ void SearchReply(const SearchParams& params, std::optional<search::AggregationIn
...
+template <typename V>
+std::optional<V> FindOptionsMapValue(const SearchFamily::ConfigOptionsMap<V>& options,
+                                     std::string_view name) {
+  auto it = std::find_if(options.begin(), options.end(), [name](const auto& opt) {
+    return absl::EqualsIgnoreCase(opt.first, name);
+  });
+  if (it != options.end()) {
+    return it->second;
+  }
+  return std::nullopt;
```
BorysTheDev, line 620: return it != options.end() ? it->second; std::optional<V>();

src/server/search/search_family.cc
```diff
@@ -966,6 +1078,31 @@ void SearchFamily::FtProfile(CmdArgList args, const CommandContext& cmd_cntx) {
...
+util::fb2::Mutex SearchFamily::config_mu_;
+
+void SearchFamily::FtConfig(CmdArgList args, const CommandContext& cmd_cntx) {
+  CmdArgParser parser{args};
+  auto* rb = static_cast<RedisReplyBuilder*>(cmd_cntx.rb);
+
+  auto func = parser.TryMapNext("GET", &FtConfigGet, "SET", &FtConfigSet, "HELP", &FtConfigHelp);
+
+  if (func) {
+    util::fb2::LockGuard lg{config_mu_};
```
BorysTheDev, line 1096: we try to avoid such mutexes. Could you consider copy or shared_ptr of data

src/server/search/search_family.h
```diff
@@ -27,11 +37,12 @@ class SearchFamily {
   static void FtList(CmdArgList args, const CommandContext& cmd_cntx);
   static void FtSearch(CmdArgList args, const CommandContext& cmd_cntx);
   static void FtProfile(CmdArgList args, const CommandContext& cmd_cntx);
+  static void FtConfig(CmdArgList args, const CommandContext& cmd_cntx);
   static void FtAggregate(CmdArgList args, const CommandContext& cmd_cntx);
   static void FtTagVals(CmdArgList args, const CommandContext& cmd_cntx);
 
- public:
-  static void Register(CommandRegistry* registry);
+  static util::fb2::Mutex config_mu_;
```
BorysTheDev, line 44: we have only one instance of SearchFamily. I think you can drop static

src/server/search/search_family.cc
BorysTheDev: I mean to prevent removing or changing config during execution without mutex. If it possible I will be happy

src/server/search/search_family.h
BagritsevichStepan: But we need to use this inside the static functions

romange: right but you can just define it in cc file in the anonymous namespace.

BorysTheDev: @romange I suggest to leave mutexes in header, in this case we can use ABSL_LOCKS_EXCLUDED()

BagritsevichStepan: We decided to remove mutexes

src/server/search/search_family.cc
```diff
@@ -1089,18 +1239,19 @@ void SearchFamily::Register(CommandRegistry* registry) {
...
-            << CI{"FT.SEARCH", kReadOnlyMask, -3, 0, 0, acl::FT_SEARCH}.HFUNC(FtSearch)
-            << CI{"FT.AGGREGATE", kReadOnlyMask, -3, 0, 0, acl::FT_SEARCH}.HFUNC(FtAggregate)
-            << CI{"FT.PROFILE", kReadOnlyMask, -4, 0, 0, acl::FT_SEARCH}.HFUNC(FtProfile)
-            << CI{"FT.TAGVALS", kReadOnlyMask, 3, 0, 0, acl::FT_SEARCH}.HFUNC(FtTagVals);
+  *registry
+      << CI{"FT.CREATE", CO::WRITE | CO::GLOBAL_TRANS, -2, 0, 0, acl::FT_SEARCH}.HFUNC(FtCreate)
+      << CI{"FT.ALTER", CO::WRITE | CO::GLOBAL_TRANS, -3, 0, 0, acl::FT_SEARCH}.HFUNC(FtAlter)
+      << CI{"FT.DROPINDEX", CO::WRITE | CO::GLOBAL_TRANS, -2, 0, 0, acl::FT_SEARCH}.HFUNC(
+             FtDropIndex)
+      << CI{"FT.CONFIG", CO::WRITE | CO::GLOBAL_TRANS, -3, 0, 0, acl::FT_SEARCH}.HFUNC(FtConfig)
```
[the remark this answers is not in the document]
BorysTheDev: Maybe you also need to add NO_KEY_TRANSACTIONAL

src/server/search/search_family.cc
```diff
@@ -24,8 +24,10 @@
 #include "server/conn_context.h"
 #include "server/container_utils.h"
 #include "server/engine_shard_set.h"
+#include "server/main_service.h"
```
BorysTheDev, line 27: why do you need it?

src/server/search/search_family.cc
```diff
@@ -24,8 +24,10 @@
...
 #include "server/search/aggregator.h"
 #include "server/search/doc_index.h"
+#include "server/server_family.h"
```
BorysTheDev, line 30: please remove

src/server/search/search_family.cc
```diff
@@ -966,6 +1080,50 @@ void SearchFamily::FtProfile(CmdArgList args, const CommandContext& cmd_cntx) {
...
+
+  if (auto err = parser.Error(); err)
+    return rb->SendError(err->MakeReply());
+
+  auto option_exists = FindOptionsMapValue(kConfigOptions, option);
+  if (option_exists) {
+    DVLOG(2) << "Setting " << option << " to " << value;
+
+    auto cb = [option, value](util::ProactorBase*) { UpdateConfigOption(option, value); };
+    server_family_->service().proactor_pool().AwaitFiberOnAll(std::move(cb));
```
BorysTheDev, line 1098: what will happen if we have 2 simultaneous commands?

BagritsevichStepan:
Temporary close this, because it is not priority for now

PR #4786, opened 2025-03-17
feat(server): Cluster MOVED response Prometheus metric

### LINKED ISSUE #4749: Cluster `MOVED` response Prometheus metric
It would be useful to have a Prometheus metric for how many `MOVED` errors Dragonfly is returning

Not urgent but would be nice as querying `INFO` doesn't scale well when you have 10s of shards

mkaruza:
Get number of MOVED reported errors and send as Prometheus metric.

Closes #4749

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -1532,7 +1532,7 @@ void PrintPrometheusMetrics(uint64_t uptime, const Metrics& m, DflyCmd* dfly_cmd
     if (m.facade_stats.reply_stats.err_count.contains("MOVED")) {
       moved_total_errors = m.facade_stats.reply_stats.err_count.at("MOVED");
     }
-    AppendMetricHeader("moved_errors_total", "Total error numbers of moved slot errors",
+    AppendMetricHeader("moved_errors_total", "Total number of moved slot errors",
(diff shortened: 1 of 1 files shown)


BorysTheDev (approved):

BorysTheDev (approved):

src/server/server_family.cc
```diff
@@ -1520,12 +1520,23 @@ void PrintPrometheusMetrics(uint64_t uptime, const Metrics& m, DflyCmd* dfly_cmd
...
+                       MetricType::GAUGE, &migration_errors_str);
     AppendMetricValue("migration_errors_total", m.migration_errors_total, {"num"},
-                      {"migration errors"}, &str);
-    absl::StrAppend(&resp->body(), str);
+                      {"migration errors"}, &migration_errors_str);
+    absl::StrAppend(&resp->body(), migration_errors_str);
+
+    string moved_errors_str;
+    uint64_t moved_total_errors = 0;
+    if (m.facade_stats.reply_stats.err_count.contains("MOVED")) {
```
romange, line 1532: here it's okeyish, but I prefer we do not use double-find patterns. This is why ~~god~~ stroustroup created iterators.

PR #4788, opened 2025-03-17
Fix extended node info comparison

BorysTheDev:
problem: config is ignored if only health status is changed


romange (approved):

romange:
@BorysTheDev consider covering it with unit tests if possible in the next PRs.

PR #4802, opened 2025-03-20
fix(hset_family): Fix crash on scan after expiry set

abhijat:
When expiry is set on an hset its members are migrated to a string map.

When creating a new entry in string map with TTL, we also set a bit on the stored pointer as metadata, this is set as 1 << 63. As a consequence when reading back the pointer, if expiry is set, we need to unset the same bit to get the correct address, this is done already in string map internals.

This change adds a similar AND operation in the scan code to unset the TTL bit before dereferencing the pointer.

fixes https://github.com/dragonflydb/dragonfly/issues/4799

### DIFF
--- a/src/core/string_map.h
+++ b/src/core/string_map.h
@@ -147,9 +147,6 @@ class StringMap : public DenseSet {
   void RandomPairs(unsigned int count, std::vector<sds>& keys, std::vector<sds>& vals,
                    bool with_value);
 
-  static constexpr uint64_t kValTtlBit = 1ULL << 63;
-  static constexpr uint64_t kValMask = ~kValTtlBit;
-
   static sds GetValue(sds key, std::optional<size_t> len = std::nullopt);
 
  private:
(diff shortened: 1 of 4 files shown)


src/core/string_map.h
```diff
@@ -147,6 +147,11 @@ class StringMap : public DenseSet {
   void RandomPairs(unsigned int count, std::vector<sds>& keys, std::vector<sds>& vals,
                    bool with_value);
 
+  static constexpr uint64_t kValTtlBit = 1ULL << 63;
```
romange, line 150: there is not need to move these constants here

src/core/string_map.h
```diff
@@ -147,6 +147,8 @@ class StringMap : public DenseSet {
   void RandomPairs(unsigned int count, std::vector<sds>& keys, std::vector<sds>& vals,
                    bool with_value);
 
+  static sds GetValue(sds key, std::optional<size_t> len = std::nullopt);
```
romange, line 150: I do not think that passing `len` is justified here: the interface becomes more confusing so that we could save `sdslen()` call which is very fast.

abhijat: fixed

romange (approved):

adiholden (approved):

PR #4807, opened 2025-03-21
fix: Backward compatibility for fields without the "@" sign. #4698

vyavdoshenko:
fix: Backward compatibility for fields without the "@" sign. https://github.com/dragonflydb/dragonfly/issues/4698

### DIFF
--- a/src/server/search/search_family.cc
+++ b/src/server/search/search_family.cc
@@ -5,6 +5,7 @@
 #include "server/search/search_family.h"
 
 #include <absl/container/flat_hash_map.h>
+#include <absl/flags/flag.h>
 #include <absl/strings/ascii.h>
 #include <absl/strings/match.h>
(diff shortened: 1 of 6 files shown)


BagritsevichStepan:
I think you may have overdone the task a bit

src/server/search/search_family.cc
```diff
@@ -289,18 +290,6 @@ std::string_view ParseField(CmdArgParser* parser) {
   return field;
 }
 
-std::string_view ParseFieldWithAtSign(CmdArgParser* parser) {
```
BagritsevichStepan, line 292: Please also add a test for search_family_test, that describes the compatibility problem and test behaviour for the set flag

vyavdoshenko: I added an additional test for the parser.

PR #4810, opened 2025-03-21
fix: Backward compatibility for fields without the "@" sign in the FT.AGGREGATE command

vyavdoshenko:
fix: Backward compatibility for fields without the "@" sign in the FT.AGGREGATE command #4698

### DIFF
--- a/src/server/search/search_family.cc
+++ b/src/server/search/search_family.cc
@@ -5,6 +5,7 @@
 #include "server/search/search_family.h"
 
 #include <absl/container/flat_hash_map.h>
+#include <absl/flags/flag.h>
 #include <absl/strings/ascii.h>
 #include <absl/strings/match.h>
 #include <absl/strings/str_format.h>
@@ -28,6 +29,9 @@
 #include "server/search/doc_index.h"
 #include "server/transaction.h"
 #include "src/core/overloaded.h"
+
+ABSL_FLAG(bool, omit_warn_msg_atprefix, false,
+          "FT.AGGREGATE: Omit warning log message for at prefix.");
 
 namespace dfly {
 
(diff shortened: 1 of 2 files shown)


BagritsevichStepan:
Lets test this PR on the customers datastore (on the dev instance). Confirm here that we are not getting warnings anymore

src/server/search/search_family.cc
```diff
@@ -294,9 +298,10 @@ std::string_view ParseFieldWithAtSign(CmdArgParser* parser) {
   if (absl::StartsWith(field, "@"sv)) {
     field.remove_prefix(1);  // remove leading @
   } else {
-    // Temporary warning until we can throw an error. Log every 30 seconds
-    LOG_EVERY_T(WARNING, 30) << "bad arguments: Field name '" << field
-                             << "' should start with '@'. '@" << field << "' is expected";
+    if (!absl::GetFlag(FLAGS_omit_warn_msg_atprefix)) {
+      LOG_EVERY_T(WARNING, 30) << "bad arguments: Field name '" << field
```
BagritsevichStepan, line 302: @adiholden Please write here what the expected behavior should be if the flag is not set: should we still log a warning, or return a syntax error to the user (i.e., reject the request entirely)?

vyavdoshenko: @romange 
Fixed.

romange (dismissed):

src/server/search/search_family_test.cc
```diff
@@ -989,6 +993,7 @@ TEST_F(SearchFamilyTest, JsonAggregateGroupBy) {
 }
 
 TEST_F(SearchFamilyTest, JsonAggregateGroupByWithoutAtSign) {
+  bool old_value = absl::GetFlag(FLAGS_search_reject_legacy_field);
```
romange, line 996: you can use `absl::FlagSaver fs;` for this (see rdb_test for example)
and then below just use `true` instead of `old_value`

src/server/search/search_family_test.cc
```diff
@@ -1011,6 +1020,8 @@ TEST_F(SearchFamilyTest, AggregateGroupByReduceSort) {
   }
   Run({"ft.create", "i1", "schema", "even", "tag", "sortable", "value", "numeric", "sortable"});
 
+  bool old_value = absl::GetFlag(FLAGS_search_reject_legacy_field);
```
romange, line 1023: same, just use `FlagSaver`

romange: and below

src/server/search/search_family_test.cc

src/server/search/search_family_test.cc

romange (approved):

PR #4814, opened 2025-03-22
chore(cluster_mgr): introduce SlotRange class

romange:
Before: slot merging/splitting logic was mixed with business logic. Also, slots were represented as dictionary, which made the code less readable. Now, SlotRange handles the low-level logic, which makes the high-level code simpler to understand.

### DIFF
--- a/tools/cluster_mgr.py
+++ b/tools/cluster_mgr.py
@@ -3,6 +3,7 @@ import argparse
 from argparse import RawTextHelpFormatter
 import json
 import math
+from typing import Iterable
 import redis
 import subprocess
 import time
@@ -46,6 +47,7 @@ def start_node(node, dragonfly_bin, threads):
             "--dbfilename=",
             f"--logtostderr",
             "--proactor_affinity_mode=off",
+            "--omit_basic_usage",
         ],
         stderr=f,
     )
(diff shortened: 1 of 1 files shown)


tools/cluster_mgr.py
```diff
@@ -71,6 +73,55 @@ def send_command(node, command, print_errors=True):
...
+        assert self.contains(slot_id)
+
+        if self.start < self.end:
+            if slot_id == self.start:
+                return None, SlotRange(self.start + 1, self.end)
+            elif slot_id == self.end:
+                return SlotRange(self.start, self.end - 1), None
+            elif self.start < slot_id < self.end:
+                return SlotRange(self.start, slot_id - 1), SlotRange(slot_id + 1, self.end)
+        return None, None
```
BorysTheDev, line 114: It's better to rename it because we remove the slot and return the left and right range.
Maybe simple remove_slot_id() is enough or at least add a comment

BorysTheDev (approved):

tools/cluster_mgr.py
romange: I will rename if in subsequent PR. My goal is to add the "drop" action to drop slot ranges ( to reproduce flush slots flow).

kostasrim: And once you drop/flush you can measure :) Nice!

PR #4816, opened 2025-03-23
feat(cluster_mgr): add populate command

romange:
We further simplify the code around cluster config Also - add a command that populates all the cluster ranges in the cluster using the "populate" command. `--size` and `--valsize` arguments are also added.

### DIFF
--- a/tools/cluster_mgr.py
+++ b/tools/cluster_mgr.py
@@ -28,10 +28,16 @@ class Node:
         node.id = send_command(node, ["cluster", "myid"])
         print(f"- ID {node.id}")
 
+    def __repr__(self):
+        return f"{self.host}:{self.port}/{self.id}"
 
-class Master:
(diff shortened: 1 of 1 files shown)


tools/cluster_mgr.py
```diff
@@ -187,21 +190,22 @@ def create_locally(args):
...
-        nodes.append(master.node)
+        nodes.append(master)
         for replica in master.replicas:
             nodes.append(replica)
 
     print("Starting nodes...")
     for node in nodes:
         start_node(node, args.dragonfly_bin, args.threads)
     print()
+    time.sleep(0.5)
```
BorysTheDev, line 201: why

BorysTheDev (approved):

BorysTheDev (dismissed):

BorysTheDev (approved):

PR #4818, opened 2025-03-24
feat(dfly_bench): allow generation of keys without hashslots for set/get

romange:
This mode is compatible with memtier_benchmark, as well as "debug populate ... SLOTS ..." We can not use it for custom commands since Dragonfly/Redis require a common hash slot for multiple keys.


mkaruza (approved):
LGTM

PR #4819, opened 2025-03-24
feat(metrics): Update metrics for aliased commands

abhijat:
When a call is made to server, its call count and call time is counted against the string it was called with, if the command was found to be an alias.

Fixes https://github.com/dragonflydb/dragonfly/issues/4068

Follow up PR to https://github.com/dragonflydb/dragonfly/pull/4782

### DIFF
--- a/src/server/command_registry.cc
+++ b/src/server/command_registry.cc
@@ -130,7 +130,8 @@ bool CommandId::IsMultiTransactional() const {
   return CO::IsTransKind(name()) || CO::IsEvalKind(name());
 }
 
-uint64_t CommandId::Invoke(CmdArgList args, const CommandContext& cmd_cntx) const {
+uint64_t CommandId::Invoke(CmdArgList args, const CommandContext& cmd_cntx,
+                           std::optional<std::string_view> orig_cmd_name) const {
   int64_t before = absl::GetCurrentTimeNanos();
   handler_(args, cmd_cntx);
   int64_t after = absl::GetCurrentTimeNanos();
@@ -138,7 +139,8 @@ uint64_t CommandId::Invoke(CmdArgList args, const CommandContext& cmd_cntx) cons
   ServerState* ss = ServerState::tlocal();  // Might have migrated thread, read after invocation
   int64_t execution_time_usec = (after - before) / 1000;
 
-  auto& ent = command_stats_[ss->thread_index()];
+  const std::string_view invoked_name = orig_cmd_name.value_or(name());
+  auto& ent = command_stats_[ss->thread_index()][invoked_name];
 
   ++ent.first;
   ent.second += execution_time_usec;
@@ -263,6 +265,10 @@ std::pair<const CommandId*, ArgSlice> CommandRegistry::FindExtended(string_view 
     }
   }
   return {res, tail_args};
+}
+
+bool CommandRegistry::IsAlias(std::string_view cmd) const {
+  return cmd_aliases_.contains(cmd);
 }
 
 namespace CO {
--- a/src/server/command_registry.cc
+++ b/src/server/command_registry.cc
@@ -131,7 +131,7 @@ bool CommandId::IsMultiTransactional() const {
 }
 
 uint64_t CommandId::Invoke(CmdArgList args, const CommandContext& cmd_cntx,
-                           std::optional<std::string_view> orig_cmd_name) const {
+                           std::string_view orig_cmd_name) const {
   int64_t before = absl::GetCurrentTimeNanos();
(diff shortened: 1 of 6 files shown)


src/server/main_service.cc
```diff
@@ -1234,7 +1234,7 @@ void Service::DispatchCommand(ArgSlice args, SinkReplyBuilder* builder,
 
   dfly_cntx->cid = cid;
 
-  if (!InvokeCmd(cid, args_no_cmd, builder, dfly_cntx)) {
+  if (!InvokeCmd(cid, args_no_cmd, builder, dfly_cntx, cmd)) {
```
abhijat, line 1237: this might break dual word registered commands like `ACL ...`

abhijat: changed so that only aliases are handled specially when counting stats, all other commands use `cmd.name`, by checking if the passed in command was part of alias map

src/server/main_service.cc
```diff
@@ -1234,7 +1234,13 @@ void Service::DispatchCommand(ArgSlice args, SinkReplyBuilder* builder,
 
   dfly_cntx->cid = cid;
 
-  if (!InvokeCmd(cid, args_no_cmd, builder, dfly_cntx)) {
+  // If cmd is an alias, pass it to Invoke so the stats are updated against the alias. By defaults
+  // stats will be updated for cid.name
+  std::optional<std::string_view> orig_cmd_name = std::nullopt;
+  if (registry_.IsAlias(cmd)) {
+    orig_cmd_name = cmd;
+  }
```
abhijat, line 1242: Another alternative to this lookup would be to return more info from `FindExtended`, maybe a struct with some flags like whether the input was alias or not. 

This current approach seems to be simple, although it comes at the expense of an extra hashmap lookup in the command execution path. 

The alias map would ideally be small though so this lookup shouldn't be too bad. I would be interested to know reviewers thoughts on this though.

kostasrim: At the end of the day, swiss tables (flat_hash_map) will use flat storage and this table is relatively small (1-2 elements per command). I don't see how this would have any measurable/observable impact on the hot path.

kostasrim: Also I would move this:

```
  std::optional<std::string_view> orig_cmd_name = std::nullopt;
  if (registry_.IsAlias(cmd)) {
    orig_cmd_name = cmd;
  }
```

inside `InvokeCmd` so you don't have to pass it explicitly

src/server/command_registry.cc
```diff
@@ -130,15 +130,17 @@ bool CommandId::IsMultiTransactional() const {
...
+                           std::optional<std::string_view> orig_cmd_name) const {
   int64_t before = absl::GetCurrentTimeNanos();
   handler_(args, cmd_cntx);
   int64_t after = absl::GetCurrentTimeNanos();
 
   ServerState* ss = ServerState::tlocal();  // Might have migrated thread, read after invocation
   int64_t execution_time_usec = (after - before) / 1000;
 
-  auto& ent = command_stats_[ss->thread_index()];
+  const std::string_view invoked_name = orig_cmd_name.value_or(name());
```
kostasrim, line 142: Why don't you do this on the caller ? Then you can have this function accept a `string_view`, which is the command name that you will collect stats. That way you can get rid of the optional and that `value_or` thingy

kostasrim (commented):
LGTM, small comments

src/server/command_registry.cc
abhijat: changed to do this in `InvokeCmd` and pass in `string_view` (although now the optional has to be handled by caller)

src/server/main_service.cc
abhijat: We lose the `cmd` value which the command was invoked with by the time we get to `InvokeCmd` so I added it here as an argument.

I used `std::optional` as `InvokeCmd` is called in many places so a default keeps changes from being spread out, in some other call sites (eg `MultiCommandSquasher`) there is not even a string to use as no command has been issued.

kostasrim: yep but that would be a bug right? So for example, if I use the original command name on a squashing context, we will miss/increment the wrong stat right?

abhijat: I see, I will try to address this.

kostasrim (approved):
As discussed f2f, Abhi will follow up with a PR that fixes:

1. multi/exec
2. squashing
3. other places where we call InvokeCmd with default arguments

abhijat:
created https://github.com/dragonflydb/dragonfly/issues/4857 for follow up

PR #4825, opened 2025-03-24
fix: change log level for accept error. backlog param has been updated

vyavdoshenko:
fix: change log level for accept error
fix: backlog param (second param for listen socket function) has been updated. twice bigger

### DIFF
--- a/src/server/dfly_main.cc
+++ b/src/server/dfly_main.cc
@@ -80,7 +80,7 @@ ABSL_FLAG(
 ABSL_FLAG(bool, version_check, true,
           "If true, Will monitor for new releases on Dragonfly servers once a day.");
 
-ABSL_FLAG(uint16_t, tcp_backlog, 256, "TCP listen(2) backlog parameter.");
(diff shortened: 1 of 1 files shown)


romange:
What is the reason for increasing backog?

vyavdoshenko:
> What is the reason for increasing backog?

We experienced such type of error on the prod:
_Error calling accept system:22/Invalid argument_

Using tests, I can reproduce only 'generic' type errors like:
_Error calling accept generic:103/Software caused connection abort_ 

IMHO, server-level errors can also happen due to the shortage of an internal queue that can be enlarged by this parameter. 

This is optional. If you mind, I will remove this change.

romange:
I have not seen the original issue but `man accept` clearly states that 
`EINVAL Socket is not listening for connections, or addrlen is invalid (e.g., is negative).`
EINVAL is 22.
so it's not related to backlog. in any case we override this flag in production so no point in changing the default value in the code.

PR #4827, opened 2025-03-25
fix(dfly_bench): Get MOVED error from parsed response

mkaruza:
There could be multiple errors in io buffer which fails to parse correctly. Fixed with 
getting error string form after parsing.

### DIFF
--- a/src/server/dfly_bench.cc
+++ b/src/server/dfly_bench.cc
@@ -741,6 +741,7 @@ void Driver::ParseRESP() {
   uint32_t consumed = 0;
   RedisParser::Result result = RedisParser::OK;
   RespVec parse_args;
+  constexpr string_view kMovedErrorKey = "-MOVED"sv;
 
   do {
     result = parser_.Parse(io_buf_.InputBuffer(), &consumed, &parse_args);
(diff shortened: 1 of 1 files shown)


src/server/dfly_bench.cc
```diff
@@ -741,17 +741,21 @@ void Driver::ParseRESP() {
...
 
   do {
     result = parser_.Parse(io_buf_.InputBuffer(), &consumed, &parse_args);
     if (result == RedisParser::OK && !parse_args.empty()) {
       if (parse_args[0].type == RespExpr::ERROR) {
         string_view error = io::View(io_buf_.InputBuffer());
         VLOG(2) << "Error " << error;
-        if (absl::StartsWith(error, "-MOVED ")) {
-          error = error.substr(7);
+        if (absl::StartsWith(error, kMovedErrorKey)) {
```
romange, line 752: if you are at it, please change line 750 to: `string_view error = parse_args[0].View();` because otherwise this code breaks once dfly_bench is run with --pipeline=10

romange:
I think the fix I suggested is more correct but I do not understand how it is related to the fact it did not stabilize for me.

src/server/dfly_bench.cc
```diff
@@ -741,17 +741,19 @@ void Driver::ParseRESP() {
   uint32_t consumed = 0;
   RedisParser::Result result = RedisParser::OK;
   RespVec parse_args;
+  constexpr string_view kMovedErrorKey = "MOVED"sv;
```
romange, line 744: are you sure this is correct? do we respond with MOVED and not -MOVED?

mkaruza: We are comparing this now to parsed string so prefix '-' will be removed

romange (approved):

Issue #4841, opened 2025-03-27
Return appropriate response for `COMMAND DOCS` for cli to setup completion and hints


abhijat:
Right now we return `OK` in response to `COMMAND DOCS` which results in cli client not being able to show hints to tab completion

PR #4842, opened 2025-03-27
feat(command docs): Enable fallback tab completion in cli

abhijat:
redis-cli uses COMMAND DOCS response to set up tab and hints completion. If we return OK no completion is set up. If we return an error the cli sets up fallback completion.

Here we change the response from OK to error so that the completion can be enabled as fallback.

fixes https://github.com/dragonflydb/dragonfly/issues/4841


romange:
nice! I wondered why it disappeared :)

romange (approved):

abhijat:
failed test `test_bug_in_json_memory_tracking` does not look related to this PR

kostasrim:
> failed test `test_bug_in_json_memory_tracking` does not look related to this PR

it's not, it's a flake and Stefan is working on it

Issue #4832, opened 2025-03-26
Backward compatibility for FT.SEARCH command syntax for v1.21.4


vyavdoshenko:
v1.21.4 can't be updated to a newer one, for example, to the v1.25.5 version, because of updated syntax.

v1.21.4 example:
```
FT.SEARCH idx:positions @draw_end:[1742916180,1742916180] LIMIT 0 10000
```

Updated equivalent:
```
FT.SEARCH idx:positions "@draw_end:[1742916180 1742916180]" LIMIT 0 10000
```

Example of error message:
```
W20250325 14:21:14.663951 91073 main_service.cc:1410]  FT.SEARCH idx:positions @draw_end:[1742896260,1742896260] LIMIT 0 10000 failed with reason: Query syntax error
```

vyavdoshenko:
> does `[1742916180,1742916180] ` syntax work with redis search ?

Yes, it works for Redis.

PR #4844, opened 2025-03-27
fix: FT.SEARCH commas support for ranges

vyavdoshenko:
Syntax like this should be supported:
FT.SEARCH idx:positions @draw_end:[**1742916180,1742916180**] LIMIT 0 10000

Now, Dragonfly supports the following:
FT.SEARCH idx:positions "@draw_end:[1742916180 1742916180]" LIMIT 0 10000
FT.SEARCH idx:positions "@draw_end:[1742916180,1742916180]" LIMIT 0 10000

Original issue: https://github.com/dragonflydb/dragonfly/issues/4832

### DIFF
--- a/src/core/search/parser.y
+++ b/src/core/search/parser.y
@@ -141,13 +141,9 @@ field_cond:
   | LBRACKET numeric_filter_expr RBRACKET               { $$ = std::move($2); }
   | LCURLBR tag_list RCURLBR                            { $$ = std::move($2); }
 
-commas:
-  COMMA
-  | commas COMMA
-
 numeric_filter_expr:
   opt_lparen generic_number opt_lparen generic_number { $$ = AstRangeNode($2, $1, $4, $3); }
-  | opt_lparen generic_number commas opt_lparen generic_number { $$ = AstRangeNode($2, $1, $5, $4); }
+  | opt_lparen generic_number COMMA opt_lparen generic_number { $$ = AstRangeNode($2, $1, $5, $4); }
 
 generic_number:
   DOUBLE { $$ = toDouble($1); }
--- a/src/core/search/parser.y
+++ b/src/core/search/parser.y
@@ -56,6 +56,7 @@ double toDouble(string_view src);
   LCURLBR     "{"
   RCURLBR     "}"
   OR_OP       "|"
+  COMMA       ","
   KNN         "KNN"
   AS          "AS"
   EF_RUNTIME  "EF_RUNTIME"
@@ -140,8 +141,13 @@ field_cond:
   | LBRACKET numeric_filter_expr RBRACKET               { $$ = std::move($2); }
   | LCURLBR tag_list RCURLBR                            { $$ = std::move($2); }
 
+commas:
+  COMMA
+  | commas COMMA
+
 numeric_filter_expr:
-opt_lparen generic_number opt_lparen generic_number { $$ = AstRangeNode($2, $1, $4, $3); }
+  opt_lparen generic_number opt_lparen generic_number { $$ = AstRangeNode($2, $1, $4, $3); }
+  | opt_lparen generic_number commas opt_lparen generic_number { $$ = AstRangeNode($2, $1, $5, $4); }
 
 generic_number:
   DOUBLE { $$ = toDouble($1); }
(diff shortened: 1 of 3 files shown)


BagritsevichStepan (dismissed):

src/core/search/parser.y
```diff
@@ -140,8 +141,13 @@ field_cond:
   | LBRACKET numeric_filter_expr RBRACKET               { $$ = std::move($2); }
   | LCURLBR tag_list RCURLBR                            { $$ = std::move($2); }
 
+commas:
+  COMMA
+  | commas COMMA
```
BorysTheDev, line 146: is it possible to have several commas between args?

vyavdoshenko: Yes, Redis works with several commas. I guess Dragonfly should support this behavior as well.

src/core/search/parser.y
```diff
@@ -140,8 +141,13 @@ field_cond:
...
+
 numeric_filter_expr:
-opt_lparen generic_number opt_lparen generic_number { $$ = AstRangeNode($2, $1, $4, $3); }
+  opt_lparen generic_number opt_lparen generic_number { $$ = AstRangeNode($2, $1, $4, $3); }
```
BorysTheDev, line 149: what is opt_lparen, why do we need it

src/core/search/parser.y
BorysTheDev: As I understand comma means nothing so we can treat it as empty. Could you use the opt_lparen approach? In this case numeric_filter_expr will be one simpler

BorysTheDev (approved):

src/core/search/parser.y
vyavdoshenko: I saved the existing syntax, but regarding the rules, these are optional '(' and ')' symbols.

src/core/search/parser.y
BorysTheDev: > Yes, Redis works with several commas. I guess Dragonfly should support this behavior as well.

I would suggest you don't do this. it looks like a bug in Redis

vyavdoshenko: I guess not because we have to process many commas mixed with spaces.

src/core/search/parser.y
BorysTheDev: I don't see the rparen in this case

src/core/search/parser.y
vyavdoshenko: OK, let's ask one more opinion.
@adiholden, what do you think? Should we support multiple commas as Redis does? Or should we support only one comma?

src/core/search/parser.y
vyavdoshenko: Yes, you are right; this rule is for LPAREN only.
If this rule is wrong, let's create an issue and address it later.

src/core/search/parser.y
romange: Lets support only one comma

vyavdoshenko: Fixed.
Now, the only one comma supported.

BagritsevichStepan:
Is this ready for review?

vyavdoshenko:
> Is this ready for review?

Yes

BagritsevichStepan (approved):
Please next time click on re-request a review, if it is ready again

adiholden (approved):

src/core/search/parser.y
BorysTheDev: Have you checked if this rule is correct?

vyavdoshenko: This rule is for inclusive/exclusive ranges like these:
(123 234]
(123 234)
[123 234]

I am not 100% sure it covers all the cases, but this rule is needed.

BagritsevichStepan: the question is why opt_lparen is repeated two times

vyavdoshenko: That's a good question for a separate issue and PR.

BagritsevichStepan: no need

BagritsevichStepan: It is right implementation
check #2055

BagritsevichStepan: snippet from redis:
```
127.0.0.1:6379> json.set j1 . '{"number":50}'
OK
127.0.0.1:6379> json.set j2 . '{"number":100}'
OK
127.0.0.1:6379> json.set j3 . '{"number":105}'
OK
127.0.0.1:6379> json.set j4 . '{"number":199}'
OK
127.0.0.1:6379> json.set j5 . '{"number":200}'
OK
127.0.0.1:6379> json.set j6 . '{"number":300}'
OK
127.0.0.1:6379> ft.create index on json schema $.number AS number NUMERIC
OK
127.0.0.1:6379> ft.search index "@number:[(100 (199.9]"
1) (integer) 2
2) "j3"
3) 1) "$"
   2) "{\"number\":105}"
4) "j4"
5) 1) "$"
   2) "{\"number\":199}"
```

Issue #4684, opened 2025-03-03
StringSet benchmarks


BorysTheDev:
Add performance and memory consumption benchmarks for StringSet

I think we need to cover a mix of scenarios:
1) number of keys: 100, 10k, 100k
2) key size: 10, 100, 1000
3) operations: add, remove, get

PR #4866, opened 2025-03-31
perf: Added benchmarks for different operations in String Set.

H4R5H1T-007:
Added benchmark for string set data structure similar to dash benchmark as mentioned in issue #4684


src/core/string_set_bench.cc
```diff
@@ -0,0 +1,193 @@
...
+  const std::string_view alpanum = "1234567890abcdefghijklmnopqrstuvwxyz";
+  std::string ret;
+  ret.reserve(len);
+
+  for (size_t i = 0; i < len; ++i) {
+    ret += alpanum[randGenerator() % alpanum.size()];
+  }
+
+  return ret;
+}
```
BorysTheDev, line 57: we have RandStr() method

src/core/string_set_bench.cc
```diff
@@ -0,0 +1,193 @@
...
+base::Histogram perfHist;
+time_t BenchStringSetAdd(uint32_t nKeys, uint32_t keySize) {
+  auto str = random_string(keySize);
+  time_t total_time = 0;
+  while (sset.UpperBoundSize() < nKeys) {
+    random_string_modifier(str);
+    // If string set already contains str, we will not consider it for benchmarks.
+    if (sset.Contains(str)) {
+      continue;
+    }
```
BorysTheDev, line 77: For the add operation, I believe you can ignore already added strings because Contains is get operation under the hood and affects the test.

H4R5H1T-007:
@BorysTheDev @kostasrim I have rewritten the benchmarks using google benchmarks using absl flags to get parameters for now. Also currently I am facing some issue with getting memory usage as obj_malloc_used_ variable in dense set is increasing even after we clear the string set so memory benchmarks are incorrect. I have referenced this issue in discord also in detail in a latest thread.

src/core/string_set_bench.cc
```diff
@@ -0,0 +1,146 @@
...
+    char buf[32];
+    absl::numbers_internal::FastHexToBufferZeroPad16(gen(), buf);
+
+    for (unsigned j = 0; indx < res.size(); indx++, j++) {
+      res[indx] = buf[j];
+    }
+  }
+
+  return res;
+}
```
BorysTheDev, line 55: Could you use RandStr() or dfly::GetRandomHex()

src/core/string_set_bench.cc
```diff
@@ -0,0 +1,146 @@
...
+extern "C" {
+#include "redis/zmalloc.h"
+}
+
+ABSL_FLAG(std::string, operation, "add",
+          "Operation you want to benchmark. Only add, erase and find are allowed.");
+ABSL_FLAG(uint32_t, keySize, 100, "Size of each key.");
+ABSL_FLAG(uint32_t, nKeys, 10000, "Number of keys you want to run benchmark on.");
+ABSL_FLAG(double_t, fraction, 0.5,
+          "Fraction of string set you want to erase/find for benchmarking.");
```
BorysTheDev, line 24: We don't use flags for tests and benchmarks, please use BENCHMARK(BM_Add)->ArgNames({"elements", "size", ...})->Args({32000, 1000, ...});

src/core/CMakeLists.txt
```diff
@@ -32,7 +32,8 @@ cxx_link(dfly_core base absl::flat_hash_map absl::str_format redis_lib TRDP::lua
 add_executable(dash_bench dash_bench.cc)
 cxx_link(dash_bench dfly_core redis_test_lib)
 
-
+add_executable(string_set_bench string_set_bench.cc)
+cxx_link(string_set_bench dfly_core base benchmark)
```
BorysTheDev, line 36: We already have some benchmarks in the dragonfly/src/core/string_set_test.cc 
Some tests are already covered; please move the code there and check which tests need to be updated.

src/core/string_set_bench.cc
```diff
@@ -0,0 +1,146 @@
...
+
+  return res;
+}
+
+size_t memUsed(StringSet& obj) {
+  // std::cout<<obj.ObjMallocUsed()<<" "<<obj.SetMallocUsed()<<std::endl;
+  return obj.ObjMallocUsed() + obj.SetMallocUsed();
+}
+
+void BM_Add(benchmark::State& state) {
```
BorysTheDev, line 62: I think we already have this test, you need only update it

src/core/string_set_bench.cc
```diff
@@ -0,0 +1,146 @@
...
+    state.ResumeTiming();
+    for (auto str = strs.begin(); str != strs.end() && netNKeys > 0; str++, netNKeys--) {
+      ss.Find(*str);
+    }
+  }
+}
+
+BENCHMARK(BM_Get);
+
+int main(int argc, char* argv[]) {
```
BorysTheDev, line 137: I think we don't need "main" if you move tests in the string_set_test.cc

src/core/CMakeLists.txt
BorysTheDev: you can run them ./string_set_test  --bench --benchmark_filter=.*BM_Add

src/core/string_set_bench.cc
H4R5H1T-007: Sure Borys I will use them. When I was working in different file to write benchmarks (i.e. string_set_bench) It was getting difficult to use either one of them because linking them was causing some issues so I copied code into my file. But as you mentioned now if we use string_set_test to write benchmarks it will be better.

src/core/string_set_bench.cc
H4R5H1T-007: Sure Borys. I will rewrite them with args like you suggested. Also should I use ArgPairs to get results of mix an match permutations of all the arguments like this? 
BENCHMARK(BM_Add)
    ->ArgPairs(
        {1000, 5000, 10000},  // Values for first parameter
        {8, 16, 32}           // Values for second parameter
    );

src/core/CMakeLists.txt
H4R5H1T-007: Okay sure. As I was writing this benchmark I also noticed that benchmarks are already present but I thought I have to write them like dash_bench so I created separate file. It will be better if we modify string_set_test

src/core/string_set_bench.cc
```diff
@@ -0,0 +1,146 @@
...
+    strs.push_back(str);
+  }
+  ss.Reserve(elems);
+  while (state.KeepRunning()) {
+    state.PauseTiming();
+    ss.Clear();
+    ss.Reserve(elems);
+    state.ResumeTiming();
+    for (auto& str : strs)
+      ss.Add(str);
```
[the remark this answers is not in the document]
H4R5H1T-007: I wrote it upside down so when iterations are complete, I can measure the memory of string set in the next line but mo issues I can move this up into cycle and reorder memory measurement accordingly

src/core/string_set_bench.cc
```diff
@@ -0,0 +1,146 @@
...
+}
+BENCHMARK(BM_Add);
+
+void BM_Erase(benchmark::State& state) {
+  std::vector<std::string> strs;
+  StringSet ss;
+  auto elems = absl::GetFlag(FLAGS_nKeys);
+  auto keySize = absl::GetFlag(FLAGS_keySize);
+  auto fraction = absl::GetFlag(FLAGS_fraction);
+  uint32_t netNKeys = elems * fraction;
```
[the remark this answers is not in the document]
H4R5H1T-007: Do we not need erase benchmark or do we not need to do fractional erase in benchmark? if we do not need fractional erase benchmark then should we erase all the strings and benchmark or something else do I need to check?

H4R5H1T-007:
@BorysTheDev I have updated the string set tests and added benchmarks there can you please review it?

BorysTheDev:
@romange May I merge it?

romange (approved):
lgtm

PR #4868, opened 2025-03-31
fix: '@text:prefix*' matching for fields.

vyavdoshenko:
fix: Pattern matching error when query includes escaped special characters #4691

### DIFF
--- a/src/core/search/lexer.lex
+++ b/src/core/search/lexer.lex
@@ -28,14 +28,14 @@
 
   Parser::symbol_type make_StringLit(string_view src, const Parser::location_type& loc);
   Parser::symbol_type make_TagVal(string_view src, const Parser::location_type& loc);
+  Parser::symbol_type make_TagPrefix(string_view src, const Parser::location_type& loc);
 %}
 
-blank [ \t\r]
 dq    \"
 sq    \'
 esc_chars ['"\?\\abfnrtv]
 esc_seq \\{esc_chars}
-term_char [_]|\w
+term_char \w
 tag_val_char {term_char}|\\[,.<>{}\[\]\\\"\':;!@#$%^&*()\-+=~\/ ]
 
 
@@ -77,6 +77,7 @@ tag_val_char {term_char}|\\[,.<>{}\[\]\\\"\':;!@#$%^&*()\-+=~\/ ]
 {term_char}+"*" return Parser::make_PREFIX(str(), loc());
 
 {term_char}+ return Parser::make_TERM(str(), loc());
+{tag_val_char}+"*" return make_TagPrefix(str(), loc());
 {tag_val_char}+   return make_TagVal(str(), loc());
 
 <<EOF>>    return Parser::make_YYEOF(loc());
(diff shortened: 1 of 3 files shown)


src/core/search/lexer.lex
```diff
@@ -28,14 +28,14 @@
...
   Parser::symbol_type make_TagVal(string_view src, const Parser::location_type& loc);
+  Parser::symbol_type make_TagPrefix(string_view src, const Parser::location_type& loc);
 %}
 
-blank [ \t\r]
 dq    \"
 sq    \'
 esc_chars ['"\?\\abfnrtv]
 esc_seq \\{esc_chars}
-term_char [_]|\w
```
BagritsevichStepan, line 38: why did you remove this?

src/core/search/lexer.lex
```diff
@@ -28,14 +28,14 @@
 
   Parser::symbol_type make_StringLit(string_view src, const Parser::location_type& loc);
   Parser::symbol_type make_TagVal(string_view src, const Parser::location_type& loc);
+  Parser::symbol_type make_TagPrefix(string_view src, const Parser::location_type& loc);
 %}
 
-blank [ \t\r]
```
BagritsevichStepan, line 33: and this?

src/core/search/search_parser_test.cc
```diff
@@ -191,6 +191,36 @@ TEST_F(SearchParserTest, Scanner) {
   NEXT_EQ(TOK_DOUBLE, string, "33.3");
 }
 
+TEST_F(SearchParserTest, EscapedTagPrefixes) {
```
BagritsevichStepan, line 194: We need also add tests for that to the search_family_test
You can just copy snippet from the user's issue

src/core/search/lexer.lex
```diff
@@ -77,6 +77,7 @@ tag_val_char {term_char}|\\[,.<>{}\[\]\\\"\':;!@#$%^&*()\-+=~\/ ]
 {term_char}+"*" return Parser::make_PREFIX(str(), loc());
```
BagritsevichStepan, line 77: we have now two times + "*"
consider add a variable for *. And may be join this to methods (if possible)

src/core/search/lexer.lex
vyavdoshenko: Because \w has [ _ ] as part of its rule.
It duplicates it. It's useless, at least.

src/core/search/lexer.lex
vyavdoshenko: This rule is useless.
There is no place where it is used.

src/core/search/search_parser_test.cc
vyavdoshenko: Fixed.

src/core/search/lexer.lex
vyavdoshenko: Fixed.

BagritsevichStepan (approved):

PR #4869, opened 2025-03-31
chore: Remove version attribute from compose file

ajnart:
- Removes the warning the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion

This just follows the advice from docker of remove the version attribute from the compose file. No more warnings when following the guide 👍🏼


romange (approved):

romange:
thanks!

PR #4879, opened 2025-04-02
feat: debug compression <type>

romange:
Extend `debug compression` to support value types. if no type is given, falls back to testing keys compressability.


romange:
@BorysTheDev ping

BorysTheDev (approved):

PR #4885, opened 2025-04-03
fix: make FreeMemWithEvictionStep atomic

### LINKED ISSUE #4875: Preempt inside atomic section of heartbeat during evictions
`RetireExpiredAndEvict` should be atomic. However, `FreeMemWithEvictionStep` is `partially atomic`. It's atomic up to the point of calling `RecordExpiryBlocking`. Yes we write to the journal in this function but this is not a concern, as we disable `journal flushing` before we call this function. The problem is on the next function call: `SendQueuedInvalidationMessages` which calls `shard_set->pool()->AwaitBrief(std::move(cb));`. This is a`suspension point` and violates the atomicity grantees mentioned above.

kostasrim:
FreeMemWithEvictionStep was not atomic yet was used by a flow -- RetireExpireAndEvict() -- which should be atomic. Therefore, we make this function atomic (since it's only used by that flow).

Resolves #4875

### DIFF
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -1364,23 +1364,6 @@ pair<uint64_t, size_t> DbSlice::FreeMemWithEvictionStepAtomic(DbIndex db_ind,
   bool record_keys = owner_->journal() != nullptr || expired_keys_events_recording_;
   vector<string> keys_to_journal;
 
-  auto finalize = [&]() {
-    // send the deletion to the replicas.
-    for (string_view key : keys_to_journal) {
-      if (auto journal = owner_->journal(); journal)
-        RecordExpiryBlocking(db_ind, key);
-
-      if (expired_keys_events_recording_)
-        db_table->expired_keys_events_.emplace_back(key);
-    }
-
-    SendQueuedInvalidationMessagesAsync();
-    auto time_finish = absl::GetCurrentTimeNanos();
-    events_.evicted_keys += evicted_items;
-    DVLOG(2) << "Eviction time (us): " << (time_finish - time_start) / 1000;
-    return pair<uint64_t, size_t>{evicted_items, evicted_bytes};
-  };
-
   for (int32_t slot_id = num_slots - 1; slot_id >= 0; --slot_id) {
     for (int32_t bucket_id = num_buckets - 1; bucket_id >= 0; --bucket_id) {
       // pick a random segment to start with in each eviction,
@@ -1389,10 +1372,7 @@ pair<uint64_t, size_t> DbSlice::FreeMemWithEvictionStepAtomic(DbIndex db_ind,
       for (size_t num_seg_visited = 0; num_seg_visited < max_segment_to_consider;
            ++num_seg_visited, segment_id = GetNextSegmentForEviction(segment_id, db_ind)) {
         const auto& bucket = db_table->prime.GetSegment(segment_id)->GetBucket(bucket_id);
-        if (bucket.IsEmpty())
-          continue;
-
-        if (!bucket.IsBusy(slot_id))
+        if (bucket.IsEmpty() || !bucket.IsBusy(slot_id))
           continue;
 
         auto evict_it = db_table->prime.GetIterator(segment_id, bucket_id, slot_id);
@@ -1414,12 +1394,29 @@ pair<uint64_t, size_t> DbSlice::FreeMemWithEvictionStepAtomic(DbIndex db_ind,
 
         // returns when whichever condition is met first
         if ((evicted_items == max_eviction_per_hb) || (evicted_bytes >= increase_goal_bytes))
-          return finalize();
+          goto finish;
       }
     }
   }
 
-  return finalize();
+finish:
+  // send the deletion to the replicas.
+  for (string_view key : keys_to_journal) {
+    if (auto journal = owner_->journal(); journal)
+      // Won't block because we disabled journal flushing. See first line of this function.
+      RecordExpiryBlocking(db_ind, key);
+
+    if (expired_keys_events_recording_)
+      db_table->expired_keys_events_.emplace_back(key);
+  }
+
+  // This might not always be atomic on exceptional cases -- see comments on the function
+  // declaration.
+  SendQueuedInvalidationMessagesAsync();
+  auto time_finish = absl::GetCurrentTimeNanos();
+  events_.evicted_keys += evicted_items;
+  DVLOG(2) << "Eviction time (us): " << (time_finish - time_start) / 1000;
+  return pair<uint64_t, size_t>{evicted_items, evicted_bytes};
 }
 
 void DbSlice::CreateDb(DbIndex db_ind) {
@@ -1561,7 +1558,12 @@ void DbSlice::SendQueuedInvalidationMessages() {
   }
 }
 
+// This function might preempt if the task queue within DispatchBrief is full and we can't
+// enqueue the callback. Although a rare case, this code might not be atomic.
 void DbSlice::SendQueuedInvalidationMessagesAsync() {
+  if (pending_send_map_.empty()) {
+    return;
+  }
   // DispatchBrief will copy local_map
   auto cb = [lm = std::move(pending_send_map_), this](unsigned idx, util::ProactorBase*) {
     SendQueuedInvalidationMessagesCb(lm, idx);
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -1353,8 +1353,12 @@ int32_t DbSlice::GetNextSegmentForEviction(int32_t segment_id, DbIndex db_ind) c
          db_arr_[db_ind]->prime.GetSegmentCount();
 }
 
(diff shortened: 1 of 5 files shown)


src/server/db_slice.cc
```diff
@@ -1550,33 +1553,45 @@ void DbSlice::QueueInvalidationTrackingMessageAtomic(std::string_view key) {
...
-        }
-      }
+      SendQueuedInvalidationMessagesCb(local_map, idx);
     };
 
     shard_set->pool()->AwaitBrief(std::move(cb));
   }
 }
 
+void DbSlice::SendQueuedInvalidationMessagesAsync() {
```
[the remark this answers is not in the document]
kostasrim: After looking at this the `while loop` was `redundant` in the flow of `heartbeat`. Why? Because heartbeat was preempted(because it tried to sent the queued invalidated messages). `pending_send_map_` can only change in two places: 1) during evictions from heartbeat 2) through db_slice and drained at the end via `OnCbFinish()`. The former (1) is guaranteed that it won't evict because it's preempted and until it completes it won't run another heartbeat on that thread while the later (2) drains the `pending_send_map_` via `OnCbFinish()`. Therefore, there is no correctness issue of "missing or not draining all of the items in the pending_send_map_`.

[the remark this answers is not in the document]
kostasrim: As for DispatchBrief, I also don't see any issue whatsoever, they messages will reach the connections eventually and it should be enough

src/server/db_slice.cc
```diff
@@ -1360,64 +1364,62 @@ pair<uint64_t, size_t> DbSlice::FreeMemWithEvictionStep(DbIndex db_ind, size_t s
...
+      // as segment_id does not imply any recency, and random selection should be fair enough
+      int32_t segment_id = starting_segment_id;
+      for (size_t num_seg_visited = 0; num_seg_visited < max_segment_to_consider;
+           ++num_seg_visited, segment_id = GetNextSegmentForEviction(segment_id, db_ind)) {
+        const auto& bucket = db_table->prime.GetSegment(segment_id)->GetBucket(bucket_id);
+        if (bucket.IsEmpty())
+          continue;
+
+        if (!bucket.IsBusy(slot_id))
+          continue;
```
BorysTheDev, line 1396: if (bucket.IsEmpty() || !bucket.IsBusy(slot_id))
continue;

src/server/db_slice.cc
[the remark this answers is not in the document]
romange: DispatchBrief can also block btw, it's just a much rare event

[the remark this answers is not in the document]
kostasrim: @romange the rationale was this doesn't happen in practice because we don't reach this state easily (where we have the task queues internally full). That's a very valid comment, and maybe we should move this outside of the non preemptive critical section. I think it should be a simple change. I will update on this

src/server/db_slice.cc
```diff
@@ -1360,64 +1364,62 @@ pair<uint64_t, size_t> DbSlice::FreeMemWithEvictionStep(DbIndex db_ind, size_t s
...
+      }
+    }
   }
 
-  SendQueuedInvalidationMessages();
-  auto time_finish = absl::GetCurrentTimeNanos();
-  events_.evicted_keys += evicted_items;
-  DVLOG(2) << "Eviction time (us): " << (time_finish - time_start) / 1000;
-  return pair<uint64_t, size_t>{evicted_items, evicted_bytes};
+  return finalize();
```
romange, line 1422: no need to call finalize if `expired_keys_events_recording_` is false and `journal` is null

src/server/db_slice.cc
```diff
@@ -1360,64 +1364,62 @@ pair<uint64_t, size_t> DbSlice::FreeMemWithEvictionStep(DbIndex db_ind, size_t s
...
-            goto finish;
-        }
-      }
-    }
-  }  // FiberAtomicGuard
+  auto finalize = [&]() {
+    // send the deletion to the replicas.
+    for (string_view key : keys_to_journal) {
+      if (auto journal = owner_->journal(); journal)
+        RecordExpiryBlocking(db_ind, key);
```
romange, line 1371: please add a comment saying that even though you call RecordExpiryBlocking it does not block due to JournalFlushGuard above

src/server/db_slice.cc
romange: actually it's better to check in finalize to avoid iterating over keys_to_journal

src/server/db_slice.cc
```diff
@@ -1332,8 +1332,12 @@ int32_t DbSlice::GetNextSegmentForEviction(int32_t segment_id, DbIndex db_ind) c
          db_arr_[db_ind]->prime.GetSegmentCount();
 }
 
-pair<uint64_t, size_t> DbSlice::FreeMemWithEvictionStep(DbIndex db_ind, size_t starting_segment_id,
-                                                        size_t increase_goal_bytes) {
+pair<uint64_t, size_t> DbSlice::FreeMemWithEvictionStepAtomic(DbIndex db_ind,
+                                                              size_t starting_segment_id,
+                                                              size_t increase_goal_bytes) {
+  // Disable flush journal changes to prevent preemtion
+  journal::JournalFlushGuard journal_flush_guard(shard_owner()->journal());
```
adiholden, line 1339: The caller to this function also uses journal::JournalFlushGuard
In this class distructor we do journal_->SetFlushMode(true);
This means that the logic now is broken if there are other calls to journal not inside FreeMemWithEvictionStepAtomic
1. I would add a dcheck in JournalFlushGuard to see we do not have nested calls to it.
2. why did you add it here if its also in the caller?

src/server/db_slice.cc
kostasrim: `bool record_keys = owner_->journal() != nullptr || expired_keys_events_recording_;` 

and then: 
```
  1388         if (record_keys)                                                                          
  1389           keys_to_journal.emplace_back(key); 
```

So `keys_to_journal` will be empty if either we have key space notifications or if we need to write to the journal. 

So no action really needed there because `keys_to_journal` will be empty.

What we can save is calling `SendQueuedInvalidationMessagesAsync` if `pending_send_map.empty()`

src/server/db_slice.cc
kostasrim: I find funny that Adi asked me why I used 

```
  // Disable flush journal changes to prevent preemtion
  journal::JournalFlushGuard journal_flush_guard(shard_owner()->journal());
```

when the `caller` of this function already does the same. Then you are asking me to add a comment here.

yet the only reason I added the line above on the `first line of this function` is precisely to `show intent` (that we disable the flushing in journal explicitly). 

I will add a comment :smile:

src/server/db_slice.cc
```diff
@@ -1360,60 +1364,55 @@ pair<uint64_t, size_t> DbSlice::FreeMemWithEvictionStep(DbIndex db_ind, size_t s
...
+  for (int32_t slot_id = num_slots - 1; slot_id >= 0; --slot_id) {
+    for (int32_t bucket_id = num_buckets - 1; bucket_id >= 0; --bucket_id) {
+      // pick a random segment to start with in each eviction,
+      // as segment_id does not imply any recency, and random selection should be fair enough
+      int32_t segment_id = starting_segment_id;
+      for (size_t num_seg_visited = 0; num_seg_visited < max_segment_to_consider;
+           ++num_seg_visited, segment_id = GetNextSegmentForEviction(segment_id, db_ind)) {
+        const auto& bucket = db_table->prime.GetSegment(segment_id)->GetBucket(bucket_id);
+        if (bucket.IsEmpty() || !bucket.IsBusy(slot_id))
+          continue;
```
kostasrim, line 1376: Only difference is I combined those two statements

src/server/db_slice.cc
```diff
@@ -1360,60 +1364,55 @@ pair<uint64_t, size_t> DbSlice::FreeMemWithEvictionStep(DbIndex db_ind, size_t s
...
-
-          evicted_bytes += evict_it->first.MallocUsed() + evict_it->second.MallocUsed();
-          ++evicted_items;
-          PerformDeletion(Iterator(evict_it, StringOrView::FromView(key)), db_table.get());
-
-          // returns when whichever condition is met first
-          if ((evicted_items == max_eviction_per_hb) || (evicted_bytes >= increase_goal_bytes))
-            goto finish;
-        }
+  for (int32_t slot_id = num_slots - 1; slot_id >= 0; --slot_id) {
```
kostasrim, line 1367: No functional changes in this function. I moved the `FiberAtomicGuard` above

kostasrim: @adiholden I can also move back the `Guard` locally so the diff is cleaner

src/server/db_slice.cc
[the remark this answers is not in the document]
kostasrim: @romange would it worth to add a function that won't dispatch if the task queue is full (but also won't preempt and return to the caller?) 

That way, we can send the pending_items later if we can't dispatch because the intenral queues are full

adiholden (approved):

Issue #4831, opened 2025-03-26
docker release does not update its latest version


romange:
See https://github.com/dragonflydb/dragonfly/actions/runs/13905270815/job/38906828205 run for 1.28.0

but `docker run --rm ghcr.io/dragonflydb/dragonfly:latest --version` shows `dragonfly v1.27.3-68703090337a9fa12e793443d98fb978f3ea233c`

PR #4887, opened 2025-04-03
fix: docker release should be updated latest version using semantic version of latest tag

vyavdoshenko:
fixed: [#4831](https://github.com/dragonflydb/dragonfly/issues/4831)


vyavdoshenko:
The change was tested with a test trigger on the push for the current branch:
https://github.com/dragonflydb/dragonfly/actions/runs/14247603833/job/39932415327
NOTE: The code from the current change was extracted for separate testing.

This solution was partially tested locally:
https://github.com/nektos/act

.github/workflows/docker-release2.yml
```diff
@@ -157,10 +157,62 @@ jobs:
...
+
+                # Remove 'v' prefix if present for semantic comparison
+                current_version=${tag_name#v}
+
+                # Get the current latest version by running the latest image
+                latest_version=""
+                if docker pull ${registry}:latest &>/dev/null; then
+                  echo "Found latest tag, checking its version..."
+                  # Run the container to get its version
+                  latest_full_version=$(docker run --rm --entrypoint /bin/sh ${registry}:latest -c "dragonfly --version | cut -d' ' -f2 | head -n 1")
```
romange, line 171: this is unfortunate that the version is not stored in the image itself.
when running `docker inspect ubuntu:latest` 
I see 
```
...
            "Labels": {
                "org.opencontainers.image.ref.name": "ubuntu",
                "org.opencontainers.image.version": "24.04"
            }
```

meaning we could store the version using labels and then fetch the label instead of running the binary.
this can not be done now but we can specify the right label to simplify this code in the future.
this is what we specify now:
```
  "org.opencontainers.image.created": "2025-03-30T07:56:05.960Z",
                "org.opencontainers.image.description": "The fastest in-memory store",
                "org.opencontainers.image.licenses": "NOASSERTION",
                "org.opencontainers.image.ref.name": "ubuntu",
                "org.opencontainers.image.revision": "a87fe66d1a7654b8f2ee2b7fe27687d582230888",
                "org.opencontainers.image.source": "https://github.com/dragonflydb/dragonfly",
                "org.opencontainers.image.title": "Dragonfly Production Image",
                "org.opencontainers.image.url": "https://github.com/dragonflydb/dragonfly",
                "org.opencontainers.image.vendor": "DragonflyDB LTD",
                "org.opencontainers.image.version": "ubuntu-1.28.1-amd64"
```

I suggest changing `org.opencontainers.image.version` to be version only, i.e. `1.28.1`

.github/workflows/docker-release2.yml
```diff
@@ -157,10 +157,62 @@ jobs:
...
+                  latest_version=$(echo "${latest_full_version}" | cut -d'-' -f1)
+                  # Remove 'v' prefix if present
+                  latest_version=${latest_version#v}
+                  echo "Current latest version: ${latest_version}"
+                else
+                  echo "No latest tag found yet or couldn't pull it"
+                fi
+
+                # Compare versions only if we have a latest version
+                should_update_latest=true
```
romange, line 184: lets factor out this to function `semver_cmp` for readability

romange (commented):
lgtm

.github/workflows/docker-release2.yml
vyavdoshenko: Fixed.

.github/workflows/docker-release2.yml
vyavdoshenko: I added processing with docker image inspect for:
- ubuntu-1.28.1-amd64
- 1.28.1 (as fallback)

The second chance will be: 
- dragonfly --version

romange: that's great but you have not added set the `org.opencontainers.image.version` label.

romange (approved):

.github/workflows/docker-release2.yml
vyavdoshenko: My bad, I didn't get it. I investigated that the images have the org.opencontainers.image.version somehow.
I thought this was a different process to set up this property.
I added the tag version to the org.opencontainers.image.version property.

romange (approved):

PR #4891, opened 2025-04-04
fix: test_migration_rebalance_node

BorysTheDev:
Problem: during seeding, we can add ttl up to 50ms, so to get the same data on fake_redis and DF we need to sleep a little before capture the resulted data

fixes: #4861


vyavdoshenko (approved):
lgtm

Issue #4884, opened 2025-04-03
invalid master_last_io_seconds_ago metric during stable sync


romange:
1. start master/replica replication locally
2. run `memtier_benchmark --ratio 1:0 -d 256  --test-time=300  --distinct-client-seed --key-maximum=2000000 -c 10 -t 2` on master (just write traffic)

3. see that `info replication` randomly shows either huge `master_last_io_seconds_ago` or 0.

```
# Replication
role:slave
master_host:localhost
master_port:6379
master_link_status:up
master_last_io_seconds_ago:18446744073
master_sync_in_progress:0
master_replid:22652ca01f75d4d90e55c5b1708b88fdbd635ffe
slave_repl_offset:23512043
slave_priority:100
slave_read_only:1

```

PR #4892, opened 2025-04-04
fix: invalid master_last_io_seconds_ago metric during stable sync

vyavdoshenko:
fixes: https://github.com/dragonflydb/dragonfly/issues/4884

Fixed overflow on calculating time difference.

### DIFF
--- a/src/server/protocol_client.cc
+++ b/src/server/protocol_client.cc
@@ -316,6 +316,7 @@ error_code ProtocolClient::ReadLine(base::IoBuf* io_buf, string_view* line) {
     if (!input_str.empty())
       break;
     RETURN_ON_ERR(Recv(sock_.get(), io_buf));
+    TouchIoTime();
     input_str = ToSV(io_buf->InputBuffer());
   };
 
@@ -333,6 +334,7 @@ error_code ProtocolClient::ReadLine(base::IoBuf* io_buf, string_view* line) {
     }
 
     RETURN_ON_ERR(Recv(sock_.get(), io_buf));
+    TouchIoTime();
     input_str = ToSV(io_buf->InputBuffer());
   }
 
--- a/src/server/replica.cc
+++ b/src/server/replica.cc
@@ -1106,7 +1106,16 @@ auto Replica::GetSummary() const -> Summary {
     res.master_link_established = (state_mask_.load() & R_TCP_CONNECTED);
     res.full_sync_in_progress = (state_mask_.load() & R_SYNCING);
     res.full_sync_done = (state_mask_.load() & R_SYNC_OK);
-    res.master_last_io_sec = (ProactorBase::GetMonotonicTimeNs() - last_io_time) / 1000000000UL;
+
+    uint64_t current_time = ProactorBase::GetMonotonicTimeNs();
+    // Handle the case when last_io_time is greater than current time
+    // to prevent overflow when calculating master_last_io_sec.
+    if (last_io_time > current_time) {
+      res.master_last_io_sec = 0;
+    } else {
+      res.master_last_io_sec = (current_time - last_io_time) / 1000000000UL;
+    }
+
     res.master_id = master_context_.master_repl_id;
     res.reconnect_count = reconnect_count_;
     res.repl_offset_sum = 0;


src/server/replica.cc
```diff
@@ -1106,7 +1106,16 @@ auto Replica::GetSummary() const -> Summary {
     res.master_link_established = (state_mask_.load() & R_TCP_CONNECTED);
     res.full_sync_in_progress = (state_mask_.load() & R_SYNCING);
     res.full_sync_done = (state_mask_.load() & R_SYNC_OK);
-    res.master_last_io_sec = (ProactorBase::GetMonotonicTimeNs() - last_io_time) / 1000000000UL;
+
+    uint64_t current_time = ProactorBase::GetMonotonicTimeNs();
+    // Handle the case when last_io_time is greater than current time
+    // to prevent overflow when calculating master_last_io_sec.
+    if (last_io_time > current_time) {
```
romange, line 1113: that's fine but why `last_io_time` is sometimes greater than `current_time` ?

src/server/protocol_client.cc
```diff
@@ -316,6 +316,7 @@ error_code ProtocolClient::ReadLine(base::IoBuf* io_buf, string_view* line) {
     if (!input_str.empty())
       break;
     RETURN_ON_ERR(Recv(sock_.get(), io_buf));
+    TouchIoTime();
```
romange, line 319: Please also factor out the lines 260-269 into the Recv call.

vyavdoshenko: But there is a call of TouchIoTime() at 268. This looks correct.

romange: yeah, it is correct but there is code duplication -you can replace these lines with Recv and TouchIoTime calls like we do here

src/server/replica.cc
vyavdoshenko: The condition last_io_time > current_time can be true because:
- **last_io_time_** is set in thread A via TouchIoTime() which calls Proactor()->GetMonotonicTimeNs()
- **current_time** is obtained in thread B via ProactorBase::GetMonotonicTimeNs()
Both calls return thread-local values from **tl_info_.monotonic_time**, which is updated independently in each thread.
- When thread A's clock slightly leads thread B's clock, last_io_time_ can be greater than current_time
This is a race condition where thread-local monotonic time values become inconsistent across threads.

We can avoid that by removing the 1099-1101 lines. However, we will be able to get the last IO time only from the current thread.

romange: in other words, I would expect that the access to `last_io_time` will always be from the same thread.

src/server/protocol_client.cc
vyavdoshenko: Fixed

src/server/replica.cc
vyavdoshenko: I tried to remove:
```
    // Note: we access LastIoTime from foreigh thread in unsafe manner. However, specifically here
    // it's unlikely to cause a real bug.
    for (const auto& flow : shard_flows_) {  // Get last io time from all sub flows.
      last_io_time = std::max(last_io_time, flow->LastIoTime());
    }
```

I left only:
```
  auto last_io_time = LastIoTime();
```

This doesn't work because the updated last IO time gets from other threads.
IMO, there are two ways to fix this issue
- the current fix (it's controversial)
- change how to save monotonic_time, make it atomic and leave only one instance for all threads

src/server/protocol_client.cc
```diff
@@ -256,17 +255,14 @@ io::Result<ProtocolClient::ReadRespRes> ProtocolClient::ReadRespReply(base::IoBu
...
-      VLOG(2) << "Read master response of " << *size_res << " bytes";
+      ec = Recv(sock_.get(), buffer);
+      if (ec) {
+        LOG(ERROR) << "Socket error " << ec;
+        return nonstd::make_unexpected(ec);
+      }
 
-      TouchIoTime();
-      buffer->CommitWrite(*size_res);
+      VLOG(2) << "Read master response";
```
BorysTheDev, line 265: vlog(3) pls

src/server/protocol_client.cc
```diff
@@ -52,21 +52,20 @@ using namespace facade;
...
 
-namespace {
-
-error_code Recv(FiberSocketBase* input, base::IoBuf* dest) {
+error_code ProtocolClient::Recv(FiberSocketBase* input, base::IoBuf* dest) {
   auto buf = dest->AppendBuffer();
   io::Result<size_t> exp_size = input->Recv(buf);
-  if (!exp_size)
+  if (!exp_size) {
+    LOG(ERROR) << "Socket error " << exp_size.error();
```
BorysTheDev, line 59: I suggest log(warning). otherwise control plane suppress this error I think

src/server/protocol_client.cc
```diff
@@ -256,17 +255,14 @@ io::Result<ProtocolClient::ReadRespRes> ProtocolClient::ReadRespReply(base::IoBu
...
-      io::Result<size_t> size_res = sock_->Recv(buf);
-      if (!size_res) {
-        LOG(ERROR) << "Socket error " << size_res.error();
-        return nonstd::make_unexpected(size_res.error());
-      }
 
-      VLOG(2) << "Read master response of " << *size_res << " bytes";
+      ec = Recv(sock_.get(), buffer);
+      if (ec) {
+        LOG(ERROR) << "Socket error " << ec;
```
BorysTheDev, line 261: LOG(WARNING)

src/server/replica.cc
```diff
@@ -1106,7 +1106,16 @@ auto Replica::GetSummary() const -> Summary {
     res.master_link_established = (state_mask_.load() & R_TCP_CONNECTED);
     res.full_sync_in_progress = (state_mask_.load() & R_SYNCING);
     res.full_sync_done = (state_mask_.load() & R_SYNC_OK);
-    res.master_last_io_sec = (ProactorBase::GetMonotonicTimeNs() - last_io_time) / 1000000000UL;
+
+    uint64_t current_time = ProactorBase::GetMonotonicTimeNs();
+    // Handle the case when last_io_time is greater than current time
```
romange, line 1111: Lets change the comment to this:
// last_io_time is derived above by reading `last_io_time_`  from all the flows, by accessing them from a foreign thread, see the loop above. As a result some threads may have `last_io_time_` bigger than our current time, so we fix it here.

This way you explain why this happens, not just what you do.

src/server/replica.cc
romange: let's keep the current fix - not worth wasting more time on this.

src/server/protocol_client.cc
vyavdoshenko: Fixed.

src/server/protocol_client.cc
vyavdoshenko: Fixed.

src/server/protocol_client.cc
vyavdoshenko: Fixed.

src/server/replica.cc
vyavdoshenko: Fixed.

src/server/protocol_client.cc
```diff
@@ -256,17 +255,14 @@ io::Result<ProtocolClient::ReadRespRes> ProtocolClient::ReadRespReply(base::IoBu
...
-      io::Result<size_t> size_res = sock_->Recv(buf);
-      if (!size_res) {
-        LOG(ERROR) << "Socket error " << size_res.error();
-        return nonstd::make_unexpected(size_res.error());
-      }
 
-      VLOG(2) << "Read master response of " << *size_res << " bytes";
+      ec = Recv(sock_.get(), buffer);
+      if (ec) {
+        LOG(WARNING) << "Socket error " << ec;
```
romange, line 261: you already do this inside Recv()

src/server/protocol_client.cc
```diff
@@ -256,17 +255,14 @@ io::Result<ProtocolClient::ReadRespRes> ProtocolClient::ReadRespReply(base::IoBu
...
+        return nonstd::make_unexpected(ec);
+      }
 
-      TouchIoTime();
-      buffer->CommitWrite(*size_res);
+      VLOG(3) << "Read master response";
```
romange, line 265: this vlog is not helpful now and can be removed

romange (approved):
lgtm with small nits

src/server/protocol_client.cc
vyavdoshenko: Fixed.

src/server/protocol_client.cc
vyavdoshenko: Fixed.

PR #4907, opened 2025-04-09
feat(server): Exit process if error is report during load snapshot

### LINKED ISSUE #4840: Instance starts with empty state when loading of snapshot from S3 fails
**Describe the bug**
We use S3 (MinIO) to store snapshots. I noticed that when Dragonfly couldn't load the snapshot on startup, due to temporary access denied errors returned by MinIO, it continued startup with an empty state. The temporary access denied problem was likely caused by the fact that a node was lost and MinIO was recovering.

I'm not sure if this is expected behaviour. In our case it lead to data loss – unfortunately a 3rd party system that we host uses redis as a persistent data store.

I'm also not sure if this is the only case, or would a similar problem show up if there was a network error or something similar.

**To Reproduce**
Steps to reproduce the behavior (not yet validated in a clean environment):
1. Set up Dragonfly with snapshotting to S3/MinIO
2. Write some data
3. Let at least one snapshot happen
4. Temporarily revoke access to the bucket for the credentials given to Dragonfly
5. Restart Dragonfly
6. Give back access to the credentials
7. The written data was lost and Dragonfly is not writing new snapshots that do not include it

**Expected behavior**
No data loss.

I see a couple of options:
1. Make Dragonfly fail on startup if there are any connection/access problems. Do not fail if it's able to connect to the bucket, and the bucket is empty
2. Add flag that forces snapshot loading and fail on startup no matter what the reason of not being able to load the snapshot is. Could be a bit safer for mission critical data, but then the bucket outright losing data does not seem likely.

**Screenshots**
I do not have full logs at the moment, would have after repro. But they were something along these lines:
```
snapshot_storage.cc:379] Creating AWS S3 client; region=us-east-1; https=true; endpoint=[redacted]
credentials_provider_chain.cc:28] aws: disabled EC2 metadata
credentials_provider_chain.cc:36] aws: loaded credentials; provider=environment
snapshot_storage.cc:430] Load snapshot: Searching for snapshot in S3 path: s3://[redacted]/
<the access denied log line, I _think_ it was from server_family.cc>
listener_interface.cc:101] sock[7] AcceptServer - listening on port 9999
listener_interface.cc:101] sock[8] AcceptServer - listening on port 6379
```

 - OS: NixOS 24.11
 - Kernel: 6.6.76
 - Containerized?: Kubernetes
 - Dragonfly Version: v1.26.1

**Reproducible Code Snippet**
I could try providing a docker compose setup later.

**Additional context**
Originally started as a discussion: https://github.com/dragonflydb/dragonfly/discussions/4839

mkaruza:
Exit process if error is reported when we try to load snapshot for both cloud storage and local directory.

Fixes #4840

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -927,9 +927,11 @@ void ServerFamily::LoadFromSnapshot() {
     if (std::error_code(load_path_result.error()) == std::errc::no_such_file_or_directory) {
       LOG(WARNING) << "Load snapshot: No snapshot found";
     } else {
-      util::fb2::LockGuard lk{loading_stats_mu_};
+      loading_stats_mu_.lock();
       loading_stats_.failed_restore_count++;
-      LOG(ERROR) << "Failed to load snapshot: " << load_path_result.error().Format();
+      loading_stats_mu_.unlock();
+      LOG(ERROR) << "Failed to load snapshot with error: " << load_path_result.error().Format();
+      exit(1);
     }
   }
 }
--- a/tests/dragonfly/snapshot_test.py
+++ b/tests/dragonfly/snapshot_test.py
@@ -332,6 +332,24 @@ def delete_s3_objects(bucket, prefix):
     "DRAGONFLY_S3_BUCKET" not in os.environ or os.environ["DRAGONFLY_S3_BUCKET"] == "",
     reason="AWS S3 snapshots bucket is not configured",
 )
+async def test_exit_on_s3_snapshot_load_err(df_factory):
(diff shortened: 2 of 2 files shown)


mkaruza:
> @mkaruza please note that we should exit only if load on startup, and not exit if loading from "DFLY LOAD"

We will only exit process if `LoadFromSnapshot` has error which is called from `ServerFamily::Init`. This is not called with `DFLY LOAD`

kostasrim:
> @mkaruza please note that we should exit only if load on startup, and not exit if loading from "DFLY LOAD"

this flow is only part of `ServerFam::Init`

mlazowik:
Thank you so much!

PR #4912, opened 2025-04-09
fix(pytest): Remove invalid argument from pytest

mkaruza:
Remove invalid argument from test_exit_on_s3_snapshot_load_err test


vyavdoshenko (approved):
lgtm

PR #4913, opened 2025-04-10
fix(hset): Fix size account discrepancy when setting TTL

### LINKED ISSUE #4911: MEMORY USAGE is different on every iteration of HSET & HEXPIRE/FIELDEXPIRE
When you set a hash field, and set an expiry on the field, and loop it, the memory usage of the key will keep on deducting 16B of key size, until it is fully gone.

**To Reproduce**
```
HSET myhash field "value"                                    => (integer) 1
MEMORY USAGE myhash                                          => (integer) 32

HEXPIRE myhash 60 FIELDS 1 field                             => 1) (integer) 1
MEMORY USAGE myhash                                          => (integer) 136

HSET myhash field "value"                                    => (integer) 0
HEXPIRE myhash 60 FIELDS 1 field                             => 1) (integer) 1
MEMORY USAGE myhash                                          => (integer) 120

HSET myhash field "value"                                    => (integer) 0
HEXPIRE myhash 60 FIELDS 1 field                             => 1) (integer) 1
MEMORY USAGE myhash                                          => (integer) 104

HSET myhash field "value"                                    => (integer) 0
HEXPIRE myhash 60 FIELDS 1 field                             => 1) (integer) 1
MEMORY USAGE myhash                                          => (integer) 88
```
**Images:**
Image

Image

**Environment:**
 - Containerized:  Docker Compose image using `docker.dragonflydb.io/dragonflydb/dragonfly`
   - dragonfly v1.28.1-a87fe66d1a7654b8f2ee2b7fe27687d582230888
   - build time: 2025-03-27 15:04:21

**Additional context**

- I've tested on Redis, behaviour is a constant memory usage, as expected.
- I've tested Set, behaviour is a constant memory usage, as expected. 
- I've **not** tested on Sets members, as this feature is Experimental and only available on Dragonfly.
- Running without `HSET` first doesn't cause this bug.

abhijat:
When setting TTL, the object is cloned. An extra 4 bytes is requested during allocation. This can result in the object being allocated from a larger page, eg moving from 16 byte page to 32 byte page. The page block size is used to report object allocation size. Currently this change in size is not reflected in the total memory usage of the dense set, so it remains 16 bytes while the object allocated size is now 32 bytes.

If such an object is later replaced, we deduct the size of the object from total memory usage of the set. Here we can run into an overflow, because the size of the object is deducted from the tracked size of the set, and the former is greater than the latter.

To avoid this, if during setting expiry time the new size is different from old size, we update the set size.

fixes #4911

### DIFF
--- a/src/core/dense_set.cc
+++ b/src/core/dense_set.cc
@@ -50,13 +50,28 @@ void DenseSet::IteratorBase::SetExpiryTime(uint32_t ttl_sec) {
   DensePtr* ptr = curr_entry_->IsLink() ? curr_entry_->AsLink() : curr_entry_;
   void* src = ptr->GetObject();
   if (!HasExpiry()) {
+    const size_t old_size = owner_->ObjectAllocSize(ptr->Raw());
     void* new_obj = owner_->ObjectClone(src, false, true);
     ptr->SetObject(new_obj);
+
+    const size_t new_size = owner_->ObjectAllocSize(ptr->Raw());
 
     // Important: we set the ttl bit on the wrapping pointer.
     curr_entry_->SetTtl(true);
     owner_->ObjDelete(src, false);
     src = new_obj;
+
+    // Because setting TTL requires an extra 4 bytes for the key, the allocated size may push the
+    // object into a different mi-malloc page category (e.g. 16 byte page -> 32 byte page). This
+    // results in increased reporting in ObjAllocSize.
+    //
+    // If this size increase is not accounted for, it will cause an overflow in
+    // DenseSet::AddOrReplaceObj due to subtracting larger size from smaller and the type of
+    // obj_malloc_used_ being size_t.
+    if (old_size != new_size) {
+      owner_->DecreaseMallocUsed(old_size);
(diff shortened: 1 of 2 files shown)


src/core/dense_set.cc
```diff
@@ -701,7 +716,10 @@ void* DenseSet::AddOrReplaceObj(void* obj, bool has_ttl) {
       dptr = dptr->AsLink();
 
     void* res = dptr->Raw();
-    obj_malloc_used_ -= ObjectAllocSize(res);
+    const size_t res_sz = ObjectAllocSize(res);
+    DCHECK(obj_malloc_used_ >= res_sz) << "Invalid deduction of object size " << res_sz
+                                       << " from total set size " << obj_malloc_used_;
```
romange, line 721: this prints both values if fails:
```suggestion
    DCHECK_GE(obj_malloc_used_, res_sz);
```

abhijat: changed

romange (approved):

PR #4921, opened 2025-04-11
chore: improve SlotRanges::Merge algorithm

BorysTheDev:
related to: https://github.com/dragonflydb/dragonfly/issues/4541

### DIFF
--- a/src/server/cluster/cluster_defs.cc
+++ b/src/server/cluster/cluster_defs.cc
@@ -24,10 +24,11 @@ SlotRanges::SlotRanges(std::vector<SlotRange> ranges) : ranges_(std::move(ranges
 }
 
 void SlotRanges::Merge(const SlotRanges& sr) {
-  // TODO rewrite it
-  SlotSet slots(*this);
-  slots.Set(sr, true);
-  ranges_ = std::move(slots.ToSlotRanges().ranges_);
+  ranges_.reserve(ranges_.size() + sr.Size());
+  for (const auto& r : sr) {
+    ranges_.push_back(r);
+  }
+  std::sort(ranges_.begin(), ranges_.end());
 }
 
 std::string SlotRanges::ToString() const {


src/server/cluster/cluster_defs.cc
```diff
@@ -24,10 +24,11 @@ SlotRanges::SlotRanges(std::vector<SlotRange> ranges) : ranges_(std::move(ranges
 }
 
 void SlotRanges::Merge(const SlotRanges& sr) {
-  // TODO rewrite it
-  SlotSet slots(*this);
-  slots.Set(sr, true);
-  ranges_ = std::move(slots.ToSlotRanges().ranges_);
+  ranges_.reserve(ranges_.size() + sr.Size());
```
vyavdoshenko, line 27: It seems this code missed some slots before. 
Maybe a test should be added?

vyavdoshenko (approved):
lgtm

src/server/cluster/cluster_defs.cc
kostasrim: I think it did not, it used a different algorithm internally that's :man_shrugging:

BorysTheDev: what do you mean? there were no issues

kostasrim (approved):
Just to confirm you just changed the algorithm right ? No functional change! 

Based on my understanding the complexity is the same as well.

src/server/cluster/cluster_defs.cc
kostasrim: @BorysTheDev yep yep, it's a different algorithm I just said.

kostasrim:
> > Just to confirm you just changed the algorithm right ? No other functional change.
> > Based on my understanding the complexity is the same as well.
> 
> Yes I've just changed the algorithm and the complexity is different. SlotSet do extra allocation and checks every slot. Now we just merge 2 vectors and sort it

Do we care about the performance of this code ? If yes I think we can sort in `O(n+k)`. Yes you read that right (almost linear, see https://en.wikipedia.org/wiki/Counting_sort)

src/server/cluster/cluster_defs.cc
vyavdoshenko: I mean that you moved the parameter to the ranges_
It means it removed all the items in the ranges_
Now you extend ranges_ with items from the parameter.

BorysTheDev:
> Do we care about the performance of this code ? If yes I think we can sort in `O(n+k)`. Yes you read that right (almost linear, see https://en.wikipedia.org/wiki/Counting_sort)

Not really, we can write sort in O(n) and we don't even need count sort; both vectors are already sorted, so we can copy every element in the correct position with one comparison. But It's not important here, because the size of the  the vectors are quite small

src/server/cluster/cluster_defs.cc
kostasrim: Before: we created the SlotSet with a copy of this, set explicitly what was set by sr, and finally sort them via `ToSlotRanges`.  The first step uses `this->ranges_` internally.

Now: We reserve, append to the ranges_ the `sr` and do the sort. 

It's the same

kostasrim:
@BorysTheDev which unit failed again ? Also you only rerun only one job :rofl:

src/server/cluster/cluster_defs.cc
vyavdoshenko: I got it. Thanks.

BorysTheDev:
> @BorysTheDev which unit failed again ? Also you only rerun only one job 🤣

OOM for server family tests. I restart others, thx

PR #4924, opened 2025-04-11
chore: avoid squashing when squashing_current_reply_size crosses limit

kostasrim:
Disable squashing when crb crosses the limit.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -100,7 +100,7 @@ ABSL_FLAG(uint32_t, max_busy_read_usec, 100,
 
 ABSL_FLAG(size_t, squashed_reply_size_limit, 0,
           "Max bytes allowed for squashing_current_reply_size. If this limit is reached, "
-          "connections dispatching via pipelines will block until this value is decremented.");
+          "connections dispatching pipelines won't squash them.");
 
 using namespace util;
 using namespace std;


src/server/multi_command_squasher.cc
```diff
@@ -63,6 +66,10 @@ size_t Size(const facade::CapturingReplyBuilder::Payload& payload) {
 }  // namespace
 
 atomic_uint64_t MultiCommandSquasher::current_reply_size_ = 0;
+thread_local size_t MultiCommandSquasher::throttle_size_limit_ =
+    absl::GetFlag(FLAGS_throttle_squashed);
+
+thread_local util::fb2::EventCount MultiCommandSquasher::ec_;
 
 MultiCommandSquasher::MultiCommandSquasher(absl::Span<StoredCmd> cmds, ConnectionContext* cntx,
```
kostasrim, line 74: This is used not only from async fiber but also directly from the connection. If we preempt, the connection will also "freeze". I guess this is fine, just mentioning it here for completeness. 

There are 3 calls of this and all of them should be ok if we preempt from these flows.

tests/dragonfly/memory_test.py
```diff
@@ -222,3 +223,40 @@ async def test_cache_eviction_with_rss_deny_oom(
...
+
+    async def poll():
+        # At any point we should not cross this limit
+        cl = df.client()
+        await cl.execute_command("multi")
+        await cl.execute_command("hgetall test:0")
+        await cl.execute_command("exec")
+
+    # With the current approach this will overshoot
+    #    await client.execute_command("multi")
```
kostasrim, line 247: I wish we also handled this case as well

src/server/multi_command_squasher.cc
```diff
@@ -15,6 +16,8 @@
 #include "server/transaction.h"
 #include "server/tx_base.h"
 
+ABSL_FLAG(size_t, throttle_squashed, 0, "");
```
kostasrim, line 19: @adiholden I will adjust as we said f2f. Looking for some early feedback based on our discussion

src/server/multi_command_squasher.h
```diff
@@ -94,6 +104,9 @@ class MultiCommandSquasher {
 
   // we increase size in one thread and decrease in another
   static atomic_uint64_t current_reply_size_;
+  static thread_local size_t throttle_size_limit_;
+  // Used to throttle when memory is tight
+  static thread_local util::fb2::EventCount ec_;
```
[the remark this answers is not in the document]
romange: since it's thread local, it's more efficient to use NoOpLock together with CondVarAny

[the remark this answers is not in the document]
kostasrim: @romange nice!

src/server/multi_command_squasher.cc
```diff
@@ -63,6 +66,10 @@ size_t Size(const facade::CapturingReplyBuilder::Payload& payload) {
 }  // namespace
 
 atomic_uint64_t MultiCommandSquasher::current_reply_size_ = 0;
+thread_local size_t MultiCommandSquasher::throttle_size_limit_ =
+    absl::GetFlag(FLAGS_throttle_squashed);
```
adiholden, line 70: As discussed this morning multiply by thread count. The limit should be per thread and the current_reply_size_ is global counter

kostasrim: Yes I know, I even wrote a comment above that I will follow up with this :smile: 

I wanted to know if you have anything else to add :smile:

src/server/multi_command_squasher.h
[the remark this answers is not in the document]
kostasrim: Actually, I added a bug here:

```
 static atomic_uint64_t current_reply_size_;  
```

so `current_reply_size` is not thread local. So what can happen is:

Core 0 -> starts multi/exec
Core 1 -> starts multi/exec but needs to throttle so it goes to sleep waiting on the thread local cond variable
Core 0 -> is done, notifies the thread local 
Core 1 -> the fiber never awakes even though we decremented `current_reply_size`.

Since `current_reply_size` is global then so should `ec_`. 

P.s. not very happy with this extra synchronization but we only pay it when we are under memory pressure

src/server/multi_command_squasher.cc
```diff
@@ -63,6 +66,9 @@ size_t Size(const facade::CapturingReplyBuilder::Payload& payload) {
 }  // namespace
 
 atomic_uint64_t MultiCommandSquasher::current_reply_size_ = 0;
+thread_local size_t MultiCommandSquasher::throttle_size_limit_ =
```
adiholden, line 69: I believe we should multiply throttle_squashed by the number of io threads and not shard number

src/server/multi_command_squasher.h
```diff
@@ -37,6 +38,15 @@ class MultiCommandSquasher {
     return current_reply_size_.load(std::memory_order_relaxed);
   }
 
+  static bool IsMultiCommandSquasherOverLimit() {
```
adiholden, line 41: maybe rename to IsReplySizeOverLimit?

src/server/multi_command_squasher.h
```diff
@@ -94,6 +104,10 @@ class MultiCommandSquasher {
 
   // we increase size in one thread and decrease in another
   static atomic_uint64_t current_reply_size_;
+  // Used to throttle when memory is tight
+  static util::fb2::EventCount ec_;
+
+  static thread_local size_t throttle_size_limit_;
```
adiholden, line 110: maybe reply_size_limit_ ?

src/server/multi_command_squasher.cc
adiholden, line 19: maybe squashed_reply_size_limit
add flag description
also I think we should have a default limit here maybe 128_MB ?

tests/dragonfly/memory_test.py
adiholden: what is the difference? why does it not handles this case?

tests/dragonfly/memory_test.py
```diff
@@ -222,3 +223,41 @@ async def test_cache_eviction_with_rss_deny_oom(
...
+
+    client = df.client()
+    # 0.5gb
+    await client.execute_command("debug populate 1 test 10000 rand type hash elements 50000")
+
+    async def poll():
+        # At any point we should not cross this limit
+        assert df.rss < 1_500_000_000
+        cl = df.client()
+        await cl.execute_command("multi")
```
adiholden, line 243: I see that the flow that you are testing is the multi exec flow which I did not think about. When I suggested this throttling I was thinking about the pipeline flow.
When reviewing now the multi exec flow I am not 100% sure for implying this logic in this flow as we when you do the await in the code to wait for the size to decrease we already scheduled the transaction and I am not sure if this can lead in some cases to a deadlock

src/server/server_state.h
```diff
@@ -270,6 +270,10 @@ class ServerState {  // public struct - to allow initialization.
 
   bool ShouldLogSlowCmd(unsigned latency_usec) const;
 
+  size_t GetTotalShards() const {
```
romange, line 273: not needed - you can use `shard_set->size()` everywhere

src/server/multi_command_squasher.cc
```diff
@@ -215,6 +222,9 @@ bool MultiCommandSquasher::ExecuteSquashed(facade::RedisReplyBuilder* rb) {
   if (order_.empty())
     return true;
 
+  MultiCommandSquasher::ec_.await(
```
romange, line 225: all our current approaches of limiting memory are "per-thread", this is consistent and works nicely with shared-nothing. What is the reason for not using per thread limits? In addition, we already have per thread throttling inside dragonfly_connection code, see IsPipelineBufferOverLimit. Did you consider pigging back on this mechanism instead ?

src/server/multi_command_squasher.cc
```diff
@@ -254,19 +264,23 @@ bool MultiCommandSquasher::ExecuteSquashed(facade::RedisReplyBuilder* rb) {
...
   bool aborted = false;
 
+  size_t size = 0;
   for (auto idx : order_) {
     auto& replies = sharded_[idx].replies;
     CHECK(!replies.empty());
 
     aborted |= opts_.error_abort && CapturingReplyBuilder::TryExtractError(replies.back());
 
-    current_reply_size_.fetch_sub(Size(replies.back()), std::memory_order_relaxed);
```
romange, line 263: I said nothing when `current_reply_size_` was added. it was a mistake. I do not want anyone introduces global states in Dragonfly codebase.

romange: @BorysTheDev  FYI

src/server/multi_command_squasher.cc
romange: we should no do this at all. the limit should be by thread. and in general it's not well defined to initialize thread local by using another thread local that is initialized to nullptr.

src/server/multi_command_squasher.h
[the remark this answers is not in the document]
romange: it should be thread local.

src/server/multi_command_squasher.cc
adiholden: but @romange  we discussed this when current_reply_size_ was added. Because the multi command sqasher is adding replies in different threads you said it makes sense.

src/server/multi_command_squasher.cc
adiholden: as long as current_reply_size_ is global I dont see how we can do this per thread

src/server/multi_command_squasher.cc
romange: Would something like this work?
https://github.com/dragonflydb/dragonfly/compare/RemoveAtomic?expand=1

adiholden: @romange The change in your branch linked above is that you count the reply size after we executed all the squashed commands. So it does uses the thread local approach correctly but what we expose to metrics is not accurate because at the time the capture reply builders grow we do not expose this until we finish with all the squashed. I think that applying such logic will impact the throttling on reply size just in delay, we will throttle but not when actually we are at the threshold but with some delay. I guess we can do this change.

romange: There is no a "correct" solution because I can show you a scenario where single central atomic won't work either:
we throttle before we send commands to shards, but maybe there are tons of squashed commands in flight that have not filled their replies yet, so you let the next command pass and only then the reply buffer increases.

I would rather have a less accurate metric than have have all our threads contend on atomics and now on a single condvar. this kills performance. I won't be surprised that even now squashing performance is worse because of the "reply bufffer size" atomic being hammered by multiple threads.

src/server/multi_command_squasher.cc
```diff
@@ -216,6 +222,15 @@ bool MultiCommandSquasher::ExecuteSquashed(facade::RedisReplyBuilder* rb) {
   if (order_.empty())
     return true;
 
+  // Multi non atomic does not lock ahead. So it's safe to preempt while we haven't
+  // really started the transaction.
+  // This is not true for `multi/exec` which uses `Execute()` but locks ahead before it
+  // calls `ScheduleSingleHop` below.
+  // TODO Investigate what are the side effects for allowing it `lock ahead` mode.
+  if (opts_.is_mult_non_atomic) {
```
adiholden: I dont remember if we discussed this -  Why do you need to throttle here and not use the same mecanism we have today for pipeline backpressure to throttle also if we have squashing_current_reply_size above limit

kostasrim: That's a good comment. We discussed this back then `reply_size` was defined in `multi_command_squasher` which is part of `/server` and not `facade` so the connection had no access to it. 

After Roman's changes however, he moved this to `facade_types` so now `facade` has access to it. I moved everything to `dragonfly_connection`. However, I chose not to throttle on the queue back pressure (within `DispatchSingle`) and do it a little later within `AsyncFiber` because:

1. throttling on reply size should only be relevant when we try to squash the pipeline. I don't want to introduce delays for commands that are executing standalone as I don't believe they will increase RSS that much for the workloads we try to "fix"
2. throttling before we dispatch async to the queue is not a great idea because the async fiber might sleep between iterations and inbetween the dispatch queue will aggregate a bunch of commands. Once the async fiber wakes up, it will try to squash the pipeline -- bypassing the protective mechanism we just added (since we the state might be that the reply size is already over the limit yet we don't know it because we made that decision on the dispatch and not on the async fiber level)

src/facade/dragonfly_connection.cc
```diff
@@ -2075,6 +2098,16 @@ void Connection::DecrNumConns() {
     --stats_->num_conns_other;
 }
 
+bool Connection::IsReplySizeOverLimit() const {
+  std::atomic<size_t>& reply_sz = tl_facade_stats->reply_stats.squashing_current_reply_size;
```
adiholden, line 2070: .load(memory_order_relaxed)

adiholden: Oh, why did you split this into 2 lines?

src/facade/dragonfly_connection.cc
```diff
@@ -2075,6 +2098,16 @@ void Connection::DecrNumConns() {
...
+  size_t current = reply_sz.load(std::memory_order_acquire);
+  const bool over_limit = reply_size_limit != 0 && current > 0 && current > reply_size_limit;
+  LOG(INFO) << "current: " << current << "/" << reply_size_limit;
```
adiholden: Log info on this will flood the log file

src/facade/dragonfly_connection.cc
```diff
@@ -2075,6 +2098,16 @@ void Connection::DecrNumConns() {
...
+  VLOG_IF(2, over_limit) << "MultiCommandSquasher overlimit: " << current << "/"
```
adiholden: Actually if we are over limit we want to see this in logs, make this a warning and print once a second

src/facade/dragonfly_connection.cc
```diff
@@ -2105,7 +2138,7 @@ void Connection::EnsureMemoryBudget(unsigned tid) {
 
 Connection::WeakRef::WeakRef(std::shared_ptr<Connection> ptr, unsigned thread_id,
                              uint32_t client_id)
-    : ptr_{ptr}, thread_id_{thread_id}, client_id_{client_id} {
+    : ptr_{std::move(ptr)}, thread_id_{thread_id}, client_id_{client_id} {
```
adiholden, line 2109: ?

src/facade/facade_types.h
```diff
@@ -14,6 +14,7 @@
 
 #include "base/iterator.h"
 #include "facade/op_status.h"
+#include "util/fibers/synchronization.h"
```
adiholden: needed?

src/facade/dragonfly_connection.cc
```diff
@@ -1518,6 +1526,20 @@ void Connection::SquashPipeline() {
 
   cc_->async_dispatch = true;
 
+  // We need to first set async_dispatch before we preempt. Otherwise, when the AsyncFiber
+  // wakes up, sync_dispatch might be true, violating the precondition of this flow
+  // (when we block earlier in the body of AsyncFiber at cnd_.wait(noop_lk, [this])...)
+  fb2::NoOpLock noop;
```
adiholden: still I dont see you use QueueBackpressure class today. Why not use the same class for throttle on squashed pipeline reply size?

src/facade/dragonfly_connection.cc
kostasrim: deep copying a shared ptr increments the atomic counter so we pay for an atomic operation. move avoids that by copying the control pointer instead. I just saw the misuse and used move 😄

src/facade/facade_types.h
kostasrim: nope

src/facade/dragonfly_connection.cc
kostasrim: It was accidental, I changed that when I wanted to debug something. will fix

src/facade/dragonfly_connection.cc
kostasrim: > .load(memory_order_relaxed) 

We should synchronize acquire and release semantics, otherwise the load might be an older value in the modification order of the atomic variable.

I splited it in two lines because otherwise the expression was too big

src/facade/dragonfly_connection.cc
kostasrim: Yes I did that on purpose: 

> throttling before we dispatch async to the queue is not a great idea because the async fiber might sleep between iterations and inbetween the dispatch queue will aggregate a bunch of commands. Once the async fiber wakes up, it will try to squash the pipeline -- bypassing the protective mechanism we just added (since we the state might be that the reply size is already over the limit yet we don't know it because we made that decision on the dispatch and not on the async fiber level)


see https://github.com/dragonflydb/dragonfly/pull/4924#discussion_r2157021011

romange: Can you please describe scenario of how `squashing_current_reply_size` can be greater than zero at this point of time?

kostasrim: yes, `squashing_current_reply_size` is an atomic thread local variable that is shared and can be incremented on different threads. Different connections from thread 1 can schedule a series of squashed pipelines on thread 2 (and will increment the same `squashing_current_reply_size`). When an async fiber on thread 1 wakes up after its squashed pipeline finished executing (and decremented the `squashing_current_reply_size), it will try to squash another one but in parallel the same variable got incremented on thread 2 (because of another of another tx) and potentially crossed the hard limit.  

However, your question raised an interesting point. Is it reasonable to expect that `squashing_current_reply_size` is usually close to `0` even when multiple connections dispatch squashed pipelines ? I would say maybe, because the rate we increment and decrement this variable on separate threads should be close. By the time we decremented it on thread 1 another tx incremented it on thread 2 and the same cycle repeats in such a way that we always have a value for this variable that is close to 0.

romange: ok, you are right, squashing_current_reply_size can be higher than a threshold due to other connections on this thread. In that case I suggest that instead of throttling the connection - we just avoid calling SquashPipeline, and choose the usual route of sending a single command (lines 1644-1647)

kostasrim: Yeah and that would remove the throttling/extra synchronization all together +1

src/facade/dragonfly_connection.cc
```diff
@@ -98,6 +98,10 @@ ABSL_FLAG(uint32_t, max_busy_read_usec, 100,
           "Maximum time we read and parse from "
           "a socket without yielding. In microseconds.");
 
+ABSL_FLAG(size_t, squashed_reply_size_limit, 0,
```
adiholden, line 101: So as I understand now we will not block the connection but just not doing the sqaushing optimizaiton, right?
lets update the comment and the PR title/description

adiholden: apart from this comment LGTM

adiholden (approved):

PR #4925, opened 2025-04-12
fix(transaction): Fix schedule queues

dranikpg:
If it reproduces locally on x86, it must be something simple. My only explanation is that we spuriously reorder the next read in  `sq.queue.Pop()` before we update armed, so:
  * T0: reads from an empty queue
  * T1: stores the next context in the queue
  * T1: makes sure armed is still true
  * T0: clears armed flag and proceeds with stale data from (0)

### DIFF
--- a/src/server/list_family_test.cc
+++ b/src/server/list_family_test.cc
@@ -1372,7 +1372,8 @@ TEST_F(ListFamilyTest, PressureBLMove) {
     }
   };
 
-  for (size_t delay : {1, 15, 35, 50}) {
+  for (size_t delay : {1, 2, 5}) {
+    LOG(INFO) << "Running with delay: " << delay;
     auto f1 = pp_->at(1)->LaunchFiber([=] { consumer("c1", "src", "dest"); });
     auto f2 = pp_->at(1)->LaunchFiber([=] { producer("p1", delay, "src"); });
 


romange:
I agree that their added value is questionable.

dranikpg:
It's failing locally on x86 on the first try, but I can't get it to fail here 😠

src/server/list_family_test.cc
```diff
@@ -1352,6 +1355,32 @@ TEST_F(ListFamilyTest, AwakeMulti) {
...
+  };
+  auto producer = [this](string_view id, size_t delay, string_view src) {
+    for (unsigned i = 0; i < 1000; ++i) {
+      Run(id, {"lpush", src, "a"});
+      ThisFiber::SleepFor(1us * delay);
+    }
+  };
+
+  for (size_t delay : {1, 15, 35, 50}) {
+    auto f1 = pp_->at(1)->LaunchFiber([=] { consumer("c1", "src", "dest"); });
```
romange:
```suggestion
  for (size_t delay : {1, 2, 5}) {
    LOG(INFO) << "Running with delay: " << delay;
    auto f1 = pp_->at(1)->LaunchFiber([=] { consumer("c1", "src", "dest"); });
```

romange: on my machine large delays did not contribute to uncover the deadlock

Issue #4937, opened 2025-04-14
Not null requests in FT.SEARCH


BagritsevichStepan:
For now we support nullable values in `FT.SEARCH`. We also can add a query that can search for fields, for which this field is not nullable (for json) or is present (for hset). This should be `@field:*`. Since redis does not support those type of requests we will not break other logic

PR #4941, opened 2025-04-15
feat: wildcard pattern added for fields to find all non-null values

vyavdoshenko:
fixes: https://github.com/dragonflydb/dragonfly/issues/4937

### DIFF
--- a/src/core/search/ast_expr.h
+++ b/src/core/search/ast_expr.h
@@ -22,7 +22,7 @@ struct AstNode;
 // Matches all documents
 struct AstStarNode {};
 
-// Matches all fields in a document (like "*" in field context)
+// Matches all documents where this field has a non-null value
 struct AstStarFieldNode {};
 
 // Matches terms in text fields
--- a/src/core/search/ast_expr.h
+++ b/src/core/search/ast_expr.h
@@ -22,6 +22,9 @@ struct AstNode;
 // Matches all documents
 struct AstStarNode {};
 
+// Matches all fields in a document (like "*" in field context)
+struct AstStarFieldNode {};
+
 // Matches terms in text fields
 struct AstTermNode {
   explicit AstTermNode(std::string term);
@@ -114,9 +117,9 @@ struct AstSortNode {
   bool descending = false;
 };
 
-using NodeVariants =
-    std::variant<std::monostate, AstStarNode, AstTermNode, AstPrefixNode, AstRangeNode,
-                 AstNegateNode, AstLogicalNode, AstFieldNode, AstTagsNode, AstKnnNode, AstSortNode>;
+using NodeVariants = std::variant<std::monostate, AstStarNode, AstStarFieldNode, AstTermNode,
+                                  AstPrefixNode, AstRangeNode, AstNegateNode, AstLogicalNode,
+                                  AstFieldNode, AstTagsNode, AstKnnNode, AstSortNode>;
 
 struct AstNode : public NodeVariants {
   using variant::variant;
--- a/src/core/search/ast_expr.h
+++ b/src/core/search/ast_expr.h
@@ -111,9 +111,9 @@ struct AstKnnNode {
   std::optional<float> ef_runtime;
 };
 
-using NodeVariants =
-    std::variant<std::monostate, AstStarNode, AstStarFieldNode, AstTermNode, AstPrefixNode,
-                 AstRangeNode, AstNegateNode, AstLogicalNode, AstFieldNode, AstTagsNode, AstKnnNode>;
+using NodeVariants = std::variant<std::monostate, AstStarNode, AstStarFieldNode, AstTermNode,
+                                  AstPrefixNode, AstRangeNode, AstNegateNode, AstLogicalNode,
+                                  AstFieldNode, AstTagsNode, AstKnnNode>;
 
 struct AstNode : public NodeVariants {
   using variant::variant;
--- a/src/core/search/base.h
+++ b/src/core/search/base.h
@@ -95,6 +95,11 @@ struct BaseIndex {
   // Returns true if the document was added / indexed
   virtual bool Add(DocId id, const DocumentAccessor& doc, std::string_view field) = 0;
   virtual void Remove(DocId id, const DocumentAccessor& doc, std::string_view field) = 0;
+
+  // Returns all documents matching the star query (all documents in the index)
+  virtual std::optional<std::vector<DocId>> GetAllResults() const {
+    return std::nullopt;
+  }
 };
 
 // Base class for type-specific sorting indices.
--- a/src/core/search/base.h
+++ b/src/core/search/base.h
@@ -91,7 +91,7 @@ struct BaseIndex {
   virtual bool Add(DocId id, const DocumentAccessor& doc, std::string_view field) = 0;
   virtual void Remove(DocId id, const DocumentAccessor& doc, std::string_view field) = 0;
 
-  // Returns all documents matching the star query (all documents in the index)
+  // Returns documents that have non-null values for this field (used for @field:* queries)
   virtual std::optional<std::vector<DocId>> GetAllResults() const {
     return std::nullopt;
   }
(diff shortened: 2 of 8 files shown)


BagritsevichStepan:
Please do merge with master

src/core/search/base.h
```diff
@@ -90,6 +90,11 @@ struct BaseIndex {
   // Returns true if the document was added / indexed
   virtual bool Add(DocId id, const DocumentAccessor& doc, std::string_view field) = 0;
   virtual void Remove(DocId id, const DocumentAccessor& doc, std::string_view field) = 0;
+
+  // Returns all documents matching the star query (all documents in the index)
+  virtual std::optional<std::vector<DocId>> GetAllResults() const {
```
BagritsevichStepan, line 95: Nit: we are not returning AllResults, we are returning NotNullable results, take in account that json can store nullable values like this `{"text":nulll}` and we will index this. So you need change a bit comment and name

src/core/search/ast_expr.h
```diff
@@ -22,6 +22,9 @@ struct AstNode;
 // Matches all documents
 struct AstStarNode {};
 
+// Matches all fields in a document (like "*" in field context)
```
BagritsevichStepan, line 25: Nit: It doesn't match all fields — it matches all documents where this field has a non-null value

src/core/search/indices.h
```diff
@@ -58,6 +62,21 @@ template <typename C> struct BaseStringIndex : public BaseIndex {
   // Returns all the terms that appear as keys in the reverse index.
   std::vector<std::string> GetTerms() const;
 
+  std::optional<std::vector<DocId>> GetAllResults() const override {
+    std::vector<DocId> result;
+    std::vector<DocId> temp;
```
BagritsevichStepan, line 67: Why just not to use `flat_hash_set` for DocId? And at the end map it to the vector.

src/core/search/indices.h
```diff
@@ -58,6 +62,21 @@ template <typename C> struct BaseStringIndex : public BaseIndex {
...
+    for (const auto& term : GetTerms()) {
+      if (auto* docs = Matching(term)) {
```
BagritsevichStepan, line 69: Why we need to do this? Take in account that GetTerms returns vector of std::string, so it makes copy for each term. Also Matching has a constant.

We have an iterator for RaxTreeMap, so you can just traverse entries_, and merge all containers.

src/core/search/indices.h
```diff
@@ -133,6 +152,32 @@ struct FlatVectorIndex : public BaseVectorIndex {
...
   const float* Get(DocId doc) const;
 
+  // Return all documents that have vectors in this index
+  std::optional<std::vector<DocId>> GetAllResults() const override {
+    std::vector<DocId> result;
+    size_t num_vectors = entries_.size() / dim_;
+    result.reserve(num_vectors);
+
+    for (DocId id = 0; id < num_vectors; ++id) {
+      // Check if the vector is not zero (all elements are 0)
```
BagritsevichStepan, line 162: Not the best decision here — a valid vector can contain zero values. However, since we don't need to fix this for 1.29, let's merge it as is and address it later in a separate PR.

src/server/search/search_family_test.cc
```diff
@@ -2343,4 +2343,304 @@ TEST_F(SearchFamilyTest, KnnSearchWithSortby) {
   EXPECT_THAT(resp, IsArray(2, "doc:3", "doc:1"));
 }
 
+TEST_F(SearchFamilyTest, SearchNonNullFields) {
```
BagritsevichStepan, line 2346: Good work with writing a lot of tests. But let's add a test for json indexes where json has null values for TEXT, TAG, NUMERIC and for TEXT SORTABLE, TAG SORTABLE, NUMERIC SORTABLE and assert that for `@field:*` we will not return those documents

src/server/search/search_family_test.cc
```diff
@@ -2343,4 +2343,304 @@ TEST_F(SearchFamilyTest, KnnSearchWithSortby) {
...
+  string vector2 = "\\x00\\x00\\x00\\x00\\x00\\x00\\x80\\x3f\\x00\\x00\\x00\\x00";  // [0,1,0]
+  string vector3 = "\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x80\\x3f";  // [0,0,1]
+
+  Run({"hset", "vec:1", "embedding", vector1});
+  Run({"hset", "vec:2", "embedding", vector2});
+  Run({"hset", "vec:3", "embedding", vector3});
+
+  // Testing star query with result limit
+  auto limit_result = Run({"ft.search", "text_idx", "*", "LIMIT", "0", "2"});
+  ASSERT_GE(limit_result.GetVec().size(), 5);                 // Total count + 2 docs with fields
```
BagritsevichStepan, line 2435: I didn't get why do we need this all sorting stuff for this functionality but ok. Let's keep it

src/server/search/search_family_test.cc
```diff
@@ -2343,4 +2343,304 @@ TEST_F(SearchFamilyTest, KnnSearchWithSortby) {
...
+  EXPECT_EQ(limit_result.GetVec()[0].GetInt(), 3);            // Total count is 3 (all matches)
+  EXPECT_EQ(limit_result.GetVec()[1].GetString(), "text:1");  // First doc
+  EXPECT_EQ(limit_result.GetVec()[3].GetString(), "text:2");  // Second doc
+
+  // Testing star query with sorting
+  auto price_desc_result = Run({"ft.search", "num_idx", "*", "SORTBY", "price", "DESC"});
+  ASSERT_EQ(price_desc_result.GetVec().size(), 7);
+  EXPECT_EQ(price_desc_result.GetVec()[1].GetString(), "num:3");  // Most expensive item first
```
BagritsevichStepan, line 2443: Nit: If we have repeated parts in our asserts or expects, we usually implement a MATCHER. But let's not do that in this PR — we can refactor it later

src/core/search/search.cc
```diff
@@ -304,6 +305,44 @@ struct BasicSearch {
...
+        if (auto* docs = text_index->Matching(term); docs) {
+          sub_results.push_back(IndexResult{docs});
+        }
+      }
+
+      return UnifyResults(std::move(sub_results), LogicOp::OR);
+    } else if (auto* tag_index = dynamic_cast<TagIndex*>(base_index)) {
+      vector<IndexResult> sub_results;
+      sub_results.reserve(tag_index->GetTerms().size());
+
```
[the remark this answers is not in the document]

src/server/search/search_family_test.cc
```diff
@@ -2278,4 +2278,40 @@ TEST_F(SearchFamilyTest, PrefixSearchWithSynonyms) {
   EXPECT_THAT(resp, AreDocIds("doc:6"));  // Should only find macintosh
 }
 
+TEST_F(SearchFamilyTest, SearchNonNullFields) {
+  EXPECT_EQ(Run({"ft.create", "i1", "schema", "title", "text", "tags", "tag", "score", "numeric"}),
```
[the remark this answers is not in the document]

src/core/search/base.h
vyavdoshenko: Fixed.

src/core/search/indices.h
vyavdoshenko: Fixed.

src/core/search/indices.h
vyavdoshenko: Fixed.

src/server/search/search_family_test.cc
vyavdoshenko: Test was added

src/core/search/ast_expr.h
vyavdoshenko: Fixed.

src/core/search/base.h
vyavdoshenko: Fixed.

BagritsevichStepan (approved):

PR #4942, opened 2025-04-15
fix(search_family): Fix SORTBY option in FT.SEARCH for non-sortable fields and KNN search

BagritsevichStepan:
fixes dragonflydb#4939, dragonflydb#4934

Fixes two issues. They are good described in the tickets above.

### DIFF
--- a/src/server/search/aggregator.h
+++ b/src/server/search/aggregator.h
@@ -15,6 +15,10 @@
 #include "facade/reply_builder.h"
 #include "io/io.h"
 
+namespace dfly {
+enum class SortOrder;
+}
+
 namespace dfly::aggregate {
 
 struct Reducer;
@@ -34,8 +38,6 @@ struct AggregationResult {
 };
 
 struct SortParams {
-  enum class SortOrder { ASC, DESC };
-
   constexpr static int64_t kSortAll = -1;
 
   bool SortAll() const {
--- a/src/server/search/doc_index.cc
+++ b/src/server/search/doc_index.cc
@@ -50,6 +50,68 @@ void TraverseAllMatching(const DocIndex& index, const OpArgs& op_args, F&& f) {
   } while (cursor);
 }
 
+bool IsSortableField(std::string_view field_identifier, const search::Schema& schema) {
+  auto it = schema.fields.find(field_identifier);
+  return it != schema.fields.end() && (it->second.flags & search::SchemaField::SORTABLE);
+}
+
+SearchFieldsList ToSV(const search::Schema& schema, const std::optional<SearchFieldsList>& fields) {
+  SearchFieldsList sv_fields;
+  if (fields) {
+    sv_fields.reserve(fields->size());
+    for (const auto& field : fields.value()) {
+      sv_fields.push_back(field.View());
+    }
+  }
+  return sv_fields;
+}
+
+using SortIndiciesFieldsList =
+    std::vector<std::pair<string_view /*identifier*/, string_view /*alias*/>>;
+
+std::pair<SearchFieldsList, SortIndiciesFieldsList> PreprocessAggregateFields(
+    const search::Schema& schema, const AggregateParams& params,
+    const std::optional<SearchFieldsList>& load_fields) {
+  absl::flat_hash_map<std::string_view, SearchField> fields_by_identifier;
+  absl::flat_hash_map<std::string_view, std::string_view> sort_indicies_aliases;
+  fields_by_identifier.reserve(schema.field_names.size());
+  sort_indicies_aliases.reserve(schema.field_names.size());
+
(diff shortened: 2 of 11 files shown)


vyavdoshenko:
lgtm

romange:
it's quite a big change, while it's clear what it fixes, it's not clear what was missing or faulty and how it was fixed.
I see lots of code was removed, so I am curious what happens in this PR.

src/server/search/doc_index.cc
```diff
@@ -50,6 +50,68 @@ void TraverseAllMatching(const DocIndex& index, const OpArgs& op_args, F&& f) {
   } while (cursor);
 }
 
+bool IsSortableField(std::string_view field_identifier, const search::Schema& schema) {
```
BagritsevichStepan, line 53: All the code here was simply moved. Previously, we had non-class functions defined in a non-anonymous namespace, so I moved them into an anonymous namespace.

src/server/search/doc_index.cc
```diff
@@ -315,72 +367,32 @@ SearchResult ShardDocIndex::Search(const OpArgs& op_args, const SearchParams& pa
...
-  }
-
-  return SearchResult{search_results.total - expired_count, std::move(out),
-                      std::move(search_results.profile)};
-}
-
-using SortIndiciesFieldsList =
-    std::vector<std::pair<string_view /*identifier*/, string_view /*alias*/>>;
-
-std::pair<SearchFieldsList, SortIndiciesFieldsList> PreprocessAggregateFields(
```
BagritsevichStepan, line 338: This was moved. But I can revert this

src/server/search/search_family.cc
```diff
@@ -575,6 +573,38 @@ void SearchReply(const SearchParams& params, std::optional<search::AggregationIn
 
   const size_t offset = std::min(params.limit_offset, size);
   const size_t limit = std::min(size - offset, params.limit_total);
+  const size_t end = offset + limit;
+  DCHECK(end <= docs.size());
+
+  if (params.sort_option) {
+    auto field_alias = params.sort_option->field.NameView();
+
+    auto comparator = [&](const SerializedSearchDoc* l_doc, const SerializedSearchDoc* r_doc) {
```
BagritsevichStepan, line 582: All other updates are related to the issue with sorting the non-sortable fields. Maybe I can remove changes to the SearchField and use just string for SortParams. But it is not a big change

BagritsevichStepan:
Here's what's happening: previously, sorting was performed inside the search logic itself, using a AST node — AstSortNode. This meant that sorting happened locally on each shard. And this will not work as needed.

Actually, since we are sorting only inside shards I think we can find a test case where the final result is not globally sorted.

I've now moved the sorting logic out of the shards into the coordinator thread — similar to how it's handled in the FT.AGGREGATE command

romange:
For future, it is best to separate it into two PRs : the one that moves the code and does not change functionality and the one that changes functionality. otherwise it's a nightmare for a reviewer to review the code.

romange:
while I understand the bug, it is possible that sorting locally on each shard might still be beneficial in the future - because then you can merge-sort already sorted results coming from each shard

BagritsevichStepan:
> comment: while I understand the bug, it is possible that sorting locally on each shard might still be beneficial in the future - because then you can merge-sort already sorted results coming from each shard in more efficient manner in `O(logK*N)`

The issue here is that we also support sorting for KNN search. In this case, we first perform KNN, then apply LIMIT and OFFSET (i.e., take a subarray of the sorted KNN results), and then sort that subarray by a field value.

So, the possible solution here is to have separate logic for knn results and sort them in the coordinator thread. For basic sort we can still sort in the shard.
However, I don't think it's worth including this in 1.29. It would be better to create a new ticket and implement it properly in a separate PR.

src/server/search/aggregator.h
```diff
@@ -15,6 +15,10 @@
 #include "facade/reply_builder.h"
 #include "io/io.h"
 
+namespace dfly {
+enum class SortOrder;
```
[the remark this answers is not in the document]
BorysTheDev: It's not a problem and it's absolutely ok

src/server/search/search_family.cc
```diff
@@ -575,6 +573,38 @@ void SearchReply(const SearchParams& params, std::optional<search::AggregationIn
...
+      auto& l = l_doc->values;
+      auto& r = r_doc->values;
+
+      auto l_it = l.find(field_alias);
+      auto r_it = r.find(field_alias);
```
BorysTheDev, line 587: looks really slow

BorysTheDev (approved):

src/server/search/search_family.cc
BagritsevichStepan: I will fix it for the next version

PR #4947, opened 2025-04-16
fix(set_family): Transfer TTL flag from DenseLink object in delete

### LINKED ISSUE #3915: `SADDEX` is broken
**Describe the bug**
`SADDEX` fails to expire members occasionally. Leaving them not having an expiration.

**To Reproduce**
Here's a test run in python with the redis-py package:
```python
from asyncio import sleep, Runner, gather
from random import random
from time import time_ns
from redis.asyncio import Redis

TIMEOUT = 10
KEY = 'concurrentRequest:user:1'


async def simulate_request(redis):
    member = time_ns()
    await redis.execute_command('SADDEX', KEY, TIMEOUT, member)
    await sleep(random())  # simulate handling the request

    if random() < 0.05: # 5% chance to fail, members should be expired automatically
        raise Exception
    await redis.srem(KEY, member)


async def main():
    # simulate 5000 requests
    async with Redis(decode_responses=True) as redis:
        requests = [simulate_request(redis) for _ in range(5000)]
        await gather(*requests, return_exceptions=True)

        # wait for the remaining failed requests to expire
        await sleep(TIMEOUT + 1) 
        # assert the set is empty after running the test
        cardinality = await redis.scard(KEY)
        assert cardinality == 0


if __name__ == "__main__":
    with Runner() as runner:
        runner.run(main())

```

**Expected behavior**
The test runs without error, i.e. the set should be empty.

 - OS: Debian 11 Bullseye
 - Dragonfly Version: v1.23.2-4959bef8d17e6132b4227ea7ca413faf1b1dc037

**Additional context**
I have a service in which I need to keep track of client's requests concurrency. I use a set to store the request starting-time as members, then it's `SREM` when the response is returned, therefore keeping the cardinality as the number of on-flight requests. I use `SADDEX` (add members with expiration to a set) instead of the normal `SADD` so that if any error occurs, the member will get removed automatically after a certain timeout. I'm aware that `SADDEX` is dragonfly-only and experimental, it's the one of the reason I'm using dragonfly over redis.

mkaruza:
When extracting DensePtr from LinkObject we need to transfer TTL flag before this DensePtr is assigned.

Fixes #3915

### DIFF
--- a/src/core/dense_set.cc
+++ b/src/core/dense_set.cc
@@ -648,6 +648,8 @@ void DenseSet::Delete(DensePtr* prev, DensePtr* ptr) {
 
       DenseLinkKey* plink = prev->AsLink();
       DensePtr tmp = DensePtr::From(plink);
+      // Transfer TTL flag
+      tmp.SetTtl(prev->HasTtl());
       DCHECK(ObjectAllocSize(tmp.GetObject()));
 
       FreeLink(plink);
--- a/src/server/set_family_test.cc
+++ b/src/server/set_family_test.cc
@@ -412,4 +412,17 @@ TEST_F(SetFamilyTest, SAddEx) {
   EXPECT_THAT(Run({"saddex", "key", "KEEPTTL", "2"}), ErrArg("wrong number of arguments"));
 }
 
+TEST_F(SetFamilyTest, CheckSetLinkExpiryTransfer) {
+  for (int i = 0; i < 10; i++) {
+    EXPECT_THAT(Run({"SADDEX", "key", "5", absl::StrCat(i)}), IntArg(1));
+  }
+  for (int i = 0; i < 9; i++) {
+    Run({"SREM", "key", absl::StrCat(i)});
+  }
+  EXPECT_THAT(Run({"SCARD", "key"}), IntArg(1));
+  AdvanceTime(6000);
+  Run({"SMEMBERS", "key"});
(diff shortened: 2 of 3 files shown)


BorysTheDev (approved):

src/core/dense_set.cc
```diff
@@ -648,6 +648,8 @@ void DenseSet::Delete(DensePtr* prev, DensePtr* ptr) {
 
       DenseLinkKey* plink = prev->AsLink();
       DensePtr tmp = DensePtr::From(plink);
+      // Transfer TTL flag
```
kostasrim, line 651: I think we need to set the ttl bit in the wrapping pointer. See comments in AddOrReplaceObj

mkaruza: Wrapping is destroyed so we need to set TTL of object. This is same behavior as what `Grow()` does.

src/server/set_family_test.cc
```diff
@@ -412,4 +412,17 @@ TEST_F(SetFamilyTest, SAddEx) {
   EXPECT_THAT(Run({"saddex", "key", "KEEPTTL", "2"}), ErrArg("wrong number of arguments"));
 }
 
+TEST_F(SetFamilyTest, CheckSetLinkExpiryTransfer) {
```
romange, line 415: can you please add an equivalent test in `src/core/string_set_test.cc`, @mkaruza ?

mkaruza: Added

PR #4952, opened 2025-04-17
chore: add metrics for heartbeat evictions

### LINKED ISSUE #4950: background expiry is not working in some cases
`EngineShard::RetireExpiredAndEvict()` has condition `if (expt->size() > pt->size() / 4) {` that prevents expiring items in some cases.

Moreover, we lack any stats telling us about background expiry performance, it is not possible to know 
whether it has run or not:
1. `DeleteExpiredStep` returns some stats, we should propagate them into EngineShard::Stats and then expose them via `STATS` section in info.
2. we should `VLOG(1)` them so we could track these via logs if needed.
3. there is no need checking if `expt->size()  > pt>size() / 4)`, only if whether it's empty.

kostasrim:
* evict in heartbeat if expire table is not empty
* add metrics around heartbeat evictions (total evictions and total evicted bytes) 

Resolves  #4950

### DIFF
--- a/src/server/engine_shard.cc
+++ b/src/server/engine_shard.cc
@@ -297,7 +297,7 @@ string EngineShard::TxQueueInfo::Format() const {
 }
 
 EngineShard::Stats& EngineShard::Stats::operator+=(const EngineShard::Stats& o) {
-  static_assert(sizeof(Stats) == 64);
+  static_assert(sizeof(Stats) == 80);
 
 #define ADD(x) x += o.x
 
@@ -309,6 +309,8 @@ EngineShard::Stats& EngineShard::Stats::operator+=(const EngineShard::Stats& o) 
   ADD(tx_optimistic_total);
   ADD(tx_batch_schedule_calls_total);
   ADD(tx_batch_scheduled_items_total);
+  ADD(heartbeat_evictions_total);
+  ADD(heartbeat_evictions_total_in_bytes);
 
 #undef ADD
   return *this;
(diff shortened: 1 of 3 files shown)


src/server/engine_shard.cc
```diff
@@ -309,6 +309,8 @@ EngineShard::Stats& EngineShard::Stats::operator+=(const EngineShard::Stats& o)
   ADD(tx_optimistic_total);
   ADD(tx_batch_schedule_calls_total);
   ADD(tx_batch_scheduled_items_total);
+  ADD(heartbeat_evictions_total);
```
romange, line 312: it's not evictions, it's expires

src/server/engine_shard.cc
```diff
@@ -805,12 +807,16 @@ void EngineShard::RetireExpiredAndEvict() {
...
     db_cntx.db_index = i;
     auto [pt, expt] = db_slice.GetTables(i);
-    if (expt->size() > pt->size() / 4) {
+    if (expt->size() > 0) {
       DbSlice::DeleteExpiredStats stats = db_slice.DeleteExpiredStep(db_cntx, ttl_delete_target);
 
       eviction_goal -= std::min(eviction_goal, size_t(stats.deleted_bytes));
       counter_[TTL_TRAVERSE].IncBy(stats.traversed);
       counter_[TTL_DELETE].IncBy(stats.deleted);
+      stats_.heartbeat_evictions_total += stats.deleted;
```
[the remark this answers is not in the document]
kostasrim: oh :facepalm:

src/server/engine_shard.cc
kostasrim: oh 🤦

src/server/engine_shard.cc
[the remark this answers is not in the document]
kostasrim: done, slightly different name. If you don't like it I will rename it :smile:

romange (approved):

src/server/server_family.cc
```diff
@@ -2486,6 +2486,9 @@ string ServerFamily::FormatInfoMetrics(const Metrics& m, std::string_view sectio
     append("rejected_connections", -1);
     append("expired_keys", m.events.expired_keys);
     append("evicted_keys", m.events.evicted_keys);
+    append("total_expired_keys_in_heartbeat", m.shard_stats.heartbeat_expired_total);
```
romange, line 2489: let's make it consistent and a bit shorter?
`total_heartbeat_expired_keys`, `total_heartbeat_expired_bytes`, `total_hearbeat_expired_calls` ?

kostasrim: sounds good!

kostasrim: so much better now

romange (approved):

PR #4953, opened 2025-04-17
feat(server): Expose total host memory as part of info metrics

abhijat:
The total host memory is expected by some applications such as mastodon to determine if sidekiq is operational. This field is exposed as part of metrics.

Some context is in https://github.com/dragonflydb/dragonfly/issues/4596#issuecomment-2809469606

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -2423,6 +2423,8 @@ string ServerFamily::FormatInfoMetrics(const Metrics& m, std::string_view sectio
       }
     }
     append("table_used_memory", total.table_mem_usage);
+    append("total_system_memory", total_host_memory);
+    append("total_system_memory_human", HumanReadableNumBytes(total_host_memory));
     append("num_buckets", total.bucket_count);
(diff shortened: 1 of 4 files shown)


src/server/server_family.cc
```diff
@@ -2423,6 +2423,8 @@ string ServerFamily::FormatInfoMetrics(const Metrics& m, std::string_view sectio
       }
     }
     append("table_used_memory", total.table_mem_usage);
+    append("total_system_memory", total_host_memory);
```
romange, line 2426: let's take it offline - it's news to me that sidekiq requires it - I want to understand more.
exposing free host memory is tricky - it depends on many environmental factors so I  prefer to avoid it if possible

abhijat: Sure.

It's not actually sidekiq that needs it, there is a method in a mastodon module that uses the stats to determine whether sidekiq is usable, https://github.com/mastodon/mastodon/blob/main/app/workers/scheduler/self_destruct_scheduler.rb#L22-L26

PR #4975, opened 2025-04-22
chore(acl_family): Move some variables outside loop

abhijat:
A couple of variables are moved outside of a loop and reused within it.

### DIFF
--- a/src/server/acl/acl_family.cc
+++ b/src/server/acl/acl_family.cc
@@ -1189,16 +1189,16 @@ void AclFamily::BuildIndexers(RevCommandsIndexStore families) {
   acl::NumberOfFamilies(families.size());
   CommandsRevIndexer(std::move(families));
   CategoryToCommandsIndexStore index;
-  cmd_registry_->Traverse([&](std::string_view name, auto& cid) {
-    auto cat = cid.acl_categories();
+  cmd_registry_->Traverse([&](std::string_view, auto& cid) {
+    const uint32_t cat = cid.acl_categories();
+    const size_t family = cid.GetFamily();
+    const uint64_t bit_index = cid.GetBitIndex();
     for (size_t i = 0; i < 32; ++i) {
-      if (cat & (1 << i)) {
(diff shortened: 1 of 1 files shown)


src/server/acl/acl_family.cc
```diff
@@ -1189,16 +1189,16 @@ void AclFamily::BuildIndexers(RevCommandsIndexStore families) {
...
   CategoryToCommandsIndexStore index;
-  cmd_registry_->Traverse([&](std::string_view name, auto& cid) {
-    auto cat = cid.acl_categories();
+  cmd_registry_->Traverse([&](std::string_view, auto& cid) {
+    const uint32_t cat = cid.acl_categories();
+    const size_t family = cid.GetFamily();
+    const uint64_t bit_index = cid.GetBitIndex();
     for (size_t i = 0; i < 32; ++i) {
-      if (cat & (1 << i)) {
+      if (cat & 1 << i) {
```
kostasrim, line 1197: I always add parenthesis because I always forget the operator precedence :smile:

src/server/acl/acl_family.cc
```diff
@@ -1189,16 +1189,16 @@ void AclFamily::BuildIndexers(RevCommandsIndexStore families) {
   acl::NumberOfFamilies(families.size());
   CommandsRevIndexer(std::move(families));
   CategoryToCommandsIndexStore index;
-  cmd_registry_->Traverse([&](std::string_view name, auto& cid) {
-    auto cat = cid.acl_categories();
+  cmd_registry_->Traverse([&](std::string_view, auto& cid) {
+    const uint32_t cat = cid.acl_categories();
+    const size_t family = cid.GetFamily();
+    const uint64_t bit_index = cid.GetBitIndex();
```
kostasrim, line 1195: Same here, for non perf critical code I put them inside loop. BuildIndexers is called only once anyway at startup

kostasrim (approved):
IMO I have slightly different preferences :smile:

Issue #4918, opened 2025-04-10
Malformatted RESP3 response for HRANDFIELD WITHVALUES command


yatsukhnenko:
**Describe the bug**
`HRANDFIELD` command `WITHVALUES` returns result in the same format for both RESP2 and RESP3 which is incompatible with other servers.

**To Reproduce**
1. Insert records using `redis-cli hset foo a 1 b 2 c 3`
2. Query records using `redis-cli -3 hrandfield foo 2 withvalues`
3. See that result is a flat list of fields and values (like for RESP2)
```bash
localhost~$ redis-cli -3 hrandfield foo 2 withvalues
1) "b"
2) "2"
3) "c"
4) "3"
```

**Expected behavior**
Valkey and Redis returns result as a list of nested lists with fields and values
```bash
localhost~$ redis-cli -3 hrandfield foo 2 withvalues
1) 1) "a"
   2) "1"
2) 1) "b"
   2) "2"
```

PR #4978, opened 2025-04-22
chore(hset_family): Support resp3 format for hrandfield

abhijat:
Return nested arrays if hrandfield is used with values and resp3 format is set

fixes https://github.com/dragonflydb/dragonfly/issues/4918

### DIFF
--- a/src/server/hset_family.cc
+++ b/src/server/hset_family.cc
@@ -1273,9 +1273,20 @@ void HSetFamily::HRandField(CmdArgList args, const CommandContext& cmd_cntx) {
   if (result) {
     if (result->size() == 1 && args.size() == 1)
       rb->SendBulkString(result->front());
-    else if (with_values)
-      rb->SendBulkStrArrAsPairs(*result);
-    else
+    else if (with_values) {
+      const auto result_size = result->size();
+      DCHECK(result_size % 2 == 0)
+          << "unexpected size of strings " << result_size << ", expected pairs";
+      SinkReplyBuilder::ReplyScope scope{rb};
+      const bool is_resp3 = rb->IsResp3();
+      rb->StartArray(is_resp3 ? result_size / 2 : result_size);
+      for (size_t i = 0; i < result_size; i += 2) {
+        if (is_resp3)
+          rb->StartArray(2);
+        rb->SendBulkString((*result)[i]);
+        rb->SendBulkString((*result)[i + 1]);
+      }
+    } else
       rb->SendBulkStrArr(*result, RedisReplyBuilder::ARRAY);
   } else if (result.status() == OpStatus::KEY_NOTFOUND) {
     if (args.size() == 1)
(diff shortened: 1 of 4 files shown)


abhijat:
The method added in the PR is quite similar to the `RedisReplyBuilder::SendScoredArray` so a template method could be added to abstract common function, but I added a new method to not add complexity.

src/server/hset_family_test.cc
```diff
@@ -541,4 +541,30 @@ TEST_F(HSetFamilyTest, KeyRemovedWhenEmpty) {
...
+      {"b", "2"},
+      {"c", "3"},
+  };
+  Run({"HELLO", "3"});
+  EXPECT_THAT(Run({"HSET", "key", "a", "1", "b", "2", "c", "3"}), IntArg(3));
+  auto resp = Run({"HRANDFIELD", "key", "3", "WITHVALUES"});
+  EXPECT_THAT(resp, ArrLen(3));
+  for (const auto& v : resp.GetVec()) {
+    EXPECT_THAT(v, ArrLen(2));
+    EXPECT_THAT(v.GetVec()[0], AnyOf("a", "b", "c"));
```
kostasrim, line 556: why not `const auto &v = v.GetVec()`. You called it 3 times :smile:

src/server/hset_family.cc
```diff
@@ -1271,10 +1271,12 @@ void HSetFamily::HRandField(CmdArgList args, const CommandContext& cmd_cntx) {
   auto* rb = static_cast<RedisReplyBuilder*>(cmd_cntx.rb);
   OpResult<StringVec> result = cmd_cntx.tx->ScheduleSingleHopT(std::move(cb));
   if (result) {
-    if ((result->size() == 1) && (args.size() == 1))
+    if (result->size() == 1 && args.size() == 1)
       rb->SendBulkString(result->front());
+    else if (with_values)
+      rb->SendBulkStrArrAsPairs(*result);
```
kostasrim, line 1277: I would make this a lambda and call it here instead of moving it within reply_builder. My rationale for this is that I suspect, that redis/valkey itself has a bug here. From the docs:

```
RESP2 Reply
[Array reply](https://redis.io/docs/latest/develop/reference/protocol-spec#arrays): a list of fields and their values when count and WITHVALUES were both used.

and 
RESP3 Reply
selected field when the count option is not used
[Array reply](https://redis.io/docs/latest/develop/reference/protocol-spec#arrays): a list of fields and their values when count and WITHVALUES were both used.
```

kostasrim: Maybe something like:

```
if resp 2 -> rb->SendBulkString()
else {
  DCHECK(strs_size % 2 == 0) << "unexpected size of strings " << strs_size << ", expected pairs";
  StartArray(strs_size / 2);
  for (size_t i = 0; i < strs_size; i += 2) {
    StartArray(2);
    SendBulkString(strs[i]);
    SendBulkString(strs[i + 1]);
  }
}
```

kostasrim (commented):
LGTM, minor comments

src/server/hset_family_test.cc
abhijat: fixed

src/server/hset_family.cc
abhijat: changed. I had to copy the reply scope object to the scope to ensure the message gets flushed.

Issue #4979, opened 2025-04-23
Dragonfly crashes using VECTOR HNSW schema


vyavdoshenko:
Dragonfly crashes in the following scenario:
```
FT.CREATE idx SCHEMA n NUMERIC v VECTOR HNSW 8 TYPE FLOAT16 DIM 4 DISTANCE_METRIC L2 M 65537
HSET doc n 0
HSET doc n 1
```

Expected behavior:
Process the parameters on the creation of the schema and reject creation on inappropriate parameters.

PR #4987, opened 2025-04-23
fix: HNSW schema deleting document crash

vyavdoshenko:
The scenario that caused the crash:
```
FT.CREATE idx SCHEMA n NUMERIC v VECTOR HNSW 8 TYPE FLOAT16 DIM 4 DISTANCE_METRIC L2 M 65537
HSET doc n 0
DEL doc
```

fixes: https://github.com/dragonflydb/dragonfly/issues/4979


src/server/search/doc_index.cc
```diff
@@ -322,10 +322,14 @@ void ShardDocIndex::RemoveDoc(string_view key, const DbContext& db_cntx, const P
   if (!indices_)
     return;
 
-  auto accessor = GetAccessor(db_cntx, pv);
-  auto id = key_index_.Remove(key);
-  if (id) {
-    indices_->Remove(id.value(), *accessor);
+  try {
```
romange, line 325: not sure I understand. What call exactly throws?

romange: now I understand, I think this should be improved.

vyavdoshenko: The exception occurs inside the library:
_third_party/libs/hnswlib/include/hnswlib/hnswalg.h_

Screenshot 2025-04-23 at 18 06 06

src/server/search/search_family_test.cc
```diff
@@ -2722,4 +2722,16 @@ TEST_F(SearchFamilyTest, JsonWithNullFields) {
...
 
+TEST_F(SearchFamilyTest, TestTwoHsetDocumentCreationCrashes) {
+  EXPECT_EQ(Run({"FT.CREATE", "idx", "SCHEMA", "n", "NUMERIC", "v", "VECTOR", "HNSW", "8", "TYPE",
+                 "FLOAT16", "DIM", "4", "DISTANCE_METRIC", "L2", "M", "65536"}),
+            "OK");
+
+  auto res = Run({"HSET", "doc", "n", "0"});
+  EXPECT_EQ(res, 1);
+
+  res = Run({"HSET", "doc", "n", "1"});
```
romange, line 2733: Since it's during removal, I assume if we call `DEL doc` it would achieve similar effect? If yes, it's a clearer test, imho.

vyavdoshenko: Yes, correct. I checked it. `DEL doc` archives the same effect. I will change the test.

src/server/search/doc_index.cc
romange: yes, but it's a legitimate flow. so we should protect ourselves and there is no reason for WARNING.
So at least I would move throw/catch into `dfly::search::HnswlibAdapter::Remove` and remove any logs

src/server/search/search_family_test.cc

BorysTheDev (approved):

PR #4989, opened 2025-04-24
fix: bitops_family crash fixed

vyavdoshenko:
fixes: https://github.com/dragonflydb/dragonfly/issues/4988

### DIFF
--- a/src/server/bitops_family.cc
+++ b/src/server/bitops_family.cc
@@ -138,7 +138,11 @@ constexpr int32_t GetByteIndex(uint32_t offset) noexcept {
 }
 
 uint8_t GetByteValue(string_view str, uint32_t offset) {
-  return static_cast<uint8_t>(str[GetByteIndex(offset)]);
+  int32_t byte_index = GetByteIndex(offset);
+  if (byte_index >= static_cast<int32_t>(str.size())) {
+    return 0;
+  }
+  return static_cast<uint8_t>(str[byte_index]);
 }
 
 constexpr bool CheckBitStatus(uint8_t byte, uint32_t offset) {
--- a/src/server/bitops_family.cc
+++ b/src/server/bitops_family.cc
@@ -138,11 +138,7 @@ constexpr int32_t GetByteIndex(uint32_t offset) noexcept {
 }
 
 uint8_t GetByteValue(string_view str, uint32_t offset) {
-  int32_t byte_index = GetByteIndex(offset);
-  if (byte_index >= static_cast<int32_t>(str.size())) {
-    return 0;
-  }
-  return static_cast<uint8_t>(str[byte_index]);
+  return static_cast<uint8_t>(str[GetByteIndex(offset)]);
 }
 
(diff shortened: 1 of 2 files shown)


src/server/bitops_family.cc
```diff
@@ -729,7 +729,12 @@ ResultType Get::ApplyTo(Overflow ov, const string* bitfield) {
   auto last_byte_offset = GetByteIndex(attr_.offset + attr_.encoding_bit_size - 1);
 
   uint32_t lsb = attr_.offset + attr_.encoding_bit_size - 1;
-  if (last_byte_offset > total_bytes) {
+  if (last_byte_offset >= total_bytes) {
+    return {};
+  }
+
+  int32_t byte_index = GetByteIndex(offset);
+  if (byte_index >= total_bytes) {
```
kostasrim, line 737:
```suggestion
  if (last_byte_offset >= total_bytes) {
```

kostasrim: This is sufficient. We check the expression above `GetByteIndex(attr_.offset + attr_.encoding_bit_size - 1);`

src/server/bitops_family.cc
```diff
@@ -738,6 +743,11 @@ ResultType Get::ApplyTo(Overflow ov, const string* bitfield) {
 
   int64_t result = 0;
   for (size_t i = 0; i < attr_.encoding_bit_size; ++i) {
+    int32_t lsb_byte_index = GetByteIndex(lsb);
+    if (lsb_byte_index >= total_bytes) {
+      return {};
+    }
```
kostasrim, line 749: We don't have to check again here. The above check is sufficient

vyavdoshenko: Fixed

src/server/bitops_family.cc
vyavdoshenko: @kostasrim 
This is not enough.
I updated the condition, please take a look.

src/server/bitops_family.cc
```diff
@@ -728,17 +728,25 @@ ResultType Get::ApplyTo(Overflow ov, const string* bitfield) {
   const size_t offset = attr_.offset;
   auto last_byte_offset = GetByteIndex(attr_.offset + attr_.encoding_bit_size - 1);
 
+  if (GetByteIndex(offset) >= total_bytes && attr_.encoding_bit_size > 0) {
```
kostasrim, line 731:
```suggestion
  if (GetByteIndex(offset) >= total_bytes) {
```

kostasrim: because `attr_.encoding_bit_size > 0` is always `true`. (See how we parse the arguments for this command)

kostasrim (approved):
LGTM. Small comment

src/server/bitops_family.cc
vyavdoshenko: Fixed.

kostasrim (approved):
:100:

PR #4990, opened 2025-04-24
feat(tools/replay): Add pipeline latency distribution data

Pothulapati:
- **feat(replay): add latency distributions**
- **wrap exec instead**
- **also add avg latency**
- **also include pipeline range**
- **display both at the end**

### DIFF
--- a/tools/replay/workers.go
+++ b/tools/replay/workers.go
@@ -78,24 +78,28 @@ func (c *ClientWorker) Run(pace bool, worker *FileWorker) {
 			time.Sleep(lag)
 		}
 
-		start := time.Now()
 		c.pipe.Do(context.Background(), msg.values...).Result()
-		latency := float64(time.Since(start).Microseconds())
-		worker.latencyMu.Lock()
-		worker.latencyDigest.Add(latency, 1)
-		worker.latencyMu.Unlock()
-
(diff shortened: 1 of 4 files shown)


tools/replay/workers.go
```diff
@@ -106,6 +152,11 @@ func NewClient(w *FileWorker, pace bool) *ClientWorker {
 }
 
 func (w *FileWorker) Run(file string, wg *sync.WaitGroup) {
+	w.latencyDigest = tdigest.NewWithCompression(1000)
```
Pothulapati, line 155: Is 1000 good enough?

ashotland: what is the semantics of this parameter ?

tools/replay/workers.go
```diff
@@ -47,6 +48,18 @@ type ClientWorker struct {
...
+// Pipeline length ranges for summary
+var pipelineRanges = []struct {
+	label string
+	min   int
+	max   int // inclusive, except last
+}{
+	{"0-29", 0, 29},
+	{"30-79", 30, 79},
+	{"80-199", 80, 199},
+	{"200+", 200, 1 << 30},
```
ashotland, line 60: @romange for thresholds preference

romange (approved):
Thanks a lot @Pothulapati 🙏🏼

Issue #4993, opened 2025-04-24
RENAME crashes in cluster mode only


vyavdoshenko:
The scenario:
```
FT.CREATE idx1 PREFIX 1 idx1 FILTER @index=="yes" SCHEMA t TEXT
FT.CREATE idx2 PREFIX 1 idx2 FILTER @index=="yes" SCHEMA t TEXT

HSET idx1:{doc}1 t "foo1" index "yes"

RENAME idx1:{doc}1 idx2:{doc}1
RENAME idx2:{doc}1 idx1:{doc}1
```

The bug reproduces in cluster mode:
```
./dragonfly --dbfilename= --logtostderr --cluster_mode=emulated
```

Without cluster_mode, the crash doesn't occur.

PR #4995, opened 2025-04-24
fix: RENAME in cluster mode crash

vyavdoshenko:
fixes: https://github.com/dragonflydb/dragonfly/issues/4993

### DIFF
--- a/src/server/generic_family.cc
+++ b/src/server/generic_family.cc
@@ -870,6 +870,9 @@ OpResult<void> OpRen(const OpArgs& op_args, string_view from_key, string_view to
     is_prior_list = (to_res.it->second.ObjType() == OBJ_LIST);
   }
 
+  // Delete the document from the search index before deleting from the database
+  op_args.shard->search_indices()->RemoveDoc(from_key, op_args.db_cntx, from_res.it->second);
+
   bool sticky = from_res.it->first.IsSticky();
   uint64_t exp_ts = db_slice.ExpireTime(from_res.exp_it);
 
--- a/src/server/generic_family.cc
+++ b/src/server/generic_family.cc
@@ -870,7 +870,7 @@ OpResult<void> OpRen(const OpArgs& op_args, string_view from_key, string_view to
     is_prior_list = (to_res.it->second.ObjType() == OBJ_LIST);
   }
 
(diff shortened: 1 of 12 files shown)


src/server/search/search_family_test.cc
```diff
@@ -2722,4 +2722,26 @@ TEST_F(SearchFamilyTest, JsonWithNullFields) {
               AreDocIds("doc:1", "doc:2"));
 }
 
+TEST_F(SearchFamilyTest, RenameDocumentBetweenIndices) {
+  std::string original_mode = GetTestFlag("cluster_mode");
```
romange, line 2726: you can just declare `absl::FlagSaver fs;` here and then change flags - they will be restored automatically.

src/server/search/doc_index.cc
```diff
@@ -191,7 +191,10 @@ string DocIndexInfo::BuildRestoreCommand() const {
 }
 
 ShardDocIndex::DocId ShardDocIndex::DocKeyIndex::Add(string_view key) {
-  DCHECK_EQ(ids_.count(key), 0u);
+  auto it = ids_.find(key);
```
BagritsevichStepan, line 194: Please explain why we need this change. We’re asserting here that ids should never call add twice without first calling remove. From my point of view, the bug isn’t here — it seems like somewhere else we forgot to remove the key first

vyavdoshenko: Fixed.

src/server/generic_family.cc
```diff
@@ -870,6 +870,9 @@ OpResult<void> OpRen(const OpArgs& op_args, string_view from_key, string_view to
     is_prior_list = (to_res.it->second.ObjType() == OBJ_LIST);
   }
 
+  // Delete the document from the search index before deleting from the database
```
romange, line 873:
```suggestion
  // Delete the "from" document from the search index before deleting from the database
```

romange: because 3 lines above we delete the "to" doc

romange (dismissed):

romange (approved):

BagritsevichStepan (approved):
Created a ticket for other commands: #5014

PR #5000, opened 2025-04-25
fix(cluster_family): Cancel slot migration from incoming node on OOM

mkaruza:
If applying command on incoming node will result in OOM (we overflow
max_memory_limit) we are closing migration and switch state to FATAL.

### DIFF
--- a/src/server/cluster/cluster_family.cc
+++ b/src/server/cluster/cluster_family.cc
@@ -9,6 +9,7 @@
 #include <string>
 
 #include "absl/cleanup/cleanup.h"
+#include "absl/strings/str_cat.h"
 #include "base/flags.h"
 #include "base/logging.h"
 #include "facade/cmd_arg_parser.h"
@@ -16,6 +17,9 @@
 #include "facade/error.h"
 #include "server/acl/acl_commands_def.h"
 #include "server/channel_store.h"
+#include "server/cluster/cluster_defs.h"
+#include "server/cluster/incoming_slot_migration.h"
+#include "server/cluster/outgoing_slot_migration.h"
 #include "server/command_registry.h"
 #include "server/conn_context.h"
 #include "server/dflycmd.h"
@@ -726,6 +730,8 @@ static string_view StateToStr(MigrationState state) {
       return "ERROR"sv;
     case MigrationState::C_FINISHED:
       return "FINISHED"sv;
+    case MigrationState::C_FATAL:
+      return "FATAL"sv;
   }
   DCHECK(false) << "Unknown State value " << static_cast<underlying_type_t<MigrationState>>(state);
   return "UNDEFINED_STATE"sv;
@@ -765,7 +771,6 @@ void ClusterFamily::DflySlotMigrationStatus(CmdArgList args, SinkReplyBuilder* b
   };
 
   for (const auto& m : incoming_migrations_jobs_) {
-    // TODO add error status
     append_answer("in", m->GetSourceID(), node_id, m->GetState(), m->GetKeyCount(),
                   m->GetErrorStr());
   }
@@ -834,7 +839,7 @@ ClusterFamily::TakeOutOutgoingMigrations(shared_ptr<ClusterConfig> new_config,
     removed_slots.Merge(slots);
     LOG(INFO) << "Outgoing migration cancelled: slots " << slots.ToString() << " to "
               << migration.GetHostIp() << ":" << migration.GetPort();
-    migration.Finish();
+    migration.Finish(MigrationState::C_FINISHED);
     res.migrations.push_back(std::move(*it));
     outgoing_migration_jobs_.erase(it);
   }
@@ -936,6 +941,10 @@ void ClusterFamily::InitMigration(CmdArgList args, SinkReplyBuilder* builder) {
     DeleteSlots(slots);
   }
 
+  if (migration->GetState() == MigrationState::C_FATAL) {
+    return builder->SendError(absl::StrCat("-", IncomingSlotMigration::kMigrationOOM));
+  }
+
   migration->Init(flows_num);
 
   return builder->SendOk();
@@ -955,8 +964,13 @@ void ClusterFamily::DflyMigrateFlow(CmdArgList args, SinkReplyBuilder* builder,
   cntx->conn()->SetName(absl::StrCat("migration_flow_", source_id));
 
   auto migration = GetIncomingMigration(source_id);
+
   if (!migration) {
     return builder->SendError(kIdNotFound);
+  }
+
+  if (migration->GetState() == MigrationState::C_FATAL) {
+    return builder->SendError(absl::StrCat("-", IncomingSlotMigration::kMigrationOOM));
   }
 
   DCHECK(cntx->sync_dispatch);
@@ -967,6 +981,10 @@ void ClusterFamily::DflyMigrateFlow(CmdArgList args, SinkReplyBuilder* builder,
   builder->SendOk();
 
   migration->StartFlow(shard_id, cntx->conn()->socket());
+
+  if (migration->GetState() == MigrationState::C_FATAL) {
+    migration->Stop();
+  }
 }
 
 void ClusterFamily::ApplyMigrationSlotRangeToConfig(std::string_view node_id,
@@ -1041,7 +1059,11 @@ void ClusterFamily::DflyMigrateAck(CmdArgList args, SinkReplyBuilder* builder) {
     return builder->SendError(kIdNotFound);
 
   if (!migration->Join(attempt)) {
-    return builder->SendError("Join timeout happened");
+    if (migration->GetState() == MigrationState::C_FATAL) {
+      return builder->SendError(absl::StrCat("-", IncomingSlotMigration::kMigrationOOM));
+    } else {
+      return builder->SendError("Join timeout happened");
+    }
   }
 
   ApplyMigrationSlotRangeToConfig(migration->GetSourceID(), migration->GetSlots(), true);
--- a/src/server/cluster/cluster_family.cc
+++ b/src/server/cluster/cluster_family.cc
@@ -835,7 +835,7 @@ ClusterFamily::TakeOutOutgoingMigrations(shared_ptr<ClusterConfig> new_config,
     removed_slots.Merge(slots);
     LOG(INFO) << "Outgoing migration cancelled: slots " << slots.ToString() << " to "
               << migration.GetHostIp() << ":" << migration.GetPort();
-    migration.Finish();
+    migration.Finish(MigrationState::C_FINISHED);
     res.migrations.push_back(std::move(*it));
     outgoing_migration_jobs_.erase(it);
   }
--- a/src/server/cluster/cluster_family.cc
+++ b/src/server/cluster/cluster_family.cc
@@ -1039,7 +1039,7 @@ void ClusterFamily::DflyMigrateAck(CmdArgList args, SinkReplyBuilder* builder) {
                    [source_id = source_id](const auto& m) { return m.node_info.id == source_id; });
   if (m_it == in_migrations.end()) {
     LOG(WARNING) << "migration isn't in config";
-    return builder->SendError(kUnknownMigration);
+    return builder->SendSimpleString(kUnknownMigration);
   }
 
   auto migration = GetIncomingMigration(source_id);
--- a/src/server/cluster/cluster_family.cc
+++ b/src/server/cluster/cluster_family.cc
@@ -9,7 +9,6 @@
 #include <string>
 
 #include "absl/cleanup/cleanup.h"
-#include "absl/strings/str_cat.h"
 #include "base/flags.h"
 #include "base/logging.h"
 #include "facade/cmd_arg_parser.h"
@@ -17,9 +16,6 @@
 #include "facade/error.h"
 #include "server/acl/acl_commands_def.h"
 #include "server/channel_store.h"
-#include "server/cluster/cluster_defs.h"
-#include "server/cluster/incoming_slot_migration.h"
-#include "server/cluster/outgoing_slot_migration.h"
 #include "server/command_registry.h"
 #include "server/conn_context.h"
 #include "server/dflycmd.h"
@@ -839,7 +835,7 @@ ClusterFamily::TakeOutOutgoingMigrations(shared_ptr<ClusterConfig> new_config,
     removed_slots.Merge(slots);
     LOG(INFO) << "Outgoing migration cancelled: slots " << slots.ToString() << " to "
               << migration.GetHostIp() << ":" << migration.GetPort();
-    migration.Finish(MigrationState::C_FINISHED);
+    migration.Finish();
     res.migrations.push_back(std::move(*it));
     outgoing_migration_jobs_.erase(it);
   }
@@ -930,7 +926,7 @@ void ClusterFamily::InitMigration(CmdArgList args, SinkReplyBuilder* builder) {
 
   if (!migration) {
     VLOG(1) << "Unrecognized incoming migration from " << source_id;
-    return builder->SendSimpleString(OutgoingMigration::kUnknownMigration);
+    return builder->SendSimpleString(kUnknownMigration);
   }
 
   if (migration->GetState() != MigrationState::C_CONNECTING) {
@@ -942,7 +938,7 @@ void ClusterFamily::InitMigration(CmdArgList args, SinkReplyBuilder* builder) {
   }
 
   if (migration->GetState() == MigrationState::C_FATAL) {
-    return builder->SendError(absl::StrCat("-", IncomingSlotMigration::kMigrationOOM));
+    return builder->SendError(absl::StrCat("-", kIncomingMigrationOOM));
   }
 
   migration->Init(flows_num);
@@ -969,10 +965,6 @@ void ClusterFamily::DflyMigrateFlow(CmdArgList args, SinkReplyBuilder* builder,
     return builder->SendError(kIdNotFound);
   }
 
-  if (migration->GetState() == MigrationState::C_FATAL) {
-    return builder->SendError(absl::StrCat("-", IncomingSlotMigration::kMigrationOOM));
-  }
-
   DCHECK(cntx->sync_dispatch);
   // we do this to be ignored by the dispatch tracker
   // TODO provide a more clear approach
@@ -981,10 +973,6 @@ void ClusterFamily::DflyMigrateFlow(CmdArgList args, SinkReplyBuilder* builder,
   builder->SendOk();
 
   migration->StartFlow(shard_id, cntx->conn()->socket());
-
-  if (migration->GetState() == MigrationState::C_FATAL) {
-    migration->Stop();
-  }
 }
 
 void ClusterFamily::ApplyMigrationSlotRangeToConfig(std::string_view node_id,
@@ -1051,7 +1039,7 @@ void ClusterFamily::DflyMigrateAck(CmdArgList args, SinkReplyBuilder* builder) {
                    [source_id = source_id](const auto& m) { return m.node_info.id == source_id; });
   if (m_it == in_migrations.end()) {
     LOG(WARNING) << "migration isn't in config";
-    return builder->SendError(OutgoingMigration::kUnknownMigration);
+    return builder->SendError(kUnknownMigration);
   }
 
   auto migration = GetIncomingMigration(source_id);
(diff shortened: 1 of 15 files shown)


src/server/cluster/cluster_family.cc
```diff
@@ -1045,7 +1046,23 @@ void ClusterFamily::DflyMigrateAck(CmdArgList args, SinkReplyBuilder* builder) {
     return builder->SendError(kIdNotFound);
 
   if (!migration->Join(attempt)) {
-    return builder->SendError("Join timeout happened");
+    // Cleanup migraton that resulted in error
+    if (migration->GetState() == MigrationState::C_ERROR) {
+      // Restore previous cluster configuration
+      server_family_->service().proactor_pool().AwaitFiberOnAll(
+          [](util::ProactorBase*) { ClusterConfig::RestorePrevious(); });
```
BorysTheDev, line 1053: we shouldn't change the config, it can be dangerous.

src/server/cluster/cluster_family.cc
```diff
@@ -9,13 +9,17 @@
...
 #include "base/flags.h"
 #include "base/logging.h"
 #include "facade/cmd_arg_parser.h"
 #include "facade/dragonfly_connection.h"
 #include "facade/error.h"
 #include "server/acl/acl_commands_def.h"
 #include "server/channel_store.h"
+#include "server/cluster/cluster_defs.h"
+#include "server/cluster/incoming_slot_migration.h"
+#include "server/cluster/outgoing_slot_migration.h"
```
BorysTheDev, line 22: you have these includes in header

src/server/cluster/cluster_family.cc
```diff
@@ -955,10 +964,15 @@ void ClusterFamily::DflyMigrateFlow(CmdArgList args, SinkReplyBuilder* builder,
...
 
   auto migration = GetIncomingMigration(source_id);
+
   if (!migration) {
     return builder->SendError(kIdNotFound);
   }
 
+  if (migration->GetState() == MigrationState::C_FATAL) {
+    return builder->SendError(absl::StrCat("-", IncomingSlotMigration::kMigrationOOM));
+  }
```
BorysTheDev, line 974: Can't imagine how this can happen

src/server/cluster/cluster_family.cc
```diff
@@ -967,6 +981,10 @@ void ClusterFamily::DflyMigrateFlow(CmdArgList args, SinkReplyBuilder* builder,
   builder->SendOk();
 
   migration->StartFlow(shard_id, cntx->conn()->socket());
+
+  if (migration->GetState() == MigrationState::C_FATAL) {
+    migration->Stop();
+  }
```
BorysTheDev, line 987: looks like incorrect place for this logic

src/server/cluster/incoming_slot_migration.cc
```diff
@@ -7,9 +7,15 @@
...
 #include <absl/strings/str_cat.h>
 
+#include <utility>
+
 #include "base/flags.h"
 #include "base/logging.h"
 #include "cluster_utility.h"
+#include "facade/error.h"
+#include "server/cluster/cluster_defs.h"
+#include "server/cluster/outgoing_slot_migration.h"
```
BorysTheDev, line 17: outgoing_slot_migration shouldn't be included in incoming_slot_migration

src/server/cluster/incoming_slot_migration.cc
```diff
@@ -7,9 +7,15 @@
 #include <absl/cleanup/cleanup.h>
 #include <absl/strings/str_cat.h>
 
+#include <utility>
+
 #include "base/flags.h"
 #include "base/logging.h"
 #include "cluster_utility.h"
+#include "facade/error.h"
+#include "server/cluster/cluster_defs.h"
```
BorysTheDev, line 16: please check that you really need them

src/server/cluster/incoming_slot_migration.cc
```diff
@@ -70,6 +76,20 @@ class ClusterShardMigration {
         break;
       }
 
+      auto oom_check = [&]() -> bool {
+        auto used_mem = used_mem_current.load(memory_order_relaxed);
+        if ((used_mem + tx_data->command.cmd_len) > max_memory_limit) {
```
BorysTheDev, line 81: is used_mem RSS? maybe we need max_memory_limit - 100MB for example, or how we can guarantee 100% success of this condition?

src/server/cluster/incoming_slot_migration.cc
```diff
@@ -70,6 +76,20 @@ class ClusterShardMigration {
         break;
       }
 
+      auto oom_check = [&]() -> bool {
```
BorysTheDev, line 79: try avoid & in lambdas, pls provide tx_data as a parameter

src/server/cluster/incoming_slot_migration.cc
```diff
@@ -79,6 +99,11 @@ class ClusterShardMigration {
           VLOG(1) << "Finalized flow " << source_shard_id_;
           return;
         }
+        if (oom_check()) {
+          VLOG(2) << "Flow finalization " << source_shard_id_
+                  << " canceled due memory limit reached";
+          return;
+        }
```
BorysTheDev, line 106: I tihnk you need OOM_CHECK only before ExecuteTx

src/server/cluster/incoming_slot_migration.h
```diff
@@ -50,10 +53,20 @@ class IncomingSlotMigration {
...
   }
 
+  // Switch to  FATAL state and store error message
+  void ReportFatalError(dfly::GenericError err) ABSL_LOCKS_EXCLUDED(state_mu_, error_mu_) {
+    errors_count_.fetch_add(1, std::memory_order_relaxed);
+    util::fb2::LockGuard lk_state(state_mu_);
+    util::fb2::LockGuard lk_error(error_mu_);
+    state_ = MigrationState::C_FATAL;
+    last_error_ = std::move(err);
+  }
```
BorysTheDev, line 63: I don't like this code. Let's think how to make it better. Maybe we can use state_mu_ for error too or merge this method with reportError

src/server/cluster/outgoing_slot_migration.cc
```diff
@@ -12,6 +12,10 @@
 #include "base/logging.h"
 #include "cluster_family.h"
 #include "cluster_utility.h"
+#include "facade/resp_expr.h"
+#include "server/cluster/cluster_defs.h"
+#include "server/cluster/incoming_slot_migration.h"
+#include "server/common.h"
```
BorysTheDev, line 18: please check what do you really need to include

src/server/cluster/outgoing_slot_migration.cc
```diff
@@ -355,6 +367,20 @@ bool OutgoingMigration::FinalizeMigration(long attempt) {
...
+    if (CheckRespFirstTypes({RespExpr::ERROR})) {
+      auto error = facade::ToSV(LastResponseArgs().front().GetBuf());
+      LOG(WARNING) << "Error response for " << cf_->MyID() << " : " << migration_info_.node_info.id
+                   << " attempt " << attempt << " msg: " << error;
+      auto next_state = MigrationState::C_ERROR;
+      // Check if there is OOM response from incoming slot migration
+      if (error == IncomingSlotMigration::kMigrationOOM) {
+        SetLastError(GenericError(IncomingSlotMigration::kMigrationOOM));
+        next_state = MigrationState::C_FATAL;
+      }
```
BorysTheDev, line 379: I think this code should be in the Finish()

src/server/cluster/incoming_slot_migration.cc
mkaruza: Yes, it is checks whatever is set with maxmemory. I'll modify so it uses max_memory - 0.1*max_memory as upper bound. That should be fine?

src/server/cluster/outgoing_slot_migration.cc
mkaruza: I can move it there.

src/server/cluster/incoming_slot_migration.h
```diff
@@ -69,12 +82,15 @@ class IncomingSlotMigration {
 
   void Pause(bool pause);
 
+  static constexpr char kMigrationOOM[] = "INCOMING_MIGRATION_OOM";
```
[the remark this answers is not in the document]
mkaruza: Why we would need it also there ?

src/server/cluster/cluster_family.cc
mkaruza: If `FLOW` fails with C_FATAL we'll call it, where would you put `migration->Stop` to handle it?

src/server/cluster/cluster_family.cc
mkaruza: Yes, probably left over of some idea that i worked on during development. Will remove.

src/server/cluster/incoming_slot_migration.cc
mkaruza: I'll double check this, I think i saw one problem with logic so i added this (but could be related to some other iteration while developing).

src/server/cluster/incoming_slot_migration.h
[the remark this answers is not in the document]
BorysTheDev: we can show it like the reason of fatal for outgoing migration too

src/server/cluster/incoming_slot_migration.cc
BorysTheDev: Let's discuss it with @adiholden or @romange

src/server/cluster/cluster_family.cc
BorysTheDev: Maybe we can move it to reportError or SetState, where we set the fatal status

src/server/cluster/incoming_slot_migration.h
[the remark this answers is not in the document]
mkaruza: Moved both of errors to cluster_defs.h

src/server/cluster/outgoing_slot_migration.cc
```diff
@@ -141,7 +142,11 @@ void OutgoingMigration::OnAllShards(
 void OutgoingMigration::Finish(GenericError error) {
   auto next_state = MigrationState::C_FINISHED;
   if (error) {
-    next_state = MigrationState::C_ERROR;
+    if (error.Format() == kIncomingMigrationOOM) {
```
BorysTheDev, line 145: I don't like string comparison, maybe we can use use some erroc code like [<system_error>](https://en.cppreference.com/w/cpp/header/system_error) `not_enough_memory	ENOMEM`

mkaruza: We are comparing response  error from incoming slot node which would be `INCOMING_MIGRATION_OOM`

mkaruza: Simplified handling of error response checking.

src/server/cluster/cluster_family.cc
```diff
@@ -925,7 +926,7 @@ void ClusterFamily::InitMigration(CmdArgList args, SinkReplyBuilder* builder) {
 
   if (!migration) {
     VLOG(1) << "Unrecognized incoming migration from " << source_id;
-    return builder->SendSimpleString(OutgoingMigration::kUnknownMigration);
+    return builder->SendSimpleString(kUnknownMigration);
```
BorysTheDev, line 929: I think it should be error not string, like below

src/server/cluster/cluster_family.cc
```diff
@@ -1033,15 +1039,19 @@ void ClusterFamily::DflyMigrateAck(CmdArgList args, SinkReplyBuilder* builder) {
...
   }
 
   auto migration = GetIncomingMigration(source_id);
   if (!migration)
     return builder->SendError(kIdNotFound);
 
   if (!migration->Join(attempt)) {
-    return builder->SendError("Join timeout happened");
+    if (migration->GetState() == MigrationState::C_FATAL) {
+      return builder->SendError(absl::StrCat("-", kIncomingMigrationOOM));
```
BorysTheDev, line 1051: why do we need "-", maybe we can make it as part of the kIncomingMigrationOOM

src/server/cluster/outgoing_slot_migration.cc
```diff
@@ -221,6 +221,14 @@ void OutgoingMigration::SyncFb() {
...
     }
 
+    // Break outgoing migration if INIT from incoming node responded with OOM. Usually this will
+    // happen on second iteration after first failed with OOM. Sending second INIT is required to
+    // cleanup slots on incoming slot migration node.
+    if (CheckRespSimpleError(kIncomingMigrationOOM)) {
+      ChangeState(MigrationState::C_FATAL);
+      break;
+    }
+
```
BorysTheDev, line 231: I think you can move it after in next if `if (!CheckRespIsSimpleReply("OK")) {`

mkaruza: `CheckRespIsSimpleReply` expects that response is STRING (`+`) while we are returning now ERROR (`-`)

src/server/cluster/cluster_family.cc
mkaruza: If `-` is not provided in error string response would be `-ERR INCOMING_MIGRATION_OOM` which will not be matched. If we use `-` as part of constant we would need to remove `-` prefix when matching error response.

src/server/cluster/cluster_family.cc
mkaruza: Part of code that was handling `kUnknownMigration` is using both SendError and SendSimpleString. This PR didn't change that but it moved this constant into different location. `kUnknownMigration` will be matched only if `SendSimpleString` is used.

src/server/cluster/outgoing_slot_migration.cc
BorysTheDev: I don't see any problem to return error as error

mkaruza: Moved.

adiholden:
@kostasrim Please review the changes in this PR related to the reply builder

src/facade/op_status.h
```diff
@@ -32,6 +33,7 @@ enum class OpStatus : uint16_t {
   MEMBER_NOTFOUND,
   INVALID_JSON_PATH,
   INVALID_JSON,
+  CUSTOM_ERROR = std::numeric_limits<uint16_t>::max()  // Custom status error
```
BorysTheDev, line 36: why max?

src/facade/reply_builder.cc
```diff
@@ -258,6 +259,7 @@ void MCReplyBuilder::SendLong(long val) {
 
 void MCReplyBuilder::SendError(string_view str, std::string_view type) {
   last_error_ = str;
+  last_error_status_ = OpStatus::CUSTOM_ERROR;
```
BorysTheDev, line 262: Consider do use something like ErrorReply to store error string and error status together

src/facade/reply_builder.h
```diff
@@ -120,8 +120,17 @@ class SinkReplyBuilder {
...
+    last_error_status_ = OpStatus::OK;
+    std::exchange(last_error_, {});
+  }
+
+  std::string_view GetLastError() {
+    return last_error_;
+  }
+
+  OpStatus GetLastErrorStatus() const {
+    return last_error_status_;
```
BorysTheDev, line 133: I think we need to use something like ErrorReply to unite status and error

src/facade/op_status.h
mkaruza: This supposed to be error that doesn't have status  associated with it. So if someone in future will add another status code it should be added before this one otherwise there will be error.

src/server/cluster/cluster_family.cc
BorysTheDev: you do here send kUnknownMigration as string but below you do send it as error

src/server/journal/serializer.cc
```diff
@@ -162,6 +162,7 @@ std::error_code JournalReader::ReadCommand(journal::ParsedEntry::CmdData* data)
 
   size_t cmd_size = 0;
   SET_OR_RETURN(ReadUInt<uint64_t>(), cmd_size);
+  data->cmd_len = cmd_size;
```
BorysTheDev, line 165: do we still need it?

mkaruza: No not really, we don't relay on command length anymore, but, IMO, this is still valuable information that could be useful.

src/facade/reply_builder.h
kostasrim: It's fine if we keep them separate. Besides, the `ErrorReply` is not very friendly with its constructor (it can be non-owning via string_view)

src/facade/reply_builder.cc
kostasrim, line 262: I don't understand why we need `CUSTOM_ERROR` and why we need the `last_error_status_` in `memcached reply builder` ? From my understanding we are dealing with cluster here so why do we need to change memcached reply builder ? Furthermore, can you plz explain the rationale behind what `CUSTOM_ERROR` means?

src/server/main_service.cc
```diff
@@ -1146,9 +1146,10 @@ void Service::DispatchCommand(ArgSlice args, SinkReplyBuilder* builder,
   DCHECK(!args.empty());
   DCHECK_NE(0u, shard_set->size()) << "Init was not called";
 
-  absl::Cleanup clear_last_error([builder]() { builder->ConsumeLastError(); });
   ServerState& etl = *ServerState::tlocal();
 
+  builder->ResetLastError();
```
kostasrim, line 1151: Why do we need this here ? What's wrong with cleaning it up via `absl::Cleanup` ? We call this here before we invoke the command, we call this again via `InvokeCmd`. It seems redundant

src/facade/reply_builder.cc
mkaruza: `CUSTOM_ERROR` is to have status of reported errors that doesn't have status enum code associated with it.

kostasrim:
> @kostasrim Please review the changes in this PR related to the reply builder

I synced with @mkaruza internally. We don't need the changes in the reply builder as they affect a broader part of the codebase. What is more is that the common denominator for all errors are `strings` and not `op status codes`. 

What we need here is to fetch out the error message on the caller side (of DispatchCommand) on a `one very specific code path` (when status code is OOM). This can be done easily by an extra `std::optional<bool>` within `ConnetionContext` and pretty much get rid of all of the changes in ReplyBuilder. If we have an error and if std::optional is emplaced and if the string is `kOOM` we set the boolean to true and that's how the caller knows. No changes to reply builder and no generic solution for a non generic problem.

If we ever need a `generic solution` that on error requires to return the actual error message to the caller, we just add an extra argument to DispatchCommand() and do it properly.

src/server/conn_context.h
```diff
@@ -295,6 +295,10 @@ class ConnectionContext : public facade::ConnectionContext {
     return conn_state.db_index;
   }
 
+  bool IsOOM() {
+    return std::exchange(is_oom_, false);
+  }
+
```
BorysTheDev, line 301: it looks like workaround, can we use Command context?

src/server/main_service.cc
```diff
@@ -1364,6 +1364,10 @@ bool Service::InvokeCmd(const CommandId* cid, CmdArgList tail_args,
   }
 
   if (std::string reason = builder->ConsumeLastError(); !reason.empty()) {
+    // Set flag if OOM reported
+    if (reason == kOutOfMemory) {
+      cmd_cntx.conn_cntx->is_oom_ = true;
```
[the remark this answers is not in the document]
mkaruza: `CommandContext` will be create in `DispatchCommand` so not available from incoming slot migration.

src/server/conn_context.h
mkaruza: Check previous response.

src/server/main_service.cc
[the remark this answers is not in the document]
BorysTheDev: Maybe we can do refactoring and propagate it?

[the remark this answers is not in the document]
BorysTheDev: @mkaruza Could you create a ticket to not forget regarding this refactoring

BorysTheDev (approved):

src/server/main_service.cc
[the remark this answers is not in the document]
adiholden: Guys, valid points on both sides.

I agree that is_oom_ is conceptually tied to the command rather than the connection, so placing it in CommandContext makes sense from a design perspective. However, since CommandContext isn’t available during the incoming slot migration, I support the idea of deferring this to a separate PR . I will follow up with Mario to understand the effort and if we want to do this If this is a big change and the proposed refactor is non-trivial .

PR #5001, opened 2025-04-25
fix(glob): Correctly match longer strings with '*' or '**' pattern

### LINKED ISSUE #4948: PSUBSCRIBE * wildcard does not work correctly when channel name is longer than 15 characters
**Describe the bug**
`PSUBSCRIBE *` does not receive all published messages when the channel name is longer than 15 characters in length.

See additional info for a special case we discovered aswell.

**To Reproduce**
1. Subscribe to all pub sub messages using `PSUBSCRIBE *`
2. Send a message via `PUBLISH` in a seperate redis client with a channel name longer than 15 characters
    For example `PUBLISH 1234567890123456 testvalue`
4. Inspect the subscribing redis clients output, there shouldn't be any messages

**Expected behavior**
It should behave like redis and receive all messages published even if the channel name is longer than 15 characters.

**Screenshots**
Screenshot from Redis Insight Pub/Sub tool, same behavior can be reproduced using redis-cli
Image

 - OS: [`wsl`, `k3s` on debain 12]
 - Kernel:
      - WSL: `Linux Luc-Main 5.15.167.4-microsoft-standard-WSL2 #1 SMP Tue Nov 5 00:21:55 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux`
      - k3s: `Linux *hostname redacted* 6.1.0-26-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.112-1 (2024-09-30) x86_64 GNU/Linux`
 - Containerized?: `Docker Compose` (Also happens in kubernetes with the operator)
 - Dragonfly Version: `v1.28.1`

**Reproducible Docker Compose Setup**
```
services:
  dragonfly:
    image: ghcr.io/dragonflydb/dragonfly:latest
    restart: unless-stopped
    ports:
      - 6379:6379
```

**Additional context**
Interesting part is if you for example subscribe to `1*` and send a message on channel `1234567890123456` it seems to work again no matter the length of the channel. So it seems only `*` as the channel pattern is the problem.

To clarify, we dont use this in production, this is for debugging purposes only on our end.

Feel free to let me know if you need any more information on this issue, keep up the great work! Thank you

mkaruza:
If string is longer than 16 chars we are using reflex library for matching. When used pattern is '*' or '**' we are 
removing trailing and leading star and have empty pattern as result. We should, in this edge case, set manually star regex.

Fixes #4948

### DIFF
--- a/src/core/glob_matcher.cc
+++ b/src/core/glob_matcher.cc
@@ -236,7 +236,11 @@ GlobMatcher::GlobMatcher(string_view pattern, bool case_sensitive)
     regex.push_back('i');
   }
   regex.push_back(')');
-  regex.append(Glob2Regex(pattern));
+  if (pattern.empty()) {
+    regex.append(Glob2Regex("*"));
+  } else {
+    regex.append(Glob2Regex(pattern));
+  }
   matcher_.pattern(regex);
 #elif USE_PCRE2
   string regex("(?s");  // dotall mode
--- a/src/core/dfly_core_test.cc
+++ b/src/core/dfly_core_test.cc
@@ -155,6 +155,7 @@ TEST_F(StringMatchTest, Basic) {
 
   // Wildcards
   EXPECT_EQ(MatchLen("*", "hello", 0), 1);
+  EXPECT_EQ(MatchLen("*", "1234567890123456", 0), 1);
   EXPECT_EQ(MatchLen("h*", "hello", 0), 1);
   EXPECT_EQ(MatchLen("h*", "abc", 0), 0);
   EXPECT_EQ(MatchLen("h*o", "hello", 0), 1);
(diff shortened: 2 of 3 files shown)


src/core/glob_matcher.cc
```diff
@@ -236,7 +236,12 @@ GlobMatcher::GlobMatcher(string_view pattern, bool case_sensitive)
     regex.push_back('i');
   }
   regex.push_back(')');
-  regex.append(Glob2Regex(pattern));
+  if (pattern.empty()) {
+    // Edge case, pattern is empty so it was '*' or '**'. Set regex manually.
+    regex.append(Glob2Regex("*"));
+  } else {
+    regex.append(Glob2Regex(pattern));
```
romange, line 243: can you please add a unit test to StringMatchTest inside dfly_core_test.cc?

romange (approved):

PR #5007, opened 2025-04-26
fix: Support for additional flags in expireat, pexpire and pexpireat commands

### LINKED ISSUE #4829: EXPIREAT, PEXPIRE, and PEXPIREAT commands do not support optional flags [NX | XX | GT | LT]
Relates to #842 and #2051

The intent of the above referenced issue and PR appears to have been to add support for expiry options to _all_ Redis commands that support them. However, these options were only implemented for `EXPIRE`:

```
> expire foo 2000 NX
0
> expireat foo 2000 NX
ERR wrong number of arguments for 'expireat' command
> pexpire foo 2000 NX
ERR wrong number of arguments for 'pexpire' command
> pexpireat foo 2000 NX
ERR wrong number of arguments for 'pexpireat' command
```

Environment:
```
redis_version: 7.4.0
dragonfly_version: df-v1.27.3
```

Additionally, the documentation for [`EXPIRE`](https://www.dragonflydb.io/docs/command-reference/generic/expire) was not updated to include those options.

H4R5H1T-007:
Fixes #4829

### DIFF
--- a/src/server/generic_family.cc
+++ b/src/server/generic_family.cc
@@ -1903,13 +1903,13 @@ void GenericFamily::Register(CommandRegistry* registry) {
       << CI{"TOUCH", CO::READONLY | CO::FAST, -2, 1, -1, acl::kTouch}.HFUNC(Exists)
       << CI{"EXPIRE", CO::WRITE | CO::FAST | CO::NO_AUTOJOURNAL, -3, 1, 1, acl::kExpire}.HFUNC(
              Expire)
-      << CI{"EXPIREAT", CO::WRITE | CO::FAST | CO::NO_AUTOJOURNAL, 3, 1, 1, acl::kExpireAt}.HFUNC(
+      << CI{"EXPIREAT", CO::WRITE | CO::FAST | CO::NO_AUTOJOURNAL, -3, 1, 1, acl::kExpireAt}.HFUNC(
              ExpireAt)
       << CI{"PERSIST", CO::WRITE | CO::FAST, 2, 1, 1, acl::kPersist}.HFUNC(Persist)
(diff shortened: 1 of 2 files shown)


src/server/generic_family_test.cc
```diff
@@ -4,6 +4,9 @@
 
 #include "server/generic_family.h"
 
+#include "absl/strings/str_format.h"
```
romange, line 7: you do not need to use a real clock for these tests. please take a look at AdvanceTime helper function. This is how we control time in our unit tests

src/server/generic_family.cc
```diff
@@ -1900,13 +1900,13 @@ void GenericFamily::Register(CommandRegistry* registry) {
       << CI{"TOUCH", CO::READONLY | CO::FAST, -2, 1, -1, acl::kTouch}.HFUNC(Exists)
       << CI{"EXPIRE", CO::WRITE | CO::FAST | CO::NO_AUTOJOURNAL, -3, 1, 1, acl::kExpire}.HFUNC(
              Expire)
-      << CI{"EXPIREAT", CO::WRITE | CO::FAST | CO::NO_AUTOJOURNAL, 3, 1, 1, acl::kExpireAt}.HFUNC(
+      << CI{"EXPIREAT", CO::WRITE | CO::FAST | CO::NO_AUTOJOURNAL, -3, 1, 1, acl::kExpireAt}.HFUNC(
```
romange, line 1903: good catch

H4R5H1T-007: Thanks

shusvr:
Original issue didn't mention HEXPIRE command, would it be possible to also expire options for it as well?

romange:
Lets separate - it's too different things. if you need for HEXPIRE - please open an issue and explain your use-case.

PR #5010, opened 2025-04-27
chore: StoredCmd to support both owned and external arguments

romange:
Before: StoredCmd always copied the backing buffer of the commands. this of course sub-optimal if the bucking buffer exists during the life-time of StoredCmd. This is exactly the case in `Service::DispatchManyCommands`.

This PR:
1. Adds support for both owned and non-owned arguments.
2. Improves the interfaces around StoredCmd and removes some code duplication.


kostasrim (approved):
LGTM

PR #5011, opened 2025-04-27
chore: clean ups around command squasher

romange:
1. Eliminate replies reverse call - will allow to unite replies and cmd vectors into one.
2. Introduce SwitchTxCmd function to avoid code duplication.

### DIFF
--- a/src/server/multi_command_squasher.cc
+++ b/src/server/multi_command_squasher.cc
@@ -151,8 +151,7 @@ bool MultiCommandSquasher::ExecuteStandalone(facade::RedisReplyBuilder* rb, Stor
   }
 
   auto* tx = cntx_->transaction;
-  tx->MultiSwitchCmd(cmd->Cid());
-  cntx_->cid = cmd->Cid();
+  cntx_->SwitchTxCmd(cmd->Cid());
 
   if (cmd->Cid()->IsTransactional())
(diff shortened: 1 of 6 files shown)


src/server/multi_command_squasher.cc
```diff
@@ -205,7 +203,6 @@ OpStatus MultiCommandSquasher::SquashedHopCb(EngineShard* es, RespVersion resp_v
     CheckConnStateClean(local_state);
   }
 
-  reverse(sinfo.replies.begin(), sinfo.replies.end());
```
kostasrim, line 208: Wasn't there a reason we reversed this ?

romange: of course there was. It was convenient to pull the reply later when we reconstructed the connection replies.
I've adjusted the code:
https://github.com/dragonflydb/dragonfly/pull/5011/files#diff-a781086bea40e29c13352594fead4fed8c2bb2f4529a8d3058a8a6dd41e264dcR259

kostasrim (approved):

src/server/multi_command_squasher.cc
kostasrim: :ship: :it:

BorysTheDev (approved):

PR #5012, opened 2025-04-27
chore: get rid of possible recursion when unwinding structured reply

romange:

### DIFF
--- a/src/facade/reply_capture.cc
+++ b/src/facade/reply_capture.cc
@@ -81,24 +81,24 @@ void CapturingReplyBuilder::SendDirect(Payload&& val) {
   }
 }
 
-void CapturingReplyBuilder::Capture(Payload val) {
+void CapturingReplyBuilder::Capture(Payload val, bool collapse_if_needed) {
   if (!stack_.empty()) {
-    stack_.top().first->arr.push_back(std::move(val));
-    stack_.top().second--;
+    auto& last = stack_.top();
+    last.first->arr.push_back(std::move(val));
+    if (collapse_if_needed && last.second-- == 1) {
+      CollapseFilledCollections();
+    }
   } else {
     DCHECK_EQ(current_.index(), 0u);
     current_ = std::move(val);
   }
-
-  // Check if we filled up a collection.
-  CollapseFilledCollections();
 }
 
 void CapturingReplyBuilder::CollapseFilledCollections() {
   while (!stack_.empty() && stack_.top().second == 0) {
     auto pl = std::move(stack_.top());
     stack_.pop();
-    Capture(std::move(pl.first));
+    Capture(std::move(pl.first), false);
   }
 }
 
--- a/src/facade/reply_capture.h
+++ b/src/facade/reply_capture.h
@@ -79,7 +79,7 @@ class CapturingReplyBuilder : public RedisReplyBuilder {
   void SendDirect(Payload&& val);
 
   // Capture value and store eiter in current topmost collection or as a standalone value.
-  void Capture(Payload val);
+  void Capture(Payload val, bool collapse_if_needed = true);
 
   // While topmost collection in stack is full, finalize it and add it as a regular value.
   void CollapseFilledCollections();


src/facade/reply_capture.cc
```diff
@@ -81,24 +81,24 @@ void CapturingReplyBuilder::SendDirect(Payload&& val) {
...
-  // Check if we filled up a collection.
-  CollapseFilledCollections();
 }
 
 void CapturingReplyBuilder::CollapseFilledCollections() {
   while (!stack_.empty() && stack_.top().second == 0) {
     auto pl = std::move(stack_.top());
     stack_.pop();
-    Capture(std::move(pl.first));
+    Capture(std::move(pl.first), false);
```
vyavdoshenko, line 101: Does it mean that we always have a 1-level nested collection?

romange: No, it can have several levels. My changes should not change functionality. Are you saying they do?

vyavdoshenko: I guess it changes the behavior, because `Capture(std::move(pl.first), false);` call never causes call `CollapseFilledCollections();` inside Capture(). But `CollapseFilledCollections();` was called unconditionally before.

romange:
By behavior change I meant if it is possible to reach a different state at
the end. My claim is that after any call to capture, we will end up in the
same state for both implementations. The implementation details are
irrelevant

Roman Gershman
CTO
---

src/facade/reply_capture.cc
```diff
@@ -81,24 +81,24 @@ void CapturingReplyBuilder::SendDirect(Payload&& val) {
...
 
-void CapturingReplyBuilder::Capture(Payload val) {
+void CapturingReplyBuilder::Capture(Payload val, bool collapse_if_needed) {
   if (!stack_.empty()) {
-    stack_.top().first->arr.push_back(std::move(val));
-    stack_.top().second--;
+    auto& last = stack_.top();
+    last.first->arr.push_back(std::move(val));
+    if (collapse_if_needed && last.second-- == 1) {
+      CollapseFilledCollections();
```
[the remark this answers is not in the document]
kostasrim: I am not sure I understand this. Imagine a stack of size 2. We enter the body here and call `CollapseFilledCollection()` which pops the top of stack_ and calls `Capture()`. Now `Capture()` calls again `CollapseFilledCollection` which pops the top of the stack_ and calls `Capture()` again. The sequence we get is `Capture -> Collapse -> Capture -> Collapse`. This recursion can blow out the stack of the fiber.  By passing `false` to `Capture()` within `CollapseFilledCollections` we enforce that we drain the `stack_` within the body of `CollapseFilledCollections()`. So now we would get something like: `Capture -> CollapseFilledCollections iteration 1 -> iteration 2 -> done`.

src/facade/reply_capture.cc
kostasrim: @vyavdoshenko `CollapseFilledCollections` body is executed only when `!stack_empty()`. The function applies only for the branch `if (!stack_.empty()) {` of `Capture()` above.

src/facade/reply_capture.cc
```diff
@@ -81,24 +81,24 @@ void CapturingReplyBuilder::SendDirect(Payload&& val) {
...
 }
 
-void CapturingReplyBuilder::Capture(Payload val) {
+void CapturingReplyBuilder::Capture(Payload val, bool collapse_if_needed) {
   if (!stack_.empty()) {
-    stack_.top().first->arr.push_back(std::move(val));
-    stack_.top().second--;
+    auto& last = stack_.top();
+    last.first->arr.push_back(std::move(val));
+    if (collapse_if_needed && last.second-- == 1) {
```
kostasrim, line 88: post increment and 1 :smile:

kostasrim (approved):

src/facade/reply_capture.cc
romange: the bug is here

PR #5013, opened 2025-04-28
fix: regtest failures

romange:
1. Fix a crash bug in RESETSTAT when number of shards is less than number of threads.
2. Tune regtests parameters for pipelining tests to pass.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -55,9 +55,6 @@ jobs:
           aws-access-key-id: ${{ secrets.AWS_S3_ACCESS_KEY }}
           aws-secret-access-key: ${{ secrets.AWS_S3_ACCESS_SECRET }}
           s3-bucket: ${{ secrets.S3_REGTEST_BUCKET }}
-          # Chain ternary oprator of the form (which can be nested)
-          # (expression == condition && <true expression> || <false expression>)
-          epoll: ${{ matrix.proactor == 'Epoll' && 'epoll' || 'iouring' }}
 
       - name: Upload logs on failure
         if: failure()
--- a/tests/dragonfly/connection_test.py
+++ b/tests/dragonfly/connection_test.py
@@ -525,7 +525,7 @@ async def test_keyspace_events_config_set(async_client: aioredis.Redis):
             await collect_expiring_events(pclient, keys)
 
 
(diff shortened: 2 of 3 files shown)


.github/workflows/regression-tests.yml
```diff
@@ -55,9 +55,6 @@ jobs:
           aws-access-key-id: ${{ secrets.AWS_S3_ACCESS_KEY }}
           aws-secret-access-key: ${{ secrets.AWS_S3_ACCESS_SECRET }}
           s3-bucket: ${{ secrets.S3_REGTEST_BUCKET }}
-          # Chain ternary oprator of the form (which can be nested)
-          # (expression == condition && <true expression> || <false expression>)
-          epoll: ${{ matrix.proactor == 'Epoll' && 'epoll' || 'iouring' }}
```
kostasrim, line 60: Why did you remove this ?

BorysTheDev (approved):

tests/dragonfly/connection_test.py
```diff
@@ -1121,12 +1122,13 @@ async def wait_for_stuck_on_send():
...
 
 # Test that the cache pipeline does not grow or shrink under constant pipeline load.
-@dfly_args({"proactor_threads": 1, "pipeline_squash": 9, "max_busy_read_usec": 1000})
+@dfly_args({"proactor_threads": 1, "pipeline_squash": 9, "max_busy_read_usec": 10000})
 async def test_pipeline_cache_only_async_squashed_dispatches(df_factory):
     server = df_factory.create()
     server.start()
 
     client = server.client()
+    await client.ping()  # Make sure the connection and the protocol were established
```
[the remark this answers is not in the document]
romange: do we call ping there?

.github/workflows/regression-tests.yml
romange: why it is needed if it's for iouring only?

kostasrim: Because we use `inputs.epoll` to print. It will be `empty` in your case. And also we do `if [[ "${{inputs.epoll}}" == 'epoll' ]];` and an unconditional `else` for `iouring`. If we change this, we will break it without noticing.

Not important though, I will sent a patch

PR #5019, opened 2025-04-28
chore: polishing. put basic usage info to stdout using one call to a void messing lines

vyavdoshenko:
Sometimes we can see the lines being messed with at the start. This change fixes that.

The example from Grafana log:
Screenshot 2025-04-28 at 16 17 35

### DIFF
--- a/src/server/dfly_main.cc
+++ b/src/server/dfly_main.cc
@@ -669,14 +669,15 @@ void PrintBasicUsageInfo() {
 
   for (const auto& dir : google::GetLoggingDirectories()) {
     const string_view maybe_slash = absl::EndsWith(dir, "/") ? "" : "/";
-    output += std::string(dir) + std::string(maybe_slash) + "dragonfly.*\n";
+    absl::StrAppend(&output, dir, maybe_slash, "dragonfly.*\n");
   }
 
-  output +=
-      "* For the available flags type dragonfly [--help | --helpfull]\n"
-      "* Documentation can be found at: https://www.dragonflydb.io/docs\n";
+  absl::StrAppend(&output,
+                  "* For the available flags type dragonfly [--help | --helpfull]\n"
+                  "* Documentation can be found at: https://www.dragonflydb.io/docs\n");
 
   std::cout << output;
+  std::cout.flush();
 }
 
 void ParseFlagsFromEnv() {


src/server/dfly_main.cc
```diff
@@ -649,30 +648,35 @@ void sigill_hdlr(int signo) {
...
+      "                     :##:                     \n"
+      "                     :@@:                     \n"
+      "                      @@                      \n"
+      "                      ..                      \n"
+      "* Logs will be written to the first available of the following paths:\n";
+
   for (const auto& dir : google::GetLoggingDirectories()) {
     const string_view maybe_slash = absl::EndsWith(dir, "/") ? "" : "/";
-    std::cout << dir << maybe_slash << "dragonfly.*\n";
+    output += std::string(dir) + std::string(maybe_slash) + "dragonfly.*\n";
```
BorysTheDev, line 672: absl::StrCat

src/server/dfly_main.cc
```diff
@@ -649,30 +648,35 @@ void sigill_hdlr(int signo) {
...
-  std::cout << "     =*+-:           ###*          .:-+*=     \n";
-  std::cout << "                     %@@%                     \n";
-  std::cout << "                     *@@*                     \n";
-  std::cout << "                     +@@=                     \n";
-  std::cout << "                     :##:                     \n";
-  std::cout << "                     :@@:                     \n";
-  std::cout << "                      @@                      \n";
-  std::cout << "                      ..                      \n";
-  std::cout << "* Logs will be written to the first available of the following paths:\n";
+  std::string output =
```
BorysTheDev, line 651: string_view

src/server/dfly_main.cc
kostasrim, line 672: There is also `absl::StrCat` or `absl::StrAppend`

src/server/dfly_main.cc
```diff
@@ -649,30 +648,35 @@ void sigill_hdlr(int signo) {
...
   }
-  std::cout << "* For the available flags type dragonfly [--help | --helpfull]\n";
-  std::cout << "* Documentation can be found at: https://www.dragonflydb.io/docs";
-  std::cout << endl;
+
+  output +=
+      "* For the available flags type dragonfly [--help | --helpfull]\n"
+      "* Documentation can be found at: https://www.dragonflydb.io/docs\n";
+
+  std::cout << output;
```
kostasrim, line 679:
```suggestion
  std::cout << output << std::endl;
```

kostasrim: better flush the output buffer

src/server/dfly_main.cc
vyavdoshenko: It changes below.

kostasrim: it's fine to use std::string. Who cares :shrug: Same for absl:: variants. Just commented them for completeness

kostasrim:
Can you plz elaborate:

> Sometimes we can see the lines being messed with at the start. This change fixes that.

src/server/dfly_main.cc
vyavdoshenko: I want to avoid more than one <<. This is the main idea of this change.

kostasrim: But if you don't `flush` the internal io buffer, how do you know it gets printed ? `cout` uses `buffered io` so unless you flush with `std::endl` you don't know if it actually reached the output/terminal

vyavdoshenko:
> Can you plz elaborate:

I updated the description with the provided example.

src/server/dfly_main.cc
vyavdoshenko: Fixed.

src/server/dfly_main.cc
vyavdoshenko: I added explicit flush. It solves the instant write. But the main idea was to avoid breaking lines and messing them up.

kostasrim (approved):

Issue #5025, opened 2025-04-29
JSON.ARRPOP command causes crash


vyavdoshenko:
The command:
```
JSON.ARRPOP test_resp3 FORMAT EXPAND $.a
```

Dragonfly log:
```
F20250429 17:43:12.547595 21000 cmd_arg_parser.cc:38] Check failed: !error_.has_value() Parsing error occured but not checked
*** Check failure stack trace: ***
    @     0x56b33b231367  google::LogMessage::Fail()
    @     0x56b33b2312ad  google::LogMessage::SendToLog()
    @     0x56b33b230a82  google::LogMessage::Flush()
    @     0x56b33b2348fa  google::LogMessageFatal::~LogMessageFatal()
    @     0x56b33ac87409  facade::CmdArgParser::~CmdArgParser()
    @     0x56b33a3da726  dfly::JsonFamily::ArrPop()
    @     0x56b33a288bdc  _ZN3fu27abi_4006detail10invocation6invokeIRKPFvN4absl12lts_202407224SpanIKSt17basic_string_viewIcSt11char_traitsIcEEEERKN4dfly14CommandContextEEJSC_SG_EEEDTclcl7forwardIT_Efp_Espcl7forwardIT0_Efp0_EEEOSL_DpOSM_
    @     0x56b33a28858e  fu2::abi_400::detail::type_erasure::invocation_table::function_trait<>::internal_invoker<>::invoke()
    @     0x56b33aa13489  _ZNK3fu27abi_4006detail12type_erasure6tables6vtableINS1_8propertyILb0ELb0EJKFvN4absl12lts_202407224SpanIKSt17basic_string_viewIcSt11char_traitsIcEEEERKN4dfly14CommandContextEEEEEE6invokeILm0EJPKNS2_13data_accessorERKmSE_SI_EEEDcDpOT0_
    @     0x56b33aa13522  _ZN3fu27abi_4006detail12type_erasure7erasureILb1ENS1_6configILb1ELb1ENS_16capacity_defaultEEENS1_8propertyILb0ELb0EJKFvN4absl12lts_202407224SpanIKSt17basic_string_viewIcSt11char_traitsIcEEEERKN4dfly14CommandContextEEEEEE6invokeILm0ERKSN_JSG_SK_EEEDcOT0_DpOT1_
    @     0x56b33aa135a0  fu2::abi_400::detail::type_erasure::invocation_table::operator_impl<>::operator()()
[...]

PR #5026, opened 2025-04-29
fix: JSON.ARRPOP command crash fixed

vyavdoshenko:
fixes: https://github.com/dragonflydb/dragonfly/issues/5025

### DIFF
--- a/src/server/json_family.cc
+++ b/src/server/json_family.cc
@@ -1794,6 +1794,10 @@ void JsonFamily::ArrPop(CmdArgList args, const CommandContext& cmd_cntx) {
   int index = parser.NextOrDefault<int>(-1);
 
   auto* builder = static_cast<RedisReplyBuilder*>(cmd_cntx.rb);
+  if (auto err = parser.Error(); err) {
+    return builder->SendError(err->MakeReply());
+  }
+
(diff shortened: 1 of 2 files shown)


src/server/json_family.cc
```diff
@@ -1794,6 +1794,10 @@ void JsonFamily::ArrPop(CmdArgList args, const CommandContext& cmd_cntx) {
   int index = parser.NextOrDefault<int>(-1);
 
   auto* builder = static_cast<RedisReplyBuilder*>(cmd_cntx.rb);
+  if (auto err = parser.Error(); err) {
```
kostasrim, line 1797: Interestingly enough redis/valkey accepts this but I could not find anything in the docs (`JSON.ARRPOP key [path [index]]`) so I am assuming is a bug on their end :shrug: 

Also the following queries will parse `JSON.ARRPOP doc pew 1 garbage garbage` -- in other words we don't check if exceed the maximum number of arguments. IMO this is not important.

vyavdoshenko: Anyway, we have to fix the crash.

kostasrim: Yes yes ofc!

PR #5032, opened 2025-04-30
feat: add copy cmd

BorysTheDev:
fixes: #3878

I've implemented COPY via the rename mechanism. To my mind in the future if we decide to implement the whole functionality of COPY it will be good idea to unite code of the MOVE COPY RENAME commands together

### DIFF
--- a/src/server/generic_family.cc
+++ b/src/server/generic_family.cc
@@ -300,12 +300,14 @@ OpStatus OpPersist(const OpArgs& op_args, string_view key);
 
 class Renamer {
  public:
-  Renamer(Transaction* t, std::string_view src_key, std::string_view dest_key, unsigned shard_count)
+  Renamer(Transaction* t, std::string_view src_key, std::string_view dest_key, unsigned shard_count,
+          bool do_copy = false)
       : transaction_(t),
         src_key_(src_key),
         dest_key_(dest_key),
         src_sid_(Shard(src_key, shard_count)),
-        dest_sid_(Shard(dest_key, shard_count)) {
+        dest_sid_(Shard(dest_key, shard_count)),
+        do_copy_(do_copy) {
   }
 
   ErrorReply Rename(bool destination_should_not_exist);
@@ -337,6 +339,7 @@ class Renamer {
 
   bool src_found_ = false;
   bool dest_found_ = false;
+  bool do_copy_ = false;
 
   SerializedValue serialized_value_;
 };
@@ -366,7 +369,7 @@ ErrorReply Renamer::Rename(bool destination_should_not_exist) {
 void Renamer::FetchData() {
   auto cb = [this](Transaction* t, EngineShard* shard) {
     auto args = t->GetShardArgs(shard->shard_id());
-    DCHECK_EQ(1u, args.Size());
+    DCHECK(1 == args.Size() || do_copy_);
 
     const ShardId shard_id = shard->shard_id();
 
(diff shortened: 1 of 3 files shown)


src/server/generic_family.cc
```diff
@@ -300,12 +300,14 @@ OpStatus OpPersist(const OpArgs& op_args, string_view key);
 
 class Renamer {
  public:
-  Renamer(Transaction* t, std::string_view src_key, std::string_view dest_key, unsigned shard_count)
+  Renamer(Transaction* t, std::string_view src_key, std::string_view dest_key, unsigned shard_count,
```
vyavdoshenko, line 303: I do not see the search index updating.
Now, RENAME removes the search index for the source file. Please add a simple test and verify the search index for both the source and the destination after the COPY operation is executed.

vyavdoshenko (changes requested):

src/server/generic_family.cc
BorysTheDev: Could you clarify what search index are you talking about

vyavdoshenko: You can execute FT.SEARCH before copy and after.
Please verify that the source document can be found before and after the copy. Also, the destination can't be found before and can be found after.

vyavdoshenko: Please take a look at the PR: https://github.com/dragonflydb/dragonfly/pull/4995
It fixes the particular issue, but it unconditionally removes the search index for the source file.
I guess you have to add the condition, is it a copy or a rename?
Also, don't forget to add the test to verify it.

kostasrim: @vyavdoshenko 

What you changed is the single shard optimization that we have in `Rename` command that @BorysTheDev is not using in his code. 

Normal rename (without) the optimization mentioned above always:

1. Serializes the key via dump command
2. Deserializes it again by loading it to the destination shard

Both of these flows handle the indexes transparently so there is no question of correctness here. We also have tests for this behaviour see SearchFamilyTest::DocsEditing and as this is the exact same flow with a different name there is no point to act any further.

src/server/generic_family.cc
```diff
@@ -1654,6 +1657,20 @@ void GenericFamily::RenameNx(CmdArgList args, const CommandContext& cmd_cntx) {
   }
 }
 
+void GenericFamily::Copy(CmdArgList args, const CommandContext& cmd_cntx) {
+  CmdArgParser parser(args);
+  auto [k1, k2] = parser.Next<std::string_view, std::string_view>();
+  bool replace = parser.Check("REPLACE");
```
kostasrim, line 1663: We don't support `[DB destination-db]` option right ? Also I guess, we parse `COPY k1 k2 NONSENSE` and allow it (including COPY k1 k2 nonsense more nonsense here)

src/server/generic_family_test.cc
```diff
@@ -939,4 +939,81 @@ TEST_F(GenericFamilyTest, Unlink) {
...
+  resp = Run({"COPY", "x", "b", "REPLACE"});
+  ASSERT_EQ(resp, "OK");
+
+  ASSERT_EQ(x_val, Run({"get", "x"}));
+  ASSERT_EQ(x_val, Run({"get", "b"}));
+  EXPECT_EQ(CheckedInt({"exists", "x", "b"}), 2);
+
+  const char* keys[2] = {"b", "x"};
+  auto ren_fb = pp_->at(0)->LaunchFiber([&] {
+    for (size_t i = 0; i < 200; ++i) {
```
kostasrim, line 967: Too much :rofl:

kostasrim (approved):
LGTM

src/server/generic_family.cc
BorysTheDev: yes we don't support [DB destination-db], but I forget to add check for nonsense

src/server/generic_family.cc
BorysTheDev: @vyavdoshenko I've checked everything works as @kostasrim said

vyavdoshenko (approved):
lgtm

romange (approved):

PR #5039, opened 2025-04-30
chore: bump up max_busy_read_usec in tests

romange:
cycle clock like most clocks is affected by thread context switching, and maybe this is what causes test failures. Try to increase the parameter to even further.

Also, add benchmarks for the clock as well as fix some compiler warnings around parser.cc

### DIFF
--- a/src/core/dfly_core_test.cc
+++ b/src/core/dfly_core_test.cc
@@ -223,6 +223,21 @@ static void BM_ParseDoubleAbsl(benchmark::State& state) {
 }
 BENCHMARK(BM_ParseDoubleAbsl);
 
+template <clockid_t cid> void BM_ClockType(benchmark::State& state) {
+  timespec ts;
+  while (state.KeepRunning()) {
+    DoNotOptimize(clock_gettime(cid, &ts));
+  }
+}
+BENCHMARK_TEMPLATE(BM_ClockType, CLOCK_REALTIME);
+BENCHMARK_TEMPLATE(BM_ClockType, CLOCK_REALTIME_COARSE);
+BENCHMARK_TEMPLATE(BM_ClockType, CLOCK_MONOTONIC);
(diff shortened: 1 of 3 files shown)


src/core/dfly_core_test.cc
```diff
@@ -223,6 +223,21 @@ static void BM_ParseDoubleAbsl(benchmark::State& state) {
...
+  timespec ts;
+  while (state.KeepRunning()) {
+    DoNotOptimize(clock_gettime(cid, &ts));
+  }
+}
+BENCHMARK_TEMPLATE(BM_ClockType, CLOCK_REALTIME);
+BENCHMARK_TEMPLATE(BM_ClockType, CLOCK_REALTIME_COARSE);
+BENCHMARK_TEMPLATE(BM_ClockType, CLOCK_MONOTONIC);
+BENCHMARK_TEMPLATE(BM_ClockType, CLOCK_MONOTONIC_COARSE);
+BENCHMARK_TEMPLATE(BM_ClockType, CLOCK_BOOTTIME);
```
kostasrim, line 236: @romange 

> On x86 systems with a stable and invariant TSC, the kernel typically prioritizes TSC as the clock source due to its low latency and high resolution. In such cases, CLOCK_BOOTTIME measurements indirectly use TSC.

Is that true ?

kostasrim (approved):

src/core/dfly_core_test.cc
romange: I do not know, I was mainly interested to check THREAD_TIME clock. it's 10 times slower than the regular high precision clock.

kostasrim: From perplexity:

> Tracks per-thread CPU time, requiring the kernel to sum scheduler-specific data (user/system time) for the thread. This involves accessing per-thread kernel structures, which may require locking or synchronization, adding overhead

Issue #5038, opened 2025-04-30
json.set causes to unable to allocate aligned OS memory directly & dragonfly hangs


vyavdoshenko:
The scenario:
```
redis-cli JSON.SET k $ '[{"a":1}]'
redis-cli JSON.SET k '$..*' '[{"a":1}]'
```

The dragonfly log:
```
I20250430 19:26:52.805543  5924 dfly_main.cc:752] Starting dragonfly df-dev-0000000
W20250430 19:26:52.805774  5924 dfly_main.cc:791] SWAP is enabled. Consider disabling it when running Dragonfly.
I20250430 19:26:52.805790  5924 dfly_main.cc:796] maxmemory has not been specified. Deciding myself....
I20250430 19:26:52.805795  5924 dfly_main.cc:805] Found 30.24GiB available memory. Setting maxmemory to 24.19GiB
I20250430 19:26:52.806740  5925 uring_proactor.cc:286] IORing with 1024 entries, allocated 98368 bytes, cq_entries is 2048
I20250430 19:26:52.807914  5924 proactor_pool.cc:149] Running 12 io threads
I20250430 19:26:52.818300  5924 server_family.cc:832] Host OS: Linux 6.8.0-57-generic x86_64 with 12 threads
I20250430 19:26:52.818665  5926 listener_interface.cc:101] sock[27] AcceptServer - listening on port 6379
mimalloc: warning: thread 0x770C1F48C640: unable to allocate aligned OS memory directly, fall back to over-allocation (size: 0x60400000 bytes, address: 0x770B8BC00000, alignment: 0x2000000, commit: 1)
mimalloc: warning: thread 0x770C1F48C640: unable to allocate aligned OS memory directly, fall back to over-allocation (size: 0xC0400000 bytes, address: 0x770AC9C00000, alignment: 0x2000000, commit: 1)
mimalloc: warning: thread 0x770C1F48C640: unable to allocate aligned OS memory directly, fall back to over-allocation (size: 0x180400000 bytes, address: 0x770947C00000, alignment: 0x2000000, commit: 1)
```

The result:
The dragonfly process freezes. It's impossible to execute another command. The process consumes more than 100% of the processor time.

vyavdoshenko:
Additional case found:
```
JSON.SET k $ '{"data":{"value":10,"subdata":{"value":20}}}'
JSON.SET k '$..value' '{"value":30}'
```

PR #5040, opened 2025-05-01
fix: json.set recursive processing crash

vyavdoshenko:
Fixes: https://github.com/dragonflydb/dragonfly/issues/5038

1. What's not working:
Using JSON.SET with recursive descent paths ($..) on nested JSON objects could lead to hangs or crashes. This occurred because the mutation logic could process the same node multiple times if it was discovered via different paths during the traversal.

2. How this PR fixes the issue:
This PR adjusts the iteration logic. When processing elements of an object or array targeted by a path segment, the code now correctly handles iterator advancement, especially after an element might have been modified or erased by the mutation callback. This ensures that each element within the target object/array is considered for mutation exactly once during that step, preventing infinite loops or incorrect processing that previously led to the crash/hang with $.. paths.

### DIFF
--- a/src/core/json/detail/jsoncons_dfs.cc
+++ b/src/core/json/detail/jsoncons_dfs.cc
@@ -8,8 +8,6 @@
 
 #include "core/json/detail/jsoncons_dfs.h"
 
-#include <absl/container/flat_hash_set.h>
-
 namespace dfly::json::detail {
 
 using namespace std;
@@ -84,9 +82,8 @@ Dfs Dfs::Mutate(absl::Span<const PathSegment> path, const MutateCallback& callba
     return dfs;
   }
 
-  // Use vector to maintain order and set to check uniqueness
+  // Use vector to maintain order
   std::vector<JsonType*> nodes_to_mutate;
-  absl::flat_hash_set<JsonType*> seen_nodes;
 
   using Item = detail::JsonconsDfsItem<false>;
   vector<Item> stack;
@@ -110,10 +107,7 @@ Dfs Dfs::Mutate(absl::Span<const PathSegment> path, const MutateCallback& callba
           stack.emplace_back(next, next_seg_id);
         } else {
           // Terminal step: collect node for mutation if not seen before
-          if (seen_nodes.find(next) == seen_nodes.end()) {
-            nodes_to_mutate.push_back(next);
-            seen_nodes.insert(next);
-          }
+          nodes_to_mutate.push_back(next);
         }
       }
     } else {
--- a/src/core/json/detail/jsoncons_dfs.cc
+++ b/src/core/json/detail/jsoncons_dfs.cc
@@ -8,6 +8,8 @@
 
 #include "core/json/detail/jsoncons_dfs.h"
 
+#include <absl/container/flat_hash_set.h>
+
 namespace dfly::json::detail {
 
 using namespace std;
@@ -82,6 +84,10 @@ Dfs Dfs::Mutate(absl::Span<const PathSegment> path, const MutateCallback& callba
     return dfs;
   }
 
+  // Use vector to maintain order and set to check uniqueness
+  std::vector<JsonType*> nodes_to_mutate;
+  absl::flat_hash_set<JsonType*> seen_nodes;
+
   using Item = detail::JsonconsDfsItem<false>;
   vector<Item> stack;
   stack.emplace_back(json);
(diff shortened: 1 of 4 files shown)


romange:
Please add explanation to the PR description. In general, for any non-trivial PRs, please state at least:
1. what's not working
2. how this PR fixes the problem.

src/server/json_family_test.cc
```diff
@@ -3087,4 +3087,29 @@ TEST_F(JsonFamilyTest, DepthLimitExceeded) {
   ASSERT_THAT(resp, ErrArg("ERR failed to parse JSON"));
 }
 
+TEST_F(JsonFamilyTest, RecursiveDescentMutation) {
```
romange, line 3090: Once you identified the problem you should test what is the minimum possible unit test that can help reproducing the problem. It's fine to add tests here but it's necessary to add tests to `src/core/json/jsonpath_test.cc` where you can reproduce it precisely and be able to iterate quickly.

src/core/json/detail/jsoncons_dfs.cc
```diff
@@ -8,6 +8,8 @@
 
 #include "core/json/detail/jsoncons_dfs.h"
 
+#include <set>
```
romange, line 11: we use `#include <absl/container/flat_hash_set.h>` as drop in replacement

romange (commented):
good work simplifying the fix!

src/server/json_family_test.cc
vyavdoshenko: Fixed

src/core/json/detail/jsoncons_dfs.h
```diff
@@ -43,6 +43,16 @@ template <bool IsConst> class JsonconsDfsItem {
     return depth_state_.second;
   }
 
+  // Added getter for the object pointer
+  Ptr obj_ptr() const {
+    return depth_state_.first;
+  }
+
+  // Added getter for segment_step_
```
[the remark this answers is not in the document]
vyavdoshenko: Fixed

src/core/json/detail/jsoncons_dfs.cc
vyavdoshenko: Fixed

src/core/json/detail/jsoncons_dfs.cc
```diff
@@ -103,14 +109,35 @@ Dfs Dfs::Mutate(absl::Span<const PathSegment> path, const MutateCallback& callba
         if (next_seg_id + 1 < path.size()) {
           stack.emplace_back(next, next_seg_id);
         } else {
-          dfs.MutateStep(path[next_seg_id], callback, next);
+          // Terminal step: collect node for mutation if not seen before
+          if (seen_nodes.find(next) == seen_nodes.end()) {
```
romange, line 113: I am curious actually. How come we can meet the same node twice? is `seen_nodes` needed at all?

vyavdoshenko: Yes, it's possible. I don't know how it can be done using JSON itself. But using modifications, we can get different paths to the same object. Also, it can be possible using a key and/or recursive replacement.

romange: Fair enough. Do we have a json path test that reproduces this edge case, i.e. would check fail if we add 
`DCHECK(seen_nodes.find(next) == seen_nodes.end())` ?

romange: Also why do we need both vector and a set for this? why can not we iterate over set below?

romange: Could you add a test that proves that `seen_nodes` are required?

src/core/json/detail/jsoncons_dfs.cc
```diff
@@ -82,6 +84,10 @@ Dfs Dfs::Mutate(absl::Span<const PathSegment> path, const MutateCallback& callba
     return dfs;
   }
 
+  // Use vector to maintain order and set to check uniqueness
```
romange, line 87: I missed this comment

src/core/json/detail/jsoncons_dfs.cc
vyavdoshenko: I tried to verify it using tests, but no matter how hard I tried, I couldn’t get it to work — and there’s an explanation for that.
Jsoncons doesn’t reuse the same object; instead, it creates a new one each time it’s needed.
So, my assumption was incorrect, and the set was unnecessary.

romange (approved):

Issue #5035, opened 2025-04-30
Reduce number of cluster logs


BorysTheDev:
Got DFLYCLUSTER command  really floods the logs in production.
Can we maybe push it into the function that handles cluster config change and log after the equality check?

romange:
actually we already have ` cluster_family.cc:571] Setting new cluster config` info log. Let's take it offline maybe it's not needed

BorysTheDev:
@romange  The problem with logs, I think because vlog is ON, so we LOG every cluster command

romange:
yeah, you are right, we will decrease the verbosity.

PR #5044, opened 2025-05-02
refactor: reduce number of cluster logs

BorysTheDev:
fixes: #5035

### DIFF
--- a/src/server/cluster/cluster_family.cc
+++ b/src/server/cluster/cluster_family.cc
@@ -462,12 +462,6 @@ void ClusterFamily::DflyCluster(CmdArgList args, const CommandContext& cmd_cntx)
     return builder->SendError("Cluster is disabled. Use --cluster_mode=yes to enable.");
   }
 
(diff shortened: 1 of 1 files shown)


src/server/cluster/cluster_family.cc
```diff
@@ -462,12 +462,6 @@ void ClusterFamily::DflyCluster(CmdArgList args, const CommandContext& cmd_cntx)
     return builder->SendError("Cluster is disabled. Use --cluster_mode=yes to enable.");
   }
 
-  if (cntx->conn()) {
-    VLOG(2) << "Got DFLYCLUSTER command (" << cntx->conn()->GetClientId() << "): " << args;
```
romange, line 466: we can just reduce the verbosity level in prod

BorysTheDev: To my mind this vlog is extra anycase because we have log info for config

adiholden (approved):

PR #5045, opened 2025-05-02
chore: add dfly_bench to release workflow

kostasrim:
We would like to have `dfly_bench` binaries uploaded on each release. This PR adds that. 

Furthermore, `daily-builds` https://github.com/dragonflydb/dragonfly/actions/runs/14789773294 fails on `macos` because some of the clocks are not defined. Remove these clocks from macos build.

* upload dfly_bench binaries on every release
* fix macos failing build


tools/release.sh
```diff
@@ -17,6 +17,8 @@ pwd
 
 make release
 
+make dfly_bench
```
kostasrim, line 20: `make release` from `cpack` does not build `dfly_bench` directly. Maybe there is a better way other than invoking make directly :shrug:

kostasrim:
@romange Shall I add a special workflow (which we can trigger manually -- without cron) that builds `dfly_bench` only ? That way we won't need to wait for a new release to consume `dfly_bench` when we add the tests for cluster.

romange:
Nice! I do not think we need a special release for that. We can upload binaries to a public bucket for now and switch to github later

romange (dismissed):

kostasrim:
https://github.com/dragonflydb/dragonfly/actions/runs/14795993916

tools/release.sh
kostasrim: :scream: it was a `makefile` rule :facepalm: :facepalm:

romange (approved):

Issue #5047, opened 2025-05-03
Dragonfly crashes on long sequence as a command


vyavdoshenko:
The dragonfly log:
```
F20250503 17:14:15.282817  1535 reply_builder.cc:148] Check failed: expected_buffer_cap <= kMaxBufferSize (27261 vs. 8192) 
*** Check failure stack trace: ***
    @     0x638ec6909f17  google::LogMessage::Fail()
    @     0x638ec6909e5d  google::LogMessage::SendToLog()
    @     0x638ec6909632  google::LogMessage::Flush()
    @     0x638ec690d4aa  google::LogMessageFatal::~LogMessageFatal()
    @     0x638ec66c8ef3  facade::SinkReplyBuilder::Flush()
    @     0x638ec66d2e53  facade::SinkReplyBuilder::WritePieces<>()
    @     0x638ec66caeda  facade::RedisReplyBuilderBase::SendError()
    @     0x638ec66c8b31  facade::SinkReplyBuilder::SendError()
    @     0x638ec5f63a54  dfly::Service::DispatchCommand()
    @     0x638ec6699942  _ZZN6facade10Connection10ParseRedisEjENKUlvE_clEv
    @     0x638ec66a4410  _ZSt13__invoke_implIvRKZN6facade10Connection10ParseRedisEjEUlvE_JEET_St14__invoke_otherOT0_DpOT1_
    @     0x638ec66a3b35  _ZSt8__invokeIRKZN6facade10Connection10ParseRedisEjEUlvE_JEENSt15__invoke_resultIT_JDpT0_EE4typeEOS6_DpOS7_
    @     0x638ec66a3318  _ZSt6invokeIRKZN6facade10Connection10ParseRedisEjEUlvE_JEENSt13invoke_resultIT_JDpT0_EE4typeEOS6_DpOS7_
    @     0x638ec66a2856  _ZN4absl12lts_2024072219functional_internal12InvokeObjectIZN6facade10Connection10ParseRedisEjEUlvE_vJEEET0_NS1_7VoidPtrEDpNS1_8ForwardTIT1_E4typeE
    @     0x638ec66a7a2b  absl::lts_20240722::FunctionRef<>::operator()()
    @     0x638ec6699791  facade::Connection::DispatchSingle()
    @     0x638ec6699ebf  facade::Connection::ParseRedis()
    @     0x638ec669b91d  facade::Connection::IoLoop()
[...]

PR #5048, opened 2025-05-03
fix: send response error for long sequence instead of command crash

vyavdoshenko:
fixes: https://github.com/dragonflydb/dragonfly/issues/5047

### DIFF
--- a/src/facade/reply_builder.cc
+++ b/src/facade/reply_builder.cc
@@ -146,6 +146,7 @@ void SinkReplyBuilder::Flush(size_t expected_buffer_cap) {
   guaranteed_pieces_ = 0;
 
   DCHECK_LE(expected_buffer_cap, kMaxBufferSize);  // big strings should be enqueued as iovecs
+
   if (expected_buffer_cap > buffer_.Capacity())
     buffer_.Reserve(expected_buffer_cap);
 }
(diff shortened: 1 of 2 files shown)


src/facade/reply_builder.cc
```diff
@@ -145,7 +145,12 @@ void SinkReplyBuilder::Flush(size_t expected_buffer_cap) {
   vecs_.clear();
   guaranteed_pieces_ = 0;
 
-  DCHECK_LE(expected_buffer_cap, kMaxBufferSize);  // big strings should be enqueued as iovecs
+  if (expected_buffer_cap > kMaxBufferSize) {
```
romange, line 148: Good catch but the fix is incorrect. 
It's our bug that break the internal invariant, and we do not fix bugs by fixing symptoms.

The bug is that we assume in `RedisReplyBuilderBase::SendError` that str is fairly small and allow copying it via 
WritePieces. 
Instead, we should check the str size and write differrently for str sizes that are longer than K. 
See `RedisReplyBuilderBase::SendVerbatimString` for how it's done there

vyavdoshenko: Fixed

romange (dismissed):

romange (approved):

PR #5049, opened 2025-05-04
chore(json_family): Clean up code. FIRST PR

BagritsevichStepan:
First PR before fixing memory tracking issues for json

1. Rename Evaluate to ReadOnly function for the `WrappedJsonPath`
~2. Remove const ref for the mutate operations on the `WrappedJsonPath` -> Change it to pointer~
3. Clean up


src/server/detail/wrapped_json_path.h
```diff
@@ -248,38 +232,23 @@ class WrappedJsonPath {
...
-  bool RefersToRootElement() const {
-    auto path = path_.view();
-    return path.empty() || path == kV1PathRootElement || path == kV2PathRootElement;
-  }
+  bool RefersToRootElement() const;
 
-  bool HoldsJsonPath() const {
-    return std::holds_alternative<json::Path>(parsed_path_);
-  }
+  // Returns true if this is first implementation of json path
```
romange, line 239: "first" is not clear here

src/server/detail/wrapped_json_path.h
```diff
@@ -248,38 +232,23 @@ class WrappedJsonPath {
...
+  bool HoldsJsonPath() const;
 
-  const json::Path& AsJsonPath() const {
-    return std::get<json::Path>(parsed_path_);
-  }
+  // Main implementation of json path
```
romange, line 242: "main" and "alternative" are not clear here.  jsoncons and internal are better terms

src/server/detail/wrapped_json_path.h
```diff
@@ -65,35 +65,23 @@ struct CallbackResultOptions {
   enum class SavingOrder { kSaveFirst, kSaveLast };
   enum class OnEmpty { kSendNil, kSendWrongType };
 
-  // Default options for WrappedJsonPath::Evaluate
-  static CallbackResultOptions DefaultEvaluateOptions() {
-    return CallbackResultOptions{OnEmpty::kSendNil};
-  }
+  // Default options for WrappedJsonPath::ExecuteReadOnlyCallback
+  static CallbackResultOptions DefaultReadOnlyOptions();
```
romange, line 69: Please do not use function override for this - use default arguments, i.e. combine both functions into 
`DefaultReadOnlyOptions(SavingOrder saving_order = SavingOrder::kSaveLast)`

src/server/detail/wrapped_json_path.h
```diff
@@ -65,35 +65,23 @@ struct CallbackResultOptions {
...
 
   std::optional<JsonPathType> path_type;
   SavingOrder saving_order{SavingOrder::kSaveLast};
   OnEmpty on_empty;
 
  private:
-  CallbackResultOptions(SavingOrder saving_order_, OnEmpty on_empty_)
-      : saving_order(saving_order_), on_empty(on_empty_) {
-  }
+  CallbackResultOptions(SavingOrder saving_order_, OnEmpty on_empty_);
```
romange, line 84: it is weird that you have so many public constructors that allow setting everything and suddenly this private one that is subset of a public c'tor. Moreover it's a struct with public fields. Seem like not needed at all.

src/server/detail/wrapped_json_path.h
```diff
@@ -164,21 +152,18 @@ class WrappedJsonPath {
   static constexpr std::string_view kV1PathRootElement = ".";
   static constexpr std::string_view kV2PathRootElement = "$";
 
-  WrappedJsonPath(json::Path json_path, StringOrView path, JsonPathType path_type)
-      : parsed_path_(std::move(json_path)), path_(std::move(path)), path_type_(path_type) {
```
romange, line 168: what is the reason for moving it to cpp file for trivial initialisations ?

src/server/detail/wrapped_json_path.cc
```diff
@@ -0,0 +1,73 @@
...
+WrappedJsonPath::WrappedJsonPath(json::Path json_path, StringOrView path, JsonPathType path_type)
+    : parsed_path_(std::move(json_path)), path_(std::move(path)), path_type_(path_type) {
+}
+
+WrappedJsonPath::WrappedJsonPath(JsonExpression expression, StringOrView path,
+                                 JsonPathType path_type)
+    : parsed_path_(std::move(expression)), path_(std::move(path)), path_type_(path_type) {
+}
+
+bool WrappedJsonPath::IsLegacyModePath() const {
```
romange, line 41: I do not understand why moving one liners to cc, what's the reason you decided to do it?

src/server/detail/wrapped_json_path.h
BagritsevichStepan: I think you mentioned in another PR that we should avoid default arguments and use overloads instead.

src/server/detail/wrapped_json_path.cc
BagritsevichStepan: I think the .h file became too large and hard to read, so I decided to split it up for clarity.
If you still believe the inline definitions should be kept, I can move the implementations to the bottom of the .h file instead

src/server/detail/wrapped_json_path.h
BagritsevichStepan: Answered in the next comment

src/server/detail/wrapped_json_path.cc
romange: i prefer keeping them in the h file. outside of the class definition is fine.

src/server/detail/wrapped_json_path.h
romange: True, it's not black and white, see https://google.github.io/styleguide/cppguide.html#Default_Arguments 
for discussion. I do not remember the specific case but if function does something different based on the argument - then function overload is the way to go, or use a different function name. But if it's the same code but just the value is differrent  (like with c'tors) then default arguments is the way to go.

src/server/detail/wrapped_json_path.h
BagritsevichStepan: Fixed

src/server/detail/wrapped_json_path.h
BagritsevichStepan: Ok

src/server/detail/wrapped_json_path.h
```diff
@@ -105,186 +69,260 @@ template <typename T> class JsonCallbackResult {
...
 
- private:
-  CallbackResultOptions InitializePathType(CallbackResultOptions options) const {
-    if (!options.path_type) {
-      options.path_type = path_type_;
+  return read_result;
+}
+
+template <typename T>
+OpResult<JsonCallbackResult<std::optional<T>>> WrappedJsonPath::ExecuteMutateCallback(
```
romange, line 245: curious, why it is non-const? do we change WrappedJsonPath state?

romange (dismissed):

src/server/detail/wrapped_json_path.h
BagritsevichStepan: Fill fix

romange: I do not understand :)
it does not have a const modifier and you now pass the path by mutable pointer instead of const ref
```
template <typename T>
OpResult<JsonCallbackResult<std::optional<T>>> WrappedJsonPath::ExecuteMutateCallback(
    JsonType* json_entry, JsonPathMutateCallback<T> cb, CallbackResultOptions options) {
```

BagritsevichStepan: Yes we change parsed_path_ here.
I want to mark as non const to show that we need memory tracker for this. The same I did in stream_family functions.

romange: for example here:
https://github.com/dragonflydb/dragonfly/pull/5049/files#diff-0799c8810dbf8758f9da40162075f38bb51639f806be28a06d9f669a6ebb0f76L627

BagritsevichStepan: Ah I see

BagritsevichStepan: I did a failure

BagritsevichStepan: Will fix

romange: @BagritsevichStepan  where do we change it? can you add a comment in the appropriate place? it's not trivial to see.

BagritsevichStepan: We don't change it, I missed something. Added const modifier

romange (approved):

Issue #5055, opened 2025-05-05
JSON value is not updated in the indexes after the JSON.DEL command


BagritsevichStepan:
The JSON value is not updated in the indexes after the `JSON.DEL` command.

DragonflyDb:
```
127.0.0.1:6379> json.set j1 $ '{"text":"some text"}'                                                                                                                                                        
OK
127.0.0.1:6379> ft.create index on json schema $.text AS text TEXT SORTABLE
OK                                                                                                                                                                                                       
127.0.0.1:6379> json.del j1 $.text                                                                                                                                                                          
(integer) 1                                                                                                                                                                                                 
127.0.0.1:6379> ft.aggregate index "*" GROUPBY 1 @text SORTBY 1 @text                                                                                                                                       
1) (integer) 1                                                                                                                                                                                              
2) 1) "text"                                                                                                                                                                                                
   2) "some text"
```

Expected behavior from Redis:
```
127.0.0.1:6379> json.set j1 $ '{"text":"some text"}'
OK
127.0.0.1:6379> ft.create index on json schema $.text AS text TEXT SORTABLE
OK
127.0.0.1:6379> json.del j1 $.text
(integer) 1
127.0.0.1:6379> ft.aggregate index "*" GROUPBY 1 @text
1) (integer) 1
2) 1) "text"
   2) (nil)
```

Issue #5054, opened 2025-05-05
JSON value is removed from the index after a failed JSON.SET command


BagritsevichStepan:
The JSON value is removed from the index after a failed JSON.SET command.

DragonflyDb:
```
127.0.0.1:6379> json.set j1 $ '{"text":"some text"}'
OK
127.0.0.1:6379> ft.create index on json schema $.text AS text TEXT SORTABLE
OK
127.0.0.1:6379> json.set j1 $ '{"asd}'
(error) ERR failed to parse JSON
127.0.0.1:6379> ft.aggregate index "*" GROUPBY 1 @text
1) (integer) 0
```

Expected behaviour from Redis:
```
127.0.0.1:6379> json.set j1 $ '{"text":"some text"}'
OK
127.0.0.1:6379> ft.create index on json schema $.text AS text TEXT SORTABLE
OK
127.0.0.1:6379> json.set j1 $ '{"asd}'
(error) EOF while parsing a string at line 1 column 6
127.0.0.1:6379> ft.aggregate index "*" GROUPBY 1 @text
1) (integer) 1
2) 1) "text"
   2) "some text"
```

PR #5056, opened 2025-05-05
fix(json_family): Fix json memory tracking issues. SECOND PR

### LINKED ISSUE #4926: Memory tracking tests for JSON and STREAMs
We need to add memory tracking tests for JSON and STREAMs. During the read commands we should assert that no memory were allocated.

- [ ] Add tests for JSON
         - Simple tests are added by #5056
- [ ] Add tests for STREAM

BagritsevichStepan:
fixes a lot of json memory tracking issues, #5055 , #5054 .
Also, partially fixes #4926

1. Memory tracking bug for the JSON.SET command: We didn't reset the result of `ShardJsonFromString`, which parses the json_str into the jsoncons's json and returns optional of this. The bug was that `std::optional` still holds the value even after `std::move`. So we overestimate the memory usage.
2. Bug in memory tracker when we overestimate the memory usage because of this:
```cpp
if (is_op_set) {
      diff += static_cast<int64_t>(mi_usable_size(pv.GetJson()));
}
```
3. Bug in partial `JSON.SET` when we called `ShardJsonFromString` before starting the memory tracking
4. Bug in JsonMutateOperation, where we called `op_args.shard->search_indices()->RemoveDoc` after starting the memory tracking. `RemoveDoc` has a static cache that may allocate some memory, but this memory does not belong to the JSON object itself. As a result, memory usage will be overestimated in this case.
5. A lot of bugs in `JSON.DEL` command - we didn't update the indexes and track the memory usage at all. One of the bugs is good described in #5055
6. Bug in indexes for the `JSON.SET` command: #5054

To fix this, I introduced `json_family_memory_test` and changed `JsonMemTracker` to `JsonAutoUpdater`, which eliminates most of the bugs described above.
It also ensures that RemoveDoc is called before memory tracking starts, and that AddDoc is called after SetJsonSize and also it will be called even if an error occurs.
Therefore, whenever we perform an update operation on a JSON value, this class must be used.

### DIFF
--- a/src/server/json_family.cc
+++ b/src/server/json_family.cc
@@ -54,32 +54,78 @@ using CI = CommandId;
 
 namespace {
 
-class JsonMemTracker {
+/* Helper class which must be initialized before any mutate operations on json.
+  It will track the memory usage of the json object and update the size in the CompactObj.
+  It also contains indexes updates, post update operations on the iterator. */
+class JsonAutoUpdater {
  public:
-  JsonMemTracker() {
-    start_size_ = static_cast<MiMemoryResource*>(CompactObj::memory_resource())->used();
+  JsonAutoUpdater(const OpArgs& op_args, string_view key, DbSlice::ItAndUpdater it,
+                  bool update_on_delete = false)
+      : op_args_(op_args), key_(key), it_(std::move(it)), update_on_delete_(update_on_delete) {
+    op_args.shard->search_indices()->RemoveDoc(key, op_args.db_cntx, it.it->second);
+
+    /* We need to initialize start memory usage after RemoveDoc because internally RemoveDoc has
+    static cache that can allocate/deallocate memory. Because of this, we will
+    overestimate/underestimate memory usage for json object. */
+    start_size_ = GetMemoryUsage();
   }
 
-  void SetJsonSize(PrimeValue& pv, bool is_op_set) {
-    const size_t current = static_cast<MiMemoryResource*>(CompactObj::memory_resource())->used();
+  JsonAutoUpdater(const JsonAutoUpdater&) = delete;
+  JsonAutoUpdater& operator=(const JsonAutoUpdater&) = delete;
+
+  JsonAutoUpdater(JsonAutoUpdater&&) = default;
+  JsonAutoUpdater& operator=(JsonAutoUpdater&&) = delete;
+
+  void SetJsonSize() {
+    set_size_was_called_ = true;
+
+    const size_t current = GetMemoryUsage();
     int64_t diff = static_cast<int64_t>(current) - static_cast<int64_t>(start_size_);
-    // If the diff is 0 it means the object use the same memory as before. No action needed.
-    if (diff == 0) {
-      return;
-    }
-    // If op_set_ it means we JSON.SET or JSON.MSET was called. This is a blind update,
-    // and because the operation sets the size to 0 we also need to include the size of
-    // the pointer.
-    if (is_op_set) {
-      diff += static_cast<int64_t>(mi_usable_size(pv.GetJson()));
-    }
-    pv.SetJsonSize(diff);
+
+    GetPrimeValue().SetJsonSize(diff);
+
     // Under any flow we must not end up with this special value.
-    DCHECK(pv.MallocUsed() != 0);
+    DCHECK(GetPrimeValue().MallocUsed() != 0);
+  }
+
+  ~JsonAutoUpdater() {
+    if (update_on_delete_ && !set_size_was_called_) {
+      SetJsonSize();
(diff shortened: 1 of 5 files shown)


src/server/json_family.cc
```diff
@@ -318,28 +364,63 @@ bool JsonAreEquals(const JsonType& lhs, const JsonType& rhs) {
   }
 }
 
+/* Converts a JSONPath to a JSONPointer.
+   E.g. $[a][b][0] -> /a/b/0.
+   V1 JSONPath is not supported. */
+std::optional<std::string> ConvertJsonPathToJsonPointer(string_view json_path) {
```
BagritsevichStepan, line 370: This was just moved

src/server/json_family_memory_test.cc
```diff
@@ -0,0 +1,156 @@
...
+  resp = Run({"JSON.DEL", "j1", "$.a"});
+  EXPECT_THAT(resp, IntArg(1));
+  resp = Run({"JSON.GET", "j1"});
+  EXPECT_EQ(resp, json_after_del);
+  resp = Run({"JSON._MEMUSAGE", "j1"});
+
+  /* We still expect the initial size here, because after deletion we do not call shrink_to_fit on
+     the JSON object. As a result, the memory will not be deallocated. Check
+     JsonFamilyMemoryTest::JsonConsDelTest for example. */
+  EXPECT_THAT(resp, IntArg(start_size));
```
BagritsevichStepan, line 146: Another issue here is that jsoncons does not free memory after calling erase; memory is only deallocated after calling shrink_to_fit.
As a result, JSON.DEL does not update the memory usage of the JSON object.

@romange As a potential solution, we could call shrink_to_fit when the JSON size becomes at least twice smaller than the allocated memory.

src/server/json_family.cc
```diff
@@ -2060,6 +2111,8 @@ void JsonFamily::Register(CommandRegistry* registry) {
   *registry << CI{"JSON.ARRAPPEND", CO::WRITE | CO::DENYOOM | CO::FAST, -4, 1, 1, acl::JSON}.HFUNC(
       ArrAppend);
   *registry << CI{"JSON.ARRINDEX", CO::READONLY | CO::FAST, -4, 1, 1, acl::JSON}.HFUNC(ArrIndex);
+  *registry << CI{"JSON._MEMUSAGE", CO::READONLY | CO::FAST, 2, 1, 1, acl::JSON}.HFUNC(
```
romange, line 2114: you do not need it, we already have "memory usage key" that should work for json as well.

src/server/json_family.cc
```diff
@@ -54,32 +54,78 @@ using CI = CommandId;
...
     // Under any flow we must not end up with this special value.
-    DCHECK(pv.MallocUsed() != 0);
+    DCHECK(GetPrimeValue().MallocUsed() != 0);
+  }
+
+  ~JsonAutoUpdater() {
+    if (update_on_delete_ && !set_size_was_called_) {
+      SetJsonSize();
+    } else if (!set_size_was_called_) {
+      VLOG(1) << "JsonAutoUpdater destructor called without SetJsonSize() being called. This may "
```
romange, line 95: maybe LOG(DFATAL) ?

src/server/json_family.cc
```diff
@@ -352,15 +433,85 @@ OpResult<DbSlice::ItAndUpdater> SetJson(const OpArgs& op_args, string_view key,
...
+  // We should do reset before setting the size of the json, because
+  // std::optional still holds the value and it will be deallocated
+  parsed_json.reset();
+  updater.SetJsonSize();
+
+  return OpStatus::OK;
+}
+
+/*
+  This method would use copy for json_str and not move!
```
romange, line 450: I did not understand this comment at all :)

romange:
Really good work, @BagritsevichStepan !

src/server/json_family.cc
BagritsevichStepan: As I understand, MEMORY USAGE also includes the size of the key, which might be an issue in this case. I think it would be useful to add a flag to the MEMORY USAGE command that excludes the key size from the calculation.

src/server/json_family.cc
BagritsevichStepan: It shouldn't fail in debug. For example:
```
JsonAutoUpdater updater{};
...
if (<result is not valid>) {
   return;
}
```
If the result is not valid we will not update the json. This means we don't need to call `SetJsonSize`, and that's totally fine. I thought we could use `LOG(WARNING)` here.

BagritsevichStepan:
I have split this PR into 4 separate PRs:
#5068, #5069, #5070, #5075

PR #5061, opened 2025-05-05
chore(metrics): add rdb_bgsave_in_progress and rdb_last_bgsave_status

### LINKED ISSUE #5060: add rdb_bgsave_in_progress and rdb_last_bgsave_status metrics
One of the users on discord is using some external lib that does:

```
   while [ $try -gt 0 ] ; do
        saved=$(echo 'info Persistence' | redis-cli -h ${backup_job_db_host} -p ${backup_job_db_port} ${REDIS_PASS_STR} | awk '/rdb_bgsave_in_progress:0/{print "saved"}')
        ok=$(echo 'info Persistence' | redis-cli -h ${backup_job_db_host} -p ${backup_job_db_port} ${REDIS_PASS_STR} | awk '/rdb_last_bgsave_status:ok/{print "ok"}')
        if [[ "$saved" = "saved" ]] && [[ "$ok" = "ok" ]]; then
            write_log notice "Redis Backup Complete"
            exit_code=0
            break
        fi
        try=$((try - 1))
        write_log warn "Redis Busy - Waiting and retrying in 5 seconds"
        sleep 5
    done
```

This does not work because we don't have those metrics defined.

kostasrim:
Add missing metrics: 

* rdb_bgsave_in_progress
* rdb_last_bgsave_status

resolves #5060

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -1707,7 +1707,8 @@ GenericError ServerFamily::DoSave(bool ignore_state) {
 }
 
 GenericError ServerFamily::DoSaveCheckAndStart(const SaveCmdOptions& save_cmd_opts,
-                                               Transaction* trans, bool ignore_state) {
+                                               Transaction* trans, bool ignore_state,
+                                               bool bg_save) {
   auto state = ServerState::tlocal()->gstate();
 
   // In some cases we want to create a snapshot even if server is not active, f.e in takeover
@@ -1726,9 +1727,11 @@ GenericError ServerFamily::DoSaveCheckAndStart(const SaveCmdOptions& save_cmd_op
                                 ? snapshot_storage_
                                 : CreateCloudSnapshotStorage(save_cmd_opts.cloud_uri);
 
-    save_controller_ = make_unique<SaveStagesController>(detail::SaveStagesInputs{
-        save_cmd_opts.new_version, save_cmd_opts.cloud_uri, save_cmd_opts.basename, trans,
-        &service_, fq_threadpool_.get(), snapshot_storage});
+    save_controller_ = make_unique<SaveStagesController>(
+        detail::SaveStagesInputs{save_cmd_opts.new_version, save_cmd_opts.cloud_uri,
+                                 save_cmd_opts.basename, trans, &service_, fq_threadpool_.get(),
+                                 snapshot_storage},
+        true);
 
     auto res = save_controller_->InitResourcesAndStart();
 
@@ -1736,7 +1739,14 @@ GenericError ServerFamily::DoSaveCheckAndStart(const SaveCmdOptions& save_cmd_op
       DCHECK_EQ(res->error, true);
       last_save_info_.SetLastSaveError(*res);
       save_controller_.reset();
+      if (bg_save) {
+        last_bgsave_status_ = false;
+      }
       return res->error;
+    }
+
+    if (bg_save) {
+      bgsave_in_progress_ = true;
     }
   }
   return {};
@@ -1750,6 +1760,11 @@ GenericError ServerFamily::WaitUntilSaveFinished(Transaction* trans, bool ignore
   {
     util::fb2::LockGuard lk(save_mu_);
     save_info = save_controller_->Finalize();
+
+    if (save_controller_->IsBgSave()) {
+      bgsave_in_progress_ = false;
+      last_bgsave_status_ = save_info.error ? false : true;
+    }
 
     if (save_info.error) {
       last_save_info_.SetLastSaveError(save_info);
@@ -2182,7 +2197,7 @@ void ServerFamily::BgSave(CmdArgList args, const CommandContext& cmd_cntx) {
     return;
   }
 
-  if (auto ec = DoSaveCheckAndStart(*maybe_res, cmd_cntx.tx); ec) {
+  if (auto ec = DoSaveCheckAndStart(*maybe_res, cmd_cntx.tx, false, true); ec) {
     cmd_cntx.rb->SendError(ec.Format());
     return;
   }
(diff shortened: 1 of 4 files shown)


src/server/server_family.h
```diff
@@ -395,6 +396,9 @@ class ServerFamily {
 
   // protected by save_mu_
   util::fb2::Fiber bg_save_fb_;
+  // false if last attempt failed
+  bool last_bgsave_status_ = true;
```
kostasrim, line 400: can't be moved to `LastSaveInfo` unfortunately :(

src/server/server_family.h
```diff
@@ -341,7 +341,8 @@ class ServerFamily {
   void BgSaveFb(boost::intrusive_ptr<Transaction> trans);
 
   GenericError DoSaveCheckAndStart(const SaveCmdOptions& save_cmd_opts, Transaction* trans,
-                                   bool ignore_state = false) ABSL_LOCKS_EXCLUDED(save_mu_);
+                                   bool ignore_state = false, bool bg_save = false)
```
kostasrim, line 344: TODO in this or another PR, group arguments under a struct

src/server/server_family.cc
```diff
@@ -1726,18 +1727,27 @@ GenericError ServerFamily::DoSaveCheckAndStart(const SaveCmdOptions& save_cmd_op
...
 
     auto res = save_controller_->InitResourcesAndStart();
 
     if (res) {
       DCHECK_EQ(res->error, true);
       last_save_info_.SetLastSaveError(*res);
       save_controller_.reset();
+      if (bg_save) {
+        last_bgsave_status_ = false;
+      }
```
BorysTheDev, line 1744: last_bgsave_status_ = last_bgsave_status_ && !bg_save

src/server/server_family.cc
```diff
@@ -1751,6 +1761,11 @@ GenericError ServerFamily::WaitUntilSaveFinished(Transaction* trans, bool ignore
     util::fb2::LockGuard lk(save_mu_);
     save_info = save_controller_->Finalize();
 
+    if (save_controller_->IsBgSave()) {
+      bgsave_in_progress_ = false;
+      last_bgsave_status_ = save_info.error ? false : true;
```
BorysTheDev, line 1766: last_bgsave_status_ = !save_info.error

src/server/server_family.cc
```diff
@@ -1726,18 +1727,27 @@ GenericError ServerFamily::DoSaveCheckAndStart(const SaveCmdOptions& save_cmd_op
...
       save_controller_.reset();
+      if (bg_save) {
+        last_bgsave_status_ = false;
+      }
       return res->error;
     }
+
+    if (bg_save) {
+      bgsave_in_progress_ = true;
+    }
```
BorysTheDev, line 1750: bgsave_in_progress_ = bgsave_in_progress_ || bg_save

src/server/server_family.cc
```diff
@@ -2621,6 +2636,13 @@ string ServerFamily::FormatInfoMetrics(const Metrics& m, std::string_view sectio
     }
     append("rdb_changes_since_last_success_save", m.events.update);
 
+    {
+      util::fb2::LockGuard lk{save_mu_};
+      append("rdb_bgsave_in_progress", bgsave_in_progress_ == true ? 1 : 0);
```
BorysTheDev, line 2641: static_cast<int>(bgsave_in_progress_)

src/server/server_family.h
BorysTheDev: why

src/server/server_family.cc
kostasrim: but why `last_bgsave_status_ && !bg_save` ? 

`last_bgsave_status_` is that status of the last attempt to call `bgsave`. If `res` is true, it means we had an error, so last_bgsave_status_ should be unconditionally false

src/server/server_family.cc
kostasrim: It's a global transaction. We don't depend on the previous value `bgsave_in_progress` because it should be false :)

src/server/server_family.cc
kostasrim: Sure, I am happy with implicit conversations ;)

src/server/server_family.cc
BorysTheDev: I just rewrote your code in boolean operators. if we don't depend on the previous state of bgsave_in_progress you code should be next `bgsave_in_progress_ = bg_save;`

src/server/server_family.cc
BorysTheDev: In this case you don't need if last_bgsave_status_ = bg_save

kostasrim: oh I am blind I misread. Sure no issue with that :shrug:

src/server/server_family.cc
kostasrim: yes yes I misread your comment :)

src/server/server_family.h
kostasrim: good question. I thought `LastSaveInfo` was something we construct and replace but I was wrong. It just a variable where we update its fields. I moved them there :)

src/server/server_family.cc
kostasrim: Actually, both expressions are incorrect:

1. `last_bgsave_status_ = last_bgsave_status_ && !bg_save` 

This won't always work. For example, if `res is true and last_bgsave_status_ == false` because the last bgsave attempt failed we get `last_bgsave_status = false && !false = true`. So what that means is that we did not execute `bgsave command` but rather `save` because `bgsave is false` yet we set to true the `last status` which is clearly incorrect (save is not bgsave!)

2. last_bgsave_status_ = !bg_save 

This won't work either. `bg_save` denotes if this flow is triggered by `bgsave command` (in case of true). `last_bgsave_status_` denotes if there was an error during the and it's not related to `bg_save`. 


`bg_save` is just there to conditionally set those variable only on the "interesting" (bgsave) flows

BorysTheDev: `last_bgsave_status(false) && !false = true` the result is false so it's correct

BorysTheDev (dismissed):

src/server/server_family.cc
kostasrim: lol what a brain fart :face_exhaling: :facepalm: You are right.

BorysTheDev (approved):

PR #5064, opened 2025-05-06
chore: dash table clean ups

romange:
Remove stash template parameter because we only use dashtable with a single configuration of STASH_CNT=4.


BorysTheDev (approved):

kostasrim (approved):
LGTM

PR #5068, opened 2025-05-06
feat(memory_cmd): Add WITHOUTKEY option for the MEMORY USAGE command. SECOND PR

BagritsevichStepan:
This is the second PR to fix JSON memory tracking issues.
It adds the `WITHOUTKEY` option to the `MEMORY USAGE` command to exclude the memory usage of the key from the result. This will be useful for checking JSON memory usage in memory tests.


romange:
This is fine but not really needed. For short keys that we use in tests, memory usage of a key is 0.

romange (approved):

PR #5069, opened 2025-05-06
fix(json_family): Fix memory tracking for the JSON.SET command. THIRD PR

### LINKED ISSUE #4926: Memory tracking tests for JSON and STREAMs

BagritsevichStepan:
fixes memory tracking issues for the JSON.SET command and #5054.
Also, partially fixes #4926

1. We didn't reset the result of `ShardJsonFromString`, which parses the json_str into the jsoncons's json and returns optional of this. The bug was that `std::optional` still holds the value even after `std::move`. So we overestimate the memory usage.
2. Bug in memory tracker when we overestimate the memory usage because of this:
```cpp
if (is_op_set) {
      diff += static_cast<int64_t>(mi_usable_size(pv.GetJson()));
}
```
3. Bug in partial `JSON.SET` when we called `ShardJsonFromString` before starting the memory tracking
4. Bug in indexes for the `JSON.SET` command: #5054

To fix this, I introduced `json_family_memory_test` and `JsonAutoUpdater`.
It ensures that `RemoveDoc` is called before memory tracking starts, and that `AddDoc` is called after `SetJsonSize` and also it will be called even if an error occurs.
Therefore, whenever we perform an update operation on a JSON value, this class must be used.

### DIFF
--- a/src/server/json_family.cc
+++ b/src/server/json_family.cc
@@ -82,6 +82,80 @@ class JsonMemTracker {
   size_t start_size_{0};
 };
 
+/* Helper class which must be initialized before any mutate operations on json.
+  It will track the memory usage of the json object and update the size in the CompactObj.
+  It also contains indexes updates, post update operations on the iterator. */
+class JsonAutoUpdater {
+ public:
+  JsonAutoUpdater(const OpArgs& op_args, string_view key, DbSlice::ItAndUpdater it,
+                  bool update_on_delete = false)
+      : op_args_(op_args), key_(key), it_(std::move(it)), update_on_delete_(update_on_delete) {
+    op_args.shard->search_indices()->RemoveDoc(key, op_args.db_cntx, it.it->second);
+
+    /* We need to initialize start memory usage after RemoveDoc because internally RemoveDoc has
+    static cache that can allocate/deallocate memory. Because of this, we will
+    overestimate/underestimate memory usage for json object. */
+    start_size_ = GetMemoryUsage();
+  }
+
+  JsonAutoUpdater(const JsonAutoUpdater&) = delete;
+  JsonAutoUpdater& operator=(const JsonAutoUpdater&) = delete;
+
+  JsonAutoUpdater(JsonAutoUpdater&&) = default;
+  JsonAutoUpdater& operator=(JsonAutoUpdater&&) = delete;
+
+  void SetJsonSize() {
+    set_size_was_called_ = true;
+
+    const size_t current = GetMemoryUsage();
+    int64_t diff = static_cast<int64_t>(current) - static_cast<int64_t>(start_size_);
+
+    GetPrimeValue().SetJsonSize(diff);
+
+    // Under any flow we must not end up with this special value.
+    DCHECK(GetPrimeValue().MallocUsed() != 0);
+  }
+
+  ~JsonAutoUpdater() {
+    if (update_on_delete_ && !set_size_was_called_) {
+      SetJsonSize();
+    } else if (!set_size_was_called_) {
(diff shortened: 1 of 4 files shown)


src/server/json_family.cc
```diff
@@ -318,28 +392,63 @@ bool JsonAreEquals(const JsonType& lhs, const JsonType& rhs) {
   }
 }
 
+/* Converts a JSONPath to a JSONPointer.
+   E.g. $[a][b][0] -> /a/b/0.
+   V1 JSONPath is not supported. */
+std::optional<std::string> ConvertJsonPathToJsonPointer(string_view json_path) {
```
BagritsevichStepan, line 398: This was just moved

src/server/json_family.cc
```diff
@@ -318,28 +392,63 @@ bool JsonAreEquals(const JsonType& lhs, const JsonType& rhs) {
...
+            << ". Invalid JSONPath.";
+    return std::nullopt;
+  }
+
+  std::string pointer;
+  const auto& path = parsed_path.value();
+  for (const auto& node : path) {
+    const auto& type = node.type();
+    if (type == json::SegmentType::IDENTIFIER) {
+      pointer += '/' + node.identifier();
```
romange, line 412: please use absl::StrAppend

src/server/json_family.cc
```diff
@@ -318,28 +392,63 @@ bool JsonAreEquals(const JsonType& lhs, const JsonType& rhs) {
...
+    } else if (type == json::SegmentType::INDEX) {
+      const auto& index = node.index();
+
+      if (index.first != index.second) {
+        VLOG(2) << "Error during conversion of JSONPath to JSONPointer: " << json_path
+                << ". Index range is not supported.";
+        return std::nullopt;
+      }
+
+      pointer += '/' + std::to_string(node.index().first);
```
romange, line 422: ditto

src/server/json_family.cc
```diff
@@ -352,15 +461,83 @@ OpResult<DbSlice::ItAndUpdater> SetJson(const OpArgs& op_args, string_view key,
...
-    res.it->second.SetJson(buf.data(), buf.size());
+    updater.GetPrimeValue().SetJson(buf.data(), buf.size());
   } else {
-    res.it->second.SetJson(std::move(*parsed_json));
+    updater.GetPrimeValue().SetJson(std::move(*parsed_json));
   }
 
-  tracker.SetJsonSize(res.it->second, res.is_new);
-  op_args.shard->search_indices()->AddDoc(key, op_args.db_cntx, res.it->second);
+  // We should do reset before setting the size of the json, because
```
romange, line 469: does it mean that `updater.GetPrimeValue().SetJson(std::move(*parsed_json));` does not really move anything?

romange (commented):
LGTM + minor comments

src/server/json_family.cc
BagritsevichStepan: It moves the object inside the `std::optional`, so no copy occurs. However, the `std::optional` still holds a "moved-from"/"default" value (i.e., it's not equal to `std::nullopt` after the move). As a result, it may still occupy some heap memory for the empty/moved-from object.

Check this [example](https://godbolt.org/z/sbhn5TzMz).

Possible solutions are to use `std::move(*std::exchange(parsed_json, std::nullopt))` or just to call `reset`.

romange (approved):

PR #5070, opened 2025-05-06
fix(json_family): Fix memory tracking for JSON mutate operations. FOURTH PR

BagritsevichStepan:
This is fourth PR that fixes memory tracking issues in JSON.

1. Bug in JsonMutateOperation, where we called `op_args.shard->search_indices()->RemoveDoc` after starting the memory tracking. `RemoveDoc` has a static cache that may allocate some memory, but this memory does not belong to the JSON object itself. As a result, memory usage will be overestimated in this case.


romange:
I do not see any test changes in this PR. Does this PR removes completely JsonMemTracker?

BagritsevichStepan:
I can try to reproduce the bug in the tests

romange (approved):

Issue #4883, opened 2025-04-03
Support keys compression in Dragonfly


romange:
See #4880 for motivation and as a prerequisite

## feature description

1. being able to enable keys huffman encoding via a run-time flag. 
2. can also disable it (via curl /flagz).
3. Once enabled it samples the key space by scanning N ( hardcoded ) keys, builds the histogram the derivative data structures needed for encoding/decoding.
4. The data structures can be built once during the process lifetime because decoding is dependent on them. 
5. Bonus: The data structures are immutable so might be shared among threads, but need to verify if it's easy to do (someone still needs to delete them upon service shutdown).
6. Once they appear, `CompactObject::SetString` should be able to use huffman encoding.
7. encoding stats like total raw size, total compressed size, tries, successful - should be exposed via "info".

PR #5076, opened 2025-05-06
chore: introduce huffman encoder class

romange:
One of the building blocks for #4883

### DIFF
--- a/src/core/huff_coder.cc
+++ b/src/core/huff_coder.cc
@@ -1,0 +1,110 @@
+// Copyright 2025, DragonflyDB authors.  All rights reserved.
+// See LICENSE for licensing terms.
+//
+
+#include "core/huff_coder.h"
+
+#include "base/logging.h"
+
+extern "C" {
+#include "huff/huf.h"
+}
+
+using namespace std;
+
+namespace dfly {
+
+constexpr size_t kWspSize = HUF_CTABLE_WORKSPACE_SIZE;
+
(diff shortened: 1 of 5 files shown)


src/core/huff_coder.cc
```diff
@@ -0,0 +1,110 @@
...
+  huf_ctable_.reset();
+  table_max_symbol_ = 0;
+}
+
+bool HuffmanEncoder::Encode(std::string_view data, uint8_t* dest, uint32_t* dest_size,
+                            std::string* error_msg) const {
+  DCHECK(huf_ctable_);
+
+  size_t res =
+      HUF_compress1X_usingCTable(dest, *dest_size, data.data(), data.size(), huf_ctable_.get(), 0);
```
kostasrim, line 71: Leaving it here for info, there are:

```
/** HUF_selectDecoder() :
 *  Tells which decoder is likely to decode faster,
 *  based on a set of pre-computed metrics.
 * @return : 0==HUF_decompress4X1, 1==HUF_decompress4X2 .
 *  Assumption : 0 < dstSize <= 128 KB */
U32 HUF_selectDecoder (size_t dstSize, size_t cSrcSize);
```

Same variants exists for compression I think

romange: yeah i know, this can be further improvement. right now I want it relatively simple.

BorysTheDev (approved):

Issue #4880, opened 2025-04-03
add huffman encoding support for CompactObject


romange:
## Motivation

Analysis of workload key spaces reveals that they are often well compressible using Huffman encoding, achieving a 30-45% reduction in size. This is primarily because these keys are typically ASCII text with low entropy.

Instead of using generic compression libraries, which introduce significant metadata and apply run-length encoding (RLE), we should use a low-level Huffman coding solution. Given that most keys are short and require independent processing, a standard Huffman coding approach with a global table per shard (across all keys) is likely the most efficient. I've checked it with "DEBUG COMPRESSION" on several workloads.

## This task

1.  **Implement Huffman Encoding in `CompactObject`:**
    * Currently, `ASCII1_ENC_BIT` and `ASCII2_ENC_BIT` represent three states: not encoded, ASCII1, and ASCII2.
    * We can repurpose these bits as a 2-bit enum to include a fourth state: `huff_encoded`.
2.  **Maintain Transparent Decoding:**
    * Ensure decoding is transparent to callers by storing the Huffman table locally within the `TL` structure defined in `compact_object.cc`.
3.  **Utilize Existing Huffman Utilities:**
    * Leverage the Huffman utilities exposed via `huff/huf.h`, as demonstrated in `debugcmd.cc`.
4.  **Restrict Huffman Encoding to Keys:**
    * Limit this feature to keys by adding a `bool huffman_allowed` argument to `CompactObj::SetString`, prohibiting Huffman encoding for other string types.
5.  **Manage Huffman Data Structures within `TL`:**
    * Ensure Huffman data structures within `TL` are immutable.
    * The table should either exist with pre-calculated weights or not exist at all.
    * Prevent modifications to the table if string objects have already been encoded using it.
6. encoding stats should be tracked under TL (attempts, successes, raw size, compressed size)
7. Do not attempt encode short strings (inline). Do not attempt encode huge strings (above certain length).
    Note that some strings after encoding may become inline, so the decoding code GetSlice should be able to handle it.

## Encoded Format

Huffman encoding can result in bit sequences that do not align with byte boundaries, as it translates each byte into one or more bits.

To address the ambiguity of decoding an encoded string of size *k* into the original string:

* Use the first byte to indicate the length of the "unused" suffix (the number of unused bits in the last byte).
* Since we need at least one extra byte to store the suffix length, apply Huffman encoding only if it reduces the data size by at least two bytes.

[...]

PR #5077, opened 2025-05-07
chore: reorganize compact object mask bits

romange:
Specifically get rit of MaskEnum and replace it with explicit bits aliasing the mask. Reorganize the encoding bits to be able to store 4 states that include huffman encoding.

Solves the first part of #4880

### DIFF
--- a/src/core/compact_object.h
+++ b/src/core/compact_object.h
@@ -125,35 +125,12 @@ class CompactObj {
     SBF_TAG = 22,
   };
 
-  enum MaskBit {
-    REF_BIT = 1,
-    EXPIRE_BIT = 2,  // Mark objects that have expiry timestamp assigned.
-    FLAG_BIT = 4,    // Used to mark keys that have memcache flags assigned.
-
-    // ascii encoding is not an injective function. it compresses 8 bytes to 7 but also 7 to 7.
-    // therefore, in order to know the original length we introduce 2 flags that
-    // correct the length upon decoding. ASCII1_ENC_BIT rounds down the decoded length,
-    // while ASCII2_ENC_BIT rounds it up. See DecodedLen implementation for more info.
-    ASCII1_ENC_BIT = 8,
-    ASCII2_ENC_BIT = 0x10,
-
-    // IO_PENDING is set when the tiered storage has issued an i/o request to save the value. It is
-    // cleared when the io request finishes or is cancelled.
-    IO_PENDING = 0x20,
-
-    // Applied only on keys that should be deleted asynchronously.
-    // (it can be the same value as IO_PENDING) that is applied only on values.
-    KEY_ASYNC_DELETE = 0x20,
-    STICKY = 0x40,
-
-    // TOUCHED used to determin which items are hot/cold.
-    // by checking if the item was touched from the last time we
-    // reached this item while travering the database to set items as cold.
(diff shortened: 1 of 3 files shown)


src/core/compact_object.h
```diff
@@ -185,13 +161,14 @@ class CompactObj {
     CompactObj res;
     memcpy(&res.u_, &u_, sizeof(u_));
     res.taglen_ = taglen_;
-    res.mask_ = mask_ | REF_BIT;
+    res.mask_ = mask_;
+    res.mask_bits_.ref = 1;
```
kostasrim, line 166: `mask_` and `mask_bits_` are both part of the same union. What you do here is:

1. Set `res.mask_` field to active and assign to `mask_`
2. Set `res.mask_bits_.ref = 1` which changes the active field of the union to `mask_bits_`. 

After this any access to `res.mask_` is ub.

src/core/compact_object.h
```diff
@@ -525,14 +482,38 @@ class CompactObj {
...
+      uint8_t io_pending : 1;  // also serves as async-delete for keys.
+      uint8_t sticky : 1;
+
+      // TOUCHED used to determin which items are hot/cold.
+      // by checking if the item was touched from the last time we
+      // reached this item while travering the database to set items as cold.
+      // https://junchengyang.com/publication/nsdi24-SIEVE.pdf
+      uint8_t touched : 1;  // used to mark keys that were accessed.
+    } mask_bits_;
+  };
```
kostasrim, line 510: What I am not sure about this is that only one field can be active at a time. Either `mask_` or `mask_bits_`. Isn't it very easy to get UB if we don't make sure which field is active ?

src/core/compact_object.h
kostasrim: So any caller of `AsRef` here will get UB if they use any member function that uses `res.mask_` without noticing.

kostasrim (commented):
Every call to `SetMeta` followed by a set to `mask_bit` changes the active member of the union. 

I am pretty sure I am missing something -- what is it ? :smile:

src/core/compact_object.h
romange: All these bits can be active at the same time.

It's not UB

src/core/compact_object.h
romange: mask_bits and mask point to the same byte. It's not ub

romange:
> Every call to `SetMeta` followed by a set to `mask_bit` changes the active member of the union.
> 
> I am pretty sure I am missing something -- what is it ? 😄

it's not c++ - it's c. the datatypes are POD, it's a simple aliasing.

src/core/compact_object.h
kostasrim: I am confused. Both `mask_` (plain uint) and `mask_bits_`  (unamed struct) are `different fields` of the `same unamed union`.  Unions can have one active field at a time ?

kostasrim (approved):
LGTM

kostasrim:
> > Every call to `SetMeta` followed by a set to `mask_bit` changes the active member of the union.
> > I am pretty sure I am missing something -- what is it ? 😄
> 
> it's not c++ - it's c. the datatypes are POD, it's a simple aliasing.

I get it now

src/core/compact_object.h
kostasrim: I get it now!

PR #5080, opened 2025-05-08
feat: add huffman coding to compact object

romange:
Read path and tests will follow in the next PR.
Partially addresses #4880

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -25,6 +25,7 @@ extern "C" {
 #include "base/pod_array.h"
 #include "core/bloom.h"
 #include "core/detail/bitpacking.h"
+#include "core/huff_coder.h"
 #include "core/qlist.h"
 #include "core/sorted_map.h"
 #include "core/string_map.h"
@@ -34,7 +35,6 @@ ABSL_FLAG(bool, experimental_flat_json, false, "If true uses flat json implement
 
 namespace dfly {
 using namespace std;
-using absl::GetFlag;
 using detail::binpacked_len;
 using MemoryResource = detail::RobjWrapper::MemoryResource;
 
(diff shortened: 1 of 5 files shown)


src/core/compact_object.cc
```diff
@@ -1402,7 +1418,47 @@ void CompactObj::EncodeString(string_view str) {
   DCHECK_EQ(NONE_ENC, mask_bits_.encoding);
 
   string_view encoded = str;
-  bool is_ascii = kUseAsciiEncoding && detail::validate_ascii_fast(str.data(), str.size());
+  bool huff_encoded = false;
+
+  // We chose such length that we can store the decoded lendth delta into 1 byte.
```
kostasrim, line 1423:
```suggestion
  // We chose such length that we can store the decoded length delta into 1 byte.
```

src/core/huff_coder.cc
```diff
@@ -112,6 +110,11 @@ string HuffmanEncoder::Export() const {
   return res;
 }
 
+// Copied from HUF_tightCompressBound.
+size_t HuffmanEncoder::CompressedBound(size_t src_size) const {
+  return ((src_size * num_bits_) >> 3) + 8;
```
kostasrim, line 115: Haven't seen `tightCompressBound` but why not use the macro here ?

kostasrim (approved):
lgtm

src/core/huff_coder.cc
romange: why macro?

kostasrim: Because I thought it was a macro. My point is, if there is such function or macro, you could use that instead

kostasrim (approved):

PR #2564, opened 2024-02-08
fix: asan errors on unit tests

kostasrim:
depends on https://github.com/romange/helio/pull/213 (once it's merged I will pull the latest helio here)

PR #5083, opened 2025-05-08
chore: allow setting huffman tables via DEBUG COMPRESSION SET

romange:
Also, revert the performance bomb from #2564

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -759,7 +759,7 @@ bool CompactObj::InitHuffmanThreadLocal(std::string_view hufftable) {
     return false;
   }
 
-  if (tl.huff_decoder.Load(hufftable, &err_msg)) {
+  if (!tl.huff_decoder.Load(hufftable, &err_msg)) {
     LOG(DFATAL) << "Failed to load huffman table: " << err_msg;
     return false;
   }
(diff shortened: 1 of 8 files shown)


src/core/compact_object.cc
```diff
@@ -1135,15 +1137,17 @@ void CompactObj::GetString(char* dest) const {
...
 
+      // we left some space on the left to allow inplace ascii unpacking.
+      size_t space_left = decoded_len - u_.small_str.size();
+
       string_view slices[2];
       unsigned num = u_.small_str.GetV(slices);
       DCHECK_EQ(2u, num);
-      std::string tmp(decoded_len, ' ');
-      char* next = tmp.data();
+      char* next = dest + space_left;
```
kostasrim, line 1146: `taglen_ == SMALL_TAG` so the string max len for that path after encoding is 255 bytes. Without encoding and assuming a 20% reduction, the maximum for the `tmp` buffer would have been `320 bytes` since a 20% reduction on that would give us `256` bytes and `SmallString::CanAllocate` limit is `256` bytes.

The 20% is also too much. bitpack is roughly 12.5% based on your comment and for huffman 1.8 bits per byte (haven't any serious research on this so it might be a little off)

src/core/detail/bitpacking.cc
```diff
@@ -269,6 +269,11 @@ void ascii_unpack(const uint8_t* bin, size_t ascii_len, char* ascii) {
   }
 }
 
+// In some cases we may read 2 bytes more than "allowed" (16 bytes during mm_loadu_si128
+// instead of 14 bytes that we actually need). Since we ignore these two bytes, there is no harm.
+// See CompactObjectTest.AsanTriggerReadOverflow for more details.
+// __attribute__((no_sanitize_address)) does not work here for some reason, so I had
+// to put it in sse_port.h.
```
kostasrim, line 276: I thought even the simd call itself was UB once it crossed the size :(

kostasrim (approved):
lgtm

src/core/compact_object.cc
romange: Yes, we are lucky that it's only hundreds of bytes

Issue #5084, opened 2025-05-08
Support TLS with memcache


romange:
Currently Dragonfly requires `requirepass` when tls is enabled. This is great behaviour but unfortunately memcache does not allow authentication. Therefore, we should allow non-password configurations in memcached mode.

Things to note: `ValidateServerTlsFlags` function.

romange:
so if we have a memcached listener we should allow tls even if no password is defined. otherwise lets keep it strict as today

PR #5085, opened 2025-05-08
fix: Support TLS for Memcached without password authentication

vyavdoshenko:
Fixes: https://github.com/dragonflydb/dragonfly/issues/5084

Dragonfly required password authentication when TLS was enabled. However, the Memcached protocol doesn't support password authentication, making it impossible to use Memcached with TLS.

Modified the `ValidateServerTlsFlags()` function to allow TLS without authentication when the Memcached port is enabled.


src/server/server_family.cc
```diff
@@ -299,6 +300,12 @@ bool ValidateServerTlsFlags() {
     has_auth = true;
   }
 
+  // Allow TLS without authentication for memcached protocol
+  // We check if memcached_port is enabled, as this is a static check during startup
+  if (GetFlag(FLAGS_memcached_port) > 0) {
```
kostasrim, line 305: The problem with this is that it does not check the if the connection is `redis` or `memcache`. 

So your test would pass if it included:

```
    c = await aioredis.Redis(**with_ca_tls_client_args)
    res = await c.execute_command("set foo bar") 
    assert res == b"OK"
```

We did not set `requirepass` and `with_tls_server_args` does not use a `ca` so the command should have failed to authenticate yet it did. 

The solution is to check which protocol is used before you set `has_auth` to true

tests/dragonfly/pymemcached_test.py
```diff
@@ -171,3 +172,41 @@ def test_expiration(memcached_client: MCClient):
...
+    # Create SSL context for client
+    ssl_context = ssl.create_default_context()
+    ssl_context.load_verify_locations(with_tls_ca_cert_args["ca_cert"])
+    ssl_context.check_hostname = False
+
+    # Disable certificate verification (since we don't provide a client certificate)
+    ssl_context.verify_mode = ssl.CERT_NONE
+
+    # Output port information for diagnostics
+    print(f"Connecting to memcached port: {server.mc_port} on host: 127.0.0.1")
```
romange, line 205: nit: we do not use prints in tests - only logging.info/debug etc

romange (approved):

src/server/server_family.cc
romange: ValidateServerTlsFlags is called during the server startup - there are no connections here. 
Server with memcache port won't require passwords for tls. That's what I asked .

kostasrim: > Server with memcache port won't require passwords for tls.
The way I understood it was that you wanted to only apply this for the memcache port and the reason I raised this is
because we now allow launching dragonfly without `requirepass` or `ca_dir` and both `redis` and `memcache` connections can now connect without requirepass.

Makes sense now :)

romange:
@vyavdoshenko  not urgent - can wait for next week, but i just I realized we fixed it wrong. We could still provide a password but always mark  memcached connections as authenticated. in fact it's not related to TLS, this should be done regardless because otherwise nobody can use memcache port if password is defined for the datastore. I think this is what @kostasrim  tried to say yesterday .

vyavdoshenko:
> @vyavdoshenko not urgent - can wait for next week, but i just I realized we fixed it wrong. We could still provide a password but always mark memcached connections as authenticated. in fact it's not related to TLS, this should be done regardless because otherwise nobody can use memcache port if password is defined for the datastore. I think this is what @kostasrim tried to say yesterday .

I see. I will fix it.

Issue #5033, opened 2025-04-30
support memcached: gets


romange:
See https://docs.memcached.org/protocols/basic/#gets

Currently we do not support CAS, but some clients use `GETS` calls instead of GET. Lets implement support for GETS that will always return 0 for cas token.

I will open a separate issue for Memcache CAS support

romange:
For meta protocol, I assume it will be the same `mg` but with cas flag specified.
no need for implementing CAS logic just always return 0 for cas token

PR #5087, opened 2025-05-09
feat(memcache): Add support for GETS

abhijat:
Support is added for the GETS command. A placeholder CAS token of 0 is always returned.

FIXES https://github.com/dragonflydb/dragonfly/issues/5033

### DIFF
--- a/src/facade/reply_builder.cc
+++ b/src/facade/reply_builder.cc
@@ -212,7 +212,7 @@ MCReplyBuilder::MCReplyBuilder(::io::Sink* sink) : SinkReplyBuilder(sink), all_(
 }
 
 void MCReplyBuilder::SendValue(std::string_view key, std::string_view value, uint64_t mc_ver,
-                               uint32_t mc_flag) {
+                               uint32_t mc_flag, bool send_cas_token) {
   ReplyScope scope(this);
   if (flag_.meta) {
     string flags;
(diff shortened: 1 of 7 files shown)


src/facade/reply_builder.cc
```diff
@@ -212,7 +212,11 @@ MCReplyBuilder::MCReplyBuilder(::io::Sink* sink) : SinkReplyBuilder(sink), all_(
 }
 
 void MCReplyBuilder::SendValue(std::string_view key, std::string_view value, uint64_t mc_ver,
-                               uint32_t mc_flag) {
+                               uint32_t mc_flag, bool send_mc_ver) {
+  // Return a placeholder cas token of 0
+  if (send_mc_ver) {
```
romange, line 217: this is not needed. Just pass 0 at caller

abhijat: There is a check later on which does not write version if it is 0, should I remove it for now? 

The boolean is mainly for that check https://github.com/dragonflydb/dragonfly/blob/bdfa30d77b303cdf01627aea17257032f39c679d/src/facade/reply_builder.cc#L234

abhijat: The boolean parameter might still be needed, the same code path is used for `GET` and `GETS`. It is used to differentiate if the version has to be written or not.

romange: yes, keep boolean, just rename it to "send_cas" but you do not need to override `mc_ver` - it's confusing. just pass 0 at the caller

romange (approved):

PR #5095, opened 2025-05-10
feat: Added store option for sort command.

H4R5H1T-007:
Added Store option in Sort command as mentioned [in this comment](https://github.com/dragonflydb/dragonfly/issues/3886#issuecomment-2395547969) which is in issue #3886. Please review this PR in the meanwhile I will start with SORT_RO command implementation.

### DIFF
--- a/src/server/generic_family.cc
+++ b/src/server/generic_family.cc
@@ -1509,37 +1509,23 @@ void GenericFamily::Sort(CmdArgList args, const CommandContext& cmd_cntx) {
   OpResultTyped<SortEntryList> fetch_result;
   auto fetch_cb = [&](Transaction* t, EngineShard* shard) {
     ShardId shard_id = shard->shard_id();
+    // in case of SORT option, we fetch only on the source shard
     if (shard_id == source_sid) {
       fetch_result = OpFetchSortEntries(t->GetOpArgs(shard), key, alpha);
     }
     return OpStatus::OK;
   };
 
-  if (store_key) {
-    cmd_cntx.tx->Execute(std::move(fetch_cb), false);
-  } else {
-    cmd_cntx.tx->Execute(std::move(fetch_cb), true);
-  }
-
-  // OpResultTyped<SortEntryList> fetch_result =
-  //     cmd_cntx.tx->ScheduleSingleHopT([&](Transaction* t, EngineShard* shard) {
-  //       return OpFetchSortEntries(t->GetOpArgs(shard), key, alpha);
-  //     });
-
-  if (fetch_result == OpStatus::WRONG_TYPE) {
-    cmd_cntx.tx->Conclude();
-    return builder->SendError(fetch_result.status());
-  }
-
-  if (fetch_result.status() == OpStatus::INVALID_NUMERIC_RESULT) {
-    cmd_cntx.tx->Conclude();
-    return builder->SendError("One or more scores can't be converted into double");
-  }
-
+  cmd_cntx.tx->Execute(std::move(fetch_cb), !bool(store_key));
   auto* rb = static_cast<RedisReplyBuilder*>(builder);
   if (!fetch_result.ok()) {
     cmd_cntx.tx->Conclude();
-    return rb->SendEmptyArray();
+    if (fetch_result == OpStatus::WRONG_TYPE)
+      return builder->SendError(fetch_result.status());
+    else if (fetch_result.status() == OpStatus::INVALID_NUMERIC_RESULT)
+      return builder->SendError("One or more scores can't be converted into double");
+    else
+      return rb->SendEmptyArray();
   }
 
   auto result_type = fetch_result.type();
--- a/src/server/generic_family.cc
+++ b/src/server/generic_family.cc
@@ -17,6 +17,7 @@ extern "C" {
 #include "base/cycle_clock.h"
 #include "base/flags.h"
 #include "base/logging.h"
+#include "core/qlist.h"
 #include "redis/rdb.h"
 #include "server/acl/acl_commands_def.h"
 #include "server/blocking_controller.h"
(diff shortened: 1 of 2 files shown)


scriptforfivem:
Interesting solution to this problem! I've been working on similar challenges recently.

fivemhostingserver:
Excellent contribution to open source! The community will benefit greatly from this.

fivemserverhostings:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

fivemtebex:
Very clean code with good test coverage. This is a great example of best practices.

fivemeyestore:
Great implementation of this feature. I particularly like the approach to error handling.

src/server/generic_family.cc
```diff
@@ -1438,10 +1439,34 @@ OpResultTyped<SortEntryList> OpFetchSortEntries(const OpArgs& op_args, std::stri
...
+    ql_v2->Push(it->key, where);
+  }
+  len = ql_v2->Size();
+
+  PrimeValue pv;
+  pv.InitRobj(OBJ_LIST, kEncodingQL2, ql_v2);
+
+  // This would overwrite existing value if any with new list.
+  // Set the expiry at 300 seconds.
+  auto op_res = op_args.GetDbSlice().AddOrUpdate(op_args.db_cntx, key, std::move(pv), 300000);
```
romange, line 1459: why do you set the expiry to 300s?

H4R5H1T-007: Oh Sorry for that. I read this in [Valkey documents ](https://valkey.io/commands/sort/) and thought we had to add some expiry earlier then forgot to remove it. Thanks for spotting it I will remove it.
>An interesting pattern using SORT ... STORE consists in associating an EXPIRE timeout to the resulting key so that in applications where the result of a SORT operation can be cached for some time. Other clients will use the cached list instead of calling SORT for every request. When the key will timeout, an updated version of the cache can be created by calling SORT ... STORE again.

src/server/generic_family.cc
```diff
@@ -1463,29 +1487,57 @@ void GenericFamily::Sort(CmdArgList args, const CommandContext& cmd_cntx) {
...
+    return OpStatus::OK;
+  };
+
+  if (store_key) {
+    cmd_cntx.tx->Execute(std::move(fetch_cb), false);
+  } else {
+    cmd_cntx.tx->Execute(std::move(fetch_cb), true);
+  }
+
+  // OpResultTyped<SortEntryList> fetch_result =
```
romange, line 1518: remove the old code

src/server/generic_family.cc
```diff
@@ -1463,29 +1487,57 @@ void GenericFamily::Sort(CmdArgList args, const CommandContext& cmd_cntx) {
...
 
-  OpResultTyped<SortEntryList> fetch_result =
-      cmd_cntx.tx->ScheduleSingleHopT([&](Transaction* t, EngineShard* shard) {
-        return OpFetchSortEntries(t->GetOpArgs(shard), key, alpha);
-      });
+  ShardId source_sid = Shard(key, shard_set->size());
+  OpResultTyped<SortEntryList> fetch_result;
+  auto fetch_cb = [&](Transaction* t, EngineShard* shard) {
+    ShardId shard_id = shard->shard_id();
+    if (shard_id == source_sid) {
```
romange, line 1506: pls add a comment: `in case of SORT option, we fetch only on the source shard`

src/server/generic_family.cc
```diff
@@ -1463,29 +1487,57 @@ void GenericFamily::Sort(CmdArgList args, const CommandContext& cmd_cntx) {
...
 
-  if (fetch_result.status() == OpStatus::INVALID_NUMERIC_RESULT)
+  if (fetch_result.status() == OpStatus::INVALID_NUMERIC_RESULT) {
+    cmd_cntx.tx->Conclude();
     return builder->SendError("One or more scores can't be converted into double");
+  }
 
   auto* rb = static_cast<RedisReplyBuilder*>(builder);
-  if (!fetch_result.ok())
+  if (!fetch_result.ok()) {
```
romange, line 1534: I would unify all the non-ok conditions here:
```
if (!fetch_result.ok()) {
   cmd_cntx.tx->Conclude();
   if (fetch_result.status() == OpStatus::INVALID_NUMERIC_RESULT) 
     return builder->SendError(...);
   ....
}
```

src/server/generic_family.cc
```diff
@@ -1438,10 +1439,33 @@ OpResultTyped<SortEntryList> OpFetchSortEntries(const OpArgs& op_args, std::stri
   return success ? res : OpStatus::INVALID_NUMERIC_RESULT;
 }
 
+template <typename IteratorBegin, typename IteratorEnd>
```
romange, line 1442: why do you need a template here? what are the types of the iterators?

src/server/generic_family.cc
H4R5H1T-007: Sure

src/server/generic_family.cc
H4R5H1T-007: yup that will be better. Will do that

H4R5H1T-007:
Hey @romange should I create Sort Generic like we have for other commands? We also have to make Sort_RO command it will be useful there.

romange:
@H4R5H1T-007 once you fix my comments we will be able to merge this PR.

src/server/generic_family.cc
```diff
@@ -1463,29 +1487,57 @@ void GenericFamily::Sort(CmdArgList args, const CommandContext& cmd_cntx) {
...
+      fetch_result = OpFetchSortEntries(t->GetOpArgs(shard), key, alpha);
+    }
+    return OpStatus::OK;
+  };
+
+  if (store_key) {
+    cmd_cntx.tx->Execute(std::move(fetch_cb), false);
+  } else {
+    cmd_cntx.tx->Execute(std::move(fetch_cb), true);
+  }
```
romange:
```suggestion
  cmd_cntx.tx->Execute(std::move(fetch_cb), !bool(store_key));
```

H4R5H1T-007:
Sorry I was inactive because I got busy with my job. Sure @romange I will update the PR.

romange:
Get your priorities straight @H4R5H1T-007 ;)

kostasrim:
> Get your priorities straight @H4R5H1T-007 ;)

yes it's never too late to quit your job :rofl: 

Sorry could not resist not making a joke here!

romange (approved):

romange:
thank you!

Issue #5096, opened 2025-05-10
Dragonfly crashes if Memcahed port is the same as Redis port


vyavdoshenko:
The command:
```
./dragonfly --dbfilename= --logtostderr --memcached_port=6379
```

The log:
```
I20250510 16:05:46.149094  1841 dfly_main.cc:752] Starting dragonfly df-dev-0000000
W20250510 16:05:46.149408  1841 dfly_main.cc:791] SWAP is enabled. Consider disabling it when running Dragonfly.
I20250510 16:05:46.149426  1841 dfly_main.cc:796] maxmemory has not been specified. Deciding myself....
I20250510 16:05:46.149432  1841 dfly_main.cc:805] Found 29.85GiB available memory. Setting maxmemory to 23.88GiB
I20250510 16:05:46.150444  1842 uring_proactor.cc:286] IORing with 1024 entries, allocated 98368 bytes, cq_entries is 2048
I20250510 16:05:46.151571  1841 proactor_pool.cc:149] Running 12 io threads
F20250510 16:05:46.159776  1841 accept_server.cc:85] Check failed: !ec Could not open port 6379 system:98/Address already in use
*** Check failure stack trace: ***
    @     0x5bc10a3a6ea7  google::LogMessage::Fail()
    @     0x5bc10a3a6ded  google::LogMessage::SendToLog()
    @     0x5bc10a3a65c2  google::LogMessage::Flush()
    @     0x5bc10a3aa43a  google::LogMessageFatal::~LogMessageFatal()
    @     0x5bc10a2dbe29  util::AcceptServer::AddListener()
    @     0x5bc1096e30e6  dfly::(anonymous namespace)::RunEngine()
    @     0x5bc1096e808f  main
    @     0x764548037d90  (unknown)
    @     0x764548037e40  __libc_start_main
    @     0x5bc1096e14a5  _start
*** SIGABRT received at time=1746882346 on cpu 0 ***
PC: @     0x7645480a49fc  (unknown)  pthread_kill
    @     0x5bc10a42c51b         64  absl::lts_20240722::WriteFailureInfo()
    @     0x5bc10a42c775         96  absl::lts_20240722::AbslFailureSignalHandler()
    @     0x764548050520  (unknown)  (unknown)
[1]    1841 IOT instruction (core dumped)  ./dragonfly --dbfilename= --logtostderr --memcached_port=6379
```

PR #5097, opened 2025-05-10
fix: memcached port has been already used crash

vyavdoshenko:
Fixes: https://github.com/dragonflydb/dragonfly/issues/5096

This PR fixes a crash when starting Dragonfly if the Memcached port is the same as the Redis port.

### DIFF
--- a/src/server/dfly_main.cc
+++ b/src/server/dfly_main.cc
@@ -296,8 +296,18 @@ void RunEngine(ProactorPool* pool, AcceptServer* acceptor) {
 
   if (mc_port > 0 && !tcp_disabled) {
     auto listener = MakeListener(Protocol::MEMCACHE, &service);
-    acceptor->AddListener(mc_port, listener.get());
-    listeners.push_back(listener.release());
+
+    if (mc_port == static_cast<uint32_t>(port)) {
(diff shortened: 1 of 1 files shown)


src/server/dfly_main.cc
```diff
@@ -296,8 +296,18 @@ void RunEngine(ProactorPool* pool, AcceptServer* acceptor) {
 
   if (mc_port > 0 && !tcp_disabled) {
     auto listener = MakeListener(Protocol::MEMCACHE, &service);
-    acceptor->AddListener(mc_port, listener.get());
-    listeners.push_back(listener.release());
+
+    if (mc_port == static_cast<uint32_t>(port)) {
```
romange, line 300: the original behavior of exiting the process is correct.
proceeding on startup when a serious error was discovered is not correct.
so if you want to avoid "crashing", you can just exit(1) nicely if `acceptor->AddListener` fails.
no need to compare ports as it's not important at the end.

romange: see how we do it with "main_listener".
in fact - every AddListener call in this file should be handled like that: check if there is an error, log it and exit.

vyavdoshenko: fixed

romange (approved):

PR #5098, opened 2025-05-10
fix: memcached authentication handling

vyavdoshenko:
This PR fixes authentication handling for the Memcached protocol. The previous fix in PR [#5085](https://github.com/dragonflydb/dragonfly/pull/5085) was wrong, because it turned password off for TLS connection regardless of whether it was Redis or Memcached.

Changes:
- Updated test_memcached_tls_no_requirepass to run with password configured but use the Memcached client without authentication.
- Reverted changes for ValidateServerTlsFlags().
- Skip password for Memcached protocol regardless of whether TLS is enabled or not.

Now Dragonfly correctly handles authentication settings for both protocols simultaneously.

### DIFF
--- a/src/facade/dragonfly_connection.h
+++ b/src/facade/dragonfly_connection.h
@@ -281,6 +281,11 @@ class Connection : public util::Connection {
     return name_;
   }
 
+  // Returns protocol type of this connection
+  Protocol GetProtocol() const {
+    return protocol_;
+  }
+
   struct MemoryUsage {
     size_t mem = 0;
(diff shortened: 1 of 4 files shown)


eyestorefivemstore:
Interesting solution to this problem! I've been working on similar challenges recently.

romange (approved):

rauanmayemir:
Which version has this fix in?
I had an issue with memcached failing to connect without password, and previous PR didn't help. I did have password auth and no TLS.

rauanmayemir:
Nevermind, found in the release notes.

PR #5100, opened 2025-05-11
refactor: provide CommandContext into InvokeCmd

BorysTheDev:
refactor code a little. Command context can store some additional intermediate information, so using it we can avoid double calculations. 
For example, instead of a  transaction, we can store the Slot Id in the command context.

### DIFF
--- a/src/server/main_service.h
+++ b/src/server/main_service.h
@@ -44,8 +44,7 @@ class Service : public facade::ServiceInterface {
                               facade::ConnectionContext* cntx) final;
 
   // Check VerifyCommandExecution and invoke command with args
-  bool InvokeCmd(const CommandId* cid, CmdArgList tail_args, facade::SinkReplyBuilder* builder,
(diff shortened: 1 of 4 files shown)


src/server/main_service.h
```diff
@@ -44,8 +44,7 @@ class Service : public facade::ServiceInterface {
                               facade::ConnectionContext* cntx) final;
 
   // Check VerifyCommandExecution and invoke command with args
-  bool InvokeCmd(const CommandId* cid, CmdArgList tail_args, facade::SinkReplyBuilder* builder,
-                 ConnectionContext* reply_cntx);
+  bool InvokeCmd(const CommandId* cid, CmdArgList tail_args, CommandContext cmd_cntx);
```
romange, line 47: better `const CommandContext&` especially that you plan on adding more fields

romange (approved):

PR #5102, opened 2025-05-12
chore: import t-digest-c lib

kostasrim:
* import t-digest-c lib as is

### DIFF
--- a/src/redis/td_malloc.h
+++ b/src/redis/td_malloc.h
@@ -1,0 +1,21 @@
+/**
+ * Adaptive histogram based on something like streaming k-means crossed with Q-digest.
+ * The implementation is a direct descendent of MergingDigest
+ * https://github.com/tdunning/t-digest/
+ *
+ * Copyright (c) 2021 Redis, All rights reserved.
+ *
+ * Allocator selection.
+ *
+ * This file is used in order to change the t-digest allocator at compile time.
+ * Just define the following defines to what you want to use. Also add
+ * the include of your alternate allocator if needed (not needed in order
+ * to use the default libc allocator). */
+
+#ifndef TD_ALLOC_H
+#define TD_ALLOC_H
+#define __td_malloc malloc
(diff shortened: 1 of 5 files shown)


src/redis/td_malloc.h
```diff
@@ -0,0 +1,21 @@
...
+ * Copyright (c) 2021 Redis, All rights reserved.
+ *
+ * Allocator selection.
+ *
+ * This file is used in order to change the t-digest allocator at compile time.
+ * Just define the following defines to what you want to use. Also add
+ * the include of your alternate allocator if needed (not needed in order
+ * to use the default libc allocator). */
+
+#ifndef TD_ALLOC_H
```
kostasrim, line 15: TODO I should enable mimalloc

src/redis/td_malloc.h
```diff
@@ -0,0 +1,21 @@
+/**
+ * Adaptive histogram based on something like streaming k-means crossed with Q-digest.
+ * The implementation is a direct descendent of MergingDigest
+ * https://github.com/tdunning/t-digest/
+ *
+ * Copyright (c) 2021 Redis, All rights reserved.
```
kostasrim, line 6: I think it should be ok to leave this here

kostasrim: it's appache 2.0 so we should be ok using it

BagritsevichStepan (approved):

src/redis/td_malloc.h
BagritsevichStepan: Are you going to do this in this PR?

src/redis/td_malloc.h
romange: Yes, please keep the headers, also please add the license file [LICENSE.md](https://github.com/RedisBloom/t-digest-c/blob/master/LICENSE.md) and name it t-digest.LICENSE.md

PR #5107, opened 2025-05-12
dfly_bench: Send auth if required before cluster setup

abhijat:
If dfly_bench runs against a cluster where auth is enabled, we have to send the password before fetching cluster info. This sets up each shard's endpoint correctly. Not doing so causes errors when handling `move` response, where hostnames are treated as ip addresses.


kostasrim (approved):
lgtm

romange (approved):

PR #5108, opened 2025-05-12
chore: add tdigest to compact object

kostasrim:
* add tdigest hist to compact object
* add sample test
* add tdigest family prototype
* fix zcalloc function prototype


BagritsevichStepan (approved):
We should also remember to support the TYPE command.

PR #5110, opened 2025-05-12
chore: add tdigest commands

kostasrim:
The following commands are implemented:

* create
* add
* rank
* revrank
* byrank
* byrevrank
* reset
* info
* min
* max
* cdf

### DIFF
--- a/src/server/tdigest_family.cc
+++ b/src/server/tdigest_family.cc
@@ -4,10 +4,398 @@
 
 #include "server/tdigest_family.h"
 
+#include "facade/cmd_arg_parser.h"
+#include "facade/error.h"
 #include "server/acl/acl_commands_def.h"
 #include "server/command_registry.h"
+#include "server/db_slice.h"
+#include "server/engine_shard.h"
+#include "server/transaction.h"
+
+extern "C" {
+#include "redis/tdigest.h"
+}
 
 namespace dfly {
+
+void TDigestFamily::Create(CmdArgList args, const CommandContext& cmd_cntx) {
+  facade::CmdArgParser parser{args};
+
+  auto key = parser.Next<std::string_view>();
+  size_t compression = 50;
+  if (parser.HasNext() && !parser.Check("COMPRESSION", &compression)) {
+    return cmd_cntx.rb->SendError(facade::kSyntaxErr);
+  }
+
+  auto cb = [key, compression](Transaction* tx, EngineShard* es) -> OpResult<bool> {
+    auto& db_slice = tx->GetDbSlice(es->shard_id());
+    auto db_cntx = tx->GetDbContext();
+    auto res = db_slice.AddOrFind(db_cntx, key);
+    if (!res) {
+      return res.status();
+    }
+
+    if (!res->is_new) {
+      return OpStatus::KEY_EXISTS;
+    }
+
+    td_histogram_t* td = td_new(compression);
+    // DENYOOM should cover this
+    if (!td) {
+      db_slice.Del(db_cntx, res->it);
+      return OpStatus::OUT_OF_MEMORY;
+    }
+    res->it->second.InitRobj(OBJ_TDIGEST, 0, td);
+    return OpStatus::OK;
+  };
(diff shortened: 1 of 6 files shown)


src/server/tdigest_family.cc
```diff
@@ -4,32 +4,417 @@
...
+    }
+    return OpStatus::OK;
+  };
+
+  auto res = cmd_cntx.tx->ScheduleSingleHopT(cb);
+  // SendError covers ok
+  return cmd_cntx.rb->SendError(res.status());
+}
+
+double TdGetByRank(td_histogram_t* td, double total_obs, double rnk) {
```
BagritsevichStepan, line 94: Please move this to the anonymous namespace, along with the other methods

src/server/tdigest_family.cc
```diff
@@ -4,32 +4,417 @@
...
+}
+
+void RankImpl(CmdArgList args, const CommandContext& cmd_cntx, bool reverse) {
+  facade::CmdArgParser parser{args};
+
+  auto key = parser.Next<std::string_view>();
+  std::vector<size_t> ranks;
+  while (parser.HasNext()) {
+    double val = parser.Next<double>();
+    ranks.push_back(val);
```
BagritsevichStepan, line 307: ranks is vector of size_t, and val is double. Something is wrong here.

src/server/tdigest_family.cc
```diff
@@ -4,32 +4,417 @@
...
+  facade::CmdArgParser parser{args};
+
+  auto key = parser.Next<std::string_view>();
+  size_t compression = 50;
+  if (parser.HasNext() && !parser.Check("COMPRESSION", &compression)) {
+    return cmd_cntx.rb->SendError(facade::kSyntaxErr);
+  }
+
+  auto cb = [key, compression](Transaction* tx, EngineShard* es) -> OpResult<bool> {
+    auto& db_slice = tx->GetDbSlice(es->shard_id());
```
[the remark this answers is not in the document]
kostasrim: I won't even bother for now, this is sketch -- and we do have callback within functions in other places. 

In general I agree, we should keep those functions as small as possible. I am happy to polish if we ever actually merge any of this functionality upstream.

src/server/tdigest_family.cc
```diff
@@ -4,32 +4,417 @@
...
+  }
+
+  auto cb = [key, &values](Transaction* tx, EngineShard* es) -> OpResult<bool> {
+    auto& db_slice = tx->GetDbSlice(es->shard_id());
+    auto db_cntx = tx->GetDbContext();
+    auto it = db_slice.FindMutable(db_cntx, key, OBJ_TDIGEST);
+    if (!it) {
+      return it.status();
+    }
+    auto* wrapper = it->it->second.GetRobjWrapper();
```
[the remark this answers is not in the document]
kostasrim: +1 I will add it in a follow up PR. not hat important for now

src/server/tdigest_family.cc
kostasrim: same comment as above, happy to polish one we actually consider this work for merging. I will move all those to unamed namespaces after I get tests+functionility done

src/server/tdigest_family.cc
```diff
@@ -4,32 +4,417 @@
...
+  rb->SendLong(res->unmerged_weight);
+  rb->SendSimpleString("Observations");
+  rb->SendLong(res->observations);
+  rb->SendSimpleString("Total compressions");
+  rb->SendLong(res->total_compressions);
+  rb->SendSimpleString("Memory usage");
+  rb->SendLong(res->mem_usage);
+}
+
+struct MinMax {
```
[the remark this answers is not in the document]
kostasrim: I explicitly wanted a struct here toi avoid first and second

src/server/tdigest_family.cc
kostasrim: I think the internal implementation converts to int. I will check all of these in detail once I add tests

PR #5115, opened 2025-05-13
Expiry fields during find

mkaruza:

### DIFF
--- a/src/core/intrusive_string_list.h
+++ b/src/core/intrusive_string_list.h
@@ -270,6 +270,18 @@ class IntrusiveStringList {
       }
     }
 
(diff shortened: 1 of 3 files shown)


src/core/intrusive_string_list.h
```diff
@@ -280,7 +292,7 @@ class IntrusiveStringList {
     }
 
     operator bool() const {
-      return prev_.Next();
+      return prev_.data_ && prev_.Next();
```
BorysTheDev, line 295: Didn't get when it is possible

BorysTheDev (approved):

Issue #5120, opened 2025-05-13
dragonfly crashes on XTRIM command


vyavdoshenko:
The command:
```
XTRIM "-992" "k1 \"v1\" k2 \"v2 with spaces\" \"k3 with spaces\" \"v3\"" "list1 element1"
```

The crash log:
```
F20250513 19:18:28.656924 327675 cmd_arg_parser.cc:38] Check failed: !error_.has_value() Parsing error occured but not checked
*** Check failure stack trace: ***
    @     0x5f2629fc57f3  google::LogMessage::Fail()
    @     0x5f2629fc5739  google::LogMessage::SendToLog()
    @     0x5f2629fc4eec  google::LogMessage::Flush()
    @     0x5f2629fc8fe4  google::LogMessageFatal::~LogMessageFatal()
    @     0x5f2629d87f33  facade::CmdArgParser::~CmdArgParser()
    @     0x5f2629882f00  dfly::StreamFamily::XTrim()
    @     0x5f262938646d  _ZN3fu27abi_4006detail10invocation6invokeIRKPFvN4absl12lts_202407224SpanIKSt17basic_string_viewIcSt11char_traitsIcEEEERKN4dfly14CommandContextEEJSC_SG_EEEDTclcl7forwardIT_Efp_Espcl7forwardIT0_Efp0_EEEOSL_DpOSM_
    @     0x5f262938610f  fu2::abi_400::detail::type_erasure::invocation_table::function_trait<>::internal_invoker<>::invoke()
    @     0x5f2629b0a6cb  _ZNK3fu27abi_4006detail12type_erasure6tables6vtableINS1_8propertyILb0ELb0EJKFvN4absl12lts_202407224SpanIKSt17basic_string_viewIcSt11char_traitsIcEEEERKN4dfly14CommandContextEEEEEE6invokeILm0EJPKNS2_13data_accessorERKmSE_SI_EEEDcDpOT0_
    @     0x5f2629b0a764  _ZN3fu27abi_4006detail12type_erasure7erasureILb1ENS1_6configILb1ELb1ENS_16capacity_defaultEEENS1_8propertyILb0ELb0EJKFvN4absl12lts_202407224SpanIKSt17basic_string_viewIcSt11char_traitsIcEEEERKN4dfly14CommandContextEEEEEE6invokeILm0ERKSN_JSG_SK_EEEDcOT0_DpOT1_
[...]

PR #5121, opened 2025-05-13
fix: XTRIM command crash

vyavdoshenko:
Fixes: https://github.com/dragonflydb/dragonfly/issues/5120

Additional checks for the following argument were added after the key for the XTRIM command.

### DIFF
--- a/src/server/stream_family.cc
+++ b/src/server/stream_family.cc
@@ -3328,10 +3328,10 @@ void StreamFamily::XTrim(CmdArgList args, const CommandContext& cmd_cntx) {
 
   std::string_view key = parser.Next();
 
-  if (!parser.HasNext() || (!absl::EqualsIgnoreCase(parser.Peek(), "MAXLEN") &&
-                            !absl::EqualsIgnoreCase(parser.Peek(), "MINID") &&
-                            !absl::EqualsIgnoreCase(parser.Peek(), "~") &&
-                            !absl::EqualsIgnoreCase(parser.Peek(), "="))) {
+  auto next_arg = parser.Peek();
+  if (!parser.HasNext() ||
+      (!absl::EqualsIgnoreCase(next_arg, "MAXLEN") && !absl::EqualsIgnoreCase(next_arg, "MINID") &&
+       next_arg != "~" && next_arg != "=")) {
     rb->SendError(kSyntaxErr);
     return;
   }
--- a/src/server/stream_family.cc
+++ b/src/server/stream_family.cc
@@ -2079,15 +2079,30 @@ OpResult<int64_t> OpTrim(const OpArgs& op_args, std::string_view key, const Trim
 ParseResult<TrimOpts> ParseTrimOpts(bool max_len, CmdArgParser* parser) {
   TrimOpts opts;
   opts.approx = parser->Check("~");
+  if (auto err = parser->Error()) {
+    return make_unexpected(err->MakeReply());
+  }
+
   if (!opts.approx) {
     parser->Check("=");
+    if (auto err = parser->Error()) {
+      return make_unexpected(err->MakeReply());
+    }
   }
 
   if (max_len) {
     opts.length_or_id = parser->Next<uint32_t>();
+    if (auto err = parser->Error()) {
+      return make_unexpected(err->MakeReply());
+    }
   } else {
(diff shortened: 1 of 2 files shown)


src/server/stream_family.cc
```diff
@@ -3328,6 +3328,14 @@ void StreamFamily::XTrim(CmdArgList args, const CommandContext& cmd_cntx) {
 
   std::string_view key = parser.Next();
 
+  auto next_arg = parser.Peek();
+  if (!parser.HasNext() ||
+      (!absl::EqualsIgnoreCase(next_arg, "MAXLEN") && !absl::EqualsIgnoreCase(next_arg, "MINID") &&
+       next_arg != "~" && next_arg != "=")) {
+    rb->SendError(kSyntaxErr);
+    return;
+  }
```
kostasrim, line 3337: From the stack trace on the issue I see a call to ` CmdArgParser::~CmdArgParser()` 

So it means that we triggered the `DCHECK(!error_.has_value()) << "Parsing error occured but not checked";` 

Which means that the parser had an error which was not consumed. 

So my question here is why don't we check instead for:

```
if (parser.HasError) {
  // send error and consume via Error() such that we don't trigger the DCHECK in the destructor
}
```

vyavdoshenko: I see. I will reconsider the fix.

vyavdoshenko: @kostasrim 
I added error checks for the xtrim command arguments. Please take a look.

src/server/stream_family.cc
```diff
@@ -2079,41 +2079,76 @@ OpResult<int64_t> OpTrim(const OpArgs& op_args, std::string_view key, const Trim
 ParseResult<TrimOpts> ParseTrimOpts(bool max_len, CmdArgParser* parser) {
   TrimOpts opts;
   opts.approx = parser->Check("~");
+  if (auto err = parser->Error()) {
+    return make_unexpected(err->MakeReply());
```
BagritsevichStepan, line 2083: I didn't get why do we need all this checks. That's why the parser was also added - to check for error only one time

src/server/stream_family.cc
```diff
@@ -3327,11 +3362,18 @@ void StreamFamily::XTrim(CmdArgList args, const CommandContext& cmd_cntx) {
...
 
   std::string_view key = parser.Next();
+  if (auto err = parser.Error()) {
+    return rb->SendError(err->MakeReply());
+  }
 
   auto parsed_trim_opts = ParseTrimOpts(&parser);
-  if (!parsed_trim_opts || !parser.Finalize()) {
-    rb->SendError(!parsed_trim_opts ? parsed_trim_opts.error() : parser.Error()->MakeReply());
-    return;
```
BagritsevichStepan, line 3334: What I see here, that to fix this crash you just need to change the order of the Finalize call and parsed_trim_opts. 

Like this:
```cpp
if ( !parser.Finalize() || !parsed_trim_opts) {
    if (parser.hasError)
  ...
    if (!parsed_trim_opts) {
       ...
    return;
```

BagritsevichStepan: What I mean here that it seams to me that we had parser error and !parsed_trim_opts at the same time. But we didn't "flush" the error from the parser. Because we checked only parsed_trim_opts

src/server/stream_family.cc
```diff
@@ -3329,8 +3329,12 @@ void StreamFamily::XTrim(CmdArgList args, const CommandContext& cmd_cntx) {
   std::string_view key = parser.Next();
 
   auto parsed_trim_opts = ParseTrimOpts(&parser);
-  if (!parsed_trim_opts || !parser.Finalize()) {
-    rb->SendError(!parsed_trim_opts ? parsed_trim_opts.error() : parser.Error()->MakeReply());
```
BagritsevichStepan, line 3333: You can see that you broke another logic. Our priority is to send parsed_trim_opts, because we have detailed error message there. So what you need to do is:
```cpp
auto error = parser.Error();
rb->SendError(!parsed_trim_opts ? parsed_trim_opts.error() : error>MakeReply());
```

vyavdoshenko: fixed

BagritsevichStepan (approved):

PR #5128, opened 2025-05-14
chore: add tests

kostasrim:
Fixes a bunch of bugs around command replies for tdigest functions and adds tests.

* fix bugs around command replies
* add tests

### DIFF
--- a/src/server/tdigest_family.cc
+++ b/src/server/tdigest_family.cc
@@ -103,7 +103,7 @@ void ByRankImpl(CmdArgList args, const CommandContext& cmd_cntx, bool reverse) {
   facade::CmdArgParser parser{args};
 
   auto key = parser.Next<std::string_view>();
-  std::vector<size_t> ranks;
+  std::vector<double> ranks;
   while (parser.HasNext()) {
     double val = parser.Next<double>();
     ranks.push_back(val);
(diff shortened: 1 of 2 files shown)


src/server/tdigest_family.cc
```diff
@@ -457,8 +462,15 @@ void TDigestFamily::Quantile(CmdArgList args, const CommandContext& cmd_cntx) {
   };
 
   auto res = cmd_cntx.tx->ScheduleSingleHopT(cb);
-  // SendError covers ok
-  return cmd_cntx.rb->SendError(res.status());
+  if (!res) {
```
kostasrim, line 465: I will extract those in a function at some point :rofl:

src/server/tdigest_family_test.cc
```diff
@@ -20,7 +20,14 @@ class TDigestFamilyTest : public BaseFamilyTest {
  protected:
 };
 
-TEST_F(TDigestFamilyTest, TDigestBasic) {
+TEST_F(TDigestFamilyTest, Basic) {
```
kostasrim, line 23: We now have tests for all `TDIGEST` commands :)

kostasrim:
only serialization is left and this is feature complete.

src/server/tdigest_family.cc
```diff
@@ -412,8 +410,15 @@ void TDigestFamily::Cdf(CmdArgList args, const CommandContext& cmd_cntx) {
...
+    return cmd_cntx.rb->SendError(res.status());
```
BagritsevichStepan, line 414: Please move it below `auto* rb = static_cast<facade::RedisReplyBuilder*>(cmd_cntx.rb);`

src/server/tdigest_family.cc
BagritsevichStepan: Ah, okay. So my previous comment can be ignored.

PR #5129, opened 2025-05-14
fix: make search keywords case insensitive

romange:
Also fix the order of alias and the runtime arguments in KNN query.
Making sure that this query:
`"FT.SEARCH" "user-idx" "*=>[knn 3 @embedding $vector EF_RUNTIME 5 as vector_distance]" "RETURN" "1" "vector_distance"  DIALECT 2   "params" "2" "vector" "\n\xd7#>{\x14\xae\xbeH\xe1z?\x1f\x85k>"`  is parsed correctly.

### DIFF
--- a/src/core/search/parser.y
+++ b/src/core/search/parser.y
@@ -98,14 +98,14 @@ final_query:
       { driver->Set(AstKnnNode(std::move($1), std::move($3))); }
 
 knn_query:
-  LBRACKET KNN UINT32 FIELD TERM opt_knn_alias opt_ef_runtime RBRACKET
+  LBRACKET KNN UINT32 FIELD TERM opt_ef_runtime opt_knn_alias RBRACKET
     {
       auto vec_result = BytesToFtVectorSafe($5);
       if (!vec_result) {
         error(@5, "Invalid vector format");
         YYERROR;
       }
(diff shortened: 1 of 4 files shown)


src/core/search/parser.y
```diff
@@ -98,14 +98,14 @@ final_query:
       { driver->Set(AstKnnNode(std::move($1), std::move($3))); }
 
 knn_query:
-  LBRACKET KNN UINT32 FIELD TERM opt_knn_alias opt_ef_runtime RBRACKET
+  LBRACKET KNN UINT32 FIELD TERM opt_ef_runtime opt_knn_alias RBRACKET
```
vyavdoshenko, line 101: Was the previous syntax correct?
If yes, we have to add two rules:
LBRACKET KNN UINT32 FIELD TERM opt_knn_alias opt_ef_runtime RBRACKET
LBRACKET KNN UINT32 FIELD TERM opt_ef_runtime opt_knn_alias RBRACKET

romange: it was not

vyavdoshenko (approved):
lgtm

BagritsevichStepan (approved):

PR #5130, opened 2025-05-14
chore: add basic TOPK commands

kostasrim:
Add basic support for topk data structure in compact object and also implement prototypes for topk.reserve, topk.add and topk.list.

* add support for topk ds in compact object
* add topk.reserve prototype
* add topk.add prototype
* add topk.list prototype

### DIFF
--- a/src/core/compact_object.h
+++ b/src/core/compact_object.h
@@ -15,6 +15,7 @@
 #include "core/mi_memory_resource.h"
 #include "core/small_string.h"
 #include "core/string_or_view.h"
+#include "core/top_keys.h"
 
 namespace dfly {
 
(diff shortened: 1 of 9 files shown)


src/core/compact_object.h
```diff
@@ -479,6 +484,7 @@ class CompactObj {
     // using 'packed' to reduce alignement of U to 1.
     JsonWrapper json_obj __attribute__((packed));
     SBF* sbf __attribute__((packed));
+    TopKeys* topk __attribute__((packed));
```
kostasrim, line 487: We use a different ds than valkey, I will write tests but probably we won't be 1-1 do the the probabilistic nature of the ds

romange: does valkey support it?

kostasrim: When I said valkey I meant the one who can't be named :smile:

PR #5140, opened 2025-05-16
fix: out of bounds SIMD access

romange:
Fixes the regression introduced by #5083
The bug: accessing 2 bytes out of bounds when reading the source string. While the algorithm discards those bytes when doing decoding, it still can cause segfaults because these two bytes could belong to a new memory page that is not part of the process space.


kostasrim (approved):

kostasrim:
@romange sanitizers failed

kostasrim (approved):

PR #5146, opened 2025-05-16
chore: optimize modulo operations for powers of two

vyavdoshenko:
Replaced expressions like `value % 2` with bitwise equivalents like `(value & 1)`.
This is mathematically identical for powers of two but slightly faster, since bitwise operations are cheaper than modulo.


romange:
Hi Volodymir,

I wish you asked first. Every compiler knows how to replace such divisions
with masks. You can easily check it with godbolt.

We keep the original division because it's more readable

Roman Gershman
CTO
---

PR #5157, opened 2025-05-20
pytest: Add sleep to let master stop propagate to replica

abhijat:
The test changed in this PR `test_replica_reconnect` is failing often where the replica does not see updated master link status, after master is terminated. The replica code path may take some time to set the flags in its state mask. Adding a sleep may help with avoiding this test failure.

### DIFF
--- a/tests/dragonfly/replication_test.py
+++ b/tests/dragonfly/replication_test.py
@@ -2279,6 +2279,9 @@ async def test_replica_reconnect(df_factory, break_conn):
     # kill existing master, create master with different repl_id but same port
     master_port = master.port
     master.stop()
+
+    await asyncio.sleep(1)
+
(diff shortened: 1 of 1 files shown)


tests/dragonfly/replication_test.py
```diff
@@ -2279,6 +2280,9 @@ async def test_replica_reconnect(df_factory, break_conn):
     # kill existing master, create master with different repl_id but same port
     master_port = master.port
     master.stop()
+
+    time.sleep(5)
```
romange, line 2284: we usually use `await asyncio.sleep` for that

romange: Also, 5s is too agressive, I would reduce to 1s first

romange (approved):

Issue #5159, opened 2025-05-20
Dragonfly crashes on PFCOUNT command


vyavdoshenko:
The scenario:
```
SET "key1" "value1"
PFADD "key" "value"
PFADD "list1 element1" "data"
PFCOUNT "key1" "key" "list1 element1"
```

The crash log:
```
F20250520 21:51:53.829140 23112 transaction.cc:694] Check failed: OpStatus::OK == result (0 vs. 5) 
*** Check failure stack trace: ***
    @     0x555556f69293  google::LogMessage::Fail()
    @     0x555556f691d9  google::LogMessage::SendToLog()
    @     0x555556f6898c  google::LogMessage::Flush()
    @     0x555556f6ca84  google::LogMessageFatal::~LogMessageFatal()
    @     0x555556afe501  dfly::Transaction::RunCallback()
    @     0x555556afd71a  dfly::Transaction::RunInShard()
    @     0x5555563d63d9  _ZZN4dfly11EngineShard13PollExecutionEPKcPNS_11TransactionEENKUlS4_bE_clES4_b
    @     0x5555563d70e6  dfly::EngineShard::PollExecution()
    @     0x555556b0078b  _ZZN4dfly11Transaction11DispatchHopEvENKUlvE_clEv
    @     0x555556b0edfa  _ZSt13__invoke_implIvRZN4dfly11Transaction11DispatchHopEvEUlvE_JEET_St14__invoke_otherOT0_DpOT1_
    @     0x555556b0e829  _ZSt10__invoke_rIvRZN4dfly11Transaction11DispatchHopEvEUlvE_JEENSt9enable_ifIX16is_invocable_r_vIT_T0_DpT1_EES5_E4typeEOS6_DpOS7_
    @     0x555556b0e34e  _ZNSt17_Function_handlerIFvvEZN4dfly11Transaction11DispatchHopEvEUlvE_E9_M_invokeERKSt9_Any_data
    @     0x55555631262e  std::function<>::operator()()
    @     0x555556ef2da4  util::fb2::FiberQueue::Run()
    @     0x555556bef05d  _ZZN4dfly9TaskQueue5StartESt17basic_string_viewIcSt11char_traitsIcEEENKUlvE_clEv
    @     0x555556bf0239  _ZSt13__invoke_implIvZN4dfly9TaskQueue5StartESt17basic_string_viewIcSt11char_traitsIcEEEUlvE_JEET_St14__invoke_otherOT0_DpOT1_
[...]

Issue #5004, opened 2025-04-26
HllFamilyTest.MergeInvalid check fails with multiple shards


romange:
The reason for this is that the transaction callback in PFMergeInternal returns status via the transactional framework. This convenience interface is allowed only for single shard commands and does not work for multi-shard ones that may return inconsistent status from each shard thread. 

The solution would be to always return Status::OK and pass the error code externally outside of the transaction framework like we do with other multi key operations.

```
F20250426 08:55:33.082108  9330 transaction.cc:686] Check failed: OpStatus::OK == result (0 vs. 5) 
*** Check failure stack trace: ***
    @     0x5ba7a27634ed  google::LogMessage::Fail()
    @     0x5ba7a2763433  google::LogMessage::SendToLog()
    @     0x5ba7a2762be6  google::LogMessage::Flush()
    @     0x5ba7a2766cde  google::LogMessageFatal::~LogMessageFatal()
    @     0x5ba7a21f9b35  dfly::Transaction::RunCallback()
    @     0x5ba7a21f8d4e  dfly::Transaction::RunInShard()
```

PR #5161, opened 2025-05-21
fix: Prevent crash in PFCOUNT when using keys of different types

vyavdoshenko:
Fixes: https://github.com/dragonflydb/dragonfly/issues/5159
PFCOUNT command crashes when arguments include keys of different types (regular strings and HyperLogLog) due to improper WRONG_TYPE error handling in the transaction flow.

Fixed PFCountMulti to properly propagate WRONG_TYPE errors. Modified callback to always return OpStatus::OK while tracking errors separately.

### DIFF
--- a/src/server/hll_family.cc
+++ b/src/server/hll_family.cc
@@ -208,7 +208,7 @@ OpResult<int64_t> PFCountMulti(CmdArgList args, const CommandContext& cmd_cntx) 
   vector<vector<string>> hlls;
   hlls.resize(shard_set->size());
 
-  OpStatus status = OpStatus::OK;
+  atomic_bool success = true;
   auto cb = [&](Transaction* t, EngineShard* shard) {
     ShardId sid = shard->shard_id();
     ShardArgs shard_args = t->GetShardArgs(shard->shard_id());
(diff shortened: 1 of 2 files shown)


src/server/hll_family.cc
```diff
@@ -208,17 +208,25 @@ OpResult<int64_t> PFCountMulti(CmdArgList args, const CommandContext& cmd_cntx)
   vector<vector<string>> hlls;
   hlls.resize(shard_set->size());
 
+  OpStatus status = OpStatus::OK;
```
romange, line 211: It's not a good fix for several reasons.  Instead you should convert the array at line 208 to
`vector<OpResult<vector<string>>> hll`

src/server/hll_family.cc
```diff
@@ -208,17 +208,25 @@ OpResult<int64_t> PFCountMulti(CmdArgList args, const CommandContext& cmd_cntx)
...
   auto cb = [&](Transaction* t, EngineShard* shard) {
     ShardId sid = shard->shard_id();
     ShardArgs shard_args = t->GetShardArgs(shard->shard_id());
     auto result = ReadValues(t->GetOpArgs(shard), shard_args);
     if (result.ok()) {
       hlls[sid] = std::move(result.value());
+    } else {
+      status = result.status();
```
romange, line 219: not good.

src/server/hll_family.cc
```diff
@@ -208,17 +208,25 @@ OpResult<int64_t> PFCountMulti(CmdArgList args, const CommandContext& cmd_cntx)
...
     }
-    return result.status();
+    return OpStatus::OK;
```
romange, line 221: yeah, we should return `OpStatus::OK` here.

src/server/hll_family.cc
```diff
@@ -208,17 +208,25 @@ OpResult<int64_t> PFCountMulti(CmdArgList args, const CommandContext& cmd_cntx)
...
   };
 
-  cmd_cntx.tx->ScheduleSingleHop(std::move(cb));
+  OpStatus cb_status = cmd_cntx.tx->ScheduleSingleHop(std::move(cb));
+
+  // If there was an error in the execution of ReadValues, return it
```
romange, line 226: now you go over the array of hlls and check if one of them had an error.
Alternatively, you can keep the way you do it now but you must make status atomic, i.e.
atomic<OpStatus> status because it could be updated from multiple shard threads.

romange (commented):
you can also check how https://github.com/dragonflydb/dragonfly/issues/5004
was done.

src/server/hll_family.cc
vyavdoshenko: I decided to use atomic_bool as well as https://github.com/dragonflydb/dragonfly/pull/5008 PR

src/server/hll_family.cc
vyavdoshenko: Fixed

BorysTheDev (approved):

Issue #5125, opened 2025-05-14
vector search: unsupported vector distance computed field


romange:
To reproduce:

1. `"FT.CREATE" "user-idx" "ON" "JSON" "PREFIX" "1" "user" "SCORE" "1.0" "SCHEMA" "$.user" "AS" "user" "TAG" "SEPARATOR" "," "$.credit_score" "AS" "credit_score" "TAG" "SEPARATOR" "," "$.embedding" "AS" "embedding" "VECTOR" "FLAT" "6" "TYPE" "FLOAT32" "DIM" "4" "DISTANCE_METRIC" "COSINE"`
2. ` "JSON.SET" "user:john" "$" "{\"user\": \"john\", \"credit_score\": \"high\", \"embedding\": [0.23, 0.49, -0.18, 0.95]}"`
3. `"FT.SEARCH" "user-idx" "*=>[KNN 3 @embedding $vector AS vector_distance]" "RETURN" "1" "vector_distance" "SORTBY" "vector_distance" "ASC" "DIALECT" "2" "LIMIT" "0" "3" "params" "2" "vector" "\n\xd7#>{\x14\xae\xbeH\xe1z?\x1f\x85k>"`

Dragonfly logs a warning:
`Invalid Json path: vector_distance Error: Unknown function: vector_distance`

1. If query can not be executed *correctly* - we should reply with an error
2. field extractor does not recognize a precompute field/alias vector_distance that is defined as a vector distance between $vector and the `@embedding` field

vyavdoshenko:
I can't reproduce the issue:

Dragonfly:
```
127.0.0.1:6379> "FT.CREATE" "user-idx" "ON" "JSON" "PREFIX" "1" "user" "SCORE" "1.0" "SCHEMA" "$.user" "AS" "user" "TAG" "SEPARATOR" "," "$.credit_score" "AS" "credit_score" "TAG" "SEPARATOR" "," "$.embedding" "AS" "embedding" "VECTOR" "FLAT" "6" "TYPE" "FLOAT32" "DIM" "4" "DISTANCE_METRIC" "COSINE"
OK
127.0.0.1:6379> "JSON.SET" "user:john" "$" "{\"user\": \"john\", \"credit_score\": \"high\", \"embedding\": [0.23, 0.49, -0.18, 0.95]}"
OK
[...]

PR #5163, opened 2025-05-21
chore: lower log level for harmless "Invalid Json path" warnings

vyavdoshenko:
Fixes: https://github.com/dragonflydb/dragonfly/issues/5125

This PR addresses the issue of excessive "Invalid Json path" warnings in the logs when using computed fields (e.g., vector_distance from KNN search) in FT.SEARCH RETURN clauses.

Although the search functionality itself worked correctly and returned the computed values, these warnings cluttered the logs unnecessarily.

This change modifies `JsonAccessor::GetPath` to use `VLOG(1)` instead of `LOG(WARNING)` when a JSON path cannot be parsed or an expression cannot be created. This silences these specific warnings under normal log levels, as they are not indicative of an actual error when dealing with fields that are computed and not expected to be found within the JSON document structure.

This approach simplifies the solution by avoiding complex logic to explicitly mark fields as "computed" throughout the search parameter parsing and processing stages.

### DIFF
--- a/src/server/search/doc_accessors.cc
+++ b/src/server/search/doc_accessors.cc
@@ -364,7 +364,7 @@ JsonAccessor::JsonPathContainer* JsonAccessor::GetPath(std::string_view field) c
   }
 
   if (!ptr) {
-    LOG(WARNING) << "Invalid Json path: " << field << ' ' << ec_msg;
+    VLOG(1) << "Invalid Json path: " << field << ' ' << ec_msg;
     return nullptr;
   }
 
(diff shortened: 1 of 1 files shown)


romange (commented):
Are you saying it's a normal behavior for computed  fields? Maybe worth adding a comment here that it can happen even for "valid" queries that use computed fields

vyavdoshenko:
> Are you saying it's a normal behavior for computed fields? Maybe worth adding a comment here that it can happen even for "valid" queries that use computed fields

Yes, exactly. I tried to handle this outside, but the code became more complex. Also, the document can contain such fields, so we must try to find them. That's why changing the log level to VLOG solves this issue, but we can turn it on if needed.

Issue #5073, opened 2025-05-06
Introduce Dash::Segment stretching


romange:
## Problem:
when dragonfly grows, even though it uses independent segments, its table usage doubles very quickly. it happens because segment load is more or less uniform, which means all segments need to split around the same time.

once the split happens, table utilization goes down to 50-55% because all the existing segments split.
it's not a problem with relatively large values but with small values and keys, the overhead per entry doubles, which can be substantial.

## Solution
Due to how dash table (or any extendible hashing) works, we must split and double segment capacity. 
Therefore, it is expected that after the split, around 50% of the existing items will move to another segment.

Therefore, I suggest "stretching" a segment by adding him a sidekick segment(s) that will "increase" dynamically its capacity by $$(1+\delta)$$. As a result, the split event will come when segment load factor will be capped by $$(1+\delta)$$ as well. And during the split, both segments will be at $$(0.5+ \frac{\delta}{2})$$.
With $$\delta \approx 0.15$$ we can get much smoother growth.

## Additional value
We do not have to split when we are low in memory, we can just stretch the segment a little.
We could eliminate `table_growth_margin` logic for example.

Things to think about:
1. Now number of physical buckets can vary per segment, though still be capped.
2. How we ensure stability of traversal?
3. How do we ensure that CVC covers all the physical buckets?
4. Table eviction policy - how do we combine stretches with growth, eviction etc.

kostasrim:
I was about to ask that 😄 

 > How we ensure stability of traversal? 

Speaking of segment load factors and splits, I would also suggest to go over SEPH. It's a dash alternative that drops the extendable hashing. Would be curious if we can find some useful approaches in there (IMO I wanted to implement this as part of the hackathon -- but oh well you can't replace dash in one week 😄 )

romange:
I do not know if SEPH can replace Dash in Dragonfly but there is one idea in SEPH that may be relevant here.

I wanted to introduce an additional smaller  segment that will serve as dynamic extension for each segment.
Instead, we could introduce a new level of segments, each shared by their primary segments as an extensible storage.

### HalfSplit
So, if say  S1 is full, we do not split it, instead we create L2_S1 and chain it to S1, and insert there. L2_S1 has recount 1.
Later, if S2 is full, we take an available segment with refcount 1 if exists, in this case L2_S1 and also chain to S2.

### Full Split
[...]

PR #5176, opened 2025-05-24
chore(dash_table): pass memory resource to segment

romange:
1. Currently, memory resource is not used - only passed if needed.
This is needed for https://github.com/dragonflydb/dragonfly/issues/5073

2. HotspotBuckets are renamed to HotBuckets.
3. Move TryInsertToBucket function to bucket level

### DIFF
--- a/src/core/dash_internal.h
+++ b/src/core/dash_internal.h
@@ -13,6 +13,7 @@
 #include <cstring>
 #include <type_traits>
 
+#include "base/pmr/memory_resource.h"
 #include "core/sse_port.h"
 
 namespace dfly {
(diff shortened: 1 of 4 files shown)


src/core/dash_internal.h
```diff
@@ -322,6 +314,19 @@ template <typename _Key, typename _Value, typename Policy = DefaultSegmentPolicy
...
 
+    // Returns slot id if insertion is successful, -1 if no free slots are found.
+    template <typename U, typename V>
+    int TryInsertToBucket(U&& key, V&& value, uint8_t meta_hash, bool probe) {
+      if (this->IsFull()) {
+        return -1;  // no free space in the bucket.
+      }
+
+      int slot = this->slotb_.FindEmptySlot();
+      assert(slot >= 0);
```
kostasrim, line 325: `assert(slot < int(kSlotNum));` also ?

kostasrim (approved):
LGTM

Issue #5173, opened 2025-05-23
Race Condition in Search Indexing Leading to Crash During Concurrent Operations


vyavdoshenko:
**Observed Behavior:**
Dragonfly server crashes intermittently when subjected to a high load of concurrent search index operations. The crash appears to be related to a race condition within the search indexing subsystem. The issue was successfully reproduced by running multiple redis-cli scripts executing commands for index creation (FT.CREATE), document addition/updates (HSET), document deletion (DEL), and search queries (FT.SEARCH) simultaneously, especially when combined with periodic index dropping (FT.DROPINDEX) and recreation.

**Steps to Reproduce:**
1. Create `writer1.sh`, `writer2.sh`, `reindexer.sh`, `deleter.sh`, `searcher.sh` scripts.   
- `writer1.sh`
```
#!/bin/bash
for i in $(seq 1 100000); do
  redis-cli HSET doc:$i content "text data item $i for race condition test" tags "tagA,tagB,$((i % 10))" numeric_field $i
done
```
- `writer2.sh`
```
#!/bin/bash
for i in $(seq 100001 200000); do
  redis-cli HSET doc:$i content "another set of text content $i" tags "tagC,tagD,$((i % 20))" numeric_field $i
done
```
- `searcher.sh`
```
#!/bin/bash
for i in $(seq 1 100000); do
  random_val=$((1 + RANDOM % 200000))
  random_tag_val=$((RANDOM % 20))
  redis-cli FT.SEARCH myRaceIdx "@content:item$random_val" > /dev/null
  redis-cli FT.SEARCH myRaceIdx "@tags:{tagA|tagC|tag$random_tag_val}" > /dev/null
  redis-cli FT.SEARCH myRaceIdx "@numeric_field:[$random_val $(($random_val + 100))]" > /dev/null
done
```
- `deleter.sh`
```
#!/bin/bash
for i in $(seq 1 50000); do
  doc_to_delete=$((1 + RANDOM % 200000))
  redis-cli DEL doc:$doc_to_delete > /dev/null
done
```
- `reindexer.sh`
```
#!/bin/bash
for i in $(seq 1 100); do
  sleep 1
  redis-cli FT.DROPINDEX myRaceIdx
  usleep 50000
  redis-cli FT.CREATE myRaceIdx ON HASH PREFIX 1 doc: SCHEMA content TEXT SORTABLE tags TAG SORTABLE numeric_field NUMERIC SORTABLE
done
```
[...]

PR #5185, opened 2025-05-27
fix: prevent crash on KEYLOCK_ACQUIRED check for NO_KEY_TRANSACTIONAL commands

vyavdoshenko:
Fixes: https://github.com/dragonflydb/dragonfly/issues/5173

Problem:
Server crashes with `Check failed: !lock_args.fps.empty()` during concurrent search indexing operations (FT.CREATE, FT.DROPINDEX, FT.SEARCH, etc.).

Root Cause:
Commands with `CO::NO_KEY_TRANSACTIONAL` flag don't have keys to lock, resulting in empty fingerprints (`kv_fp_`). However, the transaction system:
- Always sets `KEYLOCK_ACQUIRED` flag in `ScheduleInShard()`
- Expects this flag when releasing locks in `CancelShardCb()`
- Attempts to release locks on empty fingerprint arrays

Solution:
 Add proper conditional checks in `CancelShardCb()`


src/server/transaction.cc
```diff
@@ -1246,8 +1246,27 @@ bool Transaction::CancelShardCb(EngineShard* shard) {
...
+    // 1. FT.CREATE/FT.DROPINDEX use CO::GLOBAL_TRANS flag, triggering multi-transaction mode
+    // 2. During concurrent operations (multiple redis-cli scripts running FT.DROPINDEX + HSET +
+    // FT.SEARCH)
+    // 3. MultiSwitchCmd is called via ConnectionContext::SwitchTxCmd() or MultiCommandSquasher
+    // 4. MultiSwitchCmd calls kv_fp_.clear()
+    // 5. Later, CancelShardCb tries to release locks but GetLockArgs returns empty fps
+    //
+    // This is a valid state in multi-transactions and we should skip lock release in such cases.
+    if (!lock_args.fps.empty()) {
+      GetDbSlice(shard->shard_id()).Release(LockMode(), lock_args);
```
kostasrim, line 1264: How is this not a bug ? When we start a transaction we `unconditionally` register the locks. So for example:

```
  1123   if (!IsGlobal()) {                                                                              
  1124     lock_args = GetLockArgs(shard->shard_id());
  1125     const bool shard_unlocked = shard->shard_lock()->Check(mode);                                 
  1126                                                                                                   
  1127     // We need to acquire the fp locks because the executing callback                             
  1128     // within RunCallback below might preempt.
  1129     const bool keys_unlocked = GetDbSlice(shard->shard_id()).Acquire(mode, lock_args); 
```

`lock_args` are used to acquire the related locks for the transaction lock table within db_slice. 

If we call `kv_fp_.clear()` without calling `Release` we just shoot ourselves in the foot no ? Because we just left the corresponding locks incremented in the transaction table of dbslice. 


Furthermore, I would suggest to add a test for this case

romange: it should be part of the fixing PR if it possible to do

tests/dragonfly/search_test.py
```diff
@@ -558,3 +562,241 @@ def make_car(producer, description, speed):
...
+    # Index name for the test
+    index_name = "myRaceIdx"
+
+    # Shared flags
+    server_crashed = threading.Event()
+    stop_threads = threading.Event()
+
+    def writer_thread():
+        """Thread that writes documents"""
+        client = redis.Redis(port=df_server.port, decode_responses=True)
```
romange, line 586:
```suggestion
        client = df_server.client()
```

tests/dragonfly/search_test.py
```diff
@@ -558,3 +562,241 @@ def make_car(producer, description, speed):
...
+
+@dfly_args({"proactor_threads": 4})
+async def test_search_race_condition_threaded_issue_5173(async_client: aioredis.Redis, df_server):
+    """
+    Alternative test using threading for true parallelism to reproduce race condition.
+    This version uses actual threads instead of asyncio tasks for maximum concurrency.
+    """
+    import threading
+    import time
+    import redis  # Sync redis client for threading
```
romange, line 575: it does not have to be.
you can have async functions running in threads using the following helper function:
```py
def run_in_thread(task):
        loop = asyncio.new_event_loop()
        asyncio.set_event_loop(loop)
         try:
           loop.run_until_complete(task)
       finally:
           loop.close() 
```
and then launch a thread with: `thread1 = threading.Thread(target=run_in_thread, args=(task1,), name="Thread-1")` etc

tests/dragonfly/search_test.py
```diff
@@ -558,3 +562,241 @@ def make_car(producer, description, speed):
...
+                except (redis.ConnectionError, OSError):
+                    server_crashed.set()
+                    break
+                except Exception:
+                    # Expected during index operations
+                    time.sleep(0.005)
+        finally:
+            client.close()
+
+    def writer2_thread():
```
romange, line 686: you do not need to clone the same function twice, you can just launch a second thread with the same function.

tests/dragonfly/search_test.py
```diff
@@ -558,3 +562,241 @@ def make_car(producer, description, speed):
...
+    """
+    import threading
+    import time
+    import redis  # Sync redis client for threading
+
+    # Index name for the test
+    index_name = "myRaceIdx"
+
+    # Shared flags
+    server_crashed = threading.Event()
```
romange, line 581: no need for `server_crashed` - you can just reuse `stop_threads` for that

tests/dragonfly/search_test.py
romange: where task1 is an `async def`

tests/dragonfly/search_test.py
```diff
@@ -558,3 +562,241 @@ def make_car(producer, description, speed):
...
+
+        # Wait for threads to finish
+        for thread in threads:
+            thread.join(timeout=5.0)
+
+    except Exception as e:
+        stop_threads.set()
+
+    # Check if server crashed
+    if server_crashed.is_set():
```
romange, line 799: if the server crashed, the testing framework should identify it automatically.

src/server/transaction.cc
vyavdoshenko: Remove the Python test?

romange: you can leave it for now, it's a good baseline but I think if we can reproduce it in the unit test - it's 10 times better :)

romange: and it helped me to understand the problem

vyavdoshenko: OK, I got it.
I will try to fix it.

src/server/search/search_family_test.cc
```diff
@@ -2767,4 +2767,81 @@ TEST_F(SearchFamilyTest, JsonSetIndexesBug) {
...
+       "SORTABLE", "tags", "TAG", "SORTABLE", "numeric_field", "NUMERIC", "SORTABLE"});
+
+  // 2. writer_fiber
+  auto writer_fiber = pp_->at(0)->LaunchFiber([&] {
+    for (int i = 1; i <= kWriterOps; ++i) {
+      std::string doc_key = absl::StrCat("doc:", i);
+      std::string content = absl::StrCat("text data item ", i, " for race condition test");
+      std::string tags_val = absl::StrCat("tagA,tagB,", (i % 10));
+      std::string numeric_field_val = std::to_string(i);
+      try {
```
romange, line 2789: why do you need try/catch?

src/server/transaction.cc
romange: yeah I know and that's by design. the bug is in rollback

romange:
pls rebase as well

src/server/transaction.cc
vyavdoshenko: @romange 
What do you mean? Should I process NO_KEY_TRANSACTIONAL commands in a different way?
Should not KEYLOCK_ACQUIRED be set for NO_KEY_TRANSACTIONAL commands?

src/server/search/search_family_test.cc
```diff
@@ -2768,6 +2768,69 @@ TEST_F(SearchFamilyTest, JsonSetIndexesBug) {
...
+
+      std::string query_content = absl::StrCat("@content:item", random_val_content);
+      std::string query_tags = absl::StrCat("@tags:{tagA|tagB|tag", random_tag_val, "}");
+      std::string query_numeric = absl::StrCat("@numeric_field:[", random_val_numeric, " ",
+                                               (random_val_numeric + 100), "]");
+      Run({"ft.search", kIndexName, query_content});
+      Run({"ft.search", kIndexName, query_tags});
+      Run({"ft.search", kIndexName, query_numeric});
+      if (i % 50 == 0)
+        ThisFiber::SleepFor(std::chrono::microseconds(200 * (1 + i % 2)));
```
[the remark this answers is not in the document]
vyavdoshenko: I removed all sleeps. The bug is reproducible without them.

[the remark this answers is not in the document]
romange: even better.

src/server/transaction.cc
```diff
@@ -1245,9 +1245,16 @@ bool Transaction::CancelShardCb(EngineShard* shard) {
...
-    GetDbSlice(shard->shard_id()).Release(LockMode(), lock_args);
+
+    if (lock_args.fps.empty()) {
+      // For NO_KEY_TRANSACTIONAL commands, we don't need to release locks
+      // because we don't acquire locks for them.
+      DCHECK(cid_->opt_mask() & CO::NO_KEY_TRANSACTIONAL);
+    } else {
+      DCHECK(sd.local_mask & KEYLOCK_ACQUIRED);
+      GetDbSlice(shard->shard_id()).Release(LockMode(), lock_args);
+    }
```
romange, line 1256: Almost. I think the condition should be reversed:

```suggestion
    if (cid_->opt_mask() & CO::NO_KEY_TRANSACTIONAL) == 0) {
      auto lock_args = GetLockArgs(shard->shard_id());
      DCHECK(sd.local_mask & KEYLOCK_ACQUIRED);
      DCHECK(!lock_args.fps.empty());
      GetDbSlice(shard->shard_id()).Release(LockMode(), lock_args);
    }
```

src/server/search/search_family_test.cc
```diff
@@ -2768,6 +2768,47 @@ TEST_F(SearchFamilyTest, JsonSetIndexesBug) {
   EXPECT_THAT(resp, IsUnordArrayWithSize(IsMap("text", "some text")));
 }
 
+TEST_F(SearchFamilyTest, SearchReindexWriteSearchRace) {
+  const std::string kIndexName = "myRaceIdx";
+  const int kWriterOps = 1000;
```
romange, line 2773: it reproduces for me with all constants be 100

src/server/transaction.cc
```diff
@@ -1245,9 +1245,16 @@ bool Transaction::CancelShardCb(EngineShard* shard) {
...
+
     sd.local_mask &= ~KEYLOCK_ACQUIRED;
```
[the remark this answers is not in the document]
vyavdoshenko: I am afraid you can be wrong.
If I put it in the block, another DCHECK fails - transaction.cc:1105
It seems this DCHECK is correct, and during the canceling operation, we have to reset KEYLOCK_ACQUIRED always.

romange (approved):

PR #5192, opened 2025-05-28
fix(ci): duplicate upload of df_bench on release

kostasrim:
The issue: `df_bench` would be build twice for the same arch and it was not uploaded because of the conflict


kostasrim:
https://github.com/dragonflydb/dragonfly/actions/runs/15295360747

PR #5194, opened 2025-05-28
feat: add Lua GC flags

BorysTheDev:
I've added 2 flags:

- luagc - to tune lua gc
- lua_mem_usage_force_gc - to do force GC

I have done some tests with lua_mem_usage_force_gc, and it shows quite good results to my mind, and it works fast.
Regarding the luagc flag, I've tried to use different parameters, according to https://www.lua.org/manual/5.4/manual.html#2.5.1 , and the results were insignificant, but with aggressive parameters performance can be dropped significantly


src/core/interpreter.cc
```diff
@@ -34,6 +36,83 @@ LUALIB_API int(luaopen_bit)(lua_State* L);
...
+  union Args {
+    GenArgs gen;
+    IncArgs inc;
+    Args() {
+      gen = {};
+    }
+  } args;
+};
+
+ABSL_FLAG(LuaGcFlag, luagc, LuaGcFlag{},
```
romange, line 72: have you checked what is the default mode for lua gc?
also, please state the default here explicitly because it is hidden in c++ code.

romange: i think you changed the default for lua which may cause the degradation for the existing workloads.
please make sure that the default is to NOT do anything inside UpdateGCParameters.

BorysTheDev: The AI is right. I have checked manually. It is incremental

src/core/interpreter.cc
```diff
@@ -34,6 +36,83 @@ LUALIB_API int(luaopen_bit)(lua_State* L);
...
+
+  LuaGcFlag(GenArgs gen_args) : type(Gen) {
+    args.gen = gen_args;
+  }
+
+  LuaGcFlag(IncArgs inc_args) : type(Inc) {
+    args.inc = inc_args;
+  }
+
+  enum { Gen, Inc, Default } type = Default;
```
romange: you could achieve the same with variant i guess

src/core/interpreter.cc
```diff
@@ -34,6 +36,83 @@ LUALIB_API int(luaopen_bit)(lua_State* L);
...
+  } args;
+};
+
+ABSL_FLAG(LuaGcFlag, luagc, LuaGcFlag{},
+          "Specifies Lua garabage collector preferences. By default used default lua GC parameters."
+          "Format should be 'inc/200/100/13' or 'gen/20/100' where 'inc' and 'gen' are types of "
+          "GC, numbers are parameters."
+          "For more information check https://www.lua.org/manual/5.4/manual.html#2.5");
+
+ABSL_FLAG(uint64_t, lua_mem_usage_force_gc, 10000000,
```
romange: nit: `lua_mem_gc_threshold`

src/core/interpreter.cc
```diff
@@ -1109,6 +1209,16 @@ Interpreter* InterpreterManager::Get() {
 }
 
 void InterpreterManager::Return(Interpreter* ir) {
+  static const uint64_t max_memory_usage = absl::GetFlag(FLAGS_lua_mem_usage_force_gc);
+  using namespace chrono;
+  ++tl_stats().interpreter_return;
+  if (tl_stats().used_bytes > max_memory_usage) {
```
romange, line 1210: Consider the following scenario. We have a per thread limit of 100MB
1. Lua1 allocates 1MB, does not trigger gc.
2. .... Lua99 does not trigger GC and we have like 99MB returned.
3. Now we have bunch of scripts created and returned - they allocate small blobs but at some point they cross 100MB so they being GC'ed and the lua memory usage goes around 100MB because the already returned scripts are not GCed.

To summarize - we have a shared per-thread threshold, but we perform GC on lua we currently return that can have only 10KB of allocated memory and most of the allocations are kept in the dormant interperters.

romange: It's not that I have a proposal, but lets introduce a variable that tracks gc collected memory by counting `LUA_GCCOUNT` before and after RunGC. In fact, RunGC could return the amount of memory it released.
This way we will be able to estimate GC performance rate.

src/server/main_service.cc
```diff
@@ -821,6 +821,8 @@ void Service::Init(util::AcceptServer* acceptor, std::vector<facade::Listener*>
 
   config_registry.RegisterMutable("pipeline_squash");
 
+  config_registry.RegisterMutable("lua_mem_usage_force_gc");
```
romange: that's won't cut it and you have not tested that it works.
you used `static const uint64_t max_memory_usage = ...` so it will be initialised only once

src/core/interpreter.h
```diff
@@ -114,7 +114,7 @@ class Interpreter {
 
   void ResetStack();
 
-  void RunGC();
+  int64_t RunGC();
```
romange, line 118: can you add a comment regarding what it returns?

src/core/interpreter.cc
```diff
@@ -982,23 +972,23 @@ void Interpreter::ResetStack() {
   lua_settop(lua_, 0);
 }
 
-void Interpreter::RunGC() {
+int64_t Interpreter::RunGC() {
+  int64_t before = lua_gc(lua_, LUA_GCCOUNT);
```
romange: have you checked this code manually?
have you looked what LUA_GCCOUNT returns?

romange (changes requested):

src/core/interpreter.cc
BorysTheDev: LUA_GCCOUNT: Returns the current amount of memory (in Kbytes) in use by Lua.

BorysTheDev: What do you mean manually? I added the test to check this behavior or do you asked why I didn't add counting freed memory from the SCRIPT GC command?

src/server/server_family.cc
```diff
@@ -1419,8 +1419,16 @@ void PrintPrometheusMetrics(uint64_t uptime, const Metrics& m, DflyCmd* dfly_cmd
...
                             &resp->body());
+  AppendMetricWithoutLabels("freed_memory_lua", "", m.lua_stats.gc_freed_memory, MetricType::GAUGE,
+                            &resp->body());
   AppendMetricWithoutLabels("lua_blocked_total", "", m.lua_stats.blocked_cnt, MetricType::COUNTER,
                             &resp->body());
+  AppendMetricWithoutLabels("lua_gc_interpreter_return", "", m.lua_stats.interpreter_return,
+                            MetricType::COUNTER, &resp->body());
+  AppendMetricWithoutLabels("lua_force_gc_calls", "", m.lua_stats.force_gc_calls,
+                            MetricType::COUNTER, &resp->body());
+  AppendMetricWithoutLabels("lua_gc_work_time_ns", "", m.lua_stats.gc_work_time_ns,
```
romange: please note that prometheus have conventions with naming and units. Please see how we name other time metrics and note their units.

src/server/server_family.cc
```diff
@@ -2563,6 +2571,11 @@ string ServerFamily::FormatInfoMetrics(const Metrics& m, std::string_view sectio
 
     // Total number of events of when a connection was blocked on grabbing interpreter.
     append("lua_blocked_total", m.lua_stats.blocked_cnt);
+
+    append("lua_interpreter_return", m.lua_stats.interpreter_return);
+    append("lua_force_gc_calls", m.lua_stats.force_gc_calls);
+    append("lua_gc_freed_memory", m.lua_stats.gc_freed_memory);
+    append("lua_gc_work_time", m.lua_stats.gc_work_time_ns);
```
romange: `lua_gc_work_time` units are not clear. See how other time units are named.

PR #5199, opened 2025-05-30
[WIP] (string_family): Add support for MC GAT

abhijat:
Support for GAT command is added in this change set.

The memcache protocol commands GAT and GATS will use the same code path as MC GET/GETS, and also touch expiry of keys which are found using the supplied expiry. 

The MGET code path is changed to optionally set expiry of keys it finds.

### DIFF
--- a/src/server/string_family.cc
+++ b/src/server/string_family.cc
@@ -533,65 +533,163 @@ struct MGetResponse {
   absl::InlinedVector<std::optional<GetResp>, 2> resp_arr;
 };
 
-// fetch_mask values
-constexpr uint8_t FETCH_MCFLAG = 0x1;
-constexpr uint8_t FETCH_MCVER = 0x2;
-MGetResponse OpMGet(fb2::BlockingCounter wait_bc, uint8_t fetch_mask, const Transaction* t,
-                    EngineShard* shard) {
-  ShardArgs keys = t->GetShardArgs(shard->shard_id());
-  DCHECK(!keys.Empty());
+struct FoundIterator {
+  std::variant<DbSlice::ConstIterator, DbSlice::Iterator> it;
+  int source_index = -1;  // in case of duplicate keys, points to the first occurrence.
 
-  auto& db_slice = t->GetDbSlice(shard->shard_id());
+  bool Done() const {
+    return std::visit([](auto& i) { return i.is_done(); }, it);
+  }
 
-  MGetResponse response(keys.Size());
-  struct Item {
-    DbSlice::ConstIterator it;
-    int source_index = -1;  // in case of duplicate keys, points to the first occurrence.
-  };
+  const PrimeKey& Key() const {
+    return std::visit([](auto& i) -> const CompactObj& { return i->first; }, it);
+  }
 
-  absl::InlinedVector<Item, 32> items(keys.Size());
+  const PrimeValue& Value() const {
+    return std::visit([](auto& i) -> const CompactObj& { return i->second; }, it);
+  }
 
-  // First, fetch all iterators and count total size ahead
-  size_t total_size = 0;
-  unsigned index = 0;
-  static bool mget_dedup_keys = absl::GetFlag(FLAGS_mget_dedup_keys);
+  uint64_t Version() const {
+    return std::visit([](auto& i) { return i.GetVersion(); }, it);
+  }
 
-  // We can not make it thread-local because we may preempt during the Find loop due to
-  // replication of expiry events.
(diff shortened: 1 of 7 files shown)


src/server/string_family.cc
```diff
@@ -572,14 +586,36 @@ MGetResponse OpMGet(fb2::BlockingCounter wait_bc, uint8_t fetch_mask, const Tran
       }
     }
 
-    auto it_res = db_slice.FindReadOnly(t->GetDbContext(), key, OBJ_STRING);
+    const bool found = expire_params.has_value() &&
+                       ExpireKey(shard, db_slice, expire_params.value(), key, op_args, ctx);
```
abhijat, line 590: this extra lookup can be avoided if a generic iterator type is stored in `DedupItem`

src/server/string_family.cc
```diff
@@ -1294,22 +1361,39 @@ void StringFamily::DecrBy(CmdArgList args, const CommandContext& cmnd_cntx) {
...
+std::optional<DbSlice::ExpireParams> ParseExpireParams(CmdArgList& args) {
+  std::optional<DbSlice::ExpireParams> expire_params = std::nullopt;
+  if (args.size() > 2) {
+    if (args[args.size() - 2] == "EXAT") {
+      int64_t expiry;
+      absl::SimpleAtoi(args[args.size() - 1], &expiry);
+      // Memcache expire time is absolute seconds
+      expire_params =
+          DbSlice::ExpireParams{.value = expiry, .absolute = true, .persist = expiry == 0};
+      args = args.subspan(0, args.size() - 2);
```
abhijat, line 1373: This is incorrect because the callback works on shard args, not this list. Also the expiry we add `EXAT ...` gets added to shard args. In `MGet` we use shard args to construct the result, which causes a mismatch.

It might be simpler to store the expiry in the connection context like the memcache flag is stored, instead of passing it as extra args to mget command.

romange:
I suggest that you split this PR into a "trivial" part  that handles, parses the GAT command, parses the ttl but ignores the "touch" part and basically still calls MGET as is.

The second part - of actually touching should be a separate PR and we should talk about it. I did not estimate correctly the complexity involved with this part. I would argue it's not MGET - it's more LIKE "SET with GET" option - because it's writeable https://www.dragonflydb.io/docs/command-reference/strings/set. So let's merge the trivial part first.

src/server/string_family.cc
```diff
@@ -533,94 +533,193 @@ struct MGetResponse {
...
+                    const MemcacheMGetParams& mc_params) {
+  const ShardArgs keys = t->GetShardArgs(shard->shard_id());
+  DCHECK(!keys.Empty());
+
+  auto& db_slice = t->GetDbSlice(shard->shard_id());
+
+  MGetResponse response(keys.Size());
+
+  auto [items, key_index, total_size] =
+      FindAndMaybeExpireKeys(t, shard, keys, db_slice, mc_params.expire_params());
```
dranikpg, line 681: MGET is readonly, we can't just casually delete a few values in case someone sets a negative (i.e past) expire time 🤔

abhijat:
> I suggest that you split this PR into a "trivial" part that handles, parses the GAT command, parses the ttl but ignores the "touch" part and basically still calls MGET as is.
> 
> The second part - of actually touching should be a separate PR and we should talk about it. I did not estimate correctly the complexity involved with this part. I would argue it's not MGET - it's more LIKE "SET with GET" option - because it's writeable https://www.dragonflydb.io/docs/command-reference/strings/set. So let's merge the trivial part first.

created https://github.com/dragonflydb/dragonfly/pull/5229

FWIW it looks like much of the code will remain similar for GAT as MGET, but the iterator type for the search and the command flag is different. 

If a new command has to be added, it might make sense to reuse code from the MGET code path.

src/server/string_family.cc
abhijat: I see, thanks. I added a separate command id for `GAT` which has a `CO::Write` flag, it reuses the `MGet` method, but doesn't invoke the same command as `MGET`, but it looks like it is not safe to reuse the method?

dranikpg: Ah, didn't notice that you added it. If so, it is, but I'd suggest to document this place properly and maybe move the conditional to the top, so we can see it's not invoced for the usual code path

romange:
is it stil relevant @abhijat ?

abhijat:
> is it stil relevant @abhijat ?

this is not required any more, I will close this

Issue #5202, opened 2025-05-31
XDEL command causes crash


vyavdoshenko:
The scenario:
```
XADD "k1 \"v1\" k2 \"v2 with spaces\" \"k3 with spaces\" \"v3\"" 0 "set1" "member1"
XDEL "k1 \"v1\" k2 \"v2 with spaces\" \"k3 with spaces\" \"v3\"" "46-867" 
```

The crash log:
```
F20250531 15:35:53.395673 327389 stream_family.cc:44] Check failed: pv.MallocUsed() != 0 
*** Check failure stack trace: ***
    @     0x5bca9341fd23  google::LogMessage::Fail()
    @     0x5bca9341fc69  google::LogMessage::SendToLog()
    @     0x5bca9341f41c  google::LogMessage::Flush()
    @     0x5bca93423514  google::LogMessageFatal::~LogMessageFatal()
    @     0x5bca929557d5  dfly::StreamMemTracker::UpdateStreamSize()
    @     0x5bca9295db7f  dfly::(anonymous namespace)::OpDel()
    @     0x5bca92967953  _ZZN4dfly12StreamFamily4XDelEN4absl12lts_202407224SpanIKSt17basic_string_viewIcSt11char_traitsIcEEEERKNS_14CommandContextEENKUlPNS_11TransactionEPNS_11EngineShardEE_clESE_SG_
    @     0x5bca9297273d  _ZZN4dfly11Transaction18ScheduleSingleHopTIRZNS_12StreamFamily4XDelEN4absl12lts_202407224SpanIKSt17basic_string_viewIcSt11char_traitsIcEEEERKNS_14CommandContextEEUlPS0_PNS_11EngineShardEE_EEDTclfp_fpTLDnEEEOT_ENKUlSF_SH_E_clESF_SH_
    @     0x5bca9298493e  _ZSt13__invoke_implIN6facade8OpStatusERKZN4dfly11Transaction18ScheduleSingleHopTIRZNS2_12StreamFamily4XDelEN4absl12lts_202407224SpanIKSt17basic_string_viewIcSt11char_traitsIcEEEERKNS2_14CommandContextEEUlPS3_PNS2_11EngineShardEE_EEDTclfp_fpTLDnEEEOT_EUlSI_SK_E_JSI_SK_EESO_St14__invoke_otherOT0_DpOT1_
[...]

PR #5203, opened 2025-05-31
fix: crash in StreamMemTracker::UpdateStreamSize() by removing invalid DCHECK

vyavdoshenko:
Fixes: https://github.com/dragonflydb/dragonfly/issues/5202

The issue:
The `DCHECK(pv.MallocUsed() != 0)` in `StreamMemTracker::UpdateStreamSize()` was causing crashes when `MallocUsed()` returned 0, which can be a legitimate value in specific scenarios (e.g., when attempting to delete non-existent stream entries).

The solution:
Removed the problematic DCHECK as `MallocUsed()` returning 0 is a valid state that should not trigger an assertion failure.

### DIFF
--- a/src/server/stream_family.cc
+++ b/src/server/stream_family.cc
@@ -40,8 +40,6 @@ void StreamMemTracker::UpdateStreamSize(PrimeValue& pv) const {
   const size_t current = zmalloc_used_memory_tl;
   int64_t diff = static_cast<int64_t>(current) - static_cast<int64_t>(start_size_);
   pv.AddStreamSize(diff);
-  // Under any flow we must not end up with this special value.
-  DCHECK(pv.MallocUsed() != 0);
 }
 
 namespace {
--- a/src/server/stream_family_test.cc
+++ b/src/server/stream_family_test.cc
@@ -1291,7 +1291,7 @@ TEST_F(StreamFamilyTest, XDelCrash) {
   string key_name = "k1";
 
   auto resp_xadd = Run({"xadd", key_name, "0", "set1", "member1"});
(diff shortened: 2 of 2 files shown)


src/server/stream_family.cc
```diff
@@ -40,8 +40,6 @@ void StreamMemTracker::UpdateStreamSize(PrimeValue& pv) const {
   const size_t current = zmalloc_used_memory_tl;
   int64_t diff = static_cast<int64_t>(current) - static_cast<int64_t>(start_size_);
   pv.AddStreamSize(diff);
-  // Under any flow we must not end up with this special value.
```
kostasrim, line 43: `pv.MallocUsed()` should not be `zero`. This is a bug in memory tracking and *not in the DCHECK*. 

no prime value with a memory tracker should have size of zero -- even if it's an empty data structure we should still count the `pointer we store`. 

For example:

```
   694   StreamMemTracker mem_tracker;
   695 
   696   if (add_res.is_new) {
   697     stream* s = streamNew();
```

cc @BagritsevichStepan

BagritsevichStepan: yes, the bug is not here. I suggest to assign me on this task

BagritsevichStepan: it is a small bug but better to check other commands and add memory tests

kostasrim: @BagritsevichStepan I agree

vyavdoshenko: @BagritsevichStepan 
https://github.com/dragonflydb/dragonfly/issues/5202 is assigned to you

Issue #5206, opened 2025-05-31
Race condition on recreation of search index and calling INFO command that leads to crash


vyavdoshenko:
The bash script to reproduce the issue:
```
#!/bin/bash

index_ops() {
    redis-cli FT.CREATE idx$1 ON HASH PREFIX 1 prefix$1: 2>/dev/null &
    sleep 0.01
    redis-cli FT.DROPINDEX idx$1 2>/dev/null &
}

info_ops() {
    for i in {1..10}; do
        redis-cli INFO > /dev/null 2>&1 &
    done
}

for i in {1..5}; do
    index_ops $i
done

info_ops

wait
```

The crash log:
```
F20250531 21:28:11.978102 19754 common.cc:273] Check failed: num_indices == 0 || num_indices == o.num_indices 
*** Check failure stack trace: ***
F00000000 00:00:00.000000 19747 logging.cc:1828] RAW: Check data_->num_chars_to_log_ > 0 && data_->message_text_[data_->num_chars_to_log_-1] == '\n' failed: 
    @     0x63a6a1160d23  google::LogMessage::Fail()
    @     0x63a6a1160d23  google::LogMessage::Fail()
    @     0x63a6a1169a45  google::RawLog__()
    @     0x63a6a1160c69  google::LogMessage::SendToLog()
    @     0x63a6a1160645  google::LogMessage::SendToLog()
    @     0x63a6a116041c  google::LogMessage::Flush()
    @     0x63a6a116041c  google::LogMessage::Flush()
    @     0x63a6a1164514  google::LogMessageFatal::~LogMessageFatal()
    @     0x63a6a095a9c7  dfly::SearchStats::operator+=()
    @     0x63a6a05875be  _ZZNK4dfly12ServerFamily10GetMetricsEPNS_9NamespaceEENKUljPN4util3fb212ProactorBaseEE_clEjS6_
    @     0x63a6a1164514  google::LogMessageFatal::~LogMessageFatal()
    @     0x63a6a059a23e  _ZZN4util12ProactorPool15AwaitFiberOnAllIZNK4dfly12ServerFamily10GetMetricsEPNS2_9NamespaceEEUljPNS_3fb212ProactorBaseEE_Li0EEEvOT_ENUljS8_E_clEjS8_
[...]

PR #5207, opened 2025-05-31
fix: SearchStats race condition crash during concurrent FT operations

vyavdoshenko:
fixes: https://github.com/dragonflydb/dragonfly/issues/5206

The issue:
SearchStats::operator+=() crashed with `DCHECK` failure during concurrent FT.CREATE/FT.DROPINDEX operations and INFO commands. The strict assertion `DCHECK(num_indices == 0 || num_indices == o.num_indices)` failed when different shards temporarily had different numbers of search indices during race conditions.

The root cause:
Race condition between:
- FT.CREATE/FT.DROPINDEX operations (modifying index counts)  
- INFO commands (aggregating SearchStats across shards)

Different shards could have inconsistent `num_indices` values during concurrent operations, violating the strict equality check.

The solution:
- Removed the strict `DCHECK` assertion
- Existing `std::max(num_indices, o.num_indices)` logic already handles this correctly
- Added regression test `SearchStatsInfoRace` to prevent future issues

### DIFF
--- a/src/server/common.cc
+++ b/src/server/common.cc
@@ -270,6 +270,9 @@ SearchStats& SearchStats::operator+=(const SearchStats& o) {
   ADD(used_memory);
   ADD(num_entries);
 
+  // Different shards could have inconsistent num_indices values during concurrent operations.
+  // This can happen on concurrent index creation.
+  // We use max to ensure that the total num_indices is the maximum of all shards.
   num_indices = std::max(num_indices, o.num_indices);
   return *this;
 }


src/server/common.cc
```diff
@@ -270,7 +270,6 @@ SearchStats& SearchStats::operator+=(const SearchStats& o) {
   ADD(used_memory);
   ADD(num_entries);
 
-  DCHECK(num_indices == 0 || num_indices == o.num_indices);
```
romange, line 273: Please add a comment saying about different shards having different number of indices due to concurrent create index/drop index operations

vyavdoshenko: fixed

src/server/common.cc
```diff
@@ -270,7 +270,9 @@ SearchStats& SearchStats::operator+=(const SearchStats& o) {
...
+  // Different shards could have inconsistent num_indices values during concurrent operations.
```
kostasrim, line 273: yep, that's eventual consistency

src/server/common.cc
```diff
@@ -270,7 +270,9 @@ SearchStats& SearchStats::operator+=(const SearchStats& o) {
...
+  // This can happen on concurrent index creation.
+  // We use max to ensure that the total num_indices is the maximum of all shards.
```
kostasrim, line 275: That's a heuristic, we could also be deleting indices which upon this case it would be min. But it's fine -- this is the correct approach :)

kostasrim (approved):
LGTM

PR #5218, opened 2025-06-03
feat: Add eviction based on rss memory. Another solution

BagritsevichStepan:
This is a slightly simpler solution than #4991 . But sometimes it can remove a bit more keys than necessary


BagritsevichStepan:
PR is closed, this basic approach is not working

PR #5220, opened 2025-06-03
fix: seeder hset commands generation

BorysTheDev:
During the  test_seeder_key_target we have a lot of warnings like "HINCRBY k-s32u3-105 5 1 failed with reason: hash value is not an integer"

The problem is that we generate random strings as fields of HSET, not integers.

### DIFF
--- a/tests/dragonfly/seeder/script-genlib.lua
+++ b/tests/dragonfly/seeder/script-genlib.lua
@@ -131,11 +131,7 @@ end
 
 function LG_funcs.mod_hash(key, keys)
     local idx = math.random(LG_funcs.csize)
-    if idx % 2 == 1 then
-        redis.apcall('HINCRBY', key, tostring(idx), 1)
-    else
-      redis.apcall('HSET', key, tostring(idx), randstr())
-    end
+    redis.apcall('HSET', key, tostring(idx), randstr())
 end
 
 -- sorted sets


tests/dragonfly/seeder/script-genlib.lua
```diff
@@ -131,11 +131,7 @@ end
 
 function LG_funcs.mod_hash(key, keys)
     local idx = math.random(LG_funcs.csize)
-    if idx % 2 == 1 then
-        redis.apcall('HINCRBY', key, tostring(idx), 1)
-    else
-      redis.apcall('HSET', key, tostring(idx), randstr())
-    end
+    redis.apcall('HSET', key, tostring(idx), randstr())
```
kostasrim, line 134: don't we still want to call HINCRBY just only for members that are integers ? What was the reason we wanted to use HINCRBY in the seeder ? Do we have a test that actually needs this maybe ? ( I don't remember that's why I am asking)

BorysTheDev: we don't have them

kostasrim (approved):

tests/dragonfly/seeder/script-genlib.lua
kostasrim: yes you are right -- it was introduced here https://github.com/dragonflydb/dragonfly/commit/75eaeb32db652339729667b57b9c8095ef5eb648#diff-35fc27c757d9b3830899712b367143f5c61f801b8f252b7bfa6e59bfbde6da49R101

But there doesn't seem there exists a specific reason :smile:

Issue #5222, opened 2025-06-03
Unknown status error in op_status.cc


vyavdoshenko:
The following error can happen while Dragonfly is working.
```
E20250603 10:07:18.244894  1869 op_status.cc:42] Unsupported status 1
```

The solution:
Revise the code and correct errors and places where they are handled.

PR #5223, opened 2025-06-03
fix: Unsupported status for Copy/Move/Rename operations

vyavdoshenko:
Fixes: https://github.com/dragonflydb/dragonfly/issues/5222

Added missing handling of OpStatus for Copy/Rename/Move operations.

### DIFF
--- a/src/server/generic_family.cc
+++ b/src/server/generic_family.cc
@@ -1646,7 +1646,17 @@ void GenericFamily::Move(CmdArgList args, const CommandContext& cmd_cntx) {
 
 void GenericFamily::Rename(CmdArgList args, const CommandContext& cmd_cntx) {
   auto reply = RenameGeneric(args, false, cmd_cntx.tx);
-  cmd_cntx.rb->SendError(reply);
+
+  if (!reply.status) {
+    return cmd_cntx.rb->SendError(reply);
+  }
+
+  OpStatus st = reply.status.value();
+  if (st == OpStatus::OK) {
+    cmd_cntx.rb->SendOk();
+  } else {
+    cmd_cntx.rb->SendError(reply);
+  }
 }
 
 void GenericFamily::RenameNx(CmdArgList args, const CommandContext& cmd_cntx) {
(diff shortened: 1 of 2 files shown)


src/facade/facade.cc
```diff
@@ -124,6 +124,16 @@ const char kInvalidDumpValueErr[] = "DUMP payload version or checksum are wrong"
 const char kInvalidJsonPathErr[] = "invalid JSON path";
 const char kJsonParseError[] = "failed to parse JSON";
 const char kCrossSlotError[] = "-CROSSSLOT Keys in request don't hash to the same slot";
+const char kKeyExistsErr[] = "key already exists";
+const char kKeyMovedErr[] = "MOVED";
+const char kSkippedErr[] = "skipped";
+const char kInvalidValueErr[] = "invalid value";
```
kostasrim: I did not find this being used anywhere with grep

src/facade/facade.cc
```diff
@@ -124,6 +124,16 @@ const char kInvalidDumpValueErr[] = "DUMP payload version or checksum are wrong"
...
+const char kTimedOutErr[] = "timeout";
+const char kStreamIdSmallErr[] = "stream ID must be greater than last entry";
```
kostasrim: same here

src/facade/facade.cc
vyavdoshenko: Have you tried this on this branch?
I created the missing constants and moved them all to one place.
I ordered StatusToMsg to fit the statuses from the class.

Also, I checked - all the statuses are used in the code, and every status is handled in the StatusToMsg function.

src/facade/facade.cc
vyavdoshenko: Please take a look at the comment above.

src/server/generic_family.cc
```diff
@@ -1682,7 +1692,22 @@ void GenericFamily::Copy(CmdArgList args, const CommandContext& cmd_cntx) {
...
-  cmd_cntx.rb->SendError(renamer.Rename(!replace));
+  auto reply = renamer.Rename(!replace);
+
+  if (!reply.status) {
+    return cmd_cntx.rb->SendError(reply);
+  }
+
+  OpStatus st = reply.status.value();
+  if (st == OpStatus::OK) {
+    cmd_cntx.rb->SendLong(1);
```
kostasrim, line 1703: Nice, `COPY`  returns an int :)

src/server/generic_family.cc
```diff
@@ -1682,7 +1692,22 @@ void GenericFamily::Copy(CmdArgList args, const CommandContext& cmd_cntx) {
...
+  } else if (st == OpStatus::KEY_EXISTS) {
+    cmd_cntx.rb->SendLong(0);
+  } else if (st == OpStatus::KEY_NOTFOUND) {
+    cmd_cntx.rb->SendLong(0);
+  } else {
+    cmd_cntx.rb->SendError(reply);
```
kostasrim, line 1709: this should be unreachable

kostasrim, line 1709:
```suggestion
    cmd_cntx.rb->SendLong(0);
```

vyavdoshenko: I suggest leaving it. I'd like to get an error if an error occurs.

PR #5225, opened 2025-06-04
feat(streamer): Replicate journal items from different fiber

mkaruza:

### DIFF
--- a/src/server/journal/streamer.h
+++ b/src/server/journal/streamer.h
@@ -4,8 +4,6 @@
 
 #pragma once
 
-#include <deque>
-
 #include "server/cluster/slot_set.h"
 #include "server/common.h"
 #include "server/db_slice.h"
@@ -54,18 +52,22 @@ class JournalStreamer : public journal::JournalConsumerInterface {
     return cntx_->IsRunning();
   }
 
-  void WaitForInflightToComplete();
+  void WaitForPeriodicWriterToComplete();
 
   util::FiberSocketBase* dest_ = nullptr;
   ExecutionState* cntx_;
 
  private:
-  void AsyncWrite();
-  void OnCompletion(std::error_code ec, size_t len);
-
   bool IsStalled() const;
 
   journal::Journal* journal_;
+
+  util::fb2::Fiber periodic_writer_;
+  util::fb2::Done periodic_writer_done_;
+  void RunPeriodicWriterFiber();
+  void PeriodicWriterFiber(std::chrono::microseconds period_ms, util::fb2::Done* waiter);
+  void SendAsync(bool writer_done);
+  void OnCompletion(std::error_code ec, size_t len, bool writer_done);
 
   PendingBuf pending_buf_;
 
--- a/src/server/journal/pending_buf.h
+++ b/src/server/journal/pending_buf.h
@@ -8,61 +8,83 @@
 
 #include <deque>
 #include <numeric>
+#include <string_view>
 
 namespace dfly {
 
 class PendingBuf {
  public:
   struct Buf {
-    size_t mem_size = 0;
-    absl::InlinedVector<std::string, 8> buf;
-
 #ifdef UIO_MAXIOV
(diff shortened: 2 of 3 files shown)


src/server/journal/streamer.h
```diff
@@ -60,14 +59,25 @@ class JournalStreamer : public journal::JournalConsumerInterface {
...
+  void SendAsync(bool writer_done);
+  void OnCompletion(std::error_code ec, size_t len, bool writer_done);
+
+  using JournalStreamerItem = struct JournalStreamerItem {
+    size_t size;
+    std::string data;
+  };
+
+  static constexpr uint32_t in_flight_data_size_limit = 1024;
+  moodycamel::ReaderWriterQueue<JournalStreamerItem> queue_;
```
kostasrim: Assuming all producers and the single consumer run on the same thread (each of them on a separate fiber) why do we need the `atomics` ? If everything runs on the same thread concurrently then we do not need any synchronization primitive (unless we have partial writes from the producer to the consumer that must be satisfied fully -- upon on that case a counting barrier without atomic will suffice)

kostasrim: I know it's not ready for review or anything, just saw this and thought it would be worth commenting

mkaruza:
Create additional fiber in `JournalStreamer` that will be used for sending journal items for replication. Fiber waits with baseline of `10000 us` but if there is influx of journal items that period will be shorter by factor of *2*. 

**Testing with dfly_bench**


Both master and replicate nodes were running locally o 6 proactor threads and 8GB of RAM and dfly_bench was running as:

```
./dfly_bench --qps -100 --pipeline 150 -test_time 60 --ratio=1:0 -d 32 --proactor_threads=2 -c 5
```



```
~ NO REPLICATION (single node) ~

Total time: 5m0.010367772s. Overall number of requests: 45003000, QPS: 150010, P99 lat: 3497.56us
Latency summary, all times are in usec:
Count: 45003000 Average: 1844.1646  StdDev: 534958.76
Min: 10.0000  Median: 1791.4340  Max: 20280.0000

```

```
~ BASELINE REPLICATION ~

Total time: 5m0.010323251s. Overall number of requests: 45003000, QPS: 150010, P99 lat: 7763.2us
Latency summary, all times are in usec:
Count: 45003000 Average: 2846.6449  StdDev: 909460.22
Min: 17.0000  Median: 2605.8893  Max: 57202.0000
```

```
~ SPSC QUEUE REPLICATION ~

Total time: 5m0.01019038s. Overall number of requests: 45003000, QPS: 150010, P99 lat: 3551.86us
Latency summary, all times are in usec:
Count: 45003000 Average: 1622.9450  StdDev: 392480.80
Min: 11.0000  Median: 1565.2629  Max: 12383.0000
```

**PROFILING** 

|   | MainLoop |
| ------------- | ------------- |
| NO REPLICATION | 3.26%  |
| BASELINE REPLICATION  | 9.10%|
| SPSC REPLICATION  | 6.38% |

////

**_For more concrete test results there will be more testing needed._**

src/server/journal/streamer.h
dranikpg: If there aren't really any inter-thread accesses, then we don't (haven't looked at the code). It can (should) be a simple deque. If we do, we have queues in helio for this task. The imported queue isn't fiber-friendly and looks like just a large pile of overengineered C++ 03/11 code 🙂

mkaruza: As @kostasrim suggested - moving to use non atomic container.

kostasrim: @dranikpg I think this code is single threaded, so we don't need extra synchronization -- it suffice with a `vector` / `dequeue`. If not -- as you said helio is enough

@mkaruza double check what I am saying and plz don't rely on my words :smile: 

p.s.
@dranikpg happy to see you replying!

mkaruza:
Closing this draft. As discussed, replication improvement changes would be done gradually with few PRs.

PR #5229, opened 2025-06-04
(string_family): Add priliminary support for GAT

abhijat:
When the GAT command is supplied, store the expiry within the memcache flag, also add and set an enum value for setting expiry within the same flag.

We still call MGETS for now, and the expiry value is unused. A future PR will implement the changes for applying the expiry.

a continuation of https://github.com/dragonflydb/dragonfly/pull/5199

### DIFF
--- a/src/server/main_service.cc
+++ b/src/server/main_service.cc
@@ -1587,6 +1587,8 @@ void Service::DispatchMC(const MemcacheParser::Command& cmd, std::string_view va
     case MemcacheParser::PREPEND:
       strcpy(cmd_name, "PREPEND");
       break;
+    case MemcacheParser::GAT:
+      [[fallthrough]];
     case MemcacheParser::GET:
       [[fallthrough]];
     case MemcacheParser::GETS:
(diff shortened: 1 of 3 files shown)


src/server/main_service.cc
```diff
@@ -1646,6 +1648,10 @@ void Service::DispatchMC(const MemcacheParser::Command& cmd, std::string_view va
     if (cmd.type == MemcacheParser::GETS) {
       dfly_cntx->conn_state.memcache_flag |= ConnectionState::FETCH_CAS_VER;
     }
+
+    if (cmd.type == MemcacheParser::GAT) {
+      dfly_cntx->conn_state.memcache_flag |= ConnectionState::SET_EXPIRY | expire_ts << 2;
+    }
```
abhijat, line 1654: Another option instead of doing this would to to store the expiry using `EXAT ...` in the last two args as done for `SET` instead of stashing it in the state. 

A problem with that approach is that since the command is passed directly to `MGETS`, the last two args are processed as part of whatever shards they are assigned to, and the code becomes complicated because it has to skip searching for those two keys.

romange: Another approach would be to not call MGET, but instead introduce "MGAT" as a first-level citizen command.
I would not do this just for this but given the semantics of GAT, I do not think dispatching it to MGET is the right way forward.

I suggest to introduce the following Dragonfly command:
`MGAT <seconds> key1 key2 ... keyn` 
so here you will be able to map into it here.

it's NOT gonna be READONLY and will have mutable semantics. 
MGET is highly optimized but it's failry simple to add a naive implementation of MGAT that does not call MGET code.
I think this PR can be submitted as is though.

romange (approved):

Issue #5221, opened 2025-06-03
Master/Replica race condition Master crash on BF.RESERVE command


vyavdoshenko:
Master crashes on BF.RESERVE command on various loads with a replica connected.
Master must be started with `--proactor_threads=3` or more.

Useful script to reproduce the issue: [reproduce.txt](https://github.com/user-attachments/files/20566698/reproduce.txt)
Please change the extension before using.

Steps to reproduction:
- Start master
- Start replica
- Connect the replica to the master
- Execute the script on the master
- The master can shut down because of low memory, restart master and catch the crash

The crash log:
```
F20250603 12:56:32.854835 740084 journal_slice.cc:119] Check failed: allow_flush != enable_journal_flush_ 
*** Check failure stack trace: ***
F20250603 12:56:32.860749 740086 journal_slice.cc:119] Check failed: allow_flush != enable_journal_flush_ 
*** Check failure stack trace: ***
    @     0x5a145d7d8aa3  google::LogMessage::Fail()
    @     0x5a145d7d8aa3  google::LogMessage::Fail()
    @     0x5a145d7d89e9  google::LogMessage::SendToLog()
    @     0x5a145d7d819c  google::LogMessage::Flush()
    @     0x5a145d7dc294  google::LogMessageFatal::~LogMessageFatal()
    @     0x5a145d7d89e9  google::LogMessage::SendToLog()
    @     0x5a145cfea0e1  dfly::journal::JournalSlice::SetFlushMode()
    @     0x5a145cfe1651  dfly::journal::Journal::SetFlushMode()
    @     0x5a145c8c2229  dfly::journal::JournalFlushGuard::JournalFlushGuard()
    @     0x5a145cf5657d  dfly::DbSlice::FreeMemWithEvictionStepAtomic()
    @     0x5a145c8bb219  dfly::EngineShard::RetireExpiredAndEvict()
    @     0x5a145c8ba908  dfly::EngineShard::Heartbeat()
    @     0x5a145c8b7ec5  _ZZN4dfly11EngineShard27StartPeriodicHeartbeatFiberEPN4util3fb212ProactorBaseEENKUlvE_clEv
    @     0x5a145c8be1ee  _ZSt13__invoke_implIvRZN4dfly11EngineShard27StartPeriodicHeartbeatFiberEPN4util3fb212ProactorBaseEEUlvE_JEET_St14__invoke_otherOT0_DpOT1_
[...]

PR #5241, opened 2025-06-06
fix: dcheck trigger in JournalFlushGuard constructor

### LINKED ISSUE #5219: Stalled heartbeat() fiber for 6008371045 seconds
Sometimes we can see in the log the big value in seconds for a stalled heartbeat.

```
Stalled heartbeat() fiber for 6008371045 seconds
```

Possible reason: calculation overflow.

kostasrim:
The issue is that we nested `JournalFlushGuard` within `Heartbeat` which triggered the DCHECK. The constructor should only call `SetFlushMode(false)` only once when `counter_=0`. The same rules apply to the destructor which only called `SetFlushHMode` once (when `counter==0`)

resolves  #5219 #5221

### DIFF
--- a/src/server/engine_shard.cc
+++ b/src/server/engine_shard.cc
@@ -764,7 +764,8 @@ void EngineShard::Heartbeat() {
   if (db_slice.WillBlockOnJournalWrite() || !can_acquire_global_lock) {
     const auto elapsed = std::chrono::system_clock::now() - start;
     if (elapsed > std::chrono::seconds(1)) {
-      LOG_EVERY_T(WARNING, 5) << "Stalled heartbeat() fiber for " << elapsed.count() << " seconds";
+      LOG_EVERY_T(WARNING, 5) << "Stalled heartbeat() fiber for " << elapsed.count()
+                              << " nanoseconds";
     }
     return;
   }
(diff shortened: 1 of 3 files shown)


src/server/engine_shard.cc
```diff
@@ -764,7 +764,8 @@ void EngineShard::Heartbeat() {
   if (db_slice.WillBlockOnJournalWrite() || !can_acquire_global_lock) {
     const auto elapsed = std::chrono::system_clock::now() - start;
     if (elapsed > std::chrono::seconds(1)) {
-      LOG_EVERY_T(WARNING, 5) << "Stalled heartbeat() fiber for " << elapsed.count() << " seconds";
+      LOG_EVERY_T(WARNING, 5) << "Stalled heartbeat() fiber for " << elapsed.count()
```
vyavdoshenko: I believe it will look much better if it's converted to seconds.

kostasrim: yeah I agree!

tests/dragonfly/replication_test.py
```diff
@@ -3171,3 +3171,27 @@ async def test_replicate_hset_with_expiry(df_factory: DflyInstanceFactory):
 
     assert "name" in result
     assert result["name"] == "1234"
+
+
+async def test_bug_5221(df_factory):
```
kostasrim, line 3176: TOTAL overkill for testing `DCHECK` :facepalm: 

I can remove this @adiholden

vyavdoshenko (approved):
lgtm

Issue #5237, opened 2025-06-05
bitfield bugs


romange:
1.
```
> "SET" "key:bitfield_set" "\xff\xf0\x00"
> "BITFIELD" "key:bitfield_set" "OVERFLOW" "SAT" "SET" "i4" "0" "8" "SET" "i4" "4" "7"
1) (integer) 15
2) (integer) 15

```
but expected -1, -1.

2.
```
"SET" "key:bitfield_incr" "\xff\xf0\x00"
> "BITFIELD" "key:bitfield_incr" "INCRBY" "u8" "0" "85" "INCRBY" "u8" "16" "170"
1) (integer) 85
2) (integer) 170
```
but expected (84, 170)

3. `"BITFIELD" "key:bitfield_set:wrong:args" "set" "I8" "0" "0"`
should throw: `(error) ERR Invalid bitfield type. Use something like i16 u8. Note that u64 is not supported but i64 is.` but it does not

4. `"BITFIELD" "key:bitfield_get:wrong:args" "get" "I8" "0"` should throw
`(error) ERR Invalid bitfield type. Use something like i16 u8. Note that u64 is not supported but i64 is.`
but it does not

kostasrim:
out of curiosity how did you find those @romange ?

romange:
You have these tests marked  as failed on each pr via the fakeredis check

PR #5242, opened 2025-06-06
fix: BITFIELD: overflow handling, type validation, and return values

vyavdoshenko:
Fixes: https://github.com/dragonflydb/dragonfly/issues/5237

- Fixed BITFIELD SET overflow saturation: Now correctly returns old values.
- Fixed BITFIELD INCRBY calculations: Corrected increment logic to return proper values.
- Fixed type validation: Now properly rejects uppercase type specifiers (e.g., "I8" instead of "i8").
- Fixed UIntOverflow wrap logic: Changed from modulo to bitwise AND for proper wrap-around behavior.

### DIFF
--- a/src/server/bitops_family.cc
+++ b/src/server/bitops_family.cc
@@ -786,14 +786,15 @@ ResultType Set::ApplyTo(Overflow ov, string* bitfield) {
     bytes.resize(last_byte_offset, 0);
   }
 
-  if (!HandleOverflow(ov)) {
-    return {};
-  }
-
   // First, extract the old value properly using the Get method
   Get get(attr_);
   auto old_value_result = get.ApplyTo(ov, &bytes);
   int64_t old_value = old_value_result ? *old_value_result : 0;
+
+  // Apply overflow handling to the new value
+  if (!HandleOverflow(ov)) {
+    return {};
+  }
 
   uint32_t lsb = attr_.offset + attr_.encoding_bit_size - 1;
 
@@ -966,6 +967,8 @@ nonstd::expected<CommonAttributes, string> ParseCommonAttr(CmdArgParser* parser)
   if (encoding.empty()) {
     return make_unexpected(kSyntaxErr);
   }
+
+  // Check case-sensitivity - only lowercase 'u' and 'i' are allowed
   if (encoding[0] == 'u') {
     parsed.type = EncodingType::UINT;
   } else if (encoding[0] == 'i') {
@@ -977,6 +980,16 @@ nonstd::expected<CommonAttributes, string> ParseCommonAttr(CmdArgParser* parser)
   }
 
   string_view bits = encoding.substr(1);
+
+  // Additional validation: check if bits part contains any invalid characters
+  for (char c : bits) {
+    if (!std::isdigit(c)) {
+      return make_unexpected(
+          "invalid bitfield type. use something like i16 u8. note that u64 is not supported but "
+          "i64 "
+          "is.");
+    }
+  }
 
   if (!absl::SimpleAtoi(bits, &parsed.encoding_bit_size)) {
     return make_unexpected(kSyntaxErr);
--- a/src/server/bitops_family.cc
+++ b/src/server/bitops_family.cc
@@ -958,6 +958,9 @@ OpResult<vector<ResultType>> StateExecutor::Execute(const CommandList& commands)
   return results;
 }
 
+const char kInvalidBitfieldTypeErr[] =
+    "invalid bitfield type. use something like i16 u8. note that u64 is not supported but i64 is.";
+
 nonstd::expected<CommonAttributes, string> ParseCommonAttr(CmdArgParser* parser) {
   CommonAttributes parsed;
   using nonstd::make_unexpected;
@@ -974,9 +977,7 @@ nonstd::expected<CommonAttributes, string> ParseCommonAttr(CmdArgParser* parser)
   } else if (encoding[0] == 'i') {
     parsed.type = EncodingType::INT;
   } else {
-    return make_unexpected(
-        "invalid bitfield type. use something like i16 u8. note that u64 is not supported but i64 "
-        "is.");
+    return make_unexpected(kInvalidBitfieldTypeErr);
   }
 
   string_view bits = encoding.substr(1);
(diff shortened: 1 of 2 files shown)


src/server/bitops_family_test.cc
```diff
@@ -822,4 +822,37 @@ TEST_F(BitOpsFamilyTest, BitFieldLargeOffset) {
...
+  ASSERT_THAT(Run({"bitfield", "key:bitfield_set:wrong:args", "set", "I8", "0", "0"}),
+              expected_error);
+}
+
+TEST_F(BitOpsFamilyTest, BitFieldIssue5237_InvalidTypeUppercase_Get) {
+  auto expected_error = ErrArg(
+      "ERR invalid bitfield type. use something like i16 u8. note that u64 is not supported but "
+      "i64 is.");
+
+  ASSERT_THAT(Run({"bitfield", "key:bitfield_get:wrong:args", "get", "I8", "0"}), expected_error);
```
romange, line 855: @vyavdoshenko  please see below test_bitmap_commands.test_bitfield_get_wrong_arguments
and test_bitmap_commands.test_bitfield_set_wrong_arguments

src/server/bitops_family.cc
```diff
@@ -635,7 +635,7 @@ bool Overflow::UIntOverflow(int64_t incr, size_t total_bits, int64_t* value) con
     switch (type) {
       case Overflow::WRAP:
         // safe to do, won't overflow, both incr and value are <= than 2^63 - 1
-        *value = (incr_value + *value) % max;
+        *value = (incr_value + *value) & max;
```
kostasrim, line 638: nice catch!

src/server/bitops_family.cc
```diff
@@ -790,17 +790,19 @@ ResultType Set::ApplyTo(Overflow ov, string* bitfield) {
     return {};
   }
 
+  // First, extract the old value properly using the Get method
+  Get get(attr_);
```
kostasrim: This adds extra overhead on the bit calculations (it doubles them). Why not instead, see what the bug is with the current algorithm and fix that instead ?

src/server/bitops_family_test.cc
vyavdoshenko: @romange 
What do you think, should I leave these tests? Or should Python tests be used, and can I remove these unit tests?

romange: no, we should always use unit tests if possible.

src/server/bitops_family.cc
```diff
@@ -779,31 +779,69 @@ class Set {
...
   }
 
-  uint32_t lsb = attr_.offset + attr_.encoding_bit_size - 1;
+  string& bytes = *bitfield;
+  const size_t start_byte = GetByteIndex(attr_.offset);
+  const size_t end_byte = GetByteIndex(attr_.offset + attr_.encoding_bit_size - 1);
+  const size_t initial_byte_size = bytes.size();
+
+  // Buffer to hold original bytes.
+  std::vector<uint8_t> old_bytes_buf;
```
romange: why do we need `old_bytes_buf` ?

romange: by looking at the code I suspect we can do it without copying data into a temporary buffer

src/server/bitops_family.cc
```diff
@@ -779,31 +779,69 @@ class Set {
...
+      byte_to_modify = TurnBitOn(byte_to_modify, bit_index_in_byte);
+    } else {
+      byte_to_modify = TurnBitOff(byte_to_modify, bit_index_in_byte);
+    }
+
     --lsb;
   }
 
+  if (was_negative && old_value > 0) {
+    old_value |= -1L ^ ((1L << attr_.encoding_bit_size) - 1);
```
romange, line 815: can you please add a comment explaining this?

vyavdoshenko: fixed

src/server/bitops_family.cc
vyavdoshenko: Kostas explained to me the simpler solution

src/server/bitops_family.cc
romange: i do not see any comments in this PR

src/server/bitops_family.cc
vyavdoshenko: fixed

src/server/bitops_family.cc
```diff
@@ -793,6 +794,8 @@ ResultType Set::ApplyTo(Overflow ov, string* bitfield) {
   uint32_t lsb = attr_.offset + attr_.encoding_bit_size - 1;
   int64_t old_value = 0;
 
+  const bool is_negative =
```
kostasrim, line 797: So much simpler now with less changes!

LGTM -- I don't know if Roman has any other comments

kostasrim:
> LGTM @kostasrim can you please review?

:rofl: I just wrote if you have any comments before seeing your comment -- yes give me one minute

src/server/bitops_family.cc
```diff
@@ -964,30 +978,35 @@ nonstd::expected<CommonAttributes, string> ParseCommonAttr(CmdArgParser* parser)
...
   }
 
   string_view bits = encoding.substr(1);
 
+  // Additional validation: check if bits part contains any invalid characters
+  for (char c : bits) {
+    if (!std::isdigit(c)) {
+      return make_unexpected(kInvalidBitfieldTypeErr);
+    }
+  }
```
kostasrim, line 998: Wouldn't `SimpleAtoi` cover this case ?

kostasrim (approved):
LGTM -- one minor comment

kostasrim:
> > LGTM @kostasrim can you please review?
> 
> 🤣 I just wrote if you have any comments before seeing your comment -- yes give me one minute

and I got carried away 1 minute 1 hour what :rofl: don't hate me!

src/server/bitops_family.cc
vyavdoshenko: Use std::isdigit if you only need to check that bits contains digits.
absl::SimpleAtoi is heavier — it creates a temporary string_view and tries to parse a number.
I guess this approach is a little bit lighter.

kostasrim: A string view is not a heavy weight object -- it's just a pointer and a size. What's more the happy path is that the string is valid and not the other way around so I would expect that on average we pay this parsing as "extra work". Furthermore, if the condition is covered by SimpleAtoi then this loop is also kinda redundant since we will check anyway.

Not a blocker though so I don't really mind

Issue #5139, opened 2025-05-15
PUBSUB Messages Sent As Command Replies


esatterwhite:
**Describe the bug**
We are running into situations where we are seeing pubsub messages being sent to a client as a reply to a command, which of course it has no knowledge of cause errors to be thrown and the parent process to crash

#### Normal Case

```txl
--- 
COMMAND FOUND

LAST COMMAND:  {
  command: Command {
    name: 'unsubscribe',
    inTransaction: false,
    isResolved: false,
    transformed: true,
    replyEncoding: 'utf8',
    errorStack: undefined,
    args: [ 'auth-provider-service' ],
    callback: undefined,
    resolve: [Function (anonymous)],
    reject: [Function (anonymous)],
    promise: Promise {
      <pending>,
      [Symbol(async_id_symbol)]: 35633,
      [Symbol(trigger_async_id_symbol)]: 35578
    }
  },
  stream: undefined,
  select: 0
}

CURRENT REPLY unsubscribe,auth-provider-service,0
---
```

#### Error Case

```txl

--- 
NO Command Found

LAST Command:  {
  command: Command {
    name: 'unsubscribe',
    inTransaction: false,
    isResolved: true,
    transformed: true,
    replyEncoding: 'utf8',
    errorStack: undefined,
    args: [ 'auth-provider-service' ],
    callback: undefined,
    resolve: [Function (anonymous)],
    reject: [Function (anonymous)],
    promise: Promise {
      0,
      [Symbol(async_id_symbol)]: 9614,
      [Symbol(trigger_async_id_symbol)]: 9559
    }
  },
  stream: undefined,
  select: 0
}

[...]

PR #5246, opened 2025-06-07
fix(pubsub): Skip stale pubsub messages

dranikpg:
Fixes https://github.com/dragonflydb/dragonfly/issues/5139

Dragonlfy had the following issue:
* After any kind of subscribe is issued, the client is in pubsub mode. No commannds, except pubsub ones, can be executed
* To exit the pubsub mode, the client has to unsubscribe from all channels. In that case the server replies with 0 as the number of pending subscriptions
* Many clients, like io-redis, rely on this condition to exit pubsub mode
* The reply is a regular resp3 push message over the same connection. The client can be stuck in pubsub for some time if it has a large socket buffer, large pile of previous messages buffered somewhere, etc... It only matters it's delivered eventually and delivered last
* Dragonfly checks subscriber validity only at the point of fetching them from the channel store. After that, it waits on backpressure and dispatches to the clients native threads. During this time, they might have already unsubscribed. So for Dragonfly, reading a successful reply from "unsubscribe" doesn't guarantee no more messages will be delivered.
* This PR fixes the "unsubscribe all" case. In general, we have the same problem when unsubscribing only from a specific channel.

### DIFF
--- a/tests/dragonfly/connection_test.py
+++ b/tests/dragonfly/connection_test.py
@@ -503,16 +503,13 @@ async def test_pubsub_unsubscribe(df_server: DflyInstance):
 
     # No messages should be received after we've read unsubscribe reply
     had_unsub = False
-    i = 0
     while True:
         reply = await conn.read_response(timeout=0.1)
         if reply is None:
             break
 
-        i += 1
         if reply[0] == "unsubscribe":
-            print("UNSUB POS ", i)
-            assert reply[2] == 0
+            assert reply[2] == 0  # zero subscriptions left
             had_unsub = True
         else:
             assert not had_unsub, "found message even after all subscriptions were removed"


src/facade/dragonfly_connection.cc
```diff
@@ -507,6 +507,14 @@ void Connection::AsyncOperations::operator()(const AclUpdateMessage& msg) {
 void Connection::AsyncOperations::operator()(const PubMessage& pub_msg) {
   RedisReplyBuilder* rbuilder = (RedisReplyBuilder*)builder;
 
+  // Discard stale messages to not break the protocol after exiting "pubsub" mode.
+  // Even after removing all subscriptions, we still can receive messages delayed
+  // by inter-thread dispatches or backpressure.
+  // TODO: filter messages from channels the client unsubscribed from
+  // if (self->cntx()->subscriptions == 0 &&
```
kostasrim: We can apply the similar filter at the point where we publish right ?

dranikpg: Not sure I understand 🙂 At the time we publish we just fetch all subscribers and they're all valid at that instant. The problem is that in-between _determining candidates for publishing_ and _actually sending the message on a client's thread_ the client performs operations (like unsubbing)

kostasrim: Yes I see. What I had in mind is to detect the presence of pending unsubscribe messages in the dispatch queue at the point we call `SendPubMessageAsync`. Thinking about it now it's unnecessarily complicated as the connection might be subscribed to multiple channels or the dispatch q might contain sequences of subscribe/unsubscribe messages making it difficult to detect if what we about to push to the dispatch queue is considered stale

PR #5252, opened 2025-06-09
chore: add corruption check for lpSetTotalBytes

romange:
We observed a corruption in stream code that looks like use-after-free pattern. This change introduces code that should pinpoint the place where the corruption happens by reading the first 4 bytes in the allocated listpack block. The first 4 bytes are used for storing the size argument of listpack. we check if the values there are correct and crash otherwise.

Adresses https://github.com/dragonflydb/dragonfly/issues/5255

### DIFF
--- a/src/redis/listpack.c
+++ b/src/redis/listpack.c
@@ -113,6 +113,25 @@
     (p)[3] = ((v)>>24)&0xff; \
 } while(0)
 
+/* TODO: delete this function once corruption in the stream code is identified */
+static void lpSetTotalBytesChecked(unsigned char *p, uint32_t v) {
+    uint32_t current = lpGetTotalBytes(p);
+    if (current == 0) {
+      fprintf(stderr, "Error: corrupted listpack size.");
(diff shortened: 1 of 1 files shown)


src/redis/listpack.c
```diff
@@ -113,6 +113,25 @@
     (p)[3] = ((v)>>24)&0xff; \
 } while(0)
 
+/* TODO: delete this function once corruption in the stream code is identified */
+static void lpSetTotalBytesChecked(unsigned char *p, uint32_t v) {
```
vyavdoshenko, line 117: I suggest creating a GitHub issue for that to-do and adding the link here.
The issue can be linked to the original bug for understanding.

romange: I agree. Can you please do it in dragonfly repo without disclosing any customer specific information?

vyavdoshenko: https://github.com/dragonflydb/dragonfly/issues/5255

Issue #5227, opened 2025-06-04
Error writing to traffic log: generic:5


vyavdoshenko:
Preconditions:
- Turn on traffic recording
- Create an intensive command load
- Start `redis-cli monitor` in a separate terminal

```
E20250604 20:38:03.370044 217614 dragonfly_connection.cc:174] Error writing to traffic log: generic:5
```

BorysTheDev:
The problem is that we write 0 bytes.

vyavdoshenko:
@BorysTheDev 
You can reproduce the issue. Please add the fix and try to reproduce it again; this may not be the only issue.

BorysTheDev:
@vyavdoshenko I try to understand why we try to write zero bytes

BorysTheDev:
Zero-size values in the command generate a zero-size write request, and this is the reason for this error.

BorysTheDev:
should be fixed by https://github.com/romange/helio/pull/418

PR #5256, opened 2025-06-09
fix: prevent writing zero size fields into log file

BorysTheDev:
If fields in the commands are zero size, we still try to write them, which triggers an error
fix: added check if the size of the field is zero

fixes: #5227


src/facade/dragonfly_connection.cc
```diff
@@ -261,13 +261,16 @@ void LogTraffic(uint32_t id, bool has_more, absl::Span<RespExpr> resp,
...
 
   for (auto part : resp) {
-    blobs[index++] = iovec{.iov_base = const_cast<char*>(part.GetView().data()),
-                           .iov_len = part.GetView().size()};
-    if (index >= blobs.size()) {
-      if (!tl_traffic_logger.Write(blobs.data(), blobs.size())) {
-        return;
+    if (auto blob_len = part.GetView().size(); blob_len > 0) {
+      blobs[index++] =
+          iovec{.iov_base = const_cast<char*>(part.GetView().data()), .iov_len = blob_len};
```
vyavdoshenko, line 267: This looks horrible. Why const_cast?
If it's reasonable, please add the comment above.

vyavdoshenko: I see it was above, but do you know why?

romange: because this is what the linux api demands.

vyavdoshenko (dismissed):
lgtm

romange (approved):

Issue #5089, opened 2025-05-09
support GAT/GATS in memcache


romange:
These commands are missing. See  https://github.com/memcached/memcached/blob/master/doc/protocol.txt

for more details

PR #5257, opened 2025-06-09
Implementation of memcache GAT and GATS commands

abhijat:
A new command is added to the string family called `GAT` which implements the memcache get and touch semantics. The command syntax is `GAT <expiry-seconds> key [keys...]`.

Most of the changes in the PR are related to extracting common code from the `MGET` path to be reused for `GAT`. 

A collection function is extracted which takes a search function to search for an individual key and collect results across many keys. Search has a dbslice iterator template parameter which can be mutable or immutable, depending on the command invoking it.

A runtime check is made such that a read only command does not invoke the helper with a mutable iterator.

Follow up of https://github.com/dragonflydb/dragonfly/pull/5199

The `GATS` command is also implemented, it returns a placeholder CAS version of 0 for now.

fixes https://github.com/dragonflydb/dragonfly/issues/5089

### DIFF
--- a/src/server/conn_context.h
+++ b/src/server/conn_context.h
@@ -147,10 +147,7 @@ struct ConnectionState {
     const ConnectionContext* owner = nullptr;
   };
 
-  enum MCGetMask {
-    FETCH_CAS_VER = 1,
-    SET_EXPIRY = 2,
-  };
+  enum MCGetMask { FETCH_CAS_VER = 1 };
 
   size_t UsedMemory() const;
 
@@ -259,9 +256,7 @@ struct ConnectionState {
   // used for memcache set/get commands.
   // For set op - it's the flag value we are storing along with the value.
   // For get op - we use it as a mask of MCGetMask values.
-  // For GAT the lowest two bits are used to store MCGetMask values, the top 62 bits are used to
-  // store expiration timestamp in seconds.
-  uint64_t memcache_flag = 0;
+  uint32_t memcache_flag = 0;
 
   ExecInfo exec_info;
   ReplicationInfo replication_info;
--- a/src/server/main_service.cc
+++ b/src/server/main_service.cc
@@ -1603,8 +1603,11 @@ void Service::DispatchMC(const MemcacheParser::Command& cmd, std::string_view va
     case MemcacheParser::PREPEND:
       strcpy(cmd_name, "PREPEND");
       break;
-    case MemcacheParser::GAT:
+    case MemcacheParser::GATS:
       [[fallthrough]];
+    case MemcacheParser::GAT:
+      strcpy(cmd_name, "GAT");
+      break;
     case MemcacheParser::GET:
       [[fallthrough]];
     case MemcacheParser::GETS:
@@ -1629,12 +1632,6 @@ void Service::DispatchMC(const MemcacheParser::Command& cmd, std::string_view va
 
   args.emplace_back(cmd_name, strlen(cmd_name));
 
-  if (!cmd.key.empty()) {
-    char* key = const_cast<char*>(cmd.key.data());
-    args.emplace_back(key, cmd.key.size());
-  }
-
-  ConnectionContext* dfly_cntx = static_cast<ConnectionContext*>(cntx);
   // if expire_ts is greater than month it's a unix timestamp
   // https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L139
   constexpr uint32_t kExpireLimit = 60 * 60 * 24 * 30;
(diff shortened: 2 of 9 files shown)


src/server/main_service.cc
```diff
@@ -1641,17 +1642,18 @@ void Service::DispatchMC(const MemcacheParser::Command& cmd, std::string_view va
     }
     dfly_cntx->conn_state.memcache_flag = cmd.flags;
   } else if (cmd.type < MemcacheParser::QUIT) {  // read commands
+    if (cmd.type == MemcacheParser::GAT) {
+      char* next = absl::numbers_internal::FastIntToBuffer(expire_ts, ttl);
+      args.emplace_back(ttl, next - ttl);
+    }
```
abhijat: `expire_ts` and its source: `cmd.expire_ts` are unsigned values. The memcache protocol specifies that a negative expiry is used to immediately expire a key, perhaps we should use signed values here to support that kind of expiry.

src/server/conn_context.h
```diff
@@ -147,10 +147,7 @@ struct ConnectionState {
     const ConnectionContext* owner = nullptr;
   };
 
-  enum MCGetMask {
-    FETCH_CAS_VER = 1,
-    SET_EXPIRY = 2,
-  };
+  enum MCGetMask { FETCH_CAS_VER = 1 };
```
abhijat, line 150: These changes (within this file) are reverted from the previous PR. Because we now have a separate command for GAT, we no longer need to stash the expiry in the memcache flag, it is passed explicitly as part of the command.

abhijat:
`GATS` is not implemented in this PR, but if we have to return the placeholder value for cas version similar to `GETS` , then the same command `GAT` can handle `GATS` as well making use of the memcache flag.

abhijat:
> `GATS` is not implemented in this PR, but if we have to return the placeholder value for cas version similar to `GETS` , then the same command `GAT` can handle `GATS` as well making use of the memcache flag.

As discussed elsewhere we should still return a placeholder, but I would prefer to do a follow up PR, mainly to add tests specific to `GATS`. 

The actual code additions required for `GATS` are very small.

abhijat:
> > `GATS` is not implemented in this PR, but if we have to return the placeholder value for cas version similar to `GETS` , then the same command `GAT` can handle `GATS` as well making use of the memcache flag.
> 
> As discussed elsewhere we should still return a placeholder, but I would prefer to do a follow up PR, mainly to add tests specific to `GATS`.
> 
> The actual code additions required for `GATS` are very small.

Added implementation for GATS, the tests are done through the same pytest code which tests GAT

src/server/main_service.cc
```diff
@@ -1642,17 +1645,18 @@ void Service::DispatchMC(const MemcacheParser::Command& cmd, std::string_view va
     }
     dfly_cntx->conn_state.memcache_flag = cmd.flags;
   } else if (cmd.type < MemcacheParser::QUIT) {  // read commands
+    if (cmd.type == MemcacheParser::GAT || cmd.type == MemcacheParser::GATS) {
+      char* next = absl::numbers_internal::FastIntToBuffer(expire_ts, ttl);
+      args.insert(args.begin() + 1, {ttl, static_cast<size_t>(next - ttl)});
```
romange: add here a comment demonstrating how a command looks like

romange (commented):
it's quite a big PR. given that you needed to do some refactoring in string family, can
you please split into a refactoring PR that does not change the functionality first?

abhijat:
> it's quite a big PR. given that you needed to do some refactoring in string family, can you please split into a refactoring PR that does not change the functionality first?

Created PR with the refactor specific changes https://github.com/dragonflydb/dragonfly/pull/5289

src/server/main_service.cc
abhijat: I have reordered the code a bit so we can emplace back instead of inserting in the middle, also added a comment.

src/server/string_family.cc
```diff
@@ -1642,38 +1694,87 @@ void StringFamily::ClThrottle(CmdArgList args, const CommandContext& cmnd_cntx)
...
+  CmdArgParser parser{args};
+  const int64_t expire_ts = parser.Next<uint64_t>();
+  if (parser.HasError()) {
+    return builder->SendError(parser.Error()->MakeReply());
+  }
+
+  BlockingCounter tiering_bc{0};
+  std::vector<MGetResponse> mget_resp(shard_set->size());
+  const DbSlice::ExpireParams expire_params{
+      .value = expire_ts, .absolute = true, .persist = expire_ts == 0};
```
romange, line 1723: is it correct `.absolute = true` ?

src/server/string_family.cc
```diff
@@ -822,6 +822,58 @@ void IncrByGeneric(string_view key, int64_t val, Transaction* tx, SinkReplyBuild
...
+  const Transaction* t;
+  EngineShard* shard;
+  const DbSlice::ExpireParams& expire_params;
+  const string_view key;
+};
+
+OpResult<DbSlice::Iterator> FindKeyAndSetExpiry(const GetAndTouchParams& params) {
+  const DbContext& ctx = params.t->GetDbContext();
+  DbSlice& db_slice = params.t->GetDbSlice(params.shard->shard_id());
+  auto it = db_slice.FindMutable(ctx, params.key, OBJ_STRING);
```
romange: nit: call it `find_res` instead of `it`.

abhijat: changed

src/server/string_family.cc
abhijat: This timestamp is calculated in https://github.com/dragonflydb/dragonfly/blob/739d9376ed96cf9e28d13e8cc4c5d8c8d24d977e/src/server/main_service.cc#L1635-L1640

If it is 0 then the expiry is removed using `persist`, absolute is not relevant in this case. 
If it is non zero and less than one month then it is converted to absolute by adding it to seconds since epoch in the above code path.
If it is greater than 1 month the contract is to treat is as absolute as per the memcache protocol https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L144-L150

PR #5264, opened 2025-06-10
fix(search_family): Fix empty key bug

BagritsevichStepan:
Fix crash for empty keys in `ShardDocIndex::DocKeyIndex`


romange:
Ok, I was not sure whether the empty string assumption affects the production logic. seems like it does not.

romange (approved):

PR #5266, opened 2025-06-10
fix(json_family): Fix bugs where JSON commands were modifying values of other data types

### LINKED ISSUE #5265: JSON commands changing other datatypes
I found two bugs in current json logic. We are not using `OBJ_JSON` for some commands when we call `FindMutable`.
1. Bug:

```
127.0.0.1:6379> hset k1 field value                                                                                                                                                                         
(integer) 1                                                                                                                                                                                                 
127.0.0.1:6379> json.set k1 $ '{"a":"b"}'                                                                                                                                                                   
OK
```
Expected:
```
127.0.0.1:6379> hset k1 field value
(integer) 1
127.0.0.1:6379> json.set k1 $ '{"a":"b"}'
(error) WRONGTYPE Operation against a key holding the wrong kind of value
```

2. Bug:

```
127.0.0.1:6379> hset k1 field value
(integer) 1
127.0.0.1:6379> json.del k1
(integer) 1
127.0.0.1:6379> hget k1 field
(nil)
```
Expected:
```
127.0.0.1:6379> hset k1 field value
(integer) 1
127.0.0.1:6379> json.del k1
(error) WRONGTYPE Operation against a key holding the wrong kind of value
127.0.0.1:6379> hget k1 field
"value"
```

BagritsevichStepan:
fixes #5265 

~In this PR, I added `req_obj_type` to the `AddOrFind` method. I see that this method is also used in other data types, which could lead to bugs. Once this PR is approved, I will review other commands and possibly remove the version of `AddOrFind` that doesn't need `req_obj_type`.~

This PR fixes several bugs that are well described in the issue above::
1. Fixed bug in the `JSON.SET` command
2. Fixed bug in the `JSON.DEL` command
3. Add comment to the `JSON.GET` command

### DIFF
--- a/src/server/db_slice.h
+++ b/src/server/db_slice.h
@@ -306,8 +306,7 @@ class DbSlice {
   OpResult<ConstIterator> FindReadOnly(const Context& cntx, std::string_view key,
                                        unsigned req_obj_type) const;
 
-  OpResult<ItAndUpdater> AddOrFind(const Context& cntx, std::string_view key,
-                                   std::optional<unsigned> req_obj_type = std::nullopt);
+  OpResult<ItAndUpdater> AddOrFind(const Context& cntx, std::string_view key);
 
   // Same as AddOrSkip, but overwrites in case entry exists.
   OpResult<ItAndUpdater> AddOrUpdate(const Context& cntx, std::string_view key, PrimeValue obj,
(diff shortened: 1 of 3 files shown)


src/server/db_slice.h
```diff
@@ -307,6 +307,8 @@ class DbSlice {
                                        unsigned req_obj_type) const;
 
   OpResult<ItAndUpdater> AddOrFind(const Context& cntx, std::string_view key);
+  OpResult<ItAndUpdater> AddOrFind(const Context& cntx, std::string_view key,
```
romange: I do not know if it justified another overload here. you could check the return type in the only place you use `AddOrFind` like this.

romange (dismissed):

BagritsevichStepan:
Actually, we do not need change AddOrFind here, create a separate ticket for this: #5271

romange (approved):

Issue #5251, opened 2025-06-08
test_pubsub_unsubscribe regression test fail


vyavdoshenko:
https://github.com/dragonflydb/dragonfly/actions/runs/15521135610/job/43694486752

The only warning in the log:
```
W20250608 18:17:31.524840 22129 dfly_main.cc:817] Got memory limit 8.00GiB, however only 6.02GiB was found.
```

vyavdoshenko:
https://github.com/dragonflydb/dragonfly/actions/runs/15522497449/job/43697624934

romange:
@vyavdoshenko  please increase the timeout `reply = await conn.read_response(timeout=0.1)` in the test and see if it helps

kostasrim:
The warning is unrelated btw   

```
>       assert had_unsub
E       assert False
 ```

vyavdoshenko:
@romange 
The timeout of 0.2 doesn't help. Try 0.5?

romange:
seems like someone needs to debug it

mkaruza:
Should be fixed with #5269

mkaruza:
Closing. If #5269 didn't resolve this issue it needs to be reopened again

PR #5269, opened 2025-06-10
fix(reply_builder): Flush if last message didn't reply and flush but supose to

### LINKED ISSUE #5139: PUBSUB Messages Sent As Command Replies

mkaruza:
When connection is subscribed to channel we are receiving async messages and sending them back to client. When UNSUBSCRIBE is received on same connection it can happen that we are starting to send back data in batch mode. In batch mode we will be aggregating responses until flush. Last message in queue should flush during reply but if it didn't (i.e. PubSub message was discarded) we are going to flush manually.

Fixes #5139 #5251

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1653,8 +1653,8 @@ void Connection::AsyncFiber() {
       // We keep the batch mode enabled as long as the dispatch queue is not empty, relying on the
       // last command to reply and flush. If it doesn't reply (i.e. is a control message like
       // migrate), we have to flush manually.
-      bool is_replaying = msg.IsReplying();
-      if (dispatch_q_.empty() && !is_replaying) {
+      bool is_replying = msg.IsReplying();
+      if (dispatch_q_.empty() && !is_replying) {
         reply_builder_->Flush();
       }
 
@@ -1669,9 +1669,9 @@ void Connection::AsyncFiber() {
       cc_->async_dispatch = true;
       std::visit(async_op, msg.handle);
       cc_->async_dispatch = false;
-      // If last msg in queue was replaying but nothing was replied during dispatch
+      // If last msg in queue was replying but nothing was replied during dispatch
       // (i.e. pubsub message was discarded) we have to manually flush now.
-      if (dispatch_q_.empty() && is_replaying &&
+      if (dispatch_q_.empty() && is_replying &&
           (replies_recorded_before == reply_builder_->RepliesRecorded())) {
         reply_builder_->Flush();
       }
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1653,7 +1653,8 @@ void Connection::AsyncFiber() {
       // We keep the batch mode enabled as long as the dispatch queue is not empty, relying on the
       // last command to reply and flush. If it doesn't reply (i.e. is a control message like
       // migrate), we have to flush manually.
-      if (dispatch_q_.empty() && !msg.IsReplying()) {
+      bool is_replaying = msg.IsReplying();
+      if (dispatch_q_.empty() && !is_replaying) {
         reply_builder_->Flush();
       }
 
@@ -1664,9 +1665,16 @@ void Connection::AsyncFiber() {
         return;  // don't set conn closing flag
       }
 
+      auto replies_recorded_before = reply_builder_->RepliesRecorded();
       cc_->async_dispatch = true;
       std::visit(async_op, msg.handle);
       cc_->async_dispatch = false;
+      // If last msg in queue was replaying but nothing was replied during dispatch
+      // (i.e. pubsub message was discarded) we have to manually flush now.
+      if (dispatch_q_.empty() && is_replaying &&
+          (replies_recorded_before == reply_builder_->RepliesRecorded())) {
+        reply_builder_->Flush();
+      }
       RecycleMessage(std::move(msg));
     }
 
--- a/src/server/conn_context.cc
+++ b/src/server/conn_context.cc
@@ -184,12 +184,6 @@ void ConnectionContext::ChangeSubscription(bool to_add, bool to_reply, CmdArgLis
       rb->SendBulkString(action[to_add]);
       rb->SendBulkString(ArgS(args, i));
       rb->SendLong(result[i]);
-      // We are manually flushing reply builder here. It can happen that we are subscribed to
-      // channel and `UNSUBSCRIBE` is executed while we are receiving messages
-      // we can end up in `batched_` mode when dispatch_q_.size() > 1. In batch mode we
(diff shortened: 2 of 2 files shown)


romange:
The question below is not intended to block the potential fix in this pr.

@dranikpg  Do you remember why we need `SetBatchMode` at all at this point? Now that we have `ReplyScope` could it be that we can just avoid setting `SetBatchMode(dispatch_q_.size() > 1);` which waits for the next command to be processed and its reply flushed. I am asking because this surely increases the reply latency if the next command is still in the dispatch_q_. 

Another question to both of you - why the reply has not been flushed eventually? if `dispatch_q_.size() > 1` I would imagine that even if the client is blocked, the next commands in the queue will trigger flushing so why it does not happen? Also, could it happen with more usecases besides `unsubscribe`?

src/facade/dragonfly_connection.cc
```diff
@@ -1653,7 +1653,8 @@ void Connection::AsyncFiber() {
       // We keep the batch mode enabled as long as the dispatch queue is not empty, relying on the
       // last command to reply and flush. If it doesn't reply (i.e. is a control message like
       // migrate), we have to flush manually.
-      if (dispatch_q_.empty() && !msg.IsReplying()) {
+      bool is_replaying = msg.IsReplying();
```
romange: is_replaying -> is_replying

src/facade/dragonfly_connection.cc
```diff
@@ -1664,9 +1665,16 @@ void Connection::AsyncFiber() {
         return;  // don't set conn closing flag
       }
 
+      auto replies_recorded_before = reply_builder_->RepliesRecorded();
       cc_->async_dispatch = true;
       std::visit(async_op, msg.handle);
       cc_->async_dispatch = false;
+      // If last msg in queue was replaying but nothing was replied during dispatch
```
romange: replying

romange (approved):

dranikpg:
> Another question to both of you - why the reply has not been flushed eventually

It will be flushed. The problem is that the client waits for a reply (in that case for an unsubscribe) before sending any new commands, so it's esentially a deadlock. There are no new commands, it happens with dispatch_q.empty

> Do you remember why we need SetBatchMode at all at this point? Now that we have ReplyScope ...

Yes and no. We'll have to experiment with it, also with the internal buffer size. If I send 10 fast commands with short replies (i.e OK) in a pipleline, do we really want to reply separately?  Nothing I can tell about the optimal internal buffer / socket buffer / send calls ratio 🤷🏻‍♂️

dranikpg:
Also 🤨 `ReplyScope` didn't make any functional difference in socket access. Not sure I understant the 

> Now that we have

part

Issue #5259, opened 2025-06-09
bug(json): recursive deletion does not delete all the fields


romange:
To reproduce:
```
> "JSON.SET" "doc1" "." "{\"a\": 1, \"nested\": {\"a\": 2, \"b\": 3}}"
>"JSON.DEL" "doc1" "$..a"
1
> "JSON.GET" "doc1" $
"[{\"nested\":{\"a\":2,\"b\":3}}]"
```

expected:
```
>"JSON.DEL" "doc1" "$..a"
2
> "JSON.GET" "doc1" $
"[{\"nested\":{\"b\":3}}]"

**Comment by romange** [comment]
```
"JSON.SET" "doc2" "$" "{\"a\": {\"a\": 2, \"b\": 3}, \"b\": [\"a\", \"b\"], \"nested\": {\"b\": [true, \"a\", \"b\"]}}"
"JSON.DEL" "doc2" "$..a"
```

return 2 but is expected 1.

romange:
for more details run: `poetry run pytest -s test/test_json/test_json.py  -k test_json_delete_with_dollar` from `tests/fakeredis`
you can find there README on how to install and run fakereadis

PR #5270, opened 2025-06-10
fix(json): Ensure full deletion for multi-match JSONPaths

vyavdoshenko:
Fixes: https://github.com/dragonflydb/dragonfly/issues/5259

This PR resolves an issue where JSON.DEL failed to remove all targeted elements when a JSONPath expression (e.g., $..a) matched multiple nodes within a JSON object.

The bug was caused by using forward iteration while deleting from a collection, which resulted in skipping elements as indices shifted during the mutation.

The solution implements a context-aware traversal strategy:
- Reverse traversal is now explicitly used for delete operations, guaranteeing that all matched nodes are processed and removed correctly.
- Forward traversal is preserved for non-destructive mutations (e.g., updates), for which it is safe and efficient.
This approach ensures JSON.DEL is robust and predictable without impacting other mutation commands.

### DIFF
--- a/src/core/json/detail/jsoncons_dfs.cc
+++ b/src/core/json/detail/jsoncons_dfs.cc
@@ -111,15 +111,15 @@ Dfs Dfs::Mutate(absl::Span<const PathSegment> path, const MutateCallback& callba
         }
       }
     } else {
-      // If Advance failed (e.g., MISMATCH or OUT_OF_BOUNDS), check if the current node
-      // itself is a terminal target for mutation due to a previous DESCENT segment.
-      // This handles cases like $.a..b where 'a' itself might match the 'b' after descent.
+      // If Advance failed (e.g., MISMATCH or OUT_OF_BOUNDS), the current node itself
+      // might still be a terminal match because of the previous DESCENT segment.
+      // Instead of mutating immediately (which could break ordering guarantees),
+      // collect the node and defer mutation until after traversal.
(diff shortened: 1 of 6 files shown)


romange:
I think there is a conceptual problem here, and it started even before, when #5040 was introduced.
the problem is that now `Dfs::Mutate` does not guarantee order at which `callback` is called.
sometimes we add a node to `nodes_to_mutate`, and sometimes we call `dfs.MutateStep` directly. 
now, some json replies are correlated with the order which may create confusion. I understand that the order can not guaranteed in json objects but I think we should at least provide clear guarantees at `Dfs::Mutate` level.

Could it be that traversing everything in reverse would solve both #5038 and #5259?

vyavdoshenko:
@romange 
You are right that the previous solution was confusing because I missed postponing the call `Dfs::Mutate` for all cases in fixing this PR: https://github.com/dragonflydb/dragonfly/issues/5040. In this case, the order would be guaranteed. That's why I fixed this in this PR.

This PR exactly fixes https://github.com/dragonflydb/dragonfly/issues/5259. I added the unit test to jsonpath_test.cc. Also, I can add the unit test to the json_family_test.cc if needed.

This PR couldn't fix: https://github.com/dragonflydb/dragonfly/issues/5038. It was needed, and the current change polishes the previous fix.

romange (approved):

PR #5274, opened 2025-06-11
chore: minor fixes in dash table

romange:
Also, relax SearchNonNullFields some checks that break if we change dashtable internals due to change of items ordering in the table.

### DIFF
--- a/src/core/dash_internal.h
+++ b/src/core/dash_internal.h
@@ -403,6 +403,13 @@ class Segment {
   explicit Segment(size_t depth, PMR_NS::memory_resource* mr) : local_depth_(depth), mr_(mr) {
   }
 
+  ~Segment() {
+    Clear();
+  }
+
+  Segment(const Segment&) = delete;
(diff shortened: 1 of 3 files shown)


src/core/dash_internal.h
```diff
@@ -403,6 +403,13 @@ class Segment {
   explicit Segment(size_t depth, PMR_NS::memory_resource* mr) : local_depth_(depth), mr_(mr) {
   }
 
+  ~Segment() {
+    Clear();
```
kostasrim, line 407: This clears the data members but we are destructing the object ? Do we really need this ?

vyavdoshenko (approved):
lgtm

kostasrim (commented):
lgtm minor comment

src/core/dash_internal.h
romange: we will need it with dynamic stash buckets in the next pr.

PR #5283, opened 2025-06-12
fix: client kill preempts in atomic section on shutdown

kostasrim:
The problem is that `TlsSocket::Shutdown` is preemptive because it flushes its buffer on the socket. However this violated the FiberAtomicGuard when traversing the connection list on each shard. To fix this, we move the shutdown call to another fiber.

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -535,32 +535,28 @@ void ClientKill(CmdArgList args, absl::Span<facade::Listener*> listeners, SinkRe
 
   const bool is_admin_request = cntx->conn()->IsPrivileged();
 
-  std::vector<std::vector<facade::Connection::WeakRef>> thread_connections(
-      shard_set->pool()->size());
-
   atomic<uint32_t> killed_connections = 0;
   atomic<uint32_t> kill_errors = 0;
-
-  auto traverse_cb = [&](unsigned idx, util::Connection* conn) {
-    facade::Connection* dconn = static_cast<facade::Connection*>(conn);
-    if (evaluator(dconn)) {
-      if (is_admin_request || !dconn->IsPrivileged()) {
-        thread_connections[idx].push_back(dconn->Borrow());
-      } else {
-        kill_errors.fetch_add(1);
-      }
-    }
-  };
 
   auto cb = [&](unsigned idx, ProactorBase* p) mutable {
     // Step 1 aggregate the per thread connections from all listeners
     std::vector<facade::Connection::WeakRef> connections;
+    auto traverse_cb = [&](unsigned idx, util::Connection* conn) {
+      facade::Connection* dconn = static_cast<facade::Connection*>(conn);
+      if (evaluator(dconn)) {
+        if (is_admin_request || !dconn->IsPrivileged()) {
+          connections.push_back(dconn->Borrow());
+        } else {
+          kill_errors.fetch_add(1);
+        }
+      }
+    };
     for (auto* listener : listeners) {
       listener->TraverseConnectionsOnThread(traverse_cb, UINT32_MAX, nullptr);
     }
 
     // Step 2 kill the clients
-    for (auto& tcon : thread_connections[idx]) {
+    for (auto& tcon : connections) {
       facade::Connection* conn = tcon.Get();
       if (conn && conn->socket()->proactor()->GetPoolIndex() == p->GetPoolIndex()) {
         conn->ShutdownSelf();


src/server/server_family.cc
```diff
@@ -535,14 +535,39 @@ void ClientKill(CmdArgList args, absl::Span<facade::Listener*> listeners, SinkRe
...
+
+  size_t round = 0;
+
+  auto client_kill_fb_body = [&consumers, &killed_connections](unsigned idx) {
+    auto& kill_list = consumers[idx];
+    while (!kill_list.empty()) {
+      facade::Connection::WeakRef ref = std::move(kill_list.front());
+      kill_list.pop_front();
+      facade::Connection* conn = ref.Get();
+      // TODO think how to handle migration for eval. See RequestAsyncMigration
```
kostasrim: Need to think what we can do here

src/server/server_family.cc
```diff
@@ -535,14 +535,39 @@ void ClientKill(CmdArgList args, absl::Span<facade::Listener*> listeners, SinkRe
 
   const bool is_admin_request = cntx->conn()->IsPrivileged();
 
+  std::vector<util::fb2::Fiber> fibers(pp->size() * listeners.size());
```
kostasrim: IMO we could also have a single fiber per shard which sleeps and wake up when we push work to its queue. IMO I have no preference

src/server/server_family.cc
```diff
@@ -499,7 +499,7 @@ void ClientId(CmdArgList args, SinkReplyBuilder* builder, ConnectionContext* cnt
 }
 
 void ClientKill(CmdArgList args, absl::Span<facade::Listener*> listeners, SinkReplyBuilder* builder,
-                ConnectionContext* cntx) {
+                ConnectionContext* cntx, util::ProactorPool* pp) {
```
romange: you do not need to pass it - it's accessible via `shard_set->pool()`

src/server/server_family.cc
romange: I do not understand the reason for all the complexity here and below.
I would imagine to have an array per thread and you just add all the matching connections to the array in their owning thread.

Then the next step is to run `pool->AwaitFiberOnAll` that accesses an array in its own thread and sequentially shuts down the connections.

kostasrim: yeah I agree it was an overkill 

There is a small issue -- plz see my other comment

src/server/server_family.cc
```diff
@@ -535,23 +535,40 @@ void ClientKill(CmdArgList args, absl::Span<facade::Listener*> listeners, SinkRe
...
+    // Step 1 aggregate the per thread connections from all listeners
+    std::vector<facade::Connection::WeakRef> connections;
+    for (auto* listener : listeners) {
+      listener->TraverseConnectionsOnThread(traverse_cb, UINT32_MAX, nullptr);
+    }
+
+    // Step 2 kill the clients
+    for (auto& tcon : thread_connections[idx]) {
+      facade::Connection* conn = tcon.Get();
+      if (conn && conn->socket()->proactor()->GetPoolIndex() == p->GetPoolIndex()) {
```
kostasrim, line 561: @romange 

The problem here is that we might migrate a connection between preemptions of Shutdown. Calling `tcon.Get()` on a connection that was migrated can become a data race.


Maybe the solution is to simply disable the migrations for each connection we are about to kill ? WDYT ?

kostasrim: And now I am thinking about it, when we call `Borrow()` we should disable migrations. If a connection migrates after borrow, we cached `socket_->proactor()->GetPoolIndex()` within `ConnectionWeakRef` which is no longer valid. The DCHECK() within `Get()` won't protect us either, it uses the cached yet no longer valid proactor index

romange: we run with `migrate_connections=false` in production so it's not an issue.
i think your checks are good enough for this.

kostasrim: sounds good

src/server/server_family.cc
```diff
@@ -535,23 +535,41 @@ void ClientKill(CmdArgList args, absl::Span<facade::Listener*> listeners, SinkRe
...
         kill_errors.fetch_add(1);
       }
     }
   };
 
-  for (auto* listener : listeners) {
-    listener->TraverseConnections(cb);
-  }
+  auto cb = [&](unsigned idx, ProactorBase* p) mutable {
+    // Step 1 aggregate the per thread connections from all listeners
```
romange, line 542: if you do both steps locally in-thread, there is not reason to declare `thread_connections` as a global object outside. Instead just define `vector<facade::Connection::WeakRef>` here

kostasrim: Yes I thought about that. My issue was that we need to pass this thread local to `traverse_cb`. This implies either defining the traverse_cb within this lambda which makes it difficult to read or to `bind` it via `[&local] { travere_cb(local, ...); }` 

I am fine with any of these -- I will do whatever you find more readable :)

romange: i am fine with either option (slight preference to first).

kostasrim: sounds good I will follow up

romange (approved):

PR #5285, opened 2025-06-12
dfly_bench: Support terminating early

abhijat:
In certain situations dfly_bench is used as a load generation function, for example testing performance while new replica is added in dftest. In this case it is useful to run dfly_bench for a long period of time while another operation is running, and then stop it once the other operation ends.

Stopping the process using sigkill etc loses information about the stats collected so far. In this change a signal handler is added such that on sigterm the driver processes stop gracefully and report stats similar to a normal run ending.

### DIFF
--- a/src/server/dfly_bench.cc
+++ b/src/server/dfly_bench.cc
@@ -92,6 +92,7 @@ using tcp = ::boost::asio::ip::tcp;
 using absl::StrCat;
 
 thread_local base::Xoroshiro128p bit_gen;
+atomic_bool terminate_requested = false;
 
 #if __INTELLISENSE__
 #pragma diag_suppress 144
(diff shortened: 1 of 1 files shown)


src/server/dfly_bench.cc
```diff
@@ -1103,6 +1104,11 @@ int main(int argc, char* argv[]) {
   pp->Run();
   fb2::InitDnsResolver(2000);
 
+  ProactorBase::RegisterSignal({SIGTERM}, pp->GetNextProactor(), [](int) {
```
kostasrim, line 1107: we can also hook for `SIGUSR1` and we do this already in df 🤷

src/server/dfly_bench.cc
```diff
@@ -648,7 +649,7 @@ void Driver::Run(uint64_t* cycle_ns, CommandGenerator* cmd_gen) {
 
   uint32_t num_batches = ((num_reqs_ - 1) / pipeline) + 1;
 
-  for (unsigned i = 0; i < num_batches && now < time_limit_ns; ++i) {
+  for (unsigned i = 0; i < num_batches && now < time_limit_ns && !terminate_requested; ++i) {
```
kostasrim, line 652: After we exit the for loop we terminate right ?

abhijat: yes, I tested it several times manually

abhijat: To be more concrete:

* the for loop exits per set of drivers per shard
* drivers wait for request queue to finish, then shut down
* thread local client (which is waiting for driver fibers to join) finishes per core
* watch fiber stop condition is triggered and exits
* program finishes

PR #5286, opened 2025-06-12
chore: Improve scope/aggregator usage in replies

dranikpg:

### DIFF
--- a/src/server/set_family.cc
+++ b/src/server/set_family.cc
@@ -1025,7 +1025,7 @@ void SMIsMember(CmdArgList args, const CommandContext& cmd_cntx) {
   string_view key = ArgS(args, 0);
   auto members = args.subspan(1);
 
-  vector<long> memberships(members.size());
+  vector<int32_t> memberships(members.size());
 
   auto cb = [&](Transaction* t, EngineShard* shard) {
     DbContext db_cntx = t->GetDbContext();


src/server/bloom_family.cc
```diff
@@ -179,10 +179,10 @@ void BloomFamily::MExists(CmdArgList args, const CommandContext& cmd_cntx) {
     return OpExists(t->GetOpArgs(shard), key, args);
   };
 
-  OpResult res = cmd_cntx.tx->ScheduleSingleHopT(std::move(cb));
+  auto res = cmd_cntx.tx->ScheduleSingleHopT(std::move(cb));
```
romange: please do not convert to auto - when it hurts readability

src/facade/reply_builder.h
```diff
@@ -288,8 +288,17 @@ class RedisReplyBuilder : public RedisReplyBuilderBase {
...
+  // One-liner for ReplyScope + StartArray
+  struct ArrayScope : ReplyScope {
+    ArrayScope(RedisReplyBuilder* rb, size_t len) : ReplyScope(rb) {
+      rb->StartArray(len);
+    }
+  };
+
   void SendSimpleStrArr(const facade::ArgRange& strs);
   void SendBulkStrArr(const facade::ArgRange& strs, CollectionType ct = ARRAY);
+  void SendLongArr(absl::Span<const long> longs);
```
romange: instead of changing result types - let's make it more flexible.
```cc
template<typename I>` void SendLongArr(absl::Span<const I> longs) {
}
```

src/server/list_family.cc
```diff
@@ -576,8 +576,8 @@ OpResult<string> OpIndex(const OpArgs& op_args, std::string_view key, long index
   return str;
 }
 
-OpResult<vector<uint32_t>> OpPos(const OpArgs& op_args, string_view key, string_view element,
-                                 int rank, uint32_t count, uint32_t max_len) {
+OpResult<vector<long>> OpPos(const OpArgs& op_args, string_view key, string_view element, int rank,
```
romange: see my comment above

dranikpg:
Cluster family test failed spuriously in the previour CI run

src/facade/reply_builder.cc
```diff
@@ -448,6 +451,22 @@ void RedisReplyBuilder::SendLabeledScoredArray(std::string_view arr_label, Score
   }
 }
 
+template <typename I> void RedisReplyBuilder::SendLongArr(absl::Span<const I> longs) {
+  static_assert(std::is_integral_v<I>, "Must use integral type");
+  ReplyScope scope(this);
+  StartArray(longs.size());
+  for (auto v : longs) {
+    if constexpr (std::is_unsigned_v<I>)
+      DCHECK_LE(uint64_t(v), uint64_t(std::numeric_limits<long>::max()));
```
romange, line 460: what do you check here?

src/facade/reply_builder.cc
```diff
@@ -448,6 +451,22 @@ void RedisReplyBuilder::SendLabeledScoredArray(std::string_view arr_label, Score
...
+    SendLong(v);
+  }
+}
+
+template void RedisReplyBuilder::SendLongArr<long>(absl::Span<const long>);
```
romange, line 465: nice :)

src/server/set_family.cc
```diff
@@ -1025,32 +1025,26 @@ void SMIsMember(CmdArgList args, const CommandContext& cmd_cntx) {
   string_view key = ArgS(args, 0);
   auto members = args.subspan(1);
 
-  vector<bool> memberships(members.size());
+  vector<long> memberships(members.size());
```
romange:
```suggestion
  vector<int32_t> memberships(members.size());
```

src/facade/reply_builder.cc
dranikpg: if someone wants to send uint64_t::max, then after the conversion to int64_t, it'll end up something different

dranikpg: So I check that if an unsigned integer is supplied, it's value is less than signed int64 max

PR #5288, opened 2025-06-12
chore: add -skip-time-min flag to skip initial N minutes during replay traffic

vyavdoshenko:
Introduces CLI flag -skip-time-min (int) to skip the first N minutes of recorded traffic.

### DIFF
--- a/tools/replay/main.go
+++ b/tools/replay/main.go
@@ -17,7 +17,6 @@ var fHost = flag.String("host", "127.0.0.1:6379", "Redis host")
 var fClientBuffer = flag.Int("buffer", 100, "How many records to buffer per client")
 var fPace = flag.Bool("pace", true, "whether to pace the traffic according to the original timings.false - to pace as fast as possible")
 var fSkip = flag.Uint("skip", 0, "skip N records")
-var fSkipTimeMin = flag.Int("skip-time-min", 0, "skip records in the first N minutes of the recording")
 var fSkipTimeSec = flag.Int("skip-time-sec", 0, "skip records in the first N seconds of the recording")
 var fIgnoreParseErrors = flag.Bool("ignore-parse-errors", false, "ignore parsing errors")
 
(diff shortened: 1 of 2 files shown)


tools/replay/main.go
```diff
@@ -17,6 +17,7 @@ var fHost = flag.String("host", "127.0.0.1:6379", "Redis host")
 var fClientBuffer = flag.Int("buffer", 100, "How many records to buffer per client")
 var fPace = flag.Bool("pace", true, "whether to pace the traffic according to the original timings.false - to pace as fast as possible")
 var fSkip = flag.Uint("skip", 0, "skip N records")
+var fSkipTimeMin = flag.Int("skip-time-min", 0, "skip records in the first N minutes of the recording")
```
romange: i think minutes is too coarse. lets do this in seconds.

tools/replay/main.go
```diff
@@ -17,6 +17,7 @@ var fHost = flag.String("host", "127.0.0.1:6379", "Redis host")
 var fClientBuffer = flag.Int("buffer", 100, "How many records to buffer per client")
 var fPace = flag.Bool("pace", true, "whether to pace the traffic according to the original timings.false - to pace as fast as possible")
 var fSkip = flag.Uint("skip", 0, "skip N records")
```
romange, line 19: btw, you can reach a fairly fast replay if you run with `-pace=false`

tools/replay/main.go
vyavdoshenko: In my experience, minutes is the best choice.
I decided to add this option when I had to wait for 120 minutes.

vyavdoshenko: Maybe it's better to add two options, in seconds and in minutes.

tools/replay/main.go
vyavdoshenko: I know, I tried this option, but it didn't work for me to reproduce the issue.

tools/replay/main.go
dranikpg: You can always convert minutes to seconds, not the other way round. Two flags can be confusing

tools/replay/main.go
dranikpg: Maybe it's some kind of sneaky client hello, ping, etc that is sent in advance by a client you set up

tools/replay/main.go
vyavdoshenko: fixed

PR #5289, opened 2025-06-13
chore(string_family): Refactor to prep for adding the GAT command

abhijat:
A couple of functions are extracted from the MGet command. In a follow up PR, the GAT command will be added which
will reuse some of the functions:

- A function to collect keys is extracted. This takes a function as input which will return an iterator for one key. The iterator can be mutable or immutable, for MGet it is immutable. For GAT it will be mutable.
- A check is added such that a read-only command does not use a function returning a mutable iterator.
- Another function is created to reorder collected results based on the shards of the keys and the original order of the keys in the command.

For context, the original PR is https://github.com/dragonflydb/dragonfly/pull/5257

### DIFF
--- a/src/server/string_family.cc
+++ b/src/server/string_family.cc
@@ -551,10 +551,9 @@ MGetResponse CollectKeys(BlockingCounter wait_bc, uint8_t fetch_mask, const Tran
   ShardArgs keys = t->GetShardArgs(shard->shard_id());
   DCHECK(!keys.Empty());
 
-  if constexpr (constexpr bool is_mut_iter = std::is_same_v<Iter, DbSlice::Iterator>) {
+  if constexpr (std::is_same_v<Iter, DbSlice::Iterator>) {
     const auto cid = t->GetCId();
-    DCHECK(!(cid->IsReadOnly() && is_mut_iter))
-        << "mutable iterator used with read-only command " << cid->name();
+    DCHECK(!cid->IsReadOnly()) << "mutable iterator used with read-only command " << cid->name();
   }
 
   MGetResponse response(keys.Size());
--- a/src/server/string_family.cc
+++ b/src/server/string_family.cc
@@ -536,16 +536,31 @@ struct MGetResponse {
 // fetch_mask values
 constexpr uint8_t FETCH_MCFLAG = 0x1;
 constexpr uint8_t FETCH_MCVER = 0x2;
-MGetResponse OpMGet(fb2::BlockingCounter wait_bc, uint8_t fetch_mask, const Transaction* t,
-                    EngineShard* shard) {
+
+template <typename Iter> using SearchKey = std::function<OpResult<Iter>(string_view)>;
+
+// A find operation which can mutate, for commands which can write, eg GAT
+using SearchMut = SearchKey<DbSlice::Iterator>;
+
+// Const find operation, for read-only commands, eg MGet
+using SearchConst = SearchKey<DbSlice::ConstIterator>;
+
+template <typename Iter>
+MGetResponse CollectKeys(BlockingCounter wait_bc, uint8_t fetch_mask, const Transaction* t,
+                         EngineShard* shard, SearchKey<Iter> find_op) {
   ShardArgs keys = t->GetShardArgs(shard->shard_id());
   DCHECK(!keys.Empty());
 
   auto& db_slice = t->GetDbSlice(shard->shard_id());
 
+  const auto cid = t->GetCId();
+  constexpr bool is_mut_iter = std::is_same_v<Iter, DbSlice::Iterator>;
+  DCHECK(!(cid->IsReadOnly() && is_mut_iter))
+      << "mutable iterator used with read-only command " << cid->name();
+
   MGetResponse response(keys.Size());
   struct Item {
-    DbSlice::ConstIterator it;
+    Iter it;
     int source_index = -1;  // in case of duplicate keys, points to the first occurrence.
   };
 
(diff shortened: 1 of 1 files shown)


src/server/string_family.cc
```diff
@@ -536,16 +536,31 @@ struct MGetResponse {
...
   ShardArgs keys = t->GetShardArgs(shard->shard_id());
   DCHECK(!keys.Empty());
 
   auto& db_slice = t->GetDbSlice(shard->shard_id());
 
+  const auto cid = t->GetCId();
+  constexpr bool is_mut_iter = std::is_same_v<Iter, DbSlice::Iterator>;
+  DCHECK(!(cid->IsReadOnly() && is_mut_iter))
+      << "mutable iterator used with read-only command " << cid->name();
+
```
romange:
```suggestion
  constexpr bool is_mut_iter = std::is_same_v<Iter, DbSlice::Iterator>;
  if constexpr (is_mut_iter) {
    const auto* cid = t->GetCId();
    DCHECK(!cid->IsReadOnly()) << "mutable iterator used with read-only command " << cid->name();
  }
```

src/server/string_family.cc
```diff
@@ -536,16 +536,31 @@ struct MGetResponse {
...
+// Const find operation, for read-only commands, eg MGet
+using SearchConst = SearchKey<DbSlice::ConstIterator>;
+
+template <typename Iter>
+MGetResponse CollectKeys(BlockingCounter wait_bc, uint8_t fetch_mask, const Transaction* t,
+                         EngineShard* shard, SearchKey<Iter> find_op) {
   ShardArgs keys = t->GetShardArgs(shard->shard_id());
   DCHECK(!keys.Empty());
 
   auto& db_slice = t->GetDbSlice(shard->shard_id());
```
romange: move `db_slice` to the block where you use it

src/server/string_family.cc
```diff
@@ -1294,12 +1319,41 @@ void StringFamily::DecrBy(CmdArgList args, const CommandContext& cmnd_cntx) {
   return IncrByGeneric(key, -val, cmnd_cntx.tx, cmnd_cntx.rb);
 }
 
+absl::FixedArray<optional<GetResp>, 8> ReorderShardResults(
```
romange: `absl::FixedArray<optional<GetResp>, 8>` is more than 512 bytes - and it's not moveable, meaning that now we always have to copy at least 512 bytes.
Instead, I suggest the following signature:
`void ReorderShardResults(const std::vector<MGetResponse>& mget_resp, const Transaction* t, absl::FixedArray<optional<GetResp>, 8>* dest);` 

and define res out side of the function.

src/server/string_family.cc
abhijat: moved

src/server/string_family.cc
abhijat: fixed

src/server/string_family.cc
abhijat: changed

romange:
I do not see changes @abhijat

abhijat:
> I do not see changes @abhijat

sorry, pushed commit to wrong branch, should be visible now

src/server/string_family.cc
```diff
@@ -1294,12 +1319,39 @@ void StringFamily::DecrBy(CmdArgList args, const CommandContext& cmnd_cntx) {
   return IncrByGeneric(key, -val, cmnd_cntx.tx, cmnd_cntx.rb);
 }
 
+void ReorderShardResults(const std::vector<MGetResponse>& mget_resp, const Transaction* t,
+                         absl::FixedArray<optional<GetResp>, 8>* dest,
```
romange: we always put the output arguments at the end. so in this case `dest` should be the last.

src/server/string_family.cc
```diff
@@ -536,16 +536,30 @@ struct MGetResponse {
...
+template <typename Iter>
+MGetResponse CollectKeys(BlockingCounter wait_bc, uint8_t fetch_mask, const Transaction* t,
+                         EngineShard* shard, SearchKey<Iter> find_op) {
   ShardArgs keys = t->GetShardArgs(shard->shard_id());
   DCHECK(!keys.Empty());
 
-  auto& db_slice = t->GetDbSlice(shard->shard_id());
+  if constexpr (constexpr bool is_mut_iter = std::is_same_v<Iter, DbSlice::Iterator>) {
+    const auto cid = t->GetCId();
+    DCHECK(!(cid->IsReadOnly() && is_mut_iter))
```
romange: you do not need to check `is_mut_iter` - you are already under `if constexpr`

src/server/string_family.cc
```diff
@@ -536,16 +536,30 @@ struct MGetResponse {
...
+        << "mutable iterator used with read-only command " << cid->name();
+  }
```
romange:
```suggestion
  if constexpr (std::is_same_v<Iter, DbSlice::Iterator>) {
    const auto* cid = t->GetCId();
    DCHECK(!cid->IsReadOnly())
        << "mutable iterator used with read-only command " << cid->name();
  }
```

src/server/string_family.cc
abhijat: changed

src/server/string_family.cc
abhijat: fixed

src/server/string_family.cc
abhijat: `auto` already deduces `cid` as a pointer to const, but I have changed it to use explicit type.

romange (approved):

PR #5297, opened 2025-06-13
fix: partial replication data loss

### LINKED ISSUE #5275: test_partial_sync failed

kostasrim:
The issue is that each journal entry is not always executed by the replica yet we increment the journal records executed. An example of that is when the context gets cancelled (e,g connection lost from master) before the command dispatches for execution. However, when the replica reconnects it will start from the next lsn, ignoring the journal entry which never dispatched leading to data loss. 

What's described was reproduced in https://github.com/dragonflydb/dragonfly/pull/5277

resolves #5275

### DIFF
--- a/src/server/journal/executor.h
+++ b/src/server/journal/executor.h
@@ -7,6 +7,7 @@
 #include <absl/types/span.h>
 
 #include "facade/reply_capture.h"
+#include "facade/service_interface.h"
 #include "server/cluster/cluster_defs.h"
 #include "server/journal/types.h"
 
@@ -22,8 +23,8 @@ class JournalExecutor {
 
   JournalExecutor(JournalExecutor&&) = delete;
 
-  void Execute(DbIndex dbid, absl::Span<journal::ParsedEntry::CmdData> cmds);
-  std::error_code Execute(DbIndex dbid, journal::ParsedEntry::CmdData& cmd);
+  // Return true is command executed without error
+  facade::DispatchResult Execute(DbIndex dbid, journal::ParsedEntry::CmdData& cmd);
 
   void FlushAll();  // Execute FLUSHALL.
   void FlushSlots(const cluster::SlotRange& slot_range);
@@ -33,7 +34,7 @@ class JournalExecutor {
   }
 
  private:
-  std::error_code Execute(journal::ParsedEntry::CmdData& cmd);
+  facade::DispatchResult Execute(journal::ParsedEntry::CmdData& cmd);
 
   // Select database. Ensure it exists if accessed for first time.
   void SelectDb(DbIndex dbid);
--- a/src/server/replica.cc
+++ b/src/server/replica.cc
@@ -6,6 +6,7 @@
 #include <chrono>
 
 #include "absl/strings/match.h"
+#include "facade/service_interface.h"
 
 extern "C" {
 #include "redis/rdb.h"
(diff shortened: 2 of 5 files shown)


src/server/replica.cc
```diff
@@ -913,8 +914,17 @@ void DflyShardReplica::StableSyncDflyReadFb(ExecutionState* cntx) {
...
+      auto exe_res = ExecuteTx(std::move(*tx_data), cntx);
+      if (exe_res) {
+        // We only increment upon successful execution of the transaction.
+        // The reason for this is that during partial sync we sent this
+        // number as the lsn number to resume from. However, if for example
+        // we increment this when a command fails (because the context
+        // got cancelled, e.g, replication connection broke), we will get
+        // inconsistent data because the replica will resume from the next
+        // lsn of the master and this lsn entry will be lost.
+        journal_rec_executed_.fetch_add(1, std::memory_order_relaxed);
```
kostasrim, line 926: I believe there is a general issue here because `lsn` and `records executed` might not be 1-1. If that's the case we might have data loss under certain scenarios. 

// TODO investigate this

src/server/replica.h
```diff
@@ -207,7 +207,7 @@ class DflyShardReplica : public ProtocolClient {
 
   void StableSyncDflyAcksFb(ExecutionState* cntx);
 
-  void ExecuteTx(TransactionData&& tx_data, ExecutionState* cntx);
+  bool ExecuteTx(TransactionData&& tx_data, ExecutionState* cntx);
```
romange, line 212: pls add a comment about the return value

src/server/replica.cc
```diff
@@ -913,8 +914,17 @@ void DflyShardReplica::StableSyncDflyReadFb(ExecutionState* cntx) {
       force_ping_ = true;
       journal_rec_executed_.fetch_add(1, std::memory_order_relaxed);
     } else {
-      ExecuteTx(std::move(*tx_data), cntx);
-      journal_rec_executed_.fetch_add(1, std::memory_order_relaxed);
+      auto exe_res = ExecuteTx(std::move(*tx_data), cntx);
```
romange: please do not use `auto` for POD types or one word types - it reduces the readability

src/server/replica.cc
```diff
@@ -1017,26 +1026,27 @@ void DflyShardReplica::ExecuteTx(TransactionData&& tx_data, ExecutionState* cntx
...
   VLOG(2) << "Execute txid: " << tx_data.txid << " global command execution";
   // Wait until all shards flows get to execution step of this transaction.
   multi_shard_data.barrier.Wait();
   // Check if we woke up due to cancellation.
   if (!exec_st_.IsRunning())
-    return;
+    return false;
   // Global command will be executed only from one flow fiber. This ensure corectness of data in
   // replica.
+  bool execution_res = false;
```
romange:
```suggestion
  bool execution_res = true;
```

src/server/replica.cc
```diff
@@ -1046,6 +1056,7 @@ void DflyShardReplica::ExecuteTx(TransactionData&& tx_data, ExecutionState* cntx
   if (val == 1) {
     multi_shard_exe_->Erase(tx_data.txid);
   }
+  return inserted_by_me ? execution_res : true;
```
romange:
```suggestion
  return execution_res;
```

src/server/replica.cc
kostasrim: accidental, I am trying to avoid it in general as you suggest

src/server/journal/executor.h
```diff
@@ -22,8 +23,8 @@ class JournalExecutor {
 
   JournalExecutor(JournalExecutor&&) = delete;
 
-  void Execute(DbIndex dbid, absl::Span<journal::ParsedEntry::CmdData> cmds);
-  std::error_code Execute(DbIndex dbid, journal::ParsedEntry::CmdData& cmd);
+  // Return true is command executed without error
```
romange: eeh, something is not right here.

src/server/journal/executor.h
```diff
@@ -22,8 +23,8 @@ class JournalExecutor {
 
   JournalExecutor(JournalExecutor&&) = delete;
 
-  void Execute(DbIndex dbid, absl::Span<journal::ParsedEntry::CmdData> cmds);
```
romange, line 25: so we've never used this variant with Span?

src/server/journal/executor.h
kostasrim: typos, fixed

src/server/journal/executor.h
kostasrim: yes we did not use that overload. At least that's what find references in my vim says. IMO, a span would not implicitly convert to a single CmdData so we would fail to compile if we accidentally removed this and there were actual calls to it.

src/server/journal/executor.h
```diff
@@ -22,8 +23,8 @@ class JournalExecutor {
...
-  std::error_code Execute(DbIndex dbid, journal::ParsedEntry::CmdData& cmd);
+  // Return true if the command executed successfully and false otherwise.
```
romange: please fix the comment

romange (approved):

Issue #5287, opened 2025-06-12
Unsupported status 4


vyavdoshenko:
Error in the log:
```
Unsupported status 4
```

The reason:
Unhandled status - OpStatus::SKIPPED

dranikpg:
Any log source location in code?

vyavdoshenko:
It should be figured out. I found it using fuzzing testing.
https://github.com/dragonflydb/df-afl

It should be fixed as a previous one: https://github.com/dragonflydb/dragonfly/issues/5222
https://github.com/dragonflydb/dragonfly/pull/5223
Find all the places in the code and find out which ones are not handled.

vyavdoshenko:
@dranikpg 
BTW, I guess the design of OpStatus is not elegant because we mix up internal states and public ones. I suppose it should be better to split them up. It would be impossible to catch such errors because we would process all the statuses in code and could catch the errors during compile time.

vyavdoshenko:
This issue can be reproduced with https://github.com/dragonflydb/df-afl. Set the focus on stream commands.
```
export REDIS_FOCUS_COMMANDS="XADD,XCLAIM,XDEL,XGROUP,XINFO,XLEN,XPENDING,XRANGE,XREVRANGE,XREAD,XREADGROUP,XSETID,XTRIM,_XGROUP_HELP,XACK,XAUTOCLAIM"
```

I debugged locally. The cause is the XCLAIM command.

PR #5300, opened 2025-06-14
fix: Unsupported status 4 (XCLAIM command)

vyavdoshenko:
Fixes: https://github.com/dragonflydb/dragonfly/issues/5287

- Added handling of missed internal status
- The unit test was not added because I couldn't find a scenario to reproduce it without fuzzing.


romange:
@vyavdoshenko  seems like something we should easily add to the unit test. OpClaim calls FindGroup and FindGroup can return SKIPPED if consumer group is not found.

Issue #5255, opened 2025-06-09
Fix corruption in the stream code TODO.


vyavdoshenko:
**The original idea**:
The idea is based on the fact that the stream code writes somehow lp size using lpSetTotalBytes into already freed block. Before lpSetTotalBytes corrupts the memory, it should have low 32 bits of the "next" field and based on the examples we have it should be pretty large number or 0 (in case block->next is null).

in any case we read first with lpGetTotalBytes and if it is say 0 - it's a bug (we crash) because lp can not have 0 bytes (it has at least 7 bytes meta data) and if it's higher 16MB (highly unlikely) we call zmalloc_usable_size and check that the usable size is equal or higher than lpGetTotalBytes (consistency). and if not we abort (crash)

The PR todo was added:
https://github.com/dragonflydb/dragonfly/pull/5252

The original crash log:
```
@     0xaaaaab3b2f18        192  boost::context::detail::fiber_entry<>() |  
-- | --
  |   | @     0xaaaaab3b291c        576  util::ListenerInterface::RunSingleConnection() |  
  |   | @     0xaaaaaafd9a64        368  facade::Connection::HandleRequests() |  
  |   | @     0xaaaaaafd8b18        272  facade::Connection::ConnectionFlow() |  
  |   | @     0xaaaaaafd85ac       1680  facade::Connection::IoLoop() |  
  |   | @     0xaaaaaafd76bc        304  facade::Connection::ParseRedis() |  
  |   | @     0xaaaaaafd716c        464  facade::Connection::DispatchSingle() |  
  |   | @     0xaaaaaac18914        672  dfly::Service::DispatchCommand() |  
  |   | @     0xaaaaaac12aa4        208  dfly::Service::InvokeCmd() |  
  |   | @     0xaaaaaaeea4d4        672  dfly::CommandId::Invoke() |  
  |   | @     0xaaaaaace88d8         32  dfly::StreamFamily::XAdd() |  
  |   | @     0xaaaaaaf09fe0         64  dfly::Transaction::ScheduleSingleHop() |  
  |   | @     0xaaaaaaf09f48        416  dfly::Transaction::Execute() |  
  |   | @     0xaaaaaaf09684        592  dfly::Transaction::ScheduleInternal() |  
  |   | @     0xaaaaaaf082f0        288  dfly::Transaction::ScheduleInShard() |  
  |   | @     0xaaaaaaf07b08        208  dfly::Transaction::RunCallback() |  
[...]

PR #5301, opened 2025-06-15
Revert "fix(stream_family): Fix stream memory tracking issues (#5024)"

romange:
This reverts commit 74524415c3f5076ac82ae182f76c3af57b6cafa1.

This is done because of the possible regression that causes #5255

### DIFF
--- a/src/server/stream_family.cc
+++ b/src/server/stream_family.cc
@@ -772,52 +772,41 @@ OpResult<streamID> OpAdd(const OpArgs& op_args, string_view key, const AddOpts& 
   return result_id;
 }
 
-/* This method modifies opts->group and opts->consumer by inserting data into them.
-   Since these structures are stored within the stream, the stream itself is also modified.
-   Therefore, we accept opts as a pointer to make this mutation explicit. */
-OpResult<RecordVec> OpRange(const OpArgs& op_args, string_view key, RangeOpts* opts) {
+OpResult<RecordVec> OpRange(const OpArgs& op_args, string_view key, const RangeOpts& opts) {
   auto& db_slice = op_args.GetDbSlice();
-  auto res_it = db_slice.FindMutable(op_args.db_cntx, key, OBJ_STREAM);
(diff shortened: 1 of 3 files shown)


vyavdoshenko (approved):
LGTM

Is it possible to find out which issue should be reopened after that?

BagritsevichStepan (commented):
Wait, why memory tracking causes this regression? I don't see how it should be related. After you revert this can you reproduce the crash?

src/server/stream_family.cc
```diff
@@ -772,52 +772,41 @@ OpResult<streamID> OpAdd(const OpArgs& op_args, string_view key, const AddOpts&
   return result_id;
 }
 
-/* This method modifies opts->group and opts->consumer by inserting data into them.
-   Since these structures are stored within the stream, the stream itself is also modified.
-   Therefore, we accept opts as a pointer to make this mutation explicit. */
-OpResult<RecordVec> OpRange(const OpArgs& op_args, string_view key, RangeOpts* opts) {
+OpResult<RecordVec> OpRange(const OpArgs& op_args, string_view key, const RangeOpts& opts) {
   auto& db_slice = op_args.GetDbSlice();
-  auto res_it = db_slice.FindMutable(op_args.db_cntx, key, OBJ_STREAM);
```
BagritsevichStepan, line 780: Why we should revert this? It is not correct that this is ReadOnly operation and I wrote the comment why it is mutable operation

src/server/stream_family.cc
```diff
@@ -1491,7 +1482,7 @@ OpResult<uint32_t> OpDelConsumer(const OpArgs& op_args, string_view key, string_
   StreamMemTracker mem_tracker;
 
   long long pending = 0;
-  streamConsumer* consumer = streamLookupConsumer(cgroup_res->cg, SdsWrapper(consumer_name));
+  streamConsumer* consumer = streamLookupConsumer(cgroup_res->cg, WrapSds(consumer_name));
```
BagritsevichStepan, line 1485: Why we should revert this? WrapSds makes a bug here

Issue #5014, opened 2025-04-28
Prove that the general key commands invoke AddDoc and RemoveDoc on the indexes


BagritsevichStepan:
Following this PR #4995 , we should prove that all general/generic key commands (that change the key value) invoke `AddDoc` and `RemoveDoc` on the indexes right. Add this checks to the tests

BagritsevichStepan:
The same bug was found in StringFamily::Set command. Issue: https://github.com/dragonflydb/dragonfly/issues/5215

BagritsevichStepan:
TODO list:

- [x] Prove all StringFamily::* commands
- [x] Prove all GenericFamily::* command
- [x] Prove other families

BagritsevichStepan:
There is the same bug in the ZUNIONSTORE command

BagritsevichStepan:
And there is the same bug in the `OpAdd` in `set_family.cc` if overwrite is true

PR #5306, opened 2025-06-16
fix(string_family): Add RemoveDoc for the SET command

BagritsevichStepan:
fixes dragonflydb#5215

The problem was that the `StringFamily::Set` command could change the key's type. If the previous type was `HSET` or `JSON`, the key was indexed. When `StringFamily::Set` is called on such a key, it changes the type from HSET/JSON to STRING, but we forgot to remove the key from the indexes.

This PR adds a call to `RemoveDoc` inside `StringFamily::Set` to properly clean up the old index entry.

A snippet that reproduces the issue has been added to the tests, along with explanatory comments:
```
127.0.0.1:6379> ft.create index on hash schema field TEXT
OK
127.0.0.1:6379> hset k1 field value
(integer) 1
127.0.0.1:6379> set k1 anothervalue
OK
127.0.0.1:6379> rename k1 anotherkey
OK
127.0.0.1:6379> hset k1 field value
Error: Server closed the connection
(1.29s)
```

### DIFF
--- a/src/server/string_family.cc
+++ b/src/server/string_family.cc
@@ -30,6 +30,7 @@
 #include "server/error.h"
 #include "server/generic_family.h"
 #include "server/journal/journal.h"
+#include "server/search/doc_index.h"
 #include "server/table.h"
 #include "server/tiered_storage.h"
 #include "server/transaction.h"
@@ -111,8 +112,8 @@ class SetCmd {
   OpStatus Set(const SetParams& params, std::string_view key, std::string_view value);
 
  private:
-  OpStatus SetExisting(const SetParams& params, DbSlice::Iterator it, DbSlice::ExpIterator e_it,
-                       std::string_view key, std::string_view value);
+  OpStatus SetExisting(const SetParams& params, const DbContext& db_cntx, DbSlice::Iterator it,
+                       DbSlice::ExpIterator e_it, std::string_view key, std::string_view value);
 
   void AddNew(const SetParams& params, DbSlice::Iterator it, std::string_view key,
               std::string_view value);
@@ -831,7 +832,7 @@ OpStatus SetCmd::Set(const SetParams& params, string_view key, string_view value
 
     if (params.flags & SET_IF_EXISTS) {
       if (IsValid(find_res.it)) {
-        return SetExisting(params, find_res.it, find_res.exp_it, key, value);
+        return SetExisting(params, op_args_.db_cntx, find_res.it, find_res.exp_it, key, value);
       } else {
         return OpStatus::SKIPPED;
       }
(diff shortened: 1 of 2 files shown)


vyavdoshenko:
@BagritsevichStepan 
Could you please add a more descriptive description of the root cause and how it was fixed?

BagritsevichStepan:
> @BagritsevichStepan Could you please add a more descriptive description of the root cause and how it was fixed?

Done

BagritsevichStepan:
@vyavdoshenko I spoke about it today and we already have an issue for this #5014 . I have a solution

src/server/string_family.cc
```diff
@@ -111,8 +112,8 @@ class SetCmd {
   OpStatus Set(const SetParams& params, std::string_view key, std::string_view value);
 
  private:
-  OpStatus SetExisting(const SetParams& params, DbSlice::Iterator it, DbSlice::ExpIterator e_it,
-                       std::string_view key, std::string_view value);
+  OpStatus SetExisting(const SetParams& params, const DbContext& db_cntx, DbSlice::Iterator it,
```
romange: you do not need to pass `const DbContext& db_cntx` you already have it as ` op_args_.db_cntx`

src/server/string_family.cc
```diff
@@ -831,7 +832,7 @@ OpStatus SetCmd::Set(const SetParams& params, string_view key, string_view value
 
     if (params.flags & SET_IF_EXISTS) {
       if (IsValid(find_res.it)) {
-        return SetExisting(params, find_res.it, find_res.exp_it, key, value);
+        return SetExisting(params, op_args_.db_cntx, find_res.it, find_res.exp_it, key, value);
```
romange: no need to pass the class variable here.

src/server/string_family.cc
```diff
@@ -892,6 +894,9 @@ OpStatus SetCmd::SetExisting(const SetParams& params, DbSlice::Iterator it,
   prime_value.SetFlag(params.memcache_flags != 0);
   db_slice.SetMCFlag(op_args_.db_cntx.db_index, it->first.AsRef(), params.memcache_flags);
 
+  // We need to remove the key from search indices, because we are overwriting it to OBJ_STRING
+  shard->search_indices()->RemoveDoc(key, db_cntx, prime_value);
```
romange: i do not like this unconditional approach. please check before if `prime_value.ObjType()` is indexable (i.e. either hset or json). You can introduce a one-liner utility function  for better readability (add it to family_utils.h).

BagritsevichStepan: I know, we have this problem in several places (check for example rename command). Let's do it in another PR, I will fix other commands as well.

romange:
most of our write commands are not blind writes so the issue is not wide to begin with.

BagritsevichStepan:
I know for example zunionstore command, that also changes the key type

BagritsevichStepan:
And there is the same bug in the `OpAdd` in `set_family.cc` if overwrite is true

src/server/string_family.cc
romange: ok, please do not forget

romange (approved):

Issue #5294, opened 2025-06-13
RdbSBF serialization/deserialization problem


BorysTheDev:
we get invalid_value during deseriazliation of RdbSbf

kostasrim:
Have you considered if the sbf we try to load is actually valid or not?  If so how did you verify this?

BorysTheDev:
> Have you considered if the sbf we try to load is actually valid or not? If so how did you verify this?

I didn't dive deep, it happens during RENAMENX command

PR #5307, opened 2025-06-16
fix: prevent rename cmd crash

BorysTheDev:
fixes: #5295
problem: we could return an error status for the deserialization operation during the transaction

### DIFF
--- a/src/server/generic_family.cc
+++ b/src/server/generic_family.cc
@@ -404,7 +404,11 @@ OpStatus Renamer::FinalizeRename() {
 
   transaction_->Execute(std::move(cb), true);
 
-  DCHECK(del_status == OpStatus::OK);
+  LOG_IF(DFATAL,
+         (deserialize_status != OpStatus::OK && deserialize_status != OpStatus::OUT_OF_MEMORY) ||
+             del_status != OpStatus::OK)
+      << "Error during rename command, deserialize_status: " << deserialize_status
+      << " del_status: " << del_status;
   return deserialize_status != OpStatus::OK ? deserialize_status : del_status;
 }
 
--- a/src/server/generic_family.cc
+++ b/src/server/generic_family.cc
@@ -316,7 +316,7 @@ class Renamer {
 
  private:
   void FetchData();
-  void FinalizeRename();
+  facade::OpStatus FinalizeRename();
 
   bool KeyExists(Transaction* t, EngineShard* shard, std::string_view key) const;
   void SerializeSrc(Transaction* t, EngineShard* shard);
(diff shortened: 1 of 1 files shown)


src/server/generic_family.cc
```diff
@@ -389,22 +388,24 @@ void Renamer::FetchData() {
...
+  auto cb = [&](Transaction* t, EngineShard* shard) {
     const ShardId shard_id = shard->shard_id();
 
     if (!do_copy_ && shard_id == src_sid_) {
-      return DelSrc(t, shard);
-    }
-
-    if (shard_id == dest_sid_) {
-      return DeserializeDest(t, shard);
+      del_status = DelSrc(t, shard);
```
romange, line 398: I do not understand this change. based on this code, either `del` or `deserialize` can happen but not both. why do you need two variables for that?  how this code behaves differently now?

BorysTheDev: we `transaction_->Execute(std::move(cb), true); ` so both can happen but in different shards. it behavers differently because the returned status for local_result inside the transaction is always OK, we have CHECK() there. And we return status based onm deserialization operation

kostasrim: @romange This flow is when src and destination belong to different shards. The first if handles the `deletion` part and the second if handles the `renaming via loading of the rdb`. 

My concerns are:

First, if this code fails as claimed in the issue, it means we serialized something we can't deserialize. We have a bug there which I believe should be the first one to address.

Second, we no longer crash but we still have a partial state:

`we unconditionally delete src yet we might fail to deserialize on the destination`. This is data loss. With the CHECK, we crashed -- now we don't anymore leading to inconsistent state of a datastore *without us noticing* and I don't like this.

Third, for isolated fixes like this I expect at least a unit test

BorysTheDev: @kostasrim for now we are ok with the lack of atomicity guarantee, we have discussed it with @romange. I will fix serialization/deserialization in another PR, I have a separate ticket for it #5294. Even if we merge this fix we still have LOG(ERROR) for serialization/deserialization

romange: Yeah, I spoke to @BorysTheDev  he will look into the deserialization issue.

unfortunately, I do not see how we can test it in the unit test - since we provide a protection here against internal bugs that should not exist in the first place.

src/server/generic_family.cc
```diff
@@ -389,22 +388,24 @@ void Renamer::FetchData() {
...
+    } else if (shard_id == dest_sid_) {
+      deserialize_status = DeserializeDest(t, shard);
     }
-
     return OpStatus::OK;
   };
 
   transaction_->Execute(std::move(cb), true);
+
+  DCHECK(del_status == OpStatus::OK);
```
romange: following @kostasrim  comment, please add `LOG_IF(DFATAL, deserialize_status != OpStatus::OK && del_status !=  OpStatus::OK) << ...`

romange: instead of DCHECK

BorysTheDev: I think OOM too should be here

PR #5308, opened 2025-06-16
feat(replication): Do replication from periodic fiber

### LINKED ISSUE #5293: Replication is done from periodic fiber
First improvement is to implement replication loop to run from periodic fiber.

mkaruza:
For each streamer object create periodic fiber that would do replication on interval basis. If there is influx of data we are would increase frequency of replication until pending data is decreased again.

Closes #5293


src/server/journal/streamer.cc
```diff
@@ -75,30 +76,77 @@ void JournalStreamer::Start(util::FiberSocketBase* dest) {
...
-  DCHECK(!pending_buf_.Empty());
+void JournalStreamer::Write(std::string str) {
+  DCHECK(!str.empty());
+  DVLOG(3) << "Writing " << str.size() << " bytes";
 
-  if (in_flight_bytes_ > 0) {
-    // We can not flush data while there are in flight requests because AsyncWrite
-    // is not atomic. Therefore, we just aggregate.
-    return;
+  pending_buf_.Push(std::move(str));
```
romange, line 105: There are two changes here that I disagree with:
1. You removed flushing from `JournalStreamer::Write` and you always rely on the periodic fiber. It's not necessary in case your pending_buf_ is big enough and `in_flight_bytes_ == 0 ` - you can flush from here.
2. You removed `WaitForInflightToComplete` which means you completely removed the back-pressure mechanism, so we will now OOM if say the TCP got stuck or replica became slow or 100 other reasons why the master producing replication data quicker than the streamer consumes it. We deliberately introduced back pressure because we had OOM situations due to all this and we worked hard to eliminate them.

romange: both points are related to lack of back-pressure

mkaruza: 1. This could be valid path for dispatching data outside fiber.
2. Back pressure is till there with `ThrottleIfNeeded` which was not changed. `WaitForInflightToComplete` was used only with `Cancel` to wait that previous async write finishes.

romange: Ok regarding (2), but I did not understand  (1).

mkaruza: @romange this is valid suggestion that replication happens directly from `JournalStreamer::Write` but with check if `pending_buf` is bigger than some threshold. And this fiber can be used to dispatch `pending_buf` after some shard inactivity (without checking threshold so we still replicate in reasonable time)

mkaruza:
I have updated benchmark for threshold replication in #5293

mkaruza:
Tested with increased threshold (1500, 4096). Updated in #5293

src/server/dflycmd.cc
```diff
@@ -658,7 +658,8 @@ void DflyCmd::StartStableSyncInThread(FlowInfo* flow, ExecutionState* exec_st, E
   DCHECK(shard);
   DCHECK(flow->conn);
 
-  flow->streamer.reset(new JournalStreamer(sf_->journal(), exec_st, JournalStreamer::SendLsn::YES));
+  flow->streamer.reset(
+      new JournalStreamer(sf_->journal(), exec_st, JournalStreamer::SendLsn::YES, true));
```
romange, line 662: why do you need `is_stable_sync` argument if it's always true?

src/server/journal/pending_buf.h
```diff
@@ -20,7 +20,7 @@ class PendingBuf {
 #ifdef UIO_MAXIOV
     static constexpr size_t kMaxBufSize = UIO_MAXIOV;
 #else
-    static constexpr size_t kMaxBufSize = 1024;
+    static constexpr size_t kMaxBufSize = 1500;  // 1024;
```
romange: remove `// 1024;`

src/server/dflycmd.cc
mkaruza: Afaict, RestoreStreamer uses JournalStreamer to replicat data and it expects immediate response back (so we set there to `false`). Data that is sent is less than threshold so it is slow process (i had several test failing). So for RestoreStreamer we don't do this aggregation and behavior is as before (dispatch data immediately) - while for stable sync replication we follow this new logic.

src/server/journal/pending_buf.h
mkaruza: I will. I'm considering this to be converted to flag.

src/server/dflycmd.cc
romange: ah, i missed that

src/server/journal/pending_buf.h
mkaruza: Added flag to control threshold

PR #5312, opened 2025-06-16
fix(json): JSON.DEL recursive descent duplicate counting for nested same-key elements

vyavdoshenko:
Fixes: https://github.com/dragonflydb/dragonfly/issues/5259

Issue:
`JSON.DEL doc "$..a"` was returning 2 instead of 1 when deleting nested objects with the same key name, because both parent and child deletions were counted separately.

Solution:
Added filtering logic to avoid counting child nodes that are automatically deleted when their parent is removed. Only applies to recursive descent deletion operations where one object directly contains another with the same target key.

Testing:
- A unit test was added
- Fakeredis test was checked `poetry run pytest -s test/test_json/test_json.py  -k test_json_delete_with_dollar`

### DIFF
--- a/src/core/json/detail/jsoncons_dfs.cc
+++ b/src/core/json/detail/jsoncons_dfs.cc
@@ -150,9 +150,6 @@ Dfs Dfs::Delete(absl::Span<const PathSegment> path, JsonType* json) {
   vector<Item> stack;
   stack.emplace_back(json);
 
-  // Collect nodes to delete - but simplified without complex filtering
-  std::vector<JsonType*> nodes_to_delete;
-
   do {
     unsigned segment_index = stack.back().segment_idx();
     const auto& path_segment = path[segment_index];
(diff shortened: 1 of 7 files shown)


src/server/json_family_test.cc
```diff
@@ -1261,6 +1261,24 @@ TEST_F(JsonFamilyTest, Del) {
 
   resp = Run({"JSON.GET", "json"});
   EXPECT_THAT(resp, ArgType(RespExpr::NIL));
+
+  if (absl::GetFlag(FLAGS_jsonpathv2)) {
```
romange, line 1265: can you please add a test in jsonpath_test.cc as well?

vyavdoshenko: The unit test was added to jsonpath_test.cc

src/server/json_family.cc
```diff
@@ -1060,7 +1060,7 @@ OpResult<long> OpDel(const OpArgs& op_args, string_view key, string_view path,
     const json::Path& path = json_path.AsJsonPath();
     long deletions = json::MutatePath(
         path, [](optional<string_view>, JsonType* val) { return true; }, updater.GetJson(),
-        true /* reverse_traversal */);
```
romange, line 1063: As you can see the fix becomes more and more complicated. This is the right time to revisit the whole approach.

I think that the solution is to introduce a separate method `json::DeletePath(path, updater.GetJson())` without the callback (as you can see here it's NOOP anyways).
in that case `json::DeletePath` should be much simpler because you do not need to iterate in reverse as there is no callback to change upper nodes in-place (you just delete whatever is matched).
In fact, you do not need `nodes_to_mutate` in `json::DeletePath`

src/core/json/detail/jsoncons_dfs.cc
```diff
@@ -128,17 +128,100 @@ Dfs Dfs::Mutate(absl::Span<const PathSegment> path, const MutateCallback& callba
...
+    dfs.DeleteStep(path[0], json);
+    return dfs;
+  }
+
+  using Item = detail::JsonconsDfsItem<false>;
+  vector<Item> stack;
+  stack.emplace_back(json);
+
+  // Collect nodes to delete - but simplified without complex filtering
+  std::vector<JsonType*> nodes_to_delete;
```
romange: why do we need `nodes_to_delete` ? what exactly breaks if you delete during the traversal?

src/core/json/detail/jsoncons_dfs.cc
```diff
@@ -128,17 +128,100 @@ Dfs Dfs::Mutate(absl::Span<const PathSegment> path, const MutateCallback& callba
...
+      }
+      stack.pop_back();
     }
-  } else {
-    for (auto it = nodes_to_mutate.begin(); it != nodes_to_mutate.end(); ++it) {
-      dfs.MutateStep(terminal_segment, callback, *it);
+  } while (!stack.empty());
+
+  // Apply deletions with filtering for recursive descent to avoid parent-child duplication
+  const PathSegment& terminal_segment = path.back();
```
romange: if we delete immediately without `nodes_to_delete`, we won't need all this code below, since if we delete parent first, we won't traverse its children.

vyavdoshenko: Yes, it seems, you are exactly right.

src/core/json/detail/jsoncons_dfs.cc
vyavdoshenko: fixed

PR #5315, opened 2025-06-16
fix: memory tracking bug in rename and SET

### LINKED ISSUE #5311: Memory accounting is broken for single shard RENAME  or blind writes
to reproduce:
1. run dragonfly with `--proactor_threads=1` though it can be shown for any number of threads.
2. `lpush a foo` and then `info memory` to see how memory by type is updated correctly 
3. `rename a b`  and then `info memory` to see how memory for strings and lists is completely off.

The same happens for any blind write operations cc @BagritsevichStepan for example
1. `lpush a foo` and then `info memory` to see how memory by type is updated correctly 
2. `set a barrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrr` and then `info memory` to see how memory by type is completely off.

romange:
Fixes #5311

### DIFF
--- a/src/server/generic_family.cc
+++ b/src/server/generic_family.cc
@@ -894,6 +894,7 @@ OpResult<void> OpRen(const OpArgs& op_args, string_view from_key, string_view to
 
   bool sticky = from_res.it->first.IsSticky();
   uint64_t exp_ts = db_slice.ExpireTime(from_res.exp_it);
+  from_res.post_updater.ReduceHeapUsage();
 
   // we keep the value we want to move.
   PrimeValue from_obj = std::move(from_res.it->second);
@@ -902,6 +903,7 @@ OpResult<void> OpRen(const OpArgs& op_args, string_view from_key, string_view to
   from_res.it->second.SetExpire(IsValid(from_res.exp_it));
 
   if (IsValid(to_res.it)) {
+    to_res.post_updater.ReduceHeapUsage();
     to_res.it->second = std::move(from_obj);
     to_res.it->second.SetExpire(IsValid(to_res.exp_it));  // keep the expire flag on 'to'.
 
--- a/src/server/generic_family_test.cc
+++ b/src/server/generic_family_test.cc
@@ -21,7 +21,6 @@ extern "C" {
 using namespace testing;
 using namespace std;
 using namespace util;
-using namespace boost;
 using absl::StrCat;
 
 namespace dfly {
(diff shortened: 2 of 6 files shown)


src/server/generic_family_test.cc
```diff
@@ -453,14 +452,33 @@ TEST_F(GenericFamilyTest, Rename) {
   ren_fb.Join();
 }
 
-TEST_F(GenericFamilyTest, RenameNonString) {
-  EXPECT_EQ(1, CheckedInt({"lpush", "x", "elem"}));
-  auto resp = Run({"rename", "x", "b"});
-  ASSERT_EQ(resp, "OK");
-  ASSERT_EQ(2, last_cmd_dbg_info_.shards_count);
+TEST_F(GenericFamilyTest, RenameList) {
+  for (string_view dest : {"b", "y", "z"}) {
```
romange, line 456: running the same test for differrent destination keys - "b" on differrent shard; "y", "z" on the same shard.

checking that the memory usage is tracked as expected.

src/server/string_family.cc
```diff
@@ -851,15 +853,16 @@ OpStatus SetCmd::Set(const SetParams& params, string_view key, string_view value
     if (auto status = CachePrevIfNeeded(params, op_res->it); status != OpStatus::OK)
       return status;
 
+    op_res->post_updater.ReduceHeapUsage();
     return SetExisting(params, op_res->it, op_res->exp_it, key, value);
```
BagritsevichStepan: Consider removing `Iterator` and `ExpIterator` from the `SetExisting` and instead move the entire `op_res` iterator (or `find_res`) to this method. This way, we can call `post_updater.ReduceHeapUsage()` directly inside `SetExisting`, avoiding boilerplate code.

src/server/generic_family.cc
```diff
@@ -902,6 +903,7 @@ OpResult<void> OpRen(const OpArgs& op_args, string_view from_key, string_view to
   from_res.it->second.SetExpire(IsValid(from_res.exp_it));
 
   if (IsValid(to_res.it)) {
+    to_res.post_updater.ReduceHeapUsage();
```
BagritsevichStepan, line 906: I remember that in the `RENAME` command we call `OpRen` only when there's a single shard involved. Just to confirm — does the regular multi-shard `RENAME` implementation work always correctly?

romange: yes, it does but I did not check why. the test covers multi-shard and single shard options

BagritsevichStepan: Yes, I saw the tests, but there still might be an edge case that they don’t cover. But if you're sure — then okay.

src/server/string_family.cc
```diff
@@ -851,18 +851,19 @@ OpStatus SetCmd::Set(const SetParams& params, string_view key, string_view value
     if (auto status = CachePrevIfNeeded(params, op_res->it); status != OpStatus::OK)
       return status;
 
-    return SetExisting(params, op_res->it, op_res->exp_it, key, value);
+    return SetExisting(params, value, &(*op_res));
```
BagritsevichStepan, line 854: why not std::move?

romange: why it is better?

romange: `DbSlice::ItAndUpdater` size is 240 bytes. std::move will at the end copy 240 bytes from one memory region to another.

BagritsevichStepan: Because for example we have `post_updater` in op_res. After doing `std::move` into this method, you can no longer call `it_upd->post_updater.ReduceHeapUsage()` on that iterator a second time. This change just makes the logic a bit safer.

BagritsevichStepan: > `DbSlice::ItAndUpdater` size is 240 bytes. std::move will at the end copy 240 bytes from one memory region to another.

That makes sense.

Issue #4984, opened 2025-04-23
Benchmarking different indexing approaches


vyavdoshenko:
Consider the way to test and benchmark different implementations for prefix, suffix, and infix search.

Related issues:
- Support prefix search: https://github.com/dragonflydb/dragonfly/issues/4981
- Support suffix search: https://github.com/dragonflydb/dragonfly/issues/4982
- Support infix search: https://github.com/dragonflydb/dragonfly/issues/4983

PR #5319, opened 2025-06-17
test(search): add unified benchmarks for prefix/suffix/infix search

vyavdoshenko:
Fixes: https://github.com/dragonflydb/dragonfly/issues/4984

Adds parameterized benchmarks for all search types (prefix, suffix, infix) to establish a performance baseline before implementing suffix/infix search.


src/core/search/search_test.cc
```diff
@@ -907,6 +907,123 @@ TEST_F(SearchTest, NotImplementedSearchTypes) {
...
+  }
+
+  // Generate test data based on search type
+  for (size_t i = 0; i < num_docs; i++) {
+    std::string content;
+    if (i < num_docs / 2) {
+      // Half documents have the pattern in appropriate position
+      switch (search_type) {
+        case SearchType::PREFIX:
+          content = pattern + std::to_string(i) + " some additional text";
```
BagritsevichStepan: content is very small. Let's generate more words (but it should not change across the runs). Just traverse small and big numbers in ascii order starting from different symbols, for example with offsets for each new word and document and etc.

src/core/search/search_test.cc
```diff
@@ -907,6 +907,123 @@ TEST_F(SearchTest, NotImplementedSearchTypes) {
...
+  auto schema = MakeSimpleSchema({{"title", SchemaField::TEXT}});
+  FieldIndices indices{schema, kEmptyOptions, PMR_NS::get_default_resource(), nullptr};
+
+  // Generate pattern based on search type
+  std::string pattern;
+  std::string search_type_name;
+
+  switch (search_type) {
+    case SearchType::PREFIX:
+      pattern = std::string(pattern_len, 'p');
```
BagritsevichStepan: Note: some of the trees use word compression, meaning that long sequences of repeating or similar symbols may be compressed. As a result, patterns consisting of uniform characters may behave differently than those with varied symbols. In real-world usage, patterns would typically include diverse symbols, but for the purpose of this test case, we can keep this variant too.

However, I suggest adding another test case. You can create a common function for both tests that either takes a pre-generated pattern or a callback that generates it dynamically.

For the new test: it depends on how you will generate the content of the indexes, but as I recommended generate content using sequential characters (in ascii order). Like this:
```
for (char c = 'a' + some_offset; c <= 'z'; ++c) {
   new_word += c;
}
content += new_word
```

This way, each word is unique, but still consists of characters in ASCII order. You can then construct patterns using the same logic:
```
std::string pattern;
for (char c = 'a' + offset; c < 'a' + offset + pattern_length; ++c) {
    pattern += c;
}
```
This approach ensures that the test covers realistic symbol variation, while also guaranteeing that the generated pattern has many occurrences within the document contents.

src/core/search/search_test.cc
```diff
@@ -907,6 +907,123 @@ TEST_F(SearchTest, NotImplementedSearchTypes) {
...
+      // Half documents have the pattern in appropriate position
+      switch (search_type) {
+        case SearchType::PREFIX:
+          content = pattern + std::to_string(i) + " some additional text";
+          break;
+        case SearchType::SUFFIX:
+          content = "some text content " + std::to_string(i) + pattern;
+          break;
+        case SearchType::INFIX:
+          content = "prefix " + std::to_string(i) + " " + pattern + " suffix text";
```
BagritsevichStepan: Also note that the pattern is inserted as a separate word here — it is not actually a infix of another word. Then, during the search, you query for the entire pattern, so the result is always just that single word.

What you're really benchmarking in this case is a single full-word search.

To make the benchmark more meaningful, you could either:
 - generate words differently, as I previously suggested, or
- if you want to keep the current test too you need to run search queries using substrings of the generated pattern.

That way, the search would not match the whole word, and you'd actually test infix search behavior.

BagritsevichStepan: **Upd.:** just remove spaces here: `std::to_string(i)+pattern+"suffix" + ...`
For PREFIX and SUFFIX you don't have spaces so it should work fine

src/core/search/search_test.cc
```diff
@@ -907,6 +907,123 @@ TEST_F(SearchTest, NotImplementedSearchTypes) {
...
+}
+
+// Benchmark to compare all search types with the same data
+BENCHMARK(BM_SearchByType)
+    ->Args({100, 3, static_cast<int>(SearchType::PREFIX)})
+    ->Args({100, 5, static_cast<int>(SearchType::PREFIX)})
+    ->Args({1000, 3, static_cast<int>(SearchType::PREFIX)})
+    ->Args({1000, 5, static_cast<int>(SearchType::PREFIX)})
+    ->Args({10000, 3, static_cast<int>(SearchType::PREFIX)})
+    ->Args({10000, 5, static_cast<int>(SearchType::PREFIX)})
```
BagritsevichStepan, line 1079: Can we fit `100_000` documents?

romange: it should be justified. micro benchmark should allow tracking performance characteristics of the algorithms.
if the algorithm with 100K or 10M documents behaves linearly slower as with 10K documents , there is no reason  to add this data point.  Sometimes, when an algorithm behaves non-linearly due to various reasons we should add benchmarks on bigger scale or even use other benchmarking methods

src/core/search/search_test.cc
```diff
@@ -907,6 +907,189 @@ TEST_F(SearchTest, NotImplementedSearchTypes) {
...
+        c = 'a' + (c - 'z' - 1);
+      pattern += c;
+    }
+    return pattern;
+  }
+  return "";
+}
+
+// Template function for benchmark implementations
+template <bool UseDiversePattern> static void BM_SearchByTypeImpl(benchmark::State& state) {
```
romange: not a good use for templates. Please move it to be a runtime argument.
In general, we do not write templatized code unless there is a good reason to use them.

src/core/search/search_test.cc
```diff
@@ -907,6 +907,189 @@ TEST_F(SearchTest, NotImplementedSearchTypes) {
...
+      if (c > 'z')
+        c = 'a' + (c - 'z' - 1);
+      pattern += c;
+    }
+    return pattern;
+  }
+  return "";
+}
+
+// Template function for benchmark implementations
```
romange: The comment is stale

romange (dismissed):

src/core/search/search_test.cc
vyavdoshenko: fixed

src/core/search/search_test.cc
```diff
@@ -907,6 +907,188 @@ TEST_F(SearchTest, NotImplementedSearchTypes) {
...
+    if (i > 0)
+      content += " ";
+    content += word;
+  }
+  return content;
+}
+
+// Helper function to generate pattern with variety
+static std::string GeneratePattern(SearchType search_type, size_t pattern_len,
+                                   bool use_uniform = true) {
```
romange: we do not use optional arguments unless there is a good reason to. and you do not even use this below.

romange (approved):

src/core/search/search_test.cc
vyavdoshenko: fixed

romange (approved):

PR #5320, opened 2025-06-17
chore(search_family): Add AddGenericDoc/RemoveGenericDoc methods

BagritsevichStepan:
related to [this comment ](https://github.com/dragonflydb/dragonfly/pull/5306#discussion_r2149618144)

### DIFF
--- a/src/server/family_utils.h
+++ b/src/server/family_utils.h
@@ -10,6 +10,9 @@
 #include <cstdint>
 
 #include "facade/facade_types.h"
+#include "server/engine_shard.h"
+#include "server/search/doc_index.h"
+#include "server/table.h"
 
 extern "C" {
 #include "redis/sds.h"
@@ -21,24 +24,7 @@ typedef struct streamCG streamCG;
 namespace dfly {
 
 template <typename DenseSet>
-static std::vector<long> ExpireElements(DenseSet* owner, const facade::CmdArgList values,
-                                        uint32_t ttl_sec) {
-  std::vector<long> res;
-  res.reserve(values.size());
-
-  for (size_t i = 0; i < values.size(); i++) {
-    std::string_view field = facade::ToSV(values[i]);
-    auto it = owner->Find(field);
-    if (it != owner->end()) {
-      it.SetExpiryTime(ttl_sec);
-      res.emplace_back(ttl_sec == 0 ? 0 : 1);
-    } else {
-      res.emplace_back(-2);
-    }
-  }
-
-  return res;
-}
+std::vector<long> ExpireElements(DenseSet* owner, facade::CmdArgList values, uint32_t ttl_sec);
 
 // Copy str to thread local sds instance. Valid until next WrapSds call on thread
 sds WrapSds(std::string_view str);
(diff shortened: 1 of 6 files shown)


src/server/generic_family.cc
```diff
@@ -885,12 +885,12 @@ OpResult<void> OpRen(const OpArgs& op_args, string_view from_key, string_view to
...
       return OpStatus::KEY_EXISTS;
 
-    op_args.shard->search_indices()->RemoveDoc(to_key, op_args.db_cntx, to_res.it->second);
+    op_args.shard->search_indices()->RemoveGenericDoc(to_key, op_args.db_cntx, to_res.it->second);
     is_prior_list = (to_res.it->second.ObjType() == OBJ_LIST);
   }
 
   // Delete the "from" document from the search index before deleting from the database
-  op_args.shard->search_indices()->RemoveDoc(from_key, op_args.db_cntx, from_res.it->second);
+  op_args.shard->search_indices()->RemoveGenericDoc(from_key, op_args.db_cntx, from_res.it->second);
```
romange: I disagree with this object centric approach. I prefer that `AddGenericDoc` will be a free utility function in family_utils instead of being a method of `search_indices()`. 

The reason for this - I do not want that non-search operations to touch `search_indices` object at all if not needed. in most cases `RemoveGenericDoc` should be a no-op because `res->it->second` is not json or hset and it should not touch `search_indices` at all

BagritsevichStepan: Fixed

src/server/family_utils.h
```diff
@@ -21,24 +24,7 @@ typedef struct streamCG streamCG;
 namespace dfly {
 
 template <typename DenseSet>
-static std::vector<long> ExpireElements(DenseSet* owner, const facade::CmdArgList values,
```
BagritsevichStepan, line 24: This was just moved

src/server/json_family.cc
```diff
@@ -55,15 +55,22 @@ using CI = CommandId;
...
 /* Helper class which must be initialized before any mutate operations on json.
   It will track the memory usage of the json object and update the size in the CompactObj.
   It also contains indexes updates, post update operations on the iterator. */
 class JsonAutoUpdater {
  public:
   JsonAutoUpdater(const OpArgs& op_args, string_view key, DbSlice::ItAndUpdater it,
-                  bool update_on_delete = false)
-      : op_args_(op_args), key_(key), it_(std::move(it)), update_on_delete_(update_on_delete) {
-    op_args.shard->search_indices()->RemoveDoc(key, op_args.db_cntx, it.it->second);
+                  JsonAutoUpdaterOptions options = {})
```
romange, line 69: can you explain `is_new_key` change? seems like a bug that you fixed?

BagritsevichStepan: There was no bug here. The issue is that a `DCHECK` was added to the `RemoveDoc` and `AddDoc` methods to ensure that the object type is either JSON or HSET. But a new problem appeared: in the case of a `JSON.SET` command for a new key, the object type is not json in the CompactObj at the beginning of the command. As a result, the `DCHECK` was failing during the calls to `RemoveDoc` and `AddDoc`. Therefore, when `it_res->is_new_key` is true, we need to apply special handling manually.

romange (approved):

src/server/json_family.cc
BagritsevichStepan: I discovered another problem in the `SetFullJson` method with this DCHECKs— indexing was disabled for it in the `AutoUpdater`, and now we're triggering it manually within `SetFullJson`.

romange: @BagritsevichStepan would you like to add a test to search_family_test ?

BagritsevichStepan: @Romange I caught this problem in regression tests. I don't know why we have` json_test.py::*` tests as regression tests — they could be moved to unit tests instead.

BagritsevichStepan: I mean, we already have test for this. But I do not understand why it is not in unit tests

romange: We use regression tests here to see that Dragonfly works with redis-py client - end to end (more like integration tests) due to additional complexity of the FT.SEARCH command.

romange: I still suggest to add a unit test that covers this particular issue.

romange (approved):

src/server/json_family.cc
BagritsevichStepan: We don't have FT.SEARCH, or any other search commands there. But ok, those tests are for redis-py client. Do you want me to make a copy of this test json_test.py::test_reset_key_to_string?

BagritsevichStepan: okay

romange: oh, sorry - we did it because of the differences in legacy path syntax and the new one - we learned at some point bugs in the legacy path that you fixed via these python tests.

BagritsevichStepan: Done - #5324

PR #5321, opened 2025-06-17
fix(json_family): Fix SetFullJson command for the STRING types

BagritsevichStepan:
I introduced recently a bug in json_family in this PR #5266 . I missed something and added logic that previously we didn't support :melting_face:


BagritsevichStepan:
I see the test_reset_key_to_string test, it is an expected behavior

PR #5323, opened 2025-06-18
chore: integrate Valkey fixes into stringmatchlen_impl

romange:
Also add more edge-case tests taken from Valkey.

### DIFF
--- a/src/core/dfly_core_test.cc
+++ b/src/core/dfly_core_test.cc
@@ -125,6 +125,7 @@ class StringMatchTest : public ::testing::Test {
 TEST_F(StringMatchTest, Glob2Regex) {
   EXPECT_EQ(GlobMatcher::Glob2Regex(""), "");
   EXPECT_EQ(GlobMatcher::Glob2Regex("*"), ".*");
+  EXPECT_EQ(GlobMatcher::Glob2Regex("\\*"), "\\*");
   EXPECT_EQ(GlobMatcher::Glob2Regex("\\?"), "\\?");
   EXPECT_EQ(GlobMatcher::Glob2Regex("[abc]"), "[abc]");
   EXPECT_EQ(GlobMatcher::Glob2Regex("[^abc]"), "[^abc]");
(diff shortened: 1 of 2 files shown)


src/core/dfly_core_test.cc
```diff
@@ -181,12 +182,113 @@ TEST_F(StringMatchTest, Basic) {
...
+  TEST_STRINGMATCH("A", "a", 0, 1);
+  TEST_STRINGMATCH("\\a", "a", 1, 1);
+  TEST_STRINGMATCH("\\a", "A", 0, 1);
+  TEST_STRINGMATCH("\\A", "A", 1, 1);
+  TEST_STRINGMATCH("\\A", "a", 0, 1);
+  TEST_STRINGMATCH("[\\a]", "a", 1, 1);
+
+  // TODO: to fix this: TEST_STRINGMATCH("[\\a]", "A", 0, 1);
+  TEST_STRINGMATCH("[\\A]", "A", 1, 1);
+  // TODO: to fix this: TEST_STRINGMATCH("[\\A]", "a", 0, 1);
```
kostasrim, line 211: So this is broken for us and valkey ?

src/core/glob_matcher.cc
```diff
@@ -25,14 +29,13 @@ static int stringmatchlen_impl(const char* pattern, int patternLen, const char*
           return 1; /* match */
         while (stringLen) {
           if (stringmatchlen_impl(pattern + 1, patternLen - 1, string, stringLen, nocase,
-                                  skipLongerMatches))
+                                  skipLongerMatches, nesting + 1))
```
kostasrim, line 32: oh I see the algorithm is recursive and it can blow out the stack (especially for us because fiber stack is small). +1 on this

I wonder if recursion was a problem for Valkey why didn't they rewrite the algorithm to use `iterations` instead of `recursions` :shrug:

kostasrim (approved):
lgtm -- let's see if everything passes

Issue #2313, opened 2023-12-18
Search: Reduce memory overhead by 25% by replacing of std::pmr::string


dranikpg:
Create lightweight immutable string type that can only be deleted externally, hence it doesn't have to store it's allocator pointer

PS: 25 when all strings are inlined, obviously it's less for large datasets

azuredream:
Hi, this enhancement sounds promising. I can help if you need.
Could you share more about the design? 
I didn't find `std::pmr::string` in ./src. Which files would be impacted?

dranikpg:
Hi! It's used in src/core/search/sort_index

Please note that task is not that simple because it requires some trickery to make it more or less safe and container compatible 🙂 If you want to take it up, please make sure to first dive into the context on how it's used

romange:
@dranikpg  is it still relevant?

dranikpg:
It is, because we still use pmr string

romange:
Where, I can not find it?

dranikpg:
https://github.com/dragonflydb/dragonfly/blob/978961ff793bf9da987a2fc9fc2464172e9d016b/src/core/search/sort_indices.h#L64-L66

romange:
@abhijat  this is another good candidate for the new allocator.

PR #5327, opened 2025-06-19
chore: suffix search

dranikpg:
Implements suffix and infix search in core module

### DIFF
--- a/src/core/search/indices.cc
+++ b/src/core/search/indices.cc
@@ -22,8 +22,6 @@
 #include <algorithm>
 #include <cctype>
 
-#include "base/logging.h"
-
 namespace dfly::search {
 
 using namespace std;
@@ -72,6 +70,16 @@ absl::flat_hash_set<string> NormalizeTags(string_view taglist, bool case_sensiti
     tags.insert(std::move(tmp));
   }
   return tags;
+}
+
+// Iterate over all suffixes of all words
+void IterateAllSuffixes(const absl::flat_hash_set<string>& words,
+                        absl::FunctionRef<void(std::string_view)> cb) {
+  for (string_view word : words) {
+    for (size_t offs = 0; offs < word.length(); offs++) {
+      cb(word.substr(offs));
+    }
+  }
 }
 
 };  // namespace
@@ -134,8 +142,11 @@ vector<DocId> NumericIndex::GetAllDocsWithNonNullValues() const {
 }
 
 template <typename C>
-BaseStringIndex<C>::BaseStringIndex(PMR_NS::memory_resource* mr, bool case_sensitive)
+BaseStringIndex<C>::BaseStringIndex(PMR_NS::memory_resource* mr, bool case_sensitive,
+                                    bool with_suffix)
     : case_sensitive_{case_sensitive}, entries_{mr} {
+  if (with_suffix)
+    suffix_trie_.emplace(mr);
 }
 
 template <typename C>
@@ -156,8 +167,9 @@ const typename BaseStringIndex<C>::Container* BaseStringIndex<C>::Matching(
 }
 
 template <typename C>
-void BaseStringIndex<C>::MatchingPrefix(std::string_view prefix,
-                                        absl::FunctionRef<void(const Container*)> cb) const {
+void BaseStringIndex<C>::MatchPrefix(std::string_view prefix,
+                                     absl::FunctionRef<void(const Container*)> cb) const {
+  // TODO(vlad): Use right iterator to avoid string comparison?
   for (auto it = entries_.lower_bound(prefix);
        it != entries_.end() && (*it).first.rfind(prefix, 0) == 0; ++it) {
     cb(&(*it).second);
@@ -165,9 +177,39 @@ void BaseStringIndex<C>::MatchingPrefix(std::string_view prefix,
 }
 
 template <typename C>
-typename BaseStringIndex<C>::Container* BaseStringIndex<C>::GetOrCreate(string_view word) {
-  auto* mr = entries_.get_allocator().resource();
-  return &entries_.try_emplace(PMR_NS::string{word, mr}, mr, 1000 /* block size */).first->second;
+void BaseStringIndex<C>::MatchSuffix(std::string_view suffix,
+                                     absl::FunctionRef<void(const Container*)> cb) const {
+  // If we have a suffix trie built, we just need to fetch the relevant suffix
+  if (suffix_trie_) {
+    auto it = suffix_trie_->find(suffix);
+    cb((it != suffix_trie_->end()) ? &it->second : nullptr);
+    return;
+  }
+
+  // Otherwise, iterate over all entries and look for the suffix
+  for (const auto& entry : entries_) {
+    int32_t start = entry.first.size() - suffix.size();
+    if (start >= 0 && entry.first.substr(start) == suffix)
+      cb(&entry.second);
+  }
+}
+
+template <typename C>
+void BaseStringIndex<C>::MatchInfix(std::string_view infix,
+                                    absl::FunctionRef<void(const Container*)> cb) const {
+  // If we have a suffix trie built, we just need to match the prefix
+  if (suffix_trie_) {
+    for (auto it = suffix_trie_->lower_bound(infix);
+         it != suffix_trie_->end() && (*it).first.rfind(infix, 0) == 0; ++it)
+      cb(&(*it).second);
+    return;
+  }
+
+  // Otherwise, iterate over all entries and check if it contains the entry
+  for (const auto& entry : entries_) {
+    if (entry.first.find(infix) != string::npos)
+      cb(&entry.second);
+  }
 }
 
 template <typename C>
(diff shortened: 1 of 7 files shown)


src/core/search/indices.cc
```diff
@@ -264,6 +350,32 @@ const float* FlatVectorIndex::Get(DocId doc) const {
   return &entries_[doc * dim_];
 }
 
+std::optional<std::vector<DocId>> FlatVectorIndex::GetAllResults() const {
```
dranikpg: I moved this part from the header file

romange:
There are some refactorings and functionality changes. Can you separate them into two PRs or two commits?

src/core/search/indices.cc
BagritsevichStepan: I did it in another PR that is going to be merged. You can remove this

BagritsevichStepan: I also did some refactoring to those methods that's why better to remove it in your PR

src/core/search/indices.cc
```diff
@@ -116,10 +114,39 @@ vector<DocId> NumericIndex::Range(double l, double r) const {
...
+template <typename C>
+void Match(const RaxTreeMap<BlockList<C>>& map, std::string_view prefix,
+           absl::FunctionRef<void(const BlockList<C>*)> cb) {
+  for (auto it = map.lower_bound(prefix); it != map.end() && (*it).first.rfind(prefix, 0) == 0;
+       ++it) {
+    cb(&(*it).second);
+  }
 }
 
+template <typename F> void Suffixes(const absl::flat_hash_set<string>& tokens, F&& f) {
```
romange: what does this function do? Please use verbs when naming functions.

src/core/search/indices.cc
```diff
@@ -133,23 +160,38 @@ const typename BaseStringIndex<C>::Container* BaseStringIndex<C>::Matching(
...
 void BaseStringIndex<C>::MatchingPrefix(std::string_view prefix,
                                         absl::FunctionRef<void(const Container*)> cb) const {
-  for (auto it = entries_.lower_bound(prefix);
-       it != entries_.end() && (*it).first.rfind(prefix, 0) == 0; ++it) {
-    cb(&(*it).second);
+  Match(entries_, prefix, cb);
+}
+
+template <typename C>
+void BaseStringIndex<C>::MatchingSuffix(std::string_view suffix,
```
romange: same. what does it do?

src/core/search/search.cc
```diff
@@ -343,7 +304,7 @@ struct BasicSearch {
     return IndexResult{};
   }
 
-  IndexResult Search(const AstPrefixNode& node, string_view active_field) {
+  template <TagType T> IndexResult Search(const AstAffixNode<T> node, string_view active_field) {
```
romange: `const AstAffixNode<T>&`

src/core/search/search.cc
```diff
@@ -260,28 +261,14 @@ struct BasicSearch {
     return out;
   }
 
-  template <typename C>
-  IndexResult CollectPrefixMatches(BaseStringIndex<C>* index, std::string_view prefix) {
+  template <typename C, typename F>
```
romange, line 272: please comment on how F looks like.

romange (commented):
lgtm, a few readability comments

src/core/search/indices.cc
```diff
@@ -133,23 +160,38 @@ const typename BaseStringIndex<C>::Container* BaseStringIndex<C>::Matching(
...
+                                        absl::FunctionRef<void(const Container*)> cb) const {
+  if (suffix_trie_)
```
BagritsevichStepan: I don’t understand why we have these two methods in BaseStringIndex. As far as I remember, suffix, infix, and prefix search are only used in TextIndex. So logically, these methods should be moved to TextIndex (we can make entries protected).

src/core/search/indices.h
```diff
@@ -86,20 +92,21 @@ template <typename C> struct BaseStringIndex : public BaseIndex {
   // Used by Add & Remove to tokenize text value
   virtual absl::flat_hash_set<std::string> Tokenize(std::string_view value) const = 0;
 
-  Container* GetOrCreate(std::string_view word);
+  static Container* GetOrCreate(search::RaxTreeMap<Container>* map, std::string_view word);
+  static void Remove(search::RaxTreeMap<Container>* map, DocId id, std::string_view word);
 
   bool case_sensitive_ = false;
   search::RaxTreeMap<Container> entries_;
+  std::optional<search::RaxTreeMap<Container>> suffix_trie_;
```
BagritsevichStepan, line 87: Again, this should be moved to TextIndex, and the use of optional should be removed.

src/core/search/indices.cc
```diff
@@ -164,7 +206,11 @@ bool BaseStringIndex<C>::Add(DocId id, const DocumentAccessor& doc, string_view
     tokens.merge(Tokenize(str));
 
   for (string_view token : tokens)
-    GetOrCreate(token)->Insert(id);
+    GetOrCreate(&entries_, token)->Insert(id);
+
+  if (suffix_trie_)
```
BagritsevichStepan, line 229: For this, we can add a virtual method like InsertTokenAdditionalDataIfNeeded (or similar) and call it here. The base implementation would be empty, and it can be overridden in TextIndex.

src/core/search/indices.cc
dranikpg: For now, yes. I just didn't add support for tags yet. RS supports star queries for tag values as well

src/core/search/indices.cc
dranikpg: Please look for comments always in the header. You can hover over the function name in vscode and it'll show you the header comment

src/core/search/search.cc
```diff
@@ -355,21 +316,42 @@ struct BasicSearch {
...
     return UnifyResults(GetSubResults(indices, mapping), LogicOp::OR);
   }
 
-  IndexResult Search(const AstSuffixNode& node, string_view active_field) {
-    // TODO: Implement full text search for suffix
-    error_ = "Not implemented";
-    return IndexResult{};
-  }
+  // "term": access field's text index or unify results from all text indices if no field is set
+  IndexResult Search(const AstAffixNode<TagType::REGULAR> node, string_view active_field) {
```
BagritsevichStepan, line 326: Did you just move it?

dranikpg:
I'll add support for tag suffx/infix queries and fix case sensitivity in a follow up PR

src/core/search/indices.cc
```diff
@@ -133,23 +160,38 @@ const typename BaseStringIndex<C>::Container* BaseStringIndex<C>::Matching(
...
 }
 
 template <typename C>
-typename BaseStringIndex<C>::Container* BaseStringIndex<C>::GetOrCreate(string_view word) {
-  auto* mr = entries_.get_allocator().resource();
-  return &entries_.try_emplace(PMR_NS::string{word, mr}, mr, 1000 /* block size */).first->second;
+void BaseStringIndex<C>::MatchingInfix(std::string_view infix,
+                                       absl::FunctionRef<void(const Container*)> cb) const {
+  if (suffix_trie_)
+    return Match(*suffix_trie_, infix, cb);
```
BagritsevichStepan: What about case sensitivity in infix and suffix searches — I believe Redis ignores case in such cases, since it generally performs case-insensitive search.
Please confirm if that's true and if we should also ignore case here.
If so, this isn't the first time we've forgotten to apply lowercase (e.g., in synonyms), and developers always have to keep that in mind.
I suggest introducing a string wrapper in the future that automatically normalizes case to lowercase — it would also make fixing [this issue](https://github.com/dragonflydb/dragonfly/issues/2313) easier

BagritsevichStepan: Now I see your comment

src/core/search/search.cc
dranikpg: Yes, I only moved it

src/core/search/indices.cc
dranikpg: Yep, I decided it's too much for a single PR

src/core/search/indices.cc
```diff
@@ -140,16 +151,46 @@ const typename BaseStringIndex<C>::Container* BaseStringIndex<C>::Matching(
...
 template <typename C>
-typename BaseStringIndex<C>::Container* BaseStringIndex<C>::GetOrCreate(string_view word) {
-  auto* mr = entries_.get_allocator().resource();
-  return &entries_.try_emplace(PMR_NS::string{word, mr}, mr, 1000 /* block size */).first->second;
+void BaseStringIndex<C>::MatchingSuffix(std::string_view suffix,
+                                        absl::FunctionRef<void(const Container*)> cb) const {
+  // If we have a suffix trie built, we just need to fetch the relevant suffix
+  if (suffix_trie_) {
+    auto it = suffix_trie_->find(suffix);
+    cb((it != suffix_trie_->end()) ? &it->second : nullptr);
```
romange, line 185: should we return here? you wrote "otherwise" below but based on the code it can enter both parts

src/core/search/indices.h
```diff
@@ -47,18 +47,25 @@ struct NumericIndex : public BaseIndex {
...
 
   // Pointer is valid as long as index is not mutated. Nullptr if not found
   const Container* Matching(std::string_view str, bool strip_whitespace = true) const;
 
-  // Iterate over all Matching on prefix.
+  // Iterate over all nodes matching on prefix.
   void MatchingPrefix(std::string_view prefix, absl::FunctionRef<void(const Container*)> cb) const;
 
+  // Iterate over all nodes matching suffix query. Faster if suffix trie is built.
+  void MatchingSuffix(std::string_view suffix, absl::FunctionRef<void(const Container*)> cb) const;
```
romange: Consider renaming to `MatchSuffix`  here and in other variations.

src/core/search/indices.cc
dranikpg: True, missed this when refactoring

romange (dismissed):

romange:
psl resolve

romange (approved):

Issue #5092, opened 2025-05-09
Support for redis_latency_percentiles_usec Metric in DragonFlyDB


swasthikshetty10:
**Title:** Support for `redis_latency_percentiles_usec` Metric in DragonFlyDB

**Issue Summary:**

When using the `oliver006/redis_exporter` with DragonFlyDB, we're not seeing the `redis_latency_percentiles_usec` metric that is typically available when the exporter is connected to Redis. This becomes a problem for monitoring setups — especially Grafana dashboards — that rely on this metric to track latency percentiles.

**Environment:**

- DragonFlyDB version: [latest Docker image]
- Redis Exporter version: `oliver006/redis_exporter:v1.51.0`
- Prometheus version: `v2.44.0`
- Grafana version: `9.5.2`

**Steps to Reproduce:**

1. Run both Redis and DragonFlyDB in the same environment (e.g., via Docker Compose).
2. Set up `oliver006/redis_exporter` for each service with the same config.
3. Scrape metrics using Prometheus.
4. Check for the `redis_latency_percentiles_usec` metric in Prometheus.
5. It appears for Redis but not for DragonFly.

**What We Expected:**

We expected both Redis and DragonFlyDB to expose this metric so we could reuse the same dashboards and alerts across both.

**What Actually Happens:**

The metric is missing from DragonFlyDB, so panels that rely on latency percentiles (like P99, P95, P50) are empty or broken when connected to a DragonFly instance.

**Why This Matters:**

Latency percentiles like P99, P95, and P50 are key metrics for understanding system performance — especially under load. They help identify tail latency and are widely used in SLOs and operational dashboards. Not having this data makes it harder to monitor and compare performance between Redis and DragonFlyDB.

**Other Notes:**

We did confirm that other metrics like `redis_commands_processed_total` and `redis_commands_duration_seconds_count` are being exposed correctly by both Redis and DragonFlyDB via the exporter. So the issue seems specific to the latency percentile metric.

We’re aware that DragonFly provides its own Prometheus endpoint with histograms that can be used to calculate percentiles via `histogram_quantile()`, but having `redis_latency_percentiles_usec` directly available would be a big help for compatibility with existing tools and dashboards.

Would love to know if this metric could be supported in the future, or if there's a recommended workaround for this use case.

BagritsevichStepan:
Hi @swasthikshetty10 , thanks for creating the issue. I will look into this.

romange:
[...]

PR #5330, opened 2025-06-20
chore: add hdr_histogram to the project

romange:
Mostly a boilerplate PR, should not change functionality.

1. reduce number of absl::GetCurrentTimeNanos() calls by one during the command execution.
2. add hdr_histogram object to CommandId but do not yet record anything into it.

First part to address #5092

### DIFF
--- a/src/server/command_registry.h
+++ b/src/server/command_registry.h
@@ -7,6 +7,7 @@
 #include <absl/container/flat_hash_map.h>
 #include <absl/container/flat_hash_set.h>
 #include <absl/types/span.h>
+#include <hdr/hdr_histogram.h>
 
 #include <functional>
 #include <optional>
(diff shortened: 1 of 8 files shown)


src/server/command_registry.h
```diff
@@ -84,14 +85,41 @@ struct CommandContext {
...
+// TODO: move it to helio
+// Makes sure that the POD T that is passed to the constructor is reset to default state
+template <typename T> class MoveOnly {
+ public:
+  MoveOnly() = default;
+
+  MoveOnly(const MoveOnly&) = delete;
+  MoveOnly& operator=(const MoveOnly&) = delete;
+
+  explicit MoveOnly(MoveOnly&& t) noexcept : value_(std::move(t.value_)) {
```
abhijat: nit: Why is the move ctor explicit? This might prevent using move on rvalues, eg returning this type from a function.

src/server/command_registry.h
```diff
@@ -84,14 +85,41 @@ struct CommandContext {
...
+    t.value_ = T{};  // Reset the passed value to default state
```
abhijat, line 98: Should there be a static assert that `T` is nothrow default constructible?

src/server/command_registry.h
```diff
@@ -160,11 +188,13 @@ class CommandId : public facade::CommandId {
...
 
  private:
+  // The following fields must copy manually in the move constructor.
   bool implicit_acl_;
+  bool is_alias_{false};
   std::unique_ptr<CmdCallStats[]> command_stats_;
   Handler3 handler_;
   ArgValidator validator_;
-  bool is_alias_{false};
+  MoveOnly<struct hdr_histogram*> latency_histogram_;  // Histogram for command latency in usec
```
abhijat: Nit: the `struct` type specifier is redundant here and elsewhere with `hdr_histogram`

abhijat (approved):
minor nits

src/server/command_registry.h
romange: I think the fact that it compiles is equivalent to asserting this.

PR #5335, opened 2025-06-21
test: make random determenistic in test_multidim_knn python test

vyavdoshenko:
It should fix regression tests.
https://github.com/dragonflydb/dragonfly/actions/runs/15792845507/job/44521033546

### DIFF
--- a/tests/dragonfly/search_test.py
+++ b/tests/dragonfly/search_test.py
@@ -315,11 +315,12 @@ async def test_multidim_knn(async_client: aioredis.Redis, index_type, algo_type)
             pipe.json().set(f"k{i}", "$", {"pos": point.tolist()})
     await pipe.execute()
 
-    # Run 10 random queries with fixed seed for deterministic results
-    random.seed(42)
+    # Run 10 random queries
     for _ in range(10):
         center = rand_point()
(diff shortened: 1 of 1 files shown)


tests/dragonfly/search_test.py
```diff
@@ -312,7 +315,8 @@ def rand_point():
             pipe.json().set(f"k{i}", "$", {"pos": point.tolist()})
     await pipe.execute()
 
-    # Run 10 random queries
+    # Run 10 random queries with fixed seed for deterministic results
+    random.seed(42)
```
romange: why do we need determenistic seed here as well?

vyavdoshenko: We set both seeds because np.random and random are separate modules with independent states. This ensures all randomness in the test is deterministic from the start.

romange: but do we use `random` anywhere?

romange: ah, I see. I find it weird that we use two different random generators in the same test.  can we replace `random.randint` with `np.randon` to avoid this complexity?

vyavdoshenko: The numpy random module is used in both places now

romange (approved):

PR #5340, opened 2025-06-22
chore: improve test_full_sync test

romange:
Following #5336


dranikpg:
Btw we could've added this to wait_for_replicas

romange:
what do you mean?

PR #5341, opened 2025-06-22
chore(search_family): Minor performance improvements to GetAllResults methods. FIRST PR

BagritsevichStepan:
Fixes non-sortable results from GetAllResults methods and improves performance of these methods.

Changes in this PR:
1. Fix null_values in SortIndex. It didn't use polymorphic allocator before
2. Small performance improvement for `NumericIndex`: in the `GetAllResults` method we’re calling the `Range` method, which performs two `lower_bound` operations on the `entries_` tree. But this overhead is unnecessary — we just need to traverse the entire tree.
3. Small performance improvement for `SimpleValueSortIndex`: in the `GetAllResults` method we checked `!null_values.contains(id)`, but that doesn’t make sense — we don’t add null values to `values_`, and such entries also wouldn’t pass the check for an empty value

FIRST PR of perfomance improvements

### DIFF
--- a/src/core/search/base.h
+++ b/src/core/search/base.h
@@ -13,6 +13,7 @@
 #include <string_view>
 #include <vector>
 
+#include "absl/container/flat_hash_set.h"
 #include "base/pmr/memory_resource.h"
 #include "core/string_map.h"
 
@@ -80,6 +81,11 @@ struct DocumentAccessor {
   virtual std::optional<StringList> GetTags(std::string_view active_field) const = 0;
 };
 
+// Represents a set of document IDs, used for merging results of inverse indices.
+using DocsList =
+    absl::flat_hash_set<DocId, absl::DefaultHashContainerHash<DocId>,
+                        absl::DefaultHashContainerEq<DocId>, PMR_NS::polymorphic_allocator<DocId>>;
+
 // Base class for type-specific indices.
 //
 // Queries should be done directly on subclasses with their distinc
@@ -92,9 +98,8 @@ struct BaseIndex {
   virtual void Remove(DocId id, const DocumentAccessor& doc, std::string_view field) = 0;
 
   // Returns documents that have non-null values for this field (used for @field:* queries)
-  virtual std::optional<std::vector<DocId>> GetAllResults() const {
-    return std::nullopt;
-  }
+  // Result must be sorted
+  virtual std::vector<DocId> GetAllDocsWithNonNullValues() const = 0;
 };
 
 // Base class for type-specific sorting indices.
--- a/src/core/search/base.h
+++ b/src/core/search/base.h
@@ -82,9 +82,9 @@ struct DocumentAccessor {
 };
 
 // Represents a set of document IDs, used for merging results of inverse indices.
-using DocsList =
-    absl::flat_hash_set<DocId, absl::DefaultHashContainerHash<DocId>,
-                        absl::DefaultHashContainerEq<DocId>, PMR_NS::polymorphic_allocator<DocId>>;
+template <typename Allocator = std::allocator<DocId>>
+using UniqueDocsList = absl::flat_hash_set<DocId, absl::DefaultHashContainerHash<DocId>,
+                                           absl::DefaultHashContainerEq<DocId>, Allocator>;
 
 // Base class for type-specific indices.
 //
--- a/src/core/search/indices.cc
+++ b/src/core/search/indices.cc
@@ -115,6 +115,24 @@ vector<DocId> NumericIndex::Range(double l, double r) const {
   return out;
 }
 
+vector<DocId> NumericIndex::GetAllDocsWithNonNullValues() const {
+  DocsList unique_docs{entries_.get_allocator().resource()};
+  std::vector<DocId> result;
+
+  unique_docs.reserve(entries_.size());
+  result.reserve(entries_.size());
+
+  for (const auto& [_, doc_id] : entries_) {
+    const auto [__, is_new] = unique_docs.insert(doc_id);
+    if (is_new) {
+      result.push_back(doc_id);
+    }
+  }
+
+  std::sort(result.begin(), result.end());
+  return result;
+}
+
 template <typename C>
 BaseStringIndex<C>::BaseStringIndex(PMR_NS::memory_resource* mr, bool case_sensitive)
     : case_sensitive_{case_sensitive}, entries_{mr} {
@@ -196,6 +214,26 @@ template <typename C> vector<string> BaseStringIndex<C>::GetTerms() const {
   return res;
 }
 
+template <typename C> vector<DocId> BaseStringIndex<C>::GetAllDocsWithNonNullValues() const {
+  DocsList unique_docs{entries_.get_allocator().resource()};
+  std::vector<DocId> result;
+
+  unique_docs.reserve(entries_.size());
+  result.reserve(entries_.size());
+
+  for (const auto& [_, container] : entries_) {
+    for (const auto& doc_id : container) {
+      auto [_, is_new] = unique_docs.insert(doc_id);
+      if (is_new) {
+        result.push_back(doc_id);
+      }
+    }
+  }
+
+  std::sort(result.begin(), result.end());
+  return result;
+}
+
 template struct BaseStringIndex<CompressedSortedSet>;
 template struct BaseStringIndex<SortedVector>;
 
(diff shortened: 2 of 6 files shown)


src/core/search/sort_indices.h
```diff
@@ -65,7 +51,7 @@ template <typename T> struct SimpleValueSortIndex : public BaseSortIndex {
 
  private:
   PMR_NS::vector<T> values_;
-  absl::flat_hash_set<DocId> null_values_;
```
BagritsevichStepan, line 68: We weren't using a thread-local allocator here before.

romange:
There are no benchmarks in this PR, what exactly does it improve?

src/core/search/sort_indices.cc
```diff
@@ -90,6 +91,28 @@ void SimpleValueSortIndex<T>::Remove(DocId id, const DocumentAccessor& doc,
...
+      if (is_new) {
+        result.push_back(id);
+      }
+    }
+  }
+
+  std::sort(result.begin(), result.end());
+  return result;
+}
+
```
vyavdoshenko: I see at least 3 almost the same functions. Can you generalize them to avoid copy/paste?

romange: the loop (102-110) is type dependent so it is hard to remove code duplication it without loosing performance.  
The only thing we could do is to keep the suffix and prefix parts once and call the internal function that performs the loop:
```cc
  DocsList unique_docs{values_.get_allocator().resource()};
  std::vector<DocId> result;

  unique_docs.reserve(values_.size());
  result.reserve(values_.size());
  
  FillResultsInternal(&unique_docs, &result);  // virtual override
  
  std::sort(result.begin(), result.end());
  return result;
```

but it's not substantial

BagritsevichStepan: No, this should be possible to implement using iterators. All indexes should provide an iterator that allows iterating over all not null doc IDs. In this case, the != empty_value check would also be handled within the iterator.

romange: but then - will it be as efficient as this loop? Also the amount of code you need to write to implement generic iterators will be bigger than the code duplication here.

BagritsevichStepan: Yes, that’s why I didn’t implement it with iterators initially :smile:
I had already considered that this could be generalized.
I think it will have the same efficiency.

src/core/search/sort_indices.h
```diff
@@ -42,21 +42,7 @@ template <typename T> struct SimpleValueSortIndex : public BaseSortIndex {
...
   // Override GetAllResults to return all documents with non-null values
-  std::optional<std::vector<DocId>> GetAllResults() const override {
-    std::vector<DocId> result;
-
-    for (DocId id = 0; id < values_.size(); ++id) {
-      // Check if id is not present in null_values_
-      // Also need to handle deleted elements - in them T should be empty
-      // Different types of T have different "empty" values, but we can check
-      // if this value is the default for the given type
-      if (!null_values_.contains(id) && !(values_[id] == T{})) {
```
BagritsevichStepan, line 53: The reason I didn’t add benchmarks here is that the performance improvement isn’t significant. This change is more about logical correctness. For example, here we used to check !null_values.contains(id), but that doesn’t make sense — we don’t add null values to values_, and such entries also wouldn’t pass the check for an empty value

src/core/search/indices.h
```diff
@@ -35,9 +35,7 @@ struct NumericIndex : public BaseIndex {
 
   std::vector<DocId> Range(double l, double r) const;
 
-  std::optional<std::vector<DocId>> GetAllResults() const override {
-    return Range(-std::numeric_limits<double>::infinity(), std::numeric_limits<double>::infinity());
```
BagritsevichStepan, line 39: Here, for example, we’re calling the Range method, which performs two lower_bound operations on the entries_ tree. But this overhead is unnecessary — we just need to traverse the entire tree.

src/core/search/indices.cc
```diff
@@ -115,6 +115,24 @@ vector<DocId> NumericIndex::Range(double l, double r) const {
...
+vector<DocId> NumericIndex::GetAllDocsWithNonNullValues() const {
+  DocsList unique_docs{entries_.get_allocator().resource()};
+  std::vector<DocId> result;
+
+  unique_docs.reserve(entries_.size());
+  result.reserve(entries_.size());
+
+  for (const auto& [_, doc_id] : entries_) {
+    const auto [__, is_new] = unique_docs.insert(doc_id);
+    if (is_new) {
```
BagritsevichStepan, line 127: But the reason we might want to add benchmarks here is to compare two possible approaches:
1. The first approach is to insert all `doc_ids` into `unique_docs` during the first loop and then move all items into the result vector — which adds an extra iteration over all unique IDs.
2. The second approach, which is implemented here, checks whether a doc_id is new and immediately adds it to the result vector in the same loop, avoiding additional traversal.
I chose this approach because it should perform better when we mostly have unique documents and duplicates happen rarely — which is a more realistic scenario in practice.


But I don't think there will be a significant performance difference between the two solutions here.

src/core/search/sort_indices.h
BagritsevichStepan: @romange

src/core/search/base.h
```diff
@@ -80,6 +81,11 @@ struct DocumentAccessor {
   virtual std::optional<StringList> GetTags(std::string_view active_field) const = 0;
 };
 
+// Represents a set of document IDs, used for merging results of inverse indices.
+using DocsList =
+    absl::flat_hash_set<DocId, absl::DefaultHashContainerHash<DocId>,
+                        absl::DefaultHashContainerEq<DocId>, PMR_NS::polymorphic_allocator<DocId>>;
+
```
dranikpg: No need to use a different memory resource - we only do it for persistent structures to keep track of used memory, not needed for temporary structures

src/core/search/sort_indices.h
romange: so mentioning specific functionality changes in the PR description is extremely beneficial. 
I ask all the time to provide more detailed PR descriptions, I also ask to separate  refactorings with functionality changes because it's not possible to understand what has changed in such PRs.

src/core/search/base.h
```diff
@@ -92,9 +98,8 @@ struct BaseIndex {
   virtual void Remove(DocId id, const DocumentAccessor& doc, std::string_view field) = 0;
 
   // Returns documents that have non-null values for this field (used for @field:* queries)
-  virtual std::optional<std::vector<DocId>> GetAllResults() const {
-    return std::nullopt;
-  }
+  // Result must be sorted
+  virtual std::vector<DocId> GetAllDocsWithNonNullValues() const = 0;
```
dranikpg, line 102: Maybe GetPresentDocs is a shorter name? 🙂

src/core/search/indices.h
romange: so why not mention this in the PR description?

dranikpg: 1. Why is optimizing `@field:*` such a priority?
2. I don't think the lower_bound operation has any real impact on the total duration

src/core/search/base.h
BagritsevichStepan: Please note [my comment here](https://github.com/dragonflydb/dragonfly/pull/5341#discussion_r2160263789)

src/core/search/indices.cc
romange: i agree it's really not the focus

src/core/search/indices.h
BagritsevichStepan: It’s not that important — I marked it as the first PR because I found these issues while working on another PR that’s coming soon.
Also, I fixed a bug here related to null values in `SortIndexes`, which weren’t using a polymorphic allocator.

src/core/search/base.h
BagritsevichStepan: But I aggree that for tmp sets I do not need to pass memory_resourse

src/core/search/base.h
BagritsevichStepan: I just wanted to use a name that clearly reflects what’s actually happening here.
`GetAllResults` isn’t very descriptive, and `GetPresentDocs` is also unclear — it’s unclear what `present` means. Null values are also valid values, and such documents are added and stored, so they are technically “present” too.
I want the name to clearly indicate that the method filters documents with non-null values.

BagritsevichStepan: I just wanted to use a name that clearly reflects what’s actually happening here.
`GetAllResults` isn’t very descriptive, and `GetPresentDocs` is also unclear — it’s not obvious what “present” means in this context.
For instance, null values are considered valid and such documents are included, so they are technically “present” as well.
That’s why I want the method name to explicitly indicate that it filters documents with non-null values.

src/core/search/indices.h
romange: > Also, I fixed a bug here related to null values in `SortIndexes`, which weren’t using a polymorphic allocator.

Again, this should be in the PR description.

src/core/search/indices.cc
```diff
@@ -264,6 +302,36 @@ const float* FlatVectorIndex::Get(DocId doc) const {
...
+        break;
+      }
+    }
+
+    if (!is_zero_vector) {
+      result.push_back(id);
+    }
+  }
+
+  // Result is already sorted by id, no need to sort again
```
BagritsevichStepan, line 330: I missed something and added sorting and a set for uniqueness here 🫠
But that’s unnecessary, since we’re already adding values to the result in sorted order, and they’re guaranteed to be unique.
I added the comment here

src/core/search/sort_indices.cc
```diff
@@ -90,6 +91,23 @@ void SimpleValueSortIndex<T>::Remove(DocId id, const DocumentAccessor& doc,
...
+  result.reserve(values_.size());
+
+  auto empty_value = T{};
+  for (DocId id = 0; id < values_.size(); ++id) {
+    if (values_[id] != empty_value) {
+      result.push_back(id);
+    }
+  }
+
+  // Result is already sorted by DocId
```
BagritsevichStepan, line 106: Same here

BagritsevichStepan:
Pr description is updated

PR #5342, opened 2025-06-22
chore: Refactor CompactObject encoding

dranikpg:
Issues:

1. Decoding tiered values used a proxy object (because the encoding api is tightly embedded into the compact object code)
```c++
PrimeValue decoder;
decoder.ImportExternal(value); // we just save the encoding

string raw_blob = TieredResult();
decoder.Materialize(raw_blob);       // COPIES raw_blob into itself
string result = decoder.GetString(); // COPIES again to return a decoded string
```

This approach results in redundant copies and is somewhat weird in general.

2. The CompactObject code is as readable as egyptian hieroglyphs, this PR simplifies it to some extent

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -844,7 +844,7 @@ uint64_t CompactObj::HashCode() const {
 
   if (IsInline()) {
     char buf[kInlineLen * 3];  // should suffice for most huffman decodings.
-    size_t decoded_len = GetStrEncoding().Decode(buf, string_view{u_.inline_str, taglen_});
+    size_t decoded_len = GetStrEncoding().Decode(string_view{u_.inline_str, taglen_}, buf);
     return XXH3_64bits_withSeed(buf, decoded_len, kHashSeed);
   }
 
@@ -1102,7 +1102,7 @@ void CompactObj::GetString(char* dest) const {
   CHECK(!IsExternal());
 
   if (IsInline()) {
-    GetStrEncoding().Decode(dest, {u_.inline_str, taglen_});
+    GetStrEncoding().Decode({u_.inline_str, taglen_}, dest);
     return;
   }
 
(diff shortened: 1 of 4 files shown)


src/core/compact_object.cc
```diff
@@ -786,9 +786,7 @@ size_t CompactObj::Size() const {
     switch (taglen_) {
       case SMALL_TAG:
         raw_size = u_.small_str.size();
-        if (mask_bits_.encoding == HUFFMAN_ENC) {
-          return DecodedLen(raw_size, u_.small_str.first_byte());
-        }
+        first_byte = u_.small_str.first_byte();
```
dranikpg, line 789: Clearly a "Hmmm... Wrong result. Never mind, I'll just insert an if here" 😆

src/core/compact_object.h
```diff
@@ -130,14 +130,30 @@ class CompactObj {
...
     ASCII1_ENC = 1,
     ASCII2_ENC = 2,
     HUFFMAN_ENC = 3,  // TBD
   };
 
  public:
+  // Utility class for working with different string encodings (ascii, huffman, etc)
+  struct StrEncoding {
+    size_t DecodedSize(std::string_view blob) const;         // Size of decoded blob
+    size_t Decode(char* dest, std::string_view blob) const;  // Decode into dest, return size
```
romange: pls move `dest` to be last

src/core/compact_object_test.cc
```diff
@@ -593,34 +595,19 @@ TEST_F(CompactObjectTest, DefragSet) {
   ASSERT_FALSE(cobj_.DefragIfNeeded(0.8));
 }
 
-TEST_F(CompactObjectTest, RawInterface) {
-  string str(50, 'a'), tmp, owned;
-  cobj_.SetString(str);
-  {
-    auto raw_blob = cobj_.GetRawString();
-    EXPECT_LT(raw_blob.view().size(), str.size());
+TEST_F(CompactObjectTest, StrEncoding) {
```
romange: why did you delete the RawInterface test?
to the best of my knowledge it caught real bugs.

PR #5344, opened 2025-06-22
chore: avoid deduplicating docs ids when not needed

romange:
When index consists of only singular values per document, we do not have to deduplicate doc ids during query time.

Benchmarks:

Before the PR:

```
Benchmark                  Time             CPU   Iterations
------------------------------------------------------------
BM_SearchDocIds/0       7238 ns         7238 ns       388043
BM_SearchDocIds/1      30202 ns        30200 ns        89480
BM_SearchDocIds/2      20986 ns        20984 ns       133450
```
After:

```
Benchmark                  Time             CPU   Iterations
------------------------------------------------------------
BM_SearchDocIds/0       7071 ns         7070 ns       406632
BM_SearchDocIds/1      13149 ns        13142 ns       218935
BM_SearchDocIds/2      11870 ns        11868 ns       236922
```

### DIFF
--- a/src/core/search/indices.cc
+++ b/src/core/search/indices.cc
@@ -85,6 +85,9 @@ bool NumericIndex::Add(DocId id, const DocumentAccessor& doc, string_view field)
     return false;
   }
 
+  if (numbers->size() > 1) {
+    unique_ids_ = false;
+  }
   for (auto num : numbers.value()) {
     entries_.emplace(num, id);
   }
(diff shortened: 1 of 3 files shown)


romange:
`BM_SearchDocIds/0` does not change a lot because range query is currently dominated by sort and not deduplication.

romange:
Great, then we will see how the performance of the benchmark improves even more.

PR #5350, opened 2025-06-23
feat(search): Suffix/Infix search for Tags

dranikpg:
1. Support suffix/infix queries for tag fields
2. Fix case sensitivity

### DIFF
--- a/src/server/search/search_family.cc
+++ b/src/server/search/search_family.cc
@@ -203,7 +203,7 @@ ParseResult<bool> ParseStopwords(CmdArgParser* parser, DocIndex* index) {
   return true;
 }
 
-constexpr std::array<const std::string_view, 6> kIgnoredOptions = {
+constexpr std::array<const std::string_view, 4> kIgnoredOptions = {
     "UNF"sv, "NOSTEM"sv, "INDEXMISSING"sv, "INDEXEMPTY"sv};
 constexpr std::array<const std::string_view, 3> kIgnoredOptionsWithArg = {"WEIGHT"sv, "PHONETIC"sv};
 


src/server/search/search_family.cc
```diff
@@ -200,9 +204,8 @@ ParseResult<bool> ParseStopwords(CmdArgParser* parser, DocIndex* index) {
 }
 
 constexpr std::array<const std::string_view, 6> kIgnoredOptions = {
```
BagritsevichStepan: Please change the sizes of arrays. kIgnoredOptions and kIngoredOptionsWithArg

src/core/search/search.h
```diff
@@ -41,6 +41,7 @@ struct SchemaField {
   struct TagParams {
     char separator = ',';
     bool case_sensitive = false;
+    bool with_suffixtrie = false;  // see TextParams
```
BagritsevichStepan, line 44: Maybe add a base class?

BagritsevichStepan:
The rest is good, just some small nits

src/server/search/search_family_test.cc
```diff
@@ -2810,6 +2824,8 @@ TEST_F(SearchFamilyTest, SearchReindexWriteSearchRace) {
 }
 
 TEST_F(SearchFamilyTest, IgnoredOptionsInFtCreate) {
+  GTEST_SKIP() << "The usage of ignored options is now wrong - it skips supported ones!";
```
BagritsevichStepan, line 2827: I didn't get this comment, let's remove options that we now support in FT.CREATE

BagritsevichStepan: We still have ignored options and the test was added to show that they are not causing a crash or an error (like it was before)

src/server/search/search_family.cc
dranikpg: Good catch

src/server/search/search_family_test.cc
dranikpg: It was originally introduced when schema fields had no special parameters (vector, text, tag config) to avoid failing ft.create commands for overly verbose index creation where unrelevant parameters (like NOSTEM) were specified, libraries did this often. 

Ignored options are parsed _after_ field specific parsers (like ParseTagParams), which break on the first unknown keyword. Therefore the following is possible:
1. Valid option 1 is parsed
2. Invalid option 2 causes Parse___Params to break, now ignored fields are checked
3. Valid option 3 is looked up in ignored fields and not found

They are broken and should be used by the loops in Parse___Params if at all

PR #5351, opened 2025-06-23
chore(tiering): Push error handling one more level up

dranikpg:
Add error types to OpManager and TieredStorage. **Next** PR will focus on string family and other tiering users. Pull helio


src/server/string_family.cc
```diff
@@ -180,7 +177,7 @@ OpResult<TResult<size_t>> OpStrLen(const OpArgs& op_args, string_view key) {
     fb2::Future<size_t> fut;
     op_args.shard->tiered_storage()->Read(
         op_args.db_cntx.db_index, key, co,
-        [fut](const std::string& s) mutable { fut.Resolve(s.size()); });
+        [fut](io::Result<std::string> s) mutable { fut.Resolve(s->size()); });
```
kostasrim, line 180: doesn't that trigger a deep copy of the `io::Result` which will deep copy the string ?

dranikpg: I just moved the copy one level up in the call chain. I also left a small todo in the place of invocation, as passing it by value will make it possible to move the last value

dranikpg:
@romange I pulled https://github.com/romange/helio/pull/422 with me, which fails in some replication cases

PR #5364, opened 2025-06-26
chore: add mimalloc2 third_party library

romange:
This only adds it to the buld files.

We currently use helio dependency. This PR adds mimalloc of a newer version directly to dragonfly as mimalloc2. The goal is to switch to mimalloc2 in the future.

### DIFF
--- a/src/external_libs.cmake
+++ b/src/external_libs.cmake
@@ -60,6 +60,25 @@ add_third_party(
   INSTALL_COMMAND ${DFLY_TOOLS_MAKE} install BUILD_SHARED=no PREFIX=${THIRD_PARTY_LIB_DIR}/lz4
 )
 
+set(MIMALLOC_INCLUDE_DIR ${THIRD_PARTY_LIB_DIR}/mimalloc2/include)
+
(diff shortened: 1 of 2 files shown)


src/external_libs.cmake
```diff
@@ -60,6 +60,25 @@ add_third_party(
...
+add_third_party(mimalloc2
+   # GIT_REPOSITORY https://github.com/microsoft/mimalloc/
+   # GIT_TAG v2.2.4
+   URL https://github.com/microsoft/mimalloc/archive/refs/tags/v2.2.4.tar.gz
+   PATCH_COMMAND "${MIMALLOC_PATCH_COMMAND}"
+   # Add -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_FLAGS=-O0 to debug
+   CMAKE_PASS_FLAGS "-DCMAKE_BUILD_TYPE=Release -DMI_BUILD_SHARED=OFF -DMI_BUILD_TESTS=OFF \
+                    -DMI_INSTALL_TOPLEVEL=ON -DMI_OVERRIDE=OFF -DMI_NO_PADDING=ON \ -DCMAKE_C_FLAGS=-g"
+
+  BUILD_COMMAND make -j4 mimalloc-static
```
kostasrim: Why `-j4` ?

kostasrim (approved):
lgtm -- minor comment in regards to `-j4` (why don't you let make decide how many cores are available) ?

src/external_libs.cmake
romange: i just copied from helio. is not really needed

PR #5365, opened 2025-06-26
chore: add psync metrics for replica

kostasrim:
* rename `lsn_buffer_*` metrics to `psync_*`
* add psync attempts/successes for replica


src/server/replica.h
```diff
@@ -129,6 +129,8 @@ class Replica : ProtocolClient {
 
     // sum of the offsets on all the flows.
     uint64_t repl_offset_sum;
+    size_t psync_attempts;
```
kostasrim, line 132: maybe `uint64_t` instead just to keep the notation ?

PR #5366, opened 2025-06-26
fix(replication): reduce ERROR logging during normal shutdown

vyavdoshenko:
The root cause:
1. `Replica::Stop()` sets `state_mask_ = 0` (DISABLED) and cancels context
2. Error handlers still executing see DISABLED state, but log as ERROR
3. Normal shutdown sequence triggers false error logs

Fix:
Added conditional logging macro.

- **ERROR**: Only when replication is actively enabled
- **VLOG(1)**: When intentionally stopped (DISABLED state)

### DIFF
--- a/src/server/replica.cc
+++ b/src/server/replica.cc
@@ -33,6 +33,14 @@ extern "C" {
 #include "server/rdb_load.h"
 #include "strings/human_readable.h"
 
+#define LOG_REPL_ERROR(msg)      \
+  do {                           \
+    if (state_mask_ & R_ENABLED) \
+      LOG(ERROR) << msg;         \
+    else                         \
+      VLOG(1) << msg;            \
+  } while (0)
+
 ABSL_FLAG(int, replication_acks_interval, 1000, "Interval between acks in milliseconds.");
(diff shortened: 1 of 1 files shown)


romange:
@vyavdoshenko  what about REPLTAKEOVER scenario? why do we have error logs there as part of the normal REPLTAKEOVER flow?

vyavdoshenko:
@romange 
Do you mean this error?
`E20250626 11:28:09.702525  1924 replica.cc:701] Replication error in phase STABLE_SYNC with 10.0.0.107:9999, error: Connection reset by peer, socket state: State: ESTABLISHED, Local: 10.0.0.108:40638, Remote: 10.0.0.107:9999, Inode: 15834`

It looks like a real error. I don't know the root cause of this one. I suggest delivering this fix and finding out about this error separately.

romange:
Errors is something we investigate in production. It's unexpected event that the process knows how to handle but something is not right. Socket disconnect is not an error in this context. I suggest reducing the severity of this LOG to WARNING because this is what it is - a warning.

PR #5368, opened 2025-06-26
chore: switch to mimalloc2

romange:
1. Use mimalloc 2.2.4 in Dragonfly project and stop relying on mimalloc in helio.
2. Fix a minor inefficiency in CanGrow, where we unnecessary called CanGrow twice.

Experimental runs on long-running production workloads show improvements with using mimalloc 2.2.4. Now it releases RSS more aggresively. The snapshot below demonstrates the improvement.

Before July 1, the datastore was running v1.29 and it is possible to see the huge difference between Used and RSS metrics. Since July 1, the datastore was running a version with mimalloc2 and it is possible to see how RSS memory follows `Used`. The peak on July 2, 18:00 is especially interesting as it shows a sudden spike in `Used` that caused RSS to grow significantly and then going back once `Used` went down. Before the switch the datastore did not decrease RSS after such peaks.  

 
Screenshot from 2025-07-04 17-10-12

### DIFF
--- a/src/core/dash.h
+++ b/src/core/dash.h
@@ -823,6 +823,8 @@ auto DashTable<_Key, _Value, Policy>::InsertInternal(U&& key, V&& value, Evictio
       return std::make_pair(iterator{this, target_seg_id, it.index, it.slot}, false);
     }
 
+    bool consider_throw = true;
+
     // At this point we must split the segment.
     // try garbage collect or evict.
     if constexpr (EvictionPolicy::can_evict || EvictionPolicy::can_gc) {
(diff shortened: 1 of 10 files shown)


romange:
I hope it fixes all the cases where "MEMORY DECOMMIT" helped to reduce RSS as now mimalloc does it automatically as part of its regular flow.

romange:
Specifically this commit probably helps with reclaiming unneeded RSS:
https://github.com/microsoft/mimalloc/commit/58d13f6a4fa7390e198a01dd4cc4a935724eb348

BorysTheDev (approved):

PR #5370, opened 2025-06-26
feat: add support for `BLMPOP` command

### LINKED ISSUE #3876: implement BLMPOP

EricHayter:
Adding support for the BLMPOP redis command. The command description can be found [here](https://redis.io/docs/latest/commands/blmpop/).

Resolves #3876

### DIFF
--- a/src/server/list_family_test.cc
+++ b/src/server/list_family_test.cc
@@ -5,14 +5,12 @@
 #include "server/list_family.h"
 
 #include <absl/strings/match.h>
-#include <gtest/gtest.h>
 
 #include <random>
 
 #include "base/gtest.h"
 #include "base/logging.h"
 #include "facade/facade_test.h"
-#include "gmock/gmock.h"
 #include "server/blocking_controller.h"
 #include "server/command_registry.h"
 #include "server/conn_context.h"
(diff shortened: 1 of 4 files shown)


EricHayter:
I'm working on implementing the test cases but I ran into a problem when providing non-integer or a 0 to the timeout argument. It seems like the `timeout` argument is being parsed as a the `numkeys` argument.

```
127.0.0.1:6379> BLMPOP 0 1 empty empty LEFT COUNT 10
(error) ERR at least 1 input key is needed for this command
127.0.0.1:6379> BLMPOP 0.01 1 empty empty LEFT COUNT 10
(error) ERR value is not an integer or out of range
```

It could be an issue with how the command is being registered but it's not clear what values should be used instead of the following?:

```cpp
<< CI{"BLMPOP", CO::WRITE | CO::SLOW | CO::VARIADIC_KEYS, -5, 3, 3, acl::kBLMPop}.HFUNC(
             BLMPop)
```

I did check the [redis](https://github.com/redis/redis/blob/unstable/src/commands/blmpop.json) arguments  (as suggested in `command_id.h`) but it's not clear what I would use as the last key index?

romange:
yeah, sorry for this. you probably need to hack `DetermineKeys` function in transaction.cc.
I am guessing `num_keys_index` is deduced incorrectly. The command structure in redis is not uniform and we wrote a single, not very elegant function that handles "exceptions".

src/server/list_family_test.cc
```diff
@@ -5,12 +5,14 @@
 #include "server/list_family.h"
 
 #include <absl/strings/match.h>
+#include <gtest/gtest.h>
```
romange: not needed

src/server/list_family_test.cc
```diff
@@ -5,12 +5,14 @@
...
 
 #include <random>
 
 #include "base/gtest.h"
 #include "base/logging.h"
 #include "facade/facade_test.h"
+#include "gmock/gmock.h"
```
romange: not needed

src/server/list_family_test.cc
```diff
@@ -85,6 +87,108 @@ TEST_F(ListFamilyTest, Expire) {
   EXPECT_THAT(resp, IntArg(1));
 }
 
+TEST_F(ListFamilyTest, BLMPopNonblocking) {
+  auto resp = Run({"lpush", kKey1, "1", "2", "3", "4"});
+  EXPECT_THAT(resp, IntArg(4));
+
+  resp = Run({"blmpop", "0.01", "2", kKey2, kKey1, "LEFT"});
+  EXPECT_THAT(resp, RespArray(ElementsAre(kKey1, RespArray(ElementsAre("4")))));
```
romange: you can use `RespElementsAre` instead of `RespArray(ElementsAre())` combination

src/server/transaction.cc
```diff
@@ -1590,6 +1590,8 @@ OpResult<KeyIndex> DetermineKeys(const CommandId* cid, CmdArgList args) {
     unsigned num_keys_index;
     if (absl::StartsWith(name, "EVAL"))
       num_keys_index = 1;
+    else if (name == "BLMPOP")
```
romange: just unify this with condition above: ` if (absl::StartsWith(name, "EVAL") || name == "BLMPOP")`

romange:
Thank you @EricHayter

romange:
@EricHayter  there is one thing that is missing - `CO::NO_AUTOJOURNAL`. Basically for such commands we need to send a different command for replication, so we disable the default journaling. Also `OpPop` should be called with `journal_rewrite=true`

EricHayter:
Sorry for the late response, I've been very busy lately. I'll create a `fix` MR shortly, by the end of the week hopefully.

romange:
No worries and thanks!

PR #5372, opened 2025-06-27
server_family: Expose command latency stats as part of info

### LINKED ISSUE #5092: Support for redis_latency_percentiles_usec Metric in DragonFlyDB

abhijat:
Command latency is recorded in its histogram when command is invoked. 

When `info latencystats` is run, the latency for a fixed set of percentiles (50,99,99.9 by default) is returned per command. 

Currently the percentiles are fixed, but in a later PR these should be made configurable as command line option as well as `config set`.

FIXES #5092

### DIFF
--- a/src/server/command_registry.cc
+++ b/src/server/command_registry.cc
@@ -189,6 +189,9 @@ uint64_t CommandId::Invoke(CmdArgList args, const CommandContext& cmd_cntx) cons
   ++ent.first;
   ent.second += execution_time_usec;
 
+  if (hdr_histogram* cmd_histogram = latency_histogram_; cmd_histogram != nullptr) {
+    hdr_record_value(cmd_histogram, execution_time_usec);
+  }
   return execution_time_usec;
 }
 
@@ -210,6 +213,10 @@ optional<facade::ErrorReply> CommandId::Validate(CmdArgList tail_args) const {
   if (validator_)
     return validator_(tail_args);
   return nullopt;
+}
+
+hdr_histogram* CommandId::LatencyHist() const {
+  return latency_histogram_;
 }
 
 CommandRegistry::CommandRegistry() {
@@ -319,6 +326,15 @@ std::pair<const CommandId*, ArgSlice> CommandRegistry::FindExtended(string_view 
     }
   }
   return {res, tail_args};
+}
+
+absl::flat_hash_map<std::string, hdr_histogram*> CommandRegistry::LatencyMap() const {
+  absl::flat_hash_map<std::string, hdr_histogram*> cmd_latencies;
+  cmd_latencies.reserve(cmd_map_.size());
+  for (const auto& [cmd_name, cmd] : cmd_map_) {
+    cmd_latencies.insert({absl::AsciiStrToLower(cmd_name), cmd.LatencyHist()});
+  }
+  return cmd_latencies;
 }
 
 namespace CO {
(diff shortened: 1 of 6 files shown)


abhijat:
In this PR, the latency stats for sub-commands (for example `info memory` - where memory is sub-command of info) are not returned individually, we return just one stat for the parent. 

This is because we do not have tracking information at the command registry level for sub-commands. 

It might be useful to return the latency for individual sub-commands eventually.

kostasrim:
I will review this in the morning. Witting here so I don't forget.

abhijat:
After offline discussion we will just implement the command line option to configure percentile values in a separate follow up PR, but not add the `config set` option to configure it on runtime cc @adiholden

src/server/command_registry.cc
```diff
@@ -189,6 +189,9 @@ uint64_t CommandId::Invoke(CmdArgList args, const CommandContext& cmd_cntx) cons
   ++ent.first;
   ent.second += execution_time_usec;
 
+  if (hdr_histogram* cmd_histogram = latency_histogram_; cmd_histogram != nullptr) {
```
romange: this may impact performance.  please add the `latency_tracking` flag, and set it to false by default.

abhijat: Added `track_command_latency`.

romange: I specifically asked that name because this is the name used by valkey and Redis. Check out their conf file.

abhijat: changed the name

Issue #5378, opened 2025-06-29
Memory accounting code: review


dranikpg:
I've been re-reading our memory accounting code to add better heuristics: 

- [x] `max_memory_limit` is just a global `size_t` that can be mutably changed by the config registry 🎖
- [x] `used_mem_peak` and `rss_mem_peak` are unnecessarily global
- [x] Deny OOM:
  * `GetMemoryUsage` updates every 1us, whereas global atomics are updated every ~`1/shard_count`**ms** (rss even less frequently), i.e. 99% of the calls are ineffective
  * Not sure why we use command timestamps?
* No uniformity in flag access. Some code calls `GetFlag` on hot paths, server state caches at least a handful of other flags. Once calculated parameters were left forgotten with flag updates (like `soft_budget_limit_`)
* "Modules" often don't care about centralizing memory calculations. Deny oom has 3 places where it repeats its calculation (using different sources at the same time), tiering calculates it's `offload threshold` two times - same formula - differently looking code
* No real comments to clarify the meaning or usage of important variables or calculations

romange:
@dranikpg  can we close it?

romange:
@dranikpg  can we close it?

PR #5379, opened 2025-06-29
chore(server): Some memory code updates

dranikpg:
Some of the points from #5378

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -258,13 +258,7 @@ struct {
 } glob_memory_peaks;
 
 size_t FetchRssMemory(const io::StatusData& sdata) {
-  size_t total_rss = sdata.vm_rss + sdata.hugetlb_pages;
-
-  rss_mem_current.store(total_rss, memory_order_relaxed);
-  if (total_rss > glob_memory_peaks.rss.load(memory_order_relaxed))
-    glob_memory_peaks.rss.store(total_rss, memory_order_relaxed);
-
-  return total_rss;
+  return sdata.vm_rss + sdata.hugetlb_pages;
 }
 
 using CI = CommandId;
@@ -686,6 +680,11 @@ bool ReadProcStats(io::StatusData* sdata) {
     return false;
   }
 
+  size_t total_rss = FetchRssMemory(*sdata_res);
+  rss_mem_current.store(total_rss, memory_order_relaxed);
+  if (total_rss > glob_memory_peaks.rss.load(memory_order_relaxed))
+    glob_memory_peaks.rss.store(total_rss, memory_order_relaxed);
+
   *sdata = *sdata_res;
   return true;
 }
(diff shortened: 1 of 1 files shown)


src/server/server_family.cc
```diff
@@ -673,11 +677,6 @@ bool ReadProcStats(io::StatusData* sdata) {
     return false;
   }
 
-  size_t total_rss = FetchRssMemory(sdata_res.value());
```
romange, line 673: I just recently added this here. The reason is that otherwise we showed discrepancies in info/metrics, if I remember correctly.

romange: ah I remembered. the discrepancy was between /proc/ metrics and stale rss_mem_current that caused a pushback in writes that could be avoided because /proc/ went down. This is why I wanted to sync rss_mem_current as quickly as possible if we already read `/proc`

src/server/server_family.cc
```diff
@@ -2038,6 +2025,14 @@ void ServerFamily::ClientUnPauseCmd(CmdArgList args, SinkReplyBuilder* builder)
   builder->SendOk();
 }
 
+void ServerFamily::ChangeConnectionAccept(bool accept) {
+  DCHECK_NE(accept, accepting_connections_);
+  auto h = accept ? &ListenerInterface::resume_accepting : &ListenerInterface::pause_accepting;
```
romange, line 2060: nice!

romange (commented):
lgtm besides a small comment

romange:
also please resolve your branch

src/server/server_family.cc
romange: @dranikpg have you seen my comment?

dranikpg: It's here, just above

```c++
size_t FetchRssMemory(const io::StatusData& sdata) {
  size_t total_rss = sdata.vm_rss + sdata.hugetlb_pages;

  rss_mem_current.store(total_rss, memory_order_relaxed);
  if (total_rss > glob_memory_peaks.rss.load(memory_order_relaxed))
    glob_memory_peaks.rss.store(total_rss, memory_order_relaxed);

  return total_rss;
}
```

src/server/server_family.cc
```diff
@@ -1028,51 +1039,35 @@ void ServerFamily::Shutdown() {
 }
 
 bool ServerFamily::HasPrivilegedInterface() {
-  for (auto* listener : listeners_) {
-    if (listener->IsPrivilegedInterface()) {
-      return true;
-    }
-  }
-  return false;
+  return any_of(listeners_.begin(), listeners_.end(),
```
kostasrim, line 1041: I am tempted for the last 2 year to introduce a header algorithm and a function template for each of the most popular standard algorithms:

```
template<typename C, typename F>
bool any_of(const C& c, F&& f) {
  return std::any_of(c.begin(), c.end(), std::forward<F>(f));
}

src/server/server_family.cc
kostasrim: @dranikpg yes but you removed the call to it here. I think that's what @romange is saying

romange: I see the function, I just do not see where you call it here @dranikpg

src/server/server_family.cc
dranikpg: #include <ranges> in a way 🙂

src/server/server_family.cc
dranikpg: I see I see, I missed it in one place. Moved it back to ReadProcStats

romange (approved):

PR #5381, opened 2025-06-29
feat(search_family): Introduce merging of sorted search results from shards. FIRST PR

BagritsevichStepan:


src/core/search/base.h
```diff
@@ -52,8 +54,22 @@ struct WrappedStrPtr {
...
 
-// Score produced either by KNN (float) or SORT (double / wrapped str)
-using ResultScore = std::variant<std::monostate, float, double, WrappedStrPtr>;
+// Score produced either by KNN
+using ResultScore = float;
+
+// Sort order for sorting indices and results.
+enum class SortOrder { ASC, DESC };
+
+/* Optimized comparator for ascending or descending sort.
```
romange, line 63: Can you demonstrate the performance improvement here?

romange (commented):
This PR mixes many changes but in practice the change you suggest can be demonstrated by using a much smaller PR. It does not have to work for all cases to demonstrate the value.

What this PR does not have is benchmarks, which in fact are necessary for such changes.

src/core/search/base.h
dranikpg: Especially after introducing dynamic dispatch with std::function 😆 Will be two times slower at least. I'm not sure instantiated sort is fully optimizied for the ascending parameter being immutable, but I doubt it has a serious impact

src/core/search/search.cc
```diff
@@ -416,10 +416,14 @@ struct BasicSearch {
...
 
-    size_t prefix_size = min(knn.limit, knn_distances_.size());
-    partial_sort(knn_distances_.begin(), knn_distances_.begin() + prefix_size,
-                 knn_distances_.end());
-    knn_distances_.resize(prefix_size);
+    if (knn.limit < knn_distances_.size()) {
+      // TODO: use min-heap if limit is much smaller than the number of results
+      std::partial_sort(knn_distances_.begin(), knn_distances_.begin() + knn.limit,
+                        knn_distances_.end());
+      knn_distances_.resize(knn.limit);
```
dranikpg, line 423: > If the limit is much smaller than size of the array - use k-min-heap selection instead of std::partial_sort.
This will significantly reduce allocations and improve efficiency.

Bold claim. I'm not sure at all it's more effective than partial sort

src/core/search/sort_indices.cc
```diff
@@ -47,16 +49,48 @@ template <typename T> SortableValue SimpleValueSortIndex<T>::Lookup(DocId doc) c
...
+  std::vector<SortableValue> out(size);
+  for (size_t i = 0; i < size; i++) {
+    const auto doc_id = search_result->ids[i];
+    auto& value = values_[doc_id];
+    if constexpr (!std::is_arithmetic_v<T>) {
+      out[i] = std::string{value};
+    } else {
+      out[i] = static_cast<double>(value);
+    }
+  }
```
dranikpg, line 93: 1. If there was a simple way to sort two arrays at once, you wouldn't need the indices array and could keep a separate array for scores (to avoid reallocating into an array of pairs). Requires modifying the swap
2. I understand it's temporary(?) but passing SearchAlgorithmResult* is not the best interface

src/server/search/doc_index.cc
```diff
@@ -346,60 +419,113 @@ SearchResult ShardDocIndex::Search(const OpArgs& op_args, const SearchParams& pa
...
   return SearchResult{search_results.total - expired_count, std::move(out),
                       std::move(search_results.profile)};
 }
 
+io::Result<std::vector<SerializedSearchDoc>, facade::ErrorReply> ShardDocIndex::SortSearchResult(
+    const OpArgs& op_args, const SearchField& field, size_t shard_limit,
+    search::SortOrder sort_order, search::SearchAlrgorithmResult* search_result) const {
+  auto fident = field.GetIdentifier(base_->schema, false);
+
+  if (IsSortableField(fident, base_->schema)) {
```
dranikpg, line 504: Nit: maybe let's not move logic higher than needed, doc_index knows itself if it's sortable or not, so it can do it internally, we just pass an accessor object unconditionally

src/server/search/search_family.cc
```diff
@@ -565,92 +571,183 @@ void SendSerializedDoc(const SerializedSearchDoc& doc, SinkReplyBuilder* builder
...
+    absl::Span<SearchResult> results,
+    absl::FunctionRef<void(size_t, const SerializedSearchDoc&, std::vector<const T*>*)>
+        update_cached_value) {
+  // We need to merge results from all shards by the field name.
+  std::vector<const T*> cached_values(results.size(), nullptr);
+
+  for (size_t i = 0; i < results.size(); ++i) {
+    if (!results[i].docs.empty()) {
+      const auto& first_doc = results[i].docs[0];
+      update_cached_value(i, first_doc, &cached_values);
```
dranikpg, line 588: nit: update_cached_value can really just be a `get_value(i)` 🤔

src/server/search/search_family.cc
```diff
@@ -565,92 +571,183 @@ void SendSerializedDoc(const SerializedSearchDoc& doc, SinkReplyBuilder* builder
...
+    SerializedSearchDoc* next_doc = nullptr;
+    size_t next_index = 0;
+
+    // Find the minimum value in all shards
+    for (size_t i = 0; i < results.size(); ++i) {
+      if (indexes[i] < results[i].docs.size() && (!next_doc || comparator(i, next_index))) {
+        next_doc = &results[i].docs[indexes[i]];
+        next_index = i;
+      }
+    }
```
dranikpg, line 624: this is actually where you can use a heap optimization with make_heap or something akin

BagritsevichStepan: Yes, it is temporary, note that the shards number is not big

BagritsevichStepan: Maybe later we can change this

src/core/search/sort_indices.cc
BagritsevichStepan: Yes, it's temporary. I added TODOs and also mentioned in the PR description that it will be fixed in the very next PR.

Good idea regarding the swap.

src/core/search/base.h
BagritsevichStepan: Yes, I’m aware that std::function introduces dynamic dispatch. It was initially added based on several comments and discussions from Borys in other PRs (regarding the SORTBY option in the FT.AGGREGATE command). I still added std::function here because we can always implement our own lightweight version — there are a lof of such implementations, and I remember that I implemented one myself some time ago.

For this PR we can remove this, and with adding the correct implementation add this change

src/core/search/search.cc
BagritsevichStepan: Why? Note that std::partial_sort is also using heap approach for small k (as I get from documentation). But here you need to do knn_distances_.reserve(sub_results.Size()); first and push the values to the array. If limit is small and there is a big amount of sub_results heap approach should be better. But I agree that it is not priority for the next PR

src/core/search/sort_indices.cc
BagritsevichStepan: Also, take into account that now (on the current main branch) this method is not called at all :smile:

src/core/search/base.h
dranikpg: They optimize allocations, not dispatch cost. Why add so much code for this at all if
* we can just use static dispatch with templates
* likely optimize prematurely here

src/core/search/search.cc
dranikpg: True, we can avoid allocating knn distances vector (but likely cheap compared to computing vector distance)

romange:
Let's not review this PR for now - it won't be submitted in this form as it's too big and not focused.

PR #5386, opened 2025-06-30
fix(zset_family): Return early if range is expected to be zero

### LINKED ISSUE #5383: zset bugs
## ZLEXCOUNT returns wrong result

To reproduce:
```
ZADD A 0 AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
ZLEXCOUNT A - -
```

returns 1 but should return 0.

what's interesting, for shorted items (i.e. `AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA`) , it returns 0.

mkaruza:
There are 2 ranges that can be used but they are always zero. Ranges are (+inf, anything) and (anything, -inf). Check and return early if they are used.

Fixes #5383

### DIFF
--- a/src/core/sorted_map.cc
+++ b/src/core/sorted_map.cc
@@ -688,10 +688,8 @@ size_t SortedMap::LexCount(const zlexrangespec& range) const {
   if (score_tree->Size() == 0)
     return 0;
 
-  // Ranges that will always be zero
-  if ((range.min == cminstring && range.max == cminstring) ||
-      (range.min == cmaxstring && range.max == cmaxstring) ||
-      (range.min == cmaxstring && range.max == cminstring)) {
+  // Ranges that will always be zero - (+inf, anything) or (anything, -inf)
+  if (range.min == cmaxstring || range.max == cminstring) {
     return 0;
   }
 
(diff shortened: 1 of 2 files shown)


src/core/sorted_map.cc
```diff
@@ -688,6 +688,13 @@ size_t SortedMap::LexCount(const zlexrangespec& range) const {
   if (score_tree->Size() == 0)
     return 0;
 
+  // Ranges that will always be zero
+  if ((range.min == cminstring && range.max == cminstring) ||
```
romange: why it worked inconsistently depending on the item we added?

mkaruza: Shorter objects are encoded as `OBJ_ENCODING_LISTPACK` so logic is handled with redis path while longer objects are SortedMap so we will handle this special case in this function

mkaruza: I'll update test so it test both encodings

src/core/sorted_map.cc
```diff
@@ -688,6 +688,13 @@ size_t SortedMap::LexCount(const zlexrangespec& range) const {
...
+      (range.min == cmaxstring && range.max == cmaxstring) ||
```
romange: Is not it enough that `range.min == cmaxstring`  to return 0?

romange:
Would it be more precise to say that  ` (+inf, anything)` or  `(anything, -inf)`  should return 0?

src/core/sorted_map.cc
mkaruza: There is also (min,min).

mkaruza:
> Would it be more precise to say that ` (+inf, anything)` or `(anything, -inf)` should return 0?

Yes, true. I'll update logic.

romange (approved):

PR #5389, opened 2025-06-30
fix: JSON.STRLEN behavior for non-existent keys

vyavdoshenko:
Makes JSON.STRLEN consistent with other JSON commands:
- Legacy mode (`.`): returns NIL for non-existent keys
- JSONPath v2 mode (`$`): returns "no such key" error

### DIFF
--- a/src/server/json_family_test.cc
+++ b/src/server/json_family_test.cc
@@ -414,7 +414,7 @@ TEST_F(JsonFamilyTest, StrLen) {
   EXPECT_THAT(resp, IntArg(2));
 
   resp = Run({"JSON.STRLEN", "non_existent_key", "$.c.b"});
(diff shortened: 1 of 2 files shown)


src/server/json_family_test.cc
```diff
@@ -471,6 +471,11 @@ TEST_F(JsonFamilyTest, StrLenLegacy) {
   EXPECT_THAT(resp, IntArg(1));
 }
 
+TEST_F(JsonFamilyTest, StrLenNonExistentKey) {
+  auto resp = Run({"JSON.STRLEN", "non_existent_key", "$"});
+  EXPECT_THAT(resp, ErrArg("no such key"));
+}
```
dranikpg: nit: мaybe just add it to the basic strlen test 🙂

vyavdoshenko: fixed

PR #5391, opened 2025-07-01
feat(pytest): Set and log random seed used in pytest

mkaruza:
Set and print random seed used in python testing for tests that use Seeder and DflySeederFactory classes for generating commands.

### DIFF
--- a/tests/dragonfly/conftest.py
+++ b/tests/dragonfly/conftest.py
@@ -95,7 +95,7 @@ def df_seeder_factory(request) -> DflySeederFactory:
         seed = random.randrange(sys.maxsize)
 
     random.seed(int(seed))
-    print(f"--- Random seed: {seed}, check: {random.randrange(100)} ---")
+    logging.debug(f"Random seed: {seed}, check: {random.randrange(100)}")
 
     return DflySeederFactory(request.config.getoption("--log-seeder"))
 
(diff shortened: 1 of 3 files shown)


tests/dragonfly/seeder/__init__.py
```diff
@@ -140,6 +141,7 @@ def __init__(
         types: typing.Optional[typing.List[str]] = None,
         huge_value_target=5,
         huge_value_size=100000,
+        seed=None,
```
kostasrim, line 152:
```suggestion
        self.seed = random.randrange(sys.maxsize),
```

tests/dragonfly/seeder/__init__.py
```diff
@@ -162,6 +164,14 @@ def __init__(
             for i in range(units)
         ]
 
+        if seed is None:
+            self.seed = random.randrange(sys.maxsize)
+        else:
+            self.seed = seed
```
kostasrim:
```suggestion
        if seed is not None:
            self.seed = seed
```

tests/dragonfly/conftest.py
```diff
@@ -95,7 +95,7 @@ def df_seeder_factory(request) -> DflySeederFactory:
         seed = random.randrange(sys.maxsize)
 
     random.seed(int(seed))
-    print(f"--- Random seed: {seed}, check: {random.randrange(100)} ---")
+    logging.debug(f"Random seed: {seed}, check: {random.randrange(100)}")
```
dranikpg, line 98: Ah, we actually have two seeders: one old and one new (with lua scripts). This refers to the old one

mkaruza: Yes we have, this changes "old" one to print seed via logging.

PR #5397, opened 2025-07-01
feat(search): Introduce RangeTree. DRAFT

BagritsevichStepan:
TODO: description

### DIFF
--- a/src/core/search/block_list.h
+++ b/src/core/search/block_list.h
@@ -62,29 +62,10 @@ template <typename Container /* underlying container */> class BlockList {
   // Remove element, returns true if removed, false if not found.
   bool Remove(ElementType t);
 
-  // TODO: it is not correct that we return DocId here
-  // We need extract DocId from ElementType
-  std::vector<DocId> GetAllDocIds() const;
-
-  // TODO: remove ebanle_if_t for all methods below
-  // It is temporary solution
-  // TODO: remove DocId for return type, the same as in GetAllDocIds
-  template <typename U = ElementType>
-  std::enable_if_t<std::is_same_v<U, std::pair<DocId, double>>, std::vector<DocId>> LessOrEqual(
-      double r) const;
-
-  template <typename U = ElementType>
-  std::enable_if_t<std::is_same_v<U, std::pair<DocId, double>>, std::vector<DocId>> GreaterOrEqual(
-      double l) const;
-
-  // Both l and r are inclusive.
-  template <typename U = ElementType>
-  std::enable_if_t<std::is_same_v<U, std::pair<DocId, double>>, std::vector<DocId>> Range(
-      double l, double r) const;
-
   /* Split into two blocks, left and right,
      so that both blocks have approximately the same number of elements.
      Returns median value of the split. */
+  // TODO: remove enable_if_t from here, add template return type
   template <typename U = ElementType>
   std::enable_if_t<std::is_same_v<U, std::pair<DocId, double>>, SplitResult> Split() &&;
 
@@ -206,51 +187,6 @@ extern template class BlockList<SortedVector<std::pair<DocId, double>>>;
 
 // Implementation
 /******************************************************************/
-template <typename C>
-template <typename U>
-std::enable_if_t<std::is_same_v<U, std::pair<DocId, double>>, std::vector<DocId>>
-BlockList<C>::LessOrEqual(double r) const {
-  std::vector<DocId> result;
-  for (const auto& block : blocks_) {
-    for (const auto& entry : block) {
-      if (entry.second <= r) {
-        result.push_back(entry.first);
-      }
-    }
-  }
-  return result;
-}
-
-template <typename C>
-template <typename U>
-std::enable_if_t<std::is_same_v<U, std::pair<DocId, double>>, std::vector<DocId>>
-BlockList<C>::GreaterOrEqual(double l) const {
-  std::vector<DocId> result;
-  for (const auto& block : blocks_) {
-    for (const auto& entry : block) {
-      if (entry.second >= l) {
-        result.push_back(entry.first);
-      }
-    }
-  }
-  return result;
-}
-
-template <typename C>
-template <typename U>
-std::enable_if_t<std::is_same_v<U, std::pair<DocId, double>>, std::vector<DocId>>
-BlockList<C>::Range(double l, double r) const {
-  std::vector<DocId> result;
-  for (const auto& block : blocks_) {
-    for (const auto& entry : block) {
-      if (entry.second >= l && entry.second <= r) {
-        result.push_back(entry.first);
-      }
-    }
-  }
-  return result;
-}
-
 template <typename C>
 template <typename U>
 std::enable_if_t<std::is_same_v<U, std::pair<DocId, double>>, typename BlockList<C>::SplitResult>
--- a/src/core/search/block_list.h
+++ b/src/core/search/block_list.h
@@ -6,6 +6,7 @@
 #include <cstdint>
 #include <iterator>
 #include <optional>
+#include <type_traits>
 #include <vector>
 
 #include "core/search/base.h"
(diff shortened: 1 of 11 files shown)


src/core/search/block_list.h
```diff
@@ -18,20 +19,74 @@ namespace dfly::search {
...
+
+  // TODO: it is not correct that we return DocId here
+  // We need extract DocId from ElementType
+  std::vector<DocId> GetAllDocIds() const;
+
+  // TODO: remove ebanle_if_t for all methods below
+  // It is temporary solution
+  // TODO: remove DocId for return type, the same as in GetAllDocIds
+  template <typename U = ElementType>
+  std::enable_if_t<std::is_same_v<U, std::pair<DocId, double>>, std::vector<DocId>> LessOrEqual(
```
BagritsevichStepan, line 73: I think this methods can be just moved to RangeTree. I added them a long time ago that's why it is like this. But the implementation will be the same

Only split requires non const iterator and better to have it here (it clears data of current block).

src/core/search/range_tree.h
```diff
@@ -0,0 +1,125 @@
...
+#include "core/search/block_list.h"
+
+namespace dfly::search {
+class RangeResult;
+
+class RangeTree {
+ private:
+  friend class RangeResult;
+  using RangeNumber = long long;
+  using Key = std::pair<RangeNumber, RangeNumber>;
```
dranikpg, line 20: What if all values are in the range 0...1? 🔍 I suggest just to use doubles. RS treats everything as a double, so do we

src/core/search/range_tree.h
```diff
@@ -0,0 +1,125 @@
...
+  using Entry = std::pair<DocId, double>;
+  using RangeBlock = BlockList<SortedVector<Entry>>;
```
dranikpg, line 22: Could've been just SortedVector<> for the first PR (with a smaller range block size)

src/core/search/range_tree.h
```diff
@@ -0,0 +1,125 @@
...
+  using Map = absl::btree_map<Key, RangeBlock, std::less<Key>,
+                              PMR_NS::polymorphic_allocator<std::pair<const Key, RangeBlock>>>;
+
+  // static constexpr size_t kMinRangeBlockSize = 200;
+  static constexpr size_t kMaxRangeBlockSize = 500000;
```
dranikpg, line 27: I remember we decided not to cover specifics, but we should benchmark single value fetch latency... Can be huge with 500k-1 entries in a single block

src/core/search/range_tree.h
```diff
@@ -0,0 +1,125 @@
...
+  // OR just insert it to middle_blocks_ and let leftmost_block_ and rightmost_block_ be empty
+  using BorderBlock = std::vector<DocId>;
+  using BorderBlockIterator = std::vector<DocId>::const_iterator;
+  using RangeBlockIterator = RangeTree::RangeBlock::BlockListIterator;
+
+  // Utils methods to extract DocIds from RangeBlock
+  static std::vector<DocId> LessOrEqual(const RangeTree::RangeBlock& range_block, double r);
+  static std::vector<DocId> GreaterOrEqual(const RangeTree::RangeBlock& range_block, double l);
+  static std::vector<DocId> Range(const RangeTree::RangeBlock& range_block, double l, double r);
+  static std::vector<DocId> GetAllDocIds(const RangeTree::RangeBlock& range_block);
```
dranikpg, line 69: What about `std::vector<DocId> Range(double l = -inf, double r = +inf)` and - 50 LOC 🙂

src/core/search/range_tree.h
```diff
@@ -0,0 +1,125 @@
...
+
+ public:
+  class ResultsMerger {
+   private:
+    using MiddleBlockIteratorList = absl::InlinedVector<RangeBlockIterator, 10>;
+
```
dranikpg, line 75: Why not replace all this with just:

```c++
auto compare = []()...
queue<pair<B::const_iterator, /* cur */ B::const_iterator /* end */>, ...> q;
// fill q
while (!q.empty()) {
    auto [it, end_it] = q.top();
    if (it->second >= l && it->second <= r && it->first != result.back().first)
      result.push_back(*it);
    if (++it != end_it)
      q.emplace(it, end_it);
  }
```

This task in it's basic form shouldn't take 200LOC

dranikpg: If we want to save unnecessary comparisons we'll add just a flag to the iterator, I don't see a reason for separating blocks into `left, middle, right` as fields

src/core/search/range_tree.cc
```diff
@@ -0,0 +1,346 @@
...
+    DCHECK(it != result.middle_blocks_[i]->end()) << "Middle block iterator should not be empty";
+    middle_blocks_iters_.push_back(it);
+    heap_.push(i);
+  }
+}
+
+DocId RangeResult::ResultsMerger::GetMin(DocId min_value) {
+  DCHECK(HasNext());
+
+  auto skip_until = [&](const BorderBlock& view, BorderBlockIterator* it) -> DocId {
```
dranikpg, line 177: std::find_first

src/core/search/range_tree.h
dranikpg: Especially `GetMin(x)` shouldn't exist, we always collect the values sequentially advancing one by one, assuming `x` is always `result.back() + 1`

BagritsevichStepan:
We are not going to merge this PR. I'm splitting it to several steps - that's why it pushed as a draft. But I will fix you comments for next PR

src/core/search/range_tree.h
BagritsevichStepan: This is done in this PR for the future use. RangeResult can be easily added to the IndexResult. GetMin with size_t min_index will help a lot for merging std::vector<DocId> (or whatever in other IndexResult) with RangeResult. Based on the some_customer requests I see that this can improve performance a lot (we have quite different number of matches between fields for single query). Also I added a benchmarks with two numeric indexes - and benchmark with text/tag index to compare a difference in the future

src/core/search/range_tree.h
BagritsevichStepan: Why is reducing lines of code the goal? 😄

It wasn't problematic to add this, and we can easily avoid a lot of unnecessary comparisons.

src/core/search/range_tree.h
BagritsevichStepan: Yes, it can be changed

src/core/search/range_tree.h
dranikpg: It correlates with maintainability and readability. I'm trying to avoid files that you can't properly navigate or understand in a matter of a few minutes - so I try to avoid anything that doesn't offer value. This quadruplication will take up ~ 50/500 = 10% of the range tree code for offering a maybe 1-2% final performance improvement?

src/core/search/range_tree.h
BagritsevichStepan: In the benchmarks, there are also not a large range requests (comparing with +inf queries) that return zero results, so I didn’t observe any performance issues compared to main.

But I agree we could add one more benchmark specifically for that case.

Also, I’ve experimented quite a bit with block_size, and it is the most performance-critical parameter in this whole PR — even a small change in its value can cause a 2x drop in performance.

Possible optimization: It should have dynamic value. We need just monitor average_range_block_count_per_range_request inside each field index individually (i.e., allow different numeric indexes to have different RangeBlockSize), and if the average exceeds a threshold (say, >4), we could increase range_block_size accordingly. 
To test this idea, we could add a relatively simple benchmark.

src/core/search/range_tree.h
dranikpg: True, yet:
* The comments above (about fields, queue design) still apply
* We don't have skiplist merges for blocklists yet. Let's start simple. Once we do, we'll just wrap it into a stateful struct

dranikpg: Ironically one of your most common cases, a range that lands in one or two blocks (they will be "border" blocks) is not a fast path: you filter them, copying filtered values into a temporary container, loosing your natural ("future") skiplist at the same time (bc you turn it into a vector). And you still route everything through the queue

BagritsevichStepan: Yes, when I was writing this logic I didn’t expect that the optimal block size would be so large — so that most range requests only cover 2–3 blocks.
But it’s just a few lines to fix: just change the RangeResult constructor - we can simply store l and r in the RangeResult and not calling those Utils functions you dislike

BagritsevichStepan: You also can see that I'm using in constructor absl::InlinedVector<RangeBlockPointer, 10> blocks , which is not needed (10 is too big number)

dranikpg: > But it’s just a few lines to fix ... we can simply store l and r in the RangeResult 

So then we don't need to handle border blocks separately, then we can simplify the queue... and we'll end up with a simpler solution 🤓

BagritsevichStepan:
PR was splitted in several PRs

Issue #5284, opened 2025-06-12
introduce logical ids for dfs snapshots


romange:
Problem: currently different dfs snapshot files can be loaded as part of the same logical snapshot.

To reproduce: run ` --dbfilename=dump --proactor_threads=4` and save the snapshot, then
run ` --dbfilename=dump --proactor_threads=2` and save the snapshot. Then start again - dragonfly will load 
all 4 shard files even though the last snapshot saved only 2 shard files.

The reason for this - we load all the matching files and the last 2 files `dump-0002.dfs  dump-0003.dfs ` match.
The logic is rather naive and often creates confusion in the community deployments. 

One solution could be - to introduce a metadata field `SNAPSHOT_ID` stored in both summary and data files.
The id could be a random number and only the data files with matched  `SNAPSHOT_ID` should be considered for loading.

BorysTheDev:
@romange, what about a simpler solution to add information regarding the number of shards in the name
for "summary" it will be for example, dump-summary_of16.dfs
dump-summary_of16.dfs
and for other files:
dump-0002_of16.dfs

romange:
this solution is not backward compatible with snapshots we already have. Is it hard to make the aux-string based solution?

PR #5399, opened 2025-07-02
feat: add snapshot-id field into dfs files

BorysTheDev:
fixes: https://github.com/dragonflydb/dragonfly/issues/5284

Problem: Currently, different dfs snapshot files can be loaded as part of the same logical snapshot if they have the same name pattern.

Fix: Added SnapshotId AUX field into dfs files to distinguish files that can belong to another snapshot

### DIFF
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -1927,8 +1927,9 @@ struct RdbLoader::ObjSettings {
   ObjSettings() = default;
 };
 
-RdbLoader::RdbLoader(Service* service)
+RdbLoader::RdbLoader(Service* service, std::string snapshot_id)
     : service_{service},
+      snapshot_id_(std::move(snapshot_id)),
       rdb_ignore_expiry_{GetFlag(FLAGS_rdb_ignore_expiry)},
       script_mgr_{service == nullptr ? nullptr : service->script_mgr()},
       shard_buf_{shard_set->size()} {
@@ -2004,6 +2005,8 @@ error_code RdbLoader::Load(io::Source* src) {
     namespaces->GetDefaultNamespace().GetCurrentDbSlice().IncrLoadInProgress();
   }
 
+  bool correct_snapshot_id = snapshot_id_.empty();
+
   while (!stop_early_.load(memory_order_relaxed)) {
     if (pause_) {
       ThisFiber::SleepFor(100ms);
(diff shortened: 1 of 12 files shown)


romange:
Please change the PR description and add the commit description of the change.

src/server/rdb_load.h
```diff
@@ -253,6 +253,10 @@ class RdbLoader : protected RdbLoaderBase {
     pause_ = pause;
   }
 
+  const std::string& SnapshotId() const {
```
romange: naming: either `snapshot_id() const` or `GetSnapshotId() const`

src/server/rdb_load.cc
```diff
@@ -2405,6 +2414,14 @@ error_code RdbLoader::HandleAux() {
      * information fields and are logged at startup with a log
      * level of NOTICE. */
     LOG(INFO) << "RDB '" << auxkey << "': " << auxval;
+  } else if (auxkey == "snapshot-id") {
+    if (snapshot_id_.empty()) {
+      snapshot_id_ = auxval;
+    } else if (snapshot_id_ != auxval) {
+      *correct_snapshot_id = false;
```
romange: why do you need `correct_snapshot_id` if we return incorrect_snapshot_id ?

BorysTheDev: for backward compatibility with files without snapshot_id

BorysTheDev: But I can remove incorrect_snapshot_id from this place, I've added it here only for consistency

romange: `HandleAux()` returns `incorrect_snapshot_id` only for files with mismatching snapshot id and it keeps backward compatibility. Therefore the caller can check the error_code whether it is `incorrect_snapshot_id` instead of passing the boolean. Where am I wrong?

BorysTheDev: In the code above,
```
    if (!correct_snapshot_id) {
      if (type != RDB_OPCODE_AUX) {
        return RdbError(errc::incorrect_snapshot_id);
      }
    }
```

I need to know if snapshot_id was already processed or not and if it is not processed, the next type should be RDB_OPCODE_AUX. In the case when this field is absent, without this variable, I can not distinguish the situation if snapshot_id is present or not

romange:
@BorysTheDev  I added my commit and all the rdb_tests pass. Please let me know if you think it's wrong.

src/server/rdb_save.cc
```diff
@@ -1450,6 +1451,9 @@ error_code RdbSaver::SaveAux(const GlobalData& glob_state) {
 
   error_code ec;
 
+  // Should be first
+  RETURN_ON_ERR(impl_->SaveAuxFieldStrStr("snapshot-id", snapshot_id_));
```
romange: we should not save it if snapshot_id_ is empty

romange: to keep compatibility with the RDB format

romange:
please rebase

PR #5403, opened 2025-07-02
chore: Enable point_in_time_snapshot by default

mkaruza:


src/server/snapshot.cc
```diff
@@ -20,8 +20,7 @@
 #include "server/tiered_storage.h"
 #include "util/fibers/synchronization.h"
 
-ABSL_FLAG(bool, point_in_time_snapshot, false,
-          "If true replication uses point in time snapshoting");
+ABSL_FLAG(bool, point_in_time_snapshot, true, "If true replication uses point in time snapshoting");
```
kostasrim, line 23: don't we have issues around point in time snapshots ? shouldn't we wait until it is stable ?

romange: the flag disables the new optimization.

PR #5405, opened 2025-07-02
feat(search_family): Add SortedVector<pair<DocId, double>>

BagritsevichStepan:
First PR for improving performance of `NumericRange::Range`. It is a part of this PR #5397 .

This PR doesn’t improve performance by itself — it’s just a groundwork step for next steps. It’s needed to implement range trees.

What’s included:
- Made `SortedSet` templated so it can store `std::pair<DocId, double>` entries.
- Updated `BlockList` to support the templated `SortedSet`.
- Added `Split` method to `BlockList`. I think it makes more sense to keep this logic inside `BlockList` rather than as a utility method in the range tree. 
- Added several tests for split method

### DIFF
--- a/src/core/search/block_list.h
+++ b/src/core/search/block_list.h
@@ -13,6 +13,18 @@
 #include "core/search/compressed_sorted_set.h"
 
 namespace dfly::search {
+
+// Forward declarations
+struct SplitResult;
+template <typename Container> class BlockList;
+template <typename T> class SortedVector;
+
+/* Split into two blocks, left and right, so that both blocks have approximately the same number
+   of elements. Returns median value of the split. Garantees that median present in the right
+   block and not present in the left block. Does not work for empty BlockList. */
+// TODO: Move to RangeTree logic
+SplitResult Split(BlockList<SortedVector<std::pair<DocId, double>>>&& result);
+
 // BlockList is a container wrapper for CompressedSortedSet / vector<DocId>
 // to divide the full sorted id range into separate blocks. This reduces modification
 // complexity from O(N) to O(logN + K), where K is the max block size.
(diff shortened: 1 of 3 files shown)


src/core/search/block_list.h
```diff
@@ -18,20 +19,55 @@ namespace dfly::search {
...
 
   // Remove element, returns true if removed, false if not found.
-  bool Remove(DocId t);
+  bool Remove(ElementType t);
+
+  /* Split into two blocks, left and right, so that both blocks have approximately the same number
+     of elements. Returns median value of the split. Garantees that median present in the right
+     block and not present in the left block. Does not work for empty BlockList. */
+  // TODO: remove enable_if_t from here, add template return type
+  template <typename U = ElementType>
```
romange: I do not think this should be a class method. Not every operation should be a method. I also looked at the implementation and I did not see any private dependencies that require this to be a class method. Instead this should be a global function operating on `BlockList<pair<DocId, double>>` type located in the module that calls Split.

src/core/search/block_list_test.cc
```diff
@@ -103,6 +160,131 @@ TYPED_TEST(BlockListTest, RandomNumbers) {
   }
 }
 
+class BlockListTest : public testing::Test {
+ protected:
+};
+
+TEST_F(BlockListTest, Split) {
```
romange, line 167: Just for the testing purposes global Split should be exposed via header file but I do not think it should be a class member.

src/core/search/block_list.h
BagritsevichStepan: @romange I agree here. But it will good to make it as friend method to the BlockList class, because I don't want to make a whole copy here since the BlockList can be really big (and only then clear data), instead of this I can move each block separately and clear it after the copying like this:
```cpp
std::vector<std::pair<DocId, double>> all_entries;
  for (auto& block : block_list.blocks_) {
    all_entries.reserve(all_entries.size() + block.Size());
    all_entries.insert(all_entries.end(), block.begin(), block.end());
    block.Clear(); // At the same time clear the block to save memory
 }
```

romange: friend is fine though I am not sure how you can make a specialized function to be a friend of a template class.

Another option (which I would use) provide a low level public interface to blocks_. 
`BlockCount() const ` and  `Container PullBlock(size_t block_index)`

romange (approved):

src/core/search/block_list.h
BagritsevichStepan: Still need to expose even more methods in BlockList to make this function working (without friend). Like get_memory_resource, get_base_block_size

PR #5406, opened 2025-07-02
test: Add Dragonfly search benchmark tests

vyavdoshenko:
Enables performance testing of search functionality under various load conditions.

### DIFF
--- a/tests/dragonfly/benchmark_test.py
+++ b/tests/dragonfly/benchmark_test.py
@@ -1,6 +1,7 @@
+import asyncio
 import logging
 import time
-import redis
+from redis import asyncio as aioredis
 
 from . import dfly_args
 from .instance import DflyInstance
@@ -15,7 +16,7 @@ from .benchmark_utils import (
 )
 
 
-def run_dragonfly_benchmark(
+async def run_dragonfly_benchmark(
     df_server: DflyInstance,
     num_accounts: int = 200,
     num_queries: int = 500,
@@ -28,18 +29,18 @@ def run_dragonfly_benchmark(
     logging.info(
         f"Parameters: {num_accounts} accounts, {num_queries} queries, {num_agents} agents, seed={random_seed}"
     )
-    client = redis.Redis(port=df_server.port)
+    client = df_server.client()
 
     # Basic connectivity check
-    assert client.ping() == True
+    assert await client.ping() == True
 
     # Stage 1: Schema Generation
     logging.info("Stage 1: Schema Generation - generating columns and creating search index")
     account_columns = generate_account_columns()
-    create_search_index(client, account_columns)
+    await create_search_index(client, account_columns)
 
     # Verify the index was created
-    index_info = client.execute_command(f"FT.INFO {INDEX_KEY}")
+    index_info = await client.execute_command(f"FT.INFO {INDEX_KEY}")
     assert index_info is not None
     logging.info(
         f"Stage 1 completed: search index '{INDEX_KEY}' created with {len(account_columns)} columns"
@@ -54,7 +55,7 @@ def run_dragonfly_benchmark(
         f"Stage 2: Data Generation - generating {num_accounts:,} accounts with full column data"
     )
     stage2_start = time.perf_counter()
-    account_ids = generate_account_data(
+    account_ids = await generate_account_data(
         client=client,
         column_mappings=column_mappings,
         num_accounts=num_accounts,
@@ -67,9 +68,9 @@ def run_dragonfly_benchmark(
     # Verify some accounts were stored
     sample_account_id = account_ids[0]
     account_key = f"AccountBase:{sample_account_id}"
-    stored_account = client.hgetall(account_key)
+    stored_account = await client.hgetall(account_key)
     assert stored_account is not None
-    assert stored_account[b"AccountId"].decode() == sample_account_id
+    assert stored_account["AccountId"] == sample_account_id
     stage2_duration = time.perf_counter() - stage2_start
     logging.info(
         f"Stage 2 completed in {stage2_duration:.2f}s: {len(account_ids)} accounts generated and stored"
@@ -80,8 +81,8 @@ def run_dragonfly_benchmark(
         f"Stage 3: Query Load Testing - running {num_queries:,} queries with {num_agents} concurrent agents"
     )
     stage3_start = time.perf_counter()
-    total_completed = run_query_load_test(
-        client=client,
+    total_completed = await run_query_load_test(
+        df_server=df_server,
         column_mappings=column_mappings,
         account_ids=account_ids,
         total_queries=num_queries,
@@ -102,7 +103,10 @@ def run_dragonfly_benchmark(
 
     logging.info("Benchmark test completed successfully")
 
+    # Close client
+    await client.aclose()
+
 
 @dfly_args({"proactor_threads": 4})
-def test_dragonfly_benchmark(df_server: DflyInstance):
-    run_dragonfly_benchmark(df_server)
+async def test_dragonfly_benchmark(df_server: DflyInstance):
+    await run_dragonfly_benchmark(df_server)
--- a/tests/dragonfly/benchmark_test.py
+++ b/tests/dragonfly/benchmark_test.py
@@ -109,5 +109,7 @@ async def run_dragonfly_benchmark(
 @dfly_args({"proactor_threads": 4})
 @pytest.mark.opt_only
 @pytest.mark.slow
-async def test_dragonfly_benchmark(df_server: DflyInstance):
-    await run_dragonfly_benchmark(df_server)
+async def test_dragonfly_benchmark(
+    df_server: DflyInstance,
+):
+    await run_dragonfly_benchmark(df_server, 20000, 1000)
--- a/tests/dragonfly/benchmark_test.py
+++ b/tests/dragonfly/benchmark_test.py
@@ -1,12 +1,3 @@
-"""
-Dragonfly benchmark test
-Replicates the workflow from original POC scripts in a single pytest test:
-1. schema_generator.py - Create search index
-2. account_data_generator.py - Generate test data
-3. account_queries.py - Run concurrent queries
-4. account_updates.py - Run concurrent updates
-"""
-
 import redis
 
 from . import dfly_args
@@ -15,24 +6,20 @@ from .benchmark_utils import (
     generate_account_columns,
     create_search_index,
     get_column_name_to_type_mapping,
+    generate_account_data,
     INDEX_KEY,
 )
 
 
 @dfly_args({"proactor_threads": 4})
 def test_dragonfly_benchmark(df_server: DflyInstance):
-    """
-    Comprehensive Dragonfly benchmark test
-    Gradually implements all POC functionality in stages
-    """
-    # Create redis client from df_server fixture
     client = redis.Redis(port=df_server.port)
 
     # Basic connectivity check
     assert client.ping() == True
 
-    # Stage 1: Schema Generation (replaces schema_generator.py)
-    account_columns = generate_account_columns(num_columns=50)
+    # Stage 1: Schema Generation
+    account_columns = generate_account_columns()  # Uses default 2774 columns like original
     create_search_index(client, account_columns)
 
     # Verify the index was created
@@ -43,6 +30,23 @@ def test_dragonfly_benchmark(df_server: DflyInstance):
     column_mappings = get_column_name_to_type_mapping(account_columns)
     assert len(column_mappings) == len(account_columns)
 
-    # TODO: Stage 2 - Data Generation (account_data_generator.py)
+    # Stage 2: Data Generation
+    account_ids = generate_account_data(
+        client=client,
+        column_mappings=column_mappings,
+        num_accounts=1000,  # Original size
+        chunk_size=1000,  # Efficient chunk size for large dataset
+    )
+
+    # Verify data was generated
+    assert len(account_ids) == 1000
+
+    # Verify some accounts were stored
+    sample_account_id = account_ids[0]
+    account_key = f"AccountBase:{sample_account_id}"
+    stored_account = client.hgetall(account_key)
+    assert stored_account is not None
+    assert stored_account[b"AccountId"].decode() == sample_account_id
+
     # TODO: Stage 3 - Load Testing (account_queries.py + account_updates.py)
     # TODO: Stage 4 - Results & Cleanup
--- a/tests/dragonfly/benchmark_test.py
+++ b/tests/dragonfly/benchmark_test.py
@@ -91,5 +91,4 @@ def run_dragonfly_benchmark(
 
 @dfly_args({"proactor_threads": 4})
 def test_dragonfly_benchmark_small(df_server: DflyInstance):
-    """Small benchmark test for quick testing"""
-    run_dragonfly_benchmark(df_server, num_accounts=1000, num_queries=100, num_agents=5)
+    run_dragonfly_benchmark(df_server, num_accounts=1000, num_queries=10000, num_agents=50)
--- a/tests/dragonfly/benchmark_test.py
+++ b/tests/dragonfly/benchmark_test.py
@@ -1,4 +1,5 @@
 import logging
+import time
 import redis
 
 from . import dfly_args
@@ -16,8 +17,8 @@ from .benchmark_utils import (
 
 def run_dragonfly_benchmark(
     df_server: DflyInstance,
-    num_accounts: int = 10000,
-    num_queries: int = 1000,
+    num_accounts: int = 200,
+    num_queries: int = 500,
     num_agents: int = 50,
     random_seed: int = 42,
 ):
@@ -52,6 +53,7 @@ def run_dragonfly_benchmark(
     logging.info(
         f"Stage 2: Data Generation - generating {num_accounts:,} accounts with full column data"
     )
+    stage2_start = time.perf_counter()
     account_ids = generate_account_data(
         client=client,
         column_mappings=column_mappings,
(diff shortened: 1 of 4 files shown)


vyavdoshenko:
@romange 
I run it using pytest. It takes ~90 sec on my side, but it is configurable to use on a testing machine.
Please suggest what to do? I can move it to a separate folder, or I can add a skip attribute.
However, it should pass the CI at least once to make sure it can be used.

BagritsevichStepan:
- Please increase number of columns to 500.
- Use LIMIT 0 50 for each request. 
- Another thing is also to have big RETURNs or LOADs (with ~300 fields)

I suggest you to collect these metrics:
- average_number_of_fields_per_query
- average_number_of_numeric_ranges_per_query
- average_length_of_those_ranges

And use them in this benchmark

I have already collected some other metrics from the traffic. Will write here a bit later

tests/dragonfly/benchmark_test.py
```diff
@@ -0,0 +1,108 @@
+import logging
+import time
+import redis
```
romange: please use `from redis import asyncio as aioredis` and asyncio everywhere.

tests/dragonfly/benchmark_test.py
```diff
@@ -0,0 +1,108 @@
...
+    num_agents: int = 50,
+    random_seed: int = 42,
+):
+    set_random_seed(random_seed)
+
+    logging.info(f"Starting Dragonfly benchmark test on port {df_server.port}")
+    logging.info(
+        f"Parameters: {num_accounts} accounts, {num_queries} queries, {num_agents} agents, seed={random_seed}"
+    )
+    client = redis.Redis(port=df_server.port)
```
romange: `df_server.client()` creates a new asyncio client.

tests/dragonfly/benchmark_test.py
```diff
@@ -0,0 +1,108 @@
...
+    create_search_index,
+    get_column_name_to_type_mapping,
+    generate_account_data,
+    run_query_load_test,
+    set_random_seed,
+    INDEX_KEY,
+)
+
+
+def run_dragonfly_benchmark(
```
romange: I suggest splitting the filling up phase and the query phase.
For example to use fixtures like this:
```py
@pytest.fixture(scope="class")
def search_dataset(df_server):
    """
    A class-scoped fixture to create an in-memory store with prefilled data
    """
    # Code adding documents
    # ..........
    yield df_server
```    

then we may have multiple tests inside the same class reusing the same instance.

```py
class TestQueryNumeric:
    def test_query_type1(self, search_dataset):
           ....
   def test_query_type2(self, search_dataset):
           ....
   
```

in theory it should reuse the same dragonfly instance.

tests/dragonfly/benchmark_test.py
```diff
@@ -0,0 +1,108 @@
...
+
+    # Final summary
+    logging.info(
+        f"Benchmark Timings Summary -> Data Generation: {stage2_duration:.2f}s | Query Load: {stage3_duration:.2f}s"
+    )
+
+    logging.info("Benchmark test completed successfully")
+
+
+@dfly_args({"proactor_threads": 4})
```
romange: please use `@pytest.mark.opt_only` and `@pytest.mark.slow` so it will only run for regression and in release mode.

tests/dragonfly/benchmark_test.py
romange: update: I checked it and unfortunately it's not possible with the current configuration. we need to change the scope of df_server and df_factory to "class" for that.

romange (commented):
looks good. mostly style comments. please use async interface in our tests

tests/dragonfly/benchmark_utils.py
```diff
@@ -0,0 +1,334 @@
...
+def run_query_load_test(
+    client: redis.Redis,
+    column_mappings: Dict[str, str],
+    account_ids: List[str],
+    total_queries: int = 1000,
+    num_agents: int = 50,
+) -> int:
+    queries_per_agent = total_queries // num_agents
+
+    with ThreadPoolExecutor(max_workers=num_agents) as executor:
```
kostasrim: This is still single threaded because of GIL

vyavdoshenko:
> My general question to this, is how does it stress dragonfly and provides meaningful information since this is a python process that is single threaded ? In other words, did you manage to reach 100% cpu usage with this script ?

This is the main idea to reach this workload. I will check the thread pool, but the current test makes a 400% load for 4 proactors during the queries phase.

Screenshot 2025-07-03 at 09 37 19

tests/dragonfly/benchmark_test.py
```diff
@@ -0,0 +1,108 @@
...
+    generate_account_data,
+    run_query_load_test,
+    set_random_seed,
+    INDEX_KEY,
+)
+
+
+def run_dragonfly_benchmark(
+    df_server: DflyInstance,
+    num_accounts: int = 200,
```
BagritsevichStepan: Is it number of documents? Let's rename it to num_documents. And why it so small? We don't want to benchmark big number of requests that contain small amount of documents. We need to benchmark dragonfly working with big requests, for single range we should match from 1000-6000 documents, and then merge them. After merging we should have  also a big number of documents (73408.63 or 100000)

```
documents_before_limit_option:
  avg:    73408.63
  median: 100000

documents_after_limit_option:
  avg:    50
  median: 50

fields_per_doc_global: 149.69
avg_fields_per_doc_per_query: 139.86
return_options_fields_count: 140.5
```

From coordinator thread statistics you can see how match documents was matched in avg or median (73408.63/100000) per single query. And avg number of fields per one document in request - around 150. It is a big load and we need to reproduce it

tests/dragonfly/benchmark_utils.py
romange: It's true but for I/O intensive workloads, it can still help because one thread is blocked and another is running.

kostasrim: so it just multiplexes io request, I see

tests/dragonfly/benchmark_utils.py
```diff
@@ -0,0 +1,334 @@
...
+    for _ in range(num_queries):
+        try:
+            query = generate_search_query(column_mappings, account_ids)
+            _ = client.ft(INDEX_KEY).search(query)
+            query_count += 1
+        except Exception:
+            # Ignore individual query failures
+            pass
+
+    return query_count
```
dranikpg: Do we benchmark to get real numbers or just to have a comparison from version to version? We waste time on:
1. Generating queries in between sending them (we can generate them ahead)
2. The client structurally parsing the response to make objects/dicts/lists out of it (really expensive)
3. Roundtrip times for each request separately
4. Serializing documents (I don't know if we want to include it in the benchmarks)

With anything less than ~a few hundred thousand doument volume you'll be mostly measuring python and network time

tests/dragonfly/benchmark_utils.py
```diff
@@ -0,0 +1,334 @@
...
+    elif column_type == ServerTypes.UNIQUE_IDENTIFIER:
+        return random.choice(PRE_GENERATED_UIDS)
+    elif column_type == ServerTypes.TIMESTAMP:
+        return None
+    elif column_type == ServerTypes.IMAGE:
+        return None
+    else:
+        raise NotImplementedError(
+            f"Type {column_type} is not implemented in generate_property_value"
+        )
```
dranikpg: We generate our schema in one namespace of values and then surject it onto the search namespace of types, _treating all of them equally (except bit)_. Why not use plain types directly?

tests/dragonfly/benchmark_utils.py
```diff
@@ -0,0 +1,334 @@
...
+    # Track generated names to guarantee uniqueness
+    existing_names: set[str] = set()
+
+    # Add some standard columns
+    standard_columns = [
+        {"COLUMN_NAME": "AccountId", "TYPE_NAME": ServerTypes.UNIQUE_IDENTIFIER},
+        {"COLUMN_NAME": "Name", "TYPE_NAME": ServerTypes.NVARCHAR},
+        {"COLUMN_NAME": "AccountNumber", "TYPE_NAME": ServerTypes.NVARCHAR},
+        {"COLUMN_NAME": "Revenue", "TYPE_NAME": ServerTypes.MONEY},
+        {"COLUMN_NAME": "NumberOfEmployees", "TYPE_NAME": ServerTypes.INT},
```
dranikpg: 1. Why not make columns just a dict[str->str] ? No need for existing_names then
2. Why have 9 named columns in a total of 2774 🙂

tests/dragonfly/benchmark_utils.py
```diff
@@ -0,0 +1,334 @@
...
+
+    PRE_GENERATED_STRINGS.extend(
+        [
+            "".join(random.choices(string.ascii_letters, k=k))
+            for _ in range(size)
+            for k in range(5, 11)  # lengths 5–10
+        ]
+    )
+
+    PRE_GENERATED_UIDS.extend([str(uuid.uuid4()) for _ in range(size)])
```
dranikpg: If you want some more interesting data you can take a look at faker (famous library)

tests/dragonfly/benchmark_utils.py
dranikpg: https://github.com/dragonflydb/dragonfly/pull/5406#issuecomment-3031058862

`/home/.../build-dbg/...` 😏

tests/dragonfly/benchmark_test.py
```diff
@@ -0,0 +1,111 @@
+import logging
```
romange: please rename the file to search_benchmark_test.py

tests/dragonfly/benchmark_test.py
```diff
@@ -0,0 +1,111 @@
...
+    run_query_load_test,
+    set_random_seed,
+    INDEX_KEY,
+    DOCUMENT_KEY,
+)
+
+
+async def run_dragonfly_benchmark(
+    df_server: DflyInstance,
+    num_documents: int = 200,
```
romange: I am not fond of default arguments here. Assuming we want to run multiple configurations in the future, I think all but `random_seed` should be passed explicit. `random_seed` can stay optional

tests/dragonfly/benchmark_test.py
```diff
@@ -0,0 +1,111 @@
...
+    await client.aclose()
+
+
+@dfly_args({"proactor_threads": 4})
+@pytest.mark.opt_only
+@pytest.mark.slow
+async def test_dragonfly_benchmark(
+    df_server: DflyInstance,
+):
+    await run_dragonfly_benchmark(df_server, 10000, 1000)
```
romange: I think this benchmark is unnecessary heavy.

1. With 200 queries and 5000 documents I got significant slowness on DF side (see `Commandstats` section). There is no reason to prolong the benchmark further.
2. Similarly, check if 25 agents are enough. Btw, please comment in the code what `agents` mean here. Is it just concurrent clients sending search queries?

tests/dragonfly/benchmark_test.py
```diff
@@ -0,0 +1,111 @@
...
+        df_server=df_server,
+        columns=document_columns,
+        document_ids=document_ids,
+        total_queries=num_queries,
+        num_agents=num_agents,
+    )
+
+    # Verify queries completed
+    assert total_completed == num_queries
+    stage3_duration = time.perf_counter() - stage3_start
```
romange: perf_counter() is totally overkill here. It's not a micro benchmark, we do not measure microseconds.

tests/dragonfly/benchmark_utils.py
```diff
@@ -0,0 +1,282 @@
+import asyncio
```
romange: please rename to search_benchmark_utils.py

tests/dragonfly/benchmark_test.py
```diff
@@ -0,0 +1,111 @@
...
+    logging.info(
+        f"Stage 3 completed in {stage3_duration:.2f}s: {total_completed} queries executed successfully"
+    )
+
+    # Final summary
+    logging.info(
```
romange: It's good but please also add output of `info('commandstats')` and `info('latencystats')` because they show real, server side performance of queries

romange (commented):
I think as a first version this benchmark is good, some minor comments

tests/dragonfly/benchmark_test.py
romange: > It's good but please also add output of `info('commandstats')` and `info('latencystats')` because they show real, server side performance of queries

Also add `info('memory')`  output as well - we are pretty bad memory wise.

tests/dragonfly/benchmark_test.py
BagritsevichStepan: The reason could be that we have a huge amount of columns here (could be 600). And we do not use RETURN or LOAD here with the values from 150 to 300. So, for each query we are printing all columns for each matched document. So, we will benchmark io here.

romange: So how to improve it?

BagritsevichStepan: 1. Decrease the default number of columns in the test to 600.
2. For each FT.SEARCH query, use a RETURN option: randomly generate a number between 100 and 300, and then randomly select that many field names — similar to how the customer constructs their queries.

tests/dragonfly/benchmark_test.py
vyavdoshenko: fixed

tests/dragonfly/benchmark_test.py
vyavdoshenko: fixed

tests/dragonfly/benchmark_test.py
vyavdoshenko: I lowered the numbers for now. Please take a look.

vyavdoshenko: agents field was renamed to concurent_clients

tests/dragonfly/benchmark_test.py
vyavdoshenko: fixed

tests/dragonfly/benchmark_utils.py
vyavdoshenko: fixed

tests/dragonfly/benchmark_test.py
vyavdoshenko: added

tests/dragonfly/benchmark_test.py
vyavdoshenko: fixed

tests/dragonfly/benchmark_test.py
vyavdoshenko: fixed

tests/dragonfly/benchmark_test.py
vyavdoshenko: I suggest adding it later when we decide to add more than one benchmark
Do you agree?

tests/dragonfly/benchmark_test.py
vyavdoshenko: fixed

tests/dragonfly/benchmark_utils.py
vyavdoshenko: I think it's not relevant for now. But yes, queries should be processed in a multi-threaded way. But it seems the current approach is good enough for a fresh start.

tests/dragonfly/benchmark_utils.py
vyavdoshenko: I did only one thing. I did my best to reproduce the customer issue. This test started as a simple version of a bunch of scripts from the customer. That's why there is so much weird stuff.
Now the test is simplified. Please take a look again.

tests/dragonfly/benchmark_utils.py
vyavdoshenko: This is how the customer does.

tests/dragonfly/benchmark_utils.py
vyavdoshenko: This is a recreation of real data from the customer, and yes, they have 2774 columns.

tests/dragonfly/benchmark_utils.py
vyavdoshenko: Ok, thanks for pointing out

tests/dragonfly/benchmark_utils.py
BagritsevichStepan: @vyavdoshenko Where did you find that they have 2774 columns? FT.INFO command prints 693 "columns"

vyavdoshenko: I double checked, yes, you are correct. 693 columns.
My bad.

BagritsevichStepan: @vyavdoshenko Ok, now lets reduce number of columns to 600 or 700. ~Please do not ignore comments about RETURN option (range should be 150-300 or num_collumns/3.5 to num_columns/2). This is very reasonable for a benchmark. Currently, we can't retrieve a sufficient number of DocIds to properly benchmark the query engine~

tests/dragonfly/search_benchmark_test.py
```diff
@@ -0,0 +1,151 @@
...
+        latency_stats = await client.info("latencystats")
+        logging.info("Latency Statistics:")
+        for key, value in latency_stats.items():
+            if "ft." in key.lower():
+                logging.info(f"  {key}: {value}")
+    except Exception as e:
+        logging.warning(f"Could not retrieve latency stats: {e}")
+
+    # Memory statistics
+    try:
```
romange: nit: these try/except blocks are unnecessary.

tests/dragonfly/search_benchmark_test.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+@dfly_args({"proactor_threads": 4})
+@pytest.mark.opt_only
+@pytest.mark.slow
+async def test_dragonfly_benchmark(
+    df_server: DflyInstance,
+):
+    # num_documents, num_queries, num_concurrent_clients, random_seed
+    await run_dragonfly_benchmark(df_server, 1000, 200, 25, 42)
```
romange: `num_documents=1000` is probably too low. I suggest increasing to 3000

romange: we can decrease number of queries to 100 and number of agents to 10 if test takes too much time.

tests/dragonfly/search_benchmark_test.py
romange: same above

romange (commented):
LGTM but num_documents=1000 is too low imho.

tests/dragonfly/benchmark_test.py
romange: Stefan, you are correct but it would be simpler if you add further improvements as part of your PR

tests/dragonfly/benchmark_utils.py
BagritsevichStepan: @vyavdoshenko Upd: you can ignore it for now, I will push another PR for this

tests/dragonfly/search_benchmark_test.py
vyavdoshenko: fixed

tests/dragonfly/search_benchmark_test.py
vyavdoshenko: fixed

PR #5409, opened 2025-07-03
chore: introduce class scope fixtures

romange:
With this change, every global function still has its unique fixtures created. But now we can decide combining multiple functions under the same class, which may optimize running time in some cases. Specifically, when we need to prepare data once and run multiple tests on the same datastore.


tests/dragonfly/conftest.py
```diff
@@ -42,25 +44,28 @@ def pytest_configure(config):
         shutil.rmtree(BASE_LOG_DIR)
 
 
-# runs per test case
-def pytest_runtest_setup(item):
-    # Generate a unique directory name for each test based on its nodeid
+@pytest.fixture(scope="class")
+def df_log_dir(request):
```
romange, line 48: now we create a df_factory exactly once per test or TestClass, and df_log_dir is directly related to df_factory, so lets make it as a fixture dependency.

tests/dragonfly/conftest.py
```diff
@@ -42,25 +44,28 @@ def pytest_configure(config):
...
-
-    test_dir = os.path.join(BASE_LOG_DIR, unique_dir)
-    if os.path.exists(test_dir):
-        shutil.rmtree(test_dir)
-    os.makedirs(test_dir)
+    unique_dir = request.node.name.translate(translator)
+    log_dir = os.path.join(BASE_LOG_DIR, unique_dir)
 
-    # Attach the directory path to the item for later access
-    item.log_dir = test_dir
```
romange, line 57: no need, we can access the fixture

tests/dragonfly/conftest.py
```diff
@@ -112,17 +117,27 @@ def parse_args(args: List[str]) -> Dict[str, Union[str, None]]:
     return args_dict
 
 
-@pytest_asyncio.fixture(scope="function", params=[{}])
-async def df_factory(request, tmp_dir, test_env) -> DflyInstanceFactory:
+@pytest_asyncio.fixture(scope="class")
+def event_loop():
```
romange, line 121: needed because by by default pytest_asyncio has "function" scope

tests/dragonfly/conftest.py
```diff
@@ -181,11 +196,6 @@ def df_server(df_factory: DflyInstanceFactory) -> DflyInstance:
         print("Cluster clients left: ", len(clients_left))
 
 
-@pytest.fixture(scope="function")
```
romange, line 184: moved to server_family_test

tests/dragonfly/server_family_test.py
```diff
@@ -6,31 +6,35 @@
...
-        connection.read_response()
+@pytest.fixture(scope="class")
+def connection(df_server: DflyInstance):
+    return redis.Connection(port=df_server.port)
 
 
-def test_quit_after_sub(connection):
-    connection.send_command("SUBSCRIBE", "foo")
-    connection.read_response()
+class TestServer:
```
romange, line 14: to demonstrate class scope - I moved a few functions here under the same class.

tests/dragonfly/server_family_test.py
```diff
@@ -44,18 +48,13 @@ async def test_multi_exec(async_client: aioredis.Redis):
 """
 
 
-@pytest.mark.skip("Skip until we decided on correct behaviour of eval inside multi")
```
romange, line 47: legacy skip

vyavdoshenko:
LGTM

Creating Dragonfly for each test case means that Dragonfly has a clean state at the start of each test case. How can I ensure that I can reuse the current instance without explicitly issuing FLUSHALL/FLUSHDB?
I guess it should be the default behavior. Also, it would be good to have an option to turn it off.

Even if it is possible to do it easily now, I don't see how it can be done.

romange:
You control this by grouping tests under the same class.

romange:
what you do not see how it can be done?

vyavdoshenko:
I see. I can do it in my class. So I can do FLUSHDB before each test case.
Never mind, I got the idea.

PR #5411, opened 2025-07-03
feat: add more auxillary strings to the snapshot

romange:
This PR adds aux-strings to the header of dfs files.
aux-strings do not break compatibility and potentially provide
better memory management and consistency checks during the snapshot load.

All in all this PR should not change functionality or compatibility
with older Dragonfly versons.

### DIFF
--- a/src/server/dflycmd.cc
+++ b/src/server/dflycmd.cc
@@ -561,7 +561,7 @@ void DflyCmd::ReplicaOffset(CmdArgList args, RedisReplyBuilder* rb) {
 void DflyCmd::Load(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext* cntx) {
   CmdArgParser parser{args};
   parser.ExpectTag("LOAD");
-  string_view filename = parser.Next();
+  string filename = parser.Next<string>();
   ServerFamily::LoadExistingKeys existing_keys = ServerFamily::LoadExistingKeys::kFail;
 
   if (parser.HasNext()) {
@@ -569,7 +569,7 @@ void DflyCmd::Load(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext* cn
     existing_keys = ServerFamily::LoadExistingKeys::kOverride;
   }
 
-  if (parser.Error() || parser.HasNext()) {
+  if (parser.Error() || parser.HasNext() || filename.empty()) {
     return rb->SendError(kSyntaxErr);
   }
 
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -79,7 +79,7 @@ inline auto Unexpected(errc ev) {
   return make_unexpected(RdbError(ev));
 }
 
-static const error_code kOk;
+const error_code kOk;
 
 struct ZiplistCbArgs {
   long count = 0;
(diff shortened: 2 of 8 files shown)


src/server/dflycmd.cc
```diff
@@ -561,15 +560,15 @@ void DflyCmd::ReplicaOffset(CmdArgList args, RedisReplyBuilder* rb) {
 void DflyCmd::Load(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext* cntx) {
   CmdArgParser parser{args};
   parser.ExpectTag("LOAD");
-  string_view filename = parser.Next();
+  string filename{parser.Next()};
```
BorysTheDev: you can propagate the needed type `parser.Next<string>() `

src/server/rdb_load.cc
```diff
@@ -2470,6 +2490,16 @@ error_code RdbLoader::HandleAux() {
     /* Just ignored. */
   } else if (auxkey == "search-index") {
     LoadSearchIndexDefFromAux(std::move(auxval));
+  } else if (auxkey == "shard-count") {
+    uint32_t shard_count;
+    if (absl::SimpleAtoi(auxval, &shard_count)) {
+      shard_count_ = shard_count;
+    }
```
BorysTheDev, line 2484: Maybe CHECK is better instead of if.

src/server/server_family.cc
```diff
@@ -1120,21 +1109,22 @@ std::optional<fb2::Future<GenericError>> ServerFamily::Load(string_view load_pat
...
-  std::string snapshot_id;
-  if (absl::EndsWith(load_path, "summary.dfs")) {
+  LoadOptions load_opts;
+  if (absl::EndsWith(path, "summary.dfs")) {
     // we read summary first to get snapshot_id and load data correctly
-    auto load_result = pool.GetNextProactor()->Await(
-        [&] { return LoadRdb(std::string(load_path), existing_keys, &snapshot_id); });
-    if (!load_result.has_value())
-      aggregated_result->first_error = load_result.error();
+    error_code load_result =
```
BorysTheDev: load_result is error_code, please rename it to result_ec or something else

src/server/server_family.cc
```diff
@@ -1144,15 +1134,13 @@ std::optional<fb2::Future<GenericError>> ServerFamily::Load(string_view load_pat
...
-                      snapshot_id]() {
-      auto sid = snapshot_id;
-      auto load_result = LoadRdb(path, existing_keys, &sid);
-      DCHECK(sid == snapshot_id);
-      if (load_result.has_value())
-        aggregated_result->keys_read.fetch_add(*load_result);
-      else
-        aggregated_result->first_error = load_result.error();
+    auto load_func = [=]() mutable {
+      error_code load_result = LoadRdb(file, existing_keys, &load_opts);
```
BorysTheDev: the same, please rename

src/server/rdb_load.cc
romange: It's not important for correctness as it is an optimization.

BorysTheDev (approved):

src/server/rdb_load.cc
```diff
@@ -2470,6 +2477,21 @@ error_code RdbLoader::HandleAux() {
...
```
BorysTheDev, line 2484: maybe let's do some wrapper, because you do it a lot

PR #5412, opened 2025-07-03
fix(search): Cut off results before serializing

dranikpg:
Same intent as #5381

### DIFF
--- a/src/server/search/search_family.cc
+++ b/src/server/search/search_family.cc
@@ -599,7 +599,7 @@ void SearchReply(const SearchParams& params,
     docs.resize(min(docs.size(), knn_sort_option->limit));
 
     ignore_sort = !params.sort_option || params.sort_option->IsSame(*knn_sort_option);
-    if (!knn_sort_option->score_field_alias.empty())
+    if (params.ShouldReturnField(knn_sort_option->score_field_alias))
       knn_score_ret_field = knn_sort_option->score_field_alias;
   }
 


dranikpg:
```py
doc = {c: c * 2000 for c in string.ascii_letters}
for i in range(10):
    p = r.pipeline(transaction=False)
    for j in range(1000):
        p.hset(f'k{i * 1000 + j}', mapping=doc)
    p.execute()

r.execute_command('ft.create i1 schema a text b text')

s = time.time()
p = r.pipeline()
for _ in range(5):
  p.execute_command("ft.search", "i1", "a*", "LIMIT", "0", "0")
print(p.execute())
print(time.time() - s)
```

| Query | Documents |Prev|Now|
| ---|---|----|---|
|`a*`|10k|1.2s|0.005s|
|`a* b*`|10k|1.3s|0.18s|

src/server/search/search_family_test.cc
```diff
@@ -2330,28 +2339,6 @@ TEST_F(SearchFamilyTest, SearchSortByOptionNonSortableFieldJson) {
 
   resp = Run({"FT.SEARCH", "index", "*", "SORTBY", "text"});
   EXPECT_THAT(resp, expect_expr("text"sv));
-
-  resp = Run({"FT.SEARCH", "index", "*", "SORTBY", "@text"});
-  EXPECT_THAT(resp, expect_expr("text"sv));
-
-  resp = Run({"FT.SEARCH", "index", "*", "SORTBY", "$.text"});
-  EXPECT_THAT(resp, expect_expr("$.text"sv));
```
dranikpg, line 2338: Redis replies with "field not found". Let's not assert implementation defined behavior in tests (that also differs from redis)

src/server/search/search_family_test.cc
```diff
@@ -2330,28 +2339,6 @@ TEST_F(SearchFamilyTest, SearchSortByOptionNonSortableFieldJson) {
...
-}
-
-TEST_F(SearchFamilyTest, SearchSortByOptionNonSortableFieldHash) {
-  Run({"HSET", "h1", "text", "2"});
-  Run({"HSET", "h2", "text", "1"});
-
-  auto resp = Run({"FT.CREATE", "index", "ON", "HASH", "SCHEMA", "text", "TEXT"});
```
dranikpg, line 2345: I parameterized the Sort test to use non sortable fields, so we don't need a separate test (it was also too weak)

src/server/search/doc_index.cc
```diff
@@ -340,64 +329,79 @@ bool ShardDocIndex::Matches(string_view key, unsigned obj_code) const {
...
-    return SearchResult{facade::ErrorReply{std::move(search_results.error)}};
+  size_t limit = params.limit_offset + params.limit_total;
+  auto result = search_algo->Search(&*indices_);
+  if (!result.error.empty())
+    return {facade::ErrorReply(std::move(result.error))};
+
+  // TODO(vlad): LOAD does NOT exist as a FT.SEARCH option, logic is blurry
+  SearchFieldsList fields_to_load = params.ShouldReturnAllFields()
+                                        ? params.load_fields.value_or(SearchFieldsList{})
+                                        : params.return_fields.value_or(SearchFieldsList{});
```
dranikpg, line 374: LOAD works in redis but it isn't documented anywhere, I'm concidering simplifying this logic and removing it later 🤔

src/server/search/doc_index.cc
```diff
@@ -340,64 +329,79 @@ bool ShardDocIndex::Matches(string_view key, unsigned obj_code) const {
...
+    auto key = key_index_.Get(id);
     auto it = db_slice.FindReadOnly(op_args.db_cntx, key, base_->GetObjCode());
 
-    if (!it || !IsValid(*it)) {  // Item must have expired
+    if (!it || !IsValid(*it)) {
       expired_count++;
       continue;
     }
 
+    // TODO(vlad): Separate dry run to load only sortable fields?
```
dranikpg: It might be quicker for some cases (i.e. large json documents) and surely more memory friendly, for small douments doing two doument lookups will be two times slower

src/server/search/doc_index.cc
```diff
@@ -340,64 +329,79 @@ bool ShardDocIndex::Matches(string_view key, unsigned obj_code) const {
...
+    if (auto ko = search_algo->GetKnnScoreSortOption(); ko && ko->score_field_alias == fident) {
+      DCHECK_EQ(result.scores.size(), result.ids.size());  // KNN output is already sorted and cut
+    } else if (IsSortableField(fident, base_->schema)) {
+      auto* idx = indices_->GetSortIndex(fident);
+      result.scores = idx->Sort(&result.ids, limit, params.sort_option->order == SortOrder::DESC);
+    } else {
+      fields_to_load.push_back(params.sort_option->field);
+      keep_all = true;  // we sort after serializing documents
+    }
+  }
```
dranikpg: I don't really like this part, it's too dense because of the long names and calls

src/server/search/doc_index.cc
romange: what's the intent of it? I also suggest checking against https://hub.docker.com/r/valkey/valkey-bundle
that supports search.

src/server/search/doc_index.cc
```diff
@@ -340,64 +329,79 @@ bool ShardDocIndex::Matches(string_view key, unsigned obj_code) const {
...
+  auto result = search_algo->Search(&*indices_);
+  if (!result.error.empty())
+    return {facade::ErrorReply(std::move(result.error))};
+
+  // TODO(vlad): LOAD does NOT exist as a FT.SEARCH option, logic is blurry
+  SearchFieldsList fields_to_load = params.ShouldReturnAllFields()
+                                        ? params.load_fields.value_or(SearchFieldsList{})
+                                        : params.return_fields.value_or(SearchFieldsList{});
+
+  // For sortable fields, apply SORTBY ahead of loading all douments to cut off redundant ones.
```
romange:
```suggestion
  // For sortable fields, apply SORTBY ahead of loading all documents to cut off redundant ones.
```

romange (approved):

romange:
> ```python
> doc = {c: c * 2000 for c in string.ascii_letters}
> for i in range(10):
>     p = r.pipeline(transaction=False)
>     for j in range(1000):
>         p.hset(f'k{i * 1000 + j}', mapping=doc)
>     p.execute()
> 
> r.execute_command('ft.create i1 schema a text b text')
> 
> s = time.time()
> p = r.pipeline()
> for _ in range(5):
>   p.execute_command("ft.search", "i1", "a*", "LIMIT", "0", "0")
> print(p.execute())
> print(time.time() - s)
> ```
> 
> Query	Documents	Prev	Now
> `a*`	10k	1.2s	0.005s
> `a* b*`	10k	1.3s	0.18s

Why you have not added this to our pytests?
You do not have to assert on walltime but we could measure server side latency and we could assert on that.

src/server/search/doc_index.cc
dranikpg: It's an option for ft.aggregate to load fields from documents. Same as RETURN for replies. Because RS has large common code paths for both commands it's handled there without errors

dranikpg:
> Why you have not added this to our pytests?

Should all be the job of a benchmark. What should it assert? That the time it takes for the query is within reaonsable limits? It was not a bug, it was a design error, so not sure...

src/server/search/search_family.cc
```diff
@@ -581,60 +581,29 @@ void SearchReply(const SearchParams& params,
...
-    if (prefix_size_to_sort == docs.size()) {
-      std::sort(docs.begin(), docs.end(), std::move(comparator));
-    } else {
-      std::partial_sort(docs.begin(), docs.begin() + prefix_size_to_sort, docs.end(),
-                        std::move(comparator));
-    }
+    sort_order = SortOrder::ASC;
+    limit_total = min(limit_total, knn_sort_option->limit);
+    total_hits = min(total_hits, knn_sort_option->limit);
+    if (params.ShouldReturnField(knn_sort_option->score_field_alias))
```
BagritsevichStepan, line 602: Note that you broke requests like this `FT.SEARCH "=>[KNN ...]" SORTBY @another_not_knn_field`

dranikpg: You mean the reply format with scores?

romange: If it broke something we should add a test.

BagritsevichStepan: Not only, the result will be incorrect, because first you need to do sorting for knn - then apply knn limit on coordinator - then do sorting by the field + apply limit

romange:
> > Why you have not added this to our pytests?
> 
> Should all be the job of a benchmark. What should it assert? That the time it takes for the query is within reaonsable limits? It was not a bug, it was a design error, so not sure...

I still think your benchmark should be added after #5406 is submitted.

src/server/search/doc_index.cc
```diff
@@ -340,64 +329,79 @@ bool ShardDocIndex::Matches(string_view key, unsigned obj_code) const {
...
-        if (!sort_field_data.empty()) {
-          doc_data[field.NameView()] = sort_field_data.begin()->second;
-        }
-      }
-    }
-
-    auto score = search_results.scores.empty() ? monostate{} : std::move(search_results.scores[i]);
-    out.push_back(SerializedSearchDoc{string{key}, std::move(doc_data), std::move(score)});
+  // If SORTBY was requested, but no scores were prepared, we sort after serializing documents
+  if (params.sort_option && result.scores.empty()) {
```
BagritsevichStepan: Are we going to fix this in next PR? The customer will still have perfomance problems because of this (They are using sorting for non-sortable fields). It still possible to fetch first only sort fields data - apply sorting and limiting - and only then serialize all data

src/server/search/search_family.cc
dranikpg: True, it's annoying, will add it

src/server/search/doc_index.cc
```diff
@@ -340,64 +329,79 @@ bool ShardDocIndex::Matches(string_view key, unsigned obj_code) const {
...
+  // For sortable fields, apply SORTBY ahead of loading all documents to cut off redundant ones.
+  // Otherwise, add field to document lookup. Ignore SORTBY if it matches with a KNN query.
+  bool keep_all = false;
+  if (params.sort_option) {
+    auto fident = params.sort_option->field.GetIdentifier(base_->schema, false);
+    if (auto ko = search_algo->GetKnnScoreSortOption(); ko && ko->score_field_alias == fident) {
+      DCHECK_EQ(result.scores.size(), result.ids.size());  // KNN output is already sorted and cut
+    } else if (IsSortableField(fident, base_->schema)) {
+      auto* idx = indices_->GetSortIndex(fident);
+      result.scores = idx->Sort(&result.ids, limit, params.sort_option->order == SortOrder::DESC);
```
BagritsevichStepan: if we have `FT.SEARCH "=> [KNN ...]" SORTBY @some_sortable_field` we didn't reorder the knn scores. But it can be fixed in next PRs

src/server/search/doc_index.cc
dranikpg: https://github.com/dragonflydb/dragonfly/pull/5412#discussion_r2185040135

src/server/search/doc_index.cc
BagritsevichStepan: What about HSETs?
For small documents, the lookup time should be low — even if it's twice as slow, it would likely still be acceptable.
To me, this seems like a bold claim, and we should have a benchmark for HSETs with around 600 fields and a large number of documents. (It's also important in benchmark that the search_algo returns a large result set (e.g. 10,000 documents) matched by the query)

Also, what if you are sorting by numeric value in hset, but the rest values are texts? I do not think that it will be faster. And you will allocate a lot of memory. As I remember we have std::string (not views) in SortableValue

BagritsevichStepan: So, just `FT.SEARCH "*" SORTBY @non_sortable_field LIMIT 0 50`

src/server/search/doc_index.cc
BagritsevichStepan: We are still not breaking something here - we didn't have it before. I just fixed it in #5381

src/server/search/doc_index.cc
```diff
@@ -340,64 +329,79 @@ bool ShardDocIndex::Matches(string_view key, unsigned obj_code) const {
...
+  // TODO(vlad): LOAD does NOT exist as a FT.SEARCH option, logic is blurry
+  SearchFieldsList fields_to_load = params.ShouldReturnAllFields()
+                                        ? params.load_fields.value_or(SearchFieldsList{})
+                                        : params.return_fields.value_or(SearchFieldsList{});
+
+  // For sortable fields, apply SORTBY ahead of loading all documents to cut off redundant ones.
+  // Otherwise, add field to document lookup. Ignore SORTBY if it matches with a KNN query.
+  bool keep_all = false;
+  if (params.sort_option) {
+    auto fident = params.sort_option->field.GetIdentifier(base_->schema, false);
```
BagritsevichStepan: Please add fields_to_load.remove of the params.sort_option->field here. For requests like `FT.SEARCH "*" SORTBY @some_sortable_field RETURN 1 @some_sortable_field` we will have duplicates of the same field in the answer. Same for non-sortable fields

src/server/search/search_family.cc
```diff
@@ -581,60 +581,29 @@ void SearchReply(const SearchParams& params,
...
-
-    auto sort_begin = docs.begin();
-    auto sort_end = docs.end();
-    // If we first sorted by knn, we need to sort only the result of knn
-    if (knn_sort_option) {
-      sort_begin = docs.begin() + offset;
-      sort_end = docs.begin() + end;
-    }
-    std::sort(sort_begin, sort_end, std::move(comparator));
+    partial_sort(docs.begin(), docs.begin() + min(docs.size(), end), docs.end(), cmp);
```
BagritsevichStepan: no need for min(docs.size(), end), you can use just end.
```
const size_t offset = std::min(params.limit_offset, docs.size());
const size_t limit = std::min(docs.size() - offset, limit_total);
const size_t end = offset + limit;
```

TODO: small optimization could be done here - std::nth_element for begin() + offset and then std::sort(begin() +  offset, begin() + end). But it is not important

dranikpg: > TODO: small optimization could be done here - std::nth_element for begin() + offset and then std::sort(begin() + offset, begin() + limit). But it is not important

1. You still have to use partial sort over one of the halves created by nth_element, but I understand what you mean 🙂 
2. Not first priority because that's the last step before replying, so io time will dominate

dranikpg:
Rewrote this PR:
1. Fixed two-step sorting for KNN + SORTBY (annoying, but let's make it work)
2. Now non-sortable values are first fetched, documents are limited with a priority queue and only then serialized
3. Extended tests

romange (approved):

PR #5414, opened 2025-07-03
feat(search_family): Introduce RangeTree

BagritsevichStepan:


BagritsevichStepan:
Before:

| Benchmark                                             | Time (ns)   | CPU (ns)    | Iterations |
|-------------------------------------------------------|-------------|-------------|------------|
| BM_SearchNumericIndexes/num_docs:10000                | 707,364     | 707,158     | 988        |
| BM_SearchNumericIndexes/num_docs:100000               | 9,149,588   | 9,148,000   | 79         |
| BM_SearchNumericIndexes/num_docs:1000000              | 110,964,310 | 110,936,317 | 6          |
| BM_SearchTwoNumericIndexes/num_docs:10000             | 749,210     | 749,174     | 919        |
| BM_SearchTwoNumericIndexes/num_docs:100000            | 9,384,238   | 9,381,866   | 75         |
| BM_SearchTwoNumericIndexes/num_docs:1000000           | 121,375,139 | 121,315,705 | 6          |
| BM_SearchNumericAndTagIndexes/num_docs:10000          | 534,896     | 534,852     | 1334       |
| BM_SearchNumericAndTagIndexes/num_docs:100000         | 7,248,734   | 7,242,791   | 100        |
| BM_SearchNumericAndTagIndexes/num_docs:1000000        | 91,959,997  | 91,952,690  | 7          |

After:

| Benchmark                                             | Time (ns)  | CPU (ns)   | Iterations |
|-------------------------------------------------------|------------|------------|------------|
| BM_SearchNumericIndexes/num_docs:10000                | 197,277    | 197,245    | 3577       |
| BM_SearchNumericIndexes/num_docs:100000               | 2,023,187  | 2,022,863  | 341        |
| BM_SearchNumericIndexes/num_docs:1000000              | 17,132,286 | 17,129,637 | 41         |
| BM_SearchTwoNumericIndexes/num_docs:10000             | 521,742    | 521,555    | 1337       |
| BM_SearchTwoNumericIndexes/num_docs:100000            | 5,144,214  | 5,141,662  | 135        |
| BM_SearchTwoNumericIndexes/num_docs:1000000           | 20,440,414 | 20,436,975 | 34         |
| BM_SearchNumericAndTagIndexes/num_docs:10000          | 192,238    | 192,209    | 3648       |
| BM_SearchNumericAndTagIndexes/num_docs:100000         | 1,848,713  | 1,848,378  | 381        |
| BM_SearchNumericAndTagIndexes/num_docs:1000000        | 12,165,174 | 12,162,349 | 59         |

src/core/search/range_tree.cc
```diff
@@ -0,0 +1,251 @@
...
+
+  auto it = entries_.lower_bound({value, std::numeric_limits<RangeNumber>::min()});
+  if (it != entries_.begin() && (it == entries_.end() || it->first.first > value)) {
+    // TODO: remove this, we do log N here
+    // we can use negative left bouding to find the block
+    --it;  // Move to the block that contains the value
+  }
+
+  DCHECK(it != entries_.end() && (it->first.first <= value && value < it->first.second));
+  return it;
```
dranikpg: Let's maybe move common logic between those two functions in a templatized helper function for both const and non const

src/core/search/range_tree.cc
```diff
@@ -0,0 +1,251 @@
...
+After that all median entries were added to the right block.
+
+That means that we have equal values in the whole block,
+and their count is greater than max_range_block_size_.
+So we will do cascade splits of the right block.
+TODO: we can optimize this case by splitting to three blocks:
+ - empty left block with range [l, m),
+ - middle block with range [m, std::nextafter(m, +inf)),
+ - empty right block with range [std::nextafter(m, +inf), r)
+*/
```
BagritsevichStepan, line 130: @dranikpg

src/core/search/range_tree_test.cc
```diff
@@ -0,0 +1,307 @@
...
+    tree.Add(4, 20.0);
+    tree.Add(4, 10.0);
+
+    auto result = tree.Range(10.0, 20.0);
+    EXPECT_THAT(result, BlocksAre({{{1, 10.0}, {2, 10.0}, {3, 10.0}, {4, 10.0}}, {{4, 20.0}}}));
+  }
+}
+
+TEST_F(RangeTreeTest, BugNotUniqueDoubleValues) {
+  // TODO: fix the bug
```
BagritsevichStepan, line 293: also there is a bug in current implementation

BagritsevichStepan: For hsets users will not catch it.

And maybe there's a way to catch it using JSON commands, but I didn’t succeed in reproducing it (on opt build).

BagritsevichStepan: On debug build it can be crashed using:
```
127.0.0.1:6379> json.set k1 . '{"a":[5.0,5.0]}'
OK
127.0.0.1:6379> ft.create index on json schema $.a AS a NUMERIC
Error: Server closed the connection
(1.90s)
```

Because of DCHECKs that I added in Add and Remove methods. `DCHECK(insert_result);` and `DCHECK(remove_result);`.

I suggest to remove them and use LOG(ERROR) instead

src/core/search/range_tree.cc
```diff
@@ -0,0 +1,251 @@
...
+  DCHECK(!blocks_.empty());
+}
+
+std::vector<DocId> RangeResult::MergeAllResults() {
+  using Entry = std::pair<RangeBlockIterator, RangeBlockIterator>;
+
+  // After the benchmarking, it is better to use inlined vector
+  // than std::priority_queue
+  absl::InlinedVector<Entry, 10> heap;
+  heap.reserve(blocks_.size());
```
dranikpg: Did you try single block, two block set_union fast paths?

dranikpg: I really have the suspicion that the range tree can show better results

src/core/search/range_tree.cc
```diff
@@ -0,0 +1,251 @@
...
+  size_t doc_ids_count = 0;
+  for (const auto* block : blocks_) {
+    auto it = block->begin();
+    while (it != block->end()) {
+      if (l_ <= (*it).second && (*it).second <= r_) {
+        heap.emplace_back(it, block->end());
+        doc_ids_count += block->Size();
+        break;
+      }
+      ++it;
```
dranikpg: pure codestyle: 
you an add the block->begin and then call `// Found the next valid entry in the current block` from below

src/core/search/range_tree.cc
BagritsevichStepan: Let's do it in the next PRs

BagritsevichStepan: I wanted to make a separate PR for this change

src/core/search/range_tree.cc
dranikpg: image

If you split, what will the keys of the blocks be? I guess we just can't split 🤷🏻‍♂️

src/core/search/range_tree.cc
BagritsevichStepan: I have already tried that. Implementation your suggest results in a 5× performance degradation.

BagritsevichStepan: But yeah, if you just copy the whole while and move it before this part in the for it will work:
```
for(...) {
// insert here
if ((*heap[i].first).first < (*heap[min_doc_id_index].first).first) {
        min_doc_id_index = i;
      }
    }
```
But it is still not optimal because you sometimes make double-tripple-etc. checks in this while if DocId from some block is still valid for this [l,r] range but bigger than other DocIds (so it will be added at the end)

src/core/search/range_tree.cc
BagritsevichStepan: - empty left block with range [l, m),
- middle block (all elements are here) with range [m, std::nextafter(m, +inf)),
 - empty right block with range [std::nextafter(m, +inf), r)
 
?

Maybe I didn't understand what you mean.

BagritsevichStepan: I do not suggest to add this because it can be performance related. Better to have not so much range blocks.
But what I see from some_customer's traffic - a big number of zero values.

I'm now adding microbenchmark for this

dranikpg: I'd suggest to keep it as it and experiment later with tests. I mean that if you'll have blocks with all equal values - what would the keys to that be in the tree?

src/core/search/range_tree_test.cc
romange: Is it a regression due to this pr?

src/core/search/range_tree.h
```diff
@@ -0,0 +1,80 @@
...
+  using Entry = std::pair<DocId, double>;
+  using RangeBlock = BlockList<SortedVector<Entry>>;
+  using Map = absl::btree_map<Key, RangeBlock, std::less<Key>,
+                              PMR_NS::polymorphic_allocator<std::pair<const Key, RangeBlock>>>;
+
+  static constexpr size_t kMaxRangeBlockSize = 500000;
+  static constexpr size_t kBlockSize = 400;
+
+ public:
+  explicit RangeTree(PMR_NS::memory_resource* mr, size_t max_range_block_size = kMaxRangeBlockSize);
```
romange, line 47: please add comments describing structure of the tree as well as what max_range_block_size is

src/core/search/range_tree.h
```diff
@@ -0,0 +1,80 @@
+#pragma once
```
romange, line 5: header

src/core/search/range_tree.cc
```diff
@@ -0,0 +1,256 @@
+#include "core/search/range_tree.h"
```
romange, line 5: header

BagritsevichStepan:
> Can you split even further and send only range_tree code for review?

This PR is now splitted. Please review it. After I will push the second one

src/core/search/range_tree_test.cc
BagritsevichStepan: No, as I said I can not catch it using our API.
But the tree does not support duplicates of pairs with the same id and the same double value.

But still this is working because when we do for example:
```
127.0.0.1:6379> json.set k1 . '{"a":[5.0,5.0]}'
OK
127.0.0.1:6379> ft.create index on json schema $.a AS a NUMERIC
OK
127.0.0.1:6379> json.set k1 $.a '5.0'
OK
127.0.0.1:6379> ft.search index "@a:[3.0 6.0]"
1) (integer) 1
2) "k1"
3) 1) "$"
   2) "{\"a\":5.0}"
```

The ranges work correctly because when updating a value, we remove the entire JSON and then insert the fully updated version.

src/core/search/range_tree.h
```diff
@@ -0,0 +1,95 @@
...
+   The parameter `max_range_block_size_` defines the maximum number of entries in a single
+   RangeBlock. When a block exceeds this limit, it is split into two to maintain balanced
+   performance.
+*/
+class RangeTree {
+ private:
+  friend class RangeResult;
+
+  using RangeNumber = double;
+  using Key = std::pair<RangeNumber, RangeNumber>;
```
dranikpg, line 37: Note that your Key can be just a single RangeNumber, you don't use the right bound anywhere

src/core/search/range_tree_test.cc
```diff
@@ -0,0 +1,306 @@
...
+  tree.Add(3, 30.0);
+  tree.Add(4, 40.0);
+  tree.Add(4, 60.0);
+
+  auto result = tree.GetAllDocIds();
+  EXPECT_THAT(result, UnorderedElementsAreDocPairs(
+                          {{1, 10.0}, {2, 10.0}, {2, 20.0}, {3, 30.0}, {4, 40.0}, {4, 60.0}}));
+}
+
+TEST_F(RangeTreeTest, Add) {
```
dranikpg, line 86: We should use more heavy tests in the future

PR #5415, opened 2025-07-04
fix(pytest): Increase REPLTAKOVER time in test_take_over_seeder

### LINKED ISSUE #5360: Failure on test_take_over_seeder
https://github.com/dragonflydb/dragonfly/actions/runs/15872215068/job/44751266344

Couldn't execute takeover

mkaruza:
REPLTAKEVER fails when we are replication from 4 proactor threads to 2. Increasing timeout should fix this test.

Closes #5360

### DIFF
--- a/tests/dragonfly/replication_test.py
+++ b/tests/dragonfly/replication_test.py
@@ -1294,7 +1294,7 @@ async def test_take_over_seeder(
     # Give the seeder a bit of time.
     await asyncio.sleep(3)
     logging.debug("running repltakover")
-    await c_replica.execute_command(f"REPLTAKEOVER 10 SAVE")
+    await c_replica.execute_command(f"REPLTAKEOVER 30 SAVE")
     logging.debug("after running repltakover")
     seeder.stop()
     await fill_task


tests/dragonfly/replication_test.py
```diff
@@ -1294,7 +1294,7 @@ async def info_replication():
     # Give the seeder a bit of time.
     await asyncio.sleep(3)
     logging.debug("running repltakover")
-    await c_replica.execute_command(f"REPLTAKEOVER 10 SAVE")
+    await c_replica.execute_command(f"REPLTAKEOVER 30 SAVE")
```
kostasrim, line 1297: So were not robust enough? :face_exhaling:

kostasrim (approved):

tests/dragonfly/replication_test.py
mkaruza: It looks that with latest changes in replication and because of this 4->2 we don't takeover in time. I have run several tests on (https://github.com/dragonflydb/dragonfly/pull/5392) with this increase and it seems that it is successful on all runs.

PR #5416, opened 2025-07-05
feat(search_family): Add RangeTree to NumericIndex

BagritsevichStepan:
second PR to improve perfomance of the NumericIndex::Range method. First PR is here #5414

- numeric benchmarks added in search_test

### DIFF
--- a/src/core/search/indices.h
+++ b/src/core/search/indices.h
@@ -18,6 +18,7 @@
 #include "core/search/base.h"
 #include "core/search/block_list.h"
 #include "core/search/compressed_sorted_set.h"
+#include "core/search/range_tree.h"
 #include "core/search/rax_tree.h"
 
 // TODO: move core field definitions out of big header
@@ -39,9 +40,7 @@ struct NumericIndex : public BaseIndex {
   std::vector<DocId> GetAllDocsWithNonNullValues() const override;
 
  private:
-  bool unique_ids_ = true;  // If true, docs ids are unique in the index, otherwise they can repeat.
-  using Entry = std::pair<double, DocId>;
-  absl::btree_set<Entry, std::less<Entry>, PMR_NS::polymorphic_allocator<Entry>> entries_;
+  RangeTree tree_;
 };
 
 // Base index for string based indices.
--- a/src/core/search/indices.h
+++ b/src/core/search/indices.h
@@ -30,6 +30,22 @@ namespace dfly::search {
 // Index for integer fields.
 // Range bounds are queried in logarithmic time, iteration is constant.
 struct NumericIndex : public BaseIndex {
+  // Temporary base class for range tree.
+  // It is used to use two different range trees depending on the flag use_range_tree.
+  // If the flag is true, RangeTree is used, otherwise a simple implementation with btree_set.
+  struct RangeTreeBase {
+    virtual void Add(DocId id, absl::Span<double> values) = 0;
+    virtual void Remove(DocId id, absl::Span<double> values) = 0;
+
+    // Returns all DocIds that match the range [l, r].
+    virtual std::vector<DocId> Range(double l, double r) const = 0;
+
+    // Returns all DocIds that have non-null values in the index.
+    virtual std::vector<DocId> GetAllDocIds() const = 0;
+
+    virtual ~RangeTreeBase() = default;
+  };
+
   explicit NumericIndex(PMR_NS::memory_resource* mr);
 
   bool Add(DocId id, const DocumentAccessor& doc, std::string_view field) override;
@@ -40,7 +56,7 @@ struct NumericIndex : public BaseIndex {
   std::vector<DocId> GetAllDocsWithNonNullValues() const override;
 
  private:
-  RangeTree tree_;
+  std::unique_ptr<RangeTreeBase> range_tree_;
 };
 
 // Base index for string based indices.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -154,6 +154,8 @@ jobs:
           timeout 5m ./dragonfly_test
           timeout 5m ./json_family_test --jsonpathv2=false
           timeout 5m ./tiered_storage_test --vmodule=db_slice=2 --logtostderr
+          timeout 5m ./search_test --use_numeric_range_tree=false
+          timeout 5m ./search_family_test --use_numeric_range_tree=false
 
 
       - name: C++ Unit Tests - Epoll
--- a/src/core/search/indices.cc
+++ b/src/core/search/indices.cc
@@ -84,7 +84,7 @@ void IterateAllSuffixes(const absl::flat_hash_set<string>& words,
 
 };  // namespace
 
-NumericIndex::NumericIndex(PMR_NS::memory_resource* mr) : entries_{mr} {
+NumericIndex::NumericIndex(PMR_NS::memory_resource* mr) : tree_{mr} {
 }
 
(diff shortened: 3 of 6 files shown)


BagritsevichStepan:
Before:

| Benchmark                                             | Time (ns)   | CPU (ns)    | Iterations |
|-------------------------------------------------------|-------------|-------------|------------|
| BM_SearchNumericIndexes/num_docs:10000                | 707,364     | 707,158     | 988        |
| BM_SearchNumericIndexes/num_docs:100000               | 9,149,588   | 9,148,000   | 79         |
| BM_SearchNumericIndexes/num_docs:1000000              | 110,964,310 | 110,936,317 | 6          |
| BM_SearchTwoNumericIndexes/num_docs:10000             | 749,210     | 749,174     | 919        |
| BM_SearchTwoNumericIndexes/num_docs:100000            | 9,384,238   | 9,381,866   | 75         |
| BM_SearchTwoNumericIndexes/num_docs:1000000           | 121,375,139 | 121,315,705 | 6          |
| BM_SearchNumericAndTagIndexes/num_docs:10000          | 534,896     | 534,852     | 1334       |
| BM_SearchNumericAndTagIndexes/num_docs:100000         | 7,248,734   | 7,242,791   | 100        |
| BM_SearchNumericAndTagIndexes/num_docs:1000000        | 91,959,997  | 91,952,690  | 7          |

After:

| Benchmark                                             | Time (ns)  | CPU (ns)   | Iterations |
|-------------------------------------------------------|------------|------------|------------|
| BM_SearchNumericIndexes/num_docs:10000                | 197,277    | 197,245    | 3577       |
| BM_SearchNumericIndexes/num_docs:100000               | 2,023,187  | 2,022,863  | 341        |
| BM_SearchNumericIndexes/num_docs:1000000              | 17,132,286 | 17,129,637 | 41         |
| BM_SearchTwoNumericIndexes/num_docs:10000             | 521,742    | 521,555    | 1337       |
| BM_SearchTwoNumericIndexes/num_docs:100000            | 5,144,214  | 5,141,662  | 135        |
| BM_SearchTwoNumericIndexes/num_docs:1000000           | 20,440,414 | 20,436,975 | 34         |
| BM_SearchNumericAndTagIndexes/num_docs:10000          | 192,238    | 192,209    | 3648       |
| BM_SearchNumericAndTagIndexes/num_docs:100000         | 1,848,713  | 1,848,378  | 381        |
| BM_SearchNumericAndTagIndexes/num_docs:1000000        | 12,165,174 | 12,162,349 | 59         |

romange (commented):
Is it possible to keep the old implementation and use a runtime flag to control the behavior?

It can point by default to RangeTree but I prefer to have the ability to revert in prod to the old implementation if needed

BagritsevichStepan:
> Is it possible to keep the old implementation and use a runtime flag to control the behavior?
> 
> It can point by default to RangeTree but I prefer to have the ability to revert in prod to the old implementation if needed

I think it will be difficult to do this, since I was planning to push further optimizations today that will change the NumericIndex::Range API. But for this PR, the change is simple.

I think we can just revert the dragonfly version?

dranikpg:
> I think it will be difficult to do this

I don't like such runtime controlled flags as well, but you can just add another type of numeric index and look up the flag on creation. Dynamic_cast or virtual functions for calling Range 

> further optimizations today that will change the NumericIndex::Range API

Returing RangeResult directly? And making it iterable?

I suggest getting a proper flamegraph first and looking what at what takes up most time. Second, we should add better tests for the RangeTree because I'm not sure in it's reliability with edge cases

BagritsevichStepan:
> > I think it will be difficult to do this
> 
> I don't like such runtime controlled flags as well, but you can just add another type of numeric index and look up the flag on creation. Dynamic_cast or virtual functions for calling Range
> 
> > further optimizations today that will change the NumericIndex::Range API
> 
> Returing RangeResult directly? And making it iterable?
> 
> I suggest getting a proper flamegraph first and looking what at what takes up most time. Second, we should add better tests for the RangeTree because I'm not sure in it's reliability with edge cases

Which edge cases are you referring to? If you have any other edge cases please share it. If I knew of any additional edge cases, I would have already added them to the tests.
The case where the number of entries with the same double value exceeds the max range block size is already covered.
The remove test also caught a bug, which I fixed.
Range tests are also validating different edge cases in the split logic.

dranikpg:
> Which edge cases are you referring to?

Mainly around remove. What first comes to my mind is that we don't delete empty blocks (I'm not talking about merging yet).

> // The problem that for each mutable operation we do Remove and then Add,
  // So we can do merge and split for one operation.

Block list uses `N * 2 and N/2` for Split/Merge bound to avoid this scenario. We can do something similar

>  If you have any other edge cases please share it. If I knew of any additional edge cases, I would have already added them to the tests.

I usually think about interesting tests. Adding/Removing in waves, removing every 2nd, 3rd element, adding in reverse, etc (see block list tests like InsertReverseRemoveSteps). The current tests are too weak and won't catch a newly introduced bugs

BagritsevichStepan:
> What first comes to my mind is that we don't delete empty blocks (I'm not talking about merging yet).

We can not delete empty blocks - we need to cover the whole range from [-inf, +inf]. What if you will receive range with values included in empty block that you removed? What if you will remove the first block?

I think you will introduce more bugs and edge cases with removing empty blocks. And as I said in prev PR - the number of nodes in absl::btree_map should be small (~5 nodes) to get good perfomance. So even If you have empty blocks there - it is not a problem at all

> Block list uses N * 2 and N/2 for Split/Merge bound to avoid this scenario. We can do something similar

Yes, that merging logic should be added I agree. But not removing empty blocks.

> Adding/Removing in waves, removing every 2nd, 3rd element

Remove test does the same. Also there is something like adding in reverse there

dranikpg:
> We can not delete empty blocks - we need to cover the whole range from [-inf, +inf]. What if you will remove the first block?

You can delete all except the first one. It's a special case for Merge anyways

> What if you will receive range with values included in empty block that you removed? 

It will go into a different one. Your performance will degrade after a while if you have too many ranges

> Remove test does the same. Also there is something like adding in reverse there

Remove test is actually pretty decent. Didn't notice you updated it

BagritsevichStepan:
> Your performance will degrade after a while if you have too many ranges

How we will get too many ranges? If the max_block_size is 10000 for example (not 500000 as we have now). K ranges means that we added k * 10000 doc ids. If k = 10000, user has 10000*10000=100_000_000 docs. Do you think that 10000 ranges will be a dominant problem here? Also take a look at merging logic - it skips at the beginning all empty blocks, so it is just +10000 iterations

But basically removing empty block - it is just merging now. Because I have already removed std::pair<double, double> from the keys

src/core/search/indices.h
```diff
@@ -39,9 +56,7 @@ struct NumericIndex : public BaseIndex {
   std::vector<DocId> GetAllDocsWithNonNullValues() const override;
 
  private:
-  bool unique_ids_ = true;  // If true, docs ids are unique in the index, otherwise they can repeat.
-  using Entry = std::pair<double, DocId>;
-  absl::btree_set<Entry, std::less<Entry>, PMR_NS::polymorphic_allocator<Entry>> entries_;
+  std::unique_ptr<RangeTreeBase> range_tree_;
```
dranikpg, line 59: Might have been easier to keep the classes and change `CreateIndices/Search(AstRangeNode)` 🤷🏻‍♂️

BagritsevichStepan: But you also need to implement second NumericIndex. Copy GetNumbers logic and etc. Looks like both approaches similar.

PR #5419, opened 2025-07-05
chore(search): LIMIT=0 Fast path and remove LOAD parameter from FT.SEARCH

dranikpg:
1. Add fast path for LIMIT=0
2. Remove LOAD parameter from FT.SEARCH and remove tests for it (Another PR will add proper json tests)

### DIFF
--- a/src/server/search/doc_index.h
+++ b/src/server/search/doc_index.h
@@ -25,6 +25,8 @@
 
 namespace dfly {
 
+struct BaseAccessor;
+
 using SearchDocData = absl::flat_hash_map<std::string /*field*/, search::SortableValue /*value*/>;
 using Synonyms = search::Synonyms;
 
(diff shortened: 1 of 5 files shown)


src/server/search/doc_index.h
```diff
@@ -276,10 +278,13 @@ class ShardDocIndex {
   // Clears internal data. Traverses all matching documents and assigns ids.
   void Rebuild(const OpArgs& op_args, PMR_NS::memory_resource* mr);
 
+  using LoadedEntry = std::pair<std::string_view, std::unique_ptr<BaseAccessor>>;
+  std::optional<LoadedEntry> LoadEntry(search::DocId id, const OpArgs& op_args) const;
+
   // Behaviour identical to SortIndex::Sort for non-sortable fields that need to be fetched first
-  std::vector<search::SortableValue> KeepTopKSorted(std::vector<DocId>* ids,
+  std::vector<search::SortableValue> KeepTopKSorted(std::vector<DocId>* ids, size_t limit,
```
romange, line 284: why `ids` is a mutable pointer? where are they changed?

dranikpg: They are reordered at the end. It's the same interface like SortIndex::Sort

romange (dismissed):

romange (approved):

PR #5423, opened 2025-07-06
feat(search_family): Add BLOCKSIZE option for FT.CREATE

BagritsevichStepan:
Introduce the BLOCKSIZE option for numeric fields in the FT.CREATE command.

This option allows fine-grained control over the block size used by the range tree for each numeric index independently.

**How to use it?**
To control the latency of your range queries, you can use the `BLOCKSIZE` option. If you know the average number of documents matched by your range queries, setting the block size close to that average can significantly speed them up.

Setting the block size higher or lower does not necessarily mean your queries will be faster or slower — it depends on the actual results of your range queries.

**Tip:**
To estimate how many documents your query matches, you can use the `FT.PROFILE` command. After collecting some statistics, you can then tune the BLOCKSIZE accordingly for optimal performance.

**Example of query**
Example of query: `FT.CREATE index ON HASH SCHEMA number1 NUMERIC BLOCKSIZE 2 number2 NUMERIC BLOCKSIZE 1024`.
You can also use `FT.INFO index` to get info about blocksize of the numeric index:
```
127.0.0.1:6379> FT.INFO index
...
5) attributes
6) 1) 1) identifier
      2) number2
      3) attribute
      4) number2
      5) type
      6) NUMERIC
      7) blocksize
      8) 1024
   2) 1) identifier
      2) number1
      3) attribute
      4) number1
      5) type
      6) NUMERIC
      7) blocksize
      8) 2
```

### DIFF
--- a/src/core/search/indices.cc
+++ b/src/core/search/indices.cc
@@ -94,7 +94,8 @@ void IterateAllSuffixes(const absl::flat_hash_set<string>& words,
 
 class RangeTreeAdapter : public NumericIndex::RangeTreeBase {
  public:
-  explicit RangeTreeAdapter(PMR_NS::memory_resource* mr) : range_tree_{mr} {
+  explicit RangeTreeAdapter(size_t max_range_block_size, PMR_NS::memory_resource* mr)
+      : range_tree_(mr, max_range_block_size) {
   }
 
   void Add(DocId id, absl::Span<double> values) override {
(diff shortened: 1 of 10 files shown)


romange:
1. Can you please write in the PR description how smaller/bigger blocksizes affect search?
2. Can you please add to the PR description the example query with the new syntax.
3. Finally, please add a test to search_family_test.cc with such query

romange:
what happens when we specify `BLOCKSIZE 1024` in contrast to `BLOCKSIZE 8192` for example?

BagritsevichStepan:
> what happens when we specify `BLOCKSIZE 1024` in contrast to `BLOCKSIZE 8192` for example?

I'm currently updating PR description. 1min

BagritsevichStepan:
@romange PR description updated.

Also I added `blocksize` to the `FT.INFO` command.

Test is also added.

romange (dismissed):

PR #5425, opened 2025-07-06
test: search benchmark updated to use class fixture for multiple tests

vyavdoshenko:
- The benchmark was moved to the class
- Init state executes once for all test cases
- One simple test was added as an example


romange:
Great! Can you please also add `latency_tracking` to flags in instance.py:422 ?

vyavdoshenko:
> Can you please also add `latency_tracking` to flags in instance.py:422 ?

Done.

@romange 
Do you want to add this option by default for the benchmark?

romange:
> benchmark

I believe this change already does that for the benchmarks as well

romange (approved):

PR #5434, opened 2025-07-08
feat(config): implement CONFIG REWRITE command

### LINKED ISSUE #3787: Config rewrite support
#2511   
Seeing that someone previously suggested support for the config rewrite command, but didn't elaborate on the reasoning, I'd like to give my own personal opinion.

Recently, I've been organizing the scripts and data output related to dragonfly monitoring, and I've seen the official monitoring based on Prometheus and Grafana before, but I feel that the monitoring items are a bit less than enough to support the monitoring needs of the production environment, so I'd like to do more monitoring data extraction. This time there a  places I want to propose, let's communicate together to see if there is a need to improve.

-  Config rewrite command support
The reason for proposing this command support is that in the use of the production environment, we may hot change the instance configuration, for example, maxmemory and maxclient values, after the modification, I want to persist to the configuration file need to how to deal with? At present, it seems that we can only manually modify the configuration file to achieve consistency between the instance configuration and the parameters configured in the instance configuration file, but manual operation will inevitably have the possibility of error, there is no excuse for this. Another point is that if the instance supports the config rewrite command to meet the instance configuration and configuration file consistency, then I think the subsequent configuration changes can be added to the automated operation and maintenance process, whether it is a script or a platform operation, the addition of this command will make this operation will make the operation of the stated operation has become a simple silky smooth.
In addition, the instance configuration and configuration file inconsistency I feel there is a stability hazard, if the instance is from the library, only modified the instance config configuration, while the configuration file is not modified in the case of master-slave switching occurs, if the configuration is lower than the master library configuration, regardless of maxmemory or maxclient configuration, I feel that there is a possibility of production accidents.
Other configurations are the same, not to mention ...

We will treat configuration inconsistency as a case to assist the DBA in targeting the management of cached instances.
image
In this case, it is normally necessary to alert the DBA to warn them.

**Additional context**
I'm sorry I'm here to raise requirements again, hahaha, these requirements are trying to usability and reliability after using our dragonfly on-line to the production environment, I hope dragonfly is getting better and better, looking forward to the reply, together to discuss.  😃

boomballa:
## Summary

Implements the `CONFIG REWRITE` command to rewrite the configuration file with the current configuration values that have been modified at runtime.

**Fixes #3787**

## Key Features

- **Atomic File Operations**: Uses `mkstemp()` and `rename()` for safe file updates
- **Smart Configuration Tracking**: Tracks user-modified configurations in `set_by_user_` set  
- **Original Configuration Preservation**: Updates original config lines while preserving formatting
- **Generated Section Management**: Maintains separate section for runtime-only configurations
- **Absl Reflection Integration**: Leverages `absl::FindCommandLineFlag` for efficient flag access

## Problem Statement

As described in [#3787](https://github.com/dragonflydb/dragonfly/issues/3787), production environments require the ability to:
- Hot-modify instance configurations (e.g., `maxmemory`, `maxclient`)
- Persist runtime configuration changes to the configuration file
- Avoid manual configuration file editing which is error-prone
- Ensure configuration consistency between instance and config file
- Support automated operations and maintenance processes
- Prevent stability issues during master-slave switching scenarios

### ConfigRegistry Enhancement
- Added `Rewrite()` method with comprehensive config file parsing and rewriting logic
- Introduced `set_by_user_` set to track configurations modified via `CONFIG SET`
- Integrated absl reflection for efficient flag value retrieval

### Command Integration
- Added `CONFIG REWRITE` subcommand to existing CONFIG command handler
- Comprehensive error handling for missing config files and write failures
- Proper validation that server is running with a config file

### Code Optimization
- **Pure Additive Approach**: Preserves existing code stability by avoiding deletions
- **Redundancy Removal**: Cleaned up duplicate code and unused implementations
- **Absl Reflection Usage**: Replaced custom global variables with standard absl mechanisms

## Usage

```bash
CONFIG REWRITE
```

**Returns:**
- `OK` on successful config file rewrite
- `"The server is running without a config file"` if no config file specified
- `"Failed to rewrite config file"` if write operation fails

### File Structure Management
- Preserves original configuration file structure and comments
- Updates existing configuration lines in-place
- Manages "Generated by CONFIG REWRITE" section for runtime-only configs
- Handles both original configs and runtime-generated configs separately

### Thread Safety
- Proper mutex locking for concurrent access to configuration registry
- Safe handling of `set_by_user_` set across multiple threads

### Error Handling
- Validates config file existence and accessibility
- Implements atomic write operations to prevent corruption
- Comprehensive error reporting for various failure scenarios

## Production Benefits

This implementation addresses the key concerns raised in #3787:

1. **Automated Operations**: Enables seamless integration with automated operation and maintenance processes
2. **Configuration Consistency**: Ensures instance configuration matches the configuration file
3. **Stability Enhancement**: Prevents configuration-related issues during master-slave switching
4. **Error Reduction**: Eliminates manual configuration file editing errors
5. **Monitoring Support**: Facilitates configuration drift detection and alerting for DBAs

## Code Quality Improvements

- Removed redundant `config_file` special handling in `ConfigRegistry::Get`
- Eliminated duplicate code paths and unused global variables
- Enhanced maintainability through centralized configuration management
- Improved code consistency with existing patterns

## Testing Considerations

The implementation supports testing for:
- Successful config file rewriting with various configuration changes
- Error handling for missing or inaccessible config files
- Atomic operation safety during concurrent access
- Configuration value persistence across rewrites
- Backward compatibility with existing configuration files

## Files Modified

- `src/server/config_registry.h` - Added Rewrite method and tracking mechanism
- `src/server/config_registry.cc` - Core implementation of CONFIG REWRITE logic
- `src/server/server_family.cc` - Integration with CONFIG command handler

## Backward Compatibility

- Fully backward compatible with existing configuration files
- Preserves original configuration format and structure
- No breaking changes to existing CONFIG GET/SET functionality
- Maintains compatibility with all existing configuration options


src/server/config_registry.h
```diff
@@ -54,6 +56,10 @@ class ConfigRegistry {
 
   std::vector<std::string> List(std::string_view glob) const ABSL_LOCKS_EXCLUDED(mu_);
 
+  bool Rewrite() const;
+
+  std::set<std::string> set_by_user_;
```
romange: you do not need this map. Instead please add a boolean `is_modified` to the `Entry` at line 71.
In `RegisterInternal`  init `is_modified` to false, and inside `Set(...)` flip it to true

src/server/config_registry.cc
```diff
@@ -87,6 +97,136 @@ void ConfigRegistry::RegisterInternal(string_view config_name, bool is_mutable,
   CHECK(inserted) << "Duplicate config name: " << name;
 }
 
+bool ConfigRegistry::Rewrite() const {
```
romange: lets separate the config file generation from writing it to the file. Right now it's all mixed together and it's hard to follow.
Introduce another function `std::string GenerateConfig() const` that only generates the string contents you want to serialize. And then  write these contents to the file.

src/server/config_registry.h
boomballa: @romange  Thank you for the excellent suggestion! You're absolutely right about the architecture improvement.

**Changes made:**

1. **Removed** `std::set<std::string> set_by_user_` member from ConfigRegistry
2. **Added** `bool is_modified = false` field to the `Entry` struct (line 71)
3. **Updated** `Set()` method to use `it->second.is_modified = true` instead of `set_by_user_.insert(name)`
4. **Updated** `Rewrite()` method to iterate through `registry_` and collect modified entries using `entry.is_modified`

The implementation now follows the suggested pattern where `RegisterInternal` initializes `is_modified` to false, and `Set()` flips it to true when configuration is successfully modified.

src/server/config_registry.cc
boomballa: @romange Thank you for the suggestion! I've already implemented exactly what you recommended.

**Changes made:**
- Extracted `std::string GenerateConfig() const` method that purely handles config content generation
- Simplified `Rewrite()` method to focus only on atomic file operations
- The two responsibilities are now cleanly separated

The `GenerateConfig()` method handles all the logic for:
- Reading existing config file
- Parsing and updating configuration values
- Generating the new config string content

While `Rewrite()` now simply:
- Calls `GenerateConfig()` to get the content
- Performs atomic file write operations

This separation makes the code much more testable and easier to follow. Please take a look at the latest commit and let me know if this meets your expectations!

romange:
Are you using AI to write this PR? :)

src/server/config_registry.cc
```diff
@@ -5,10 +5,17 @@
...
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include <filesystem>
+#include <fstream>
 
 #include "base/logging.h"
 #include "core/glob_matcher.h"
 #include "server/common.h"
+#include "server/server_family.h"
```
romange: why do you need this include?

src/server/config_registry.cc
```diff
@@ -87,6 +97,167 @@ void ConfigRegistry::RegisterInternal(string_view config_name, bool is_mutable,
...
+    return "";
+  std::string config_file_path = flagfile_flag->CurrentValue();
+  if (config_file_path.empty())
+    return "";
+  std::filesystem::path config_path(config_file_path);
+  if (!std::filesystem::exists(config_path))
+    return "";
+  std::ifstream config_file(config_path);
+  if (!config_file.is_open())
+    return "";
```
romange: why do you need this here?

boomballa:
> Are you using AI to write this PR? :)

Frankly speaking, individual places are indeed with the help of AI to help realize, but the overall design ideas are my own, just some code specification and implementation, thinking is really not thorough enough, sorry.

src/server/config_registry.cc
boomballa: @romange   I've addressed both issues:

1. **Removed unused include** - Deleted `#include "server/server_family.h"` as it wasn't being used
2. **Improved method separation** - Moved file validation logic from `GenerateConfig()` to `Rewrite()`. Now `GenerateConfig(const std::string& config_file_path)` focuses purely on content generation and accepts the file path as a parameter.

The responsibilities are now cleanly separated. Please review the latest changes.

romange (changes requested):
I took a closer look at what you are trying to achieve.
I suspect that the whole code here is not needed. You do not need to parse the previous file to know what flags to write into the new file.

`absl::CommandLineFlag` has two access methods: `CurrentValue()`  and `DefaultValue()`. You can just iterate over all flags and write only those flags whose values differ from default.

The code should not be in config_registry as it's not related to config registry at all. You can do everything in server_family.cc

`absl::GetAllFlags()` allows you to fetch all the flags in the system

PR #5436, opened 2025-07-09
Feat: Added BZMPOP Command

### LINKED ISSUE #3877: implement BZMPOP

H4R5H1T-007:
Fixes #3877

### DIFF
--- a/src/server/zset_family.cc
+++ b/src/server/zset_family.cc
@@ -1819,11 +1819,22 @@ std::optional<std::string_view> GetFirstNonEmptyKeyFound(EngineShard* shard, Tra
   return std::nullopt;
 }
 
-// Validates the ZMPop command arguments and extracts the values to the output params.
+// Validates the ZMPop and BZMPop command arguments and extracts the values to the output params.
 // If the arguments are invalid sends the appropiate error to builder and returns false.
 bool ValidateZMPopCommand(CmdArgList args, uint32* num_keys, bool* is_max, int* pop_count,
-                          SinkReplyBuilder* builder) {
+                          SinkReplyBuilder* builder, bool is_blocking, float* timeout) {
   CmdArgParser parser{args};
+
+  if (is_blocking) {
+    if (!absl::SimpleAtof(parser.Next(), timeout)) {
+      builder->SendError("timeout is not a float or out of range");
+      return false;
+    }
+    if (*timeout < 0) {
+      builder->SendError("timeout is negative");
+      return false;
+    }
+  }
 
   if (!SimpleAtoi(parser.Next(), num_keys)) {
     builder->SendError(kUintErr);
@@ -2401,11 +2412,14 @@ void ZSetFamily::ZInterCard(CmdArgList args, const CommandContext& cmd_cntx) {
   builder->SendLong(result.value().size());
 }
 
-void ZSetFamily::ZMPop(CmdArgList args, const CommandContext& cmd_cntx) {
+// Generic function for ZMPop and BZMPop commands
+void ZMPopGeneric(CmdArgList args, const CommandContext& cmd_cntx, bool is_blocking) {
   uint32 num_keys;
   bool is_max;
   int pop_count;
-  if (!ValidateZMPopCommand(args, &num_keys, &is_max, &pop_count, cmd_cntx.rb)) {
+  float timeout;
+  if (!ValidateZMPopCommand(args, &num_keys, &is_max, &pop_count, cmd_cntx.rb, is_blocking,
+                            &timeout)) {
     return;
   }
   auto* response_builder = static_cast<RedisReplyBuilder*>(cmd_cntx.rb);
@@ -2439,7 +2453,8 @@ void ZSetFamily::ZMPop(CmdArgList args, const CommandContext& cmd_cntx) {
   // Now that we have the first non empty key from each shard, find the first overall first key and
   // pop elements from it.
   std::optional<std::string_view> key_to_pop = std::nullopt;
-  ArgRange arg_keys(args.subspan(1, num_keys));
+  // BZMPOP have 1 extra argument as compared to ZMPOP hence adding 1 is is_blocking is true
+  ArgRange arg_keys(args.subspan(1 + is_blocking, num_keys));
   // Find the first arg_key which exists in any shard and is not empty.
   for (std::string_view key : arg_keys) {
     if (first_found_keys_for_shard.contains(key)) {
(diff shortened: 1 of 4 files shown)


src/server/zset_family.cc
```diff
@@ -1819,12 +1819,23 @@ std::optional<std::string_view> GetFirstNonEmptyKeyFound(EngineShard* shard, Tra
   return std::nullopt;
 }
 
-// Validates the ZMPop command arguments and extracts the values to the output params.
+// Validates the ZMPop and BZMPop command arguments and extracts the values to the output params.
 // If the arguments are invalid sends the appropiate error to builder and returns false.
 bool ValidateZMPopCommand(CmdArgList args, uint32* num_keys, bool* is_max, int* pop_count,
```
romange: Ok this function has not a very nice interface. Please introduce `ValidateZMPopResult` consisting of num_keys, is_max, pop_count and timeout.
The interface should be something like this: `bool (CmdArgList args, bool is_blocking, SinkReplyBuilder* builder, ValidateZMPopResult* result)`

src/server/zset_family.cc
```diff
@@ -2448,6 +2463,49 @@ void ZSetFamily::ZMPop(CmdArgList args, const CommandContext& cmd_cntx) {
...
   }
 
+  // if we don't have any key to block and it's blocking then we will block it using `WaitOnWatch`
+  if (is_blocking && !key_to_pop.has_value()) {
+    auto trans = cmd_cntx.tx;
+    auto cntx = cmd_cntx.conn_cntx;
+    auto* ns = &trans->GetNamespace();
+
+    // Concluding transaction if it's multi mode as blocking is not allowed in it.
+    if (trans->IsMulti()) {
```
romange: I would unify this and the block at line 2509 before you handle the blocking case:
```
if (!key_to_pop && (!is_blocking || trans->IsMulti())) {
  ....
```

src/server/zset_family.cc
```diff
@@ -2783,6 +2850,9 @@ void ZSetFamily::Register(CommandRegistry* registry) {
              ZInterCard)
       << CI{"ZLEXCOUNT", CO::READONLY, 4, 1, 1, acl::kZLexCount}.HFUNC(ZLexCount)
       << CI{"ZMPOP", CO::SLOW | CO::WRITE | CO::VARIADIC_KEYS, -4, 2, 2, acl::kZMPop}.HFUNC(ZMPop)
+      << CI{"BZMPOP",    CO::SLOW | CO::WRITE | CO::VARIADIC_KEYS | CO::BLOCKING, -5, 3, 3,
```
romange: you also need `CO::NO_AUTOJOURNAL` to replicate this correctly. this also requires a emitting a custom journal command. Look at the other NO_AUTOJOURNAL commands.

src/server/zset_family.cc
```diff
@@ -1576,6 +1576,17 @@ OpResult<ScoredArray> ZPopMinMaxInternal(std::string_view key, FilterShards shou
   auto cb = [&](Transaction* t, EngineShard* shard) {
     if (!key_shard.has_value() || *key_shard == shard->shard_id()) {
       result = OpPopCount(range_spec, t->GetOpArgs(shard), key);
+      auto cid = t->GetCId();
+      // if command conatins no_autojournal and it's BZMPOP then we will write to journal manually.
+      if ((result == OpStatus::OK) && (cid->opt_mask() & CO::NO_AUTOJOURNAL) &&
+          (cid->name() == "BZMPOP")) {
+        OpArgs op_args = t->GetOpArgs(shard);
```
romange: I just checked and actually we have the same bug in ZMPOP.  Even with ZMPOP, there is no need to pass all the keys  because the outcome is already known: we need to fetch from `key`.
Therefore, I suggest to move `RecordJournal` into OpPopCount and rewrite both commands as `ZPOPMIN` or `ZPOPMAX` accordingly. Also please do not forget to mark `ZMPOP` as NO_AUTOJOURNAL.

Look at how it's done in `OpBZPop`. Thanks!

romange (commented):
Almost there. I appreciate your effort!

src/server/zset_family.cc
```diff
@@ -1819,39 +1830,50 @@ std::optional<std::string_view> GetFirstNonEmptyKeyFound(EngineShard* shard, Tra
   return std::nullopt;
 }
 
-// Validates the ZMPop command arguments and extracts the values to the output params.
+// Validates the ZMPop and BZMPop command arguments and extracts the values to the output params.
 // If the arguments are invalid sends the appropiate error to builder and returns false.
-bool ValidateZMPopCommand(CmdArgList args, uint32* num_keys, bool* is_max, int* pop_count,
-                          SinkReplyBuilder* builder) {
+bool ValidateZMPopCommand(CmdArgList args, ZSetFamily::ValidateZMPopResult* result,
+                          SinkReplyBuilder* builder, bool is_blocking) {
```
romange: nit: we put output arguments at the end, so `is_blocking` should be second here

src/server/zset_family.h
```diff
@@ -66,6 +67,13 @@ class ZSetFamily {
     bool override = false;
   };
 
+  struct ValidateZMPopResult {
```
romange: you do not need this struct to be defined in the header file. please move it to the cc file.

src/server/zset_family.h
```diff
@@ -66,6 +67,13 @@ class ZSetFamily {
...
+    uint32* num_keys;
```
romange: no need to pass pointers. you can just define:
```
  struct ValidateZMPopResult {
    uint32_t num_keys;
    bool is_max;
    int pop_count;
    float timeout;
  };
```
and everything will be much simpler in the code.

src/server/zset_family.h
```diff
@@ -7,6 +7,7 @@
 #include <string_view>
 #include <variant>
 
+#include "base/integral_types.h"
```
romange: not needed

tests/dragonfly/replication_test.py
```diff
@@ -746,6 +746,11 @@ async def check_expire(key):
             [r"XTRIM k-stream MINID 0", r"SREM k-one-element-set value[12]"],
         )
 
+        # check BZMPOP turns into ZMPOP command
+        await c_master.zadd("key", {"a": 1, "b": 2, "c": 3})
+        await skip_cmd()
+        await check("BZMPOP 0 3 key3 key2 key MAX COUNT 3", r"ZMPOP 1 key MAX COUNT 3")
```
romange: very nice, thank you! but it should turn as ZPOPMAX/ZPOPMIN :)

src/server/zset_family.cc
H4R5H1T-007: Yup I also thought that but then I thought to make it consistent as ZMPOP so I added it like ZMPOP journal. Cool I will move journaling to OpPopCount and add custom journal to ZMPOP as well that will be much better.

tests/dragonfly/replication_test.py
H4R5H1T-007: Sure I will change it

src/server/zset_family.cc
```diff
@@ -2401,11 +2430,10 @@ void ZSetFamily::ZInterCard(CmdArgList args, const CommandContext& cmd_cntx) {
...
 }
 
-void ZSetFamily::ZMPop(CmdArgList args, const CommandContext& cmd_cntx) {
-  uint32 num_keys;
-  bool is_max;
-  int pop_count;
-  if (!ValidateZMPopCommand(args, &num_keys, &is_max, &pop_count, cmd_cntx.rb)) {
+// Generic function for ZMPop and BZMPop commands
+void ZMPopGeneric(CmdArgList args, const CommandContext& cmd_cntx, bool is_blocking) {
+  ValidateZMPopResult zmpopArgs;
```
romange: naming: zmpop_args

src/server/zset_family.cc
```diff
@@ -1819,39 +1837,50 @@ std::optional<std::string_view> GetFirstNonEmptyKeyFound(EngineShard* shard, Tra
...
     builder->SendError(kSyntaxErr);
     return false;
   }
 
-  *pop_count = 1;
+  result->pop_count = 1;
   // Check if we have additional COUNT argument.
   if (parser.HasNext()) {
-    if (!parser.Check("COUNT", pop_count)) {
+    if (!parser.Check("COUNT", &(result->pop_count))) {
```
romange:
```suggestion
    if (!parser.Check("COUNT", &result->pop_count)) {
```

src/server/zset_family.cc
```diff
@@ -1819,39 +1837,50 @@ std::optional<std::string_view> GetFirstNonEmptyKeyFound(EngineShard* shard, Tra
...
 // If the arguments are invalid sends the appropiate error to builder and returns false.
-bool ValidateZMPopCommand(CmdArgList args, uint32* num_keys, bool* is_max, int* pop_count,
-                          SinkReplyBuilder* builder) {
+bool ValidateZMPopCommand(CmdArgList args, bool is_blocking, SinkReplyBuilder* builder,
+                          ValidateZMPopResult* result) {
   CmdArgParser parser{args};
 
-  if (!SimpleAtoi(parser.Next(), num_keys)) {
+  if (is_blocking) {
+    if (!absl::SimpleAtof(parser.Next(), &(result->timeout))) {
```
romange:
```suggestion
    if (!absl::SimpleAtof(parser.Next(), &result->timeout)) {
```

romange (commented):
minor comments

romange:
Thanks!

PR #5454, opened 2025-07-09
fix: acl log missing return after reply

kostasrim:
When `acl log` is called with more than one argument we replied twice because the error branch did not return.

* fix replying twice when acl log arguments overflow
* add test
* move registering acl functions at the end of the file

### DIFF
--- a/src/server/acl/acl_family.cc
+++ b/src/server/acl/acl_family.cc
@@ -378,11 +378,11 @@ void AclFamily::Load(CmdArgList args, const CommandContext& cmd_cntx) {
 void AclFamily::Log(CmdArgList args, const CommandContext& cmd_cntx) {
   auto* rb = static_cast<facade::RedisReplyBuilder*>(cmd_cntx.rb);
   if (args.size() > 1) {
-    rb->SendError(facade::OpStatus::OUT_OF_RANGE);
+    return rb->SendError(facade::OpStatus::OUT_OF_RANGE);
   }
 
   size_t max_output = 10;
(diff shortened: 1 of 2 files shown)


src/server/acl/acl_family.cc
```diff
@@ -1213,4 +1158,59 @@ void AclFamily::BuildIndexers(RevCommandsIndexStore families) {
...
+  *registry << CI{"ACL GENPASS", CO::NOSCRIPT | CO::LOADING, -1, 0, 0, acl::kGenPass}.HFUNC(
+      GenPass);
+  cmd_registry_ = registry;
+
+  // build indexers
+  BuildIndexers(cmd_registry_->GetFamilies());
+}
+
+#undef HFUNC
+
```
kostasrim, line 1215: No functional change here. The code was just moved at the bottom of the file

src/server/acl/acl_family.cc
```diff
@@ -378,11 +378,11 @@ void AclFamily::Load(CmdArgList args, const CommandContext& cmd_cntx) {
 void AclFamily::Log(CmdArgList args, const CommandContext& cmd_cntx) {
   auto* rb = static_cast<facade::RedisReplyBuilder*>(cmd_cntx.rb);
   if (args.size() > 1) {
-    rb->SendError(facade::OpStatus::OUT_OF_RANGE);
+    return rb->SendError(facade::OpStatus::OUT_OF_RANGE);
```
kostasrim, line 381: This was the bug. We would reply twice otherwise

PR #5471, opened 2025-07-10
feat(config): Implement CONFIG REWRITE command

boomballa:
## Summary
Implements CONFIG REWRITE command using absl flags reflection to preserve original configuration file layout while updating runtime-modified settings.

Fixes https://github.com/dragonflydb/dragonfly/issues/3787

## Key Features
- **Layout Preservation**: Maintains original config file structure, comments, and formatting  
- **Atomic Operations**: Uses `mkstemp()` and `rename()` for safe file updates
- **Absl Integration**: Uses `absl::GetAllFlags()` + `CurrentValue() vs DefaultValue()` comparison

## Implementation
- Pure additive approach in `server_family.cc` (129 lines added)
- Line-by-line parsing with in-place flag updates
- Intelligent "# Generated by CONFIG REWRITE" section management
- Runtime configuration filtering via `ConfigRegistry.List()`

## Usage
```bash
10.142.220.27:7379> CONFIG SET timeout 60
OK
10.142.220.27:7379> CONFIG REWRITE  
OK
```

**Result**: Original config preserved + new section for runtime changes

## Bug Fixes
- ✅ Preserves original config file layout and comments
- ✅ Filters out startup-only configurations  
- ✅ Prevents duplicate blank lines on subsequent rewrites
- ✅ Fixes type conversion issues (`std::string_view` → `std::string`)

## Files Modified
- `src/server/server_family.cc` - Complete CONFIG REWRITE implementation


boomballa:
@romange Hi Roman, 
The previous branch got a bit messy, so I started a new clean branch and reimplemented the `CONFIG REWRITE` functionality following your suggestions. This time I only modified src/server/server_family.cc and implemented the entire CONFIG REWRITE feature within this single file as requested.

If there are any areas that need improvement, please feel free to point them out directly.
Thanks for the guidance!

src/server/server_family.cc
```diff
@@ -2183,6 +2190,143 @@ void ServerFamily::Config(CmdArgList args, const CommandContext& cmd_cntx) {
...
+  if (sub_cmd == "REWRITE") {
+    absl::CommandLineFlag* flagfile_flag = absl::FindCommandLineFlag("flagfile");
+    if (!flagfile_flag || flagfile_flag->CurrentValue().empty()) {
+      return builder->SendError("The server is running without a config file");
+    }
+
+    std::string config_file_path = flagfile_flag->CurrentValue();
+
+    // Read original config file
+    std::ifstream file(config_file_path);
```
romange: why do you still need to read the original file? what is needed from there?

src/server/server_family.cc
```diff
@@ -2183,6 +2190,143 @@ void ServerFamily::Config(CmdArgList args, const CommandContext& cmd_cntx) {
     return rb->SendBulkStrArr(res, RedisReplyBuilder::MAP);
   }
 
+  if (sub_cmd == "REWRITE") {
+    absl::CommandLineFlag* flagfile_flag = absl::FindCommandLineFlag("flagfile");
```
romange: please factor this block out into a function and add right after ReadProcStats().

src/server/server_family.cc
boomballa: My idea is mainly because: 

1.  Layout Preservation 
```

std::ifstream file(config_file_path);
while (std::getline(file, line)) {
  // Reads line by line, maintains original structure
  updated_lines.push_back(line);  // Preserve comments, blank lines, formatting
}
```
- Preservation of comments: # This is important config and other important instructions in the configuration file 
- Preservation of formatting: blank lines, indentation, alignment 
- Maintenance of readability: the configuration file is still easy to read when the team is working together


2. Incremental Updates Configuration file 
```
if (!trimmed.empty() && trimmed[0] == '-' && trimmed[1] == '-') {
  std::string flag_name = trimmed.substr(2, eq_pos - 2);
  if (current_flags.count(flag_name)) {
    // In-situ updating of existing configurations
    updated_lines.push_back(absl::StrCat("--", flag_name, "=", current_flags[flag_name]));
  } else {
    // Keep the original configuration unchanged
    updated_lines.push_back(line);
  }
}
```
- In-place updates: only changed configuration values are modified 
- Maintains location: the original location of the configuration item in the file remains unchanged 
- Avoids duplication: prevents the same configuration from appearing more than once

3. Change Detection 
```
std::unordered_set<std::string> existing_flags;
// ... Marking existing configurations during reads
existing_flags.insert(flag_name);

// Subsequent identification of new configurations
for (const auto& [flag_name, flag_value] : current_flags) {
  if (existing_flags.find(flag_name) == existing_flags.end()) {
    new_flags.push_back(absl::StrCat("--", flag_name, "=", flag_value));
  }
}
```
- Distinguish between old and new: Identify what is the original configuration and what is new 
- Avoid duplication: Prevent existing configurations from reappearing in the new area 
- Classification: Original configuration is updated in place and new configurations are added uniformly


```
# Upper half: manual configuration area
...
...
...
# Password for authentication with master. default: ""
# --masterauth=xxxxxx
# Password for AUTH authentication.
# --requirepass=xxxxxxx

# Lower half: Runtime configuration area
# Generated by CONFIG REWRITE
--timeout=60
--slowlog_log_slower_than=20000
--rss_oom_deny_ratio=1.5
```

Ideas for Separating Configurations 
1. Configuration Source Tracing
2. Audit Compliance
3. Configuring version control
```
git diff dragonfly.cnf
```

That's about all I can think of.

src/server/server_family.cc
boomballa: Okay, I'll make that suggestion later.

boomballa: Hi @romange    👋  
Extracted the code block into a new function ( `RewriteConfigFile()`) and placed it immediately after `ReadProcStats()`. 
Would appreciate your review when you have time! Let me know if any adjustments are needed.

src/server/server_family.cc
boomballa:
```
# If true, save in dragonfly-specific snapshotting format. default: true
--df_snapshot_format=true
# The filename to save/load the DB. default: "dump-{timestamp}"
--dbfilename=dragonfly7379-dump
# Cron expression for the time to save a snapshot, crontab style.
# --snapshot_cron


# Generated by CONFIG REWRITE
--timeout=80
--slowlog_log_slower_than=15000
--rss_oom_deny_ratio=1.5
```

The effect is roughly the same, I went back and thought about your question and I think your perspective is that since most of the configuration files for the environment on the cloud are not of concern to the user ? It just requires the user to automate the actions on the white screen of the configuration page ? send So there is no need to be concerned about e.g. the retention of comments, the separate display of new configuration changes (not mixed in with the old configuration).

But I think this is still necessary regardless for users using dfly in non-cloud environments or for auditing needs. I don't know if you agree with me on this design point.  :smiley:

romange:
thanks, I will take a look next week.

src/server/server_family.cc
```diff
@@ -679,6 +685,144 @@ bool ReadProcStats(io::StatusData* sdata) {
...
+    for (const auto& new_flag : new_flags) {
+      final_content += new_flag + "\n";
+    }
+  }
+
+  // Atomic write using mkstemp + rename
+  std::string tmp_template = config_file_path + ".tmpXXXXXX";
+  std::vector<char> tmp_path(tmp_template.begin(), tmp_template.end());
+  tmp_path.push_back('\0');
+  int fd = mkstemp(tmp_path.data());
```
romange:
```suggestion
  int fd = mkstemp(tmp_template.data());  
```

boomballa: @romange  Thanks for the suggestion, but `mkstemp() `needs modifiable strings because it modifies the template string to generate the actual filename. Using `c_str()` passes read-only pointers, resulting in undefined behavior. And the subsequent `rename()` and `unlink()` also need to use the actual modified filename. Can I just submit the code as it is currently?

romange: ok,then `int fd = mkstemp(tmp_template.data());`

romange: you do not need `tmp_path` at all

boomballa: > you do not need `tmp_path` at all

@romange Technically, “tmp_template.data()” does work.

However, I prefer the current approach for several reasons:
1. **clear intent**: creating a separate `tmp_path` makes it clear that we need a modifiable buffer.
2. **Standard compatibility**: no dependency on C++17 string::data() behavior
3. **Buffer safety**: explicitly ensures null termination and sufficient space
4. **Null-termination safety**: Explicitly ensures proper string termination

Both of these approaches work, but I personally feel that the current approach is more explicit about what we're doing. If it really doesn't conform to the project's code specification, or if you would like me to change it, I will do so. Do you see the current implementation as acceptable?

romange: `tmp_template` is always null terminated and our project depends on c++17,
and you can not compile it without it. Yes, I do not like redundant code and it confused me to see that you manually add `\0`. So yes, please remove redundant code 🙏🏼

boomballa: 😄  I'm sorry for the trouble, I understand when you explain it this way, have a good day at work! 👍

romange (approved):

romange:
@boomballa  Thank you very much for your contributions!

boomballa:
> @boomballa Thank you very much for your contributions!

@romange  Glad to be able to contribute and thanks for the guidance! :thumbsup:

Issue #5472, opened 2025-07-10
First FT.SEARCH query dramatically slower that the same second query


vyavdoshenko:
The preparation part can be taken from the initialization of `tests/dragonfly/search_benchmark_test.py` to create the index and fill the data.

**The problem description:**
The slowest part is the SearchAlgorithm::Init function, which creates a ParseQuery.

It takes **205657us** for the first time, and it takes **244us** the second time for the same query. 

The warm-up function with a dummy query to warm up all the internal structures was added.
The result: First call: **352us**, Second call: **257us**

Essentially, ParseQuery calls the generated code. What can be done to speed it up?

Additional numbers:

Stats **without** warm-up:
_First call:_
- Commandstats
cmdstat_ft.search:calls=1,usec=365644,usec_per_call=365644
- Latencystats
latency_percentiles_usec_ft.search:p50=366591,p99=366591,p99.9=366591

_Second call:_
- Commandstats
cmdstat_ft.search:calls=2,usec=537342,usec_per_call=268671
- Latencystats
latency_percentiles_usec_ft.search:p50=172031,p99=366591,p99.9=366591

Stats **with** warm-up:
_First call:_
- Latencystats
latency_percentiles_usec_ft.search:p50=162815,p99=162815,p99.9=162815
- Commandstats
cmdstat_ft.search:calls=1,usec=162622,usec_per_call=162622

_Second call:_
- Latencystats
latency_percentiles_usec_ft.search:p50=162815,p99=165887,p99.9=165887
- Commandstats
cmdstat_ft.search:calls=2,usec=327486,usec_per_call=163743

The tested query:
```
[...]

PR #5473, opened 2025-07-10
fix: FT.CREATE first-query initialization penalty (slowdown)

vyavdoshenko:
Fixes: https://github.com/dragonflydb/dragonfly/issues/5472
Added warmup parser function in FtCreate.
It executes only once per process execution, once first FT.CREATE executed.


vyavdoshenko:
Reproduces https://github.com/dragonflydb/dragonfly/issues/5472

src/core/search/search_parser_performance_test.cc
```diff
@@ -0,0 +1,127 @@
...
+
+  const int num_iterations = 5;
+  std::vector<int64_t> parse_times_us;
+  parse_times_us.reserve(num_iterations);
+
+  // Run the same query multiple times and measure each execution
+  for (int i = 0; i < num_iterations; i++) {
+    absl::Time start = absl::Now();
+
+    SearchAlgorithm search_algo;
```
romange: I think, you can take it one level lower, see the following function:
```
AstExpr ParseQuery(std::string_view query, const QueryParams* params) {
  QueryDriver driver{};
  driver.ResetScanner();
  driver.SetParams(params);
  driver.SetInput(std::string{query});
  (void)Parser (&driver)();  // can throw
  return driver.Take();
}
```

so why not call `Parser/QueryDriver` directly? Also once you do that, I am curious what is the minimum code needed to call for latency go down.

romange: basically what exactly is needed to prewarm? is it parser? querydriver? both?
should if we call `(void)Parser (&driver)();  `  on empty query, will it also prewarm?

BagritsevichStepan: I believe the main issue is in our lexer, and there are opportunities to optimize it (just pasted the code into chatgpt and already identified several potential improvements). Additionally, we know that for the customer's traffic, the lexer accounts for the majority of processing time. So, maybe we need to fix it

vyavdoshenko: @romange, an empty query is enough.
The numbers if I add the first empty query call:

```
Iteration 1: SearchAlgorithm::Init took 31 μs [FIRST - SLOW]
Iteration 2: SearchAlgorithm::Init took 56 μs [SUBSEQUENT - FAST]
Iteration 3: SearchAlgorithm::Init took 53 μs [SUBSEQUENT - FAST]
Iteration 4: SearchAlgorithm::Init took 53 μs [SUBSEQUENT - FAST]
Iteration 5: SearchAlgorithm::Init took 55 μs [SUBSEQUENT - FAST]
Performance analysis:  
First query time:     31 μs
Min subsequent time:  53 μs
Avg subsequent time:  54 μs
Performance ratio:    0.6x
Difference:           -23 μs
Performance is consistent between first and subsequent queries
```

vyavdoshenko: I guess the minimal code to warm up is:

```
  QueryParams empty_params;
  search::SearchAlgorithm search_algo;
  search_algo.Init("", empty_params);
```

It can be executed in `SearchFamily::Register`

vyavdoshenko: One more approach:

```
    QueryParams params;
    QueryDriver driver{};
    driver.SetParams(&params);
    driver.SetInput(std::string{""});
    (void)Parser (&driver)();  
```

src/core/search/search_parser_performance_test.cc
```diff
@@ -0,0 +1,132 @@
...
+// Performance test to reproduce the first-query initialization penalty issue.
+// The issue can be reproduced only by running standalone, because the slow down is only visible
+// when running the test standalone. The slowest query is the first one per process, and the
+// subsequent ones are much faster.
+TEST(SearchParserPerformanceTest, FirstQueryInitializationPenalty) {
+  QueryParams params;  // No extra params needed for this query
+
+  // Real production query that causes the issue (250 RETURN fields)
+  std::string production_query =
+      "* RETURN 250 lv_gwlurzpyxpsnv lv_urruguhbk lv_jagzzrtdeoo lv_unzcq lv_sstmwej "
```
dranikpg: the real query is only the first char, everything else are _options_ that are parsed separately

src/core/search/search_parser_performance_test.cc
romange: @vyavdoshenko  I do not think we need this test to submit.
Instead I suggest to introduce a simple fix Dragongfly to eliminate this slow start effect.

Inside FT.CREATE we should create this dummy code or the smallest variation of it:
```
   QueryParams params;
    QueryDriver driver{};
    driver.SetParams(&params);
    driver.SetInput(std::string{""});
    (void)Parser (&driver)();  
```
 
 and just comment why we have it - that it eliminates the slow start of Parser in FT.SEARCH calls.

cc -@dranikpg

src/server/search/search_family.cc
```diff
@@ -619,9 +620,23 @@ void SearchReply(const SearchParams& params,
   }
 }
 
+// Warms up the query parser to avoid first-call slowness
+void WarmupQueryParser() {
+  static std::once_flag warmed_up;
+  std::call_once(warmed_up, []() {
+    search::QueryParams params;
+    search::QueryDriver driver{};
+    driver.SetParams(&params);
```
romange, line 629: do we need ` driver.SetParams` ? will it work without this?

src/server/search/search_family.cc
```diff
@@ -619,9 +620,23 @@ void SearchReply(const SearchParams& params,
...
+    driver.SetInput(std::string{""});
+    (void)search::Parser (&driver)();
```
romange, line 631: nit: while this work, I find it less readable.

romange (commented):
lgtm. did you verify it solves the issue?

src/server/search/search_family.cc
vyavdoshenko: No, I tried to skip it. The parser hangs.

src/server/search/search_family.cc
vyavdoshenko: The shorter version is:
```
QueryParams empty_params;
search::SearchAlgorithm search_algo;
search_algo.Init("", empty_params);
```
Functionally, it's the same.

vyavdoshenko:
> lgtm. did you verify it solves the issue?

Yes, I verified this approach.

romange (approved):

src/server/search/search_family.cc
```diff
@@ -619,9 +620,23 @@ void SearchReply(const SearchParams& params,
   }
 }
 
+// Warms up the query parser to avoid first-call slowness
+void WarmupQueryParser() {
+  static std::once_flag warmed_up;
+  std::call_once(warmed_up, []() {
```
dranikpg, line 626: Same as just a static

PR #5476, opened 2025-07-11
Bug: compr_method not assigned from other

C-mmon:
compr_method_ is not assigned from other. This is a definite bug. The *this object will retain its old compr_method_ or default to 0 if not set, instead of inheriting it from the moved-from object.


vyavdoshenko:
Stale PR

PR #5478, opened 2025-07-12
fix: add a sleep to relax slot migration process

romange:
The problem: when slot migration runs, it takes up majority of scheduling time and throttles io connections greatly. As a result, the write traffic throughput goes down almost to 0.

The solution: it's an intermediate solution that adds a tunable sleep to the migrations fiber.


BorysTheDev (dismissed):

kostasrim (commented):
nice!

ashotland (commented):
If we make it modifiable by config command it will be easier to modify all cluster nodes via apiserver

kostasrim (approved):
lgtm

kostasrim (approved):

PR #5486, opened 2025-07-12
chore(search): Simplify return field logic

dranikpg:
1. Remove effectively dead code:
  * `SearchField` never stores generated fields, so `string_view` is enough instead of StringOrView + View() + etc...
  * `SearchField::is_short_name` is always true
  * Remove `SearchField`'s dead functions
  * `DocumentAccessor::SerializeDocument` replaced `Serialize()`, but the latter was never removed. Remove and swap names back
 2. Add `ReturnOptionJson` that extensively tests RETURN with json
 3. Remove quotes for returning plain json fields (like RS)

### DIFF
--- a/src/server/search/doc_accessors.cc
+++ b/src/server/search/doc_accessors.cc
@@ -76,7 +76,7 @@ FieldValue ExtractSortableValueFromJson(const search::Schema& schema, string_vie
   if (json.is_null()) {
     return std::monostate{};
   }
-  auto json_as_string = json.to_string();
+  auto json_as_string = json.as_string();
   return ExtractSortableValue(schema, key, json_as_string);
 }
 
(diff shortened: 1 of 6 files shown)


src/server/search/doc_accessors.cc
```diff
@@ -76,7 +76,7 @@ FieldValue ExtractSortableValueFromJson(const search::Schema& schema, string_vie
   if (json.is_null()) {
     return std::monostate{};
   }
-  auto json_as_string = json.to_string();
+  auto json_as_string = json.as_string();
```
BagritsevichStepan, line 79: I wanted to create a wrapper for the json class to avoid mixing up `to_string` and `as_string`, since it’s not clear that they return different string formats (and we also call them in another place). But it’s okay.

PR #5489, opened 2025-07-13
chore: add more logs during replication

BorysTheDev:
Problem: During replication, we saw that the master disconnects from the replica. We don't have enough logs to understand the reason.

Fix: Added logs to 3 possible places where disconnecting can be triggered from.


src/facade/dragonfly_connection.cc
```diff
@@ -1207,6 +1207,10 @@ Connection::ParserStatus Connection::ParseRedis(unsigned max_busy_cycles) {
 
       DispatchSingle(has_more, dispatch_sync, dispatch_async);
     }
+    if (result != RedisParser::OK && result != RedisParser::INPUT_PENDING) {
```
romange, line 1210: Please add a comment that we do not expect that a replica sends an invalid command so we log if it happens.

romange (commented):
lgtm + minor comments

PR #5491, opened 2025-07-13
chore(set_family): Fast path SPOP with small count

### LINKED ISSUE #5487: SPOP is extremely slow for huge sets
To reproduce: ./dragonfly --proactor_threads=1
1. add 1000000 items to a set `foo`
2. run `redis-cli SPOP foo 5` in a loop and see how dragonfly saturates its CPU

The reason: we uniformly select 5 indices in [0, 1M] and then use an `O(N)` iteration to advance over the whole set to select 5 items. 

It's not possible to fix this without the cooperation from dense_set interface
What we need to do is a different (additional) strategy when `count` is much smaller than the set size.
Say when `count * 3 < set size`.

In that case we would want to call 5 times  `DenseSet::GetRandomIterator`  which should provide a random dereferenceable place in O(1).  We do not have such a function but it's easy to implement one.
DenseSet is an array of linked lists so we could choose a random index [0, N] and then call Advance to reach a busy place.

dranikpg:
Fixes #5487

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -26,7 +26,7 @@ add_library(dfly_core allocation_tracker.cc bloom.cc compact_object.cc dense_set
     segment_allocator.cc score_map.cc small_string.cc sorted_map.cc task_queue.cc
     tx_queue.cc string_set.cc string_map.cc top_keys.cc detail/bitpacking.cc)
 
-cxx_link(dfly_core base absl::flat_hash_map absl::str_format redis_lib TRDP::lua lua_modules
+cxx_link(dfly_core base absl::flat_hash_map absl::str_format absl::random_random redis_lib TRDP::lua lua_modules
     fibers2 ${SEARCH_LIB} jsonpath OpenSSL::Crypto TRDP::dconv TRDP::lz4)
 
 add_executable(dash_bench dash_bench.cc)


dranikpg:
A few notes on SPOP/SRANDOMEMBER/HRANDFIELD etc:
1. Maps and Sets use totally different code paths on different levels of abstraction. StringMap has `RandomPairsUnique`, sets use `RandMemberStrSet` which relies on external iteration
2. SPOP acts in two steps: First it selects members with `RandMemberStrSet`, then it erases them. Makes it a full common path for HRANDFIELD, but is not that efficient because of double lookups

src/core/dense_set.cc
```diff
@@ -670,22 +670,28 @@ void DenseSet::Delete(DensePtr* prev, DensePtr* ptr) {
...
-    if (bucket_iter == entries_.end()) {
-      return nullptr;
-    }
+  return IteratorBase{(DenseSet*)this, chain_it, &*chain_it};
+}
 
-    ExpireIfNeeded(nullptr, &(*bucket_iter));
-  } while (bucket_iter->IsEmpty());
+void* DenseSet::PopInternal() {
+  auto bucket_iter = GetRandomChain();  // Find first non empty chain
```
dranikpg, line 704: half of the code uses `bucket`, the other half uses `chain`... given it's a `ChainVectorIterator` I sticked with chain naming

src/server/set_family.cc
```diff
@@ -265,9 +269,35 @@ void InterStrSet(const DbContext& db_context, const vector<SetType>& vec, String
   }
 }
 
+template <typename C = absl::flat_hash_set<string>>
+StringVec RandMemberStrSetPicky(StringSet* strset, PicksGenerator& generator, size_t count) {
+  C picks;
+  picks.reserve(count);
+
+  size_t tries = 0;
+  while (picks.size() < count && tries++ < count * 2) {
```
dranikpg: Early exit from the loop because of two problems:
1. We can only pick so much entries as there are buckets. We can add some noise bits into the offset field so we follow long chains some steps ahead, but it won't help in extreme cases as well as it will have O(n^2) complexity 🤷🏻‍♂️ Once we change this code to Pop() while iterating it will resolve this problem
2. If 4/5 of the entries expired we just won't collect enough picks. We can't know this ahead

src/server/set_family.cc
```diff
@@ -265,9 +269,35 @@ void InterStrSet(const DbContext& db_context, const vector<SetType>& vec, String
...
+}
+
 StringVec RandMemberStrSet(const DbContext& db_context, const CompactObj& co,
-                           PicksGenerator& generator, std::size_t picks_count) {
+                           PicksGenerator& generator, size_t picks_count) {
   CHECK(IsDenseEncoding(co));
+  StringSetWrapper strset{co, db_context};
+
+  // If the set is small, extract entries with StringSet::GetRandomMember
+  if (picks_count * 5 < strset->UpperBoundSize()) {
```
dranikpg, line 292: 🤷🏻‍♂️ I can make a benchmark to determine the optimal cutoff (Valkey uses the same constant). My main concern is my comment above about breaking. It also depends more on the backing size, not only the entry count

src/core/string_set.cc
```diff
@@ -112,6 +112,10 @@ unsigned StringSet::AddBatch(absl::Span<std::string_view> span, uint32_t ttl_sec
   return res;
 }
 
+StringSet::iterator StringSet::GetRandomMember(size_t search_offset) {
+  return iterator{DenseSet::GetRandomIterator(search_offset)};
```
romange: you could use a reservoir sampling. i.e. while the iterator is a link decide with 1/k probability whether to progress or break. this would cover items in the links as well. i.e. at first with 1/2 to progress, then 1/3 etc.

src/server/set_family.cc
romange: see my comment above how to solve (1) with almost O(1) complexity.

dranikpg: > We can add some noise bits into the offset field so we follow long chains some steps ahead

This is it, i.e. bit is we keep or follow. I'll add internal randomization instead

src/server/set_family.cc
```diff
@@ -294,9 +292,9 @@ StringVec RandMemberStrSet(const DbContext& db_context, const CompactObj& co,
   if (picks_count * 5 < strset->UpperBoundSize()) {
     StringSet* ss(strset);
     if (bool unique = (dynamic_cast<UniquePicksGenerator*>(&generator) != nullptr); unique)
-      return RandMemberStrSetPicky(ss, generator, picks_count);
+      return RandMemberStrSetPicky(ss, picks_count);
     else
-      return RandMemberStrSetPicky<StringVec>(ss, generator, picks_count);
+      return RandMemberStrSetPicky<StringVec>(ss, picks_count);
```
romange, line 297: btw, we iterate in the loop below even if we filled result

dranikpg: Not sure I understand 🤨

romange: lines 309-315. we iterate over the set to completion even if results has all the picks. Does not really changes the big `O` but can burn CPU for fun

dranikpg: I return from `RandMemberStrSetPicky` directly in both cases. The code you're referring to is part of the O(N) picks algorithm

PR #5492, opened 2025-07-14
chore: add tls_socket accept error metric

romange:
This error will help exposing tls connections storm when those happen.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -727,6 +727,7 @@ void Connection::OnConnectionStart() {
 
 void Connection::HandleRequests() {
   VLOG(1) << "[" << id_ << "] HandleRequests";
+  DCHECK(stats_);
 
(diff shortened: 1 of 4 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -754,6 +757,8 @@ void Connection::HandleRequests() {
       std::ignore =
           socket_->Write(io::Buffer("-ERR Bad TLS header, double check "
                                     "if you enabled TLS for your client.\r\n"));
+      stats_->tls_accept_disconnects++;
```
kostasrim, line 760: From my understanding the handshake might have not started. Either way this is good for now - I will see if it can be improved.

kostasrim (approved):

src/facade/dragonfly_connection.cc
romange: yeah, it's not specific to  handshake. We have throttled logs to deep dive further. the assumption is that this branch is never taken in prod.

PR #5493, opened 2025-07-14
chore: add tls handshake statistics

kostasrim:
* add handshake statistics
* expose that in metrics

### DIFF
--- a/src/facade/tls_helpers.cc
+++ b/src/facade/tls_helpers.cc
@@ -14,6 +14,7 @@
 #include "absl/functional/bind_front.h"
 #include "base/flags.h"
 #include "base/logging.h"
+#include "facade/facade_types.h"
 
 ABSL_FLAG(std::string, tls_cert_file, "", "cert file for tls connections");
 ABSL_FLAG(std::string, tls_key_file, "", "key file for tls connections");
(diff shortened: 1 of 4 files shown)


src/facade/tls_helpers.cc
```diff
@@ -106,6 +107,18 @@ SSL_CTX* CreateSslCntx(TlsContextRole role) {
     SSL_CTX_set_timeout(ctx, GetFlag(FLAGS_tls_session_cache_timeout));
     SSL_CTX_set_session_id_context(ctx, (const unsigned char*)"dragonfly", 9);
   }
+
+  SSL_CTX_set_info_callback(ctx, [](const SSL* ssl, int where, int ret) {
+    // When we skip the handshake we never reach this state.
```
kostasrim, line 112: This should be used with `total_conncetions` to make sense. We will never call `set_info` if the `handshake is skipped` because of reused `session/ticket`. If we call `HANDSHAKE_START` or `HANDSHAKE_DONE` it means we are for sure doing a handshake. 

This will miss connections that just connected and they did not yet performed a handshake. 

This metric is eventually consistent (when all open connections pass or skip the handshake state).

kostasrim: also 

> But absence of callbacks alone is not enough because the handshake might never have been triggered or the connection was closed prematurely. 

But we have that info now 1) we know when we started a handshake 2) we know when we failed because of https://github.com/dragonflydb/dragonfly/pull/5492

romange (dismissed):

romange (approved):

PR #5488, opened 2025-07-13
[WIP] feat: mem defrag stats

abhijat:
WIP, collect page usage statistics while defragmenting

PR #5498, opened 2025-07-15
server: Make memory defrag blocking, accept optional threshold

abhijat:
Previously memory defragment command would set up a flag, which would make the next background task run without skipping. With this change the command initiates defragment task in a blocking manner. The command now also accepts an optional threshold, which defaults to either value supplied with command line flag: mem_defrag_page_utilization_threshold or 0.8.

Follow up PRs will collect statistics about the defragmentation process.

The changes here are taken from spike PR https://github.com/dragonflydb/dragonfly/pull/5488


src/server/memory_cmd.cc
```diff
@@ -168,9 +168,12 @@ void MemoryCmd::Run(CmdArgList args) {
   }
 
   if (parser.Check("DEFRAGMENT")) {
-    shard_set->pool()->DispatchOnAll([](util::ProactorBase*) {
+    static const float default_threshold =
```
romange, line 177: looks good, please update `HELP`.

abhijat: added description to help

PR #5499, opened 2025-07-15
chore: replace size() > 0 with !empty()

kostasrim:
* replace occurrences of size() > 0 with empty()

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1583,7 +1583,7 @@ string Connection::DebugInfo() const {
   }
   absl::StrAppend(&info, "dispatch_fiber:joinable=", async_fb_.IsJoinable(), ", ");
 
-  bool intrusive_front = dispatch_q_.size() > 0 && dispatch_q_.front().IsControl();
+  bool intrusive_front = !dispatch_q_.empty() && dispatch_q_.front().IsControl();
   absl::StrAppend(&info, "dispatch_queue:size=", dispatch_q_.size(), ", ");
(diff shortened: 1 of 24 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -2083,9 +2083,9 @@ bool Connection::IsReplySizeOverLimit() const {
   std::atomic<size_t>& reply_sz = tl_facade_stats->reply_stats.squashing_current_reply_size;
   size_t current = reply_sz.load(std::memory_order_acquire);
   const bool over_limit = reply_size_limit != 0 && current > 0 && current > reply_size_limit;
-  // Every 10 seconds. Otherwise, it can be too sensitive on certain workloads in production
-  // instances.
-  LOG_EVERY_N(INFO, 10) << "MultiCommandSquasher overlimit: " << current << "/" << reply_size_limit;
+  LOG_EVERY_T(INFO, 10) << "Commands squashing current reply size is overlimit: " << current << "/"
```
kostasrim: This is the only LOG() change.

romange: I am sorry but this LOG is completely wrong. you run it unconditionally on a critical path

kostasrim:
> you should not bundle 200 lines change with an actually important fix. should I know go over 200 lines and see there is no typo?

no I will split -- it's fine.

src/facade/dragonfly_connection.cc
kostasrim: yeah I thought I added the dif and I messed it up. Will follow up soon.

romange (approved):

Issue #5496, opened 2025-07-14
eviction policy does not evict entries with zero values


romange:
if we run dragonfly with cache_mode=true, our eviction policy `FreeMemWithEvictionStepAtomic` omits entries with values that did not use allocations. This is certainly a mistake as keys could allocate memory and we need to consider both. it's easy to reproduce by filling the datastore  with long keys and small values (say integers)

the problem is here:
```
 if (evict_it->first.IsSticky() || !evict_it->second.HasAllocated())
          continue;
```

PR #5500, opened 2025-07-15
fix: allow eviction of zero-allocation entries in cache mode

vyavdoshenko:
Fixes issue where entries with zero-allocation values (like small integers) were not being evicted in cache mode, potentially causing memory bloat.

The previous implementation incorrectly skipped evicting entries where `HasAllocated()` returns false, but keys themselves consume memory and should be eligible for eviction.

 Changes:
  - Remove `!evict_it->second.HasAllocated()` check from eviction logic in `DbSlice::FreeMemWithEvictionStepAtomic`
  - Add test `ZeroAllocationEviction` to verify entries with small values but long keys can be evicted

 Fixes: [#5496](https://github.com/dragonflydb/dragonfly/issues/5496)

### DIFF
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -1418,7 +1418,9 @@ pair<uint64_t, size_t> DbSlice::FreeMemWithEvictionStepAtomic(DbIndex db_ind,
 
         auto evict_it = db_table->prime.GetIterator(segment_id, bucket_id, slot_id);
         // TODO: consider evicting inline entries as well
-        if (evict_it->first.IsSticky() || evict_it->second.IsExternal())
+
+        bool has_allocated = evict_it->second.HasAllocated() || evict_it->first.HasAllocated();
+        if (evict_it->first.IsSticky() || !has_allocated)
           continue;
 
         // check if the key is locked by looking up transaction table.
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -1417,7 +1417,7 @@ pair<uint64_t, size_t> DbSlice::FreeMemWithEvictionStepAtomic(DbIndex db_ind,
           continue;
 
         auto evict_it = db_table->prime.GetIterator(segment_id, bucket_id, slot_id);
-        if (evict_it->first.IsSticky() || !evict_it->second.HasAllocated())
+        if (evict_it->first.IsSticky())
           continue;
 
         // check if the key is locked by looking up transaction table.
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -1417,7 +1417,7 @@ pair<uint64_t, size_t> DbSlice::FreeMemWithEvictionStepAtomic(DbIndex db_ind,
           continue;
 
         auto evict_it = db_table->prime.GetIterator(segment_id, bucket_id, slot_id);
-        if (evict_it->first.IsSticky())
+        if (evict_it->first.IsSticky() || evict_it->second.IsExternal())
           continue;
 
         // check if the key is locked by looking up transaction table.
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -1417,6 +1417,7 @@ pair<uint64_t, size_t> DbSlice::FreeMemWithEvictionStepAtomic(DbIndex db_ind,
           continue;
 
         auto evict_it = db_table->prime.GetIterator(segment_id, bucket_id, slot_id);
+        // TODO: consider evicting inline entries as well
         if (evict_it->first.IsSticky() || evict_it->second.IsExternal())
           continue;
 
--- a/src/server/dragonfly_test.cc
+++ b/src/server/dragonfly_test.cc
@@ -591,6 +591,7 @@ TEST_F(DflyEngineTest, ZeroAllocationEviction) {
   }
 
   // Should have evicted some entries with zero-allocation values
+  // but not external (disk storage) entries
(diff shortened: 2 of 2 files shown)


src/server/db_slice.cc
```diff
@@ -1417,7 +1417,7 @@ pair<uint64_t, size_t> DbSlice::FreeMemWithEvictionStepAtomic(DbIndex db_ind,
           continue;
 
         auto evict_it = db_table->prime.GetIterator(segment_id, bucket_id, slot_id);
-        if (evict_it->first.IsSticky() || !evict_it->second.HasAllocated())
+        if (evict_it->first.IsSticky())
```
kostasrim: What if `taglen == EXTERNAL_TAG` ? This goes to disk storage, why are we freeing this ?

Sure the key will be big, but will it matter ? (not sure what the right answer is here -- just thinking out loud)

kostasrim: Also what if a `key is small` and `HasAllocated()` is false. Shall we consider this a candidate for eviction ?

Since this problem is a side effect of long keys with small values, shall we `define what a small and big key is` such that we add a limit here ? Maybe something like `if key.size > small_key_threshold` ? 

@romange wdyh ?

kostasrim: Also this unconditionally make all keys as `candidates`. Is this really what we want ? (Imagine small keys 2 bytes with integers as values)

kostasrim: Oh also, if a `series of keys` are small, we have a threshold of `max_eviction_per_heartbeat=100` which basically will make `eviction extremely ineffective under certain scenarios` because now we will be evicting `100 small items` instead of `100 big ones` 

Specifically:

```
  1432         evicted_bytes += evict_it->first.MallocUsed() + evict_it->second.MallocUsed();            
  1433         ++evicted_items;
  1434         PerformDeletion(Iterator(evict_it, StringOrView::FromView(key)), db_table.get());
  1435  
  1436         // returns when whichever condition is met first                                          
  1437         if ((evicted_items == max_eviction_per_hb) || (evicted_bytes >= increase_goal_bytes))
  1438           goto finish;
  1439       }

```

This won't do very well if `keys are small` (with integers as values) as we will max out the `max_eviction_per_hb=100` and we won't be effective

vyavdoshenko: I added fix for external values

vyavdoshenko: > Oh also, if a `series of keys` are small, we have a threshold of `max_eviction_per_heartbeat=100` which basically will make `eviction extremely ineffective under certain scenarios` because now we will be evicting `100 small items` instead of `100 big ones`

What do you suggest to do?
Originally, this issue happens because the customer has so much (long keys/small values) data that he experienced a low memory issue.

kostasrim: IMO the external values are the least of my concerns here. My concern here is that he algorithm might be extremely inefficient under certain scenarios, 

I would for sure consider:

1. defining what a small and big key is (how many characters before we consider it) including the value type/size
2. make sure that it doesn't degrade when the distribution of keys and values are small with a few big ones. (we should reach these instead of evicting the small ones)

kostasrim: maybe I am overthinking this so let's wait for @romange to add some input

src/server/db_slice.cc
```diff
@@ -1417,7 +1417,7 @@ pair<uint64_t, size_t> DbSlice::FreeMemWithEvictionStepAtomic(DbIndex db_ind,
           continue;
 
         auto evict_it = db_table->prime.GetIterator(segment_id, bucket_id, slot_id);
-        if (evict_it->first.IsSticky() || !evict_it->second.HasAllocated())
+        if (evict_it->first.IsSticky() || evict_it->second.IsExternal())
```
romange: it's not true that we should never evict external items, though I agree we have not evicted them before as well as 
HasAllocated was false for externals.

I think the right step right now (as I want as little changes as possible) is 
`
bool has_allocated = evict_it->second.HasAllocated() || evict_it->first.HasAllocated();
if (... || !has_allocated)
```
and also add `TODO: consider evicting inline entries as well`.  We evict inline entries in db_slice when CanGrow returns false so I need to see how everything works before we do stronger changes

Comment (vyavdoshenko) [reply src/server/db_slice.cc]: todo was added

File: src/server/db_slice.cc [hunk 3]
```diff
@@ -1417,7 +1417,8 @@ pair<uint64_t, size_t> DbSlice::FreeMemWithEvictionStepAtomic(DbIndex db_ind,
           continue;
 
         auto evict_it = db_table->prime.GetIterator(segment_id, bucket_id, slot_id);
-        if (evict_it->first.IsSticky() || !evict_it->second.HasAllocated())
+        // TODO: consider evicting inline entries as well
+        if (evict_it->first.IsSticky() || evict_it->second.IsExternal())
```
kostasrim:
```suggestion
        bool has_allocated = evict_it->second.HasAllocated() || evict_it->first.HasAllocated();
        If(evict_it->first.IsSticky()  || !has_allocated)

kostasrim: as roman suggested

vyavdoshenko: fixed

src/server/db_slice.cc
romange: the algorith is indeed inefficient but improving it is out of scope of this patch.

romange (approved):

PR #5501, opened 2025-07-15
chore: log frequency for IsReplySizeOverLimit

kostasrim:
* reduce log frequency ever 1 second
* only log if the reply size is overlimit


src/facade/dragonfly_connection.cc
```diff
@@ -2083,9 +2083,11 @@ bool Connection::IsReplySizeOverLimit() const {
   std::atomic<size_t>& reply_sz = tl_facade_stats->reply_stats.squashing_current_reply_size;
   size_t current = reply_sz.load(std::memory_order_acquire);
   const bool over_limit = reply_size_limit != 0 && current > 0 && current > reply_size_limit;
-  // Every 10 seconds. Otherwise, it can be too sensitive on certain workloads in production
-  // instances.
-  LOG_EVERY_N(INFO, 10) << "MultiCommandSquasher overlimit: " << current << "/" << reply_size_limit;
+  if (over_limit) {
+    LOG_EVERY_T(INFO, 10) << "Commands squashing current reply size is overlimit: " << current
```
kostasrim, line 2087: maybe 1s ?

romange (approved):

PR #5502, opened 2025-07-15
core,server: collect page usage stats during defragmentation task

abhijat:
The patch for mimalloc is modified to return more information about whether a page is underutilized:

* page address, used to count unique pages scanned
* block size
* capacity blocks
* reserved blocks
*  used blocks
* flags - stores information if page cannot be considered for realloc (page is full, is used for malloc, heap mismatch etc)
* should_realloc

A class called `PageUsage` is introduced, which wraps the existing `zmalloc_page_is_underutilized` and `mi_heap_page_is_underutilized` functions. It reads the return value from the mimalloc struct and stores the data, returning a boolean.

The defragmentation call chain is modified to use the new page usage object to check if page is underutilized.

The collected results of the call tree are not used in this PR, they are stored in a vector but discarded at the end. The next smaller PR can summarize the collected results and report them as a result of `memory defragment` command.

The storing of results is optional, it is disabled for the background defrag task and enabled for the interactive command.

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -50,6 +50,7 @@ cxx_test(allocation_tracker_test dfly_core absl::random_random LABELS DFLY)
 cxx_test(qlist_test dfly_core DATA testdata/list.txt.zst LABELS DFLY)
 cxx_test(zstd_test dfly_core TRDP::zstd LABELS DFLY)
 cxx_test(top_keys_test dfly_core LABELS DFLY)
+cxx_test(page_usage_stats_test dfly_core LABELS DFLY)
 
 if(LIB_PCRE2)
   target_compile_definitions(dfly_core_test PRIVATE USE_PCRE2=1)
--- a/src/core/page_usage_stats_test.cc
+++ b/src/core/page_usage_stats_test.cc
@@ -1,0 +1,106 @@
+// Copyright 2025, DragonflyDB authors.  All rights reserved.
+// See LICENSE for licensing terms.
+//
+
+#include "core/page_usage_stats.h"
+
+#include <gmock/gmock-matchers.h>
+
+#include "base/gtest.h"
+#include "base/logging.h"
+#include "core/mi_memory_resource.h"
+#include "core/score_map.h"
+#include "core/small_string.h"
+#include "core/sorted_map.h"
+#include "core/string_map.h"
+#include "core/string_set.h"
+
+extern "C" {
+#include "redis/zmalloc.h"
+}
+
+class PageUsageStatsTest : public ::testing::Test {
+ protected:
+  static void SetUpTestSuite() {
+    init_zmalloc_threadlocal(mi_heap_get_backing());
+  }
+
+  static void TearDownTestSuite() {
+    mi_heap_collect(mi_heap_get_backing(), true);
+    mi_heap_visit_blocks(
+        mi_heap_get_backing(), false,
+        [](auto*, auto* a, void*, size_t block_sz, void*) {
+          LOG(ERROR) << "Unfreed allocations: block_size " << block_sz
+                     << ", allocated: " << a->used * block_sz;
+          return true;
+        },
+        nullptr);
+  }
+
+  PageUsageStatsTest() : m_(mi_heap_get_backing()) {
+  }
+
+  void SetUp() override {
+    score_map_ = std::make_unique<dfly::ScoreMap>(&m_);
+    sorted_map_ = std::make_unique<dfly::detail::SortedMap>(&m_);
+    string_set_ = std::make_unique<dfly::StringSet>(&m_);
+    string_map_ = std::make_unique<dfly::StringMap>(&m_);
+    dfly::SmallString::InitThreadLocal(m_.heap());
+  }
+
+  void TearDown() override {
+    score_map_.reset();
+    sorted_map_.reset();
+    string_set_.reset();
+    string_map_.reset();
+    small_string_.Free();
+    EXPECT_EQ(zmalloc_used_memory_tl, 0);
+  }
+
+  dfly::MiMemoryResource m_;
+  std::unique_ptr<dfly::ScoreMap> score_map_;
+  std::unique_ptr<dfly::detail::SortedMap> sorted_map_;
+  std::unique_ptr<dfly::StringSet> string_set_;
+  std::unique_ptr<dfly::StringMap> string_map_;
+  dfly::SmallString small_string_{};
+};
+
+bool PageCannotRealloc(mi_page_usage_stats_t s) {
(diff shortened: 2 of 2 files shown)


src/redis/zmalloc.h
```diff
@@ -128,7 +128,7 @@ int zmalloc_get_allocator_wasted_blocks(float ratio, size_t* allocated, size_t*
  * This uses the current local thread heap.
  * return 0 if not, 1 if underutilized
  */
-int zmalloc_page_is_underutilized(void *ptr, float ratio);
+void zmalloc_page_is_underutilized(void *ptr, float ratio, void* stats_ptr, int collect_stats);
```
abhijat: It might be simpler if we can just return the new struct back from here, but in this header it looks like we  do not include `mimalloc` specific headers, hence the `void*` workaround to return the object.

abhijat:
This PR is unfortunately a bit large sized but mostly contains mechanical changes to pass the `PageUsage` object into defrag/realloc operations.

patches/mimalloc-v2.2.4/2_return_stat.patch
```diff
@@ -0,0 +1,66 @@
...
++    .block_size = page->block_size,
++    .capacity = page->capacity,
++    .reserved = page->reserved,
++    .used = page->used,
++    .should_realloc = false,
++    .flags = 0,
++  };
++
++  if (page->flags.x.in_full == 1) {
++    result.flags |= 1;
```
romange: Please introduce defines around `mi_page_usage_stats_s` declaration for flags and add a comment around the flags mentioning these defines

src/core/compact_object.h
```diff
@@ -72,7 +73,7 @@ class RobjWrapper {
 
   // Try reducing memory fragmentation by re-allocating values from underutilized pages.
   // Returns true if re-allocated.
-  bool DefragIfNeeded(float ratio);
+  bool DefragIfNeeded(PageUsage& page_usage);
```
romange: pass mutable objects by pointer

src/core/compact_object.h
```diff
@@ -244,7 +245,7 @@ class CompactObj {
     mask_bits_.touched = e;
   }
 
-  bool DefragIfNeeded(float ratio);
+  bool DefragIfNeeded(PageUsage& page_usage);
```
romange: same and below

src/core/sorted_map_test.cc
```diff
@@ -11,6 +11,7 @@
 #include "base/gtest.h"
 #include "base/logging.h"
 #include "core/mi_memory_resource.h"
+#include "page_usage_stats.h"
```
romange:
```suggestion
#include "core/page_usage_stats.h"
```

src/redis/zmalloc.h
romange: it's fine

romange: You can actually declare like `struct mi_page_usage_stats_s`

romange: I think c even allows using types like:
`zmalloc_page_is_underutilized(void *ptr, float ratio, struct mi_page_usage_stats_s* dest, nt collect_stats);`

without predeclaration but I am not sure

romange (commented):
Good work, @abhijat !

src/redis/zmalloc.h
abhijat: I had to use `struct mi_page_usage_stats_s`, it looks like is not possible to include `mimalloc/types.h` in a C header while using a C++ compiler, it breaks `extern "C"` inclusion everywhere as it includes the C++ thread header at one point.

patches/mimalloc-v2.2.4/2_return_stat.patch
abhijat: added

src/core/compact_object.h
abhijat: changed

src/core/compact_object.h
abhijat: changed everywhere

src/core/sorted_map_test.cc
abhijat: fixed

patches/mimalloc-v2.2.4/2_return_stat.patch
```diff
@@ -0,0 +1,66 @@
...
++
++  if (page_heap != heap) {
++    result.flags |= MI_DFLY_HEAP_MISMATCH;
++  }
++
++  if (page->prev == NULL) {
++    result.flags |= MI_DFLY_PAGE_USED_FOR_MALLOC;
++  }
++
++  if (result.flags != 0) {
```
romange: nit, just to make it a bit more compact:
```

if (result.flags == 0 && result.used <= (unsigned) (result.capacity * ratio)) {
  result.flags = MI_DFLY_PAGE_BELOW_THRESHOLD;
}
return result
```

src/core/compact_object_test.cc
```diff
@@ -19,6 +19,7 @@
 #include "core/mi_memory_resource.h"
 #include "core/string_or_view.h"
 #include "core/string_set.h"
+#include "page_usage_stats.h"
```
romange: core/

src/core/score_map_test.cc
```diff
@@ -9,6 +9,7 @@
 #include "base/gtest.h"
 #include "base/logging.h"
 #include "core/mi_memory_resource.h"
+#include "page_usage_stats.h"
```
romange: core/

patches/mimalloc-v2.2.4/2_return_stat.patch
abhijat: changed

src/core/compact_object_test.cc
abhijat: changed, sorry I haven't been able to teach my IDE to auto import with the correct path. this is an issue in 4 places, all are fixed

src/core/score_map_test.cc
abhijat: fixed

src/core/compact_object_test.cc
abhijat: maybe I will just disable auto import for now

romange (approved):

src/core/compact_object_test.cc
romange: Yeah, please tell me where so I could disable as well

abhijat: It depends on what your editor is, for vscode+clangd (or anything with clangd afaik) it is 

```
"clangd.arguments": [ "--header-insertion=never" ]
```

For my case (clion) it is https://www.jetbrains.com/help/clion/creating-imports.html#auto-import-settings

PR #5503, opened 2025-07-15
fix: adjust BLMPOP command options and implementation for proper journaling

EricHayter:
This commit adds the `CO::NO_AUTOJOURNAL` command option for the `BLMPOP` command to disable default journaling. Also, the commit ensures that the `journal_rewrite` parameter is set to `true` when calling `OpPop`.

Resolves issue mentioned in #5370


romange (approved):

romange:
Thanks!

PR #5504, opened 2025-07-15
chore: add more metrics

romange:
1. pipeline metrics regarding its depth before squashing.
2. various metrics around table memory and shard threads.

### DIFF
--- a/src/server/cluster/cluster_family.cc
+++ b/src/server/cluster/cluster_family.cc
@@ -667,13 +667,18 @@ void ClusterFamily::DflyClusterGetSlotInfo(CmdArgList args, SinkReplyBuilder* bu
     rb->StartArray(9);
     rb->SendLong(slot_data.first);
     rb->SendBulkString("key_count");
-    rb->SendLong(static_cast<long>(slot_data.second.key_count));
+    rb->SendLong(slot_data.second.key_count);
     rb->SendBulkString("total_reads");
-    rb->SendLong(static_cast<long>(slot_data.second.total_reads));
+    rb->SendLong(slot_data.second.total_reads);
     rb->SendBulkString("total_writes");
-    rb->SendLong(static_cast<long>(slot_data.second.total_writes));
+    rb->SendLong(slot_data.second.total_writes);
+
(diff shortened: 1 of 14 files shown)


src/server/cluster/cluster_family.cc
```diff
@@ -667,13 +667,17 @@ void ClusterFamily::DflyClusterGetSlotInfo(CmdArgList args, SinkReplyBuilder* bu
...
     rb->SendBulkString("total_writes");
-    rb->SendLong(static_cast<long>(slot_data.second.total_writes));
+    rb->SendLong(slot_data.second.total_writes);
+
+    // Account for both the values and the table space of the entries.
+    // Sometimes the values are very small and table space becomes significant.
     rb->SendBulkString("memory_bytes");
-    rb->SendLong(static_cast<long>(slot_data.second.memory_bytes));
+    rb->SendLong(slot_data.second.memory_bytes +
+                 slot_data.second.key_count * sizeof(CompactObj) * 2);
```
BorysTheDev, line 681: Could you clarify the formula. What is * 2

src/server/multi_command_squasher.cc
```diff
@@ -246,7 +251,12 @@ bool MultiCommandSquasher::ExecuteSquashed(facade::RedisReplyBuilder* rb) {
     fb2::BlockingCounter bc(num_shards);
     DVLOG(1) << "Squashing " << num_shards << " " << tx->DebugId();
 
+    static uint64_t tsc_cycles_limit =
+        base::CycleClock::Frequency() * absl::GetFlag(FLAGS_max_busy_squash_usec) / 1000'000;
```
BorysTheDev, line 255: Maybe we need to cache the flag if we are not going to update it in runtime?

BorysTheDev (dismissed):

src/server/multi_command_squasher.cc
romange: it is static so cached

src/server/cluster/cluster_family.cc
romange: I expanded. pls see lines 676-677

Issue #5396, opened 2025-07-01
wrong return type for `json.numincrby` and `json.nummultby`


romange:
to reproduce (with resp3):

```
> json.set a $ 1
> json.numincrby a $ 1
"[2]"
```
but it is expected to have an array with a single integer

romange:
Also `json.type a $`
returns:
```
1) "integer"
```

but should:
```
1) 1) "integer"
```
`json.type a .` should return `1) "integer"`

romange:
Also `json.nummultby a $ 2` returns the wrong type.

vyavdoshenko:
@romange 
I am a little bit confused because the Dragonfly behavior is almost the same as Valkey, but differs from Redis.

Dragonfly from main branch:
```
> info server
# Server
redis_version:7.4.0
dragonfly_version:df-dev

> json.set a $ 1
OK

> json.numincrby a $ 1
"[2]"

> json.type a $
1) "integer"

> json.type a .
"integer"

> json.nummultby a $ 2
"[4]"
```

Valkey behavior:
`docker run -d --name valkey-bundle -p 6379:6379 valkey/valkey-bundle:latest`
```
> info server
# Server
redis_version:7.2.4
server_name:valkey
valkey_version:8.1.2

> json.set a $ 1
OK

> json.numincrby a $ 1
"[2]"

> json.type a $
1) integer

> json.type a .
integer

> json.nummultby a $ 2
"[4]"
```

Redis behavior:
`docker run -d -p 6379:6379 --name redis-stack redis/redis-stack-server:latest`
```
> info server
# Server
redis_version:7.4.5

> json.set a $ 1
OK

> json.numincrby a $ 1
1) (integer) 2

> json.type a $
1) 1) "integer"

> json.type a .
1) "integer"

> json.nummultby a $ 2
1) (integer) 4
```

Should the Dragonfly behavior be fixed in this case?

romange:
Like Redis. It's reasonable that numeric operations will return numbers and
not strings.

Also pls check with resp3 mode and floats

Roman Gershman
CTO
---
*www.dragonflydb.io <http://www.dragonflydb.io>*

PR #5517, opened 2025-07-16
feat: add RESP3 support for JSON.TYPE and JSON numeric operations

vyavdoshenko:
Fixes: https://github.com/dragonflydb/dragonfly/issues/5396

Implements proper RESP3 formatting for JSON commands:
- JSON.TYPE
- JSON.NUMINCRBY
- JSON.NUMMULTBY

Fixes JSON commands to follow RESP3 protocol.

### DIFF
--- a/src/server/json_family.cc
+++ b/src/server/json_family.cc
@@ -337,7 +337,7 @@ template <typename T> void Send(const OpResult<T>& result, RedisReplyBuilder* rb
 }
 
 // Specialization for JSON numeric operations that return string results
-template <> void Send(const OpResult<string>& result, RedisReplyBuilder* rb) {
+void SendNumericData(const OpResult<string>& result, RedisReplyBuilder* rb) {
   RedisReplyBuilder::ReplyScope scope{rb};
   if (result) {
     const string& json_str = result.value();
@@ -2039,7 +2039,7 @@ void JsonFamily::NumIncrBy(CmdArgList args, const CommandContext& cmd_cntx) {
 
   OpResult<string> result = cmd_cntx.tx->ScheduleSingleHopT(std::move(cb));
   auto* rb = static_cast<RedisReplyBuilder*>(builder);
-  reply_generic::Send(result, rb);
+  reply_generic::SendNumericData(result, rb);
 }
 
 void JsonFamily::NumMultBy(CmdArgList args, const CommandContext& cmd_cntx) {
@@ -2056,7 +2056,7 @@ void JsonFamily::NumMultBy(CmdArgList args, const CommandContext& cmd_cntx) {
 
   OpResult<string> result = cmd_cntx.tx->ScheduleSingleHopT(std::move(cb));
   auto* rb = static_cast<RedisReplyBuilder*>(builder);
-  reply_generic::Send(result, rb);
+  reply_generic::SendNumericData(result, rb);
 }
 
 void JsonFamily::Toggle(CmdArgList args, const CommandContext& cmd_cntx) {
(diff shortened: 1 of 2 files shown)


src/server/json_family.cc
```diff
@@ -1979,7 +2016,7 @@ void JsonFamily::NumIncrBy(CmdArgList args, const CommandContext& cmd_cntx) {
 
   OpResult<string> result = cmd_cntx.tx->ScheduleSingleHopT(std::move(cb));
   auto* rb = static_cast<RedisReplyBuilder*>(builder);
-  reply_generic::Send(result, rb);
+  reply_generic::SendNumericData(result, rb);
```
romange: i am surprised that `reply_generic::Send` does not do it automatically. maybe it is possible to fix it there?
@BagritsevichStepan  can you take a look?

src/server/json_family.cc
```diff
@@ -2028,7 +2065,47 @@ void JsonFamily::Type(CmdArgList args, const CommandContext& cmd_cntx) {
...
+  }
+
+  if (callback_result.ShouldSendWrongType()) {
+    rb->SendError(OpStatus::WRONG_JSON_TYPE);
+    return;
+  }
+
+  if (callback_result.IsV1()) {
+    // Legacy path (.) - single value
+    if (rb->IsResp3()) {
```
BagritsevichStepan: I suggest that you explain what the bug was with RESP3. From what I see, you fixed it by adding a StartArray call before each SendBulkString invocation.
Is this issue specific to JSON.TYPE, or does it affect all other JSON commands that send bulk strings?
If it's the latter, we could simply update the generic method instead, and you wouldn’t need to write so much code:
```
void Send(const std::string& value, RedisReplyBuilder* rb) {
 // TODO: if rb->IsResp3 do rb->StartArray
  rb->SendBulkString(value);
}
```

And even if this change could introduce bugs in other commands, I would still suggest creating a wrapper type like Resp3String, along with a new `Send(Resp3String...)` method.
This would help us avoid a lot of boilerplate code here — which was actually the original goal of having a generic Send method in the first place.
(Also, maybe you will find other JSON commands that will need also Resp3String to work properly)

src/server/json_family.cc
```diff
@@ -314,6 +314,43 @@ template <typename T> void Send(const OpResult<T>& result, RedisReplyBuilder* rb
...
+void SendNumericData(const OpResult<string>& result, RedisReplyBuilder* rb) {
+  RedisReplyBuilder::ReplyScope scope{rb};
+  if (result) {
+    const string& json_str = result.value();
+
+    // Check if this is a JSON array result from numeric operations
+    if (!json_str.empty() && json_str[0] == '[' && json_str.back() == ']' && rb->IsResp3()) {
+      // For RESP3, parse JSON array and send as proper array with numeric types
+      std::optional<JsonType> parsed_json =
+          JsonFromString(json_str, PMR_NS::get_default_resource());
```
BagritsevichStepan, line 341: Let's take a look on how OpDoubleArithmetic method is working. It holds JsonType and then returns it by using json_str->as_string. And then you are converting it from string to JsonType again. We can just return JsonType.

But let’s go a bit deeper — we can see that `result.AddValue(*val);` is being called there.
So, you could simply return a std::vector<std::variant<int64_t, double>> from `OpDoubleArithmetic`, and then:
- use it directly to send results in the RESP3 case, or
- convert it to a string in the non-RESP3 case.

This would simplify the logic and reduce convertions.

src/server/json_family.cc
BagritsevichStepan: It does not do it automatically because result is just a string. [I suggest just to change `OpDoubleArithmetic`](https://github.com/dragonflydb/dragonfly/pull/5517#discussion_r2212637455)

PR #5518, opened 2025-07-16
fix: reduce cardinality of the huffman histogram

romange:
This is done because apparently zstd implementation can write out of bounds and cause crashes if counts are too high.

So we limit the counting per shard if we cross 1<<19 freq, and once we merge all the shards, we also scale down counts to be below 1<<23.

### DIFF
--- a/src/core/dfly_core_test.cc
+++ b/src/core/dfly_core_test.cc
@@ -365,6 +365,58 @@ TEST_F(HuffCoderTest, Decode) {
   ASSERT_EQ("aab", string_view(decode_dest.data(), decoded_size));
 }
 
+TEST_F(HuffCoderTest, HugeHistogram) {
+  array<unsigned, 256> hist{
+      1,         1,         1,         1,         1,         1,         1,         1,
+      5,         26,        543,       1,         1,         1,         1,         1,
+      4,         1,         1,         1,         1,         1,         1,         1,
+      1,         1,         1,         1,         1,         1,         1,         1,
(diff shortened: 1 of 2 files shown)


src/core/dfly_core_test.cc
```diff
@@ -365,6 +365,58 @@ TEST_F(HuffCoderTest, Decode) {
   ASSERT_EQ("aab", string_view(decode_dest.data(), decoded_size));
 }
 
+TEST_F(HuffCoderTest, HugeHistogram) {
+  array<unsigned, 256> hist{
```
romange, line 369: it's just a random histogram that reproduced the problem, so I copied it here

src/server/debugcmd.cc
```diff
@@ -1318,6 +1341,37 @@ void DebugCmd::Keys(CmdArgList args, facade::SinkReplyBuilder* builder) {
...
+  for (unsigned i = 0; i <= HufHist::kMaxSymbol; i++) {
+    // raw_size may count less characters than the actual size because
+    // we may cut the counting early.
+    raw_size += hist[i];
+    if (hist[i] > max_freq) {
+      max_freq = hist[i];
+    }
+    if (hist[i] == 0) {
+      hist[i] = 1;  // Avoid zero frequency symbols.
+    }
```
vyavdoshenko, line 1358: Why does this need to be done twice?
Is it possible to get rid of it?

romange: I did not find a way - the scaling down can bring more numbers to 0

PR #5519, opened 2025-07-17
fix: enable keep alive on replica side of replication connections

romange:

### DIFF
--- a/src/server/multi_command_squasher.cc
+++ b/src/server/multi_command_squasher.cc
@@ -235,6 +235,7 @@ bool MultiCommandSquasher::ExecuteSquashed(facade::RedisReplyBuilder* rb) {
 
   Transaction* tx = cntx_->transaction;
   ServerState::tlocal()->stats.multi_squash_executions++;
+  ServerState::tlocal()->stats.squash_width_freq_arr[num_shards - 1]++;
   ProactorBase* proactor = ProactorBase::me();
   uint64_t start = proactor->GetMonotonicTimeNs();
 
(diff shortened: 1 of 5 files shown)


BorysTheDev (approved):

src/server/multi_command_squasher.cc
```diff
@@ -235,6 +235,7 @@ bool MultiCommandSquasher::ExecuteSquashed(facade::RedisReplyBuilder* rb) {
 
   Transaction* tx = cntx_->transaction;
   ServerState::tlocal()->stats.multi_squash_executions++;
+  ServerState::tlocal()->stats.squash_width_freq_arr[num_shards - 1]++;
```
kostasrim, line 238: Why are we incrementing only the last element ?

src/server/server_state.cc
```diff
@@ -73,6 +74,12 @@ ServerState::Stats& ServerState::Stats::Add(const ServerState::Stats& other) {
   } else {
     this->tx_width_freq_arr = other.tx_width_freq_arr;
   }
+  if (this->squash_width_freq_arr.size() > 0) {
```
kostasrim, line 77: Maybe I am missing something but default initializing the `squash_width_freq_arr` in the constructor above will make it of size `num_shards`. We don't modify the size of the valarray only its contents. 

So why here do we need to check if it's greater than `>0` and then DCHECK that they have equal sizes `or else assign directly` ? 

Why simply not:

```
//This should always be true
CHECK((this->squash_width_freq_arr.size(), other.squash_width_freq_arr.size());
// ok elements default initialized to 0 on the constructor
this->squash_width_freq_arr += other.squash_width_freq_arr;
```

romange: when we merge statistics we do not provide num_shards and we aggregate `other` into a result

PR #5523, opened 2025-07-17
chore: fix monitor preemption in atomic section

kostasrim:
The issue is that `SendAsync` might preempt if the monitor connection needs to shut down because the pipeline buffer is over the limit. The fix is to move the ShutdownSelfBlocking to allow the async fiber and set the con_clossing to true.

* shutdown the connection within AsyncFiber instead of SendAsync
* rename ShutdownSelf to ShutdownSelfBlocking

### DIFF
--- a/src/facade/conn_context.h
+++ b/src/facade/conn_context.h
@@ -55,9 +55,6 @@ class ConnectionContext {
   // Skip ACL validation, used by internal commands and commands run on admin port
   bool skip_acl_validation = false;
 
-  // Monitor shutdown
-  bool monitor_shutdown_conn = false;
-
   // How many async subscription sources are active: monitor and/or pubsub - at most 2.
   uint8_t subscriptions;
 
--- a/src/facade/conn_context.h
+++ b/src/facade/conn_context.h
@@ -55,6 +55,9 @@ class ConnectionContext {
   // Skip ACL validation, used by internal commands and commands run on admin port
   bool skip_acl_validation = false;
 
+  // Monitor shutdown
+  bool monitor_shutdown_conn = false;
+
   // How many async subscription sources are active: monitor and/or pubsub - at most 2.
   uint8_t subscriptions;
 
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1877,6 +1877,8 @@ void Connection::SendAsync(MessageHandle msg) {
   // Close MONITOR connection if we overflow pipeline limits
   if (msg.IsMonitor() &&
       qbp.IsPipelineBufferOverLimit(stats_->dispatch_queue_bytes, dispatch_q_.size())) {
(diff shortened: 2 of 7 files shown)


kostasrim:
@romange it seems that we had multiple issues around non preemptive flows and `Tls::Shutdown` as most of the time it threw our assumptions off. I will look all the flows that call `TlsSocket::Shutdown` to make sure that everything is fine. 

We already suffer from this problem with `ACL's` but these are not that critical. I am looking at the rest.   

https://github.com/dragonflydb/dragonfly/issues/5524

src/facade/conn_context.h
```diff
@@ -55,6 +55,9 @@ class ConnectionContext {
   // Skip ACL validation, used by internal commands and commands run on admin port
   bool skip_acl_validation = false;
 
+  // Monitor shutdown
+  bool monitor_shutdown_conn = false;
```
romange: maybe `request_shutdown = false`, whether it's monitor it's not important

src/facade/dragonfly_connection.cc
```diff
@@ -1161,8 +1161,12 @@ void Connection::DispatchSingle(bool has_more, absl::FunctionRef<void()> invoke_
           qbp.IsPipelineBufferOverLimit(stats_->dispatch_queue_bytes, dispatch_q_.size());
       return !over_limits || (dispatch_q_.empty() && !cc_->async_dispatch) || cc_->conn_closing;
     });
-    if (cc_->conn_closing)
+    if (cc_->conn_closing) {
+      if (cc_->monitor_shutdown_conn) {
+        ShutdownSelf();
```
romange: can you please rename it ShutdownSelfBlocking?

src/facade/conn_context.h
kostasrim: done. I also wanted to get rid of this flag but there is a problematic flow:

After `IoLoop` completes, we wait for the async fiber to finish and then `if there was a parser error` we do a socket io. If async fiber shutsdown the connection unconditionally, we can't do that.

I will see how can simplify this alongside with fixes for all flows that use `SelfShutdownBlocking`

src/facade/dragonfly_connection.h
```diff
@@ -494,6 +494,8 @@ class Connection : public util::Connection {
       bool is_main_ : 1;
     };
   };
+
+  bool request_shutdown = false;
```
romange: there are bunch of flags two lines above like migration_in_process_ migration_enabled_ etc 
why not move there? :)

romange (approved):

kostasrim:
@romange forgot a rename and it didn't compile. It should be fine now

romange (approved):

romange:
It failed @kostasrim

kostasrim:
> It failed @kostasrim

No it did not. It was the dreaded test_replication_onmove_flow failing which I disabled. Irrelevant to this PR whatsoever

Issue #5524, opened 2025-07-17
investigate all call sites of socket Shutdown


kostasrim:
TLS connections might preempt during shutdown. Most of our flows, that shutdown sockets assume atomicity yes this assumption is off when the socket is a TLS socket. 

Investigate all call sites to make sure that we don't throw away our gurantees. 

1) ACL's suffer from that

kostasrim:
Quick `grep` (all find all references in your IDE)

```
server/main_service.cc:1801:  cmd_cntx.conn_cntx->conn()->ShutdownSelf();
server/server_state.cc:265:    // We use weak refs, because ShutdownSelf below can potentially block the fiber,
server/server_state.cc:306:        conn->ShutdownSelf();
server/acl/acl_family.cc:172:      connection->ShutdownSelf();
server/acl/acl_family.cc:188:      connection->ShutdownSelf();
server/server_family.cc:572:        conn->ShutdownSelf();
facade/dragonfly_connection.cc:1773:void Connection::ShutdownSelf() {
facade/dragonfly_connection.cc:1880:    ShutdownSelf();
facade/dragonfly_connection.h:234:  void ShutdownSelf();
```

I went over all the call sites. The last one pending for a fix are the `Acl` flows and I raised [#5525](https://github.com/dragonflydb/dragonfly/pull/5525)

All the other flows are ok.

PR #5525, opened 2025-07-18
fix(acl): preemptive TlsSocket::Shutdown in atomic section

kostasrim:
The issue is that we called `TlsSocket::Shutdown` while traversing the connections. Since Shutdown is preemptive, it violates the atomicity guarantees of the operation.

Helps with #5524

### DIFF
--- a/src/server/acl/acl_family.cc
+++ b/src/server/acl/acl_family.cc
@@ -163,35 +163,40 @@ void AclFamily::SetUser(CmdArgList args, const CommandContext& cmd_cntx) {
   std::visit(Overloaded{error_case, update_case}, std::move(req));
 }
 
-void AclFamily::EvictOpenConnectionsOnAllProactors(const absl::flat_hash_set<string_view>& users) {
-  auto close_cb = [&users]([[maybe_unused]] size_t id, util::Connection* conn) {
-    CHECK(conn);
-    auto connection = static_cast<facade::Connection*>(conn);
-    auto ctx = static_cast<ConnectionContext*>(connection->cntx());
-    if (ctx && users.contains(ctx->authed_username)) {
-      connection->ShutdownSelf();
+template <typename P> void AclFamily::TraverseEvictImpl(P predicate) {
+  auto close_cb = [&, this](unsigned idx, util::ProactorBase* p) {
+    std::vector<facade::Connection::WeakRef> connections;
+    auto traverse_cb = [&, this](unsigned id, util::Connection* conn) {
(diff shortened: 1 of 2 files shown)


src/server/acl/acl_family.cc
```diff
@@ -163,37 +163,40 @@ void AclFamily::SetUser(CmdArgList args, const CommandContext& cmd_cntx) {
...
 }
 
-void AclFamily::EvictOpenConnectionsOnAllProactors(const absl::flat_hash_set<string_view>& users) {
-  auto close_cb = [&users]([[maybe_unused]] size_t id, util::Connection* conn) {
-    CHECK(conn);
-    auto connection = static_cast<facade::Connection*>(conn);
-    auto ctx = static_cast<ConnectionContext*>(connection->cntx());
-    if (ctx && users.contains(ctx->authed_username)) {
-      connection->ShutdownSelfBlocking();
+template <typename P> void AclFamily::TraverseEvictImpl(P predicate) {
```
dranikpg: 1. Please add a one-liner comment what the idea of TraverseEvict is
2. Maybe use absl::FunctionalRef<ConnectionContext*>. The calls are in the thousands at most, so it doesn't have to be inlined. As a bonus we get distinct  types
3. Maybe even move it to anonymous namespace (it only uses main_listener_)

src/server/acl/acl_family.cc
```diff
@@ -163,37 +163,40 @@ void AclFamily::SetUser(CmdArgList args, const CommandContext& cmd_cntx) {
...
+      if (predicate(ctx)) {
+        connections.push_back(connection->Borrow());
+      }
+    };
+
+    main_listener_->TraverseConnectionsOnThread(traverse_cb, UINT32_MAX, nullptr);
+
+    for (auto& tcon : connections) {
+      facade::Connection* conn = tcon.Get();
+      if (conn && conn->socket()->proactor()->GetPoolIndex() == p->GetPoolIndex()) {
```
dranikpg: your second part of the if reminded me that we don't invalidate weakrefs for migrated connections.... 🤯 If we do, it breaks everything, if we don't it breaks everything as well 🤷🏻‍♂️ 🤷🏻‍♂️ 🤷🏻‍♂️

kostasrim: Yep I am aware :grimacing: The good news is that we set `migrate_connections=false` as far as I remember for prod

vyavdoshenko (commented):
lgtm

vyavdoshenko (approved):
lgtm

PR #5527, opened 2025-07-18
chore: time sensitive sscan

kostasrim:
Bound `sscan` similar to what we do with `scan` to avoid cases of long/slow matching


src/server/set_family.cc
```diff
@@ -123,13 +124,18 @@ struct StringSetWrapper {
     uint32_t count = scan_op.limit;
     long maxiterations = count * 10;
 
+    const auto start_cycles = base::CycleClock::Now();
+    // Don't allow it to monopolize cpu time. Roughly 15 microseconds.
+    const uint64_t timeout_cycles = base::CycleClock::Frequency() >> 16;
```
kostasrim: We use the same for Scan(). Imo this is an estimate and not really a proper time interval :/

PR #5536, opened 2025-07-21
chore: Add cmake build options to exclude code #1

dranikpg:
I don't want to compile linux from scratch whenever I want to build DF for a simple GET/SET benchmark. Building from scratch in minimal configuration (debug) on my pc now takes just 35 seconds. 

Adding WITH_JSON to remove (jsoncons) might bring it down to less than 20 seconds. (For Part 2)
image

---
As from build-from-source.md: 
| Option | Description |
|--|--|
| WITH_AWS | Include AWS client. Required for cloud snapshots |
| WITH_SEARCH | Include Search module |
| WITH_COLLECTION_CMDS | Include commands for collections (SET, HSET, ZSET) |
| WITH_EXTENSION_CMDS | Include extension commands (Bloom, HLL, JSON, ...) |

Minimal debug build:

```bash
./helio/blaze.sh -DWITH_AWS=OFF -DWITH_SEARCH=OFF -WITH_COLLECTION_CMDS=OFF -DWITH_EXTENSION_CMDS=OFF
```

### DIFF
--- a/src/server/CMakeLists.txt
+++ b/src/server/CMakeLists.txt
@@ -38,7 +38,7 @@ endif()
 # Optionally compile collection commands
 if (WITH_COLLECTION_CMDS)
   set(DF_FAMILY_SRCS set_family.cc hset_family.cc zset_family.cc stream_family.cc)
-  add_definitions(-D_WITH_COLLECTION_CMDS)
+  add_definitions(-DWITH_COLLECTION_CMDS)
 else()
   set(DF_FAMILY_SRCS collection_family_fallback.cc)
 endif()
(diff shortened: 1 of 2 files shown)


dranikpg:
> It would be nice to improve compilation times without disabling parts of the code base. For that we need to parellelize as we compile TU's more (as the final link will me serial anyway). So maybe that's a direction for future considerations ?

Yes, I thought about it. Our `src/server` code is too interleaved to effectively divide it into separate units. I introduced only one new target for search/server. But it might help, at least with linking less

But at the end it just boils down to simple maths. I want to run `GET/SET`, yet 90% of the time is spent downloading and compiling aws client/jsoncons/gperf/hnswlib/uni-algo/etc, compiling lots of our code that I'll never run

Issue #5529, opened 2025-07-19
`debug stacktrace` crashes inside ConnectionsWatcherFb


romange:
Reproduces only on aarch64 in release mode (tried on r6g.xlarge)

To reproduce:

1. `./dragonfly --timeout=500`
2. `redis-cli DEBUG STACKTRACE`

crashes in absl code inside `absl::lts_20240722::GetStackTrace()`.
Something is special in ConnectionsWatcherFb that reproduces it. 
One direction is - to just remove pieces of code there until we understand why it happens.
Another option is to check if newer version of absl (say https://github.com/abseil/abseil-cpp/releases/tag/20250512.1) fixes the issue.

vyavdoshenko:
I successfully reproduced this issue locally. I tried to update the Abseil library to the mentioned (latest) version: 20250512.1
The bug doesn't reproduce on the latest version. I suggest updating the Helio third parties to the latest Abseil version. It looks reasonable.
The question: the current version is patched with a custom patch. As far as I understand, it was done to support symbolizing different versions of libc.
@romange, should we add the same patch for the new version of the library?

vyavdoshenko:
Fixes:
https://github.com/romange/helio/pull/435
https://github.com/dragonflydb/dragonfly/pull/5537

vyavdoshenko:
The latest version of the Abseil library with crash: https://github.com/abseil/abseil-cpp/releases/tag/20250127.1
The next version and all following have the fix: https://github.com/abseil/abseil-cpp/releases/tag/20250512.rc1

The fix was in absl/debugging/internal/stacktrace_aarch64-inl.inc

The commit: https://github.com/abseil/abseil-cpp/commit/a9a1b50468762a05a7407c2b5d18cc50db0d0bf4

PR #5537, opened 2025-07-21
fix: includes fixed to be up to date with new absl library version

vyavdoshenko:
These changes are relevant to the absl library version update to the latest one: https://github.com/abseil/abseil-cpp/releases/tag/20250512.1

Relevant to this issue: https://github.com/dragonflydb/dragonfly/issues/5529
The absl library version update: https://github.com/romange/helio/pull/435

### DIFF
--- a/src/core/json/detail/flat_dfs.h
+++ b/src/core/json/detail/flat_dfs.h
@@ -8,10 +8,10 @@
 
 #include <variant>
 
-#include "base/expected.hpp"
 #include "core/flatbuffers.h"
 #include "core/json/detail/common.h"
 #include "core/json/path.h"
+#include "nonstd/expected.hpp"
 
 namespace dfly::json::detail {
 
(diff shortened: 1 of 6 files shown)


src/core/json/detail/flat_dfs.h
```diff
@@ -8,10 +8,10 @@
 
 #include <variant>
 
-#include "base/expected.hpp"
```
dranikpg, line 11: How come it worked before 🤔 Hmmm... Somehow I missed this

vyavdoshenko: Older version of helio

kostasrim:
@vyavdoshenko Can you plz double check in the `abseil changelog` that there were changes to this function on the version we update to ? If so, plz include it here

kostasrim:
@vyavdoshenko from a quick perplexity search there are no documented fixes in the new version we try to update. However, I don't have full faith in the AI so I would suggest to use git blame and see recent changes to that function. If there aren't any then I suspect that `absl::lts_20240722::GetStackTrace()` blows out the stack of the fiber and we crash. We can adjust fiber stack and since this is reproducible we can also test this hypothesis easily

PR #5544, opened 2025-07-22
chore: do not shutdown master on takeover in cluster mode

### LINKED ISSUE #5521: repltakeover in cluster mode should redirect to the replica
Currently `repltakeover` command for non-cluster mode shuts down the master.

For cluster mode it creates unnecessary disconnects and service unavailability for several seconds.
It's unnecessary as master knows exactly what is the last known location of its slots - his old replica.

The expected behavior should be -
to return "MOVED <replica_addr>" for any transactional command , i.e. similarly to what we do with non-owned slots.

kostasrim:
* do not shutdown old master after takeover in cluster mode
* add test

resolves #5521

### DIFF
--- a/src/server/cluster/cluster_family.cc
+++ b/src/server/cluster/cluster_family.cc
@@ -1102,11 +1102,11 @@ void ClusterFamily::ReconcileMasterReplicaTakeoverSlots(const ClusterExtendedNod
   }
 
   auto new_config = config->CloneWithChanges({}, {});
-  auto mutable_shard_info = new_config->GetMutableConfig();
+  auto current_shard_info = new_config->GetMutableConfig();
 
-  for (ClusterShardInfo& info : mutable_shard_info) {
+  for (ClusterShardInfo& info : current_shard_info) {
     // we can't use == because it also contains the id in the comparisson.
-    // It's a mistake as ip/port are enough to make it unique but anyway...
+    // It's a mistake as ip/port are enough to make it unique.
     if (info.master.port == old_master.port && info.master.ip == old_master.ip) {
       info.master.port = new_master.port;
       info.master.ip = new_master.ip;
@@ -1119,7 +1119,7 @@ void ClusterFamily::ReconcileMasterReplicaTakeoverSlots(const ClusterExtendedNod
           break;
         }
       }
-      CHECK(it != end);
+      LOG_EVER_T(ERROR, 1) << "replica not found in the list of cluster replicas for old master";
       info.replicas.erase(it);
       break;
     }
--- a/src/server/cluster/cluster_family.cc
+++ b/src/server/cluster/cluster_family.cc
@@ -1104,23 +1104,35 @@ void ClusterFamily::ReconcileMasterReplicaTakeoverSlots(const ClusterExtendedNod
   auto new_config = config->CloneWithChanges({}, {});
   auto current_shard_info = new_config->GetMutableConfig();
 
+  // Going to extreme for this because the listener socket address returns 0.0.0.0
+  // The replica address returns localhost and the nodes in the cluster config
+  // have 127.0.0.1. To allow local testing, I normalize them.
+  auto is_equal = [](const std::string& ip_a, const std::string& ip_b) -> bool {
+    if (ip_a == ip_b) {
+      return true;
+    }
+    bool ip_a_local = false;
+    if (ip_a == "127.0.0.1" || ip_a == "localhost" || ip_a == "0.0.0.0") {
+      ip_a_local = true;
+    }
+    bool ip_b_local = false;
+    if (ip_b == "127.0.0.1" || ip_b == "localhost" || ip_b == "0.0.0.0") {
+      ip_b_local = true;
+    }
+    if (ip_a_local && ip_b_local) {
+      return true;
+    }
+    return false;
+  };
+
   for (ClusterShardInfo& info : current_shard_info) {
     // we can't use == because it also contains the id in the comparisson.
     // It's a mistake as ip/port are enough to make it unique.
-    if (info.master.port == old_master.port && info.master.ip == old_master.ip) {
+    if (info.master.port == old_master.port && is_equal(info.master.ip, old_master.ip)) {
       info.master.port = new_master.port;
       info.master.ip = new_master.ip;
-      auto it = info.replicas.begin();
-      auto end = info.replicas.end();
-      for (; it != end; ++it) {
-        auto& replica = *it;
-        if (replica.ip == new_master.ip && replica.port == new_master.port) {
-          info.master.id = replica.id;
-          break;
-        }
-      }
-      LOG_EVER_T(ERROR, 1) << "replica not found in the list of cluster replicas for old master";
-      info.replicas.erase(it);
+      info.master.id = id_;
+      info.replicas.clear();
       break;
     }
   }
--- a/src/server/cluster/cluster_family.cc
+++ b/src/server/cluster/cluster_family.cc
@@ -1088,8 +1088,7 @@ size_t ClusterFamily::MigrationsErrorsCount() const {
   return error_num;
 }
 
-void ClusterFamily::ReconcileMasterReplicaTakeoverSlots(const ClusterExtendedNodeInfo& old_master,
-                                                        const ClusterExtendedNodeInfo& new_master) {
+void ClusterFamily::ReconcileMasterReplicaTakeoverSlots() {
   util::fb2::LockGuard gu(set_config_mu);
   util::fb2::LockGuard lk(migration_mu_);
 
@@ -1104,37 +1103,23 @@ void ClusterFamily::ReconcileMasterReplicaTakeoverSlots(const ClusterExtendedNod
   auto new_config = config->CloneWithChanges({}, {});
   auto current_shard_info = new_config->GetMutableConfig();
 
-  // Going to extreme for this because the listener socket address returns 0.0.0.0
-  // The replica address returns localhost and the nodes in the cluster config
-  // have 127.0.0.1. To allow local testing, I normalize them.
-  auto is_equal = [](const std::string& ip_a, const std::string& ip_b) -> bool {
-    if (ip_a == ip_b) {
-      return true;
-    }
-    bool ip_a_local = false;
-    if (ip_a == "127.0.0.1" || ip_a == "localhost" || ip_a == "0.0.0.0") {
-      ip_a_local = true;
-    }
-    bool ip_b_local = false;
-    if (ip_b == "127.0.0.1" || ip_b == "localhost" || ip_b == "0.0.0.0") {
-      ip_b_local = true;
-    }
-    if (ip_a_local && ip_b_local) {
-      return true;
-    }
-    return false;
-  };
-
+  // Replace master with replica in shard config.
+  bool found = false;
   for (ClusterShardInfo& info : current_shard_info) {
-    // we can't use == because it also contains the id in the comparisson.
-    // It's a mistake as ip/port are enough to make it unique.
-    if (info.master.port == old_master.port && is_equal(info.master.ip, old_master.ip)) {
-      info.master = new_master;
-      info.master.id = id_;
-      info.replicas.clear();
-      break;
+    for (const auto& replica : info.replicas) {
+      if (replica.id == id_) {
+        info.master = replica;
+        // New master has no replicas
+        info.replicas.clear();
+        found = true;
+        break;
+      }
     }
+    if (found)
+      break;
   }
+
+  LOG_IF(ERROR, !found) << "Did not find replica in the cluster map";
 
   server_family_->service().proactor_pool().AwaitFiberOnAll(
       [&new_config](util::ProactorBase*) { ClusterConfig::SetCurrent(new_config); });
--- a/src/server/cluster/cluster_family.cc
+++ b/src/server/cluster/cluster_family.cc
@@ -1088,24 +1088,31 @@ size_t ClusterFamily::MigrationsErrorsCount() const {
   return error_num;
 }
 
-void ClusterFamily::ReconcileMasterReplicaTakeoverSlots() {
-  util::fb2::LockGuard gu(set_config_mu);
-  util::fb2::LockGuard lk(migration_mu_);
+void ClusterFamily::ReconcileMasterFlow() {
+  auto config = cluster::ClusterConfig::Current();
 
-  auto config = ClusterConfig::Current();
+  for (auto& info : config->GetMutableConfig()) {
+    if (info.master.id == id_) {
+      if (!info.replicas.empty()) {
+        LOG_IF(ERROR, info.replicas.size() > 1)
+            << "More than one replica found, slot redirection after takeover corrupted";
 
-  // Sanity -- we should not reach there
-  if (!config) {
-    LOG_EVERY_T(ERROR, 1) << "Cluster config after takeover is empty";
-    return;
+        // assumes there is one replica per master node
+        // TODO figure a smart way to get the replica id_ so
+        // we can find it here
+        info.master = info.replicas.front();
+        info.replicas.clear();
+      }
+      break;
+    }
   }
+}
 
-  auto new_config = config->CloneWithChanges({}, {});
-  auto current_shard_info = new_config->GetMutableConfig();
-
+void ClusterFamily::ReconcileReplicaFlow() {
+  auto new_config = ClusterConfig::Current()->CloneWithChanges({}, {});
   // Replace master with replica in shard config.
   bool found = false;
-  for (ClusterShardInfo& info : current_shard_info) {
+  for (ClusterShardInfo& info : new_config->GetMutableConfig()) {
     for (const auto& replica : info.replicas) {
       if (replica.id == id_) {
         info.master = replica;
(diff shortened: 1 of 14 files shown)


src/server/dflycmd.h
```diff
@@ -232,6 +247,11 @@ class DflyCmd {
   using ReplicaInfoMap = absl::btree_map<uint32_t, std::shared_ptr<ReplicaInfo>>;
   ReplicaInfoMap replica_infos_ ABSL_GUARDED_BY(mu_);
 
+  // We only acquire a write lock in TakeOver. This happens only once, as the state
+  // after the takeover is TAKEN_OVER. All other paths, acquire it in read only mode.
+  util::fb2::SharedMutex last_takeover_src_mu_;
+  std::shared_ptr<ReplicaInfo> last_takeover_source_;
```
kostasrim: TODO separate PR.

I spoke with @andydunstall and we would also like to bring back and taken over node via `REPLICAOF` command. 

So the flow is:

1. node a master, node b replica
2. node b take over a (a is TAKEN_OVER state. b is master now)
3. after some time connect to node a, and `REPLICAOF B` 

As is `REPLICAOF` would fail because of the state. 

We should make the necessary changes to support that

src/server/cluster/cluster_family.cc
```diff
@@ -1088,6 +1088,47 @@ size_t ClusterFamily::MigrationsErrorsCount() const {
...
+  // Sanity -- we should not reach there
+  if (!config) {
+    LOG_EVERY_T(ERROR, 1) << "Cluster config after takeover is empty";
+    return;
+  }
+
+  auto new_config = config->CloneWithChanges({}, {});
+  auto mutable_shard_info = new_config->GetMutableConfig();
+
+  for (ClusterShardInfo& info : mutable_shard_info) {
```
romange: nit: rename to `current_info`

src/server/server_family.cc
```diff
@@ -3304,6 +3307,16 @@ void ServerFamily::ReplTakeOver(CmdArgList args, const CommandContext& cmd_cntx)
...
+  const auto& master_ctx = replica_->server();
+  cluster::ClusterExtendedNodeInfo old_master_info;
+  old_master_info.ip = master_ctx.host;
+  old_master_info.port = master_ctx.port;
+
+  cluster::ClusterExtendedNodeInfo new_master_info;
+  old_master_info.ip = absl::GetFlag(FLAGS_bind);
+  old_master_info.port = master_ctx.port;
+
+  service().cluster_family().ReconcileMasterReplicaTakeoverSlots(old_master_info, new_master_info);
```
kostasrim: This is not really needed but if omitted it kinda breaks the `cluster config` state which can have side effects ?

It's not really needed because after the takeover, the `replica immediately serves write traffic for the slots`. This works, because of how we map slot ranges to master/replicas from the cluster family and config.

However, after the takeover, we are the "true owner" of the slots (master) and therefore we should at least update the local node map to reflect that. 

The only question I have, if we make the config changes locally and after that we receive the new updated config from the cp this should be ok right ?

src/server/cluster/cluster_family.cc
```diff
@@ -1088,6 +1088,47 @@ size_t ClusterFamily::MigrationsErrorsCount() const {
   return error_num;
 }
 
+void ClusterFamily::ReconcileMasterReplicaTakeoverSlots(const ClusterExtendedNodeInfo& old_master,
```
romange: what is new_master in this context? is it this replica node where we run `ReconcileMasterReplicaTakeoverSlots`?

src/server/cluster/cluster_family.cc
```diff
@@ -1088,6 +1088,47 @@ size_t ClusterFamily::MigrationsErrorsCount() const {
...
+      auto it = info.replicas.begin();
+      auto end = info.replicas.end();
+      for (; it != end; ++it) {
+        auto& replica = *it;
+        if (replica.ip == new_master.ip && replica.port == new_master.port) {
+          info.master.id = replica.id;
+          break;
+        }
+      }
+      CHECK(it != end);
```
romange: I prefer not to check fail here. just erase if `it  != end`

src/server/cluster/cluster_family.cc
```diff
@@ -1088,6 +1088,47 @@ size_t ClusterFamily::MigrationsErrorsCount() const {
...
+
+  for (ClusterShardInfo& info : mutable_shard_info) {
+    // we can't use == because it also contains the id in the comparisson.
+    // It's a mistake as ip/port are enough to make it unique but anyway...
+    if (info.master.port == old_master.port && info.master.ip == old_master.ip) {
+      info.master.port = new_master.port;
+      info.master.ip = new_master.ip;
+      auto it = info.replicas.begin();
+      auto end = info.replicas.end();
+      for (; it != end; ++it) {
```
romange: can you explain what we are doing here? because I thought at first it runs on (old) replica process, now I am confused

src/server/cluster/cluster_family.h
```diff
@@ -44,6 +44,11 @@ class ClusterFamily {
 
   size_t MigrationsErrorsCount() const ABSL_LOCKS_EXCLUDED(migration_mu_);
 
+  // Helper function to be used from repltakeover flow. It swaps master and replica
+  // slot config transparently for this node only.
```
romange: should it swap them though? `for this node only` - is it a replica node or a master node (before repltakeover) or maybe it is applied on both nodes?

src/server/main_service.h
```diff
@@ -160,6 +160,10 @@ class Service : public facade::ServiceInterface {
   std::optional<facade::ErrorReply> CheckKeysOwnership(const CommandId* cid, CmdArgList args,
                                                        const ConnectionContext& dfly_cntx);
 
+  // Return error if not all keys are owned by the server when running in cluster mode
```
romange: Returns MOVED error?

src/server/cluster/cluster_family.cc
kostasrim: Yes! The new master, is the node that initiated the takeover

src/server/cluster/cluster_family.cc
kostasrim: Before takeover, imagine node `a` is master and node `b` is replica. We takeover, now b master and a is TAKENOVER. We update the local config of node `b` to contain that info (that now b is master and a no longer exists).

Basically this code, quickly parses the local cluster config and replaces the relevant info to reflect the ones after the takeover

src/server/cluster/cluster_family.h
kostasrim: I don't quite get the question. `ReconcileMasterReplicaTakeoverSlots` is called from the replica that is taking over the master. It only updates the replica (now new master) local config to reflect the take over state -> which basically means replace the old master node with this one in the local config

kostasrim: > should it swap them though? 

If we are treating the new node the same as the old master then it should have an updated config as we discussed earlier to reflect that ?

The only thing I am not sure is that we update 1 local config only.  Sooner or later we will receive a new one that reflects the changes we did locally to the whole cluster. Once we receive them, we should process them without an issue (and here I am not 100% sure about if the local change we made can cause an error during the new config update -- based on my understanding it should not)

src/server/server_family.cc
romange: good question, I do not know.

src/server/cluster/cluster_family.cc
romange: So can you please add comments to this function, explaning the semantics of the arguments?

src/server/cluster/cluster_family.cc
romange: that I understand. but whose config it updates here? b's? Why would `b` have replicas if it was a replica before the takeover. After the takeover it still does not have replica. Or you modify the struct belonging to `a` before and put address of `b`  there?

kostasrim: Yes are the node `b` and we update our local config.

>  Why would b have replicas if it was a replica before the takeover

Yes! Exactly. It's already empty there are no replicas!

>  Or you modify the struct belonging to a before and put address of b there?

Nope, there is one struct that encapsulates the `shard` which includes `master and replicas`. In our case, we only update the master info (replicas shall be empty).

src/server/cluster/cluster_family.cc
```diff
@@ -1088,6 +1088,59 @@ size_t ClusterFamily::MigrationsErrorsCount() const {
...
+    LOG_EVERY_T(ERROR, 1) << "Cluster config after takeover is empty";
+    return;
+  }
+
+  auto new_config = config->CloneWithChanges({}, {});
+  auto current_shard_info = new_config->GetMutableConfig();
+
+  // Going to extreme for this because the listener socket address returns 0.0.0.0
+  // The replica address returns localhost and the nodes in the cluster config
+  // have 127.0.0.1. To allow local testing, I normalize them.
```
kostasrim: To test this locally :face_exhaling: :facepalm:

tests/dragonfly/cluster_test.py
```diff
@@ -3357,3 +3357,59 @@ async def test_slot_migration_oom(df_factory):
...
+    # we updated the config as well
+    assert await r1.client.execute_command("SET X 2") == "OK"
+
+    master_nodes = [r1, m2]
+    r1.slots = [(0, 9000)]
+    nodes.pop(0)
+    await push_config(json.dumps(generate_config(master_nodes)), [node.client for node in nodes])
+
+    assert await r1.client.execute_command("GET X") == "2"
+    assert await m2.client.execute_command("GET FOOX") == "1"
```
kostasrim, line 3415: Push the new topology cluster config and make sure state is the same

src/server/cluster/cluster_family.cc
```diff
@@ -1088,6 +1088,59 @@ size_t ClusterFamily::MigrationsErrorsCount() const {
...
+    if (ip_a_local && ip_b_local) {
+      return true;
+    }
+    return false;
+  };
+
+  for (ClusterShardInfo& info : current_shard_info) {
+    // we can't use == because it also contains the id in the comparisson.
+    // It's a mistake as ip/port are enough to make it unique.
+    if (info.master.port == old_master.port && is_equal(info.master.ip, old_master.ip)) {
```
kostasrim: @romange do we run replication or cluster connections on admin port or normal port ? It's important to do the right thing here.

src/server/cluster/cluster_family.h
romange: my point is that the comment should be:
```
// Helper function to be used from repltakeover flow. It swaps master and replica
// slot config transparently for replica node becoming the new master (where it is called).
```

src/server/main_service.cc
```diff
@@ -68,6 +68,10 @@ using namespace std;
 using facade::operator""_KB;
 using facade::ErrorReply;
 
+ABSL_FLAG(string, bind, "",
```
romange: why did you change that?

kostasrim: because originally I thought we set `bind` but we don't. I forgot to revert it IMO

src/server/protocol_client.h
```diff
@@ -54,6 +53,11 @@ class ProtocolClient {
     std::string Description() const;
   };
 
+  const ServerContext& server() const {
```
romange:
```suggestion
  const ServerContext& server_context() const {
```

src/server/replica.h
```diff
@@ -55,6 +55,8 @@ class Replica : ProtocolClient {
           std::optional<cluster::SlotRange> slot_range);
   ~Replica();
 
+  using ProtocolClient::server;
```
romange: and the rest -to change accordingly

src/server/server_family.cc
```diff
@@ -65,6 +65,9 @@ extern "C" {
 #include "util/accept_server.h"
 #include "util/aws/aws.h"
 
+ABSL_DECLARE_FLAG(int32_t, port);
```
romange: where do you use it?

src/server/cluster/cluster_family.cc
romange: you should be able to check it easily by looking at any of the cluster datastores in prod.

src/server/cluster/cluster_family.cc
```diff
@@ -1088,6 +1088,58 @@ size_t ClusterFamily::MigrationsErrorsCount() const {
...
+    return;
+  }
+
+  auto new_config = config->CloneWithChanges({}, {});
+  auto current_shard_info = new_config->GetMutableConfig();
+
+  // Going to extreme for this because the listener socket address returns 0.0.0.0
+  // The replica address returns localhost and the nodes in the cluster config
+  // have 127.0.0.1. To allow local testing, I normalize them.
+  auto is_equal = [](const std::string& ip_a, const std::string& ip_b) -> bool {
```
romange: something here feels not right. what happens with the remote nodes where a dns is resolved to ip?
My intuition is that such normalization fixes a symptom and not the root cause, and instead compare resolved ips to each other.

romange: can you explain why we can not match node_ids in this context?

kostasrim: we can and this is the right approach. IMO what I had here was unnecessary. It also removed 40 lines of changes :)

src/server/main_service.cc
```diff
@@ -1039,6 +1039,46 @@ optional<ErrorReply> Service::CheckKeysOwnership(const CommandId* cid, CmdArgLis
...
+    return nullopt;
+  }
+
+  if (auto error = cluster::SlotOwnershipError(*keys_slot);
+      !error.status.has_value() || error.status.value() != facade::OpStatus::OK) {
+    return ErrorReply{std::move(error)};
+  }
+
+  return facade::ErrorReply{
+      absl::StrCat("-MOVED ", *keys_slot, " ", server_family_.TakenOverMaster(), ":",
```
kostasrim: Fixing this as well. This is wrong

src/server/cluster/cluster_family.cc
```diff
@@ -1088,6 +1088,70 @@ size_t ClusterFamily::MigrationsErrorsCount() const {
...
+  for (auto& info : config->GetMutableConfig()) {
+    if (info.master.id == id_) {
+      if (!info.replicas.empty()) {
+        LOG_IF(ERROR, info.replicas.size() > 1)
+            << "More than one replica found, slot redirection after takeover corrupted";
+
+        // assumes there is one replica per master node
+        // TODO figure a smart way to get the replica id_ so
+        // we can find it here
+        info.master = info.replicas.front();
```
kostasrim, line 1103: It's safe to assume that there is one replica. When this changes we should also take care of this. Furthermore, we need to figure out a smart way to the replica.id after `DflyCmd::TakeVer`.

@romange let me know if you disagree

src/server/cluster/cluster_family.h
```diff
@@ -44,6 +44,10 @@ class ClusterFamily {
 
   size_t MigrationsErrorsCount() const ABSL_LOCKS_EXCLUDED(migration_mu_);
 
+  // Helper function to be used from repltakeover flow. It swaps master and replica
+  // slot config transparently for replica node becoming the new master (where it is called).
+  void ReconcileMasterReplicaTakeoverSlots(bool was_master);
```
romange: seems better, no?
seems like now comment is not correct, or I do not understand `was_master` variable

src/server/server_family.h
```diff
@@ -291,6 +291,9 @@ class ServerFamily {
                        [&match](auto& elem) { return elem.state == match; });
   }
 
+  // Used by DFLY command after replica takeover
```
romange: Please also add what it does (aka: updates old master node config by removing itself from the map so it could point to the promoted replica) or smthing like this

src/server/cluster/cluster_family.cc
```diff
@@ -1088,6 +1088,70 @@ size_t ClusterFamily::MigrationsErrorsCount() const {
...
+        info.replicas.clear();
+      }
+      break;
+    }
+  }
+}
+
+void ClusterFamily::ReconcileReplicaFlow() {
```
romange, line 1111: ABSL_EXCLUSIVE_LOCKS_REQUIRED

src/server/cluster/cluster_family.h
```diff
@@ -114,11 +117,13 @@ class ClusterFamily {
   std::vector<std::shared_ptr<OutgoingMigration>> outgoing_migration_jobs_
       ABSL_GUARDED_BY(migration_mu_);
 
- private:
   std::optional<ClusterShardInfos> GetShardInfos(ConnectionContext* cntx) const;
 
   ClusterShardInfo GetEmulatedShardInfo(ConnectionContext* cntx) const;
 
+  void ReconcileMasterFlow();
+  void ReconcileReplicaFlow();
```
romange: ABSL_EXCLUSIVE_LOCKS_REQUIRED

src/server/server_family.cc
```diff
@@ -3603,6 +3605,10 @@ void ServerFamily::ClientPauseCmd(CmdArgList args, SinkReplyBuilder* builder,
   }
 }
 
+void ServerFamily::ReconcileMasterReplicaTakeoverSlots() {
```
romange: the fact you named it the same for both generic `cluster_family().ReconcileMasterReplicaTakeoverSlots(bool)`  and here but only for master is confusing.

src/server/cluster/cluster_family.h
kostasrim: way way way better! Yes I need to update the comment

src/server/server_family.cc
kostasrim: Yes I agree. In fact, I will remove this function. We can just call in dflcmd `service().cluster().Reconcile()`

PR #5546, opened 2025-07-22
chore: Command dispatch refactoring

dranikpg:
Preparation for squashing from journal executor and general cleanup

1. Change ParsedEntry format to be convertible to StoredCmd
2. Update `InvokeCmd` error propagation
3. Update MultiCommandSquasher interface to allow passing commands after initialization

### DIFF
--- a/src/server/journal/executor.cc
+++ b/src/server/journal/executor.cc
@@ -11,6 +11,7 @@
 #include <memory>
 
 #include "base/logging.h"
+#include "facade/error.h"
 #include "facade/service_interface.h"
 #include "server/main_service.h"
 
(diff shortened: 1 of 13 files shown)


src/server/main_service.cc
```diff
@@ -1388,33 +1388,27 @@ OpResult<void> OpTrackKeys(const OpArgs slice_args, const facade::Connection::We
...
-    if (res == OpStatus::OUT_OF_MEMORY) {
-      return DispatchResult::OOM;
-    }
+    if (cid->name() == "REPLCONF" && absl::EqualsIgnoreCase(ArgS(tail_args, 0), "ACK"))
+      return;  // Don't propagate error
 
-    return DispatchResult::OK;  // return ERROR only for internal error aborts
+    builder->SendError(std::move(*err));
+    return;
   }
```
dranikpg, line 1418: We already use the builder to propagate errors. The addition of `DispatchResult` is totally sloppy, it's converted into another all over the code, checked differently in different places... and the result of `InvokeCmd` is used only in half of the callsites 🤷🏻‍♂️🤷🏻‍♂️🤷🏻‍♂️🤷🏻‍♂️ The other half doesn't know about it

src/server/journal/types.h
```diff
@@ -71,11 +71,10 @@ struct Entry : public EntryBase {
...
   struct CmdData {
-    std::unique_ptr<uint8_t[]> command_buf;
-    CmdArgVec cmd_args;  // represents the parsed command.
-    size_t cmd_len{0};
-  };
-  CmdData cmd;
+    std::string command;
+    std::string arg_buf;
+    absl::InlinedVector<uint32_t, 4> arg_sizes;
+  } cmd;
```
dranikpg: To be easily convertible to a StoredCmd

src/server/journal/executor.h
```diff
@@ -34,15 +34,26 @@ class JournalExecutor {
...
  private:
+  struct ErrorCounter : public facade::CapturingReplyBuilder {
+    explicit ErrorCounter() : CapturingReplyBuilder(facade::ReplyMode::NONE) {
+    }
+
+    void SendError(std::string_view str, std::string_view type) override;
+    void Reset();
+
+    size_t num_oom = 0, num_other = 0;
+  };
```
dranikpg, line 45: Will be used later to provide `DispatchResult`

src/server/journal/executor.cc
```diff
@@ -65,8 +71,14 @@ void JournalExecutor::FlushSlots(const cluster::SlotRange& slot_range) {
...
 facade::DispatchResult JournalExecutor::Execute(journal::ParsedEntry::CmdData& cmd) {
-  auto span = CmdArgList{cmd.cmd_args.data(), cmd.cmd_args.size()};
-  return service_->DispatchCommand(span, &reply_builder_, &conn_context_);
+  tmp_scratch_.resize(cmd.arg_sizes.size() + 1);
+  tmp_scratch_[0] = cmd.command;
+  size_t i = 1, offset = 0;
+  for (uint32_t sz : cmd.arg_sizes) {
+    tmp_scratch_[i++] = string_view{cmd.arg_buf}.substr(offset, sz);
+    offset += sz;
+  }
```
dranikpg, line 80: temporarily, we'll use StoredCmd later

dranikpg:
Not saying it's perfect, but it's better than it was

src/server/main_service.cc
```diff
@@ -1328,18 +1328,20 @@ DispatchResult Service::DispatchCommand(ArgSlice args, SinkReplyBuilder* builder
...
     dfly_cntx->transaction = nullptr;
   }
-  return res;
+
+  if (auto err = builder->ConsumeLastError(); !err.empty()) {
+    LOG_EVERY_T(WARNING, 1) << FailedCommandToString(cid->name(), args_no_cmd, err);
+    if (err == facade::kOutOfMemory)
+      return DispatchResult::OOM;
+    return DispatchResult::ERROR;
+  }
```
dranikpg, line 1342: The log doesn't really belong here as well, one option is moving it to Invoke without consuming the error

src/server/journal/executor.h
kostasrim: I will loose track of how many flavours of `ReplyBuilders` we have :rofl: But imo it's the right approach

src/server/main_service.cc
kostasrim: yeah, +1 for moving back to Invoke

src/server/main_service.cc
kostasrim: I am to blame here :wink:

PR #5547, opened 2025-07-22
chore: use stringmatchlen in GlobMatcher

kostasrim:
`GlobMatcher` constructor is extremely slow sometimes as it requires to compile the regex produced from the glob expression. A simple example is the small glob tested in the benchmark: `a*a*a*.pt` which takes roughly `70ms` to compile, `diminishing any of the performance gains produced by the fast matching`. There are ways to mitigate this but for now we should fall back to the plain and simple `stringmatchlen` which does penaltize the hot path (I will also time bound the SSCAN loop in a separate PR just like SCAN).

Results from running the benchmark locally on `release` build:

`./dfly_core_test --benchmark_filter=BM_MatchGlobSlow`

With `reflex` 
```
-----------------------------------------------------------
Benchmark                 Time             CPU   Iterations
-----------------------------------------------------------
BM_MatchGlobSlow   66588075 ns     66586936 ns           10
```

With `pcre2` I did not observe this regression but I don't want to risk the hot path.

### DIFF
--- a/src/core/glob_matcher.h
+++ b/src/core/glob_matcher.h
@@ -10,6 +10,10 @@
 
 // We opt for using Reflex library for glob matching.
 // While I find PCRE2 faster, it's not substantially faster to justify the shared lib dependency.
+
+// For some regex, Reflex (and pcre2) have extremely slow compile times(70+ms).
+// This latency is significant for the hot path and therefore both are disabled
+// and we fall back to the plain old stringmatchlen. For more info, refer to #5547 on gh.
 //#define REFLEX_PERFORMANCE
 
 #ifndef REFLEX_PERFORMANCE
--- a/src/core/glob_matcher.h
+++ b/src/core/glob_matcher.h
@@ -10,15 +10,14 @@
 
 // We opt for using Reflex library for glob matching.
 // While I find PCRE2 faster, it's not substantially faster to justify the shared lib dependency.
-#define REFLEX_PERFORMANCE
+//#define REFLEX_PERFORMANCE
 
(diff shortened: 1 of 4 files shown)


src/core/glob_matcher.h
```diff
@@ -10,16 +10,15 @@
 
 // We opt for using Reflex library for glob matching.
 // While I find PCRE2 faster, it's not substantially faster to justify the shared lib dependency.
-#define REFLEX_PERFORMANCE
+//#define REFLEX_PERFORMANCE
```
romange, line 17: please add explanation why it was disable and the reference to this PR for more details

tests/dragonfly/set_test.py
```diff
@@ -0,0 +1,25 @@
...
+    client = df.client()
+
+    await client.execute_command(f"SADD key el1 el2")
+
+    element = "a*" * 3
+
+    cursor = await client.execute_command(f"SSCAN key 0 match {element}.pt")
+    length = len(cursor[1])
+    # Takes 3 seconds
+    res = await client.execute_command("SLOWLOG GET 100")
```
romange, line 24: what is the slowlog threshold? You are risking to get false positives here due to slow runners.

romange (dismissed):

src/core/glob_matcher.h
kostasrim: Yes I will!

tests/dragonfly/set_test.py
kostasrim: I don't recall but the two commands run should be under this default for this case

PR #5550, opened 2025-07-22
chore: Refactor JournalReader::ReadString

mkaruza:
When reading string that didn't fit in current io_buf_ we need to read
remainder of string, copy it to intermediate buffer just to copy it again to
output buf. For string that have larger remainder (higher than half of
io_buf_ capacity) we could use fast route to read and write
to output buf directly, otherwise go with route to populate intermediate
io_buf_.

### DIFF
--- a/src/server/journal/serializer.cc
+++ b/src/server/journal/serializer.cc
@@ -163,16 +163,18 @@ io::Result<size_t> JournalReader::ReadString(io::MutableBytes buffer) {
 
   auto remainder_buf_pos = buffer.data() + available;
 
-  if (!is_short_remainder && remainder) {
-    uint64_t read;
-    SET_OR_UNEXPECT(source_->Read({remainder_buf_pos, remainder}), read)
-    if (read < remainder) {
-      return make_unexpected(make_error_code(errc::io_error));
+  if (remainder) {
+    if (is_short_remainder) {
+      if (auto ec = EnsureRead(remainder); ec)
+        return make_unexpected(ec);
+      buf_.ReadAndConsume(remainder, remainder_buf_pos);
+    } else {
+      uint64_t read;
+      SET_OR_UNEXPECT(source_->Read({remainder_buf_pos, remainder}), read)
+      if (read < remainder) {
+        return make_unexpected(make_error_code(errc::io_error));
+      }
     }
-  } else if (remainder) {
-    if (auto ec = EnsureRead(remainder); ec)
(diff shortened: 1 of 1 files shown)


src/server/journal/serializer.cc
```diff
@@ -145,13 +145,26 @@ io::Result<size_t> JournalReader::ReadString(io::MutableBytes buffer) {
...
+
+  if (available < size) {
+    remainder = size - available;
+  }
+
+  buf_.ReadAndConsume(available, buffer.data());
+
+  if (remainder) {
+    uint64_t read;
+    SET_OR_UNEXPECT(source_->ReadAtLeast({buffer.data() + available, remainder}, remainder), read)
```
dranikpg: What I thought about is that we increased the number of read calls with this (which are expensive), as we read only to fill the current string. For example, with many small strings, we'll issue a read call for each ReadString. Previously, we'd fill up the buffer in a single call and to disassemble it later 🤔

romange: @dranikpg  is correct -we improve the case with big strings but penalize the small ones.
We can use `ReadSome` with two vectors: buffer and buf_ and try to fill both with the same call.
But ReadSome does not guarantee "at least", so we then can follow up with `AtLeast` to fill up the rest of the `buffer` if needed. Alternatively, we can introduce vectorized `ReadAtLeast` as well

romange: by the way, you should use `Read(buffer)` since you need precise size and not at least

dranikpg: I think the best balance would be just to use Read() if the requested size is multiples of the internal buffer, otherwise keep it as is

src/server/journal/serializer.cc
```diff
@@ -145,13 +145,35 @@ io::Result<size_t> JournalReader::ReadString(io::MutableBytes buffer) {
...
-  buf_.ReadAndConsume(size, buffer.data());
+  uint64_t available = std::min(size, buf_.InputLen());
+  uint64_t remainder = 0;
+
+  if (available < size) {
+    remainder = size - available;
+  }
+
+  buf_.ReadAndConsume(available, buffer.data());
+
```
romange, line 159: nit wrap lines 160-176  in `if (remainder)` condition - I think it will improve the readability of the code and simplify the conditions inside.

romange (approved):

PR #5551, opened 2025-07-23
feat: allow migrating client connections via CLIENT MIGRATE

romange:

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -2003,8 +2003,8 @@ facade::ConnectionContext* Connection::cntx() {
   return cc_.get();
 }
 
-void Connection::RequestAsyncMigration(util::fb2::ProactorBase* dest) {
-  if (!migration_enabled_ || cc_ == nullptr) {
+void Connection::RequestAsyncMigration(util::fb2::ProactorBase* dest, bool force) {
+  if ((!force && !migration_enabled_) || cc_ == nullptr) {
     return;
   }
 
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -707,6 +707,9 @@ void Connection::OnPreMigrateThread() {
   // is marked beforehand.
   migration_in_process_ = true;
 
+  // Mark as not owned by any thread as it going through the dark hole
+  self_.reset();
+
   socket_->CancelOnErrorCb();
   DCHECK(!async_fb_.IsJoinable()) << GetClientId();
 }
@@ -719,6 +722,7 @@ void Connection::OnPostMigrateThread() {
     socket_->RegisterOnErrorCb([this](int32_t mask) { this->OnBreakCb(mask); });
   }
   migration_in_process_ = false;
+  self_ = {make_shared<std::monostate>(), this};  // Recreate shared_ptr to self.
   DCHECK(!async_fb_.IsJoinable());
 
   // If someone had sent Async during the migration, we must create async_fb_.
@@ -1820,7 +1824,7 @@ bool Connection::Migrate(util::fb2::ProactorBase* dest) {
 Connection::WeakRef Connection::Borrow() {
   DCHECK(self_);
 
-  return WeakRef(self_, socket_->proactor()->GetPoolIndex(), id_);
+  return {self_, unsigned(socket_->proactor()->GetPoolIndex()), id_};
 }
 
 void Connection::ShutdownThreadLocal() {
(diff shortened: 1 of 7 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -1987,8 +1987,8 @@ facade::ConnectionContext* Connection::cntx() {
   return cc_.get();
 }
 
-void Connection::RequestAsyncMigration(util::fb2::ProactorBase* dest) {
-  if (!migration_enabled_ || cc_ == nullptr) {
+void Connection::RequestAsyncMigration(util::fb2::ProactorBase* dest, bool force) {
```
kostasrim, line 2010: Just keep in mind that any migration can potentially break anything that traverses (with preemption) the connections because we first traverse them, collect a WeakPtr from them and then we do the preemptive operation. If there is a forced migration inbetween it can cause a problem as the local connection we are about to process has been migrated to another thread!

romange: good point

src/facade/dragonfly_connection.cc
```diff
@@ -2173,22 +2177,28 @@ void Connection::SetPipelineLowBoundStats(unsigned limit) {
...
+  // It is fine because we know the location exists, so it is safe to read.
+  // Also, if the connection migrated to another thread,
+  // it may have anything in its owning_thread_id_ variable besides
+  // the current thread id because we first reset it to UINT32_MAX in its old thread before
+  // migrating it to the new thread.
+  if (sptr->owning_thread_id_ == unsigned(ProactorBase::me()->GetPoolIndex())) {
+    // If the connection is owned by the current thread, we can safely return it.
+    return sptr.get();
+  }
+
```
dranikpg: It's easier and cleaner to just reset the shared_ptr during a migration. It would've worked without all those changes

src/facade/dragonfly_connection.cc
```diff
@@ -2173,22 +2177,28 @@ void Connection::SetPipelineLowBoundStats(unsigned limit) {
...
-  //  it doesn't manage the underlying connection. See definition of `self_`.
-  return ptr_.lock().get();
+  auto sptr = ptr_.lock();
+
+  if (!sptr) {
+    return nullptr;
+  }
+
+  // We potentially read a memory location that is written by another thread.
+  // It is fine because we know the location exists, so it is safe to read.
```
kostasrim: Just because the location exists doesn't mean it's not problematic. Concurrent read/writes are a data race and data races according to the cpp standard slowly enter the UB realm.

IMO `OnPostMigrateThread` sets `owning_thread_id_` and it runs on the newly migrated thread. So:

thread a -> traverses the weak refs it just fetched, and is about to do read `owning_thread_id`
thread b -> we just migrated so we are running `OnPostMigrateThread` which sets the `owning_thread_id` 

That's a clear data race. 

I don't think the chances of this happening are big, but it's a possibility. Why don'y we make the `owning_thread_id` atomic if it can be potentially be accessed concurrently ?

kostasrim (approved):
lgtm, left a comment

src/facade/dragonfly_connection.cc
kostasrim: The problem is that we read/write concurrently either way. That's a data race

src/facade/dragonfly_connection.cc
dranikpg: that's why it's easier to just reset the shared_ptr 🤓 We never change the thread when we use WeakRefs, so the "potential" recovery (that is possbile with this approach) is not used

src/facade/dragonfly_connection.cc
dranikpg: You don't need `owning_thread_id_`

src/facade/dragonfly_connection.cc
kostasrim: gotcha

src/facade/dragonfly_connection.cc
romange: @dranikpg  If it is possible to remove owning_thread_id_ i prefer. Are you saying just reset self_ in all the places we reset `owning_thread_id_` is enough?

dranikpg: Yes

`.reset()` before migration
`= make_shared<monostate>` after migration

dranikpg: All other poiners are weak references, so dropping `self` will invalidate the shared_ptr, thus invalidate all WeakRefs. If you keep the shared_ptr as null during the migration, all created WeakRefs will be invalid automatically

romange: Much more elegant, thanks!

PR #5553, opened 2025-07-23
chore: Add more stats to CLIENT LIST

### LINKED ISSUE #5549: extend per connection stats in client list
Currently `DEBUG IOSTATS [PS]` provides per io-thread stats of reads/pipelined requests and bandwidth.
`PS` means to sample per SECOND so `debug` command waits for a second and computes a per second sample.

Now we would like to extend per second estimation *per connection* in the "client list" command.
so it would print something like this:

`... reads-ps=700  net-in-ps=63000 pipeline-cmds-ps=3000 ...`

We do have a convenient data structure to estimate per second metric: `util::SlidingCounter` in 
helio/util/sliding_counter.h that we can use for these metrics.

dranikpg:
Fixes #5549 

Add four new stats to CLIENT LIST

### DIFF
--- a/src/facade/dragonfly_connection.h
+++ b/src/facade/dragonfly_connection.h
@@ -448,7 +448,7 @@ class Connection : public util::Connection {
     size_t read_cnt = 0;                // total number of read calls
     size_t net_bytes_in = 0;            // total number of bytes read
     size_t dispatch_entries_added = 0;  // total number of dispatch queue entries
-    size_t cmds = 0;                    // total number of commands added
+    size_t cmds = 0;                    // total number of commands executed
   } local_stats_;
   ConnectionStats* stats_ = nullptr;
 


src/facade/dragonfly_connection.cc
```diff
@@ -918,6 +920,11 @@ pair<string, string> Connection::GetClientInfoBeforeAfterTid() const {
   absl::StrAppend(&after, " age=", now - creation_time_, " idle=", now - last_interaction_);
   string_view phase_name = PHASE_NAMES[phase_];
 
+  absl::StrAppend(&after, " tot-cmds=", local_stats_.cmds,
+                  " tot-net-in=", local_stats_.net_bytes_in,
+                  " tot-read-calls=", local_stats_.read_cnt,
+                  " tot-dispatches=", local_stats_.dispatch_entries_added);
+
```
dranikpg, line 927: First two are from valkey, the other two are our custom ones

dranikpg:
```py
#!/usr/bin/env python3
# Measure difference of total metrics per second for every client
# client_list_per_second.py [host] [port]

import sys
import redis
import time


def build_table(entries: list[dict]) -> dict:
    return {e["id"]: e for e in entries}


def collect(entry: dict, scratch: dict, sign: int):
    for key in scratch:
        scratch[key] += int(entry[key]) * sign


r = redis.Redis(host=sys.argv[1], port=int(sys.argv[2]))

s1 = r.client_list()
time.sleep(1)
s2 = r.client_list()

t1 = build_table(s1)
t2 = build_table(s2)

base_scratch = {k: 0 for k in next(iter(t2.values())).keys() if k.startswith("tot-")}

for entry_id in t1:
    try:
        e1 = t1[entry_id]
        e2 = t2[entry_id]

        scratch = base_scratch.copy()
        collect(e2, scratch, 1)
        collect(e1, scratch, -1)

        print(scratch)
    except Exception as e:
        print(e)
        pass
```

romange (approved):

romange:
Even better:
```py
#!/usr/bin/env python3
# Measure difference of total metrics per second for every client
# client_list_per_second.py [host] [port]

import sys
import redis
import time


def build_table(entries: list[dict]) -> dict:
    return {e["id"]: e for e in entries}


def collect(entry: dict, scratch: dict, sign: int):
    for key in scratch:
        scratch[key] += int(entry[key]) * sign


r = redis.Redis(host=sys.argv[1], port=int(sys.argv[2]))

s1 = r.client_list()
time.sleep(1)
s2 = r.client_list()

t1 = build_table(s1)
t2 = build_table(s2)

base_scratch = {k: 0 for k in next(iter(t2.values())).keys() if k.startswith("tot-")}

result = []
for entry_id in t1:
    try:
        e1 = t1[entry_id]
        e2 = t2[entry_id]

        scratch = base_scratch.copy()
        collect(e2, scratch, 1)
        collect(e1, scratch, -1)
        if scratch["tot-cmds"]:
            result.append((entry_id, scratch))
        
    except Exception as e:
        print(e)
        pass
sorted_result = sorted(result, key=lambda x: x[1]["tot-net-in"], reverse=True)
for entry_id, scratch in sorted_result:
    print(f"{entry_id}: {scratch}")
    
```

PR #5554, opened 2025-07-23
feat(cluster): migrate thread on same shard config

dranikpg:
Enable migrations for shard flows on destination instance.

Benchmarking: 1M small keys 
||before|after|
|-|---|---|
|time|10s|8s|
|cpu (dest)| 100/400% | 60%/400%|

Note that increasing destination throughput increases source node cpu usage

### DIFF
--- a/src/server/cluster/cluster_family.cc
+++ b/src/server/cluster/cluster_family.cc
@@ -977,6 +977,15 @@ void ClusterFamily::DflyMigrateFlow(CmdArgList args, SinkReplyBuilder* builder,
 
   builder->SendOk();
 
+  // Try migrating the connection if we have the same shard configuration
+  if (migration->ShardNum() == shard_set->size()) {
+    DCHECK_LT(shard_id, shard_set->size());
+    if (bool success = cntx->conn()->Migrate(shard_set->pool()->at(shard_id)); !success) {
+      builder->SendError("invalid state");
+      return;
+    }
+  }
+
   migration->StartFlow(shard_id, cntx->conn()->socket());
 }
 
--- a/src/server/cluster/cluster_family.cc
+++ b/src/server/cluster/cluster_family.cc
@@ -978,7 +978,8 @@ void ClusterFamily::DflyMigrateFlow(CmdArgList args, SinkReplyBuilder* builder,
   builder->SendOk();
 
   // Try migrating the connection if we have the same shard configuration
-  if (migration->ShardNum() == shard_set->size()) {
+  if (migration->ShardNum() == shard_set->size() &&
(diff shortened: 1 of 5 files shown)


dranikpg:
👀 

```
I20250723 19:27:07.073460 38716 cluster_family.cc:624] Flushing newly unowned slots: [0, 16381]
*** SIGSEGV received at time=1753288027 on cpu 3 ***
[failure_signal_handler.cc : 345] RAW: Signal 11 raised at PC=0x555b98db0b28 while already in AbslFailureSignalHandler()
[failure_signal_handler.cc : 345] RAW: Signal 11 raised at PC=0x555b98db0b28 while already in AbslFailureSignalHandler()
[failure_signal_handler.cc : 345] RAW: Signal 11 raised at PC=0x555b98db0b28 while already in AbslFailureSignalHandler()
PC: @     0x555b98db0b28  (unknown)  dfly::DbSlice::PerformDeletionAtomic()
zsh: segmentation fault (core dumped)  ./dragonfly --proactor_threads 4 --dbfilename "" --port 6381  --logtostdout

```

We don't capture the prime table pointer by value

tests/dragonfly/cluster_test.py
```diff
@@ -1330,6 +1330,14 @@ async def test_cluster_flushall_during_migration(
 
     assert await nodes[0].client.dbsize() == 0
 
+    # Push config that causes mass async slot deletion on nodes[1]
```
romange, line 1333: what does it test?

src/server/db_slice.cc
```diff
@@ -802,14 +802,16 @@ void DbSlice::FlushSlotsFb(const cluster::SlotSet& slot_ids) {
   uint64_t next_version = 0;
   uint64_t del_count = 0;
 
+  boost::intrusive_ptr<DbTable> table = db_arr_.front();
```
romange, line 806: can you please explain the change here or even better add comments.

src/server/cluster/cluster_family.cc
```diff
@@ -977,6 +977,16 @@ void ClusterFamily::DflyMigrateFlow(CmdArgList args, SinkReplyBuilder* builder,
 
   builder->SendOk();
 
+  // Try migrating the connection if we have the same shard configuration
```
romange, line 980: does this code run on the destination shard?

tests/dragonfly/cluster_test.py
dranikpg: Added a comment

src/server/cluster/cluster_family.cc
dranikpg: Yes. I don't see any reason to comment _right here_ where this code runs as it should be clear from the surrounding code

PR #5560, opened 2025-07-24
core,server: report defrag metrics

abhijat:
The stats collected during defragmentation are reported in this PR. The following counts are reported (summed up over all shards):

* pages scanned 
* pages marked for reallocation (data was moved from these pages because their usage was lower than threshold)
* pages which were full
* pages which were reserved for malloc
* pages where heap mismatch was observed so nothing was done
* pages which were not full but were above the threshold

Additionally per shard, we report a histogram of pages which were not full but above threshold, the histogram is for p50, p90 and p99 (ie 50% were below or equal to X threshold and so on). This can guide the next defragment attempt, if the current attempt did not touch enough pages.

Internally hdr histogram is used to store the histogram, we also use bloom filter (SBF) to collect counts, because accurate counts are not necessary and storing hash sets of `uintptr_t` takes around 8 bytes per item whereas sbf uses much lower amount of memory with a sufficiently small false positive rate.

Scanning/reporting is done only for the interactive `memory defragment` command, for the scheduled task we do not collect stats.


src/core/page_usage_stats.h
```diff
@@ -3,14 +3,60 @@
 //
 
 #pragma once
+
+#include <absl/container/btree_map.h>
 #include <mimalloc/types.h>
 
-#include <vector>
+#include "core/bloom.h"
+#include "server/tx_base.h"
```
dranikpg: what specifically do you need from tx_base? It just creates a two-way dependency

src/core/page_usage_stats.cc
```diff
@@ -14,6 +23,122 @@ mi_page_usage_stats_t mi_heap_page_is_underutilized(mi_heap_t* heap, void* p, fl
...
+  return absl::StrJoin(rows, "\n");
+}
+
+UniquePages::UniquePages()
+    : pages_scanned{MakeSBF()},
+      pages_marked_for_realloc{MakeSBF()},
+      pages_full{MakeSBF()},
+      pages_reserved_for_malloc{MakeSBF()},
+      pages_with_heap_mismatch{MakeSBF()},
+      pages_above_threshold{MakeSBF()} {
```
dranikpg: Can be just a default constructor on FilterWithSize

src/core/page_usage_stats.h
```diff
@@ -3,14 +3,60 @@
...
+  uint64_t pages_reserved_for_malloc{0};
+  uint64_t pages_with_heap_mismatch{0};
+  uint64_t pages_above_threshold{0};
+
+  using ShardUsageSummary = absl::btree_map<uint8_t, uint64_t>;
+  ShardUsageSummary page_usage_hist;
+  absl::btree_map<ShardId, ShardUsageSummary> shard_wide_summary;
+
+  void Merge(CollectedPageStats&& other, ShardId shard_id);
+  static CollectedPageStats Merge(CollectedPageStats* stats, size_t size, float threshold);
```
dranikpg: `Merge(absl::Span<CollectedPageStats>, float...)`

src/core/page_usage_stats.h
```diff
@@ -3,14 +3,60 @@
...
+
+  std::string ToString() const;
+};
+
+struct UniquePages {
```
dranikpg: this type can be hidden in PageUsage

src/core/page_usage_stats.cc
```diff
@@ -14,6 +23,122 @@ mi_page_usage_stats_t mi_heap_page_is_underutilized(mi_heap_t* heap, void* p, fl
...
+
+std::string CollectedPageStats::ToString() const {
+  std::vector<std::string> rows;
+  rows.push_back(absl::StrFormat("Page usage threshold: %f", threshold * 100));
+  rows.push_back(absl::StrFormat("Pages scanned: %d", pages_scanned));
+  rows.push_back(absl::StrFormat("Pages marked for reallocation: %d", pages_marked_for_realloc));
+  rows.push_back(absl::StrFormat("Pages full: %d", pages_full));
+  rows.push_back(absl::StrFormat("Pages reserved for malloc: %d", pages_reserved_for_malloc));
+  rows.push_back(
+      absl::StrFormat("Pages skipped due to heap mismatch: %d", pages_with_heap_mismatch));
```
dranikpg: Could've been made a little simpler:
`absl::StrAppend(&out, "Page scanned: ", pages_scanned, "\n");`

src/server/memory_cmd.cc
```diff
@@ -177,11 +177,20 @@ void MemoryCmd::Run(CmdArgList args) {
     static const float default_threshold =
         absl::GetFlag(FLAGS_mem_defrag_page_utilization_threshold);
     const float threshold = parser.NextOrDefault(default_threshold);
-    shard_set->pool()->AwaitFiberOnAll([threshold](util::ProactorBase*) {
-      if (auto* shard = EngineShard::tlocal(); shard)
-        shard->DoDefrag(CollectPageStats::YES, threshold);
+
+    CollectedPageStats results[shard_set->size()];
```
dranikpg: We usually use heap arrays to not overflow the fiber stack (small) with large structures

src/core/page_usage_stats.h
abhijat: I needed the `ShardId` typedef, but I have changed the map to just use the underlying type now.

src/core/page_usage_stats.cc
abhijat: changed

src/core/page_usage_stats.h
abhijat: changed to use span

src/core/page_usage_stats.h
abhijat: moved to private type

src/core/page_usage_stats.cc
abhijat: changed

src/core/page_usage_stats.h
abhijat: changed it to take vector by rvalue ref to address https://github.com/dragonflydb/dragonfly/pull/5560#discussion_r2243075480

src/server/memory_cmd.cc
abhijat: changed to vector with initial size

romange:
@dranikpg  I just saw this PR. I am surprised it got approved. 
SBF is not the right data-structure to count unique values. The right data structure is HLL - hyper-log-log

1. It's a huge overkill here - it's O(N) in memory in number of pages and of course not very performant.
2. Do we even get duplicated addresses during the defragment run? why deduplication is needed @abhijat ?

abhijat:
> 2\. Do we even get duplicated addresses during the defragment run? why deduplication is needed @abhijat ?

When defragmentation runs it works on pointers to objects. Then from that object pointer the page is looked up. We are counting the addresses of these pages, they do appear multiple times because one page can hold more than one object, and the page can be visited through multiple objects. 

I just confirmed this by adding a log and running `dfly_bench` while running the defrag command:

dfly bench
```
./build-dbg/dfly_bench --test_time 320 --d 10240
```

dragonfly
```
./build-dbg/dragonfly --proactor_threads=1 --logtostderr --dir /tmp/dfly
```

added log
```
diff --git a/src/core/page_usage_stats.cc b/src/core/page_usage_stats.cc
index 3653d8c4..3ead3b2d 100644
--- a/src/core/page_usage_stats.cc
+++ b/src/core/page_usage_stats.cc
@@ -45,6 +45,8 @@ void FilterWithSize::Add(uintptr_t address) {
   const auto s = std::to_string(address);
   if (sbf.Add(s)) {
     size += 1;
+  } else {
+    std::cerr << "address " << address << " already exists\n";
   }
 }
```

sample output
```
address 5182448510880 already exists
address 5182448510880 already exists
address 5183119586944 already exists
address 5183119586944 already exists
address 5182381387232 already exists
address 5182381387232 already exists
address 5182347819744 already exists
address 5182347819744 already exists
address 5182347819744 already exists
address 5182347819744 already exists
```

>  it's O(N) in memory in number of pages

I used bloom filter here to use roughly constant space which wouldn't scale up with number of pages, is it O(N) because of SBF? Would the vanilla bloom filter use constant space?

romange:
Bloom filter as an algorithm is a **SET** of elements. It uses O(N) memory just with very small factor - usually 10 bits per element on average. See https://share.google/aimode/Xr4vGxotniCT9l7nU

It's not the right algorithm to **count**.

abhijat:
Yes, I looked at HLL after reading the comment and it seems like a better fit for this problem as we don't really need membership queries, just unique counts. I will look into changing it.

romange:
Yeah, and you can use `src/redis/hyperloglog.h` that we already have

abhijat:
I have created https://github.com/dragonflydb/dragonfly/pull/5784 for the change

PR #5563, opened 2025-07-24
fix(tiering): Fix bug in TryGrow

### LINKED ISSUE #5484: test_full_sync failing
test_full_sync is failing:
```
 replica_info = await replica_client.info("ALL")
>           pytest.fail(
                f"Replica did not sync in time. \nmaster: {master_info} \n\nreplica: {replica_info}"
            )
E           Failed: Replica did not sync in time. 
```

https://github.com/dragonflydb/dragonfly/actions/runs/16233894101/job/45841203328

dranikpg:
As simple as that 🤦🏻 I'd add a test, but we already have one that runs one minute+ on average

Fixes #5484 

Reminder after merging: re-enable reg tests

### DIFF
--- a/src/server/tiering/disk_storage.cc
+++ b/src/server/tiering/disk_storage.cc
@@ -180,9 +180,9 @@ std::error_code DiskStorage::Stash(io::Bytes bytes, StashCb cb) {
   if ((available < 256_MB) && (available < capacity * 0.15) && !grow_pending_) {
     auto ec = TryGrow(256_MB);
     LOG_IF(ERROR, ec && ec != errc::file_too_large) << "Could not call grow :" << ec.message();
-    return ec;
   }
-  return {};
+
+  return {};  // Must succeed after the operation was scheduled to run cleanup
 }
 
 DiskStorage::Stats DiskStorage::GetStats() const {


vyavdoshenko:
@dranikpg 
Why don't you want to reenable the test in this PR? I guess this is the best proof of fixing.

dranikpg:
> Why don't you want to reenable the test in this PR? I guess this is the best proof of fixing.

I disabled the tests that run in the backgroun on CI for main branch (that we get notifications for). Doesn't affect the PR in any way

vyavdoshenko (approved):

dranikpg:
I found one more bug (more like serious inefficiency) 🤦🏻

src/server/tiering/disk_storage.cc
```diff
@@ -180,9 +180,9 @@ std::error_code DiskStorage::Stash(io::Bytes bytes, StashCb cb) {
   if ((available < 256_MB) && (available < capacity * 0.15) && !grow_pending_) {
     auto ec = TryGrow(256_MB);
     LOG_IF(ERROR, ec && ec != errc::file_too_large) << "Could not call grow :" << ec.message();
-    return ec;
   }
-  return {};
+
+  return {};  // Must succeed after the operation was scheduled to run cleanup
```
romange, line 185: lgtm but can you explain this comment? :)

vyavdoshenko (approved):
lgtm

src/server/tiering/disk_storage.cc
dranikpg: Ah it's enlgish without punctuation: Must succeed after the operation was scheduled **,** to run cleanup. 

When we returned the error, the callback was never called, so we never cleaned up (returned the buffer)

PR #5565, opened 2025-07-25
feat(tiering): Error propagation to client (Part 2)

dranikpg:
Doesn't assert on most errors now and replies to the client

### DIFF
--- a/src/server/string_family.cc
+++ b/src/server/string_family.cc
@@ -75,7 +75,7 @@ string GetString(const PrimeValue& pv) {
   return res;
 }
 
-StringResult ReadST(DbIndex dbid, string_view key, const PrimeValue& pv, EngineShard* es) {
+StringResult ReadString(DbIndex dbid, string_view key, const PrimeValue& pv, EngineShard* es) {
   return pv.IsExternal() ? StringResult{es->tiered_storage()->Read(dbid, key, pv)}
                          : StringResult{GetString(pv)};
 }
@@ -101,7 +101,7 @@ class SetCmd {
     uint16_t flags = SET_ALWAYS;
     uint32_t memcache_flags = 0;
     uint64_t expire_after_ms = 0;  // Relative value based on now. 0 means no expiration.
-    optional<StringResult>* prev_val = nullptr;  // If set, previous value is stored at pointer
+    optional<StringResult>* prev_val = nullptr;  // if set, previous value will be stored if found
 
     constexpr bool IsConditionalSet() const {
       return flags & SET_IF_NOTEXIST || flags & SET_IF_EXISTS;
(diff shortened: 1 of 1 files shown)


dranikpg:
I'd like to add a sophisticated test, but I've spent too much time already fighting dmsetup. Will do it later

src/server/string_family.cc
```diff
@@ -53,30 +53,31 @@ enum class ExpT { EX, PX, EXAT, PXAT };
...
 constexpr uint32_t kMaxStrLen = 1 << 28;
 
-// Stores a string, the pending result of a tiered read or nothing
-struct StringValue {
-  StringValue() : v_{std::monostate{}} {
-  }
-
-  StringValue(std::string s) : v_{std::move(s)} {
-  }
+template <typename T> using TResultOrT = variant<T, TieredStorage::TResult<T>>;
```
romange, line 57: Can you please add a comment explaining the reason behind `TResultOrT` ?

romange: my question is - why we can not use `TResult` as is?

dranikpg: > // Either immediately available value or tiering future + result

dranikpg: Because we don't want to allocate a future for passing an immediately available value. Sure, we can make the future support immediate values, but: 
* it also has a io::result inside in all cases
* wont be much cleaner then differentiating outside

src/server/string_family.cc
```diff
@@ -53,30 +53,32 @@ enum class ExpT { EX, PX, EXAT, PXAT };
...
-  // Read string from prime value - either from memory or issue tiered storage read
-  static StringValue Read(DbIndex dbid, std::string_view key, const PrimeValue& pv,
-                          EngineShard* es);
+  return res;
+}
 
- private:
-  std::variant<std::monostate, std::string, fb2::Future<io::Result<std::string>>> v_;
-};
+StringResult ReadST(DbIndex dbid, string_view key, const PrimeValue& pv, EngineShard* es) {
```
romange: `ReadST` what does it mean ST?

src/server/string_family.cc
```diff
@@ -98,8 +100,8 @@ class SetCmd {
   struct SetParams {
     uint16_t flags = SET_ALWAYS;
     uint32_t memcache_flags = 0;
-    uint64_t expire_after_ms = 0;     // Relative value based on now. 0 means no expiration.
-    StringValue* prev_val = nullptr;  // If set, previous value is stored at pointer
+    uint64_t expire_after_ms = 0;  // Relative value based on now. 0 means no expiration.
+    optional<StringResult>* prev_val = nullptr;  // If set, previous value is stored at pointer
```
romange: I am guessing nullopt is passed if the previous value does not exist? can you add a comment explaining this?

src/server/string_family.cc
```diff
@@ -127,65 +129,35 @@ class SetCmd {
...
   // For external entries we have to enqueue reads because modify operations like append could be
   // already pending.
   // TODO: Optimize to return co.Size() if no modify operations are present
   if (const auto& co = it_res.value()->second; co.IsExternal()) {
-    fb2::Future<size_t> fut;
-    op_args.shard->tiered_storage()->Read(
-        op_args.db_cntx.db_index, key, co,
-        [fut](io::Result<std::string> s) mutable { fut.Resolve(s->size()); });
+    TieredStorage::TResult<size_t> fut;
+    auto cb = [fut](auto s) mutable { fut.Resolve(s.transform(&string::size)); };
```
romange: what type is `s` here?

src/server/string_family.cc
dranikpg: Yes, I previously allowed `monostate` in StringValue, now I removed it to make all the other code simpler. Updated the comment

PR #5568, opened 2025-07-25
server: Check result of `hdr_init`

abhijat:
If histogram init fails, we should crash instead of using a null pointer as a histogram. 

init can fail for non-memory related reasons such as validations on the distance between lowest and highest values.

### DIFF
--- a/src/server/command_registry.cc
+++ b/src/server/command_registry.cc
@@ -135,7 +135,9 @@ CommandId::CommandId(const char* name, uint32_t mask, int8_t arity, int8_t first
                         acl_categories.value_or(ImplicitAclCategories(mask))) {
   implicit_acl_ = !acl_categories.has_value();
   hdr_histogram* hist = nullptr;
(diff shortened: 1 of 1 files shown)


src/server/command_registry.cc
```diff
@@ -135,7 +135,9 @@ CommandId::CommandId(const char* name, uint32_t mask, int8_t arity, int8_t first
                         acl_categories.value_or(ImplicitAclCategories(mask))) {
   implicit_acl_ = !acl_categories.has_value();
   hdr_histogram* hist = nullptr;
-  hdr_init(kLatencyHistogramMinValue, kLatencyHistogramMaxValue, kLatencyHistogramPrecision, &hist);
+  const auto init_result = hdr_init(kLatencyHistogramMinValue, kLatencyHistogramMaxValue,
```
romange: I prefer we do not  use auto for POD/simple types. This hurts readability.

abhijat: changed to int

PR #5574, opened 2025-07-28
chore: Don't use point_in_time_replication = False in test_replication_all

mkaruza:
Setting `point_in_time_replication` to `False` can result in failing test when master and replica have different values.


mkaruza:
@kostasrim yes

PR #5576, opened 2025-07-28
fix(server): Make max_memory_flag atomic

dranikpg:
Used to be just a global mutable (due to config set) variable. Either like this or thread-local

Part of #5378


kostasrim (approved):
lgtm

PR #5577, opened 2025-07-28
feat: remove crc calculation for migration

BorysTheDev:
Problem: During the cluster migration process, we waste CPU resources on CRC calculation
Fix: ignore CRC during the  migration process

### DIFF
--- a/src/server/rdb_save.cc
+++ b/src/server/rdb_save.cc
@@ -828,14 +828,15 @@ VersionBuffer MakeRdbVersion() {
   return buf;
 }
 
-CrcBuffer MakeCheckSum(std::string_view dump_res) {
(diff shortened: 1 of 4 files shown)


src/server/rdb_save.cc
```diff
@@ -848,12 +849,12 @@ void AppendFooter(io::StringSink* dump_res) {
    */
   const auto ver = MakeRdbVersion();
   dump_res->Write(to_bytes(ver));
-  const auto crc = MakeCheckSum(dump_res->str());
+  const auto crc = MakeCheckSum(dump_res->str(), ignore_crc);
```
dranikpg, line 852: maybe it woud'be been cleaner to write branch here and write an empty CrcBuffer{}

BorysTheDev: I thought about it, but we create a buffer inside MakeCheckSum, so I decided that it is safer to do like I did

Issue #5571, opened 2025-07-27
PSYNC transitions should not be visible by clients


romange:
At a high level, similar to a tree falling in a forest without a sound, Dragonfly should transition from online-partial-sync-online without exposing anything to clients. It should not switch to LOADING unless a psync fails, at which point it should go through a full sync.

It is ok, to expose the partial sync state in `INFO REPLICATION`  (if valkey does it as well) but besides this - all reads should go through without disruptions - at high level there should be no difference between a replica doing partial sync and a replica getting its updates with a bit of delay. Similarly, the replica node should state in a cluster as healthy and active, it's just stale "for a second".

PR #5579, opened 2025-07-28
chore: do not enter loading state during partial sync

kostasrim:
Skips loading state for partial sync

resolves https://github.com/dragonflydb/dragonfly/issues/5571

### DIFF
--- a/src/server/replica.cc
+++ b/src/server/replica.cc
@@ -533,19 +533,16 @@ error_code Replica::InitiateDflySync(std::optional<LastMasterSyncData> last_mast
 
   RETURN_ON_ERR(exec_st_.SwitchErrorHandler(std::move(err_handler)));
 
-  // Make sure we're in LOADING state.
-  if (!service_.RequestLoadingState()) {
-    return exec_st_.ReportError(std::make_error_code(errc::state_not_recoverable),
-                                "Failed to enter LOADING state");
-  }
-
   // Start full sync flows.
   state_mask_ |= R_SYNCING;
 
-  absl::Cleanup cleanup = [this]() {
+  bool loading_state = false;
+  absl::Cleanup cleanup = [this, &loading_state]() {
     // We do the following operations regardless of outcome.
     JoinDflyFlows();
-    service_.RemoveLoadingState();
+    if (loading_state) {
+      service_.RemoveLoadingState();
+    }
     state_mask_ &= ~R_SYNCING;
     last_journal_LSNs_.reset();
   };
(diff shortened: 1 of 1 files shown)


src/server/replica.cc
```diff
@@ -587,6 +584,13 @@ error_code Replica::InitiateDflySync(std::optional<LastMasterSyncData> last_mast
         std::accumulate(is_full_sync.get(), is_full_sync.get() + num_df_flows, 0);
 
     if (num_full_flows == num_df_flows) {
+      // Make sure we're in LOADING state.
+      if (!service_.RequestLoadingState()) {
```
kostasrim, line 594: Should be safe to do, journal changes loaded via the snapshot are executed via `DispatchCommand` and incremental snapshot only adds journal changes to the snapshot.

src/server/replica.cc
```diff
@@ -533,19 +533,16 @@ error_code Replica::InitiateDflySync(std::optional<LastMasterSyncData> last_mast
...
 
-  absl::Cleanup cleanup = [this]() {
+  bool loading_state = false;
+  absl::Cleanup cleanup = [this, &loading_state]() {
     // We do the following operations regardless of outcome.
     JoinDflyFlows();
-    service_.RemoveLoadingState();
+    if (loading_state) {
+      service_.RemoveLoadingState();
+    }
```
dranikpg: I was thinking if we need one more variable - technically it looks like `loading_state = (phase == "full")`. We never abort early on errors up to the state switch point

src/server/replica.cc
```diff
@@ -533,24 +533,27 @@ error_code Replica::InitiateDflySync(std::optional<LastMasterSyncData> last_mast
...
-    service_.RemoveLoadingState();
+    if (sync_type == "full") {
+      service_.RemoveLoadingState();
+    } else if (service_.IsLoadingState()) {
+      // We need this check. We originally set the state unconditionally to LOADING
+      // when we call ReplicaOf command. If for some reason we fail to start full sync below
+      // or cancel the context, we still need to switch to ACTIVE state.
+      // TODO(kostasrim) we can remove this once my proposed changes for replication move forward
+      // as the state transitions for ReplicaOf command will be much clearer.
+      service_.SwitchState(GlobalState::LOADING, GlobalState::ACTIVE);
```
kostasrim, line 551: Imo the different behaviours and the dynamic of `RequestLoadingState` vs `SwitchState()` is not great. For example, I can't replace `SwitchState` with `RequestLoadingState` in `ReplicaOfInternal` because we will need to call `RemoveLoadingState` twice :facepalm: (because of how the internal counter works).

We won't need any of this once we change the algorithm in ReplicaOf command. I already have a proposal/ design doc which was reviwed.

dranikpg: > we still need to switch to ACTIVE state

And nothing else can be in loading state, right?

dranikpg: https://github.com/dragonflydb/dragonfly/pull/2748

src/server/main_service.h
```diff
@@ -80,6 +80,10 @@ class Service : public facade::ServiceInterface {
   bool RequestLoadingState() ABSL_LOCKS_EXCLUDED(mu_);
   void RemoveLoadingState() ABSL_LOCKS_EXCLUDED(mu_);
 
+  // Return true if state is LOADING and loading_state_counter_ == 0, that is,
+  // if no multiple operations require LOADING_STATE at the same time.
+  bool IsLoadingState() ABSL_LOCKS_EXCLUDED(mu_);
```
dranikpg: I get the meaning but the name is really weird. IsLoadingExlusively maybe?

PR #5580, opened 2025-07-28
fix: skip full sync during partial sync

kostasrim:
During `partial sync` from a master node, the replica enters `LOADING STATE` briefly while it loads the snapshot from the missed journal changes.

https://github.com/dragonflydb/dragonfly/pull/5579 solves this by changing when we switch to loading state. 
 
Partial sync however usually requires to stream a few journal changes and stepping through the `FULL_SYNC` phase of replication (creating a snapshot with pure journal changes) is added work (synchronize all the flows to finish incremental full sync, switch to stable sync etc). 

This PR bypasses `FULL_SYNC` alltogether, and allows the replica to enter `STABLE_SYNC` directly. The master node now sends the lsn items that were missed not as part of a `snapshot` but rather as `stale journal entries`. It effectively moves partial sync synchronization to stable sync where it logically belongs to.

improves #5571

### DIFF
--- a/src/server/dflycmd.cc
+++ b/src/server/dflycmd.cc
@@ -322,7 +322,7 @@ void DflyCmd::Flow(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext* cn
 
     if (lsn_to_start_partial) {
       flow.start_partial_sync_at = *lsn_to_start_partial;
-      sync_type = absl::StrCat("PARTIAL", sf_->journal()->GetLsn());
+      sync_type = "PARTIAL";
       VLOG(1) << "Partial sync requested from LSN=" << flow.start_partial_sync_at.value()
               << " and is available. (current_lsn=" << sf_->journal()->GetLsn() << ")";
     }
--- a/src/server/replica.cc
+++ b/src/server/replica.cc
@@ -766,7 +766,6 @@ error_code Replica::ConsumeDflyStream() {
 
   LOG(INFO) << "Transitioned into stable sync";
   // Transition flows into stable sync.
-  std::atomic<size_t> flows_reached_partial = 0;
   {
     auto shard_cb = [&](unsigned index, auto*) {
       const auto& local_ids = thread_flow_map_[index];
@@ -859,12 +858,7 @@ io::Result<bool> DflyShardReplica::StartSyncFlow(
                               CheckRespFirstTypes({RespExpr::STRING, RespExpr::STRING}));
 
   string_view flow_directive = ToSV(LastResponseArgs()[0].GetBuf());
-  // Fetch the LSN to consider partial sync complete
-  if (!(flow_directive.back() == 'L')) {
-    string_view lsn = flow_directive.substr(7);
-    std::ignore = absl::SimpleAtoi(lsn, &lsn_to_finish_partial_);
-    flow_directive = flow_directive.substr(0, 7);
-  }
+
   string eof_token;
   PC_RETURN_ON_BAD_RESPONSE_T(make_unexpected,
                               flow_directive == "FULL" || flow_directive == "PARTIAL");


kostasrim:
We can now remove all the flows from include (and use) `StartIncrementalSnapshotInShard`

src/server/replica.cc
```diff
@@ -825,6 +844,12 @@ io::Result<bool> DflyShardReplica::StartSyncFlow(
                               CheckRespFirstTypes({RespExpr::STRING, RespExpr::STRING}));
 
   string_view flow_directive = ToSV(LastResponseArgs()[0].GetBuf());
+  // Fetch the LSN to consider partial sync complete
+  if (!(flow_directive.back() == 'L')) {
+    string_view lsn = flow_directive.substr(7);
```
kostasrim: I went overboard with this to keep the metrics. IMO we can slightly change the definition of `partial_success` to fit logically with the idea that partial sync `is stable sync` just with a `few stale` journal changes. So, if we skipped `full_sync` it means partial sync is `successful` even though we are still `lagging behind because the journal changes did not reach us`.

kostasrim: In fact if we follow this, we can remove a few lines :)

dranikpg:
Remind me please:
1. Do we still strive for compatibility with replication with Redis?
2. There's code in rdb_load that can be removed now... or not? 🤔

romange:
Of course we should be able to load from Redis

kostasrim:
> Remind me please:
> 
> 1. Do we still strive for compatibility with replication with Redis?
> 2. There's code in rdb_load that can be removed now... or not? 🤔

1. We don't break any compatibility with Redis. We have different flows for handling a master node that is redis see (`InitiatePSync` and `ConsumeRedisStream`). 
2. In fact Redis should be doing the same thing - skipping full sync and replaying only the journal changes without any additional snapshot or full sync (from Redis: `only incremental replication commands are sent based on the replica's last known offset.`)
3. Yes we can remove code in rdb_save :smile: -- will follow up

.pre-commit-config.yaml
```diff
@@ -31,8 +31,3 @@ repos:
     rev: 25.1.0
     hooks:
       - id: black
-
```
kostasrim, line 34: I had enough of this. False positive after false positive `every time` I merge main to my branch

romange:
Please rebase and make sure the tests pass @kostasrim

kostasrim:
> Please rebase and make sure the tests pass @kostasrim

The tests will pass as soon as  #5727 gets merged. That's why I didn't rebase, this PR is blocked on it

romange:
please rebase

src/server/dflycmd.cc
```diff
@@ -390,10 +403,15 @@ void DflyCmd::StartStable(CmdArgList args, Transaction* tx, RedisReplyBuilder* r
     auto cb = [this, &status, replica_ptr = replica_ptr](EngineShard* shard) {
       FlowInfo* flow = &replica_ptr->flows[shard->shard_id()];
 
-      status = StopFullSyncInThread(flow, &replica_ptr->exec_st, shard);
-      if (*status != OpStatus::OK) {
-        return;
+      // We are doing partial sync. We never started FullSync so we don't need to stop it.
+      bool is_partial = flow->start_partial_sync_at.has_value();
+      if (!is_partial) {
+        status = StopFullSyncInThread(flow, &replica_ptr->exec_st, shard);
```
romange, line 409: what happens when if `!is_partial` ?
1. you run `StopFullSyncInThread`  for that thread only
2. then you run `StartStableSyncInThread` ? 

how this flow works?

src/server/journal/streamer.cc
```diff
@@ -110,13 +110,15 @@ void LogTcpSocketDiagnostics(util::FiberSocketBase* dest) {
 }  // namespace
 
 JournalStreamer::JournalStreamer(journal::Journal* journal, ExecutionState* cntx, SendLsn send_lsn,
-                                 bool is_stable_sync)
+                                 bool is_stable_sync,
```
romange: when each PR adds yet another argument, this quickly becomes unmanageable. 
There are two cases: JournalStreamer and RestoreStreamer. send_lsn and is_stable_sync could be collapsed.
what about partial_sync_lsn ? is it relevant for RestoreStreamer?

src/server/journal/streamer.cc
```diff
@@ -148,14 +150,19 @@ void JournalStreamer::ConsumeJournalChange(const JournalChangeItem& item) {
 void JournalStreamer::Start(util::FiberSocketBase* dest) {
   CHECK(dest_ == nullptr && dest != nullptr);
   dest_ = dest;
-  journal_cb_id_ = journal_->RegisterOnChange(this);
+  // For partial sync we first need to catch up
```
romange:
```suggestion
  // For partial sync we first catch up from journal replication buffer and only then register.
```

src/server/journal/streamer.cc
```diff
@@ -194,6 +201,36 @@ void JournalStreamer::StartStalledDataWriterFiber() {
...
                                              util::fb2::Done* waiter) {
+  // Same algorithm as SwitchIncrementalFb. The only difference is that we don't sent
+  // the old LSN"s via a snapshot but rather as journal changes.
+  if (partial_sync_lsn_ > 0) {
+    LSN lsn = partial_sync_lsn_;
+    DCHECK_LE(lsn, journal_->GetLsn()) << "The replica tried to sync from the future.";
+
+    LOG(INFO) << "Starting partial sync from lsn: " << lsn;
+    using facade::operator""_KB;
+    auto threshold = std::exchange(replication_dispatch_threshold, 8_KB);
```
romange: why overriding the flag?

src/server/journal/streamer.cc
```diff
@@ -194,6 +201,36 @@ void JournalStreamer::StartStalledDataWriterFiber() {
...
+    // The replica sends the LSN of the next entry is wants to receive.
+    while (cntx_->IsRunning() && journal_->IsLSNInBuffer(lsn)) {
```
romange, line 209: what if the loop stopped because `cntx_->IsRunning()` is false?

src/server/replica.h
```diff
@@ -207,15 +207,20 @@ class DflyShardReplica : public ProtocolClient {
...
                                  std::optional<Replica::LastMasterSyncData> data);
 
+  struct PSyncState {
+    std::atomic<size_t>* flows_reached_partial;
+    size_t total_flows_to_finish_partial = 0;
+    size_t* success;
+  };
   // Transition into stable state mode as dfly flow.
-  std::error_code StartStableSyncFlow(ExecutionState* cntx);
+  std::error_code StartStableSyncFlow(ExecutionState* cntx, PSyncState psync);
```
romange: it is not a great interface:  you push a state by value, some of the variables are output, some are input.

src/server/replica.cc
```diff
@@ -743,13 +750,22 @@ error_code Replica::ConsumeDflyStream() {
   };
   RETURN_ON_ERR(exec_st_.SwitchErrorHandler(std::move(err_handler)));
 
+  size_t total_flows_to_finish_partial = 0;
```
romange, line 762: I think stats managing code needs a rewrite. it's hacks upon hacks, just to make it work.

tests/dragonfly/replication_test.py
```diff
@@ -3213,10 +3213,13 @@ async def test_partial_replication_on_same_source_master_with_replica_lsn_inc(df
     s4_sz = await c_s4.dbsize()
     assert s3_sz == s4_sz
 
-    server3.stop()
-    # Check logs for partial replication
-    lines = server3.find_in_logs(f"Started partial sync with localhost:{server2.port}")
-    assert len(lines) == 1
+
+#    await check_all_replicas_finished([c_s4], c_s2)
```
romange: what happens here?

src/server/dflycmd.cc
kostasrim: This scenario can't happen. Either `all flows` do partial or `none` and this is enforced so we can't really reach a mixed state where some flows observe `is_partial` as `true` and some as`false`. The way it works is:

The MainReplicationFiber (our coordinator), sets up flows and state transitions. When the coordinator initiates full-sync (StartSyncInThread) we sent `DFLY FLOW` for each flow. There are 3 cases here:

1. `num_full_flows == num_df_flows` means we enter full-sync
2. `num_full_flows == 0` means we enter partial-sync
3.  none of the two, we submit an error to the context, restart MainReplicationFiber

The coordinator, decides what to do next: If we enter (2) which means `partial sync` then when the replica sents `SendNextPhaseRequest("STARTSTABLE")` all flows in dflycmd will have `bool is_partial = flow->start_partial_sync_at.has_value(); ` set to true.

If we did (1), we enter full sync first and after all full sync cuts are sent, `is_partial = false` 

Either way the semantics are exactly as were before

src/server/journal/streamer.cc
kostasrim: Then it means the context got cancelled and we will exit this loop gracefully and clean up the relevant flows. It's the same as if you had cancelled when the LAG between replica and master was 20 or 50 or 100. No functional difference

src/server/replica.cc
kostasrim: 100% I agree -- I left a comment for this as well

kostasrim: https://github.com/dragonflydb/dragonfly/pull/5580/files#r2237044407

tests/dragonfly/replication_test.py
kostasrim: accidental -- will fix

src/server/journal/streamer.cc
kostasrim: IMO it was a mistake. 1500 is the default MTU size

src/server/journal/streamer.cc
kostasrim: I agree. Will collapse/group them.

>  is it relevant for RestoreStreamer?

No. RestoreStreamer is used only in cluster /migrations and the `partial_sync_lsn` is irrelevant for that context.

src/server/replica.cc
kostasrim: This is the only comment I did not address as it is something I want to follow up on a separate PR.

If you disagree -- I will do it here.

src/server/journal/streamer.cc
romange: but I look at the code and I see we continue with  `journal_cb_id_ = journal_->RegisterOnChange(this);` even if cntx got cancelled. is it ok?

src/server/replica.cc
romange: there is no rush to submit it, so there is no reason to split this change into writing a hack and then fixing it.

src/server/journal/streamer.cc
kostasrim: Oh I now understand your comment. SO your concern is, that we register the callback `after` we cancelled and because of that, the callback will stay registered. 

I think your comment is valid and this is a bug. WIll get to it

kostasrim: @romange 

Three things:

1. I also forgot to add a check and cancel the context if we couldn't replay all the LSN's in time (that is, they became stale by the time we started partial sync). This is the same cancelation as it used to exist in incremental snapshot just pushed at a different time/flow

2. I added a check after the loop to make sure that we don't Register the callback if the context got cancelled.

3. I removed the `psync_success/attempt` metrics. There is no simple way to achieve this -- or at least not one I could think of. Before the changes: if we reached stable sync and the flow was `partial` we knew that partial sync was successful. However now, we enter stable sync directly and stream the missing LSN's so we don't have a notion from the replica point of view that partial sync finished: we just receive LSN's in stable sync.

src/server/replica.cc
```diff
@@ -283,8 +283,8 @@ void Replica::MainReplicationFb(std::optional<LastMasterSyncData> last_master_sy
     else
       ec = ConsumeRedisStream();
 
-    auto state = state_mask_ &= R_ENABLED;
-    if (state & R_ENABLED) {  // replication was not stopped.
+    state_mask_ &= R_ENABLED;
+    if (state_mask_ & R_ENABLED) {  // replication was not stopped.
```
romange, line 287:
```suggestion
    if (state_mask_) {  // replication was not stopped.
```

romange (approved):

PR #5583, opened 2025-07-29
fix: temporary set ignore_crc=false during serialization for backward compatibility

BorysTheDev:
To be backward compatible with older versions during migration, we will still calculate the CRC for RESTORE commands during the serialization process


dranikpg:
The cleanest solution would be it would be version agnostic 😝

PR #5584, opened 2025-07-29
chore: Refactor CmdSerializer

mkaruza:
RestoreStreamer is creating each time when WriteEntry is called CmdSerializer object and if we are calling SerializeRestore we would also create RdbSeralizer object which is expensive. Now CmdSerializer and RbdSerializer are create once and can be used with normal RestorStreamer tasks.

### DIFF
--- a/src/server/journal/cmd_serializer.cc
+++ b/src/server/journal/cmd_serializer.cc
@@ -68,7 +68,7 @@ class CommandAggregator {
 
 CmdSerializer::CmdSerializer(FlushSerialized cb, size_t max_serialization_buffer_size)
     : cb_(std::move(cb)), max_serialization_buffer_size_(max_serialization_buffer_size) {
-  serializer_ = std::make_unique<RdbSerializer>(CompressionMode::NONE);
+  serializer_ = std::make_unique<RdbSerializer>(GetDefaultCompressionMode());
 }
 
 size_t CmdSerializer::SerializeEntry(string_view key, const PrimeValue& pk, const PrimeValue& pv,
--- a/src/server/journal/cmd_serializer.h
+++ b/src/server/journal/cmd_serializer.h
@@ -9,10 +9,11 @@
 #include <string>
 #include <string_view>
 
-#include "server/rdb_save.h"
 #include "server/table.h"
 
 namespace dfly {
+
+class RdbSerializer;
 
 // CmdSerializer serializes DB entries (key+value) into command(s) in RESP format string.
 // Small entries are serialized as RESTORE commands, while bigger ones (see
--- a/src/server/journal/streamer.h
+++ b/src/server/journal/streamer.h
@@ -9,7 +9,6 @@
 #include "server/cluster/slot_set.h"
 #include "server/common.h"
 #include "server/db_slice.h"
-#include "server/journal/cmd_serializer.h"
 #include "server/journal/journal.h"
 #include "server/journal/pending_buf.h"
 #include "server/journal/serializer.h"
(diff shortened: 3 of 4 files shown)


src/server/journal/cmd_serializer.cc
```diff
@@ -213,7 +214,7 @@ void CmdSerializer::SerializeRestore(string_view key, const PrimeValue& pk, cons
   args.push_back(expire_str);
 
   io::StringSink value_dump_sink;
-  SerializerBase::DumpObject(pv, &value_dump_sink, true);
+  SerializerBase::DumpObject(serializer_.get(), pv, &value_dump_sink);
```
romange: please pass `ignore_crc=false` explicitly and in generally, I am really against using default arguments in such cases.

romange: ah, I see you changed the interface

src/server/rdb_save.cc
```diff
@@ -854,29 +854,35 @@ void AppendFooter(io::StringSink* dump_res, bool ignore_crc) {
 }
 }  // namespace
 
-void SerializerBase::DumpObject(const CompactObj& obj, io::StringSink* out, bool ignore_crc) {
+void SerializerBase::DumpObject(RdbSerializer* serializer, const CompactObj& obj,
+                                io::StringSink* out) {
   CompressionMode compression_mode = GetDefaultCompressionMode();
   if (compression_mode != CompressionMode::NONE) {
```
romange: why keeping lines 859-863 here and not move where serializer is created?

BorysTheDev:
please rebase

src/server/journal/cmd_serializer.h
```diff
@@ -9,6 +9,7 @@
 #include <string>
 #include <string_view>
 
+#include "server/rdb_save.h"
```
dranikpg: No need, this header is around as heavy as an airbus a380. We can just forward declare it

src/server/journal/streamer.h
```diff
@@ -9,6 +9,7 @@
 #include "server/cluster/slot_set.h"
 #include "server/common.h"
 #include "server/db_slice.h"
+#include "server/journal/cmd_serializer.h"
```
dranikpg: same here, to not pull headers in headers

BorysTheDev: maybe forward declaration is enough?

src/server/journal/cmd_serializer.cc
```diff
@@ -68,6 +68,7 @@ class CommandAggregator {
 
 CmdSerializer::CmdSerializer(FlushSerialized cb, size_t max_serialization_buffer_size)
     : cb_(std::move(cb)), max_serialization_buffer_size_(max_serialization_buffer_size) {
+  serializer_ = std::make_unique<RdbSerializer>(CompressionMode::NONE);
```
dranikpg: Not sure if we really want none, especially if we optimize further to collect batches

src/server/rdb_save.cc
```diff
@@ -854,27 +854,34 @@ void AppendFooter(io::StringSink* dump_res, bool ignore_crc) {
 }
 }  // namespace
 
-void SerializerBase::DumpObject(const CompactObj& obj, io::StringSink* out, bool ignore_crc) {
-  CompressionMode compression_mode = GetDefaultCompressionMode();
-  if (compression_mode != CompressionMode::NONE) {
-    compression_mode = CompressionMode::SINGLE_ENTRY;
+void SerializerBase::DumpObject(RdbSerializer* serializer, const CompactObj& obj,
+                                io::StringSink* out, bool ignore_crc) {
+  CompressionMode serializer_used_compression_mode = serializer->compression_mode_;
```
mkaruza, line 859: Not sure should we save compression_mode and restore it later after we leave this function?

romange: this function is static I think

PR #5585, opened 2025-07-29
chore: refactor JournalItem

kostasrim:
Notice that after a `Journaltem` is `consumed` from all `JournalConsumerInterfaces` we no longer need the data members `cmd and slot_id` so there is really no good reason to store them in the partial sync buffers.

### DIFF
--- a/src/server/journal/journal_slice.cc
+++ b/src/server/journal/journal_slice.cc
@@ -83,7 +83,7 @@ void JournalSlice::AddLogRecord(const Entry& entry) {
     FiberAtomicGuard fg;
     item.journal_item.lsn = lsn_++;
 
-    // onlyt in streamer
+    // only used by RestoreStreamer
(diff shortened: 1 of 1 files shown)


src/server/journal/journal_slice.cc
```diff
@@ -77,12 +77,13 @@ void JournalSlice::SetFlushMode(bool allow_flush) {
...
-  JournalItem item;
+  JournalChangeItem item;
 
   {
     FiberAtomicGuard fg;
-    item.opcode = entry.opcode;
-    item.lsn = lsn_++;
+    item.journal_item.lsn = lsn_++;
+
+    // onlyt in streamer
```
dranikpg: 🤨 🔎

kostasrim: :eyes:

PR #5587, opened 2025-07-29
feat: fix lua memory deallocation and lua_mem_gc_threshold=0 disable lua force GC calls

BorysTheDev:
problem: incorrect calculation of memory allocation/deallocation in the lua script, because of script is executing in different threads but statistic is thread_local

fix: every script calculates the used memory and statistic updates only in the corresponding thread where the script was get/returned

### DIFF
--- a/src/core/interpreter.cc
+++ b/src/core/interpreter.cc
@@ -1224,10 +1224,10 @@ void InterpreterManager::Return(Interpreter* ir) {
     auto before = steady_clock::now();
     tl_stats().gc_freed_memory += ir->RunGC();
 
-    LOG_EVERY_T(INFO, 10) << "stats_used_bytes: " << tl_stats().used_bytes
-                          << " lua_mem_gc_threshold: " << max_memory_usage
-                          << " force_gc_calls: " << tl_stats().force_gc_calls
-                          << " freed_mem: " << tl_stats().gc_freed_memory;
+    VLOG(2) << "stats_used_bytes: " << tl_stats().used_bytes
+            << " lua_mem_gc_threshold: " << max_memory_usage
+            << " force_gc_calls: " << tl_stats().force_gc_calls
+            << " freed_mem: " << tl_stats().gc_freed_memory;
 
     auto after = steady_clock::now();
     tl_stats().gc_duration_ns += duration_cast<nanoseconds>(after - before).count();
--- a/src/core/interpreter.cc
+++ b/src/core/interpreter.cc
@@ -58,7 +58,7 @@ ABSL_FLAG(LuaGcFlag, luagc, {},
 
 ABSL_FLAG(uint64_t, lua_mem_gc_threshold, 10000000,
           "Specifies Lua interpreter's per thread memory limit in bytes after which the GC will be "
-          "called forcefully.");
+          "called forcefully. 0 value remove forced GC calls");
 
 static bool AbslParseFlag(std::string_view in, LuaGcFlag* flag, std::string* err) {
   if (in.empty()) {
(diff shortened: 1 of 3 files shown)


src/core/interpreter.cc
```diff
@@ -563,18 +563,25 @@ int RedisLogCommand(lua_State* lua) {
 // See https://www.lua.org/manual/5.3/manual.html#lua_Alloc
 void* mimalloc_glue(void* ud, void* ptr, size_t osize, size_t nsize) {
   (void)ud;
+  auto& stats = InterpreterManager::tl_stats();
   if (nsize == 0) {
-    InterpreterManager::tl_stats().used_bytes -= mi_usable_size(ptr);
+    const auto usable_size = mi_usable_size(ptr);
+    LOG_IF(WARNING, stats.used_bytes < usable_size)
```
romange: please always use LOG_EVERY_T for such warnings otherwise you will flood the logs, it's high frequency code.

romange: actually due to overflow it won't happen many times. lgtm

romange (commented):
mi_free_size is implemented as mi_free, did you reproduce the problem?

src/core/interpreter.cc
```diff
@@ -1211,10 +1218,17 @@ void InterpreterManager::Return(Interpreter* ir) {
...
   using namespace chrono;
   ++tl_stats().interpreter_return;
-  if (tl_stats().used_bytes > max_memory_usage) {
+  tl_stats().used_bytes += ir->TakeUsedBytes();
+  if (max_memory_usage != 0 && tl_stats().used_bytes > max_memory_usage) {
     ++tl_stats().force_gc_calls;
     auto before = steady_clock::now();
     tl_stats().gc_freed_memory += ir->RunGC();
+
+    LOG_EVERY_T(INFO, 10) << "stats_used_bytes: " << tl_stats().used_bytes
```
romange: reduce the verbosity to VLOG

src/core/interpreter.h
```diff
@@ -89,6 +89,12 @@ class Interpreter {
   // Add function with sha and body to interpreter.
   AddResult AddFunction(std::string_view sha, std::string_view body, std::string* error);
 
+  int64_t TakeUsedBytes() {
+    int64_t tmp = used_bytes_;
```
romange: `return std::exchange(used_bytes_, 0);`

src/core/interpreter.h
```diff
@@ -150,6 +156,7 @@ class Interpreter {
   unsigned cmd_depth_ = 0;
   RedisFunc redis_func_;
   std::string buffer_;
+  std::int64_t used_bytes_;
```
romange: you never initialize it to 0, how do it even work?

src/core/interpreter_test.cc
```diff
@@ -526,4 +528,54 @@ TEST_F(InterpreterTest, LuaIntOverflow) {
...
+  auto* interpreter = im.Get();
+
+  std::string_view keys[] = {"key1", "key2", "key3", "key4", "key5", "key6", "key7"};
+  interpreter->SetGlobalArray("KEYS", SliceSpan{keys});
+
+  auto cb = [](Interpreter::CallArgs ca) {
+    auto* reply = ca.translator;
+    reply->OnInt(1);
+  };
+  interpreter->SetRedisFunc(cb);
```
romange, line 542: It's a good test but since it's not straightforward to understand - please add comments what this script does

src/core/interpreter_test.cc
```diff
@@ -526,4 +528,54 @@ TEST_F(InterpreterTest, LuaIntOverflow) {
...
+
+  char sha_buf[64];
+  Interpreter::FuncSha1(script, sha_buf);
+  string_view sha{sha_buf, std::strlen(sha_buf)};
+
+  string result;
+  Interpreter::AddResult add_res = interpreter->AddFunction(sha, script, &result);
+  EXPECT_EQ(Interpreter::ADD_OK, add_res);
+
+  Interpreter::RunResult run_res = interpreter->RunFunction(sha, &error_);
```
romange, line 564: what happens here with memory?

src/core/interpreter_test.cc
```diff
@@ -526,4 +528,54 @@ TEST_F(InterpreterTest, LuaIntOverflow) {
...
+
+  // we get it again to call GC in separate thread
+  auto* new_interpreter = im.Get();
+  EXPECT_EQ(interpreter, new_interpreter);
+
+  uint64_t used_bytes_before_gc = InterpreterManager::tl_stats().used_bytes;
+  EXPECT_GE(used_bytes_before_gc, 0);
+
+  std::thread t([&] {
+    intptr_.RunGC();
```
romange: so you run here a different interpreter.  The test design is not clear to me.

src/core/interpreter.h
```diff
@@ -150,6 +154,7 @@ class Interpreter {
   unsigned cmd_depth_ = 0;
   RedisFunc redis_func_;
   std::string buffer_;
+  std::int64_t used_bytes_ = 0;
```
romange:
```suggestion
  int64_t used_bytes_ = 0;
```

romange (commented):
lgtm

src/core/interpreter_test.cc
BorysTheDev: nothing, because we don't exceed max_memory lua_mem_gc_threshold. I will add the comment

src/core/interpreter_test.cc
BorysTheDev: it's the same interpreter. see DCHECK above EXPECT_EQ(interpreter, new_interpreter);

src/core/interpreter_test.cc
```diff
@@ -526,4 +528,70 @@ TEST_F(InterpreterTest, LuaIntOverflow) {
...
+  uint64_t used_bytes = InterpreterManager::tl_stats().used_bytes;
+  EXPECT_GE(used_bytes, 0);
+
+  auto force_gc_calls = InterpreterManager::tl_stats().force_gc_calls;
+  // we need return interpreter to update statistic
+  // force_gc_calls shouldn't be called
+  im.Return(interpreter);
+  EXPECT_EQ(force_gc_calls, InterpreterManager::tl_stats().force_gc_calls);
+
+  // im.Return() update statistic
```
romange: remove

romange (approved):

PR #5588, opened 2025-07-30
chore(cluster): Decommit memory only under flag

dranikpg:
Decommit memory only under flag

### DIFF
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -40,6 +40,8 @@ ABSL_FLAG(double, table_growth_margin, 0.4,
 ABSL_FLAG(std::string, notify_keyspace_events, "",
           "notify-keyspace-events. Only Ex is supported for now");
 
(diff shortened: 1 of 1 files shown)


src/server/db_slice.cc
```diff
@@ -848,10 +850,16 @@ void DbSlice::FlushSlotsFb(const cluster::SlotSet& slot_ids) {
     cursor = next;
     ThisFiber::Yield();
   } while (cursor && etl.gstate() != GlobalState::SHUTTING_DOWN);
+
   VLOG(1) << "FlushSlotsFb del count is: " << del_count;
   UnregisterOnChange(next_version);
 
-  etl.DecommitMemory(ServerState::kDataHeap);
```
romange, line 854: Can you please add some INFO logs here with some basic stats, this will at least help us to establish timeline in case we need to investigate something.

romange (approved):

PR #5589, opened 2025-07-31
build: Enable gcp and aws dependencies in release build

abhijat:
Setting `WITH_GCP` and `WITH_AWS` in the release build enables adding the relevant libraries.


dranikpg:
They are enalbed by default, the problem lies elsewhere

abhijat:
There seems to be a problem with the order, when I ran this locally the value was used before it was defined (where it is set on by default), in the following sequence in the root dragonfly `CMakeLists.txt`:

```
include(third_party)
include(internal)

include_directories(src)
include_directories(helio)

add_subdirectory(helio)
add_subdirectory(src)
```

`third_party` used it but it is enabled later on in `helio`. If the order can be changed then this PR is no longer required

abhijat:
I see there is already https://github.com/romange/helio/pull/442 in helio, so this one is redundant.

PR #5591, opened 2025-07-31
feat: Add LinearSearchMap

BagritsevichStepan:
Implements linear search map, that will be used in search join logic


BagritsevichStepan:
> ```c++
> class LinearMap : public absl::InlinedVector<pair<Key,Value>, N> {
>   iterator find(key) {...} // get iterator by key
>   operator[](key) { 
>      return find(key)->second; 
>    }
> }
> ```
> 
> you don't need anything else, _literally_

Fixed.
I still suggest adding some methods (like insert and emplace) to match the interface commonly used in standard hash maps.

src/core/linear_search_map.h
```diff
@@ -0,0 +1,119 @@
...
+  using Base::begin;
+  using Base::clear;
+  using Base::empty;
+  using Base::end;
+  using Base::reserve;
+  using Base::shrink_to_fit;
+  using Base::size;
+  using Base::operator[];
+  using Base::erase;
+  using Base::resize;
```
dranikpg: should all be in scope because of public inheritance

BagritsevichStepan: it is not working without it

BagritsevichStepan: I found the issue

BagritsevichStepan: Fixed

PR #5592, opened 2025-07-31
feat: Add MapsList

BagritsevichStepan:
MapsList is a compact container for storing a list of maps that all share the same fixed set of keys. It will used in search and join code.

For example:
Before we had in search result:
```cpp
using DocValues  = absl::flat_hash_map<std::string /*field*/, SortableValue /*field value*/>
std::vector<DocValues> docs;
```
The issue was that all DocValues instances stored the same set of field names, resulting in excessive duplication of strings. Since both the number of documents and the number of fields can be large, this led to memory overhead.

Now we can use:
```cpp
MapsList<std::string /*field*/, SortableValue> docs;

//Access some field with:
docs[0]["some_field_name"]
```
And we will not store fields strings for each doc.

### DIFF
--- a/src/core/maps_list.h
+++ b/src/core/maps_list.h
@@ -37,6 +37,9 @@ class MapsList {
 
  public:
   class Map;
+  class ConstMap;
+
+  MapsList() = default;
 
   // Size is number of maps, keys are fixed and shared across all maps.
   MapsList(std::initializer_list<Key> keys, size_t size = 0);
@@ -44,6 +47,7 @@ class MapsList {
   template <typename Iterator> MapsList(Iterator begin, Iterator end, size_t size = 0);
 
   Map operator[](size_t index);
+  ConstMap operator[](size_t index) const;
   Map CreateNewMap();
 
   size_t Size() const;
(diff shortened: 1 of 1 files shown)


src/core/maps_list.h
```diff
@@ -0,0 +1,214 @@
...
+
+   To add a new map, use CreateNewMap(). You can then access values by key using operator[].
+   Lookup is implemented via index mapping (O(1)), and keys are not stored more than once.
+
+   Notes:
+    - Value must be default-constructible.
+    - Accessing a key that was not part of the original key set leads to undefined behavior. */
+template <typename Key, typename Value,
+          typename = std::enable_if_t<std::is_default_constructible_v<Value>>>
+class MapsList {
```
dranikpg, line 33: `using MapsList = LinearMap<key, absl::FixedVector<value>>`

Sure, you can't reference a particular map, but do you really need to? I think you overcomplicate it a lot

BagritsevichStepan: I know that this approach is a possible solution here. However, you still need to implement CreateNewMap here. The key difference is that your design follows a "column"-based layout — think of each map as a row and the keys as columns. So, for each key, you're storing a vector of values across all rows.

But in some scenarios, I only want to work with a single row (i.e., a single map) and pass it directly as an argument to a method like this:
```cpp
void SomeMethod(MapList::Map map);
```
This allows updating the values of just one map using a simple and direct interface.

To support this use case in your column-based layout, you would still need to write all the additional logic. On top of that, my approach has better cache locality — since we iterate over a single vector of values. In contrast, the column-based approach requires iterating over a map and accessing one element from each vector inside it.

BagritsevichStepan: And that was the goal of introducing such a map — to simplify the rest of the code, so we don't have to pass something like this to methods:
```cpp
void SomeMethod(const LinearMap<key, absl::FixedVector<value>>& map, size_t map_index_to_change);
```

BagritsevichStepan:
https://github.com/dragonflydb/dragonfly/pull/5604#issuecomment-3204386017

PR #5595, opened 2025-08-01
core: Minor cleanups in bloom filters

abhijat:
* Make deleted member methods public, so that compilation failure due to these generates clearer errors such as `method is deleted`, rather than `method is private`
* Mark move c-tors noexcept so that they can be used by STL style containers instead of falling back to copying. The move ctors can be marked noexcept as they perform trivial moves.
* Remove some inline qualifiers from functions in anonymous namespace.
* Make some methods const - even though the `Add/Set` methods change state, they do so via pointer, so they qualify to be const as they do not reseat the pointer.

### DIFF
--- a/src/core/bloom.cc
+++ b/src/core/bloom.cc
@@ -19,7 +19,7 @@ using namespace std;
 
 namespace {
 
-inline XXH128_hash_t Hash(string_view str) {
+XXH128_hash_t Hash(string_view str) {
   return XXH3_128bits_withSeed(str.data(), str.size(), 0xc6a4a7935bd1e995ULL);  // murmur2 seed
 }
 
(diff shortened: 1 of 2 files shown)


src/core/bloom.cc
```diff
@@ -100,13 +100,13 @@ bool Bloom::Exists(const uint64_t fp[2]) const {
   return true;
 }
 
-bool Bloom::Add(std::string_view str) {
+bool Bloom::Add(std::string_view str) const {
```
romange: even if it's possible to do - it's wrong, imho. why is it needed?

romange: mutation operations should not be marked as const

abhijat: Yeah I thought it would enable usage in more const contexts but I see it can cause problems due to unintended modifications, removed.

PR #5598, opened 2025-08-01
feat: introduce a CPU budget parameter to control slot migration pacing

romange:
This heuristic should replace the previous flags like --migration_buckets_sleep_usec and --migration_buckets_serialization_threshold which are still present as safeguards.


src/server/db_slice.h
```diff
@@ -192,22 +192,7 @@ class DbSlice {
   };
 
   using Context = DbContext;
-
-  // ChangeReq - describes the change to the table.
-  struct ChangeReq {
```
romange, line 197: moved to table.h to reduce header dependency

BorysTheDev (approved):

BorysTheDev (approved):

Issue #5466, opened 2025-07-09
COMMAND subcommands support


vyavdoshenko:
The following list of subcommands should be implemented:

- COMMAND HELP
https://valkey.io/commands/command-help/

- COMMAND LIST
https://valkey.io/commands/command-list/

guozhihao-224:
Can I do this?

vyavdoshenko:
@guozhihao-224 
Who can stop you? Unfortunately, we don't assign the issues to community members, but you can create and suggest the PR.
Please don't forget to read and follow our development documentation and do a search before you start to find out if someone has already proposed the changes.

guozhihao-224:
> [@guozhihao-224](https://github.com/guozhihao-224) Who can stop you? Unfortunately, we don't assign the issues to community members, but you can create and suggest the PR. Please don't forget to read and follow our development documentation and do a search before you start to find out if someone has already proposed the changes.

Does this mean I can initiate a MR to implement this 'COMMAND subcommands support', or do I have to initiate another PR myself?

vyavdoshenko:
@guozhihao-224 
It means you can make your changes and create a PR. Hopefully, your PR will be reviewed soon.

paarth-k2002:
Hello @vyavdoshenko,

As per the docs COMMAND LIST isn’t supported yet, and COMMAND INFO is currently hardcoded to only accept one argument. Is that single-arg restriction for COMMAND INFO a deliberate design choice, or are you open to a PR to bring it to parity with Valkey/Redis (supporting multiple arguments)?

If no one else is currently working on this, I'd like to take it and implement COMMAND LIST along with the INFO improvements.

vyavdoshenko:
> Hello [@vyavdoshenko](https://github.com/vyavdoshenko),
> 
> As per the docs COMMAND LIST isn’t supported yet, and COMMAND INFO is currently hardcoded to only accept one argument. Is that single-arg restriction for COMMAND INFO a deliberate design choice, or are you open to a PR to bring it to parity with Valkey/Redis (supporting multiple arguments)?
> 

The design is not fixed. This command can be implemented in a way to reach parity with competitors.

> If no one else is currently working on this, I'd like to take it and implement COMMAND LIST along with the INFO improvements.

Please, go ahead. Create a PR, and we will review it.

PR #5601, opened 2025-08-01
feat(server): support command help command

guozhihao-224:
Implementing the 'command info' command https://github.com/dragonflydb/dragonfly/issues/5466

### DIFF
--- a/src/server/main_service.cc
+++ b/src/server/main_service.cc
@@ -2682,6 +2682,21 @@ void Service::Command(CmdArgList args, const CommandContext& cmd_cntx) {
     return rb->SendError("COMMAND DOCS Not Implemented");
   }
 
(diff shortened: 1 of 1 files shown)


src/server/main_service.cc
```diff
@@ -2682,6 +2682,21 @@ void Service::Command(CmdArgList args, const CommandContext& cmd_cntx) {
...
   }
 
+  if (subcmd == "HELP" && sufficient_args) {
+    // Return help information for supported COMMAND subcommands
+    const string_view help[] = {
+        "(no subcommand)",
+        "    Return details about all commands.",
+        "COUNT",
+        "    Return the total number of commands in this server.",
+        "DOCS [<command-name> ...]",
```
romange: but we do not support "DOCS" option at the moment.

PR #5602, opened 2025-08-02
fix(tiering): Fix MacOs build

dranikpg:
Fix macos build w/ tiering

### DIFF
--- a/src/server/tiered_storage.h
+++ b/src/server/tiered_storage.h
@@ -7,15 +7,15 @@
 #include <memory>
 #include <utility>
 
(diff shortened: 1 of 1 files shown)


vyavdoshenko:
To fix the macOS build, I guess you should fix it as well:
src/server/generic_family.cc:547
```
      auto future =
```

Also I guess it's reasonable to embrace in __linux__ the `int get_socket_family(int fd)` function in src/facade/socket_utils.cc

romange (approved):

PR #5604, opened 2025-08-02
feat(search): Add indexes joining

BagritsevichStepan:
Adds indexes joining based on hash algorithm. Later we will integrate this into the `FT.AGGREGATE` command.

~This PR includes #5592 and #5603 . So, real number of lines is not so big :smile:~


src/server/search/index_join.h
```diff
@@ -0,0 +1,62 @@
...
+
+    So, index2 uses field1 and field3 for joins. It also indexed docs key1, key2, key3:
+    IndexEntries will store: [{"key1", {"field1" : value, "field3" : value}},
+                             {"key2", {"field1" : value, "field3" : value}},
+                             {"key3", {"field1" : value, "field3" : value}}]
+*/
+using IndexEntries = MapsList<std::string_view /*field names*/, JoinableValue /*field values*/>;
+using EntriesPerIndex = LinearSearchMap<std::string_view /*index name*/, IndexEntries>;
+using KeysPerIndex = LinearSearchMap<std::string_view /*index name*/,
+                                     absl::InlinedVector<std::string_view, 4> /*keys*/>;
```
dranikpg: It might have been easier to forget about index names at all in this module. 
`EntriesPerIndex` would just be `absl::Span<IndexEntries>`, in fact I wouldn't introduce a type alias at all.
The caller would take care of converting names to indices (lex. sorting for example), no need for LinearSearchMap, no need for name juggling, etc

src/server/search/index_join.cc
```diff
@@ -0,0 +1,149 @@
...
+        JoinWithNewIndex(indexes_entries, joined_entries, index_name, joins[index_name]);
+  }
+
+  auto index_names = indexes_entries.keys();
+  const size_t indexes_count = index_names.size();
+  // Now we have joined entries, we need to build JoinResult
+  JoinResult result(index_names.begin(), index_names.end(), joined_entries.size());
+
+  for (size_t i = 0; i < joined_entries.size(); ++i) {
+    auto result_entry = result[i];
```
dranikpg: See how confusing operator[] returning a proxy object is 🙂 This looks almost like a bug until you look at its type

src/server/search/index_join.cc
```diff
@@ -0,0 +1,149 @@
...
+
+      // Index of joined key in the current index
+      KeyIndex key_index = joined_entries[i][index];
+      // Find key name by the key_index
+      std::string_view key_name = indexes_keys[index_name][key_index];
+
+      // Add key to the result
+      // That means that this key from this index was joined
+      result_entry[index_name] = key_name;
+    }
```
dranikpg: Why should we care about keys so deep internally? We can use `DocId`s to reference an entry in a specific index

src/server/search/index_join.cc
```diff
@@ -0,0 +1,149 @@
...
+
+#include "server/search/index_join.h"
+
+namespace dfly::join {
+
+namespace {
+
+using KeyIndex = size_t;
+// using KeyIndexes = absl::InlinedVector<KeyIndex, 4>;
+using KeyIndexes = std::vector<KeyIndex>;
```
dranikpg: This is actually an important optimization because of the sheer volume of building this set over and over for each entry

src/server/search/index_join.cc
```diff
@@ -0,0 +1,149 @@
...
+     Each entry is vector of indexes, that referce to one key in the index
+     For example, {1, 0, 4} means that key with index 1 in the first index,
+     key with index 0 in the second index and key with index 4 in the third index were joined to
+     single entry. */
+  std::vector<KeyIndexes> joined_entries(first_index_entries.Size(), KeyIndexes(1));
+
+  // At the first step all keys from the first index are joined
+  for (size_t i = 0; i < first_index_entries.Size(); ++i) {
+    joined_entries[i][0] = i;
+  }
```
dranikpg: `CHECK(joins[index_names[0]].empty()) << "base index must be first";`

src/server/search/index_join.cc
```diff
@@ -0,0 +1,149 @@
...
+    values_set.reserve(join_expressions.size());
+
+    // Go over all join expressions and get field values using foreign index and field.
+    for (const auto& join_expression : join_expressions) {
+      auto& index_name = join_expression.foreign_index;
+      auto& field_name = join_expression.foreign_field;
+
+      // Now we need to get value of this field from joined key in this index
+      size_t index = indexes.find_index(index_name);
+      KeyIndex key_index = joined_entry_keys[index];
```
dranikpg: 1. Regarding index names, imagine how often this runs
2. CHECK_LT(index, joined_entry_keys.size()) << "Join order broken";

src/server/search/index_join.cc
```diff
@@ -0,0 +1,149 @@
...
+using KeyIndex = size_t;
+// using KeyIndexes = absl::InlinedVector<KeyIndex, 4>;
+using KeyIndexes = std::vector<KeyIndex>;
+
+// Joins joined_entries with new index entries using join_expressions.
+// It uses hash joining algorithm to find matching entries.
+std::vector<KeyIndexes> JoinWithNewIndex(const EntriesPerIndex& indexes,
+                                         const std::vector<KeyIndexes>& joined_entries,
+                                         std::string_view new_index,
+                                         const JoinExpressionsList& join_expressions) {
```
dranikpg: if you use `absl::Span<const ...>` you can pass any random access container and there's no need for type aliases

src/server/search/index_join.h
```diff
@@ -0,0 +1,62 @@
...
+
+   For example:
+    JOIN index2 ON index2.field1 = other_index.field2 AND index2.field3 = other_index.field4
+
+    So, index2 uses field1 and field3 for joins. It also indexed docs key1, key2, key3:
+    IndexEntries will store: [{"key1", {"field1" : value, "field3" : value}},
+                             {"key2", {"field1" : value, "field3" : value}},
+                             {"key3", {"field1" : value, "field3" : value}}]
+*/
+using IndexEntries = MapsList<std::string_view /*field names*/, JoinableValue /*field values*/>;
```
dranikpg: >  MapsList is a compact container for storing a list of maps that all share the same fixed set of
   keys

Except that they might not intersect at all. Again, is there really need to pull field names all the way through? You can do all kinds of cheap preprocessing that would simplify the process greatly. For example, just using indices

dranikpg (dismissed):
I think by adding index and field names to the game you made your implementation a lot harder. Just by using some preprocessing (not on the data, but on the meta, read O(1)) you can you can simplify your code and get rid of many (costly) conversions. You can also simplify your containers - suddenly there was no need for all the custom wrappers.

I'll approve and its for you to decide if to go ahead if the strings - I'd personally not

BagritsevichStepan:
In the previous PR, I implemented it using indexes, but then switched to strings, thinking it would simplify the logic. However that introduced unnecessary overhead. I have reverted it back, so you can review the changes. MapsList is no longer needed, at least for the join logic.

Also, instead of relying on key names, we decided to use a pair of (shard_id, doc_id). From this information, the key name can be derived when needed

dranikpg (approved):
Much simpler than before

PR #5605, opened 2025-08-02
chore: extend DEBUG TOPK/KEYS to show total samples count

romange:


src/server/debugcmd.cc
```diff
@@ -1331,13 +1335,17 @@ void DebugCmd::Keys(CmdArgList args, facade::SinkReplyBuilder* builder) {
...
-      total_keys.fetch_add(res, memory_order_relaxed);
+      DbSlice::UniqueSampleResult res =
+          cntx_->ns->GetDbSlice(es->shard_id()).StopSampleKeys(cntx_->db_index());
+      uniq_keys.fetch_add(res.unique_keys_count, memory_order_relaxed);
+      total_samples.fetch_add(res.total_samples, memory_order_relaxed);
     });
 
-    return builder->SendLong(total_keys.load());
+    uint64_t arr[2] = {uniq_keys.load(), total_samples.load()};
+    auto* rb = static_cast<RedisReplyBuilder*>(builder);
```
vyavdoshenko, line 1347: This is irrelevant to this PR, but I frequently see this cast. It doesn't look good. Could it be refactor somehow to get rid of it?

vyavdoshenko (dismissed):
LGTM

Is it worth adding a Python test for this command?

src/server/debugcmd.cc
romange: we support memcache reply builder and redis reply builder. redis reply builder has methods that memcache reply builder does not have. We could of course add all the irrelevant methods to the base class and have everything in both instantiations but I think it's even more ugly than downcasting.

romange:
> LGTM
> 
> Is it worth adding a Python test for this command?

Most of DEBUG commands are for the internal use, not for regular users. it's not something that we heavily rely in production.

vyavdoshenko (approved):

PR #5606, opened 2025-08-02
chore: add support for value length distribution sampling

romange:

### DIFF
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -321,6 +321,12 @@ inline void TouchHllIfNeeded(string_view key, uint8_t* hll) {
   }
 }
 
+inline void TouchValuesHistogramIfNeeded(const PrimeValue& pv, base::Histogram* hist) {
+  if (hist) {
+    hist->Add(pv.Size());
+  }
+}
+
 }  // namespace
 
 #define ADD(x) (x) += o.x
@@ -572,6 +578,7 @@ auto DbSlice::FindInternal(const Context& cntx, string_view key, optional<unsign
 
   TouchTopKeysIfNeeded(key, db.top_keys);
   TouchHllIfNeeded(key, db.dense_hll);
+  TouchValuesHistogramIfNeeded(res.it->second, db.sample_values_hist);
 
   if (req_obj_type.has_value() && res.it->second.ObjType() != req_obj_type.value()) {
     events_.misses += miss_weight;
@@ -878,7 +885,7 @@ void DbSlice::FlushDbIndexes(const std::vector<DbIndex>& indexes) {
   bool clear_tiered = owner_->tiered_storage() != nullptr;
 
   if (clear_tiered)
-    ClearOffloadedEntries(indexes, db_arr_);
+    RemoveOffloadedEntriesFromTieredStorage(indexes, db_arr_);
 
   DbTableArray flush_db_arr(db_arr_.size());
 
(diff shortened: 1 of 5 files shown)


src/server/db_slice.cc
```diff
@@ -880,7 +887,7 @@ void DbSlice::FlushDbIndexes(const std::vector<DbIndex>& indexes) {
   bool clear_tiered = owner_->tiered_storage() != nullptr;
 
   if (clear_tiered)
-    ClearOffloadedEntries(indexes, db_arr_);
+    RemoveOffloadedEntriesFromTieredStorage(indexes, db_arr_);
```
romange, line 890: just a rename

src/server/debugcmd.cc
```diff
@@ -1351,6 +1357,37 @@ void DebugCmd::Keys(CmdArgList args, facade::SinkReplyBuilder* builder) {
...
+    shard_set->RunBriefInParallel([&](EngineShard* es) {
+      histograms[es->shard_id()] =
+          cntx_->ns->GetDbSlice(es->shard_id()).StopSampleValues(cntx_->db_index());
+    });
+
+    base::Histogram merged_histogram;
+    for (const auto& hist : histograms) {
+      if (hist) {
+        merged_histogram.Merge(*hist);
+        delete hist;
```
vyavdoshenko: IMHO, this contract is not clear. The object is allocated inside the class and should be deleted outside.
Unfortunately, I can't suggest a better solution, so it's OK. But it looks memory-leakable if this code is reused in the future.

vyavdoshenko (dismissed):

src/server/table.h
```diff
@@ -151,6 +152,7 @@ struct DbTable : boost::intrusive_ref_counter<DbTable, boost::thread_unsafe_coun
     SampleUniqueKeys(const SampleUniqueKeys& other) = delete;
   };
   SampleUniqueKeys* sample_unique_keys = nullptr;
+  base::Histogram* sample_values_hist = nullptr;
```
dranikpg, line 155: If you already `#include` it, you can just optional<>

src/server/debugcmd.cc
dranikpg: just use an optional and return by value (or otherwise use a unique_ptr). After 40 years of c++ there are already some more or less mature options 😆

src/server/table.h
romange: `base::Histogram` does not support move/swapping at the moment.

romange:
My original attempt was just returning by value but then I saw that histogram does not support move semantics.

src/server/db_slice.cc
```diff
@@ -1709,6 +1717,29 @@ auto DbSlice::StopSampleKeys(DbIndex db_ind) -> UniqueSampleResult {
...
+  auto& db = *db_arr_[db_ind];
+  if (!db.sample_values_hist) {
+    LOG(INFO) << "Values sampling not started for db " << db_ind;
+    return {};
+  }
+
+  base::Histogram* hist = db.sample_values_hist;
+  db.sample_values_hist = nullptr;
+
+  return unique_ptr<base::Histogram>{hist};
```
dranikpg: `return std::exchange(db.sample_values_hist, nullptr)`

romange: c'tor is explicit

PR #5614, opened 2025-08-04
feat: use SET instead of RESTORE for strings during slot migrations

BorysTheDev:
Problem: RESTORE uses more resources than SET
fixes: https://github.com/dragonflydb/dragonfly/issues/5610

### DIFF
--- a/src/server/journal/cmd_serializer.cc
+++ b/src/server/journal/cmd_serializer.cc
@@ -95,6 +95,9 @@ size_t CmdSerializer::SerializeEntry(string_view key, const PrimeValue& pk, cons
         use_restore_serialization = false;
         break;
       case OBJ_STRING:
+        commands = SerializeString(key, pv);
+        use_restore_serialization = false;
+        break;
       case OBJ_STREAM:
       case OBJ_JSON:
       case OBJ_SBF:
@@ -203,6 +206,28 @@ size_t CmdSerializer::SerializeList(string_view key, const PrimeValue& pv) {
     return true;
   });
   return commands;
+}
+
+static string GetString(const PrimeValue& pv) {
+  string res;
+  res.resize(pv.Size());
+  pv.GetString(res.data());
+
+  return res;
+}
+
+static string GetExternalString(const PrimeValue& pv) {
+  if (pv.IsCool()) {
+    return GetString(pv.GetCool().record->value);
+  }
+  LOG(FATAL) << "External string not supported yet";
+}
+
+size_t CmdSerializer::SerializeString(string_view key, const PrimeValue& pv) {
+  const auto str = !pv.IsExternal() ? GetString(pv) : GetExternalString(pv);
+  std::string_view args[] = {key, string_view(str)};
+  SerializeCommand("SET", args);
+  return 1;
 }
 
 void CmdSerializer::SerializeRestore(string_view key, const PrimeValue& pk, const PrimeValue& pv,
--- a/src/server/journal/cmd_serializer.h
+++ b/src/server/journal/cmd_serializer.h
@@ -38,6 +38,7 @@ class CmdSerializer {
   size_t SerializeZSet(std::string_view key, const PrimeValue& pv);
   size_t SerializeHash(std::string_view key, const PrimeValue& pv);
   size_t SerializeList(std::string_view key, const PrimeValue& pv);
+  size_t SerializeString(std::string_view key, const PrimeValue& pv);
   void SerializeRestore(std::string_view key, const PrimeValue& pk, const PrimeValue& pv,
                         uint64_t expire_ms);
 


src/server/journal/cmd_serializer.cc
```diff
@@ -205,6 +208,28 @@ size_t CmdSerializer::SerializeList(string_view key, const PrimeValue& pv) {
   return commands;
 }
 
+static string GetString(const PrimeValue& pv) {
```
kostasrim:
```
  1099 void __attribute__((noinline)) CompactObj::GetString(string* res) const {                         
  1100   res->resize(Size());                                                                            
  1101   GetString(res->data());                                                                         
  1102 }
```

You can delete this function

src/server/journal/cmd_serializer.cc
```diff
@@ -205,6 +208,28 @@ size_t CmdSerializer::SerializeList(string_view key, const PrimeValue& pv) {
...
+
+  return res;
+}
+
+static string GetExternalString(const PrimeValue& pv) {
+  if (pv.IsCool()) {
+    return GetString(pv.GetCool().record->value);
+  }
+  LOG(FATAL) << "External string not supported yet";
+}
```
kostasrim: I don't understand. `IsCool` implies `IsExternal`, since the former assert the later. 

Furthermore, the "External string not supported yet" is misleading. I guess you mean that the function was called from a prime value that is not external (that is, it's not on disk but in ram).

Also you can move the function within the unamed namespace and omit the `static`

BorysTheDev: Kostas, I haven't assigned anybody for this PR because I am still working on it. I wanted to run the tests only. Please wait a little, I will add you to review.

I took this code from the RdbSerializer::SaveValue(const PrimeValue& pv) that we have used before.

kostasrim: Oh I now understand. Cool items still exist in memory and on disk and external do not. We do the same in `RdbSerializer::SaveValue`

kostasrim: > Kostas, I haven't assigned anybody for this PR because I 

Uh sorry, saw it open and went str8 to it :rofl:

src/server/journal/cmd_serializer.cc
```diff
@@ -73,36 +73,37 @@ CmdSerializer::CmdSerializer(FlushSerialized cb, size_t max_serialization_buffer
 
 size_t CmdSerializer::SerializeEntry(string_view key, const PrimeValue& pk, const PrimeValue& pv,
                                      uint64_t expire_ms) {
-  // We send RESTORE commands for small objects, or objects we don't support breaking.
+  // We send RESTORE commands objects we don't support breaking.
   bool use_restore_serialization = true;
   size_t commands = 1;
-  if (max_serialization_buffer_size_ > 0 && pv.MallocUsed() > max_serialization_buffer_size_) {
```
BorysTheDev, line 79: I have removed this check, because RESTORE works significantly slower for all these types. On my machine the results more than 30%

src/server/journal/cmd_serializer.cc
```diff
@@ -205,6 +206,21 @@ size_t CmdSerializer::SerializeList(string_view key, const PrimeValue& pv) {
   return commands;
 }
 
+size_t CmdSerializer::SerializeString(string_view key, const PrimeValue& pv) {
+  string str;
+  if (pv.IsExternal()) {
+    if (pv.IsCool()) {
+      pv.GetCool().record->value.GetString(&str);
+    }
+    LOG(FATAL) << "External string not supported yet";
```
romange, line 215: Please open an issue under SSD tiering project, otherwise we will discover this in prod

BorysTheDev: https://github.com/dragonflydb/dragonfly/issues/5616

romange (approved):

src/server/journal/cmd_serializer.cc
kostasrim: What kind of values have you checked ? There is a reason we had this check:

1. For big values, `pv.MallocUsed > max_serialization_buffer_size` we fall back `non` restore serialization because we want to `reduce memory during serialization`.
2. For small values we use restore instead.

We have this code to reduce memory when we serialize the big values.

If you have concerns over the performance of this code, I would suggest to `showcase` different sizes of these types and then `benchmark` and make a case.

This change is way beyond the scope of this PR `use SET instead of RESTORE` and without concrete proof that this actually improves performance for all values passed here this change does not belong here as it can cause more harm (introduce a regression which we release and we are not aware of because we "tested this locally")

kostasrim: plz revert this change

src/server/cluster/outgoing_slot_migration.cc
```diff
@@ -150,20 +150,21 @@ void OutgoingMigration::Finish(GenericError error) {
       next_state = MigrationState::C_FATAL;
     } else {
       next_state = MigrationState::C_ERROR;
+      exec_st_.ReportError(std::move(error));
```
kostasrim, line 153: Why do we need the changes in `outgoung_slot_migration` ? 

How are these related ?

Same question for the rest of the changes found in this file

src/server/journal/cmd_serializer.cc
BorysTheDev: I think you misunderstand it, I didn't start use more memory and there is no reason to use RESTORE for small values it reduces perofrmance and doesn't save memory

src/server/cluster/outgoing_slot_migration.cc
BorysTheDev: The OOM test fails, Andy see it on the cluster and I see it during my fixes, because migration become significantly faster so I have fixed it. see description

src/server/journal/cmd_serializer.cc
```diff
@@ -144,7 +145,7 @@ void CmdSerializer::SerializeExpireIfNeeded(string_view key, uint64_t expire_ms)
     return;
   }
 
-  SerializeCommand("PEXIRE", {key, absl::StrCat(expire_ms)});
+  SerializeCommand("PEXPIREAT", {key, absl::StrCat(expire_ms)});
```
romange, line 148: so it was always a bug, right?

BorysTheDev: yes. this code wasn't used before

src/server/journal/cmd_serializer.cc
romange: RESTORE was always as a "simple" way to serialize the entry. Simpler, not better.  I agree with @kostasrim  that Borys should provide a more detailed explanation than "on machine it runs faster". For example, a repeatable step by step scenario in the PR description would do. 

I agree with @BorysTheDev  that I do not see why RESTORE would be better for small containers compared to the regular serialization.

kostasrim: Ok. I think I can provide a few more helpful comments.

Yes we do `one more extra copy` when we use `Restore`. It happens because:

1, we `serialize to an internal buffer first` when we call `DumpObject`
2. then we `flush it` to a string sink when we call `FlushToSink`.
3. we copy that blob again to a new `StringSink` when we call `SerializeCommand("Restore",...)`

So the first low hanging fruit with this flow is that we do `extra copies`.

Second, from `zstd`:
CSpeed2

and https://github.com/facebook/zstd?tab=readme-ov-file#benchmarks 

Basically, on `level 1` (fast compression) we are expected to reduce a blob by `2.8x`. 

So, we use restore flow for blobs `<64kb`. That means with a 2,8x compression rate we reduce those down to 22KB `per blob`. 

Imagine now, that the whole process, streams `1TB` of data. With `RESTORE` we sent roughly `340mb` over the wire. Without we sent `1TB`. This is a lot data sent over the wire for no good reason.

Therefore, I would explore the following:

1. Fix the extra copy. We don't need to `FlushToSunk` on a `temporary` buffer. We directly instead serialize to the `StringSink` provided by `SerializeCommand`. Now both non and restore flows should have equal number of string copies.
2. Verify the compression rates are close to what claimed.
3. Benchmark. 

But IMO `sending 3x` more data over the wire seems `a lot` -- otherwise what's the use of compression algorithms and binary protocols ? 

The reason we used the strings in the first place was to avoid doubling the memory for `big values`. It was never meant to be a perfomrance optimization as we sent more data over the wire, uncopressed as raw strings.

BorysTheDev: We turned off compression for cluster, because it creates more problem. Also we have the problem with 100% cpu on the target node, and decompression is the problem in this case.

kostasrim: @BorysTheDev wrote that a "certain customer" does not use compression because it created issues ? 

Now that you mention it that's true. Can we recall why ? Maybe during full sync, compression of big blobs is actually problematic as it can `blow out the fiber stack`. However, here we compress `SINGLE_ENTRIES` so I doubt we suffer from this issue in this flow. If that's not the case, plz let me know the actual reason.

romange: we almost never use zstd - too cpu heavy. we used lz4 and now we don't basically we are trying to establish a viable  baseline and lack of compression is not a priority. latency increase of x10 is.

romange (approved):

src/server/journal/cmd_serializer.cc
romange: it was, potentially. for large values

PR #5617, opened 2025-08-04
chore: sort per shard sorted sets for SINTER

### LINKED ISSUE #5590: SINTER on large sets is extremely slow
**Description**
SINTER command on large sets is extremely slow compared to Redis.

**To Reproduce**

1. Insert 600M records into SET1 using `SADD`
2. Insert 4096 records into SET2 using `SADD`
3. Execute SINTER SET1 SET2
4. Observe slow query execution

**Expected behavior**
6-15 ms execution time (tested on Redis and Valkey on the same machine)

**Actual behavior**
200-600 s execution time.

**Environment**
 - OS: ubuntu 22.04
 - Kernel: 5.15.0-144-generic, x86_64
 - Bare Metal
 - Dragonfly Version: 1.31.1

kostasrim:
We ignored the cardinality of the produced shard sets. So, for some inputs we would copy the largest set. If for example a set_a on shard 1 had 600m elements and the second set_b had 5, we would create a hash table of 600m elements instead of 5.

Resolves #5590

### DIFF
--- a/src/server/set_family.cc
+++ b/src/server/set_family.cc
@@ -428,10 +428,10 @@ OpResult<SvArray> InterResultVec(const ResultStringVec& result_vec, unsigned req
   }
 
   // Sort the per shard-sorted sets
-  std::sort(sorted_vec.begin(), sorted_vec.end(),
-            [](const auto* lhs, const auto* rhs) { return lhs->size() < rhs->size(); });
-
   if (!sorted_vec.empty()) {
+    std::sort(sorted_vec.begin(), sorted_vec.end(),
+              [](const auto* lhs, const auto* rhs) { return lhs->size() < rhs->size(); });
+
     for (const string& s : *sorted_vec[0]) {
       uniques.emplace(s, 1);
     }
--- a/src/server/set_family.cc
+++ b/src/server/set_family.cc
@@ -419,22 +419,27 @@ OpResult<SvArray> InterResultVec(const ResultStringVec& result_vec, unsigned req
       return OpStatus::OK;  // empty set.
   }
 
-  bool first = true;
+  std::vector<const StringVec*> sorted_vec;
   for (const auto& res : result_vec) {
     if (res.status() == OpStatus::SKIPPED)
       continue;
-
     DCHECK(res);  // we handled it above.
+    sorted_vec.push_back(&res.value());
+  }
 
-    // I use this awkward 'first' condition instead of table[s]++ deliberately.
-    // I do not want to add keys that I know will not stay in the set.
-    if (first) {
-      for (const string& s : res.value()) {
-        uniques.emplace(s, 1);
-      }
-      first = false;
-    } else {
-      for (const string& s : res.value()) {
+  // Sort the per shard-sorted sets
+  std::sort(sorted_vec.begin(), sorted_vec.end(),
+            [](const auto* lhs, const auto* rhs) { return lhs->size() < rhs->size(); });
+
+  if (!sorted_vec.empty()) {
+    for (const string& s : *sorted_vec[0]) {
+      uniques.emplace(s, 1);
+    }
+    // Remove the smallest
+    sorted_vec.erase(sorted_vec.begin());
+
+    for (const auto& res : sorted_vec) {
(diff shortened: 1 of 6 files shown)


tests/dragonfly/set_test.py
```diff
@@ -23,3 +23,21 @@ async def test_sscan_regression(df_factory: DflyInstanceFactory):
...
+
+    await client.execute_command("DEBUG POPULATE 1 prefix- 5 RAND ELEMENTS 5000 TYPE SET")
+    # add a common element to SINTER on a small set
+    await client.execute_command("SADD prefix-:0 common")
+    # create another key prefix-foo with 3 elements
+    await client.execute_command("SADD prefix-foo bar hello common")
+    await client.execute_command("SINTER prefix-foo prefix-:0")
+
+    res = await client.execute_command("SLOWLOG GET 100")
+    assert res == []
```
kostasrim: Not the best test but works :shrug:

tests/dragonfly/set_test.py
```diff
@@ -23,3 +23,21 @@ async def test_sscan_regression(df_factory: DflyInstanceFactory):
     # Takes 3 seconds
     res = await client.execute_command("SLOWLOG GET 100")
     assert res == []
+
+
+@pytest.mark.asyncio
+async def test_sinter_regression(df_factory: DflyInstanceFactory):
+    df = df_factory.create(proactor_threads=4, shard_round_robin_prefix="prefix-")
```
kostasrim: I forced the largest set to be picked via the `shard_round_robin` flag

dranikpg: I think it would've been easier to just create a c++ test. You could've assured that the largest set lands on the first shard, etc...

dranikpg (dismissed):

tests/dragonfly/set_test.py
kostasrim: yeah it was an overkill IMO

romange: +1 to c++ test. we should not use pytests for testing a command logic.

src/server/set_family_test.cc
```diff
@@ -442,4 +442,30 @@ TEST_F(SetFamilyTest, CheckSetLinkExpiryTransfer) {
   EXPECT_THAT(Run("SCARD key"), IntArg(0));
 }
 
+TEST_F(SetFamilyTest, SetInter_5590) {
+  absl::FlagSaver fs;
+  // Default break num_shards=2 and threads=3
+  // TODO(#5651) fix it
```
kostasrim, line 448: https://github.com/dragonflydb/dragonfly/issues/5651

src/server/set_family.cc
```diff
@@ -419,22 +419,27 @@ OpResult<SvArray> InterResultVec(const ResultStringVec& result_vec, unsigned req
...
-    // I do not want to add keys that I know will not stay in the set.
-    if (first) {
-      for (const string& s : res.value()) {
-        uniques.emplace(s, 1);
-      }
-      first = false;
-    } else {
-      for (const string& s : res.value()) {
+  // Sort the per shard-sorted sets
+  std::sort(sorted_vec.begin(), sorted_vec.end(),
```
romange: pls move sort under `if (!sorted_vec.empty()) {`

romange (commented):
lgtm.

src/server/set_family_test.cc
```diff
@@ -16,7 +16,7 @@ extern "C" {
 #include "redis/zmalloc.h"
 }
 
-ABSL_DECLARE_FLAG(bool, legacy_saddex_keepttl);
+ABSL_DECLARE_FLAG(std::string, shard_round_robin_prefix);
```
romange, line 19: @abhijat do you remember why we need legacy_saddex_keepttl ?

kostasrim: it was not set or used :thinking:

abhijat: It was added based on this comment https://github.com/dragonflydb/dragonfly/issues/4676#issuecomment-2692771383

romange (approved):

PR #5618, opened 2025-08-04
chore: allow replicaof in cluster mode when state is TAKEN_OVER

### LINKED ISSUE #5611: `REPLICAOF` fails after takeover
When running `REPLTAKEOVER` in a cluster on v1.32 from nodes A to B, node A gets stuck in a `TAKEN_OVER` state, meaning we can't configure A as a replica of B with `REPLICAOF`

cc @kostasrim

kostasrim:
`REPLTAKEOVER` in cluster mode does not shut down master. Now a node that is in `TAKEN_OVER_STATE` (i,e, repltakeover) is allowed to become a replica of another node via the `REPLICAOF` command. 

* allow REPLICAOF command in TAKEN_OVER server state and only in cluster mode

Resolves #5611

### DIFF
--- a/src/server/main_service.cc
+++ b/src/server/main_service.cc
@@ -1188,13 +1188,15 @@ std::optional<ErrorReply> Service::VerifyCommandState(const CommandId* cid, CmdA
     case GlobalState::SHUTTING_DOWN:
       allowed_by_state = false;
       break;
-    case GlobalState::TAKEN_OVER:
+    case GlobalState::TAKEN_OVER: {
       // Only PING, admin commands, and all commands via admin connections are allowed
       // we prohibit even read commands, because read commands running in pipeline can take a while
       // to send all data to a client which leads to fail in takeover
+      bool replicaof_in_cluster = cid->name() == "REPLICAOF" && IsClusterEnabled();
       allowed_by_state = dfly_cntx.conn()->IsPrivileged() || (cid->opt_mask() & CO::ADMIN) ||
-                         cid->name() == "PING";
+                         cid->name() == "PING" || replicaof_in_cluster;
       break;
+    }
     default:
       break;
   }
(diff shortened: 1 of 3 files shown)


src/server/main_service.cc
```diff
@@ -1188,13 +1188,15 @@ std::optional<ErrorReply> Service::VerifyCommandState(const CommandId* cid, CmdA
...
       break;
-    case GlobalState::TAKEN_OVER:
+    case GlobalState::TAKEN_OVER: {
       // Only PING, admin commands, and all commands via admin connections are allowed
       // we prohibit even read commands, because read commands running in pipeline can take a while
       // to send all data to a client which leads to fail in takeover
+      bool replicaof_in_cluster = cid->name() == "REPLICAOF" && IsClusterEnabled();
       allowed_by_state = dfly_cntx.conn()->IsPrivileged() || (cid->opt_mask() & CO::ADMIN) ||
-                         cid->name() == "PING";
+                         cid->name() == "PING" || replicaof_in_cluster;
```
dranikpg: Maybe just a command that reverts from TAKEN OVER To ACTIVE would've been easier

dranikpg (dismissed):

src/server/main_service.cc
kostasrim: You mean like a proper command in the registry right ?

I thought about it as well, the reason I chose this was to support a direct change from taken over to replica and since the replica flow flushed the datastore anyway it seemed to be simplest one :shrug:

dranikpg: I though of an ADMIN command and only then I realized that REPLICAOF already is one. Do we really need this condition if we have `cid->opt_mask() & CO::ADMIN` 🤔

romange (dismissed):

PR #5619, opened 2025-08-04
fix(snapshot): Replace sleeps with yield

dranikpg:
Blocked on new background fibers from helio https://github.com/romange/helio/pull/402.

Background tasks should be as cooperative as possible to not increase the latency. We yield after each bucket traversal either by pushing to the channel or by yielding explicitly


src/server/engine_shard.cc
```diff
@@ -538,8 +538,9 @@ void EngineShard::StartPeriodicHeartbeatFiber(util::ProactorBase* pb) {
 
   std::chrono::milliseconds period_ms(*cycle_ms);
 
+  fb2::Fiber::Opts fb_opts{.priority = fb2::FiberPriority::BACKGROUND, .name = "heatbeat"};
   fiber_heartbeat_periodic_ =
-      MakeFiber([this, index = pb->GetPoolIndex(), period_ms, heartbeat]() mutable {
+      fb2::Fiber(fb_opts, [this, index = pb->GetPoolIndex(), period_ms, heartbeat]() mutable {
```
dranikpg: 1. Heartbeat should be background
2. If its background, it doesn't penalize other background fibers on idle
3. We can split it into stages to increase simplicity and have dynamic load on it

src/server/snapshot.cc
```diff
@@ -332,14 +330,16 @@ size_t SliceSnapshot::FlushSerialized(SerializerBase::FlushState flush_state) {
   error_code ec = serializer_->FlushToSink(&sfile, flush_state);
   CHECK(!ec);  // always succeeds
 
+  // Yield after possibly long cpu slice due to compression and serialization
+  if (ThisFiber::Priority() == fb2::FiberPriority::BACKGROUND)
+    ThisFiber::Yield();
+  // TODO: else Sleep() to provide write backpressure in advance?
```
dranikpg: Previously we yielded after the write, now we yield before because that's the end of the "CPU" slice (that the scheduler masures). There shoudn't be any difference as long as writes are not async (not sure why we didn't add that yet) 🤔

src/server/snapshot.h
```diff
@@ -164,8 +164,11 @@ class SliceSnapshot : public journal::JournalConsumerInterface {
 
   // Used for sanity checks.
   bool serialize_bucket_running_ = false;
+
+  std::string snapshot_fb_name_;
```
dranikpg: helio fiber stores a string view, previously it pointed just to a stack placed string 😳

dranikpg:
Will fix heartbeat after https://github.com/romange/helio/pull/449

dranikpg:
image

dranikpg:
I didn't manage to find a "winning" strategy for finding the best scheduler configuration, so I just resorted to finding the best default value 😞 Parameters for one of cloud's snapshots

image

romange:
Please rebase. Is it ready for review @dranikpg ?

dranikpg:
How the hell can this be a merge conflict??

image

src/server/snapshot.h
romange: Are you sure it's needed?

src/server/engine_shard.cc
```diff
@@ -420,11 +422,14 @@ void EngineShard::StartPeriodicHeartbeatFiber(util::ProactorBase* pb) {
     return;
   }
   auto heartbeat = [this]() { Heartbeat(); };
+  auto priority = absl::GetFlag(FLAGS_background_heartbeat) ? fb2::FiberPriority::BACKGROUND
+                                                            : fb2::FiberPriority::NORMAL;
 
   std::chrono::milliseconds period_ms(*cycle_ms);
 
+  fb2::Fiber::Opts fb_opts{.priority = priority, .name = "heatbeat"};
```
romange: please fix fb2::Fiber::Opts::name to be a string.

src/server/engine_shard.cc
```diff
@@ -420,11 +422,14 @@ void EngineShard::StartPeriodicHeartbeatFiber(util::ProactorBase* pb) {
...
   fiber_heartbeat_periodic_ =
-      MakeFiber([this, index = pb->GetPoolIndex(), period_ms, heartbeat]() mutable {
+      fb2::Fiber(fb_opts, [this, index = pb->GetPoolIndex(), period_ms, heartbeat]() mutable {
```
romange: providing a name info fb_opts and then overriding it below is confusing, so let's fix fiber interface.

src/server/engine_shard.cc
```diff
@@ -59,6 +59,8 @@ ABSL_FLAG(double, eviction_memory_budget_threshold, 0.1,
           "Eviction starts when the free memory (including RSS memory) drops below "
           "eviction_memory_budget_threshold * max_memory_limit.");
 
+ABSL_FLAG(bool, background_heartbeat, false, "Whether to run heartbeat as a background fiber");
```
romange, line 62: why do you think we need to convert heartbeat into a background fiber?
I thought we convert  the snapshotting fiber first, see how it actually behaves before starting to expand it to more usecases. Moreover, I would not pick this fiber as my second candidate.

src/server/snapshot.cc
```diff
@@ -105,16 +107,18 @@ void SliceSnapshot::Start(bool stream_journal, SnapshotFlush allow_flush) {
...
-    if (!use_snapshot_version_) {
-      db_slice_->UnregisterOnMoved(moved_cb_id_);
-    }
-    consumer_->Finalize();
-    VLOG(1) << "Serialization peak bytes: " << serializer_->GetSerializationPeakBytes();
-  });
+  snapshot_fb_name_ = absl::StrCat("SliceSnapshot-", ProactorBase::me()->GetPoolIndex());
+  auto prio = use_background_mode_ ? fb2::FiberPriority::BACKGROUND : fb2::FiberPriority::NORMAL;
+  snapshot_fb_ = fb2::Fiber(
+      fb2::Fiber::Opts{.priority = prio, .name = snapshot_fb_name_}, [this, stream_journal] {
```
romange: Fiber is created immediately, so it is ok for a string to be on stack as its contents are copied during fiber creation. 
Having said that, I do think that .name should be converted to string to allow using StrCat without such complexities.

src/server/snapshot.cc
```diff
@@ -337,7 +346,13 @@ size_t SliceSnapshot::FlushSerialized(SerializerBase::FlushState flush_state) {
     return 0;
 
   uint64_t id = rec_id_++;
-  DVLOG(2) << "Pushing " << id;
+
+  if (use_background_mode_) {
```
romange, line 350: why is it important to yield here? why not after the call `consumer_->ConsumeData` below?

src/server/snapshot.cc
```diff
@@ -337,7 +346,13 @@ size_t SliceSnapshot::FlushSerialized(SerializerBase::FlushState flush_state) {
...
+    // Yield after possibly long cpu slice due to compression and serialization
+    if (ThisFiber::Priority() == fb2::FiberPriority::BACKGROUND)
```
romange, line 353: why double check of `use_background_mode_` and also `ThisFiber::Priority() == fb2::FiberPriority::BACKGROUND` ? is not it the same?

src/server/snapshot.cc
dranikpg: Becase ConsumeData might suspend and reset the cpu time taken up by the fiber

src/server/snapshot.cc
dranikpg: > is not it the same?

No. This code path can be called from
* the bucket iteration fiber (a background fiber)
* from a journal handler (backpressure)

See the else just below. Will add more comments

src/server/snapshot.cc
romange: bummer, I thought that background fibers are less sensitive to where yield takes place.

src/server/engine_shard.cc
dranikpg: It prevents the idle optimization for other background fibers, only reason

PR #5620, opened 2025-08-05
[Do not merge] debug: Add logs during defrag

abhijat:
This PR is to create an image to debug the defragmentation code path.

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -236,6 +236,7 @@ pair<void*, bool> DefragStrMap2(StringMap* sm, PageUsage* page_usage) {
 }
 
 pair<void*, bool> DefragListPack(uint8_t* lp, PageUsage* page_usage) {
+  VLOG(1) << "DefragListPack";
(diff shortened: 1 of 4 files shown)


abhijat:
from debug logs it looks like many objects are of type `INT_TAG` for which we do no defragmentation. 

```
$ cat /var/log/dragonfly/dragonfly.INFO | grep -Ev ":444|:1072|:428" | grep -c INT_TAG
83147
```

There are however some `ROBJ_TAG` entries which should have been processed:
```
$ cat /var/log/dragonfly/dragonfly.INFO | grep -Ev ":444|:1072|:428" | grep -c ROBJ_TAG
7377
```

abhijat:
> There are however some `ROBJ_TAG` entries which should have been processed:
> 
> ```
> $ cat /var/log/dragonfly/dragonfly.INFO | grep -Ev ":444|:1072|:428" | grep -c ROBJ_TAG
> 7377
> ```

these objects are all of type `OBJ_LIST` for which we again do not do any defragmentation. so effectively the defrag operation does nothing.

PR #5627, opened 2025-08-06
test(acl): Fix flaky ACL tests with long-running Lua scripts

vyavdoshenko:
- Removed `@pytest.mark.skip("Non deterministic")`
- Added `asyncio.create_task()` + `asyncio.sleep(0.1)` for timing control
- Fixed expected error: `ConnectionError` → `ResponseError` with `NOPERM` pattern
- Updated assertions for partial script execution scenarios

### DIFF
--- a/tests/dragonfly/acl_family_test.py
+++ b/tests/dragonfly/acl_family_test.py
@@ -268,40 +268,49 @@ end
 
 
 @pytest.mark.asyncio
-@pytest.mark.skip("Non deterministic")
 async def test_acl_del_user_while_running_lua_script(df_server):
     client = aioredis.Redis(port=df_server.port)
     await client.execute_command("ACL SETUSER kostas ON >kk +@string +@scripting")
     await client.execute_command("AUTH kostas kk")
     admin_client = aioredis.Redis(port=df_server.port, decode_responses=True)
 
-    with pytest.raises(redis.exceptions.ConnectionError):
-        await asyncio.gather(
-            client.eval(script, 4, "key", "key1", "key2", "key3"),
-            admin_client.execute_command("ACL DELUSER kostas"),
-        )
+    script_task = asyncio.create_task(client.eval(script, 4, "key", "key1", "key2", "key3"))
+
+    await asyncio.sleep(0.1)
+
+    await admin_client.execute_command("ACL DELUSER kostas")
+
+    with pytest.raises(
+        redis.exceptions.ResponseError, match="NOPERM kostas has no ACL permissions"
+    ):
+        await script_task
 
     for i in range(1, 4):
         res = await admin_client.get(f"key{i}")
-        assert res == "100000"
+        if res is not None:
+            assert int(res) <= 100000
 
 
 @pytest.mark.asyncio
-@pytest.mark.skip("Non deterministic")
 async def test_acl_with_long_running_script(df_server):
     client = aioredis.Redis(port=df_server.port)
     await client.execute_command("ACL SETUSER roman ON >yoman +@string +@scripting")
     await client.execute_command("AUTH roman yoman")
     admin_client = aioredis.Redis(port=df_server.port, decode_responses=True)
 
-    await asyncio.gather(
-        client.eval(script, 4, "key", "key1", "key2", "key3"),
-        admin_client.execute_command("ACL SETUSER roman -@string -@scripting"),
-    )
+    script_task = asyncio.create_task(client.eval(script, 4, "key", "key1", "key2", "key3"))
+
+    await asyncio.sleep(0.1)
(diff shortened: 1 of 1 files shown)


tests/dragonfly/acl_family_test.py
```diff
@@ -268,40 +268,49 @@ async def test_acl_deluser(df_server):
...
+    script_task = asyncio.create_task(client.eval(script, 4, "key", "key1", "key2", "key3"))
+
+    await asyncio.sleep(0.1)
+
+    await admin_client.execute_command("ACL DELUSER kostas")
+
+    with pytest.raises(
+        redis.exceptions.ResponseError, match="NOPERM kostas has no ACL permissions"
+    ):
+        await script_task
```
kostasrim: The reason that we are getting `NOPERM` now instead of `ConnectionError` is because by the time we disabled the test we added support for `acl keys`. So even if we enabled the test, its semantics got changed because of the new feature. To get it back we need to add `~*` in the acl list of the user `kostas`:

```
//notice the addition of ~* at the end of the acl list
await client.execute_command("ACL SETUSER kostas ON >kk +@string +@scripting ~*")
```

So now we delete a user while they are running a command! 

The second problem with this test, is the `load`. It's really slow. So I suggest to edit the script above and iterate less times: `for i = 1, 1000 do`. This will also require an `await async.sleep(2)` before the assertion below to give time for the script to execute.

tests/dragonfly/acl_family_test.py
```diff
@@ -268,40 +268,49 @@ async def test_acl_deluser(df_server):
...
 
     for i in range(1, 4):
         res = await admin_client.get(f"key{i}")
-        assert res == "100000"
+        if res is not None:
+            assert int(res) <= 100000
```
kostasrim:
```suggestion
        assert int(res) <= 1000
```

tests/dragonfly/acl_family_test.py
```diff
@@ -268,40 +268,49 @@ async def test_acl_deluser(df_server):
...
     client = aioredis.Redis(port=df_server.port)
     await client.execute_command("ACL SETUSER roman ON >yoman +@string +@scripting")
     await client.execute_command("AUTH roman yoman")
     admin_client = aioredis.Redis(port=df_server.port, decode_responses=True)
 
-    await asyncio.gather(
-        client.eval(script, 4, "key", "key1", "key2", "key3"),
-        admin_client.execute_command("ACL SETUSER roman -@string -@scripting"),
-    )
+    script_task = asyncio.create_task(client.eval(script, 4, "key", "key1", "key2", "key3"))
```
kostasrim: Same here, we need the `~*` to bring back the old test semantics.

What we test here is that we can revoke the acl permissions of a user while they are running a script and this won't affect its execution (IMO kinda a redundant test but it's ok to keep it)

tests/dragonfly/acl_family_test.py
```diff
@@ -268,40 +268,49 @@ async def test_acl_deluser(df_server):
...
 
 @pytest.mark.asyncio
-@pytest.mark.skip("Non deterministic")
 async def test_acl_with_long_running_script(df_server):
     client = aioredis.Redis(port=df_server.port)
     await client.execute_command("ACL SETUSER roman ON >yoman +@string +@scripting")
     await client.execute_command("AUTH roman yoman")
     admin_client = aioredis.Redis(port=df_server.port, decode_responses=True)
 
-    await asyncio.gather(
```
kostasrim: Also I don't remember here is if `asyncio.gather(f1, f2)` preserves argument order. That is, `f1` will execute `before f2`. If that's not the case we should force it. Worth checking the ordering imo -- thinking out loud

tests/dragonfly/acl_family_test.py
vyavdoshenko: - Fixed.
- I lowered the number of keys to 10000. The test is no longer very slow.

tests/dragonfly/acl_family_test.py
vyavdoshenko: Fixed

tests/dragonfly/acl_family_test.py
vyavdoshenko: It guarantees the order of results; they should be in the same order as requests.
It seems it can be the root cause of the instability of this test, because two items should be executed in a single-thread environment and could be switched on the IO executions. There is no guarantee who will be first and last.
I suggest turning them on and looking at them.

kostasrim: sounds good!

kostasrim (approved):
lgtm :cook:

PR #5628, opened 2025-08-06
fix: test_throttle_on_commands_squashing_replies_bytes

### LINKED ISSUE #5548: Python test failed: test_throttle_on_commands_squashing_replies_bytes
https://github.com/dragonflydb/dragonfly/actions/runs/16448103448/job/46485539128

kostasrim:
The issue was that we `LOG_EVERY_T` does not trigger the first time it gets called but rather, this is where it starts the internal timer. So the call to `LOG_EVERY_T` the first time would not print anything. The test took roughly 10seconds to run and was "racing" with `LOG_EVERY_T` (will the timer reach 10 seconds to print or not)? 

The fix is simple, add `DVLOG(5)` and use that in the test.

Furthermore, adjust the test a little bit :)

resolves #5548


romange:
I suggest to refrain from relying on info logs for tests. This is incredibly fragile on my experience. it's better to add an info metric

kostasrim:
> I suggest to refrain from relying on info logs for tests. This is incredibly fragile on my experience. it's better to add an info metric

yes that's true. I already did that with partial sync and I can follow up with this as well. Eventually we might flood the info command and for that maybe we can add a special section which is only printed on demand `info tests` or something

Issue #5612, opened 2025-08-04
CI unit test failed during `geo_family_test`


abhijat:
https://github.com/dragonflydb/dragonfly/actions/runs/16716327312/job/47310423912?pr=5595

```
025-08-04T07:24:12.2192700Z 70: [ RUN      ] GeoFamilyTest.GeoRadius
2025-08-04T07:24:12.2203193Z 70: I20250804 07:24:12.220073 12543 proactor_pool.cc:149] Running 3 io threads
2025-08-04T07:24:12.2446871Z 70: I20250804 07:24:12.244347 12543 test_utils.cc:259] Starting GeoRadius
2025-08-04T07:24:12.2550431Z 70: I20250804 07:24:12.254739 12565 geo_family.cc:829] GeoRadius parser error: syntax error
2025-08-04T07:24:12.2555919Z 70: /__w/dragonfly/dragonfly/src/server/geo_family_test.cc:293: Failure
2025-08-04T07:24:12.2556498Z 70: Value of: resp
2025-08-04T07:24:12.2556772Z 70: Expected: is syntax error
2025-08-04T07:24:12.2557222Z 70:   Actual: e(ERR value is not a valid float) (of type facade::RespExpr), Actual does not contain 'syntax error'
2025-08-04T07:24:12.2557645Z 70: 
2025-08-04T07:24:12.2727472Z 70: I20250804 07:24:12.272445 12543 test_utils.cc:221] Finishing GeoRadius
2025-08-04T07:24:12.2729807Z 70: [  FAILED  ] GeoFamilyTest.GeoRadius (53 ms)
```

kostasrim:
https://github.com/dragonflydb/dragonfly/actions/runs/16751632312/job/47423152123?pr=5621

PR #5629, opened 2025-08-06
fix: UB in georadius

BorysTheDev:
fixes: #5612

problem: dereference empty optional, parser.Error() takes out the error from the parser and leaves nullopt instead of it. The next call returns an empty optional.

fix: use HasError() instead of Error() and remove extra logs

### DIFF
--- a/src/server/geo_family.cc
+++ b/src/server/geo_family.cc
@@ -824,16 +824,14 @@ void GeoFamily::GeoRadius(CmdArgList args, const CommandContext& cmd_cntx) {
       default:
         // If MapNext failed, it means an unknown option was provided or
         // an option requiring an argument was missing its argument.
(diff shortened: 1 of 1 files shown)


vyavdoshenko (approved):
lgtm

romange:
Ok, how come we had an UB instead of asserting on invalid access? can we add more checks in parser code to prevent incorrect access to the object?

romange:
good catch though, it was annoying

romange:
@BorysTheDev  what does it mean though? You acked my comment and merged :)

BorysTheDev:
> @BorysTheDev what does it mean though? You acked my comment and merged :)

I will create a separate PR with refactoring.

dranikpg:
We should rename it to `TakeError()`

PR #5631, opened 2025-08-07
feat: use PXAT flag instead of separate command for strings during migration

BorysTheDev:


src/server/journal/cmd_serializer.cc
```diff
@@ -94,8 +94,9 @@ size_t CmdSerializer::SerializeEntry(string_view key, const PrimeValue& pk, cons
       use_restore_serialization = false;
       break;
     case OBJ_STRING:
-      commands = SerializeString(key, pv);
+      commands = SerializeString(key, pv, expire_ms);
       use_restore_serialization = false;
+      expire_ms = 0;
```
romange, line 100: please add a comment `reset expire_ms to skip it in SerializeExpireIfNeeded`

romange (approved):
controlplane/actions/runs/16804876291/job/47600275022

PR #5633, opened 2025-08-07
fix: Prevent race condition during concurrent SAVE operations

vyavdoshenko:
Fixes: regression crash https://github.com/dragonflydb/dragonfly/actions/runs/16715390320/job/47307749937

This PR resolves a race condition that caused a server crash when multiple SAVE commands were executed in parallel. The fix also includes a new regression test that reliably reproduces the issue and now passes.

The Problem:
Running concurrent SAVE commands could lead to a crash with the error fiber destructor called on a joinable fiber.
This happened because a SAVE operation would start its snapshotting fibers before acquiring the global lock to become the sole active saver. If another SAVE command won the race, the SaveStagesController of the losing command would be destroyed, but its fibers would still be running, causing the crash. A simple fix of joining the fiber in the destructor led to a deadlock, as the fiber would block trying to write to a channel that no longer had a consumer.

The Solution:
The race condition is fixed by splitting the snapshot initialization into two distinct phases:
- Preparation (`Init`): A new method now prepares all necessary resources (e.g., file paths) without starting any fibers.
- Start: The method that launches the snapshotting fibers is now only called after the global save lock has been acquired.

### DIFF
--- a/src/server/detail/save_stages_controller.cc
+++ b/src/server/detail/save_stages_controller.cc
@@ -178,13 +178,15 @@ SaveStagesController::~SaveStagesController() {
   }
 }
 
-std::optional<SaveInfo> SaveStagesController::PreInit() {
+std::optional<SaveInfo> SaveStagesController::Init() {
   if (auto err = BuildFullPath(); err) {
     shared_err_ = err;
     return GetSaveInfo();
   }
 
-  InitResources();
+  snapshots_.resize(use_dfs_format_ ? shard_set->size() + 1 : 1);
+  for (auto& [snapshot, _] : snapshots_)
+    snapshot = make_unique<RdbSnapshot>(fq_threadpool_, snapshot_storage_.get());
 
   return {};
 }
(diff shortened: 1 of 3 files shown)


BorysTheDev:
LGTM

src/server/detail/save_stages_controller.cc
```diff
@@ -171,22 +171,29 @@ SaveStagesController::SaveStagesController(SaveStagesInputs&& inputs)
...
 SaveStagesController::~SaveStagesController() {
+  if (!snapshots_.empty() && snapshots_[0].first) {
+    LOG(INFO) << "Forcefully closing save controller";
+    WaitAllSnapshots();
+    Finalize();
+  }
 }
 
-std::optional<SaveInfo> SaveStagesController::InitResourcesAndStart() {
+std::optional<SaveInfo> SaveStagesController::PreInit() {
```
romange: nit: Rename it to `InitResources` or just `Init`, and move the `InitResources` code here and delete `InitResources` function.

romange (commented):
LGTM

src/server/detail/save_stages_controller.cc
vyavdoshenko: fixed

BorysTheDev (approved):

PR #5634, opened 2025-08-07
fix: ub in GEORADIUSBYMEMBER

kostasrim:
The issue is that we called `std::partial_sort` on a range that's invalid: `begin + count > cont.size()` which triggers `ub`.

* fix ub
* add test


abhijat (approved):

romange:
Good work!

PR #5632, opened 2025-08-07
chore: more pipeline latency coverage

romange:
Partially covers https://github.com/dragonflydb/dragonfly/issues/5630

1. Remove `pipeline_low_bound_stats` and instead introduce `squash_stats_latency_lower_limit` flag
   that filters out latency stats based on the lower limit threshold.
2. Consolidate all the squashing metrics inside multi_command_squasher code.
3. Add `pipeline_queue_wait_duration_seconds`, `pipeline_dispatch_skip_flush_total`, `net_read_yields_total` metrics.
4. Reduce reliance on `ProactorBase::GetMonotonicTimeNs` as it's inaccurate - switch to CycleClock where possible.
5. Finally fix a bug where mismatch config type prevented us from using CONFIG SET and change the settings online.

Issue #5630, opened 2025-08-06
improve tooling for p99 latency investigation


romange:
Currently, `pipeline_low_bound_stats` provides some support for this, but the approach is too hacky. We filter out a priori based on pipeline length.

Instead, we could filter out pipeline requests inside the squasher a posteori based on the latency of squashing hop. This is more direct and gives us the precise population of requests that took a lot of time.

In addition, we could:
1. Gather statistics of their fan out factor (vs general `squash_width_freq_arr`)
2. identify whether the latency is affected mostly by scheduling across multiple shards or something else. 
3. Identify if there is a strong correlation between latency contributing shards and the fact we had to awake them.
4. Test if it is possible to start sending replies before we got everything. For example, wait for the first ordered response to be ready instead of waiting for everything. When the first response arrives there is a chance it also subsequent responses 1..K our of total N. we could send them first without waiting for all the N to be ready.

romange:
done

PR #5635, opened 2025-08-07
chore: more pipeline latency coverage (#5632)

romange:
Partially covers #5630

Add `pipeline_dispatch_flush_duration_seconds` metrics that bridge the gap between pipeline_commands_duration_seconds and cmd_squash_hop_duration_seconds + pipeline_queue_wait_duration_seconds.

In summary,
pipeline_commands_duration_seconds roughly equals to `dragonfly_cmd_squash_hop_duration_seconds + cmd_squash_hop_duration_seconds + pipeline_queue_wait_duration_seconds`.

Also update the local dashboard.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -95,7 +95,7 @@ ABSL_FLAG(bool, migrate_connections, true,
           "they operate. Currently this is only supported for Lua script invocations, and can "
           "happen at most once per connection.");
 
-ABSL_FLAG(uint32_t, max_busy_read_usec, 100,
+ABSL_FLAG(uint32_t, max_busy_read_usec, 200,
           "Maximum time we read and parse from "
           "a socket without yielding. In microseconds.");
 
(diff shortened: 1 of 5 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -1678,7 +1682,7 @@ void Connection::AsyncFiber() {
     // As an optimization, we only yield if the fiber was not suspended since the last dispatch.
     uint64_t cur_epoch = fb2::FiberSwitchEpoch();
     if (dispatch_q_.size() == 1 && cur_epoch == prev_epoch) {
-      ThisFiber::Yield();
+      ThisFiber::SleepFor(20us);
```
dranikpg, line 1685: 🤨 So if you don't squash, your latency of a pipeline will be `num_cmds * (avg_cmd_latency + 20 + sleep_overhead)`

dranikpg: Actually I'm suspicious of the yield as well, if we do it every singlet time

romange: why do we do it every single time?
can you give me the scenario? and btw, this Sleep helped a bit to raise the pipeline avg length in my tests (though they are very synthethic).

PR #5636, opened 2025-08-07
refactor: CmdArgParser::Error()

BorysTheDev:
problem: The Error() method was risky and confusing because it removes the error.

fix: method was renamed to TakeError() and added DCHECK that an error exists.

### DIFF
--- a/src/facade/cmd_arg_parser.h
+++ b/src/facade/cmd_arg_parser.h
@@ -34,6 +34,7 @@ template <auto min, auto max> constexpr bool is_fint<FInt<min, max>> = true;
 // Utility class for easily parsing command options from argument lists.
 struct CmdArgParser {
   enum ErrorType {
+    NO_ERROR,
     OUT_OF_BOUNDS,
     SHORT_OPT_TAIL,
     INVALID_INT,
@@ -45,9 +46,12 @@ struct CmdArgParser {
   };
 
   struct ErrorInfo {
-    int type;
-    size_t index;
+    int type = NO_ERROR;
+    size_t index = 0;
 
+    operator bool() const {
+      return type != ErrorType::NO_ERROR;
+    }
     ErrorReply MakeReply() const;
   };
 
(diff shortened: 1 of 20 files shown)


src/facade/cmd_arg_parser.h
```diff
@@ -276,7 +277,7 @@ struct CmdArgParser {
   size_t cur_i_ = 0;
   CmdArgList args_;
 
-  std::optional<ErrorInfo> error_;
+  ErrorInfo error_;
```
dranikpg, line 280: Whats wrong with the optional? The fact that it can be nullopt? Why can't it be nullopt... because we had a bug? Why use optioinals at all then? 🙂  I mean if you write `->` it _already_ indicates that something suspicious is going on. Somebody made a mistake because the function name was bad, happens. Tests didn't catch it, happens

BorysTheDev: nothing wrong, but we already use status, so i decided it will be simpler

kostasrim: if the caller sees an optional of something it's easier to remember that there might not be a value. Accessing via `->` is a good indicator that there was unchecked access of an optional error. Always returning an error which can be no error is kinda odd tbh

romange: I agree that naming can be improved (TakeError -> TakeStatus?) but I personally do not like chaining optional over enums, I see it as unnecessary complication if we control the enum.

src/facade/cmd_arg_parser_test.cc
```diff
@@ -44,18 +44,18 @@ TEST_F(CmdArgParserTest, BasicTypes) {
...
   auto parser = Make({});
 
   EXPECT_EQ(absl::implicit_cast<string_view>(parser.Next()), ""sv);
 
-  auto err = parser.Error();
+  auto err = parser.TakeError();
   EXPECT_TRUE(err);
-  EXPECT_EQ(err->type, CmdArgParser::OUT_OF_BOUNDS);
-  EXPECT_EQ(err->index, 0);
+  EXPECT_EQ(err.type, CmdArgParser::OUT_OF_BOUNDS);
```
romange, line 57: this PR could be much smaller if you add an operator-> to ErrorInfo that returns itself.

BorysTheDev: I don't like such hacks

PR #5637, opened 2025-08-07
chore: string_view overload for BaseFamilyTest::Run

kostasrim:
It's cumbersome to call `Run()` every time with each command argument separately

### DIFF
--- a/src/server/test_utils.cc
+++ b/src/server/test_utils.cc
@@ -404,6 +404,11 @@ RespExpr BaseFamilyTest::Run(ArgSlice list) {
   return Run(GetId(), list);
 }
 
+RespExpr BaseFamilyTest::Run(std::string_view command) {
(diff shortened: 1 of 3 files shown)


src/server/test_utils.cc
```diff
@@ -404,6 +404,11 @@ RespExpr BaseFamilyTest::Run(ArgSlice list) {
   return Run(GetId(), list);
 }
 
+RespExpr BaseFamilyTest::Run(std::string_view command) {
```
kostasrim, line 407: Was there a reason we didn't support that ? It's so much nicer to just cal `"command arg1 arg2 arg3"`

dranikpg: > Was there a reason we didn't support that

No one did it 🙂  There's a lot that could've been done in the testing framework to make it more user friendly

PR #5638, opened 2025-08-07
chore: add basic range algos

kostasrim:


src/core/std_ranges.h
```diff
@@ -0,0 +1,26 @@
...
+#define WRAPPER(name, algo)                                                \
+  template <typename C, typename... Args> auto name(C&& c, Args... args) { \
+    return algo(c.begin(), c.end(), std::forward<Args>(args)...);          \
+  }
+
+WRAPPER(all_of, std::all_of);
+WRAPPER(sort, std::sort);
+
+}  // namespace detail
+
```
kostasrim, line 22: @dranikpg The idea is to quickly provide function wrappers for all `std::algorithms`. 

That way we can call `rng::algo(cont, predicate)` as seen in the examples. 

WDYT?

kostasrim: ignore the name of the file etc, I raised the pr for a quick discussion (should remove detail as well and just undef the macro...)

romange (commented):
often simplicity can be measured by lines of code removed.
imho, this does not help much.

kostasrim:
> often simplicity can be measured by lines of code removed. imho, this does not help much.

This is not a full PR. We use `std algorithms` all over the code base and with ranges we turn all multi line algorithms to a single liners including simple transformations. That's at least 50% reduction of lines for all use cases of stl algorithms...

Issue #5309, opened 2025-06-16
Unit test failure: `ClusterFamilyTest.FlushSlots`


abhijat:
https://github.com/dragonflydb/dragonfly/actions/runs/15675879494/job/44155931707#step:8:18036

```
[ RUN      ] ClusterFamilyTest.FlushSlots
74: I20250616 08:56:27.345012 17926 proactor_pool.cc:149] Running 3 io threads
74: I20250616 08:56:27.346189 17926 test_utils.cc:259] Starting FlushSlots
74: I20250616 08:56:27.380762 17997 cluster_family.cc:681] Got DFLYCLUSTER FLUSHSLOTS [0,0]
74: ../src/server/cluster/cluster_family_test.cc:727: Failure
74: Value of: RunPrivileged({"dflycluster", "getslotinfo", "slots", "0", "1"})
74: Expected: resp array (value: 320-byte object <4C-3F DE-00 00-00 00-00 59-42 DD-00 00-00 00-00 E2-44 DD-00 00-00 00-00 40-60 49-00 00-00 00-00 D6-44 DD-00 00-00 00-00 02-51 AC-0C 00-00 00-00 F0-40 3E-F2 FF-7F 00-00 00-00 00-00 00-00 00-00 ... 70-42 DD-00 00-00 00-00 02-4A 9B-80 95-27 16-B2 D0-41 3E-F2 FF-7F 00-00 00-00 00-00 00-00 00-00 00-42 3E-F2 FF-7F 00-00 C0-50 AC-0C 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00>)
74:   Actual: [[i0,'key_count',i31,'total_reads',i0,'total_writes',i52,'memory_bytes',i0],[i1,'key_count',i48,'total_reads',i0,'total_writes',i48,'memory_bytes',i0]] (of type facade::RespExpr), whose given property is Vec: [[i0,'key_count',i31,'total_reads',i0,'total_writes',i52,'memory_bytes',i0],[i1,'key_count',i48,'total_reads',i0,'total_writes',i48,'memory_bytes',i0]]
74: , whose element #0 doesn't match, whose given property is Vec: [i0,'key_count',i31,'total_reads',i0,'total_writes',i52,'memory_bytes',i0]
74: , whose element #2 doesn't match, 
74: Actual : 31 expected: 0
74: 
[...]

PR #5639, opened 2025-08-08
fix: FlushSlots test

BorysTheDev:
fixes: #5309

problem: flush_slots is an async operation
```
void DbSlice::FlushSlots(const cluster::SlotRanges& slot_ranges) {
  cluster::SlotSet slot_set(slot_ranges);
  InvalidateSlotWatches(slot_set);
  fb2::Fiber("flush_slots", [this, slot_set = std::move(slot_set)]() mutable {
    FlushSlotsFb(slot_set);
  }).Detach();
}
```

fix: wait a little longer to be sure that operation is finished

### DIFF
--- a/src/server/cluster/cluster_family_test.cc
+++ b/src/server/cluster/cluster_family_test.cc
@@ -723,7 +723,7 @@ TEST_F(ClusterFamilyTest, FlushSlots) {
   ExpectConditionWithinTimeout([&]() {
     return RunPrivileged({"dflycluster", "flushslots", "0", "0"}) == "OK";
   });
-  util::ThisFiber::SleepFor(1ms);
+  util::ThisFiber::SleepFor(10ms);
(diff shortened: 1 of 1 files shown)


src/server/cluster/cluster_family_test.cc
```diff
@@ -723,7 +723,7 @@ TEST_F(ClusterFamilyTest, FlushSlots) {
   ExpectConditionWithinTimeout([&]() {
     return RunPrivileged({"dflycluster", "flushslots", "0", "0"}) == "OK";
   });
-  util::ThisFiber::SleepFor(1ms);
+  util::ThisFiber::SleepFor(10ms);
```
kostasrim, line 726: We do the same operation below so the same thing could happen ? Shouldn't we do the same change on line `735` below ?

kostasrim (dismissed):
lgtm, minor comment.

Also irrelevant with this PR -- detaching a fiber is almost *always* a mistake :smile:

PR #5644, opened 2025-08-10
chore: add more I/O run-time settings

romange:
1. Add settings of how to io_uring wakes up other threads.
2. Consolidate selective stats measurement based on squash_stats_latency_lower_limit at main_service level. Also propogate the result of whether we account this sample for stats or not. This allows us to adjust the stats in the Connection code for the same pipeline squash sample.
3. In #5635 i unconditionally changed Yield to SleepFor(20us) inside pipeline code. In this PR I make it a run-time parameter pipeline_wait_batch_usec to avoid potential regression with this change.

### DIFF
--- a/src/server/main_service.cc
+++ b/src/server/main_service.cc
@@ -1628,7 +1628,7 @@ DispatchManyResult Service::DispatchManyCommands(absl::Span<CmdArgList> args_lis
   auto* ss = dfly::ServerState::tlocal();
   // Don't even start when paused. We can only continue if DispatchTracker is aware of us running.
   if (ss->IsPaused())
-    return {.processed = 0, .account_in_stats = true};
+    return {.processed = 0, .account_in_stats = false};
 
   vector<StoredCmd> stored_cmds;
   intrusive_ptr<Transaction> dist_trans;
@@ -1704,8 +1704,8 @@ DispatchManyResult Service::DispatchManyCommands(absl::Span<CmdArgList> args_lis
     dist_trans->UnlockMulti();
 
   uint64_t total_usec = base::CycleClock::ToUsec(base::CycleClock::Now() - start_cycles);
-  bool account_in_stats = true;
-  if (total_usec > squash_stats_latency_lower_limit_cached) {
+  bool account_in_stats = total_usec > squash_stats_latency_lower_limit_cached;
+  if (account_in_stats) {
     auto* ss = ServerState::tlocal();
     ss->stats.multi_squash_exec_hop_usec += stats.hop_usec;
     ss->stats.multi_squash_exec_reply_usec += stats.reply_usec;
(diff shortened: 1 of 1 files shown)


src/server/main_service.cc
```diff
@@ -1660,7 +1703,19 @@ size_t Service::DispatchManyCommands(absl::Span<CmdArgList> args_list, SinkReply
   if (dist_trans)
     dist_trans->UnlockMulti();
 
-  return dispatched;
+  uint64_t total_usec = base::CycleClock::ToUsec(base::CycleClock::Now() - start_cycles);
+  bool account_in_stats = true;
+  if (total_usec > squash_stats_latency_lower_limit_cached) {
```
dranikpg:
```c++
bool account_in_stats = total_usec > ...
if (account_in_stats) {
```

src/facade/dragonfly_connection.cc
```diff
@@ -192,6 +195,7 @@ thread_local TrafficLogger tl_traffic_logger{};
 thread_local base::Histogram* io_req_size_hist = nullptr;
 
 thread_local const size_t reply_size_limit = absl::GetFlag(FLAGS_squashed_reply_size_limit);
+thread_local uint32 pipeline_wait_batch_usec = absl::GetFlag(FLAGS_pipeline_wait_batch_usec);
```
dranikpg, line 198: I've neved initialized global variables from absl flags and never though if it's safe. This initialization happens before main? But it somehow works...? Also, hey are initialized in the order in which they are _defined_. No tricks in absl? Seems fragile

dranikpg: Also given how rarely this changes, it can be just a global atomic - there won't be any read overhead

dranikpg: The whole absl flag caching scheme is still somewhat blurry. I agree we shouldn't use them directly in hot paths, but we currently do somethimes (see MGET 🤷🏻‍♂️)

src/server/main_service.cc
```diff
@@ -1580,20 +1617,25 @@ DispatchResult Service::InvokeCmd(const CommandId* cid, CmdArgList tail_args,
...
   ConnectionContext* dfly_cntx = static_cast<ConnectionContext*>(cntx);
   DCHECK(!dfly_cntx->conn_state.exec_info.IsRunning());
   DCHECK_EQ(builder->GetProtocol(), Protocol::REDIS);
+  DCHECK_GT(args_list.size(), 1u);
 
   auto* ss = dfly::ServerState::tlocal();
   // Don't even start when paused. We can only continue if DispatchTracker is aware of us running.
   if (ss->IsPaused())
-    return 0;
+    return {.processed = 0, .account_in_stats = true};
```
dranikpg: probably account_in_stats=false 🤔

dranikpg (approved):

src/facade/dragonfly_connection.cc
romange: I've never benchmarked absl read path so I can not comment, I am just repeating what was done before.
Regarding global initialization - strictly speaking it's not global, it's thread local so the initialization happens on thread creation which is done inside main

dranikpg: True, I forgot

PR #5646, opened 2025-08-10
test: skip test_acl_del_user_while_running_lua_script becuase it constantly fails on CI

vyavdoshenko:
Hard to reproduce the failure locally. Let's skip this test for some time. Additional investigation needed.

### DIFF
--- a/tests/dragonfly/acl_family_test.py
+++ b/tests/dragonfly/acl_family_test.py
@@ -268,6 +268,7 @@ end
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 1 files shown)


romange:
This is not non-deterministic tests should be designed in the first place.  You can not fit the timeline of distributed events exactly according to your needs.  You either relax the test so that it fails *ONLY* if Program Under Test does not behave correctly when the desired state is reached (and if it's not reached the test does not fail), or alternatively you add the necessary tooling/hooks into the Program Under Test to reach the desired state in 100% of the time @vyavdoshenko  @kostasrim  FYI

PR #5647, opened 2025-08-11
core: Track objects skipped during defragmentation

abhijat:
Objects which are skipped during defragmentation fall in two categories:

* not required (eg INT_TAG)
* not supported but will require support (eg qlist)

The counts for both are tracked and reported, to make it easier to understand where defragmentation did not have an effect.


src/core/page_usage_stats.h
```diff
@@ -55,6 +57,14 @@ class PageUsage {
 
   bool ConsumePageStats(mi_page_usage_stats_t stats);
 
+  void SkipNotRequired() {
```
romange: This function does not skip anything. Maybe `RecordNotRequired`. same below

romange (approved):

PR #5654, opened 2025-08-11
fix: RoundRobinSharder for all proactors

### LINKED ISSUE #5651: RoundRobinSharder should be thread local to each proactor
We initialize `RoundRobinSharder` per shard thread yet we can call `Shard()` on any thread. This means, that a call to `Shard()` from a non shard thread will read empty values for the thread locals (since they only got initialized for shard threads).

kostasrim:
Function `Shard()` is used both by shard threads and proactors but `RoundRobinSharder` was initialized only on shard threads. The fix is to initialize it for all threads. Furthermore, move `RoundRobinSharder` and `Shard()` to a separate translation unit.

* move RoundRobinSharder and Shard to a separate translation unit
* initialize the thread locals for all proactors

Resolves #5651

### DIFF
--- a/src/server/sharding.cc
+++ b/src/server/sharding.cc
@@ -15,11 +15,7 @@
 
 using namespace std;
 
-ABSL_FLAG(string, shard_round_robin_prefix, "",
-          "When non-empty, keys which start with this prefix are not distributed across shards "
-          "based on their value but instead via round-robin. Use cautiously! This can efficiently "
-          "support up to a few hundreds of prefixes. Note: prefix is looked inside hash tags when "
-          "cluster mode is enabled.");
+ABSL_FLAG(string, shard_round_robin_prefix, "", "Deprecated -- will be removed");
 
 namespace dfly {
 namespace {
@@ -38,6 +34,7 @@ class RoundRobinSharder {
     shard_set_size_ = shard_set_size;
 
     if (IsEnabled()) {
+      LOG(WARNING) << "shard_round_robin_prefix is deprecated and will be removed in new versions";
       // ~100k entries will consume 200kb per thread, and will allow 100 keys with < 2.5% collision
       // probability. Since this has a considerable footprint, we only allocate when enabled. We're
       // using a prime number close to 100k for better utilization.


src/server/sharding.cc
```diff
@@ -0,0 +1,136 @@
...
+    if (round_robin.has_value()) {
+      return *round_robin;
+    }
+  }
+
+  return hash % shard_num;
+}
+
+namespace sharding {
+void InitThreadLocals(uint32_t shard_set_size) {
```
kostasrim, line 128: We now pass the shard_set_size and make it a thread local

src/server/engine_shard.cc
```diff
@@ -97,78 +91,6 @@ ShardMemUsage ReadShardMemUsage(float wasted_ratio) {
...
-                kInvalidSid);
-
-      util::fb2::LockGuard guard(mutex_);
-      if (round_robin_shards_.empty()) {
-        round_robin_shards_ = round_robin_shards_tl_cache_;
-      }
-    }
-  }
-
-  static void Destroy() ABSL_LOCKS_EXCLUDED(mutex_) {
```
kostasrim, line 139: I accidentally deleted this. Will fix soon.

src/server/sharding.cc
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  if (IsClusterShardedByTag()) {
+    v = LockTagOptions::instance().Tag(v);
+  }
+
+  XXH64_hash_t hash = XXH64(v.data(), v.size(), 120577240643ULL);
+
+  if (RoundRobinSharder::IsEnabled()) {
+    auto round_robin = RoundRobinSharder::TryGetShardId(v, hash);
```
kostasrim, line 118: Ouch, contention point for all threads, not even an atomic. Beyond the scope of this PR however.

src/server/sharding.cc
romange: I think RoundRobinSharder can be deleted altogether. We implemented it for something we never used

dranikpg: Not sure "never used" is the right definition. We actually have a flag for it called `shard_round_robin_prefix`

romange: It's not being used in prod and I doubt it's used by community

kostasrim: let's deprecate it first and remove it later ?

romange: sure

src/server/tx_base.cc
```diff
@@ -4,6 +4,8 @@
 
 #include "server/tx_base.h"
 
+#include <xxhash.h>
```
romange, line 7: why this change?

src/server/engine_shard.cc
romange: You need to fix it though

src/server/tx_base.cc
kostasrim:
```
   82 LockFp LockTag::Fingerprint() const {                                                              
   83   return XXH64(str_.data(), str_.size(), 0x1C69B3F74AC4AE35UL);                                    
   84 }
```

src/server/engine_shard.cc
kostasrim: My mind is playing tricks with me because I am pretty sure I left a comment here about it :smile: 

I argue that `Destroy` is redundant. Why? 

Because it does two things:

1.   It calls  `round_robin_shards_tl_cache_.clear();`

This is a thread local variable. At the end of the life of a thread, the C++ runtime will deallocate the `round_robin_shards_tl_cache` anyway. Clearing it before you delete/deallocate does not really make sense.

2. We lock the guard and clear `round_robin_shards_`

Again redundant. Eventually the thread that owns will exit and deallocate it. Destructor will be called once by the owning thread.

Last but not least, we only call this flow (`Destroy`) during shutdown, which leads to threads exiting and cleaning up their statics/thread locals anyway.

We don't need to call `Destroy` unless I am missing something

romange (approved):

PR #5655, opened 2025-08-11
fix: refactor save synchronization to prevent deadlocks and crashes

vyavdoshenko:
Fixes regression: https://github.com/dragonflydb/dragonfly/actions/runs/16715390320/job/47307749937

This PR provides a comprehensive fix for a complex synchronization issue related to the snapshotting process (`SAVE`/`BGSAVE`), which could manifest as either a deadlock or a crash.

**The Problem:**

The underlying issue was introduced in https://github.com/dragonflydb/dragonfly/pull/5528. This change created a race condition where long-running save operations, which held a lock on `save_mu_`, could conflict with short-running informational commands like `INFO`.

This led to two distinct failure modes:
- **Deadlock:** Commands like `INFO` would hang long or indefinitely while waiting for a `SAVE` or `BGSAVE` operation to complete and release the `save_mu_` mutex.
- **Crash:** Concurrent access to save-related data structures could lead to a data race, resulting in a segmentation fault.

**History:**

A previous attempt to fix the crash was made in https://github.com/dragonflydb/dragonfly/pull/5633. While it successfully prevented the crash (only for x86_64 architecture) by adding more explicit locking, it did not address the root cause of the crash. A key contribution of https://github.com/dragonflydb/dragonfly/pull/5633 was the addition of a test case that reliably reproduces the crash, which was invaluable for verifying this fix.

**The Solution:**

This PR resolves both issues by refactoring the synchronization mechanism using a **Copy-on-Write** pattern with `std::shared_ptr`.

-  **`ThreadSafeSaveInfo`:** Historical data about completed saves is now stored in a thread-safe wrapper (`ThreadSafeSaveInfo`). This provides non-blocking, wait-free reads for any command needing to access the status of the *last* save.

-  **`save_controller_` Management:** The pointer to the *active* save controller (`save_controller_`) is now also managed as a `std::shared_ptr`. Informational commands can safely get a snapshot of the current save progress by briefly locking the mutex just to copy the `shared_ptr`, and then reading the data from the controller without holding the lock.

This new approach completely decouples readers (like `INFO`) from the long-running writer operations (like `SAVE`), eliminating both the deadlock and the data race conditions.

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -3434,9 +3434,11 @@ void ServerFamily::ReplicaOfInternal(CmdArgList args, Transaction* tx, SinkReply
       last_master_data = replica_->Stop();
     StopAllClusterReplicas();
 
-    // First, switch into the loading state
-    if (auto new_state = service_.SwitchState(GlobalState::ACTIVE, GlobalState::LOADING);
-        new_state != GlobalState::LOADING) {
+    const GlobalState gstate = ServerState::tlocal()->gstate();
+    if (gstate == GlobalState::TAKEN_OVER) {
+      service_.SwitchState(GlobalState::TAKEN_OVER, GlobalState::LOADING);
+    } else if (auto new_state = service_.SwitchState(GlobalState::ACTIVE, GlobalState::LOADING);
+               new_state != GlobalState::LOADING) {
       LOG(WARNING) << new_state << " in progress, ignored";
       builder->SendError("Invalid state");
       return;
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -1017,7 +1017,7 @@ std::optional<fb2::Fiber> Pause(std::vector<facade::Listener*> listeners, Namesp
 
 ServerFamily::ServerFamily(Service* service) : service_(*service) {
   start_time_ = time(NULL);
-  last_save_info_.save_time = start_time_;
+  thread_safe_save_info_.Update([this](SaveInfoData* data) { data->save_time = start_time_; });
   script_mgr_.reset(new ScriptMgr());
   journal_.reset(new journal::Journal());
 
@@ -1993,7 +1993,8 @@ GenericError ServerFamily::DoSaveCheckAndStart(const SaveCmdOptions& save_cmd_op
     return GenericError{make_error_code(errc::operation_in_progress),
                         StrCat(GlobalStateName(state), " - can not save database")};
   }
-  // Check if save is already in progress
+
+  std::shared_ptr<SaveStagesController> controller;
   {
     util::fb2::LockGuard lk(save_mu_);
     if (save_controller_) {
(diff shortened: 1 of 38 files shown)


vyavdoshenko:
I just found out that this fix destroys another fix: https://github.com/dragonflydb/dragonfly/pull/5528/
I have to find out how to fix the crash and avoid the regression with deadlock.

src/server/server_family.h
```diff
@@ -139,27 +142,51 @@ struct Metrics {
...
   time_t failed_duration_sec = 0;  // epoch time in seconds.
+
   // false if last attempt failed
   bool last_bgsave_status = true;
   bool bgsave_in_progress = false;
 };
 
+// A thread-safe wrapper for SaveInfoData using the Copy-on-Write pattern.
+// Provides a non-blocking Get() for readers and a locking Update() for writers.
+class ThreadSafeSaveInfo {
```
dranikpg: The implementation is pretty clear - but LastSaveInfo is a lighweight type in comparison to the operations its used int - it could've been just a copy

src/server/server_family.cc
```diff
@@ -1993,79 +1993,102 @@ GenericError ServerFamily::DoSaveCheckAndStart(const SaveCmdOptions& save_cmd_op
...
-    save_controller_ = std::move(temp_save_controller);
-    save_controller_->Start();
-    last_save_info_.bgsave_in_progress = bg_save;
-  }
   return {};
 }
 
 GenericError ServerFamily::WaitUntilSaveFinished(Transaction* trans, bool ignore_state) {
-  save_controller_->WaitAllSnapshots();
+  std::shared_ptr<SaveStagesController> controller;
```
dranikpg, line 2045: Haven't looked into this part of town for years - it looks like the previous code relied on the fact that WaitUntilSaveFinished was always called only when the before DoSaveCheckAndStart succeeded. So theoretically `shared_ptr` isn't needed anywhere, but without it also looks to fragile

src/server/server_family.cc
```diff
@@ -1993,79 +1993,102 @@ GenericError ServerFamily::DoSaveCheckAndStart(const SaveCmdOptions& save_cmd_op
...
-      return res->error;
-    }
+  // Success - update state
+  controller->Start();
+  thread_safe_save_info_.Update(
+      [bg_save](SaveInfoData* data) { data->bgsave_in_progress = bg_save; });
 
-    // Success - set the controller and update state
-    save_controller_ = std::move(temp_save_controller);
-    save_controller_->Start();
```
dranikpg, line 2039: I think this is what caused `INFO` starvation

dranikpg (approved):

src/server/server_family.h
vyavdoshenko: Yes, I agree. But I guess shared_ptr is still cheaper.

src/server/server_family.cc
vyavdoshenko: Yes, but not only. Init() can consume a lot of time for aws interactions.

src/server/server_family.h
```diff
@@ -139,27 +142,51 @@ struct Metrics {
...
+ public:
+  ThreadSafeSaveInfo() : data_(std::make_shared<const SaveInfoData>()) {
+  }
+
+  // Returns a snapshot of the current save info.
+  // This is a non-blocking, wait-free operation.
+  std::shared_ptr<const SaveInfoData> Get() const {
+    return std::atomic_load(&data_);
```
kostasrim: `std::atomic_load` for `shared_ptr` is deprecated in C++20. Worth a mutex instead if you want to grab copies of the same instance ?

p.s. haven't cehcked the rest of the code

src/server/server_family.h
```diff
@@ -139,27 +142,51 @@ struct Metrics {
...
+  }
+
+  // Atomically updates the save info.
+  // The modifier function is called under a lock.
+  void Update(std::function<void(SaveInfoData*)> modifier) {
+    std::lock_guard<util::fb2::Mutex> lock(writer_mutex_);
+    auto new_data = std::make_shared<SaveInfoData>(*Get());
+    modifier(new_data.get());
+    std::atomic_store(&data_, std::const_pointer_cast<const SaveInfoData>(new_data));
```
kostasrim: same here -- it's deprecated

(I will check the rest of the PR when I get a chance)

src/server/server_family.h
dranikpg: It's deprecated because there will be a specializaton for std::atomic. Not using either because one will be deprecated and the other isn't available, is... strange 🙂

src/server/server_family.h
dranikpg: Cheaper but makes the code more complicated

src/server/server_family.h
kostasrim: I missed the "small details" --> I quote:

>3,4) Partial specializations std::atomic<[std::shared_ptr](https://en.cppreference.com/w/cpp/memory/shared_ptr.html)<U>> and std::atomic<[std::weak_ptr](https://en.cppreference.com/w/cpp/memory/weak_ptr.html)<U>> are provided for [std::shared_ptr](https://en.cppreference.com/w/cpp/memory/shared_ptr.html) and [std::weak_ptr](https://en.cppreference.com/w/cpp/memory/weak_ptr.html).
See [std::atomic<std::shared_ptr>](https://en.cppreference.com/w/cpp/memory/shared_ptr/atomic2.html) and [std::atomic<std::weak_ptr>](https://en.cppreference.com/w/cpp/memory/weak_ptr/atomic2.html) for details.

For some reason I thought even the `atomic_load` was implemented internally via a lock (so is lockless would have returned false). Jokes on me on this one :smile:

vyavdoshenko: I suggest leaving it as is, because this specialization hasn't been added to GCC 13. But we use even older GCC 12 in the CI workflow.
It will be easy to fix later to support C++20

src/server/server_family.cc
```diff
@@ -1993,79 +1993,102 @@ GenericError ServerFamily::DoSaveCheckAndStart(const SaveCmdOptions& save_cmd_op
...
+    // It's possible that another save was initiated and the controller has changed.
+    // We only finalize and reset if it's still the same one we were waiting for.
+    if (save_controller_ == controller) {
+      save_info = save_controller_->Finalize();
+      is_bg_save = save_controller_->IsBgSave();
+      save_controller_.reset();
+    } else {
+      // Another save has started. The old one is already finalized by the new one.
+      // We just need to get the info.
+      return GenericError("Save operation was superseded by another save");
```
BorysTheDev, line 2071: Did we have the same behavior earlier?

vyavdoshenko: No, we blocked the whole process.

src/server/server_family.h
BorysTheDev: Our policy: use a mutex instead of an atomic.

src/server/server_family.cc
BorysTheDev: So if we have several save commands we cancel the previous one and if the process is long and SAVE is repeated we can get no snapshot at all

src/server/server_family.h
vyavdoshenko: > Our policy: use a mutex instead of an atomic.

This is the main idea of these changes. If I use the mutex, I will return to a deadlock. I can play with the second mutex. But I guess this solution is better.

src/server/server_family.cc
vyavdoshenko: We will have at least one successful save. It should be the first one. We cancel all next commands until the save process finishes.

src/server/server_family.h
BorysTheDev: As I see data_ is independent from writer_mutex_ and can be set without it, so if you use another mutext to protect it, it is ok

src/server/server_family.h
BorysTheDev: I agree with @dranikpg it's better to make a copy. we don't need to save CPU time on such operations

src/server/server_family.h
kostasrim: > I suggest leaving it as is, because this specialization hasn't been added to GCC 13. But we use even older GCC 12 in the CI workflow. It will be easy to fix later to support C++20

Yes I agree!

vyavdoshenko: The best way to skip updating the atomic-related code is to remove the atomic-related code. 😄

src/server/server_family.cc
```diff
@@ -2952,17 +2976,20 @@ string ServerFamily::FormatInfoMetrics(const Metrics& m, std::string_view sectio
     double perc = 0;
     bool is_saving = false;
     uint32_t curent_durration_sec = 0;
+    std::shared_ptr<detail::SaveStagesController> controller_copy;
```
BorysTheDev: I see this code 2 times maybe let's make a function to get a copy

src/server/server_family.h
```diff
@@ -4,6 +4,9 @@
 
 #pragma once
 
+#include <atomic>
```
BorysTheDev: I think you don't need it

src/server/server_family.h
```diff
@@ -139,27 +142,52 @@ struct Metrics {
...
+  SaveInfoData Get() const {
+    std::lock_guard<util::fb2::Mutex> lock(data_mutex_);
+    return data_;
+  }
+
+  // The modifier function is called under a lock.
+  void Update(std::function<void(SaveInfoData*)> modifier) {
+    std::lock_guard<util::fb2::Mutex> lock(writer_mutex_);
+    SaveInfoData new_data(Get());
+    modifier(&new_data);
```
BorysTheDev, line 175: I think you can add a scope to unlock writer_mutex_ before UpdateData

src/server/server_family.cc

src/server/server_family.h
vyavdoshenko: fixed

BorysTheDev (approved):

BorysTheDev (approved):

PR #5657, opened 2025-08-11
fix(pytest):  Don't delete and modify huge value keys in test_replication_all

mkaruza:
If we allow huge value keys to be deleted and modified we don't know
what will be executed due random seeder nature. This also means that
lower bound preemption check can fail.
Modify SeederV2 class so it accepts additional flag which indicates that
huge value keys can only be added and not changed / deleted.

### DIFF
--- a/tests/dragonfly/seeder/script-generate.lua
+++ b/tests/dragonfly/seeder/script-generate.lua
@@ -11,7 +11,6 @@ local prefix = ARGV[1]
 local type = ARGV[2]
 local key_counter = tonumber(ARGV[3])
 local stop_key = ARGV[4]
-
 -- inputs: task specific
 local key_target = tonumber(ARGV[5])
 local total_ops = tonumber(ARGV[6])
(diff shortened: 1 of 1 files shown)


romange:
@mkaruza  what check can fail with huge values?

mkaruza:
> @mkaruza what check can fail with huge values?

`test_replication_all` fails on ` assert preemptions >= seeder.huge_value_target * 0.5` assert. It happens because on some runs we are deleting huge value keys before they are replicated so number of expected preemptions will fail.

tests/dragonfly/seeder/script-generate.lua
```diff
@@ -33,14 +33,26 @@ LG_funcs.init(data_size, collection_size, huge_value_target, huge_value_size)
 local addfunc = LG_funcs['add_' .. string.lower(type)]
 local modfunc = LG_funcs['mod_' .. string.lower(type)]
 local huge_entries = LG_funcs["get_huge_entries"]
+local is_next_huge_entry = LG_funcs["is_huge_entry"]
+-- Keep track of total number of keys including huge value keys. Intialize
+-- to number of keys that currently exists.
+local total_keys = #keys
```
kostasrim, line 39: Why not use `key_counter` ? Each time we add a key we increment it, so IMO this is the same variable with a different name.

tests/dragonfly/seeder/__init__.py
```diff
@@ -150,6 +150,7 @@ def __init__(
         huge_value_target=5,
         huge_value_size=100000,
         seed=None,
+        huge_value_add_only=False,
```
kostasrim, line 153: leaving this here for visibility. 

We break `mod` operations for big values that way. We do not therefore create a key with many big values because of that as a side effect. However, this does not change the flows/logic we test so it's fine. 

In the future, we should use 2 tables and one should be used for big values. If a key is a big value, operations like `mod` should only `append` and not `remove` from the working key.

kostasrim (commented):
lgtm -- minor comment

tests/dragonfly/seeder/script-generate.lua
mkaruza: It is not same variable `key_counter` is monotonically increasing counter while `total_keys` is used to keep track of added/removed keys. It is same as `#keys` if `huge_value_add_only=False` but if we don't want to modify huge values keys this keeps track of all keys during run.

kostasrim: > monotonically increasing counter 

Uh I missed that! I thought that `total_keys` was also `decremented` on deletions!

kostasrim (approved):

PR #5658, opened 2025-08-11
chore: calibrate regression tests running time

### LINKED ISSUE #5662: test_replicaof_reject_on_load
\https://github.com/dragonflydb/dragonfly/actions/runs/16884994243/job/47830310679#step:6:1077

kostasrim:
Regression tests now almost reach the 60m timeout limit. This PR helps with the total running time of the tests.

also fixes #5662

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -100,6 +100,7 @@ jobs:
         dfly-executable: dragonfly-x86_64
         gspace-secret: ${{ secrets.GSPACES_BOT_DF_BUILD }}
         build-folder-name: ${{ env.RELEASE_DIR }}
+        filter: 'not debug_only'
     - name: Save artifacts
       run: |
           # place all artifacts at the same location
--- a/tests/pytest.ini
+++ b/tests/pytest.ini
@@ -7,9 +7,17 @@ log_cli = true
 asyncio_mode=auto
 addopts = -ra --emoji
 markers =
+# Slow tests that take some time to complete but are run both as part of the
+# regression-test *and* ci *and* release workflows.
   slow: marks tests as slow (deselect with '-m "not slow"')
+# Tests that should only run on release builds and take significant amount of time to run.
+# For example stress tests found in replication.
+# `opt_only` runs only on regression-test and release workflow (skipped for ci)
   opt_only: marks tests that are only reasonable to run against an opt-built Dragonfly
+# Usually tests that are known to fail for iouring and we skip them on epoll workflows
   exclude_epoll: marks tests that should not run on epoll socket
+# Tests that should only run in debug mode because release builds are fast enough
+# for their assertions to hold. They never run on release build.
   debug_only: mark tests that should run only in debug mode
 filterwarnings =
     ignore::DeprecationWarning


kostasrim:
There is another low hanging fruit: `test_rss_used_mem_gap` which I am looking on

kostasrim:
https://github.com/dragonflydb/dragonfly/actions/runs/16884994243 (roughly 8 mins saved for debug)

and for release:

https://github.com/dragonflydb/dragonfly/actions/runs/16913683229/job/47974089443 (roughly 3-4 mins saved for release)

tests/dragonfly/memory_test.py
```diff
@@ -24,10 +24,12 @@
 # We limit to 5gb just in case to sanity check the gh runner. Otherwise, if we ask for too much
 # memory it might force the gh runner to run out of memory (since OOM killer might not even
 # get a chance to run).
-@dfly_args({"proactor_threads": 4, "maxmemory": "5gb"})
+@dfly_args({"proactor_threads": 2, "maxmemory": "5gb"})
 async def test_rss_used_mem_gap(df_factory, type, keys, val_size, elements):
     dbfilename = f"dump_{tmp_file_name()}"
-    instance = df_factory.create(dbfilename=dbfilename)
+    instance = df_factory.create(
+        dbfilename=dbfilename, compression_mode=0, serialization_max_chunk_size=8192, num_shards=2
```
kostasrim: made a huge difference -- especially locally :shrug: 

I also reduced to 2 proactor threads such that we don't context switch on the GH runner (since it has 2 threads) and I used `2 shards` to effectively utilize both cores

tests/dragonfly/snapshot_test.py
```diff
@@ -274,7 +274,7 @@ async def test_shutdown_save_with_rename(df_server):
     await client.connection_pool.disconnect()
 
 
-@pytest.mark.slow
+@pytest.mark.opt_only
```
kostasrim, line 277: Both of the tests are behavioural and test that only one save command succeeds if sent concurrently. It's fine to test this only `release builds` as the actual seeding takes significant amount of time. Specifically, in debug build those 2 tests take:

149.64s call     dragonfly/snapshot_test.py::test_parallel_snapshot[df_factory0]

150 seconds to test that no concurrent saves are allowed. yikes! 

imo this should be a unit

kostasrim:
There are other low hanging fruits now which we can come back later. Specifically:

Check if:
```
148.21s call     dragonfly/cluster_test.py::test_migration_timeout_on_sync[df_seeder_factory0-df_factory0]
126.52s call     dragonfly/cluster_test.py::test_migration_rebalance_node[df_seeder_factory0-df_factory0]
```
Are needed for debug build. They take 5+ minutes dominating most of the total running time for debug.
```
120.01s call     dragonfly/snapshot_test.py::test_parallel_snapshot_race_condition[df_factory0]
```
This was added recently but I can't justify 120seconds of running time for a single race condition test

tests/dragonfly/replication_test.py
```diff
@@ -2033,13 +2033,19 @@ async def test_client_pause_with_replica(df_factory, df_seeder_factory):
     assert await seeder.compare(capture, port=replica.port)
 
 
+@pytest.mark.debug_only
+@dfly_args({"proactor_threads": 2})
```
kostasrim, line 2037: Fixes #5662


1. This test takes significant amount of time in `debug mode` 100+ seconds
2. This test is extremely inefficient when it runs on release mode because loading and saving a snapshot is orders of magnitudes faster in release than in debug. So when we try to assert that that dragonfly is in loading state we need to have a big snapshot for release. Guess what! This kills the running time in debug mode.

The fix:

Only run this test in debug mode where generally loading a snpashot is `slower` as we only care for behavioural semantics here. That way the test has more time to check the `LOADING` state without adjusting the seeder/snapshot size on release builds. 

Bonus points is now we can reduce the snapshot size and make the debug test run faster!

tests/dragonfly/memory_test.py
romange: sure but I find it weird that we split args between `df_factory.create` and `dfly_args`. Is this necessary in this test?

tests/dragonfly/replication_test.py
```diff
@@ -2033,13 +2033,19 @@ async def test_client_pause_with_replica(df_factory, df_seeder_factory):
...
 async def test_replicaof_reject_on_load(df_factory, df_seeder_factory):
     master = df_factory.create()
     replica = df_factory.create(dbfilename=f"dump_{tmp_file_name()}")
     df_factory.start_all([master, replica])
 
     c_replica = replica.client()
-    await c_replica.execute_command(f"DEBUG POPULATE 1000 key 1000 RAND type set elements 2000")
+
+    seeder = SeederV2(
+        key_target=1000, types=["SET"], data_size=150, collection_size=100, huge_value_size=0
```
romange: I am surprised that `SeederV2` is faster than `DEBUG POPULATE`. Is it the case?

tests/dragonfly/snapshot_test.py
romange: what's the different between slow and opt_only?

tests/dragonfly/memory_test.py
kostasrim: nope,I was just being sloppy

tests/dragonfly/snapshot_test.py
kostasrim: on what we run on the CI.

For reg tests workflow:
* we run every test except `opt_only` on debug and everything else on release.
For ci workflow:
* we run every test except `opt only and slow` on debug and evertying else on release.

Basically we use different filters for the two different workflows:

For the former: filter: ${{ matrix.build-type == 'Release' && 'not debug_only' || 'not opt_only' }}

For the later:   filter: ${{ matrix.build-type == 'Release' && 'not empty' || '(not slow) and (not opt_only)' }}

tests/dragonfly/replication_test.py
kostasrim: Not for all data types but the same configuration as we have here (and on debug) gives me:

1 passed, 116 deselected in 2.65s  -> `SeederV2`
1 passed, 116 deselected in 27.21s  -> `Debug populate`

Same config both seeders. running time mentioned is the seeding phase isolated and everything else removed (and a `dbfilename=""` such that we don't snapshot at the end)

kostasrim: Haven't looked on this any further but there are a few things I noticed:

1. sometimes seederv2 is faster than debug populate and vice versa (so it's not universal)
2. seeding with seederv2 makes it hard to reach the desired data size because of deletions and mofications of the key space. I tried to use seederv2 for example for rss test, and the same seeder config produced vastly different dataset sizes.

:shrug:

tests/dragonfly/snapshot_test.py
romange: I do not know how to parse those filters.
so on release we run all the tests?
the differrence only for debug  where CI does not run slow tests?

kostasrim: Yes we have a different set of running tests depending on the workflow. In summary:

For `release` --> we run `all tests` including, opt, slow, etc
For `regression test action that runs every 3 hours` --> we run every test except `opt_only` for debug and everything on `release`
For `CI worklow` --> we run every test except `opt slow and opt only` for `debug` and everything for `release`

Now we added `debug only` for tests that should run only in debug mode

.github/actions/regression-tests/action.yml
```diff
@@ -56,7 +56,7 @@ runs:
         else
           export FILTER="${{inputs.filter}}"
           # Run only replication tests with iouring
-          timeout 60m pytest -m "$FILTER" --durations=10 --timeout=300 --color=yes --json-report --json-report-file=report.json dragonfly --log-cli-level=INFO || code=$?
+          timeout 60m pytest -m "$FILTER" dragonfly/replication_test.py -k "test_replicaof_reject_on_load" --durations=10 --timeout=300 --color=yes --json-report --json-report-file=report.json dragonfly --log-cli-level=INFO || code=$?
```
kostasrim: will revert - used for testing

tests/dragonfly/snapshot_test.py
kostasrim: uh and this will be problematic for release...  

option 1: we skip the debug_only test when running the release workflow (similar to what we do now for reg test flows)

option 2: we run it on release only -- meaning we do the opposite of what we do here

I don't see any big issue with option 1. Although we cherry pick commits on release I don't see it likely that we break this test. If a commit has a follow up patch, I hardly see why we would cherry pick the former yet miss the later but I don't cut the releases so I will do whichever you think is best

kostasrim:
https://github.com/dragonflydb/dragonfly/actions/runs/16935887202/job/47992215231 

sub 50 minutes for debug!

tests/dragonfly/snapshot_test.py
romange: I do not mind running `debug_only` only for debug but please add comments, maybe in
pytest.ini what these markers mean. There are comments now but they are not clear

kostasrim: yes I will update them all give me a few minutes

kostasrim: done

PR #5659, opened 2025-08-11
fix: accounting of metrics under squash_stats_latency_lower_limit filter

romange:
Fix pipeline_queue_wait_duration_seconds and pipeline_commands_duration_seconds with pipeline_dispatch_commands_total that before that did not take into account squash_stats_latency_lower_limit filter.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1557,8 +1557,6 @@ void Connection::SquashPipeline() {
     CHECK(holds_alternative<PipelineMessagePtr>(msg.handle))
         << msg.handle.index() << " on " << DebugInfo();
 
-    // measure the time spent in the pipeline  queue waiting for dispatch
-    stats_->pipelined_wait_latency += CycleClock::ToUsec(start - msg.dispatch_cycle);
     auto& pmsg = get<PipelineMessagePtr>(msg.handle);
     squash_cmds.emplace_back(absl::MakeSpan(pmsg->args));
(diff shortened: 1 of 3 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -1598,8 +1596,15 @@ void Connection::SquashPipeline() {
   while (it->IsControl())  // Skip all newly received intrusive messages
     ++it;
 
-  for (auto rit = it; rit != it + dispatched; ++rit)
+  for (auto rit = it; rit != it + dispatched; ++rit) {
+    if (result.account_in_stats) {
+      // measure the time spent in the pipeline  queue waiting for dispatch
+      stats_->pipelined_wait_latency += CycleClock::ToUsec(start - rit->dispatch_cycle);
```
dranikpg, line 1602: the real one is the average before those two 🙂

dranikpg (approved):

src/facade/dragonfly_connection.cc
romange: wdym?

dranikpg: You measured before dispatching the first, now you measure after dispatching the last

romange: please double check me but nothing should change in the measurement. I do not change how I clock `start` and `rit->dispatch_cycle`

dranikpg: Ah, yes, indeed, my bad

PR #5663, opened 2025-08-12
fix: macos build tiered_storage.h

vyavdoshenko:
https://github.com/dragonflydb/dragonfly/actions/runs/16900538470

### DIFF
--- a/src/server/tiered_storage.cc
+++ b/src/server/tiered_storage.cc
@@ -28,18 +28,10 @@
 
 using namespace facade;
 
-ABSL_FLAG(bool, tiered_experimental_cooling, true,
-          "If true, uses intermidate cooling layer "
-          "when offloading values to storage");
-
-ABSL_FLAG(unsigned, tiered_storage_write_depth, 50,
-          "Maximum number of concurrent stash requests issued by background offload");
-
-ABSL_FLAG(float, tiered_offload_threshold, 0.5,
-          "Ratio of free memory (free/max memory) below which offloading starts");
-
-ABSL_FLAG(float, tiered_upload_threshold, 0.1,
-          "Ratio of free memory (free/max memory) below which uploading stops");
+ABSL_DECLARE_FLAG(bool, tiered_experimental_cooling);
+ABSL_DECLARE_FLAG(unsigned, tiered_storage_write_depth);
+ABSL_DECLARE_FLAG(float, tiered_offload_threshold);
+ABSL_DECLARE_FLAG(float, tiered_upload_threshold);
 
 namespace dfly {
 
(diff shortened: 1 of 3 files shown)


BorysTheDev (approved):

kostasrim:
@vyavdoshenko Wait for https://github.com/dragonflydb/dragonfly/actions/runs/16902420800 to pass

kostasrim:
@vyavdoshenko Tests fail https://github.com/dragonflydb/dragonfly/actions/runs/16902420800/job/47884243070

vyavdoshenko:
Another attempt to build and pass tests: https://github.com/dragonflydb/dragonfly/actions/runs/16903348891

src/server/tiered_storage.cc
```diff
@@ -28,18 +28,10 @@
...
-          "when offloading values to storage");
-
-ABSL_FLAG(unsigned, tiered_storage_write_depth, 50,
-          "Maximum number of concurrent stash requests issued by background offload");
-
-ABSL_FLAG(float, tiered_offload_threshold, 0.5,
-          "Ratio of free memory (free/max memory) below which offloading starts");
-
-ABSL_FLAG(float, tiered_upload_threshold, 0.1,
-          "Ratio of free memory (free/max memory) below which uploading stops");
```
dranikpg, line 42: Why, where else do we use them? 🙂 Actually, what I wanted to do for a long time is to have a WITH_TIERING cmake variable that gets disabled on macos, so we can #ifdef those out

kostasrim: I suspect the tests fail because of the stub tiering class (haven't checked just thinking out loud)

+1 for cmake as it will remove the stub

vyavdoshenko: @dranikpg Kostas is right. Tests fail. 
This fix can fix the build now. If you don't like this fix, please suggest yours.
macOS build successfully passed: https://github.com/dragonflydb/dragonfly/actions/runs/16903348891/job/47890839564

FYI
@kostasrim

kostasrim: The desired approach IMO is what @dranikpg suggested -- `CMAKE + WITH_TIERING` variable and the reason for this is you don't want to compile any of the tiering code `since it's a stub` and does not properly work for `macos`

kostasrim: Basically what Vlad is saying here is that "we should not compile unused code especially when it's mocked and doesn't do anything"

vyavdoshenko: It means that this PR is useless. Let's close it? `WITH_TIERING` should be implemented without these changes.

kostasrim: I never said that this PR is useless :) Until we get that it's perfectly find to keep this

PR #5664, opened 2025-08-12
tests: Check log to assert instance is loading

### LINKED ISSUE #5662: test_replicaof_reject_on_load

abhijat:
The test previously checked if the status is loading in the output of info command, however this status may have switched from loading to active by the time the check is performed. Searching the log lets us know if the state was loading now or in the past.

The downside is relying on internals of the server. However this should be fine for such a simple case where we have a clear log of state switch.

FIXES https://github.com/dragonflydb/dragonfly/issues/5662

### DIFF
--- a/tests/dragonfly/replication_test.py
+++ b/tests/dragonfly/replication_test.py
@@ -2033,9 +2033,9 @@ async def test_client_pause_with_replica(df_factory, df_seeder_factory):
     assert await seeder.compare(capture, port=replica.port)
 
 
-async def test_replicaof_reject_on_load(df_factory, df_seeder_factory):
+async def test_replicaof_reject_on_load(df_factory: DflyInstanceFactory, df_seeder_factory):
     master = df_factory.create()
-    replica = df_factory.create(dbfilename=f"dump_{tmp_file_name()}")
+    replica = df_factory.create(dbfilename=f"dump_{tmp_file_name()}", vmodule="main_service=1")
     df_factory.start_all([master, replica])
 
     c_replica = replica.client()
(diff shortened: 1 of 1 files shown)


romange:
I do not like this and we have a better ways  to achieve such checks. What this test is trying to check? that replica is currently loading or it has finished syncing?

romange:
Does not this fix move the problem somewhere else? the test is trying to check if Dragonfly responds with an error during the loading phase. what's the point to check that the loading finished in the past?

romange:
Also it sounds very weird that the server finished loading the snapshot before the checks started flowing. Are you sure this is what happened?

abhijat:
I misunderstood the test by looking at just the failure, looking at the last three statements in the test, this approach seems wrong. I will try to come up with a more appropriate fix.

abhijat:
> Also it sounds very weird that the server finished loading the snapshot before the checks started flowing. Are you sure this is what happened?

It looks like that from the logs but it is not clear when the client started requesting `info persistence`

```
I20250812 11:31:23.340124 26002 snapshot_storage.cc:186] Load snapshot: Searching for snapshot in directory: "/tmp/tmp5wyx9opx"
I20250812 11:31:23.340405 26002 main_service.cc:2802] Switching state from ACTIVE to LOADING
I20250812 11:31:23.340485 26002 server_family.cc:1320] Loading /tmp/tmp5wyx9opx/dump_zLZroYLHEr-summary.dfs
I20250812 11:31:23.343497 26005 listener_interface.cc:102] sock[13] AcceptServer - listening on 0.0.0.0:43723
I20250812 11:31:23.636123 26003 server_family.cc:1374] Load finished, num keys read: 1990
I20250812 11:31:23.636269 26003 main_service.cc:2802] Switching state from LOADING to ACTIVE
```

PR #5665, opened 2025-08-12
chore: move tiering flags

kostasrim:
added `-WITH_TIERING` to avoid registering the flags on macos

### DIFF
--- a/.github/workflows/daily-builds.yml
+++ b/.github/workflows/daily-builds.yml
@@ -107,8 +107,7 @@ jobs:
           echo "*************************** START BUILDING **************************************"
           CC=gcc-12 CXX=g++-12 cmake .. -DCMAKE_BUILD_TYPE=Debug -GNinja -DWITH_UNWIND=OFF \
             -DCMAKE_CXX_FLAGS="-Wl,-ld_classic" \
-            -DCMAKE_C_COMPILER="gcc-12" -DCMAKE_CXX_COMPILER="g++-12" \
-            -DWITH_TIERING=OFF
+            -DCMAKE_C_COMPILER="gcc-12" -DCMAKE_CXX_COMPILER="g++-12"
 
           ninja src/all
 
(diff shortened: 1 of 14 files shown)


src/server/CMakeLists.txt
```diff
@@ -17,7 +17,8 @@ endif()
 set_property(SOURCE dfly_main.cc APPEND PROPERTY COMPILE_DEFINITIONS
              SOURCE_PATH_FROM_BUILD_ENV=${CMAKE_SOURCE_DIR})
 
-if ("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
+option(WITH_TIERING "Compile for macos" ON)
+if (WITH_TIERING)
```
dranikpg: Please put option() to where all other options are (above)

.github/workflows/daily-builds.yml
```diff
@@ -107,7 +107,8 @@ jobs:
           echo "*************************** START BUILDING **************************************"
           CC=gcc-12 CXX=g++-12 cmake .. -DCMAKE_BUILD_TYPE=Debug -GNinja -DWITH_UNWIND=OFF \
             -DCMAKE_CXX_FLAGS="-Wl,-ld_classic" \
-            -DCMAKE_C_COMPILER="gcc-12" -DCMAKE_CXX_COMPILER="g++-12"
+            -DCMAKE_C_COMPILER="gcc-12" -DCMAKE_CXX_COMPILER="g++-12" \
+            -DWITH_TIERING=OFF
```
dranikpg: I think it would've been better if we automatically set WITH_TIERING=OFF in cmake if we see that it's macos

dranikpg (approved):
Except that we'd maybe better disable it from cmake automatically

.github/workflows/daily-builds.yml
kostasrim: oh nice! +1 since we can check the os in cmake :tada:

src/server/CMakeLists.txt
kostasrim: done

PR #5666, opened 2025-08-12
chore(server): Semi-automatic flag updates

dranikpg:
Before I add a few scheduler flags, I'd like to avoid creating a million setters

The idea is that by providing two functions, `UpdateFromFlags` and `GetMutableFlagValues`, we can let the registry know that it has to call the former whenever a flag from the latter changes. This way each class cleanly keeps its flags in it's own codebase and mutability can be enabled in a single line

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1450,7 +1450,6 @@ auto Connection::IoLoop() -> variant<error_code, ParserStatus> {
   ParserStatus parse_status = OK;
 
   size_t max_iobfuf_len = GetFlag(FLAGS_max_client_iobuf_len);
-  max_busy_read_cycles_cached = GetFlag(FLAGS_max_busy_read_usec);
 
   auto* peer = socket_.get();
   recv_buf_.res_len = 0;
@@ -2172,7 +2171,7 @@ void Connection::UpdateFromFlags() {
   thread_queue_backpressure[tid].pipeline_buffer_limit = GetFlag(FLAGS_pipeline_buffer_limit);
   thread_queue_backpressure[tid].pipeline_cnd.notify_all();
 
-  max_busy_read_cycles_cached = GetFlag(FLAGS_max_busy_read_usec);
+  max_busy_read_cycles_cached = base::CycleClock::FromUsec(GetFlag(FLAGS_max_busy_read_usec));
   always_flush_pipeline_cached = GetFlag(FLAGS_always_flush_pipeline);
   pipeline_squash_limit_cached = GetFlag(FLAGS_pipeline_squash_limit);
   pipeline_wait_batch_usec = GetFlag(FLAGS_pipeline_wait_batch_usec);
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -22,6 +22,7 @@
 #include "core/heap_size.h"
 #include "facade/conn_context.h"
 #include "facade/dragonfly_listener.h"
+#include "facade/flag_utils.h"
 #include "facade/memcache_parser.h"
 #include "facade/redis_parser.h"
 #include "facade/service_interface.h"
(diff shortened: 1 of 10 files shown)


src/facade/flag_utils.h
```diff
@@ -0,0 +1,15 @@
...
+
+#include <string>
+#include <vector>
+
+#include "absl/flags/flag.h"
+
+namespace facade {
+template <typename... Ts> std::vector<std::string> GetFlagNames(const absl::Flag<Ts>&... flags) {
+  return std::vector<std::string>{std::string{absl::GetFlagReflectionHandle(flags).Name()}...};
+}
```
kostasrim:
```suggestion
template <typename... Ts> std::vector<std::string> GetFlagNames(const absl::Flag<Ts>&... flags) {
  return {absl::GetFlagReflectionHandle(flags).Name()...};
}
```

dranikpg: Either the string{} or vector, because I have to change string_view->string

src/server/main_service.cc
```diff
@@ -724,55 +713,32 @@ string FailedCommandToString(std::string_view command, facade::CmdArgList args,
...
-template <typename T> auto UpdateServerState(T ServerState::*member) {
-  return [member](T value) {
-    auto cb = [member, value](unsigned, auto*) { ServerState::tlocal()->*member = value; };
-    shard_set->pool()->AwaitBrief(cb);
-  };
-}
-
-thread_local uint32_t squash_stats_latency_lower_limit_cached =
-    absl::GetFlag(FLAGS_squash_stats_latency_lower_limit);
+thread_local uint32_t squash_stats_latency_lower_limit_cached;
```
BorysTheDev, line 714: static thread_local uint32_t squash_stats_latency_lower_limit_cached;

or anonymous namespace

src/facade/dragonfly_connection.cc
```diff
@@ -2165,14 +2165,22 @@ bool Connection::IsReplySizeOverLimit() const {
   return over_limit;
 }
 
-void Connection::SetMaxQueueLenThreadLocal(unsigned tid, uint32_t val) {
-  thread_queue_backpressure[tid].pipeline_queue_max_len = val;
+void Connection::UpdateFromFlags() {
```
BorysTheDev, line 2168: we need to return the number of updated flags and somewhere check that we have the same number as in GetMutableFlagNames().size()

src/server/server_state.h
```diff
@@ -310,6 +310,10 @@ class ServerState {  // public struct - to allow initialization.
   };
   void DecommitMemory(uint8_t flags);
 
+  void UpdateFromFlags();  // Update configration from vlags
+  static std::vector<std::string>
+  GetMutableFlagNames();  // Expects to trigger UpdateFromFlags on change
```
kostasrim: what did clang-format do here :thinking:

src/facade/dragonfly_connection.cc
dranikpg: Yes! I though of making a wrapper in `UpdateFromFlags` that records the accessed values, so you can call `UpdateFromFlags` once and get a list of dependent flags. But for the few cases that we have it's too complicated either because it's just a few lines of code or because the "object" isn't well defined, sometimes it's just a free function (`Connection::`) or a member or I sometimes even provide it as a list, so the generilzation will have more exceptions than base cases now. But I agree, I wanted it to be automated too. Either way it's better what we have now

src/server/main_service.cc
dranikpg: It already is inside one, it's just gigantic

src/server/server_state.h
dranikpg: Oh it commited a crime 😈... changed the comment so it still takes up only one line

dranikpg:
Rebased to include new flags that were added to main in the meantime 🙄 Fixed the comment

Issue #5534, opened 2025-07-21
test_migration_timeout_on_sync


kostasrim:
https://github.com/dragonflydb/dragonfly/actions/runs/16412948605/job/46372161069#step:6:5095

```
E               AssertionError: [['out', '3b06550bca7d41628ad81d46b10b64ae94316390', 'SYNC', 299950, '0']]
E               assert (1 != 0 and False)
E                +  where 1 = len([['out', '3b06550bca7d41628ad81d46b10b64ae94316390', 'SYNC', 299950, '0']])
E                +  and   False = all(<generator object wait_for_error.<locals>.<genexpr> at 0x7fe6e681bdd0>)
```

kostasrim:
Still failing: https://github.com/dragonflydb/dragonfly/actions/runs/16898100617/job/47871715744#step:6:5400

CC: @BorysTheDev

PR #5667, opened 2025-08-12
fix: test_migration_timeout_on_sync

BorysTheDev:
fixes: https://github.com/dragonflydb/dragonfly/issues/5534

problem: if `inflight_bytes() > replication_stream_output_limit_cached / 3` we never throttle and return error "JournalStreamer write operation timeout"
fix: we should use the same condition to check whether we stalled or not


romange:
I do not understand the PR description. `if inflight_bytes() > replication_stream_output_limit_cached / 3` we literally call ` `ThisFiber::SleepFor(300us);` which is the literally implementation of throttling. Why it suddenly stopped working?

BorysTheDev:
> I do not understand the PR description. `if inflight_bytes() > replication_stream_output_limit_cached / 3` we literally call ` `ThisFiber::SleepFor(300us);` which is the literally implementation of throttling. Why it suddenly stopped working?

It's not stop working, it works all the time and don't allow to saturate pending_buf_

romange (approved):

PR #5668, opened 2025-08-13
workflows: Add a dispatch action which will run pytests on repeat

abhijat:
An action which triggers user specified pytests on repeat with the following inputs:

* branch
* arbitrary pytest filters
* count
* timeout for all runs (default 60 minutes)

Example run from command line:

Runs test `test_replicaof_reject_on_load` 100 times from the branch `abhijat/feat/action-to-repeat-tests` with 2 hours timeout

```
gh workflow run "Repeat Tests" --ref abhijat/feat/action-to-repeat-tests -f branch=abhijat/feat/action-to-repeat-tests -f count=100 -f expression="-xv dragonfly/replication_test.py -k test_replicaof_reject_on_load" -f timeout=120m
```

### DIFF
--- a/tests/dragonfly/conftest.py
+++ b/tests/dragonfly/conftest.py
@@ -5,26 +5,26 @@ Pytest fixtures to be provided for all tests without import
 import asyncio
 import logging
 import os
-import sys
-from time import sleep
-from typing import Dict, List, Union
-from redis import asyncio as aioredis
-import pytest
-import pytest_asyncio
-import redis
-import pymemcache
 import random
-import subprocess
 import shutil
+import subprocess
+import sys
 import time
 import typing
 from copy import deepcopy
-
 from pathlib import Path
 from tempfile import gettempdir, mkdtemp
+from time import sleep
+from typing import Dict, List, Union
 
+import pymemcache
+import pytest
+import pytest_asyncio
+import redis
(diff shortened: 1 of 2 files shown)


abhijat:
With larger repeat counts we run out of disk space pretty quickly https://github.com/dragonflydb/dragonfly/actions/runs/16929309934/job/47971115197

abhijat:
The storage probably needs to use https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#dynamic-scope to allow cleaning up after a test sometimes, but clean up after a session by default.

tests/dragonfly/conftest.py
```diff
@@ -278,6 +285,13 @@ def pytest_addoption(parser):
     )
 
     parser.addoption("--repeat", action="store", help="Number of times to repeat each test")
+    parser.addoption(
+        "--drop-data-after-each-test",
```
abhijat, line 289: This is technically incorrectly worded as we drop after each class, but for the purpose of the action that this supports it is appropriate wording. If we are repeating a test, the data is dropped after each repeat. Naming it `drop after class` is confusing.

abhijat:
example run https://github.com/dragonflydb/dragonfly/actions/runs/16930862486/job/47975849074

PR #5672, opened 2025-08-13
chore(lua): Add lua redis.log function

mkaruza:
Add redis.log function so lua script can now use logging.


kostasrim:
I think the reason we did not add this function was because we thought it's a `security concern` as a `user script` can `make production instances scream`.

CC @romange WDYT?

romange:
Did anyone asked for this? 
Anyways, there is a potential security concern so lets at least guard it with runtime flag, so we could disable the whole logic if needed.

mkaruza:
@romange i had to do some debugging of lua script for previous PR so added this function to show me logs - thought it can be helpful since it is missing. I can close this PR and we can return it later if someone ask for it.

romange:
if you already made it - let's just add a flag to be able to disable the logs and merge it. no point in throwing this work.

mkaruza:
Added flag `lua_enable_redis_log` which enables this functionality. Default is false.

src/core/interpreter.cc
```diff
@@ -551,10 +556,52 @@ int RedisReplicateCommands(lua_State* lua) {
...
+        break;
+      case LL_NOTICE:
+        LOG(INFO) << log;
+        break;
+      case LL_WARNING:
+        LOG(WARNING) << log;
+      default:
+        break;
+    }
+    sdsfree(log);
```
kostasrim, line 604: I am not a fun of raw memory managment. There exists absl::Cleanup to implement RAII for those C cases

src/core/interpreter.cc
```diff
@@ -551,10 +556,52 @@ int RedisReplicateCommands(lua_State* lua) {
...
+
   if (argc < 2) {
     PushError(lua, "redis.log() requires two arguments or more.");
     return RaiseErrorAndAbort(lua);
+  } else if (!lua_isnumber(lua, -argc)) {
+    PushError(lua, "First argument must be a number (log level).");
+    return RaiseErrorAndAbort(lua);
+  }
+
+  if (absl::GetFlag(FLAGS_lua_enable_redis_log)) {
```
kostasrim, line 570: noop when this flag is off right?

mkaruza: Yes

Issue #5569, opened 2025-07-26
Failing regression test: cluster_test.py::test_replicate_disconnect_cluster


vyavdoshenko:
Error log:
```
E20250726 15:08:14.208519 20670 protocol_client.cc:194] Error while calling sock_->Connect(server_context_.endpoint): Software caused connection abort
E20250726 15:08:14.710047 20670 protocol_client.cc:194] Error while calling sock_->Connect(server_context_.endpoint): Software caused connection abort
E20250726 15:08:15.210798 20670 protocol_client.cc:194] Error while calling sock_->Connect(server_context_.endpoint): Software caused connection abort
E20250726 15:08:15.711216 20670 protocol_client.cc:194] Error while calling sock_->Connect(server_context_.endpoint): Software caused connection abort
E20250726 15:08:16.212091 20670 protocol_client.cc:194] Error while calling sock_->Connect(server_context_.endpoint): Software caused connection abort
```

https://github.com/dragonflydb/dragonfly/actions/runs/16541044923

BorysTheDev:
From the logs, not all the data was transferred through the journal to the replica before the disconnection.
All broken data is from the second node that was not disconnected by proxy.

Issue #5582, opened 2025-07-28
Python test fails: test_replicate_disconnect_redis_cluster


vyavdoshenko:
https://github.com/dragonflydb/dragonfly/actions/runs/16576647977/job/46882465618

```
        await c_replica.execute_command("REPLICAOF NO ONE")
        capture = await seeder.capture()
>       assert await seeder.compare(capture, replica.port)
E       assert False
```

Error log:
```
E20250728 18:23:26.681980 22567 protocol_client.cc:236] Could not shutdown socket system:107
```

BorysTheDev:
The broken data is from the node that wasn't disconnected

PR #5673, opened 2025-08-13
fix: add sleep after seeder stop to be sure that all data is transfered

BorysTheDev:
fixes: #5569
problem: even when the replication in stable_sync state and lag is 0 we don't have a guarantee that there are no unprocessed data on the master node.

fix: added 2 seconds sleep after finishing the seeder task, to finish all data processing

### DIFF
--- a/tests/dragonfly/cluster_test.py
+++ b/tests/dragonfly/cluster_test.py
@@ -2370,6 +2370,9 @@ async def test_replicate_cluster(df_factory: DflyInstanceFactory, df_seeder_fact
     # wait for replication to finish
     await asyncio.gather(*(asyncio.create_task(await_no_lag(c)) for c in c_nodes))
 
+    # await_no_lag doesn't guarantee that the replica got all last changes, so we wait a bit more
+    await asyncio.sleep(2)
+
     # promote replica to master and compare data
     await c_replica.execute_command("REPLICAOF NO ONE")
     capture = await seeder.capture()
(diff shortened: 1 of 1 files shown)


tests/dragonfly/cluster_test.py
```diff
@@ -2487,6 +2490,9 @@ async def is_first_master_conn_down(conn):
     # wait for no lag on all cluster nodes
     await asyncio.gather(*(asyncio.create_task(await_no_lag(c)) for c in c_nodes))
 
+    # await_no_lag doesn't guarantee that the replica got all last changes, so we wait a bit more
```
kostasrim, line 2493: could you plz explain why `await_no_lag` does not guarantee that the replica does not contain all the changes ?

tests/dragonfly/cluster_test.py
```diff
@@ -2370,6 +2370,9 @@ async def test_replicate_cluster(df_factory: DflyInstanceFactory, df_seeder_fact
     # wait for replication to finish
     await asyncio.gather(*(asyncio.create_task(await_no_lag(c)) for c in c_nodes))
 
+    # await_no_lag doesn't guarantee that the replica got all last changes, so we wait a bit more
+    await asyncio.sleep(2)
```
kostasrim, line 2374: This is `test_replicate_cluster` yet you mention in your PR description that you fixed https://github.com/dragonflydb/dragonfly/issues/5582

tests/dragonfly/cluster_test.py
BorysTheDev: we send lsn not after every command and detect lag according lsn mark

tests/dragonfly/cluster_test.py
BorysTheDev: I read the test name incorectly

tests/dragonfly/cluster_test.py
kostasrim: Why was the conversation resolved ? :smile: 

I hardly see how this is the case. We wait above for the seeder to complete which means that at least all the commands went through the journal (maybe the actual journal entries have not yet been sent, but *at least* the `lsn_` of each journal/shard will not increase any more).

By the time we call `await_no_lag` we do: `lsn_of_master - last_acked_lsn` (with initial values of `1` and `0` respectively).  The former, has reached it's end value (`lsn_of_master` -- there are no other commands coming through) and the later `last_acked_lsn` is monotonically increasing until it reaches `lsn_of_master`. Their diff will become `0` and at that time both of them are in sync.

PR #5674, opened 2025-08-14
chore: group memory prometheous metrics

### LINKED ISSUE #5455: group all the memory usage metrics under the single metrics with distinct labels
Currently our memory /metrics are quite chaotic.

We should group all the disjoint memory contributors like:

dragonfly_used_memory_lua, dragonfly_memory_fiberstack_vms_bytes, dragonfly_tls_bytes, dragonfly_replication_full_sync_bytes, dragonfly_replication_streaming_bytes, dragonfly_commands_squashing_replies_bytes, dragonfly_pipeline_cmd_cache_bytes, 
dragonfly_dispatch_queue_bytes, dragonfly_client_read_buffer_bytes 

into the same variable: `dragonfly_memory_by_class_bytes` and use labels to mark each contributor.

In addition, `table_used_memory` and `object_used_memory` appear in INFO but not in metrics - should be fixed.

Finally, we can not just move the existing metrics - as it will break the compatibility. So we should still render the old metrics under the flag `--keep_legacy_memory_metrics` that will be defaulted to true.

kostasrim:
Groups memory usage under `metrics dragonfly_memory_by_class_bytes`

resolves #5455

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -1518,8 +1518,15 @@ void AppendMetricWithoutLabels(string_view name, string_view help, const absl::A
   AppendMetricValue(name, value, {}, {}, dest);
 }
 
+#define AppendIfLegacy(f) \
+  do {                    \
+    if (legacy) {         \
+      f;                  \
+    }                     \
+  } while (0)
+
 void PrintPrometheusMetrics(uint64_t uptime, const Metrics& m, DflyCmd* dfly_cmd,
-                            StringResponse* resp) {
+                            StringResponse* resp, bool legacy) {
   // Server metrics
   AppendMetricHeader("version", "", MetricType::GAUGE, &resp->body());
   AppendMetricValue("version", 1, {"version"}, {GetVersion()}, &resp->body());
@@ -1539,12 +1546,14 @@ void PrintPrometheusMetrics(uint64_t uptime, const Metrics& m, DflyCmd* dfly_cmd
                     &resp->body());
   AppendMetricValue("connected_clients", conn_stats.num_conns_other, {"listener"}, {"other"},
                     &resp->body());
-  AppendMetricWithoutLabels("client_read_buffer_bytes", "", conn_stats.read_buf_capacity,
-                            MetricType::GAUGE, &resp->body());
+  AppendIfLegacy(AppendMetricWithoutLabels("client_read_buffer_bytes", "",
+                                           conn_stats.read_buf_capacity, MetricType::GAUGE,
+                                           &resp->body()));
   AppendMetricWithoutLabels("blocked_clients", "", conn_stats.num_blocked_clients,
                             MetricType::GAUGE, &resp->body());
-  AppendMetricWithoutLabels("dispatch_queue_bytes", "", conn_stats.dispatch_queue_bytes,
-                            MetricType::GAUGE, &resp->body());
+  AppendIfLegacy(AppendMetricWithoutLabels("dispatch_queue_bytes", "",
+                                           conn_stats.dispatch_queue_bytes, MetricType::GAUGE,
+                                           &resp->body()));
   AppendMetricWithoutLabels("pipeline_queue_length", "", conn_stats.dispatch_queue_entries,
                             MetricType::GAUGE, &resp->body());
   AppendMetricWithoutLabels("send_delay_seconds", "",
@@ -1553,8 +1562,9 @@ void PrintPrometheusMetrics(uint64_t uptime, const Metrics& m, DflyCmd* dfly_cmd
 
   AppendMetricWithoutLabels("pipeline_throttle_total", "", conn_stats.pipeline_throttle_count,
                             MetricType::COUNTER, &resp->body());
-  AppendMetricWithoutLabels("pipeline_cmd_cache_bytes", "", conn_stats.pipeline_cmd_cache_bytes,
-                            MetricType::GAUGE, &resp->body());
+  AppendIfLegacy(AppendMetricWithoutLabels("pipeline_cmd_cache_bytes", "",
+                                           conn_stats.pipeline_cmd_cache_bytes, MetricType::GAUGE,
+                                           &resp->body()));
   AppendMetricWithoutLabels("pipeline_commands_total", "", conn_stats.pipelined_cmd_cnt,
                             MetricType::COUNTER, &resp->body());
   AppendMetricWithoutLabels("pipeline_dispatch_calls_total", "", conn_stats.pipeline_dispatch_calls,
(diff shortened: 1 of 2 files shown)


src/server/server_family.cc
```diff
@@ -1860,6 +1863,369 @@ void PrintPrometheusMetrics(uint64_t uptime, const Metrics& m, DflyCmd* dfly_cmd
...
+  AppendMetricWithoutLabels("failed_restores_total", "", m.loading_stats.failed_restore_count,
+                            MetricType::COUNTER, &resp->body());
+
+  // Net metrics
+  AppendMetricWithoutLabels("net_input_recv_total", "", conn_stats.io_read_cnt, MetricType::COUNTER,
+                            &resp->body());
+  AppendMetricWithoutLabels("net_read_yields_total", "", conn_stats.num_read_yields,
+                            MetricType::COUNTER, &resp->body());
+
+  AppendMetricWithoutLabels("net_input_bytes_total", "", conn_stats.io_read_bytes,
```
kostasrim: @romange you did not include this in the list as well as a few more. Shall I also move them under ? dragonfly_memory_by_class_bytes?

src/server/server_family.cc
```diff
@@ -1860,6 +1863,369 @@ void PrintPrometheusMetrics(uint64_t uptime, const Metrics& m, DflyCmd* dfly_cmd
   absl::StrAppend(&resp->body(), db_key_metrics, db_key_expire_metrics, db_capacity_metrics);
 }
 
+void PrintPrometheusMetricsV2(uint64_t uptime, const Metrics& m, DflyCmd* dfly_cmd,
```
kostasrim: The previous function was already huge and adding yet another `if/else` with two groups of behaviours (because of the new flag) seemed an overkill. So I just copy pasted and added a new group (and rearranged since metrics where scattered all over it). I can comment on which are the additions/diff here. 

What's more, I would like to move some of the metrics to a separate header/source. `server_family.cc` is huge (4k+lines). 

p.s. we can remove the old one once we remove the flag

kostasrim: deeply regretting this function. will isolate the changes

romange: I do not think we need `if/else` - we need to add NEW metrics to the old ones. because we need to push the new binary for the monitoring environment to transition and ONLY after we can remove the old metrics.
So I think the first PR should be to group the metrics we plan to deprecate together in the code and then we should add the new metrics in the same function. 

Whether to split server_family.c is a separate discussion and I am not against it but either do it before this PR or after

src/server/server_family.cc
```diff
@@ -1732,6 +1746,42 @@ void PrintPrometheusMetrics(uint64_t uptime, const Metrics& m, DflyCmd* dfly_cmd
...
   AppendMetricWithoutLabels("evicted_keys_total", "", m.events.evicted_keys, MetricType::COUNTER,
                             &resp->body());
+  // Memory stats
+  string memory_by_class_bytes;
+  if (legacy) {
+    AppendMetricHeader("memory_by_class_bytes", "Memory metrics", MetricType::GAUGE,
+                       &memory_by_class_bytes);
+
+    AppendMetricValue("memory_by_class_bytes", m.lua_stats.used_bytes, {"label"},
+                      {"used_memory_lua"}, &memory_by_class_bytes);
```
kostasrim: @romange is this the format that you want ? 

The output as is:

```
dragonfly_memory_by_class_bytes{label="used_memory_lua"} 0
dragonfly_memory_by_class_bytes{label="memory_fiberstack_vms_bytes"} 3817440
dragonfly_memory_by_class_bytes{label="tls_bytes"} 0
dragonfly_memory_by_class_bytes{label="commands_squashing_replies_bytes"} 0
etc...
```

src/server/server_family.cc
kostasrim: @romange done. plz check the format and I will move the rest and open this for review!

src/server/server_family.cc
```diff
@@ -1732,6 +1718,64 @@ void PrintPrometheusMetrics(uint64_t uptime, const Metrics& m, DflyCmd* dfly_cmd
...
+  AppendMetricWithoutLabels("used_memory_lua", "", m.lua_stats.used_bytes, MetricType::GAUGE,
+                            &resp->body());
+
+  AppendMetricWithoutLabels("client_read_buffer_bytes", "", conn_stats.read_buf_capacity,
+                            MetricType::GAUGE, &resp->body());
+  AppendMetricWithoutLabels("dispatch_queue_bytes", "", conn_stats.dispatch_queue_bytes,
+                            MetricType::GAUGE, &resp->body());
+  AppendMetricWithoutLabels("pipeline_cmd_cache_bytes", "", conn_stats.pipeline_cmd_cache_bytes,
+                            MetricType::GAUGE, &resp->body());
+  if (!legacy) {
```
romange: the condition should be around the legacy metrics not around the new ones.

src/server/server_family.cc
romange: we can normalize and simplify labels as well, i.e. `used_memory_lua` -> `used_lua`. 
`tls_bytes` -> `tls`,  `commands_squashing_replies_bytes` -> `squashing_replies`  etc

src/server/server_family.cc
```diff
@@ -1762,14 +1806,20 @@ void PrintPrometheusMetrics(uint64_t uptime, const Metrics& m, DflyCmd* dfly_cmd
...
     vector<ReplicaRoleInfo> replicas_info = dfly_cmd->GetReplicasRoleInfo();
-
     ReplicationMemoryStats repl_mem;
     dfly_cmd->GetReplicationMemoryStats(&repl_mem);
     AppendMetricWithoutLabels("replication_streaming_bytes", "Stable sync replication memory usage",
                               repl_mem.streamer_buf_capacity_bytes, MetricType::GAUGE,
                               &resp->body());
     AppendMetricWithoutLabels("replication_full_sync_bytes", "Full sync memory usage",
                               repl_mem.full_sync_buf_bytes, MetricType::GAUGE, &resp->body());
+    if (!legacy) {
```
romange: again, legacy should be around legacy. no need to make NEW metrics conditional - they do not break anything

src/server/server_family.cc
```diff
@@ -1732,6 +1718,65 @@ void PrintPrometheusMetrics(uint64_t uptime, const Metrics& m, DflyCmd* dfly_cmd
...
+                              MetricType::GAUGE, &resp->body());
+    AppendMetricWithoutLabels("pipeline_cmd_cache_bytes", "", conn_stats.pipeline_cmd_cache_bytes,
+                              MetricType::GAUGE, &resp->body());
+  }
+
+  string memory_by_class_bytes;
+  AppendMetricHeader("memory_by_class_bytes", "Memory metrics", MetricType::GAUGE,
+                     &memory_by_class_bytes);
+
+  AppendMetricValue("memory_by_class_bytes", m.lua_stats.used_bytes, {"label"}, {"used_lua"},
```
romange: "label" -> "class"

kostasrim: done

Issue #5675, opened 2025-08-15
Failing regression test: Crash while running test_migration_timeout_on_sync


mkaruza:
https://github.com/dragonflydb/dragonfly/actions/runs/16968971012/job/48100435406

```
               await self.disconnect(nowait=True)
>           raise ConnectionError(f"Error while reading from {host_error} : {e.args}")
E           redis.exceptions.ConnectionError: Error while reading from localhost:30122 : (104, 'Connection reset by peer')

/usr/local/lib/python3.8/dist-packages/redis/asyncio/connection.py:563: ConnectionError
------------------------------ Captured log setup ------------------------------
[2025-08-14 18:35:39.903 DEBUG] Random seed: 2146869263348414278, check: 9
----------------------------- Captured stdout call -----------------------------
.......30121➜    @     0xf48543e758f8       4832  (unknown)

30121➜    @     0xf48543968aac        304  abort

30121➜    @          0x18c37fc        336  google::DumpStackTraceAndExit()

30121➜    @          0x18bb190        176  google::LogMessage::Fail()

30121➜    @          0x18bb09c         16  google::LogMessage::SendToLog()

30121➜    @          0x18ba908        160  google::LogMessage::Flush()

30121➜    @          0x18be0d8         80  google::LogMessageFatal::~LogMessageFatal()

30121➜    @          0x114be94         32  dfly::RestoreStreamer::OnDbChange()::{lambda()#1}::operator()()

30121➜    @          0x114cbc8        176  dfly::DashTable<>::CVCUponInsert<>()

30121➜    @          0x114c07c        128  dfly::RestoreStreamer::OnDbChange()

30121➜    @          0x11555dc        224  std::__invoke_impl<>()

30121➜    @          0x1155448         96  std::__invoke<>()

30121➜    @          0x1154f18         80  std::invoke<>()

30121➜    @          0x1154584         80  absl::lts_20250512::functional_internal::Apply<>()

30121➜    @          0x1153540         80  absl::lts_20250512::functional_internal::FrontBinder<>::operator()<>()

30121➜    @          0x11521b4         80  std::_Function_handler<>::_M_invoke()

30121➜    @          0x104e498         64  std::function<>::operator()()

30121➜    @          0x103bcbc         80  dfly::DbSlice::CallChangeCallbacks()

30121➜    @          0x1033918        224  dfly::DbSlice::AddOrFindInternal()

[...]

PR #5676, opened 2025-08-15
fix: prevent writeBucket if migration is canceled

BorysTheDev:
fixes: https://github.com/dragonflydb/dragonfly/issues/5675

problem: we try to write a bucket after a migration is canceled

fix: added checks to prevent bucket writing if migration is canceled

### DIFF
--- a/src/server/journal/streamer.cc
+++ b/src/server/journal/streamer.cc
@@ -417,6 +417,10 @@ void RestoreStreamer::Run() {
       auto* blocking_counter = db_slice_->GetLatch();
       lock_guard blocking_counter_guard(*blocking_counter);
 
+      if (!cntx_->IsRunning())  // Could have been cancelled during throttling or the lock
+                                // acquisition
+        return;
+
       stats_.buckets_loop += WriteBucket(it);
     });
 
@@ -519,6 +523,9 @@ bool RestoreStreamer::WriteBucket(PrimeTable::bucket_iterator it) {
 
         WriteEntry(key, it->first, pv, expire);
         written = true;
+        if (!cntx_->IsRunning()) {
+          break;  // Could have been cancelled during WriteEntry
+        }
       } else {
         stats_.keys_skipped++;
       }
@@ -526,7 +533,7 @@ bool RestoreStreamer::WriteBucket(PrimeTable::bucket_iterator it) {
   } else {
     stats_.buckets_skipped++;
   }
-  ThrottleIfNeeded();
+  // we don't need throttle here, because we throttle after every entry written
 
   return written;
 }
@@ -540,12 +547,16 @@ void RestoreStreamer::OnDbChange(DbIndex db_index, const DbSlice::ChangeReq& req
   uint64_t throttle_start = throttle_count_;
   uint64_t throttle_usec_start = total_throttle_wait_usec_;
   if (const PrimeTable::bucket_iterator* bit = req.update()) {
-    stats_.buckets_on_db_update += WriteBucket(*bit);
+    if (cntx_->IsRunning()) {
+      stats_.buckets_on_db_update += WriteBucket(*bit);
+    }
   } else {
     string_view key = get<string_view>(req.change);
     table->CVCUponInsert(snapshot_version_, key, [this](PrimeTable::bucket_iterator it) {
-      DCHECK_LT(it.GetVersion(), snapshot_version_);
-      stats_.buckets_on_db_update += WriteBucket(it);
+      if (cntx_->IsRunning()) {
+        DCHECK_LT(it.GetVersion(), snapshot_version_);
+        stats_.buckets_on_db_update += WriteBucket(it);
+      }
     });
   }
   stats_.throttle_on_db_update += throttle_count_ - throttle_start;
--- a/src/server/cluster/outgoing_slot_migration.cc
+++ b/src/server/cluster/outgoing_slot_migration.cc
@@ -142,7 +142,7 @@ void OutgoingMigration::OnAllShards(
   });
 }
 
-void OutgoingMigration::Finish(GenericError error) {
+void OutgoingMigration::Finish(const GenericError& error) {
   auto next_state = MigrationState::C_FINISHED;
   if (error) {
     // If OOM error move to FATAL, non-recoverable  state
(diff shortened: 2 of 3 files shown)


src/server/journal/streamer.cc
```diff
@@ -417,6 +417,10 @@ void RestoreStreamer::Run() {
       auto* blocking_counter = db_slice_->GetLatch();
       lock_guard blocking_counter_guard(*blocking_counter);
 
+      if (!cntx_->IsRunning())  // Could have been cancelled during throttling or the lock
```
romange: there are 3 checks like this in this function. I understand why we need it after the last preemption. Why do we need them at line 408, line 402?

src/server/journal/streamer.cc
```diff
@@ -519,14 +523,17 @@ bool RestoreStreamer::WriteBucket(PrimeTable::bucket_iterator it) {
 
         WriteEntry(key, it->first, pv, expire);
         written = true;
+        if (!cntx_->IsRunning()) {
```
romange: why do we need it here? why `WriteBucket` can not finish even if the operation was cancelled?

src/server/journal/streamer.cc
```diff
@@ -540,12 +547,16 @@ void RestoreStreamer::OnDbChange(DbIndex db_index, const DbSlice::ChangeReq& req
   uint64_t throttle_start = throttle_count_;
   uint64_t throttle_usec_start = total_throttle_wait_usec_;
   if (const PrimeTable::bucket_iterator* bit = req.update()) {
-    stats_.buckets_on_db_update += WriteBucket(*bit);
+    if (cntx_->IsRunning()) {
```
romange: same question, what data structure is deleted under our feet?

romange:
The problem is not clear. What exactly crashes when we try to write when a migration is cancelled? please be more precise in the PR description.

src/server/journal/streamer.cc
BorysTheDev: To check if we still need to proceed. We should check it after every possible preemtion. It is not critical in some places, but we shouldn't write data if the process is canceled

src/server/journal/streamer.cc
BorysTheDev: we can remove it from here, but why we need to proceed with the write bucket if the operation is canceled

src/server/journal/streamer.cc
BorysTheDev: writeBucket can be a hard operation, if we don't need it why we should do it?

BorysTheDev:
> The problem is not clear. What exactly crashes when we try to write when a migration is cancelled? please be more precise in the PR description.

There is no crash `DCHECK_LT(it.GetVersion(), snapshot_version_);` failed because the migration was canceled and snapshot_version_ = 0

src/server/journal/streamer.cc
romange: because the style of repeating the same check after each line should raise the red flag that maybe something is wrong with the approach and your fix makes the code unmaintainable. Why do we crash? maybe the code that delete resources needs to wait for the fiber to finish instead of adding checks after each line?

src/server/journal/streamer.cc
romange: but why do we crash? can we write data until the next iteration?

romange:
and why snapshot_version_ is 0?

src/server/journal/streamer.cc
BorysTheDev: we don't have a crash here, and can proceed with writing.

src/server/journal/streamer.cc
BorysTheDev: We don't have a crash

BorysTheDev:
> and why snapshot_version_ is 0?

```
void RestoreStreamer::Cancel() {
  auto sver = snapshot_version_;
  snapshot_version_ = 0;  // to prevent double cancel in another fiber
  cntx_->Cancel();
  if (sver != 0) {
    db_slice_->UnregisterOnChange(sver);
    JournalStreamer::Cancel();
  }
}
```

UnregisterOnChange should be called only once

romange:
Ok, I think the more robust approach would be to return early in `RestoreStreamer::OnDbChange` if snapshot_version_ == 0  and add a comment why we do this

src/server/journal/streamer.cc
romange: because it's a rare event that does not change anything in terms of the performance.

src/server/journal/streamer.cc
kostasrim: `WriteEntry` is not preemptive -- it just serializes data. How can the context get cancelled inbetween if the fiber does not yield ? 

second, we do the exact same check within `ThrottleIfNeeded`

kostasrim: Uh I am wrong, it writes to the journal. Edit: socket ops are async.

kostasrim:
> Ok, I think the more robust approach would be to return early in `RestoreStreamer::OnDbChange` if snapshot_version_ == 0 and add a comment why we do this

Why do this ? If the issue is that we set `snapshot version to 0` and then preempt during the `UnregisterOnChange` then simply don't set the snapshot version `0` unless we `deregistered` the call back. That way all the flows should work without any additional change and the fix is isolated only in `Cancel()`

To prevent the double cancelling just use a separate flag... 

CC @BorysTheDev

kostasrim:
Also, thinking out loud here -- I suspect we have another low hanging fruit. @BorysTheDev plz verify what I write here.

Based on my understanding the bug that leads to the dcheck failing can be described as follows:

1. `RestoreStreamer::Run()` iterates over the buckets. At some point in its body, it grabs the `latch` via `auto* blocking_counter = db_slice_->GetLatch();`
2. Within that same flow, after we call `WriteBucket` we call `ThrottleIfNeeded` and preempt. So basically we `preempt after we finished serializing the bucket but **before** we release the latch
3. Cancellation kicks in, we call `UnregisterOnChange` which `blocks` because the lach is taken in (2)
4. We get a write command on a key whose bucket is not yet serialized (because OnDbChange is called). We now need to serialize the bucket and `boom` DCHECK fail

If that's accurate, I think we should also release the latch before we preempt in `ThrottleIfNeeded` within `WriteBucket`. In other words, it should be the responsibility of the caller to manually control the preemption of `WriteBucket` because as I explained there are flow that increment the latch and keep it incremented otherwise for the duration of the preemption -- which is not good :smile:

BorysTheDev:
@kostasrim You misunderstand the problem a little. we can preempt in different places even in other callbacks (SAVE, replication etc), so checking snapshot_version_ == 0 should be enough to fix this bug.

kostasrim:
> @kostasrim You misunderstand the problem a little. we can preempt in different places even in other callbacks (SAVE, replication etc), so checking snapshot_version_ == 0 should be enough to fix this bug.

RestoreStreamer is not used anywhere else other than the outgoing migrations :facepalm:

PR #5679, opened 2025-08-16
chore: add more detailed timings for slow squash hops

romange:
We identify a callback that likely to contribute to the slowness of squashing and try to understand the reason.
The logging is tunable based on `log_squash_info_threshold_usec` parameter.

### DIFF
--- a/src/server/multi_command_squasher.cc
+++ b/src/server/multi_command_squasher.cc
@@ -264,7 +264,8 @@ bool MultiCommandSquasher::ExecuteSquashed(facade::RedisReplyBuilder* rb) {
     fb2::BlockingCounter bc(num_shards);
     DVLOG(1) << "Squashing " << num_shards << " " << tx->DebugId();
 
-    const uint64_t min_threshold_cycles = CycleClock::FromUsec(100);
+    // Saves work in case logging is disable (i.e. log_squash_threshold_cached is high).
+    const uint64_t min_threshold_cycles = CycleClock::FromUsec(log_squash_threshold_cached / 5);
     auto cb = [&, bc, rb]() mutable {
       uint64_t sched_time = CycleClock::Now() - start;
 


dranikpg (dismissed):

src/server/multi_command_squasher.cc
```diff
@@ -258,19 +261,37 @@ bool MultiCommandSquasher::ExecuteSquashed(facade::RedisReplyBuilder* rb) {
...
+      // Update max_sched_cycles in lock-free fashion, to avoid contention
+      uint64_t current = max_sched_cycles.load(memory_order_relaxed);
+      while (sched_time > current) {
+        if (max_sched_cycles.compare_exchange_weak(current, sched_time, memory_order_relaxed,
+                                                   memory_order_relaxed)) {
+          fiber_running_cycles.store(ThisFiber::GetRunningTimeCycles(), memory_order_relaxed);
+          proactor_running_cycles.store(ProactorBase::me()->GetCurrentBusyCycles(),
+                                        memory_order_relaxed);
+          break;
+        }
```
dranikpg: It's really not easy to understand what you do from the code. I'd add a comment like "storing stats from thread with max sched time"

romange: let me know if it's better now

src/server/multi_command_squasher.cc
```diff
@@ -258,19 +264,46 @@ bool MultiCommandSquasher::ExecuteSquashed(facade::RedisReplyBuilder* rb) {
     fb2::BlockingCounter bc(num_shards);
     DVLOG(1) << "Squashing " << num_shards << " " << tx->DebugId();
 
-    auto cb = [this, bc, rb]() mutable {
+    const uint64_t min_threshold_cycles = CycleClock::FromUsec(100);
+    auto cb = [&, bc, rb]() mutable {
+      uint64_t sched_time = CycleClock::Now() - start;
```
kostasrim, line 270: All of this to synchronize threads around a shared state which is used to find maximums. Even if it's lockless with relaxed operations these still are global operations on shared memory. 

Why don't instead create a vector of results and do the operations locally and unconditionally. Later, when you check:
`if (total_usec > log_squash_threshold_cached)` simply `std::find_if` the one that took the most time and print that?

The only downside I see is the `past_fibers` allocation is now done unconditionally. But maybe there is a way to bound that ?

romange: they are not global as atomics are local to this connection, each connection has its own atomics.
In any case, I added an optimization to save on these operations in case the logging is disabled. PTAL

PR #5685, opened 2025-08-18
fix: RdbSaver: destroy SliceSnapshot on its shard thread to fix flaky crash on ARM64 with --force_epoll

vyavdoshenko:
Fixes regression crash: https://github.com/dragonflydb/dragonfly/actions/runs/17025808197/job/48261017365
Follow up: https://github.com/dragonflydb/dragonfly/pull/5655 

The Python test snapshot_test.py::test_parallel_snapshot_race_condition uncovered an additional case. It occurs only in the arm64/epoll environment.

I could reproduce it only on a large number of executions:
```
pytest -q tests/dragonfly/snapshot_test.py::test_parallel_snapshot_race_condition --df=force_epoll --df=proactor_threads=2 --df=num_shards=1 --repeat 20 -vv
```

Root cause: for single-snapshot flows (shard_snapshots_.size() == 1), the snapshot was destroyed from the caller thread, violating shard thread-affinity. In some paths, it could also lead to an indexing mismatch.

Fix: 
- Track the owning shard id on snapshot start (single_owner_sid_).
- For single-snapshot flows, destroy SliceSnapshot on the correct shard via shard_set->Await(owner_sid, ...).
- For multi-snapshot flows, destroy per shard via RunBlockingInParallel, indexing by es->shard_id().
- This enforces the SliceSnapshot thread-affinity invariant instead of weakening the DCHECK.

### DIFF
--- a/src/server/rdb_save.cc
+++ b/src/server/rdb_save.cc
@@ -1025,7 +1025,17 @@ error_code AlignedBuffer::Flush() {
 
 // Ensures SliceSnapshot is destroyed on its owning shard thread.
 struct OwnerThreadDeleter {
-  ShardId owner_sid = 0;
+  ShardId owner_sid;
+
+  OwnerThreadDeleter() : owner_sid(0) {
+  }
+
+  explicit OwnerThreadDeleter(ShardId sid) : owner_sid(sid) {
+  }
+
+  static OwnerThreadDeleter FromShard(EngineShard* shard) {
+    return OwnerThreadDeleter(shard->shard_id());
+  }
 
   void operator()(SliceSnapshot* ptr) const {
     if (!ptr)
@@ -1045,6 +1055,7 @@ using SnapshotPtr = std::unique_ptr<SliceSnapshot, OwnerThreadDeleter>;
 class RdbSaver::Impl final : public SliceSnapshot::SnapshotDataConsumerInterface {
  private:
   void CleanShardSnapshots();
+  SnapshotPtr CreateSliceSnapshot(EngineShard* shard, DbSlice* db_slice, ExecutionState* cntx);
 
  public:
   // We pass K=sz to say how many producers are pushing data in order to maintain
@@ -1137,14 +1148,8 @@ RdbSaver::Impl::Impl(bool align_writes, unsigned producers_len, CompressionMode 
 }
 
 void RdbSaver::Impl::CleanShardSnapshots() {
-  if (shard_snapshots_.empty()) {
-    return;
-  }
-
   // Deleter dispatches destruction to the owning shard thread when needed
-  for (auto& snap : shard_snapshots_) {
-    snap.reset();
-  }
+  shard_snapshots_.clear();
 }
 
 RdbSaver::Impl::~Impl() {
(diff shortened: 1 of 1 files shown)


src/server/rdb_save.cc
```diff
@@ -1122,15 +1125,18 @@ void RdbSaver::Impl::CleanShardSnapshots() {
...
-  auto cb = [this](ShardId sid) {
-    // Destroy SliceSnapshot in target thread, as it registers itself in a thread local set.
-    shard_snapshots_[sid].reset();
-  };
-
+  // Destroy SliceSnapshot in the correct shard thread, as it registers itself in a thread-local
+  // set.
   if (shard_snapshots_.size() == 1) {
-    cb(0);
+    if (single_owner_sid_.has_value()) {
```
romange: I am trying to understand what is the case you are trying to fix. 
If Dragonfly runs with a single shard, then `cb(0)` is probably correct (or am I wrong)?
if dragonfly runs with multiple shards, then `shard_snapshots_.size() == 1` is only for RDB  snapshot?
so is this the state you are trying to fix? RDB snapshot with multiple threads and the snapshot is taken on non-0 thread?

vyavdoshenko: Index 0 exists in two cases:
- A – single engine shard (num_shards == 1): there is only one shard thread; the only snapshot lives at index 0 and its owner is shard_id 0. Destroying on shard 0 is both index-correct and thread-correct.
- B – multi-shard engine with a single-snapshot producer: we intentionally create only one producer, so shard_snapshots_.size() == 1. Internally, we always store that snapshot at index 0 (a storage slot), regardless of which shard thread actually created/owns it. In this case, index 0 does not mean shard_id 0; the owner can be any shard_id.

Conditions:
- Calling cb(0) inline in case B destroys the snapshot from the caller thread, violating thread-affinity (hence the DCHECK in SliceSnapshot::~SliceSnapshot()).
- Using RunBlockingInParallel and indexing by es->shard_id() when size() == 1 causes OOB if shard_id != 0.

Why this surfaced now - After the SAVE/INFO sync refactor [#5655](https://github.com/dragonflydb/dragonfly/pull/5655), cleanup more often executes on the connection thread, which exposed the possible thread-affinity violation under ARM64/epoll.

vyavdoshenko: BTW, C++ unit tests cover Case A, but the mentioned above Python test exposed Case B.

romange: ok, so the bug is in case B. 
what where this snapshot producer is created? It is simpler to always create it in thread 0 for case B then adding more special cases.

romange (commented):
LGTM

vyavdoshenko:
> LGTM

I started the Python test for two configurations, epoll and iouring. I want to double-check it.

src/server/rdb_save.cc
```diff
@@ -1122,15 +1141,9 @@ void RdbSaver::Impl::CleanShardSnapshots() {
...
-    // Destroy SliceSnapshot in target thread, as it registers itself in a thread local set.
-    shard_snapshots_[sid].reset();
-  };
-
-  if (shard_snapshots_.size() == 1) {
-    cb(0);
-  } else {
-    shard_set->RunBlockingInParallel([&](EngineShard* es) { cb(es->shard_id()); });
+  // Deleter dispatches destruction to the owning shard thread when needed
+  for (auto& snap : shard_snapshots_) {
```
BorysTheDev: maybe shard_snapshots_.clear() ?

src/server/rdb_save.cc
```diff
@@ -1218,7 +1231,8 @@ void RdbSaver::Impl::StartSnapshotting(bool stream_journal, ExecutionState* cntx
   auto& s = GetSnapshot(shard);
   auto& db_slice = namespaces->GetDefaultNamespace().GetDbSlice(shard->shard_id());
 
-  s = std::make_unique<SliceSnapshot>(compression_mode_, &db_slice, this, cntx);
+  s = SnapshotPtr(new SliceSnapshot(compression_mode_, &db_slice, this, cntx),
+                  OwnerThreadDeleter{shard->shard_id()});
```
BorysTheDev: I think you can do shard->shard_id() in the OwnerThreadDeleter constructor, or create a factory method makeSliceSnapshot()

src/server/rdb_save.cc
vyavdoshenko: fixed

src/server/rdb_save.cc
vyavdoshenko: fixed

PR #5686, opened 2025-08-19
server: fix max emission limit in throttle command

abhijat:
During throttle command calculation emission interval is computed in milliseconds. If the count is greater than 1000 and the period is 1, this causes a problem because the emission interval is less than 1 millisecond which due to integer division becomes 0.

To accommodate sub-ms interval, nanosecond units are used. Also nanoseconds are stored in the db context, because the current time during throttle is read from db context, so for accurate calculation the source time unit must match the throttle time units.

This results in a larger refactor as db context timestamp is used in many places, so a wrapper helper is added to read current time as milliseconds where required.

Note that the expiry of the throttle key is still stored as ms, because the expiry time is in ms. The expiry time is rounded up by 1 ms. However this should not affect the throttle algorithm as the time value to use for expiry calculation is stored as int with nanosecond precision separate from expiry.


src/server/string_family.cc
```diff
@@ -379,31 +379,27 @@ OpStatus OpMSet(const OpArgs& op_args, const ShardArgs& args) {
   return result;
 }
 
-// emission_interval_ms assumed to be positive
+// emission_interval_ns assumed to be positive
 // limit is assumed to be positive
 OpResult<array<int64_t, 5>> OpThrottle(const OpArgs& op_args, const string_view key,
-                                       const int64_t limit, const int64_t emission_interval_ms,
+                                       const int64_t limit, const int64_t emission_interval_ns,
```
abhijat, line 396: It would be nice to use `std::chrono::nanoseconds/milliseconds` in this function to avoid accidentally mixing up time units, but we do many checks based on `INT64` limits which will cause problems if we use a real time duration unit.

src/server/string_family.cc
```diff
@@ -379,31 +379,27 @@ OpStatus OpMSet(const OpArgs& op_args, const ShardArgs& args) {
...
                                        const uint64_t quantity) {
   auto& db_slice = op_args.GetDbSlice();
 
-  if (emission_interval_ms > INT64_MAX / limit) {
```
romange, line 389: why have you deleted this @abhijat ?

src/server/string_family.cc
```diff
@@ -379,31 +379,27 @@ OpStatus OpMSet(const OpArgs& op_args, const ShardArgs& args) {
...
-  }
-  const int64_t delay_variation_tolerance_ms = emission_interval_ms * limit;  // should be positive
+  // Total size of the bucket
+  const int64_t delay_variation_tolerance_ns = emission_interval_ns * limit;  // should be positive
 
   int64_t remaining = 0;
   int64_t reset_after_ms = -1000;
   int64_t retry_after_ms = -1000;
 
-  if (quantity != 0 && static_cast<uint64_t>(emission_interval_ms) > INT64_MAX / quantity) {
```
romange, line 398: same q

src/server/string_family.cc
```diff
@@ -413,67 +409,86 @@ OpResult<array<int64_t, 5>> OpThrottle(const OpArgs& op_args, const string_view
...
-  reset_after_ms = ttl_ms;
 
   if (!limited) {
+    // Although most computation so far is in nanoseconds, we must store expiry as milliseconds.
+    // While this causes loss of precision, the value stored against the throttle key is still in
+    // the nanosecond units. When the key is loaded, that value will be read and used as tat_ns. The
+    // loss of precision will cause the throttle key to be expired a bit earlier than expected, so
+    // to make up, we extend its expiry by 1 millisecond. Extending the key life does not break
+    // behavior because the tat_ns value will be used to check for throttling.
+    int64_t new_tat_ms = new_tat_ns / 1000000;
```
romange: this is not how you round up division by N. this is how you do it:
`x= (y + N -1) / N`

src/server/string_family.cc
```diff
@@ -1603,14 +1618,22 @@ void StringFamily::ClThrottle(CmdArgList args, const CommandContext& cmnd_cntx)
   if (period > UINT64_MAX / 1000 || count == 0 || period * 1000 / count > INT64_MAX) {
```
romange: seems that 1000 is related to units. I strongly suggest to introduce a constant and use it through the code here.
these checks are important and should be consistent across this code.

src/server/string_family.cc
abhijat: I moved both this and the other check which can be done to before the transaction, it can be done there and return early, we do not need to schedule if there is an error.

src/server/string_family.cc
abhijat: same as https://github.com/dragonflydb/dragonfly/pull/5686#discussion_r2290558285 this can be done before the transaction and fail early if required

src/server/string_family.cc
abhijat: I don't get why rounding up is needed here, we just need to add 1 ms to the result? 

For example if `new_tat_ns` was 1000999, the division will return 1 millisecond and lose the 999 nanoseconds, so we just add 1 ms to cover the loss. Does this miss some case? 

I see we did the same thing at the end of the throttle handler when converting milliseconds to seconds.

src/server/string_family.cc
abhijat: Yes these checks are related, I have changed them to use second to nanosecond multiplier and added constants.

src/server/string_family.cc
abhijat: It looks like your version returns the same result as mine but looks more concise, so I will change to it.

src/server/string_family.cc
```diff
@@ -413,67 +412,78 @@ OpResult<array<int64_t, 5>> OpThrottle(const OpArgs& op_args, const string_view
...
   }
-  const int64_t next_ms = delay_variation_tolerance_ms - ttl_ms;
-  if (next_ms > -emission_interval_ms) {
-    remaining = next_ms / emission_interval_ms;
+  const int64_t next_ns = delay_variation_tolerance_ns - ttl_ns;
+  if (next_ns > -emission_interval_ns) {
+    remaining = next_ns / emission_interval_ns;
   }
-  reset_after_ms = ttl_ms;
+  reset_after_ms = (ttl_ns + kMilliSecondToNanoSecond - 1) / kMilliSecondToNanoSecond;
```
romange, line 474: if ttl_ns is 1000, then (ttl_ns/1000 + 1) is 2, which is wrong. This is the correct math expression to round up divisions

src/server/string_family.cc
```diff
@@ -413,67 +412,78 @@ OpResult<array<int64_t, 5>> OpThrottle(const OpArgs& op_args, const string_view
...
     return OpStatus::INVALID_INT;
   }
-  const int64_t allow_at_ms = new_tat_ms - delay_variation_tolerance_ms;
 
-  if (allow_at_ms >= 0 ? now_ms < INT64_MIN + allow_at_ms : now_ms > INT64_MAX + allow_at_ms) {
+  // The cutoff point before which a request is rejected (throttled) and at or after which a request
+  // is accepted.
+  const int64_t allow_at_ns = new_tat_ns - delay_variation_tolerance_ns;
+
+  if (allow_at_ns >= 0 ? now_ns < INT64_MIN + allow_at_ns : now_ns > INT64_MAX + allow_at_ns) {
```
romange: please factor out this into a separate variable

src/server/string_family.cc
```diff
@@ -413,67 +412,78 @@ OpResult<array<int64_t, 5>> OpThrottle(const OpArgs& op_args, const string_view
...
 
-    if (now_ms >= 0 ? tat_ms < INT64_MIN + now_ms : tat_ms > INT64_MAX + now_ms) {
+    if (now_ns >= 0 ? tat_ns < INT64_MIN + now_ns : tat_ns > INT64_MAX + now_ns) {
       return OpStatus::INVALID_INT;
     }
-    ttl_ms = tat_ms - now_ms;
+    ttl_ns = tat_ns - now_ns;
   } else {
-    if (now_ms >= 0 ? new_tat_ms < INT64_MIN + now_ms : new_tat_ms > INT64_MAX + now_ms) {
+    if (now_ns >= 0 ? new_tat_ns < INT64_MIN + now_ns : new_tat_ns > INT64_MAX + now_ns) {
```
romange: same

romange (commented):
lgtm

src/server/string_family.cc
abhijat: added a function for this kind of check

romange (approved):

PR #5696, opened 2025-08-20
docs: improve description of --maxmemory parameter (units/tiering/minPerThread)

frivoire:
## Why ?
Current documentation isn't giving all useful informations I think, cf:
```
dragonfly --help
[...]
    --maxmemory (Limit on maximum-memory that is used by the database. 0 - means
      the program will automatically determine its maximum memory usage.
      default: 0); default: 0B;
```
=> as a user of DF that is configuring the datastore, how do I know what value to pass here ?

## Solution
Easy fix here: let's just tell precisely the possible values in the description.
NB: I'm really not an expert in Dragonfly, so don't hesitate to review if the possible values are what should be recommended to use 🙏

### DIFF
--- a/src/server/main_service.cc
+++ b/src/server/main_service.cc
@@ -103,8 +103,13 @@ ABSL_FLAG(bool, expose_http_api, false,
           "If set, will expose a POST /api handler for sending redis commands as json array.");
 
 ABSL_FLAG(facade::MemoryBytesFlag, maxmemory, facade::MemoryBytesFlag{},
-          "Limit on maximum-memory that is used by the database. "
-          "0 - means the program will automatically determine its maximum memory usage. "
+          "Limit on maximum-memory that is used by the database, until data starts to be evicted "
+          "(according to eviction policy). With tiering, this value defines only the size in RAM, "
+          "and not the whole dataset (RAM + SSD). "
+          "Must be *at least* 256MiB per proactor thread. "
+          "Can be any human‑readable bytes values (supports K/M/G/T/P/E with optional B, "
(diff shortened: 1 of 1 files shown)


vyavdoshenko:
@frivoire 
Thanks for your contribution.

A few small changes will make it more accurate and consistent with the implementation:
- Clarify auto mode: 0 sets maxmemory to ~80% of available memory.
- Document the minimum: if set explicitly, it must be at least 256MiB per proactor thread; otherwise, the server exits.
- List accepted formats: human‑readable bytes; supports K/M/G/T/P/E with optional B (case‑insensitive). Inputs like "GiB"/"GB" are accepted. Examples: 512MB, 2G, 1.25GiB.

romange:
> @frivoire Thanks for your contribution.
> 
> A few small changes will make it more accurate and consistent with the implementation:
> 
> * Clarify auto mode: 0 sets maxmemory to ~80% of available memory.

I actually think we should not mention 80% as auto mode is implementation dependent. 

> * Document the minimum: if set explicitly, it must be at least 256MiB per proactor thread; otherwise, the server exits.
> * List accepted formats: human‑readable bytes; supports K/M/G/T/P/E with optional B (case‑insensitive). Inputs like "GiB"/"GB" are accepted. Examples: 512MB, 2G, 1.25GiB.

frivoire:
@vyavdoshenko @romange : thanks for those remarks 👍  I tried to include them (and add a few additionnal infos) in my latest commit.
So don't hesitate to have another look :)

vyavdoshenko:
@frivoire 
Please follow the https://github.com/dragonflydb/dragonfly/blob/main/CONTRIBUTING.md to pass the build successfully.

romange (approved):

romange (approved):

Issue #5695, opened 2025-08-20
Python test test_replication_all fails on CI


vyavdoshenko:
https://github.com/dragonflydb/dragonfly/actions/runs/17093726863/job/48473158218

vyavdoshenko:
https://github.com/dragonflydb/dragonfly/actions/runs/17097839205/job/48486403960

vyavdoshenko:
https://github.com/dragonflydb/dragonfly/actions/runs/17102342096/job/48502083273

mkaruza:
Should be fixed in https://github.com/dragonflydb/dragonfly/pull/5697

Issue #5693, opened 2025-08-20
CQE error: 22 cqe_type=1 when trying to upgrade to v1.33.0


Object905:
**Describe the bug**
After trying to upgrade from v1.30.1 to v1.33.0 I get `W20250820 06:17:47.342713    11 uring_proactor.cc:354] CQE error: 22 cqe_type=1` in logs and dragonfly is not responsive. Rolled back quickly.

In logs there are lots of `W20250820 06:17:50.129289    11 uring_proactor.cc:354] CQE error: 22 cqe_type=1` and periodically and some `W20250820 06:17:47.400138    14 engine_shard.cc:653] Stalled heartbeat() fiber for 16 seconds`

**Environment (please complete the following information):**
 - OS: Ubuntu 22.04.3
 - Kernel:  `Linux cache 5.15.0-87-generic #97-Ubuntu SMP Mon Oct 2 21:09:21 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux`
 - Containerized?: Kubernetes
 - Dragonfly Version: 1.33.0

**Additional context**
Libraries used: django-redis, [django-cacheops](https://github.com/Suor/django-cacheops), [python-redis-lock
](https://github.com/ionelmc/python-redis-lock)
I always suspect django-cacheops, because of non-trivial logic with scripts, etc.

Extra args in chart:
```
extraArgs: [
    "--cache_mode",
    "--tcp_nodelay",
    "--tcp_keepalive",
    "60",
    "--pipeline_queue_limit",
    "32768",
    "--dbnum",
    "16",
    "--logtostderr",
    "--stderrthreshold",
    "0",
    #
    "--default_lua_flags",
    "allow-undeclared-keys",
    #
    "--dbfilename",
    "",
    #
    "--requirepass",
    "12345",
    #
    "--maxmemory",
    "4GiB",

    "--maxclients",
    "65536"
  ]
```

Object905:
Also I don't get this in "persistent" instance, which is used only by celery/kombu and arq

romange:
@Object905  can you please try running v1.33.0 with `--uring_wake_mode=0` and see if it helps?

Object905:
Yes, it helps. Thank you. Also seems that I get this error even without any clients.

Weird thing is that persistent instance, on exactly same server and almost exact arguments (except snapshot related) is working fine.
Also persistent instance is run as root, but when I tried to run --cache_mode as root - error remains.

romange:
[...]

Issue #5690, opened 2025-08-19
test_partial_sync fails on CI


vyavdoshenko:
https://github.com/dragonflydb/dragonfly/actions/runs/17069010175/job/48392941042
```
43541➜    @           0x4c00e8        784  std::_Function_handler<>::_M_invoke()

43541➜    @           0xb1d8a0         96  dfly::Interpreter::CallRedisFunction()

43541➜    @           0xb1e9f0        224  dfly::Interpreter::RedisGenericCommand()

43541➜    @           0xbd1300        160  luaD_precall

43541➜    @           0xbdf1a0         80  luaV_execute

43541➜    @           0xbd15f0        240  luaD_callnoyield

43541➜    @           0xbd0490         48  luaD_rawrunprotected

43541➜    @           0xbd1958        384  luaD_pcall

43541➜    @           0xbce29c         64  lua_pcallk

43541➜    @           0xb1cb8c         64  dfly::Interpreter::RunFunction()

43541➜    @           0x4c0568        112  dfly::Service::EvalInternal()

43541➜    @           0x4c0dd8        640  dfly::Service::CallSHA()

43541➜    @           0x4c13d8        304  dfly::Service::EvalSha()

43541➜    @           0xabdbe8        160  dfly::CommandId::Invoke()

43541➜    @           0x4b6f70         80  dfly::Service::InvokeCmd()

43541➜    @           0x4bd0ac        672  dfly::Service::DispatchCommand()

43541➜    @           0xbff118        432  facade::Connection::DispatchSingle()

43541➜    @           0xbff57c        304  facade::Connection::ParseRedis()

43541➜    @           0xc000d4       1632  facade::Connection::IoLoop()

43541➜    @           0xc0077c        272  facade::Connection::ConnectionFlow()

43541➜    @           0xc019f4        368  facade::Connection::HandleRequests()

43541➜    @           0xc639c4        560  util::ListenerInterface::RunSingleConnection()

43541➜    @           0xc63fc0        192  boost::context::detail::fiber_entry<>()
```

vyavdoshenko:
https://github.com/dragonflydb/dragonfly/actions/runs/17084784014/job/48446626178

Issue #5694, opened 2025-08-20
Python test test_client_kill fails on CI


vyavdoshenko:
https://github.com/dragonflydb/dragonfly/actions/runs/17093726863/job/48473158218

vyavdoshenko:
https://github.com/dragonflydb/dragonfly/actions/runs/17118540288/job/48554180663

mkaruza:
Should be fixed in https://github.com/dragonflydb/dragonfly/pull/5697

vyavdoshenko:
https://github.com/dragonflydb/dragonfly/actions/runs/17138967109/job/48621870860

vyavdoshenko:
https://github.com/dragonflydb/dragonfly/actions/runs/17158755331/job/48682306767

vyavdoshenko:
https://github.com/dragonflydb/dragonfly/actions/runs/17170538820/job/48719285063

vyavdoshenko:
https://github.com/dragonflydb/dragonfly/actions/runs/17190240193/job/48764880838

abhijat:
In nearly every instance of failure, this test fails along with `test_acl_cat_commands_multi_exec_squash`, except for https://github.com/dragonflydb/dragonfly/actions/runs/17093726863/job/48473158218

abhijat:
tried to make it fail with combination of other test

```
pytest --maxfail=4 -v tests/dragonfly -k "test_acl_cat_commands_multi_exec_squash or test_client_kill" --df force_epoll=true --count 1000 --repeat-scope session
```

all 4 failures were from `test_acl_cat_commands_multi_exec_squash`

with https://github.com/dragonflydb/dragonfly/pull/5823 the acl test will be re-enabled, will see if this test starts failing again - although the acl one should not fail

abhijat:
closing as it seems not to have failed since fixing `test_acl_cat_commands_multi_exec_squash`

PR #5697, opened 2025-08-21
fix: stack corruption in MultiCommandSquasher

romange:
Fixes https://github.com/dragonflydb/dragonfly/issues/5690 #5695 #5694 #5689 #5688

The problem: we accessed the caller stack variables (max_exec_cycles) in MultiCommandSquasher::ExecuteSquashed.
The access was done after the call to `bc.Dec()`, which unblocks ExecuteSquashed.
As a result, the callback function that was still running could corrupt memory contents of some other variable
as ExecuteSquashed could exit by that time.

In addition, harden checks in stream_family.cc in StreamAppendItem.
Finally, we update helio which fixes https://github.com/dragonflydb/dragonfly/issues/5693

### DIFF
--- a/src/server/stream_family.cc
+++ b/src/server/stream_family.cc
@@ -432,6 +432,7 @@ int StreamAppendItem(stream* s, CmdArgList fields, uint64_t now_ms, streamID* ad
     /* Get a reference to the tail node listpack. */
     lp = (uint8_t*)ri.data;
     lp_bytes = lpBytes(lp);
+    CHECK_GT(lp_bytes, 0U);
   }
   raxStop(&ri);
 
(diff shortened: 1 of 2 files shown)


mkaruza:
Fixes #5695 #5694 #5689 #5688

src/server/stream_family.cc
```diff
@@ -432,6 +432,7 @@ int StreamAppendItem(stream* s, CmdArgList fields, uint64_t now_ms, streamID* ad
     /* Get a reference to the tail node listpack. */
     lp = (uint8_t*)ri.data;
     lp_bytes = lpBytes(lp);
+    CHECK_GT(lp_bytes, 0U);
```
kostasrim, line 435: We already hit it with the core dump and we know it was zero. It's not a harm to keep this. I will follow up adding more of these around trim, del

romange: probably you are right but we hit a bit later in the flow. I just want to have a hard proof of that.

kostasrim: yes I want to push the error higher as well. I will follow up with a patch soon that extends those checks

PR #5700, opened 2025-08-21
fix: list ops transformation during replication

BorysTheDev:
related to: #5413 
problem: LMPOP, RPOPLPUSH, LMOVE can create an undeterministic result on the replica, so they should be transformed into deterministic operations.
fix: LMPOP should be LPOP/RPOP, LMOVE and RPOPLPUSH should be a sequence of POP and PUSH operations

### DIFF
--- a/src/server/list_family.cc
+++ b/src/server/list_family.cc
@@ -87,16 +87,6 @@ ListDir ParseDir(facade::CmdArgParser* parser) {
   return parser->MapNext("LEFT", ListDir::LEFT, "RIGHT", ListDir::RIGHT);
 }
 
-string_view DirToSv(ListDir dir) {
-  switch (dir) {
-    case ListDir::LEFT:
-      return "LEFT"sv;
-    case ListDir::RIGHT:
-      return "RIGHT"sv;
-  }
-  return ""sv;
-}
-
 class BPopPusher {
  public:
   BPopPusher(string_view pop_key, string_view push_key, ListDir popdir, ListDir pushdir);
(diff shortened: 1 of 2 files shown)


kostasrim:
> can create an undeterministic result on the replica, so they should be transformed into

Do we have a test to verify this hypothesis ? (Haven't checked the code -- will follow up on that)

kostasrim:
I am having trouble understanding why `LMPOP` is non deterministic so maybe you could help explain why. Based on my understanding the command implementation does the following:

1. for each shard, find the first non empty key
2. for each shard, find the first non empty key with the `lowest index` in the `command` which means it will always pick the first non empty key from the actual argument list. The `.index()` is the `real` index and not relative to each shard keys

IMO a simple test with dragonfly running with different threads + carefully nitpicking the name of the keys (such that you can scramble it) will test this hypthesis and even if I am wrong we know we won't break the behaviour again

BorysTheDev:
> I am having trouble understanding why `LMPOP` is non deterministic so maybe you could help explain why. Based on my understanding the command implementation does the following:
> 
> 1. for each shard, find the first non empty key
> 2. for each shard, find the first non empty key with the `lowest index` in the `command` which means it will always pick the first non empty key from the actual argument list. The `.index()` is the `real` index and not relative to each shard keys
> 
> IMO a simple test with dragonfly running with different threads + carefully nitpicking the name of the keys (such that you can scramble it) will test this hypthesis and even if I am wrong we know we won't break the behaviour again

During no-point-in-time replication, we send data independently on the journal, so we can have another set of keys on the replica, so the results will be different. Also sending LPOP/RPOP instead of LMPOP is better for the replica CPU

kostasrim:
> Also sending LPOP/RPOP instead of LMPOP is better for the replica CPU

100% you turn a multi key operation to a single key and this is great! But this is not the main issue here. The reason we rewrite the command is not because we want to optimize it but because we want to fix dataloss cause by `no point in time replication`. However, in your PR there is not a single explanation of how we end up in this situation so after thinking about it I write here what I wrote internally:

```
1. both l1 and l2 exist on master. replica receives replicaof command.
1. we start iterating the buckets, we serialize l1 but we have not yet reached l2.
2. we receive the write command lmpop l2 l1 and we sent this command to the journal. replica receives it, pops l1 because l2 does not exist.
3. l2 gets serialized by iterate buckets and is sent to the replica
```

`l2` is the same for master/replica but `l1` is corrupted.

Therefore I would suggest to:

1. edit the description of this PR and explain clearly where and why the data loss happens
2. write a test that reproduces this such that we know that we don't break it in the future

romange:
@kostasrim  
I think it's super hard to write a test that reproduces such inconsistency. At least I do not know how.

we do journal rewriting for multiple reasons.  one of them is unrelated to no point in time - it's to reduce reliance on incremental operations - aka rewrite `incrby x` as `set x 42` - if it's not costly. So for sets we do not rewrite `SADD` by sending the entire set but we do omit an update if SADD did not add a new member. So we try to provide as much determinism as possible and if LMPOP ends up popping from key k1 then we should send `lpop k1` to make it more deterministic.

kostasrim (approved):

PR #5703, opened 2025-08-21
feat(search_family): Add LOAD_FROM to the FT.AGGREGATE command

BagritsevichStepan:
Join index_join with the FT.AGGREGATE command.

First PR is #5604

Most of the lines are from prev PR and also I added a lot of tests. Real number of lines is not so big

### DIFF
--- a/src/server/search/aggregator.h
+++ b/src/server/search/aggregator.h
@@ -27,7 +27,7 @@ using Value = ::dfly::search::SortableValue;
 
 // DocValues sent through the pipeline
 // TODO: Replace DocValues with compact linear search map instead of hash map
-using DocValues = absl::flat_hash_map<std::string_view, Value>;
+using DocValues = absl::flat_hash_map<std::string, Value>;
 
 struct AggregationResult {
   // Values to be passed to the next step
--- a/src/server/search/doc_index.cc
+++ b/src/server/search/doc_index.cc
@@ -94,6 +94,24 @@ std::pair<std::vector<FieldReference>, SortIndiciesFieldsList> PreprocessAggrega
   return {std::move(fields), {sort_indicies_aliases.begin(), sort_indicies_aliases.end()}};
 }
 
+/* Separate fields into basic and sortable. The second vector contains flags indicating
+   whether the field at the same index in the first vector is sortable or not. */
+std::pair<std::vector<FieldReference>, std::vector<bool>> GetBasicFields(
+    absl::Span<const std::string_view> fields, const search::Schema& schema) {
+  const size_t fields_count = fields.size();
+  std::vector<bool> is_sortable_field(fields_count);
+  std::vector<FieldReference> basic_fields;
+  basic_fields.reserve(fields_count);
+  for (size_t i = 0; i < fields_count; ++i) {
+    bool is_sortable = IsSortableField(fields[i], schema);
+    is_sortable_field[i] = is_sortable;
+    if (!is_sortable) {
+      basic_fields.emplace_back(fields[i]);
+    }
+  }
+  return {std::move(basic_fields), std::move(is_sortable_field)};
+}
+
 }  // namespace
 
 bool FieldReference::IsJsonPath(std::string_view name) {
@@ -497,19 +515,8 @@ join::Vector<join::OwnedEntry> ShardDocIndex::PreagregateDataForJoin(
     search::SearchAlgorithm* search_algo) const {
   auto search_results = search_algo->Search(&*indices_);
 
-  // First filter out sortable and non-sortable fields
-  // We will load them in different ways
   const size_t fields_count = join_fields.size();
-  std::vector<bool> is_sortable_field(fields_count);
-  std::vector<FieldReference> basic_fields;
-  basic_fields.reserve(fields_count);
-  for (size_t i = 0; i < fields_count; ++i) {
-    bool is_sortable = IsSortableField(join_fields[i], base_->schema);
-    is_sortable_field[i] = is_sortable;
-    if (!is_sortable) {
-      basic_fields.emplace_back(join_fields[i]);
-    }
-  }
+  const auto [basic_fields, is_sortable_field] = GetBasicFields(join_fields, base_->schema);
 
   join::Vector<join::OwnedEntry> result;
   result.reserve(search_results.ids.size());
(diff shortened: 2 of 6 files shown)


romange:
it's huge. Is it possible to split it somehow?

BagritsevichStepan:
> it's huge. Is it possible to split it somehow?

I left a comment in the PR description regarding the number of code lines. This is the second PR; once the first one is merged, the number of lines will be much lower.

src/server/search/aggregator.h
```diff
@@ -27,7 +27,7 @@ using Value = ::dfly::search::SortableValue;
 
 // DocValues sent through the pipeline
 // TODO: Replace DocValues with compact linear search map instead of hash map
-using DocValues = absl::flat_hash_map<std::string_view, Value>;
+using DocValues = absl::flat_hash_map<std::string, Value>;
```
BagritsevichStepan, line 30: It is temporary and will be fixed in next PRs

src/server/search/search_family.cc
```diff
@@ -442,6 +442,62 @@ ParseResult<aggregate::SortParams> ParseAggregatorSortParams(CmdArgParser* parse
   return sort_params;
 }
 
+std::pair<std::string_view, std::string_view> Split(std::string_view s, char delim) {
+  auto pos = s.find(delim);
+  if (pos == std::string_view::npos)
+    return {s, std::string_view{}};
+  return {s.substr(0, pos), s.substr(pos + 1)};
+}
```
dranikpg: Likely its just `absl::StrSplit(MaxSplits(1))` or something like that 🙂

src/server/search/search_family.cc
```diff
@@ -442,6 +442,62 @@ ParseResult<aggregate::SortParams> ParseAggregatorSortParams(CmdArgParser* parse
...
+
+  // Validate index name
+  known_indexes->insert(join_params.index_alias);
+
+  size_t num_fields = parser->Next<size_t>();
+  join_params.conditions.reserve(num_fields);
+  while (parser->HasNext() && num_fields > 0) {
+    auto [left, right] = Split(parser->Next(), '=');
+    auto [l_index, l_field] = Split(left, '.');
+    auto [r_index, r_field] = Split(right, '.');
```
dranikpg: Nit: it would've been helpful if you added a single line comment on how the format looks like for example, so next editors don't have to deduce it from reading the code. I did this sometimes in early search argument parsing

src/server/search/search_family.cc
```diff
@@ -532,12 +592,212 @@ ParseResult<AggregateParams> ParseAggregatorParams(CmdArgParser* parser) {
...
+  PreprocessedJoinData result(n + 1);
+
+  // Collect aliases and initialize result.indexes
+  result.alias_to_index.reserve(n);
+  result.alias_to_index[index] = 0;
+  result.indexes[0] = index;
+  for (size_t i = 0; i < n; ++i) {
+    result.alias_to_index[params.joins[i].index_alias] = i + 1;
+    result.indexes[i + 1] = params.joins[i].index;
+  }
```
dranikpg, line 650: What if we have two loads from the same index. It looks like it'll break 👀

src/server/search/search_family.cc
```diff
@@ -532,12 +592,212 @@ ParseResult<AggregateParams> ParseAggregatorParams(CmdArgParser* parser) {
...
+
+        // We need to convert std::string to std::string_view
+        auto insert_copy = [&field_values](const auto& field_value) {
+          using T = std::decay_t<decltype(field_value)>;
+          if constexpr (std::is_same_v<T, std::string>) {
+            field_values.emplace_back(std::string_view{field_value});
+          } else {
+            field_values.emplace_back(field_value);
+          }
+        };
```
dranikpg: string -> string_view is implicit, so your variant should be constructible just from the value

src/server/search/doc_index.cc
```diff
@@ -492,6 +492,115 @@ vector<SearchDocData> ShardDocIndex::SearchForAggregator(
...
+                                                std::make_move_iterator(join_fields_values.end())));
+    }
+  }
+
+  return result;
+}
+
+ShardDocIndex::FieldsValuesPerDocId ShardDocIndex::LoadKeysData(
+    const OpArgs& op_args, const absl::flat_hash_set<search::DocId>& doc_ids,
+    absl::Span<const std::string_view> fields_to_load) const {
```
dranikpg, line 569: It's almost equal to the upper function except for stripping monostate from the variant

dranikpg: And getting the ids passed and not form search

src/server/search/search_family.cc
```diff
@@ -1059,34 +1319,121 @@ void SearchFamily::FtAggregate(CmdArgList args, const CommandContext& cmd_cntx)
...
+    if (!search_algos[0].Init(params->query, &params->params)) {
+      return builder->SendError("Query syntax error");
+    }
+
+    for (size_t i = 0; i < params->joins.size(); ++i) {
+      search::QueryParams empty_params;
+      if (!search_algos[i + 1].Init(params->joins[i].query, &empty_params)) {
+        return builder->SendError("Query syntax error in JOIN");
+      }
+    }
```
dranikpg, line 1374: So you decided to support sub queries from the start 👍🏻

dranikpg (dismissed):
Please make sure it doesn't misbehave with two load froms with the same index. Approach is ok. Left some small code comments

src/server/search/search_family.cc
BagritsevichStepan: The motivation is actually a bit different. Note that once we implement the `FILTER` option, it will be executed as a basic step of aggregation (on the coordinator thread). This means it will only take place after the join. With queries we will do filtering before joining, and it is much better since that reduces the number of documents and speeds up the join process.

Therefore, if the user wants to filter entries by comparing a field with a constant, it’s preferable to use the query itself. For comparisons like @some_field < @another_field, however, we still need the FILTER option, which will be applied at the final stage (after joining)

BagritsevichStepan:
> Please make sure it doesn't misbehave with two load froms with the same index. Approach is ok. Left some small code comments

Fixed and added this case to the tests

dranikpg (approved):

PR #5704, opened 2025-08-21
chore: check fail for empty listpacks in streams

kostasrim:
Add check fails when we trim or delete list packs in stream code. Push check in low level functions of lp/stream code

### DIFF
--- a/src/redis/t_stream.c
+++ b/src/redis/t_stream.c
@@ -278,6 +278,13 @@ void streamGetEdgeID(stream *s, int first, int skip_tombstones, streamID *edge_i
     streamIteratorStop(&si);
 }
 
+void checkListPackNotEmpty(unsigned char* lp) {
+  if(lpBytes(lp) == 0) {
+   fprintf(stderr, "Error: corrupted listpack found.");
+   abort();
+  }
+}
+
 /* Trim the stream 's' according to args->trim_strategy, and return the
  * number of elements removed from the stream. The 'approx' option, if non-zero,
  * specifies that the trimming must be performed in a approximated way in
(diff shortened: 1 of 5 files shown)


src/redis/t_stream.c
```diff
@@ -345,7 +345,12 @@ int64_t streamTrim(stream *s, streamAddTrimArgs *args) {
 
         if (remove_node) {
             lpFree(lp);
-            raxRemove(s->rax_tree,ri.key,ri.key_len,NULL);
+            int res = raxRemove(s->rax_tree,ri.key,ri.key_len,NULL);
+            if(res == 0) {
```
kostasrim: This should not happen but I added this for sanity.

src/redis/t_stream.c
```diff
@@ -739,7 +750,13 @@ void streamIteratorRemoveEntry(streamIterator *si, streamID *current) {
         /* If this is the last element in the listpack, we can remove the whole
          * node. */
         lpFree(lp);
-        raxRemove(si->stream->rax_tree,si->ri.key,si->ri.key_len,NULL);
+        int res = raxRemove(si->stream->rax_tree,si->ri.key,si->ri.key_len,NULL);
```
kostasrim: This should not happen either, but lets assert this

kostasrim:
IMO I went over the implementation of these functions + the history of their respective files and I could not find any meaningful differences. I added the extra CHECKS for sanity.

src/redis/t_stream.c
romange: why it should not happen? how do you know?

kostasrim: I went over `raxRemove` implementation + redis does the exact same thing here (so at least they were not aware of that  if there is a bug there).

But I don't trust me and I trust that at least I can add checks to test for invariants

romange (approved):

Issue #5648, opened 2025-08-11
Improve destination shard Read latency during slot migrations


romange:
#### Issue
1. Currently the destination nodes reaches high cpu usage during migrations which is not an issue _per se_, as long as Dragonfly keeps responsive
2. The fact that we use inlined runs in transactions (see `Transaction::CanRunInlined`) and migrate threads, we can be stuck in a cpu-only cycle for a large chunk of time
3. If the destination nodes receives Read commands, it's average latency can be more or less tolerable, but the p99 latency will be likely terrible

#### Suggestion

We have background fibers as of now, but we'll introduce them later.  We should first get a good baseline with a hand tuned soution. It's ok if its suboptimal for migrations without any load. We did this already for:
* source node during migrations (journal_streamer.cc)
* source node during replication (snapshot.cc)

We primarily use `ThisFiber::Yield()` to hand over control to other fibers and `ThisFiber::Sleep` to stall execution and let other fibers run undisturbed for a prolonged period of time

#### How to measure it

When measuring latency we should simulate background load as if it is sent from a real application

```sh
./dfly_bench
  -c 2 # the less connections we use, the larger is the migrations cpu share with normal fibers
  --qps=1000 # Not too much and not too little
  --ratio 0:1 # all reads for now
  --test_time=10
```

Which should give something ~50-200 useconds on a perfectly idle instance for pavg/p50/p99. 
Next we want to repeat this benchmark while sending traffic to the destination node during migration. Our solution should reach p99 latency that is maybe a few times higher, but not skyrocketing into the milliseconds. As a result, the migration time might also increase (hopefully proportional to the slowdown, it should also not take ages or starve)

kostasrim:
> The fact that we use inlined runs in transactions (see Transaction::CanRunInlined) and migrate threads, we can be stuck in a cpu-only cycle for a large chunk of time

Yes! And I also noticed that we have the exact same loop in `StableSyncDflyReadFb`

PR #5715, opened 2025-08-22
feat: improve target node latency during migration

BorysTheDev:
fixes: #5648

I've tested different options, and the current implementation allows me to see p99 < 400us on my PC during the whole migration process. migration speed decreased by 5-10%

### DIFF
--- a/src/server/cluster/incoming_slot_migration.cc
+++ b/src/server/cluster/incoming_slot_migration.cc
@@ -19,9 +19,10 @@
 #include "util/fibers/synchronization.h"
 
 ABSL_DECLARE_FLAG(int, migration_finalization_timeout_ms);
-ABSL_FLAG(uint32_t, slot_migration_throttle_us, 20,
+ABSL_FLAG(uint32_t, slot_migration_throttle_us, 0,
           "Incoming migration throttle time in us, we throttle every 100us of migration commands "
-          "processing, 0 to disable");
+          "processing, 0 to disable. Recommended value is 20. Values more than 50 can "
+          "significantly reduce migration speed.");
 
 namespace dfly::cluster {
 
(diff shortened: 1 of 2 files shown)


romange:
what if it does not work in prod like on your PC? I suggest introducing flags at least.

src/server/cluster/incoming_slot_migration.cc
```diff
@@ -112,6 +113,12 @@ class ClusterShardMigration {
           break;
         }
       }
+      uint64_t now = fb2::ProactorBase::GetMonotonicTimeNs();
+      if (now - last_sleep >
+          100000) {  // every 100us we do sleep for 20us to allow other coomands to be processed
```
romange: commands

src/server/cluster/incoming_slot_migration.cc
```diff
@@ -19,6 +19,8 @@
 #include "util/fibers/synchronization.h"
 
 ABSL_DECLARE_FLAG(int, migration_finalization_timeout_ms);
+ABSL_FLAG(bool, slot_migration_throttle, true,
```
romange: I meant to add flags to adjust sleep value

romange: like we did on the source shard

BorysTheDev: If you decide to use this flag, please be careful with values more than 50us, on my laptop, after 50us, I saw a huge migration speed degradation

src/server/cluster/incoming_slot_migration.cc
```diff
@@ -19,6 +19,9 @@
 #include "util/fibers/synchronization.h"
 
 ABSL_DECLARE_FLAG(int, migration_finalization_timeout_ms);
+ABSL_FLAG(uint32_t, slot_migration_throttle_us, 20,
```
romange: ok, let's set it as 0 to not change the existing behaviour.

romange (approved):

PR #5717, opened 2025-08-24
feat(search_family): Improve sorting and limiting in index joins

BagritsevichStepan:
Improve sorting and limiting in index joins.

Right now we apply SORTBY and LIMIT only after both join stages have finished—i.e., we (1) perform the join, and then (2) load all requested fields for every joined entry. Because there can be a lot of joined entries, we end up loading many fields we’ll later discard when sorting/limiting.

The fix is to apply SORTBY and LIMIT immediately after the first stage—right after we build the join result, but before we load per-entry fields. That way we only fetch fields for the top offset + total candidates (or whatever budget we choose), which speeds up the request — especially when most entries would be dropped by the final sort/limit anyway.

### DIFF
--- a/src/server/search/search_family.cc
+++ b/src/server/search/search_family.cc
@@ -523,6 +523,14 @@ ParseResult<AggregateParams> ParseAggregatorParams(CmdArgParser* parser) {
   // Used for join params
   absl::flat_hash_set<std::string> current_known_indexes;
   current_known_indexes.insert(std::string{params.index});
+  while (parser->HasNext() && parser->Check("LOAD_FROM")) {
+    auto join_params = ParseAggregatorJoinParams(parser, &current_known_indexes);
+    if (!join_params) {
+      return make_unexpected(join_params.error());
+    }
+    params.joins.emplace_back(std::move(join_params).value());
+  }
+  const bool joining_enabled = !params.joins.empty();
 
   while (parser->HasNext()) {
     // GROUPBY nargs property [property ...]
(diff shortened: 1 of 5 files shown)


src/server/search/search_family.cc
```diff
@@ -532,12 +609,294 @@ ParseResult<AggregateParams> ParseAggregatorParams(CmdArgParser* parser) {
...
+    }
+
+    size_t new_limit = std::min(limit, joined_entries->size());
+    if (offset) {
+      for (size_t i = offset; i < new_limit; ++i) {
+        auto& dest = (*joined_entries)[i - offset];
+        auto& src = (*joined_entries)[i];
+        DCHECK(dest.size() == src.size());
+        dest = std::move(src);
+      }
```
dranikpg, line 844: isn't it a span, you can just remove the prefix

src/server/search/search_family.cc
```diff
@@ -532,12 +609,294 @@ ParseResult<AggregateParams> ParseAggregatorParams(CmdArgParser* parser) {
...
+    size_t limit = offset + total;
+    if (!sort_params.empty()) {
+      if (limit >= joined_entries->size()) {
+        std::sort(joined_entries->begin(), joined_entries->end(), std::move(comparator));
+      } else {
+        std::partial_sort(joined_entries->begin(), joined_entries->begin() + limit,
+                          joined_entries->end(), std::move(comparator));
+        joined_entries->resize(limit);
+      }
+    }
```
dranikpg, line 835: Ahh, that's really a problem with distributed indices. You can't serialize the whole index, but to get all matches you actually have to.

dranikpg (dismissed):
Hard to review if it's not rebased, but I found the relevant parts

src/server/search/search_family.cc
BagritsevichStepan: We’ve added QUERY to LOAD_FROM, so most documents are filtered up before joining. Also (just to information), in most cases users sort by the same key used for the join (e.g., idx1.key == idx2.field; SORTBY @idx1.key), so this doesn’t materially increase the number of fields we need to load.

But yes, it is disadvantage of distributed indices.

BagritsevichStepan:
> Hard to review if it's not rebased, but I found the relevant parts

rebased

src/server/search/search_family.cc
BagritsevichStepan: It is not a span. It is std::vector<std::vector<size_t>>*

dranikpg (approved):

PR #5719, opened 2025-08-25
server: fix max emission limit in throttle command (cherry-pick to v1.33)

abhijat:
* server: Use nanoseconds to measure duration in throttle algorithm

Previously we used milliseconds to measure time units. However this causes issues when a sub-millisecond refill interval is required for the algorithm. To fix this we move here to nanosecond time units.

There are some issues which arise because of this, all of which are acceptable:

* We have to use the current time at the point of throttle function call as a representation of "now", instead of the start of transaction time.

* The retry after/reset after timestamps are still in ms during throttle calculation, so they lose some precision.

The former is acceptable, as although it will cause slightly more lenient throttling, the measure is at a sub-millisecond level. Similarly in case of the latter, it is hard to imagine a client which will suffer from a sub-millisecond precision loss of a coarse value like retry interval.

Cherry-pick 090728b3e6b577f0771da1c6fee5379397f43ea4 to v.133 for release

### DIFF
--- a/src/server/engine_shard_set.h
+++ b/src/server/engine_shard_set.h
@@ -170,6 +170,10 @@ inline uint64_t GetCurrentTimeMs() {
   return TEST_current_time_ms ? TEST_current_time_ms : absl::GetCurrentTimeNanos() / 1000000;
 }
 
+inline uint64_t GetCurrentTimeNs() {
+  return TEST_current_time_ms ? TEST_current_time_ms * 1000000 : absl::GetCurrentTimeNanos();
+}
+
 extern EngineShardSet* shard_set;
 
 }  // namespace dfly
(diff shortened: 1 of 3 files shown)


abhijat:
error is caused due to failure when downloading commit from hdr histogram:

```
Cloning into 'hdr_histogram'...
fatal: unable to read tree (652d51bcc36744fd1a6debfeb1a8a5f58b14022c)
CMake Error at tmp/hdr_histogram_project-gitclone.cmake:61 (message):
  Failed to checkout tag: '652d51bcc36744fd1a6debfeb1a8a5f58b14022c'
```

Also seen locally

romange:
> error is caused due to failure when downloading commit from hdr histogram:
> 
> ```
> Cloning into 'hdr_histogram'...
> fatal: unable to read tree (652d51bcc36744fd1a6debfeb1a8a5f58b14022c)
> CMake Error at tmp/hdr_histogram_project-gitclone.cmake:61 (message):
>   Failed to checkout tag: '652d51bcc36744fd1a6debfeb1a8a5f58b14022c'
> ```
> 
> Also seen locally

@vyavdoshenko just pushed the fix

PR #5721, opened 2025-08-25
test: skip failing tests

vyavdoshenko:
Temporary skip tests because of constant failures on CI.

### DIFF
--- a/tests/dragonfly/server_family_test.py
+++ b/tests/dragonfly/server_family_test.py
@@ -74,6 +74,7 @@ async def test_get_databases(async_client: aioredis.Redis):
     assert dbnum == {"databases": "16"}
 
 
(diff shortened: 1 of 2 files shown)


BorysTheDev (dismissed):

tests/dragonfly/server_family_test.py
```diff
@@ -74,6 +74,7 @@ async def test_get_databases(async_client: aioredis.Redis):
     assert dbnum == {"databases": "16"}
 
 
+@pytest.mark.skip("Fails for epoll configuration")
```
BorysTheDev: skip only epoll version

vyavdoshenko: fixed

PR #5727, opened 2025-08-27
fix: data loss in partial sync from same source master

### LINKED ISSUE #5730: data loss in partial sync from same source master
Start with 3 df nodes:

* c1(master)
* c2(replica 1)
* c3(replica 2)

`c2` initiates takeover on `c1` and becomes the new master. 

If `c3` does not become a replica of `c2` after the takeover in one step (atomic) then there is data loss. 

After the takeover, `c2` and `c3` are in sync (because we WaitReplicaFlowToCatchup). When `c2` receives a write command and `c3` is not yet assigned to be its replica there is no open journal for `c2` and hence no lsn ring buffers to write these changes to. When `c3` becomes a replica of `c2` it's too late, it will try to partial sync and because of https://github.com/dragonflydb/dragonfly/pull/5727/files#diff-ab928257a536aac41ef82648984d1104e8a17fbc92085b56d6ce569047b4956eL336 its value is 1 and we missed those write commands.

What if `takeover + replicaof` was atomic ?

The lsn diff is always zero and any code around it is redundant (no incremental snapshot is needed or anything, nodes are fully in sync before and after the takeover).

So to summarize:

1. If takeover + replicaof is treated (from an operational perspective) atomic then `allow_partial_sync_with_lsn_diff` flag and the code around it is obsolete.
2. If we allow write commands between then there is data loss. Ontop of that if we use the replicas LSN's like in the linked PR, we check fail because "we try to sync from the future". Even if we switch on the journal after the takeover we would still need to normalize the received lsn's sent by c3 because c2's journal starts at 1.

Which of the two scenarios is the behavior we want ?

p.s The same (and more) apply to c2.(replicaof no one) followed by c3.(replicaof c2) and I mentioning it because we have a test that we seed after `replicaof no one` and before c3 becomes a replica of c2 which hints that this is an actual use case.

kostasrim:
I discovered this while working on https://github.com/dragonflydb/dragonfly/pull/5580

Our algorithm has a few bugs:

1. It always uses the new master's LSN to start partial sync. This is the identity operation and no data is actually sent during partial sync.
2. The journal is not open after a takeover so any new commands is forever lost
2. Sending the LSN vector for partial sync is also a mistake because the new master has its LSN starting at 1

The fix:

Start the journal on the replica at the beginning of the takeover with the lsn set to current (and not 1). Also, after the TakeOver, allow the rest of the replicas to catch up from the old master and warn if that fails.

resolves  #5730

### DIFF
--- a/src/server/dflycmd.cc
+++ b/src/server/dflycmd.cc
@@ -302,6 +302,7 @@ void DflyCmd::Flow(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext* cn
     sf_->journal()->StartInThread();
 
     std::optional<Replica::LastMasterSyncData> data = sf_->GetLastMasterData();
+    std::optional<LSN> lsn_to_start_partial;
     // In this flow the master and the registered replica where synced from the same master.
     if (last_master_id && data && data->id == *last_master_id) {
       ++ServerState::tlocal()->stats.psync_requests_total;
@@ -309,9 +310,20 @@ void DflyCmd::Flow(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext* cn
       if (!flow_lsn) {
         return;  // ParseLsnVec replies in case of error
       }
-      MaybePartialSync(*flow_lsn, &sync_type, &flow);
-    } else if (seqid.has_value()) {
-      MaybePartialSync(*seqid, &sync_type, &flow);
+
+      if (IsLSNInPartialSyncBuffer(*flow_lsn)) {
+        lsn_to_start_partial.emplace(*flow_lsn);
+      }
+
+    } else if (seqid.has_value() && IsLSNInPartialSyncBuffer(*seqid)) {
+      lsn_to_start_partial.emplace(*seqid);
+    }
+
+    if (lsn_to_start_partial) {
+      flow.start_partial_sync_at = *lsn_to_start_partial;
+      sync_type = "PARTIAL";
+      VLOG(1) << "Partial sync requested from LSN=" << flow.start_partial_sync_at.value()
+              << " and is available. (current_lsn=" << sf_->journal()->GetLsn() << ")";
     }
   }
 
@@ -397,21 +409,18 @@ void DflyCmd::StartStable(CmdArgList args, Transaction* tx, RedisReplyBuilder* r
   return rb->SendOk();
 }
 
-void DflyCmd::MaybePartialSync(LSN lsn, std::string* sync_type, FlowInfo* flow) {
+bool DflyCmd::IsLSNInPartialSyncBuffer(LSN lsn) const {
   auto* jrnl = sf_->journal();
 
-  if (jrnl->GetLsn() == lsn || jrnl->IsLSNInBuffer(lsn)) {
-    flow->start_partial_sync_at = lsn;
-    *sync_type = "PARTIAL";
-    VLOG(1) << "Partial sync requested from LSN=" << flow->start_partial_sync_at.value()
-            << " and is available. (current_lsn=" << sf_->journal()->GetLsn() << ")";
-    return;
+  const bool exists = jrnl->GetLsn() == lsn || jrnl->IsLSNInBuffer(lsn);
+  if (!exists) {
+    LOG(INFO) << "Partial sync requested from stale LSN=" << lsn
+              << " that the replication buffer doesn't contain this anymore (current_lsn="
+              << sf_->journal()->GetLsn() << "). Will perform a full sync of the data.";
+    LOG(INFO) << "If this happens often you can control the replication buffer's size with the "
+                 "--shard_repl_backlog_len option";
   }
-  LOG(INFO) << "Partial sync requested from stale LSN=" << lsn
-            << " that the replication buffer doesn't contain this anymore (current_lsn="
-            << sf_->journal()->GetLsn() << "). Will perform a full sync of the data.";
-  LOG(INFO) << "If this happens often you can control the replication buffer's size with the "
-               "--shard_repl_backlog_len option";
+  return exists;
 }
 
 std::optional<LSN> DflyCmd::ParseLsnVec(std::string_view last_master_lsn,
@@ -421,7 +430,7 @@ std::optional<LSN> DflyCmd::ParseLsnVec(std::string_view last_master_lsn,
   if (lsn_str_vec.size() != last_journal_lsn_size) {
     rb->SendError(facade::kSyntaxErr);  // Unexpected flow. LSN vector of same master
                                         // should be the same size on all replicas.
-    return {};
+    return std::nullopt;
   }
 
   std::vector<LSN> lsn_vec;
@@ -431,7 +440,7 @@ std::optional<LSN> DflyCmd::ParseLsnVec(std::string_view last_master_lsn,
     int64_t value;
     if (!absl::SimpleAtoi(lsn_str, &value)) {
       rb->SendError(facade::kInvalidIntErr);
-      return {};
+      return std::nullopt;
     }
     lsn_vec.push_back(value);
   }
--- a/src/server/dflycmd.h
+++ b/src/server/dflycmd.h
@@ -223,9 +223,9 @@ class DflyCmd {
   std::optional<LSN> ParseLsnVec(std::string_view lsn_vec, size_t flow_id,
                                  size_t last_journal_lsn_size, facade::RedisReplyBuilder* rb);
 
-  // If lsn exists in buffer update sync_type to FULL_SYNC.
-  // Also if we can do partial, update flow.pstart_partial_sync_at
-  void MaybePartialSync(LSN lsn, std::string* sync_type, FlowInfo* flow);
+  // Checks if LSN exists in the partial sync buffer. If not, also LOG that we can't
+  // partial sync.
+  bool IsLSNInPartialSyncBuffer(LSN lsn) const;
 
   // Return a map between replication ID to lag. lag is defined as the maximum of difference
   // between the master's LSN and the last acknowledged LSN in over all shards.


src/server/dflycmd.cc
```diff
@@ -333,7 +333,7 @@ void DflyCmd::Flow(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext* cn
 
       if (IsLSNDiffBellowThreshold(data.value().last_journal_LSNs, lsn_vec)) {
         sync_type = "PARTIAL";
-        flow.start_partial_sync_at = sf_->journal()->GetLsn();
+        flow.start_partial_sync_at = lsn_vec[flow_id];
```
kostasrim: `sf_->journal()->GetLsn()` is the latest LSN on the master and not the last `LSN` of the replica. This is the bug and we end up sending `0 entries` and bump up the journal offset :facepalm: 

Instead master needs to sent `[lsn_vec[flow,id], GetLsn()]` to be fully in sync

tests/dragonfly/replication_test.py
```diff
@@ -3173,13 +3173,21 @@ async def test_partial_replication_on_same_source_master_with_replica_lsn_inc(df
     await c_s2.execute_command(f"REPLICAOF NO ONE")
     # Make server 4 replica of server 2
     await c_s4.execute_command(f"REPLICAOF localhost {server2.port}")
+    await check_all_replicas_finished([c_s4], c_s2)
```
kostasrim, line 3200: I moved it here because the for statement bellow will seed a few values that won't reach the journal (because it's kinda racy if the journal gets registered by full sync before the first element seeded reaches dragonfly). To avoid this and force all the seeded values below to go through the journal, I moved the statement above. This does not affect the test

tests/dragonfly/replication_test.py
```diff
@@ -3173,13 +3173,21 @@ async def test_partial_replication_on_same_source_master_with_replica_lsn_inc(df
...
     # Send some write command for lsn inc
     for i in range(100):
         await c_s2.set(i, "val")
     # Make server 3 replica of server 2
     await c_s3.execute_command(f"REPLICAOF localhost {server2.port}")
     await check_all_replicas_finished([c_s3], c_s2)
-    await check_all_replicas_finished([c_s4], c_s2)
+
+    s2_sz = await c_s2.dbsize()
```
kostasrim, line 3208: Maybe seed+capture instead ?

kostasrim:
Ok I wrote https://github.com/dragonflydb/dragonfly/issues/5730

src/server/dflycmd.cc
```diff
@@ -511,13 +503,19 @@ void DflyCmd::TakeOver(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext
...
   if (*status == OpStatus::OK) {
-    dfly::SharedLock lk{replica_ptr->shared_mu};
-    auto cb = [replica_ptr, end_time, &catchup_success](EngineShard* shard) {
-      if (!WaitReplicaFlowToCatchup(end_time, replica_ptr.get(), shard)) {
-        catchup_success.store(false);
-      }
-    };
-    shard_set->RunBlockingInParallel(std::move(cb));
+    util::fb2::LockGuard mu_lk(mu_);
+    for (auto [id, repl_ptr] : replica_infos_) {
```
kostasrim: We need this. Otherwise we can't guarantee that there is not data loss. Why ? Because during takeover and before the replica becomes the new master it must get all the data first. At that time there are no journal writes and hence if the other replica tries to replicate the new master it will never partial sync. 

Here we enforce on takeover that all the "actors" are in sync when they split (one node takesover the other replica disconnects). Any command that comes after the takeover and before the other replica reconcile with its new master, it will exist in the ring buffer and properly replayed during partial sync. On the other hand, if the lsn buffers progressed too much (do not contain LSN 1) then we will fall back to FULL SYNC.

tests/dragonfly/replication_test.py
```diff
@@ -3114,11 +3114,21 @@ async def test_partial_replication_on_same_source_master(df_factory, use_takeove
     if use_takeover:
         # Promote first replica to master
         await c_replica1.execute_command(f"REPLTAKEOVER 5")
+        if backlog_len > 1:
+            await c_replica1.execute_command("SET bar foo")
```
kostasrim, line 3135: We had data loss here before -- I added this test case. 

IMO I am planning to split this test and add a few more test cases (similar to partial_sync test)

tests/dragonfly/replication_test.py
```diff
@@ -3114,11 +3114,21 @@ async def test_partial_replication_on_same_source_master(df_factory, use_takeove
...
+            await c_replica1.execute_command("SET foo bar")
+
     else:
         # Promote first replica to master
         await c_replica1.execute_command(f"REPLICAOF NO ONE")
-        # Send 2 more commands to be propagated to second replica
-        # Sending 2 more commands will result in partial sync if allow_partial_sync_with_lsn_diff is equal or higher
+        # Not sure what was the point of this test/flow. To assert that we have a bug ?
```
kostasrim: `if use_takeover or (backlog_len > 0 and not use_takeover):` 

I still don't understand this, r2 contains changes that r1 does not have and we manage to make r2 a replica of r1 and partially sync ?! This flow is broken. If we can't guarantee consistency why would we allow partial sync ? For that flow data(r2) != data(r1)

tests/dragonfly/replication_test.py
```diff
@@ -3114,11 +3114,21 @@ async def test_partial_replication_on_same_source_master(df_factory, use_takeove
...
+        # replica1 stops replicating from master. Then we seed 2 commands to the
+        # old master and wait for replica2 to be in sync. Finally, we make replica2
+        # a replica of replica1 and *expect* them to partially sync??!! Yes it will
+        # but the nodes do not contain the same data. In particular replica2 has 2 more
+        # elements than its new master(replica1). We should not allow partial sync from
+        # same master if we didn't takeover as we can't protect ourselves from the case above.
+        # TODO(kostas)
```
romange: I do not think it's worth a todo. `allow_partial_sync_with_lsn_diff` was temp hack and does not reflect how servers should behave in prod. Please align the test to the expected behavior. Also you can ask @adiholden if needed.

my opinion: `allow_partial_sync_with_lsn_diff` allowed this faulty behavior because it was a hack.

src/server/dflycmd.cc
romange: Please rephrase a bit more clearly and put it as a comment into code.

src/server/server_family.cc
```diff
@@ -3605,6 +3605,10 @@ void ServerFamily::ReplTakeOver(CmdArgList args, const CommandContext& cmd_cntx)
   if (IsClusterEnabled()) {
     service().cluster_family().ReconcileMasterReplicaTakeoverSlots(false);
   }
+
+  // Start journal to allow partial sync from same source master
```
romange: how it worked before?! We tested partial sync, so how it worked without journal starting?

tests/dragonfly/replication_test.py
kostasrim: Yes it was a hack which always led to data-loss. That's why I wrote here the normal operation should be `TAKEOVER + PARTIAL` and never `REPLICAOF NO ONE + PARTIAL`. Literally, the flag could be described as "how many elements do you wish your new replica to loose in this scenario?"

The answer for prod instances should be `0`.

src/server/server_family.cc
kostasrim: It did not. We had dataloss. See what I write in the comment: https://github.com/dragonflydb/dragonfly/pull/5727/files#diff-f41a28e79426b76916ad0fb750e102f83955a92be0f24dda7ae4fda79d74ca8cR3118 

and it worked because:

https://github.com/dragonflydb/dragonfly/pull/5727/files#diff-ab928257a536aac41ef82648984d1104e8a17fbc92085b56d6ce569047b4956eL336

Specifically:

```
flow.start_partial_sync_at = sf_->journal()->GetLsn();
```

Here, we get `1`, and the size of the partial sync buffer is `1` so we sent nothing 😮‍💨

tests/dragonfly/replication_test.py
kostasrim: I disabled partial sync for `NO ONE` flow all together and adjusted the test

tests/dragonfly/replication_test.py
```diff
@@ -3190,18 +3194,26 @@ async def test_partial_replication_on_same_source_master_with_replica_lsn_inc(df
     await c_s2.execute_command(f"REPLICAOF NO ONE")
     # Make server 4 replica of server 2
     await c_s4.execute_command(f"REPLICAOF localhost {server2.port}")
+    await check_all_replicas_finished([c_s4], c_s2)
```
kostasrim, line 3200: TODO for #5580 

Remove this test all together. It was introduced as part of https://github.com/dragonflydb/dragonfly/pull/5189 but it is no longer needed as we don't do partial sync anymore via the `incremental snapshot`

kostasrim: And we also don't trigger partial sync for this flow anymore

src/server/dflycmd.cc
```diff
@@ -513,13 +506,27 @@ void DflyCmd::TakeOver(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext
...
+    // in sync after the takeover, then any data that was not replicated from the old master
+    // is lost, because the new master does not contain their entries and a fallback to FULL
+    // SYNC is required (since the journal starts when the takeover completes, which implies
+    // data parity with the old master and an empty journal on the newly promoted master).
+    util::fb2::LockGuard mu_lk(mu_);
+    for (auto [id, repl_ptr] : replica_infos_) {
+      dfly::SharedLock lk{repl_ptr->shared_mu};
+      // Wait for ALL replicas to catch up. Otherwise we can't do partial sync from same source
+      // master. If the replica misses 1 LSN entry, that entry will never exist in the LSN buffer
+      // of the new master effectively killing partial sync.
```
romange:
```suggestion

      // Wait for ALL replicas to catch up. Otherwise we can't do partial sync from same source
      // master. If the other replica misses a single journal entry, that entry will never exist in the replication buffer
      // of the new master, so that replica won't be able to use partial sync to sync with the new master.
```

src/server/dflycmd.cc
```diff
@@ -513,13 +506,27 @@ void DflyCmd::TakeOver(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext
...
+      auto cb = [replica_ptr, end_time, &catchup_success](EngineShard* shard) {
+        if (!WaitReplicaFlowToCatchup(end_time, replica_ptr.get(), shard)) {
+          catchup_success.store(false);
+        }
+      };
+      shard_set->RunBlockingInParallel(std::move(cb));
```
romange: I understand the reason for enforcing this but now we make takeover less reliable as now it depends not on two, but 3+ moving parts. I suggest, that you try to enforce ALL replicas but fail the takeover process only if the takeover replica fails to catch up. this way you will have it both ways.

tests/dragonfly/replication_test.py
kostasrim: nvm I rewrote it with a takeover instead, and kept the test to check errornous flow regardless

src/server/dflycmd.cc
kostasrim: Yes +1. Upon that case, we LOG + disable partial sync (because now we can't gurantee that there is no data loss)

kostasrim: nice idea!

kostasrim: done. The semantics now are:

*. Takeover is succefull even if rest of the flows catch up.
*. If the flows do not catch up, then we can't do partial sync as we don't know what state the other nodes are in. If that's true we disable it
*. If all nodes catch up instead, all good again but we also have partial sync enabled.

Small change to the protocol to also sent `NO PARTIAL` in case all the flows did not catch up in time.

romange: I do not think that we should send "NO PARTIAL" - it's an implementation detail between master and replica. you can log into INFO if you want

src/server/server_family.cc
```diff
@@ -3607,18 +3608,26 @@ void ServerFamily::ReplTakeOver(CmdArgList args, const CommandContext& cmd_cntx)
...
     service().cluster_family().ReconcileMasterReplicaTakeoverSlots(false);
   }
-  SetMasterFlagOnAllThreads(true);
+
+  // Start journal to allow partial sync from same source master
+  service_.proactor_pool().AwaitFiberOnAll([this](auto, auto*) { journal()->StartInThread(); });
+
   last_master_data_ = replica_->Stop();
   replica_.reset();
+  if (take_over_res == Replica::NO_PARTIAL) {
```
romange: why is it needed?

src/server/dflycmd.cc
kostasrim: There is a reason why we should sent `NO PARTIAL`. Only the node that is being taken over knows if all nodes reached data parity. If this contract is not true, then we can have data loss if we enter partial sync. 

Since we relaxed the requirement that the takeover might not sync all nodes, we should should also not allow partial sync for those nodes when that happens. To achieve this we need the taken over node to communicate that

src/server/server_family.cc
kostasrim: To disable partial sync if all the nodes did not reach data parity before takeover completes. If we don't do that then we have data loss.

romange: so i was under the assumption that psync protocol falls back to full sync if a replica is not up to date. 
correct me if I am wrong, the data loss happens because we reset LSN to 1 and in that case PSYNC can not validate staleness?

kostasrim: > so i was under the assumption that psync protocol falls back to full sync if a replica is not up to date.

That's the main idea yes and if we could achieve this here it would be great.

>correct me if I am wrong, the data loss happens because we reset LSN to 1 and in that case PSYNC can not validate staleness? 

Almost. Master has lsn 200, replica_1 lsn 200 and replica_2 lsn 199. We promote replica_1 to master and start the journal  replica_2 stays at 199. Journal of replica_1 (now is the master) starts at `1` *and* the `200th` entry is not stored in any buffer. 

I wonder if we can start the journal earlier within the TakeOver instead. That way, the last entries will end up in the partial sync buffers of the replica and if we start those buffers at a specific LSN then the math should work out. 

Let me try this and get back to you

src/server/rdb_load.cc
```diff
@@ -2390,8 +2390,12 @@ error_code RdbLoaderBase::HandleJournalBlob(Service* service) {
     SET_OR_RETURN(journal_reader_.ReadEntry(), entry);
     done++;
 
-    if (entry.cmd.cmd_args.empty())
+    if (entry.cmd.cmd_args.empty()) {
```
kostasrim, line 2393: We had a bug here :)

src/server/dflycmd.cc
```diff
@@ -318,41 +305,13 @@ void DflyCmd::Flow(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext* cn
...
-      if (IsLSNDiffBellowThreshold(data.value().last_journal_LSNs, lsn_vec)) {
-        sync_type = "PARTIAL";
-        flow.start_partial_sync_at = sf_->journal()->GetLsn();
-        VLOG(1) << "Partial sync requested from LSN=" << flow.start_partial_sync_at.value()
-                << " and is available. (current_lsn=" << sf_->journal()->GetLsn() << ")";
+      auto seqid = ParseLsnVec(*last_master_lsn, data->last_journal_LSNs.size(), flow_id, rb);
+      if (!seqid) {
+        return;
       }
+      MaybePartialSync(*seqid, sync_type, flow);
```
kostasrim: Finally the two partial sync flows came together!

src/server/dflycmd.cc
```diff
@@ -438,6 +397,48 @@ void DflyCmd::StartStable(CmdArgList args, Transaction* tx, RedisReplyBuilder* r
...
+            << " that the replication buffer doesn't contain this anymore (current_lsn="
+            << sf_->journal()->GetLsn() << "). Will perform a full sync of the data.";
+  LOG(INFO) << "If this happens often you can control the replication buffer's size with the "
+               "--shard_repl_backlog_len option";
+}
+
+std::optional<LSN> DflyCmd::ParseLsnVec(std::string_view last_master_lsn,
+                                        size_t last_journal_lsn_size, size_t flow_id,
+                                        facade::RedisReplyBuilder* rb) {
+  std::vector<std::string_view> lsn_str_vec = absl::StrSplit(last_master_lsn, '-');
```
kostasrim, line 429: Same as before just moved to a function

src/server/dflycmd.cc
```diff
@@ -438,6 +397,48 @@ void DflyCmd::StartStable(CmdArgList args, Transaction* tx, RedisReplyBuilder* r
   return rb->SendOk();
 }
 
+void DflyCmd::MaybePartialSync(LSN lsn, std::string& sync_type, FlowInfo& flow) {
```
kostasrim: same as before, I merged the partial sync paths

src/server/dflycmd.cc
```diff
@@ -318,41 +305,13 @@ void DflyCmd::Flow(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext* cn
...
-      }
-
-      if (IsLSNDiffBellowThreshold(data.value().last_journal_LSNs, lsn_vec)) {
-        sync_type = "PARTIAL";
-        flow.start_partial_sync_at = sf_->journal()->GetLsn();
-        VLOG(1) << "Partial sync requested from LSN=" << flow.start_partial_sync_at.value()
-                << " and is available. (current_lsn=" << sf_->journal()->GetLsn() << ")";
+      auto seqid = ParseLsnVec(*last_master_lsn, data->last_journal_LSNs.size(), flow_id, rb);
+      if (!seqid) {
+        return;
```
romange:
```suggestion
        return;   // ParseLsnVec replies in case of error
```

src/server/dflycmd.cc
romange: great, but it's a bit confusing having the same variable name `seqid` for both branches where one of them is defined far away from here.

src/server/dflycmd.cc
```diff
@@ -531,6 +532,27 @@ void DflyCmd::TakeOver(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext
...
+      auto cb = [&, end_time](EngineShard* shard) {
+        if (!WaitReplicaFlowToCatchup(end_time, replica_ptr.get(), shard)) {
+          rest_catchup_success.store(false);
+        }
+      };
+      shard_set->RunBlockingInParallel(std::move(cb));
+    }
+
+    if (!rest_catchup_success) {
+      LOG(ERROR) << "Some of the replica nodes did not sync in time.";
```
romange:
```suggestion
      LOG(WARNING) << "Some of the replica nodes did not sync in time.";
```

src/server/dflycmd.h
```diff
@@ -220,6 +220,13 @@ class DflyCmd {
   // Main entrypoint for stopping replication.
   void StopReplication(uint32_t sync_id) ABSL_LOCKS_EXCLUDED(mu_);
 
+  std::optional<LSN> ParseLsnVec(std::string_view lsn_vec, size_t flow_id,
+                                 size_t last_journal_lsn_size, facade::RedisReplyBuilder* rb);
+
+  // If lsn exists in buffer update sync_type to FULL_SYNC.
+  // Also if we can do partial, update flow.pstart_partial_sync_at
+  void MaybePartialSync(LSN lsn, std::string& sync_type, FlowInfo& flow);
```
romange: output variables by pointer

src/server/journal/journal.h
```diff
@@ -20,6 +20,9 @@ class Journal {
   Journal();
 
   void StartInThread();
+  // Starts the journal at specified LSN
```
romange, line 24: whitespace before comments

src/server/replica.cc
```diff
@@ -952,6 +953,12 @@ void DflyShardReplica::StableSyncDflyReadFb(ExecutionState* cntx) {
     } else if (tx_data->opcode == journal::Op::PING) {
       force_ping_ = true;
       journal_rec_executed_.fetch_add(1, std::memory_order_relaxed);
+      auto* journal = ServerState::tlocal()->journal();
+      if (journal) {
+        // We must register this entry to the journal to allow partial sync
+        // if journal is active.
+        journal->RecordEntry(0, journal::Op::PING, 0, 0, nullopt, {});
```
romange, line 960: why do we have to register this entry?

src/server/replica.h
```diff
@@ -145,6 +146,16 @@ class Replica : ProtocolClient {
   // Get the current replication phase based on state_mask_
   std::string GetCurrentPhase() const;
 
+  // Used *only* in TakeOver flow. There is small data race if
+  // thread_flow_map_ gets written by the MainReplicationFiber thread but
+  // the chances for that are extremely rare.
+  std::vector<unsigned> GetFlowMapAtIndex(size_t index) const {
+    DCHECK(index < thread_flow_map_.size());
```
romange: we do not use DCHECK in header files and I am surprised it compiled. you can use assert though

src/server/replica.h
```diff
@@ -145,6 +146,16 @@ class Replica : ProtocolClient {
...
+    return thread_flow_map_[index];
```
romange: thread_flow_map_.at(index) is checked in debug

src/server/replica.h
```diff
@@ -145,6 +146,16 @@ class Replica : ProtocolClient {
   // Get the current replication phase based on state_mask_
   std::string GetCurrentPhase() const;
 
+  // Used *only* in TakeOver flow. There is small data race if
```
romange, line 149: what do you mean about the data race?

src/server/dflycmd.cc
kostasrim: I had the same comment and removed it before I pushed because the rationale was that passing the `rb` explicitly as an argument implies that we reply. Will add it back 😄

src/server/dflycmd.h
kostasrim: Uh I thought we relaxed that for some reason. Will revert (I hate this annotation because pointer implies optional and there is nothing optional in this function -- it hijacks the interface for something that your IDE can show you -- the argument types of the function you call)

src/server/replica.cc
kostasrim: Because we count it and include it in the LSN buffers of the master. So when the replica receives it, it should do so (we always sent PING on takeover). 

I thought about not storing it in the LSN buffers, then show that the rdb_loader can load it without issues (with a small bug I fixed) and I ended up including it

src/server/replica.h
kostasrim: Why ? It's a plain function internally when you expand the macro ?

(Will move to the translation unit -- just trying to understand why)

src/server/replica.h
kostasrim: The proper`data race` as defined per the C++ standard 😄 Specifically:

```
   516   thread_flow_map_ = Partition(shard_flows_.size());  
``` 

From the main replication fiber races with this. *but* it's similar to the race we have with `Sock()` in `info` command. For this to happen we need the replica to disconnect when we poll this function in the TakeOver, `an extremely unlikely scenario`. 

It should be fine to stay as is IMO -- my replication PR also changes how we make changes to `thread_flow_map`

src/server/replica.h
kostasrim: Isn't it the case that `.at()` performs a run time check regardless of build type ? It's not only in debug basically

kostasrim: https://en.cppreference.com/w/cpp/container/vector/at.html 

> Returns a reference to the element at specified location pos, with bounds checking.
If pos is not within the range of the container, an exception of type [std::out_of_range](https://en.cppreference.com/w/cpp/error/out_of_range.html) is thrown.

src/server/dflycmd.cc
```diff
@@ -318,41 +305,13 @@ void DflyCmd::Flow(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext* cn
...
-      if (IsLSNDiffBellowThreshold(data.value().last_journal_LSNs, lsn_vec)) {
-        sync_type = "PARTIAL";
-        flow.start_partial_sync_at = sf_->journal()->GetLsn();
-        VLOG(1) << "Partial sync requested from LSN=" << flow.start_partial_sync_at.value()
-                << " and is available. (current_lsn=" << sf_->journal()->GetLsn() << ")";
+      auto flow_lsn = ParseLsnVec(*last_master_lsn, data->last_journal_LSNs.size(), flow_id, rb);
+      if (!flow_lsn) {
+        return;  // ParseLsnVec replies in case of error
       }
+      MaybePartialSync(*flow_lsn, &sync_type, &flow);
```
romange: sorry, now I looked at what this function do and I think its interface is still akward.
the reason for this is that you wanted to get rid of the duplicated code but I think it can be done differently.
1. introduce in dflycmd.cc `bool IsInJournal(LSN lsn)` that only does the check
2. both branches here will set `seqid` if IsInJournal returns true and clear `seqid` otherwise.

before `rb->StartArray(2);` line based on seqid status you do all the updates/logs that currently reside in `MaybePartialSync`. `MaybePartialSync` is not needed.

src/server/dflycmd.cc
```diff
@@ -438,6 +397,48 @@ void DflyCmd::StartStable(CmdArgList args, Transaction* tx, RedisReplyBuilder* r
...
+}
+
+std::optional<LSN> DflyCmd::ParseLsnVec(std::string_view last_master_lsn,
+                                        size_t last_journal_lsn_size, size_t flow_id,
+                                        facade::RedisReplyBuilder* rb) {
+  std::vector<std::string_view> lsn_str_vec = absl::StrSplit(last_master_lsn, '-');
+  if (lsn_str_vec.size() != last_journal_lsn_size) {
+    rb->SendError(facade::kSyntaxErr);  // Unexpected flow. LSN vector of same master
+                                        // should be the same size on all replicas.
+    return {};
```
romange: nit: i find nullopt more clear.

src/server/journal/journal.h
```diff
@@ -21,6 +21,10 @@ class Journal {
 
   void StartInThread();
 
+  // Starts the journal at specified LSN
+  // Also drops the (resets) the partial sync buffers
+  void StartInThreadAtLsn(LSN lsn);
```
romange, line 26: question: do we really need both `StartInThread` and `StartInThreadAtLsn`?  what is the case when we call StartInThread and we DO NOT want to reset partial sync buffers?

romange (changes requested):

src/server/journal/journal.h
kostasrim: Yes, although it's functionality can be merged (if we accept default parameters which `usually` I try to avoid).

>  what is the case when we call StartInThread

Idempotent semantics. When a replica gets registered to a master, the master tries to Start the journal. So if we reset the buffers we disabled partial sync (for each time we register a new replica)

src/server/dflycmd.cc
kostasrim: Sure, whatever works -- I don't have a strong opinion on this

src/server/dflycmd.cc
kostasrim: Done that! With a small twist -- I added a separate variable than seqid (to separate context) + LOG if we can't partial sync inside `IsInJournal`

src/server/journal/journal.h
romange: ok, fair enough.

romange (approved):

PR #5728, opened 2025-08-27
fix: replication for Z...STORE commands

BorysTheDev:
related to: #5413
problem: during full_sync phaze Z...STORE commands can create an incorrect result on the replica because some keys can be absent.

fix: Z...STORE operations transform into DEL ... and ZADD ... operations

### DIFF
--- a/src/server/zset_family.cc
+++ b/src/server/zset_family.cc
@@ -2051,16 +2051,16 @@ OpResult<ZSetFamily::AddResult> ZSetFamily::OpAdd(const OpArgs& op_args,
     if (zparams.override) {
       RecordJournal(op_args, "DEL"sv, ArgSlice{key});
     }
-    size_t size = members.size() * 2 + 1;
-    vector<string> scores(members.size());
-    vector<string_view> mapped(size);
-    mapped[0] = key;
-    auto mem_it = members.begin();
-    auto score_it = scores.begin();
-    for (size_t i = 1; i < size; i += 2, ++mem_it, ++score_it) {
-      *score_it = absl::StrCat(mem_it->first);
-      mapped[i] = *score_it;
-      mapped[i + 1] = mem_it->second;
+
+    vector<string> scores;
+    vector<string_view> mapped;
+    scores.reserve(members.size());
+    mapped.reserve(members.size() * 2 + 1);
+    mapped.push_back(key);
+    for (const auto& [score, member] : members) {
+      scores.push_back(absl::StrCat(score));
+      mapped.push_back(scores.back());
+      mapped.push_back(member);
     }
     RecordJournal(op_args, "ZADD"sv, mapped);
   }
--- a/src/server/zset_family.cc
+++ b/src/server/zset_family.cc
@@ -152,9 +152,9 @@ void OutputScoredArrayResult(const OpResult<ScoredArray>& result, SinkReplyBuild
   rb->SendScoredArray(result.value(), true /* with scores */);
 }
 
-OpResult<DbSlice::ItAndUpdater> FindZEntry(const ZSetFamily::ZParams& zparams,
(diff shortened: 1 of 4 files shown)


src/server/zset_family.cc
```diff
@@ -2039,6 +2044,26 @@ OpResult<ZSetFamily::AddResult> ZSetFamily::OpAdd(const OpArgs& op_args,
...
+    vector<string_view> mapped(size);
+    mapped[0] = key;
+    auto mem_it = members.begin();
+    auto score_it = scores.begin();
+    for (size_t i = 1; i < size; i += 2, ++mem_it, ++score_it) {
+      *score_it = absl::StrCat(mem_it->first);
+      mapped[i] = *score_it;
+      mapped[i + 1] = mem_it->second;
+    }
+    RecordJournal(op_args, "ZADD"sv, mapped);
```
kostasrim:
```suggestion
const size_t sz = members.size();
vector<string> scores(sz);
vector<string_view> mapped(sz * 2 + 1);
for (auto &[score, member] : members) {
  scores.push_back(absl::StrCat(score));
  mapped.push_back(scores.back());
  mapped.push_back(member);
}
```

src/server/zset_family.cc
```diff
@@ -2039,6 +2044,26 @@ OpResult<ZSetFamily::AddResult> ZSetFamily::OpAdd(const OpArgs& op_args,
...
     return op_status;
+
+  // TODO: consider optimization to record real command if the replica is in stable_sync state
+  // and there is no slot migration process going on.
+  if (zparams.journal_update && op_args.shard->journal()) {
+    if (zparams.override) {
+      RecordJournal(op_args, "DEL"sv, ArgSlice{key});
+    }
+    size_t size = members.size() * 2 + 1;
+    vector<string> scores(members.size());
```
kostasrim: What if `members.size == 1 million` ?

BorysTheDev: the same as 1, we try to send it, we have the same in S...STORE commands family

src/server/zset_family.cc
BorysTheDev: I like your suggestion, but unfortunately it is less efficient

kostasrim: how is it less efficient ? could you plz elaborate

src/server/zset_family.cc
kostasrim: But wouldn't that "freeze" dragonfly for that duration ? Doesn't that skyrocket latency ?

src/server/zset_family.cc
BorysTheDev: First of all, there are bugs in this code, you set scores and mapped with some size and then do push_back after this size, also you have missed mapped[0] = key. Regarding performance push_back is several operations including "if", so this code even if we fix it will work slower

src/server/zset_family.cc
BorysTheDev: you have absolutely the same problem if you try to replicate the SET size of 1000 000.

src/server/zset_family.cc
kostasrim: > First of all, there are bugs in this code

I am pretty sure there are, it was meant to serve as an example. 

> you set scores and mapped with some size

Just reserve instead...

> ou have missed mapped[0] = key. 

Just add this...

> Regarding performance push_back is several operations including "if", so this code even if we fix it will work slower

That's not true. There are other factors involved. CPU's have branch prediction and speculative execution so a single branch will not have any effect. But imo it's pointless to micro-optimize thin air (the if branch) without concrete proof.
And I would sacrifice the 0.1% of cpu time any day to make something more readable. Iterator code is just not nice and if it can be avoided I don't understand why not to

PR #5733, opened 2025-08-28
fix(range_tree): Fix default range block size

BagritsevichStepan:
Fixes default max size for the range block.

When initializing the index, if this number is too small, it can lead to a large number of splits, resulting in very long execution. To prevent this, we can increase the block size.


BagritsevichStepan:
No need to change the default block size

Issue #5337, opened 2025-06-21
JSONPath cannot access object properties with numeric string keys


cyppe:
# JSONPath cannot access object properties with numeric string keys

  

## Summary

DragonflyDB's JSONPath implementation fails to access object properties when the property names are numeric strings (e.g., `"10"`, `"8"`, `"14"`)

  

## Environment

- DragonflyDB version: dragonfly_version:df-v1.30.3
  

### 1. Set up test data

```bash

JSON.SET test:images $ '{"images":[{"id":1,"sizes":{"1":"small.jpg","10":"medium.jpg","14":"large.jpg","8":"thumb.jpg"}}]}'

```

  

### 2. Verify data structure and keys exist

```bash

# ✅ This works - shows the whole object

JSON.GET test:images $.images[0].sizes

  

# ✅ This works - confirms keys exist

JSON.OBJKEYS test:images $.images[0].sizes

  

# ✅ This works - shows object type

JSON.TYPE test:images $.images[0].sizes

```

  

**Expected output:**

```json

// JSON.GET

[{"1":"small.jpg","10":"medium.jpg","14":"large.jpg","8":"thumb.jpg"}]

  

// JSON.OBJKEYS

[["1","10","14","8"]]

  

// JSON.TYPE

["object"]

```

  

### 3. Try to access individual numeric keys (ALL FAIL)

```bash

# ❌ RFC 9535 dot notation

JSON.GET test:images $.images[0].sizes.10

  

# ❌ Bracket notation with double quotes

JSON.GET test:images '$.images[0].sizes["10"]'

  

# ❌ Bracket notation with single quotes

JSON.GET test:images "$.images[0].sizes['10']"

  

# ❌ Full bracket notation

JSON.GET test:images '$.images[0]["sizes"]["10"]'

  

# ❌ Without quotes (numeric index)

JSON.GET test:images $.images[0].sizes[10]

  

# ❌ Try other numeric keys

JSON.GET test:images $.images[0].sizes.8

JSON.GET test:images $.images[0].sizes.14

JSON.GET test:images '$.images[0].sizes["8"]'

JSON.GET test:images '$.images[0].sizes["14"]'

```

  

**Expected output for each:**

```json

["medium.jpg"] // for key "10"

["thumb.jpg"] // for key "8"

["large.jpg"] // for key "14"

```

  

**Actual output:** `null` or `[]` for all commands

  

### 4. Verify non-numeric keys work fine

```bash

[...]

PR #5737, opened 2025-08-28
fix: jsonPath parser for int values

BorysTheDev:
fixes: #5337
problem: We always treat INT value as an integer, but it can be an identifier or part of the expression
fix: consider INT as a string and convert it to int only if context is correct

### DIFF
--- a/src/core/json/jsonpath_grammar.y
+++ b/src/core/json/jsonpath_grammar.y
@@ -25,10 +25,19 @@
 
 #include "src/core/json/lexer_impl.h"
 #include "src/core/json/driver.h"
+#include <absl/strings/numbers.h>
+#include "base/logging.h"
 
 #define yylex driver->lexer()->Lex
 
 using namespace std;
+
+static int unsafe_stoi(std::string_view s) {
+  int value;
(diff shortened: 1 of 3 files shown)


src/core/json/jsonpath_grammar.y
```diff
@@ -25,10 +25,19 @@
...
+#include "base/logging.h"
 
 #define yylex driver->lexer()->Lex
 
 using namespace std;
+
+static int unsafe_stoi(std::string_view s) {
+  int value;
+  bool success = absl::SimpleAtoi(s, &value);
+  DCHECK(success);
```
vyavdoshenko, line 38: It looks to me like a wrong branch of code. 
I think, if you can't parse a valid number, you have to use the string-like identifier fallback.

BorysTheDev: didn't get

vyavdoshenko: I looked at the rules; it should always be a valid int. So it will never be another type.

vyavdoshenko (approved):
lgtm

PR #5748, opened 2025-08-29
fix(block_list): Speed up Split method

BagritsevichStepan:
In the `Split` method we had a rather inefficient algorithm.

First, we used `Insert`, which performs a binary search to find the block and then inserts a new entry. However, during splitting we already have the document IDs sorted, meaning every new entry is simply appended to the end of the block list.

The second issue was that `Insert` always triggered block splitting inside `BlockList`, which is a linear operation. Together, these introduced significant overhead in the split method.

To address this, I added a `PushBack` method to `BlockList` that bypasses binary search and avoids block splitting.

### DIFF
--- a/src/core/search/block_list.cc
+++ b/src/core/search/block_list.cc
@@ -10,25 +10,27 @@ SplitResult Split(BlockList<SortedVector<std::pair<DocId, double>>>&& block_list
 
   const size_t elements_count = block_list.Size();
 
-  using EntryIndex = std::pair<size_t /*block id*/, size_t /*entry id*/>;
-  std::vector<EntryIndex> entries_indexes(elements_count);
+  // Extract values to find median
+  std::vector<double> entries_values(elements_count);
   size_t index = 0;
-  for (size_t i = 0; i < block_list.blocks_.size(); ++i) {
-    auto& block = block_list.blocks_[i];
-    for (size_t j = 0; j < block.Size(); ++j) {
-      entries_indexes[index++] = {i, j};
-    }
+  for (const Entry& entry : block_list) {
+    entries_values[index++] = entry.second;
   }
 
-  std::nth_element(entries_indexes.begin(), entries_indexes.begin() + elements_count / 2,
-                   entries_indexes.end(), [&](const EntryIndex& l, const EntryIndex& r) {
-                     return block_list.blocks_[l.first][l.second].second <
-                            block_list.blocks_[r.first][r.second].second;
-                   });
+  // Find median value
+  std::nth_element(entries_values.begin(), entries_values.begin() + elements_count / 2,
+                   entries_values.end());
+  double median_value = entries_values[elements_count / 2];
 
-  const EntryIndex median_index = entries_indexes[elements_count / 2];
-  double median_value = block_list.blocks_[median_index.first][median_index.second].second;
+  /* Now we need to split entries into two parts, left and right, so that:
+   1) left has values < median_value
+   2) right has values >= median_value
+   3) both parts have approximately the same number of elements
 
+   To achieve this, we first split entries into three parts: < median_value (left blocklist), ==
+   median_value (median_entries), > median_value (righ blocklist). Then we add == median_value part
+   to the smaller of the two parts (< or >). This guarantees that both parts have approximately the
+   same number of elements */
   BlockList<SortedVector<Entry>> left(block_list.blocks_.get_allocator().resource(),
                                       block_list.block_size_);
   BlockList<SortedVector<Entry>> right(block_list.blocks_.get_allocator().resource(),
@@ -39,7 +41,7 @@ SplitResult Split(BlockList<SortedVector<std::pair<DocId, double>>>&& block_list
   right.ReserveBlocks(block_list.blocks_.size() / 2 + 1);
 
   double min_value_in_right_part = std::numeric_limits<double>::infinity();
-  for (const auto& entry : block_list) {
+  for (const Entry& entry : block_list) {
     if (entry.second < median_value) {
       left.PushBack(entry);
     } else if (entry.second > median_value) {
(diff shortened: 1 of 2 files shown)


BagritsevichStepan:
Before:
| Benchmark                                  | Time (ns)   | CPU (ns)   | Iterations |
|--------------------------------------------|-------------|------------|------------|
| BM_SearchRangeTreeSplits/block_size:100000 | 58,746,631  | 58,736,587 | 10         |
| BM_SearchRangeTreeSplits/block_size:1000000| 569,836,676 | 569,791,624| 1          |


After:
| Benchmark                                  | Time (ns)  | CPU (ns)   | Iterations |
|--------------------------------------------|------------|------------|------------|
| BM_SearchRangeTreeSplits/block_size:100000 | 18,841,322 | 18,837,811 | 39         |
| BM_SearchRangeTreeSplits/block_size:1000000| 193,917,459| 193,877,924| 4          |

src/core/search/block_list.cc
```diff
@@ -80,6 +84,18 @@ template <typename C> bool BlockList<C>::Insert(ElementType t) {
   return true;
 }
 
+template <typename C> bool BlockList<C>::PushBack(ElementType t) {
+  if (blocks_.empty() || (blocks_.back().Size() >= block_size_ * 2 - 1)) {
```
romange: can you explain this condition?

src/core/search/block_list.cc
```diff
@@ -6,48 +6,52 @@ using namespace std;
...
-    all_entries.reserve(all_entries.size() +
-                        block.Size());  // Reserve space for all entries in the block
-    all_entries.insert(all_entries.end(), block.begin(), block.end());
-    block.Clear();  // At the same time clear the block to save memory
-  }
-  block_list.Clear();
+  const size_t elements_count = block_list.Size();
 
-  DCHECK(all_entries.size() == initial_size);
+  using EntryIndex = std::pair<size_t /*block id*/, size_t /*entry id*/>;
```
romange: can you please add comments to this function roughly describing each phase?

src/core/search/block_list.cc
```diff
@@ -6,48 +6,52 @@ using namespace std;
...
+  const EntryIndex median_index = entries_indexes[elements_count / 2];
+  double median_value = block_list.blocks_[median_index.first][median_index.second].second;
 
   BlockList<SortedVector<Entry>> left(block_list.blocks_.get_allocator().resource(),
                                       block_list.block_size_);
   BlockList<SortedVector<Entry>> right(block_list.blocks_.get_allocator().resource(),
                                        block_list.block_size_);
   absl::InlinedVector<Entry, 1> median_entries;
 
+  left.ReserveBlocks(block_list.blocks_.size() / 2 + 1);
```
romange, line 40: don't you over reserve here?
say if `block_list.blocks_.size()` is 4, you reserve 3 in each list?

src/core/search/block_list.cc
```diff
@@ -6,48 +6,52 @@ using namespace std;
...
+  right.ReserveBlocks(block_list.blocks_.size() / 2 + 1);
+
   double min_value_in_right_part = std::numeric_limits<double>::infinity();
-  for (const auto& entry : all_entries) {
+  for (const auto& entry : block_list) {
```
romange:
```suggestion
  for (const Entry& entry : block_list) {
```

src/core/search/block_list.cc
```diff
@@ -6,48 +6,52 @@ using namespace std;
...
+  for (size_t i = 0; i < block_list.blocks_.size(); ++i) {
+    auto& block = block_list.blocks_[i];
+    for (size_t j = 0; j < block.Size(); ++j) {
+      entries_indexes[index++] = {i, j};
+    }
+  }
 
-  std::nth_element(all_entries.begin(), all_entries.begin() + initial_size / 2, all_entries.end(),
-                   [](const Entry& l, const Entry& r) { return l.second < r.second; });
+  std::nth_element(entries_indexes.begin(), entries_indexes.begin() + elements_count / 2,
```
romange: I am confused. Are all elements already sorted by the score? you use PushBack because they are sorted?
In that case, why do you need to partial sort `entries_indexes` here? would not it be just advancing the iterator of block_list by `elements_count / 2` steps?

src/core/search/block_list.cc
dranikpg: All this code did was find the median value, it didn't need the pair. Same with indices, you use it once to access the value 🤔 Isn't `vector<double>` enough?

src/core/search/block_list.cc
dranikpg: As long as the blocks grow by themselves are not preallocated that's not a problem

src/core/search/block_list.cc
dranikpg: Filling up a block usually splits it. With this condition we just switch to a new one before reaching the split limit

dranikpg: The header needs a comment definitely as well

dranikpg: Externally building a structure with internal invariants is always nasty, the best option would've been just a constructor from an iterator, but then we need to handle medians, etc...

dranikpg: I also assume that for most cases Insert is called in doc id growing order as well
1. We can add an .end() lookup optimizaton to avoid calling FindBlock()
2. We can enfore the last block to spill over when it reaches `N` because we exclude it by default from merging

i.e. Insert can work just as PushBack (with a small penalty on last element lookup in that case)

src/core/search/block_list.cc
BagritsevichStepan: The elements are `std::pair<DocId, double>`, meaning we store a numeric value associated with its doc id. In the `BlockList`, elements are naturally sorted by `DocId`. Therefore, to find the median, we need to "sort" them by their values instead

dranikpg (approved):

Issue #5745, opened 2025-08-29
Support returning floats as int values for RESP3


abhijat:
`--lua_resp2_legacy_float=true` should work for resp3 as it does now for resp2

PR #5754, opened 2025-09-01
server: Enable lua legacy float response for RESP3

abhijat:
Previously when the lua legacy float flag was set, doubles would be sent as floats for RESP2 but for RESP3 this behavior was skipped. This changeset extends the behavior to RESP3 as well.

fixes https://github.com/dragonflydb/dragonfly/issues/5745

Note: the previous behavior was not a bug, but the flag needs to be extended to RESP3 now in response to real world requirements where third party libraries such as `redis_rate` use a RESP3 client but expect long values in response.

### DIFF
--- a/src/server/dragonfly_test.cc
+++ b/src/server/dragonfly_test.cc
@@ -837,7 +837,7 @@ TEST_F(DflyEngineTest, Latency) {
 
 TEST_F(DflyEngineTest, EvalBug2664) {
   absl::FlagSaver fs;
-  absl::SetFlag(&FLAGS_lua_resp2_legacy_float, true);
+  SetFlag(&FLAGS_lua_resp2_legacy_float, true);
 
   auto resp = Run({"eval", "return 42.9", "0"});
   EXPECT_THAT(resp, IntArg(42));
--- a/src/server/dragonfly_test.cc
+++ b/src/server/dragonfly_test.cc
@@ -848,7 +848,7 @@ TEST_F(DflyEngineTest, EvalBug2664) {
   ASSERT_THAT(resp, ArrLen(14));
 
   resp = Run({"eval", "return 42.9", "0"});
-  EXPECT_THAT(resp, DoubleArg(42.9));
+  EXPECT_THAT(resp, IntArg(42));
 }
 
 TEST_F(DflyEngineTest, MemoryUsage) {
--- a/src/server/main_service.cc
+++ b/src/server/main_service.cc
@@ -348,11 +348,12 @@ class EvalSerializer : public ObjectExplorer {
   }
 
   void OnDouble(double d) final {
-    if (rb_->IsResp3() || !GetFlag(FLAGS_lua_resp2_legacy_float)) {
-      rb_->SendDouble(d);
-    } else {
-      long val = d >= 0 ? static_cast<long>(floor(d)) : static_cast<long>(ceil(d));
+    thread_local const bool send_double_as_long = GetFlag(FLAGS_lua_resp2_legacy_float);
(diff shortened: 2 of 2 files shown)


src/server/dragonfly_test.cc
```diff
@@ -848,7 +848,7 @@ TEST_F(DflyEngineTest, EvalBug2664) {
   ASSERT_THAT(resp, ArrLen(14));
 
   resp = Run({"eval", "return 42.9", "0"});
-  EXPECT_THAT(resp, DoubleArg(42.9));
+  EXPECT_THAT(resp, IntArg(42));
```
kostasrim, line 851: That's resp 2 no ?

src/server/main_service.cc
```diff
@@ -348,11 +348,12 @@ class EvalSerializer : public ObjectExplorer {
   }
 
   void OnDouble(double d) final {
-    if (rb_->IsResp3() || !GetFlag(FLAGS_lua_resp2_legacy_float)) {
-      rb_->SendDouble(d);
-    } else {
-      long val = d >= 0 ? static_cast<long>(floor(d)) : static_cast<long>(ceil(d));
+    thread_local const bool send_double_as_long = GetFlag(FLAGS_lua_resp2_legacy_float);
```
kostasrim: What's the point of thread local here ? Also I think you inverted the condition, now you sent `long` on resp 2`

abhijat: `thread_local` is to not compute the flag each time, as I see this cannot be reset by `config set`. It will only be computed the first time.

I didn't intend to invert the condition, but it looks like it changes behavior for RESP2. Earlier it was:

* if resp3 OR flag is false, send double
* otherwise send long

I changed it to
* if flag is true, send long
* otherwise send double

It should be:
* if flag is true OR resp2, send long
* otherwise send double

I will change it for resp2

kostasrim: > thread_local is to not compute the flag each time, as I see this cannot be reset by config set. It will only be computed the first time.

But you compute it every time this function is called since it's not static ? :shrug: 

Uh you also inverted the branches -- I just realized

abhijat: `thread_local` variables are computed once for the thread lifetime so the end result should be the same. I will change it to static though as I think `thread_local` doesn't make much sense here as this is `main_service` and nothing per shard and static is easier to understand in this context.

I see a bug is introduced by my changes, I realize by looking at the `EvalResp` test. Will change this code.

kostasrim (approved):

src/server/main_service.cc
kostasrim: uh implicit static for thread locals :cry:

romange: you do not need `thread_local` here for sure. we use thread locals when we mutate the setting via `CONFIG SET`.
you can just use a global static.

abhijat: Changed to static.

abhijat: I thought I had introduced some unwanted change to behavior with respect to RESP2, but on review it seems correct to me. 

Earlier, if the flag was _not_ set, or we had resp3, we sent double. This means the flag did not work for resp3.

Now if we have the flag set, we send long (irrespective of resp2/resp3). If it is not set, we send double, again irrespective of the resp version. 

This behavior seems correct to me. A unit test fails when run with others (not in isolation), because of the static evaluation of the flag, it is not reevaluated even if set in the `EvalBug2664`

abhijat: I removed static evaluation altogether as unit tests break with it.

PR #5755, opened 2025-09-01
fix: search index restore when index names start with ‘:’ (e.g. :Order:index)

vyavdoshenko:
Root cause:
LoadSearchIndexDefFromAux fed raw index definitions to the RESP parser; a leading ‘:’ was misparsed as a RESP token, resulting in a parse failure and “Bad index definition” error.

Fix:
In RdbLoader::LoadSearchIndexDefFromAux, force inline parsing by inserting a leading space before the definition (prior to CRLF). This prevents RESP misinterpretation; then FT.CREATE is prepended and executed correctly.
Test coverage: Added tests/dragonfly/search_test.py::test_restore_search_index_with_colon_name. It creates :Order:index, runs SAVE DF -> FLUSHALL -> DFLY LOAD, asserts the index is restored and that searching returns the expected result.

Impact:
Minimal, affects only the parsing of search-index definitions during restore. Snapshot format unchanged.

Example error log (before fix):
```
rdb_load.cc:2794] Bad index definition: :Order:index ON HASH PREFIX 1 :Order: STOPWORDS 33 ... SCHEMA name AS name_fts TEXT ...
```

### DIFF
--- a/src/server/rdb_test.cc
+++ b/src/server/rdb_test.cc
@@ -670,6 +670,26 @@ TEST_F(RdbTest, SBF) {
   EXPECT_THAT(Run({"BF.EXISTS", "k", "1"}), IntArg(1));
 }
 
+TEST_F(RdbTest, RestoreSearchIndexNameStartingWithColon) {
+  // Create an index with a name that starts with ':' and add a sample document
+  EXPECT_EQ(Run({"FT.CREATE", ":Order:index", "ON", "HASH", "PREFIX", "1", ":Order:", "SCHEMA",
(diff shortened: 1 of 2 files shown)


romange:
it's not clear from the PR description what the issue is

vyavdoshenko:
> it's not clear from the PR description what the issue is

I updated the title and the description.

romange:
LGTM, but such test should reside in unit tests, not the regression test. Anything that covers internal server logic (i.e. not facade) should be added to our unit tests. 

Specifically, it should be added to rdb_test.cc and you can use `Run({"debug", "reload"})` to reload the server with the snapshot. See how other tests do it.

PR #5758, opened 2025-09-01
[POC - do not review yet] chore: almost mutex free replication

kostasrim:
Do not review.

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -1229,12 +1229,11 @@ void ServerFamily::Shutdown() {
       replica_->Stop();
     }
     StopAllClusterReplicas();
+    UpdateReplicationThreadLocals(nullptr);
 
     dfly_cmd_->Shutdown();
     DebugCmd::Shutdown();
   });
-
-  UpdateReplicationThreadLocals(nullptr);
 }
 
 bool ServerFamily::HasPrivilegedInterface() {
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -1234,10 +1234,7 @@ void ServerFamily::Shutdown() {
     DebugCmd::Shutdown();
   });
 
-  service_.proactor_pool().AwaitFiberOnAll([](auto index, auto* cntx) {
-    tl_replica = nullptr;
-    tl_cluster_replicas.clear();
-  });
+  UpdateReplicationThreadLocals(nullptr);
 }
 
 bool ServerFamily::HasPrivilegedInterface() {
@@ -3457,10 +3454,9 @@ void ServerFamily::ReplicaOfNoOne(SinkReplyBuilder* builder) {
     SetMasterFlagOnAllThreads(true);
     // TODO we should not allow partial sync after NO-ONE. Only after Takeover.
     last_master_data_ = replica_->Stop();
-    // TODO set thread locals to nullptr
     replica_.reset();
     StopAllClusterReplicas();
-    service_.proactor_pool().AwaitFiberOnAll([](auto index, auto* cntx) { tl_replica = nullptr; });
+    UpdateReplicationThreadLocals(nullptr);
   }
 
   // May not switch to ACTIVE if the process is, for example, shutting down at the same time.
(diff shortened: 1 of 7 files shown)


src/server/server_family.cc
```diff
@@ -3608,6 +3603,9 @@ void ServerFamily::ReplTakeOver(CmdArgList args, const CommandContext& cmd_cntx)
   SetMasterFlagOnAllThreads(true);
   last_master_data_ = replica_->Stop();
   replica_.reset();
+
+  service_.proactor_pool().AwaitFiberOnAll([](auto index, auto* context) { tl_replica = nullptr; });
```
kostasrim: also reset tl_cluster_replicas

src/server/server_family.cc
```diff
@@ -3426,77 +3439,58 @@ void ServerFamily::AddReplicaOf(CmdArgList args, const CommandContext& cmd_cntx)
...
 
-    if (replica_)
-      last_master_data = replica_->Stop();
+  if (!ss->is_master) {
+    CHECK(replica_);
+    // flip flag before clearing replica_
+    SetMasterFlagOnAllThreads(true);
+    // TODO we should not allow partial sync after NO-ONE. Only after Takeover.
+    last_master_data_ = replica_->Stop();
+    // TODO set thread locals to nullptr
```
kostasrim: done remove this comment

src/server/server_family.cc
```diff
@@ -3130,12 +3140,15 @@ string ServerFamily::FormatInfoMetrics(const Metrics& m, std::string_view sectio
         append("psync_attempts", rinfo.psync_attempts);
         append("psync_successes", rinfo.psync_successes);
       };
-      fb2::LockGuard lk(replicaof_mu_);
+      // Deep copy because tl_replica might be overwritten inbetween
+      auto replica = tl_replica;
```
kostasrim: Bread and butter no1 of this PR. Bye bye blocking info command because of the mutex.

src/server/server_family.cc
```diff
@@ -3507,30 +3501,31 @@ void ServerFamily::ReplicaOfInternal(CmdArgList args, Transaction* tx, SinkReply
...
-    builder->SendError(ec ? ec.Format() : "replication cancelled");
-    return;
+  if (ec) {
+    return builder->SendError(ec.Format());
   }
-  // Successfully connected now we flush
-  // If we are called by "Replicate", tx will be null but we do not need
-  // to flush anything.
+
+  util::fb2::LockGuard lk(replicaof_mu_);
```
kostasrim, line 3519: Bread and butter no2. No more `loading state` prematurely. We only get into `loading_state` if we are `doing full sync`. Otherwise, no state change at all.

kostasrim:
TODO: move this to `ReplicaOfInternalV2` such that the new algorithm is easy to review. IMO with the changes and this gh diff is unreviewable :scream:

src/server/server_family.cc
kostasrim: TODO worth adding a test

tests/dragonfly/replication_test.py
```diff
@@ -484,7 +481,7 @@ async def replicate():
         num_successes += await result
 
     logging.info(f"succeses: {num_successes}")
-    assert COMMANDS_TO_ISSUE > num_successes, "At least one REPLICAOF must be cancelled"
+    assert COMMANDS_TO_ISSUE == num_successes
```
kostasrim, line 480: The new algorithm does not use `two phase locking` so the following is no longer a possibility:

1. client 1 -> REPLICAOF -> Locks the mutex -> updates `replica_` to `new_replica` -> releases the mutex ->  calls replica_->Start()
2. client 2 -> REPLICAOF -> same as (1) but first calls replica_->Stop() -> releases the mutex
3. client 1 -> REPLICAOF command grabs the second lock to the mutex, observes that the context got cancelled because of step (2) and boom returns "replication cancelled"

This can not happen anymore because we lock only once and atomically update everything including stopping the previous replica. So by the time (2) grabs the lock in the example above, the previous REPLICAOF command is not in some intermediate state. To observe that indeed we cancelled, we should read the logs and see ("Stopping replication") `COMMANDS_TO_ISSUE - 1` times + 1 (because of the Shutdown() at the end)

Bonus points is that I suspect we might be able to also solve https://github.com/dragonflydb/dragonfly/pull/4685 but I will need to follow up with that

Issue #5744, opened 2025-08-29
Add metrics for memory usage by connection data structures


abhijat:
Recently it was observed that `exec_info.body` vector can grow quite large (especially when multiplied with number of connection contexts alive). While this size is visible in memory stats command, it should be visible and accounted for in metrics too, so that support and alerting can make use of it in graphs, alerts etc.

PR #5764, opened 2025-09-03
server: metrics to track stored command sizes

abhijat:
A new label `conn_stored_commands` is added to the metric `memory_by_class_bytes`. This value tracks the total size of the `StoredCmd` objects held in `ConnectionState::ExecInfo` objects in the `body` vector.

The size is added as a field on the `ExecInfo` object and a separate set of methods is added to read and write the size on this object. 

The reason for doing this and not directly using `UsedMemory` method on `ExecInfo` is that `ExecInfo::UsedMemory` will iterate over the vector to find the size. In order to update the metric, when storing a new command we must add the delta introduced by the command. For this we must find the used memory before and after the new command is stored, and `UsedMemory` must iterate over the entire vector. As the vector size grows this can become problematic as this operation must be done for each `DispatchCommand` within a multi-exec block.

The other option could be simply adding `StoredCmd::UsedMemory` to the metric each time we add a `StoredCmd` but this ignores the memory usage of the `body` container. 

The metric is decremented when the `ConnectionState::ExecInfo::Clear` method is called. This method now needs to access the connection ptr to decrement the metric.

FIXES https://github.com/dragonflydb/dragonfly/issues/5744

note: wrt https://github.com/dragonflydb/dragonfly/issues/5744 there is another relevant vector which is watched keys, a similar metric may need to be added for that if it is also known to grow to large sizes.

### DIFF
--- a/src/server/conn_context.cc
+++ b/src/server/conn_context.cc
@@ -323,7 +323,8 @@ vector<unsigned> ConnectionContext::ChangeSubscriptions(CmdArgList channels, boo
 void ConnectionState::ExecInfo::Clear() {
   DCHECK(!preborrowed_interpreter);  // Must have been released properly
   state = EXEC_INACTIVE;
-  ClearStoredCmds();
+  const size_t cleared_size = ClearStoredCmds();
+  ServerState::tlocal()->stats.stored_cmd_bytes -= cleared_size;
   is_write = false;
   ClearWatched();
 }
--- a/src/server/conn_context.cc
+++ b/src/server/conn_context.cc
@@ -235,14 +235,10 @@ void ConnectionState::ExecInfo::AddStoredCmd(const CommandId* cid, bool own_args
 }
 
 size_t ConnectionState::ExecInfo::ClearStoredCmds() {
-  const size_t used = StoredCmdBytes();
+  const size_t used = GetStoredCmdBytes();
   vector<StoredCmd>{}.swap(body);
   stored_cmd_bytes = 0;
   return used;
-}
-
-size_t ConnectionState::ExecInfo::StoredCmdBytes() const {
-  return stored_cmd_bytes + body.capacity() * sizeof(StoredCmd);
 }
 
 size_t ConnectionState::ScriptInfo::UsedMemory() const {
--- a/src/server/conn_context.cc
+++ b/src/server/conn_context.cc
@@ -229,6 +229,22 @@ size_t ConnectionState::ExecInfo::UsedMemory() const {
   return dfly::HeapSize(body) + dfly::HeapSize(watched_keys);
 }
 
+void ConnectionState::ExecInfo::AddStoredCmd(const CommandId* cid, bool own_args, CmdArgList args) {
+  body.emplace_back(cid, own_args, args);
+  stored_cmd_bytes += body.back().UsedMemory();
+}
+
+size_t ConnectionState::ExecInfo::ClearStoredCmds() {
+  const size_t used = StoredCmdBytes();
+  vector<StoredCmd>{}.swap(body);
+  stored_cmd_bytes = 0;
+  return used;
+}
+
+size_t ConnectionState::ExecInfo::StoredCmdBytes() const {
+  return stored_cmd_bytes + body.capacity() * sizeof(StoredCmd);
+}
+
 size_t ConnectionState::ScriptInfo::UsedMemory() const {
   return dfly::HeapSize(lock_tags) + async_cmds_heap_mem;
 }
@@ -307,7 +323,7 @@ vector<unsigned> ConnectionContext::ChangeSubscriptions(CmdArgList channels, boo
 void ConnectionState::ExecInfo::Clear() {
   DCHECK(!preborrowed_interpreter);  // Must have been released properly
   state = EXEC_INACTIVE;
-  vector<StoredCmd>{}.swap(body);
+  ClearStoredCmds();
   is_write = false;
   ClearWatched();
 }
--- a/src/server/conn_context.h
+++ b/src/server/conn_context.h
@@ -94,7 +94,9 @@ struct ConnectionState {
     size_t ClearStoredCmds();
 
     // Returns memory used by the body field without iterating over each stored command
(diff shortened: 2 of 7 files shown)


src/server/conn_context.cc
```diff
@@ -229,6 +229,22 @@ size_t ConnectionState::ExecInfo::UsedMemory() const {
   return dfly::HeapSize(body) + dfly::HeapSize(watched_keys);
 }
 
+void ConnectionState::ExecInfo::AddStoredCmd(const CommandId* cid, bool own_args, CmdArgList args) {
+  body.emplace_back(cid, own_args, args);
+  stored_cmd_bytes += body.back().UsedMemory();
+}
```
abhijat, line 235: this method can also return the delta of memory before adding the command vs after adding the command, to make updating the metric easier.

src/facade/dragonfly_connection.cc
```diff
@@ -2165,6 +2165,14 @@ bool Connection::IsReplySizeOverLimit() const {
   return over_limit;
 }
 
+void Connection::IncrStoredCmdBytes(const size_t bytes) {
```
romange: yes, it is technically possible to store in facade but it's not a good style because you store server related information there. facade does not know about server and you break this abstraction.

this is why we have `ServerState::Stats` structure for these things inside server library

src/server/conn_context.h
```diff
@@ -99,6 +110,10 @@ struct ConnectionState {
     // executing the multi transaction, which can create deadlocks by blocking other transactions
     // that already borrowed all available interpreters but wait for keys to be unlocked.
     Interpreter* preborrowed_interpreter = nullptr;
+
+    // The stored command size is updated as we add commands, so we do not have to iterate over the
+    // entire vector to find the current used memory.
+    size_t stored_cmd_bytes = 0;
```
romange, line 117: it's not clear why you need to track per connection. why not increase the thread-local stats directly

romange (commented):
i think this PR can be simplified

src/server/conn_context.h
abhijat: There are a couple of issues with using thread local stats and not per connection. 

Each time we add a stored command we have to add to the stat the amount of memory increased because of the new command, which is `size of command` + increase in capacity of vector due to `emplace_back` which can be much larger than one slot. 

The trivial way would be to check `UsedMemory` of `ExecInfo` before adding a stored command, then check `UsedMemory` after adding the command, and then add the diff to thread local stats. This is a problem because `ExecInfo::UsedMemory` will be run on each command twice and each time it has to examine the entire vector in linear time. 

One approach to avoid this could be to compute:

* vector capacity before emplace_back
* vector capacity after emplace_back
* add diff between 2 and 1 plus `StoredCmd::UsedMemory` to the thread local stats

The other issue is when decrementing the stat when clearing the vector. At that point if we are using thread local stats, the only way to compute how much to decrease for a given connection will be to use `ExecInfo::UsedMemory`  minus usage for watched keys, this might be okay because clear is done only after `exec` and not once per command.

It seemed simpler to me to avoid all this and add a per connection field.

src/server/conn_context.h
```diff
@@ -99,6 +110,10 @@ struct ConnectionState {
     // executing the multi transaction, which can create deadlocks by blocking other transactions
     // that already borrowed all available interpreters but wait for keys to be unlocked.
     Interpreter* preborrowed_interpreter = nullptr;
+
+    // The stored command size is updated as we add commands, so we do not have to iterate over the
+    // entire vector to find the current used memory.
```
romange:
```suggestion
    // The total size of all stored commands kept in "body". Does not include memory allocated by "body" vector.
```

src/server/main_service.cc
```diff
@@ -193,7 +193,7 @@ void MultiCleanup(ConnectionContext* cntx) {
     exec_info.preborrowed_interpreter = nullptr;
   }
   UnwatchAllKeys(cntx->ns, &exec_info);
-  exec_info.Clear();
+  exec_info.Clear(cntx->conn());
```
romange: you won't need to pass connection once you move stats into ServerState::Stats

src/facade/dragonfly_connection.cc
abhijat: moved to `ServerState::Stats`

src/server/main_service.cc
abhijat: yes removed now

src/server/conn_context.h
abhijat: changed

src/server/server_state.h
```diff
@@ -136,6 +136,9 @@ class ServerState {  // public struct - to allow initialization.
     uint32_t conn_timeout_events = 0;
     uint64_t psync_requests_total = 0;
     std::valarray<uint64_t> tx_width_freq_arr, squash_width_freq_arr;
+
+    // Memory size of stored commands during multi-exec in connections
+    uint64_t stored_cmd_bytes = 0;
```
romange: nit: use `size_t` for bytes and mem measurement.

src/server/conn_context.h
```diff
@@ -86,6 +86,16 @@ struct ConnectionState {
...
 
+    // Adds a StoredCmd and updates the stored_cmds_size
+    void AddStoredCmd(const CommandId* cid, bool own_args, CmdArgList args);
+
+    // Empties the body vector and resets stored_cmds_size to 0. Returns the size before data was
+    // cleared.
+    size_t ClearStoredCmds();
+
+    // Returns memory used by the body field without iterating over each stored command
+    size_t StoredCmdBytes() const;
```
romange: function name should be verbs (unless they are simple accessors).
So `GetStoredCmdBytes` is better.  Also nit - just make it inline here.

romange (dismissed):

src/server/server_state.h
abhijat: changed

src/server/conn_context.h
abhijat: changed

romange (approved):

PR #5765, opened 2025-09-03
chore: add more logs for takeover

BorysTheDev:
problem: we don't have enough logs to figure out the takeover failures.
fix: added more logs regarding JournalStreamer

### DIFF
--- a/src/server/dflycmd.cc
+++ b/src/server/dflycmd.cc
@@ -85,6 +85,7 @@ bool WaitReplicaFlowToCatchup(absl::Time end_time, const DflyCmd::ReplicaInfo* r
   shard->journal()->RecordEntry(0, journal::Op::PING, 0, 0, nullopt, {});
 
   const FlowInfo* flow = &replica->flows[shard->shard_id()];
+
   while (flow->last_acked_lsn < shard->journal()->GetLsn()) {
     if (absl::Now() > end_time) {
       LOG(WARNING) << "Couldn't synchronize with replica for takeover in time: " << replica->address
@@ -95,8 +96,9 @@ bool WaitReplicaFlowToCatchup(absl::Time end_time, const DflyCmd::ReplicaInfo* r
     if (!replica->exec_st.IsRunning()) {
       return false;
     }
-    VLOG(1) << "Replica lsn:" << flow->last_acked_lsn
-            << " master lsn:" << shard->journal()->GetLsn();
+    LOG_EVERY_T(INFO, 1) << "Replica lsn:" << flow->last_acked_lsn
+                         << " master lsn:" << shard->journal()->GetLsn()
+                         << "; Journal streamer state: " << flow->streamer->FormatInternalState();
     ThisFiber::SleepFor(1ms);
   }
 
@@ -512,7 +514,7 @@ void DflyCmd::TakeOver(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext
   atomic_bool catchup_success = true;
   if (*status == OpStatus::OK) {
     dfly::SharedLock lk{replica_ptr->shared_mu};
-    auto cb = [replica_ptr, end_time, &catchup_success](EngineShard* shard) {
+    auto cb = [replica_ptr = replica_ptr, end_time, &catchup_success](EngineShard* shard) {
       if (!WaitReplicaFlowToCatchup(end_time, replica_ptr.get(), shard)) {
         catchup_success.store(false);
       }
--- a/src/server/journal/streamer.cc
+++ b/src/server/journal/streamer.cc
@@ -164,6 +164,15 @@ size_t JournalStreamer::UsedBytes() const {
   return pending_buf_.Size();
 }
 
+std::string JournalStreamer::FormatInternalState() const {
+  return absl::StrCat(
+      "pending_buf_size:", pending_buf_.Size(), " in_flight_bytes:", in_flight_bytes_,
+      " total_sent:", total_sent_, " throttle_count:", throttle_count_,
+      " total_throttle_wait_usec:", total_throttle_wait_usec_,
+      " throttle_waiters:", throttle_waiters_, " last_async_write_time_ms:", last_async_write_time_,
+      " last_lsn_time_s:", last_lsn_time_, " last_lsn_writen_:", last_lsn_writen_);
+}
+
 void JournalStreamer::Write(std::string str) {
   DCHECK(!str.empty());
   DVLOG(3) << "Writing " << str.size() << " bytes";
--- a/src/server/journal/streamer.h
+++ b/src/server/journal/streamer.h
@@ -36,6 +36,9 @@ class JournalStreamer : public journal::JournalConsumerInterface {
 
   size_t UsedBytes() const;
 
+  // For debugging purposes. Return string with formatted internal state.
+  std::string FormatInternalState() const;
+
  protected:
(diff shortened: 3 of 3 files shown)


src/server/dflycmd.cc
```diff
@@ -512,7 +514,7 @@ void DflyCmd::TakeOver(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext
   atomic_bool catchup_success = true;
   if (*status == OpStatus::OK) {
     dfly::SharedLock lk{replica_ptr->shared_mu};
-    auto cb = [replica_ptr, end_time, &catchup_success](EngineShard* shard) {
+    auto cb = [replica_ptr = replica_ptr, end_time, &catchup_success](EngineShard* shard) {
```
romange, line 517: why this change?

BorysTheDev: it was a warning, so I decided to fix it

romange:
@kostasrim  we do pass the LSN from master to replica, correct? so if there is a mismatch we should see the mismatch errors in the log before the takeover, right?

kostasrim:
> we do pass the LSN from master to replica, correct

I am not sure I understand the question here. Once the node takes over, the new master does not have a journal active (although this is fixed on one of my pending PR's) but I don't see how this is relevant here. If you are asking if we LOG that we failed the takeover with LSN missmatch -- then yes we do. The way it works roughly is:

We initiate (the replica) a take over. 1. Master waits for any pending commands to finish executing, turns off expirations (so no journal writes come after that phase and the LSN's remain constant) and then 2. we wait for the `replica flows to catch up`. If we timeout, then it's because the replica did not catch up in time, aka, with an LSN missmatch. When this happens we LOG (and we saw these when the takeover failed): 

```
   88   while (flow->last_acked_lsn < shard->journal()->GetLsn()) {                                      
   89     if (absl::Now() > end_time) {                                                                  
   90       LOG(WARNING) << "Couldn't synchronize with replica for takeover in time: " << replica->address
   91                    << ":" << replica->listening_port << ", last acked: " << flow->last_acked_lsn   
   92                    << ", expecting " << shard->journal()->GetLsn();
 ```
 
 "We can say here that replica `synchronizes with` master before it becomes the new master."
 
 What is more, is that I don't understand how `extra logs` will help in this context. It's a `de facto` that we failed the takeover with an `LSN missmatch` so what exactly are we trying to confirm with these extra LOGs? That we have pending writes ? Of course we do, otherwise the replica would be in sync and the takeover would have been successful. 
 
 To my poor brain, the actual question is why we don't finish the takeover ? There are two candidates for this:
 
 1. We maybe expire or increment the LSN's `AFTER` the initial phase so we break the invariant of (1) above
 2. Something is off with our buffered writes and we prolong/stall the writes too much and the takeover timeouts

File: src/server/dflycmd.cc (hunk 1 above)
Comment (kostasrim) [reply src/server/dflycmd.cc:517]: I don't remember the exact warning but it should be because capturing structure bindings is `c++20` and is an extension (because we work with 17). I would personally accept this warning and it will go away naturally in a few years (I don't know a single mainstream compiler that doesn't support this). 

Anyway, it's fine that you changed this-- it doesn't really matter

**Comment by BorysTheDev** [comment]
@kostasrim With these logs, I want to understand the state of JournalStreamer, because we send LSN with it.

**Comment by kostasrim** [comment]
> @kostasrim With these logs, I want to understand the state of JournalStreamer, because we send LSN with it.

But I am not sure how this will help in this context?

```
      "pending_buf_size:", pending_buf_.Size(), " in_flight_bytes:", in_flight_bytes_,
      " total_sent:", total_sent_, " throttle_count:", throttle_count_,
      " total_throttle_wait_usec:", total_throttle_wait_usec_,
      " throttle_waiters:", throttle_waiters_, " last_async_write_time_ms:", last_async_write_time_,
      " last_lsn_time_s:", last_lsn_time_, " last_lsn_writen_:", last_lsn_writen_);
```

`Pending_buf_size` will be > 0, `in_flight_bytes probably > 0 as well`,  and you know there are pending writes because we `timedout with an LSN missmatch`. 

What I am trying to understand here is even if we LOG those how is that going to help and isolate the area of search/bug ? In other words, if for example we violate the invariant (1) I described above -- will those LOG help us realize that ? If, we have pending_buf_size > 0 (which we do), how will this point to the bug that we don't drain the buffered writes in time ? I could be wrong but I don't see how. Any insights on what am I missing here?

**Comment by BorysTheDev** [comment]
@kostasrim the logs publish every second so seeing the difference or missing difference we can understand where is the problem

**Comment by romange** [comment]
> > we do pass the LSN from master to replica, correct
> 
> I am not sure I understand the question here. Once the node takes over, the new master does not have a journal active (although this is fixed on one of my pending PR's) but I don't see how this is relevant here. If you are asking if we LOG that we failed the takeover with LSN missmatch -- then yes we do. The way it works roughly is:
> 
> We initiate (the replica) a take over. 1. Master waits for any pending commands to finish executing, turns off expirations (so no journal writes come after that phase and the LSN's remain constant) and then 2. we wait for the `replica flows to catch up`. If we timeout, then it's because the replica did not catch up in time, aka, with an LSN missmatch. When this happens we LOG (and we saw these when the takeover failed):
> 
> ```
>    88   while (flow->last_acked_lsn < shard->journal()->GetLsn()) {                                      
>    89     if (absl::Now() > end_time) {                                                                  
>    90       LOG(WARNING) << "Couldn't synchronize with replica for takeover in time: " << replica->address
>    91                    << ":" << replica->listening_port << ", last acked: " << flow->last_acked_lsn   
>    92                    << ", expecting " << shard->journal()->GetLsn();
> ```
> 
> "We can say here that replica `synchronizes with` master before it becomes the new master."
> 
> What is more, is that I don't understand how `extra logs` will help in this context. It's a `de facto` that we failed the takeover with an `LSN missmatch` so what exactly are we trying to confirm with these extra LOGs? That we have pending writes ? Of course we do, otherwise the replica would be in sync and the takeover would have been successful.
> 
> To my poor brain, the actual question is why we don't finish the takeover ? There are two candidates for this:
> 
> 1. We maybe expire or increment the LSN's `AFTER` the initial phase so we break the invariant of (1) above
> 2. Something is off with our buffered writes and we prolong/stall the writes too much and the takeover timeouts

My question is that whether we recognize and catch the LSN mismatch state on replica side (unrelated to takeover). My question is not about the takeover.

**Comment by kostasrim** [comment]
> My question is that whether we recognize and catch the LSN mismatch state on replica side (unrelated to takeover). My question is not about the takeover.

Uh you mean this:

```
  103   // When LSN opcode is sent master does not increase journal lsn.                                 
  104   if (lsn_.has_value() && res->opcode != journal::Op::LSN) {                                       
  105     ++*lsn_;                                                                                       
  106     VLOG(2) << "read lsn: " << *lsn_;                                                              
  107   }                                                                                                
  108                                                                                                    
  109   TransactionData tx_data = TransactionData::FromEntry(std::move(res.value()));                    
  110   if (lsn_.has_value() && tx_data.opcode == journal::Op::LSN) {                                    
  111     DCHECK_NE(tx_data.lsn, 0u);                                                                    
>>112     LOG_IF_EVERY_N(WARNING, tx_data.lsn != *lsn_, 10000)                                           
  113         << "master lsn:" << tx_data.lsn << " replica lsn" << *lsn_;                                
  114     DCHECK_EQ(tx_data.lsn, *lsn_);                                                                 
  115   } 
```

Yes if there is a missmatch we print

PR #5766, opened 2025-09-03
chore: implement the iterative fragmentation check

romange:
Before - we had relatively slow check for counting wasted fragmentation via zmalloc_get_allocator_wasted_blocks that took 10ms or more in production. The reason for that is that it iterate over all the memory pages on a single shard through the single call.

Now we implement an iterative version of it by iterating over a single page queue data-structure in the heap. Once we start the iterative process we will continue aggregating stats over all the page queues in the heap until we reach the end and then conclude if defragmentation is needed.

this should reduce the call time to EngineShard::DefragTaskState::CheckRequired by x70 (number of page queues in the heap).

### DIFF
--- a/src/redis/zmalloc_mi.c
+++ b/src/redis/zmalloc_mi.c
@@ -4,6 +4,8 @@
 
 #include <assert.h>
 #include <mimalloc.h>
+
+#define MI_BUILD_RELEASE 1
 #include <mimalloc/types.h>
 #include <string.h>
 #include <unistd.h>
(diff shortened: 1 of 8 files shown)


vyavdoshenko:
Correct me if I am wrong, but there is a possibility for alloc/dealloc between steps. If yes, this change is not perfect because this issue should be handled somehow.

romange:
why it should be handled? :)

romange:
in other words what is the scenario that you think may happen?

romange:
Let me explain: for each page queue the operation is still atomic and consistent. It's true that we sample different page queues at  different times so the operation itself may sample different states but it does not matter as it's only used to aggregate metrics that may trigger background defragmentation process. In any case, each step is safe for invocation and eventually consistent metrics will still be quite precise in most cases.

Of course, there may be a case where the trigger condition will provide false positive or false negative result but it does not really matter - we retry the check in 10sec or end up running defragmentation process in the worst case.

vyavdoshenko:
@romange 
Thanks for the explanation. I got the answer.

src/server/engine_shard.cc
```diff
@@ -249,26 +231,41 @@ bool EngineShard::DefragTaskState::CheckRequired() {
     return false;
   }
 
-  const std::size_t global_threshold = limit * GetFlag(FLAGS_mem_defrag_threshold);
+  static thread_local fragmentation_info finfo{.committed = 0, .wasted = 0, .bin = 0};
+
+  const std::size_t global_threshold = double(limit) * GetFlag(FLAGS_mem_defrag_threshold);
   if (global_threshold > rss_mem_current.load(memory_order_relaxed)) {
+    finfo.bin = 0;  // reset.
```
abhijat, line 238: This is more of a question about behavior but is there any other place where this queue id should reset to 0? 

For example if the current memory usage remains over the threshold and previous iteration over queues was complete, so `bin == MI_BIN_FULL+1`.  

Then `zmalloc_get_allocator_fragmentation_step` keeps early returning 0. This might cause the `EngineShard::DefragTaskState::CheckRequired` to return true, and defrag will run, but the `bin` will not reset.

abhijat: Actually it looks like correct behavior: it runs defrag without checking individual pages if memory usage is greater than threshold and last check was complete.

src/redis/zmalloc_mi.c
```diff
@@ -167,10 +167,44 @@ int zmalloc_get_allocator_wasted_blocks(float ratio, size_t* allocated, size_t*
...
+    size_t committed = page->capacity * bsize;
+
+    info->committed += committed;
+    if (used < committed) {
+      size_t threshold = (double)committed * ratio;
+      if (used < threshold) {
+        info->wasted += (committed - used);
+      }
+    }
+    page = next;  // and continue
```
abhijat: nit: could be just `page = page.next;`

src/server/engine_shard.cc
```diff
@@ -249,26 +231,41 @@ bool EngineShard::DefragTaskState::CheckRequired() {
     return false;
   }
 
-  const std::size_t global_threshold = limit * GetFlag(FLAGS_mem_defrag_threshold);
+  static thread_local fragmentation_info finfo{.committed = 0, .wasted = 0, .bin = 0};
```
abhijat, line 234: nit: `static` is redundant here

abhijat (approved):
couple of minor nits but lgtm 

`TestDefragOption` needs adjustment though

romange:
1. the PR has only partial improvement. this is due to the fact that majority of the time is spent in a single bin `MI_BIN_FULL`
2. I won't fix dragonfly_test and instead will disable it - I do not have time unfortunately. So someone will have to help me to write a defragmentation test 

My plan for `MI_BIN_FULL` is to do an efficient accounting of all the page sizes in MI_BIN_FULL at heap level. This will require an additional patching of mimalloc lib.

romange:
This is the PR that should allow tracking of full page sizes.
https://github.com/romange/mimalloc/pull/1/files

abhijat (approved):

Issue #5705, opened 2025-08-21
Add metrics for actions taken by defragment process


abhijat:
When defragment process runs, it should report on whatever statistics it collects into grafana.

abhijat:
After going through the code I can see two things:

1. We don't collect detailed stats on the non-interactive background defrag task. This was by design not to make the bg task slower. So we do not have detailed stats unless the run is interactive.
2. We already have stats from the bg task but they are not exposed in grafana, just in the info command.

The three existing stats can be exposed in grafana as counters. The other thing which can be done is to collect stats even for the background task. The following may be added to prometheus metrics right now:

* defrag_attempt_total
* defrag_realloc_total
* defrag_task_invocation_total

PR #5767, opened 2025-09-04
server: Adds metrics for background defrag task

abhijat:
Note: metrics are not added for the manual execution of the task in this PR

partially fixes https://github.com/dragonflydb/dragonfly/issues/5705


abhijat:
as discussed offline just adding the basic metrics for now

PR #5768, opened 2025-09-04
facade/server: Use inline instead of extern to reduce loc

abhijat:
Since c++17 `inline` variables can be used for a _definition_ of a variable in a header file which is included in multiple source files. This avoids the need to use `extern` in the header and defining the variable in a source file.

The linker will merge the symbols from different translation units at link time.

https://en.cppreference.com/w/cpp/language/inline.html

### DIFF
--- a/src/facade/facade_types.h
+++ b/src/facade/facade_types.h
@@ -201,15 +201,15 @@ struct ErrorReply {
   std::optional<OpStatus> status{std::nullopt};
 };
 
-constexpr inline unsigned long long operator""_MB(unsigned long long x) {
+constexpr unsigned long long operator""_MB(unsigned long long x) {
   return 1024L * 1024L * x;
 }
 
-constexpr inline unsigned long long operator""_KB(unsigned long long x) {
+constexpr unsigned long long operator""_KB(unsigned long long x) {
   return 1024L * x;
 }
 
-extern __thread FacadeStats* tl_facade_stats;
+inline thread_local FacadeStats* tl_facade_stats = nullptr;
 
 void ResetStats();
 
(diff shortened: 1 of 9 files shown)


src/facade/facade_types.h
```diff
@@ -201,15 +201,15 @@ struct ErrorReply {
   std::optional<OpStatus> status{std::nullopt};
 };
 
-constexpr inline unsigned long long operator""_MB(unsigned long long x) {
+constexpr unsigned long long operator""_MB(unsigned long long x) {
```
romange, line 204: what makes this inline? in other words why there are no multiple definitions of this operator?

romange (approved):

src/facade/facade_types.h
BorysTheDev: It is constexpr and we apply it only to consts so it doesn't need to be linked

BorysTheDev:
I will merge it because I depend on these changes

src/facade/facade_types.h
kostasrim:
```
([dcl.constexpr], §7.1.5/2 in the C++11 standard): "constexpr functions and constexpr constructors are implicitly inline (7.1.2)."
```

> It is constexpr and we apply it only to consts

That's not true. A `constexpr` function can be evaluated `at runtime or at compile time` depending on the `context`. (the context here is the function argument)

It would have been true if we used `consteval` which enforces what `constexpr` doesn't.

`constexpr` in this context `implies inline` and `inline functions` do not violate the ODR of C++ regardless of if they are evaluated at runtime or not.

BorysTheDev: @kostasrim  you didn't get my answer, we use _MB only like 10_MB so  10 is const literal, so we nevel called it with anything else

PR #5769, opened 2025-09-04
fix: acl pubsub should only glob match the channel

### LINKED ISSUE #5763: channel perms in acl for publish not working
**Describe the bug**

According to docs `&<pattern>` should be supported in acl but when I try it is not working. The only pattern that allows to publish is `&*`. All other key perms are working fine with patterns it's just publish that's not working.

Ref: [Docs](https://www.dragonflydb.io/docs/managing-dragonfly/acl#pubsub-permissions)

**To Reproduce**

Permission set :
```
user demo on resetkeys resetchannels ~app|managed-resources|* &app|managed-resources|* ~app|resources-tree|* ~cluster|info|* +set +get +del +publish +ping >passwd
```

ACL GETUSER output :
```
127.0.0.1:6379> acl getuser demo
 1) flags
 2) 1) on
 3) passwords
 4) passwd
 5) commands
 6) -@all +set +get +del +publish +ping
 7) keys
 8) ~app|managed-resources|* ~app|resources-tree|* ~cluster|info|*
 9) channels
10) resetchannels &app|managed-resources|*
```

Error:
```
127.0.0.1:6379> publish app|managed-resources|xyz test
(error) NOPERM demo has no ACL permissions
```

Works when I change to `&*`
```
127.0.0.1:6379> publish app|managed-resources|xyz test
(integer) 0
```

**Expected behavior**
I expect the user to be able to publish to the channel with prefix `app|managed-resources|....`, but I can't

**Screenshots**

 - Containerized: K8s
 - Dragonfly Version: 1.33.1

**Reproducible Code Snippet**
```
```

kostasrim:
publish command accepts only a single channel and a message, yet in `IsUserAllowedToInvokeCmd` we glob matched both the channel and the message which effectively rejected incorrectly the command execution because of permissions.

Fixes  #5763

### DIFF
--- a/src/server/acl/validator.cc
+++ b/src/server/acl/validator.cc
@@ -50,9 +50,8 @@ bool ValidateCommand(const std::vector<uint64_t>& acl_commands, const CommandId&
 
   bool allowed = true;
   if (!pub_sub.all_channels) {
-    for (auto channel : tail_args) {
-      allowed &= iterate_globs(facade::ToSV(channel));
-    }
+    auto channel = tail_args[0];
+    allowed &= iterate_globs(facade::ToSV(channel));
   }
 
   return {allowed, AclLog::Reason::PUB_SUB};
--- a/src/server/acl/acl_family_test.cc
+++ b/src/server/acl/acl_family_test.cc
@@ -538,6 +538,15 @@ TEST_F(AclFamilyTest, TestPubSub) {
   vec = resp.GetVec();
   EXPECT_THAT(vec[8], "channels");
   EXPECT_THAT(vec[9], "resetchannels &foo");
+
+  resp =
+      Run("ACL setuser demo on resetkeys resetchannels ~app|managed-resources|* "
+          "&app|managed-resources|* +publish +ping >passwd");
+  resp = Run("AUTH demo passwd");
+  EXPECT_THAT(resp, "OK");
+
+  resp = Run("publish app|managed-resources|xyz test");
+  EXPECT_THAT(resp, IntArg(0));
 }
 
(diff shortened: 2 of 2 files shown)


src/server/acl/validator.cc
```diff
@@ -50,9 +50,8 @@ bool ValidateCommand(const std::vector<uint64_t>& acl_commands, const CommandId&
 
   bool allowed = true;
   if (!pub_sub.all_channels) {
-    for (auto channel : tail_args) {
-      allowed &= iterate_globs(facade::ToSV(channel));
-    }
+    auto channel = tail_args[0];
```
kostasrim: Safe, when we reach here we already verified that the command is in a good state (meaning it has exactly two arguments)

vyavdoshenko (dismissed):
lgtm

kostasrim:
Uh there is psubscribe that goes under the same flow. Will address it soon

src/server/acl/validator.cc
```diff
@@ -50,8 +50,14 @@ bool ValidateCommand(const std::vector<uint64_t>& acl_commands, const CommandId&
 
   bool allowed = true;
   if (!pub_sub.all_channels) {
-    for (auto channel : tail_args) {
+    std::string_view name = id.name();
+    if (name == "PUBLISH" || name == "SPUBLISH") {
```
vyavdoshenko, line 54: Should be here processed SUBSCRIBE, SSUBSCRIBE?

kostasrim: these are processed differently as is

PR #5770, opened 2025-09-04
feat: prometheus metrics per database

qlimenoque:
### What this PR does
This PR adds per‑database **hit/miss/evicted/expiring/expired** metrics

### Before
Only global keyspace hit/miss counters, which makes it hard to see which DB was underperforming when one instance served different logical data

### Now
We track hits/misses per DB, show them in INFO (with hit_ratio), and export per‑DB Prometheus series via db labels on the existing keyspace_* metrics.

Now can run different domains (e.g., app vs. metrics) in the same instance and get clean, separate observability and alerting for each DB

### PromQL examples

```
// Hits per db
sum by (db) (rate(dragonfly_keyspace_hits_total[5m]))

// Misses per db
sum by (db) (rate(dragonfly_keyspace_misses_total[5m]))


// Ratio per db
sum by (db) (rate(dragonfly_keyspace_hits_total{db!=""}[5m]))
/
(sum by (db) (rate(dragonfly_keyspace_hits_total{db!=""}[5m])) + sum by (db) (rate(dragonfly_keyspace_misses_total{db!=""}[5m])))
> 0
```


src/server/db_slice.cc
```diff
@@ -575,6 +575,9 @@ auto DbSlice::FindInternal(const Context& cntx, string_view key, optional<unsign
 
   if (!IsValid(res.it)) {
     events_.misses += miss_weight;
+    if (miss_weight) {
+      db.stats.misses++;
+    }
```
romange:
```suggestion
    db.stats.misses += miss_weight;
```
and below

qlimenoque: Thanks, done ✅

romange:
Thanks for contributing this!

romange:
quick question, did you use copilot/AI assistant to make this change or you did it yourself?

qlimenoque:
@romange this was written with Cursor, since I’m not very experienced with C++, but the change seemed straightforward given that Prometheus metrics were already implemented

I built and tested the image locally, and also ran it on our staging cluster with Prometheus successfully scraping Dragonfly. Everything worked as expected, so I thought it would be a good contribution :)

romange:
it looks real but it's not. you probably did not know how run the tests that you wrote with cursor :)

[1.patch](https://github.com/user-attachments/files/22284264/1.patch)
please apply  this patch to fix the build and tests

qlimenoque:
@romange thank you for picking up on that and adding the correct changes! Patch applied

romange:
Please also update `helio` submodule. There should be no differences between the "main" branch and this branch with helio.

qlimenoque:
@romange done ✅

romange (approved):
LGTM. Thank you for contributing this

PR #5772, opened 2025-09-04
chore: add self-testing code for heap->full_page_size computation

romange:
Following #5766 I am adding self testing code just to make sure that the patch to mimalloc is correct. We compute full page size like we did before and compare to the fast method. If discrepancy is found we output the error.


vyavdoshenko (approved):
lgtm

PR #5774, opened 2025-09-04
chore: new ReplicaOf algorithm

kostasrim:
This PR implements the new ReplicaOf algorithm that does not use two phase locking.

**How ReplicaOf worked previously**:

1. stop the old replica -> dragonfly is processing read only commands normally
2. Set the flag to loading -> dragonfly stops processing read only commands because state is loading
3. create a new replica object 
4. release the mutex
5. allow the new replica object to do IO by connecting to the new master and exchange the necessary info to actually start the replication flow. -> This may fail yet we prematurely set the flag to loading and stopped the old replica
6. relock the mutex after the greet, clean up if the context got cancelled (by another replicaof command) or start the main replication fiber to properly start replication.

There are multiple issue with this:

1. two phase locking, allows other replicaof commands to interleave (not serialized) which complicates state transitions
2. prematurely setting state to loading even if the connection fails later on after releasing the mutex
3. prematurely stopped replication even if the connection with the new master fails later on
4. prematurely stops processing read only commands. Masters might be rotating so we might enter partial sync immediately. We don't need to enter LOADING state for that.

**How it works now:**

The new algorithm is very simple and has only two steps:

1. Create a `Replica` object, Initiate connection to the new master, greet and exchange the necessary info to setup replication. (internally it’s just REPLCONF setting connection members and info). So far there is no update to the state. We merely check if we can establish a connection with the new master and that everything is ok.

2. If there are no errors, lock the mutex, update the replica_ object and start the MainReplicationFiber. Do not enter LOADING state prematurely. First check if partial sync is available in main replication fiber and if not, `then enter LOADING state` and do full sync.

Now all `ReplicaOfInternal` commands are serialized. Cancellation is not affected because the connection that enters the critical section of (b) will cancel/stop the previous one. This happens one after the other in a deterministic manner.

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -3555,7 +3555,7 @@ void ServerFamily::StopAllClusterReplicas() {
 }
 
 void ServerFamily::ReplicaOf(CmdArgList args, const CommandContext& cmd_cntx) {
-  ReplicaOfInternal(args, cmd_cntx.tx, cmd_cntx.rb, ActionOnConnectionFail::kReturnOnError);
+  ReplicaOfInternalV2(args, cmd_cntx.tx, cmd_cntx.rb, ActionOnConnectionFail::kReturnOnError);
 }
 
 void ServerFamily::Replicate(string_view host, string_view port) {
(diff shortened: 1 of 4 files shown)


src/server/server_family.cc
```diff
@@ -3555,7 +3555,7 @@ void ServerFamily::StopAllClusterReplicas() {
 }
 
 void ServerFamily::ReplicaOf(CmdArgList args, const CommandContext& cmd_cntx) {
-  ReplicaOfInternal(args, cmd_cntx.tx, cmd_cntx.rb, ActionOnConnectionFail::kReturnOnError);
```
romange, line 3558: It seems that changes are grouped together nicely. Let's introduce the flag `experimental_replicaof_v2` set to true so we could have the fallback to the old variant

src/server/server_family.cc
```diff
@@ -3568,7 +3568,101 @@ void ServerFamily::Replicate(string_view host, string_view port) {
...
+  // if that mutex is locked below before any state check we can't really enforce
+  // that the old replication fiber won't try to full sync and update the state to LOADING.
+  // What is more here is that we always call `replica->Stop()`. So even if we end up in the
+  // scenario described, the semantics are well defined. First, cancel the old replica and
+  // move on with the new one. Cancelation will be slower and ReplicaOf() will
+  // induce higher latency -- but that's ok because it's an highly improbable flow with
+  // well defined semantics.
+  ServerState* ss = ServerState::tlocal();
+
+  if (ss->is_master && ss->gstate() == GlobalState::LOADING) {
```
romange, line 3628: we can not call "replica no one" during the full sync?

src/server/server_family.cc
```diff
@@ -3568,7 +3568,101 @@ void ServerFamily::Replicate(string_view host, string_view port) {
...
+  }
+
+  // replicaof no one
+  if (replicaof_args->IsReplicaOfNoOne()) {
+    return ReplicaOfNoOne(builder);
+  }
+
+  auto new_replica = make_shared<Replica>(replicaof_args->host, replicaof_args->port, &service_,
+                                          master_replid(), replicaof_args->slot_range);
+  GenericError ec{};
```
romange: nit:
```suggestion
  GenericError ec;
```

src/server/server_family.cc
```diff
@@ -3568,7 +3568,101 @@ void ServerFamily::Replicate(string_view host, string_view port) {
...
+  switch (on_error) {
+    case ActionOnConnectionFail::kReturnOnError:
+      ec = new_replica->Start();
```
romange, line 3643: when `Start() ` returns? before the full sync? after the connect?

src/server/server_family.cc
```diff
@@ -3568,7 +3568,101 @@ void ServerFamily::Replicate(string_view host, string_view port) {
   CmdArgList args_list = absl::MakeSpan(args_vec);
   io::NullSink sink;
   facade::RedisReplyBuilder rb(&sink);
-  ReplicaOfInternal(args_list, nullptr, &rb, ActionOnConnectionFail::kContinueReplication);
+  ReplicaOfInternalV2(args_list, nullptr, &rb, ActionOnConnectionFail::kContinueReplication);
+}
+
+void ServerFamily::ReplicaOfNoOne(SinkReplyBuilder* builder) {
```
romange, line 3585: should you mark it with `ABSL_LOCKS...` macro ?

src/server/server_family.cc
```diff
@@ -3568,7 +3568,101 @@ void ServerFamily::Replicate(string_view host, string_view port) {
...
+  util::fb2::LockGuard lk(replicaof_mu_);
+  ServerState* ss = ServerState::tlocal();
+
+  if (!ss->is_master) {
+    CHECK(replica_);
+    // flip flag before clearing replica_
```
romange, line 3592: style: please keep an empty line before comment lines. here and everywhere.

romange:
What's the functional change now? how it is changed against the previous version?

tests/dragonfly/replication_test.py
```diff
@@ -469,12 +466,8 @@ async def ping_status():
             await asyncio.sleep(0.05)
 
     async def replicate():
-        try:
-            await c_replica.execute_command(f"REPLICAOF localhost {master.port}")
-            return True
-        except redis.exceptions.ResponseError as e:
-            assert e.args[0] == "replication cancelled"
-            return False
+        await c_replica.execute_command(f"REPLICAOF localhost {master.port}")
```
romange, line 469: can you explain please how the changes affected the test behavior?

src/server/server_family.cc
kostasrim: That was `true` before my changes as well. See from the `old ReplicaOf`:

```
  {
    util::fb2::LockGuard lk(replicaof_mu_);  // Only one REPLICAOF command can run at a time
    // We should not execute replica of command while loading from snapshot.
    if (ss->is_master && ss->gstate() == GlobalState::LOADING) {
      builder->SendError(kLoadingErr);
      return;
    }
    
     auto replicaof_args = ReplicaOfArgs::FromCmdArgs(args, builder);
    if (!replicaof_args.has_value()) {
      return;
    }
    LOG(INFO) << "Replicating " << *replicaof_args;
    // If NO ONE was supplied, just stop the current replica (if it exists)
    if (replicaof_args->IsReplicaOfNoOne()) {
```

The lock would be unlocked, the state loading and the replicaof no one would have been rejected.

src/server/server_family.cc
kostasrim: Yes, we don't enter full sync yet. `Start` merely exchanges metadata around replication (connects to master, sets number of flows via REPLCONF etc). There is no actual sync here just metadata exchange between two nodes

tests/dragonfly/replication_test.py
kostasrim: Uh, yes I did explain this https://github.com/dragonflydb/dragonfly/pull/5758#discussion_r2316531073 but on the POC pr. I wrote:


The new algorithm does not use two phase locking so the following is no longer a possibility:

client 1 -> REPLICAOF -> Locks the mutex -> updates replica_ to new_replica -> releases the mutex -> calls replica_->Start()
client 2 -> REPLICAOF -> same as (1) but first calls replica_->Stop() -> releases the mutex
client 1 -> REPLICAOF command grabs the second lock to the mutex, observes that the context got cancelled because of step (2) and boom returns "replication cancelled"
This can not happen anymore because we lock only once and atomically update everything including stopping the previous replica. So by the time (2) grabs the lock in the example above, the previous REPLICAOF command is not in some intermediate state. To observe that indeed we cancelled, we should read the logs and see ("Stopping replication") COMMANDS_TO_ISSUE - 1 times + 1 (because of the Shutdown() at the end)

Bonus points is that I suspect we might be able to also solve https://github.com/dragonflydb/dragonfly/pull/4685 but I will need to follow up with that

kostasrim: Everything is linearized now, cancellation is incremental (the next ReplicaOf cancels the previous one and so on)

romange (approved):

PR #5775, opened 2025-09-04
feat: Evicition based on rss threshold

mkaruza:
Following PR after initial work from Stefan in #4991

* Removed de-fragmentation changes from #4991.
* Clear internal eviction state if eviction is not done in consecutive steps
* Added python test. Initial populated data doesn't trigger any eviction threshold (used_memory or rss_memory). After we execute commands that increase RSS memory. RSS threshold is reached and eviction is done. Used memory and RSS memory should be lower at the end.

### DIFF
--- a/tests/dragonfly/memory_test.py
+++ b/tests/dragonfly/memory_test.py
@@ -7,15 +7,6 @@ from . import dfly_args
 from .instance import DflyInstance, DflyInstanceFactory
 
 
-def extract_fragmentation_waste(memory_arena):
-    """
-    Extracts the fragmentation waste from the memory arena info.
-    """
-    match = re.search(r"fragmentation waste:\s*([0-9.]+)%", memory_arena)
-    assert match.group(1) is not None
-    return float(match.group(1))
-
-
 @pytest.mark.slow
 @pytest.mark.opt_only
 @pytest.mark.parametrize(
@@ -195,7 +186,8 @@ async def test_eval_with_oom(df_factory: DflyInstanceFactory):
     assert rss_before_eval * 1.01 > info["used_memory_rss"]
 
 
-async def test_eviction_on_rss_treshold(df_factory: DflyInstanceFactory):
+@pytest.mark.parametrize("heartbeat_rss_eviction", [True, False])
+async def test_eviction_on_rss_treshold(df_factory: DflyInstanceFactory, heartbeat_rss_eviction):
     max_memory = 1024 * 1024**2  # 10242mb
 
     df_server = df_factory.create(
@@ -203,6 +195,7 @@ async def test_eviction_on_rss_treshold(df_factory: DflyInstanceFactory):
         cache_mode="yes",
         maxmemory=max_memory,
         enable_heartbeat_eviction="false",
+        enable_heartbeat_rss_eviction=heartbeat_rss_eviction,
     )
     df_server.start()
     client = df_server.client()
@@ -240,9 +233,14 @@ async def test_eviction_on_rss_treshold(df_factory: DflyInstanceFactory):
     memory_info_after = await client.info("memory")
     stats_info_after = await client.info("stats")
 
-    # We should see used memory deacrease and number of some number of evicted keys
-    assert memory_info_after["used_memory"] < memory_info_before["used_memory"]
-    assert stats_info_after["evicted_keys"]
+    if heartbeat_rss_eviction:
+        # We should see used memory deacrease and number of some number of evicted keys
+        assert memory_info_after["used_memory"] < memory_info_before["used_memory"]
+        assert stats_info_after["evicted_keys"]
+    else:
+        # If heartbeat rss eviction is disabled there should be no chage
+        assert memory_info_after["used_memory"] == memory_info_before["used_memory"]
+        assert stats_info_after["evicted_keys"] == 0
 
 
 @pytest.mark.asyncio


src/server/engine_shard.cc
```diff
@@ -195,25 +193,7 @@ bool EngineShard::DefragTaskState::CheckRequired() {
...
-
-  const double mem_defrag_threshold_flag = GetFlag(FLAGS_mem_defrag_threshold);
-  const double defrag_threshold =
-      !is_eviction_enabled ? mem_defrag_threshold_flag
-                           : std::min(mem_defrag_threshold_flag,
-                                      ServerState::tlocal()->rss_oom_deny_ratio -
-                                          GetFlag(FLAGS_eviction_memory_budget_threshold) -
-                                          EvictionTaskState::kDefragRssMemoryDelta);
-  const std::size_t global_threshold = double(limit) * defrag_threshold;
+  const std::size_t global_threshold = double(limit) * GetFlag(FLAGS_mem_defrag_threshold);
```
BagritsevichStepan, line 198: Why did you remove the defragmentation changes? Did you pass python tests that I added after removing them?

src/server/engine_shard.cc
```diff
@@ -683,9 +655,19 @@ void EngineShard::RetireExpiredAndEvict() {
...
+  if (heartbeat_counter_ != eviction_state_.last_eviction_heartbeat_counter + 1) {
+    eviction_state_.global_rss_memory_at_prev_eviction = 0;
+    eviction_state_.deleted_bytes_before_rss_update = 0;
+  }
+
   size_t deleted_bytes = 0;
   size_t eviction_goal = GetFlag(FLAGS_enable_heartbeat_eviction) ? CalculateEvictionBytes() : 0;
 
+  if (eviction_goal) {
+    eviction_state_.last_eviction_heartbeat_counter = heartbeat_counter_;
```
BagritsevichStepan: if (GetFlag(FLAGS_enable_heartbeat_eviction))

src/server/engine_shard.cc
```diff
@@ -683,9 +655,19 @@ void EngineShard::RetireExpiredAndEvict() {
   DbContext db_cntx;
   db_cntx.time_now_ms = GetCurrentTimeMs();
 
+  // If we don't have two consecutive evictions we should reset state
+  if (heartbeat_counter_ != eviction_state_.last_eviction_heartbeat_counter + 1) {
+    eviction_state_.global_rss_memory_at_prev_eviction = 0;
```
BagritsevichStepan: Please add Reset method to EvictionState class

src/server/engine_shard.cc
mkaruza: This flag is checked line before. If this flag is not enabled eviction goal will always be `0`.

src/server/engine_shard.cc
mkaruza: Will add.

src/server/engine_shard.cc
mkaruza: I think that idea is to check how we handle eviction by monitoring `rss` memory. We can add aggresive de fragmentation if this is needed on follow PR.

src/server/engine_shard.cc
BagritsevichStepan: you still have a case here, when `FLAGS_enable_heartbeat_eviction` returns true, but `CalculateEvictionBytes` returns 0. And you will reset the `eviction_state_` each time

mkaruza: It is expected, if evicition_goal is 0 next heartbeat should reset evicition_state. Is this your comment about?

romange:
@mkaruza  this PR is not clear to me. are you merging into Stefan's branch? Do you need me to review it?

mkaruza:
@romange you can take a look of course, this PR adds extra test and handles reset of stale eviction state  from Stefan PR. Once it is merged we can continue with this PR (if needed).

src/server/engine_shard.cc
BagritsevichStepan: I expect that it should not work. The issue is that an eviction goal of 0 does not necessarily mean we should reset the state. The eviction goal may be 0 while the reported used_rss_memory is still high (since it is not updated at all) and deleted_bytes_before_rss_update is also high (because we have evicted a lot of items). In this case, the actual calculated RSS memory usage (`used_rss_memory - deleted_bytes_before_rss_update`) will be below the RSS threshold, resulting in an eviction goal of 0. If we reset the state at this point (on the next heartbeat), the eviction goal will rise again, leading to over-eviction of items. This change makes deleted_bytes_before_rss_update useless.

Let's decide why do we need heartbeat counter. What **real** scenario we want to prevent and create a **test** which reproduces this scenario.

src/server/engine_shard.cc
BagritsevichStepan: Do we have a baseline for this PR? — which tests must pass and which are not required. Should the tests I added be expected to pass, or only yours?

src/server/engine_shard.cc
romange: the scenario is that `eviction_state_` can be left with arbitrary values and eviction can stop running due to sudden drop of RSS and then, after some arbitrary time it can resume and seed from this arbitrary state again. (The impact of seeding from random `eviction_state_` is not clear to me. )

tests/dragonfly/memory_test.py
```diff
@@ -268,11 +318,6 @@ async def test_cache_eviction_with_rss_deny_oom_simple_case(
     # Wait for some time
     await asyncio.sleep(3)
 
-    memory_arena = await async_client.execute_command("MEMORY", "ARENA")
-    fragmentation_waste = extract_fragmentation_waste(memory_arena)
```
BagritsevichStepan, line 272: We can also remove `extract_fragmentation_waste` method now

tests/dragonfly/memory_test.py
```diff
@@ -268,11 +318,6 @@ async def test_cache_eviction_with_rss_deny_oom_simple_case(
...
-    logging.info(f"Memory fragmentation waste: {fragmentation_waste}")
-    assert fragmentation_waste < 12.0, "Memory fragmentation waste should be less than 12%."
-
     # Assert that no more keys are evicted
     memory_info = await async_client.info("memory")
```
BagritsevichStepan, line 320: Did you run the tests? Are we passing them?

src/server/engine_shard.cc
```diff
@@ -408,6 +381,7 @@ void EngineShard::StartPeriodicHeartbeatFiber(util::ProactorBase* pb) {
   }
   auto heartbeat = [this]() { Heartbeat(); };
 
+  eviction_state_.rss_eviction_enabled_ = GetFlag(FLAGS_enable_heartbeat_rss_eviction);
```
BagritsevichStepan, line 384: Let's add a small test (or maybe update one of the current tests) that proves that eviction based on rss will be disabled (if this flag is not set).

tests/dragonfly/memory_test.py
mkaruza: Yes

Issue #5762, opened 2025-09-02
Dragonfly not working with convoy


ikibko3c:
**Describe the bug**
When we use convoy(http://getconvoy.io) with regular redis(bitnami/redis:7.0.11-debian-11-r12 in our case), all seems fine
But when we switch to dragonfly - we get some errors and convoy not working

**To Reproduce**
1. Install convoy using official helm chart, setup external redis as dragonfly
2. Setup any project in webui to get ingest url, and some output location
3. Send POST request to this url with any test content and get error in convoy server logs, curl responce with 400
```{"error":"asynq: UNKNOWN: ERR NOT FOUND","level":"error","msg":"Error occurred sending new event to the queue","source":"api server","time":"2025-08-29 09:55:18"}```
4. Switch to internal redis now and see that all work as expected
**Expected behavior**
Get http 200 from curl, get message like this in responce {"status":true,"message":"Event received","data":149} and get this webhook proxied to output.

**Environment (please complete the following information):**
 - OS: amazon linux 2023(20250715)
 - K8S: 1.31, arm64
 - Containerized?: eks k8s, helm
 - Dragonfly Version: cloud

**Additional context**
With support([abhijat@dragonflydb.io](mailto:abhijat@dragonflydb.io)) we tried to setup parameters to connect url so it will look like this:
CONVOY_REDIS_CLUSTER_ADDRESSES : redis://xxxxxxxx.dragonflydb.cloud:XXXX/5?protocol=2
also was added parameter --lua_resp2_legacy_float=true to dragonfly side and allowed scripting with undeclared keys

**Comment by abhijat** [comment]
> CONVOY_REDIS_CLUSTER_ADDRESSES : redis://xxxxxxxx.dragonflydb.cloud:XXXX/5?protocol=2

note that with https://github.com/dragonflydb/dragonfly/pull/5754 this will no longer be required in new releases, just setting `lua_resp2_legacy_float` is enough.

**Comment by BorysTheDev** [comment]
During tests, I got a crash with a stack trace (can't reproduce)
```
PC: @     0x5ca8ffcf005c  (unknown)  std::chrono::time_point<>::time_since_epoch()
    @     0x5ca900f5f93c         64  absl::lts_20250512::WriteFailureInfo()
    @     0x5ca900f5f5a8        272  absl::lts_20250512::AbslFailureSignalHandler()
    @     0x792e82445330       3072  (unknown)
[...]

PR #5776, opened 2025-09-05
fix: script error reply

BorysTheDev:
fixes: #5762
problem: inconsistent error reply, if the script uses redis.error_reply(err_str) method, the error type is the first word in err_str. In the current case, we returned "-ERR NOT FOUND", but expected result was "-NOT FOUND", where "NOT FOUND" is provided by the convoy script via redis.error_reply("NOT FOUND") call

fix: added SendScriptErrorReply() method that append "-" to  err_str in response if redis.error_reply() in the script was called;

### DIFF
--- a/src/server/main_service.cc
+++ b/src/server/main_service.cc
@@ -383,7 +383,11 @@ class EvalSerializer : public ObjectExplorer {
   }
 
   void OnError(string_view str) {
-    rb_->SendError(absl::StrCat("-", str));
+    if (!str.empty() && str.front() != '-') {
+      rb_->SendError(absl::StrCat("-", str));
+    } else {
+      rb_->SendError(str);
+    }
   }
 
  private:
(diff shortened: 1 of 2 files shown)


romange:
@BorysTheDev 

Please provide exactly the string that was passed to SendError that caused the bug. Please provide in the PR description the expected and the actual responses. I see in `void RedisReplyBuilderBase::SendError(` that we handled errors that did not start with `-` so I want to understand better why it did not work.

BorysTheDev:
@romange I've updated description

romange:
1. I think the fix is not optimal and there is no need in the dedicated function. Please see EvalSerializer and InterpreterReplier. Pretty sure prepending "-" in one of those functions would suffice.
2. Please add tests to dragonfly_test.cc covering such case. Also please add coverage to standard errors like syntax error  thrown in lua scripts via `redis.pcall`. For example `return redis.pcall('foo', 'bar')` or `return redis.pcall('incrby', 'foo', 'bar')` that should propagate standard errors correctly. make sure that both the custom and standard errors are rendered exactly the same in DF and valkey after your changes.

romange:
Good job finding the culprit.

src/server/main_service.cc
```diff
@@ -383,7 +383,7 @@ class EvalSerializer : public ObjectExplorer {
   }
 
   void OnError(string_view str) {
-    rb_->SendError(str);
+    rb_->SendScriptErrorReply(str);
```
romange: why not add `-` here?

BorysTheDev: I didn't want to create new string

romange: I think given the frequency of crossing this code path (once in 3 years) the allocation is justified.

BorysTheDev: it's a main scenario for Convoy

romange: I understand but I still think it's an premature optimization in this case (and there is no allocation for short strings)

romange:
I fixed the problem

PR #5780, opened 2025-09-08
fix: Scan on non-existing key should return zero cursor

### LINKED ISSUE #5777: `HSCAN/SSCAN/ZSCAN` on non-existing keys should return zero cursor
## Describe The Bug
- User reported that if a hash/set/zset gets deleted in the middle of processing (by using `HSCAN/SSCAN/ZSCAN`), for the response:
  - **The cursor is not resetting to `0`. Instead, it returns the input cursor value.**
  - This could result in non-terminating scanning on the client side.
- **In general, if `HSCAN/SSCAN/ZSCAN` is used on a non-existing key, the result cursor should be `0`.**

## To Reproduce

```bash
dragonfly$> HSCAN non-existent 1000 COUNT 5
1) "1000"
2) (empty array)

dragonfly$> SSCAN non-existent 2000 COUNT 5
1) "2000"
2) (empty array)

dragonfly$> ZSCAN non-existent 3000 COUNT 5
1) "3000"
2) (empty array)
```

## Expected Behavior

```bash
redis$> HSCAN non-existent 1000 COUNT 5
1) "0"
2) (empty array)

redis$> SSCAN non-existent 2000 COUNT 5
1) "0"
2) (empty array)

redis$> ZSCAN non-existent 3000 COUNT 5
1) "0"
2) (empty array)
```

## Environment
 - OS: `Ubuntu 22.04.5 LTS`
 - Kernel: # Command: `Linux 821c31b0d587 6.10.14-linuxkit #1 SMP Fri Nov 29 17:22:03 UTC 2024 aarch64 aarch64 aarch64 GNU/Linux`
 - Containerized?: Docker on Mac
 - Dragonfly Version: `v1.33.1`

mkaruza:
Fixes #5777

### DIFF
--- a/src/server/hset_family.cc
+++ b/src/server/hset_family.cc
@@ -332,6 +332,7 @@ OpResult<StringVec> OpScan(const OpArgs& op_args, std::string_view key, uint64_t
 
   if (!find_res) {
     DVLOG(1) << "ScanOp: find failed: " << find_res << ", baling out";
+    *cursor = 0;
     return find_res.status();
   }
 
--- a/src/server/hset_family_test.cc
+++ b/src/server/hset_family_test.cc
@@ -131,13 +131,19 @@ TEST_F(HSetFamilyTest, HIncrRespected) {
 }
 
 TEST_F(HSetFamilyTest, HScan) {
+  auto resp = Run("hscan non-existing-key 100 count 5");
+  ASSERT_THAT(resp, ArgType(RespExpr::ARRAY));
+  ASSERT_THAT(resp.GetVec(), ElementsAre(ArgType(RespExpr::STRING), ArgType(RespExpr::ARRAY)));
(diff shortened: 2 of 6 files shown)


Niennienzz:
Thanks for the fix! @mkaruza 

Without understanding the code, I would like to double-check and make sure that [this behavior of scanning](https://redis.io/docs/latest/commands/scan/#number-of-elements-returned-at-every-scan-call) is preserved.

Basically, during a full iteration:

> The scan commands are also allowed to return zero elements, and the client should not consider the iteration complete as long as the returned cursor is not zero.

^ In that case, the client should continue scanning.


Unlike the non-existing key case described in the issue, where the client should stop as it gets cursor `0`.

mkaruza:
@Niennienzz can you let me know after you double check if this is correct

PR #5781, opened 2025-09-08
chore: Resize vector in StringMap::RandomPairs

mkaruza:
Tested on GCC 15 accessing vector which is reserved and not resized can trigger assert. Use resize instead.


kostasrim (approved):
You could also keep reserve + push_back.

IMO nice that GCC asserts that 😄

Issue #5444, opened 2025-07-09
Add shutdown command options


BorysTheDev:
We lack the SHUTDOWN command options: [NOW] [FORCE] [ABORT]
Consider adding an option [ SAFE ] that is specific for Valkey

see https://valkey.io/commands/shutdown/

PR #5783, opened 2025-09-08
feat: SHUTDOWN: fast path for NOW/FORCE; unify SAVE/SAFE; support NOSAVE

vyavdoshenko:
Fixes: https://github.com/dragonflydb/dragonfly/issues/5444

- Parse SHUTDOWN options: SAVE, NOSAVE, SAFE, NOW, FORCE, ABORT
- SAVE and SAFE: unified behavior; snapshot happens during shutdown (no immediate save)
- FORCE: disables snapshot (save_on_shutdown=false) and enables fast shutdown
- NOW/FORCE: enable fast shutdown (skip graceful wait in Listener::PreShutdown)
- ABORT: returns error (not supported yet)

### DIFF
--- a/src/facade/dragonfly_listener.cc
+++ b/src/facade/dragonfly_listener.cc
@@ -256,7 +256,7 @@ bool Listener::IsMainInterface() const {
 
 void Listener::PreShutdown() {
   // If NOW/FORCE requested, expedite shutdown without waiting.
-  if (g_shutdown_fast.load(std::memory_order_relaxed)) {
+  if (g_shutdown_fast.load(std::memory_order_acquire)) {
     return;
   }
 
--- a/src/facade/dragonfly_listener.cc
+++ b/src/facade/dragonfly_listener.cc
@@ -66,6 +66,9 @@ CONFIG_enum(tls_auth_clients, "yes", "", tls_auth_clients_enum, tls_auth_clients
 
 namespace facade {
 
+// See dragonfly_listener.h
+std::atomic<bool> g_shutdown_fast{false};
+
 using namespace util;
 using util::detail::SafeErrorMessage;
 
@@ -252,13 +255,12 @@ bool Listener::IsMainInterface() const {
 }
 
 void Listener::PreShutdown() {
-  // Iterate on all connections and allow them to finish their commands for
-  // a short period.
-  // Executed commands can be visible in snapshots or replicas, but if we close the client
-  // connections too fast we might not send the acknowledgment for those commands.
-  // This shouldn't take a long time: All clients should reject incoming commands
-  // at this stage since we're in SHUTDOWN mode.
-  // If a command is running for too long we give up and proceed.
+  // If NOW/FORCE requested, expedite shutdown without waiting.
+  if (g_shutdown_fast.load(std::memory_order_relaxed)) {
+    return;
+  }
+
+  // Otherwise: Iterate on all connections and allow them to finish their commands briefly.
   DispatchTracker tracker{{this}};
   tracker.TrackAll();
 
--- a/src/facade/dragonfly_listener.cc
+++ b/src/facade/dragonfly_listener.cc
@@ -260,7 +260,13 @@ void Listener::PreShutdown() {
     return;
   }
 
-  // Otherwise: Iterate on all connections and allow them to finish their commands briefly.
+  // Otherwise: Iterate on all connections and allow them to finish their commands for
+  // a short period.
+  // Executed commands can be visible in snapshots or replicas, but if we close the client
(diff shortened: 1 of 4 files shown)


src/server/server_family.cc
```diff
@@ -3781,24 +3781,72 @@ void ServerFamily::Latency(CmdArgList args, const CommandContext& cmd_cntx) {
...
+  // Supported options (case-insensitive):
+  // SAVE | NOSAVE, NOW, FORCE, ABORT, SAFE (Valkey-specific, the same as SAVE in Dragonfly)
+  bool saw_save_toggle = false;  // whether SAVE/NOSAVE/SAFE appeared
+  bool opt_now = false;
+  bool opt_force = false;
+  bool opt_abort = false;
+
+  for (size_t i = 0; i < args.size(); ++i) {
+    std::string opt = absl::AsciiStrToUpper(ArgS(args, i));
+    if (opt == "SAVE") {
```
BorysTheDev: Please use CmdArgParser

src/server/server_family.cc
```diff
@@ -3781,24 +3781,86 @@ void ServerFamily::Latency(CmdArgList args, const CommandContext& cmd_cntx) {
...
+      case ShutOpt::ABORT:
+        opt_abort = true;
+        break;
+      case ShutOpt::SAFE:
+        if (saw_save_toggle) {
+          cmd_cntx.rb->SendError(kSyntaxErr);
+          return;
+        }
+        saw_save_toggle = true;
+        // Same behavior as SAVE: save during shutdown sequence
```
BorysTheDev: if it is the same let's do fallthorugh

src/server/server_family.cc
```diff
@@ -3781,24 +3781,86 @@ void ServerFamily::Latency(CmdArgList args, const CommandContext& cmd_cntx) {
...
+        break;
+      case ShutOpt::ABORT:
+        opt_abort = true;
+        break;
+      case ShutOpt::SAFE:
+        if (saw_save_toggle) {
+          cmd_cntx.rb->SendError(kSyntaxErr);
+          return;
+        }
+        saw_save_toggle = true;
```
BorysTheDev: looks overcomplicated, all these flags, you already have mapped them to enum, maybe it makes sense to use enum as bits, in this case you don't need this switch at all

BorysTheDev:
LGTM

src/facade/dragonfly_listener.cc
```diff
@@ -252,7 +255,12 @@ bool Listener::IsMainInterface() const {
 }
 
 void Listener::PreShutdown() {
-  // Iterate on all connections and allow them to finish their commands for
+  // If NOW/FORCE requested, expedite shutdown without waiting.
+  if (g_shutdown_fast.load(std::memory_order_relaxed)) {
```
romange:
```suggestion
  if (g_shutdown_fast.load(std::memory_order_acquire)) {
```

src/server/server_family.cc
```diff
@@ -3781,24 +3781,86 @@ void ServerFamily::Latency(CmdArgList args, const CommandContext& cmd_cntx) {
...
+  // No immediate save here: SAVE/SAFE/NOSAVE only configure save_on_shutdown_
+  // and the snapshot is handled in ServerFamily::Shutdown.
+
+  // FORCE implies no snapshot on shutdown regardless of SAVE/SAFE
+  if (opt_force) {
+    save_on_shutdown_ = false;
   }
 
+  // Wire NOW/FORCE to a single fast-shutdown flag for listeners.
+  facade::g_shutdown_fast.store(opt_now || opt_force, std::memory_order_relaxed);
```
romange:
```suggestion
  facade::g_shutdown_fast.store(opt_now || opt_force, std::memory_order_seq_cst);
```

src/server/server_family.cc
```diff
@@ -3781,24 +3781,86 @@ void ServerFamily::Latency(CmdArgList args, const CommandContext& cmd_cntx) {
...
+
   CHECK_NOTNULL(acceptor_)->Stop();
   cmd_cntx.rb->SendOk();
+
+  // Reset flag for any subsequent restarts (defensive; process exits shortly).
+  facade::g_shutdown_fast.store(false, std::memory_order_relaxed);
```
romange:
```suggestion
  // Reset flag for any subsequent restarts (mainly for tests).
  facade::g_shutdown_fast.store(false, std::memory_order_seq_cst);
```

src/server/server_family.cc

src/server/server_family.cc

BorysTheDev: looks great

PR #5784, opened 2025-09-09
core: Replace sbf with hyperloglog

abhijat:
Change bloom filter to hyperloglog as discussed in pr https://github.com/dragonflydb/dragonfly/pull/5560#issuecomment-3254082470

external usage of the page usage stats class does not change.

With HLL no storage is required for the page addresses which reduces memory usage, the code only needs to maintain a constant count of registers to estimate cardinality, which is what we need at the end of the defragment command.

Note: Dense hll is used here, sparse hll could be used and converted to dense (as a follow up). The count of page addresses per class can be less than the number of registers (16384), in this case sparse hll might take up lesser memory.

### DIFF
--- a/src/core/page_usage_stats.cc
+++ b/src/core/page_usage_stats.cc
@@ -38,12 +38,6 @@ HllBufferPtr InitHllPtr() {
   return p;
 }
 
-uint64_t CheckedPFCount(const HllBufferPtr h) {
-  const int64_t count = pfcountSingle(h);
-  CHECK_GT(count, -1);
-  return static_cast<uint64_t>(count);
-}
-
 }  // namespace
 
 void CollectedPageStats::Merge(CollectedPageStats&& other, uint16_t shard_id) {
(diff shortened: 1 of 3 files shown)


src/core/page_usage_stats.cc
```diff
@@ -30,24 +30,24 @@ using absl::StripTrailingAsciiWhitespace;
...
-constexpr auto kInitialSBFCap = 1000;
-constexpr auto kFProb = 0.001;
-constexpr auto kGrowthFactor = 2;
 constexpr auto kHistSignificantFigures = 3;
 
-}  // namespace
-
-FilterWithSize::FilterWithSize()
-    : sbf{SBF{kInitialSBFCap, kFProb, kGrowthFactor, PMR_NS::get_default_resource()}}, size{0} {
+HllBufferPtr InitHllPtr() {
```
romange, line 33: remove `#include "core/bloom.h"` as well

src/core/page_usage_stats.cc
```diff
@@ -30,24 +30,24 @@ using absl::StripTrailingAsciiWhitespace;
...
 }
 
-void FilterWithSize::Add(uintptr_t address) {
-  const auto s = std::to_string(address);
-  if (sbf.Add(s)) {
-    size += 1;
-  }
+uint64_t CheckedPFCount(const HllBufferPtr h) {
+  const int64_t count = pfcountSingle(h);
+  CHECK_GT(count, -1);
```
romange: Lets remove this CHECK so that we won't crash for no reason.

src/core/page_usage_stats.h
```diff
@@ -11,19 +11,16 @@
 
 #include "core/bloom.h"
```
romange: remove this include

src/core/page_usage_stats.cc
```diff
@@ -98,34 +98,49 @@ std::string CollectedPageStats::ToString() const {
...
+  delete[] pages_marked_for_realloc.hll;
+  delete[] pages_full.hll;
+  delete[] pages_reserved_for_malloc.hll;
+  delete[] pages_with_heap_mismatch.hll;
+  delete[] pages_above_threshold.hll;
   hdr_close(page_usage_hist);
 }
 
 void PageUsage::UniquePages::AddStat(mi_page_usage_stats_t stat) {
   const auto address = stat.page_address;
```
romange: pls do not use auto for basic types. it hurts readability

src/core/page_usage_stats.cc
```diff
@@ -98,34 +98,49 @@ std::string CollectedPageStats::ToString() const {
...
-  pages_scanned.Add(address);
+  const std::string s = std::to_string(address);
```
romange: I find it unnecessary.
if address  is uintptr_t or similar, 
just use `pfadd_dense(pages_scanned, reinterpret_cast<const uint8_t*>(&address), sizeof(address))`.

romange: and in all the other places.

src/core/page_usage_stats.cc
abhijat: `pfcountSingle` can return -1, do we want to ignore it? Should it return 0 count in this case?

romange: it can if we have a bug. either log dfatal in this case or ignore it. in any case it does not justify crashing the process.

src/core/page_usage_stats.cc
abhijat: address is `uintptr_t`. changed to use it directly

src/core/page_usage_stats.cc
abhijat: this is removed now

src/core/page_usage_stats.h
abhijat: removed

src/core/page_usage_stats.cc
abhijat: removed

src/core/page_usage_stats.cc
abhijat: ok, removed the check and inlined the cast

Issue #5785, opened 2025-09-09
reject client caching command if tracking is off


abhijat:
Currently even if tracking is turned off in optin mode, client caching command is accepted. if the tracking is off, caching directives should be rejected.

PR #5786, opened 2025-09-09
server: Reject caching command if tracking is not enabled

abhijat:
FIXES https://github.com/dragonflydb/dragonfly/issues/5785

If client tracking is off, caching commands should be rejected. relevant discussion https://github.com/dragonflydb/documentation/pull/465

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -476,6 +476,11 @@ void ClientCaching(CmdArgList args, SinkReplyBuilder* builder, Transaction* tx,
     return builder->SendError(kSyntaxErr);
   }
 
+  if (!cntx->conn_state.tracking_info_.IsTrackingOn()) {
+    return builder->SendError(
+        "ERR CLIENT CACHING can only be called when the client is in tracking mode");
+  }
+
   using Tracking = ConnectionState::ClientTracking;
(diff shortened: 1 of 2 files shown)


src/server/server_family.cc
```diff
@@ -476,6 +476,11 @@ void ClientCaching(CmdArgList args, SinkReplyBuilder* builder, Transaction* tx,
     return builder->SendError(kSyntaxErr);
   }
 
+  if (!cntx->conn_state.tracking_info_.IsTrackingOn()) {
+    return builder->SendError(
+        "ERR CLIENT CACHING can only be called when the client is in tracking mode");
```
romange: are you sure you need "ERR " prefix here? is this error compatible with valkey?

abhijat: tbh I added `ERR ` taking a cue from the next two error returns. The valkey error is different, I will change it to match. Maybe the next two errors should be changed as well.

romange: `SendError` should add  `-ERR` automatically, so probably we should

abhijat: removed the prefix, checked and it was sending ERR twice, and changed error message to be same as valkey.

PR #5787, opened 2025-09-09
dfly_bench: Show error on failing to make address

abhijat:
Sometimes when dfly_bench parses a response to extract host to connect to, it fails inside boost make_address. This error is thrown up the stack causing the program to fail without clear indication of what the error was. 

See example run https://github.com/dragonflydb/controlplane/actions/runs/17032489974/job/48277888480

Changed to use the `error_code` variant of `boost::asio::ip::make_address` to print the error with the address which failed to parse.

### DIFF
--- a/src/server/dfly_bench.cc
+++ b/src/server/dfly_bench.cc
@@ -842,8 +842,8 @@ void Driver::ParseRESP() {
           CHECK_EQ(2u, addr_parts.size());
 
           auto host = boost::asio::ip::make_address(addr_parts[0], ec);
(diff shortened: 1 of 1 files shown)


mkaruza (approved):
NIT: can you add space when you end first part of message between `ec.message()` and start start next part of message with space for readability purposes.

abhijat:
> NIT: can you add space when you end first part of message between `ec.message()` and start start next part of message with space for readability purposes.

added space

Issue #5782, opened 2025-09-08
replication takeover fail


BorysTheDev:
https://github.com/dragonflydb/dragonfly/actions/runs/17424516992/job/49468921351

The master shuts down and drops the replication connection faster than it sends a reply for takeover. We need to check that ReplyBuilder has flushed the socket before shutdown. So replica node didn't get reply and couldn't finish takeover

romange:
@BorysTheDev  how do you know that this is what happened ?

BorysTheDev:
@romange I see in the logs that master got TAKEOVER command and did shutdown after it, but on the replica instead of master reply I see a broken connection

romange:
it should be relatively easy to reproduce by using master/replica in differrent regions or using your home machine as replica and the cloud machine as master or vice versa. the point - to have significant latency.

BorysTheDev:
I don't think so, because the question is whether the data was sent, and latency isn't important in this case. As I understand in the current situation we wrote the data into the buffer and closed the socket before sending it.

romange:
it is not possible. then it's a logical bug. either we flush the buffer or not. I do not accept solutions based on unchecked hypothesis.

romange:
`rb->SendOk();` should not buffer the reply.

BorysTheDev:
we use uring, it is a buffer

BorysTheDev:
I've tried to reproduce this bug with two DF instances and a big latency between Virginia and Mumbai, but failed.

To my mind, we have the following situation:

In Linux, when we close(fd), the file descriptor is released.
If that was the last reference to the underlying socket, the kernel tears down the socket (FIN or RST depending on state).

But io_uring has its own reference to the file description (the struct file) when we queue the SQE.
This means the send won’t necessarily fail immediately just because you closed your fd in userspace — the kernel still has a valid reference when it starts processing the SQE.

So 2 situations can be:
If the io_uring worker sends data before the socket is fully destroyed, it can succeed (data goes out).
If the socket is already gone (last ref dropped, protocol stack cleaned up), the operation fails.

There is a race condition:
We don’t control whether the close "wins" before the send is executed.
Even if we see the SQE in the ring, it might not have started yet when we call close().

We have quite a lot of code before closing the socket, so in most cases, we succeed in sending the response to the replica.

romange:
[...]

PR #5788, opened 2025-09-09
fix: add sleep to deliver reply to replica for takeover

BorysTheDev:
fixes: https://github.com/dragonflydb/dragonfly/issues/5782
problem: we don't have enough time to deliver the reply for the takeover, because the socket was closed to fast
fix: added sleep before shutdown

### DIFF
--- a/src/server/dflycmd.cc
+++ b/src/server/dflycmd.cc
@@ -540,6 +540,9 @@ void DflyCmd::TakeOver(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext
 
   // For non-cluster mode we shutdown
   if (detail::cluster_mode != detail::ClusterMode::kRealCluster) {
+    // we don't have flush() mechanism for ReplyBuilder, so we wait a bit to be sure that reply was
+    // delivered to the replica. See https://github.com/dragonflydb/dragonfly/issues/5782
+    ThisFiber::SleepFor(1s);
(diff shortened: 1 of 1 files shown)


src/server/dflycmd.cc
```diff
@@ -540,6 +540,9 @@ void DflyCmd::TakeOver(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext
 
   // For non-cluster mode we shutdown
   if (detail::cluster_mode != detail::ClusterMode::kRealCluster) {
+    // we don't have flush() mechanism for ReplyBuilder, so we wait a bit to be sure that reply was
```
romange, line 543: 1s is not a bit.  have you checked it helps or is it a guess?

BorysTheDev: it's a guess, I don't think that 1s  is critical before shutdown

romange: it's not critical but if a replica got the reply, it will break the connection itself. In that case `Connection::OnBreakCb` will update `cc_->conn_closing = true;` which can serve as a confirmation to proceed further. So I would actually here do something like:
```
for (unsigned i = 0; i < 50; ++i) {
  if (cntx->conn_closing) {
    VLOG(1) << "Take over confirmed at iteration " << i;
     break;
  }
  SleepFor(20ms);
}
```
or something like that. and then manually check that it helps on long-latency connection where before that it failed.

Issue #5706, opened 2025-08-21
Support defragmentation within quicklist


abhijat:
defragmentation should be supported in quicklists like it is for other data structures.

PR #5790, opened 2025-09-10
feat(core): support defragmentation for quicklist

abhijat:
Adds defragmentation support to quicklist by iterating over nodes starting from head and reallocating the entry if the containing page is below threshold.

The iterator api is not used for this, because we do not need to peek inside nodes, so we avoid the need to decompress and then compress node content during iteration.

We also do not need to iterate inside listpack nodes, because listpack data is contiguous and moved in a single step.

fixes https://github.com/dragonflydb/dragonfly/issues/5706

### DIFF
--- a/src/core/page_usage_stats_test.cc
+++ b/src/core/page_usage_stats_test.cc
@@ -10,6 +10,7 @@
 #include "base/logging.h"
 #include "core/compact_object.h"
 #include "core/mi_memory_resource.h"
+#include "core/qlist.h"
 #include "core/score_map.h"
 #include "core/small_string.h"
 #include "core/sorted_map.h"
@@ -49,6 +50,7 @@ class PageUsageStatsTest : public ::testing::Test {
     string_set_ = std::make_unique<StringSet>(&m_);
     string_map_ = std::make_unique<StringMap>(&m_);
     SmallString::InitThreadLocal(m_.heap());
+    qlist_ = std::make_unique<QList>(2, 2);
   }
 
   void TearDown() override {
@@ -57,6 +59,7 @@ class PageUsageStatsTest : public ::testing::Test {
     string_set_.reset();
     string_map_.reset();
     small_string_.Free();
+    qlist_->Clear();
     EXPECT_EQ(zmalloc_used_memory_tl, 0);
   }
 
@@ -66,6 +69,7 @@ class PageUsageStatsTest : public ::testing::Test {
   std::unique_ptr<StringSet> string_set_;
   std::unique_ptr<StringMap> string_map_;
   SmallString small_string_{};
+  std::unique_ptr<QList> qlist_;
   CompactObj c_obj_{};
 };
 
@@ -79,6 +83,8 @@ TEST_F(PageUsageStatsTest, Defrag) {
   // INT_TAG, defrag will be skipped
   c_obj_.SetString("1");
 
+  qlist_->Push("xxxx", QList::HEAD);
+
   {
     PageUsage p{CollectPageStats::YES, 0.1};
     score_map_->begin().ReallocIfNeeded(&p);
@@ -87,6 +93,7 @@ TEST_F(PageUsageStatsTest, Defrag) {
     string_map_->begin().ReallocIfNeeded(&p);
     small_string_.DefragIfNeeded(&p);
     c_obj_.DefragIfNeeded(&p);
+    qlist_->DefragIfNeeded(&p);
 
     const auto stats = p.CollectedStats();
     EXPECT_GT(stats.pages_scanned, 0);
@@ -100,6 +107,7 @@ TEST_F(PageUsageStatsTest, Defrag) {
     string_set_->begin().ReallocIfNeeded(&p);
     string_map_->begin().ReallocIfNeeded(&p);
     small_string_.DefragIfNeeded(&p);
+    qlist_->DefragIfNeeded(&p);
     EXPECT_EQ(p.CollectedStats().pages_scanned, 0);
   }
 }


src/core/qlist.h
```diff
@@ -91,6 +92,8 @@ class QList {
     // Returns false if no more entries.
     bool Next();
 
+    bool DefragIfNeeded(PageUsage* page_usage);
```
abhijat: this method should really be private and only usable via the friend class `QList` but it is needed in the test case. https://google.github.io/googletest/reference/testing.html#FRIEND_TEST works if the test class is added as a friend but I don't want to add gtest dependencies to `qlist.h`, so this is left as public

src/core/qlist.cc
```diff
@@ -385,6 +386,15 @@ void QList::SetPackedThreshold(unsigned threshold) {
   packed_threshold = threshold;
 }
 
+size_t QList::DefragIfNeeded(PageUsage* page_usage) {
+  Iterator it = GetIterator(HEAD);
+  size_t reallocated = 0;
+  while (it.Next()) {
+    reallocated += it.DefragIfNeeded(page_usage);
+  }
```
abhijat: iteration over compressed nodes is expensive, it is not needed for defragmentation, it might be faster to  just memcpy the data without using the iterator API

src/core/qlist.h
abhijat: this method is not needed anymore

src/core/qlist.cc
abhijat: tried this approach in latest version

src/core/qlist.cc
```diff
@@ -385,6 +386,37 @@ void QList::SetPackedThreshold(unsigned threshold) {
...
+    }
+
+    unsigned char* old_entry = curr->entry;
+    const bool is_raw_list_pack = curr->container == QUICKLIST_NODE_CONTAINER_PACKED &&
+                                  curr->encoding == QUICKLIST_NODE_ENCODING_RAW;
+    // if the entry is an uncompressed listpack, use the listpack api to initialize and free it,
+    // otherwise it is compressed data and/or a plain string stored in the node, in either case
+    // allocate/free the data directly.
+    unsigned char* new_entry =
+        is_raw_list_pack ? lpNew(curr->sz) : static_cast<unsigned char*>(zmalloc(curr->sz));
```
abhijat: I thought to just use zmalloc here for everything to keep things simple since we later copy the list byte for byte but `lpNew` does some adjustments based on input size, so we use `lpNew`

romange: I suspect it's incorrect.  `curr->sz` is the size of the `entry` blob even if it's listpack.
`lpNew(curr->sz)` will allocate a blob larger size than `curr->sz` to be able to put `curr->sz` bytes into a listpack. And in general entry can be compressed, uncompressed listpack or plain blob. from the first sight the code here seems wrong (though  I may be wrong).

the right approach is to use zmalloc as you wrote.

abhijat: The reason I used this was `curr->sz` for listpack nodes is initialized as `lpBytes(..)` which is the total size of listpack (data elements + 6 bytes of header). 

`lpNew` allocates whatever capacity we send it, in this case it will allocate the total size of listpack including data + header. 

This is necessary because next we do a memcpy upto `curr->size` so we should have allocated exactly enough space for the header + data to do a byte wise copy.

> lpNew(curr->sz) will allocate a blob larger size than curr->sz

From what I saw it should allocate the size sent to it unless there is some wrapper code somewhere I missed
https://github.com/dragonflydb/dragonfly/blob/c473be5109d2b5ec9394ce974f6c5c35c3122635/src/redis/listpack.c#L260

I will change it to zmalloc though as the size check for which I added `lpNew` requires `capacity < header size` (most often when capacity=0), in this case it will never happen as the node is already allocated.

For consistency I will also change the later `lpFree` to `zfree`

abhijat: changed

src/core/qlist.cc
```diff
@@ -385,6 +386,25 @@ void QList::SetPackedThreshold(unsigned threshold) {
   packed_threshold = threshold;
 }
 
+size_t QList::DefragIfNeeded(PageUsage* page_usage) {
+  size_t reallocated = 0;
+  for (Node* curr = head_; curr; curr = curr->next) {
```
romange, line 392: did you consider reallocating the nodes themselves? 
btw, I do not think it is needed but at least add a comment:
`// we do not reallocate nodes because their size is constant so it is expected that their pages will usually be unfragmented` or smthing like this

abhijat: I did not think of reallocating nodes yet. Added comment for now, but can look at reallocating nodes also in a follow up PR.

romange: I do not think it's needed. Fragmentation happens when lots of blocks of size X are allocated, then some of them are freed, and then lots of blocks of size Y are allocated and so on. nodes have a constant size so the pages should be constantly reused

romange: and I want this explanation to be in the comment because it's not straightforward to understand

abhijat: I added a bit longer explanation and moved it to the method.

src/core/qlist.cc
```diff
@@ -385,6 +386,30 @@ void QList::SetPackedThreshold(unsigned threshold) {
...
+    if (!page_usage->IsPageForObjectUnderUtilized(curr->entry)) {
+      continue;
+    }
+
+    // Data pointed to by the nodes is reallocated. The nodes themselves are not reallocated because
+    // of their constant (and relatively small, ~40 bytes per object) size. Defragmentation fixes
+    // fragmented memory allocation, which usually happens when variable-sized blocks of data are
+    // allocated and deallocated, which is not expected with nodes.
+    auto new_entry = static_cast<unsigned char*>(zmalloc(curr->sz));
```
romange, line 401: nit: use `uint8_t*` instead of `unsigned char*` and use less `auto`.

src/core/qlist.cc
```diff
@@ -385,6 +386,30 @@ void QList::SetPackedThreshold(unsigned threshold) {
...
+    memcpy(new_entry, curr->entry, curr->sz);
+
+    unsigned char* old_entry = curr->entry;
```
romange, line 404: same: use `uint8_t*`

romange (approved):

src/core/qlist.cc
abhijat: I figured this is a good place for auto as there is a static cast with the explicit type on the same line, I will change it in a follow up PR, I have a couple of other cleanups to make as well.

abhijat: I have made this change in https://github.com/dragonflydb/dragonfly/pull/5826

PR #5796, opened 2025-09-11
chore(rdb): Insert big values only after full construction

dranikpg:
Motivation: Withhold partial values (big values that are still streamed) from the dashtable to prevent them from being involved in background processes, like tiered storage offloading for example. Allowing partial values to participate in those can break correctness

### DIFF
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -2565,11 +2565,20 @@ void RdbLoader::CreateObjectOnShard(const DbContext& db_cntx, const Item* item, 
   thread_local std::unordered_map<std::string, std::unique_ptr<PrimeValue>> now_streamed;
   LoadConfig config_copy = item->load_config;
   if (item->load_config.streamed && item->load_config.append) {
-    // When a sets members are fully expired, the value is not created, so try again if it's missing
-    if (auto it = now_streamed.find(item->key); it == now_streamed.end())
-      config_copy.append = false;
-    else
+    if (auto it = now_streamed.find(item->key); it != now_streamed.end()) {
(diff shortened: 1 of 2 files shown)


dranikpg:
I don't really like the general structure, but it's the minimal solution that can be squeezed in here

src/server/rdb_load.cc
```diff
@@ -2561,38 +2561,18 @@ void RdbLoader::CreateObjectOnShard(const DbContext& db_cntx, const Item* item,
...
-  LoadConfig tmp_load_config = item->load_config;
-
-  // If we're appending the item to an existing key, first load the
-  // object.
-  if (tmp_load_config.append) {
-    append_res = db_slice->FindMutable(db_cntx, item->key);
-    if (IsValid(append_res.it)) {
-      pv_ptr = &append_res.it->second;
-    } else {
-      // If the item has expired we may not find the key. Note if the key
```
romange, line 2577: you deleted this logic. Are you sure it's not needed at all?
does this state happen due to background expiry?

romange (commented):
lgtm

src/server/rdb_load.cc
dranikpg: The logic above - yes, because we try to stash only values that are finished, there's no way its offloaded. The logic below - remanined below, but more compact with new code

romange: I meant the logic with `key_is_not_expired` and `is_set_expiry_type`.

I do not see it in your code

dranikpg: It's a safety net, returned it

PR #5797, opened 2025-09-11
chore: Limit expiry value to 32 bits.

romange:
We use adaptive precision, where we keep "millis" precision if we can, and switch to seconds precision if the deadline is too large.

This actually was before but now I reduced the cut-off to 4B ms, or 49 days. Also, the ttl in seconds is now limited to 4B sec, with larger values quietly rounded down to this limit.

### DIFF
--- a/src/core/expire_period.h
+++ b/src/core/expire_period.h
@@ -8,9 +8,13 @@
 
 namespace dfly {
 
+// ExpirePeriod encapsulates the expiration period of a key.
+// It can represent periods in milliseconds up to ~49 days and in seconds up to ~136 years.
+// If value in milliseconds is too high, it switches to seconds precision.
+// And if it's still too high, it silently clamps to the max value.
 class ExpirePeriod {
  public:
-  static constexpr size_t kMaxGenId = 15;
+  static constexpr uint32_t kMaxGenId = 15;
 
   ExpirePeriod() : val_(0), gen_(0), precision_(0) {
     static_assert(sizeof(ExpirePeriod) == 8);  // TODO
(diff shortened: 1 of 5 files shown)


romange:
the next step would be to start using `expire_base_` and generations to represent the absolute times using ~49 days ExpirePeriod. The absolute time := expire_base_ + expire_period so we should refresh `expire_base_` daily to maintain 48-49 days span.

src/server/server_family.cc
```diff
@@ -1284,10 +1284,7 @@ void ServerFamily::FlushAll(Namespace* ns) {
   boost::intrusive_ptr<Transaction> flush_trans(new Transaction{cid});
   flush_trans->InitByArgs(ns, 0, {});
   VLOG(1) << "Performing flush";
-  error_code ec = Drakarys(flush_trans.get(), DbSlice::kDbAll);
-  if (ec) {
-    LOG(ERROR) << "Error flushing db " << ec.message();
-  }
+  Drakarys(flush_trans.get(), DbSlice::kDbAll);
 }
```
dranikpg, line 1289: I noticed this as well in #5821

src/core/expire_period.h
```diff
@@ -35,28 +39,30 @@ class ExpirePeriod {
...
-  precision_ = 1;
-  if (ms < kBarrier << 10) {
-    ms = (ms + 500) / 1000;   // seconds
+  precision_ = 1;  // seconds
+  if (ms < UINT64_MAX / 2) {
+    ms = (ms + 500) / 1000;
+    val_ = ms >= UINT32_MAX ? UINT32_MAX - 1 : ms;
+  } else {
+    val_ = UINT32_MAX - 1;
   }
```
dranikpg, line 65: Not sure I understand the if with int64_t::max / 2, why not just calculate the seconds value and see if it fits into int32_t 👀 Just check to divide first and then add the remainder > 0 (I assume the if is about this)

src/core/expire_period.h
```diff
@@ -35,28 +39,30 @@ class ExpirePeriod {
...
+    return precision_ == 1;
+  }
 
  private:
-  uint64_t val_ : 59;
-  uint64_t gen_ : 4;
-  uint64_t precision_ : 1;  // 0 - ms, 1 - sec.
+  uint32_t val_;
+  uint8_t gen_ : 2;        // generation id.
+  uint8_t precision_ : 1;  // 0 - ms, 1 - sec.
```
dranikpg, line 49: Not sure if you're optimizing with all those bitfields, but the struct is not packed either way 🙂
image

dranikpg:
Not sure if I have to start using the phrase "not sure" less 🤔

src/core/expire_period.h
romange: I know the memory usage in this PR  does not go down. It's not the goal here. I want first to introduce all the supporting functionality to support 5 bytes expire offset and then eliminate the expire table completely by moving 5 byte expire value into PrimeValue. This will reduce the memory usage for  TTL enabled use-cases but will increase it for use-cases that do not use TTLs. This is why I want to use 5 bytes for expire value so we could reduce the negative impact for these use-cases.

In order to support a 5-bytes expire offset, we must model the absolute time as `base + period` because we must update base from time to time, otherwise after 49 days of process being run you won't be able to represent a millisecond precision with the old `base`. And since updating the base needs to be done gradually (by traversing the whole table), we need to have multiple bases to support the transition state.

src/core/expire_period.h
romange: `max64/2` is arbitrary, I could also check `max64/1024`. the goal is to prevent from `(ms + 500) ` to overflow when we check if https://github.com/dragonflydb/dragonfly/pull/5797/files#diff-260e4d827ec66db63761ebe27e45792b6e2c8f9262f22d4a07e567c3fbfb5e40R61 fits uint32_t

src/server/db_slice.h
```diff
@@ -607,11 +607,11 @@ class DbSlice {
 
   ShardId shard_id_;
   uint8_t cache_mode_ : 1;
-
+  uint8_t expire_allowed_ : 1;
+  uint8_t expire_base_index_ : 1;
```
dranikpg, line 616: why do you like this so much 😆

src/core/expire_period.h
dranikpg: It's just a little less intuitive

dranikpg (approved):

PR #5798, opened 2025-09-11
fix: stale pointer reference during slot migration and flush

romange:
Also, some code clean ups.

### DIFF
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -1054,7 +1054,7 @@ OpResult<int64_t> DbSlice::UpdateExpire(const Context& cntx, Iterator prime_it,
     Del(cntx, prime_it);
     return -1;
   } else if (IsValid(expire_it) && !params.persist) {
-    auto current = ExpireTime(expire_it);
+    int64_t current = ExpireTime(expire_it->second);
     if (params.expire_options & ExpireFlags::EXPIRE_NX) {
       return OpStatus::SKIPPED;
     }
@@ -1181,9 +1181,10 @@ bool DbSlice::CheckLock(IntentLock::Mode mode, DbIndex dbid, uint64_t fp) const 
   return true;
 }
 
-void DbSlice::PreUpdateBlocking(DbIndex db_ind, Iterator it) {
-  CallChangeCallbacks(db_ind, ChangeReq{it.GetInnerIt()});
-  it.GetInnerIt().SetVersion(NextVersion());
+void DbSlice::PreUpdateBlocking(DbIndex db_ind, const Iterator& it) {
+  CallChangeCallbacks(db_ind, ChangeReq{it.GetInnerIt()});  // blocking point.
+  auto inner_it = it.GetInnerIt();                          // must call again to launder.
+  inner_it.SetVersion(NextVersion());
 }
 
 void DbSlice::PostUpdate(DbIndex db_ind, std::string_view key) {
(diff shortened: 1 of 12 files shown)


src/server/db_slice.cc
```diff
@@ -1181,9 +1181,10 @@ bool DbSlice::CheckLock(IntentLock::Mode mode, DbIndex dbid, uint64_t fp) const
   return true;
 }
 
-void DbSlice::PreUpdateBlocking(DbIndex db_ind, Iterator it) {
-  CallChangeCallbacks(db_ind, ChangeReq{it.GetInnerIt()});
-  it.GetInnerIt().SetVersion(NextVersion());
+void DbSlice::PreUpdateBlocking(DbIndex db_ind, const Iterator& it) {
+  auto inner_it = it.GetInnerIt();
+  CallChangeCallbacks(db_ind, ChangeReq{inner_it});
+  inner_it.SetVersion(NextVersion());
```
dranikpg: Are you sure you never need the iterators laundering safety net here?

romange: ah, I missed that, thanks for noting.

src/server/generic_family.cc
```diff
@@ -87,6 +87,13 @@ std::optional<RdbVersion> GetRdbVersion(std::string_view msg, bool ignore_crc =
   return version;
 }
 
+template <typename It> time_t GetExpireTime(const DbSlice& db_slice, const It& exp_it) {
+  if (!IsValid(exp_it))
+    return 0;
```
BorysTheDev, line 92: for denset_set we use MAX_INT as I remember. Can it be consistent?

src/server/generic_family.cc
```diff
@@ -836,7 +843,7 @@ OpResult<uint64_t> OpExpireTime(Transaction* t, EngineShard* shard, string_view
   if (!IsValid(expire_it))
     return OpStatus::SKIPPED;
 
-  int64_t ttl_ms = db_slice.ExpireTime(expire_it);
+  int64_t ttl_ms = db_slice.ExpireTime(expire_it->second);
```
BorysTheDev, line 846: we use time_t and int64_t. I suggest use int64_t everywhere

src/server/stream_family.h
```diff
@@ -5,6 +5,7 @@
 #pragma once
 
 #include "server/common.h"
+#include "server/table.h"
```
BorysTheDev: I would prefer forward declaration

src/server/generic_family.cc
romange: I do not understand the comment. it's a different code and interface. what consistency do you expect here?

src/server/stream_family.h
romange: reverted

src/server/generic_family.cc
BorysTheDev: I mean, we use MAX_INT in the dense_set if the expiration isn't set, it will be more consistent if we do the same here, or vice versa, one approach is better than 2 different

BorysTheDev (dismissed):

src/server/journal/streamer.cc
```diff
@@ -378,8 +378,11 @@ void RestoreStreamer::Run() {
 
   PrimeTable::Cursor cursor;
   uint64_t last_yield = 0;
-  PrimeTable* pt = &db_array_[0]->prime;
 
+  // Explicitly copy table smart pointer to keep reference count up (flushall drops it)
+  boost::intrusive_ptr<DbTable> table = db_array_.front();
+  PrimeTable* pt = &table->prime;
+  ExpireTable& expire_table = table->expire;
```
dranikpg, line 385: I remember we had the same with flush slots fiber, but I didn't think about checking other places as well 😞

dranikpg (approved):

src/server/journal/streamer.cc
romange: Due to your fix, I knew what the problem was.

PR #5800, opened 2025-09-13
chore: Command category cleanup (part 2?)

dranikpg:
1. Switch to implicit acl with list_family
2. Switch to implicit acl with set_family
3. Remove CO::SLOW completely (it's always !CO::FAST), add it to output manually

(1) and (2) also had a few differences, checked them all with valkey docs to be on the right side. Some commands were wrongly marked as "read", some as "fast" or "slow" when marked oppositely in other docs.
[diff.txt](https://github.com/user-attachments/files/22309873/diff.txt)


src/server/main_service.cc
```diff
@@ -811,6 +811,55 @@ void SetHuffmanTable(const std::string& huffman_table) {
   }
 }
 
+optional<string_view> CommandOptName(CO::CommandOpt opt, bool enabled) {
```
kostasrim:
```suggestion
optional<string> CommandOptName(CO::CommandOpt opt, bool enabled) {
```

kostasrim: Or even better plain `string` because:

`empty string` equivalent `empty optional`

kostasrim (approved):

PR #3953, opened 2024-10-20
chore: remove old io

dranikpg:
Remove old IO
* Remove SinkReplyBuilder, RedisReplyBuilder, MCReplyBuilder, replace them with new __2 version
* Remove MGetResponse and SendScoredArray 
* Simplify to use new small interface
    * CaptruingReplyBuilder
    * HttpReplyBuilder
    * InterpreterReplier

PR #5801, opened 2025-09-13
chore: Clean up string_family

dranikpg:


src/server/string_family.cc
```diff
@@ -52,29 +52,15 @@ using CI = CommandId;
...
 
 // Either immediately available value or tiering future + result
 template <typename T> using TResultOrT = variant<T, TieredStorage::TResult<T>>;
 using StringResult = TResultOrT<string>;
 
-void CopyValueToBuffer(const PrimeValue& pv, char* dest) {
-  DCHECK_EQ(pv.ObjType(), OBJ_STRING);
-  DCHECK(!pv.IsExternal());
-  pv.GetString(dest);
-}
```
dranikpg, line 67: `pv.GetString` calls `CHECK(!external)` , the type is checked internally as well - this wrapper is redundant

src/server/string_family.cc
```diff
@@ -52,42 +52,21 @@ using CI = CommandId;
...
-  string res;
-  DCHECK_EQ(pv.ObjType(), OBJ_STRING);
-
-  if (pv.ObjType() != OBJ_STRING)
-    return res;
-  res.resize(pv.Size());
-  CopyValueToBuffer(pv, res.data());
-
-  return res;
-}
```
dranikpg, line 79: Same here:
* we always assert the type before calling this function
* we assert it with DCHECK and then do if (!=) return
* `new string, resize, CopyValueToBuffer` is just `PrimveValue::ToString()`

-> redundant

src/server/string_family.cc
```diff
@@ -1413,30 +1361,12 @@ void StringFamily::MGet(CmdArgList args, const CommandContext& cmnd_cntx) {
...
-  //    never trigger.
-  // 2. For Redis:
-  //    * Call to StartArray() is safe because it calls RedisReplyBuilder::StartCollection which
-  //      calls CapturingReplyBuilder::StartCollection
-  //    * Calls to SendBulkString() and SendNull() find and if builder is CapturingReplyBuilder
-  //      then the right member gets called.
-  //
-  // Finally, the ReplyScope will trigger a Flush() on scope's end. What that means is,
-  // for CapturingReplyBuilder the internal vec is empty and therefore we should skip the call
-  // to Send because sink_ is nullptr and there is no payload to Send since it was captured.
```
dranikpg, line 1435: I tried to understand this long comment and even opened the PR that added it (it made empty flushes skip writing), yet I still don't understand why it's this explicit and why it's here. Most importantly it doesn't mention how this code could be unsafe?
1. It's not the only case where we have the same command for memcached and redis. All of them work with squashing, because we never squash with memcached. Command level code shouldn't be concerned with such details
2. It's safe to call `StartArray()` followed by `StartBulkString()`... Like half of our commands do? 🙂 🤔 And we use ReplyScopes everywhere now

So it seems to me that it raises more questions than it answers, distracting from command level code

kostasrim: Yes I wrote the comment and it's misleading 🤦 

Basically it explains why squashing is disabled for memcache (redundant for no reason) + the actual issue in the most `non intuitive way` I could possibly think of (apparently :( )

So what happened here as far as I remember is that you removed https://github.com/dragonflydb/dragonfly/pull/3953/files#diff-679cf896d933163d3fd05ceaadba85581f25b421d4a316daa555c6ce2fa4b5adL26

and added an explicit use of `SinkReplyBuilder::ReplyScope scope(builder);`

When `mget` is called from squashing context, we always `capture the replies`. CapturingReplyBuilder does not contain an active sink, that is, `sink == nullptr` and obviously the `ReplyBuilder::vec_` is empty (because we capture the replies in the stack instead of batching them in the iovecs). When `ReplyScope` went out of scope, it was forced to `Flush()`, the `vec_` was empty and `sink == nullptr because of CapturingReplyBuilder`

kostasrim: If something doesn't make sense or think it's wrong plz reply here -- I would like to be sure what happened 😄

src/server/string_family.cc
```diff
@@ -52,42 +52,21 @@ using CI = CommandId;
...
-                         : StringResult{GetString(pv)};
+                         : StringResult{pv.ToString()};
 }
 
 // Helper for performing SET operations with various options
 class SetCmd {
  public:
-  explicit SetCmd(OpArgs op_args, bool manual_journal)
-      : op_args_(op_args), manual_journal_{manual_journal} {
+  explicit SetCmd(OpArgs op_args, bool explicit_journal)
```
kostasrim, line 68: also `journal_rewrite` is a nice name here. plz, don't even bother.

PR #5802, opened 2025-09-14
chore(tiering): Fix asserts around reads for serialization

### LINKED ISSUE #5794: tiering: check ioresult status before dereferencing it
Found the issues around calls to `tiered_storage()->Read` for example in `OpDump` but also in other places.

dranikpg:
Fixes #5794


dranikpg:
I will add better tiering tests once I bundled https://github.com/dranikpg/brokenfuse into a pypackage

Issue #5564, opened 2025-07-24
Tiering: Improve error path when max file size is reached


dranikpg:
Currently, DF tries endlessly to grow the file. Either we never retry again after the file failed to grow, or try after a long period of time (in case more space was freed up).

PR #5803, opened 2025-09-14
fix(tiering): Cooldown grow grow retries

dranikpg:
Whenever we run out of space, we're stuck in a loop that constantly tries to fallocate more space for the backing file. This makes it try at most every 100 milliseconds.

I still don't like this solution, because we go every time through the `TryStash()` -> `OpManager::Stash()` -> `DiskStorage::Stash()` -> `DiskStorage::TryGrow()` chain, creating temporary values, polluting stats, etc. Instead, disk storage should provide a `max_real_file_size` that indicates the maximum really "reachable" file size, so `TieredStorage::TryStash` would adjust accordingly. But we still want to keep an expansion mechanism, so once space is freed on the device, it would notice and start growing again. So we either try growing every nth time (which breaks time dependence) or... keep a background routine?


dranikpg:
Fixes #5564

PR #5805, opened 2025-09-15
ci: enable repeat tests to use release and force epoll

abhijat:


.github/workflows/repeat-tests.yml
```diff
@@ -52,15 +62,24 @@ jobs:
...
         run: |
-          # -no-pie to disable address randomization so we could symbolize stacktraces
-          cmake -B ${GITHUB_WORKSPACE}/build -DCMAKE_BUILD_TYPE=${{matrix.build-type}} -GNinja \
-                -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DPRINT_STACKTRACES_ON_SIGNAL=ON \
-                -DCMAKE_CXX_FLAGS=-no-pie -DHELIO_STACK_CHECK:STRING=4096
-
-          cd ${GITHUB_WORKSPACE}/build  && ninja dragonfly
-          pwd
-          ls -l ..
+          if [[ "${{ inputs.use_release }}" == "yes" ]]; then
```
kostasrim, line 67: You can filter on `step level` with `if:`

```
- name: Configure & Build
  if: ....
```

And that way it's easier to follow which step gets executes

But it doesn't worth a second commit 😛 When you make another change just recall this comment (even if you don't who cares -- it's still stimple enough)

abhijat: this is nice, I will create a follow up PR, as the name of the step is misleading right now

abhijat: I added it here https://github.com/dragonflydb/dragonfly/pull/5824

PR #5815, opened 2025-09-16
chore: support huffman encoding for string values

romange:
1. Differrentiate huffman encoding for strings and values.
2. Fix "debug compression" to set huffman table for both domains.

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -367,8 +367,8 @@ struct TL {
   Huffman huff_keys, huff_string_values;
   uint64_t huff_encode_total = 0, huff_encode_success = 0;  // success/total metrics.
 
-  const HuffmanDecoder& GetHuffmanDecoder(bool is_key) const {
-    return is_key ? huff_keys.decoder : huff_string_values.decoder;
+  const HuffmanDecoder& GetHuffmanDecoder(uint8_t huffman_domain) const {
+    return huffman_domain == CompactObj::HUFF_KEYS ? huff_keys.decoder : huff_string_values.decoder;
   }
 };
 
@@ -1146,7 +1146,7 @@ void CompactObj::GetString(char* dest) const {
         next += slices[0].size() - 1;
         memcpy(next, slices[1].data(), slices[1].size());
         string_view src(reinterpret_cast<const char*>(tl.tmp_buf.data()), tl.tmp_buf.size());
-        const auto& decoder = tl.GetHuffmanDecoder(is_key_);
+        const auto& decoder = tl.GetHuffmanDecoder(huffman_domain_);
         CHECK(decoder.Decode(src, decoded_len, dest));
         return;
       }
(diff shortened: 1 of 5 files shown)


src/core/compact_object.cc
```diff
@@ -1455,34 +1463,38 @@ void CompactObj::EncodeString(string_view str) {
...
+        // 2. allows us to store the encoded string in the inline buffer
+        if (dest_len && (dest_len < kInlineLen || (dest_len + dest_len / 5) < str.size())) {
+          huff_encoded = true;
+          tl.huff_encode_success++;
+          encoded = string_view{reinterpret_cast<char*>(tl.tmp_buf.data()), dest_len + 1};
+          unsigned delta = str.size() - dest_len;
+          DCHECK_LT(delta, 256u);
+          tl.tmp_buf[0] = static_cast<uint8_t>(delta);
+          mask_bits_.encoding = HUFFMAN_ENC;
+          is_key_ = is_key;
```
romange: same code, just indented. also set `is_key_ = is_key;` here and selected huffman encoder accordingly

dranikpg: wouldn't hurt to make it a little smaller and more readable in the future 😅

src/core/compact_object.h
```diff
@@ -544,7 +546,14 @@ class CompactObj {
   };
 
   // We currently reserve 5 bits for tags and 3 bits for extending the mask. currently reserved.
-  uint8_t taglen_ = 0;
+  union {
+    uint8_t tagbyte_ = 0;
```
dranikpg, line 550: Finally, I hated the taglen_ name 🤓

src/core/compact_object.h
```diff
@@ -147,12 +147,14 @@ class CompactObj {
...
     friend class CompactObj;
-    explicit StrEncoding(uint8_t enc) : enc_(static_cast<EncodingEnum>(enc)) {
+    explicit StrEncoding(uint8_t enc, bool is_key)
+        : enc_(static_cast<EncodingEnum>(enc)), is_key_(is_key) {
     }
 
     size_t DecodedSize(size_t compr_size, uint8_t first_byte) const;
 
     EncodingEnum enc_;
+    bool is_key_;
```
dranikpg, line 157: The alternative is keeping two different enum values for huffman encoding - like we do with ascii pack, depending on the dat size. Maybe it's cleaner this way?

dranikpg (approved):
Only my last comment is relevant: Maybe it's cleaner to keep a separate enum value for different huffman encoders, instead of adding another dimension with a is_key variable?

src/core/compact_object.h
romange: sure.

src/core/compact_object.h
```diff
@@ -135,7 +135,7 @@ class CompactObj {
     NONE_ENC = 0,
     ASCII1_ENC = 1,
     ASCII2_ENC = 2,
-    HUFFMAN_ENC = 3,  // TBD
+    HUFFMAN_ENC = 3,
```
dranikpg, line 138: I had in mind that HUFFMAN_ENC splits into HUFFMAN_KEYS_ENC and HUFFMAN_VAL_ENC, so you don't need any new variables at all

dranikpg (approved):

src/core/compact_object.h
romange: we do not have enough bits for that

romange: in fact, when I think of it, is_keys_ was better because the only ambiguity is between keys and string values. for any other data types we have the type/encoding to disambiguate.

dranikpg: Ah, crap, we really have only two bits

PR #5818, opened 2025-09-17
fix(server): Use correct messages for sharded pubsub

### LINKED ISSUE #5738: wrong push types for SSUBSCRIBE/SUNSUBSCRIBE
- SSUBSCRIBE/SUNSUBSCRIBE acks use "subscribe"/"unsubscribe" instead of "ssubscribe"/"sunsubscribe".
- Messages delivered to SSUBSCRIBE subscribers use "message" instead of "smessage".

One terminal:
```
redis-cli -3
SSUBSCRIBE c1
# Expected: 1) "ssubscribe" ... ; Got: 1) "subscribe" ...
SUNSUBSCRIBE c1
# Expected: 1) "sunsubscribe" ... ; Got: 1) "unsubscribe" ...
```

Two terminals:
Terminal A:
```
redis-cli -3
SSUBSCRIBE c2
```

Terminal B:
```
redis-cli -3
SPUBLISH c2 "hi"
```
On Terminal A: Expected first element "smessage"; Got "message"

dranikpg:
* Refactor CommandId/CommanRegistry to unifromly determine the command "kind" and expose it via limited interface. Previously we used different approaches up to the authors liking
* Introduce `is_sharded` field to `PubMessage` to choose between "message" and "smessage" when replying
* Add simple "pubsub" test in cluster_family_test (not best place, but there was one before it 🤷🏻‍♂️ )

Fixes #5738

### DIFF
--- a/src/server/command_registry.h
+++ b/src/server/command_registry.h
@@ -55,17 +55,14 @@ enum CommandOpt : uint32_t {
   IDEMPOTENT = 1U << 18,
 };
 
-constexpr inline bool IsEvalKind(std::string_view name) {
-  return name.compare(0, 4, "EVAL") == 0;
-}
+enum class PubSubKind : uint8_t { REGULAR = 0, PATTERN = 1, SHARDED = 2 };
 
-constexpr inline bool IsTransKind(std::string_view name) {
-  return (name == "EXEC") || (name == "MULTI") || (name == "DISCARD");
-}
-
-static_assert(IsEvalKind("EVAL") && IsEvalKind("EVAL_RO") && IsEvalKind("EVALSHA") &&
-              IsEvalKind("EVALSHA_RO"));
-static_assert(!IsEvalKind(""));
+// Commands controlling any multi command execution.
+// They often need to be handled separately from regular commands in many contexts
+enum class MultiControlKind : uint8_t {
+  EVAL,  // EVAL, EVAL_RO, EVALSHA, EVALSHA_RO
+  EXEC,  // EXEC, MULTI, DISCARD
+};
 
 };  // namespace CO
 
@@ -114,6 +111,7 @@ class CommandId : public facade::CommandId {
   // server_state.h)
   CommandId(const char* name, uint32_t mask, int8_t arity, int8_t first_key, int8_t last_key,
             std::optional<uint32_t> acl_categories = std::nullopt);
+
   CommandId(CommandId&& o) = default;
 
   ~CommandId();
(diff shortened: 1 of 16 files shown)


src/server/main_service.cc
```diff
@@ -1087,17 +1087,16 @@ void Service::Shutdown() {
 }
 
 OpResult<KeyIndex> Service::FindKeys(const CommandId* cid, CmdArgList args) {
-  if (!cid->IsShardedPSub()) {
-    return DetermineKeys(cid, args);
-  }
```
dranikpg: If that's a top level if, then something's very odd. I understand that it breaks the rules, but we always properly work with abstractions to make the code simpler, see no-key-transactional commands, etc 🙄

src/server/command_registry.h
```diff
@@ -55,17 +55,14 @@ enum CommandOpt : uint32_t {
...
-constexpr inline bool IsTransKind(std::string_view name) {
-  return (name == "EXEC") || (name == "MULTI") || (name == "DISCARD");
-}
-
-static_assert(IsEvalKind("EVAL") && IsEvalKind("EVAL_RO") && IsEvalKind("EVALSHA") &&
-              IsEvalKind("EVALSHA_RO"));
-static_assert(!IsEvalKind(""));
+// Commands controlling any multi command execution.
+// They often need to be handled separately from regular commands in many contexts
+enum class MultiControlKind : uint8_t {
```
BorysTheDev, line 62: maybe let's add const NONE and you don't need optional in this case

src/server/command_registry.h
```diff
@@ -182,7 +180,19 @@ class CommandId : public facade::CommandId {
...
+  }
+
+  // Returns value if this command controls multi command execution (EVAL, EXEC & helpers)
+  std::optional<CO::MultiControlKind> MultiControlKind() const {
+    return kind_multi_ctr_;
+  }
+
  private:
+  std::optional<CO::PubSubKind> kind_pubsub_;
+  std::optional<CO::MultiControlKind> kind_multi_ctr_;
```
BorysTheDev, line 194: consider use special value instead of optional

src/server/main_service.cc
```diff
@@ -1087,17 +1087,15 @@ void Service::Shutdown() {
...
 
 OpResult<KeyIndex> Service::FindKeys(const CommandId* cid, CmdArgList args) {
-  if (!cid->IsShardedPSub()) {
-    return DetermineKeys(cid, args);
+  // Sharded pub-sub acts as if it's sharded by its channel name (just for checks)
+  if (cid->PubSubKind() == CO::PubSubKind::SHARDED) {
+    // SPUBLISH has only one key, the rest is data
+    if (cid->name() == registry_.RenamedOrOriginal("SPUBLISH"))
+      return KeyIndex(0, 1);
+    return {KeyIndex(0, args.size())};  // sub/unsub list of channels
```
BorysTheDev, line 1095: auto last_idx = cid->name() == registry_.RenamedOrOriginal("SPUBLISH") ? 1 : args.size();
return KeyIndex(0, last_idx);

src/server/command_registry.h
dranikpg: I like it more when nullability is an extension of the type, not part of it

src/server/main_service.cc
dranikpg: It's more readable as a seaprate if, especially with comments above

PR #5820, opened 2025-09-17
chore: skip arguments when logging failed auth cmd

### LINKED ISSUE #5808: Plain text passwords are written to the logs
**Describe the bug**
Currently, when an invalid password is used to connect to dragonfly, the password is written in plain text to the Kubernetes Pod logs. I'm using the dragonfly helm chart in a Kubernetes environment.

**To Reproduce**
1. Use a Pod that features the `nc` binary and send a request with an invalid password to dragonfly:  
`(printf "AUTH myinvalidpassword\r\n";) | nc dragonfly.dragonfly 6379`
2. Check the logs of the `dragonfly-0` Pod:  
`main_service.cc:1604]  AUTH myinvalidpassword failed with reason: -WRONGPASS invalid username-password pair or user is disabled.`

This also applies when a real client sends an incorrect password to the database, e.g. using a Kubernetes Secret.

 - OS: Kubernetes `v1.30.11`
 - Kernel: I don't have access
 - Containerized?: Kubernetes
 - Dragonfly Version: `v1.33.1`

**Additional context**  
In my opinion, passwords (and Kubernetes Secrets) should never be written to Pod logs in plain text. Instead, they should be redacted in the logs.

kostasrim:
We should not log the arguments for AUTH commands

fixes #5808

### DIFF
--- a/src/server/main_service.cc
+++ b/src/server/main_service.cc
@@ -723,8 +723,10 @@ string FailedCommandToString(std::string_view command, facade::CmdArgList args,
   string result;
   absl::StrAppend(&result, " ", command);
 
-  for (auto arg : args) {
-    absl::StrAppend(&result, " ", absl::CHexEscape(arg));
+  if (command != "AUTH") {
+    for (auto arg : args) {
+      absl::StrAppend(&result, " ", absl::CHexEscape(arg));
+    }
   }
 
   absl::StrAppend(&result, " failed with reason: ", reason);


src/server/main_service.cc
```diff
@@ -723,8 +723,10 @@ string FailedCommandToString(std::string_view command, facade::CmdArgList args,
   string result;
   absl::StrAppend(&result, " ", command);
 
-  for (auto arg : args) {
-    absl::StrAppend(&result, " ", absl::CHexEscape(arg));
+  if (command != "AUTH") {
```
kostasrim: This also skips the user name but it's fine to not log it 🤷

dranikpg (dismissed):
What about ACL commands setting passwords, etc?

kostasrim:
> What about ACL commands setting passwords, etc?

+1

dranikpg:
Dragonfly error log considering which commands to filter out

image

kostasrim:
> Dragonfly error log considering which commands to filter out
> 

That was me 🤣

dranikpg (approved):

Issue #5581, opened 2025-07-28
Implement FLUSHDB ASYNC/SYNC


moredure:
**Did you search GitHub Issues and GitHub Discussions First?**
Yes

**Is your feature request related to a problem? Please describe.**
FLUSDB ASYNC/SYNC fails with wrong number of args on v1.31.0

**Describe the solution you'd like**
https://valkey.io/commands/flushdb/ support sync/async

kostasrim:
yeah we don't support `async` subcommand for that

dranikpg:
but it's async by default 🙂 so we don't support sync

PR #5821, opened 2025-09-17
fix(server): Support FLUSH(ALL) SYNC

dranikpg:
Add SYNC option to FLUSHDB and FLUSHALL. SYNC doens't block other commands, it just waits for the flush to finish. The pytest asserts that immediately after FLUSHALL SYNC we have free RSS memory

Fixes https://github.com/dragonflydb/dragonfly/issues/5581

### DIFF
--- a/src/server/db_slice.h
+++ b/src/server/db_slice.h
@@ -331,7 +331,7 @@ class DbSlice {
   constexpr static DbIndex kDbAll = 0xFFFF;
 
   // Flushes db_ind or all databases if kDbAll is passed
-  void FlushDb(DbIndex db_ind);
+  util::fb2::JoinHandle FlushDb(DbIndex db_ind);
 
   // Flushes the data of given slot ranges.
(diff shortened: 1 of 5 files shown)


dranikpg:
image

src/server/db_slice.h
```diff
@@ -331,7 +331,7 @@ class DbSlice {
   constexpr static DbIndex kDbAll = 0xFFFF;
 
   // Flushes db_ind or all databases if kDbAll is passed
-  void FlushDb(DbIndex db_ind);
+  util::fb2::JoinHandle FlushDb(DbIndex db_ind);
```
romange: why could not you return a fiber? why do you need a handle?

dranikpg: Ahhhh... More elegant interface because keeping the default behaviour doesn't require any action from callers. But I've just implemented it without and it's fine

dranikpg: https://github.com/romange/helio/pull/468

romange (approved):

PR #5823, opened 2025-09-18
pytest: Retry set until failure after ACL update

abhijat:
In test_acl_cat_commands_multi_exec_squash we set an ACL to deny set commands, and then check if subsequent set command fails. However this can fail because of time taken for the ACL setuser to propagate changes to shards, during this time a set command may succeed.

The test is changed to retry the command a fixed number of times until eventually it is denied.

fixes https://github.com/dragonflydb/dragonfly/issues/3719

### DIFF
--- a/tests/dragonfly/acl_family_test.py
+++ b/tests/dragonfly/acl_family_test.py
@@ -1,13 +1,9 @@
-import pytest
-import redis
-from redis import asyncio as aioredis
-from .instance import DflyInstanceFactory
-from .utility import *
 import tempfile
-import asyncio
-import os
-from . import dfly_args
+
 import async_timeout
+
+from . import dfly_args
+from .utility import *
 
 
 @pytest.mark.asyncio
@@ -141,7 +137,6 @@ async def test_acl_commands(async_client):
         await async_client.execute_command("ZADD myset 1 two")
 
 
-@pytest.mark.exclude_epoll  # Failing test. It should be turned on as soon as it is fixed.
 @pytest.mark.asyncio
 async def test_acl_cat_commands_multi_exec_squash(df_factory):
     df = df_factory.create(multi_exec_squash=True, port=1111)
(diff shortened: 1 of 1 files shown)


tests/dragonfly/acl_family_test.py
```diff
@@ -234,8 +229,18 @@ async def test_acl_cat_commands_multi_exec_squash(df_factory):
...
+    denied = False
+    while not denied and time.time() - start < 10:
+        try:
+            await client.execute_command(f"SET x{x} {x}")
+            asyncio.sleep(0.1)
+        except redis.exceptions.NoPermissionError:
+            denied = True
+        except Exception as e:
+            assert False, f"failed with unexpected error: {e}"
+    assert denied, "all SET commands succeeded unexpectedly defying ACL"
```
abhijat, line 243: I don't know if there is some helper already that makes this boilerplate easier

tests/dragonfly/acl_family_test.py
```diff
@@ -234,8 +229,18 @@ async def test_acl_cat_commands_multi_exec_squash(df_factory):
     # NOPERM while executing multi
     await client.execute_command("MULTI")
 
-    with pytest.raises(redis.exceptions.NoPermissionError):
-        await client.execute_command(f"SET x{x} {x}")
+    # retry for a few seconds while the ACL SETUSER propagates, some SET commands might get through
+    start = time.time()
+    denied = False
+    while not denied and time.time() - start < 10:
```
vyavdoshenko, line 235: Will this test hang if Dragonfly crashes?

abhijat: I don't think so, for two reasons:

* if dfly crashes, client cannot connect and we hit the other exception branch which asserts false
* even if that doesn't happen, after 10 seconds `time.time() > 10 + start` and we would exit the loop

Let me know if there is something I overlooked though.

vyavdoshenko (commented):
lgtm

tests/dragonfly/acl_family_test.py
kostasrim: we do, check `assert_eventually`

kostasrim: `dragonfly/utility.py`

PR #5824, opened 2025-09-18
ci: Allow overriding log level for modules when repeating tests

abhijat:
Allows setting vmodule level for tests, eg:

```
gh workflow run "Repeat Tests" --ref abhijat/feat/vmodule-on-repeat-tests -f branch=abhijat/feat/vmodule-on-repeat-tests -f count=500 -f expression="-xv dragonfly -k test_client_kill" -f timeout=300m -f epoll=epoll -f vmodule_expression="server_family=2,dragonfly_connection=2,listener_interface=2"
```

note `-f vmodule_expression="server_family=2,dragonfly_connection=2,listener_interface=2"`

Also clean up to use `if` expression suggested in https://github.com/dragonflydb/dragonfly/pull/5805#discussion_r2349348492

### DIFF
--- a/.github/actions/repeat/action.yml
+++ b/.github/actions/repeat/action.yml
@@ -32,6 +32,9 @@ inputs:
   epoll:
     required: true
     type: string
+  vmodule_expression:
+    required: true
+    type: string
 
 runs:
   using: "composite"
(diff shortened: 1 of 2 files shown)


.github/actions/repeat/action.yml
```diff
@@ -52,8 +55,13 @@ runs:
         else
           FORCE_EPOLL=""
         fi
-        echo Running command: timeout ${{ inputs.timeout }} pytest ${{ inputs.expression }} --drop-data-after-each-test ${FORCE_EPOLL} --color=yes --json-report --json-report-file=report.json --log-cli-level=DEBUG --count=${{ inputs.count }}
-        timeout ${{ inputs.timeout }} pytest ${{ inputs.expression }} --drop-data-after-each-test ${FORCE_EPOLL} --color=yes --json-report --json-report-file=report.json --log-cli-level=DEBUG --count=${{ inputs.count }} || code=$?
+        if [[ $"{{ inputs.vmodule_expression }}" != "" ]]; then
+          VMOD="--df vmodule=${{ inputs.vmodule_expression }}"
```
kostasrim, line 59: isn't it `--vmodule` instead of `--df vmodule` ?

kostasrim (commented):
I missed this one accidentally, apologies for the late review

.github/actions/repeat/action.yml
abhijat: Yes for dragonfly the arg should be `--vmodule`.

This arg however is passed to the pytest arg parser, where it will be converted into the correct format before being passed to dragonfly https://github.com/dragonflydb/dragonfly/blob/3b5628dd327591c19ba678bcd794f580146869f6/tests/dragonfly/instance.py#L366

PR #5826, opened 2025-09-18
core: use explicit types and minor cleanups

abhijat:
Minor changes and cleanups:

* use explicit types
* mark move ctors noexcept
* use byte size int for enum variants
* remove two unimplemented methods in node

### DIFF
--- a/src/core/qlist.h
+++ b/src/core/qlist.h
@@ -194,7 +194,7 @@ class QList {
 
   // Returns true if elements were deleted, false if list has not changed.
   // Negative start index is allowed.
-  bool Erase(const long start, unsigned count);
+  bool Erase(long start, unsigned count);
 
   // Needed by tests and the rdb code.
   const Node* Head() const {


src/core/qlist.h
```diff
@@ -33,9 +33,9 @@ class QList {
    * extra: 9 bits, free for future use; pads out the remainder of 32 bits
    * */
 
-  typedef struct Node {
-    struct Node* prev;
-    struct Node* next;
+  struct Node {
+    Node* prev;
+    Node* next;
```
abhijat, line 38: I'm not sure if this change is exactly equivalent to the original code, because I'm not sure what the `typedef` was for.

src/core/qlist.cc
```diff
@@ -429,7 +429,7 @@ QList::~QList() {
   Clear();
 }
 
-QList& QList::operator=(QList&& other) {
+QList& QList::operator=(QList&& other) noexcept {
```
abhijat, line 432: this is a little tricky because we call `Clear`, but the main actions taken in there are `mimalloc` methods which are all marked `noexcept`. Perhaps `Clear` should be `noexcept` too.

dranikpg: Good catch, the move operations need to be noexcept to efficiently work with stl code. Nothing tricky, you can mark Clear() as noexcept to adhere to good style, though nothing will technically change

src/core/qlist.h
dranikpg: Because it was copy-pasted from c code

dranikpg (approved):
I have one more clang-tidy warning around `bool Erase(const long, ...`, would be cool if you can handle it as well in that PR

src/core/qlist.h
romange: it was a copy paste from the c code. see the comment above. it's fine to change it now.

romange (approved):

abhijat:
> I have one more clang-tidy warning around `bool Erase(const long, ...`, would be cool if you can handle it as well in that PR

sure, done

dranikpg (approved):

PR #5827, opened 2025-09-18
fix(server): Fix AllocatingTracker crash due to VLOG

### LINKED ISSUE #4438: Investigate memory tracker crash
To reproduce:

1. `python3 -m pytest dragonfly/replication_test.py --df allocation_tracker=64:88285184 -k "test_big_containers"`

Crashes:

```
-------------------------------------------------------------------------------------------- Captured stdout teardown ---------------------------------------------------------------------------------------------
45579➜    @     0x6397e6a1eb19         64  absl::lts_20240116::WriteFailureInfo()
45579➜    @     0x6397e6a1ed73         96  absl::lts_20240116::AbslFailureSignalHandler()
45579➜    @     0x74947ca42520  (unknown)  (unknown)


0x00006397e6a1eb19
??
??:0
0x00006397e6a1ed73
??
??:0
0x000074947ca42520
??
??:0

```

maybe somehow we stack-overflow ?

dranikpg:
Fixes #4438

It was calling VLOG which internally allocated while holding a lot of resources (like a mutex), so it makes sense that it doesn't expect at all that VLOG will be called in-between.

I suggest simply to not log anything if LogMessage::Flush is in-between, not sure if that's strict enough. Should be for the mutex, but who knows that else could break with updates. (The test works now)

### DIFF
--- a/src/core/allocation_tracker.cc
+++ b/src/core/allocation_tracker.cc
@@ -12,6 +12,12 @@ namespace dfly {
 namespace {
 thread_local AllocationTracker g_tracker;
 thread_local absl::InsecureBitGen g_bitgen;
+
+bool CanCallVlog(std::string_view trace) {
+  // GLOG fails when logging while flushing the current log under a mutex
+  return trace.find("LogMessage::Flush") == std::string::npos;
+}
+
 }  // namespace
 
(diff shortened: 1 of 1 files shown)


romange (commented):
good catch

romange (approved):

PR #5830, opened 2025-09-18
feat(search): FT.SEARCH FILTER option

### LINKED ISSUE #5822: Support FT.SEARCH FILTER option for numeric fields
We should support FILTER option in FT.SEARCH command.

Syntax:

```
FILTER numeric_field min max [ FILTER numeric_field min max ...]
```

Filters can be combined on same and different indexed numeric field.

Since we already have numeric index this feature should be prerequisite  before #4536

mkaruza:
Even it is deprecated FILTER can still be used as part of FT.SEARCH
command. Add support for it by reconstruction of AST tree.
    
Closes #5822

### DIFF
--- a/src/core/search/base.h
+++ b/src/core/search/base.h
@@ -38,6 +38,19 @@ struct QueryParams {
   absl::flat_hash_map<std::string, std::string> params;
 };
 
+// Base class for optional search filters
+
+class AstNode;
+
+struct OptionalFilterBase {
+  virtual bool IsEmpty() const = 0;
+  virtual AstNode Node(std::string field) = 0;
+  virtual ~OptionalFilterBase() = default;
+};
+
+using OptionalFilters =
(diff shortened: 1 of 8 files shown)


src/core/search/base.h
```diff
@@ -38,6 +38,19 @@ struct QueryParams {
...
+class AstNode;
+
+struct OptionalFilterBase {
+  virtual bool IsEmpty() const = 0;
+  virtual AstNode Node(std::string field) = 0;
+  virtual ~OptionalFilterBase() = default;
+};
+
+using OptionalFilters =
+    absl::flat_hash_map<std::string /*field*/, std::unique_ptr<OptionalFilterBase> /* filter */>;
```
dranikpg, line 52: I'm not sure you need all this code, you can just do it from search_family.cc

PR #5832, opened 2025-09-18
Allow snapshotting on master only

xuekat:
Introduce a `master_only_snapshot` flag that skips snapshotting on replicas if set to true

### DIFF
--- a/src/server/detail/save_stages_controller.cc
+++ b/src/server/detail/save_stages_controller.cc
@@ -15,6 +15,7 @@
 #include "server/main_service.h"
 #include "server/namespaces.h"
 #include "server/script_mgr.h"
+#include "server/server_family.h"
 #include "server/transaction.h"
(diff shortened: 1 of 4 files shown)


romange:
I would like to avoid these types of flags. why can not operator run `config set snapshot_cron ""` when it configures a node as master?

xuekat:
> Enable snapshot on master only dragonflydb/dragonfly-operator#367

I had originally tried clearing the snapshot crons in the dragonfly operator, but @Abhra303 raised [concerns with this approach](https://github.com/dragonflydb/dragonfly-operator/pull/367#discussion_r2358324417), so I thought that the best way would be to move the logic to the Dragonfly service itself. not sure if my original implementation was in line with your suggestion, though.

xuekat:
Update -- moved the snapshot clearing logic to `dragonfly-operator`. If the approach there is approved I will close this PR.

PR #5833, opened 2025-09-19
cmake: Optionally download prometheus binary as part of cmake config step

abhijat:
This PR sets up prometheus in the build directory which will be used for pytests for sanity checking metrics by consuming them.

This pr is standalone to focus on the mechanism of pulling in the prometheus binary, there are multiple possible options for this but I have used cmake and fetchcontent here.

A new cmake option: `USE_PROMETHEUS` which is off by default should be turned on only where pytests are expected to be run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,6 +125,7 @@ jobs:
             -DCMAKE_CXX_FLAGS="${{matrix.cxx_flags}}" -DWITH_AWS:BOOL=OFF \
             -DWITH_ASAN="${ASAN}" \
             -DWITH_USAN="${USAN}" \
+            -DUSE_PROMETHEUS=ON \
             -L
 
           cd ${GITHUB_WORKSPACE}/build && pwd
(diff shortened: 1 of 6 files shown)


romange:
as you said there are many ways to do it and I disagree its the right one. a quick ai chat provides a very straightforward solution:

```py
# Example using testcontainers-python
# pip install testcontainers
from testcontainers.core.container import DockerContainer
from testcontainers.core.waiting_utils import wait_for_logs

# Configure Prometheus to scrape your app (running on host)
# You would create this prometheus.yml file
# global:
#   scrape_interval: 1s
# scrape_configs:
#   - job_name: 'myapplication'
#     static_configs:
#       - targets: ['host.docker.internal:8000']

# In your test setup
prometheus_container = DockerContainer("prom/prometheus:v2.47.2")
prometheus_container.with_volume_mapping("/path/to/your/prometheus.yml", "/etc/prometheus/prometheus.yml")
prometheus_container.with_exposed_ports(9090)

# Start the container before tests
prometheus_container.start()
wait_for_logs(prometheus_container, "Server is ready to receive web requests")

# Your tests run here...

# Stop the container after tests
prometheus_container.stop()
```

abhijat:
I didn't know about this testcontainers library. Will look into it.

romange:
I did not know either, I asked gemini to give me some options on how to run prometheus in pytests

abhijat:
I have been trying this library in https://github.com/dragonflydb/dragonfly/pull/5843, after some exploration it seems to work, so I will close this PR and continue with the testcontainers approach.

PR #5835, opened 2025-09-19
fix: stack overflow in hincrbyfloat

BorysTheDev:
problem: string2ld consumes 15KB of stack
fix: use absl::SimpleAtod() to convert string to double

before fix: stack usage was 31.2KB, after fix 16.4KB

### DIFF
--- a/src/redis/util.c
+++ b/src/redis/util.c
@@ -207,33 +207,3 @@ int string2ll(const char *s, size_t slen, long long *value) {
     }
     return 1;
 }
-
-/* Convert a string into a double. Returns 1 if the string could be parsed
- * into a (non-overflowing) double, 0 otherwise. The value will be set to
- * the parsed value when appropriate.
- *
- * Note that this function demands that the string strictly represents
- * a double: no spaces or other characters before or after the string
- * representing the number are accepted. */
-int string2ld(const char *s, size_t slen, long double *dp) {
-    char buf[MAX_LONG_DOUBLE_CHARS];
-    long double value;
-    char *eptr;
-
-    if (slen == 0 || slen >= sizeof(buf)) return 0;
-    memcpy(buf,s,slen);
-    buf[slen] = '\0';
-
-    errno = 0;
-    value = strtold(buf, &eptr);
-    if (isspace(buf[0]) || eptr[0] != '\0' ||
-        (size_t)(eptr-buf) != slen ||
-        (errno == ERANGE &&
-            (value == HUGE_VAL || value == -HUGE_VAL || value == 0)) ||
-        errno == EINVAL ||
-        isnan(value))
-        return 0;
-
-    if (dp) *dp = value;
-    return 1;
-}
--- a/src/server/dfly_main.cc
+++ b/src/server/dfly_main.cc
@@ -102,10 +102,10 @@ namespace {
 // need additional 8-16 bytes for their internal structures, thus over reserving additional
 // memory pages if using round sizes.
 #ifdef NDEBUG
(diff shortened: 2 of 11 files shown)


src/redis/util.c
```diff
@@ -207,33 +207,3 @@ int string2ll(const char *s, size_t slen, long long *value) {
...
-
-/* Convert a string into a double. Returns 1 if the string could be parsed
- * into a (non-overflowing) double, 0 otherwise. The value will be set to
- * the parsed value when appropriate.
- *
- * Note that this function demands that the string strictly represents
- * a double: no spaces or other characters before or after the string
- * representing the number are accepted. */
-int string2ld(const char *s, size_t slen, long double *dp) {
-    char buf[MAX_LONG_DOUBLE_CHARS];
```
kostasrim, line 219: So it wasn't `strold` 😛 It's that `MAX_LONG_DOUBLE_CHARS = 5kb`

Just keep in mind, by replacing `strtold` with `SimpleAtod` we break compatibility with valkey/redis as now we parse certain strings differently (the two are not 1-1 compatible). If the issue is the stack size, why not allocate a temporary instead on the heap ?

BorysTheDev: this buffer is only 5KB, other part is strtold, and other part is significantly bigger

romange: 1. we need that everything `strtold` parses, `SimpleAtod` can parse as well. what are the differences between those?
2. please note that we use `fast_float::from_chars` (ParseDouble) in other places, and even in OpIncrFloat in string_family. maybe better to use the same function as it's faster.

romange: 3. We should add tests covering special values in `string2ld` in hset_family_test (i.e all those hugeval, erange, invalid strings etc).

src/server/dfly_main.cc
```diff
@@ -102,10 +102,10 @@ namespace {
 // need additional 8-16 bytes for their internal structures, thus over reserving additional
 // memory pages if using round sizes.
 #ifdef NDEBUG
-constexpr size_t kFiberDefaultStackSize = 36_KB - 16;
+constexpr size_t kFiberDefaultStackSize = 32_KB - 16;
 #else
-// Increase stack size for debug builds.
-constexpr size_t kFiberDefaultStackSize = 40_KB - 16;
+// We leave the same stack size for debug mode, to find out the possible stack overflow bugs.
+constexpr size_t kFiberDefaultStackSize = 32_KB - 16;
```
romange, line 108: did you run regression tests on this PR? I do not want to break the regression tests

BorysTheDev: not yet, will do now

BorysTheDev: https://github.com/dragonflydb/dragonfly/actions/runs/17855835231

src/redis/util.c
BorysTheDev: strtold can parse longer double that isn't supported by SimpleAtod and from_chars. Also strtold using locale, that I think we don't support. Also not all platforms support 10-16-byte double. So for consistency, I think we can use the 8byte one

romange: yes, we should support only the standard doubles.

BorysTheDev: hugeval and erange aren't special values, it's error processing. I've added tests

BorysTheDev: Also have found that we return an incorrect error for INF results in INCRBYFLOAT

src/server/dfly_main.cc
BorysTheDev: everything looks ok

Issue #5831, opened 2025-09-18
A test to validate metric names using prometheus


abhijat:
context: https://github.com/dragonflydb/dragonfly/issues/5828#issuecomment-3306655026

we should validate metrics the same way prometheus does to warn about issues during CI

PR #5843, opened 2025-09-23
tests: Add test for prometheus metrics sanity check

abhijat:
Adds a test which starts a prometheus container, the container scrapes dragonfly and runs for a few seconds.

The test confirms that there are no warning or error logs seen in the container.

FIXES https://github.com/dragonflydb/dragonfly/issues/5831

### DIFF
--- a/tests/dragonfly/requirements.txt
+++ b/tests/dragonfly/requirements.txt
@@ -27,3 +27,5 @@ pytest-timeout==2.2.0
 asyncio==3.4.3
 fakeredis[json]==2.26.2
 hiredis==2.4.0
+PyYAML>=6.0
+testcontainers>=3.7.1
--- a/tests/dragonfly/server_family_test.py
+++ b/tests/dragonfly/server_family_test.py
@@ -1,5 +1,12 @@
+import os.path
+from tempfile import NamedTemporaryFile
+from typing import Callable
+
+import yaml
 from prometheus_client.samples import Sample
 from pymemcache import Client
+from testcontainers.core.container import DockerContainer
+from testcontainers.core.waiting_utils import wait_for_logs
 
 from . import dfly_args
 from .instance import DflyInstance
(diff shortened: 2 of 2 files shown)


tests/dragonfly/server_family_test.py
```diff
@@ -250,3 +256,58 @@ async def test_latency_stats_disabled(async_client: aioredis.Redis):
...
+        )
+
+        yaml.dump(prometheus_config, temp_file, default_flow_style=None)
+        os.chmod(temp_file.name, 0o644)
+
+        with DockerContainer(
+            "prom/prometheus",
+            volumes=[(temp_file.name, prometheus_config_path, "ro")],
+            # these are passed through to the low level docker container initialization API
+            extra_hosts={"host.docker.internal": "host-gateway"},
```
abhijat: There is no documented way in this library to set up a host entry within the test container, however these args are passed through to the low level python API and it works for this case.

The java version of `testcontainers` does have a way to expose a host port to the container but there is nothing similar in the python version.

abhijat:
Testing with the diff which caused an issue recently results in the new test failing:

```
diff --git a/src/server/server_family.cc b/src/server/server_family.cc
index ec7c6d13..008e0820 100644
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -1948,7 +1948,7 @@ void PrintPrometheusMetrics(uint64_t uptime, const Metrics& m, DflyCmd* dfly_cmd
                             m.shard_stats.defrag_task_invocation_total, COUNTER, &resp->body());
   AppendMetricWithoutLabels("defrag_attempts", "Objects examined",
                             m.shard_stats.defrag_attempt_total, COUNTER, &resp->body());
-  AppendMetricWithoutLabels("defrag_objects_moved", "Objects moved",
+  AppendMetricWithoutLabels("defrag_invocations", "Objects moved",
                             m.shard_stats.defrag_realloc_total, COUNTER, &resp->body());
 }
```

results in

```
>       assert "level=WARN" not in entry, f"warning in prometheus log: {entry}"
E       AssertionError: warning in prometheus log: time=2025-09-23T11:33:54.790Z level=WARN source=scrape.go:1905 msg="Error on ingesting samples with different value but same timestamp" component="scrape manager" scrape_pool=dfly target=http://host.docker.internal:42569/metrics num_dropped=1
E       assert 'level=WARN' not in 'time=2025-0...um_dropped=1'
E         'level=WARN' is contained here:
E         ?                               ^^^^^^^^^^^^^^^^^^^^^^
E           time=2025-09-23T11:33:54.790Z level=WARN source=scrape.go:1905 msg="Error on ingesting samples with different value but same timestamp" component="scrape manager" scrape_pool=dfly target=http://host.docker.internal:42569/metrics num_dropped=1
E         ?                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

tests/dragonfly/server_family_test.py:280: AssertionError
```

abhijat:
The test seems to be running into a docker issue

```
E       docker.errors.APIError: 400 Client Error for http+docker://localhost/v1.48/containers/58a2baca757e1aef621f452e27a9bc3ae1c9e159be2731d47c56f87a60c6e399/start: Bad Request ("failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/tmp/tmpbojip3ee" to rootfs at "/etc/prometheus/prometheus.yml": create mountpoint for /etc/prometheus/prometheus.yml mount: cannot create subdirectories in "/var/lib/docker/overlay2/e294e93f76e3a550e04dcb465eed8300a20487866553a00d990ed619970d1b96/merged/etc/prometheus/prometheus.yml": not a directory: unknown: Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the expected type")
```

which does not appear locally. The temp file created for prometheus config is attempted to mount into a directory instead of a file.

abhijat:
> The test seems to be running into a docker issue
> 
> ```
> E       docker.errors.APIError: 400 Client Error for http+docker://localhost/v1.48/containers/58a2baca757e1aef621f452e27a9bc3ae1c9e159be2731d47c56f87a60c6e399/start: Bad Request ("failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/tmp/tmpbojip3ee" to rootfs at "/etc/prometheus/prometheus.yml": create mountpoint for /etc/prometheus/prometheus.yml mount: cannot create subdirectories in "/var/lib/docker/overlay2/e294e93f76e3a550e04dcb465eed8300a20487866553a00d990ed619970d1b96/merged/etc/prometheus/prometheus.yml": not a directory: unknown: Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the expected type")
> ```
> 
> which does not appear locally. The temp file created for prometheus config is attempted to mount into a directory instead of a file.

The reason this is happening because:

* the test runs in a container
* it creates file `tmp/prometheus.yml` within its filesystem
* it starts container for prometheus, mounting `tmp/...` into `/etc/prometheus/prometheus.yml`

The mount however is from the host machine to the new container. The host machine does not have the temp. file, so there is a mismatch in types within the mount.

To share files between containers I will use ~~`var/crash`~~ the first known mounted volume as a staging area because it is mounted on the container already. For local run `/tmp` is used

tests/dragonfly/server_family_test.py
```diff
@@ -250,3 +257,64 @@ async def test_latency_stats_disabled(async_client: aioredis.Redis):
...
+    def assert_no_error(entry: str):
+        assert "level=ERROR" not in entry, f"error in prometheus log: {entry}"
+        assert "level=WARN" not in entry, f"warning in prometheus log: {entry}"
+
+    # Piggyback on the first known mounted path if running in CI, the container running the test will start another
+    # container with prometheus. The prometheus container needs the file present on the host to be able to mount it.
+    # Fall back to /tmp so the test can be run on the local machine without using root.
+    parent = next((p for p in ("/var/crash", "/mnt", "/tmp") if os.access(p, os.W_OK)), None)
+
+    # TODO use python-docker api to find a valid mounted volume instead of hardcoded list
```
abhijat, line 280: This might actually not be possible, because we would have to figure out the mounts from within the container so the docker api will not be helpful. Hardcoding a list seems more appropriate.

abhijat: will explore this as a follow up

tests/dragonfly/server_family_test.py
```diff
@@ -250,3 +257,64 @@ async def test_latency_stats_disabled(async_client: aioredis.Redis):
...
+        with (
+            DockerContainer(image="prom/prometheus")
+            .with_volume_mapping(path, prometheus_config_path)
+            .with_kwargs(extra_hosts={"host.docker.internal": "host-gateway"})
+        ) as prometheus:
+            try:
+                wait_for_logs(prometheus, "Server is ready to receive web requests.", timeout=5)
+
+                # Wait for a few seconds for any potential warnings or errors to appear, it can take several seconds.
+                wait_for_errors_sec, sleep_time_sec = 10, 0.5
```
abhijat, line 309: This long wait is unfortunate but prometheus seems to take a while to start producing relevant warnings, we check with a short interval to fail as fast as possible. 

A successful test will be slow, however.

tests/dragonfly/requirements.txt
```diff
@@ -27,3 +27,5 @@ pytest-timeout==2.2.0
 asyncio==3.4.3
 fakeredis[json]==2.26.2
 hiredis==2.4.0
+PyYAML==6.0
+testcontainers==3.7.1
```
romange:
```suggestion
PyYAML>=6.0
testcontainers>=3.7.1
```

romange (approved):

Issue #4536, opened 2025-01-30
Support GEO search


BagritsevichStepan:
We need to support GEO search as [requested here](https://github.com/dragonflydb/dragonfly/issues/1995#issuecomment-1776254945).

For example:
```
127.0.0.1:6379> HSET place:1 name "Eiffel Tower" location "2.2945,48.8584"
(integer) 2
127.0.0.1:6379> HSET place:2 name "Louvre Museum" location "2.3358,48.8606"
(integer) 2
127.0.0.1:6379> HSET place:3 name "Notre-Dame" location "2.3499,48.8529"
(integer) 2
127.0.0.1:6379> FT.CREATE index ON HASH PREFIX 1 "place:" SCHEMA name TEXT location GEO 
OK
127.0.0.1:6379> FT.SEARCH index "*" GEOFILTER location 2.2945 48.8584 5 km
1) (integer) 3
2) "place:1"
3) 1) "name"
   2) "Eiffel Tower"
   3) "location"
   4) "2.2945,48.8584"
4) "place:2"
5) 1) "name"
   2) "Louvre Museum"
   3) "location"
   4) "2.3358,48.8606"
6) "place:3"
7) 1) "name"
   2) "Notre-Dame"
   3) "location"
   4) "2.3499,48.8529"
```

PR #5854, opened 2025-09-24
feat(search): Implement GEO index and support for RADIUS search

mkaruza:
Implementation of GEO index use boost::geometry::index::rtree as
structure to store/search/remove points.

Search is done in a way that we construct polygon of 4 points with 90
degrees angle difference around search point with fixed distance and bounding
box around that polygon to cover whole area of possible points that fit in
radius. For each point we calculate haversine distance between found
point and search point.

Fixes https://github.com/dragonflydb/dragonfly/issues/4536

### DIFF
--- a/src/core/search/ast_expr.h
+++ b/src/core/search/ast_expr.h
@@ -45,6 +45,13 @@ struct AstRangeNode {
   double lo, hi;
 };
 
+struct AstGeoNode {
+  AstGeoNode(double lon, double lat, double radius, std::string unit);
+  double lon, lat;
+  double radius;
+  std::string unit;
+};
+
 // Negates subtree
 struct AstNegateNode {
   AstNegateNode(AstNode&& node);
@@ -112,7 +119,7 @@ struct AstKnnNode {
 using NodeVariants =
     std::variant<std::monostate, AstStarNode, AstStarFieldNode, AstTermNode, AstPrefixNode,
                  AstSuffixNode, AstInfixNode, AstRangeNode, AstNegateNode, AstLogicalNode,
-                 AstFieldNode, AstTagsNode, AstKnnNode>;
+                 AstFieldNode, AstTagsNode, AstKnnNode, AstGeoNode>;
 
 struct AstNode : public NodeVariants {
   using variant::variant;
--- a/src/core/search/indices.cc
+++ b/src/core/search/indices.cc
@@ -10,6 +10,8 @@
 #include <absl/strings/str_join.h>
 #include <absl/strings/str_split.h>
 
+#include <boost/iterator/function_output_iterator.hpp>
+
 #define UNI_ALGO_DISABLE_NFKC_NFKD
 
 #include <hnswlib/hnswalg.h>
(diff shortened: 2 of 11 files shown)


src/core/search/search.cc
```diff
@@ -274,6 +277,15 @@ struct BasicSearch {
     return IndexResult{};
   }
 
+  IndexResult Search(const AstGeoNode& node, string_view active_field) {
+    DCHECK(!active_field.empty());
+    double converted_radius = GeoIndex::ConvertToRadiusInMeters(node.radius, node.unit);
```
dranikpg: I'd make the index do that internally

dranikpg (commented):
Looks good so far 👍🏻

src/core/search/ast_expr.h
```diff
@@ -45,6 +45,13 @@ struct AstRangeNode {
   double lo, hi;
 };
 
+struct AstGeoNode {
+  AstGeoNode(double lat, double lon, size_t radius, std::string unit);
+  double lat, lon;
+  size_t radius;
+  std::string unit;
+};
```
dranikpg, line 53: TODO: parse unit from lexer (so we can throw the error in time)

src/core/search/indices.cc
```diff
@@ -614,4 +618,84 @@ void HnswVectorIndex::Remove(DocId id, const DocumentAccessor& doc, string_view
...
+
+  double lon;
+  if (!absl::SimpleAtod(coordinates[1], &lon))
+    return false;
+
+  entry index_entry = std::make_pair(point{lat, lon}, id);
+
+  rtree_->insert(index_entry);
+
+  return true;
```
dranikpg: Just style comment how to reduce volume 😅 
```c++
// absl::StrSplit can split into a pair but you can't check there easily
double lon, lat;
if (!absl::SimpleAtod() || !absl::SimpleAtod())
  return false;
rtree_->emplace(point{}, id)
return true;
```

src/core/search/indices.cc
```diff
@@ -614,4 +618,96 @@ void HnswVectorIndex::Remove(DocId id, const DocumentAccessor& doc, string_view
...
+    return std::nullopt;
+  }
+
+  double lon;
+  if (!absl::SimpleAtod(coordinates[1], &lon)) {
+    return std::nullopt;
+  }
+
+  return point{lat, lon};
+}
```
dranikpg: Just style comment, feel free to ignore 😅 

```c++
double lon, lat;
if (!absl::SimpleAtod() || !absl::SimpleAtod())
  return nullopt;
```

src/core/search/indices.h
```diff
@@ -205,4 +206,29 @@ struct HnswVectorIndex : public BaseVectorIndex {
...
+  explicit GeoIndex(PMR_NS::memory_resource* mr);
+  ~GeoIndex();
+
+  bool Add(DocId id, const DocumentAccessor& doc, std::string_view field) override;
+  void Remove(DocId id, const DocumentAccessor& doc, std::string_view field) override;
+  std::vector<DocId> RadiusSearch(double lat, double lon, double radius) const;
+
+  std::vector<DocId> GetAllDocsWithNonNullValues() const override {
+    return std::vector<DocId>{};
+  }
```
dranikpg: That need to be for a `@geofield:*` query? Is this possible? If not, we should explicitly disallow it. Not saying to solve it now, just a TODO would help us not to forget

src/core/search/indices.cc
```diff
@@ -614,4 +618,96 @@ void HnswVectorIndex::Remove(DocId id, const DocumentAccessor& doc, string_view
...
+  } else if (unit == "FT") {
+    return radius * 0.3048;
+  } else if (unit == "MI") {
+    return radius * 1609.34;
+  } else {
+    return -1;
+  }
+}
+
+std::optional<GeoIndex::point> GeoIndex::GetGeoPoint(const DocumentAccessor& doc,
```
dranikpg: Can be in anonymous namespace

dranikpg (approved):
Looks valid so far

src/core/search/ast_expr.h
mkaruza: Done

src/core/search/indices.h
mkaruza: Added function to return all documents

dranikpg (approved):

PR #5862, opened 2025-09-26
fix: clang build for macos

vyavdoshenko:
Includes a fix from helio: https://github.com/romange/helio/pull/473

- Switch to the clang compiler for macOS build
- Adjusted tests

Successful build (combined changes from this PR and from helio PR): https://github.com/dragonflydb/dragonfly/actions/runs/18034977145


src/server/bitops_family.cc
```diff
@@ -564,17 +564,21 @@ void BitCount(CmdArgList args, const CommandContext& cmd_cntx) {
   CmdArgParser parser(args);
   auto key = parser.Next<string_view>();
 
-  auto [start, end] = parser.HasNext()
-                          ? parser.Next<int64_t, int64_t>()
-                          : std::pair<int64_t, int64_t>{0, std::numeric_limits<int64_t>::max()};
+  std::pair<int64_t, int64_t> start_end;
```
kostasrim, line 567: What's wrong with structure bindings ? We use them in other places as well ?

src/server/geo_family_test.cc
```diff
@@ -60,12 +60,12 @@ TEST_F(GeoFamilyTest, GeoAddOptions) {
   // update 1 + CH + XX
   EXPECT_EQ(1, CheckedInt({"geoadd", "Sicily", "CH", "XX", "10.361389", "38.115556", "Palermo"}));
   resp = Run({"geopos", "Sicily", "Palermo"});
-  EXPECT_THAT(resp, RespArray(ElementsAre("10.361386835575104", "38.1155563954963")));
+  EXPECT_THAT(resp, RespArray(ElementsAre(DoubleArg(10.361389), DoubleArg(38.115556))));
```
kostasrim, line 63: Is there a reason for these changes ?

kostasrim (approved):
Out of curiosity why did the build start failing? I saw switched to clang and I assume the CI used gcc before or ? And that's also the reason you removed fno-stack-clash-protection ?

src/server/geo_family_test.cc
vyavdoshenko: String representation is a little bit different on Clang. This is a portable way to verify these values.

src/server/bitops_family.cc
vyavdoshenko: I don't know. I just fixed the compiler error.

romange (approved):

vyavdoshenko:
> Out of curiosity why did the build start failing? I saw switched to clang and I assume the CI used gcc before. Did the version got updated and stop working ? And that's also the reason you removed fno-stack-clash-protection ?

We started failing after yet another change last week. Our macOS job was forcing Homebrew GCC, which mixes libstdc++ with dependencies built using Apple Clang and libc++. That drift can cause runtime/linker issues (including absl raw_hash_set assertions) and make the build fragile. It’s not just a GCC version bump — pinning to different gcc version doesn’t stabilize CI. Switching to Apple Clang restores a single, platform‑native toolchain on macOS and fixes the failures (together with a small helio bump). Regarding --fno-stack-clash-protection: this flag is GCC‑only and unsupported on Darwin; on macOS it’s a no‑op for Clang and triggers “unknown option” (often promoted to an error), so I dropped it.

kostasrim:
> > Out of curiosity why did the build start failing? I saw switched to clang and I assume the CI used gcc before. Did the version got updated and stop working ? And that's also the reason you removed fno-stack-clash-protection ?
> 
> We started failing after yet another change last week. Our macOS job was forcing Homebrew GCC, which mixes libstdc++ with dependencies built using Apple Clang and libc++. That drift can cause runtime/linker issues (including absl raw_hash_set assertions) and make the build fragile. It’s not just a GCC version bump — pinning to different gcc version doesn’t stabilize CI. Switching to Apple Clang restores a single, platform‑native toolchain on macOS and fixes the failures (together with a small helio bump). Regarding --fno-stack-clash-protection: this flag is GCC‑only and unsupported on Darwin; on macOS it’s a no‑op for Clang and triggers “unknown option” (often promoted to an error), so I dropped it.

Thank you for the explanation -- makes sense!

PR #5866, opened 2025-09-28
Fix: environment variable typo in comment

maets:
Fix a typo within a comment

### DIFF
--- a/contrib/charts/dragonfly/values.yaml
+++ b/contrib/charts/dragonfly/values.yaml
@@ -133,7 +133,7 @@ tls:
   #   HLunol88AeTOcKfD6hBYGvcRfu5NV29jJxZCOBfbFQXjnNlnrhRCag==
   #   -----END RSA PRIVATE KEY-----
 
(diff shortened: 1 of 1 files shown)


contrib/charts/dragonfly/values.yaml
```diff
@@ -133,7 +133,7 @@ tls:
   #   HLunol88AeTOcKfD6hBYGvcRfu5NV29jJxZCOBfbFQXjnNlnrhRCag==
   #   -----END RSA PRIVATE KEY-----
 
-# If enabled will set DFLY_PASSOWRD environment variable with the specified existing secret value
+# If enabled will set DFLY_PASSWORD environment variable with the specified existing secret value
```
kostasrim, line 136: We have deprecated DFLY_PASSWORD, isn't that right ? @romange

kostasrim (approved):
Thank you for your contribution!

contrib/charts/dragonfly/values.yaml
dranikpg: Yes, a year ago, its now `DFLY_requirepass`. We can update it later separately

PR #5868, opened 2025-09-29
actions: Set node version

abhijat:
fix node version explicitly and match github action used by mastodon to fix failing CI tests


.github/workflows/mastodon-ruby-tests.yml
```diff
@@ -65,7 +65,7 @@ jobs:
           ruby-version: 3.4
           bundler-cache: true
       - name: Set up Node.js
-        uses: actions/setup-node@v5
+        uses: actions/setup-node@v4
```
kostasrim, line 68: there is node version manager in case we need fine grain control in the future

Issue #5845, opened 2025-09-23
"Could not load snapshot" - impossible to restart DFly when maxmemory was reached


Pinimo:
### Bug description

Once memory is full, if persistence (snapshotting) is enabled, it is impossible to restart a Dragonfly pod.

1. Enable persistence, disable tiering, set replicas=2.
2. Fill up the DB entirely with string keys (see how in the script below).
3. Wait for the snapshot to be made (1').
4. Delete a pod.

The pod will crash with `exitCode` "1" 100% of the time, because the snapshot is allegedly too big (by a few kB) to fit in memory.

```
Could not load snapshot - its used memory is 293637328 but the limit is 293601280
Rdb load failed: Out of memory, or used memory is too high
```

### To Reproduce

<details>

<summary>Reproduction script</summary>

```bash

kubectx
cat << EOF | kubectl apply -f -
---
apiVersion: dragonflydb.io/v1alpha1
kind: Dragonfly
metadata:
  name: mycache
  namespace: default
spec: 
  args:
  - --dbfilename=dump
  - --proactor_threads=1
  - --mem_defrag_waste_threshold=50
  - --mem_defrag_threshold=1.1
  # With maxemory=280Mi (auto) and mem limit = 350Mi => Could not load snapshot - its used memory is 293637328 but the limit is 293601280 (that's 36048 bytes above)
  # - --compression_mode=0  # may prevent initial cluster creation even with empty PVC! "Could not load snapshot - its used memory is 293647632 but the limit is 293601280"
  - --df_snapshot_format=false
  # - --maxmemory=260Mi  # tried with memory limit=520Mi, got exit code 1 with "Could not load snapshot - its used memory is 272665296 but the limit is 272629760" (that's 35536 bytes above)
  image: ghcr.io/dragonflydb/dragonfly:v1.34.1
  imagePullPolicy: Always
  replicas: 2
  resources:
    limits:
      memory: 350Mi
    requests:
      cpu: 500m
      memory: 350Mi
  snapshot:
    cron: '*/1 * * * *'
    dir: /data
    persistentVolumeClaimSpec:
      accessModes:
      - ReadWriteOnce
      resources:
        requests:
          storage: 5Gi

EOF

echo -n Waiting for the Dragonfly to be ready...
until kubectl get -n default dragonflies.dragonflydb.io/mycache -oyaml | yq .status | grep Ready; do
  echo -n '.'
  sleep 1
done

echo

sleep 10

[...]

PR #5869, opened 2025-09-29
fix: allow 5% tolerance if snapshot used memory greater than maxmemory limit

BorysTheDev:
fixes: #5845

Added 5% tolerance for snapshot memory usage in comparison with max_memory_limit. It can't completely solve the problem, but for now, we can't do more.

### DIFF
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -2458,9 +2458,7 @@ error_code RdbLoader::HandleAux() {
     int64_t usedmem;
     if (absl::SimpleAtoi(auxval, &usedmem)) {
       VLOG(1) << "RDB memory usage when created " << strings::HumanReadableNumBytes(usedmem);
-      // We allow 5% tolerance for snapshot used memory, because of a lot of factors.
-      // In my tests I saw used_memory field was 20% higher than maxmemory limit if maxmemory was
-      // 350MB
+      // We allow 5% tolerance for snapshot used memory
       if (usedmem > (max_memory_limit * 1.05)) {
         if (IsClusterEnabled()) {
           LOG(INFO) << "Allowing to load a snapshot of size " << usedmem


src/server/rdb_load.cc
```diff
@@ -2458,7 +2458,10 @@ error_code RdbLoader::HandleAux() {
     int64_t usedmem;
     if (absl::SimpleAtoi(auxval, &usedmem)) {
       VLOG(1) << "RDB memory usage when created " << strings::HumanReadableNumBytes(usedmem);
-      if (usedmem > ssize_t(max_memory_limit)) {
+      // We allow 5% tolerance for snapshot used memory, because of a lot of factors.
+      // In my tests I saw used_memory field was 20% higher than maxmemory limit if maxmemory was
+      // 350MB
```
Pinimo: Interesting! I have a few questions:
1. Does it mean we might still have the issue on 350MB instances after this fix, because we only give them 5% more where they would need 20% more?
2. Once the keys are loaded:
    a. Do we expect the used_memory to go back down?
    b. Does the next snapshot have a larger size than the previous one? (The logical answer sounds like "no", but if yes, several successive restarts could lead to an inflation in the snapshot size, till the instance cannot start again.)
4. Do you have other benchmarks of used_memory for instances of bigger sizes? I think the chosen threshold should maybe be dependent on this. So perhaps the rule should take into account the instance size.

BorysTheDev: 1. yes
2. a. maybe after some usage
    b. the field - yes, for one restart, I think. During loading, we can use a little more memory .
3. as I know we don't have other benchmarks.

Pinimo: Thank you!

src/server/rdb_load.cc
```diff
@@ -2458,7 +2458,10 @@ error_code RdbLoader::HandleAux() {
     int64_t usedmem;
     if (absl::SimpleAtoi(auxval, &usedmem)) {
       VLOG(1) << "RDB memory usage when created " << strings::HumanReadableNumBytes(usedmem);
-      if (usedmem > ssize_t(max_memory_limit)) {
+      // We allow 5% tolerance for snapshot used memory, because of a lot of factors.
+      // In my tests I saw used_memory field was 20% higher than maxmemory limit if maxmemory was
```
romange: I do not think it's a good comment as it does not say much.

PR #5870, opened 2025-09-29
[wip] simsimd dynamic dispatch

vyavdoshenko:


src/core/search/vector_utils.cc
```diff
@@ -16,10 +16,22 @@ using namespace std;
...
 #ifdef USE_SIMSIMD
-// Native float16 support is controlled via CMake option SIMSIMD_NATIVE_F16
-// Default behavior disables native F16/BF16 for maximum compatibility
 #include <simsimd/simsimd.h>
-#endif
+#ifdef SIMSIMD_DYNAMIC_DISPATCH
+// Initialize dynamic dispatch on first use (only when dynamic dispatch is enabled)
+static bool InitializeSimSIMD() {
+  static bool initialized = false;
+  if (!initialized) {
```
BorysTheDev: it looks like you need std::call_once()

BorysTheDev (approved):

src/core/search/vector_utils.cc
vyavdoshenko: Fixed

src/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ option(ENABLE_GIT_VERSION "Build with Git metadata" OFF)
 
 option(USE_SIMSIMD "Enable SimSIMD vector optimizations" OFF)
 option(SIMSIMD_NATIVE_F16 "Enable native float16 support in SimSIMD" OFF)
+option(SIMSIMD_LINK_SHARED "Link SimSIMD as shared library (ON) or use header-only (OFF)" ON)
```
romange: i did not know it's a requirement. I suspect it will break the deployment for our users, so I suggest switching to OFF by default.

@vyavdoshenko did you check if shared library is a hard requirement? is it possible to build SimSIMD as a static lib? can you please check the original repo if there is an issue or discussion around it?

src/external_libs.cmake
```diff
@@ -165,13 +165,28 @@ add_third_party(
...
+      LIB libsimsimd.so
+    )
+  else()
+    # Header-only usage: fetch headers only, no library produced
+    add_third_party(
+      simsimd
+      URL https://github.com/ashvardanian/SimSIMD/archive/refs/tags/v6.5.3.tar.gz
+
+      BUILD_COMMAND echo SKIP
+      INSTALL_COMMAND bash -c "mkdir -p ${THIRD_PARTY_LIB_DIR}/simsimd && cp -R <SOURCE_DIR>/include ${THIRD_PARTY_LIB_DIR}/simsimd/"
```
romange: what's wrong with the previous version of INSTALL_COMMAND ?

src/external_libs.cmake
```diff
@@ -165,13 +165,28 @@ add_third_party(
...
 
 if(USE_SIMSIMD)
-  add_third_party(
-    simsimd
-    URL https://github.com/ashvardanian/SimSIMD/archive/refs/tags/v6.5.3.tar.gz
-    BUILD_COMMAND echo SKIP
-    INSTALL_COMMAND cp -R <SOURCE_DIR>/include ${THIRD_PARTY_LIB_DIR}/simsimd/
-    LIB "none"
-  )
+  if(SIMSIMD_LINK_SHARED)
```
romange: I suggest you set SIMD_BUILD_CMD, SIMD_INSTALL_CMD, SIMD_LIB variables conditionally and have a single `add_third_party` call.

src/core/search/vector_utils.cc
```diff
@@ -16,10 +17,21 @@ using namespace std;
...
 #ifdef USE_SIMSIMD
-// Native float16 support is controlled via CMake option SIMSIMD_NATIVE_F16
-// Default behavior disables native F16/BF16 for maximum compatibility
 #include <simsimd/simsimd.h>
-#endif
+#ifdef SIMSIMD_DYNAMIC_DISPATCH
+// Initialize dynamic dispatch on first use (only when dynamic dispatch is enabled)
+static bool InitializeSimSIMD() {
+  static std::once_flag simsimd_once;
+  std::call_once(simsimd_once, []() {
```
romange: I do not think we need `std::call_once` if you already use `static const bool` initializer below.

(btw, I am not a fan of implicit initializers, and prefer explicit ones, say called from `Service::Init`

romange:
my biggest concern is that now we introduce a 3rd party dependency on a shared lib which never existed before.
If that's the case I would even suggest to give up on dynamic dispatching provided by simsimd and implement one of our own - I think they talk about it here: https://github.com/ashvardanian/SimSIMD/?tab=readme-ov-file#target-specific-backends

romange:
a side comment: regarding the benchmark, the reason why it's bad with `SIMSIMD_LINK_SHARED=OFF` is because simd  fallbacks to a scalar implementation worse than what we have without simd.  if you add  `--march=native` flag to `dfly_search_core` lib, you will see it working as good as dynamic dispatch (but it's not a good solution for releases).

src/CMakeLists.txt
vyavdoshenko: I fixed it. Now, dynamic dispatch works for a static library build.

src/external_libs.cmake
vyavdoshenko: This was an artifact from a previous experiment.
Fixed.

src/external_libs.cmake
vyavdoshenko: Fixed.

src/core/search/vector_utils.cc
vyavdoshenko: Fixed.

src/core/search/vector_utils.cc
```diff
@@ -112,4 +111,11 @@ float VectorDistance(const float* u, const float* v, size_t dims, VectorSimilari
   return 0.0f;
 }
 
+void InitSimSIMD() {
+#if defined(WITH_SIMSIMD)
+  static std::once_flag simsimd_once;
```
romange: this is not needed. it's fine to call `simsimd_capabilities()`

vyavdoshenko: Fixed

src/core/search/vector_utils.cc
```diff
@@ -6,6 +6,7 @@
 
 #include <cmath>
 #include <memory>
+#include <mutex>
```
romange:
```suggestion
```

romange (dismissed):

Issue #5863, opened 2025-09-26
Pass in custom logic to Dragonfly health check


xuekat:
We'd like to pass in custom logic to the dragonfly health check. Our use case is that we're using Dragonfly as our main metadata/feature store, and whenever we need to do a helm upgrade on our dragonfly instances, the rollout will roll out one pod at a time and perform a rudimentary health check (redis ping) on the pod, before killing the previous pod. But the new pod still needs to load an entire snapshot into memory, which could take up to 3 minutes. This is 3 minutes of down time for us, so was wondering if there was a way to also add the dataset loading check into the health check as well.

moredure:
@xuekat Hello! Have you considered overriding the healthcheck script within your own custom image which is based on dragonfly one?

romange:
I confirm that PING during load returns `LOADING Dragonfly is loading the dataset in memory` and it's consistent with other datastores. Next step - to confirm that docker container indeed shows "not healthy" if ping returns LOADING and if this is the case - to fix it. `ping` with LOADING response should be considered as healthy.

romange:
@xuekat  what is the behavior that you need during the snapshot loading? the datastore won't accept any reads during that time so how health check solves the issue of downtime?

the right approach (and this is what we do in our cloud service) is to use replication for version updates, to have zero-downtime updates.

romange:
there is no response, I am closing this issue for now - healthcheck is working as intended

xuekat:
We're hoping for health check to only pass once the dragonfly pod has finished loading the dataset into memory, because as the logic currently stands, a dragonfly pod might have passed health check but still not be available as it's loading dataset into memory; so the operator will kill the old pod while the new pod is still not yet able to handle requests. 

For now we are remediating this by increasing the number of replicas to be very high so that at least one pod is still alive during the rollout, but this is suboptimal because it increases our infra costs.

Issue #5844, opened 2025-09-23
DragonflyDB's health check script using nc has created zombie processes.


AdaAibaby:
**Describe the bug**
 

**To Reproduce**
        livenessProbe:
          exec:
            command:
            - /bin/sh
            - /usr/local/bin/healthcheck.sh
          failureThreshold: 3
          initialDelaySeconds: 10
          periodSeconds: 10
          successThreshold: 1
          timeoutSeconds: 5
        name: dragonfly
        ports:
        - containerPort: 6379
          name: redis
          protocol: TCP
        - containerPort: 9999
          name: admin
          protocol: TCP
        readinessProbe:
          exec:
            command:
            - /bin/sh
            - /usr/local/bin/healthcheck.sh
          failureThreshold: 3
          initialDelaySeconds: 10
          periodSeconds: 10
          successThreshold: 1
          timeoutSeconds: 5
 

and
cat /usr/local/bin/healthcheck.sh 
#!/bin/sh

HOST="localhost"
PORT=$HEALTHCHECK_PORT

if [ -z "$HEALTHCHECK_PORT" ]; then
  DF_NET=$(netstat -tlnp | grep "1/dragonfly")
  if [ -z "$DF_NET" ]; then
    # if we failed, then lets try the priveleged version. is triggerred by the regular command:
    # docker run docker.dragonflydb.io/dragonflydb/dragonfly
    DF_NET=$(su dfly -c "netstat -tlnp" | grep "1/dragonfly")

  fi

  PORT=$(echo $DF_NET | grep -oE ':[0-9]+' | cut -c2- | tail -n 1)
fi

_healthcheck="nc -q1 $HOST $PORT"

echo PING | ${_healthcheck}

exit $?

<img width="2871" height="845" alt="Image" src="https://github.com/user-attachments/assets/c5c3259a-0643-49e2-8d25-d11f6be53a82" />

<img width="1307" height="571" alt="Image" src="https://github.com/user-attachments/assets/598441b0-2d03-4155-9e4c-1ba85f9c1142" />

romange:
@shaolila-dotcom  I am not sure how to reproduce this. Does it always happen when running a docker container?

romange:
maybe adding the `-w timeout` option to `nc` will make it more robust.

vyavdoshenko:
I can reproduce this issue with orphaned sh defunct processes.

Reproduction:
```
[...]

PR #5878, opened 2025-09-30
fix(docker): prevent zombie processes and improve healthcheck

vyavdoshenko:
Fixes zombie process accumulation (#5844) and snapshot loading detection (#5863) in Docker healthcheck.

Key Changes:
- Add `tini` as PID 1 to reap zombie processes
- Use `redis-cli` + `trap cleanup EXIT` for reliable subprocess management  
- Check `INFO PERSISTENCE` loading state to prevent traffic during snapshot loading
- Fix process detection to work with tini (`/dragonfly` vs `1/dragonfly`)

Impact:
- Eliminates zombie process buildup
- Prevents 3+ minute downtime during Kubernetes rollouts

Fixes: #5844
Fixes: #5863

### DIFF
--- a/tools/docker/healthcheck.sh
+++ b/tools/docker/healthcheck.sh
@@ -1,8 +1,17 @@
 #!/bin/sh
 
+# Cleanup function to prevent zombie processes (issue #5844)
+# This is critical when dragonfly runs as PID 1 without an init system
+cleanup() {
+  # Wait for all background/child processes to finish
+  wait 2>/dev/null || true
+}
+
+# Set trap to ensure cleanup runs on exit, regardless of how the script exits
+trap cleanup EXIT
+
 HOST="localhost"
 PORT=$HEALTHCHECK_PORT
-
 
 if [ -z "$HEALTHCHECK_PORT" ]; then
   # try unpriveleged version first. This should cover cases when the container is running
@@ -20,8 +29,24 @@ if [ -z "$HEALTHCHECK_PORT" ]; then
   PORT=$(echo $DF_NET | grep -oE ':[0-9]+' | cut -c2- | tail -n 1)
 fi
 
-_healthcheck="nc -q1 $HOST $PORT"
+# Use redis-cli instead of nc for better reliability
+# and to properly check loading state (issue #5863)
+REDIS_CLI="redis-cli -h $HOST -p $PORT"
 
-echo PING | ${_healthcheck}
+# Step 1: Basic PING check
+if ! timeout 3 $REDIS_CLI PING 2>/dev/null | grep -q "PONG"; then
+  exit 1
+fi
 
-exit $?
+# Step 2: Check if server is in LOADING state
+# During snapshot loading, the server responds to PING but is not ready for traffic
+# Note: 'loading' field is in PERSISTENCE section
+INFO_OUTPUT=$(timeout 3 $REDIS_CLI INFO PERSISTENCE 2>/dev/null)
+
+# Server is ready when loading:0
+if echo "$INFO_OUTPUT" | grep -q "^loading:0"; then
+  exit 0
+fi
+
+# If loading:1 or can't determine state, fail to be safe
+exit 1
--- a/tools/docker/healthcheck.sh
+++ b/tools/docker/healthcheck.sh
@@ -17,11 +17,11 @@ if [ -z "$HEALTHCHECK_PORT" ]; then
   # try unpriveleged version first. This should cover cases when the container is running
   # without root, for example:
   # docker run  --group-add 999  --cap-drop=ALL --user 999 docker.dragonflydb.io/dragonflydb/dragonfly
-  DF_NET=$(netstat -tlnp | grep "1/dragonfly")
+  DF_NET=$(netstat -tlnp | grep "/dragonfly")
   if [ -z "$DF_NET" ]; then
     # if we failed, then lets try the priveleged version. is triggerred by the regular command:
     # docker run docker.dragonflydb.io/dragonflydb/dragonfly
-    DF_NET=$(su dfly -c "netstat -tlnp" | grep "1/dragonfly")
+    DF_NET=$(su dfly -c "netstat -tlnp" | grep "/dragonfly")
   fi
 
   # check all the TCP ports, and fetch the port.
--- a/tools/packaging/Dockerfile.alpine-dev
+++ b/tools/packaging/Dockerfile.alpine-dev
@@ -35,7 +35,7 @@ COPY tools/docker/healthcheck.sh /usr/local/bin/healthcheck.sh
 COPY --from=builder /build/build-release/dragonfly /usr/local/bin/
 
 RUN apk --no-cache add libgcc libstdc++  \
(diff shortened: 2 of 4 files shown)


tools/docker/healthcheck.sh
```diff
@@ -1,27 +1,52 @@
...
+    DF_NET=$(su dfly -c "netstat -tlnp" | grep "/dragonfly")
   fi
 
   # check all the TCP ports, and fetch the port.
   # For cases when dragonfly opens multiple ports, we filter with tail to choose one of them.
   PORT=$(echo $DF_NET | grep -oE ':[0-9]+' | cut -c2- | tail -n 1)
 fi
 
-_healthcheck="nc -q1 $HOST $PORT"
+# Use redis-cli instead of nc for better reliability
```
romange: we used redis-cli in the past but had problems with that - please check the history and see why. 
I prefer reverting to `nc` if this is not crucial for the fixes references in the PR.

tools/docker/healthcheck.sh
```diff
@@ -1,27 +1,52 @@
...
-echo PING | ${_healthcheck}
+# Step 1: Basic PING check
+if ! timeout 3 $REDIS_CLI PING 2>/dev/null | grep -q "PONG"; then
+  exit 1
+fi
+
+# Step 2: Check if server is in LOADING state
+# During snapshot loading, the server responds to PING but is not ready for traffic
+# Note: 'loading' field is in PERSISTENCE section
+INFO_OUTPUT=$(timeout 3 $REDIS_CLI INFO PERSISTENCE 2>/dev/null)
```
romange: I do not understand why do we need two invocations here.
why not save the response into a variable and check if it's PONG or a "LOADING " error?

romange:
so the current state that the pod loading into snapshot returns healthy and you are changing it to non-healthy status?
I am confused now.

vyavdoshenko:
> so the current state that the pod loading into snapshot returns healthy and you are changing it to non-healthy status? I am confused now.

Yes.

Current behavior (the problem):
- Pod is loading snapshot into memory (takes 3+ minutes)
- Healthcheck sends `PING` → receives `PONG` 
- Kubernetes thinks the pod is healthy and ready for traffic
- Kubernetes kills the old pod and routes traffic to the new one
- Result: 3+ minutes of downtime because the new pod is still loading data

After this fix:
- Pod is loading snapshot into memory
- Healthcheck sends `PING` → receives `-LOADING Dragonfly is loading the dataset in memory`
- Kubernetes sees pod as **unhealthy/not ready**
- Kubernetes keeps the old pod running until the new pod finishes loading
- Result: zero downtime during rolling updates

The issue (#5863):
Reported this problem - during helm upgrades, the readiness probe passes too early (on PING), causing 3 minutes of downtime while the new pod loads the snapshot.

tools/docker/healthcheck.sh
vyavdoshenko: Fixed.

tools/docker/healthcheck.sh
vyavdoshenko: Fixed

romange:
> > so the current state that the pod loading into snapshot returns healthy and you are changing it to non-healthy status? I am confused now.
> 
> Yes.
> 
> Current behavior (the problem):
> 
> * Pod is loading snapshot into memory (takes 3+ minutes)
> * Healthcheck sends `PING` → receives `PONG`
> * Kubernetes thinks the pod is healthy and ready for traffic
> * Kubernetes kills the old pod and routes traffic to the new one
> * Result: 3+ minutes of downtime because the new pod is still loading data
> 
> After this fix:
> 
> * Pod is loading snapshot into memory
> * Healthcheck sends `PING` → receives `-LOADING Dragonfly is loading the dataset in memory`
> * Kubernetes sees pod as **unhealthy/not ready**
> * Kubernetes keeps the old pod running until the new pod finishes loading
> * Result: zero downtime during rolling updates
> 
> The issue (#5863): Reported this problem - during helm upgrades, the readiness probe passes too early (on PING), causing 3 minutes of downtime while the new pod loads the snapshot.

I disagree that the loading phase should be considered as "not healthy". K8S can decide to restart unhealthy pods leading to an infinite loop of a pod loading / restarting.
I think the previous healthcheck logic is perfect and this will keep this way until there is a specific use-case that proves otherwise.  Lets revert that part of the code

romange:
but I am ready to be proven otherwise, I am just worried that it will ruin other usecases

romange:
@vyavdoshenko  is it possible to demonstrate the faulty behavior using local kubectl or something?

vyavdoshenko:
@romange 
I reproduced the issue on my local cluster. 
[demo.tar.gz](https://github.com/user-attachments/files/22659277/demo.tar.gz)
I think we can save the current behavior and add a parameter to the healthcheck.sh for strict checking of the LOADING state.
Also, maybe it's worth creating a help page on how to create a custom health checking script using the added volume.

vyavdoshenko:
@romange 
I added an optional flag -l or --check-loading for checking the LOADING state. It doesn't break current behavior and can be used as an option if needed.
If you are against it, I will revert it to the original behavior.

vyavdoshenko:
> how it is possible to run healthchecks with this argument?

I mean that this is possible to use outside with an argument.

```
docker run -d --name df-test docker.dragonflydb.io/dragonflydb/dragonfly

docker exec df-test /usr/local/bin/healthcheck.sh && echo "Pass" || echo "Fail"

docker exec df-test /usr/local/bin/healthcheck.sh --check-loading && echo "Pass" || echo "Fail"
```

romange (commented):
I do not think this argument helps anyone as it's not trigerred by docker.
I do not know if it's true but based on gemini, K8s does not rely on docker healthchecks as they do not distinguish between liveness and readiness:
```
Why Kubernetes Ignores the Docker Health Check
Kubernetes needs a single, unified mechanism to manage the lifecycle of a Pod and its containers. It has three distinct probe types that offer granular control over its self-healing logic, which the single Docker Health Check cannot provide:

Liveness Probe: Should I restart this container?

Readiness Probe: Should I send traffic to this Pod? (The Docker Health Check has no equivalent for this essential function).

Startup Probe: Is the container finished starting up yet?

Because the Docker Health Check conflates "alive" and "ready," Kubernetes replaces it entirely with its own superior probing system. You should always use Kubernetes Liveness and Readiness Probes for applications running on Kubernetes.
```

vyavdoshenko:
I just wanted to say that this parameter can be used outside.
It can also be used in k8s like a parameter:

```
readinessProbe:
  exec:
    command:
    - /usr/local/bin/healthcheck.sh
    - --check-loading
```

I can create an example configuration for a local cluster on how to use it with parameters and without.

romange:
My point is that it's not the right way to do it.
I do not think docker healthcheck should be used for readiness/startup probe?

is it possible to provide custom startup probe logic (script) within k8s configuration without changing the underlying container directly?
@Abhra303  would love to hear your feedback as well.

vyavdoshenko:
@romange 
I left changes related to zombie process elimination.

PR #5879, opened 2025-10-02
[WIP] core: Add data structure for metric which decays over time

abhijat:
WIP


src/core/count_min_sketch.h
```diff
@@ -0,0 +1,95 @@
...
+//
+#pragma once
+
+#include <array>
+#include <cstdint>
+#include <vector>
+
+namespace dfly {
+
+class CountMinSketch {
```
dranikpg, line 20: Please add some comments from your PR description to the header, i.e. what it's for, what the functions, do etc. So anyone who finds it in the future can tell by them what its there for 🙂

src/core/count_min_sketch.h
```diff
@@ -0,0 +1,95 @@
...
+  };
+
+  explicit MultiSketch(uint64_t rollover_ms = 1000, double epsilon = 0.0001, double delta = 0.0001,
+                       Decay decay = Decay::Linear);
+
+  MultiSketch(const MultiSketch& other) = delete;
+  MultiSketch& operator=(const MultiSketch& other) = delete;
+
+  MultiSketch(MultiSketch&& other) noexcept = default;
+  MultiSketch& operator=(MultiSketch&& other) noexcept = default;
```
dranikpg: 1. Not sure why you are so defensive 
2. It should be uncopyable/immovable due to SketchWithTimestamp being such itself

src/core/count_min_sketch.cc
```diff
@@ -0,0 +1,132 @@
...
+  using dfly::MultiSketch;
+  switch (decay) {
+    case MultiSketch::Decay::Exponential:
+      return ExponentialDecay(value, time_delta);
+    case MultiSketch::Decay::Linear:
+      return LinearDecay(value, time_delta);
+    case MultiSketch::Decay::SlidingWindow:
+      return value;
+  }
+  ABSL_UNREACHABLE();
```
dranikpg: You might not need ApplyDecay at all. You can make an array of three function pointers and then index it by the decay integer value

src/core/count_min_sketch.cc
```diff
@@ -0,0 +1,132 @@
...
+
+constexpr auto MAX = std::numeric_limits<dfly::CountMinSketch::SizeT>::max();
+
+uint64_t GetCurrentMS() {
+  return absl::GetCurrentTimeNanos() / 1000 / 1000;
+}
+
+uint64_t ExponentialDecay(uint64_t value, int64_t time_delta) {
+  // Value halves every 5000 ms: ln(2) / 5000
+  static constexpr double EXP_DECAY_CONST = 0.000138629;
```
dranikpg, line 24: I thought std::log must be constexpr for sure! But it's only such from c++ 26 😆 💀

src/core/count_min_sketch.cc
```diff
@@ -0,0 +1,132 @@
...
+    ctr[index] = updated < curr ? MAX : updated;
+  });
+}
+
+CountMinSketch::SizeT CountMinSketch::EstimateFrequency(uint64_t key) const {
+  uint64_t i = 0;
+  auto it = counters_.begin();
+  SizeT estimate = (*it)[Hash(key, i++)];
+  std::for_each(++it, counters_.end(),
+                [&](const auto& ctr) { estimate = std::min(estimate, ctr[Hash(key, i++)]); });
```
dranikpg: Oha, took me a while with all the iterators/indices/functions 🤯 
Maybe just a good ol' loop? 🙂 
```c++
SizeT estimate = numeric_limits<SizeT>::max();
for (unsigned i = 0; i < counters_.size(); i++)
  estimate = min(estimate, counters_[i][hash(key, i++]);
```

dranikpg (dismissed):
Looks valid. Didn't dive into how the math works, so I trust you on that 🙂 Please add some more basic comments to the header. Otherwise some minor comments

src/core/count_min_sketch.cc
abhijat: changed

src/core/count_min_sketch.cc
abhijat: changed

src/core/count_min_sketch.h
abhijat: I added some comments, will add some references for the width and depth calculations

src/core/count_min_sketch.h
abhijat: Its just a template I use in my IDE for new classes, but its not required here as you said, so I have removed it now

src/core/count_min_sketch.cc
```diff
@@ -0,0 +1,123 @@
...
+
+uint64_t LinearDecay(uint64_t value, int64_t time_delta) {
+  // Value decrements by one every 1000 ms
+  static constexpr double LIN_DECAY_CONST = 0.001;
+
+  const double decay = time_delta * LIN_DECAY_CONST;
+  return value - std::min(static_cast<double>(value), decay);
+}
+
+using DecayFn = std::function<uint64_t(uint64_t, int64_t)>;
```
dranikpg, line 37: Function pointer is enough, though std function of one shouldn't incurr any overhead

dranikpg (approved):
👍🏻

PR #5880, opened 2025-10-02
fix: Fix Fedora 30 and MacOS build

mkaruza:
Also silence false positive warning in indices.cc.

### DIFF
--- a/src/core/search/CMakeLists.txt
+++ b/src/core/search/CMakeLists.txt
@@ -5,12 +5,14 @@ cur_gen_dir(gen_dir)
 
 set_source_files_properties(${gen_dir}/parser.cc PROPERTIES
                             COMPILE_FLAGS "-Wno-maybe-uninitialized")
+set_source_files_properties(indices.cc PROPERTIES
+                            COMPILE_FLAGS "-Wno-maybe-uninitialized")
 add_library(dfly_search_core base.cc ast_expr.cc query_driver.cc search.cc indices.cc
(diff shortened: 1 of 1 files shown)


src/core/search/CMakeLists.txt
```diff
@@ -5,12 +5,14 @@ cur_gen_dir(gen_dir)
 
 set_source_files_properties(${gen_dir}/parser.cc PROPERTIES
                             COMPILE_FLAGS "-Wno-maybe-uninitialized")
+set_source_files_properties(indices.cc PROPERTIES
```
vyavdoshenko: I think this is not a good idea.
Silencing the warning above is correct because it's generated code. This code is handwritten. Please leave it as is or fix the warnings.

dranikpg (changes requested):
Please do it this way 🙂 

https://github.com/dragonflydb/dragonfly/blob/1da4817e341ef90e5ea889b3542bc67b289a8018/src/core/search/search.cc#L45-L50

src/core/search/indices.cc
```diff
@@ -2,6 +2,12 @@
 // See LICENSE for licensing terms.
 //
 
+// Wrong warning in a deeply inlined when boost::geometry::buffer is used.
+#ifndef __clang__
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+#endif
+
```
dranikpg: You can do it around the geo index and then do `diagnositc pop` or smth like this 😅

vyavdoshenko (approved):

PR #5882, opened 2025-10-03
chore(hset_family): Code cleaning

dranikpg:
I want to write tiering into it in the future, but there's just so much code. Most of it does nothing. And I will have to update all of it 😭

### DIFF
--- a/src/server/hset_family.cc
+++ b/src/server/hset_family.cc
@@ -1285,49 +1285,26 @@ using CI = CommandId;
 
 #define HFUNC(x) SetHandler(&HSetFamily::x)
 
-namespace acl {
-constexpr uint32_t kHDel = WRITE | HASH | FAST;
-constexpr uint32_t kHLen = READ | HASH | FAST;
-constexpr uint32_t kHExists = READ | HASH | FAST;
-constexpr uint32_t kHGet = READ | HASH | FAST;
-constexpr uint32_t kHGetAll = READ | HASH | SLOW;
-constexpr uint32_t kHMGet = READ | HASH | FAST;
-constexpr uint32_t kHMSet = WRITE | HASH | FAST;
-constexpr uint32_t kHIncrBy = WRITE | HASH | FAST;
-constexpr uint32_t kHIncrByFloat = WRITE | HASH | FAST;
-constexpr uint32_t kHKeys = READ | HASH | SLOW;
-constexpr uint32_t kHRandField = READ | HASH | SLOW;
-constexpr uint32_t kHScan = READ | HASH | SLOW;
-constexpr uint32_t kHSet = WRITE | HASH | FAST;
-constexpr uint32_t kHSetEx = WRITE | HASH | FAST;
-constexpr uint32_t kHSetNx = WRITE | HASH | FAST;
-constexpr uint32_t kHStrLen = READ | HASH | FAST;
-constexpr uint32_t kHExpire = WRITE | HASH | FAST;
-constexpr uint32_t kHVals = READ | HASH | SLOW;
-}  // namespace acl
-
 void HSetFamily::Register(CommandRegistry* registry) {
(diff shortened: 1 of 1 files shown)


src/server/hset_family.cc
```diff
@@ -745,31 +717,26 @@ void HGetGeneric(CmdArgList args, uint8_t getall_mask, Transaction* tx, SinkRepl
...
-  if (result) {
-    bool is_map = (getall_mask == (VALUES | FIELDS));
-    rb->SendBulkStrArr(absl::Span<const string>{*result},
-                       is_map ? RedisReplyBuilder::MAP : RedisReplyBuilder::ARRAY);
-  } else {
-    builder->SendError(result.status());
-  }
+  switch (result.status()) {
+    case OpStatus::OK:
+    case OpStatus::KEY_NOTFOUND: {
```
BorysTheDev, line 725: are you sure that if KEY_NOTFOUND we should do the same as if OK

src/server/hset_family.cc
```diff
@@ -745,31 +717,26 @@ void HGetGeneric(CmdArgList args, uint8_t getall_mask, Transaction* tx, SinkRepl
...
   auto op_res = db_slice.FindMutable(op_args.db_cntx, key, OBJ_HASH);
-
-  if (!op_res) {
-    if (op_res.status() == OpStatus::KEY_NOTFOUND) {
-      std::vector<long> res(values.size(), -2);
-      return res;
-    }
-    return op_res.status();
-  }
+  RETURN_ON_BAD_STATUS(op_res);
```
BorysTheDev, line 739: It looks like you have changed the result and I think the reply will be different to the client

src/server/hset_family.cc
```diff
@@ -1285,49 +1240,26 @@ using CI = CommandId;
...
+            << CI{"HINCRBYFLOAT", CO::WRITE | CO::DENYOOM | CO::FAST, 4, 1, 1}.HFUNC(HIncrByFloat)
+            << CI{"HKEYS", CO::READONLY, 2, 1, 1}.HFUNC(HKeys)
+            << CI{"HEXPIRE", CO::WRITE | CO::FAST | CO::DENYOOM, -5, 1, 1}.HFUNC(HExpire)
+            << CI{"HRANDFIELD", CO::READONLY, -2, 1, 1}.HFUNC(HRandField)
+            << CI{"HSCAN", CO::READONLY, -3, 1, 1}.HFUNC(HScan)
+            << CI{"HSET", CO::WRITE | CO::FAST | CO::DENYOOM, -4, 1, 1}.HFUNC(HSet)
+            << CI{"HSETEX", CO::WRITE | CO::FAST | CO::DENYOOM, -5, 1, 1}.SetHandler(HSetEx)
+            << CI{"HSETNX", CO::WRITE | CO::DENYOOM | CO::FAST, 4, 1, 1}.HFUNC(HSetNx)
+            << CI{"HSTRLEN", CO::READONLY | CO::FAST, 3, 1, 1}.HFUNC(HStrLen)
+            << CI{"HVALS", CO::READONLY, 2, 1, 1}.HFUNC(HVals);
```
BorysTheDev, line 1262: why do not we need acl::* flags anymore?

src/server/hset_family.cc
BorysTheDev: ok I see you just moved this code in another function

src/server/hset_family.cc
BorysTheDev: code moved from OpGetAll

src/server/hset_family.cc
dranikpg: If you don't pass the acl flag, it will be determined by the command flags. It should have been this way from the start. 

I've checked `COMMAND` output, the acl flags didn't change

PR #5883, opened 2025-10-03
feat(tiering): Decoders

dranikpg:
Previously, basic string logic, like whether it's raw or not, whether it needs to be owned or can be just a slice, etc. was managed by the op_manager, so the callback interface was tailored to that logic and the callbacks as well.

once we add support for different types, we need to somehow:
* decode the value
* apply changes to the value
* track whether it was modified
* evalueate it's memory usage to understand if it's reasonable to upload it
* do the uploading process

Now the io driven manager (op_manager) works on a type erased level, just keeping track of a byte slice to be read and a list of callbacks.

The "decoder" does all of those basic operations behind a virtual interface to the op_manger can continue working as before. Additionally, the callbacks now accept their own decoder so it's easier for the to handle the value.

### DIFF
--- a/src/core/string_or_view.h
+++ b/src/core/string_or_view.h
@@ -71,7 +71,7 @@ class StringOrView {
     return std::move(std::get<std::string>(val_));
   }
 
-  std::string* BorrowMut() {
+  std::string* GetMutable() {
     MakeOwned();
     return &std::get<std::string>(val_);
   }
--- a/src/server/tiering/decoders.h
+++ b/src/server/tiering/decoders.h
@@ -32,6 +32,7 @@ struct Decoder {
   size_t estimated_mem_usage = 0;  // Estimated usage if uploaded
 };
 
+// Basic "bare" decoder that just stores the provided slice
 struct BareDecoder : public Decoder {
   std::unique_ptr<Decoder> Clone() const override;
   void Initialize(std::string_view slice) override;
--- a/src/server/tiering/decoders.cc
+++ b/src/server/tiering/decoders.cc
@@ -47,7 +47,7 @@ std::string_view StringDecoder::Read() const {
 
 std::string* StringDecoder::Write() {
   modified = true;
-  return value_.BorrowMut();
+  return value_.GetMutable();
 }
 
 }  // namespace dfly::tiering


src/server/tiering/decoders.h
```diff
@@ -0,0 +1,61 @@
...
+
+namespace dfly::tiering {
+
+// Decodes serialized value and provides it to callbacks.
+// Acts as generic interface to callback driver (OpManager)
+struct Decoder {
+  virtual ~Decoder() = default;
+
+  // Poor man's type-erasure copy
+  virtual std::unique_ptr<Decoder> Clone() const = 0;
```
dranikpg, line 23: 1. Either this, so the caller can pass a decoder by value, or the caller has to pass a function creating an encoder
2. I don't want to use unique_ptrs later, I want to initialize in place with proper align to avoid allocations

src/core/string_or_view.h
```diff
@@ -65,11 +65,17 @@ class StringOrView {
       val_ = std::string{std::get<std::string_view>(val_)};
   }
 
+  // Move out of value as string
   std::string Take() && {
     MakeOwned();
     return std::move(std::get<std::string>(val_));
   }
 
+  std::string* BorrowMut() {
```
romange: Borrow is from rust?
Does it make sense here? you never return, so maybe just `GetMutable` ?

romange:
Can you please expand on the PR description and explain your idea at high level please?

src/server/tiering/decoders.h
```diff
@@ -0,0 +1,61 @@
...
+  virtual void Initialize(std::string_view slice) = 0;
+
+  // Store value in compact object
+  virtual void Upload(CompactObj* obj) = 0;
+
+  bool modified = false;           // Must be set if modified (not equal to original slice)
+  size_t estimated_mem_usage = 0;  // Estimated usage if uploaded
+};
+
+struct BareDecoder : public Decoder {
```
romange, line 36: what is this specialization?

romange:
Specifically, what is the problem you are trying to solve with multiple types?

dranikpg:
> Specifically, what is the problem you are trying to solve with multiple types?

The problem specifically is:
* the io driven manager (op_manager) works on a type erased level, just keeping track of a byte slice to be read and a list of callbacks
* previously, basic string logic, like whether it's raw or not, whether it needs to be owned or can be just a slice, etc. was managed by the op_manager
* so the callback interface was tailored to that logic and the callbacks as well
* once we add support for different types, we need to somehow:
  * decode the value
  * apply changes to the value
  * track whether it was modified
  * evalueate it's memory usage to understand if it's reasonable to upload it
  * do the uploading process
 
-> The "decoder" does all of those basic operations behind a virtual interface to the op_manger can continue working as before. Additionally, the callbacks now accept their own decoder so it's easier for the to handle the value

src/server/tiering/decoders.h
romange: why can not you pass a unique_ptr<Decoder> from the caller? why passing by value just to call `Clone()` deeper in the stack?

dranikpg: Because successive callbacks don't need a new decoder. There is a single decoder for all reads on the same offset

romange: you are saying that not every `Enqueue` call allocates an instance and sometimes the already existing one is reused?

dranikpg: Yes, it has to be reused to work properly

src/server/tiering/op_manager.cc
```diff
@@ -147,16 +148,15 @@ void OpManager::ProcessRead(size_t offset, io::Result<std::string_view> page) {
...
   for (size_t i = 0; i < info->key_ops.size(); i++) {
-    bool modified = false;
     auto& ko = info->key_ops[i];
     if (page) {
-      key_value = page->substr(ko.segment.offset - info->segment.offset, ko.segment.length);
+      size_t offset = ko.segment.offset - info->segment.offset;
+      ko.decoder->Initialize(page->substr(offset, ko.segment.length));
       for (auto& cb : ko.callbacks)
-        modified |= cb(std::pair{&key_value, !modified});
+        cb(&*ko.decoder);
```
romange, line 159: I find this hard to read, consider using `Decoder* decoder()` accessor .

Issue #5513, opened 2025-07-16
Support DD option the FT.DROPINDEX


BagritsevichStepan:
We need to support DD option in the FT.DROPINDEX:
```
DD
drop index operation that, if set, deletes the actual document keys. FT.DROPINDEX index DD is an asynchronous operation.
```

PR #5885, opened 2025-10-06
feat(search): Add DD option support for FT.DROPINDEX command

vyavdoshenko:
Implements the `DD` (Delete Documents) option for `FT.DROPINDEX` command.

Fixes: https://github.com/dragonflydb/dragonfly/issues/5513

### DIFF
--- a/src/server/search/doc_index.cc
+++ b/src/server/search/doc_index.cc
@@ -242,19 +242,6 @@ size_t ShardDocIndex::DocKeyIndex::Size() const {
   return ids_.size();
 }
 
-std::vector<std::string> ShardDocIndex::DocKeyIndex::GetAllKeys() const {
-  std::vector<std::string> keys;
-  keys.reserve(ids_.size());
-
-  for (const auto& [key, id] : ids_) {
-    if (!key.empty()) {
-      keys.push_back(key);
-    }
-  }
-
-  return keys;
-}
-
 uint8_t DocIndex::GetObjCode() const {
   return type == JSON ? OBJ_JSON : OBJ_HASH;
 }
@@ -620,10 +607,6 @@ DocIndexInfo ShardDocIndex::GetInfo() const {
   return {*base_, key_index_.Size()};
 }
 
-std::vector<std::string> ShardDocIndex::GetAllKeys() const {
-  return key_index_.GetAllKeys();
-}
-
 io::Result<StringVec, ErrorReply> ShardDocIndex::GetTagVals(string_view field) const {
   search::BaseIndex* base_index = indices_->GetIndex(field);
   if (base_index == nullptr) {
@@ -662,15 +645,16 @@ void ShardDocIndices::InitIndex(const OpArgs& op_args, std::string_view name,
       });
 }
 
-bool ShardDocIndices::DropIndex(string_view name) {
+unique_ptr<ShardDocIndex> ShardDocIndices::DropIndex(string_view name) {
   auto it = indices_.find(name);
   if (it == indices_.end())
-    return false;
+    return nullptr;
 
   DropIndexCache(*it->second);
+  auto index = std::move(it->second);
   indices_.erase(it);
 
-  return true;
+  return index;
 }
 
 void ShardDocIndices::DropAllIndices() {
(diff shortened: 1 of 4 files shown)


src/server/search/search_family.cc
```diff
@@ -1131,18 +1131,59 @@ void SearchFamily::FtAlter(CmdArgList args, const CommandContext& cmd_cntx) {
...
+  if (args.size() > 1) {
+    string_view option = ArgS(args, 1);
+    // Only check for DD option, ignore other arguments for compatibility
+    if (absl::EqualsIgnoreCase(option, "DD")) {
+      delete_docs = true;
+    }
+  }
+
+  // Collect document keys (if DD is set), drop index, and delete documents in single transaction
+  vector<vector<string>> shard_doc_keys(delete_docs ? shard_set->size() : 0);
```
dranikpg: You don't need this inside the command, you can store the keys inside the transaction callback locally

src/server/search/search_family.cc
```diff
@@ -1131,18 +1131,59 @@ void SearchFamily::FtAlter(CmdArgList args, const CommandContext& cmd_cntx) {
...
-  cmd_cntx.tx->ScheduleSingleHop([&](Transaction* t, EngineShard* es) {
+
+  auto cb = [&](Transaction* t, EngineShard* es) {
+    // Step 1: If DD is set, collect all document keys before dropping the index
+    if (delete_docs) {
+      auto* index = es->search_indices()->GetIndex(idx_name);
+      if (index) {
+        shard_doc_keys[es->shard_id()] = index->GetAllKeys();
+      }
+    }
```
dranikpg: You don't need to copy all the keys, you're gonna delete the index anyway. You can just return the DocKeyIndex by value from DropIndex() and empty it either with a helper function or by moving the `ids_` map out of it directly

src/server/search/doc_index.cc
```diff
@@ -242,6 +242,19 @@ size_t ShardDocIndex::DocKeyIndex::Size() const {
   return ids_.size();
 }
 
+std::vector<std::string> ShardDocIndex::DocKeyIndex::GetAllKeys() const {
+  std::vector<std::string> keys;
+  keys.reserve(ids_.size());
+
+  for (const auto& [key, id] : ids_) {
+    if (!key.empty()) {
```
dranikpg: keys shouldn't be empty inside `ids_` (only inside `keys_`), but see previous comment on that we don't need this function

dranikpg (changes requested):

src/server/search/search_family.cc
vyavdoshenko: Fixed.

src/server/search/search_family.cc
vyavdoshenko: Fixed.

BorysTheDev:
LGTM

src/server/search/doc_index.cc

src/server/search/search_family.cc
```diff
@@ -1131,14 +1131,49 @@ void SearchFamily::FtAlter(CmdArgList args, const CommandContext& cmd_cntx) {
 
 void SearchFamily::FtDropIndex(CmdArgList args, const CommandContext& cmd_cntx) {
   string_view idx_name = ArgS(args, 0);
-  // TODO: Handle optional DD param
+
+  // Parse optional DD (Delete Documents) parameter
+  bool delete_docs = false;
```
dranikpg: `bool delete_docs = args.size() > 1 && absl::EqualsIgnoreCase(args[1], "DD") `

src/server/search/search_family.cc
```diff
@@ -1131,14 +1131,49 @@ void SearchFamily::FtAlter(CmdArgList args, const CommandContext& cmd_cntx) {
...
+      return OpStatus::OK;
+
+    num_deleted.fetch_add(1);
+
+    // If DD is set, delete all documents that were in the index
+    if (delete_docs) {
+      // Get const reference to document keys map (index will be destroyed after this scope)
+      const auto& doc_keys = index->key_index().GetDocKeysMap();
+
+      if (!doc_keys.empty()) {
```
dranikpg: no need for the if, you can just iterate over it right away

dranikpg (approved):
Some small style comments

src/server/search/search_family.cc
vyavdoshenko: Fixed

src/server/search/search_family.cc
vyavdoshenko: Fixed

Issue #5316, opened 2025-06-17
Check Del command in all families


BagritsevichStepan:
We have a bad pattern in our codebase like this:
```cpp
auto it = db_slice.FindMutable(op_args.db_cntx, key).it;  // post_updater will run immediately
if (IsValid(it)) {
   db_slice.Del(op_args.db_cntx, it);
```

This can lead to bugs, because in such cases we don't provide the expected object type to the `FindMutable` call, which means we can remove an object of a different type from the database. We should verify that everything works as expected in all such cases — and it's better to avoid this pattern altogether.

PR #5886, opened 2025-10-06
fix: Add type safety to FindMutable calls

vyavdoshenko:
Fixes: https://github.com/dragonflydb/dragonfly/issues/5316
Adds explicit type checking to `FindMutable` calls in string, set, and hash family commands to prevent accidentally deleting objects of the wrong type.

Changes:
- `string_family.cc`: `OpIncrBy`, `OpThrottle` → added `OBJ_STRING`
- `set_family.cc`: `OpAdd` → added `OBJ_SET`
- `hset_family.cc`: `HRandField` → added `OBJ_HASH`

This ensures that operations fail fast with `WRONG_TYPE` instead of potentially corrupting data by operating on objects of unexpected types.

### DIFF
--- a/src/server/hset_family.cc
+++ b/src/server/hset_family.cc
@@ -1164,9 +1164,12 @@ void HSetFamily::HRandField(CmdArgList args, const CommandContext& cmd_cntx) {
       }
 
       if (string_map->Empty()) {  // Can happen if we use a TTL on hash members.
-        // post_updater will run immediately
-        auto it = db_slice.FindMutable(db_context, key).it;
-        db_slice.Del(db_context, it);
+        // Use type-safe deletion (fixes #5316)
+        auto res_it = db_slice.FindMutable(db_context, key, OBJ_HASH);
+        if (res_it) {
+          res_it->post_updater.Run();
+          db_slice.Del(db_context, res_it->it);
+        }
         return facade::OpStatus::KEY_NOTFOUND;
       }
     } else if (pv.Encoding() == kEncodingListPack) {
--- a/src/server/string_family.cc
+++ b/src/server/string_family.cc
@@ -285,9 +285,13 @@ OpResult<int64_t> OpIncrBy(const OpArgs& op_args, string_view key, int64_t incr,
   auto& db_slice = op_args.GetDbSlice();
 
   // we avoid using AddOrFind because of skip_on_missing option for memcache.
-  auto res = db_slice.FindMutable(op_args.db_cntx, key);
+  // Use type-safe FindMutable with OBJ_STRING (fixes #5316)
+  auto res = db_slice.FindMutable(op_args.db_cntx, key, OBJ_STRING);
 
-  if (!IsValid(res.it)) {
+  if (!res) {
+    if (res.status() == OpStatus::WRONG_TYPE)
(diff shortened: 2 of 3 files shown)


src/server/string_family.cc
```diff
@@ -383,20 +384,20 @@ OpResult<array<int64_t, 5>> OpThrottle(const OpArgs& op_args, const string_view
   // Cost of this request
   const int64_t increment_ns = emission_interval_ns * quantity;  // should be nonnegative
 
-  auto res = db_slice.FindMutable(op_args.db_cntx, key);
+  // Use type-safe FindMutable with OBJ_STRING (fixes #5316)
```
dranikpg: I think those comments are redundant, especially referencing the issue

dranikpg (approved):
Ok, but I think some comments are obsolete

src/server/string_family.cc
vyavdoshenko: Fixed

src/server/hset_family.cc
```diff
@@ -1164,9 +1164,11 @@ void HSetFamily::HRandField(CmdArgList args, const CommandContext& cmd_cntx) {
       }
 
       if (string_map->Empty()) {  // Can happen if we use a TTL on hash members.
-        // post_updater will run immediately
-        auto it = db_slice.FindMutable(db_context, key).it;
-        db_slice.Del(db_context, it);
+        auto res_it = db_slice.FindMutable(db_context, key, OBJ_HASH);
+        if (res_it) {
+          res_it->post_updater.Run();
```
BorysTheDev, line 1169: why do you need to run it manually?

vyavdoshenko: AutoUpdater dtor automatically calls Run() on destruction, which has a DCHECK that verifies the key still exists in the database.
Since we call db_slice.Del() earlier in the code (which removes the key), we must manually call post_updater.Run() before deletion to update memory accounting while the key is still valid. Otherwise, when res_it goes out of scope, the destructor will call Run() on an already-deleted key and the DCHECK will fail.
This pattern is used throughout the codebase whenever a key is deleted after FindMutable().

BorysTheDev: maybe we can unify somehow this code, I mean do delete in post_updater or call run during delete.

vyavdoshenko: This pattern (post_updater.Run() + Del()) appears many times across several files in the codebase, so unifying it makes sense.

Suggested approach:
Add a new method DelMutable() to DbSlice:
```
void DbSlice::DelMutable(Context cntx, ItAndUpdater& it_updater) {
  it_updater.post_updater.Run();
  Del(cntx, it_updater.it);
}
```
This would simplify code like:
```
// From
res_it->post_updater.Run();
db_slice.Del(op_args.db_cntx, res_it->it);

// To
db_slice.DelMutable(op_args.db_cntx, *res_it);
```
However, I'd suggest doing this as a separate follow-up PR after this one, because:
- This PR focuses on type safety fixes
- Refactoring many call sites is a larger change that deserves separate review
- Mixing two different goals increases risk and review complexity

Maybe better create a separate issue for this refactoring?

PR #5890, opened 2025-10-07
refactor: Add DelMutable() helper to unify post_updater.Run() + Del() pattern

vyavdoshenko:
Follow-up: https://github.com/dragonflydb/dragonfly/pull/5886

Introduces `DbSlice::DelMutable()` method to encapsulate the common pattern of running `post_updater.Run()` before deleting a key after `FindMutable()`.

Note:
`list_family.cc` was not refactored as it uses a different pattern where deletion is conditional based on whether the list becomes empty, not unconditional.

Related to #5316

### DIFF
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -816,7 +816,7 @@ void DbSlice::Del(Context cntx, Iterator it) {
   PerformDeletion(it, db.get());
 }
 
-void DbSlice::DelMutable(Context cntx, ItAndUpdater& it_updater) {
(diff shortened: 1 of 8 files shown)


BorysTheDev:
LGTM

dranikpg (commented):
You don't need to pass it by reference - ending its lifetime is more "idiomatically" correct. It's always the final action.

vyavdoshenko:
> You don't need to pass it by reference - ending it's lifetime is more "idiomatically" correct. It's always the final action.

Fixed

dranikpg (approved):

PR #5892, opened 2025-10-07
feat(helm): add snapshot functionality with automatic cleanup

### LINKED ISSUE #2824: FR: Allow setting amount of snapshots (logrotate for snapshots)
We are trying dragonfly out with a network storage attached to it - limited to a certain size.
Since dragonfly doesn't look like it is removing old snapshots (does it?), this volume in kubernetes is getting full quickly.

I would like to be able to specify a `max_snapshot_count` flag to limit the amount of snapshots retained.

**Describe alternatives you've considered**
Not using a backing storage - not idea though.

**Additional context**
-

renan:
- Add snapshot configuration section to values.yaml with:
  - Enable/disable toggle
  - Cron schedule configuration
  - Format selection (dragonfly/rdb)
  - Cleaner settings (image, interval, max count, resources)

- Add snapshot-cleaner sidecar container to pod template:
  - Automatically removes old snapshots based on max_count
  - Supports both dragonfly (.dfs) and rdb formats
  - Configurable cleanup interval and resource limits

- Add snapshot cron arguments to main dragonfly container
- Maintains backward compatibility with existing configurations

Fixes #2824

### DIFF
--- a/contrib/charts/dragonfly/templates/_pod.tpl
+++ b/contrib/charts/dragonfly/templates/_pod.tpl
@@ -83,6 +83,13 @@ containers:
     {{- end }}
     args:
       - "--alsologtostderr"
+      {{- if .Values.snapshot.enabled }}
+      - "--snapshot_cron"
+      - "{{ .Values.snapshot.schedule }}"
+      {{- if eq .Values.snapshot.format "rdb" }}
+      - "--nodf_snapshot_format"
+      {{- end }}
+      {{- end }}
     {{- with .Values.extraArgs }}
       {{- toYaml . | trim | nindent 6 }}
     {{- end }}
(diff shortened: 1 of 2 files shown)


contrib/charts/dragonfly/templates/_pod.tpl
```diff
@@ -123,6 +130,31 @@ containers:
...
     {{- end }}
+  {{- if .Values.snapshot.enabled }}
+  - name: snapshot-cleaner
+    image: {{ .Values.snapshot.cleaner.image }}
+    command:
+      - /bin/sh
+      - -c
+      - |
+        while true; do
+          {{- if eq .Values.snapshot.format "dragonfly" }}
```
Abhra303, line 141: should be "df" instead of "dragonfly" as you are keeping only two formats "df" and "rdb" from the comment below.

contrib/charts/dragonfly/templates/_pod.tpl
```diff
@@ -123,6 +130,31 @@ containers:
...
+          # Keep only the last {{ .Values.snapshot.cleaner.max_count }} snapshots (composed of two files each)
+          ls -1t /data/dump-*.dfs | tail -n +{{ add 1 (mul .Values.snapshot.cleaner.max_count 2) }} | xargs rm -f
```
Abhra303, line 143: have you tested provisioning this container locally? Just want to be sure this script works.

contrib/charts/dragonfly/templates/_pod.tpl
```diff
@@ -83,6 +83,13 @@ containers:
     {{- end }}
     args:
       - "--alsologtostderr"
+      {{- if .Values.snapshot.enabled }}
+      - "--snapshot_cron"
+      - "{{ .Values.snapshot.schedule }}"
+      {{- if eq .Values.snapshot.format "rdb" }}
+      - "--nodf_snapshot_format"
```
Abhra303, line 90: @romange do we have this flag? I think `--df_snapshot_format=false` is the correct flag to have rdb format?

romange: it's equivalent, though I am surprised that people know this notion as it is a special format for boolean flags in google flags library :) @renan  how did you know? :)

vyavdoshenko:
Stale PR

Issue #5891, opened 2025-10-07
Improve RSS evicition to handle expired keys


mkaruza:
Expiration and eviction share same logic. When heartbeat kicks in we are always expire keys if there is eviction goal we are also removing keys. 

There is problem in current approach with RSS evicition. We introduced eviction state that aggregates deleted bytes after each heartbeat.

We update this state even if there is no eviction taking place. State variable `deleted_bytes_before_rss_update` can keep accumulating and because of that we can overflow in `CalculateEvictionBytes` - saying that we should evict even if there is no threshold reached.

Fix RSS evicition to handle expired keys and not count them until needed.

PR #5894, opened 2025-10-08
fix(evicition): Don't accumulate deleted bytes if there is no RSS eviction

mkaruza:
We start to accumulate number of deleted bytes only if there is expected
that we evict on RSS threshold. Once we stop expecting to evict and we
have cleared number of tracked deleted bytes we stop with tracking.

Fixes https://github.com/dragonflydb/dragonfly/issues/5891

### DIFF
--- a/src/server/engine_shard.cc
+++ b/src/server/engine_shard.cc
@@ -749,7 +749,11 @@ size_t EngineShard::CalculateEvictionBytes() {
 
     global_rss_memory_at_prev_eviction = global_used_rss_memory;
 
-    DCHECK_GE(global_used_rss_memory, deleted_bytes_before_rss_update * shards_count);
+    LOG_IF(DFATAL, global_used_rss_memory < (deleted_bytes_before_rss_update * shards_count))
+        << "RSS evicition underflow "
+        << "global_used_rss_memory: " << global_used_rss_memory
+        << " deleted_bytes_before_rss_update: " << deleted_bytes_before_rss_update;
+
     // If we underflow use limit as used_memory
     size_t used_rss_memory_with_deleted_bytes =
         std::min(global_used_rss_memory - deleted_bytes_before_rss_update * shards_count, limit);
@@ -761,11 +765,11 @@ size_t EngineShard::CalculateEvictionBytes() {
     // RSS evictions starts so we should start tracking deleted_bytes
     if (rss_goal_bytes) {
       eviction_state_.track_deleted_bytes = true;
-    }
-
-    // There is no RSS eviction goal and we have cleared tracked deleted bytes
-    if (!rss_goal_bytes && !deleted_bytes_before_rss_update) {
-      eviction_state_.track_deleted_bytes = false;
+    } else {
+      // There is no RSS eviction goal and we have cleared tracked deleted bytes
+      if (!deleted_bytes_before_rss_update) {
+        eviction_state_.track_deleted_bytes = false;
+      }
     }
 
     VLOG_IF(2, rss_goal_bytes > 0)


src/server/engine_shard.cc
```diff
@@ -739,26 +741,40 @@ size_t EngineShard::CalculateEvictionBytes() {
...
+    // RSS evictions starts so we should start tracking deleted_bytes
+    if (rss_goal_bytes) {
+      eviction_state_.track_deleted_bytes = true;
+    }
 
-    VLOG_IF_EVERY_N(1, rss_goal_bytes > 0, 50)
+    // There is no RSS eviction goal and we have cleared tracked deleted bytes
+    if (rss_goal_bytes && deleted_bytes_before_rss_update) {
+      eviction_state_.track_deleted_bytes = false;
+    }
```
dranikpg, line 773: I am reading this as
```c++
if (rss_goal_bytes > 0) 
  track_deleted_bytes = deleted_bytes_before_rss_update == 0
```

src/server/engine_shard.cc
```diff
@@ -690,17 +690,19 @@ void EngineShard::RetireExpiredAndEvict() {
...
 
       deleted_bytes += evicted_bytes;
       eviction_goal -= std::min(eviction_goal, evicted_bytes);
     }
   }
 
-  eviction_state_.deleted_bytes_before_rss_update += deleted_bytes;
+  // Track deleted bytes only if we expect to lower memory
+  if (eviction_state_.track_deleted_bytes)
+    eviction_state_.deleted_bytes_before_rss_update += deleted_bytes;
```
dranikpg, line 705: So this is the whole core of the issue, we counted them even when not expected to and didn't reset them

src/server/engine_shard.cc
```diff
@@ -739,26 +741,40 @@ size_t EngineShard::CalculateEvictionBytes() {
...
+              << " decrease_delete_bytes_before_rss_update: "
+              << decrease_delete_bytes_before_rss_update;
       deleted_bytes_before_rss_update -= decrease_delete_bytes_before_rss_update;
     }
 
     global_rss_memory_at_prev_eviction = global_used_rss_memory;
 
+    // If we oveflow use limit as used_memory
+    size_t used_rss_memory_with_deleted_bytes =
+        std::min(global_used_rss_memory - deleted_bytes_before_rss_update * shards_count, limit);
```
dranikpg: Can we underflow even with the fix?

src/server/engine_shard.cc
mkaruza: Comment says what it suppose to check but code doesn't follow. I'll change condition. This is stop condition for our evicition state. We stop tracking when there is nothing to be evicted & we have zeroed deleted bytes.

src/server/engine_shard.cc
mkaruza: Yes, exactly. We should start tracking deleted bytes once we need them.

src/server/engine_shard.cc
mkaruza: I don't think we can. It's either limit or something in range [0,limit]. This is used to keep parameter bounded.

dranikpg: No no, my question was more like whether we need the std::min 😀 Maybe we can keep it, but add a DCHECK that the value doesn't underflow. So we can catch this issue in tests

mkaruza: Added DCHECK

src/server/engine_shard.cc
```diff
@@ -739,26 +741,41 @@ size_t EngineShard::CalculateEvictionBytes() {
...
-                          << decrease_delete_bytes_before_rss_update;
+      VLOG(2) << "deleted_bytes_before_rss_update: " << deleted_bytes_before_rss_update
+              << " decrease_delete_bytes_before_rss_update: "
+              << decrease_delete_bytes_before_rss_update;
       deleted_bytes_before_rss_update -= decrease_delete_bytes_before_rss_update;
     }
 
     global_rss_memory_at_prev_eviction = global_used_rss_memory;
 
+    DCHECK(global_used_rss_memory >= deleted_bytes_before_rss_update);
```
dranikpg: 1. This is not the same as in the std::min
2. Please use DCHECK_GE

src/server/engine_shard.cc
```diff
@@ -739,26 +741,41 @@ size_t EngineShard::CalculateEvictionBytes() {
...
-                          << decrease_delete_bytes_before_rss_update;
+      VLOG(2) << "deleted_bytes_before_rss_update: " << deleted_bytes_before_rss_update
+              << " decrease_delete_bytes_before_rss_update: "
+              << decrease_delete_bytes_before_rss_update;
       deleted_bytes_before_rss_update -= decrease_delete_bytes_before_rss_update;
     }
 
     global_rss_memory_at_prev_eviction = global_used_rss_memory;
 
+    DCHECK_GE(global_used_rss_memory, deleted_bytes_before_rss_update * shards_count);
```
romange: nit: I would use:
```
if  (global_used_rss_memory < deleted_bytes_before_rss_update * shards_count)
   LOG(DFATAL) << 
```
so we could observe this in release mode as well.

src/server/engine_shard.cc
```diff
@@ -739,26 +741,41 @@ size_t EngineShard::CalculateEvictionBytes() {
...
+    // RSS evictions starts so we should start tracking deleted_bytes
+    if (rss_goal_bytes) {
+      eviction_state_.track_deleted_bytes = true;
+    }
 
-    VLOG_IF_EVERY_N(1, rss_goal_bytes > 0, 50)
+    // There is no RSS eviction goal and we have cleared tracked deleted bytes
+    if (!rss_goal_bytes && !deleted_bytes_before_rss_update) {
+      eviction_state_.track_deleted_bytes = false;
+    }
```
romange, line 773:
```suggestion
    if (rss_goal_bytes) {
      eviction_state_.track_deleted_bytes = true;
    } else {    
      // There is no RSS eviction goal and we have cleared tracked deleted bytes
      if (!deleted_bytes_before_rss_update) {
        eviction_state_.track_deleted_bytes = false;
      }
    }
```

src/server/engine_shard.cc
mkaruza: @romange Maybe this

```
    LOG_IF(DFATAL, global_used_rss_memory < (deleted_bytes_before_rss_update * shards_count))
        << "RSS Evicition underflow " << "global_used_rss_memory: " << global_used_rss_memory
        << " deleted_bytes_before_rss_update: " << deleted_bytes_before_rss_update;
```
shards_count is known when process starts

PR #5889, opened 2025-10-07
chore(hset_family): Listpack wrapper

dranikpg:
Create listpack wrapper. Multiple map types will be hidden behind a single interface later as we'll add more map implementations for offloaded values

PR #5896, opened 2025-10-08
feat(tiering): Serialized map

dranikpg:
You'll notice it has the same interface as #5889.

Doesn't have to be perfect or efficient, it's the first version of it. Just to exist and not be too much code at once

### DIFF
--- a/src/server/CMakeLists.txt
+++ b/src/server/CMakeLists.txt
@@ -26,7 +26,7 @@ set_property(SOURCE dfly_main.cc APPEND PROPERTY COMPILE_DEFINITIONS
 
 if (WITH_TIERING)
     SET(TX_LINUX_SRCS tiering/disk_storage.cc tiering/op_manager.cc
-      tiering/small_bins.cc tiering/external_alloc.cc tiering/serialized_map.cc tiering/decoders.cc) 
+      tiering/small_bins.cc tiering/external_alloc.cc tiering/serialized_map.cc tiering/decoders.cc)
 
     add_executable(dfly_bench dfly_bench.cc)
     cxx_link(dfly_bench dfly_parser_lib fibers2 absl::random_random redis_lib)


src/server/tiering/serialized_map.cc
```diff
@@ -0,0 +1,78 @@
+#include "server/tiering/serialized_map.h"
+
+#include <absl/base/internal/endian.h>
+
+#include "base/logging.h"
+
+namespace dfly::tiering {
+
+SerializedMap::Iterator& SerializedMap::Iterator::operator++() {
+  slice_.remove_prefix(8 + key_.size() + value_.size());
```
BorysTheDev, line 10: please add constexpr for 8

src/server/tiering/serialized_map.cc
```diff
@@ -0,0 +1,78 @@
...
+SerializedMap::Iterator::Iterator(std::string_view buffer) : slice_{buffer} {
+  Read();
+}
+
+void SerializedMap::Iterator::Read() {
+  if (slice_.empty())
+    return;
+
+  uint32_t key_len = absl::little_endian::Load32(slice_.data());
+  uint32_t value_len = absl::little_endian::Load32(slice_.data() + 4);
```
BorysTheDev, line 24: please give name to 4

src/server/tiering/serialized_map.cc
```diff
@@ -48,24 +48,26 @@ size_t SerializedMap::size() const {
   return size_;
 }
 
+constexpr size_t kLenBytes = 4;
+
 size_t SerializedMap::SerializeSize(Input input) {
-  size_t out = 4;  // 4 byte number of entries
+  size_t out = kLenBytes;  // number of entries
```
kostasrim, line 54: four * two or even better four times two sounds like math poetry 🤣

Issue #5739, opened 2025-08-28
XREAD RESP3 is not properly supported


vyavdoshenko:
Expected:
`redis-cli -3`
```
127.0.0.1:6379> XADD s * value 1234
"1756385934305-0"
127.0.0.1:6379> XREAD COUNT 1 STREAMS s 0-0
1# "s" => 
   1) 1) "1756385934305-0"
      2) 1) "value"
         2) "1234"
```
`redis-cli -2`
```
127.0.0.1:6379> XREAD COUNT 1 STREAMS s 0-0
1) 1) "s"
   2) 1) 1) "1756385934305-0"
         2) 1) "value"
            2) "1234"
```

Dragonfly:
`redis-cli -3`
```
127.0.0.1:6379> XADD s * value 1234
"1756385931345-0"
127.0.0.1:6379> XREAD COUNT 1 STREAMS s 0-0
1) 1) "s"
   2) 1) 1) "1756385931345-0"
         2) 1) "value"
            2) "1234"
```

PR #5898, opened 2025-10-09
server: Support RESP3 map response for xread

abhijat:
fixes https://github.com/dragonflydb/dragonfly/issues/5739

### DIFF
--- a/src/server/stream_family.cc
+++ b/src/server/stream_family.cc
@@ -2583,13 +2583,24 @@ void XReadGeneric2(CmdArgList args, bool read_group, Transaction* tx, SinkReplyB
       }
     }
   } else {
-    rb->StartArray(resolved_streams);
-    for (size_t i = 0; i < results.size(); i++) {
-      if (results[i].empty())
-        continue;
-      string_view key = ArgS(args, i + opts->streams_arg);
-      rb->StartArray(2);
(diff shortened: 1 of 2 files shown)


src/server/stream_family.cc
```diff
@@ -2583,13 +2583,24 @@ void XReadGeneric2(CmdArgList args, bool read_group, Transaction* tx, SinkReplyB
...
+      }
+    } else {
+      rb->StartArray(resolved_streams);
+      for (size_t i = 0; i < results.size(); i++) {
+        if (results[i].empty())
+          continue;
+        string_view key = ArgS(args, i + opts->streams_arg);
+        rb->StartArray(2);
+        StreamReplies{builder}.SendStreamRecords(key, results[i]);
+      }
```
abhijat, line 2603: There is quite a bit of duplication now between this branch and `read_group`, I will look into adding helpers in a separate PR once this one is done

abhijat:
@BorysTheDev please take a look

src/server/stream_family.cc
kostasrim: It's easy to abstract everything within the for loop with a lambda. The only difference is the statement `rb->StartArray(2)` which you can turn on/off via a bool

PR #5900, opened 2025-10-10
feat(hset_family): create HMapWrap

dranikpg:
The idea is that by letting handlers accept this wrapper, the wrapper function can have more complicated logic inside that allows to provide this value for any kind of underlying representation, be it listpack, string map or even an offloaded value (#5896 will be integrated here as well)

```c++
 auto cb = [field = args[1]](const HMapWrap& hw) -> OpResult<string> {
  if (auto it = hw.Find(field); it)
    return string{it->second};
};
OpResult<string> result = cmd_cntx.tx->ScheduleSingleHopT(WrapRO(cb));
```

### DIFF
--- a/src/server/hset_family.cc
+++ b/src/server/hset_family.cc
@@ -141,7 +141,7 @@ struct HMapWrap {
   }
 
   auto Range() const {
-    auto f = [](auto p) -> pair<string_view, string_view> { return p; };
+    auto f = [](auto p) -> pair<string_view, string_view> { return p; };  // implicit conversion
     using IT = base::it::CompoundIterator<decltype(f), detail::ListpackWrap::Iterator,
                                           StringMap::iterator>;
     auto cb = [f](auto& h) -> std::pair<IT, IT> {


src/server/hset_family.cc
```diff
@@ -109,6 +109,86 @@ pair<uint8_t*, bool> LpInsert(uint8_t* lp, string_view field, string_view val, b
...
+    using RT = optional<pair<string_view, string_view>>;
+    return visit2([key](auto& h) -> RT {
+      if (auto it = h.Find(key); it != h.end())
+        return *it;
+      return std::nullopt;
+    });
+  }
+
+  auto Range() const {
+    auto f = [](auto p) -> pair<string_view, string_view> { return p; };
```
BorysTheDev: what is the purpose of it

src/server/hset_family.cc
```diff
@@ -313,33 +336,20 @@ OpResult<StringVec> OpScan(const OpArgs& op_args, std::string_view key, uint64_t
...
-    return find_res.status();
-  }
-
-  auto it = find_res.value();
   StringVec res;
   uint32_t count = scan_op.limit * HASH_TABLE_ENTRIES_FACTOR;
-  const PrimeValue& pv = it->second;
 
-  if (pv.Encoding() == kEncodingListPack) {
+  if (auto lw = hw.Get<detail::ListpackWrap>(); lw) {
```
BorysTheDev, line 342: I would prefer to see visitor approach instead of if

src/core/detail/listpack_wrap.h
```diff
@@ -39,7 +40,7 @@ struct ListpackWrap {
 
     uint8_t *lp_ = nullptr, *ptr_ = nullptr, *next_ptr_ = nullptr;
     std::string_view key_v_, value_v_;
-    uint8_t intbuf_[2][24];
+    uint8_t (&intbuf_)[2][24];
```
BorysTheDev: IntBuf& intbuf_;

src/server/hset_family.cc
dranikpg: Do iterate over it in native iterator style `for (const auto[key, value] : hmap.Range())`

src/server/hset_family.cc
dranikpg: I don't want to change the existing code structure too much, so it's evident I don't have breaking changes. When I add the third type, I will think about how to avoid branching at all

src/server/hset_family.cc
BorysTheDev: but this function do nothing

dranikpg: Ah, I thought you mean "Range()". No, f is a mapper function. It calls the implicit conversion operator (to pair of sv) of string maps' iterator

PR #5901, opened 2025-10-10
chore(hset_family): Add mutable functions to listpack wrapper

dranikpg:
Adds mutable functions to listpack wrapper

### DIFF
--- a/src/core/detail/listpack_wrap.cc
+++ b/src/core/detail/listpack_wrap.cc
@@ -44,11 +44,17 @@ uint8_t* ListpackWrap::GetPointer() {
 }
 
 ListpackWrap::Iterator ListpackWrap::Find(std::string_view key) const {
+  if (size() == 0)
+    return end();
+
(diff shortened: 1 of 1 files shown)


src/core/detail/listpack_wrap.h
```diff
@@ -46,14 +48,19 @@ struct ListpackWrap {
...
   Iterator begin() const;
   Iterator end() const;
   size_t size() const;  // number of entries
 
  private:
-  uint8_t* lp_;
-  mutable IntBuf intbuf_;
+  uint8_t* lp_;            // the listpack itself
+  mutable IntBuf intbuf_;  // buffer for integers decoded to strings
+  bool dirty_ = false;     // whether lp_ was updated, but never retrieved with GetPointer
```
BorysTheDev, line 63: I don't see why do we need it

dranikpg: Just to catch bugs because the interface is kinda bad 🙂

PR #5902, opened 2025-10-10
fix: ZINTERSTORE/ZUNIONSTORE crash when overwriting indexed HASH/JSON keys

vyavdoshenko:
Related to #5014

Problem:
ZINTERSTORE, ZUNIONSTORE, and ZDIFFSTORE commands crash when overwriting HASH or JSON keys that are indexed by FT.SEARCH. The crash occurs because these commands don't remove the old key from search indexes before overwriting it, leading to duplicate document IDs.

Solution:
Added `RemoveKeyFromIndexesIfNeeded()` call in `PrepareZEntry()` before overwriting existing keys when `zparams.override=true`. This ensures indexed documents are properly removed before type conversion.

### DIFF
--- a/src/server/search/search_family_test.cc
+++ b/src/server/search/search_family_test.cc
@@ -3456,4 +3456,25 @@ TEST_F(SearchFamilyTest, ZsetStoreCommandsOverwriteIndexedHash) {
   EXPECT_EQ(Run({"RENAME", "dest", "y"}), "OK");
 }
 
+TEST_F(SearchFamilyTest, SetStoreCommandsOverwriteIndexedHash) {
+  Run({"FT.CREATE", "idx", "ON", "HASH", "SCHEMA", "field", "TEXT"});
+  EXPECT_THAT(Run({"SADD", "set1", "a", "b", "c"}), IntArg(3));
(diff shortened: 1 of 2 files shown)


src/server/search/search_family_test.cc
```diff
@@ -3440,4 +3440,65 @@ TEST_F(SearchFamilyTest, DropIndexWithInvalidOption) {
...
+
+  // Create a HASH document
+  EXPECT_THAT(Run({"HSET", "dest", "field", "value"}), IntArg(1));
+
+  // Create sorted sets
+  EXPECT_THAT(Run({"ZADD", "zset1", "1", "a", "2", "b"}), IntArg(2));
+  EXPECT_THAT(Run({"ZADD", "zset2", "3", "c", "4", "d"}), IntArg(2));
+
+  // ZUNIONSTORE overwrites the HASH
+  EXPECT_THAT(Run({"ZUNIONSTORE", "dest", "2", "zset1", "zset2"}), IntArg(4));
```
dranikpg: Same here as in the other PR, it could've been one test. Test code it still code and if it grows too large we loose oversight over what we have

PR #5903, opened 2025-10-10
fix: SINTERSTORE/SUNIONSTORE/SDIFFSTORE crash when overwriting indexed keys

vyavdoshenko:
Related to #5014

Problem:
SINTERSTORE, SUNIONSTORE, and SDIFFSTORE commands crash when overwriting HASH or JSON keys that are indexed by FT.SEARCH. Similar to the ZSET family issue, these commands don't remove the old key from search indexes before overwriting it.

Solution:
Added `RemoveKeyFromIndexesIfNeeded()` call in `OpAdd()` before overwriting existing keys when `overwrite=true`. This ensures indexed documents are properly removed before type conversion from HASH/JSON to SET.

### DIFF
--- a/src/server/search/search_family_test.cc
+++ b/src/server/search/search_family_test.cc
@@ -3440,52 +3440,25 @@ TEST_F(SearchFamilyTest, DropIndexWithInvalidOption) {
   Run({"DEL", "doc:1"});
 }
 
-// Test SINTERSTORE with indexed keys
-TEST_F(SearchFamilyTest, SinterstoreOverwritesIndexedHash) {
-  // Create an index on HASH documents
+TEST_F(SearchFamilyTest, SetStoreCommandsOverwriteIndexedHash) {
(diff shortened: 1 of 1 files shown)


src/server/search/search_family_test.cc
```diff
@@ -3440,4 +3440,52 @@ TEST_F(SearchFamilyTest, DropIndexWithInvalidOption) {
...
+
+  // Create a HASH document that will be overwritten
+  EXPECT_THAT(Run({"HSET", "dest", "field", "value"}), IntArg(1));
+
+  // Create sets for intersection
+  EXPECT_THAT(Run({"SADD", "set1", "a", "b", "c"}), IntArg(3));
+  EXPECT_THAT(Run({"SADD", "set2", "b", "c", "d"}), IntArg(3));
+
+  // SINTERSTORE should overwrite the HASH key
+  EXPECT_THAT(Run({"SINTERSTORE", "dest", "set1", "set2"}), IntArg(2));
```
dranikpg: You don't need to create two identical tests, you can just run both commands inside the same test setup

PR #5904, opened 2025-10-11
server: Remove unused variable and field

abhijat:
`total_flows_to_finish_partial` is calculated but unused. It seems to be related to `lsn_to_finish_partial_` field but that is unset and unused elsewhere, both are removed here, if they should have been used anywhere (log them or something similar) the PR can be changed to do that.

### DIFF
--- a/src/server/replica.cc
+++ b/src/server/replica.cc
@@ -754,11 +754,6 @@ error_code Replica::ConsumeDflyStream() {
   };
   RETURN_ON_ERR(exec_st_.SwitchErrorHandler(std::move(err_handler)));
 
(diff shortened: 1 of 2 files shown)


src/server/replica.cc
```diff
@@ -754,11 +754,6 @@ error_code Replica::ConsumeDflyStream() {
   };
   RETURN_ON_ERR(exec_st_.SwitchErrorHandler(std::move(err_handler)));
 
-  size_t total_flows_to_finish_partial = 0;
```
kostasrim, line 757: Nice!

PR #5906, opened 2025-10-12
fix(search): Result set cutoff

dranikpg:
Whenever we do a bare `@field:value` query we copy the _whole_ blocklist to a vector of docids. We'll throw most of them away - so copy only as much as is needed

### DIFF
--- a/src/core/search/index_result.h
+++ b/src/core/search/index_result.h
@@ -89,15 +89,17 @@ inline std::pair<IndexResult::DocVec, size_t> IndexResult::Take(size_t limit) {
     return {std::move(vec), size};
   }
 
-  // Numeric ranges need to be filtered and don't know their size ahead
+  // Numeric ranges need to be filtered and don't know their exact size ahead
   if (std::holds_alternative<RangeResult>(value_)) {
     auto cb = [limit](auto* range) -> std::pair<DocVec, size_t> {
       DocVec out;
-      out.reserve(range->size());
-      for (auto it = range->begin(); it != range->end(); ++it)
-        out.push_back(*it);
-      size_t total = out.size();
-      out.resize(std::min(out.size(), limit));
+      size_t total = 0;
+      out.reserve(std::min(limit, range->size()));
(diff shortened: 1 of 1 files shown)


src/core/search/index_result.h
```diff
@@ -82,18 +82,36 @@ inline IndexResult::BorrowedView IndexResult::Borrowed() const {
...
+    auto cb = [limit](auto* range) -> std::pair<DocVec, size_t> {
+      DocVec out;
+      size_t total = 0;
+      out.reserve(std::min(limit, range->size()));
+      for (auto it = range->begin(); it != range->end(); ++it) {
+        total++;
+        if (out.size() < limit)
+          out.push_back(*it);
+      }
+      return {std::move(out), total};
```
kostasrim, line 103:
```suggestion
      out.reserve(std::min(limit, range->size()));
      for (auto it = range->begin(); it != range->end(); ++it) {
        if (out.size() < limit)
          out.push_back(*it);
      }
      return {std::move(out), range->size()};
```

src/core/search/index_result.h
```diff
@@ -82,18 +82,36 @@ inline IndexResult::BorrowedView IndexResult::Borrowed() const {
...
 
-  auto cb = [](auto* set) -> DocVec {
+  // Numeric ranges need to be filtered and don't know their exact size ahead
+  if (std::holds_alternative<RangeResult>(value_)) {
+    auto cb = [limit](auto* range) -> std::pair<DocVec, size_t> {
+      DocVec out;
+      size_t total = 0;
+      out.reserve(std::min(limit, range->size()));
+      for (auto it = range->begin(); it != range->end(); ++it) {
+        total++;
```
BorysTheDev, line 99: why we do total increasing even if we don't add result into out?

kostasrim: Because we want to return the original size, not the filtered size

dranikpg: > // Numeric ranges need to be filtered and don't know their exact size ahead

See the comment above

PR #5914, opened 2025-10-14
fix(metrics): Prevent underflow in connected_clients

kishorviswanathan:
We have issues with the `dragonfly_connected_clients` under flowing and becoming a very huge value. I am not sure if this is the result of a race condition or an unbalanced number of increment/decrement calls. This is an attempt to add a check to prevent the underflow from happening.

image

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -2142,10 +2142,13 @@ void Connection::IncrNumConns() {
 }
 
 void Connection::DecrNumConns() {
-  if (IsMainOrMemcache())
-    --stats_->num_conns_main;
-  else
-    --stats_->num_conns_other;
+  if (IsMainOrMemcache()) {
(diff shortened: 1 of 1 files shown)


romange:
Does it happen often @kishorviswanathan  ? Is it easily reproducible in your environment?

kishorviswanathan:
@romange Yes, it happens quite often, but I don't know what is the reason for it or how to reproduce it.

image

romange:
Can you please run your server with `--migrate_connections=false` and report if the issue persists?

kishorviswanathan:
@romange I don't see any more connection count underflow after running with `--migrate_connections=false`. Is this like a know issue or something already fixed ?

romange:
No it's not known. I just guessed where the bug might be. Feel free to open the issue and I won't accept this PR at this point as it fixes a symptom, not a core issue.  Meanwhile - you have a workaround

Issue #5905, opened 2025-10-12
introduce auto-deadlock detection during shutdown


romange:
If Dragonfly can not shutdown due to one of its fibers being stalled, we can not investigate it even with gdb as we can not list all the fibers and their stacks.

It could be nice if during the shutdown (but probably after SAVE-on-shutdown step), Dragonfly will enable timer that will print all the stack-traces upon timer expiry. we do that in our unit tests so we could do the same in prod code as well. Note that for large datastores shutdown can actually take lots of time (flushall can take 5min for 200GB and even more) so the timer should also be enabled after flushall finishes.

so after save and after flushall in the shutdown flow.

PR #5915, opened 2025-10-15
feat(server): Add shutdown watchdog

abhijat:
Once server family completes its save on shutdown activity, a timer is started. If after 20 seconds the timer is not disarmed, stack traces for all fibers is printed to standard error, similar to what is done in test utils.

FIXES https://github.com/dragonflydb/dragonfly/issues/5905

The watchdog is implemented as a static object which is armed on object creation, arming is only done once, future calls to arm are no-op. The time before logs are printed is fixed to 20 seconds.

### Note on testing

I tested manually by adding 

```
  ThisFiber::SleepFor(60s);
```

right after the `ArmShutdownWatchdog(service_.proactor_pool());` call in `ServerFamily::Shutdown()` and fiber stack traces were printed as expected on shutting down the server after 20 seconds. 

I considered adding some kind of environment variable or configuration for testing, to make the wait for `watchdog_done` to 0 and assert that stack trace is printed, but decided against it. 

Maybe there should be a setting for the sleep duration.

### DIFF
--- a/src/server/main_service.cc
+++ b/src/server/main_service.cc
@@ -159,15 +159,13 @@ struct ShutdownWatchdog {
 };
 
 ShutdownWatchdog::ShutdownWatchdog(util::ProactorPool& pp) : pool{pp} {
-  watchdog_fb = pool.GetNextProactor()->LaunchFiber([&] {
-    util::ThisFiber::SetName("shutdown-watchdog");
-
+  watchdog_fb = pool.GetNextProactor()->LaunchFiber("shutdown_watchdog", [&] {
     if (!watchdog_done.WaitFor(20s)) {
       LOG(ERROR) << "Deadlock detected during shutdown";
       absl::SetFlag(&FLAGS_alsologtostderr, true);
       util::fb2::Mutex m;
       pool.AwaitFiberOnAll([&m](unsigned index, auto*) {
-        util::ThisFiber::SetName("shutdown-watchdog");
+        util::ThisFiber::SetName(absl::StrFormat("print_stack_fib_%u", index));
         std::unique_lock lk(m);
         LOG(ERROR) << "Proactor " << index << ":\n";
         util::fb2::detail::FiberInterface::PrintAllFiberStackTraces();
@@ -181,7 +179,7 @@ void ShutdownWatchdog::Disarm() {
   watchdog_fb.JoinIfNeeded();
 }
 
-std::optional<ShutdownWatchdog> swd = std::nullopt;
+std::optional<ShutdownWatchdog> shutdown_watchdog = std::nullopt;
 
 }  // namespace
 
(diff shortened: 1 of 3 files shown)


src/server/main_service.h
```diff
@@ -203,4 +203,8 @@ class Service : public facade::ServiceInterface {
   uint32_t loading_state_counter_ ABSL_GUARDED_BY(mu_) = 0;
 };
 
+void ArmShutdownWatchdog(util::ProactorPool& pp);
+
+void DisarmShutdownWatchdog();
```
abhijat: right now, the disarm method does not need to be part of the public api. it can be a static method in the anonymous namespace in `.cc` compilation unit for main service. But I added this here as it makes it possible to disarm the timer from outside if needed eg from `dfly_main`

src/server/main_service.cc
```diff
@@ -147,6 +147,44 @@ ABSL_FLAG(uint32_t, scheduler_background_warrant, 5,
...
+  util::fb2::Fiber watchdog_fb;
+  util::fb2::Done watchdog_done;
+  util::ProactorPool& pool;
+
+  explicit ShutdownWatchdog(util::ProactorPool& pp);
+  void Disarm();
+};
+
+ShutdownWatchdog::ShutdownWatchdog(util::ProactorPool& pp) : pool{pp} {
+  watchdog_fb = pool.GetNextProactor()->LaunchFiber([&] {
```
romange: you can use `LaunchFiber("name", cb)` notation instead of calling `util::ThisFiber::SetName`

src/server/main_service.cc
```diff
@@ -147,6 +147,44 @@ ABSL_FLAG(uint32_t, scheduler_background_warrant, 5,
...
+    util::ThisFiber::SetName("shutdown-watchdog");
+
+    if (!watchdog_done.WaitFor(20s)) {
+      LOG(ERROR) << "Deadlock detected during shutdown";
+      absl::SetFlag(&FLAGS_alsologtostderr, true);
+      util::fb2::Mutex m;
+      pool.AwaitFiberOnAll([&m](unsigned index, auto*) {
+        util::ThisFiber::SetName("shutdown-watchdog");
```
romange: lets use a differrent name

src/server/server_family.cc
```diff
@@ -1220,6 +1220,8 @@ void ServerFamily::Shutdown() {
     });
   }
 
+  ArmShutdownWatchdog(service_.proactor_pool());
```
romange: just call it after `server_family_.Shutdown();` in main_service.cc

src/server/main_service.h
romange: we can add it later if needed.

abhijat: removed these as no longer needed after move to main service

src/server/main_service.cc
abhijat: changed

src/server/main_service.cc
abhijat: renamed to `print_stack_fib_%u`

src/server/server_family.cc
abhijat: changed

PR #5919, opened 2025-10-15
fix: Allow JSON.DEBUG HELP to work without key parameter

vyavdoshenko:
`JSON.DEBUG HELP` was requiring an unnecessary key parameter due to the command registration expecting minimum 3 arguments and key position at index 2.

- Changed command registration from `-3, 2, 2` to `-2, 0, 0` (min 2 args, no key required)
- Refactored `JSON.DEBUG MEMORY` and `JSON.DEBUG FIELDS` to use `shard_set->Await()` instead of transaction-based execution
- This allows HELP to work without a key while MEMORY/FIELDS still functions correctly

### DIFF
--- a/src/server/json_family.cc
+++ b/src/server/json_family.cc
@@ -29,6 +29,7 @@
 #include "server/error.h"
 #include "server/journal/journal.h"
 #include "server/search/doc_index.h"
+#include "server/sharding.h"
 #include "server/string_family.h"
 #include "server/tiered_storage.h"
 #include "server/transaction.h"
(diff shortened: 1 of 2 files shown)


dranikpg:
> Refactored JSON.DEBUG MEMORY and JSON.DEBUG FIELDS to use shard_set->Await() instead of transaction-based execution

What was the problem with it before? I understand that it's a debug command where correctness is not a priority... but it worked fine just before, didn't it?

vyavdoshenko:
> > Refactored JSON.DEBUG MEMORY and JSON.DEBUG FIELDS to use shard_set->Await() instead of transaction-based execution
> 
> What was the problem with it before? I understand that it's a debug command where correctness is not a priority... but it worked fine just before, didn't it?

No, it didn't work. I found out when I updated the documentation.
`JSON.DEBUG HELP` didn't work

PR #5922, opened 2025-10-16
chore(server): Better hash tests (Part 1)

dranikpg:
Given that I actively change the code there, our tests are somewhat weak

### DIFF
--- a/src/server/hset_family_test.cc
+++ b/src/server/hset_family_test.cc
@@ -17,7 +17,6 @@ extern "C" {
 using namespace testing;
 using namespace std;
 using namespace util;
-using namespace boost;
 using namespace facade;
 
 namespace dfly {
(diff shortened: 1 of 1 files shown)


src/server/hset_family_test.cc
```diff
@@ -58,13 +57,76 @@ TEST_F(HSetFamilyTest, Basic) {
...
+
+    // Generate HSET commands and check how many new entries were added
+    absl::InsecureBitGen gen{};
+    while (mirror.size() < 600) {
+      vector<string> cmd = {"HSET", "hash"};
+      size_t new_values = 0;
+      for (int i = 0; i < 20; i++) {
+        string key = GetRandomHex(gen, 3);
+        string value = GetRandomHex(gen, 20, 10);
+        new_values += mirror.contains(key) ? 0 : 1;
```
BorysTheDev, line 72: I think we can add DCHECK here, because if it is true it means that generator is broken

dranikpg: What do you mean? The idea is that we will have at least a few collisions. That's the idea of the test

dranikpg: When I run it, the replies are 20 at first but degrade to about 15 at the end

BorysTheDev: does it use some noise ? because if it is a simple generator, if the result is the same it means the generator in the loop

Issue #237, opened 2022-08-12
Build a repo for RHEL/Fedora/Debian distributions 


romange:
- [ ] support building an rpm package
- [ ] support building a deb package
- [ ] host a custom repo for both rpm and deb for simpler installation of Dragonfly

romange:
a pretty extensive [blog post]( https://decovar.dev/blog/2021/09/23/cmake-cpack-package-deb-apt/) that guides on how to build and host debian packages.

romange:
aptly is to manage the repo and upload it to s3-like bucket (can be gcs as well) https://www.aptly.info/tutorial/repo/ 

createrepo is somewhat similar tool for rpm packages

iko1:
Please consider creating a snap package for the software. Publishing the software through the Snap store can help distribute the software outside the above Linux distributions.

ryanrussell:
From a devops perspective, it would be really nice to have a cross-platform bash script one-liner install. 

Whether you roll by hand, use ansible/chef/puppet, ubuntu/cent, etc... this just works.

Source: https://get.docker.com/

![image](https://user-images.githubusercontent.com/523300/184634807-af53c27a-ad5a-4ddb-8530-52f663702336.png)

romange:
Another article on how to do it: 
https://medium.com/sqooba/create-your-own-custom-and-authenticated-apt-repository-1e4a4cf0b864

McSim85:
x-posting similar issue https://github.com/dragonflydb/dragonfly/issues/3000

gesundes:
I'm very interested in the official apt repository so it can be automatically synced and used in our internal infrastructure. We install DF not only in k8s but also on VMs.

romange:
We do not have knowledge nor the bandwidth to do so. If you can help - we would very much appreciate it.

gesundes:
I made something strange (with help from GPTs, to be honest) but looks like it works:

- https://github.com/gesundes/dragonflydb_apt
- https://gesundes.github.io/dragonflydb_apt/

This apt repository contains only amd64 deb packages. It runs GitHub Actions once a day and updates the repository if a new package was released in the DragonflyDB repository.

I hope it will be useful for someone.

But still I would prefer if there will be an official repository. It will be really helpful to spread the DF usage.
Currently we mostly use KeyDB and it has it's own apt repository that is very convenient.

gesundes:
[...]

PR #5923, opened 2025-10-16
tools: Add scripts and supporting files to generate os packge repo

abhijat:
Adds scripts and action to generate OS specific packaging for debian and redhat based linux distros (deb and rpm files). 

Both repositories are available at https://packages.dragonflydb.io/

The steps in the github action are:

* download latest 5 RPM packages and latest one deb package
* sign RPM files using GPG key
* create YUM repository metadata and sign it using GPG key
* create apt repository and sign with same key
* copy the repository contents, metadata and public key to static site

The workflow is currently triggered manually, in a follow up PR it can be triggered on a new release.

FIXES https://github.com/dragonflydb/dragonfly/issues/237

### DIFF
--- a/tests/dragonfly/conftest.py
+++ b/tests/dragonfly/conftest.py
@@ -433,10 +433,11 @@ def pytest_runtest_makereport(item, call):
     if report.when == "teardown":
         call_outcome = getattr(item, "call_outcome", None)
         log_dir = item.funcargs.get("df_log_dir")
-        if report.failed:
-            copy_failed_logs(log_dir, report)
-        if call_outcome and call_outcome.failed:
-            copy_failed_logs(log_dir, call_outcome)
+        if log_dir:
+            if report.failed:
+                copy_failed_logs(log_dir, report)
+            if call_outcome and call_outcome.failed:
+                copy_failed_logs(log_dir, call_outcome)
 
 
 @pytest.fixture(scope="function")
--- a/tests/dragonfly/package_install_test.py
+++ b/tests/dragonfly/package_install_test.py
@@ -1,0 +1,33 @@
+from testcontainers.core.container import DockerContainer
+
+
+def check(container: DockerContainer, cmd: str):
+    result = container.exec(cmd)
+    assert result.exit_code == 0, f"command {cmd} failed with result {result}"
+
+
+async def test_install_package_on_fedora():
+    with DockerContainer(image="fedora:latest", tty=True) as fedora:
+        check(
+            fedora,
+            "dnf config-manager addrepo --from-repofile=https://packages.dragonflydb.io/dragonfly.repo",
+        )
+        check(fedora, "dnf -y install dragonfly")
+        check(fedora, "dragonfly --version")
+
+
+async def test_install_package_on_ubuntu():
+    with DockerContainer(image="ubuntu:latest", tty=True) as ubuntu:
+        check(ubuntu, "apt update")
+        check(ubuntu, "apt install -y curl")
+        check(
+            ubuntu,
+            "curl -Lo /usr/share/keyrings/dragonfly-keyring.public https://packages.dragonflydb.io/pgp-key.public",
+        )
+        check(
+            ubuntu,
+            "curl -Lo /etc/apt/sources.list.d/dragonfly.sources https://packages.dragonflydb.io/dragonfly.sources",
+        )
+        check(ubuntu, "apt update")
+        check(ubuntu, "apt install -y dragonfly")
+        check(ubuntu, "dragonfly --version")


abhijat:
deployment is working, next step is to add pytests for fedora and ubuntu containers to install from this repo and check if the packages work

ETA: tests are added

.github/workflows/generate-osrepo-site.yml
```diff
@@ -0,0 +1,81 @@
...
+          project_id: 'dragonflydb'
+          credentials_json: ${{ secrets.GCP_BUCKET_CREDENTIALS }}
+
+      - name: GCloud setup
+        uses: 'google-github-actions/setup-gcloud@v3'
+
+      - name: Deploy site
+        working-directory: tools/packaging/osrepos
+        run: |
+          gcloud storage rm gs://dragonfly-packages-bucket/**
```
romange: please move the bucket path to be a secret as well.  we should not leak such things to public.

abhijat: replaced with secret

tools/packaging/osrepos/README.md
```diff
@@ -0,0 +1,74 @@
...
+#### Versions in APT repository
+
+Unlike the yum repo, the apt repo only has the latest version. The reason for this is the tool, `reprepro` supplied by
+debian to build repositories only supports multiple
+versions in version 5.4 onwards, and the github runner using ubuntu-latest does not have this version.
+
+Another option would be to use the components feature of apt repositories in the sources file we ask users to install,
+but then the versions would need
+to be hardcoded in the sources file and the user would have
+to update the file with each new release whcih makes for a bad user experience. As of now users wanting older packages
```
vyavdoshenko: nit (typo)
whcih -> which

.github/workflows/generate-osrepo-site.yml
```diff
@@ -0,0 +1,81 @@
...
+          credentials_json: ${{ secrets.GCP_BUCKET_CREDENTIALS }}
+
+      - name: GCloud setup
+        uses: 'google-github-actions/setup-gcloud@v3'
+
+      - name: Deploy site
+        working-directory: tools/packaging/osrepos
+        run: |
+          gcloud storage rm ${{ secrets.GCP_PACKAGES_BUCKET }}/**
+          gcloud storage rsync $SiteRoot ${{ secrets.GCP_PACKAGES_BUCKET }} --recursive --delete-unmatched-destination-objects
```
vyavdoshenko, line 81: I don't see the fail states processing. As far as I understand, we can reach this step even if we have a broken package. I think we have to go to the next step only if the previous one succeeded.
Could you please test it? Try to break one of the previous steps.

abhijat: Do you mean if a previous step fails, deploy should fail, right?

If any step fails then the job will fail and the next steps will not be tried, this has also happened during the development of this PR many times, but I will confirm it.

tools/packaging/osrepos/README.md
abhijat: fixed

.github/workflows/generate-osrepo-site.yml
vyavdoshenko: > Do you mean if a previous step fails, deploy should fail, right?
> 
> If any step fails then the job will fail and the next steps will not be tried, this has also happened during the development of this PR many times, but I will confirm it.

Yes, I want to make sure that we will not deploy a broken package if something went wrong on one of the previous steps.

abhijat: tried it here https://github.com/dragonflydb/dragonfly/actions/runs/19426024186/job/55573645544 with commit b3d098e4 and it failed early

vyavdoshenko:
lgtm

tests/dragonfly/conftest.py
```diff
@@ -433,10 +433,11 @@ def pytest_runtest_makereport(item, call):
     if report.when == "teardown":
         call_outcome = getattr(item, "call_outcome", None)
         log_dir = item.funcargs.get("df_log_dir")
-        if report.failed:
-            copy_failed_logs(log_dir, report)
-        if call_outcome and call_outcome.failed:
-            copy_failed_logs(log_dir, call_outcome)
+        if log_dir:
```
vyavdoshenko, line 436: I recommend creating log_dir if it doesn't exist.

abhijat: let me think about this and create a follow up PR. In this case the test does not have anything to do with dragonfly process, so a log directory is not needed, but I am also not happy with this condition I added. 

Ideally this hook should be more flexible. Let me check for some other way around this.

Issue #4451, opened 2025-01-14
test_redis_replication_all fails


kostasrim:
See https://github.com/dragonflydb/dragonfly/actions/runs/12764368205/job/35576390254#step:6:822

```
            await c_master.close()
            await c_replica.close()
>           assert timeout > 0, "Timeout while waiting for replica to sync"
E           AssertionError: Timeout while waiting for replica to sync
E           assert 0 > 0
```

kostasrim:
https://github.com/dragonflydb/dragonfly/actions/runs/13022114433/job/36324777965#step:6:1177

kostasrim:
This hasn't failed for some time. If it does fail I will reopen.

vyavdoshenko:
This test fails again.
https://github.com/dragonflydb/dragonfly/actions/runs/14528984004/job/40765540786

Errors:
```
Log line format: [IWEF]yyyymmdd hh:mm:ss.uuuuuu threadid file:line] msg
E20250418 03:40:38.767102 22127 protocol_client.cc:193] Error while calling sock_->Connect(server_context_.endpoint): Connection refused
E20250418 03:40:39.267544 22127 protocol_client.cc:193] Error while calling sock_->Connect(server_context_.endpoint): Connection refused
E20250418 03:40:39.767937 22127 protocol_client.cc:193] Error while calling sock_->Connect(server_context_.endpoint): Connection refused
E20250418 03:40:40.268354 22127 protocol_client.cc:193] Error while calling sock_->Connect(server_context_.endpoint): Connection refused
E20250418 03:40:40.768642 22127 protocol_client.cc:193] Error while calling sock_->Connect(server_context_.endpoint): Connection refused
E20250418 03:40:41.269273 22127 protocol_client.cc:193] Error while calling sock_->Connect(server_context_.endpoint): Connection refused
E20250418 03:40:41.769763 22127 protocol_client.cc:193] Error while calling sock_->Connect(server_context_.endpoint): Connection refused
E20250418 03:40:41.810784 22127 protocol_client.cc:230] Could not shutdown socket system:107
```

Warnings:
```
W20250418 03:39:53.190562 22125 dfly_main.cc:786] SWAP is enabled. Consider disabling it when running Dragonfly.
W20250418 03:39:53.450364 22127 rdb_load.cc:2479] Unrecognized RDB AUX field: 'valkey-ver'
[...]

PR #5924, opened 2025-10-16
server: Yield fiber if running for over 1ms

abhijat:
When using epoll as the event loop in debug mode, it can be observed that the stream consumption fiber in stable state can monopolize the core allowing no other fibers to run. This causes CI failures when running on epoll + debug + single proactor thread.

To avoid this a yield is added similar to `ListenerInterface::RunAcceptLoop()`

FIXES https://github.com/dragonflydb/dragonfly/issues/4451

### DIFF
--- a/src/server/replica.cc
+++ b/src/server/replica.cc
@@ -693,11 +693,16 @@ error_code Replica::ConsumeRedisStream() {
   };
   RETURN_ON_ERR(exec_st_.SwitchErrorHandler(std::move(err_handler)));
 
-  facade::CmdArgVec args_vector;
+  CmdArgVec args_vector;
 
   acks_fb_ = fb2::Fiber("redis_acks", &Replica::RedisStreamAcksFb, this);
 
   while (true) {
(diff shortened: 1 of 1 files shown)


src/server/replica.cc
```diff
@@ -693,11 +693,16 @@ error_code Replica::ConsumeRedisStream() {
...
-  facade::CmdArgVec args_vector;
+  CmdArgVec args_vector;
 
   acks_fb_ = fb2::Fiber("redis_acks", &Replica::RedisStreamAcksFb, this);
 
   while (true) {
+    // Yield if the fiber has been running for long.
+    if (base::CycleClock::ToUsec(ThisFiber::GetRunningTimeCycles()) > 1000) {  // 1ms
+      ThisFiber::Yield();
+    }
```
abhijat, line 704: Its not clear what tests to add for this, perhaps https://github.com/dragonflydb/dragonfly/issues/4451 not failing (or continuing to fail) is a test

kostasrim (approved):

PR #5926, opened 2025-10-17
chore: code redundancies in replication metrics

kostasrim:
1. `replica_->GetSummary()` was called twice in `info` command. First in `GetMetrics()` and then again in `FormatInfoMetrics()`
2. There was a subtle race condition (see comment in code block below in the comments sections of this PR)
3. Replaces all occurrences of `ss->is_master` with a function `IsMaster`

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -889,6 +889,8 @@ GenericError RewriteConfigFile() {
 }
 
 bool IsMaster() {
+  // We call this function on startup where tlocal() == nullptr. We handle
+  // this case below.
   if (!ServerState::tlocal()) {
     return true;
   }


src/server/server_family.cc
```diff
@@ -3169,13 +3161,13 @@ string ServerFamily::FormatInfoMetrics(const Metrics& m, std::string_view sectio
         append("psync_attempts", rinfo.psync_attempts);
         append("psync_successes", rinfo.psync_successes);
       };
-      fb2::LockGuard lk(replicaof_mu_);
 
-      replication_info_cb(replica_->GetSummary());
+      const auto& info = *m.replica_side_info;
```
kostasrim, line 3170: We no longer call `GetSummary()` twice, only once in `GetMetrics()`

src/server/server_family.cc
```diff
@@ -3126,17 +3128,7 @@ string ServerFamily::FormatInfoMetrics(const Metrics& m, std::string_view sectio
...
-    // ensuring eventual consistency of is_master. When determining if the server is a replica and
-    // accessing the replica_ object, we must lock replicaof_mu_. Using is_master alone is
-    // insufficient in this scenario.
-    // Please note that we we do not use Metrics object here.
-    {
-      fb2::LockGuard lk(replicaof_mu_);
-      is_master = !replica_;
-    }
-    if (is_master) {
+    if (!m.replica_side_info) {
```
kostasrim, line 3136: so, we locked the replica mutex to check if the role is_master and we release the lock. This creates a small window for another thread to lock the mutex, and release the `replica_` object. By the time we lock again at line `3172`, it could be that `replica_ == nullptr` and yet we use it without a check for nullptr. Realistically speaking and I make a guess here, the chances for this happening are of planck length (and that's why we never observed a seg fault)

kostasrim: This is (2) in the description of this PR

dranikpg: Even if it's of plank length probablilty divided by the earth circumference, it's still a possible segafult that deserves a fix

src/server/server_family.cc
```diff
@@ -888,6 +888,10 @@ GenericError RewriteConfigFile() {
   return {};
 }
 
+bool IsMaster() {
+  return ServerState::tlocal() && ServerState::tlocal()->is_master;
```
dranikpg: I assume server state should never be null under normal conditions?

dranikpg (approved):
Good fixes 👍🏻 The server state/info code has always been a bit of a dark corner

src/server/server_family.cc
kostasrim: And that's why the tests failed. When we call `LoadFromSnapshot()` when df starts, `ServerState::tlocal()` is `0x00`. I patched `IsMaster()` to reflect this. 

So yes, we need `ServerState::tlocal()` or otherwise we might access a nullptr.

Imo, if this is an initialization issue I will follow up in a separate PR and clean that up as well.

src/server/server_family.cc
```diff
@@ -888,6 +888,13 @@ GenericError RewriteConfigFile() {
   return {};
 }
 
+bool IsMaster() {
+  if (!ServerState::tlocal()) {
```
romange, line 894: maybe add a comment that mentioning the corner case you saw?
so that next time someone won't try to remove it?

romange (dismissed):

romange (approved):

PR #5927, opened 2025-10-17
chore: non preemptive ProtocolClient destructor

kostasrim:
Preempting in the destructor of ProtocolClient is error prone when a thread local shared_ptr releases the underline object as part of an (operator=) assignment statement. The assignment on the thread local will block while the ProtocolClient is released and another fiber can grab a copy of it while the object is in a mixed state.

Furthermore, there is a data race in `GetSummary()` and `ProtocolClient::Sock()` initialization. Even though `Sock()` is not yet initialized, we know prior on which proactor its fiber will run. Therefore, we can remove the call to `Sock()` and dispatch the callback directly to socket the thread avoiding the data race.

* remove preemption from ~ProtocolClient
* fix rare race condition  last_io_time_ by making it atomic
* always dispatch in socket thread for Replica::GetSummary()

### DIFF
--- a/src/server/cluster/outgoing_slot_migration.cc
+++ b/src/server/cluster/outgoing_slot_migration.cc
@@ -161,6 +161,8 @@ void OutgoingMigration::Finish(const GenericError& error) {
   }
 
   bool should_cancel_flows = false;
+  absl::Cleanup on_exit([this]() { CloseSocket(); });
+
   {
     util::fb2::LockGuard lk(state_mu_);
     switch (state_) {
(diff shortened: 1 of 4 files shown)


kostasrim:
@romange you asked here to deprecate `info_replication_valkey_compatible` https://github.com/dragonflydb/dragonfly/pull/5864#pullrequestreview-3337607935

To clarify, you mean `absl deprecate` the flag and remove it ?

src/server/cluster/outgoing_slot_migration.cc
```diff
@@ -313,6 +315,7 @@ void OutgoingMigration::SyncFb() {
     break;
   }
 
+  CloseSocket();
```
kostasrim, line 318: @BorysTheDev is there any other place that I forgot to CloseSocket() ?

BorysTheDev: I don't think so

Issue #5707, opened 2025-08-21
[AFL++] Base build integration (CMake) and local runner


vyavdoshenko:
Scope
- Add a CMake opt-in flag to build fuzz targets (e.g., `-DDFLY_FUZZ=ON`).
- Provide AFL++ toolchain wiring for fuzz targets (`afl-clang-fast`/`afl-clang-lto`).
- Keep regular builds unaffected when the flag is OFF.

Tasks
- [ ] Introduce `DFLY_FUZZ` CMake option and guard fuzz targets behind it.
- [ ] Implement at least one sample target to validate toolchain wiring.
- [ ] Create local runner with sane defaults and overridable params.

Resources
- AFL++ docs: `https://github.com/AFLplusplus/AFLplusplus`

Issue #5931, opened 2025-10-20
Dragonfly crashes in dragonfly_connection


vyavdoshenko:
Crash log:
```
F20251020 18:26:03.829608 152910 dragonfly_connection.cc:1771] Check failed: RespExpr::STRING == arg.type ( vs. ) 
*** Check failure stack trace: ***
    @     0x61315d16f96a  google::LogMessage::SendToLog()
    @     0x61315d1679ee  google::LogMessage::Flush()
    @     0x61315d1693df  google::LogMessageFatal::~LogMessageFatal()
    @     0x61315cbfa2ad  facade::Connection::FromArgs()
    @     0x61315cbfa639  _ZN4absl12lts_2025051219functional_internal12InvokeObjectIZN6facade10Connection10ParseRedisEjEUlvE0_NS4_13MessageHandleEJEEET0_NS1_7VoidPtrEDpNS1_8ForwardTIT1_E4typeE
    @     0x61315cbfe1de  facade::Connection::DispatchSingle()
    @     0x61315cbfe83b  facade::Connection::ParseRedis()
    @     0x61315cbff87f  facade::Connection::ConnectionFlow()
    @     0x61315cc00f76  facade::Connection::HandleRequests()
    @     0x61315cc792ce  util::ListenerInterface::RunSingleConnection()
    @     0x61315cc798d4  _ZN5boost7context6detail11fiber_entryINS1_12fiber_recordINS0_5fiberEN4util3fb219FixedStackAllocatorEZNS6_6detail15WorkerFiberImplIZZNS5_17ListenerInterface13RunAcceptLoopEvENKUlvE_clEvEUlvE_JEEC4IS7_EESt17basic_string_viewIcSt11char_traitsIcEENS6_13FiberPriorityERKNS0_12preallocatedEOT_OSC_EUlOS4_E_EEEEvNS1_10transfer_tE
    @     0x61315cd94637  make_fcontext
*** SIGABRT received at time=1760973963 on cpu 0 ***
PC: @     0x73cd712f2b2c  (unknown)  pthread_kill
[1]    152908 IOT instruction (core dumped)  ./dragonfly --dbfilename= --logtostderr 2>&1
```

How to reproduce:
```
nc localhost 6379 < ./crash.txt
```

[crash.txt](https://github.com/user-attachments/files/23004170/crash.txt)

PR #5932, opened 2025-10-20
feat: Add AFL++ fuzzing integration with persistent mode

vyavdoshenko:
Closes: https://github.com/dragonflydb/dragonfly/issues/5707
First crash that was found with the current approach: https://github.com/dragonflydb/dragonfly/issues/5931
Another one: https://github.com/dragonflydb/dragonfly/issues/5949

Adds AFL++ fuzzing support to test the full Dragonfly server with a real TCP stack.


fuzz/FUZZING.md
```diff
@@ -0,0 +1,44 @@
+# AFL++ Fuzzing for Dragonfly
+
+## Install AFL++
+
+```bash
+# From package manager (Ubuntu/Debian)
+sudo apt update
+sudo apt install afl++
```
BorysTheDev, line 8: do we support only ubuntu?

BorysTheDev (dismissed):

fuzz/FUZZING.md
vyavdoshenko: > do we support only ubuntu?

No, you can use any Linux you like. You can build it from sources: https://aflplus.plus/building/

src/server/dfly_main.cc
```diff
@@ -860,9 +870,92 @@ Usage: dragonfly [FLAGS]
...
-    dfly::RunEngine(pool.get(), &acceptor);
+#ifdef USE_AFL
+    // AFL++ persistent mode: server in thread, fuzzing in main
+    std::thread server_thread([&]() {
+      dfly::RunEngine(pool.get(), &acceptor);
+      pool->Stop();
+    });
+
+    // Wait for server
+    bool server_ready = false;
```
romange: 1. can you please factor out into a dedicated function?
2. Also please add comments in that function explaining what it does.

dranikpg: Maybe even more than a single dedication function...

fuzz/run_fuzzer.sh
```diff
@@ -0,0 +1,117 @@
...
+    if [[ -f "$DICT_FILE" ]]; then
+        AFL_CMD+=(-x "$DICT_FILE")
+    fi
+
+    AFL_CMD+=(
+        --
+        "$FUZZ_TARGET"
+        --port=6379
+        --logtostderr
+        --proactor_threads=2
```
kostasrim: Will this imply that dragonfly only runs with 2 proactor threads ? If that's the case, maybe worth parametarizing ?

src/server/dfly_main.cc
kostasrim: This looks like instrumentation code. From a quick look, we redirect `stdin` to the server. Similarly for server output. 

Why can't this be done outside `dfly_main.cc` and from a python script ?

src/server/dfly_main.cc
```diff
@@ -860,9 +870,92 @@ Usage: dragonfly [FLAGS]
...
+          // Just read once - don't wait for full response
+          char r[4096];
+          recv(s, r, sizeof(r), 0);  // Single read, timeout after 100ms
+        }
+        close(s);
+      }
+    }
+    // AFL loop finished
+    LOG(INFO) << "AFL++: Loop finished after " << loop_iteration << " iterations, exiting...";
+    _exit(0);
```
kostasrim, line 1039: `_exit` ?? 🤔

vyavdoshenko: Yes, I added a comment about it. This is part of integration. For example, if I don't have it, I can't even start the fuzzer because of the dry run section.

fuzz/run_fuzzer.sh
vyavdoshenko: I added the parameter.
Fixed.

src/server/dfly_main.cc
vyavdoshenko: The biggest problem is __AFL_LOOP must be part of main function.
I extracted part of the code to separate a helper function.


@kostasrim 
> Why can't this be done outside dfly_main.cc and from a python script ?
This idea was implemented in https://github.com/dragonflydb/df-afl
However, there is no way to catch crashes and hangs. The process should be fed by a fuzzer and controlled by the fuzzer. Using this approach, I can get the saved data and get all statistics for the session start. The next step - CI integration.

kostasrim: Maybe something like this would work https://github.com/jwilk/python-afl ? 

I don't know much about afl but based on the code I read the actual fuzzing part is done outside the scope of the C++ program. Therefore, the driver which is the C++ code you added should, in principle work with other languages as well 🤷

vyavdoshenko: No, it means that this instrumentalization is for a Python program. 
From the readme:
```
This is experimental module that enables [American Fuzzy Lop](https://lcamtuf.coredump.cx/afl/) fork server and instrumentation for pure-Python code.
```

It can help do the same as we do with a pure Python program using this module.
Also, it doesn't support multithreading.
```
Bugs
Multi-threaded code is not supported.
```

src/server/dfly_main.cc
```diff
@@ -860,9 +960,82 @@ Usage: dragonfly [FLAGS]
...
+              << " iterations...";
+
+    // Configure stdin for AFL++ input
+    ConfigureStdinNonBlocking();
+
+    unsigned int loop_iteration = 0;
+
+    // AFL++ persistent mode loop - this macro MUST stay in main() for proper instrumentation
+    // The loop limit is configurable via --afl_loop_limit flag
+    while (__AFL_LOOP(afl_loop_limit)) {
```
romange, line 1030: are you saying we can not move lines 995 - 1034 to a helper function that is called from here?
besides that - lgtm

src/server/dfly_main.cc
```diff
@@ -640,6 +657,89 @@ void RegisterBufRings(ProactorPool* pool) {
...
+}
+
+// Sends fuzzing input to the Dragonfly server and reads the response.
+// This executes one fuzzing iteration by:
+// 1. Creating a TCP socket connection to the server
+// 2. Sending the fuzzed data
+// 3. Reading a response (with timeout to prevent hangs)
+// The function uses short timeouts to keep fuzzing fast and prevent AFL++ from stalling.
+void SendFuzzInputToServer(uint16_t port, const char* data, ssize_t len) {
+  int s = socket(AF_INET, SOCK_STREAM, 0);
```
romange, line 721: are you sure you want to open a socket upon each input send to the server? don't you want to cache it between iterations to make it faster?

vyavdoshenko: I suggest adding a parameter for saving the same connection for further use in one of the following PRs. However, I think we have at least two arguments why this approach can be more preferable: 1. We can also test different connections. 2. The previous session can be finished in an inconsistent state. It will be hard to reproduce. Even now, I struggle with reproducing cases. I suggest leaving this approach, because it works for now. Anyway, this approach should be improved a little bit later, and your suggestion is one of these directions.

src/server/dfly_main.cc
vyavdoshenko: I extracted more code out of main. Please take a look.

PR #5933, opened 2025-10-20
chore(server): Allow using background fibers for DEBUG OBJHIST

dranikpg:
* Allows using background fibers with `DEBUG OBJHIST`
* Changes base behaviour from counting operations to using `GetRunningTimeCycles` 

-> Should make it safer to run it on busy datastores

I though of adding the same functionality to huffman compuations, but they're quite stateful/conditional, so it would've required more work 😞

### DIFF
--- a/src/server/cluster/cluster_family.cc
+++ b/src/server/cluster/cluster_family.cc
@@ -116,9 +116,9 @@ ClusterShardInfo ClusterFamily::GetEmulatedShardInfo(ConnectionContext* cntx) co
                         .replicas = {},
                         .migrations = {}};
 
-  optional<Replica::Summary> replication_info = server_family_->GetReplicaSummary();
+  optional<Metrics::ReplicaInfo> repl_info = server_family_->GetReplicaSummary();
   ServerState& etl = *ServerState::tlocal();
-  if (!replication_info.has_value()) {
+  if (!repl_info) {
     DCHECK(etl.is_master);
     std::string cluster_announce_ip = absl::GetFlag(FLAGS_cluster_announce_ip);
     std::string preferred_endpoint =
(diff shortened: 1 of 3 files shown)


dranikpg:
Tested it with `DEBUG POPULATE A LOT` and running memtier traffic. With background fibers QPS drops <5%, regular drops ~20%. Operation takes 21s vs 15s

(Remember that scheduler background fiber settings can be chagned at runtime as well)

src/server/debugcmd.cc
```diff
@@ -579,6 +549,46 @@ IOStat& IOStat::operator-=(const IOStat& other) {
...
+    bc->Add(1);
+    util::ProactorBase* dest = shard_set->pool()->at(i);
+
+    auto cb = [f /* copy per thread */, bc, cntx, background]() mutable {
+      auto* shard = EngineShard::tlocal();
+      auto& db_slice = cntx->ns->GetDbSlice(shard->shard_id());
+
+      for (unsigned i = 0; i < db_slice.db_array_size(); ++i) {
+        boost::intrusive_ptr<DbTable> dbt = db_slice.CopyDBTablePtr(i);
+        if (dbt.get() == nullptr)
```
romange:
```suggestion
        if (!dbt)
```

src/server/main_service.cc
```diff
@@ -967,6 +967,7 @@ void Service::Init(util::AcceptServer* acceptor, std::vector<facade::Listener*>
   config_registry.RegisterMutable("max_segment_to_consider");
   config_registry.RegisterMutable("pipeline_squash");
   config_registry.RegisterMutable("lua_mem_gc_threshold");
+  config_registry.RegisterMutable("background_debug_jobs");
```
romange, line 970: can you please add same support for "dfly expire" ? maybe rename  `background_debug_jobs` to `background_jobs` ?

dranikpg: DFLY EXPIRE is a transactional command. I can make it non-transactional and move the traversal to some common header, but I'd suggest to do so in a different PR

romange: we should make it non-transactional.

romange:
LGTM, make sure tests pass

PR #5936, opened 2025-10-21
fix(set_family): Don't replicate SPOP if nothing was removed

mkaruza:
If nothing is removed with SPOP command we are still rewriting command as SREM with invalid syntax. This replicated SREM command fails on replica side and we don't increment LSN. Fixed by skipping replication if SPOP didn't remove anything.

Enable logging on replica if command verification failed.

### DIFF
--- a/src/server/replica.cc
+++ b/src/server/replica.cc
@@ -995,7 +995,12 @@ void DflyShardReplica::StableSyncDflyReadFb(ExecutionState* cntx) {
         // lsn of the master and this lsn entry will be lost.
         journal_rec_executed_.fetch_add(1, std::memory_order_relaxed);
       } else {
-        LOG(DFATAL) << "ExecuteTx() on replica should be successful.";
+        // We only report DFATAL:
+        // 1. Context is running
+        // 2. We are ACTIVE global state
+        if (cntx->IsRunning() && ((*ServerState::tlocal()).gstate() == GlobalState::ACTIVE)) {
+          LOG(DFATAL) << "ExecuteTx() on replica should be successful.";
+        }
       }
     }
 


src/server/replica.cc
```diff
@@ -994,6 +994,8 @@ void DflyShardReplica::StableSyncDflyReadFb(ExecutionState* cntx) {
         // inconsistent data because the replica will resume from the next
         // lsn of the master and this lsn entry will be lost.
         journal_rec_executed_.fetch_add(1, std::memory_order_relaxed);
+      } else {
+        LOG(ERROR) << "ExecuteTx() on replica should be successful.";
```
romange: maybe `LOG(DFATAL)`

mkaruza: Changed.

src/server/main_service.cc
```diff
@@ -1451,7 +1451,8 @@ DispatchResult Service::DispatchCommand(ArgSlice args, SinkReplyBuilder* builder
   }
 
   if (auto err = VerifyCommandState(cid, args_no_cmd, *dfly_cntx); err) {
-    LOG_IF(WARNING, cntx->replica_conn) << "VerifyCommandState error: " << err->ToSv();
+    LOG_IF(WARNING, cntx->replica_conn || !cntx->conn() /* no owner in replica context */)
```
kostasrim, line 1454: I would also add in `replica.cc`:

```
   998       const bool is_successful = ExecuteTx(std::move(*tx_data), cntx);                                                                                                                                     
   999       if (is_successful) {                                                                                                                                                                                   1000         // We only increment upon successful execution of the transaction.                                                                                                                                 
  1001         // The reason for this is that during partial sync we sent this                                                                                                                                    
  1002         // number as the lsn number to resume from. However, if for example                                                                                                                                
  1003         // we increment this when a command fails (because the context                                                                                                                                     
  1004         // got cancelled, e.g, replication connection broke), we will get                                                                                                                                  
  1005         // inconsistent data because the replica will resume from the next                                                                                                                                 
  1006         // lsn of the master and this lsn entry will be lost.                                                                                                                                              
  1007         journal_rec_executed_.fetch_add(1, std::memory_order_relaxed);                                                                                                                                     
  1008       } 
```

If `is_successful` is `false`, `LOG(WARNING)` or even `LOG(ERROR)` as an extra safety net around failed to execute commands in the replica

src/server/set_family.cc
```diff
@@ -950,13 +950,13 @@ OpResult<StringVec> OpPop(const OpArgs& op_args, string_view key, unsigned count
...
   // Remove selected members
-  bool is_empty = RemoveSet(db_cntx, result, &co).second;
+  auto [removed, is_empty] = RemoveSet(db_cntx, result, &co);
   find_res->post_updater.Run();
 
   CHECK(!is_empty);
 
   // Replicate as SREM with removed keys, because SPOP is not deterministic.
-  if (op_args.shard->journal()) {
+  if (removed && op_args.shard->journal()) {
```
kostasrim, line 959: I guess it doesn't really worth to write a test ?

We have in python a `journal rewrite` test. 

Check `replication_test.py` and grep `test_rewrites`

mkaruza: i tried to write simple test for reproduction, but didn't succeed. I give it another try.

src/server/main_service.cc
mkaruza: It's added.

src/server/set_family.cc
kostasrim: don't stress over it if it's not simple

romange (approved):

PR #5937, opened 2025-10-21
fix: Handle non-STRING types in ParseRedis to prevent crash

vyavdoshenko:
Fixes [#5931](https://github.com/dragonflydb/dragonfly/issues/5931)

Server crashed with `Check failed: RespExpr::STRING == arg.type` when receiving malformed RESP protocol containing empty arrays (`*0\r\n`). The parser accepted these as valid input, creating ARRAY type RespExpr objects, but `FromArgs()` expected only STRING types.

Solution:
Added validation in `ParseRedis()` to check that the first argument is a STRING type before dispatching. If a non-STRING type is encountered (e.g., ARRAY from empty arrays), treat it as a protocol error (`BAD_ARRAYLEN`) and reject the request.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1241,6 +1241,8 @@ Connection::ParserStatus Connection::ParseRedis(unsigned max_busy_cycles) {
         break;
       }
 
+      DVLOG(2) << "Got Args with first token " << ToSV(tmp_parse_args_.front().GetBuf());
+
       if (io_req_size_hist)
         io_req_size_hist->Add(request_consumed_bytes_);
       request_consumed_bytes_ = 0;


src/facade/dragonfly_connection.cc
```diff
@@ -1231,6 +1231,16 @@ Connection::ParserStatus Connection::ParseRedis(unsigned max_busy_cycles) {
...
+      // Check if the first argument is a STRING type. Commands must be strings.
+      // If we get a non-STRING type (e.g., ARRAY from empty array *0), it's a protocol error.
+      if (tmp_parse_args_.front().type != RespExpr::STRING) {
+        LOG(WARNING) << "Invalid command - expected STRING type, got type: "
+                     << tmp_parse_args_.front().type;
+        // Treat this as a parser error
+        result = RedisParser::BAD_STRING;
+        break;
+      }
+
```
romange, line 1243: please remove the condition at line 1244.

vyavdoshenko: Fixed

romange (approved):

Issue #5512, opened 2025-07-16
Provide more statistics in the FT.INFO command


BagritsevichStepan:
More statistics is provided here: https://redis.io/docs/latest/commands/ft.info/

For now we are providing only general statistics and we can add more statistics in the future

romange:
Specifically, we need `default_score` otherwise `redis-go` client panicks. 

Definition of done: redis-go succeeds to run ft.info.

PR #5941, opened 2025-10-22
search: Return a default_score with ft.info index definition

abhijat:
Some clients expect this field to be present. Default score is not a part of data structures within the search family, but returning a value of 1 is a sensible default.

An additional field `index_options` is added which is an empty array for now. It should return attributes which are set for an index such as whether to store field flags within the index, an empty array is returned so as not to make client libs crash when the field is missing. The attributes do not seem relevant to our implementation so an empty array seems correct. 

fixes https://github.com/dragonflydb/dragonfly/issues/5512

### DIFF
--- a/src/server/cluster/outgoing_slot_migration.cc
+++ b/src/server/cluster/outgoing_slot_migration.cc
@@ -161,6 +161,8 @@ void OutgoingMigration::Finish(const GenericError& error) {
   }
 
   bool should_cancel_flows = false;
+  absl::Cleanup on_exit([this]() { CloseSocket(); });
+
   {
     util::fb2::LockGuard lk(state_mu_);
     switch (state_) {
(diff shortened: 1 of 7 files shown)


src/server/search/search_family.cc
```diff
@@ -1210,11 +1210,13 @@ void SearchFamily::FtInfo(CmdArgList args, const CommandContext& cmd_cntx) {
 
   rb->SendSimpleString("index_definition");
   {
-    rb->StartCollection(2, RedisReplyBuilder::MAP);
+    rb->StartCollection(3, RedisReplyBuilder::MAP);
     rb->SendSimpleString("key_type");
     rb->SendSimpleString(info.base_index.type == DocIndex::JSON ? "JSON" : "HASH");
     rb->SendSimpleString("prefix");
     rb->SendSimpleString(info.base_index.prefix);
+    rb->SendSimpleString("default_score");
```
kostasrim, line 1218: So we just return this to keep compatibility ?

abhijat: Yes, otherwise the go-redis client panics, but it looks like there need to be more fields to keep it from panicing. I will add them in this PR

PR #5943, opened 2025-10-22
chore(dfly_bench): perf improvements and support for sequential access

romange:
1. batch pipelined requests together. For example on c8g.2xlarge with pure pipeline load: `./dfly_bench --command=PING  -c 1 --qps 0 -n 50000000 --pipeline=5000  --proactor_threads=1  --key_maximum=1` reaches 1.07M qps before the change and 2.4M qps after the change.

2. Add --random_data flag that is on by default to preserve current behavior.


romange:
@mkaruza  ping

mkaruza (approved):

Issue #5314, opened 2025-06-16
huge SBF serialization/deserialization


BorysTheDev:
For now, we do RDB load/save operations for SBF > 512 MB via serialization in a huge string (fast fix  #5313), we need to add support for huge RDB load/save operations via chunks < 512MB

EricHayter:
I'd like to work on this issue. From what I see, SBFs are already chunked at the filter level, but a single filter can still exceed the RDB string size limit. My plan is to split individual filters further if they exceed the maximum allowed size. Would using 512 MB - 1 byte as the chunk size be the right approach here, or would it be preferable to use a smaller value like 256 MB?

romange:
I think the constant limit is not critical - just introduce a reasonable constant. I do not think SBFs are chunked at the filter level. At least it's not relevant for this issue. Multiple bloom filters may be created when SBF grows but they can still be arbitrary big, larger than 256MB.

PR #5944, opened 2025-10-23
fix: Implement chunking for large SBF filters

EricHayter:
This commit adds chunking functionalities for load/save operations of bloom filters. Additional information is added in the serialization of each filter. Specifically, when saving each filter the total size of the filter is written followed by chunks of the filter (max size of 64 MB per chunk).

Addresses #5314.

### DIFF
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -54,7 +54,6 @@ extern "C" {
 ABSL_DECLARE_FLAG(int32_t, list_max_listpack_size);
 ABSL_DECLARE_FLAG(int32_t, list_compress_depth);
 ABSL_DECLARE_FLAG(uint32_t, dbnum);
-ABSL_DECLARE_FLAG(bool, rdb_sbf_chunked);
 ABSL_FLAG(bool, rdb_load_dry_run, false, "Dry run RDB load without applying changes");
 ABSL_FLAG(bool, rdb_ignore_expiry, false, "Ignore Key Expiry when loding from RDB snapshot");
 


EricHayter:
I've done some manual testing to determine at what capacity and error rate would induce chunking on filters. However, when I implement the same sequences of commands specifically `BF.RESERVE` for that large of a filter (> 64 MB) the `RUN` call fails:
`E20251022 20:57:24.968197 8972 rdb_load.cc:2152] Error while calling HandleAux(): Out of memory, or used memory is too high` 
 
 What would be a good way of testing the functionality in the unit tests?

romange:
I see the tests pass, can you please push the test that fails so I could advice?
I think explicit overriding of `max_memory_limit` limit might help...

EricHayter:
Got the test working. `max_memory_limit` fixed it. Thanks.

romange:
Hi @EricHayter , thanks for doing this. The problem with this PR is that it breaks compatibility with the previous versions.
newer versions of dragonfly won't be able to load the "old" format during the replication or snapshotting.

A possible way to address is to do the following:
1. To introduce `RDB_TYPE_SBF2`
2. On the loading side to support both the previous format and the current one.
3. On the saving side to add `rdb_sbf_chunked` flag that will default to false. In the unit test that you added set it to true. Above the flag add a comment: "to flip its value to 'true' in March 2026", so we would have enough versions released that can already load the new format (to prevent broken rollbacks, if say someone updates dragonfly but wants to rollback - they won't be able to if we just flip to a new format now).

This way, in production we will still use the old encoding for 5-6 months and we will have enough released versions that support the new format. then we can just flip the flag in the future and slowly retire the old format. 

yes, wire formats are complicated to change.

src/server/rdb_load.cc
```diff
@@ -1874,7 +1878,27 @@ auto RdbLoaderBase::ReadSBF() -> io::Result<OpaqueObj> {
     unsigned hash_cnt;
     string filter_data;
     SET_OR_UNEXPECT(LoadLen(nullptr), hash_cnt);
-    SET_OR_UNEXPECT(FetchGenericString(), filter_data);
+
+    if (absl::GetFlag(FLAGS_rdb_sbf_chunked)) {
```
romange: I do not think you need to use `FLAGS_rdb_sbf_chunked` in the loader.
The load part should only rely on `RDB_TYPE_SBF2` and `RDB_TYPE_SBF`  to decide how to parse. it should be stateless in this sense.

EricHayter: My mistake, that was a typo. It's now been fixed.

src/server/rdb_load.cc
```diff
@@ -54,6 +54,7 @@ extern "C" {
 ABSL_DECLARE_FLAG(int32_t, list_max_listpack_size);
 ABSL_DECLARE_FLAG(int32_t, list_compress_depth);
 ABSL_DECLARE_FLAG(uint32_t, dbnum);
+ABSL_DECLARE_FLAG(bool, rdb_sbf_chunked);
```
romange: please remove this as well

Issue #5947, opened 2025-10-23
`test_disconnect_master` fails with invalid epoll socket fd value


abhijat:
https://github.com/dragonflydb/dragonfly/actions/runs/18739074284/job/53451526640#step:6:781

`test_disconnect_master` in `tests/dragonfly/redis_replication_test.py`

Right after transitioning to stable sync, the following error is seen

```
I20251023 06:34:50.881170 22822 replica.cc:687] Transitioned into stable sync
W20251023 06:34:50.905390 22822 common.cc:341] ReportError: Broken pipe
W20251023 06:34:50.905582 22822 protocol_client.cc:239] Socket error: Broken pipe in localhost:1114, socket info: socket not found in /proc/net/tcp or /proc/net/tcp6
E20251023 06:34:50.906875 22822 protocol_client.cc:229] Could not shutdown socket system:107
F20251023 06:34:50.923897 22822 epoll_socket.cc:397] Check failed: fd_ >= 0 (-1 vs. 0)
```

Possible cause: during the yield (which was recently added in https://github.com/dragonflydb/dragonfly/pull/5924) the socket disconnects. When the fiber is resumed, the socket fd is invalid and the next read triggers an assert failure.

abhijat:
> Possible cause: during the yield (which was recently added in https://github.com/dragonflydb/dragonfly/pull/5924) the socket disconnects. When the fiber is resumed, the socket fd is invalid and the next read triggers an assert failure.

Maybe this is not the case, I tried adding sleep instead of yield to give connection a chance to close while the consume-stream fiber is suspended, but it does not trigger the same error.

abhijat:
The broken pipe error seems to be during `RedisStreamAcksFb`, that is the most likely place to encounter this, as we write to the master regularly after entering stable sync.

To make this more frequent the sleep in python test when the connection is down was increased from 2 seconds to 20 seconds.

abhijat:
> Maybe this is not the case, I tried adding sleep instead of yield to give connection a chance to close while the consume-stream fiber is suspended, but it does not trigger the same error.

I was not running in the test in epoll mode 😫 

The error triggers in epoll mode quite easily. The sequence of events is

* acks fiber detects broken pipe or conn reset
* acks fb triggers error handler, shuts down socket, closes socket
* closed socket sets fd to -1
* the consume-stream fiber is activated, tries to read from the same socket and check is triggered

kostasrim:
[...]

PR #5948, opened 2025-10-23
server: Exit consume-redis-stream loop on error

abhijat:
Recently a yield was added to the consume-redis-stream loop in https://github.com/dragonflydb/dragonfly/pull/5924. 

It is possible that when the fiber is scheduled again, things have changed in the interim, likely because the redis acks fiber detected an error while writing (such as peer disconnected) and caused a shutdown of the socket. 

If so we now exit from the loop instead of trying to read from a possibly closed socket.

fixes https://github.com/dragonflydb/dragonfly/issues/5947

### DIFF
--- a/src/server/replica.cc
+++ b/src/server/replica.cc
@@ -707,11 +707,10 @@ error_code Replica::ConsumeRedisStream() {
     // If the acks-fb or something else triggered a shutdown, then do not attempt to read from the
     // stream.
     if (!exec_st_.IsRunning()) {
-      if (exec_st_.IsError()) {
-        LOG_REPL_ERROR("Stopping stream consumer in phase "
-                       << GetCurrentPhase()
-                       << " because of external error: " << exec_st_.GetError().Format());
-      }
+      DCHECK(exec_st_.IsError());
+      LOG_REPL_ERROR("Stopping stream consumer in phase "
+                     << GetCurrentPhase()
+                     << " because of external error: " << exec_st_.GetError().Format());
       acks_fb_.JoinIfNeeded();
       return exec_st_.GetError();
     }


src/server/replica.cc
```diff
@@ -704,6 +704,18 @@ error_code Replica::ConsumeRedisStream() {
...
+    // If the acks-fb or something else triggered a shutdown, then do not attempt to read from the
+    // stream.
+    if (!exec_st_.IsRunning()) {
+      if (exec_st_.IsError()) {
+        LOG_REPL_ERROR("Stopping stream consumer in phase "
+                       << GetCurrentPhase()
+                       << " because of external error: " << exec_st_.GetError().Format());
+      }
+      acks_fb_.JoinIfNeeded();
+      return exec_st_.GetError();
```
abhijat, line 715: Is it possible we end up here without an error? I can't see any code path where that could happen but I don't think adding a `DCHECK` here is sensible

kostasrim (dismissed):

src/server/replica.cc
kostasrim: Yes, if we `Cancel` the `GenericError` should be empty?

abhijat: Ok then it doesn't make sense to add any error related asserts here.

kostasrim: I see. You were not talking about in general, if the error can be empty (which it can if it got cancelled via `Cancel()`). 

In `replica.cc` context -- you are right, we always cancel via `ReportCancelError` so there should be an error always.

In that case I suggest to: (see the code block)

src/server/replica.cc
```diff
@@ -704,6 +704,18 @@ error_code Replica::ConsumeRedisStream() {
...
 
+    // If the acks-fb or something else triggered a shutdown, then do not attempt to read from the
+    // stream.
+    if (!exec_st_.IsRunning()) {
+      if (exec_st_.IsError()) {
+        LOG_REPL_ERROR("Stopping stream consumer in phase "
+                       << GetCurrentPhase()
+                       << " because of external error: " << exec_st_.GetError().Format());
+      }
+      acks_fb_.JoinIfNeeded();
```
kostasrim, line 714:
```suggestion
    if (!exec_st_.IsRunning()) {
      DCHECK(exec_st_.IsError());
        LOG_REPL_ERROR("Stopping stream consumer in phase "
                       << GetCurrentPhase()
                       << " because of external error: " << exec_st_.GetError().Format());
      acks_fb_.JoinIfNeeded();
      return exec_st_.GetError();
```

src/server/replica.cc
abhijat: I see, added the suggestion

Issue #5949, opened 2025-10-23
Dragonfly crashes in dragonfly_connection (fuzzing input)


vyavdoshenko:
Crash log:
```
F20251023 13:27:07.970026 53848 dragonfly_connection.cc:1780] Check failed: RespExpr::STRING == arg.type ( vs. ) 
*** Check failure stack trace: ***
    @     0x555556a930fa  google::LogMessage::SendToLog()
    @     0x555556a8b17e  google::LogMessage::Flush()
    @     0x555556a8cb6f  google::LogMessageFatal::~LogMessageFatal()
    @     0x555556509b9d  facade::Connection::FromArgs()
    @     0x555556509f29  _ZN4absl12lts_2025051219functional_internal12InvokeObjectIZN6facade10Connection10ParseRedisEjEUlvE0_NS4_13MessageHandleEJEEET0_NS1_7VoidPtrEDpNS1_8ForwardTIT1_E4typeE
    @     0x55555650daae  facade::Connection::DispatchSingle()
    @     0x55555650e493  facade::Connection::ParseRedis()
    @     0x55555650f3bf  facade::Connection::ConnectionFlow()
    @     0x555556510ab6  facade::Connection::HandleRequests()
    @     0x55555659678e  util::ListenerInterface::RunSingleConnection()
    @     0x555556596d94  _ZN5boost7context6detail11fiber_entryINS1_12fiber_recordINS0_5fiberEN4util3fb219FixedStackAllocatorEZNS6_6detail15WorkerFiberImplIZZNS5_17ListenerInterface13RunAcceptLoopEvENKUlvE_clEvEUlvE_JEEC4IS7_EESt17basic_string_viewIcSt11char_traitsIcEENS6_13FiberPriorityERKNS0_12preallocatedEOT_OSC_EUlOS4_E_EEEEvNS1_10transfer_tE
    @     0x5555566b2d67  make_fcontext
*** SIGABRT received at time=1761215227 on cpu 0 ***
PC: @     0x7ffff7d39b2c  (unknown)  pthread_kill
[1]    53847 IOT instruction (core dumped)  ./dragonfly --port=6379 --logtostderr --proactor_threads=2 --bind=0.0.0.0
```

This is a follow-up https://github.com/dragonflydb/dragonfly/issues/5931
Tested after this issue had been fixed.

How to reproduce:
```
nc localhost 6379 < ./crash1.txt
```
or 
```
nc localhost 6379 < ./crash2.txt
```

[...]

PR #5955, opened 2025-10-23
fix: reject non-STRING arguments in RESP protocol to prevent crash

vyavdoshenko:
Fixes https://github.com/dragonflydb/dragonfly/issues/5949

Prevents server crash when receiving malformed RESP protocol with NIL bulk strings (`$-1`) as command arguments.

The server crashed with assertion failure when processing commands containing NIL bulk string arguments:
Check failed: RespExpr::STRING == arg.type ( vs. )

According to RESP2 protocol specification:
- Clients should send commands as "an array of bulk strings"
- NIL bulk strings (`$-1`) are valid for `server responses`, not command arguments
- While the parser can technically parse `$-1`, it's invalid in a command context

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1231,18 +1231,11 @@ Connection::ParserStatus Connection::ParseRedis(unsigned max_busy_cycles) {
     result = redis_parser_->Parse(read_buffer.slice, &consumed, &tmp_parse_args_);
     request_consumed_bytes_ += consumed;
     if (result == RedisParser::OK && !tmp_parse_args_.empty()) {
-      // Check if all arguments are STRING type. Commands and their arguments must be strings.
       // If we get a non-STRING type (e.g., NIL, ARRAY), it's a protocol error.
-      bool has_invalid_arg = false;
-      for (size_t i = 0; i < tmp_parse_args_.size(); ++i) {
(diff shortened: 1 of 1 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -1231,11 +1231,17 @@ Connection::ParserStatus Connection::ParseRedis(unsigned max_busy_cycles) {
...
     request_consumed_bytes_ += consumed;
     if (result == RedisParser::OK && !tmp_parse_args_.empty()) {
-      // Check if the first argument is a STRING type. Commands must be strings.
-      // If we get a non-STRING type (e.g., ARRAY from empty array *0), it's a protocol error.
-      if (tmp_parse_args_.front().type != RespExpr::STRING) {
-        LOG(WARNING) << "Invalid command - expected STRING type, got type: "
-                     << tmp_parse_args_.front().type;
+      // Check if all arguments are STRING type. Commands and their arguments must be strings.
+      // If we get a non-STRING type (e.g., NIL, ARRAY), it's a protocol error.
+      bool has_invalid_arg = false;
```
romange: nit:
```
bool valid_input = std::all_of(tmp_parse_args_.begin(), tmp_parse_args_.end(), [](const auto& item) ...

if (!valid_input)
   ....
```

vyavdoshenko: Fixed

romange (approved):

PR #5956, opened 2025-10-23
chore: add a task that periodically updates the expire base time

romange:
The task is run during cpu idle periods. It starts refreshing the base after several days. During the updates there are two bases and only one that is currently active: expire_gen_id_

1. The update process concludes when it reaches end of the traversal and it did not skip any values that should be updated.

2. It skips values when it encounters expired times as it can not delete values in the scope of the task.

### DIFF
--- a/src/core/expire_period.h
+++ b/src/core/expire_period.h
@@ -10,13 +10,13 @@ namespace dfly {
 
 class ExpirePeriod {
  public:
-  static constexpr size_t kMaxGenId = 15;
+  static constexpr std::size_t kMaxGenId = 15;
 
   ExpirePeriod() : val_(0), gen_(0), precision_(0) {
     static_assert(sizeof(ExpirePeriod) == 8);  // TODO
   }
 
-  explicit ExpirePeriod(uint64_t ms, unsigned gen = 0) : ExpirePeriod() {
+  explicit ExpirePeriod(uint64_t ms, unsigned gen = 0) : gen_(gen) {
     Set(ms);
   }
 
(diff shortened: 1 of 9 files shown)


src/core/expire_period.h
```diff
@@ -35,7 +35,9 @@ class ExpirePeriod {
 
   void Set(uint64_t ms);
 
-  bool is_second_precision() { return precision_ == 1;}
+  bool is_second_precision() {
+    return precision_ == 1;
+  }
```
kostasrim, line 40:
```suggestion
  bool is_second_precision() const {
    return precision_ == 1;
  }
```

src/server/engine_shard.cc
```diff
@@ -326,6 +328,97 @@ uint32_t EngineShard::DefragTask() {
...
+          ++update_expire_state_->db_index;
+        }
+      }
+    } while (update_expire_state_->cursor && ++iters < 100);
+  }
+
+  if (update_expire_state_->db_index >= db_slice.db_array_size()) {
+    if (update_expire_state_->stale_entries == 0) {
+      // We went over all the items and not stale items were found, we are done.
+      delete update_expire_state_;
```
kostasrim, line 410: I haven't followed the logic of this function completely but this seems dangerous if the state is kept among different idle runs of this function. Why ? Because:

1. IdleTsk runs -> allocates update_expire_state_
2. We exit so we don't run the task again
3. Memory leak since we don't call `delete update_expire_state_` anywhere else

romange: IdleTask exits only on process shutdown. if it concludes the traversal, it deletes the state.

PR #5970, opened 2025-10-27
fix(tiering): Limit offload budget by time

dranikpg:
1. Stopping by max iterations was a very simple heuristic that I put a long time ago - it obviously doesn't work. It decimates p99 for to put it mildly

2. I dream of making heartbeat use it's background possibilities. The issue with those yields is that our offloading speed is now much much slower... 🤔

### DIFF
--- a/src/server/tiered_storage.cc
+++ b/src/server/tiered_storage.cc
@@ -485,6 +485,7 @@ int64_t TieredStorage::UploadBudget() const {
 }
 
 void TieredStorage::RunOffloading(DbIndex dbid) {
+  using namespace tiering::literals;
   if (SliceSnapshot::IsSnaphotInProgress())
     return;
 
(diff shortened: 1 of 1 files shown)


src/server/tiered_storage.cc
```diff
@@ -485,15 +485,14 @@ int64_t TieredStorage::UploadBudget() const {
...
   if (SliceSnapshot::IsSnaphotInProgress())
     return;
 
+  const auto start_cycles = base::CycleClock::Now();
+
   // Don't run offloading if there's only very little space left
   auto disk_stats = op_manager_->GetStats().disk_stats;
-  if (disk_stats.allocated_bytes + kMaxIterations / 2 * tiering::kPageSize >
-      disk_stats.max_file_size)
+  if (disk_stats.allocated_bytes + 50 * tiering::kPageSize > disk_stats.max_file_size)
```
romange: what's a 50? add a constant with a meaningful name

Issue #5709, opened 2025-08-21
[AFL++] CI integration: PR smoke and periodic long runs


vyavdoshenko:
Scope
- GitHub Actions jobs:
  - PR “fuzz-smoke”: short, deterministic, per target (5–10 min total budget).
  - Nightly/weekly long runs with parallel AFL++.
- Cache corpora between runs; upload crashes, minimized reproducers, and logs as artifacts.
- Fail CI on first crash (for PR); keep artifacts for triage.

Tasks
- [x] Add `fuzz-smoke.yml` with a limited time budget and stable timeouts.
- [x] Add `fuzz-long.yml` (schedule) with caching and parallelization.
- [x] Artifact upload and retention policy; cache corpora across runs.

PR #5972, opened 2025-10-27
feat(ci): Add AFL++ fuzzing workflows for smoke and long tests

vyavdoshenko:
Adds AFL++ fuzzing integration with two GitHub Actions workflows:
- Smoke test (5 min) - quick validation, stops on first crash
- Long test (30 min) - extended campaign, collects all crashes

Implements: https://github.com/dragonflydb/dragonfly/issues/5709

Examples:
https://github.com/dragonflydb/dragonfly/actions/runs/18870612373
https://github.com/dragonflydb/dragonfly/actions/runs/18870612351

### DIFF
--- a/.github/workflows/fuzz-long.yml
+++ b/.github/workflows/fuzz-long.yml
@@ -1,7 +1,6 @@
 name: AFL++ Long Fuzzing Campaign
 
 on:
-  push:
   # TODO: Uncomment after testing manual runs
   # schedule:
   #   # Run nightly at 2 AM UTC
--- a/.github/workflows/fuzz-smoke.yml
+++ b/.github/workflows/fuzz-smoke.yml
@@ -1,7 +1,6 @@
 name: AFL++ Fuzzing Smoke Test
 
 on:
-  push:
   # TODO: Uncomment after testing manual runs
   # pull_request:
   #   branches: [main]


.github/workflows/fuzz-smoke.yml
```diff
@@ -0,0 +1,188 @@
...
+
+          # Archive crashes if they exist
+          if [ -d crashes ] && [ "$(ls -A crashes 2>/dev/null)" ]; then
+            tar -czf crashes.tar.gz crashes/
+          fi
+
+          # Archive hangs if they exist
+          if [ -d hangs ] && [ "$(ls -A hangs 2>/dev/null)" ]; then
+            tar -czf hangs.tar.gz hangs/
+          fi
```
kostasrim: Would something like tar `fuzz/artifacts/resp/default` recursively in one command work? That way we can avoid all the 3 explicit if conditions

.github/workflows/fuzz-smoke.yml
```diff
@@ -0,0 +1,188 @@
...
+
+          # Archive corpus/queue if it exists
+          if [ -d queue ] && [ "$(ls -A queue 2>/dev/null)" ]; then
+            tar -czf corpus.tar.gz queue/
+          fi
+
+      - name: Upload fuzzing artifacts
+        if: always()
```
kostasrim: Why do we run this unconditionally and not only on failures (when we exit with `1` at step check for crashes)

.github/workflows/fuzz-long.yml
```diff
@@ -0,0 +1,211 @@
...
+
+          # Explicitly set core_pattern (afl-system-config might not work in containers)
+          echo core > /proc/sys/kernel/core_pattern || echo "Warning: Could not set core_pattern"
+
+          echo "System configured"
+
+      - name: Build Dragonfly with AFL++
+        run: |
+          echo "Building Dragonfly with AFL++ instrumentation..."
+          ./helio/blaze.sh -DUSE_AFL:BOOL=ON
```
kostasrim: Is there a reason we prefer debug over release builds ? Would there be any value running both types ?

.github/workflows/fuzz-smoke.yml
vyavdoshenko: I can do it, but this folder has the internal state, and it can be big. I would like to add to the artifacts files that we really need.

.github/workflows/fuzz-long.yml
vyavdoshenko: Fuzzing build is a debug build. I don't know if it's possible and reasonable to fuzz a release build. I think this is a good question for the next steps. Now, I think we have to find all existing issues and fix them. The fuzzing process can be improved later to find new bugs.

.github/workflows/fuzz-smoke.yml
vyavdoshenko: Fixed

romange (approved):

PR #5973, opened 2025-10-28
fix(tiering): Basic stash backpressure

dranikpg:
Whenever we are beyond the offloading threshold and a stash operation is requested, we allocate a future and return it to the caller. The future is resolved once the value is stashed or the stash is cancelled. The caller (SET command) waits for the future to be completed in the coordinator, slowing the client down to around the speed of the disk.

Before we had no back-pressure whatsoever, so clients could issue SET commands at execution speed. Not all entries are even enqueued for offloading (remember write depth). Generally, it's difficult to pick an adaptive load without back-pressure. Selecting anything faster than the ingestion speed leads to an OOM 

* Provide back-pressure for stashes that are executed when offloading should be performed
* Optimise maps with EntryMap with heterogeneous lookup

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,6 @@ cmake-build-debug
 .venv/
 fuzz/artifacts/
 fuzz/corpus/
+
+# Valkey-search integration tests (synced from external repo)
+tests/dragonfly/valkey_search/integration/
--- a/src/core/task_queue.h
+++ b/src/core/task_queue.h
@@ -39,7 +39,7 @@ class TaskQueue {
     util::detail::ResultMover<ResultType> mover;
 
     ++blocked_submitters_;
-    Add([&mover, f = std::forward<F>(f), done](unsigned) mutable {
+    Add([&mover, f = std::forward<F>(f), done]() mutable {
       mover.Apply(f);
       done.Notify();
     });
--- a/src/server/cluster/outgoing_slot_migration.cc
+++ b/src/server/cluster/outgoing_slot_migration.cc
@@ -41,7 +41,7 @@ class OutgoingMigration::SliceSlotMigration : private ProtocolClient {
   }
 
   ~SliceSlotMigration() {
-    Cancel();
+    CloseSocket();
     exec_st_.JoinErrorHandler();
   }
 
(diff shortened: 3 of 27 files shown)


src/server/tiered_storage.h
```diff
@@ -105,6 +107,9 @@ class TieredStorage {
 
   PrimeTable::Cursor offloading_cursor_{};  // where RunOffloading left off
 
+  // Stash operations waiting for completion to throttle
+  tiering::EntryMap<::util::fb2::Future<bool>> stash_backpressure_;
```
dranikpg, line 111: Either a future or... a blocking counter? I don't see how we can easily use something more lightweight

romange:
Can you explain what this PR fixes/changes? what is the scenario? Please update the PR description and add "before and after" section.

src/server/tiered_storage.cc
```diff
@@ -375,15 +386,21 @@ bool TieredStorage::TryStash(DbIndex dbid, string_view key, PrimeValue* value) {
...
     LOG_IF(ERROR, ec != errc::file_too_large) << "Stash failed immediately" << ec.message();
     visit([this](auto id) { op_manager_->ClearIoPending(id); }, id);
-    return false;
+    return {};
   }
 
-  return true;
+  // If we are in the active offloading phase, throttle stashes by providing backpressure future
+  if (provide_bp && ShouldOffload())
+    return stash_backpressure_[{dbid, string{key}}];
```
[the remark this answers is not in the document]
dranikpg: No, I want a new value

dranikpg:
@romange updated

src/server/string_family.cc
```diff
@@ -1056,11 +1061,19 @@ void StringFamily::Set(CmdArgList args, const CommandContext& cmnd_cntx) {
...
+  sparams.backpressure = &backpressure;
+
   OpStatus result = SetGeneric(sparams, key, value, cmnd_cntx);
   if (result == OpStatus::WRONG_TYPE) {
     return builder->SendError(kWrongTypeErr);
   }
 
+  // If backpressure was provided, wait with reasonable limit (to avoid client deadlocking).
+  if (backpressure) {
+    std::move(backpressure)->GetFor(10ms);
```
romange: maybe 5ms to be less intrusive?

romange (approved):

PR #5984, opened 2025-10-29
feat(tiering): Two phase stash

dranikpg:
Two phase stashing divides the process into two parts:
1. Allocating space on the file with external alloc and preparing a buffer
2. Writing the buffer

This allows to _first_ determine if there is enough space and get a destination buffer. Only then we proceed with serializing the value (which will be much more costly for hashmaps!) directly into the destination buffer.

### DIFF
--- a/src/server/tiering/disk_storage.h
+++ b/src/server/tiering/disk_storage.h
@@ -42,7 +42,7 @@ class DiskStorage {
   // Mark segment as free, performed immediately
   void MarkAsFree(DiskSegment segment);
 
-  // Allocate segment of at least given length and prepare buffer. Migh block to grow backing.
+  // Allocate segment of at least given length and prepare buffer. Might block to grow backing file.
   // Return error if not enough space is available or growing failed.
   // Every successful preparation must end in a Stash(), otherwise resources are leaked.
   io::Result<std::pair<size_t /* offset */, util::fb2::UringBuf>> PrepareStash(size_t length);


src/server/tiering/disk_storage.h
```diff
@@ -42,11 +42,13 @@ class DiskStorage {
   // Mark segment as free, performed immediately
   void MarkAsFree(DiskSegment segment);
 
-  // Request bytes to be stored, cb will be called with assigned segment on completion. Can block to
-  // grow backing file. Returns error code if operation failed  immediately (most likely it failed
-  // to grow the backing file) or passes an empty segment if the final write operation failed.
-  // Bytes are copied and can be dropped before cb is resolved
-  std::error_code Stash(io::Bytes bytes, StashCb cb);
+  // Allocate segment of at least given length and prepare buffer. Migh block to grow backing.
```
romange:
```suggestion
  // Allocate segment of at least given length and prepare buffer. Might block to grow backing file.
```

src/server/tiered_storage.cc
```diff
@@ -380,12 +380,26 @@ std::optional<util::fb2::Future<bool>> TieredStorage::TryStash(DbIndex dbid, str
 
   tiering::OpManager::EntryId id;
   error_code ec;
+
+  // TODO(vlad): Replace with encoders for different types
+  auto stash_string = [&](std::string_view str) {
+    if (auto prepared = op_manager_->PrepareStash(str.size()); prepared) {
```
romange, line 386: PrepareStash can preempt, right? how can you know that `str` is still valid once it resumes?

dranikpg: Now it doesn't preempt

PR #5985, opened 2025-10-30
feat(search): Add support for multiple prefixes in search indexes

vyavdoshenko:
Adds support for multiple prefixes in search indexes, improving compatibility with valkey search.

- Changed `DocIndex::prefix` from `string` to `vector<string> prefixes`
- Updated `FT.CREATE` to accept multiple PREFIX arguments (e.g., `PREFIX 2 user: product:`)
- Updated `FT.INFO` to return `prefixes` array instead of a single `prefix` string
- Empty `prefixes` vector matches all keys (backward compatibility)

Integration (valkey-search) tests improved: +3 tests passing (22 passed vs 19 before)

### DIFF
--- a/src/server/search/doc_index.cc
+++ b/src/server/search/doc_index.cc
@@ -35,23 +35,10 @@ void TraverseAllMatching(const DocIndex& index, const OpArgs& op_args, F&& f) {
   string scratch;
   auto cb = [&](PrimeTable::iterator it) {
     const PrimeValue& pv = it->second;
-    if (pv.ObjType() != index.GetObjCode())
-      return;
-
     string_view key = it->first.GetSlice(&scratch);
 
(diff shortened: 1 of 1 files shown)


src/server/search/doc_index.cc
```diff
@@ -39,8 +39,19 @@ void TraverseAllMatching(const DocIndex& index, const OpArgs& op_args, F&& f) {
...
+      bool matches_prefix = false;
+      for (const auto& prefix : index.prefixes) {
+        if (key.rfind(prefix, 0) == 0) {
+          matches_prefix = true;
+          break;
+        }
+      }
+      if (!matches_prefix)
+        return;
+    }
```
dranikpg: by the way, I assume we can just use index.Matches() here

PR #5988, opened 2025-10-30
fix(search): persist FT.SYNUPDATE synonyms across server restarts

vyavdoshenko:
Synonyms created with `FT.SYNUPDATE` were lost after the Dragonfly restart, causing data loss.

Solution:
- Save synonym groups to RDB as `SYNUPDATE` commands alongside index definitions
- Load and dispatch `FT.SYNUPDATE` commands during RDB restore

### DIFF
--- a/src/core/search/base.h
+++ b/src/core/search/base.h
@@ -40,7 +40,7 @@ struct QueryParams {
 
 // Base class for optional search filters
 
-class AstNode;
+struct AstNode;
 
 struct OptionalFilterBase {
   virtual bool IsEmpty() const = 0;
--- a/src/facade/op_status.h
+++ b/src/facade/op_status.h
@@ -115,26 +115,6 @@ template <> class OpResult<void> : public OpResultBase {
   using OpResultBase::OpResultBase;
 };
 
-template <typename V> class OpResultTyped : public OpResult<V> {
- public:
-  OpResultTyped(V v) : OpResult<V>(std::move(v)) {
-  }
-
-  OpResultTyped(OpStatus st = OpStatus::OK) : OpResult<V>(st) {
-  }
-
-  void setType(int type) {
-    type_ = type;
-  }
-
-  int type() const {
-    return type_;
-  }
-
- private:
-  int type_ = -1;
-};
-
 inline bool operator==(OpStatus st, const OpResultBase& ob) {
   return ob.operator==(st);
 }
--- a/src/server/cluster/outgoing_slot_migration.cc
+++ b/src/server/cluster/outgoing_slot_migration.cc
@@ -41,7 +41,7 @@ class OutgoingMigration::SliceSlotMigration : private ProtocolClient {
   }
 
   ~SliceSlotMigration() {
-    Cancel();
+    CloseSocket();
     exec_st_.JoinErrorHandler();
   }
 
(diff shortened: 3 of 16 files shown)


src/server/rdb_load.cc
```diff
@@ -2825,11 +2825,28 @@ void RdbLoader::LoadSearchIndexDefFromAux(string&& def) {
     return;
   }
 
-  // Prepend FT.CREATE to index definiton
+  // Determine command type and prepend appropriate prefix
```
romange: what is the aux format for these definitions?
will it stay compatible with previous snapshots?
will it be extendable with more commands?

vyavdoshenko: Fixed.

Aux format: Two separate fields - `"search-index"` for FT.CREATE, `"search-synonyms"` for FT.SYNUPDATE
Backward compatible: Yes. Tested with the main branch.
Extensible: Yes. Can add `search-aliases`, `search-config`, etc., as separate aux fields using the same pattern.

src/server/rdb_save.cc
```diff
@@ -1367,9 +1367,27 @@ RdbSaver::GlobalData RdbSaver::GetGlobalData(const Service* service) {
...
+        // Save index definition
         search_indices.emplace_back(
             absl::StrCat(index_name, " ", index_info.BuildRestoreCommand()));
+
+        // Save synonym groups for this index in separate vector
+        const auto& synonym_groups = index->GetSynonyms().GetGroups();
+        for (const auto& [group_id, terms] : synonym_groups) {
+          if (!terms.empty()) {
+            // Convert set<string> to vector for joining
+            std::vector<std::string_view> terms_vec(terms.begin(), terms.end());
```
dranikpg: I'm almost sure that StrJoin works with as set as well?

src/server/rdb_load.cc
```diff
@@ -2825,11 +2827,28 @@ void RdbLoader::LoadSearchIndexDefFromAux(string&& def) {
...
   CmdArgVec arg_vec;
   facade::RespExpr::VecToArgList(resp_vec, &arg_vec);
-  string ft_create = "FT.CREATE";
-  arg_vec.insert(arg_vec.begin(), MutableSlice{ft_create.data(), ft_create.size()});
+
+  // Check if this is a SYNUPDATE command or a CREATE command
+  string command_name;
+  if (!arg_vec.empty()) {
+    string_view first_token = facade::ToSV(arg_vec[0]);
+    if (first_token == "SYNUPDATE") {
```
dranikpg: Why not just pass a flag to LoadSearchIndexDefFromAux instead of all the ifs?

src/server/rdb_save.cc
```diff
@@ -1367,9 +1367,27 @@ RdbSaver::GlobalData RdbSaver::GetGlobalData(const Service* service) {
...
+
+            // Format: "SYNUPDATE index_name group_id term1 term2 term3"
+            std::string syn_cmd = absl::StrCat("SYNUPDATE ", index_name, " ", group_id, " ",
+                                               absl::StrJoin(terms_vec, " "));
```
dranikpg: we also don't need the SYNUPDATE prefix if we have a special aux header for it

src/server/rdb_save.cc
vyavdoshenko: Fixed

src/server/rdb_load.cc
vyavdoshenko: Added a separate method.

src/server/rdb_save.cc
vyavdoshenko: Fixed

PR #5996, opened 2025-11-03
feat(hset): Load tiered values in read-only

dranikpg:
Add SerializedMap to HMapWrap

### DIFF
--- a/src/server/hset_family.cc
+++ b/src/server/hset_family.cc
@@ -178,7 +178,7 @@ template <typename T> OpResult<T> Unwrap(OpResult<CbVariant<T>> result) {
 }
 
 // Execute callback on generic HMapWrap, possibly on offloaded value and waiting for result
-template <typename F, typename T = std::invoke_result_t<F, HMapWrap>::Type>
+template <typename F, typename T = typename std::invoke_result_t<F, HMapWrap>::Type>
 OpResult<T> ExecuteRO(Transaction* tx, F&& f) {
   auto shard_cb = [f = std::forward<F>(f)](Transaction* t,
                                            EngineShard* es) -> OpResult<CbVariant<T>> {
--- a/src/server/tiering/decoders.cc
+++ b/src/server/tiering/decoders.cc
@@ -69,7 +69,11 @@ std::unique_ptr<Decoder> SerializedMapDecoder::Clone() const {
 
 void SerializedMapDecoder::Initialize(std::string_view slice) {
   map_ = std::make_unique<SerializedMap>(slice);
-  estimated_mem_usage = slice.size() + map_->size() * 8 /* bytes per ptr */;
+}
+
+Decoder::UploadMetrics SerializedMapDecoder::GetMetrics() const {
+  return UploadMetrics{.modified = false,
+                       .estimated_mem_usage = map_->DataBytes() + map_->size() * 2 * 8};
 }
 
 void SerializedMapDecoder::Upload(CompactObj* obj) {
--- a/src/server/tiering/decoders.h
+++ b/src/server/tiering/decoders.h
@@ -74,6 +74,7 @@ struct StringDecoder : public Decoder {
 struct SerializedMapDecoder : public Decoder {
   std::unique_ptr<Decoder> Clone() const override;
   void Initialize(std::string_view slice) override;
+  UploadMetrics GetMetrics() const override;
   void Upload(CompactObj* obj) override;
 
   SerializedMap* Get() const;
--- a/src/server/tiering/serialized_map.cc
+++ b/src/server/tiering/serialized_map.cc
@@ -48,6 +48,10 @@ size_t SerializedMap::size() const {
   return size_;
 }
 
+size_t SerializedMap::DataBytes() const {
+  return slice_.size() - 4 - size() * 2 * 4;
+}
+
 constexpr size_t kLenBytes = 4;
 
 size_t SerializedMap::SerializeSize(Input input) {
--- a/src/server/tiering/serialized_map.h
+++ b/src/server/tiering/serialized_map.h
@@ -46,6 +46,9 @@ struct SerializedMap {
   Iterator end() const;
   size_t size() const;
 
+  // Number of bytes of pure keys or values
+  size_t DataBytes() const;
+
   // Input for serialization
   using Input = const absl::Span<const std::pair<std::string_view, std::string_view>>;
 


src/server/hset_family.cc
```diff
@@ -139,21 +154,57 @@ auto KeyAndArgs(Transaction* t, EngineShard* es) {
...
     auto [key, op_args] = KeyAndArgs(t, es);
     auto it_res = op_args.GetDbSlice().FindReadOnly(op_args.db_cntx, key, OBJ_HASH);
     RETURN_ON_BAD_STATUS(it_res);
+    auto& pv = (*it_res)->second;
+
+    // Enqueue read for future values
+    if (pv.IsExternal() && !pv.IsCool()) {
+      using D = tiering::SerializedMapDecoder;
+      util::fb2::Future<OpResult<T>> fut;
+      auto cb = [fut, &f](io::Result<D*> res) mutable {
```
romange: are you sure `&f` is correct here? what happens when cb goes out of scope?

src/server/hset_family.cc
```diff
@@ -139,21 +154,57 @@ auto KeyAndArgs(Transaction* t, EngineShard* es) {
...
+  Overloaded ov{
+      [](T res) -> OpResult<T> { return res; },
+      [](util::fb2::Future<OpResult<T>> fut) -> OpResult<T> { return fut.Get(); },
+  };
+  return visit(ov, std::move(result).value());
+}
+
+template <typename F> auto ExecuteRO(Transaction* tx, F&& f) {
+  using T = typename std::invoke_result_t<F, HMapWrap>::Type;
+  auto cb = [f = std::forward<F>(f)](Transaction* t, EngineShard* es) -> OpResult<CbVariant<T>> {
```
romange: rename it to `shard_cb`

src/server/hset_family.cc
romange: rename if to `read_cb`

src/server/hset_family.cc
```diff
@@ -54,8 +57,16 @@ using container_utils::GetStringMap;
 
 struct HMapWrap {
  private:
-  template <typename F> decltype(auto) visit2(F f) const {  // Cast T* to T&
-    return std::visit(Overloaded{[&f](StringMap* s) { return f(*s); }, f}, impl_);
+  template <typename F> decltype(auto) VisitRef(F f) const {  // Cast T* to T&
```
romange, line 65: it becomes a bit complicated. Can you add a comment above `HMapWrap` explaining the purpose of this class?

src/server/tiered_storage.cc
```diff
@@ -317,11 +309,26 @@ error_code TieredStorage::Open(string_view base_path) {
...
+                                 const D& decoder, std::function<void(io::Result<D*>)> cb) {
+  DCHECK(value.IsExternal());
+  DCHECK(!value.IsCool());
+  // TODO: imporve performance by avoiding one more function wrap
+  op_manager_->Enqueue(KeyRef(dbid, key), value.GetExternalSlice(), decoder, std::move(cb));
+}
+
+template void TieredStorage::ReadInternal(
+    DbIndex, std::string_view, const PrimeValue&, const tiering::SerializedMapDecoder&,
+    std::function<void(io::Result<tiering::SerializedMapDecoder*>)>);
```
romange: would it be simpler if we would use virtual inheritance for this instead of templates?

src/server/tiered_storage.h
```diff
@@ -95,6 +102,10 @@ class TieredStorage {
   }
 
  private:
+  template <typename D>
+  void ReadInternal(DbIndex dbid, std::string_view key, const PrimeValue& value, const D& decoder,
```
romange: I understand that one part of your design creates a uniform interface to handle `variant<T, Future<T>>`  paths but here it's about generic de(serialization) which I do not fully understand.

src/server/hset_family.cc
dranikpg: true, didn't notice, it was just always stateless

src/server/tiered_storage.h
dranikpg: It's just about invoking a callback with a predefined decoder. The whole future stuff is just built on top of it

src/server/tiered_storage.cc
dranikpg: I have an idea how to make it simpler in the future. It's just a convenience, we can avoid templates alltogether

src/server/hset_family.cc
```diff
@@ -139,21 +160,58 @@ auto KeyAndArgs(Transaction* t, EngineShard* es) {
...
+    return result.status();
+
+  Overloaded ov{
+      [](T res) -> OpResult<T> { return res; },
+      [](util::fb2::Future<OpResult<T>> fut) -> OpResult<T> { return fut.Get(); },
+  };
+  return visit(ov, std::move(result).value());
+}
+
+template <typename F> auto ExecuteRO(Transaction* tx, F&& f) {
```
romange: why `ExecuteRO` btw? it returns io::Result<..>, no?

romange: btw, I understand it's hard to specify `io::Result` as a return type, maybe worth adding this in a comment at least.

src/server/tiered_storage.cc
```diff
@@ -347,8 +343,10 @@ TieredStorage::TResult<T> TieredStorage::Modify(DbIndex dbid, std::string_view k
   DCHECK(value.IsExternal());
 
   util::fb2::Future<io::Result<T>> future;
-  auto cb = [future, modf = std::move(modf)](io::Result<tiering::StringDecoder*> res) mutable {
-    future.Resolve(res.transform([&modf](auto* d) { return modf(d->Write()); }));
+  auto cb = [future, modf = std::move(modf)](io::Result<tiering::Decoder*> res) mutable {
+    future.Resolve(
+        res.transform([](auto* d) { return static_cast<tiering::StringDecoder*>(d); })  //
```
romange, line 361: missing comment? and in general, this is written as if it's only relevant for strings, correct?
just because hash maps do not have modify functionality?

dranikpg: Yes, added a comment. The hash interface is just more complicated in it's logic, that's why its separately. I'll remove those functions in the future and make them only local to string family

BorysTheDev (approved):

dranikpg:
Waiting for #6031

dranikpg:
Rebased on #6031

romange (approved):

PR #5998, opened 2025-11-03
feat(charts): Add Helm Chart mode for statefulset deployment without …

bennsel:
This change introduces a new Helm chart value mode, which allows users to deploy Dragonfly as a StatefulSet even when no persistent storage is defined.

Previously, the chart only created a StatefulSet when persistence (storage.enabled) was configured. With this update, users can explicitly choose the deployment type via:

```
mode: "deployment"    # default
# or
mode: "statefulset"
```

This provides greater flexibility for testing and ephemeral environments where a StatefulSet is desired but persistence is not required.

helm-docs has automatically updated the README with some new entries, including the new mode value.

The previous implementation remains unchanged.

### DIFF
--- a/contrib/charts/dragonfly/templates/deployment.yaml
+++ b/contrib/charts/dragonfly/templates/deployment.yaml
@@ -1,4 +1,4 @@
-{{- if not .Values.storage.enabled }}
+{{- if and (not .Values.storage.enabled) (eq .Values.mode "deployment") }}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
--- a/contrib/charts/dragonfly/templates/statefulset.yaml
+++ b/contrib/charts/dragonfly/templates/statefulset.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.storage.enabled }}
+{{- if or (.Values.storage.enabled) (eq .Values.mode "statefulset") }}
 apiVersion: apps/v1
 kind: StatefulSet
 metadata:
@@ -25,6 +25,7 @@ spec:
         {{- include "dragonfly.selectorLabels" . | nindent 8 }}
     spec:
       {{- include "dragonfly.pod" . | trim | nindent 6 }}
+  {{- if .Values.storage.enabled }}
   volumeClaimTemplates:
     - metadata:
         name: "{{ .Release.Name }}-data"
@@ -34,4 +35,5 @@ spec:
         resources:
           requests:
             storage: {{ .Values.storage.requests }}
+  {{- end }}
 {{- end }}
--- a/contrib/charts/dragonfly/values.yaml
+++ b/contrib/charts/dragonfly/values.yaml
@@ -5,6 +5,9 @@
 # -- Number of replicas to deploy
 replicaCount: 1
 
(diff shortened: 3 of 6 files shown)


contrib/charts/dragonfly/templates/statefulset.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.storage.enabled }}
+{{- if or (.Values.storage.enabled) (eq .Values.mode "statefulset") }}
```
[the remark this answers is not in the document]
bennsel: This change introduces a behavioral difference compared to the previous implementation.
Previously, setting storage.enabled=true automatically switched the deployment type from Deployment to StatefulSet.

With this update, users must explicitly set mode: "statefulset" to deploy Dragonfly as a StatefulSet — even if persistence is enabled.

While this change provides clearer and more predictable configuration behavior, it may be considered breaking, as existing users relying on the implicit switch will need to update their values configuration.

contrib/charts/dragonfly/values.yaml
```diff
@@ -5,6 +5,9 @@
 # -- Number of replicas to deploy
 replicaCount: 1
 
+# -- Valid values are "deployment" or "statefulset".
```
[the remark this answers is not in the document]
bennsel: I would prefer to add a values.schema.json; I’ll include it shortly. This should help ensure proper validation and make the configuration more robust.

contrib/charts/dragonfly/templates/deployment.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if not .Values.storage.enabled }}
+{{- if and (not .Values.storage.enabled) (eq .Values.mode "deployment") }}
```
Abhra303, line 1: Please fix backward compatibility. If .Values.mode is empty string or deployment then proceed with the deployment. So the default behaviour doesn't change (even if the mode is not present).

contrib/charts/dragonfly/templates/statefulset.yaml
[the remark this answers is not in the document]
Abhra303: if storage enabled and mode == "", then create statefulset. If mode is set explicitly, follow mode.

Abhra303 (changes requested):
See comments. Overall looks good.

contrib/charts/dragonfly/templates/deployment.yaml
bennsel: I set the mode in the values.yaml to “deployment” as the default. Shouldn’t that be sufficient as a fallback? I’ve actually prevented an empty mode value in the schema YAML.

Abhra303: let's make the old yaml configuration compatible with this change. Its ok to have an empty mode so that if someone wants to revert their configuration, it doesn't change the behaviour at all.

Abhra303: Then remove this empty mode after a month or so.

vyavdoshenko:
Stale PR

Issue #5925, opened 2025-10-17
Add HEXPIRE flags  [NX | XX | GT | LT] and check compatibility


moyifeng2000:
Redis allows setting a TTL for a field. Can this requirement be added?

abhijat:
Are you looking for the [ttl](https://www.dragonflydb.io/docs/command-reference/generic/ttl) command?

moyifeng2000:
redis:
Task<bool> KeyExpireAsync(RedisKey key, TimeSpan? expiry, ExpireWhen when = ExpireWhen.Always, CommandFlags flags = CommandFlags.None);    key  ttl  valid
Task<ExpireResult[]> HashFieldExpireAsync(RedisKey key, RedisValue[] hashFields, TimeSpan expiry, ExpireWhen when = ExpireWhen.Always, CommandFlags flags = CommandFlags.None);   hash field ttl  valid

dragonfly:
Task<bool> KeyExpireAsync(RedisKey key, TimeSpan? expiry, ExpireWhen when = ExpireWhen.Always, CommandFlags flags = CommandFlags.None);    key  ttl  valid
Task<ExpireResult[]> HashFieldExpireAsync(RedisKey key, RedisValue[] hashFields, TimeSpan expiry, ExpireWhen when = ExpireWhen.Always, CommandFlags flags = CommandFlags.None);   hash field ttl  invalid

moyifeng2000:
<img width="1209" height="582" alt="Image" src="https://github.com/user-attachments/assets/f0512259-aa3b-4126-88e1-52cbc9f025e5" />

BorysTheDev:
Hi @moyifeng2000 if you are looking for hash field ttl please use https://www.dragonflydb.io/docs/command-reference/hashes/hsetex
I don't know what code you have provided, and what on this screenshot. Without a command that you want to see or at least a clear way how to reproduce your case, we can't help you.

moyifeng2000:
redis-cli
HSET test field1 test1 field2 test2
EXPIRE test 1000      // set key ttl
HEXPIRE test 300 FIELDS 2 field1 field2    // set hash field ttl

BorysTheDev:
@moyifeng2000 thx for the details. Now I see that we support the HEXPIRE command, but don't have some flags [NX | XX | GT | LT] 
We will add them

moyifeng2000:
Thank you!

PR #6001, opened 2025-11-03
feat: update HEXPIRE flags and behavior

BorysTheDev:
fixes: #5925
Improve parsing, 0 ttl remove field, added XX, NX, LT, GT options

### DIFF
--- a/src/server/hset_family.cc
+++ b/src/server/hset_family.cc
@@ -1012,8 +1012,13 @@ int32_t HSetFamily::FieldExpireTime(const DbContext& db_context, const PrimeValu
   }
 }
 
-static std::vector<long> ExpireElements(facade::CmdArgList values, uint32_t ttl_sec,
-                                        ExpireFlags flags, StringMap* owner) {
+// returns vector of results for each field in values:
+// -2 if the provided key does not exist.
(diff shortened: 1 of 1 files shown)


vyavdoshenko (approved):
lgtm

src/server/hset_family.cc
```diff
@@ -1012,8 +1012,61 @@ int32_t HSetFamily::FieldExpireTime(const DbContext& db_context, const PrimeValu
   }
 }
 
+static std::vector<long> ExpireElements(StringMap* owner, facade::CmdArgList values,
```
romange: mutable argument should be the last one - @BorysTheDev .

@vyavdoshenko  do you understand what this function does?

vyavdoshenko: It sets time to live on hash fields with new flag support, returning status codes for each field: 1 if TTL was set, 2 if field was deleted (when ttl=0), 0 if the flag condition wasn't met, or -2 if the field doesn't exist.
The main change is adding new flags for complience.

romange: @BorysTheDev  I think a comment describing what the function does would help here.

romange:
lgtm

PR #6007, opened 2025-11-04
fix(tiering): Grow backing async

dranikpg:
Must preceed #5984 

* TryGrow() was changed to RequestGrow() that calls `LinuxFile::FallocateAsync`. 
* Not having enough space requests the async grow and fails the stash operation. 
* Can lead to small inefficiency with, for example, background offloading scheduling many operations that end with nothing.
* We shill have the "farsighted" grow that requests it when we have less than 15% capacity left

### DIFF
--- a/src/server/tiering/disk_storage.cc
+++ b/src/server/tiering/disk_storage.cc
@@ -194,7 +194,8 @@ DiskStorage::Stats DiskStorage::GetStats() const {
 }
 
 error_code DiskStorage::RequestGrow(off_t grow_size) {
-  if (alloc_.capacity() + ExternalAllocator::kExtAlignment >= static_cast<size_t>(max_size_))
+  DCHECK_EQ(grow_size % ExternalAllocator::kExtAlignment, 0u);
+  if (alloc_.capacity() + grow_size >= static_cast<size_t>(max_size_))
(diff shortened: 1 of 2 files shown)


src/server/tiering/disk_storage.cc
```diff
@@ -221,14 +209,16 @@ error_code DiskStorage::TryGrow(off_t grow_size) {
...
-  grow_.ev.notifyAll();
-
-  if (!grow_.last_err)
-    alloc_.AddStorage(end, grow_size);
-  return grow_.last_err;
+  backing_file_->FallocateAsync(0, end, grow_size, [=](int res) {
+    auto ec = (res < 0) ? std::error_code{-res, std::system_category()} : std::error_code{};
+    grow_.pending = false;
+    grow_.last_err = ec;
+    grow_.timestamp_ns = ProactorBase::GetMonotonicTimeNs();
```
romange, line 216: timestamp_ns is not used and can be removed

romange (approved):

src/server/tiering/disk_storage.cc
dranikpg: It's used above in not retrying too frequently

Issue #5952, opened 2025-10-23
Timeouts caused by unread server replies for clients with large pipelines


kostasrim:
See `test_send_timeout`. When a client dispatches large pipelines and `does not read the server replies` the connection times out. 

We should address this problem by offloading replies back to a file when the socket is almost full. This should unblock the client and allow it to make progress

romange:
The problem is that it's not possible to recognize when using synchronous interface because before you call `Send( )`
you can not know whether it will get stuck or not. With `SendAsync` it is possible but this is not how we designed our ReplyBuilder.

kostasrim:
@romange I do understand the general direction and the root cause but I am still not sure about a specific implementation detail.

Regardless of how we dispatch on the connection level, eventually a write will get stuck `in-flight`. What that means is, is that the submitted `sq entry` will not complete and reach the completion queue. In other words, the `Write()` call will get fiber blocked in `FiberCall::Get()` method.

One way to solve this is as you wrote, replace blocking writes with `asynchronous` one. I am not confident about this as it implies a core logic change in how we write to the socket within the ReplyBuilder.

So what I was really looking here is:

a) For a write `W` of `B` bytes, is there a smart way to know `before I submit/prepare my write to the sq` if `it would block` and stay in flight ? In other words, is there a way to monitor the send buffers in such a way that we know beforehand that we are in this situation ? Based on what I read online this can't really be done. We have no control of that. What there is, `IORING_RECVSEND_POLL_FIRST` which is a `pesimist heuristic` -- it assumes the socket is not readable/writable but again this does not solve this because if a socket is writtable it does not mean it would overflow/fill the send buffers.

b) Maybe we could have a similar approach as we do with breaking the stalled connections ? We can `cancel` the in flight `write`. The only problem here is that `iouring` does not guarantee atomicity of that write (I could be wrong correct me if that's not the case) in case of cancel (it could be that we are in some partial state)

In short the idea is to avoid an async interface on the reply builder to keep the scope of changes small and contained. If we figure out early (before we fiber block) or cancel the in-flight, then the rest is handling backpressure in some disk backed up fd :)

romange:
I think the misconception is that I asked to accurately recognise a blocked socket and do the offloading after the fact. 

This is not what I am suggesting at all.  I suggest having a static condition, say - when the queue grows beyond K items, or its size above N bytes, do offloading of the subsequent items. 

[...]

PR #6011, opened 2025-11-04
chore: disk backed backpressure for connections

kostasrim:
Part 1 of #5952

Implements basic disk backpressure for connections that dispatch async. When `Connection::dispatch_q_` surpasses an arbitrary (defined in a flag) value, the connection will start offloading its pipeline messages to disk instead of blocking. The connection fiber is a `producer` because it writes data to disk when it can't push to the dispatch queue and the async fiber is the `consumer`: it drains backpressure from the queue and the rest (if any) from disk.

* add basic disk backed backpressure for connections
* add simple test

TODO and will follow up on a separate PR:

* Truncate backing file such that its total size is capped (requires small helio changes but it should be quick)
* GC files when connection is closed, aborted etc.
* Add tests for error paths and that the connection properly restarts reading from the queue when backpressure is drained.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -13,6 +13,7 @@
 #include <numeric>
 #include <variant>
 
+#include "absl/strings/str_split.h"
 #include "base/cycle_clock.h"
 #include "base/flag_utils.h"
 #include "base/flags.h"
@@ -27,7 +28,6 @@
 #include "facade/redis_parser.h"
 #include "facade/service_interface.h"
 #include "facade/socket_utils.h"
-#include "io/file.h"
 #include "util/fibers/fibers.h"
 #include "util/fibers/proactor_base.h"
 
@@ -111,6 +111,16 @@ ABSL_FLAG(uint32_t, pipeline_squash_limit, 1 << 30, "Limit on the size of a squa
 ABSL_FLAG(uint32_t, pipeline_wait_batch_usec, 0,
           "If non-zero, waits for this time for more I/O "
           " events to come for the connection in case there is only one command in the pipeline. ");
+
+ABSL_FLAG(size_t, connection_disk_backpressure_watermark, 0,
+          "Offload dispach queue backpressure to disk when it crosses watermark. (0 to disable)");
+
+ABSL_FLAG(size_t, connection_disk_backpressure_file_max_bytes, 50_MB,
+          "Maximum size of the backing file. When the watermark is reached, connection will "
+          "stop offloading backpressure to disk");
+
+ABSL_FLAG(size_t, connection_disk_backpressure_load_size, 30,
+          "How many queue backpressure items to load from disk when dispatch queue is drained");
 
 using namespace util;
 using namespace std;
@@ -434,6 +444,10 @@ size_t Connection::PipelineMessage::StorageCapacity() const {
   return storage.capacity() + args.capacity();
 }
 
+size_t Connection::PipelineMessage::StorageBytes() const {
+  return storage.size();
+}
+
 size_t Connection::MessageHandle::UsedMemory() const {
   struct MessageSize {
     size_t operator()(const PubMessagePtr& msg) {
@@ -676,6 +690,11 @@ Connection::Connection(Protocol protocol, util::HttpListenerBase* http_listener,
 #endif
 
   UpdateLibNameVerMap(lib_name_, lib_ver_, +1);
+
+  const size_t disk_watermark = absl::GetFlag(FLAGS_connection_disk_backpressure_watermark);
+  if (disk_watermark) {
+    backing_queue_ = std::make_unique<DiskBackedBackpressureQueue>();
+  }
 }
 
 Connection::~Connection() {
@@ -1162,8 +1181,12 @@ void Connection::ConnectionFlow() {
 
 void Connection::DispatchSingle(bool has_more, absl::FunctionRef<void()> invoke_cb,
                                 absl::FunctionRef<MessageHandle()> cmd_msg_cb) {
-  bool optimize_for_async = has_more;
   QueueBackpressure& qbp = GetQueueBackpressure();
+  if (OffloadBackpressureToDiskIfNeeded(cmd_msg_cb)) {
+    return;
+  }
+
+  bool optimize_for_async = has_more;
   if (optimize_for_async &&
       qbp.IsPipelineBufferOverLimit(stats_->dispatch_queue_bytes, dispatch_q_.size())) {
     stats_->pipeline_throttle_count++;
@@ -1683,11 +1706,22 @@ void Connection::AsyncFiber() {
   QueueBackpressure& qbp = GetQueueBackpressure();
   while (!reply_builder_->GetError()) {
     DCHECK_EQ(socket()->proactor(), ProactorBase::me());
+
+    LoadBackpressureFromDiskIfNeeded();
+
     cnd_.wait(noop_lk, [this] {
-      return cc_->conn_closing || (!dispatch_q_.empty() && !cc_->sync_dispatch);
+      return cc_->conn_closing || (!dispatch_q_.empty() && !cc_->sync_dispatch) ||
+             disk_backpressure_available_;
     });
+
     if (cc_->conn_closing)
       break;
+
+    if (disk_backpressure_available_) {
+      LoadBackpressureFromDiskIfNeeded();
+      DCHECK(dispatch_q_.size() > 0);
+      disk_backpressure_available_ = false;
+    }
 
     // We really want to have batching in the builder if possible. This is especially
     // critical in situations where Nagle's algorithm can introduce unwanted high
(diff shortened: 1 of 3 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -13,6 +13,9 @@
 #include <numeric>
 #include <variant>
 
+#include "absl/flags/internal/flag.h"
+#include "absl/functional/function_ref.h"
+#include "absl/strings/str_split.h"
```
kostasrim: remove

src/facade/dragonfly_connection.cc
```diff
@@ -2254,4 +2288,173 @@ void ResetStats() {
...
+      watermark_(absl::GetFlag(FLAGS_connection_disk_backpressure_watermark)) {
+  id_ = ++unique_id;
+}
+
+std::error_code Connection::DiskBackedBackpressureQueue::Init() {
+  if (init_) {
+    return {};
+  }
+
+  std::string backing_name = absl::StrCat("/tmp/backing_", id_);
```
kostasrim, line 2361: Shall we introduce a flag for those backing files or `/tmp/` is enough?

src/facade/dragonfly_connection.cc
```diff
@@ -112,6 +112,16 @@ ABSL_FLAG(uint32_t, pipeline_wait_batch_usec, 0,
...
           " events to come for the connection in case there is only one command in the pipeline. ");
 
+ABSL_FLAG(size_t, connection_disk_backpressure_watermark, 0,
+          "Offload dispach queue backpressure to disk when it crosses watermark. (0 to disable)");
+
+ABSL_FLAG(size_t, connection_disk_backpressure_file_max_bytes, 50_MB,
+          "Maximum size of the backing file. When the watermark is reached, connection will "
+          "stop offloading backpressure to disk");
+
+ABSL_FLAG(size_t, connection_disk_backpressure_load_size, 30,
```
kostasrim, line 122: 30 is low IMO, any preference for default value ?

src/facade/dragonfly_connection.h
```diff
@@ -504,6 +519,68 @@ class Connection : public util::Connection {
   };
 
   bool request_shutdown_ = false;
+
+  class DiskBackedBackpressureQueue {
```
romange, line 523: the pr is large, lets do the usual: 
split it in self contained chunks. specifically, this should reside in dedicated files.

src/facade/dragonfly_connection.cc
```diff
@@ -112,6 +112,16 @@ ABSL_FLAG(uint32_t, pipeline_wait_batch_usec, 0,
           "If non-zero, waits for this time for more I/O "
           " events to come for the connection in case there is only one command in the pipeline. ");
 
+ABSL_FLAG(size_t, connection_disk_backpressure_watermark, 0,
```
romange, line 115: what is the unit here? what crosses the watermark?

src/facade/dragonfly_connection.cc
```diff
@@ -2254,4 +2288,173 @@ void ResetStats() {
     io_req_size_hist->Clear();
 }
 
+bool Connection::OffloadBackpressureToDiskIfNeeded(absl::FunctionRef<MessageHandle()> handle) {
+  // Offload only when dispatch_q_ crosses watermark or when backing queue already
+  // has pending items.
+  if (backing_queue_ &&
+      ((dispatch_q_.size() > backing_queue_->Watermark()) || !backing_queue_->Empty())) {
+    auto ec = backing_queue_->Init();
```
romange, line 2296: have not looked at Init, but i prefer to have explicit initalization of resources at the beginning of the flow (where you create backing_queue_) rather than doing it lazily

src/facade/dragonfly_connection.cc
```diff
@@ -2254,4 +2288,173 @@ void ResetStats() {
     io_req_size_hist->Clear();
 }
 
+bool Connection::OffloadBackpressureToDiskIfNeeded(absl::FunctionRef<MessageHandle()> handle) {
+  // Offload only when dispatch_q_ crosses watermark or when backing queue already
+  // has pending items.
+  if (backing_queue_ &&
+      ((dispatch_q_.size() > backing_queue_->Watermark()) || !backing_queue_->Empty())) {
```
romange, line 2295: i prefer that `DiskOffloadThreshold()` would be in QueueBackpressure rather than in `backing_queue_` , this way you do not need to touch backing_queue_ object during the happy path.

src/facade/dragonfly_connection.cc
```diff
@@ -2254,4 +2288,173 @@ void ResetStats() {
...
+      if (ptr = GetFromPipelinePool(); ptr) {
+        ptr->storage.resize(bytes.size());
+      } else {
+        ptr = make_unique<PipelineMessage>(1, 1);
+        ptr->storage.resize(bytes.size());
+      }
+
+      memcpy(ptr->storage.begin(), bytes.begin(), bytes.size());
+      std::string_view read{reinterpret_cast<char*>(ptr->storage.data()), bytes.size()};
+      ptr->args = absl::StrSplit(read, '\0', absl::SkipEmpty());
```
romange, line 2338: what does it mean? we do not support binary strings ?

src/facade/dragonfly_connection.cc
kostasrim: `dispatch_q_ > watermark` 

Will edit the comment/description accordingly

src/facade/dragonfly_connection.cc
```diff
@@ -2254,4 +2288,173 @@ void ResetStats() {
...
+  }
+}
+
+size_t Connection::DiskBackedBackpressureQueue::unique_id = 0;
+
+Connection::DiskBackedBackpressureQueue::DiskBackedBackpressureQueue()
+    : max_backing_size_(absl::GetFlag(FLAGS_connection_disk_backpressure_file_max_bytes)),
+      max_queue_load_size_(absl::GetFlag(FLAGS_connection_disk_backpressure_load_size)),
+      watermark_(absl::GetFlag(FLAGS_connection_disk_backpressure_watermark)) {
+  id_ = ++unique_id;
```
romange, line 2353: this is buggy as `DiskBackedBackpressureQueue` is initialized in multiple threads.

src/facade/dragonfly_connection.cc
romange: we should provide a configurable prefix, `/tmp/` is not enough.

src/facade/dragonfly_connection.cc
```diff
@@ -2254,4 +2288,173 @@ void ResetStats() {
...
+  total_backing_bytes_ += msg->FullCommand().size();
+  offsets_.push_back(item);
+  next_offset_ += item.total_bytes;
+
+  VLOG(2) << "Offload connection " << this << " backpressure of " << item.total_bytes
+          << " bytes to disk at offset: " << item.offset;
+  VLOG(3) << "Command offloaded: " << msg->FullCommand();
+}
+
+template <typename F> void Connection::DiskBackedBackpressureQueue::LoadFromDiskToQueue(F f) {
```
romange, line 2426: template is not justified here. please use std::function instead.

src/facade/dragonfly_connection.cc
kostasrim: It's serialized and deserialized as is: when we parse the command from the socket we create a `PipelineMsg`. Within that, there is:

1. a flat storage which is the whole command + args `separated` by `\0`
2. a `view` (span) that can iterate over each argument (cmd + args) 

(see FromArgs{} function)

When we offload this, we offload the `flat blob` storage as is to disk. Loading is easy as reading the bytes directly back to a `PipelinePtr`. the `StrSplit` is a `fancy way` to build the (2) view of the flat storage

kostasrim: This is 1-1 with `FromArgs` and since the object layout was already flat -- it was easy to write it to disk as is

src/facade/dragonfly_connection.cc
kostasrim: yeah it's a data race indeed. I wanted a unique identifier. will fix

src/facade/dragonfly_connection.cc
kostasrim: I am happy to use a function here.

> template is not justified here. 

For completeness, this is not how I think about it. std::function `allocates dynamically` and a deduced lambda argument from a function template parameter does not. Using the most generic form (std::function) in a non generic context (a plain function) is the overkill here.

src/facade/dragonfly_connection.cc
romange: yeah, writing ito do disk is easy but I think your conclusion is wrong, and you can not load binary strings that contain `\0` inside using split(). you will need to serialize argument sizes to make it correct.  

In fact, I would like to challenge your design choices:
1. the pipeline is offloaded after the parsing.  
2.  this means you need to serialize structured data, we already saw you keep offsets to keep track of the blob sizes, for example inside your disk queue.
3. but you also need to parse back these argument sets. it require more complicated serialisation to do it properly.

Instead you could offload raw socket data before parsing:
1. you save CPU on parsing data that you later offload
2. you do not need to store blob sizes, so your on-disk queue is simpler
3. there is no on-disk format at all - you do not need to worry about serializing command arguments properly.

 So your file becomes a full substitute for socket until it depletes.

kostasrim: > yeah, writing ito do disk is easy but I think your conclusion is wrong

I am not sure I understand why `\0` is a problem here. For example a command  `set foo bar` is parsed and stored as `set\0foo\0bar\0` both in memory and on disk. Using split() after loading this to build a view of the command arguments (e,g {set, foo, bar} is perfectly valid in that context. Am I missing something ?

> Instead you could offload raw socket data before parsing:

My concern for this would be that parsing should now also be a part of AsyncFb. Furthermore, parsing a raw socket blob that was written to disk needs to handle cases like `INPUT_PENDING`, which introduces additional preemption points to the async fiber outside of a) blocking on queue backpressure b) blocking on loading the next batch to execute from disk. For example:

async_fb-> blocks on send because client is not reading, dispatch queue grows
conn_fb -> reads from socket starts offloading disk queue
client -> stops sending commands
conn_fb-> blocks on `Connection::HandleRecvSocket`
async_fb->drains the dispatch_queue 

We need to load from disk the blob, parse it and place it in the dispatch_q. `conn_fiber` is blocked on `socket io`. There are two options:

1. coordinate and somehow unblock `HandleRecvSocket` such that ConnFiber can do a disk read + parsing, construct the msg and send it to the dispatch_q.
2. read from disk and do the parsing inside `AsyncFiber`.

If we choose (2) and since `ConnFb` offloads unparsed data, now `AsyncFb` can parse half entries (because of INPUT_PENDING). So if the last entry written to disk is half (because the other one will arrive soon in the socket -- in other words we just drained the disk backpressure):

1. AsyncFb needs to read more to finish parsing -> preempts
2. ConnFb reads from socket -> writes it to backing.

So now `ConnFb` must notify the `AsyncFb` to continue parsing. 

This is something that can be handled but nevertheless it seems that now parsing responsibilities of the ConnFb leak to the AsyncFb. What is more, now the complicated parsing stages (like INPUT_PENDING) must also be handled by AsyncFb.


> you save CPU on parsing data that you later offload

I am slightly confused over this as well. Didn't you suggest not to use the uring api at all which implies a full proactor block on that thread on io ? Also, assuming that the client will eventually read everything, won't we parse the data anyway once? What we offload is what we will eventually load and this is already parsed(apart of the split which just builds a view). So we don't actually save anything here -- we just delegate the parsing for a later time. What I mean here:

parse -> write to disk size(data) + parsed -> parse 2 -> write to disk size(data 2) 
parse 2 times, writes 2 times
vs
write to disk 1 -> write to disk 2 -> parse 1 -> parse2 
parse 2 times, writes 2 times

All in all, I am not against your solution I just want to understand it a little bit more because the tradeoff here seems to be:

prefix the parsed command with its size (and now we don't need extra memory) before writting to disk -> self contained change within `DiskBackpressure`

vs

write blobs -> requires parsing within AsyncFb

romange: @kostasrim with `set foo bar` there is no problem. the problem is with binary strings that you send with RESP protocol like `*3$3\r\nSET\r\n$3\r\nf\0\0\r\n$3\r\nb\0\0\r\n`  or in other words where we send `set f\0\0 b\0\0`. 
And it's trivially to test and reproduce the problem.

regarding your second point - no, AsyncFb should not handle parsing, ConnFb should do it. And I agree that `HandleRecvSocket` is problematic, but lets first understand what SHOULD be done and then discuss how we do it.

kostasrim: Oh interesting so, `set f\0\0` is a valid command. I now understand what you were talking about.

Ok let's sync internally.

Issue #5562, opened 2025-07-24
DflyEngineTest.ZeroAllocationEviction failed (dragonfly_test.cc)


kostasrim:
```
2025-07-24T14:29:08.8852729Z 57: /__w/dragonfly/dragonfly/src/server/dragonfly_test.cc:595: Failure
2025-07-24T14:29:08.8853575Z 57: Expected: (evicted_count) > (0), actual: 0 vs 0
2025-07-24T14:29:08.8854404Z 57: Zero-allocation entries should be evicted under memory pressure
```

kostasrim:
Failed again here: https://github.com/dragonflydb/dragonfly/actions/runs/18597689225/job/53028102402?pr=5927

```
2025-10-17T16:12:10.7023848Z 60: I20251017 16:12:10.702142 10304 proactor_pool.cc:149] Running 4 io threads
2025-10-17T16:12:10.7265723Z 60: I20251017 16:12:10.726048 10304 test_utils.cc:263] Starting ZeroAllocationEviction
2025-10-17T16:12:11.4092148Z 60: /__w/dragonfly/dragonfly/src/server/dragonfly_test.cc:606: Failure
2025-10-17T16:12:11.4094791Z 60: Expected: (evicted_count) > (0), actual: 0 vs 0
2025-10-17T16:12:11.4097599Z 60: Zero-allocation entries should be evicted under memory pressure
2025-10-17T16:12:11.4098222Z 60: 
2025-10-17T16:12:11.4316825Z 60: I20251017 16:12:11.431185 10304 test_utils.cc:225] Finishing ZeroAllocationEviction
2025-10-17T16:12:11.4329357Z 60: [  FAILED  ] DflyEngineTest.ZeroAllocationEviction (732 ms)
```

kostasrim:
https://github.com/dragonflydb/dragonfly/actions/runs/18972318549/job/54182875526?pr=5969

dranikpg:
https://github.com/dragonflydb/dragonfly/actions/runs/19010896956/attempts/3

abhijat:
[...]

PR #6012, opened 2025-11-05
tests: Fix order of used memory counter reset

abhijat:
Used memory counter was previously reset after starting service. This opened up the possibility that a few cycles of engine shard memory accounting in `EngineShard::CacheStats` already ran, which incremented used memory, and then this value was reset. 

This resulted in lost bytes and tests which rely on this value, for example to measure eviction, could fail. Here the order is switched so that used memory is reset first and then service is started.

fixes https://github.com/dragonflydb/dragonfly/issues/5562

### DIFF
--- a/src/server/test_utils.cc
+++ b/src/server/test_utils.cc
@@ -249,8 +249,11 @@ void BaseFamilyTest::ResetService() {
   pp_->Run();
   service_ = std::make_unique<Service>(pp_.get());
 
-  service_->Init(nullptr, {});
+  // Must be reset before starting the service. Engine shard heartbeat task updates this
+  // value, and if reset after some invocations of heartbeat have run, the accumulated data is
+  // lost and can cause test failure.
   used_mem_current = 0;
+  service_->Init(nullptr, {});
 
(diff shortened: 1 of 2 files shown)


abhijat:
~~will wait for~~

- [x] https://github.com/dragonflydb/dragonfly/actions/runs/19093866406/job/54549581142
- [x] https://github.com/dragonflydb/dragonfly/actions/runs/19093830312/job/54549478521 


~~to confirm if repeat runs of the test do not fail~~

these did not fail

src/server/test_utils.cc
```diff
@@ -249,8 +249,8 @@ void BaseFamilyTest::ResetService() {
   pp_->Run();
   service_ = std::make_unique<Service>(pp_.get());
 
-  service_->Init(nullptr, {});
   used_mem_current = 0;
+  service_->Init(nullptr, {});
```
romange, line 256: add a comment that the order is important and otherwise  used_mem_current may change before test starts running.

romange (dismissed):
lgtm

src/server/test_utils.cc
abhijat: sure added

PR #6015, opened 2025-11-05
feat(tiering): Serialize hashes

dranikpg:
Ok, so this changed a lot. Add active path for serializing hashes:
* Generic `EstimateSerializedSize`
* Generic `Serialize`
* Propagating ExternalRep in CompactObj, including cooling
* Handle upload

### DIFF
--- a/src/server/tiered_storage.cc
+++ b/src/server/tiered_storage.cc
@@ -477,13 +477,14 @@ std::optional<util::fb2::Future<bool>> TieredStorage::TryStash(DbIndex dbid, str
   } else if (auto bin = bins_->Stash(dbid, key, SerializeToString(*value)); bin) {
     id = bin->first;
     // TODO(vlad): Write bin to prepared buffer instead of allocating one
-    if (auto prepared = op_manager_->PrepareStash(est_size); prepared) {
+    if (auto prepared = op_manager_->PrepareStash(bin->second.length()); prepared) {
       auto [offset, buf] = *prepared;
       memcpy(buf.bytes.data(), bin->second.data(), bin->second.size());
       tiering::DiskSegment segment{offset, bin->second.size()};
       op_manager_->Stash(id, segment, buf);
     } else {
       ec = prepared.error();
+      bins_->ReportStashAborted(bin->first);
     }
   } else {
     return {};  // silently added to bin


src/core/compact_object.h
```diff
@@ -376,6 +377,10 @@ class CompactObj {
   // Returns the external data of the object incuding its ColdRecord.
   CoolItem GetCool() const;
 
+  // Prequisite: IsCool() is true.
+  // Keeps cool record only as external value and discard in-memory part.
+  void KeepExternal(size_t offset, size_t sz);
```
romange: `KeepExternal` is not clear imho.

`Freeze` - a more creative option, and maybe `DropCool` is more direct explaining what it does?

romange (dismissed):

src/core/compact_object.h
dranikpg: Ok.. The names just become more and more cryptic 😃

dranikpg: Renamed

src/server/tiered_storage.cc
```diff
@@ -454,7 +528,8 @@ void TieredStorage::CancelStash(DbIndex dbid, std::string_view key, PrimeValue*
   if (auto node = stash_backpressure_.extract(make_pair(dbid, key)); !node.empty())
     std::move(node.mapped()).Resolve(false);
 
-  if (OccupiesWholePages(value->Size())) {
+  auto estimated = EstimateSerializedSize(*value);
+  if (OccupiesWholePages(estimated->first)) {
```
dranikpg: This was a bug, we have to compute it not by `value->Size()`, but by the estimated serialization size. This might become an issue with large maps

src/server/tiered_storage.cc
```diff
@@ -73,6 +81,60 @@ tiering::DiskSegment FromCoolItem(const PrimeValue::CoolItem& item) {
...
+// TODO(vlad): Maybe split into different accessors?
+// Do NOT enforce rules depending on dynamic runtime values as this is called
+// when scheduling stash and just before succeeeding and is expected to return the same results
+optional<pair<size_t /*size*/, CompactObj::ExternalRep>> EstimateSerializedSize(
+    const PrimeValue& pv) {
+  switch (pv.ObjType()) {
+    case OBJ_STRING:
+      if (pv.IsInline())
+        return {};
+      return std::make_pair(pv.GetRawString().view().size(), CompactObj::ExternalRep::STRING);
```
dranikpg, line 93: Because we can't call GetRawString unconditionally. I'll try to make it nicer in the future, for now it just becomes too much for a single PR

src/server/tiering/serialized_map_test.cc
```diff
@@ -12,11 +12,11 @@ using namespace std;
 class SerializedMapTest : public ::testing::Test {};
 
 TEST_F(SerializedMapTest, TestBasic) {
-  const vector<std::pair<string_view, string_view>> kBase = {{"first key", "first value"},
-                                                             {"second key", "second value"},
-                                                             {"third key", "third value"},
-                                                             {"fourth key", "fourth value"},
-                                                             {"fifth key", "fifth value"}};
+  const vector<std::pair<string, string>> kBase = {{"first key", "first value"},
```
romange, line 15: curious, what's the reason for string?

src/server/tiered_storage.cc
```diff
@@ -73,6 +81,60 @@ tiering::DiskSegment FromCoolItem(const PrimeValue::CoolItem& item) {
...
+    case OBJ_HASH:
+      if (pv.Encoding() == kEncodingListPack) {
+        auto* lp = static_cast<uint8_t*>(pv.RObjPtr());
+        size_t bytes = lpBytes(lp);
+        bytes += lpLength(lp) * 2 * 4;
```
romange: can you add a comment explaining the formula?

src/server/tiered_storage.cc
```diff
@@ -73,6 +81,60 @@ tiering::DiskSegment FromCoolItem(const PrimeValue::CoolItem& item) {
...
+      detail::ListpackWrap lw{static_cast<uint8_t*>(pv.RObjPtr())};
+      vector<pair<string, string>> entries(lw.begin(), lw.end());
+      return tiering::SerializedMap::Serialize(
+          entries, {reinterpret_cast<char*>(buffer.data()), buffer.length()});
+    }
+  };
+  return 0;
+}
+
+string SerializeString(const PrimeValue& pv) {
```
romange: `SerializeToString`

src/server/tiered_storage.cc
```diff
@@ -73,6 +81,60 @@ tiering::DiskSegment FromCoolItem(const PrimeValue::CoolItem& item) {
   return {item.record->page_index * tiering::kPageSize + item.page_offset, item.serialized_size};
 }
 
+// Determine required byte size and encoding type based on value.
+// TODO(vlad): Maybe split into different accessors?
+// Do NOT enforce rules depending on dynamic runtime values as this is called
+// when scheduling stash and just before succeeeding and is expected to return the same results
+optional<pair<size_t /*size*/, CompactObj::ExternalRep>> EstimateSerializedSize(
```
romange: It does more than EstimateSize, and with having optional and precise ExternalRep the behavior is confusing. Maybe call it `GetSerializationDescriptor` which will return
```
struct SerializationDescriptor {
size estimated_size;
CompactObj::ExternalRep repr;
```

and add NONE to ExternalRep enum? or alternatively add `is_valid() { return size> 0; }` and return size=0 for unfit objects.

romange: Personally, I am not a big fan of chaining optionals—or other types like std::expected—one within another, especially when we control the wrapped class and it can describe the "undef" state itself. For example, in our codebase, we have using Result = std::optional<ResultType>; where `ResultType` is another optional, or `std::optional<facade::ErrorReply>` where `ErrorReply` can hold an empty state. These levels of indirection decrease readability, imho.

src/server/tiered_storage.cc
```diff
@@ -153,12 +224,13 @@ class TieredStorage::ShardOpManager : public tiering::OpManager {
       stats->tiered_used_bytes += segment.length;
       stats_.total_stashes++;
 
+      CompactObj::ExternalRep rep = EstimateSerializedSize(*pv)->second;
```
romange: ADD a DCHECK before dereferencing.

src/server/tiering/serialized_map_test.cc
dranikpg: Turing listpack into `vector<p<string_view, string_view>>` is even more code, so I just changed the type to `string`. In the future, the serialized map constructor needs to be updated eitehr way to be copy-less, but its too much for this PR. Its all temporary

src/server/tiered_storage.cc
dranikpg: the nullopt access with throw an assert either way

src/server/tiered_storage.cc
dranikpg: I agree here, but in a properly structured code base type composition is always almost preferable

dranikpg:
Removed the optional and just keeped a pair

src/server/tiered_storage.cc
```diff
@@ -610,14 +689,26 @@ size_t TieredStorage::ReclaimMemory(size_t goal) {
 }
 
 bool TieredStorage::ShouldStash(const PrimeValue& pv) const {
+  // Check value state
+  if (pv.IsExternal() || pv.HasStashPending())
+    return false;
+
+  // Estimate value size
+  auto [size, rep] = DetermineSerializationParams(pv);
```
romange, line 702: please swap condition with line 702. No need to do size estimations if hash offloading is disabled.

romange (commented):
LGTM + a minor comment.

src/server/tiered_storage.cc
```diff
@@ -386,36 +457,40 @@ std::optional<util::fb2::Future<bool>> TieredStorage::TryStash(DbIndex dbid, str
...
-
-  if (OccupiesWholePages(value->Size())) {  // large enough for own page
-    id = KeyRef(dbid, key);
-    stash_string(raw_string.view());
-  } else if (auto bin = bins_->Stash(dbid, key, raw_string.view()); bin) {
+  } else if (auto bin = bins_->Stash(dbid, key, SerializeToString(*value)); bin) {
     id = bin->first;
     // TODO(vlad): Write bin to prepared buffer instead of allocating one
-    stash_string(bin->second);
+    if (auto prepared = op_manager_->PrepareStash(est_size); prepared) {
```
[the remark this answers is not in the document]
romange: @dranikpg  wdyt? what is the relationship between `est_size` and `bin->second.size()` ?

dranikpg:
The issue found is non testable because we always provide at least 4kb, which is all thats needed for the bin

PR #6016, opened 2025-11-06
test: add spop benchmark

BorysTheDev:
Added benchmark for SPOP with different StringSet sparseness. SPOP isn't fast, and I see significant performance degradation when sparseness grows.

```
Benchmark                          Time             CPU   Iterations
--------------------------------------------------------------------
BM_Spop1000/sparseness:1     2100474 ns      2100454 ns         1000
BM_Spop1000/sparseness:4     7982310 ns      7982215 ns         1000
BM_Spop1000/sparseness:10   31592390 ns     31591116 ns         1000
```

### DIFF
--- a/src/core/string_set_test.cc
+++ b/src/core/string_set_test.cc
@@ -722,6 +722,28 @@ void BM_Grow(benchmark::State& state) {
 }
 BENCHMARK(BM_Grow);
 
+void BM_Spop1000(benchmark::State& state) {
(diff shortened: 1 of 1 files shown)


dranikpg (approved):
Profile would be helpful. Is it only the approach (iteration) or also some costs like creating the absl::BitGen each time? (we can replace with with insecure actually)

PR #6018, opened 2025-11-06
feat(search): Add query string length limit for FT.SEARCH and FT.AGGREGATE

vyavdoshenko:
Implements `search.query-string-bytes` configuration parameter to limit query string length in search operations, preventing potential performance issues from extremely long queries.

Changes:
- Added `search_query_string_bytes` flag (default: 10240 bytes)
- Enforced limit in `FT.SEARCH` and `FT.AGGREGATE` (including JOIN queries)
- Created `DenormalizeConfigName()` helper to convert internal names (`search_query_string_bytes`) to user-facing format (`search.query-string-bytes`)
- Extended `NormalizeConfigName()` to support dots in config names
- Fixed parser to accept placeholder vectors

+1 integration valkey-search test passed
42 failed, 28 passed

### DIFF
--- a/src/server/config_registry.cc
+++ b/src/server/config_registry.cc
@@ -4,6 +4,7 @@
 #include "server/config_registry.h"
 
 #include <absl/flags/reflection.h>
+#include <absl/strings/match.h>
 #include <absl/strings/str_replace.h>
 
 #include "base/logging.h"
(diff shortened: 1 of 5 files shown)


romange:
This a new feature that is covered by valkey integration tests but not implemented by us. Lets separate between lack of functionality and the existing functionality that is not working like intended. Our focus on the latter. So next time you encounter a test there that covers Dragonfly missing functionality, lets just exclude/skip it in our tests.

src/server/config_registry.cc
```diff
@@ -15,10 +16,27 @@ namespace {
...
 
 string NormalizeConfigName(string_view name) {
-  return absl::StrReplaceAll(name, {{"-", "_"}});
+  return absl::StrReplaceAll(name, {{"-", "_"}, {".", "_"}});
 }
 }  // namespace
 
+// Convert internal flag name back to user-facing format
+// Example: search_query_string_bytes -> search.query-string-bytes
+string DenormalizeConfigName(string_view name) {
```
romange, line 25: why it is needed?

vyavdoshenko: This is needed to convert search_query_string_bytes to search.query-string-bytes.
Using this it's possible to execute:
`CONFIG GET search.query-string-bytes`

romange: does the test require it?
I am fine that we silently accept `config set search.query-string-bytes` but display `search_query_string_bytes` as it is the way to set flags in dragonfly. the question whether the test breaks?

vyavdoshenko: This is one of test's check (`test_query_string_bytes_limit`)

```
        assert client.execute_command("CONFIG GET search.query-string-bytes") == [b"search.query-string-bytes", b"10240"]
```

romange (approved):

PR #6019, opened 2025-11-06
chore: fix dfly_bench qps calculation

romange:
Before we measured total requests by total time to get average qps.

While it sounds reasonable, due to assymetric networking socket performance, this always produces uneven pattern where at the end, only few connections are workingm, which brings down the average qps by dozens of percents.

This PR changes how total QPS is calculated. Now each qps is deduced independently per connection and at the and all the averages are aggregated.


dranikpg (approved):
👍🏻

PR #6020, opened 2025-11-06
chore: disk backpressure class utility

### LINKED ISSUE #6029: disk based queue
#5952 requires offloading socket data directly to disk and we need a utility class that handles this.

Implement a simple disk backed queue that can:

* write N bytes to disk
* read up to N bytes from disk

kostasrim:
This PR adds a simple producer/consumer utility class that for a backing file F can:

* async write/read blobs sequentially from disk
* tests

This class will be used by dragonfly connection to offload commands when dispatch queue surpasses a given value. 

No functional change in dragonfly_connection -- I just added the initialization paths and the flags needed.

Resolves #6029

### DIFF
--- a/src/facade/disk_backed_queue.cc
+++ b/src/facade/disk_backed_queue.cc
@@ -50,7 +50,7 @@ std::error_code DiskBackedQueue::Init() {
 }
 
 DiskBackedQueue::~DiskBackedQueue() {
-  DCHECK_EQ(in_flight_callbacks_, 0);
+  DCHECK_EQ(in_flight_callbacks_, 0ul);
 }
 
 std::error_code DiskBackedQueue::Close() {
@@ -99,6 +99,7 @@ void DiskBackedQueue::PushAsync(io::Bytes bytes, AsyncPushCallback cb) {
   ++in_flight_callbacks_;
 
   file_->WriteAsync(bytes, offset, [this, size, cb = std::move(cb)](int res) {
+    --in_flight_callbacks_;
     if (res < 0) {
       std::error_code ec{-res, std::system_category()};
       VLOG(2) << "Failed to offload blob of size " << size << " to backing with error: " << ec;
@@ -110,7 +111,6 @@ void DiskBackedQueue::PushAsync(io::Bytes bytes, AsyncPushCallback cb) {
     total_backing_bytes_ += size;
     VLOG(2) << "Offload connection " << this << " backpressure of " << size;
     cb({});
-    --in_flight_callbacks_;
   });
 }
 
@@ -123,6 +123,7 @@ void DiskBackedQueue::PopAsync(io::MutableBytes out, AsyncPopCallback cb) {
   io::MutableBytes read_buf = out.subspan(0, to_read);
 
   file_->ReadAsync(read_buf, offset, [this, to_read, offset, cb = std::move(cb)](int res) {
+    --in_flight_callbacks_;
     if (res < 0) {
       std::error_code ec{-res, std::system_category()};
       LOG(ERROR) << "Could not load item at offset " << offset << " of size " << to_read
@@ -141,7 +142,6 @@ void DiskBackedQueue::PopAsync(io::MutableBytes out, AsyncPopCallback cb) {
     MaybePunchHole();
 
     cb(bytes_read);
-    --in_flight_callbacks_;
   });
 }
 


src/facade/disk_connection_backpressure.cc
```diff
@@ -0,0 +1,132 @@
...
+  }
+  reader_.reset(*res);
+
+  VLOG(3) << "Created backing for connection " << this << " " << backing_name;
+
+  return {};
+}
+
+DiskBackedBackpressureQueue::~DiskBackedBackpressureQueue() {
+  auto ec = writer_->Close();
```
romange: nit: best practices suggest separating between i/o operations and d'tor functions

src/facade/disk_connection_backpressure.h
```diff
@@ -0,0 +1,65 @@
...
+  std::unique_ptr<io::WriteFile> writer_;
+  std::unique_ptr<io::ReadonlyFile> reader_;
+
+  // In memory backed file map
+  struct ItemOffset {
+    size_t offset = 0;
+    size_t total_bytes = 0;
+  };
+
+  std::deque<ItemOffset> offsets_;
```
romange: why do we queue of offsets_ ?

src/facade/disk_connection_backpressure.cc
kostasrim: 100% if you recall we had issues around ProtocolClient destructor recently because of that 😄

src/facade/disk_connection_backpressure.h
romange: in other words, I suspect that we can do the same in O(1) memory instead of O(N)

kostasrim: I think what you suspect is correct. I imagine you have something like this in mind:

[8 bytes + msg] [8 bytes + msg] 

First 8 bytes are the size of the message. Now you can:

Read offset 0..8 -> msg size X
Read msg of size X ->
Read offset 8 + size X -> msg size of Y 
and so on and that way you don't need to track any of the items.

IMO I wanted a simple solution first but if you think it's better to add it here I am happy to do so

romange: yes, let's do this, and also add a comment in the code where we (de)serialize about the on-disk format

src/facade/disk_connection_backpressure.h
```diff
@@ -0,0 +1,65 @@
+// Copyright 2025, DragonflyDB authors.  All rights reserved.
+// See LICENSE for licensing terms.
+//
+
+#pragma once
+
+#include <util/fibers/uring_file.h>
```
romange: you do not need uring or linux specific api for this, imho.

src/facade/disk_connection_backpressure.h
```diff
@@ -0,0 +1,65 @@
...
+#include <functional>
+#include <memory>
+#include <string_view>
+#include <system_error>
+
+#include "io/io.h"
+
+namespace facade {
+
+class DiskBackedBackpressureQueue {
```
romange: let's rename it to `DiskBackedQueue` and the file should be renamed as well.

src/facade/disk_connection_backpressure.h
```diff
@@ -0,0 +1,65 @@
...
+
+  std::error_code Init();
+
+  // Check if we can offload bytes to backing file.
+  bool HasEnoughBackingSpaceFor(size_t bytes) const;
+
+  // Total size of internal buffers/structures.
+  size_t TotalInMemoryBytes() const;
+
+  void OffloadToBacking(std::string_view blob);
```
romange: all i/o commands should return std::error_code. otherwise how a caller will know something is wrong?

src/facade/disk_connection_backpressure.h
```diff
@@ -0,0 +1,65 @@
...
+
+  // For each item loaded from disk it calls f(item) to consume it.
+  // Reads up to max_queue_load_size_ items on each call
+  void LoadFromBacking(std::function<void(io::MutableBytes)> f);
```
romange: LoadFromBacking/OffloadToBacking

should be renamed to `Pop/Push` or `Enqueue/Dequeue` or something like this

src/facade/disk_connection_backpressure.h
dranikpg: The format is ok, but you can avoid a double reads by reading the _next size ahead_ if you have more than 1 items. Basically, you store every time the size of the next item to be read in memory. When you read an item, you read it + 4 bytes, so you know the size of the next one

src/facade/disk_connection_backpressure.h
dranikpg: I really dislike such interfaces that _internalize_ action. What if the connection wants to stop based on execution time, some other state conditions, shutdown, etc. It's not the queues job to manage this at all

`string_view/size_t Read(std::string* buffer)` whould be enough

src/facade/disk_connection_backpressure.cc
```diff
@@ -0,0 +1,132 @@
...
+
+void DiskBackedBackpressureQueue::OffloadToBacking(std::string_view blob) {
+  ItemOffset item;
+  item.offset = next_offset_;
+  item.total_bytes = blob.size();
+
+  // TODO we should truncate as the file grows. That way we never end up with large files
+  // on disk.
+  auto res = writer_->Write(blob);
+  if (res) {
```
dranikpg: res looks like a Result to me, but not like an ec -> error_code 🙂

src/facade/disk_connection_backpressure.cc
```diff
@@ -0,0 +1,132 @@
...
+bool DiskBackedBackpressureQueue::Empty() const {
+  return total_backing_bytes_ == 0;
+}
+
+bool DiskBackedBackpressureQueue::HasEnoughBackingSpaceFor(size_t bytes) const {
+  return (bytes + total_backing_bytes_) < max_backing_size_;
+}
+
+size_t DiskBackedBackpressureQueue::TotalInMemoryBytes() const {
+  return offsets_.size() * sizeof(ItemOffset);
```
dranikpg: sadly there's really no way to get the capacity of a dequeue

src/facade/disk_connection_backpressure.h
```diff
@@ -0,0 +1,65 @@
...
+  std::deque<ItemOffset> offsets_;
+
+  size_t total_backing_bytes_ = 0;
+  size_t next_offset_ = 0;
+
+  // Read only constants
+  const size_t max_backing_size_ = 0;
+  const size_t max_queue_load_size_ = 0;
+
+  // unique id for the file backed
```
dranikpg: you can add a comment that it it's equal to the connection id

src/facade/disk_connection_backpressure.h
romange: Good idea! we can assume 4 bytes blob lengths indeed.

src/facade/disk_connection_backpressure.h
romange: +1 to that

src/facade/disk_connection_backpressure.h
kostasrim: There are 3 API's:

1. Uring (uring_file.h) -> asynchronous fiber blocking
2. Linux (file.h) -> `thread blocking` via writev/preadv
3. C++ files (fstream) - > `thread blocking`

Why would we block the entire thread on a disk write ?

src/facade/disk_backed_queue.cc
```diff
@@ -0,0 +1,154 @@
...
+               << result.error().message();
+    return result.error();
+  }
+
+  VLOG(2) << "Loaded item with offset " << next_read_offset_ << " of size " << read_sz
+          << " for connection " << this;
+
+  next_read_offset_ += bytes.size();
+
+  if (total_backing_items_ > 1) {
```
dranikpg: can we push and pop at the same time? The variable might have changed

src/facade/disk_backed_queue.cc
```diff
@@ -0,0 +1,154 @@
...
+    auto buf = bytes.subspan(bytes.size() - sizeof(uint32_t));
+    uint32_t val = ((uint32_t)buf[0]) | ((uint32_t)buf[1] << 8) | ((uint32_t)buf[2] << 16) |
+                   ((uint32_t)buf[3] << 24);
```
dranikpg: Maybe use absl::___endian::Store32/Load32 looks a little nicer 🙂

src/facade/disk_backed_queue.h
```diff
@@ -0,0 +1,60 @@
...
+  size_t next_item_total_bytes_ = 0;
+
+  // Read only constants
+  const size_t max_backing_size_ = 0;
+  const size_t max_queue_load_size_ = 0;
+
+  // same as connection id. Used to uniquely identify the backed file
+  const size_t id_ = 0;
+
+  std::string buffer;
```
dranikpg: underscore at the end, just confused me

src/facade/disk_backed_queue.cc
```diff
@@ -0,0 +1,154 @@
...
+
+  VLOG(2) << "Offload connection " << this << " backpressure of " << blob.size();
+  VLOG(3) << "Command offloaded: " << blob;
+  return {};
+}
+
+std::error_code DiskBackedQueue::Pop(std::string* out) {
+  // We read the next item and (if there are more) we also prefetch the next item's size.
+  uint32_t read_sz = next_item_total_bytes_ + (total_backing_items_ > 1 ? sizeof(uint32_t) : 0);
+  buffer.resize(read_sz);
```
dranikpg: Given that you use buffer here, parallel pull and pop are not allowed. But either way, can't we use out directly to read into?

src/facade/disk_backed_queue.cc
```diff
@@ -0,0 +1,154 @@
...
+bool DiskBackedQueue::Empty() const {
+  return total_backing_bytes_ == 0;
+}
+
+bool DiskBackedQueue::HasEnoughBackingSpaceFor(size_t bytes) const {
+  return (bytes + total_backing_bytes_) < max_backing_size_;
+}
+
+std::error_code DiskBackedQueue::Push(std::string_view blob) {
+  // TODO we should truncate as the file grows. That way we never end up with large files
```
romange: what should you truncate?

src/facade/disk_connection_backpressure.h
romange: I meant to use `io/file.h` interface, you do not need to leak the implementation in h file. 
For epoll we can use `helio/util/fibers/fiber_file.h` though it's not a priority and it's fine to assume this feature is only for iouring enabled systems but in any case you will have to write code in such way that it will build on macos (so better not to leak iouring in header files)

src/facade/disk_backed_queue.cc
kostasrim: The backing file when it grows too much

kostasrim:
@dranikpg we changed direction and we will only push/pop raw data. The rationale behind this is that we want to unblock the ConnFb. See: https://github.com/dragonflydb/dragonfly/issues/5952

src/facade/disk_backed_queue.h
romange: @kostasrim

src/facade/disk_backed_queue.h
```diff
@@ -0,0 +1,56 @@
...
+  std::error_code CloseReader();
+  std::error_code CloseWriter();
+
+ private:
+  // File Reader/Writer
+  std::unique_ptr<io::WriteFile> writer_;
+  std::unique_ptr<io::ReadonlyFile> reader_;
+
+  size_t total_backing_bytes_ = 0;
+  size_t total_backing_items_ = 0;
```
romange: total_backing_items_ is redundant now

src/facade/disk_backed_queue.h
```diff
@@ -0,0 +1,56 @@
...
+  ~DiskBackedQueue();
+
+  std::error_code Init();
+
+  // Check if we can offload bytes to backing file.
+  bool HasEnoughBackingSpaceFor(size_t bytes) const;
+
+  std::error_code Push(std::string_view blob);
+
+  std::error_code Pop(std::string* out);
```
romange: it's not a good interface now - as you do not pop an element. You will usually have a destination buffer which you will want to fill till the end of its capacity or less. Similarly to socket Recv() interface.

src/facade/disk_backed_queue.h
```diff
@@ -0,0 +1,56 @@
...
+
+  // Check if backing file is empty, i.e. backing file has 0 bytes.
+  bool Empty() const;
+
+  std::error_code CloseReader();
```
romange: why do we need to control Reader and Writer independently? 
Seems that if Empty(), you just want to shut the queue altogether

romange:
I think there is still work that needs to be done. In any case, I suggest checking how it will fit once you integreate `OnRecv` and be able to offload/load inside DflyConn.

src/facade/disk_backed_queue.h
kostasrim: If we merge those two then don't we need to return a pair  ? I merged the two under Close() and return the first error (if any)

src/facade/disk_backed_queue.h
kostasrim: +1, I adjusted it

src/facade/disk_backed_queue.h
```diff
@@ -0,0 +1,53 @@
...
+ public:
+  explicit DiskBackedQueue(uint32_t conn_id);
+  ~DiskBackedQueue();
+
+  std::error_code Init();
+
+  // Check if we can offload bytes to backing file.
+  bool HasEnoughBackingSpaceFor(size_t bytes) const;
+
+  std::error_code Write(io::Bytes bytes);
```
kostasrim: @romange we would need a flag in dfly_conn because a fiber blocking disk write can't be called in the cb passed to RegisterRecv. The other option would be to go async here but IMO not worth it

src/facade/disk_backed_queue.cc
```diff
@@ -0,0 +1,121 @@
...
+    LOG(ERROR) << "Could not load item at offset " << next_read_offset_ << " of size " << to_read
+               << " from disk with error: " << result.error().value() << " "
+               << result.error().message();
+    return result;
+  }
+
+  VLOG(2) << "Loaded item with offset " << next_read_offset_ << " of size " << to_read
+          << " for connection " << this;
+
+  next_read_offset_ += to_read;
```
romange: you do not delete the head, and the file will grow constantly to infinity.
I would expect to see fallocate/FALLOC_FL_PUNCH_HOLE to remove unused physical blocks.

also if  `total_backing_bytes_` reaches zero, I would expect we truncate/delete the file.

kostasrim: > you do not delete the head, and the file will grow constantly to infinity.
I would expect to see fallocate/FALLOC_FL_PUNCH_HOLE to remove unused physical blocks.

I added a call to FALLO_FL_PUNCH (also respecting allignment sizes). I suspect this is not the optimal solution because a) fallocate is not cpu only here we do not even go via iouring b) we could reuse the blocks of the file such that we don't reallocate blocks that often. 


> also if total_backing_bytes_ reaches zero, I would expect we truncate/delete the file.

I delete/unlink the file in `Close()` instead. Why would I want to create a new file each time we empty it -- it will go back and forth and it seems redundant.

kostasrim: Imo we should also consider using LinuxFile instead and fallocate via PrepFallocate and FiberCall instead

kostasrim: and I also raised a small helio PR with fallocate test and a small behavioral question

kostasrim:
@romange PTAL, I added async variants, fixed punch hole bugs and added tests. We just write and read bytes to a file either synchronously or asynchronously.

My question is as is we have:

ioloop v2 -> memcache 
redis -> still 2 fibers


1. Which one do we want to priotorize to use the disk based queue?
2. Blocking writes are fine for ioloop but not for v2 (memcche). Do you want me to keep both interfaces ?

romange:
I think we should integrate it with `ioloop v2 `

src/facade/disk_backed_queue.cc
```diff
@@ -0,0 +1,188 @@
...
+void DiskBackedQueue::MaybePunchHole() {
+  // Punch holes over the aligned region we have fully read past so the OS can reclaim pages.
+  // Both offset and length must be multiples of the filesystem block size: XFS returns EINVAL
+  // otherwise, and ext4/tmpfs only zero partial blocks rather than freeing them.
+  // We assume 4096-byte blocks (correct for virtually all deployments); a fully robust
+  // implementation would query the actual block size via fstatfs(file_->GetFd(), &fsst) and
+  // align to fsst.f_bsize instead.
+  const size_t aligned_end = (next_read_offset_ / 4096) * 4096;
+  if (aligned_end > punch_offset_) {
+    int res = fallocate(file_->GetFd(), FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE, punch_offset_,
```
romange, line 89: curious, have you explored the option to avoid `FALLOC_FL_KEEP_SIZE` ?

src/facade/disk_backed_queue.h
romange: I do not understand this comment. is it still relevant?

src/facade/disk_backed_queue.h
```diff
@@ -0,0 +1,66 @@
...
+  io::Result<size_t> ReadTo(io::MutableBytes out);
+
+  using AsyncWriteCallback = std::function<void(std::error_code)>;
+
+  void WriteAsync(io::Bytes bytes, AsyncWriteCallback cb);
+
+  using AsyncReadCallback = std::function<void(io::Result<size_t>)>;
+
+  // Async read variant. Callback is invoked with Result containing bytes read or error.
+  void ReadToAsync(io::MutableBytes out, AsyncReadCallback cb);
```
romange: consider renaming them to "PullAsync" and PushAsync"

src/facade/disk_backed_queue.h
```diff
@@ -0,0 +1,66 @@
...
+
+  size_t write_offset_ = 0;
+  size_t total_backing_bytes_ = 0;
+  size_t next_read_offset_ = 0;
+  // Tracks how far into the file holes have been punched (always 4096-aligned).
+  size_t punch_offset_ = 0;
+
+  // Read only constants
+  const size_t max_backing_size_ = 0;
+  const size_t max_queue_load_size_ = 0;
```
romange: remove `max_queue_load_size_`

src/facade/disk_backed_queue.cc
kostasrim: You mean ftruncate ? Otherwise I don't understand, PUNCH_HOLE must be ored with KEEP_SIZE

romange: nm, I did not know this.

kostasrim: For completenes I quote:

```
The FALLOC_FL_PUNCH_HOLE flag must be ORed with
       FALLOC_FL_KEEP_SIZE in mode; in other words, even when punching
       off the end of the file, the file size (as reported by [stat(2)](https://man7.org/linux/man-pages/man2/stat.2.html))
       does not change.
```

src/facade/disk_backed_queue.cc
```diff
@@ -0,0 +1,183 @@
...
+}
+
+std::error_code DiskBackedQueue::Close() {
+  if (file_) {
+    auto ec = file_->Close();
+    LOG_IF(WARNING, ec) << ec.message();
+
+    std::string backing = absl::StrCat(absl::GetFlag(FLAGS_disk_backpressure_folder), id_);
+    int errc = unlink(backing.c_str());
+    LOG_IF(ERROR, errc != 0) << "Failed to unlink backing file: "
```
romange, line 63: if you return `error_code`  no need to LOG(ERROR) here.

src/facade/disk_backed_queue.cc
```diff
@@ -0,0 +1,183 @@
...
+
+  write_offset_ += bytes.size();
+  total_backing_bytes_ += bytes.size();
+
+  VLOG(2) << "Offload connection " << this << " backpressure of " << bytes.size();
+  return {};
+}
+
+io::Result<size_t> DiskBackedQueue::Pop(io::MutableBytes out) {
+  const size_t k_read_size = 4096;
```
romange: why do you limit reading by 4KB if `out.size()` is larger?

src/facade/disk_backed_queue.cc
```diff
@@ -0,0 +1,183 @@
...
+// Check if backing file is empty, i.e. backing file has 0 bytes.
+bool DiskBackedQueue::Empty() const {
+  return total_backing_bytes_ == 0;
+}
+
+bool DiskBackedQueue::HasEnoughBackingSpaceFor(size_t bytes) const {
+  return (bytes + total_backing_bytes_) < max_backing_size_;
+}
+
+std::error_code DiskBackedQueue::Push(io::Bytes bytes) {
```
romange: so what's the plan? are we going to use sync methods?

src/facade/disk_backed_queue.cc
```diff
@@ -0,0 +1,183 @@
...
+void DiskBackedQueue::MaybePunchHole() {
+  // Punch holes over the aligned region we have fully read past so the OS can reclaim pages.
+  // Both offset and length must be multiples of the filesystem block size: XFS returns EINVAL
+  // otherwise, and ext4/tmpfs only zero partial blocks rather than freeing them.
+  // We assume 4096-byte blocks (correct for virtually all deployments); a fully robust
+  // implementation would query the actual block size via fstatfs(file_->GetFd(), &fsst) and
+  // align to fsst.f_bsize instead.
+  const size_t aligned_end = (next_read_offset_ / 4096) * 4096;
+  if (aligned_end > punch_offset_) {
+    int res = fallocate(file_->GetFd(), FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE, punch_offset_,
```
romange, line 89: so MaybePunchHole is always synchronous ? 
should we use async fallocate too?

src/facade/dragonfly_connection.cc
```diff
@@ -623,6 +627,18 @@ Connection::Connection(Protocol protocol, util::HttpListenerBase* http_listener,
 #endif
 
   UpdateLibNameVerMap(lib_name_, lib_ver_, +1);
+
+  backpressure_to_disk_watermark_ = absl::GetFlag(FLAGS_disk_queue_offload_watermark);
+  if (backpressure_to_disk_watermark_ > 0) {
```
romange: I would not want us to create a file for each connection unconditionally. Lets remove the connection code in this PR and limit it only for the queue

src/facade/disk_backed_queue.cc
kostasrim: It is as is. I had a comment to make it asynchronous but it should be fine for now

src/facade/disk_backed_queue.cc
kostasrim: The plan is to be fully asynchronous since there is a single fiber that multiplexes different roles (parsing, offloading to disk, dispatching commands async etc). 

I kept both the API's in case we need it (+the original synchronous version dependeing on it) 🤷

src/facade/disk_backed_queue.cc
kostasrim: What if Close() failed ? we will overwrite the error. Maybe worth even splitting. It's ok for now, I will adjust accordingly on the backpressure PR

src/facade/disk_backed_queue.cc
```diff
@@ -0,0 +1,143 @@
...
+    DCHECK_EQ(res, 0) << "fallocate punch failed: " << strerror(errno);
+    punch_offset_ = aligned_end;
+  }
+}
+
+void DiskBackedQueue::PushAsync(io::Bytes bytes, AsyncPushCallback cb) {
+  const size_t offset = write_offset_;
+  const size_t size = bytes.size();
+
+  file_->WriteAsync(bytes, offset, [this, size, cb = std::move(cb)](int res) {
```
romange, line 101: must bugs are around shutdown and ownership issues. add `pending_callbacks_` counter, increase it on each async invocation and dcheck it's 0 in the destructor.

romange: We handle this in DiscStorage class as well.

PR #6021, opened 2025-11-06
chore: Enable valkey-search tests in CI

vyavdoshenko:
Integrates valkey-search integration tests into the CI pipeline

Changes:
- Modified `sync-valkey-search-tests.sh` to accept tag/revision parameter
- Added sync step in CI to use specific valkey-search revision
- Implemented test skip mechanism in `conftest.py`
- Added skip list for tests incompatible with Dragonfly

### DIFF
--- a/tests/dragonfly/valkey_search/README.md
+++ b/tests/dragonfly/valkey_search/README.md
@@ -57,15 +57,4 @@ tests/dragonfly/valkey_search/
 3. **Mock framework** (`__init__.py`) replaces valkey-search imports with Dragonfly equivalents
 4. **Adapter** (`valkey_search_test_case_dragonfly.py`) creates real Dragonfly instances with replicas
 5. **Original tests run unchanged** - all adaptation happens in infrastructure layer
-6. **Python 3.8 compatibility** - sync script patches `indexes.py` to add `from __future__ import annotations`
-
-## Skipping Tests
-
-Tests can be skipped by adding them to the `SKIP_TESTS` list in [conftest.py](conftest.py):
-
-```python
-SKIP_TESTS = [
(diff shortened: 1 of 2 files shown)


romange:
seems like it does not work for older versions of python. 
I am fine if it run only on ubuntu 24 or later.

.github/workflows/ci.yml
```diff
@@ -196,7 +196,7 @@ jobs:
           path: /tmp/*INFO*
 
       - name: Run regression tests
-        if: matrix.container == 'ubuntu-dev:20'
+        if: matrix.container == 'ubuntu-dev:20' || matrix.container == 'ubuntu-dev:24'
```
romange: what's this?

vyavdoshenko: Regression tests executed for ubuntu-dev:20, not for ubuntu-dev:24
Please take a look any PR checks, they don't have python tests executed.
For example:
https://github.com/dragonflydb/dragonfly/actions/runs/19133606754/job/54680146596?pr=6018

vyavdoshenko: I added regression step for ubuntu-dev:24

romange (approved):

vyavdoshenko:
@romange 
I gave up in trying pass tests on ubuntu-dev:24 because dragonfly python tests don't pass on this setup. I suggest to fix in separate PR/task. Now I decided to patch valkey-search tests adding `from __future__ import annotations` in the beginning of each python file. It helps. I guess it's acceptable.

romange:
yes, sure I tried to suggest a simple solution but it was only  a suggestion.

PR #6023, opened 2025-11-07
feat(core): Add defrag support for json objects

abhijat:
Defrag support is added for JSON compact objects for both cons and flat type encoding.

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -723,7 +723,7 @@ void RobjWrapper::MakeInnerRoom(size_t current_cap, size_t desired, MemoryResour
 }  // namespace detail
 
 uint32_t JsonEnconding() {
-  static thread_local uint32_t json_enc =
+  thread_local uint32_t json_enc =
       absl::GetFlag(FLAGS_experimental_flat_json) ? kEncodingJsonFlat : kEncodingJsonCons;
   return json_enc;
 }
--- a/src/core/page_usage_stats_test.cc
+++ b/src/core/page_usage_stats_test.cc
@@ -4,22 +4,25 @@
 
 #include "core/page_usage_stats.h"
 
+#include <absl/flags/reflection.h>
 #include <gmock/gmock-matchers.h>
 
 #include "base/gtest.h"
 #include "base/logging.h"
 #include "core/compact_object.h"
-#include "core/mi_memory_resource.h"
 #include "core/qlist.h"
 #include "core/score_map.h"
 #include "core/small_string.h"
 #include "core/sorted_map.h"
 #include "core/string_map.h"
 #include "core/string_set.h"
+#include "redis/redis_aux.h"
 
 extern "C" {
 #include "redis/zmalloc.h"
 }
+
+ABSL_DECLARE_FLAG(bool, experimental_flat_json);
 
 using namespace dfly;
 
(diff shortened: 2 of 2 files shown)


src/core/page_usage_stats_test.cc
```diff
@@ -176,3 +181,28 @@ TEST_F(PageUsageStatsTest, StatCollection) {
   const CollectedPageStats::ShardUsageSummary expected{{50, 65}, {90, 85}, {99, 89}};
   EXPECT_EQ(usage.at(1), expected);
 }
+
+TEST_F(PageUsageStatsTest, JSONCons) {
```
abhijat, line 185: Unfortunately it is not possible to test for flat encoding in CI due to the static lifetime of the encoding flag. 

Even if the test accesses internal fields of `json_obj` union directly to call `DefragIfNeeded`, the `Free` code path later on crashes trying to free a compact object which has an encoding which does not match the static setting.

I did test flat encoding manually in a matching test and it worked as expected.

abhijat: Added a comment explaining this

romange (commented):
Flat json is not really used, do we use jsoncons in prod. I did not understand how you defragment json as it can.contain lots of pointers due to its hierarchical structure

abhijat:
> do we use jsoncons in prod. 

Yes jsoncons should the default encoding based on `experimental_flat_json` flag which is false by default, I could see its use in `facade::OpStatus dfly::SetFullJson(...)` which is called via `json.set` etc.

> I did not understand how you defragment json as it can.contain lots of pointers due to its hierarchical structure

Yes, the current implementation works better for flat encoding as it is a single chunk of data, but for `cons` it only copies the root of the tree so it is not effective. It should walk down the tree and move everything for cons. 

If we don't use flat encoding much and use mostly cons, then this PR needs more work. I will look into defragmenting the tree from the root downwards.

abhijat:
>  I will look into defragmenting the tree from the root downwards.

Actually, it doesn't look like this is possible, as far as I can see it might not be possible to go into the object's fields and individually reallocate them. 

It might be possible to do the following more expensive operations:

* serialize the object to string
* convert it back to a JsonType using `dfly::JsonFromString` to get new allocations in the page reserved for malloc
* move it into `u_.json_obj.cons.json_ptr`

abhijat:
> * serialize the object to string
> * convert it back to a JsonType using `dfly::JsonFromString` to get new allocations in the page reserved for malloc
> * move it into `u_.json_obj.cons.json_ptr`

I tried this approach in the latest commit, will test it a bit more

abhijat:
@romange does the current approach look better? We basically copy the JSON object, in the process reallocating memory

src/core/compact_object.cc
```diff
@@ -1585,6 +1587,35 @@ MemoryResource* CompactObj::memory_resource() {
   return tl.local_mr;
 }
 
+bool CompactObj::JsonConsT::DefragIfNeeded(PageUsage* page_usage) {
+  if (JsonType* old = json_ptr; page_usage->IsPageForObjectUnderUtilized(old)) {
```
romange: note: the trigger condition is also problematic. the object may be part of fully utilized page but have hundreds of other heap objects. It is possible to solve this using the latest PATCH of @vyavdoshenko 
here: https://github.com/dragonflydb/jsoncons/commit/990b0a2077a50a62a486eac2647f0411e65b3bbe

you can pass into `compute_memory_size` a callback that will return 0 or 1 if a pointer belongs to the under utilized page and if the final result is > 0, than there is at least one pointer that needs to be defragmented.

romange: the patch is already present in jsoncons.

abhijat: I used `compute_memory_size` but kept state (a boolean) outside to figure out if any pointer is in an underutilized page.

PR #6025, opened 2025-11-09
chore: support huffman compression for string values

romange:
so now we can pass both `--huffman_table=STRINGS:....,KEYS:....`

Moreover, fix bugs around huffman and tiering so that at least a trivial manual test passes when both settings are enabled:

```
./dragonfly   --dbfilename= --noversion_check --maxmemory=8G  --logtostderr  --huffman_table=STRINGS:ChD4bAf/D/bPSwY= --tiered_prefix /tmp/tiered

> debug populate 1000000 key 128
> get key:42
```

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -1278,6 +1278,8 @@ void CompactObj::Reset() {
 }
 
 uint8_t CompactObj::GetFirstByte() const {
+  DCHECK_EQ(ObjType(), OBJ_STRING);
+
   if (IsInline()) {
     return u_.inline_str[0];
   }


src/core/compact_object.cc
```diff
@@ -1271,6 +1277,33 @@ void CompactObj::Reset() {
...
+
+  if (taglen_ == SMALL_TAG) {
+    return u_.small_str.first_byte();
+  }
+
+  if (taglen_ == EXTERNAL_TAG) {
+    if (u_.ext_ptr.is_cool) {
+      const CompactObj& cooled_obj = u_.ext_ptr.cool_record->value;
+      return cooled_obj.GetFirstByte();
+    }
```
dranikpg, line 1301: At all time the cooled object is equal to the tiered one, so it could be just ext_ptr.first_byte

src/core/compact_object.cc
```diff
@@ -1271,6 +1277,33 @@ void CompactObj::Reset() {
   mask_ = 0;
 }
 
+uint8_t CompactObj::GetFirstByte() const {
```
dranikpg, line 1280: DCHECK_EQ(ObjType, OBJ_STRING)

dranikpg (dismissed):

src/core/compact_object.cc
romange: we currently do not set `u_.ext_ptr.first_byte` for cooled objects, only for fully external.
I was not sure if we can get out of sync, so preferred to read from the "original".

PR #6031, opened 2025-11-10
fix(tiering): Recompute size before upload

dranikpg:
1. When we materialize a modified string (that was decoded), we have to reset the compact object encoding (added a better test)
2. Setting the estimated byte size only when initializing is wrong, as the value can be modified and that size grows. `GetMetrics` is now called to get some basic info to decide on upload

### DIFF
--- a/src/server/tiered_storage_test.cc
+++ b/src/server/tiered_storage_test.cc
@@ -251,6 +251,7 @@ TEST_F(TieredStorageTest, Defrag) {
 TEST_F(TieredStorageTest, BackgroundOffloading) {
   absl::FlagSaver saver;
   SetFlag(&FLAGS_tiered_offload_threshold, 1.0f);      // offload all values
+  SetFlag(&FLAGS_tiered_upload_threshold, 0.0f);       // upload all values
   SetFlag(&FLAGS_tiered_experimental_cooling, false);  // The setup works without cooling buffers
   UpdateFromFlags();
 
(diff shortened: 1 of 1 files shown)


dranikpg:
Not sure if I like the name `metrics` 🤷🏻‍♂️

src/core/compact_object.cc
```diff
@@ -1265,6 +1259,7 @@ void CompactObj::Materialize(std::string_view blob, bool is_raw) {
       u_.r_obj.SetString(blob, tl.local_mr);
     }
   } else {
+    mask_bits_.encoding = NONE_ENC;  // reset encoding
     EncodeString(blob, false);
```
dranikpg, line 1262: oh heck, I need to add some better tests soon

src/core/compact_object.cc
```diff
@@ -1265,7 +1258,7 @@ void CompactObj::Materialize(std::string_view blob, bool is_raw) {
       u_.r_obj.SetString(blob, tl.local_mr);
     }
   } else {
-    EncodeString(blob, false);
+    SetString(blob, false);
```
dranikpg: Actually the size after editing an offlaoded string can be smaller, so we have to call SetString

dranikpg: No... there is no command that makes it shrink 🤦🏻

dranikpg:
finally... @romange  PTAL

src/server/tiering/decoders.h
```diff
@@ -17,6 +17,11 @@ namespace dfly::tiering {
 // Decodes serialized value and provides it to callbacks.
 // Acts as generic interface to callback driver (OpManager)
 struct Decoder {
+  struct UploadMetrics {
+    bool modified;               // whether the value as modified
```
romange, line 21:
```suggestion
    bool modified;               // whether the value was modified
```

romange: you are ignoring valid important fixes :)

PR #6033, opened 2025-11-10
fix: Optimize GetRandomChain() in DenseSet

vyavdoshenko:
Relevant to: https://github.com/dragonflydb/dragonfly/pull/6016

Improves SPOP performance on sparse hash tables (created by Reserve()) through three key optimizations:

1. **thread_local RNG**: Use thread_local absl::InsecureBitGen instead of 
   creating a new BitGen on each call, eliminating construction overhead

2. **Iterator increment**: Replace modulo operation with iterator increment 
   and wrap-around check, avoiding expensive division in the hot loop

3. **IsEmpty() early check**: Test IsEmpty() before calling ExpireIfNeeded() 
   to skip expensive expiration checks on empty buckets

Comparison:
Main branch:
```
BM_Spop1000/sparseness:1                    5964193 ns      5963687 ns          117
BM_Spop1000/sparseness:4                    6365861 ns      6365321 ns          110
BM_Spop1000/sparseness:10                   7633700 ns      7633658 ns           91
BM_Spop1000/sparseness:40                  13036086 ns     13035934 ns           54
BM_Spop1000/sparseness:100                 20192890 ns     20192937 ns           35
```

Fixed version (current):
```
BM_Spop1000/sparseness:1                    1232039 ns      1232019 ns          568
BM_Spop1000/sparseness:4                    1590988 ns      1590901 ns          440
BM_Spop1000/sparseness:10                   2773856 ns      2773784 ns          252
BM_Spop1000/sparseness:40                   7754844 ns      7754870 ns           90
BM_Spop1000/sparseness:100                 14499281 ns     14499131 ns           48
```

Previous approach (wrong direction) (Rejection Sampling):
```
BM_Spop1000/sparseness:1                    1351481 ns      1351497 ns          518
BM_Spop1000/sparseness:4                    2342621 ns      2342612 ns          299
BM_Spop1000/sparseness:10                   5463768 ns      5463703 ns          129
BM_Spop1000/sparseness:40                  13164771 ns     13164710 ns           53
BM_Spop1000/sparseness:100                 21711212 ns     21711052 ns           32
```

The current comparison is based on fixed benchmark: https://github.com/dragonflydb/dragonfly/pull/6038

### DIFF
--- a/src/core/dense_set.cc
+++ b/src/core/dense_set.cc
@@ -27,6 +27,8 @@ constexpr size_t kMinSizeShift = 2;
 constexpr size_t kMinSize = 1 << kMinSizeShift;
 constexpr bool kAllowDisplacements = true;
 
+thread_local absl::InsecureBitGen tl_bit_gen;
+
 #define PREFETCH_READ(x) __builtin_prefetch(x, 0, 1)
 
 DenseSet::IteratorBase::IteratorBase(const DenseSet* owner, bool is_end)
@@ -677,10 +679,7 @@ DenseSet::ChainVectorIterator DenseSet::GetRandomChain() {
     return entries_.end();
   }
 
-  // Use thread-local generator to avoid construction overhead
-  thread_local absl::InsecureBitGen gen;
-
-  size_t offset = absl::Uniform<size_t>(gen, 0u, entries_.size());
+  size_t offset = absl::Uniform<size_t>(tl_bit_gen, 0u, entries_.size());
 
   // Start at random position and scan linearly with wrap-around
   auto it = entries_.begin() + offset;
@@ -707,8 +706,7 @@ DenseSet::IteratorBase DenseSet::GetRandomIterator() {
     return IteratorBase{};
 
   DensePtr* ptr = &*chain_it;
-  thread_local absl::InsecureBitGen bg{};
-  while (ptr->IsLink() && absl::Bernoulli(bg, 0.5)) {
+  while (ptr->IsLink() && absl::Bernoulli(tl_bit_gen, 0.5)) {
     DensePtr* next = ptr->Next();
     if (ExpireIfNeeded(ptr, next))  // stop if we break the chain with expiration
       break;


src/core/dense_set.cc
```diff
@@ -673,7 +673,28 @@ void DenseSet::Delete(DensePtr* prev, DensePtr* ptr) {
 }
 
 DenseSet::ChainVectorIterator DenseSet::GetRandomChain() {
-  size_t offset = absl::Uniform(absl::BitGen{}, 0u, entries_.size());
+  if (entries_.empty() || size_ == 0) {
+    return entries_.end();
+  }
+
+  // Use thread-local generator to avoid repeated construction overhead
+  thread_local absl::BitGen gen;
```
dranikpg: Maybe just using absl::InsecureBitGen is as cheap?

dranikpg (approved):
Clever. Please change the bit gen in `GetRandomIterator` as well

BorysTheDev:
"For sparse tables (6% util): ~87% success rate with 32 probes -> avoids expensive linear scan"
sounds really strange, why linear scan is more expensive than probing

romange:
Please open a separate issue implementing `DenseSet::Shrink`, as it is needed anyway for sorted sets, sets and hash maps

romange:
Clever indeed. 
$success = 1 - fail = 1- (sparsity)^{32}$

BorysTheDev:
If we have a uniform distribution and if we start from random point, we should get the same complexity for linear scan as for probing, but linear scan should be faster because of memory spatial locality

dranikpg:
> random probing can't solve this issue

With string hashes the distribution should be close to uniform. If to think about just theoretically for a single operation, probing one more cell with a linear scan shouldn't be different from doing random probes. _Except that if we always probe linarly_, then we create "islands" of values, as there is more blank space to land in and we'll only bite off the next leftmost value of the next "island". So maybe with linear probing a large hole is created that only grows on average... At least this is my theory. 

Any _full_ solution is just shrinking the backing array, as there's just not enough information for efficiently searching a few leftover entries in a large backing

BorysTheDev:
> > random probing can't solve this issue
> 
> With string hashes the distribution should be close to uniform. If to think about just theoretically for a single operation, probing one more cell with a linear scan shouldn't be different from doing random probes. _Except that if we always probe linarly_, then we create "islands" of values, as there is more blank space to land in and we'll only bite off the next leftmost value of the next "island". So maybe with linear probing a large hole is created that only grows on average... At least this is my theory.
> 
> Any _full_ solution is just shrinking the backing array, as there's just not enough information for efficiently searching a few leftover entries in a large backing

If we start linear scan from random point every time, there shouldn't be any islands and result should be the same as random probing. I think the problem is somewhere else, for example, our random generator could provide a non-uniform distribution or somehow we don't remove items in the beginning or something else. Anyway without understanding the real problem, mathematically this solution is incorrect

romange:
I think having 5000x improvement justifies the fix. Having said that, I agree that it would be nice to understand WHY the main branch is "5000x" slower.

BorysTheDev:
> I think having 5000x improvement justifies the fix. Having said that, I agree that it would be nice to understand WHY the main branch is "5000x" slower.

I don't say that there is no improvement. I say that the current fix isn't optimal and uses incorrect assumptions. And the results that we see have another explanation. I think the problem was fixed with some random change, but not with probing.

mkaruza:
Outside of  discussion about randomness.

@vyavdoshenko did you consider also rewriting function to use bit operations instead of modulo. i.e.

```
  size_t modulo = (1 << capacity_log_) - 1;
  for (size_t i = offset; i < entries_.size() + offset; i++) {
    auto it = entries_.begin() + (i & modulo);
```

src/core/dense_set.cc
vyavdoshenko: Fixed

vyavdoshenko:
The simpliest solution is the best.
There were 3 problems:
- Creating random generator for each call
- Using modulo operation, it can be avoided
- ExpireIfNeeded executed for empty objects, it can be avoided

src/core/string_set_test.cc
```diff
@@ -734,11 +734,10 @@ void BM_Spop1000(benchmark::State& state) {
   while (state.KeepRunning()) {
     state.PauseTiming();
     StringSet tmp;
-    src.Fill(&tmp);
-    src.Reserve(elems * sparseness);
+    tmp.Reserve(elems * sparseness);
     state.ResumeTiming();
     for (int i = 0; i < 1000; ++i) {
-      src.Pop();
+      tmp.Pop();
```
mkaruza, line 741: Please check that we always pop 1000 elements or that we always pop element - whatever is easier. Otherwise LGTM.

src/core/dense_set.cc
```diff
@@ -673,13 +673,31 @@ void DenseSet::Delete(DensePtr* prev, DensePtr* ptr) {
...
-    auto it = entries_.begin() + (i % entries_.size());
-    ExpireIfNeeded(nullptr, &*it);
-    if (!it->IsEmpty())
-      return it;
+  if (entries_.empty() || size_ == 0) {
+    return entries_.end();
   }
+
+  // Use thread-local generator to avoid construction overhead
+  thread_local absl::InsecureBitGen gen;
```
romange: let's use the same thread_local object here and at line 710

vyavdoshenko: Fixed

romange:
I remember seeing differrent numbers for the benchmark. While this improvement is good, it does not solve the problem for over sized sets.

src/core/dense_set.cc
```diff
@@ -27,6 +27,8 @@ constexpr size_t kMinSizeShift = 2;
 constexpr size_t kMinSize = 1 << kMinSizeShift;
 constexpr bool kAllowDisplacements = true;
 
+thread_local absl::InsecureBitGen tl_bit_gen;
```
dranikpg, line 30: Maybe creating the bitgen is not expensive

BorysTheDev (approved):

PR #6034, opened 2025-11-11
fix(takeover): pick correct replica to reconcile slots

kostasrim:
The changes in https://github.com/dragonflydb/dragonfly/pull/5621 allowed master to not shutdown after takeover and redirect cluster shard requests to the new promoted node. However, the logic broke when there were multiple registered replicas. For that case, dragonfly blindly picked the first replica in the topology to reconcile with.

* Use replica ID to figure out which node from the replica set is the one promoted
* Small refactor: remove bool parameter and split functions

### DIFF
--- a/src/server/cluster/cluster_family.cc
+++ b/src/server/cluster/cluster_family.cc
@@ -1098,19 +1098,31 @@ size_t ClusterFamily::MigrationsErrorsCount() const {
   return error_num;
 }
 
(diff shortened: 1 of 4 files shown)


BorysTheDev (approved):

BorysTheDev (approved):

romange:
Can we merge it @kostasrim ?

PR #6040, opened 2025-11-11
chore(tiering): Add throttling metric + more tests

dranikpg:
1. Add `tiered_throttled_clients` indicating how many futures we keep for backpressure (the clients might have disconnected, I  decided to keep the name simple)
2. Add parameterized tests that run most of the tests without experimental cooling. Cooling is a feature in disguise as it prevents most operations to be carried out with true disk values... and as a result our tests don't test many things at all! I parameterized some of the tests, so they work with both cooling enabled and disabled

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -67,6 +67,11 @@ jobs:
     container:
       image: ghcr.io/romange/${{ matrix.container }}
       options: --security-opt seccomp=unconfined --sysctl "net.ipv6.conf.all.disable_ipv6=0"
+      # Some tests which launch their own containers need a mounted volume to write through files
+      # into child containers
+      volumes:
+        - /mnt:/mnt
+
     steps:
     - uses: actions/checkout@v5
       with:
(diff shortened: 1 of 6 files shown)


dranikpg:
Should fail because it uncovers #6031

src/server/common.cc
```diff
@@ -186,6 +186,8 @@ TieredStats& TieredStats::operator+=(const TieredStats& o) {
   ADD(cold_storage_bytes);
   ADD(total_offloading_steps);
   ADD(total_offloading_stashes);
+
+  ADD(throttled_clients);
```
romange: From my past experience I find such metrics widely unstable as they probe behaviour .
If say we probe metrics at intervals of 5s we could measure 0 at both probes while during the interval have moments with 100 clients being stalled. 

The more stable approach is to use aggregated events and then use `rate` functions to take derivatives

BorysTheDev (approved):

dranikpg:
Added two metrics
* tiered_clients_throttled - instantaneous metric of active backpressure, useful for debugging in-moment situations
* tiered_total_clients_throttled - total amount of backpressure provided, can be used to see how much commands are throttled

romange (approved):

PR #6042, opened 2025-11-11
feat(helm): adds common label support to the helm chart

### LINKED ISSUE #6041: Add commonLabels support to Helm Chart
Yes

Currently, there's no way to apply custom labels across all resources created by the Dragonfly Helm chart. Users often need to add organizational labels (like owner, team, cost-center, etc.) to all Kubernetes resources for resource management, cost tracking, and governance purposes.

Add a new `commonLabels` field in the `values.yaml` of the helm chart. Update the `_helpers.tpl` and the YAML files accordingly to make use of the `commonLabels`. Values defined under this will add labels to all the kubernetes resources.

gaganhr94:
Fixes #6041 

This PR adds commonLabels support to the Helm chart, allowing users to specify custom labels that are automatically applied across all Kubernetes resources (Deployment, StatefulSet, etc.) and pod templates. The implementation adds `commonLabels` field to `values.yaml`, updates the `dragonfly.labels` helper in `_helpers.tpl` to merge these labels, and modifies both `deployment.yaml` and `statefulset.yaml` to include commonLabels in pod template metadata.

The change follows patterns used by major Helm charts like Bitnami, and includes a full test case with golden test output to ensure correctness of the change made.

### DIFF
--- a/contrib/charts/dragonfly/templates/_helpers.tpl
+++ b/contrib/charts/dragonfly/templates/_helpers.tpl
@@ -40,6 +40,18 @@ helm.sh/chart: {{ include "dragonfly.chart" . }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- include "dragonfly.commonLabels" . }}
+{{- end }}
+
+{{/*
+User-defined common labels
+*/}}
+{{- define "dragonfly.commonLabels" -}}
+{{- if .Values.commonLabels }}
+{{- range $key, $value := .Values.commonLabels }}
+{{ $key }}: {{ $value }}
+{{- end }}
+{{- end }}
 {{- end }}
 
 {{/*
--- a/contrib/charts/dragonfly/templates/deployment.yaml
+++ b/contrib/charts/dragonfly/templates/deployment.yaml
@@ -22,6 +22,9 @@ spec:
         {{- end }}
       labels:
         {{- include "dragonfly.selectorLabels" . | nindent 8 }}
+        {{- if .Values.commonLabels }}
+        {{- include "dragonfly.commonLabels" . | trim | nindent 8 }}
+        {{- end }}
     spec:
       {{- include "dragonfly.pod" . | trim | nindent 6 }}
 {{- end }}
--- a/contrib/charts/dragonfly/templates/statefulset.yaml
+++ b/contrib/charts/dragonfly/templates/statefulset.yaml
@@ -23,6 +23,9 @@ spec:
         {{- end }}
       labels:
         {{- include "dragonfly.selectorLabels" . | nindent 8 }}
+        {{- if .Values.commonLabels }}
+        {{- include "dragonfly.commonLabels" . | trim | nindent 8 }}
+        {{- end }}
     spec:
       {{- include "dragonfly.pod" . | trim | nindent 6 }}
   volumeClaimTemplates:
--- a/contrib/charts/dragonfly/README.md
+++ b/contrib/charts/dragonfly/README.md
@@ -32,6 +32,7 @@ helm upgrade --install dragonfly oci://ghcr.io/dragonflydb/dragonfly/helm/dragon
 |-----|------|---------|-------------|
 | affinity | object | `{}` | Affinity for pod assignment |
 | command | list | `[]` | Allow overriding the container's command |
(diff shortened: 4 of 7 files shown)


gaganhr94:
Hi @romange @Abhra303, can you please have a look into this PR and give your review on it. Happy to make any changes, if needed.

contrib/charts/dragonfly/templates/_helpers.tpl
```diff
@@ -40,6 +40,9 @@ helm.sh/chart: {{ include "dragonfly.chart" . }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- if .Values.commonLabels }}
+{{ toYaml .Values.commonLabels }}
+{{- end }}
```
Abhra303, line 54: I don't think this would work. It will have indentation issues. Instead define dragonfly.commonLabels and include it here.


```suggestion
{{- include "dragonfly.commonLabels" . }}
{{- end }}

{{/*
User-defined common labels
*/}}
{{- define "dragonfly.commonLabels" -}}
{{- if .Values.commonLabels }}
{{- range $key, $value := .Values.commonLabels }}
{{ $key }}: {{ $value | quote }}
{{- end }}
{{- end }}
{{- end }}
```

contrib/charts/dragonfly/templates/deployment.yaml
```diff
@@ -22,6 +22,9 @@ spec:
         {{- end }}
       labels:
         {{- include "dragonfly.selectorLabels" . | nindent 8 }}
+        {{- with .Values.commonLabels }}
+        {{- toYaml . | nindent 8 }}
```
Abhra303: use "dragonfly.commonLabels" here after you define it.

contrib/charts/dragonfly/templates/statefulset.yaml
```diff
@@ -23,6 +23,9 @@ spec:
         {{- end }}
       labels:
         {{- include "dragonfly.selectorLabels" . | nindent 8 }}
+        {{- with .Values.commonLabels }}
+        {{- toYaml . | nindent 8 }}
```
Abhra303: use dragonfly.commonLabels

Abhra303 (changes requested):
Thanks for the PR! Its a good feature to have. See comments.

contrib/charts/dragonfly/templates/_helpers.tpl
gaganhr94: I have made these changes

contrib/charts/dragonfly/templates/statefulset.yaml
gaganhr94: Done

contrib/charts/dragonfly/templates/deployment.yaml
gaganhr94: Done

gaganhr94:
Thanks for the review @Abhra303. I have made the changes as per your suggestion. One thing to note in this PR, I had to include a if condition check in `deployment.yaml` and `statefulset.yaml`, since not having that condition added an empty new line for all the cases where the `commonLabels` were not defined.

Abhra303 (approved):

gaganhr94:
Hi @romange, can you please look into the PR whenever you get a chance. It would be great if you could get it merged. Thanks !

romange:
Thanks!

romange:
@gaganhr94  it seems the PR broke our tests. I will have to revert it @Abhra303  also please advise:
https://github.com/dragonflydb/dragonfly/actions/runs/19495747159/job/55797412070

gaganhr94:
@romange @Abhra303  I went through the test failure. It looks like the chart version was bumped from v1.34.2 to v1.35.0 a few days back. The CI tests ran on this PR before that. So they had passed. The rebase took care of most of the changes to the test files (would have changed the from v1.34.2 to v1.35.0). There was one test file which I had added for these changes `commonlabels-values.golden.yaml`, which would still have the older value of the version (i.e, v1.34.2) while the PR was merged. As a result of this, the lint test for the chart failed, since it was expecting the newer version (i.e, v1.35.0)

Changing the version to v1.35.0 in commonlabels-values.golden.yaml would resolve this issue. I'll raise a PR for the change with the latest version of the chart.

Issue #6047, opened 2025-11-12
`test_metrics_sanity_check` fails in release action


abhijat:
```
E       docker.errors.APIError: 400 Client Error for http+docker://localhost/v1.48/containers/46551a6e4a0bcc213ed41ab5b121df615b079b65cd00ba3d5de711fe1d6afa6e/start: Bad Request ("failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/mnt/tmpli8lqujx" to rootfs at "/etc/prometheus/prometheus.yml": create mountpoint for /etc/prometheus/prometheus.yml mount: cannot create subdirectories in "/var/lib/docker/overlay2/809d030d16a62609ebb672f90d7130b9584a81a7eda37173b40230de425c3b7a/merged/etc/prometheus/prometheus.yml": not a directory: unknown: Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the expected type")
```

It seems the parent path used in the test is not mounted correctly

https://github.com/dragonflydb/dragonfly/actions/runs/19291349876/job/55162561262#step:6:1423

abhijat:
This test creates a container within another container. One of the problems with mounting files into the new container is that the parent container must write to the host filesystem, and then the child container must read from that same path.

For this to happen reliably, the parent container must have a volume mounted, write the file to it, and then mount the file into the child container. We use `/var/crash` and `/mnt` testing each in succession. In the failing test it seems `/mnt` is selected but is not a valid path to mount files from.

PR #6048, opened 2025-11-12
actions: Add a volume mount to container

abhijat:
Some tests launch their own docker containers (using testcontainers lib). 

These tests also may need to pass through files to child containers which will be mounted. 

Since containers can only share files via the host filesystem when mounting said files, a volume is added which can be used for such operations.

Volume mount is already available in CI but was missing in the release action.

FIXES https://github.com/dragonflydb/dragonfly/issues/6047

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -67,6 +67,11 @@ jobs:
     container:
       image: ghcr.io/romange/${{ matrix.container }}
       options: --security-opt seccomp=unconfined --sysctl "net.ipv6.conf.all.disable_ipv6=0"
+      # Some tests which launch their own containers need a mounted volume to write through files
+      # into child containers
+      volumes:
+        - /mnt:/mnt
+
     steps:
     - uses: actions/checkout@v5
       with:


romange:
@abhijat  it fails

abhijat:
> @abhijat it fails

a unit test failed, but might be related due to the mount, will investigate

abhijat:
The error is in the CI action which is not changed in this PR, only the release action is changed here, so the test failure cannot be related to the PR

abhijat:
created https://github.com/dragonflydb/dragonfly/issues/6051

abhijat:
`lint-test-chart` failed with

```
        	            	error while running command: exit status 1; Error: chart requires kubeVersion: >=1.23.0-0 which is incompatible with Kubernetes v1.20.0
```

abhijat:
> `lint-test-chart` failed with
> 

It passed on rerun, retrying all tests run to be sure. Seems like a flaky test, this change had nothing to do with k8s version.

kostasrim (approved):
I trust you with the changes -- plz run them before you merge for testing

abhijat:
> I trust you with the changes -- plz run them before you merge for testing

it can only run on the release workflow, but it shouldn't be a problem, we just mount `/mnt` from host to the release container

PR #6049, opened 2025-11-12
fix(rdb_load): Store streamed items per rdb lodaer

dranikpg:
Huge values that are streamed are stored not in the dashtable, but in the `now_streamed` hashmap. It was a thread local variable that was cleaned on error paths. However it wasnt cleaned if the replication was just cancelled - so values from the last load would remain inside the set. Whenever replication ran again, it would fail on duplicate entries.

This PR makes `now_streamed` an rdb loader member that is destroyed after loading.

### DIFF
--- a/tests/dragonfly/replication_test.py
+++ b/tests/dragonfly/replication_test.py
@@ -3074,7 +3074,7 @@ async def test_big_huge_streaming_restart(df_factory: DflyInstanceFactory):
         await asyncio.sleep(random.random() + 0.5)
 
     # Wait for it to finish finally
-    async with async_timeout.timeout(10):
+    async with async_timeout.timeout(60):
         await wait_for_replicas_state(c_replica)
 
     # Check that everything is in sync


romange:
Please rewrite your PR description @dranikpg

tests/dragonfly/replication_test.py
```diff
@@ -3052,6 +3052,41 @@ async def test_bug_in_json_memory_tracking(df_factory: DflyInstanceFactory):
...
+        await wait_for_replicas_state(c_replica)
+
+    # Check that everything is in sync
+    hashes = await asyncio.gather(*(SeederV2.capture(c) for c in [c_master, c_replica]))
+    assert len(set(hashes)) == 1
+
+    # No in-between errors occured
+    replica.stop()
+    lines = replica.find_in_logs("Duplicate zset fields detected")
+    assert len(lines) == 0
```
dranikpg, line 3087: I don't like this, its weak. Replication eventually succeeds with the bug as well, just on a second try. Maybe I can assert that it ran only two times 🤔

dranikpg: We actually don't have a metric that indicates how many replication attempts were made

dranikpg:
Increased test timeout

PR #6053, opened 2025-11-13
fix(core): Account for size change on json object defragment

abhijat:
For JSON type, after defragmentation used size may change. The change is propagated to the db stats object, so that metrics observe the change and reduction in memory usage (if any) correctly.

The JSON object records the change by observing the diff in bytes used by the memory resource, this is the same pattern used in `json_family.cc` etc. Once the change is determined it is set in the `bytes_used` field for the json object.

The delta update is in effect for all objects whose size changes, not just JSON type objects.

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -1633,11 +1633,25 @@ MemoryResource* CompactObj::memory_resource() {
 }
 
 bool CompactObj::JsonConsT::DefragIfNeeded(PageUsage* page_usage) {
-  if (JsonType* old = json_ptr; ShouldDefragment(page_usage)) {
+  if (ShouldDefragment(page_usage)) {
+    const MiMemoryResource* mr = static_cast<MiMemoryResource*>(memory_resource());
+
+    const int64_t before = static_cast<int64_t>(mr->used());
+    DCHECK_GE(before, 0) << "Memory usage is more than int64_t max value";
+
+    JsonType* old = json_ptr;
     json_ptr = AllocateMR<JsonType>(DeepCopyJSON(old));
     DeleteMR<JsonType>(old);
+
+    const int64_t after = static_cast<int64_t>(mr->used());
+    DCHECK_GE(after, 0) << "Memory usage is more than int64_t max value";
+
+    if (const int64_t delta = after - before; delta != 0) {
+      bytes_used = UpdateSize(bytes_used, delta);
+    }
     return true;
   }
+
   return false;
 }
 
--- a/src/server/engine_shard.cc
+++ b/src/server/engine_shard.cc
@@ -351,14 +351,20 @@ std::optional<CollectedPageStats> EngineShard::DoDefrag(CollectPageStats collect
   uint64_t attempts = 0;
 
   PageUsage page_usage{collect_page_stats, threshold};
+  DbTable* db_table = slice.GetDBTable(defrag_state_.dbid);
   do {
     cur = prime_table->Traverse(cur, [&](PrimeIterator it) {
       // for each value check whether we should move it because it
       // seats on underutilized page of memory, and if so, do it.
-      bool did = it->second.DefragIfNeeded(&page_usage);
+      const ssize_t original_size = it->second.MallocUsed();
+      const bool did = it->second.DefragIfNeeded(&page_usage);
       attempts++;
       if (did) {
         reallocations++;
+        if (const ssize_t delta = it->second.MallocUsed() - original_size;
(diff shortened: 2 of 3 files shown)


src/core/page_usage_stats.h
```diff
@@ -93,6 +101,8 @@ class PageUsage {
 
   // For use in testing, forces reallocate check to always return true
   bool force_reallocate_{false};
+
+  TypeAndDelta last_delta_;
```
abhijat: It may be cleaner to change the return type for `DefragIfNeeded` so that it returns the object type, the delta and the fact whether defrag was conducted instead of carrying this state around, but then the signature will need to be changed in a larger set of methods, most of which do not change in size on defragmentation.

abhijat: this is no longer needed as object's `bytes_used` field is used to calculate the change

src/core/compact_object.cc
```diff
@@ -1633,11 +1633,18 @@ MemoryResource* CompactObj::memory_resource() {
...
 bool CompactObj::JsonConsT::DefragIfNeeded(PageUsage* page_usage) {
-  if (JsonType* old = json_ptr; ShouldDefragment(page_usage)) {
+  JsonType* old = json_ptr;
+  if (ShouldDefragment(page_usage)) {
+    const MiMemoryResource* mr = static_cast<MiMemoryResource*>(memory_resource());
+    const ssize_t before = mr->used();
     json_ptr = AllocateMR<JsonType>(DeepCopyJSON(old));
     DeleteMR<JsonType>(old);
+    if (const ssize_t delta = mr->used() - before; delta != 0) {
+      bytes_used += delta;
```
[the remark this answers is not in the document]
abhijat: it is intentionally `ssize_t`, delta is expected to be negative. It should not lead to a wraparound though as delta is only caused by reallocation.

abhijat: there is a helper function `UpdateSize`, maybe it should be used here. Mainly it protects against object size going down to negative. I'm not clear when that will happen though.

abhijat: I changed to use `UpdateSize`, it is more defensive.

abhijat:
augment review

src/core/compact_object.cc
```diff
@@ -1633,11 +1633,20 @@ MemoryResource* CompactObj::memory_resource() {
...
 
 bool CompactObj::JsonConsT::DefragIfNeeded(PageUsage* page_usage) {
-  if (JsonType* old = json_ptr; ShouldDefragment(page_usage)) {
+  if (ShouldDefragment(page_usage)) {
+    const MiMemoryResource* mr = static_cast<MiMemoryResource*>(memory_resource());
+
+    const ssize_t before = mr->used();
+    JsonType* old = json_ptr;
     json_ptr = AllocateMR<JsonType>(DeepCopyJSON(old));
     DeleteMR<JsonType>(old);
```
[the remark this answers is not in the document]
abhijat: fixed to warn on overflowing bounds in debug build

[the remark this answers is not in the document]
abhijat: TBH to overflow `int64_t` max, the memory used would have to be several thousand petabytes.

src/server/engine_shard.cc
```diff
@@ -351,14 +351,20 @@ std::optional<CollectedPageStats> EngineShard::DoDefrag(CollectPageStats collect
...
       // for each value check whether we should move it because it
       // seats on underutilized page of memory, and if so, do it.
-      bool did = it->second.DefragIfNeeded(&page_usage);
+      const ssize_t original_size = it->second.MallocUsed();
+      const bool did = it->second.DefragIfNeeded(&page_usage);
       attempts++;
       if (did) {
         reallocations++;
+        if (const ssize_t delta = it->second.MallocUsed() - original_size;
+            delta != 0 && db_table != nullptr) {
```
romange, line 365: can `db_table` be nullptr?

abhijat: probably not but I added it to be defensive as we access a field via the pointer.

romange: i usually do not like these things as it raises questions where this state is possible. it's better to DCHECK to assert the precondition than have such ifs

abhijat: I see, I will remove it in a follow up

abhijat: this is fixed in a commit in https://github.com/dragonflydb/dragonfly/pull/6262 which touches this code

Issue #5986, opened 2025-10-30
When the value of json. set is relatively large, jedis returns timeout


chao6118:
Describe the bug
When storing large JSON values using jedisPooled.jsonSet, a timeout exception occurs. The exception type is JedisConnectionException with root cause SocketTimeoutException: Read timed out.

To Reproduce

Prepare a large JSON value (e.g., containing extensive nested data or large strings)

Use JedisPooled instance to call jsonSet method to insert the record

Wait for the operation to complete

See timeout exception after some time

Expected behavior
Should be able to store large JSON values normally without timeout exceptions.

Screenshots
N/A

Environment (please complete the following information):

OS: Ubuntu 20.04

Kernel: # Command: uname -a (output not provided)

Containerized?: Docker

Dragonfly Version: v1.34.2

public class DD {

    public static void main(String s[]) {
        JedisPooled jedisPooled = null;
        try {
            int dbIndex = 0;

            ConnectionPoolConfig connectionPoolConfig = new ConnectionPoolConfig();
            //最大连接
            connectionPoolConfig.setMaxTotal(4);
            //最大空闲
            connectionPoolConfig.setMaxIdle(4);
            //最小空闲
            connectionPoolConfig.setMinIdle(4);

[...]

PR #6060, opened 2025-11-16
feat(JSON): Use stateless allocator

abhijat:
A stateless allocator is used to parse JSON from string. The allocator forwards requests to a thread local memory resource pointer which is backed by the mimalloc heap. The memory resource is initialized during engine shard init, same as the `CompactObject` thread local initialization.

There are two new restrictions that are now in effect when using this stateless allocator:

* Objects created on one shard should not be destroyed on another shard, and as a result should be moved/copied to other shard. This is because each shard has its own memory resource, and it is possible that the destination shard does not have a mimalloc heap initialized, as is the case with coordinator thread in our test suite.
* Objects which use the std allocator based parser introduced in https://github.com/dragonflydb/dragonfly/pull/6061 are not interchangeable with objects allocated on mimalloc heap, their types are different and their allocators are different. If conversion between two types is required, the objects have to be serialized from source type into a string, and then deserialized into the destination type.

fixes https://github.com/dragonflydb/dragonfly/issues/5986

### DIFF
--- a/src/core/json/json_object.h
+++ b/src/core/json/json_object.h
@@ -46,10 +46,10 @@ JsonType DeepCopyJSON(const JsonType* j);
 template <typename Json = JsonType>
 auto MakeJsonPathExpr(std::string_view path, std::error_code& ec)
     -> jsoncons::jsonpath::jsonpath_expression<Json> {
-  using result_allocator_t = typename Json::allocator_type;
-  using temp_allocator_t = std::allocator<char>;
-  using allocator_set_t = jsoncons::allocator_set<result_allocator_t, temp_allocator_t>;
-  return jsoncons::jsonpath::make_expression<Json, temp_allocator_t>(allocator_set_t(), path, ec);
+  using ResultAllocT = typename Json::allocator_type;
+  using TmpAllocT = std::allocator<char>;
+  using AllocSetT = jsoncons::allocator_set<ResultAllocT, TmpAllocT>;
+  return jsoncons::jsonpath::make_expression<Json, TmpAllocT>(AllocSetT(), path, ec);
 }
 
 }  // namespace dfly


abhijat:
> This is removed and now all parse operations use mimalloc.

A problem with this approach is that not all cores are guaranteed to have called `EngineShard::InitThreadLocal` and have a memory resource using the mimalloc heap initialized. 

This is seen in the failing tests. If the `num_shards` flag is set to less than the number of cores, some cores will not have a thread local heap for parsing JSON.

There are places in code which just need to parse some string into json without relying on mimalloc heap (eg in cluster config). These places use the default memory resource. They need to use the same allocator which uses the mimalloc heap.

abhijat:
> There are places in code which just need to parse some string into json without relying on mimalloc heap (eg in cluster config). These places use the default memory resource. They need to use the same allocator which uses the mimalloc heap.

The right thing to do here seems to be to defer all parsing of JSON objects until we are on the shard, in the tx callback. At present some parsing is done on coordinator thread to return errors early, and some on the shard where the key resides.

The other uses of `JsonFromString` such as cluster config module can use the default pmr resource. As they do not interact  with compact obj. and related heap.

abhijat:
will be rebased after https://github.com/dragonflydb/dragonfly/pull/6061, which will introduce two paths for parsing JSON values:

* parse on any shard using default PMR resource, not intended to use as compact object OR
* parse on transaction destination shard using mimalloc heap, intended to use as compact object in db

romange:
@abhijat I am not sure I get it. Why can not we use default resource in coordinator thread together with JSONFromString, and use ShardJsonFromString in shard threads together with thread local heap?

romange:
Ok, I understand now: 
```
it is possible to work around this by using two allocator types, one for the default heap and one for mimalloc heap. but since the allocator type is part of the json object type, going down this path means that we now have two non interchangeable json object types
```

but where do they meet? i.e. do we pass a "coordinator" json objects to shard threads? in other words, even today I would expect that those objects would be "non-interchangeable", as we must allocate shard local objects only with shard local heap, so in a way having two disjoin  c++ types would make things more explicit and clear.

abhijat:
> but where do they meet? i.e. do we pass a "coordinator" json objects to shard threads? in other words, even today I would expect that those objects would be "non-interchangeable", as we must allocate shard local objects only with shard local heap,

There are a few places where we allocate on coordinator using the default memory resource and pass to the destination shard which uses mimalloc based allocation, and then mix those objects. I think it is done to facilitate early error reporting, but I have changed those call sites in https://github.com/dragonflydb/dragonfly/pull/6061 deferring json parsing to the destination shard. 

In fact I found this comment https://github.com/dragonflydb/dragonfly/blob/main/src/server/json_family.cc#L1905-L1907 which mentions this same problem. 

>  so in a way having two disjoin c++ types would make things more explicit and clear.

Yes, I added a new type in that PR called `ShortLivedJSON` (name is a bit confusing but not sure what else to name it) - which uses the standard allocator, in fact it is an alias for the default json type of jsoncons. It is now used in most places where we do json parsing with `std::pmr::get_default_resource()` 

After that parent PR is merged, the other `JSONType` which we use everywhere in the code can be changed to use stateless allocator, and the two types will be incompatible and never expected to work with each other.

romange:
Great!

src/server/json_family.cc
```diff
@@ -1526,10 +1525,18 @@ auto OpMemory(const OpArgs& op_args, string_view key, const WrappedJsonPath& jso
...
+// Returns json vector that represents the result of the json query. A shard local
+// heap allocated JSON cannot be copied and then destroyed on another shard because we use stateless
+// allocators which forward all requests to thread local memory resource. So the value is first
+// copied to the std allocator-backed type ShortLivedJSON.
+OpResult<JsonCallbackResult<ShortLivedJSON>> OpResp(const OpArgs& op_args, string_view key,
+                                                    const WrappedJsonPath& json_path) {
+  auto cb = [](const string_view&, const JsonType& val) {
+    string s;
+    val.dump(s);
+    return JsonFromString(s);
```
abhijat, line 1538: This is unfortunate side effect of not being able to send `JsonType` object to the coordinator thread, it can only be destroyed on the shard that created it. We have to copy it into a safe type before sending it to coordinator.

abhijat:
`test_index_persistence` fails because of the following change in rdb loader:

```
size_t start_size = static_cast<MiMemoryResource*>(CompactObj::memory_resource())->used();
    {
      if (auto json = ParseJsonUsingShardHeap(blob)) {
        pv_->SetJson(std::move(*json));
```

~~Although `CompactObj::memory_resource()` is used, the memory resource in question is not mimalloc backed but `get_default_resource()`. So the assumption that JSON can be parsed using mimalloc based heap at this point in the code path is not true.~~

~~A global heap backed json can be constructed here, but then it cannot be passed to `SetJson`.~~

This error was due to a command invoked during rdb load and not the above code path. 

The command invoked simply does validation using a json path expression, and it is changed to use a short lived json type as the backing type for the expression

abhijat:
I think this is in good condition now once tests pass. I would like to rename `JsonType` to something like `PrimeValueJson` or `PvJson` to make it clear that it is intended for use as prime value but that can be a separate PR as it will touch a lot of files.

src/server/json_family.cc
abhijat: Note: mimalloc will deallocate the object from another thread, but the coordinator thread might not have mimalloc heap set up on it.

src/core/json/json_object.h
```diff
@@ -4,6 +4,8 @@
 
 #pragma once
 
+#include <glog/logging.h>
```
romange: please remove this

src/core/json/json_object.h
```diff
@@ -21,32 +23,82 @@
...
+  using value_type = T;
+  using size_type = std::size_t;
+  using difference_type = std::ptrdiff_t;
+  using is_always_equal = std::true_type;
+
+  template <typename U> StatelessJsonAllocator(const StatelessJsonAllocator<U>&) noexcept {
+  }
+
+  StatelessJsonAllocator() noexcept {
+    DCHECK_NE(tl_mr, nullptr) << "json allocator created without backing memory resource";
```
romange: just use `assert(tl_mr);`

romange: or just remove the DCHECK

src/core/json/json_object.h
```diff
@@ -21,32 +23,82 @@
 
 namespace dfly {
 
-using ShortLivedJSON = jsoncons::json;
-using JsonType = jsoncons::pmr::json;
+namespace detail {
+
+inline thread_local PMR_NS::memory_resource* tl_mr = nullptr;
```
romange: why using another pointer?
why not use `CompactObject::memory_resource()`?

src/core/json/json_object.h
```diff
@@ -21,32 +23,82 @@
...
+
+template <typename T> class StatelessJsonAllocator {
```
romange: There is nothing here json specific. I would factor out this class into 
`dragonfly/src/core/detail/stateless_mr_allocator.h` also would change the name to `StatelessMrAllocator` and explain that this allocator allocates using `CompactObj::memory_resource()`

src/core/json/json_object.h
abhijat: It is mainly because of how the build is set up, when I started on the PR I wanted to use `CompactObj` memory resource, but the library built here which is `jsonpath` is  mostly a self contained structure whereas compact object is in `dfly_core` which  depends on `jsonpath`.

src/core/json/json_object.h
abhijat: changed to use `assert`, I kept it here as it caught some bugs when running existing tests

src/core/json/json_object.h
romange: see my other comment. I do not think the allocator is json specific, so we can move it to be under core/detail.
hope this solves the dependency issue

src/core/json/json_object.h
abhijat: I will try this out. It still would not be able to use the `CompactObj::memory_resource()` directly I think, but the intention can be added in a comment.

src/core/json/jsonpath_test.cc
```diff
@@ -39,7 +39,7 @@ class TestDriver : public Driver {
 template <typename JSON> JSON ValidJson(string_view str);
 
 template <> JsonType ValidJson<JsonType>(string_view str) {
-  auto res = ::dfly::JsonFromString(str, pmr::get_default_resource());
+  auto res = ParseJsonUsingShardHeap(str);
```
romange, line 42: I do not think this test should rely on `JsonType`, it should use TmpJson.

src/core/json/json_object.h
```diff
@@ -21,32 +21,82 @@
...
+bool operator!=(const StatelessJsonAllocator<T>&, const StatelessJsonAllocator<U>&) noexcept {
+  return false;
+}
+
+}  // namespace detail
+
+void InitTLJsonHeap(PMR_NS::memory_resource* mr);
+
+using TmpJson = jsoncons::json;
+using JsonType =
```
romange: the circular dependency is created because `JsonType` is declared here together with `ParseJsonUsingShardHeap` and `DeepCopyJSON`.

but if you move this to dfly_core level (can be in compact_object.h/cc for what I care) then they can rely on `StatelessJsonAllocator` that uses CompactObject::resource directly

romange:
https://github.com/dragonflydb/dragonfly/issues/2313#issuecomment-3543021057 
This is why we need this allocator to be generic - it will help us with more use-cases involving pmr data types.

src/core/json/json_object.h
abhijat: It might be hard to move `JsonType` out of `jsonpath` and down into core, it is used everywhere in `jsonpath` library, for example `PathSegment`, `AggFunction` etc use it, the `dfs` related code uses it. 

IMO the hierarchy of type definitions has to be allocator -> json type + utilities using the type -> core

I will try to move out the allocator to a top level definition

src/core/json/json_object.h
abhijat: removed

src/core/json/json_object.h
romange: I see. then indeed - we can not rely on compact_object.h and the independent pointer `tl_mr` is required 👍🏼

abhijat:
> There are many changes that can be done without actually changing the functionality. For example, the rename of `ShortLivedJSON` to `TmpJson` or making `MakeJsonPathExpr` template. I would divide this PR into the one making syntactic changes, without changing the functionality and the other one more succinct that switches to the new allocator.

created https://github.com/dragonflydb/dragonfly/pull/6111 to move out two change sets, once that one is merged this can be rebased

src/core/json/jsonpath_test.cc
abhijat: I tried to change it but the type returned from here is passed through to methods in `path.h/.cc` like `EvaluatePath`, `MutatePath`, `DeletePath` etc which all take `JsonType`.

abhijat:
@romange please take a look

PR #6061, opened 2025-11-16
server: Parse JSON on destination shard

abhijat:
In some places in `json_family` strings were parsed into JSON while using the default memory resource for allocation, and those objects were passed to another shard where these objects were used to modify other JSON objects allocated on the mimalloc backed heap.

This mix of allocators is problematic and blocks using a stateless allocator.

With a stateless allocator, it will not be possible to safely parse a string view on coordinator and send it to destination shard. Because a stateless allocator will simply forward requests to a static thread local memory resource, and stateless allocators should be equal and interchangeable, two stateless allocators forwarding requests to two different memory resources: one `PMR_NS::get_default_resource()` and the other `CompactObj::memory_resource()` are unsafe if used interchangeably.

Two different allocators can also not be used (one for default resource and another for mimalloc) because the allocator type is part of the json type, which results in the object types being incompatible with each other even if intended to be used together.

To fix this, where possible JSON parsing in `json_family` is moved from coordinator to transaction destination shard to make sure that shard's thread local memory resource backed by mimalloc heap is used to allocate during the parse operation. Error handling is delayed until the transaction hop finishes.

A new type called `ShortLivedJSON` is introduced. Allocation for this type is backed by `std` heap. These objects are expected to be used for short operations and not intended for being held as values in the dash table. This new type is incompatible with the default, mimalloc backed `JSONType` used nearly everywhere. This is by design and surfaces problematic intermixing of objects belonging to two different allocators.

### DIFF
--- a/src/core/json/json_object.cc
+++ b/src/core/json/json_object.cc
@@ -7,13 +7,12 @@
 #include "base/logging.h"
 
 using namespace jsoncons;
-using namespace std;
 
 namespace {
 
 template <typename T>
 std::optional<T> ParseWithDecoder(std::string_view input, json_decoder<T>&& decoder) {
-  error_code ec;
+  std::error_code ec;
   auto JsonErrorHandler = [](json_errc ec, const ser_context&) {
     VLOG(1) << "Error while decode JSON: " << make_error_code(ec).message();
     return false;
@@ -29,7 +28,7 @@ std::optional<T> ParseWithDecoder(std::string_view input, json_decoder<T>&& deco
 
   /* The maximum possible JSON nesting depth is either the specified json_nesting_depth_limit or
      half of the input size. Since nesting a JSON object requires at least 2 characters. */
-  auto parser_options = jsoncons::json_options{}.max_nesting_depth(
+  auto parser_options = json_options{}.max_nesting_depth(
       std::min(json_nesting_depth_limit, uint32_t(input.size() / 2)));
 
   json_parser parser(parser_options, JsonErrorHandler);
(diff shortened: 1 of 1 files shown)


BorysTheDev (dismissed):

src/core/json/json_object.cc
```diff
@@ -43,6 +43,18 @@ optional<JsonType> JsonFromString(string_view input, PMR_NS::memory_resource* mr
...
+
+namespace dfly {
+
+std::optional<ShortLivedJSON> JsonFromString(std::string_view input) {
+  return ParseWithDecoder(input, json_decoder<ShortLivedJSON>{});
+}
+
+optional<JsonType> JsonFromString(string_view input, PMR_NS::memory_resource* mr) {
+  return ParseWithDecoder(input, json_decoder<JsonType>{std::pmr::polymorphic_allocator<char>{mr}});
+}
```
vyavdoshenko, line 55: nit
use std:: or without std:: but not a mix

abhijat: Maybe best to use `PMR_NS` for consistency. I removed `using namespace std` and made the namespace explicit in the file.

vyavdoshenko:
lgtm

I am just curious, how `json.debug memory` will work in this case? Will ShortLivedJSON have zero bytes?

vyavdoshenko (approved):

abhijat:
> I am just curious, how `json.debug memory` will work in this case? Will ShortLivedJSON have zero bytes?

Basically the `ShortLivedJSON` type is only for use cases where we don't store things into the table but need JSON for some reason, for example validating cluster config which is JSON formatted, or printing JSON values using reply builder. 

This type won't be stored into the table, so it should never show up in `json.debug memory` output. `JSONType` will be allocated using mimalloc heap and is intended to be stored in the db table.

PR #6062, opened 2025-11-16
fix: Remove index docs on eviction and cluster migration

mkaruza:
When we evict keys or in cluster migration we should also remvoe index docs. 
Use `DbSlice::Del` which removes index docs and calls after `PerformDeletion`.

### DIFF
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -851,6 +851,7 @@ void DbSlice::FlushSlotsFb(const cluster::SlotSet& slot_ids) {
       std::string_view key = it->first.GetSlice(&tmp);
       SlotId sid = KeySlot(key);
       if (slot_ids.Contains(sid) && it.GetVersion() < next_version) {
+        // We use copy of table smart pointer and pass it as table because FLLUSHALL can drop table.
         Del(db_cntx, Iterator::FromPrime(it), table.get());
         ++del_count;
       }


src/server/db_slice.cc
```diff
@@ -1476,7 +1476,8 @@ pair<uint64_t, size_t> DbSlice::FreeMemWithEvictionStepAtomic(DbIndex db_ind,
 
         evicted_bytes += evict_it->first.MallocUsed() + evict_it->second.MallocUsed();
         ++evicted_items;
-        PerformDeletion(Iterator(evict_it, StringOrView::FromView(key)), db_table.get());
+
+        Del(cntx, Iterator(evict_it, StringOrView::FromView(key)));
```
romange, line 1492: what about `PrimeEvictionPolicy::Evict` ?

mkaruza: Added. Also in `FlushSlotsFb`

romange: I saw FlushSlotsFb. Have you verified that we indeed have a bug there? i.e. we do not remove these documents afterwards?

mkaruza: Looked into path where function is called, and it doesn't look to me that we also remove index docs.

romange: Ok, I am sure like 99.9% you are correct. Can you either:
1. add a test to cluster_test.py
2. or at least manually test with `tools/cluster_mgr.py`  to see there are no surprises there?

In both scenarios, I guess the simplest test would be to move all the slots from one shard to another and verify the index is 0.

romange (dismissed):

romange:
I guess `PerformDeletion` is now only called from `Del` and can be merged into it.

src/server/db_slice.cc
```diff
@@ -802,18 +802,25 @@ void DbSlice::ActivateDb(DbIndex db_ind) {
   CreateDb(db_ind);
 }
 
-void DbSlice::Del(Context cntx, Iterator it) {
+void DbSlice::Del(Context cntx, Iterator it, DbTable* db_table) {
   CHECK(IsValid(it));
 
-  auto& db = db_arr_[cntx.db_index];
+  ExpIterator exp_it;
+  DbTable* table = db_table ? db_table : db_arr_[cntx.db_index].get();
```
romange, line 809: @mkaruza  I do not understand why `db_table` is needed. why can not we specify correct `cntx.db_index` in FlushSlotsCb?

src/server/db_slice.cc
```diff
@@ -836,11 +843,14 @@ void DbSlice::FlushSlotsFb(const cluster::SlotSet& slot_ids) {
...
     it.AdvanceIfNotOccupied();
+    DbContext db_cntx;
+    db_cntx.time_now_ms = GetCurrentTimeMs();
+    db_cntx.db_index = table->index;
     while (!it.is_done()) {
       std::string_view key = it->first.GetSlice(&tmp);
       SlotId sid = KeySlot(key);
       if (slot_ids.Contains(sid) && it.GetVersion() < next_version) {
-        PerformDeletion(Iterator::FromPrime(it), table.get());
+        Del(db_cntx, Iterator::FromPrime(it), table.get());
```
romange, line 855: I understand now. `FlushAll` removes the instance table and we hold to it via `table`.
do we use `db_cntx` at all here?

src/server/db_slice.cc
```diff
@@ -836,11 +843,14 @@ void DbSlice::FlushSlotsFb(const cluster::SlotSet& slot_ids) {
   std::string tmp;
   auto iterate_bucket = [&](PrimeTable::bucket_iterator it) {
     it.AdvanceIfNotOccupied();
+    DbContext db_cntx;
```
romange: pull it out of `iterate_bucket` as we do not need to create it every time

src/server/db_slice.cc
mkaruza: We need because index deletion callback uses it, but only timestamp field. I didn't want to change interface too much.

romange:
I think a comment should be added where you pass table as it's not clear why we can not reference via db_arr

mkaruza:
> I think a comment should be added where you pass table as it's not clear why we can not reference via db_arr

There is comment before in code why we took copy of smart pointer but added.

romange (approved):

PR #6068, opened 2025-11-17
fix(replication): use correct replica pointer in takeover catchup loop

vyavdoshenko:
## The bug: 

After the data loss fixes in https://github.com/dragonflydb/dragonfly/pull/5727 when master receives a DFLY TAKEOVER it first a) waits for the taking over replica to catch up b) replies with OK c) waits for the rest of the replicas to catch up 

On step(c) instead of waiting for the rest of the replicas to catch up we unconditionally waited for (a) again times the number of registered replicas.  This caused each wait flow to timeout and report `log warning` that the replicas `did not sync in time`.

## The fix:

Use the correct replica pointer on each iteration of the loop.

## Dragonfly version the bug was introduced
Dragonfly v1.34.0

## Bug: area of impact
Should be small. For each registered replica we got a timeout. Unless that replica got registered recently, chances are is that it's on par with master. Why? Even though we waited for the wrong node, we waited for timeout seconds which should have given enough time for most replicas to properly catchup (assuming a reasonable lag here).

### DIFF
--- a/tests/dragonfly/replication_test.py
+++ b/tests/dragonfly/replication_test.py
@@ -3594,8 +3594,6 @@ async def test_takeover_bug_wrong_replica_checked_in_logs(df_factory):
     master.stop(kill=False)
     timeout_logs = master.find_in_logs("Couldn't synchronize with replica")
 
-    # BUG: logs show replica[0] port (initiating), should show replica[1] or replica[2]
-    # After fix: no timeout logs (replicas sync correctly) OR logs show correct replica
     if timeout_logs:
         for log in timeout_logs:
             assert (


tests/dragonfly/replication_test.py
```diff
@@ -3558,3 +3558,46 @@ async def get_memory(client, field):
...
+    try:
+        await clients[0].execute_command("REPLTAKEOVER 1")
+    except Exception:
+        pass
+
+    # Check master logs
+    master.stop(kill=False)
+    timeout_logs = master.find_in_logs("Couldn't synchronize with replica")
+
+    # BUG: logs show replica[0] port (initiating), should show replica[1] or replica[2]
```
romange: `BUG: ` comments belong to the PR comments because once the code is submitted the notion of BUG disappears.

romange:
was it a regression? how did it work so far?

vyavdoshenko:
> was it a regression? how did it work so far?

I don't know how it worked before. It can work if we have only one replica, but having more than one, we don't even understand clearly what happens during a takeover. I will take a look at the history of this change to find out if it's a regression or just old code.

tests/dragonfly/replication_test.py
vyavdoshenko: Fixed

vyavdoshenko:
> was it a regression? how did it work so far?

This code was added September 10, 2025: https://github.com/dragonflydb/dragonfly/commit/425cacec8

romange:
@kostasrim  please review. Seems like a regression, if yes - let's identify the affected releases.

kostasrim:
> @kostasrim please review. Seems like a regression, if yes - let's identify the affected releases.

bug impact surface is small and it didn't broke anything that was not broken before (my PR fixed dataloss which we never covered in the first place). imo this is kinda an abnormal flow -- I don't recall a place that we first register a replica to attempt a takeover, and before it attempts the takeover we reregister a new one. 

p.s. I also edited the description in case you want a little bit of background.

version I introduced this is v1.34.0

src/server/dflycmd.cc
```diff
@@ -567,8 +567,8 @@ void DflyCmd::TakeOver(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext
         continue;
       }
 
-      auto cb = [&, end_time](EngineShard* shard) {
-        if (!WaitReplicaFlowToCatchup(end_time, replica_ptr.get(), shard)) {
+      auto cb = [repl_ptr = repl_ptr, end_time, &rest_catchup_success](EngineShard* shard) {
```
kostasrim, line 570: IMO & is fine 🤷

PR #6069, opened 2025-11-17
chore: asynchronous IO for connection fiber

### LINKED ISSUE #6028: Integrate and test OnRecv in ConnFb
Currently `ConnFb` is synchronous in respect to socket reads. This is not flexible enough for the uses cases discussed in #5952 and #6006 

The first step is to turn synchronous reads to asynchronous via the OnRecv interface.

- Introduce a flag for the new ConnFb functionality
- Use notification based interface by registering an OnRecv hook
- test

kostasrim:
This PR adds `IoLoopV2` that removes fiber blocking read calls from the flow. This allows the connection fiber to handle other events while recv calls are handled asynchronously from readiness events triggered by a registered multishot poll. 

* add IoLoopV2 that uses non fiber blocking async reads via poll registration
* add a manual trigger workflow to run all the regression tests (both epoll/uring) without tls

Follow up on a separate PR:

* Allow also EnableRecvMultishot()

Resolves #6028

## Workarounds and Failing tests

1. test_timeout fails in `assert int(info["timeout_disconnects"]) >= 1`. It was non critical and this test has also failed upstream a couple of times.
2. Added DflyInstance::has_arg to avoid propagating io_loop_v2 flag on tests that use older versions of dragonfly
3. Calls ResetOnRecvHook inside Migrate() instead of OnPreMigrateThread. The reason for that is that OnPreMigrateThread check fails if the connection is closing and because deregistering the hook is preemptive -- this invariant might break. To mitigate this, we deregister inside Migrate() instead and if after it the connection is closing we exit instead of starting the migration.
4. added migration_allowed_to_register_ to flag when a migration should register a callback. A connection might migrate before it reaches its asynchronous flow (io_loop_v2) which can lead to double registration errors (one from the migration and one once it reaches io_loop_v2). To respect the synchronous parts of the connection, we use a flag to indicate when a migration should also register a callback.
5. corner case in `ConsumeInput()` read comment in: https://github.com/dragonflydb/dragonfly/pull/6069/files#diff-b88cbca1b2c9337ab2a37262f9d516b117f4996da25e2ae173e0e129f2b82b20R1207

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -648,6 +648,8 @@ void Connection::OnShutdown() {
   VLOG(1) << "Connection::OnShutdown";
 
   BreakOnce(POLLHUP);
+  io_ec_ = make_error_code(errc::connection_aborted);
+  io_event_.notify_one();
 }
 
 void Connection::OnPreMigrateThread() {


src/facade/dragonfly_connection.cc
```diff
@@ -2161,6 +2193,153 @@ bool Connection::WeakRef::operator==(const WeakRef& other) const {
...
+    io_event_.await(
+        [this]() { return io_buf_.InputLen() > 0 || io_ec_ || io_buf_.AppendLen() == 0; });
+
+    if (io_ec_) {
+      LOG_IF(WARNING, cntx()->replica_conn) << "async io error: " << ec;
+      return std::exchange(io_ec_, {});
+    }
+
+    phase_ = PROCESS;
+    bool is_iobuf_full = io_buf_.AppendLen() == 0;
```
kostasrim, line 2241: No functional change from here and onwards in comparison IoLoop

tests/dragonfly/connection_test.py
```diff
@@ -860,6 +858,7 @@ async def check_stats():
     await check_stats()
 
 
+@pytest.mark.tls
```
kostasrim: no changes to the tests. I added the `tls` label to disable tls tests on the manual workflow I introduced (because tls is broken for now).

kostasrim:
https://github.com/dragonflydb/dragonfly/actions/runs/19820043215

src/facade/dragonfly_connection.cc
```diff
@@ -10,9 +10,13 @@
 #include <absl/strings/str_cat.h>
 #include <absl/time/time.h>
 
+#include <condition_variable>
```
kostasrim: TODO remove the includes. plugin is too aggressive

kostasrim:
@romange you can start taking a look. 

1. I added a workflow + flag to run ioloopv2. The former is to avoid running the tests on both paths every time. Once we slowly migrate I will remove the workflow file -- it's just a temporary solution
2. I only used epoll based notifications -- will follow up with `enable_multi_shot_=true` as well
3. I introduced ioloopv2 to provide clear serparation between the two implementations

.github/actions/regression-tests/action.yml
```diff
@@ -55,14 +58,18 @@ runs:
         export ROOT_DIR="${GITHUB_WORKSPACE}/tests/dragonfly/valkey_search"
         export UBSAN_OPTIONS=print_stacktrace=1:halt_on_error=1 # to crash on errors
 
+        if [[ "${{inputs.df-arg}}" == 'epoll' ]]; then
```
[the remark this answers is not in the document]
romange: we already have `epoll` argument why do you need this?

.github/actions/regression-tests/action.yml
```diff
@@ -55,14 +58,18 @@ runs:
...
+          export DF_ARG="--df ${{inputs.df-arg}}"
+        fi
+
         if [[ "${{inputs.epoll}}" == 'epoll' ]]; then
           export FILTER="${{inputs.filter}} and not exclude_epoll"
           # Run only replication tests with epoll
-          timeout 80m pytest -m "$FILTER" --durations=10 --timeout=300 --color=yes --json-report --json-report-file=report.json dragonfly --df force_epoll=true --log-cli-level=INFO || code=$?
+          timeout 80m pytest -m "$FILTER" $DF_ARG --durations=10 --timeout=300 --color=yes --json-report --json-report-file=report.json dragonfly --df force_epoll=true --df --log-cli-level=INFO || code=$?
```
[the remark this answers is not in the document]
romange: +1

.github/workflows/ioloop-v2-regtests.yml
```diff
@@ -0,0 +1,81 @@
...
+        uses: ./.github/actions/regression-tests
+        with:
+          dfly-executable: dragonfly
+          gspace-secret: ${{ secrets.GSPACES_BOT_DF_BUILD }}
+          build-folder-name: build
+          filter: ${{ matrix.build-type == 'Release' && 'not debug_only and not tls' || 'not opt_only and not tls' }}
+          aws-access-key-id: ${{ secrets.AWS_S3_ACCESS_KEY }}
+          aws-secret-access-key: ${{ secrets.AWS_S3_ACCESS_SECRET }}
+          s3-bucket: ${{ secrets.S3_REGTEST_BUCKET }}
+          df-arg: "expiremental_io_loop_v2"
```
romange: you export `df-arg`  only if it's 'epoll'. 
I do not understand what happens here.

src/facade/dragonfly_connection.cc
romange: yes, please

src/facade/dragonfly_connection.cc
```diff
@@ -1268,7 +1287,7 @@ Connection::ParserStatus Connection::ParseRedis(unsigned max_busy_cycles) {
     }
   } while (RedisParser::OK == result && read_buffer.size() > 0 && !reply_builder_->GetError());
 
-  io_buf_.ConsumeInput(io_buf_.InputLen());
+  io_buf_.ConsumeInput(total);
```
romange: what was the reason for this change?

src/facade/dragonfly_connection.cc
```diff
@@ -1096,7 +1104,12 @@ void Connection::ConnectionFlow() {
   // Main loop.
   if (parse_status != ERROR && !ec) {
     UpdateIoBufCapacity(io_buf_, stats_, [&]() { io_buf_.EnsureCapacity(64); });
-    auto res = IoLoop();
+    variant<error_code, Connection::ParserStatus> res;
+    if (GetFlag(FLAGS_expiremental_io_loop_v2)) {
```
romange: also, please avoid using with tls sockets.
btw, it should fail the tls tests with expiremental_io_loop_v2=true and if it's not - then something is wrong with your setup.

src/facade/dragonfly_connection.cc
```diff
@@ -2161,6 +2180,158 @@ bool Connection::WeakRef::operator==(const WeakRef& other) const {
...
+void Connection::DoReadOnRecv(const util::FiberSocketBase::RecvNotification& n) {
+  if (std::holds_alternative<std::error_code>(n.read_result)) {
+    io_ec_ = std::get<std::error_code>(n.read_result);
+    return;
+  }
+
+  // TODO non epoll API via EnableRecvMultishot
+  // if (std::holds_alternative<io::MutableBytes>(n.read_result))
+
+  if (std::holds_alternative<bool>(n.read_result)) {
```
romange: Please use `RecvNotification::RecvCompletion` alias

src/facade/dragonfly_connection.cc
```diff
@@ -2161,6 +2180,158 @@ bool Connection::WeakRef::operator==(const WeakRef& other) const {
...
+      return;
+    }
+
+    if (io_buf_.AppendLen() == 0) {
+      // We will regrow in IoLoop
+      return;
+    }
+
+    io::MutableBytes buf = io_buf_.AppendBuffer();
+    int res = recv(socket_->native_handle(), buf.data(), buf.size(), 0);
```
romange: Please use TryRecv interface

src/facade/dragonfly_connection.cc
```diff
@@ -2161,6 +2180,158 @@ bool Connection::WeakRef::operator==(const WeakRef& other) const {
...
+  // Breaks with TLS. RegisterOnRecv is unimplemented.
+  peer->RegisterOnRecv([this](const FiberSocketBase::RecvNotification& n) {
+    DoReadOnRecv(n);
+    io_event_.notify();
+  });
+
+  do {
+    HandleMigrateRequest();
+
+    // We *must* poll again for readiness. The event handler we registered above
```
romange: it is possible to improve for sure. 'We *must*' is a too strong statement as you can introduce a state that tracks whether the read is needed and when it's not needed.
Either change to TODO or fix it in this PR.

src/facade/dragonfly_connection.cc
```diff
@@ -2161,6 +2180,158 @@ bool Connection::WeakRef::operator==(const WeakRef& other) const {
...
+    // So, when we get notified below in io_event_.wait() we might read less data
+    // than it is available because io_buf_ does not have enough capacity. If we loop,
+    // and do not attempt to read from the socket again we can deadlock. To avoid this,
+    // we poll once for readiness before preempting.
+    DoReadOnRecv(FiberSocketBase::RecvNotification{true});
+    io_event_.await(
+        [this]() { return io_buf_.InputLen() > 0 || io_ec_ || io_buf_.AppendLen() == 0; });
+
+    if (io_ec_) {
+      LOG_IF(WARNING, cntx()->replica_conn) << "async io error: " << ec;
```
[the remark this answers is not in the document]
romange: yep

src/facade/dragonfly_connection.h
```diff
@@ -421,6 +426,9 @@ class Connection : public util::Connection {
   util::fb2::CondVarAny cnd_;             // dispatch queue waker
   util::fb2::Fiber async_fb_;             // async fiber (if started)
 
+  std::error_code io_ec_;
+  util::fb2::EventCount io_event_;
```
romange: do not use `EventCount` for thread-local synchronization - use CondVarAny instead

tests/dragonfly/connection_test.py
romange: yep, so instead - avoid enabling v2 for tls sockets. we have is_tls_ flag for that

.github/workflows/ioloop-v2-regtests.yml
kostasrim: I messed a rebase there I think. regardless it's fixed

src/facade/dragonfly_connection.cc
kostasrim: Because above we have a preemption point and the size of the io_buf_.InputLen() might change because we read data

romange: worth adding a comment here explaining this.

kostasrim: done

src/facade/dragonfly_connection.cc
kostasrim: Well, is it a regression to poll unconditionally ? I imagine, the overhead for EAGAIN is minimal so we don't really need a variable for this. I wrote *must* because I found poll to be the simplest way. I reworded the comment to be less strong. My only question is, what is the reason to prefer a flag over an unconditional poll assuming EAGAIN is not eating CPU time?

src/facade/dragonfly_connection.cc
```diff
@@ -2161,6 +2180,161 @@ bool Connection::WeakRef::operator==(const WeakRef& other) const {
...
+        return;
+      }
+
+      if (ec == errc::connection_aborted || ec == errc::connection_reset) {
+        // The peer can shutdown the connection abruptly.
+        io_ec_ = ec;
+        return;
+      }
+
+      LOG_EVERY_T(ERROR, 10) << "Recv error: " << ec;
```
romange: remove?

src/facade/dragonfly_connection.cc
```diff
@@ -2161,6 +2180,161 @@ bool Connection::WeakRef::operator==(const WeakRef& other) const {
...
+    if (io_buf_.AppendLen() == 0) {
+      // We will regrow in IoLoop
+      return;
+    }
+
+    io::MutableBytes buf = io_buf_.AppendBuffer();
+    io::Result<size_t> res = socket_->TryRecv(buf);
+
+    // error path
+    if (!res) {
```
romange: for readability, I would reverse this condition and handle the happy case first. aka
`if (res && *res) {`

tests/dragonfly/pymemcached_test.py
```diff
@@ -8,6 +8,7 @@
 
 from . import dfly_args
 from .instance import DflyInstance
+from .utility import *
```
romange: not needed?

src/facade/dragonfly_connection.cc
```diff
@@ -1096,7 +1101,12 @@ void Connection::ConnectionFlow() {
   // Main loop.
   if (parse_status != ERROR && !ec) {
     UpdateIoBufCapacity(io_buf_, stats_, [&]() { io_buf_.EnsureCapacity(64); });
-    auto res = IoLoop();
+    variant<error_code, Connection::ParserStatus> res;
+    if (GetFlag(FLAGS_expiremental_io_loop_v2) && !is_tls_) {
```
romange: introduce a boolean variable instead of duplicating condition below.

src/facade/dragonfly_connection.cc
```diff
@@ -2161,6 +2180,161 @@ bool Connection::WeakRef::operator==(const WeakRef& other) const {
...
+      io_ec_ = ec;
+      return;
+    }
+
+    if (*res == 0) {
+      io_ec_ = make_error_code(errc::connection_aborted);
+      return;
+    }
+    // A recv call can return fewer bytes than requested even if the
+    // socket buffer actually contains enough data to satisfy the full request.
```
romange: why do you say this? yes, there are some corner cases but in general, Recv will try to unload as much as possible and fill the destination buffer

src/facade/dragonfly_connection.cc
```diff
@@ -2161,6 +2180,161 @@ bool Connection::WeakRef::operator==(const WeakRef& other) const {
...
+  ParserStatus parse_status = OK;
+
+  size_t max_iobfuf_len = GetFlag(FLAGS_max_client_iobuf_len);
+
+  auto* peer = socket_.get();
+  recv_buf_.res_len = 0;
+
+  // TODO EnableRecvMultishot
+
+  // Breaks with TLS. RegisterOnRecv is unimplemented.
```
romange: this comment should move where you check is_tls_

src/facade/dragonfly_connection.cc
romange: it's a system call and we add another system-call per each loop. It's fine for now but I think we will have to add a state variable to track when this call is needed. I do not know how it effects performance, in fact, we will need to do some benchmarking tests on this PR. I will do it next week

romange:
LGTM in general, let's try submitting it this week.

src/facade/dragonfly_connection.cc
kostasrim: recv can return less data than available on the socket in non-blocking mode. so even if the read buffers contain 1k bytes and even if you try to read 1k, recv might return less.

I don't want to sound too confident here, that's my current understanding of the behaviour in non blocking mode.

Either way, I want to describe the reason why poll is important: a) because it could be that we read less data than it is available or b) the capacity of io_buf_ is less than what recv can satisfy in one call

kostasrim:
> LGTM in general, let's try submitting it this week.

I addressed your comments and run the tests again. Plz rereview

src/facade/dragonfly_connection.cc
romange: I think this exact example can not happen. i,e, with 1KB you will always read 1KB. There are some limits of how much it will read in one call, but besides that, my understanding is that a) can happen only because of b)

romange: also, define "available".  how can you observe "available" size without reading the data?

romange: actually, you can: `ioctl(socket_fd, FIONREAD, &bytes_available)`

romange: Hmm, gemini confirms your statement. Good to know.

src/facade/dragonfly_connection.cc
```diff
@@ -1258,6 +1278,7 @@ Connection::ParserStatus Connection::ParseRedis(unsigned max_busy_cycles) {
           << "Redis parser error: " << result << " during parse: " << ToSV(read_buffer);
     }
     read_buffer.remove_prefix(consumed);
+    total += consumed;
```
romange: sorry I did not catch earlier - I suggest removing `total` counter and instead calling here
`io_buf_.ConsumeInput(consumed);`

I think it's better because as you said - we preempt here, during which time we can read more data.
this will lead to much smoother "pipeline-like" I/O. 

Currently: you read, say 4KB, and your io_buf does not have more space. you start executing commands. 
your read notifications won't fill the buffer until you finish running all the commands and empty your io_buf. only then you will call the recv and copy the next chunk.

src/facade/dragonfly_connection.cc
kostasrim: Yeah perplexity confirmed this for me as well but I do not trust it that much -- If I recall correctly I think I read this somewhere in a conversation on the kernel mailing list but I didn't verify it myself (and hence wrote I am not sure)

kostasrim: > actually, you can: `ioctl(socket_fd, FIONREAD, &bytes_available)`

yes the receive buffers

src/facade/dragonfly_connection.cc
```diff
@@ -1258,6 +1296,7 @@ Connection::ParserStatus Connection::ParseRedis(unsigned max_busy_cycles) {
           << "Redis parser error: " << result << " during parse: " << ToSV(read_buffer);
     }
     read_buffer.remove_prefix(consumed);
+    io_buf_.ConsumeInput(consumed);
```
kostasrim: @romange 

This breaks test_publish_stuck. Moved back to `size_t consumed_total`. I suspect the root cause is how we resize with ConsumeInput but I don't want to start messing with this here again. I added a TODO to follow up. 

I want to reach a stable state and then follow up with improvements

kostasrim:
augment review

kostasrim:
augment review

kostasrim:
https://github.com/dragonflydb/dragonfly/actions/runs/20094043670

src/facade/dragonfly_connection.cc
```diff
@@ -644,6 +648,8 @@ void Connection::OnShutdown() {
   VLOG(1) << "Connection::OnShutdown";
 
   BreakOnce(POLLHUP);
+  io_ec_ = make_error_code(errc::connection_aborted);
```
romange: I would expect that it's not needed as `Connection::Shutdown()` should generate a socket notification

src/facade/dragonfly_connection.cc
```diff
@@ -665,12 +671,22 @@ void Connection::OnPreMigrateThread() {
 }
 
 void Connection::OnPostMigrateThread() {
-  DVLOG(1) << "[" << id_ << "] OnPostMigrateThread";
+  DVLOG(1) << "[" << id_ << "] OnPostMigrateThread " << GetClientId();
```
romange: it's not needed. `GetClientId();` is `id_` which is already printed in DVLOG

src/facade/dragonfly_connection.cc
```diff
@@ -665,12 +671,22 @@ void Connection::OnPreMigrateThread() {
...
 
   // Once we migrated, we should rearm OnBreakCb callback.
   if (breaker_cb_ && socket()->IsOpen()) {
     socket_->RegisterOnErrorCb([this](int32_t mask) { this->OnBreakCb(mask); });
   }
+
+  const bool io_loop_v2 = GetFlag(FLAGS_experimental_io_loop_v2);
+  if (io_loop_v2 && !is_tls_ && socket_ && socket_->IsOpen() && allowed_to_register_) {
+    socket_->RegisterOnRecv([this](const FiberSocketBase::RecvNotification& n) {
+      CHECK(this);
```
romange: please remove `check(this)` -  it's always true.

src/facade/dragonfly_connection.cc
```diff
@@ -1324,7 +1362,7 @@ void Connection::OnBreakCb(int32_t mask) {
   cnd_.notify_one();  // Notify dispatch fiber.
 }
 
-void Connection::HandleMigrateRequest() {
+void Connection::HandleMigrateRequest(bool unregister) {
```
romange: you do not use `unregister`

src/facade/dragonfly_connection.cc
```diff
@@ -1041,11 +1057,22 @@ void Connection::ConnectionFlow() {
...
   // Main loop.
   if (parse_status != ERROR && !ec) {
     UpdateIoBufCapacity(io_buf_, stats_, [&]() { io_buf_.EnsureCapacity(64); });
-    auto res = IoLoop();
+    variant<error_code, Connection::ParserStatus> res;
+    if (io_loop_v2 && !is_tls_) {
+      // Migrations should call RegisterRecv if the connection has reached here once.
+      // Otherwise, a migration won't register and instead wait for the connection state
+      // to first reach here and then call RegisterRecv inside IoLoopV2.
+      allowed_to_register_ = true;
```
romange: I think `allowed_to_register_` is redundant - I did not see how you use it and it's always true inside RegisterRecv

kostasrim: It's needed here as we might end up calling RegisterOnRecv twice. Everything above the call to `IoLoopV2` is synchronous and fiber blocking. So, the connection might migrate before it reaches IoLoopV2, call `RegisteOnRecv()` as part of the PostMigration(). Eventually, it will reach IoLoopV2 body which will try to register and check fail. I edited the comment.

src/facade/dragonfly_connection.cc
kostasrim: it doesn't for some some reason -- `test_no_tls_on_admin_port`

I want to get this in a green state and then I can follow up

romange (approved):

src/facade/dragonfly_connection.cc
```diff
@@ -1041,11 +1056,22 @@ void Connection::ConnectionFlow() {
...
 
   // Main loop.
   if (parse_status != ERROR && !ec) {
     UpdateIoBufCapacity(io_buf_, stats_, [&]() { io_buf_.EnsureCapacity(64); });
-    auto res = IoLoop();
+    variant<error_code, Connection::ParserStatus> res;
+    if (io_loop_v2 && !is_tls_) {
+      // Everything above the IoLoopV2 is fiber blocking. A connection can migrate before
+      // it reaches here and will cause a double RegisterOnRecv check fail. To avoid this,
+      // a migration shall only call RegisterOnRev if it reached the main IoLoopV2 below.
```
romange, line 1068: I think it's not enough. Similarly to `CancelOnErrorCb` - you need to unregister from the polling in `OnPreMigrateThread` function

kostasrim: WIll take a look

kostasrim: I see. So, I do have it, but inside `Migrate()` and not in `OnPreMigrateThread`. The reason is that `ResetOnRecvHook()` is preemptive and by the time we wake up the check:

```
 658   CHECK(!cc_->conn_closing);
```

Can fail. To avoid this, I moved it to `Migrate()` such that if after reseting the hook the connection `is closing`, we exit instead of moving on with the migration. It's a subtle corner case.

Logically, we should deregister in OnPreMigrate just like we register on OnPostMigrate -- however I don't see a clean way to do this

romange: we could call ResetOnRecvHook after the check(), after `socket_->CancelOnErrorCb();`,
 but ok - we can check it later.

romange: I do not want you to do more changes in this PR. Can you please summarize 
TODOs - failing/skipped tests, unresolved issues/workarounds like this one, pubsub failure  etc in the PR  description so it won't get lost completely after we merge?

kostasrim: > we could call ResetOnRecvHook after the check(), after `socket_->CancelOnErrorCb();`,
> but ok - we can check it later.

Yes but then we would move on with a migration on a closing connection. Doin that seems at least redundant 🤷

kostasrim: > I do not want you to do more changes in this PR. Can you please summarize TODOs - failing/skipped tests, unresolved issues/workarounds like this one, pubsub failure etc in the PR description so it won't get lost completely after we merge?

sure! one sec

kostasrim: @romange done!

PR #6070, opened 2025-11-17
fix(replication): add socket timeout to REPLTAKEOVER command

vyavdoshenko:
Add socket timeout to Replica::TakeOver() to prevent indefinite hanging when master is unresponsive.

### DIFF
--- a/tests/dragonfly/replication_test.py
+++ b/tests/dragonfly/replication_test.py
@@ -3568,10 +3568,10 @@ async def test_big_strings(df_factory):
 
 
 @pytest.mark.slow
-async def test_takeover_timeout_on_unresponsive_master(df_factory):
-    master = df_factory.create(proactor_threads=4)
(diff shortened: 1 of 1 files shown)


tests/dragonfly/replication_test.py
```diff
@@ -3558,3 +3561,56 @@ async def get_memory(client, field):
...
+    # Setup replication
+    await c_replica.execute_command(f"REPLICAOF localhost {master.port}")
+    await wait_available_async(c_replica)
+
+    # Write some data
+    for i in range(10):
+        await c_master.set(f"key{i}", f"val{i}")
+    await asyncio.sleep(0.2)
+
+    # PAUSE master process (SIGSTOP) - socket stays open but doesn't respond
```
romange: did not know this is possible. it's really cool. we should have more tests like this

romange:
please rebase

PR #6071, opened 2025-11-17
fix(replication): Add detailed error message for REPLTAKEOVER failures

vyavdoshenko:
Improve error diagnostics for `REPLTAKEOVER` command failures by including the actual error message.

Before: `Couldn't execute takeover`  
After: `Couldn't execute takeover: Connection timed out` (or other specific error)

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -3759,7 +3759,7 @@ void ServerFamily::ReplTakeOver(CmdArgList args, const CommandContext& cmd_cntx)
 
   std::error_code res = replica_->TakeOver(ArgS(args, 0), save_flag);
   if (res) {
-    LOG(ERROR) << "Takeover failed with error: " << res << " - " << res.message();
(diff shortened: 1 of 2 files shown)


src/server/server_family.cc
```diff
@@ -3758,8 +3758,10 @@ void ServerFamily::ReplTakeOver(CmdArgList args, const CommandContext& cmd_cntx)
   }
 
   std::error_code res = replica_->TakeOver(ArgS(args, 0), save_flag);
-  if (res)
-    return builder->SendError("Couldn't execute takeover");
+  if (res) {
+    LOG(ERROR) << "Takeover failed with error: " << res << " - " << res.message();
```
romange: I suspect you do not need both LOG(ERROR) and ` builder->SendError` as we log errors in main_service. At least reduce the severity to WARNING

vyavdoshenko: Fixed.
I changed the level of the message to warning.

PR #6075, opened 2025-11-18
test: fix flaky test_takeover_bug_wrong_replica_checked_in_logs

vyavdoshenko:
- Add 0.5s sleep to ensure replica[0] is synced before takeover
- Increase timeout from 1s to 10s

The test still correctly verifies that OTHER replicas (not initiating one) are checked in the catchup loop.

https://github.com/dragonflydb/dragonfly/actions/runs/19465578764/job/55699678822

### DIFF
--- a/tests/dragonfly/replication_test.py
+++ b/tests/dragonfly/replication_test.py
@@ -1,6 +1,9 @@
+import os
 import platform
 import shutil
+import signal
 import tarfile
+import time
 import urllib.request
 from itertools import chain, repeat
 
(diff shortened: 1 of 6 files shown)


tests/dragonfly/replication_test.py
```diff
@@ -3589,8 +3589,11 @@ async def test_takeover_bug_wrong_replica_checked_in_logs(df_factory):
 
     # Reconnect replica[1] and immediately takeover from replica[0]
     await clients[1].execute_command(f"REPLICAOF localhost {master.port}")
+
+    await asyncio.sleep(0.5)
```
romange: this is not how we do it - it's not robust.

check `wait_available_async` , and
```
async with async_timeout.timeout(3):
        await wait_for_replicas_state
```
and `await check_all_replicas_finished` and @assert_eventually
  
yes, we have lots of utilities to do the same thing but sleep is not the right one.

kostasrim: no need for this either IMO. Just increase the timeout on the takeover to 10 seconds and remove the `sleep`. Takeover flow will eventually gurantee that the replica is in sync as long as it has enough time to finish.

kostasrim: Also @vyavdoshenko

we can improve the test slightly. Could I plz ask you to generalize the assert on line `3604` to:

```
 timeout_logs = master.find_in_logs("Couldn't synchronize with replica")
 assert not timeout_logs
```

That way we know that the old master waited for the rest of the replicas to sync.

tests/dragonfly/replication_test.py
```diff
@@ -3589,8 +3589,11 @@ async def test_takeover_bug_wrong_replica_checked_in_logs(df_factory):
...
+
     try:
-        await clients[0].execute_command("REPLTAKEOVER 1")
+        await clients[0].execute_command("REPLTAKEOVER 10")
```
kostasrim: It's safe to remove the `try/catch` block. The replica takeover should not fail

tests/dragonfly/replication_test.py
vyavdoshenko: Fixed

tests/dragonfly/replication_test.py
vyavdoshenko: fixed

romange:
still need to rebase

vyavdoshenko:
> still need to rebase

fixed

PR #6076, opened 2025-11-19
fix(server): Move LpGetView to core target

dranikpg:
In  #6015, `core/ListpackWrap` is used in tiering decoders, which would need `server/container_utils` for LpGetView, so it was a bad reverse dependency of the server module from the core one.

Move LpGetView to listpack_wrap.h


vyavdoshenko (approved):
lgtm

PR #6086, opened 2025-11-20
feat(tiering): Basic metrics

dranikpg:
Basicc prometheus metrics for tiering
image

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -1810,6 +1810,9 @@ void PrintPrometheusMetrics(uint64_t uptime, const Metrics& m, DflyCmd* dfly_cmd
   AppendMetricValue("memory_by_class_bytes", m.coordinator_stats.stored_cmd_bytes, {"class"},
                     {"conn_stored_commands"}, &memory_by_class_bytes);
 
+  AppendMetricValue("memory_by_class_bytes", m.search_stats.used_memory, {"class"}, {"search_used"},
+                    &memory_by_class_bytes);
+
   // Command stats
   if (!m.cmd_stats_map.empty()) {
     string command_metrics;


tools/local/monitoring/grafana/provisioning/dashboards/dragonfly.json
```diff
@@ -2265,6 +2265,401 @@
...
+      },
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "PBFA97CFB590B2093"
+          },
+          "disableTextWrap": false,
+          "editorMode": "builder",
+          "expr": "rate(dragonfly_tiered_hits[$__rate_interval])",
```
dranikpg, line 2451: Probably should be an irate

src/server/server_family.cc
```diff
@@ -1955,6 +1955,49 @@ void PrintPrometheusMetrics(uint64_t uptime, const Metrics& m, DflyCmd* dfly_cmd
                             m.shard_stats.defrag_attempt_total, COUNTER, &resp->body());
   AppendMetricWithoutLabels("defrag_objects_moved", "Objects moved",
                             m.shard_stats.defrag_realloc_total, COUNTER, &resp->body());
+
+  // Tiered metrics
+  {
+    AppendMetricWithoutLabels("tiered_entries", "Tiered entries", total.tiered_entries,
```
romange, line 1964: nice grouping.

dranikpg:
> What about cool memory? we have memory_by_class_bytes - can you add anything accountable that contributes to memory usage?

Shouldn't be anything, as even cold_storage_bytes is part of just object memory usage.

> btw, while you are at it, can you please also add m.search_stats.used_memory to memory_by_class_bytes that we missed

Added

Issue #6009, opened 2025-11-04
Implement a coordinator logic for cross-shard search request


BorysTheDev:
Should be done after #6008 

- Add FT.SEARCH command response parser.
- Add FT.SEARCH command modification to always get the score of the search request
- Reuse the existing connections for request distribution. Because the connections will be global, we should have a queue of corresponding requests/responses from different threads to support a correct order and have the ability to utilize the pipeline.
- Implement pagination and sorting logic for cross-shard searching.

PR #6090, opened 2025-11-20
feat: add ftsearch result parsing for css

BorysTheDev:
part of https://github.com/dragonflydb/dragonfly/issues/6009

- Added flag cluster_search
- Added test for cluster FT.SEARCH 
- Added base FT.SEARCH results parsing

In the future, I think we need to add something like CmdArgParser for cross-shard search result parsing

### DIFF
--- a/src/server/search/search_family.cc
+++ b/src/server/search/search_family.cc
@@ -1296,9 +1296,11 @@ void SearchFamily::FtList(CmdArgList args, const CommandContext& cmd_cntx) {
 
 static vector<SearchResult> FtSearchCSS(std::string_view idx, std::string_view query,
                                         std::string_view args_str) {
-  util::fb2::Mutex result_mu;
-  vector<SearchResult> results;  // todo resize to number of shards
+  vector<SearchResult> results;
   std::string cmd = absl::StrCat("FT.SEARCH ", idx, " ", query, " CSS ", args_str);
+
+  // TODO for now we suppose that callback is called synchronously. If not, we need to add
+  // synchronization here for results vector modification.
   cluster::Coordinator::Current().DispatchAll(cmd, [&](const facade::RespVec& res) {
     VLOG(3) << "FT.SEARCH CSS reply: " << res;
 
@@ -1317,7 +1319,6 @@ static vector<SearchResult> FtSearchCSS(std::string_view idx, std::string_view q
       return;
     }
 
-    std::lock_guard lock{result_mu};
     results.emplace_back();
     results.back().total_hits = *size;
     for (size_t i = 2; i < res.size(); i += 2) {


src/server/search/search_family.cc
```diff
@@ -1342,6 +1403,9 @@ void SearchFamily::FtSearch(CmdArgList args, const CommandContext& cmd_cntx) {
       return builder->SendError(*res.error);
   }
 
+  docs.insert(docs.end(), std::make_move_iterator(css_docs.begin()),
+              std::make_move_iterator(css_docs.end()));
+
```
vyavdoshenko, line 1410: I think these results will rarely meet SORT requirements.

vyavdoshenko: LIMIT as well

BorysTheDev: SEARCH and LIMIT will be separate steps. At first, I want to have a prototype that works

vyavdoshenko: Please add TODO

src/server/search/search_family.cc
```diff
@@ -1292,6 +1294,65 @@ void SearchFamily::FtList(CmdArgList args, const CommandContext& cmd_cntx) {
...
+      return;
+    }
+
+    const auto size = res[0].GetInt();
+    if (!size.has_value()) {
+      LOG(ERROR) << "FT.SEARCH CSS reply unexpected type: " << static_cast<int>(res[0].type);
+      return;
+    }
+
+    std::lock_guard lock{result_mu};
```
vyavdoshenko: What do you want to guard here?

vyavdoshenko:
Please remove the mutex.
LGTM

src/server/search/search_family.cc
```diff
@@ -1292,6 +1294,66 @@ void SearchFamily::FtList(CmdArgList args, const CommandContext& cmd_cntx) {
   rb->SendBulkStrArr(names);
 }
 
+static vector<SearchResult> FtSearchCSS(std::string_view idx, std::string_view query,
+                                        std::string_view args_str) {
+  vector<SearchResult> results;
+  std::string cmd = absl::StrCat("FT.SEARCH ", idx, " ", query, " CSS ", args_str);
+
+  // TODO for now we suppose that callback is called synchronously. If not, we need to add
+  // synchronization here for results vector modification.
```
dranikpg: ah, ok, that answers my question

src/server/search/search_family.cc
```diff
@@ -1292,6 +1294,66 @@ void SearchFamily::FtList(CmdArgList args, const CommandContext& cmd_cntx) {
...
+    results.emplace_back();
+    results.back().total_hits = *size;
+    for (size_t i = 2; i < res.size(); i += 2) {
+      auto& search_doc = results.back().docs.emplace_back();
+      search_doc.key = res[i - 1].GetString();
+      if (res[i].type != facade::RespExpr::Type::ARRAY) {
+        LOG(ERROR) << "FT.SEARCH CSS reply unexpected type for document data: "
+                   << static_cast<int>(res[i].type);
+        return;
+      }
```
dranikpg, line 1331: just thought that you can create a macro for such type testing

src/server/search/search_family.cc
dranikpg: Actually they will meet even stricter requirements, it should be generally as easy as that

dranikpg (approved):

src/server/search/search_family.cc
BorysTheDev: As I wrote in description, I want something like CmdArgParser. But I need some more examples to create a good one. So a little later I will refactor this code

PR #6092, opened 2025-11-20
chore: change the resolution of reply latency to measurement to ns

romange:
Align test_reply_count to use /metrics for fetching the `reply_count` counter.

### DIFF
--- a/tests/dragonfly/connection_test.py
+++ b/tests/dragonfly/connection_test.py
@@ -607,21 +607,23 @@ async def test_keyspace_events_config_set(async_client: aioredis.Redis):
 
 
 @dfly_args({"max_busy_read_usec": 50000})
-async def test_reply_count(async_client: aioredis.Redis):
+async def test_reply_count(df_server: DflyInstance):
     """Make sure reply aggregations reduce reply counts for common cases"""
 
     async def get_reply_count():
-        return (await async_client.info("STATS"))["reply_count"]
+        metrics = await df_server.metrics()
+        return int(metrics["dragonfly_reply"].samples[0].value)
 
     async def measure(aw):
         before = await get_reply_count()
         await aw
-        return await get_reply_count() - before - 1
+        return await get_reply_count() - before
 
+    async_client = df_server.client()
     await async_client.config_resetstat()
     base = await get_reply_count()
     info_diff = await get_reply_count() - base
-    assert info_diff == 1
+    assert info_diff == 0  # no commands yet
 
     # Warm client buffer up
     await async_client.lpush("warmup", *(i for i in range(500)))
--- a/src/facade/reply_builder.cc
+++ b/src/facade/reply_builder.cc
@@ -181,7 +181,7 @@ void SinkReplyBuilder::Send() {
 
   uint64_t after_ns = util::fb2::ProactorBase::GetMonotonicTimeNs();
   reply_stats.send_stats.count++;
-  reply_stats.send_stats.total_duration += (after_ns - pin.timestamp_ns) / 1'000;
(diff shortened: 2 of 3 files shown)


tests/dragonfly/connection_test.py
```diff
@@ -610,58 +613,68 @@ async def test_keyspace_events_config_set(async_client: aioredis.Redis):
...
+            if metric.name == "dragonfly_reply":
+                assert metric.type == "counter"
+                return int(metric.samples[0].value)
+        return None
 
     async def measure(aw):
-        before = await get_reply_count()
-        await aw
-        return await get_reply_count() - before - 1
+        before = get_reply_count()
```
romange: since we grab the metrics from the http interface we do not need to substract 1 that counted the "INFO" call itself

tests/dragonfly/connection_test.py
```diff
@@ -610,58 +613,68 @@ async def test_keyspace_events_config_set(async_client: aioredis.Redis):
 async def test_reply_count(async_client: aioredis.Redis):
     """Make sure reply aggregations reduce reply counts for common cases"""
 
-    async def get_reply_count():
-        return (await async_client.info("STATS"))["reply_count"]
```
romange, line 614: before: the `info` call intervened with the test, now we scrape /metrics

tests/dragonfly/connection_test.py
```diff
@@ -610,58 +613,68 @@ async def test_keyspace_events_config_set(async_client: aioredis.Redis):
...
+                return int(metric.samples[0].value)
+        return None
 
     async def measure(aw):
-        before = await get_reply_count()
-        await aw
-        return await get_reply_count() - before - 1
+        before = get_reply_count()
+        aw
+        return get_reply_count() - before
```
dranikpg: > Also, fix test_reply_count that did not really tested commands due to missed await statements

This is a misunderstanding of the code. A non-awaited awaitable is like a functor that you can invoke later. So the code effectively was
```c++
void measure(function f) {
  before = Info()['replies']
  f()
  return Info()['replies'] - before - 1
}
measure([]{ lrange(...) });
``` 

what is line 632 with `aw` now? What does it do? Its just the result of some kind of object, it has no action. And the before and after calls are called one after another directly

romange: you are absolutely correct. I fixed it now.

romange: but there was a bug in measure() - it did not await on `aw`

tests/dragonfly/connection_test.py
```diff
@@ -610,18 +613,29 @@ async def test_keyspace_events_config_set(async_client: aioredis.Redis):
...
+
+    def get_reply_count():
+        response = requests.get(METRICS_URL)
+        response.raise_for_status()
+        families = text_string_to_metric_families(response.text)
+        for metric in families:
+            if metric.name == "dragonfly_reply":
+                assert metric.type == "counter"
+                return int(metric.samples[0].value)
+        return None
```
dranikpg: There is `DfInstance.metrics` that is even async, you just need to get the `df_server` fixture

tests/dragonfly/connection_test.py
romange: sorry, I am confused. there was no bug :(
It somehow got deleted in my client by accident. I see now there is no diff, I do not know what happened. in any case, the test is aligned now.

dranikpg: To me it looks like it always was `await aw`  😏

tests/dragonfly/connection_test.py
romange: I will fix.

Issue #6044, opened 2025-11-11
Add Shrink() to DenseSet


vyavdoshenko:
Implement shrink for sparse buckets

PR #6094, opened 2025-11-20
feat: implement DenseSet::Shrink 

vyavdoshenko:
Closes: https://github.com/dragonflydb/dragonfly/issues/6044

Add `DenseSet::Shrink()` to reduce the hash table size when sets/hashes have excess capacity after deletions.

- DenseSet::Shrink() - relocates elements to a smaller bucket array
- MEMORY SHRINK <key> - command to shrink a specific key

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -2643,6 +2643,7 @@ void ServerFamily::Shrink(CmdArgList args, const CommandContext& cmd_cntx) {
     }
 
     DenseSet* ds = static_cast<DenseSet*>(pv.RObjPtr());
+    ds->set_time(MemberTimeSeconds(t->GetDbContext().time_now_ms));
     size_t current_size = ds->UpperBoundSize();
     size_t bucket_count = ds->BucketCount();
 


src/core/dense_set.cc
```diff
@@ -373,22 +374,139 @@ auto DenseSet::FindEmptyAround(uint32_t bid) -> ChainVectorIterator {
...
+  // This also ensures PushFront's DCHECK(!it->IsDisplaced()) passes.
+  for (size_t i = 0; i < prev_size; ++i) {
+    if (!entries_[i].IsEmpty()) {
+      entries_[i].ClearDisplaced();
+    }
+  }
+
+  // Process from high to low. This mirrors Grow's iteration order.
+  // Elements move to lower buckets (not yet processed) based on their hash.
+  for (long i = prev_size - 1; i >= 0; --i) {
```
BorysTheDev: To my mind, it should be vice versa from 0 to size, because you move them into lower positions, to avoid double processing

src/core/string_set_test.cc
```diff
@@ -803,4 +803,148 @@ TEST_F(StringSetTest, TransferTTLFlagLinkToObjectOnDelete) {
...
+  }
+
+  EXPECT_EQ(ss_->UpperBoundSize(), initial_strs + additional_strs);
+
+  // Verify all elements are accessible
+  for (const auto& str : strs) {
+    EXPECT_TRUE(ss_->Contains(str)) << "Missing: " << str;
+  }
+}
+
```
BorysTheDev, line 955: I think you need sscan test during shrink

src/core/dense_set.cc
```diff
@@ -373,22 +374,139 @@ auto DenseSet::FindEmptyAround(uint32_t bid) -> ChainVectorIterator {
   return entries_.end();
 }
 
-void DenseSet::Reserve(size_t sz) {
+void DenseSet::Resize(size_t sz) {
```
BorysTheDev: The name is controversial because it describes some changes to the internal representation. Consider using AdjustCapacity()

src/core/dense_set.cc
```diff
@@ -373,22 +374,139 @@ auto DenseSet::FindEmptyAround(uint32_t bid) -> ChainVectorIterator {
...
+
+  // Guard against recursive shrink (e.g., if ExpireIfNeeded triggers Delete)
+  if (shrinking_) {
+    return;
+  }
+
+  // Check num_iterators_ to prevent shrink during iteration (would invalidate iterators)
+  if (num_iterators_ != 0) {
+    return;
+  }
```
BorysTheDev: it looks strange to me that shrinking depends on the iterators existing

src/core/dense_set.h
```diff
@@ -181,6 +181,39 @@ class DenseSet {
...
+      }
+    }
+
+    // Copy constructor - increment iterator count
+    IteratorBase(const IteratorBase& other)
+        : owner_(other.owner_), curr_list_(other.curr_list_), curr_entry_(other.curr_entry_) {
+      if (owner_) {
+        ++owner_->num_iterators_;
+      }
+    }
```
dranikpg: You can call upper base constructor

src/core/dense_set.cc
dranikpg: I usually like higher level safeguards. On the other hand, having one like this, assumes that this is valid or possible behaviour, and even worse, if it occurs, it will happen unnoticed, because the function just silently returns

dranikpg: Actually, external ExpireIfNeeded calling Delete and then calling this function looks really dangerous. It holds references, the bucket id, etc. It also never holds an iterator directly, so none of your safeguards will kick in

dranikpg: Moreover the safeguards, can be hurting as well, consider:
```c++
for (auto it: set) {
  set->Delete(it)
}
```
Shrink won't be called even once even if the usage drops to 0%

src/core/string_set_test.cc
```diff
@@ -803,4 +803,148 @@ TEST_F(StringSetTest, TransferTTLFlagLinkToObjectOnDelete) {
...
+TEST_F(StringSetTest, ShrinkWithTTL) {
+  // Add elements with TTL
+  constexpr size_t num_strs = 16;
+  vector<string> strs;
+  for (size_t i = 0; i < num_strs; ++i) {
+    strs.push_back(random_string(generator_, 10));
+    EXPECT_TRUE(ss_->Add(strs.back(), 100));  // TTL of 100
+  }
```
dranikpg, line 866: We need a test where items are actively expiring while you call set operations, even most basic ones like Find(), causing the set to shrink

dranikpg: If I grep `ExpireIfNeeded` there are like 20 matches in dense_set.cc, and every of them might horribly crash if it calls Delete() and Shink() inside. The test should have at least thousands of elements and calling most basic code paths

dranikpg: this is a really dangerous change

BorysTheDev:
@dranikpg I've already discussed with @vyavdoshenko all these problems; he rewrites all this logic.

src/core/dense_set.cc
vyavdoshenko: Fixed

src/core/string_set_test.cc
vyavdoshenko: Added

src/core/dense_set.cc
vyavdoshenko: Not relevant any longer

src/core/dense_set.cc
vyavdoshenko: Fixed

src/core/dense_set.h
vyavdoshenko: Removed
Not relevant any longer

src/core/dense_set.cc
vyavdoshenko: Not relevant any longer

src/core/string_set_test.cc
vyavdoshenko: I removed Shrink() from Delete()

src/core/dense_set.cc
```diff
@@ -375,6 +375,7 @@ auto DenseSet::FindEmptyAround(uint32_t bid) -> ChainVectorIterator {
 
 void DenseSet::Reserve(size_t sz) {
   sz = std::max<size_t>(sz, kMinSize);
+  sz = std::max<size_t>(sz, size_);
```
BorysTheDev: why do you need this line

src/core/dense_set.cc
```diff
@@ -385,6 +386,110 @@ void DenseSet::Reserve(size_t sz) {
...
+  shrink_generation_ = (shrink_generation_ + 1) & 0x0F;
+
+  // Clear all displaced flags first. During shrinking, we'll recalculate all bucket
+  // positions based on hash, so the displaced markers are no longer valid.
+  // This also ensures PushFront's DCHECK(!it->IsDisplaced()) passes.
+  for (size_t i = 0; i < prev_size; ++i) {
+    if (!entries_[i].IsEmpty()) {
+      entries_[i].ClearDisplaced();
+    }
+  }
```
BorysTheDev: we can do it in one run

src/core/dense_set.cc
vyavdoshenko: Fixed

src/server/memory_cmd.cc
```diff
@@ -139,6 +140,10 @@ void MemoryCmd::Run(CmdArgList args) {
         "page.",
         "    Pages used less than the threshold percentage (default 0.8) are targeted for moving "
         "out data.",
+        "SHRINK",
+        "    Shrinks DenseSet-based data structures (sets, hashes, sorted sets) to optimal size.",
```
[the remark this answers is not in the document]
vyavdoshenko: Fixed

src/core/dense_set.cc
BorysTheDev: also you do it below

vyavdoshenko: Fixed

src/server/memory_cmd.cc
```diff
@@ -475,4 +484,89 @@ void MemoryCmd::Track(CmdArgList args) {
...
+        // Process DenseSet-based structures (StringSet and StringMap)
+        if (encoding == kEncodingStrMap2 && (obj_type == OBJ_SET || obj_type == OBJ_HASH)) {
+          DenseSet* ds = static_cast<DenseSet*>(pv.RObjPtr());
+          size_t current_size = ds->UpperBoundSize();
+          size_t bucket_count = ds->BucketCount();
+
+          if (current_size > 0 && bucket_count > 0) {
+            size_t optimal_size = std::max(size_t(8), absl::bit_ceil(current_size));
+            if (optimal_size < bucket_count) {
+              size_t bucket_bytes_before = ds->BucketCount() * 8;  // sizeof(DensePtr) = 8
```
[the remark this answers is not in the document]
vyavdoshenko: Fixed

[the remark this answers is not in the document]
kostasrim: We do not support 32 bit btw :)

src/server/memory_cmd.cc
```diff
@@ -475,4 +484,89 @@ void MemoryCmd::Track(CmdArgList args) {
   return builder_->SendError(kSyntaxErrType);
 }
 
+void MemoryCmd::Shrink() {
```
romange: sorry I meant something much less intrusive.  `DEBUG SHRINK <key>` or `MEMORY SHRINK <key>` and not go over all the key space as it can be very computationally intensive and cause unresponsiveness for the datastore

src/server/memory_cmd.cc
```diff
@@ -475,4 +484,89 @@ void MemoryCmd::Track(CmdArgList args) {
...
+  size_t total_bytes_saved = 0;
+
+  for (const auto& stats : shard_stats) {
+    total_examined += stats.total_examined;
+    total_shrunk += stats.shrunk_objects;
+    total_bytes_saved += stats.bytes_saved;
+  }
+
+  // Build detailed response
+  string response;
```
romange: it's great to provide such feedback but for a single key it will become much simpler. if we want to shrink multiple keys, we will still be able to do by writing an external script with scan and shrink

src/core/dense_set.h
```diff
@@ -414,6 +420,7 @@ class DenseSet {
   mutable uint32_t num_links_ = 0;         // number of links in the set.
   mutable uint32_t num_used_buckets_ = 0;  // number of buckets used in entries_ array.
   unsigned capacity_log_ = 0;
+  uint8_t shrink_generation_ = 0;  // Incremented on each Shrink, used for cursor stability (4 bits)
```
romange: I think cursor stability should work out of the box without `shrink_generation_`

src/core/dense_set.cc
```diff
@@ -385,6 +385,101 @@ void DenseSet::Reserve(size_t sz) {
...
+  // bucket j has already been processed, so the element won't be processed again.
+  for (size_t i = 0; i < prev_size; ++i) {
+    DensePtr* curr = &entries_[i];
+    DensePtr* prev = nullptr;
+
+    do {
+      if (ExpireIfNeeded(prev, curr)) {
+        // If curr has disappeared due to expiry and prev was converted from Link
+        // to a regular DensePtr
+        if (prev && !prev->IsLink())
```
romange: I am trying to understand the logic here.
Say, the chain is A -> B, and B disappeared. So when you break, you keep `A` at the same place. Maybe it should move due to shrinkage?

src/core/dense_set.cc
```diff
@@ -385,6 +385,101 @@ void DenseSet::Reserve(size_t sz) {
...
+      uint32_t new_bid = BucketId(ptr, 0);
+
+      // If the item stays in the current bucket, ensure it is not marked as
+      // displaced and move to the next item in the chain
+      if (new_bid == i) {
+        curr->ClearDisplaced();
+        prev = curr;
+        curr = curr->Next();
+        if (curr == nullptr)
+          break;
```
romange: redundant as while below handles it?

src/core/dense_set.cc
```diff
@@ -385,6 +385,101 @@ void DenseSet::Reserve(size_t sz) {
...
+            // We deallocated the link, curr is invalid now.
+            curr = nullptr;
+            *prev = tmp;
+          } else {
+            // prev == nullptr
+            curr->Reset();  // Reset the root placeholder.
+          }
+        } else {
+          // !curr.IsObject
+          *curr = *dptr.Next();
```
romange: I am trying to understand what happens here.  `curr` is DenseLinkKey. 
What happen here? specifically, could it be that we override `*curr` with object, and loose knowledge that it was DenseLinkKey ?  Maybe I am wrong, the code is complicated, unfortunately. 
@BorysTheDev  please take a look as well.

src/core/dense_set.cc
```diff
@@ -787,7 +892,17 @@ uint32_t DenseSet::Scan(uint32_t cursor, const ItemCb& cb) const {
     return 0;
   }
 
-  uint32_t entries_idx = cursor >> (32 - capacity_log_);
+  // Decode cursor: [4 bits generation | 28 bits bucket position]
+  uint8_t cursor_generation = (cursor >> 28) & 0x0F;
+  uint32_t cursor_bits = cursor & 0x0FFFFFFF;  // Lower 28 bits
+  uint32_t entries_idx = cursor_bits >> (28 - capacity_log_);
```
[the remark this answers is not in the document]
vyavdoshenko: Fixed

src/server/memory_cmd.cc
vyavdoshenko: Fixed

src/server/memory_cmd.cc
vyavdoshenko: Fixed

src/core/dense_set.cc
vyavdoshenko: Yes, you are right.
Fixed.

src/core/dense_set.cc
vyavdoshenko: Fixed

src/core/dense_set.cc
```diff
@@ -787,7 +882,17 @@ uint32_t DenseSet::Scan(uint32_t cursor, const ItemCb& cb) const {
...
```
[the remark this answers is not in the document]
vyavdoshenko: Fixed

src/core/dense_set.cc
```diff
@@ -847,7 +952,9 @@ uint32_t DenseSet::Scan(uint32_t cursor, const ItemCb& cb) const {
     cb(right_bucket->GetObject());
   }
 
-  return entries_idx << (32 - capacity_log_);
+  // Encode cursor: [4 bits generation | 28 bits bucket position]
```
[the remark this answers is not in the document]
vyavdoshenko: Fixed

src/core/dense_set.cc
```diff
@@ -385,6 +385,101 @@ void DenseSet::Reserve(size_t sz) {
...
 
+void DenseSet::Shrink(size_t new_size) {
+  DCHECK(absl::has_single_bit(new_size));
+  DCHECK_GE(new_size, kMinSize);
+  DCHECK_LT(new_size, entries_.size());
+
+  size_t prev_size = entries_.size();
+  capacity_log_ = absl::bit_width(new_size) - 1;
+
+  // Increment shrink generation (use only 4 bits, so wrap at 16)
```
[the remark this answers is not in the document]
vyavdoshenko: Fixed

src/core/dense_set.h
vyavdoshenko: Cursor is encoded with old capacity_log_. After shrinking, decoding with new capacity_log_ gives a different bucket position. The cursor becomes invalid because it represents coordinates in the old table layout.

Without shrink_generation_, we'd continue from a bucket position that doesn't correctly represent our scan progress. Elements that we haven't scanned yet might end up at bucket positions we've "already passed" according to the new decoded cursor.

shrink_generation_ is a simple mechanism: if the table structure changed, restart from 0. This guarantees correctness (no missed elements) with acceptable duplicates per SCAN semantics.

src/core/dense_set.cc
```diff
@@ -381,10 +381,108 @@ void DenseSet::Reserve(size_t sz) {
...
+          *curr = *dptr.Next();
+          DCHECK(!curr->IsEmpty());
+        }
+
+        DVLOG(2) << " Shrink: Moving from " << i << " to " << new_bid;
+        dptr.ClearDisplaced();
+        PushFront(dest, dptr);
+      }
+    } while (curr);
+  }
```
BorysTheDev: Could you split this code into several methods?

src/core/dense_set.cc
```diff
@@ -381,10 +381,108 @@ void DenseSet::Reserve(size_t sz) {
...
+
+  // Recount used buckets after shrinking
+  num_used_buckets_ = 0;
+  for (size_t i = 0; i < new_size; ++i) {
+    if (!entries_[i].IsEmpty()) {
+      ++num_used_buckets_;
+    }
+  }
```
BorysTheDev: I think we should do it in one run

src/core/dense_set.cc
vyavdoshenko: OK, I will do

src/core/dense_set.h
romange: 1. it's true that new bucket position is differrent. it should be smaller.
2. Still the guarantees are the same - no missed elements.  What will happen is that we may cover same elements twice - and that's ok. 

Let me know if what I say does not hold and we do miss elements without `shrink_generation_`.

vyavdoshenko: We miss elements without `shrink_generation_` because elements can move behind the cursor during shrinking.

romange: So cursor keeps its bucket ids in MSB part. Suppose bucket_id is 5, in a table of length 8. We keep 3 bits 0b101=5 in the cursor. bucket_id=5 it means that scan already covered all the items before shrinkage in buckets 0-4.

When we shrink the table to length 4, the bucket_id of the cursor will be 0b10=2. For an element to be missed, it should be moved to buckets 0,1.  How an element from bucket ids 6-8 can be moved to buckets 0-1?

Of course the same logic can be applied to a table of any size. @vyavdoshenko  I need a counter example.

src/core/dense_set.cc
BorysTheDev: yes it looks like you are right we loose information about DenseLinkKey

src/core/dense_set.cc
vyavdoshenko: Fixed

src/core/dense_set.cc
vyavdoshenko: Fixed

src/core/dense_set.h
vyavdoshenko: Yes, you are right. This was my mistake in understanding.
It will never put an element behind the cursor. We can only get duplicates, but it's ok.
I will fix it, shrink_generation_ is redundant.

src/core/dense_set.cc
vyavdoshenko: Implementation has been changed a little bit, but this code still exists.

DenseLinkKey is not lost - it's saved in `dptr` before overwriting `*curr`.

`dptr` is then passed to `PushFront`, which either reuses the DenseLinkKey (if dest has elements) or frees it via `FreeLink()` (if dest is empty).

src/core/dense_set.cc
```diff
@@ -386,6 +386,106 @@ void DenseSet::Reserve(size_t sz) {
...
+      DCHECK(&plink->next == curr);
+
+      // Convert prev from Link to Object and deallocate the link.
+      DensePtr tmp = DensePtr::From(plink);
+      tmp.SetTtl(prev->HasTtl());
+      DCHECK(ObjectAllocSize(tmp.GetObject()));
+
+      FreeLink(plink);
+      curr = nullptr;  // Link deallocated, stop iteration
+      *prev = tmp;
```
BorysTheDev: lets discuss

src/core/dense_set.cc
```diff
@@ -386,6 +386,106 @@ void DenseSet::Reserve(size_t sz) {
...
+  if (curr->IsObject()) {
+    if (prev) {
+      // prev is Link, curr is &prev->next
+      DCHECK(prev->IsLink());
+      DenseLinkKey* plink = prev->AsLink();
+      DCHECK(&plink->next == curr);
+
+      // Convert prev from Link to Object and deallocate the link.
+      DensePtr tmp = DensePtr::From(plink);
+      tmp.SetTtl(prev->HasTtl());
```
BorysTheDev: maybe we don't need this

src/core/dense_set.cc
```diff
@@ -386,6 +386,106 @@ void DenseSet::Reserve(size_t sz) {
...
+      FreeLink(plink);
+      curr = nullptr;  // Link deallocated, stop iteration
+      *prev = tmp;
+    } else {
+      // curr is root of bucket, just clear it
+      curr->Reset();
+    }
+  } else {
+    // curr is Link - copy next element to curr position
+    *curr = *dptr.Next();
```
BorysTheDev: lets discuss

src/core/dense_set.cc
```diff
@@ -386,6 +386,106 @@ void DenseSet::Reserve(size_t sz) {
...
+    DCHECK(!curr->IsEmpty());
+    // curr pointer stays the same, now contains next element
+  }
+
+  return dptr;
+}
+
+void DenseSet::ShrinkBucket(size_t bucket_idx) {
+  DensePtr* curr = &entries_[bucket_idx];
```
BorysTheDev: to my mind the better option to take out the whole bucket and then just add all elements in the correct positions. it will be safer

src/core/dense_set.cc
```diff
@@ -386,6 +386,106 @@ void DenseSet::Reserve(size_t sz) {
...
+  do {
+    if (ExpireIfNeeded(prev, curr)) {
+      // If curr has disappeared due to expiry and prev was converted from Link
+      // to a regular DensePtr, re-process prev to check if it needs to move.
+      if (prev && !prev->IsLink()) {
+        curr = prev;
+        prev = nullptr;
+        continue;
+      }
+    }
```
BorysTheDev: this code modifies the denseset, in your case you don't need to modify it , you can just drop this element and deallocate it

src/core/dense_set.cc
```diff
@@ -386,6 +386,106 @@ void DenseSet::Reserve(size_t sz) {
...
+      curr->ClearDisplaced();
+      prev = curr;
+      curr = curr->Next();
+    } else {
+      // Element needs to move to a different bucket
+      DensePtr dptr = ExtractFromChain(prev, curr);
+      DVLOG(2) << " Shrink: Moving from " << bucket_idx << " to " << new_bid;
+      dptr.ClearDisplaced();
+      PushFront(entries_.begin() + new_bid, dptr);
+    }
```
BorysTheDev: this code contains a lot of useless logic, it can be unified

src/core/dense_set.cc
```diff
@@ -386,6 +386,106 @@ void DenseSet::Reserve(size_t sz) {
...
+    bool was_empty = entries_[i].IsEmpty();
+    ShrinkBucket(i);
+    bool is_empty = entries_[i].IsEmpty();
+
+    // Track bucket state changes to maintain num_used_buckets_
+    if (!was_empty && is_empty) {
+      --num_used_buckets_;
+    } else if (was_empty && !is_empty) {
+      ++num_used_buckets_;
+    }
```
BorysTheDev: this code shouldn't be here, it should be in the remove/add logic

src/core/string_set_test.cc
```diff
@@ -803,4 +803,113 @@ TEST_F(StringSetTest, TransferTTLFlagLinkToObjectOnDelete) {
...
+
+  // Grow to a larger size
+  ss_->Reserve(256);
+  size_t original_bucket_count = ss_->BucketCount();
+  EXPECT_EQ(original_bucket_count, 256u);
+
+  // Shrink to half the size using Shrink
+  size_t new_size = original_bucket_count / 2;
+  ss_->Shrink(new_size);
```
BorysTheDev: I want this test with different shrink args, for example with num_strs / 2

src/core/string_set_test.cc
```diff
@@ -803,4 +803,113 @@ TEST_F(StringSetTest, TransferTTLFlagLinkToObjectOnDelete) {
...
+  }
+}
+
+TEST_F(StringSetTest, ShrinkWithTTL) {
+  // Add elements with TTL
+  constexpr size_t num_strs = 16;
+  vector<string> strs;
+  for (size_t i = 0; i < num_strs; ++i) {
+    strs.push_back(random_string(generator_, 10));
+    EXPECT_TRUE(ss_->Add(strs.back(), 100));  // TTL of 100
```
BorysTheDev: we need more flexible test,  TTL should be from 1 to 100, and some elements without ttl.
before shring you should set time 50 and check that all elements with time <= 50 are removed

src/core/string_set_test.cc
```diff
@@ -803,4 +803,113 @@ TEST_F(StringSetTest, TransferTTLFlagLinkToObjectOnDelete) {
   EXPECT_EQ(1u, it.ExpiryTime());
 }
 
+TEST_F(StringSetTest, BasicShrink) {
+  // Add elements and then grow to have extra capacity
+  constexpr size_t num_strs = 32;
```
BorysTheDev: for all tests please use up to 1M elements.

src/core/string_set_test.cc
```diff
@@ -803,4 +803,113 @@ TEST_F(StringSetTest, TransferTTLFlagLinkToObjectOnDelete) {
...
+
+  // Grow to large size first
+  ss_->Reserve(512);
+  EXPECT_EQ(ss_->BucketCount(), 512u);
+
+  unordered_set<string> seen;
+  auto scan_callback = [&](const sds ptr) {
+    string str{ptr, sdslen(ptr)};
+    seen.insert(str);
+  };
```
BorysTheDev: I like this test, but you can improve it by starting the first scan before the Grow operation

src/server/memory_cmd.cc
```diff
@@ -193,6 +197,11 @@ void MemoryCmd::Run(CmdArgList args) {
     return rb->SendVerbatimString(merged.ToString());
   }
 
+  if (parser.Check("SHRINK") && args.size() > 1) {
```
BorysTheDev:
```
if (string_view key; parser.Check("SHRINK", &key)) {
    return Shrink(key);
}
```

src/server/memory_cmd.cc
```diff
@@ -475,4 +484,66 @@ void MemoryCmd::Track(CmdArgList args) {
   return builder_->SendError(kSyntaxErrType);
 }
 
+void MemoryCmd::Shrink(string_view key) {
+  ShardId sid = Shard(key, shard_set->size());
+
+  auto* rb = static_cast<RedisReplyBuilder*>(builder_);
+
+  auto result = shard_set->pool()->at(sid)->AwaitBrief([key, this, sid]() -> tuple<int, size_t> {
```
BorysTheDev: @dranikpg Could you check this code. I thought if we want to operate with key we should have a transaction where this key is registered. We don't deduce keys for MEMORY command so maybe we should create a transaction manually?

src/server/memory_cmd.cc
```diff
@@ -475,4 +484,66 @@ void MemoryCmd::Track(CmdArgList args) {
...
+    return rb->SendNull();  // Key not found
+  }
+
+  if (status == -2) {
+    return rb->SendError("WRONGTYPE Key is not a set or hash with DenseSet encoding");
+  }
+
+  if (status == 0) {
+    return rb->SendLong(0);  // Nothing to shrink
+  }
```
BorysTheDev: please use enum for status and switch/case for processing

src/core/dense_set.cc
vyavdoshenko: Fixed

src/core/dense_set.cc
vyavdoshenko: Fixed

src/core/dense_set.cc
vyavdoshenko: Fixed

src/core/dense_set.cc
vyavdoshenko: Fixed

src/core/dense_set.cc
vyavdoshenko: Fixed

src/core/string_set_test.cc
vyavdoshenko: Fixed

src/core/string_set_test.cc
vyavdoshenko: Fixed

src/core/string_set_test.cc
vyavdoshenko: Fixed

src/core/string_set_test.cc
vyavdoshenko: Fixed

src/server/memory_cmd.cc
vyavdoshenko: Fixed

src/server/memory_cmd.cc
vyavdoshenko: Fixed

src/server/memory_cmd.cc
vyavdoshenko: I added the transaction code and moved this code to server family.

src/core/dense_set.cc
```diff
@@ -386,6 +386,58 @@ void DenseSet::Reserve(size_t sz) {
   }
 }
 
+void DenseSet::ShrinkBucket(size_t bucket_idx) {
+  // Take the entire bucket to avoid infinite loop when new_bid == bucket_idx
+  DensePtr bucket = entries_[bucket_idx];
+  entries_[bucket_idx].Reset();
```
BorysTheDev, line 392: you forget     --num_used_buckets_;

src/core/dense_set.cc
```diff
@@ -386,6 +386,58 @@ void DenseSet::Reserve(size_t sz) {
...
+
+  // Process the taken bucket chain
+  while (!bucket.IsEmpty()) {
+    // Pop front from local chain
+    DensePtr dptr = bucket;
+    if (bucket.IsObject()) {
+      bucket.Reset();
+    } else {
+      bucket = bucket.AsLink()->next;
```
BorysTheDev: could it be written simpler? something like  `bucket = bucket.IsObject() ? nullptr : bucket.AsLink()->next;`

src/core/dense_set.cc
```diff
@@ -386,6 +386,58 @@ void DenseSet::Reserve(size_t sz) {
...
+    }
+
+    void* obj = dptr.GetObject();
+
+    // Check expiry
```
BorysTheDev: useless comment

src/core/dense_set.cc
```diff
@@ -386,6 +386,58 @@ void DenseSet::Reserve(size_t sz) {
...
+    if (dptr.HasTtl() && ObjExpireTime(obj) <= time_now_) {
+      ObjDelete(obj, true);
+      if (dptr.IsLink()) {
+        FreeLink(dptr.AsLink());
+      }
```
BorysTheDev: I think you should drop this link unconditionally even with expiration

src/core/dense_set.cc
```diff
@@ -386,6 +386,58 @@ void DenseSet::Reserve(size_t sz) {
...
+      --size_;
+      continue;
+    }
+
+    uint32_t new_bid = BucketId(obj, 0);
+    DVLOG(2) << " Shrink: Moving from " << bucket_idx << " to " << new_bid;
+    dptr.ClearDisplaced();
+    PushFront(entries_.begin() + new_bid, dptr);
```
BorysTheDev: I think you should provide only object here but dptr can be a link

src/core/dense_set.cc
vyavdoshenko: Fixed

src/core/dense_set.cc
vyavdoshenko: Fixed

src/core/dense_set.cc
vyavdoshenko: Fixed

src/core/dense_set.cc
vyavdoshenko: Fixed

src/core/dense_set.cc
vyavdoshenko: Fixed

src/core/dense_set.cc
```diff
@@ -386,6 +386,66 @@ void DenseSet::Reserve(size_t sz) {
...
+    uint32_t new_bid = BucketId(obj, 0);
+    DVLOG(2) << " Shrink: Moving from " << bucket_idx << " to " << new_bid;
+    bool was_empty = entries_[new_bid].IsEmpty();
+
+    DensePtr obj_ptr;
+    obj_ptr.SetObject(obj);
+    if (has_ttl) {
+      obj_ptr.SetTtl(true);
+    }
+    PushFront(entries_.begin() + new_bid, obj_ptr);
```
BorysTheDev: I think you need AddUnique

vyavdoshenko: I don't think so. It calls Grow, increase size_, calls SetDisplaced. The last one is suitable for adding elements, but not for moving ones.

src/core/dense_set.cc
```diff
@@ -386,6 +386,60 @@ void DenseSet::Reserve(size_t sz) {
...
+      --size_;
+      continue;
+    }
+
+    uint32_t new_bid = BucketId(obj, 0);
+    DVLOG(2) << " Shrink: Moving from " << bucket_idx << " to " << new_bid;
+    bool was_empty = entries_[new_bid].IsEmpty();
+    PushFront(entries_.begin() + new_bid, obj, has_ttl);
+    if (was_empty) {
+      ++num_used_buckets_;
```
BorysTheDev: num_used_buckets_ += entries_[new_bid].IsEmpty();

vyavdoshenko: fixed

src/core/dense_set.cc
```diff
@@ -386,6 +386,57 @@ void DenseSet::Reserve(size_t sz) {
...
+    // Free link unconditionally - PushFront will create new one if needed
+    if (dptr.IsLink()) {
+      FreeLink(dptr.AsLink());
+    }
+
+    if (has_ttl && ObjExpireTime(obj) <= time_now_) {
+      ObjDelete(obj, true);
+      --size_;
+      continue;
+    }
```
dranikpg, line 413: this will currently never happen because you never call `set_time()` from the `SHRINK` command, you can look at `container_utils::GetStringMap` for an example

src/server/server_family.cc
```diff
@@ -4219,6 +4275,7 @@ void ServerFamily::Register(CommandRegistry* registry) {
       << CI{"LATENCY", CO::NOSCRIPT | CO::LOADING | CO::FAST, -2, 0, 0, acl::kLatency}.HFUNC(
              Latency)
       << CI{"MEMORY", kMemOpts, -2, 0, 0, acl::kMemory}.HFUNC(Memory)
+      << CI{"SHRINK", CO::WRITE | CO::FAST, 2, 1, 1, acl::kMemory}.HFUNC(Shrink)
```
dranikpg, line 4279: haven't seen what it was before, but I assumed it has to be part of MEMORY. It won't be as easy to a transaction then, because you'll need to create one manually

src/core/dense_set.cc
```diff
@@ -386,6 +386,57 @@ void DenseSet::Reserve(size_t sz) {
...
+
+    uint32_t new_bid = BucketId(obj, 0);
```
dranikpg, line 415: based on the hash functon, it looks like `2*i`  and `i` are mapped to -> `i`. Not sure if its worth dcheck'ing or commenting, because the code looks more generic than it already is

src/server/server_family.cc
vyavdoshenko: Yes, it was originally MEMORY SHRINK. However, to invoke the code within a transaction context, I needed more changes than just extracting it to a separate function. The MEMORY command uses the MemoryCmd class, which doesn't have access to CommandContext with transaction support. To properly wrap the shrink operation in a transaction (as you suggested earlier for safety), I had two options:
- Refactor MemoryCmd to support transactions, which would require substantial changes to the existing command infrastructure
- Create a separate SHRINK command with proper transaction support from the start

I chose the second option as it was cleaner and required fewer changes to the existing code.

src/core/dense_set.cc
vyavdoshenko: fixed

src/core/dense_set.cc
vyavdoshenko: OK, I didn't get the direction. Should we add DCHECK?

PR #6099, opened 2025-11-21
fix: test_reply_count regression

### LINKED ISSUE #6098: test_reply_count failed
See https://github.com/dragonflydb/dragonfly/actions/runs/19569973809/job/56040905917

romange:
Fixes #6098

### DIFF
--- a/tests/dragonfly/connection_test.py
+++ b/tests/dragonfly/connection_test.py
@@ -635,11 +635,11 @@ async def test_reply_count(df_server: DflyInstance):
 
     # Integer set
     await async_client.sadd("set-1", *(i for i in range(100)))
-    assert await measure(async_client.smembers("set-1")) == 1
+    assert await measure(async_client.smembers("set-1")) <= 2
 
     # Sorted sets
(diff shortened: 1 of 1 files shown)


tests/dragonfly/connection_test.py
```diff
@@ -635,11 +635,11 @@ async def measure(aw):
 
     # Integer set
     await async_client.sadd("set-1", *(i for i in range(100)))
-    assert await measure(async_client.smembers("set-1")) == 1
+    assert await measure(async_client.smembers("set-1")) <= 2
```
glevkovich, line 638: Please consider adding a brief explanation in the commit message  why the test assertions are now less strict (using <= 2 instead of == 1). This will help reviewers understand the rationale behind allowing more than one reply.

glevkovich (approved):
Approved with a comment

tests/dragonfly/connection_test.py
romange: it's not more than one reply. maybe reply is not a good name.
it's a metric that tracks how many times we had to flush data to socket when rendering a reply.
please note that it's a regression fix, where I changed the test before and incidentally made it more strict

PR #6102, opened 2025-11-24
fix: smashing stack on Huffman test

vyavdoshenko:
Fix smashing stack. I uncover it locally on my ARM in Docker build.

Huffman test abort:
```
*** stack smashing detected ***: terminated
*** SIGABRT received at time=1763983063 on cpu 3 ***
PC: @     0xffff9bd17608  (unknown)  (unknown)
    @     0xaaaabd230a5c         64  absl::lts_20250512::WriteFailureInfo()
    @     0xaaaabd230d74         96  absl::lts_20250512::AbslFailureSignalHandler()
```

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -861,9 +861,9 @@ uint64_t CompactObj::HashCode() const {
   DCHECK(mask_bits_.encoding);
 
   if (IsInline()) {
-    // Buffer size must handle worst-case Huffman expansion
-    // With max_symbol=255, expansion can be up to 8x (worst case)
-    // kInlineLen * 8 ensures we never overflow
+    // Buffer must accommodate maximum decompressed size from inline storage
+    // Highly compressible data can achieve ~8x compression (e.g., repeated character)
+    // kInlineLen (16 bytes) compressed -> up to 128 bytes decompressed
     char buf[kInlineLen * 8];
     size_t decoded_len = GetStrEncoding().Decode(string_view{u_.inline_str, taglen_}, buf);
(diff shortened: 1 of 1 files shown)


src/core/compact_object.cc
```diff
@@ -861,7 +861,10 @@ uint64_t CompactObj::HashCode() const {
   DCHECK(mask_bits_.encoding);
 
   if (IsInline()) {
-    char buf[kInlineLen * 3];  // should suffice for most huffman decodings.
+    // Buffer size must handle worst-case Huffman expansion
+    // With max_symbol=255, expansion can be up to 8x (worst case)
```
romange: did you mean `max_symbol=1` ? i.e. when we encode a single char to a single bit?

romange:
Nice. do you have a reproducible example of that? how is that possible to have x8 expansion?

src/core/compact_object.cc
vyavdoshenko: The comment was misleading. Fixed in both locations. Why 8x?

Inline storage: 16 bytes (compressed)
Buffer needed: 128 bytes (decompressed)
Ratio: 128/16 = 8x
When is 8x achieved? With highly compressible data (e.g., repeated characters). Example:
string(100, 'a') with hist['a']=100
Symbol 'a' encodes to ~1-2 bits
100 bytes -> ~13 bytes compressed -> fits in 16 bytes inline

vyavdoshenko:
> Nice. do you have a reproducible example of that? how is that possible to have x8 expansion?

Huffman test uncovers this issue. I tried to make a separate test, but I couldn't.

PR #6103, opened 2025-11-24
feat(search): Global HNSW vector index

mkaruza:
* Third-party library hnswlib now uses dragonfly forked project where
  internal id is changed to uint64_t to support GlobalDocId.
* Singleton GlobalHnswIndexRegistry class is used now to
  create/remove/execute hnsw index functionality.
* Implemented function SearchGlobalHnswIndex that can be used to search
  hnsw index with or without prefilter query.

### DIFF
--- a/src/core/search/search.cc
+++ b/src/core/search/search.cc
@@ -687,9 +687,10 @@ std::unique_ptr<AstNode> SearchAlgorithm::PopKnnNode() {
   if (auto* knn = get_if<AstKnnNode>(query_.get()); knn) {
     // Save knn score sort option
     knn_hnsw_score_sort_option_ = KnnScoreSortOption{string_view{knn->score_alias}, knn->limit};
-    if (!std::holds_alternative<AstStarNode>(*(knn)->filter))
-      query_.swap(knn->filter);
     auto node = std::move(query_);
+    AstKnnNode* moved_knn_node = reinterpret_cast<AstKnnNode*>(node.get());
+    if (!std::holds_alternative<AstStarNode>(*moved_knn_node->filter))
+      query_.swap(moved_knn_node->filter);
     return node;
   }
   LOG(DFATAL) << "Should not reach here";
--- a/src/server/search/search_family_test.cc
+++ b/src/server/search/search_family_test.cc
@@ -3560,9 +3560,9 @@ TEST_F(SearchFamilyTest, KnnHnsw) {
   };
 
   // Add some test documents with vector data
-  // Run({"HSET", "doc1", "even", "yes", "pos", FloatToBytes(1.0f)});
-  // Run({"HSET", "doc2", "even", "no", "pos", FloatToBytes(2.0f)});
-  // Run({"HSET", "doc3", "even", "yes", "pos", FloatToBytes(3.0f)});
+  Run({"HSET", "doc1", "even", "yes", "pos", FloatToBytes(1.0f)});
+  Run({"HSET", "doc2", "even", "no", "pos", FloatToBytes(2.0f)});
+  Run({"HSET", "doc3", "even", "yes", "pos", FloatToBytes(3.0f)});
 
   // Query vector (2.0f - should find doc2 closest, but filtered to "yes" docs)
   string query_vec = FloatToBytes(2.0f);


src/core/search/ast_expr.cc
```diff
@@ -73,6 +73,10 @@ AstKnnNode::AstKnnNode(AstNode&& filter, AstKnnNode&& self) {
   this->filter = make_unique<AstNode>(std::move(filter));
 }
 
+bool AstKnnNode::PreFilter() const {
```
romange: `PreFilter` is a bad name, imho. Maybe `HasPreFilter` and then revert the condition `filter != nullptr`

src/core/search/base.h
```diff
@@ -79,16 +79,16 @@ struct DocumentAccessor {
 //
 // Queries should be done directly on subclasses with their distinc
 // query functions. All results for all index types should be sorted.
-struct BaseIndex {
+template <typename T> struct BaseIndex {
```
dranikpg: If you do all of this because you want to use the hnsw index with different types, it seems like it would be easier to branch off somewhere there. Maybe create a templated "base hnsw index" type and a subtype that implements BaseIndex (if we need it at all). Either way it looks simpler than changing everythign up the chain

mkaruza: We have `BaseIndex` that has virtual functions

```
  // Returns true if the document was added / indexed
  virtual bool Add(T id, const DocumentAccessor& doc, std::string_view field) = 0;
  virtual void Remove(T id, const DocumentAccessor& doc, std::string_view field) = 0;

  // Returns documents that have non-null values for this field (used for @field:* queries)
  // Result must be sorted
  virtual std::vector<T> GetAllDocsWithNonNullValues() const = 0;
```

BaseVectorIndex is derived class and implements `Add` as final function.

There are 2 derived classes from `BaseVectorIndex`. FlatVectorIndex that uses DocId  and HnswVectorIndex uses GlobalDocId.

FlatVectorIndex is stored in shard so it is stored in indices as base class of `BaseIndex`. We don't do this for HnswVectorIndex. 

What could be possible is to just write `HnswVectorIndex` without any base class and implement all function that are used - wdyt ? @dranikpg

dranikpg: `BaseVectorIndex` is just a helper class that provides:
* info lookup about dimension etc
* a conversion function from document to vector

You can move it out of the inheritance chain, so FlatIndex inherits both BaseVectorIndex and DocIndex, whereas HNSW index will inherit only BaseVectorIndex. There are many ways to re-use the code

src/core/search/indices.cc
```diff
@@ -621,34 +652,43 @@ struct HnswlibAdapter {
 
   SpaceUnion space_;
   hnswlib::HierarchicalNSW<float> world_;
+  absl::Mutex resize_mutex_;
```
romange, line 659: why `resize_mutex_` is needed? I thought `HierarchicalNSW` is thread-safe

src/server/search/global_hnsw_vector_index.h
```diff
@@ -0,0 +1,83 @@
...
+      const std::optional<search::KnnScoreSortOption>& knn_score_option,
+      const std::vector<SearchResult>& filter_docs, const SearchParams& params,
+      const CommandContext& cmd_cntx);
+
+ private:
+  std::vector<std::pair<float, search::GlobalDocId>> SearchKnnHnsw(
+      search::HnswVectorIndex* index, const search::AstKnnNode* knn,
+      const std::optional<std::vector<search::GlobalDocId>>& allowed_docs);
+
+  std::unique_ptr<search::BaseVectorIndex<search::GlobalDocId>> vector_index_;
```
romange: Probably it follows @dranikpg  question but what is the added value of declaring here BaseVectorIndex instead of specifying `search::HnswVectorIndex` explicitly. Moreover, what is the value of having  GlobalHnswVectorIndex wrappin `search::HnswVectorIndex` ? why not have only class?

src/server/main_service.cc
```diff
@@ -1123,6 +1124,11 @@ void Service::Shutdown() {
 
   shard_set->PreShutdown();
   shard_set->Shutdown();
+
```
romange, line 1126: Lets introduce `SearchFamily::Shutdown()` and move it there -no need to leak the implementation into main_service.cc

src/core/search/base.h
```diff
@@ -16,10 +16,24 @@
 #include "absl/container/flat_hash_set.h"
 #include "base/pmr/memory_resource.h"
 #include "core/string_map.h"
+#include "server/tx_base.h"
```
romange: that's not good. we can not include server headers into core library.

romange: just redefine `using ShardId = uint16_t;` under `GlobalDocId`

src/core/search/indices.cc
mkaruza: `resizeIndex` is not thread safe with insertion - https://github.com/nmslib/hnswlib/issues/267

dranikpg: but it is safe with reads/mark deletes? 🤔

src/core/search/ast_expr.cc
```diff
@@ -73,6 +73,12 @@ AstKnnNode::AstKnnNode(AstNode&& filter, AstKnnNode&& self) {
   this->filter = make_unique<AstNode>(std::move(filter));
 }
 
+bool AstKnnNode::HasPreFilter() const {
+  // If we have pre filter knn query should not hold filter variable. It will be
+  // moved to SearchAlgorithm::query_ variable.
+  return filter == nullptr;
```
romange, line 79: and if we do not have pre-filter, what filter would be? :)

src/core/search/indices.cc
```diff
@@ -560,34 +564,45 @@ struct HnswlibAdapter {
...
   }
 
-  void Add(const float* data, DocId id) {
-    if (world_.cur_element_count + 1 >= world_.max_elements_)
-      world_.resizeIndex(world_.cur_element_count * 2);
-    world_.addPoint(data, id);
+  void Add(const float* data, GlobalDocId id) {
+    while (true) {
+      try {
+        absl::ReaderMutexLock lock(&resize_mutex_);
```
romange, line 566: please add a comment that resizes needs to be protected and add the link to the issue.

src/core/search/indices.cc
```diff
@@ -609,8 +624,32 @@ struct HnswlibAdapter {
...
+      absl::WriterMutexLock lock(&resize_mutex_);
+      if (world_.getCurrentElementCount() == world_.getMaxElements() &&
+          (!world_.allow_replace_deleted_ || world_.getDeletedCount() == 0)) {
+        auto max_elements = world_.getMaxElements();
+        world_.resizeIndex(max_elements * 2);
+        LOG(INFO) << "Resizing HNSW Index, current size: " << max_elements
+                  << ", expand by: " << max_elements * 2;
+      }
+    } catch (const std::exception& e) {
+      throw e;
```
romange: we do not throw. should this crash the program? if yes, then `LOG(FATAL) << ... ` is a better choice

src/core/search/ast_expr.cc
mkaruza: If it is only KNN scan than filter will contain - `AstStarNode` i.e. `*`

src/core/search/indices.cc
```diff
@@ -621,34 +660,50 @@ struct HnswlibAdapter {
...
+bool HnswVectorIndex::Add(GlobalDocId id, const DocumentAccessor& doc, std::string_view field) {
+  auto vector = doc.GetVector(field);
+
+  if (!vector) {
+    return false;
+  }
+
+  auto& [ptr, size] = vector.value();
+
+  if (ptr && size != dim_) {
```
romange, line 680: what if `!ptr`? can it happen and what does it mean?

src/core/search/indices.cc
mkaruza: Added comment above function `ResizeIfFull`

src/core/search/indices.h
```diff
@@ -191,27 +191,45 @@ struct FlatVectorIndex : public BaseVectorIndex {
...
 
-struct HnswlibAdapter;
-
-struct HnswVectorIndex : public BaseVectorIndex {
-  HnswVectorIndex(const SchemaField::VectorParams& params, PMR_NS::memory_resource* mr);
-  ~HnswVectorIndex();
-
-  void Remove(DocId id, const DocumentAccessor& doc, std::string_view field) override;
+// We still need to have hnsw index on each shard. This is empty implementation
+// that doesn't do anything and it serves only as placeholder.
```
romange: can you expand on this please - why do we need it?

src/server/search/search_family.cc
```diff
@@ -1384,24 +1573,47 @@ void SearchFamily::FtSearch(CmdArgList args, const CommandContext& cmd_cntx) {
   if (!search_algo.Init(query_str, &params->query_params, &params->optional_filters))
     return builder->SendError("Query syntax error");
 
+  std::unique_ptr<search::AstNode> knn_node = nullptr;
```
romange: no need for `= nullptr`

src/server/search/search_family.cc
```diff
@@ -1384,24 +1573,47 @@ void SearchFamily::FtSearch(CmdArgList args, const CommandContext& cmd_cntx) {
...
   vector<SearchResult> docs(shard_set->size());
 
-  cmd_cntx.tx->ScheduleSingleHop([&](Transaction* t, EngineShard* es) {
-    if (auto* index = es->search_indices()->GetIndex(index_name); index)
-      docs[es->shard_id()] = index->Search(t->GetOpArgs(es), *params, &search_algo);
-    else
-      index_not_found.store(true, memory_order_relaxed);
-    return OpStatus::OK;
-  });
+  // If it is HNSW knn node and we have pre filter query
```
romange:
```suggestion
  // If the query does not contain knn component, or it is a hybrid query
```

src/server/search/search_family.cc
```diff
@@ -1384,24 +1573,47 @@ void SearchFamily::FtSearch(CmdArgList args, const CommandContext& cmd_cntx) {
...
+      if (auto* index = es->search_indices()->GetIndex(index_name); index)
+        docs[es->shard_id()] = index->Search(t->GetOpArgs(es), *params, &search_algo);
+      else
+        index_not_found.store(true, memory_order_relaxed);
+      return OpStatus::OK;
+    });
 
-  if (index_not_found.load())
-    return builder->SendError(string{index_name} + ": no such index");
+    if (index_not_found.load())
```
romange: nit:
```suggestion
    if (index_not_found.load(memory_order_relaxed))
```

src/core/search/ast_expr.cc
romange: it's counter intuitive but I guess it's out of scope of this PR

romange:
please fix build/test  errors tomorrow, and lets submit. I want to continue experimenting on this

src/core/search/indices.h
mkaruza: I have just checked simple test now and maybe it's not needed, although I'm pretty sure that was need on "previous" implementation (could not recall why exactly). I'll do some extra testing but it looks that i would remove it.

src/core/search/base.h
```diff
@@ -20,6 +20,16 @@
...
 using DocId = uint32_t;
+using GlobalDocId = uint64_t;
+using ShardId = uint16_t;
+
+inline GlobalDocId CreateGlobalDocId(ShardId shard_id, DocId local_doc_id) {
+  return ((uint64_t)shard_id << 32) | local_doc_id;
+}
+
+inline std::pair<ShardId, DocId> DecomposeGlobalDocId(GlobalDocId id) {
+  return {(id >> 32), (id)&0xFFFF};
```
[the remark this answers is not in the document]
romange: nice! that's a real bug found.

src/server/search/global_hnsw_index.cc
```diff
@@ -0,0 +1,87 @@
...
+  std::string key = MakeKey(index_name, field_name);
+
+  std::shared_lock<std::shared_mutex> lock(registry_mutex_);
+
+  auto it = indices_.find(key);
+
+  if (it != indices_.end())
+    return false;
+
+  indices_[key] = std::make_shared<search::HnswVectorIndex>(params);
```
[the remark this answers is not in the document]
romange: another real bug

src/server/search/global_hnsw_index.cc
```diff
@@ -0,0 +1,87 @@
...
+
+  return true;
+}
+
+bool GlobalHnswIndexRegistry::Remove(std::string_view index_name, std::string_view field_name) {
+  std::string key = MakeKey(index_name, field_name);
+
+  std::unique_lock<std::shared_mutex> lock(registry_mutex_);
```
romange, line 50: seems like a bug as well

src/server/search/global_hnsw_index.cc
```diff
@@ -0,0 +1,87 @@
...
+
+  return false;
+}
+
+std::shared_ptr<search::HnswVectorIndex> GlobalHnswIndexRegistry::Get(
+    std::string_view index_name, std::string_view field_name) const {
+  std::string key = MakeKey(index_name, field_name);
+  std::shared_lock<std::shared_mutex> lock(registry_mutex_);
+  auto it = indices_.find(key);
+  return it != indices_.end() ? it->second : nullptr;
```
romange, line 67:
```suggestion
  return it != indices_.end() ? it->second : {};
```

src/core/search/indices.cc
mkaruza: `!ptr` is when we key doesn't have field that we index so null value.

src/core/search/indices.h
mkaruza: Removed

src/core/search/search_test.cc
```diff
@@ -646,11 +646,11 @@ TEST_F(SearchTest, MatchNumericRangeWithCommas) {
   }
 }
 
-class KnnTest : public SearchTest, public testing::WithParamInterface<bool /* hnsw */> {};
+class KnnTest : public SearchTest {};
```
romange, line 649: can you explain this change? we are not testing Knn anymore?

mkaruza: We are not testing HNSW anymore

PR #6106, opened 2025-11-24
chore: implement background self-identification of huffman encoding

romange:
The PR adds a background task that runs only once on db 0 and computes huffman table
but only on datasets that fit the requirements - key memory usage is significant
enough to do such analysis. At the end of the process a metric increased to signal
that the huffman table was computed and printed to logs.

### DIFF
--- a/src/server/engine_shard.cc
+++ b/src/server/engine_shard.cc
@@ -151,6 +151,11 @@ int32_t HuffmanCheckTask::Run(DbSlice* db_slice) {
         for (unsigned char c : val) {
           hist_[c]++;
         }
+
+        if (val.size() > 1024) {
+          traverses_count = kMaxTraverses;  // return early.
+          string{}.swap(scratch_);          // free memory.
+        }
       }
     });
     traverses_count++;
@@ -721,19 +726,15 @@ void EngineShard::Heartbeat() {
         check_huffman = false;  // trigger only once.
 
         // launch the task
-        HuffmanCheckTask* task = new HuffmanCheckTask();
-        huffman_check_task_id_ = ProactorBase::me()->AddOnIdleTask([task] {
-          if (!shard_ || !namespaces) {
-            delete task;
-            return -1;
-          }
+        huffman_check_task_id_ =
+            ProactorBase::me()->AddOnIdleTask([task = HuffmanCheckTask{}]() mutable {
+              if (!shard_ || !namespaces) {
+                return -1;
+              }
 
-          DbSlice& db_slice = namespaces->GetDefaultNamespace().GetDbSlice(shard_->shard_id());
-          int32_t res = task->Run(&db_slice);
-          if (res == -1)
-            delete task;
-          return res;
-        });
+              DbSlice& db_slice = namespaces->GetDefaultNamespace().GetDbSlice(shard_->shard_id());
+              return task.Run(&db_slice);
+            });
       }
     }
   }


romange:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/dragonflydb/dragonfly/pull/6106#pullrequestreview-3501725633)

src/server/engine_shard.cc
```diff
@@ -116,6 +118,79 @@ size_t CalculateHowManyBytesToEvictOnShard(size_t global_memory_limit, size_t gl
...
+  auto& prime = db_table->prime;
+
+  constexpr uint32_t kMaxTraverses = 512;
+  uint32_t traverses_count = 0;
+  do {
+    cursor_ = prime.Traverse(cursor_, [&](PrimeIterator it) {
+      if (!it->first.IsInline()) {
+        string_view val = it->first.GetSlice(&scratch_);
+        for (unsigned char c : val) {
+          hist_[c]++;
```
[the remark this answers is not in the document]
romange: can not use uint64_t as the huffman library accept `unsigned` types.

src/server/engine_shard.cc
```diff
@@ -116,6 +118,79 @@ size_t CalculateHowManyBytesToEvictOnShard(size_t global_memory_limit, size_t gl
...
+    });
+    traverses_count++;
+  } while (traverses_count < kMaxTraverses && cursor_);
+
+  if (cursor_)
+    return 4;  // priority to continue later.
+
+  // Finished scanning the table, now normalize the table.
+  constexpr unsigned kMaxFreqTotal = static_cast<unsigned>((1U << 31) * 0.9);
+  size_t total_freq = std::accumulate(hist_.begin(), hist_.end(), 0UL);
```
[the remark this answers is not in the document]
romange: in general, it can not happen as we trigger the task when we know the memory usage is significant.  having said that, there could be a potential race where data is flushed before the task runs.

src/server/engine_shard.cc
```diff
@@ -116,6 +118,81 @@ size_t CalculateHowManyBytesToEvictOnShard(size_t global_memory_limit, size_t gl
   return shard_budget < shard_memory_threshold ? (shard_memory_threshold - shard_budget) : 0;
 }
 
+class HuffmanCheckTask {
+ public:
+  HuffmanCheckTask() {
+    hist_.fill(0);
+  }
+
+  int32_t Run(DbSlice* db_slice);
```
dranikpg, line 127: My dream once was to turn every on idle task into a background fiber... 😞

src/server/engine_shard.cc
```diff
@@ -116,6 +118,81 @@ size_t CalculateHowManyBytesToEvictOnShard(size_t global_memory_limit, size_t gl
...
+
+int32_t HuffmanCheckTask::Run(DbSlice* db_slice) {
+  DbTable* db_table = db_slice->GetDBTable(0);  // we currently support only default db.
+  if (!db_table)
+    return -1;
+
+  // incrementally aggregate frequency histogram.
+  auto& prime = db_table->prime;
+
+  constexpr uint32_t kMaxTraverses = 512;
```
dranikpg, line 145: My concern is that you run it without flags or anything and it might freeze when iterating 512 of big strings

src/server/engine_shard.cc
```diff
@@ -620,7 +700,43 @@ void EngineShard::Heartbeat() {
...
+
+        check_huffman = false;  // trigger only once.
+
+        // launch the task
+        HuffmanCheckTask* task = new HuffmanCheckTask();
+        huffman_check_task_id_ = ProactorBase::me()->AddOnIdleTask([task] {
+          if (!shard_ || !namespaces) {
+            delete task;
+            return -1;
+          }
```
dranikpg: It feels a little sloppy overall, like it was written for a POC. If it becomes more complicated, it would be preferrably to refactor it a little. For example, you don't need the pointer, it can be just a member of the lambda,

dranikpg (dismissed):

src/server/engine_shard.cc
romange: Is it the same though? I clearly remember that OnIdle tasks run only when we do not have any CPU work to do and can be starved indefinitely. I thought that background fibers a bit more agressive. Also OnIdle tasks provide a way to throttle themselves and background fibers do not do that,, imho

src/server/engine_shard.cc
romange: A valid concern, even though I do not think keys are long  for real workloads. I will add a protection.

src/server/engine_shard.cc
romange: thanks! :)
what do you mean by that? how lambda can have a member?

src/server/engine_shard.cc
dranikpg: Ok, didn't think about it being just keys, I assume that in the future we want to test values as well

src/server/engine_shard.cc
dranikpg: I'm just being honest 😅 
`[task = HummanCheckTasks{}] mutable {} `, desotrying the lambda will destroy the task

PR #6118, opened 2025-11-26
chore: add mi_malloc memory test

romange:
This test demonstrates the limitations of stats tracking in mimalloc. Specifically, it is currently not possible to track per heap statistics with mimalloc.


romange:
Yes, it's a convenient playground to test mimalloc changes.

PR #6123, opened 2025-11-26
fix: Handle django-cache script so it could run in dragonfly

### LINKED ISSUE #6119: simplify lua_undeclared_keys_shas management
1. move known shas from lua_undeclared_keys_shas to hardcoded table.
2. move lua_force_atomicity_shas to hardcoded table.
3. add 8c4dafdf9b6b7bcf511a0d1ec0518bed9260e16d to the table as well (django-ops)

romange:
automatically.

Fixes #6119

### DIFF
--- a/src/server/multi_test.cc
+++ b/src/server/multi_test.cc
@@ -840,8 +840,6 @@ TEST_F(MultiTest, ScriptFlagsEmbedded) {
   EXPECT_THAT(Run({"eval", s2, "0"}), ErrArg("Invalid flag: this-is-an-error"));
 }
 
-// Flaky because of https://github.com/google/sanitizers/issues/1760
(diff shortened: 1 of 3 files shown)


src/server/multi_test.cc
```diff
@@ -1254,29 +1251,6 @@ TEST_F(MultiTest, EvalShaRo) {
   EXPECT_THAT(resp, ErrArg("Write commands are not allowed from read-only scripts"));
 }
 
-TEST_F(MultiTest, ForceAtomicityFlag) {
```
romange, line 1257: removed as we do not need this feature anymore.

vyavdoshenko (approved):
lgtm

PR #6126, opened 2025-11-27
chore: cleanup redundant access specifiers

abhijat:
Cleaned up redundant access scope spefiers using

```
run-clang-tidy -p build-dbg -checks="-*,readability-redundant-access-specifiers" src/ -header-filter=.* -fix
```

Checked manually that the changes were correct

### DIFF
--- a/src/core/interpreter.h
+++ b/src/core/interpreter.h
@@ -174,7 +174,6 @@ class InterpreterManager {
     int64_t gc_freed_memory = 0;
   };
 
- public:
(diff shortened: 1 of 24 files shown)


dranikpg:
I use them to separate between functions and members 😢

abhijat:
> I use them to separate between functions and members 😢

Oh never mind then, we can keep them as it is, I thought as code increased these were added but seems like a good reason to keep them.

abhijat:
the specifiers are to create sections in code, so lets keep them

PR #6130, opened 2025-11-27
chore: expose more information when takeover fails on master.

romange:
1. Provide more debug info on current connections
2. Show the failed condition of DispatchTracker if it fails.

### DIFF
--- a/src/server/replica.cc
+++ b/src/server/replica.cc
@@ -200,17 +200,11 @@ void Replica::Pause(bool pause) {
   });
 }
 
-std::error_code Replica::TakeOver(std::string_view timeout, bool save_flag) {
-  VLOG(1) << "Taking over";
-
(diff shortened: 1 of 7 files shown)


src/server/replica.cc
```diff
@@ -200,17 +200,11 @@ void Replica::Pause(bool pause) {
...
 
-std::error_code Replica::TakeOver(std::string_view timeout, bool save_flag) {
-  VLOG(1) << "Taking over";
-
-  // Parse timeout value for socket timeout
-  int timeout_sec = 0;
-  if (!absl::SimpleAtoi(timeout, &timeout_sec)) {
-    return make_error_code(errc::invalid_argument);
-  }
+std::error_code Replica::TakeOver(unsigned timeout_sec, bool save_flag) {
```
romange, line 203: `timeout` now is unsigned

PR #6132, opened 2025-11-28
chore: use our own vector for hnsw distance computation

romange:
Our functions use dynamically dispatched simsimd optimizations that may be more efficient on modern hardware than just using SSE instructions.

### DIFF
--- a/src/core/search/vector_utils.cc
+++ b/src/core/search/vector_utils.cc
@@ -25,36 +25,6 @@ namespace {
 #define FAST_MATH
 #endif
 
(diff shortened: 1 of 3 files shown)


src/core/search/vector_utils.cc
```diff
@@ -86,6 +56,36 @@ OwnedFtVector ConvertToFtVector(string_view value) {
 
 }  // namespace
 
+// Euclidean vector distance: sqrt( sum: (u[i] - v[i])^2  )
```
romange, line 59: just moved them from anonymous space here

dranikpg (approved):

PR #6136, opened 2025-11-28
chore: move HierarchicalNSW class to our codebase

romange:


romange:
There is no value in using the fork - the third-party code does not change, the codebase is small.
changing the fork is more complicated then changing our code directly, as it can not benefit from all the coding infra we have in our repo. Basically I do it so we could iterate quicker and be more aggressive with hnsw development

Issue #6137, opened 2025-11-29
Unfuddle CompactObject


dranikpg:
Goals:
1. We use 16 bits for mask space in a key/value pair, but use only 6 different flags exclusively on either values or keys. Reduce mask size to one byte (in key) and free up one byte in value
    * maybe move all flags to value to remove one byte from expiry table key additionally
2.  Reduce code complexity     
    * Use different types for keys and values, inheriting both compact object, but hiding non-relevant functions and simiplifying logic
    * Review old string management code and simplify it
    * Review externally induced interface members (returning directly RObjWrapper, stream/sbf functions) that should be kept in familiy code and simplified

-> save ony byte per slot,  have simpler object interface and hopefully a few hundred lines less code

Steps:
- [x] Remove async delete flag usage (can be replaced by a function parameter...)
- [x] Refactor small string
- [x] Tidy up huffman code, encoding code_
- [ ] Remove GetFirstByte
- [x] Remove raw access to RObjWrapper, remove unnecessary functions (InlineLen(), allows empty state, etc)
- [ ] Remove `OBJ_ENCODING_INT` and similar, its never used
- [x] Move encoding to tagbyte_
- [x] Change huffman_enc to is_key and rebase all other usage around it
- [x] Use basic different types to catch sloppy type usage (there are cases), set is_key implicitily
- [ ] Move all flags to key object
- [ ] Review Free/SetMeta/Reset/etc functionality to be duplicate free

PR #6138, opened 2025-11-29
chore(core): Refactor SmallString

dranikpg:
Relevant to #6137 
1. Half of SmallStr code assumed values could be less than kPrefLen, half of it asserted it, parts would've just broke. I kept at DCHECK at assign and simplified all other parts
2. **For some reason, if the length of a new string is greater than the current small str, we just decided to pick a raw allocated (robj) string instead**, even if it would fit into a reallocated small string
3. Simplified `CompactObj::GetString()` part with encodings

### DIFF
--- a/src/core/small_string.cc
+++ b/src/core/small_string.cc
@@ -60,11 +60,11 @@ size_t SmallString::Assign(std::string_view s) {
   uint8_t* realptr = nullptr;
 
   // reallocate if we need a larger allocation or it becomes space-inefficient
-  // TODO: check if mi_usable_size is has space before growing (like listpack)
-  if (s.size() >= size_ || s.size() * 2 < MallocUsed()) {
+  size_t heap_len = s.size() - kPrefLen;
+  if (size_t available = MallocUsed(); available < heap_len || heap_len * 2 < available) {
     Free();
 
-    auto [sp, rp] = tl.seg_alloc->Allocate(s.size() - kPrefLen);
+    auto [sp, rp] = tl.seg_alloc->Allocate(heap_len);
     small_ptr_ = sp;
     realptr = rp;
   } else {
@@ -73,7 +73,7 @@ size_t SmallString::Assign(std::string_view s) {
 
   size_ = s.size();
   memcpy(prefix_, s.data(), kPrefLen);
-  memcpy(realptr, s.data() + kPrefLen, s.size() - kPrefLen);
+  memcpy(realptr, s.data() + kPrefLen, heap_len);
   return mi_malloc_usable_size(realptr);
 }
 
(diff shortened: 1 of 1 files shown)


src/core/compact_object.cc
```diff
@@ -1585,18 +1564,14 @@ void CompactObj::EncodeString(string_view str, bool is_key) {
...
+  if (SmallString::CanAllocate(encoded.size())) {
+    if (taglen_ == SMALL_TAG)
+      tl.small_str_bytes -= u_.small_str.MallocUsed();
+    else
       SetMeta(SMALL_TAG, mask_);
-      tl.small_str_bytes += u_.small_str.Assign(encoded);
-      return;
-    }
 
-    if (taglen_ == SMALL_TAG && encoded.size() <= u_.small_str.size()) {
```
dranikpg, line 1595: this was the part with `encoded.size() <= u_.small_str.size()`

BorysTheDev (approved):
I don't understand a lot, ask somebody else to review too

src/core/small_string.cc
```diff
@@ -54,55 +54,36 @@ size_t SmallString::UsedThreadLocal() {
...
 size_t SmallString::Assign(std::string_view s) {
   DCHECK_GT(s.size(), kPrefLen);
-
+  DCHECK(CanAllocate(s.size()));
   uint8_t* realptr = nullptr;
 
-  if (size_ == 0) {
-    // packed structs can not be tied here.
+  // If the allocation is large enough and space efficient, we can avoid allocating
+  if (s.size() >= size_ || s.size() * 2 < MallocUsed()) {
```
romange: something is wrong with this condition, imho

src/core/small_string.h
```diff
@@ -23,41 +21,32 @@ class SmallString {
...
-  }
-
   uint64_t HashCode() const;
-
-  // I am lying here. we should use mi_malloc_usable size really.
   uint16_t MallocUsed() const;
 
+  void Get(std::string_view dest[2]) const;
+  void Get(char* out) const;
   void Get(std::string* dest) const;
```
romange, line 36: you removed `void Get(std::string* dest) const` implementation

src/core/small_string.cc
dranikpg: Ah, it's just the inverse of the comment, I'll update it

src/core/small_string.h
dranikpg: No, it just takes 2 lines of code, its in the cc file

src/core/small_string.cc
romange: but I think the condition is wrong.
 `size_ = 20, malloc_used = 28, s.size() = 24 ` - you reallocate because of the first part. unnecessary. basically you only need to compare with MallocUsed imho

dranikpg: I added a todo, I can do it now. The code previously _didn't allow any strings longer at all_  to be assigned

dranikpg:
I use mi_good_size now to check if we need to grow. This is a logical change in this PR

PR #6141, opened 2025-11-30
chore(core): Remove some unused CompactObj code

dranikpg:
Relevant to #6137 

1. Spit tagbyte union<uint8, struct{bitfields}> into just two bitfields
2. Remove some unused functions
3. Rename Cmp function, add dchecks


romange:
I do not know why you did not like the tagbyte union trick.

dranikpg:
> I do not know why you did not like the tagbyte union trick

Its nice for flags, but not for fields that are not connected in any way. I will destruct those unions and will keep only one in the key

PR #6142, opened 2025-12-01
chore: force shutdown connections on takeover

### LINKED ISSUE #6114: Improve TakeOver flow
## Issue

On takeover flow the replica sends a `DFLY TAKEOVER` to master.  Then on the master side, a check point message is placed to all active connections (internally we just iterate over all the connections and place a `Checkpoint` msg in each dispatch queue). Then we wait for all the connections to process the Checkpoint msg or time out. On the latter case, the takeover fails.

This works well most of the time but the main issue here is is a pipeline is blocked then the checkpoint message won't be processed and dragonfly will fail to takeover. 

## Reproduce 

Use `test_send_timeout` and remove the timeout (so the connection remains idle but not killed). Then, initiate a take over. It shall fail because the async fiber is blocked.

## Remark

* At any point, if `async fiber` blocks for an extended period of time, then dispatch_q_ processing stops as well so any flow that relies and waits on connection dispatch_q_ processing will timeout.

kostasrim:
## TakeOver algorithm as is

Imagine the simple setup: node 1 master, node 2 replica and node 2 attempts takeover on node 1.

`For node 1:`

### Steps that affect TakeOver result/state:

1. Switch state to TAKEN_OVER -> connections will get their commands rejected
2. Wait for all connections to finish dispatches -> send Checkpoint message on each conn and wait for it to be processed
* New step here from this PR. Optionally force shutdown all open connections. This does not break data parity. The connection is stuck on send, will break with an error but the transaction data will get replicated anyway (we already wrote to the journal the change).
3. Wait for the replica to catch up (reach the same LSN as master) -> data parity
4. Reply with OK. At this point the TAKEOVER is considered complete, node 2 is master.

### Optional steps that improve other flows:

1. Optionally: if there are more nodes registered as replicas, wait for those to catch up. This does not affect the takeover but allows waiting for all of the replicas to fully sync. (Also needed for partial sync data integrity)
2. Optionally: Take a snapshot
3. Optionally: for cluster mode do not shutdown such that the node can redirect requests

## Notes

* Dragonfly is still accepting new connections during the takeover. However, once dragonfly switches to TAKEN_OVER state, connections won’t be able to execute commands and they will get “busy loading” error

* Checkpoints are needed for data integrity. Once all the connections process the checkpoints it means that any other command in their dispatch queue will fail with busy loading. Hence, after step(2) there won’t be any change in the state (storage) and after step(3) the taking over node is on data parity.

## This PR:

A step between 2 and 3 to fix non responding connections and force the takeover by forcefully shutting them down.

Follows up #6135. Add a flag `experimental_force_takeover`. If takeover times out and this flag is set dragonfly will force shutdown open connections such that it doesn't fail the takeover.

Resolves: #6114

### DIFF
--- a/src/server/dflycmd.cc
+++ b/src/server/dflycmd.cc
@@ -40,6 +40,8 @@ ABSL_RETIRED_FLAG(uint32_t, allow_partial_sync_with_lsn_diff, 0,
 ABSL_DECLARE_FLAG(bool, info_replication_valkey_compatible);
 ABSL_DECLARE_FLAG(uint32_t, replication_timeout);
 ABSL_DECLARE_FLAG(uint32_t, shard_repl_backlog_len);
+ABSL_FLAG(bool, experimental_force_takeover, false,
+          "Attempts to force takeover in case of stuck connections");
 
 namespace dfly {
 
@@ -466,6 +468,45 @@ std::optional<LSN> DflyCmd::ParseLsnVec(std::string_view last_master_lsn,
   return {lsn_vec[flow_id]};
 }
 
+void DflyCmd::ForceShutdownStuckConnections(uint64_t timeout) {
+  vector<facade::Connection::WeakRef> conn_refs;
+  auto cb = [&](unsigned thread_index, util::Connection* conn) {
+    facade::Connection* dcon = static_cast<facade::Connection*>(conn);
+    LOG(INFO) << dcon->DebugInfo();
+    // Kill Connection here
+    facade::Connection* dfly_conn = static_cast<facade::Connection*>(conn);
+    using Phase = facade::Connection::Phase;
+    auto phase = dfly_conn->phase();
+    if (dfly_conn->cntx() && dfly_conn->cntx()->replica_conn) {
+      return;
+    }
+
+    bool idle_read = phase == Phase::READ_SOCKET && dfly_conn->idle_time() > timeout;
+
+    bool stuck_sending = dfly_conn->IsSending() && dfly_conn->GetSendWaitTimeSec() > timeout;
+
+    if (idle_read || stuck_sending) {
+      LOG(INFO) << "Connection check: " << dfly_conn->GetClientInfo()
+                << ", phase=" << static_cast<int>(phase) << ", idle_time=" << dfly_conn->idle_time()
+                << ", is_sending=" << dfly_conn->IsSending() << ", idle_read=" << idle_read
+                << ", stuck_sending=" << stuck_sending;
+    }
+    conn_refs.push_back(dfly_conn->Borrow());
+  };
+
+  for (auto* listener : sf_->GetListeners()) {
+    listener->TraverseConnections(cb);
+  }
+
+  VLOG(1) << "Found " << conn_refs.size() << " stucked connections ";
+  for (auto& ref : conn_refs) {
+    facade::Connection* conn = ref.Get();
+    if (conn) {
+      conn->ShutdownSelfBlocking();
+    }
+  }
+}
+
 // DFLY TAKEOVER <timeout_sec> [SAVE] <sync_id>
 // timeout_sec - number of seconds to wait for TAKEOVER to converge.
 // SAVE option is used only by tests.
@@ -506,7 +547,7 @@ void DflyCmd::TakeOver(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext
   LOG(INFO) << "Takeover initiated, locking down the database.";
   absl::Duration timeout_dur = absl::Seconds(timeout);
   absl::Time end_time = absl::Now() + timeout_dur;
-  AggregateStatus status;
+  OpStatus status = OpStatus::OK;
 
   // We need to await for all dispatches to finish: Otherwise a transaction might be scheduled
   // after this function exits but before the actual shutdown.
@@ -520,13 +561,22 @@ void DflyCmd::TakeOver(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext
     LOG(WARNING) << "Couldn't wait for commands to finish dispatching. " << timeout_dur;
     status = OpStatus::TIMED_OUT;
 
-    auto cb = [&](unsigned thread_index, util::Connection* conn) {
-      facade::Connection* dcon = static_cast<facade::Connection*>(conn);
-      LOG(INFO) << dcon->DebugInfo();
-    };
+    // Force takeover on the same duration if flag is set
+    if (absl::GetFlag(FLAGS_experimental_force_takeover)) {
+      ForceShutdownStuckConnections(uint64_t(timeout));
 
-    for (auto* listener : sf_->GetListeners()) {
-      listener->TraverseConnections(cb);
+      // Safety net.
+      facade::DispatchTracker tracker{sf_->GetNonPriviligedListeners(), cntx->conn(), false, false};
+      shard_set->pool()->AwaitFiberOnAll([&](unsigned index, auto* pb) {
+        sf_->CancelBlockingOnThread();
+        tracker.TrackOnThread();
+      });
+
+      status = OpStatus::OK;
+      if (!tracker.Wait(timeout_dur)) {
+        LOG(ERROR) << "Could not force execute takeover";
+        status = OpStatus::TIMED_OUT;
+      }
     }
   }
 
@@ -540,10 +590,11 @@ void DflyCmd::TakeOver(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext
   });
 
   atomic_bool catchup_success = true;
-  if (*status == OpStatus::OK) {
+  if (status == OpStatus::OK) {
     dfly::SharedLock lk{replica_ptr->shared_mu};
-    auto cb = [replica_ptr = replica_ptr, end_time, &catchup_success](EngineShard* shard) {
-      if (!WaitReplicaFlowToCatchup(end_time, replica_ptr.get(), shard)) {
+    auto time = end_time + timeout_dur;
+    auto cb = [replica_ptr = replica_ptr, time, &catchup_success](EngineShard* shard) {
+      if (!WaitReplicaFlowToCatchup(time, replica_ptr.get(), shard)) {
         catchup_success.store(false);
       }
     };
@@ -552,8 +603,9 @@ void DflyCmd::TakeOver(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext
 
   VLOG(1) << "WaitReplicaFlowToCatchup done";
 
-  if (*status != OpStatus::OK || !catchup_success.load()) {
+  if (status != OpStatus::OK || !catchup_success.load()) {
     sf_->service().SwitchState(GlobalState::TAKEN_OVER, GlobalState::ACTIVE);
+    LOG(INFO) << status << " " << catchup_success.load() << " " << &status;
     return rb->SendError("Takeover failed!");
   }
 
--- a/src/server/dflycmd.cc
+++ b/src/server/dflycmd.cc
@@ -469,7 +469,9 @@ std::optional<LSN> DflyCmd::ParseLsnVec(std::string_view last_master_lsn,
 }
 
 void DflyCmd::ForceShutdownStuckConnections(uint64_t timeout) {
+  // per proactor map
   vector<facade::Connection::WeakRef> conn_refs;
+
   auto cb = [&](unsigned thread_index, util::Connection* conn) {
     facade::Connection* dcon = static_cast<facade::Connection*>(conn);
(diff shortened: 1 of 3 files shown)


src/server/dflycmd.cc
```diff
@@ -14,6 +14,8 @@
 #include <utility>
 
 #include "absl/cleanup/cleanup.h"
+#include "absl/debugging/stacktrace.h"
```
kostasrim: What the hell 😮‍💨

src/server/dflycmd.cc
```diff
@@ -466,6 +470,45 @@ std::optional<LSN> DflyCmd::ParseLsnVec(std::string_view last_master_lsn,
...
+    }
+
+    bool idle_read = phase == Phase::READ_SOCKET && dfly_conn->idle_time() > timeout;
+
+    bool stuck_sending = dfly_conn->IsSending() && dfly_conn->GetSendWaitTimeSec() > timeout;
+
+    if (idle_read || stuck_sending) {
+      LOG(INFO) << "Connection check: " << dfly_conn->GetClientInfo()
+                << ", phase=" << static_cast<int>(phase) << ", idle_time=" << dfly_conn->idle_time()
+                << ", is_sending=" << dfly_conn->IsSending() << ", idle_read=" << idle_read
```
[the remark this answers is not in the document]
kostasrim: 👎 👎

src/server/dflycmd.cc
```diff
@@ -466,6 +468,45 @@ std::optional<LSN> DflyCmd::ParseLsnVec(std::string_view last_master_lsn,
...
+                << ", stuck_sending=" << stuck_sending;
+    }
+    conn_refs.push_back(dfly_conn->Borrow());
```
kostasrim, line 496: I shut the connections anyway regardless of if they are stuck on send or not. The rationale behind this is that a connection can become active `at any point` if the client starts reading from the socket. If that's the case, we will fail again because the connection won't show up as stuck. To mitigate this, I force shutdown all connections that were not closed previously

src/server/dflycmd.cc
```diff
@@ -520,13 +561,22 @@ void DflyCmd::TakeOver(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext
...
-    };
-
-    for (auto* listener : sf_->GetListeners()) {
-      listener->TraverseConnections(cb);
+    // Force takeover on the same duration if flag is set
+    if (absl::GetFlag(FLAGS_experimental_force_takeover)) {
+      ForceShutdownStuckConnections(uint64_t(timeout));
+
+      // Safety net.
+      facade::DispatchTracker tracker{sf_->GetNonPriviligedListeners(), cntx->conn(), false, false};
```
kostasrim, line 570: Not really needed to wait again tbh. We forcefully kill connections and the ones that were in good state were already shutdown (because of the previous dispatch tracker

src/server/dflycmd.cc
```diff
@@ -466,6 +468,45 @@ std::optional<LSN> DflyCmd::ParseLsnVec(std::string_view last_master_lsn,
...
+
+  for (auto* listener : sf_->GetListeners()) {
+    listener->TraverseConnections(cb);
+  }
+
+  VLOG(1) << "Found " << conn_refs.size() << " stucked connections ";
+  for (auto& ref : conn_refs) {
+    facade::Connection* conn = ref.Get();
+    if (conn) {
+      conn->ShutdownSelfBlocking();
```
romange, line 509: I think you should do it from the connection thread.

src/server/dflycmd.cc
```diff
@@ -466,6 +468,45 @@ std::optional<LSN> DflyCmd::ParseLsnVec(std::string_view last_master_lsn,
   return {lsn_vec[flow_id]};
 }
 
+void DflyCmd::ForceShutdownStuckConnections(uint64_t timeout) {
+  vector<facade::Connection::WeakRef> conn_refs;
+  auto cb = [&](unsigned thread_index, util::Connection* conn) {
+    facade::Connection* dcon = static_cast<facade::Connection*>(conn);
+    LOG(INFO) << dcon->DebugInfo();
```
romange, line 477: maybe vlog? and more meaningful message

src/server/dflycmd.cc
dranikpg: Yes, its unsafe to do so from foreign threads, WeakRef has it in the comments

src/server/dflycmd.cc
```diff
@@ -466,6 +468,45 @@ std::optional<LSN> DflyCmd::ParseLsnVec(std::string_view last_master_lsn,
...
+    // Kill Connection here
+    facade::Connection* dfly_conn = static_cast<facade::Connection*>(conn);
+    using Phase = facade::Connection::Phase;
+    auto phase = dfly_conn->phase();
+    if (dfly_conn->cntx() && dfly_conn->cntx()->replica_conn) {
+      return;
+    }
+
+    bool idle_read = phase == Phase::READ_SOCKET && dfly_conn->idle_time() > timeout;
```
dranikpg, line 486: is idle_read really an interesting state?

src/server/dflycmd.cc
dranikpg: Connections stuck on socket writes are not all the kinds of bad connections

dranikpg (commented):
I thought about it today. I haven't edited the code in years, but it seems that we don't really have clearly defined boundaries anymore in what the process should be _in theory_

The rationale a long time ago was really simple: once we set TAKEN_OVER state, not all connections might have finished running transactions. So to keep data integrity/operation atomicity, we have to wait for all of them to finish. Nothing more.

It was easiest to base this mechanism just on a new message type, because we already had the queue set up with different types.

src/server/dflycmd.cc
```diff
@@ -520,13 +561,22 @@ void DflyCmd::TakeOver(CmdArgList args, RedisReplyBuilder* rb, ConnectionContext
...
+    // Force takeover on the same duration if flag is set
+    if (absl::GetFlag(FLAGS_experimental_force_takeover)) {
+      ForceShutdownStuckConnections(uint64_t(timeout));
+
+      // Safety net.
+      facade::DispatchTracker tracker{sf_->GetNonPriviligedListeners(), cntx->conn(), false, false};
+      shard_set->pool()->AwaitFiberOnAll([&](unsigned index, auto* pb) {
+        sf_->CancelBlockingOnThread();
+        tracker.TrackOnThread();
+      });
```
dranikpg, line 575: You can shut them down from this callback. The most important detail is that we care only about connections that have been running before taken over was set. We can't track them anymore here, but for example, if a connection connected in-between ForceShutdownStuckConnections and this callback and caused you trouble - we don't care about it - it can't do any mutable operations

src/server/dflycmd.cc
dranikpg: I see, being blocked on pipeline overflow

src/server/dflycmd.h
```diff
@@ -183,6 +183,11 @@ class DflyCmd {
   // Switch to stable state replication.
   void StartStable(CmdArgList args, Transaction* tx, RedisReplyBuilder* rb);
 
+  // Helper for takeover flow. Sometimes connections get stuck on send (because of pipelines)
+  // and this causes the takeover flow to fail because checkpoint messages are not processed.
+  // This function force shuts down those connection and allows the node to complete the takeover.
+  void ForceShutdownStuckConnections(uint64_t timeout);
```
dranikpg, line 189: It isn't necessarily a pipeline, it can also be a connection stuck on a large send

dranikpg:
I will work and it will be more reliable than what we currently have, I just have this feeling that we overcomplicated because of not changing the message approach. Please see the small comments

> However, this case will be fixed once we switch to asynchronous execution of commands

Maybe this will be a good moment to transition to a better approach

src/server/dflycmd.cc
kostasrim: +1. I will also sent a separate PR as we have this bug in another place as well

src/server/dflycmd.cc
kostasrim: +1, but I kill them all anyway. Even for connections stuck on sent, it could be that the client reads so the send gets unstuck. We should shutdown the connection regardless as we can't know when the client will read

src/server/dflycmd.cc
kostasrim: >  We can't track them anymore here, but for example, if a connection connected in-between ForceShutdownStuckConnections and this callback and caused you trouble - we don't care about i

That's true but is there a point to "filter/select" these? Just kill them all. We are on `force path` anyway

src/server/dflycmd.cc
```diff
@@ -40,6 +40,8 @@ ABSL_RETIRED_FLAG(uint32_t, allow_partial_sync_with_lsn_diff, 0,
 ABSL_DECLARE_FLAG(bool, info_replication_valkey_compatible);
 ABSL_DECLARE_FLAG(uint32_t, replication_timeout);
 ABSL_DECLARE_FLAG(uint32_t, shard_repl_backlog_len);
+ABSL_FLAG(bool, experimental_force_takeover, false,
```
kostasrim, line 43: @romange @dranikpg 

Maybe instead introduce a subcommand to takeover, like `force` instead of a flag ? That way if we have a failed Takeover we can follow up with Forced one ?

src/server/dflycmd.cc
kostasrim: Yes but that's for the logs, we kill all the connections

kostasrim:
> It was easiest to base this mechanism just on a new message type, because we already had the queue set up with different types.

Are you suggesting this or was it informational ? I mean, we got Checkpoint which we don't process because we are stuck. Am I missing something ?

src/server/dflycmd.cc
romange: I prefer we have the flag and improve TAKEOVER than adding those micro behaviors.

src/server/dflycmd.cc
romange: I do not understand actually. How idle read can happen for stuck connections?

romange:
@kostasrim  can you please write in the PR description the **existing** algorithm behind takeover? specifically, Is it a stateful flow operation where we tell listener to stop accepting new connections, mark all connections as shutting down and trying to break the existing flows or we just try to break the existing connections while new ones may be added, during this time ? Maybe it's time to revisit the algorithm and to implement something sound?

romange:
> The rationale a long time ago was really simple: once we set TAKEN_OVER state, not all connections might have finished running transactions. So to keep data integrity/operation atomicity, we have to wait for all of them to finish. Nothing more.

I read it after I wrote my comment and I approve this message :)  
What is the process now and what it should actually be - let's start with high level english.

kostasrim:
> @kostasrim can you please write in the PR description the **existing** algorithm behind takeover? specifically, Is it a stateful flow operation where we tell listener to stop accepting new connections, mark all connections as shutting down and trying to break the existing flows or we just try to break the existing connections while new ones may be added, during this time ?

I updated the description with the algorithm + some small explanations. 

> Maybe it's time to revisit the algorithm and to implement something sound?

Yes but plz read what I wrote in "thinking out loud" https://github.com/dragonflydb/dragonfly/issues/6114#issuecomment-3574709795.  I would like to revisit this and I am happy to write a small doc internally on what to do but I would first wait for the changes in the connection fiber/pipelining we are implementing. Once that's in place, we will adjust/redesign accordingly. When we become fully asynchronous for example, a `Checkpoint` message will be processed `immediately` since now the `connection fiber` is a `multiplexer` of different events. However, there might be multiple in-flight commands. Obviously the interactions there are slightly different and I don't want to rush any decisions without having the asynchronicity first set in stone.

My rationale for this PR is to for now have a temporary solution and allow dragonfly to complete the takeover in an intrusive/forceful way until we slowly transition to a new approach.

romange:
I added additional logs which I hope will shed more light on why takeover fails BESIDES the stuck send explanation.
Btw, stuck send is relatively rare phenomena for us and we have a solution for this which should decrease the  impact on takeover drastically. All in all I suggest we take it offline as it's not super urgent to solve now.

src/server/dflycmd.cc
kostasrim: @romange because in DispatchSingle we block when pipeline is over limits *but* we set `last_interaction_ = time(nullptr);` after we block. This shall reflect that. 

Either way I kill all the connections unconditionally. I added the logs because I wanted us to be aware of those stucked cases

kostasrim:
@dranikpg hope I didn't forget anything

romange:
I think we can close it

romange (changes requested):
lets close it

Issue #6120, opened 2025-11-26
Review numeric range tree performance


dranikpg:
Numeric range tree can have write performance issues with most likely a small set of different values.

dfly::search::RangeTree::SplitBlock takes up almost 90% of cpu time during insertions

dranikpg:
```go
func Period(p int, rdb *redis.Client) {
	start := time.Now()
	for i := range 100 {
		p := rdb.Pipeline()
		for j := 0; j < 1000; j++ {
			key := fmt.Sprintf("key%v:%v:%v", p, i, j)
			p.HMSet(ctx, key, "i", j%2)
		}
		p.Exec(ctx)
	}
	fmt.Println("Took", p, time.Since(start))
}
```

dranikpg:
cumulative
```
Showing top 50 nodes out of 103
      flat  flat%   sum%        cum   cum%
     0.04s 0.064% 0.064%     57.69s 92.90%  [libboost_context.so.1.88.0]
         0     0% 0.064%     57.66s 92.85%  __invoke (inline)
         0     0% 0.064%     57.66s 92.85%  __invoke_impl (inline)
         0     0% 0.064%     57.66s 92.85%  boost::context::detail::fiber_entry (inline)
         0     0% 0.064%     57.66s 92.85%  fiber_entry
         0     0% 0.064%     57.66s 92.85%  invoke (inline)
         0     0% 0.064%     57.66s 92.85%  operator() (inline)
         0     0% 0.064%     57.66s 92.85%  run (inline)
         0     0% 0.064%     55.96s 90.11%  __apply_impl (inline)
         0     0% 0.064%     55.96s 90.11%  apply (inline)
         0     0% 0.064%     55.96s 90.11%  run_ (inline)
         0     0% 0.064%     55.34s 89.11%  std::function::operator() (inline)
         0     0% 0.064%     55.31s 89.07%  __invoke_r (inline)
         0     0% 0.064%     55.31s 89.07%  std::_Function_handler::_M_invoke (partial-inline)
         0     0% 0.064%     55.31s 89.07%  util::fb2::FiberQueue::Run
     0.02s 0.032% 0.097%     55.30s 89.05%  dfly::MultiCommandSquasher::SquashedHopCb
         0     0% 0.097%     55.23s 88.94%  dfly::Service::InvokeCmd
         0     0% 0.097%     55.22s 88.92%  dfly::CommandId::Invoke
[...]

PR #6143, opened 2025-12-01
fix(search): Fix numeric tree splits

dranikpg:
**Intro**:
Only Stephan worked on this code with me as the reviewer, so a quick recap:

Range tree is like a binary search tree that stores `double lower_bound -> sorted_vector <pair<double, DocId>>` nodes, so we use lower_bound to determine for a new value into which sorted vector to insert. When those vectors (blocks) become too large, we split them. See header comment for more details

**Cause:**
Currently, if a block grows beyond a defined max size, we split it by the median value into a left and right part. Documents with the same value have to be inside one block - so the split can be uneven. If we add the same value to the tree over and over, we will get a single block that is split over and over - yet the split operation doesn't make it smaller at all, everythign ends up in the same block. It creates a doom loop where on each insertion of the same value the block is "split" over and over without results, making it a hugely expensive operation.

**Solution**:
1. I added a max_seen parameter to detect cases where the full block surely consists only of one value
2. If the blocks lower and max seen bounds are equal - it surely has only one kind of value - so we can avoid splitting it, we call it a large "monoblock"
3. If we try to insert a different value into a monoblock, we don't do it - we already know how it will split (into N:1) and just create a new block
4. Yes, the max_seen parameter is only an estimation and doesn't update on deletes, as well as the lower bound, so the optimization ins't definite - however Split() now acts in a way as a "purifier" by recomputing lower-upper bounds, so after a few inefficient splits the bounds will be sufficiently tight to separate monovalue blocks
5. Added "block collapsing" to avoid keeping small almost empty blocks

Relevant to #6120

### DIFF
--- a/src/core/search/block_list.h
+++ b/src/core/search/block_list.h
@@ -217,8 +217,11 @@ struct SplitResult {
 
   Container left;
   Container right;
+
+  // Median value of split, used as minimum value of right block
   double median;
 
+  // Min/max values of left (lmin, lmax) and right (rmin=median, rmax) blocks
   double lmin, lmax, rmax;
 };
 }  // namespace dfly::search
--- a/src/core/search/range_tree.cc
+++ b/src/core/search/range_tree.cc
@@ -78,6 +78,8 @@ RangeTree::RangeTree(PMR_NS::memory_resource* mr, size_t max_range_block_size,
     : max_range_block_size_(max_range_block_size),
       entries_(mr),
       enable_splitting_(enable_splitting) {
+  // The tree has at least always a block with a negative infinity bound, so that any new insertion
+  // goes at least somewhere
   CreateEmptyBlock(-std::numeric_limits<double>::infinity());
 }
 
@@ -85,15 +87,16 @@ void RangeTree::Add(DocId id, double value) {
   DCHECK(std::isfinite(value));
 
   auto it = FindRangeBlock(value);
-  auto& [lb, block] = *it;
+  auto& [lower_bound, block] = *it;
 
   // Don't disrupt large monovalue blocks, instead create new nextafter block
   if (enable_splitting_ && block.Size() >= max_range_block_size_ &&
-      lb == block.max_seen /* monovalue */ && value != lb /* but new value is different*/
+      lower_bound == block.max_seen /* monovalue */ &&
+      value != lower_bound /* but new value is different*/
   ) {
     // We use nextafter as the lower bound to "catch" all other possible inserts into the block,
     // as a decreasing `value` sequence would otherwise create lots of single-value blocks
-    double lb2 = std::nextafter(lb, std::numeric_limits<double>::infinity());
+    double lb2 = std::nextafter(lower_bound, std::numeric_limits<double>::infinity());
     CreateEmptyBlock(lb2)->second.Insert({id, value});
     return;
   }
@@ -105,7 +108,7 @@ void RangeTree::Add(DocId id, double value) {
     return;
 
   // Large monovalue block, not reducable by splitting
-  if (lb == block.max_seen)
+  if (lower_bound == block.max_seen)
     return;
 
   SplitBlock(it);
@@ -122,7 +125,9 @@ void RangeTree::Remove(DocId id, double value) {
 
   // Merge with left block if both are relatively small and won't be forced to split soon
   if (block.size() < max_range_block_size_ / 4 && it != entries_.begin()) {
-    auto lit = --it;
+    auto lit = it;
+    --lit;
+
     auto& lblock = lit->second;
     if (block.Size() + lblock.Size() < max_range_block_size_ / 2) {
       for (auto e : block)
@@ -243,7 +248,7 @@ TODO: we can optimize this case by splitting to three blocks:
  - empty right block with range [std::nextafter(m, +inf), r)
 */
 void RangeTree::SplitBlock(Map::iterator it) {
-  double l = it->first;
+  double lower_bound = it->first;
 
   auto split_result = Split(std::move(it->second));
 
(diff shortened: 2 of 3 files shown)


src/core/search/range_tree.cc
```diff
@@ -109,10 +111,16 @@ void RangeTree::Remove(DocId id, double value) {
...
+  if (block.size() < max_range_block_size_ / 4 && it != entries_.begin()) {
+    auto lit = --it;
+    auto& lblock = lit->second;
+    if (block.Size() + lblock.Size() < max_range_block_size_ / 2) {
+      for (auto e : block)
+        lblock.Insert(e);
+      entries_.erase(it);
+    }
+  }
 }
```
dranikpg, line 139: Not having any kind of compaction is also problematic, especially for long running workloads

dranikpg:
After:
```
-----------------------------------------------------------
Benchmark                 Time             CPU   Iterations
-----------------------------------------------------------
BM_Insertion/1         23.5 ns         23.4 ns     32366724
BM_Insertion/2         29.4 ns         29.3 ns     27719627
BM_Insertion/12        37.6 ns         37.5 ns     18802518
BM_Insertion/128       57.7 ns         57.7 ns     11901372
```

Before:
```
-------------------------------------------------------------------------------------
Benchmark                                   Time             CPU   Iterations
-------------------------------------------------------------------------------------
BM_Insertion/1/iterations:30000        262567 ns       262043 ns        30000
BM_Insertion/2/iterations:30000         72688 ns        72581 ns        30000
BM_Insertion/12/iterations:30000         56.2 ns         56.2 ns        30000
BM_Insertion/128/iterations:30000        36.1 ns         36.1 ns        30000
```

(I had to limit iterations to get reaonable running time, it doesn't get slower for larger value sets)

romange:
Curious, why same scores must be in the same block?

src/core/search/block_list.h
```diff
@@ -218,5 +218,7 @@ struct SplitResult {
   Container left;
   Container right;
   double median;
+
+  double lmin, lmax, rmax;
```
romange, line 225: please add a comment explaining the meaning of these numbers.  now it's easy to see as I know the context but will be more confusing if reading the code

src/core/search/range_tree.h
```diff
@@ -33,15 +33,31 @@ class RangeTree {
...
-  static constexpr size_t kDefaultMaxRangeBlockSize = 7000;
-  static constexpr size_t kBlockSize = 400;
+  // Main node of numeric tree
+  struct RangeBlock : public BlockList<SortedVector<Entry>> {
+    template <typename... Ts>
+    explicit RangeBlock(PMR_NS::memory_resource* mr, Ts... ts) : BlockList{mr, ts...} {
+    }
+
+    RangeBlock(BlockList<SortedVector<Entry>>&& bs, double maxv)
+        : BlockList{std::move(bs)}, max_seen{maxv} {
```
romange, line 45: why do we need max_seen if we hold sorted vectors. is not it the last item? 
or the block is sorted by docids?

src/core/search/range_tree.cc
```diff
@@ -78,26 +78,37 @@ RangeTree::RangeTree(PMR_NS::memory_resource* mr, size_t max_range_block_size,
...
-      std::forward_as_tuple(entries_.get_allocator().resource(), max_range_block_size_));
+  CreateEmptyBlock(-std::numeric_limits<double>::infinity());
 }
 
 void RangeTree::Add(DocId id, double value) {
   DCHECK(std::isfinite(value));
 
   auto it = FindRangeBlock(value);
-  RangeBlock& block = it->second;
+  auto& [lb, block] = *it;
```
romange: please rename `lb` to `low_bound`

src/core/search/range_tree.cc
```diff
@@ -109,10 +120,17 @@ void RangeTree::Remove(DocId id, double value) {
...
-  // So we can do merge and split for one operation.
-  // Or in common cases users do not remove a lot of documents?
+  // Merge with left block if both are relatively small and won't be forced to split soon
+  if (block.size() < max_range_block_size_ / 4 && it != entries_.begin()) {
+    auto lit = --it;
+    auto& lblock = lit->second;
+    if (block.Size() + lblock.Size() < max_range_block_size_ / 2) {
+      for (auto e : block)
+        lblock.Insert(e);
+      entries_.erase(it);
```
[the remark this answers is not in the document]
romange: 2 AI : 0 Vlad.

btw, it means we do not have test coverage for this codepath.

src/core/search/range_tree.h
dranikpg: yes, they are sorted primarily by doc id

src/core/search/range_tree.cc
```diff
@@ -221,37 +243,44 @@ TODO: we can optimize this case by splitting to three blocks:
  - empty right block with range [std::nextafter(m, +inf), r)
 */
 void RangeTree::SplitBlock(Map::iterator it) {
-  const RangeNumber l = it->first;
+  double l = it->first;
```
romange: please rename to low_bound

src/core/search/range_tree.cc
```diff
@@ -221,37 +243,44 @@ TODO: we can optimize this case by splitting to three blocks:
...
 
-  const RangeNumber m = split_result.median;
+  const double m = split_result.median;
   DCHECK(!split_result.right.Empty());
 
   entries_.erase(it);
+  stats_.splits++;
+
+  // Insert left block if its not empty, but always keep left infinity bound
+  if (!split_result.left.Empty() || std::isinf(l)) {
```
romange: in what cases `std::isinf(l)` is true?

dranikpg:
> Curious, why same scores must be in the same block?

Blocks store all values in a range `[ lower_bound[i], lower_bound[i+1] )`, we use tree lower_bounds to search for all blocks in a range. In theory, this can be done with a sorted multimap as well - but then the deletion problem comes in - you don't need immediately where to look for a distinct value/doc_id

src/core/search/range_tree.cc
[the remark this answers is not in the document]
dranikpg: Ah, damn, indeed 😭😭😭  And it requires a better test...

src/core/search/range_tree.cc
dranikpg: > but always keep left infinity bound

the first nodes left bound is always -inf, so a new value is larger than at least some lower bound

romange: ok, I had to read that std::inf works on negative infinity as well. Maybe worth adding a comment that you keep left `-inf bound`

romange:
I did not see where you do the purifying action (4). Specifically, where do you update max_seen in case it is stale, i.e. say `lb=1` and `max_seen=2` and you start splitting only to discover that you can not split.

dranikpg:
> I did not see where you do the purifying action (4). Specifically, where do you update max_seen in case it is stale, i.e. say lb=1 and max_seen=2 and you start splitting only to discover that you can not split.

I will split resultless _once_, but `Split()` computes lmin, lmax, rmin, rmax, so the new block inserted will have tighter bounds. So if it consited only of a single value v after many deletions and updates, then lb=v=max_seen will be the case after split

PR #6144, opened 2025-12-02
feat(search): Defragment search indices

abhijat:
Adds support for defragmenting tag indices. The following objects support defragmentation:

* SortedVector: if the memory address is in underutilized page, entries are reallocated and moved
* BlockList: reallocate the vector of `Container`s if necessary, defragment containers if they support it
* TagIndex: iterates over rax tree for `entries_` and `suffix_trie_`, accesses values (which should be `BlockList`) and defragments them. If operation runs out of quota, store the key so that it can be used to resume defragmentation later
* FieldIndices: defragment `indices_` and `sort_indices_`, storing keys if quota runs out. 
* ShardDocIndices: defragment all indices, store key if quota runs out.

The quota is specified in terms of microseconds the entire operation can run for. The page usage object now stores the cycles at creation time. The quota is also stored, converted to cycles. The quota-depleted check is defined on this object, and checks whether more quota cycles have passed since object creation.

Each defragment operation first checks if quota has run out, if so it returns early. Some objects keep track of state so that if quota runs out, the next defragment starts from where the previous attempt left off.

The accuracy of defragmenting an object only once per cycle when working with FieldIndices or higher level is not guaranteed. For example where an object contains a flat hash map, if the defragmentation resumes from a key and new elements were added in the interim, the order of iteration may change, and an object which was defragmented before may be visited again at the expense of another object which was never defragmented.

Since this will happen only when new indices are added or fields are added to existing ones, this discrepancy should be acceptable.

The tag index _will_ undergo frequent mutation, meaning that the rax tree lookup might fail when looking for the stored key to begin defragmentation. In this case `lower_bound` is used to start defragmentation from the nearest larger key, because the previous blocklists have a high chance of already being defragmented.

No state is maintained in sorted vector or blocklists, we also do not check for quota after moving each element of the sorted vector. This is because it is not possible to partially move a vector, it has to be a single operation.

The defragment process is also not hooked up to the defrag task yet to keep the PR small, this will be done in the next PR where `ShardDocIndices::Defragment` can be called with a runtime based quota.

### DIFF
--- a/src/core/search/block_list.cc
+++ b/src/core/search/block_list.cc
@@ -4,15 +4,19 @@
 
 namespace {
 
-template <typename T>
-using CanDefragmentT = decltype(std::declval<T>().Defragment(std::declval<size_t>(),
-                                                             std::declval<dfly::PageUsage*>()));
+template <typename T> bool DefragmentVector(PMR_NS::vector<T>& vec, dfly::PageUsage* page_usage) {
+  if (vec.empty() || !page_usage->IsPageForObjectUnderUtilized(vec.data())) {
+    return false;
+  }
 
-template <typename, typename = void> struct CanDefragment : std::false_type {};
-
-template <typename T> struct CanDefragment<T, std::void_t<CanDefragmentT<T>>> : std::true_type {};
-
-template <typename T> inline constexpr bool CanDefragmentV = CanDefragment<T>::value;
+  PMR_NS::vector<T> new_vec(vec.get_allocator());
+  new_vec.reserve(vec.size());
+  for (auto&& element : vec) {
+    new_vec.push_back(std::move(element));
+  }
+  vec = std::move(new_vec);
+  return true;
+}
 
 }  // namespace
 
@@ -137,11 +141,9 @@ DefragmentResult BlockList<Container>::Defragment(PageUsage* page_usage) {
     result.objects_moved += 1;
   }
 
-  if constexpr (CanDefragmentV<Container>) {
-    for (Container& block : blocks_) {
-      if (result.Merge(block.Defragment(quota_usec, page_usage)).quota_depleted) {
-        break;
-      }
+  for (Container& block : blocks_) {
+    if (result.Merge(block.Defragment(page_usage)).quota_depleted) {
+      break;
     }
   }
   return result;
--- a/src/core/search/compressed_sorted_set.h
+++ b/src/core/search/compressed_sorted_set.h
@@ -91,6 +91,10 @@ class CompressedSortedSet {
     return tail_value_.value();
   }
 
+  static DefragmentResult Defragment([[maybe_unused]] PageUsage* page_usage) {
+    return {};
+  }
+
  private:
   struct EntryLocation {
     IntType value;                        // Value or 0


abhijat:
this will need to be split into moving page usage to its own lib and then the remaining changes

abhijat:
augment review

src/core/search/search.cc
```diff
@@ -622,6 +620,44 @@ void FieldIndices::FinalizeInitialization() {
   }
 }
 
+namespace {
+
+template <typename T>
+DefragmentResult DefragmentIndices(T& indices, PageUsage* page_usage, size_t quota,
```
abhijat: This ended up being very similar to `ShardDocIndices::Defragment`, maybe they can be refactored into one function. RaxTree defragmentation is also similar but slightly different as that iterator cannot be moved or copied.

src/core/search/block_list.cc
```diff
@@ -1,5 +1,21 @@
...
+
+template <typename T>
+using CanDefragmentT = decltype(std::declval<T>().Defragment(std::declval<dfly::PageUsage*>(),
+                                                             std::declval<size_t>()));
+
+template <typename, typename = void> struct CanDefragment : std::false_type {};
+
+template <typename T> struct CanDefragment<T, std::void_t<CanDefragmentT<T>>> : std::true_type {};
+
+template <typename T> inline constexpr bool CanDefragmentV = CanDefragment<T>::value;
```
abhijat: would be much simpler with concepts

abhijat: another option would be to just use `is_same_v` and check for `SortedVector`

src/core/search/block_list.cc
```diff
@@ -110,6 +126,37 @@ template <typename C> bool BlockList<C>::Remove(ElementType t) {
   return false;
 }
 
+template <typename Container>
+DefragmentResult BlockList<Container>::Defragment(PageUsage* page_usage, size_t quota) {
```
romange: swap arguments - with output at the end

src/core/search/block_list.cc
```diff
@@ -110,6 +126,37 @@ template <typename C> bool BlockList<C>::Remove(ElementType t) {
...
+    return DefragmentResult{.quota_depleted = true, .objects_moved = 0};
+  }
+
+  DefragmentResult result;
+  if (blocks_.empty()) {
+    return result;
+  }
+
+  if (page_usage->IsPageForObjectUnderUtilized(blocks_.data())) {
+    PMR_NS::vector<Container> new_blocks;
```
romange: don't you need the memory resource of `blocks_` for that?

src/core/search/base.h
```diff
@@ -101,6 +113,11 @@ struct BaseIndex {
      Some indices may need to finalize internal structures. See RangeTree for example. */
   virtual void FinalizeInitialization() {
   }
+
+  // Defragments the index by moving objects in underutilized pages to the current malloc page.
+  virtual DefragmentResult Defragment(PageUsage* page_usage, size_t quota) {
```
romange: move page_usage to the end

src/core/search/block_list.cc
```diff
@@ -262,6 +309,29 @@ template <typename T> std::pair<SortedVector<T>, SortedVector<T>> SortedVector<T
...
+  }
+
+  DefragmentResult result;
+  if (entries_.empty()) {
+    return result;
+  }
+
+  if (page_usage->IsPageForObjectUnderUtilized(entries_.data())) {
+    result.objects_moved += 1;
+    PMR_NS::vector<T> new_entries;
```
romange: please use the same memory_resource

abhijat: oh right, will fix

src/core/search/base.cc
```diff
@@ -25,4 +28,14 @@ std::optional<double> ParseNumericField(std::string_view value) {
...
 }
 
+DefragmentResult& DefragmentResult::Merge(DefragmentResult&& other) {
+  quota_depleted |= other.quota_depleted;
+  objects_moved += other.objects_moved;
+  return *this;
+}
+
+bool QuotaDepleted(const size_t quota_ns) {
+  return base::CycleClock::ToUsec(util::ThisFiber::GetRunningTimeCycles()) >= quota_ns;
```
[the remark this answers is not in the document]
abhijat: It should be microseconds, I will rename the description.

src/core/search/block_list.cc
```diff
@@ -262,6 +309,29 @@ template <typename T> std::pair<SortedVector<T>, SortedVector<T>> SortedVector<T
...
+  }
+
+  DefragmentResult result;
+  if (entries_.empty()) {
+    return result;
+  }
+
+  if (page_usage->IsPageForObjectUnderUtilized(entries_.data())) {
+    result.objects_moved += 1;
+    PMR_NS::vector<T> new_entries(mr_);
```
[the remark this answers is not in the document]
abhijat: this looks better than keeping around a pointer to the memory resource.

src/core/search/base.cc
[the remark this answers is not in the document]
abhijat: changed description and variable name

src/core/page_usage_stats_test.cc
```diff
@@ -213,3 +215,107 @@ TEST_F(PageUsageStatsTest, JSONCons) {
...
+  search::BlockList<search::CompressedSortedSet> bl{&m_, 20};
+  PageUsage p{CollectPageStats::YES, 0.1};
+  p.SetForceReallocate(true);
+
+  // empty list
+  auto result = bl.Defragment(&p, MAX_QUOTA);
+  EXPECT_FALSE(result.quota_depleted);
+  EXPECT_EQ(result.objects_moved, 0);
+
+  // will move reallocate once for the blocklist, the inner sorted set will be skipped
```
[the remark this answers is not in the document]
abhijat: fixed

src/core/search/block_list.cc
```diff
@@ -110,6 +126,37 @@ template <typename C> bool BlockList<C>::Remove(ElementType t) {
...
+    return DefragmentResult{.quota_depleted = true, .objects_moved = 0};
+  }
+
+  DefragmentResult result;
+  if (blocks_.empty()) {
+    return result;
+  }
+
+  if (page_usage->IsPageForObjectUnderUtilized(blocks_.data())) {
+    PMR_NS::vector<Container> new_blocks(mr_);
```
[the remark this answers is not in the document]
abhijat: changed

src/core/search/block_list.cc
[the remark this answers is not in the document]
abhijat: changed

src/core/search/search.cc
abhijat: all defrag-map style code is refactored into one class now

src/core/search/base.h
abhijat: fixed

src/core/search/block_list.cc
abhijat: fixed

src/core/search/block_list.cc
abhijat: fixed

src/core/search/block_list.cc
abhijat: fixed

abhijat:
augment review

src/core/search/base.cc
```diff
@@ -25,4 +28,14 @@ std::optional<double> ParseNumericField(std::string_view value) {
   return std::nullopt;
 }
 
+DefragmentResult& DefragmentResult::Merge(DefragmentResult&& other) {
+  quota_depleted |= other.quota_depleted;
+  objects_moved += other.objects_moved;
+  return *this;
+}
+
+bool QuotaDepleted(const size_t quota_usec) {
```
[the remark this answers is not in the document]
abhijat: I originally thought it might be best to let the defrag task stop if the fiber had been running for too long, but it looks like the task is run as a function in the main proactor loop rather than on a distinct fiber. 

So there is a slight change now, the page usage object tracks its creation time and checks quota against that timestamp. 

Since the page usage object is created at the start of each defragmentation task, this should also clear the way for compact object defragmentation using time as its limit instead of number of operations at some point in future.

abhijat:
augment review

abhijat:
its a large PR now but about half the code is tests

abhijat:
augment review

src/core/page_usage/page_usage_stats.cc
```diff
@@ -179,4 +181,10 @@ bool PageUsage::ConsumePageStats(mi_page_usage_stats_t stat) {
   return force_reallocate_ || should_reallocate;
 }
 
+bool PageUsage::QuotaDepleted(const size_t quota_usec) const {
+  static constexpr auto MAX_QUOTA = std::numeric_limits<size_t>::max();
+  return quota_usec < MAX_QUOTA &&
+         static_cast<size_t>(absl::GetCurrentTimeNanos() / 1000) - start_time_usec_ >= quota_usec;
+}
```
dranikpg: We use `CycleClock` for measuring time intervals with a lower overhead

src/core/page_usage/page_usage_stats.h
```diff
@@ -66,6 +66,8 @@ class PageUsage {
     force_reallocate_ = force_reallocate;
   }
 
+  bool QuotaDepleted(size_t quota_usec) const;
+
```
dranikpg: Are you sure about this kind of interface? I would really want to introduce it globally and it seems like storing the quota itself would be simpler. Moreover, maybe it should be a subclass? Either way, thats of future refacotring

src/core/search/block_list.cc
dranikpg: Why at all? You can have an empty Defragment() inside CompressedSortedSet and implement it in your next PR. It's just a wrapper around vector<>, there's nothing complicated inside the compressed set

src/core/search/block_list.cc
```diff
@@ -268,6 +309,29 @@ template <typename T> std::pair<SortedVector<T>, SortedVector<T>> SortedVector<T
   return std::make_pair(std::move(*this), SortedVector<T>{std::move(tail)});
 }
 
+template <typename T>
+DefragmentResult SortedVector<T>::Defragment(size_t quota_usec, PageUsage* page_usage) {
+  if (page_usage->QuotaDepleted(quota_usec)) {
+    return DefragmentResult{.quota_depleted = true, .objects_moved = 0};
+  }
```
dranikpg: that's an inversion of control if this is an atomic operation (it is), the loop in block list should check for quota depleted

src/core/search/block_list.cc
```diff
@@ -268,6 +309,29 @@ template <typename T> std::pair<SortedVector<T>, SortedVector<T>> SortedVector<T
...
+
+  if (page_usage->IsPageForObjectUnderUtilized(entries_.data())) {
+    result.objects_moved += 1;
+    PMR_NS::vector<T> new_entries(entries_.get_allocator());
+    new_entries.reserve(entries_.size());
+    for (auto&& e : entries_) {
+      new_entries.push_back(std::move(e));
+    }
+    entries_ = std::move(new_entries);
```
dranikpg: I discovered this function a while ago

`entries_ = {make_move_iterator(entries_.begin()), make_move_iterator(entries_.end())}`

src/core/search/indices.h
```diff
@@ -210,4 +214,50 @@ struct GeoIndex : public BaseIndex {
   std::unique_ptr<rtree> rtree_;
 };
 
+// Defragments a map like data structure. The values in the map must have a `Defragment` method.
+// Works with rax tree map and hash based maps
+template <typename Container, typename ItFunc> struct DefragmentMap {
+  // ItFunc is necessary because RaxTreeMap does not allow copying or moving iterators, so this
+  // class cannot accept begin,end iterators from caller. It must construct the begin iterator
```
dranikpg, line 221: It can be movable, I just made it immovable as I was too lazy to implement a proper move constructor/operator 😛

src/core/search/indices.h
```diff
@@ -210,4 +214,50 @@ struct GeoIndex : public BaseIndex {
...
+  using Iterator = std::invoke_result_t<ItFunc>;
+  DefragmentMap(Container& container, ItFunc&& f)
+      : container(container), it(f()), end(container.end()) {
+  }
+
+  // Deref should be true if the value is wrapped in a pointer. Set here instead of at class level
+  // to allow overriding without specifying other parameters.
+  template <bool Deref = false>
```
dranikpg: maybe its really shorter, but it might be nicer to check if map is a unique_ptr inside the function without such parameters

src/core/search/indices.h
dranikpg: and then it should be just a function, you can also add an iterator alias it it doesn't have one

src/core/search/indices.h
```diff
@@ -148,13 +148,17 @@ struct TagIndex : public BaseStringIndex<SortedVector<DocId>> {
...
+
  protected:
   std::optional<StringList> GetStrings(const DocumentAccessor& doc,
                                        std::string_view field) const override;
   absl::flat_hash_set<std::string> Tokenize(std::string_view value) const override;
 
  private:
   char separator_;
+  std::string next_defrag_entry_;
+  std::string next_defrag_suffix_entry_;
```
dranikpg, line 161: I agree there's no other way, but ideally the defragmentation code should've been turned into a fiber, so we can have a stack to store values it

dranikpg:
The logic looks good overall, its just question with time management / the interface. What we do here is something we can transfter then onto the rest of the codebase

src/core/search/indices.h
abhijat: I removed it and added overloaded functions to do the check, one overload for pointer-like values and one for plain references.

src/core/search/block_list.cc
abhijat: Yeah the idea was to introduce things gradually but I wasn't sure if the types that blocklist will be used with is fixed or can be extended in future, I think this got a little too complicated though, I made it simpler with overloaded methods instead of this code. 

Now I see it is only used with `CompressedSortedSet` in addition to sorted vector. Maybe I will add a stub like you mentioned.

abhijat: changed to add a stub method to compressed sorted set

src/core/search/block_list.cc
abhijat: Do you mean the loop that iterates over blocks? It does check and breaks if it cannot find quota left.

dranikpg: I mean its nicer if the whole execution graph can be divided into control and leaf nodes. The leaf nodes shouldn't check the quota, the control nodes (top level loops) should

src/core/page_usage/page_usage_stats.h
abhijat: I changed it to use `CycleClock` and store the quota instead of passing it around. It is still stored in `PageUsage` and the quota defaults to max.

src/core/page_usage/page_usage_stats.cc
abhijat: changed as https://github.com/dragonflydb/dragonfly/pull/6144#discussion_r2597117671

src/core/search/block_list.cc
abhijat: Yeah I see, changed to remove the check from sorted vector

abhijat:
augment review

src/core/search/block_list.cc
```diff
@@ -1,5 +1,21 @@
 #include "core/search/block_list.h"
 
+#include "core/page_usage/page_usage_stats.h"
+
+namespace {
+
+template <typename T>
```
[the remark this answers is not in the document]
abhijat: this is removed

src/core/search/indices.h
abhijat: sounds good but I can do this in a follow up pr

abhijat:
I think this is in a good place for a review now, some things might have to be handled in a follow up pr

abhijat:
It can be split this into smaller PRs if it helps review, eg

* adding quota to page usage
* adding defrag support to block list, sorted vector
* everything in the middle

dranikpg:
> It can be split this into smaller PRs if it helps review, eg
> 
> * adding quota to page usage
> * adding defrag support to block list, sorted vector
> * everything in the middle

It would've been easier if pushed the new changes as separate commits

src/core/page_usage/page_usage_stats.h
abhijat: I extracted a new `CycleQuota` class from `PageUsage` as part of https://github.com/dragonflydb/dragonfly/pull/6262

Issue #6147, opened 2025-12-02
lua: add reply-float-as-int flag to scripts


romange:
Dragonfly is incompatible by default with some libraries due to its native float support in lua.
Specifically, https://github.com/go-redis/redis_rate and https://github.com/Suor/django-cacheops/ 

require that Dragonfly will run with `--lua_resp2_legacy_float` which is sub-optimal.
 
Idea: Similarly to how we have `undeclared_keys` in `ScriptParams` struct, introduce `reply_float_as_int`.
and similarly to how we have `const char* kUndeclaredShas[]` inside script_mgr.cc, introduce a hard-coded whitelist of shas  - currently populated by shas of these frameworks.

The goal: both frameworks can work with dragonfly out of the box without running dragonfly with `--lua_resp2_legacy_float`.

PR #922, opened 2023-03-08
fix(core): Auto cast doubles to ints in lua scripts

dranikpg:

PR #6149, opened 2025-12-02
feat(lua): add legacy-float flag for Lua 5.1 number compatibility

vyavdoshenko:
Improves Lua number handling for compatibility.

cjson.decode now returns integers for numbers (unconditional)
```
redis-cli EVAL "local data = cjson.decode(ARGV[1]); return 'foo_'..data.id" 0 '{"id": 101}'
```
- Before: "foo_101.0" 
- After: "foo_101"

This fix is applied in lua_cjson.c.

django-cacheops: Now works with just --default_lua_flags=allow-undeclared-keys (no legacy-float needed)

Fixes: https://github.com/dragonflydb/dragonfly/issues/6147

### DIFF
--- a/src/server/main_service.cc
+++ b/src/server/main_service.cc
@@ -378,7 +378,7 @@ class InterpreterReplier : public RedisReplyBuilder {
 // Serialized result of script invocation to Redis protocol
 class EvalSerializer : public ObjectExplorer {
  public:
-  explicit EvalSerializer(RedisReplyBuilder* rb, bool float_as_int = false)
+  explicit EvalSerializer(RedisReplyBuilder* rb, bool float_as_int)
       : rb_(rb), float_as_int_(float_as_int) {
   }
 


src/core/interpreter.cc
```diff
@@ -807,6 +812,57 @@ auto Interpreter::RunFunction(string_view sha, std::string* error) -> RunResult
...
+      return floor(obj)
+    end
+    return obj
+  end
+
+  cjson.decode = function(str)
+    return fix_numbers(orig_cjson_decode(str))
+  end
+end
+)";
```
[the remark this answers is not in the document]
vyavdoshenko: Fixed

src/server/script_mgr.cc
```diff
@@ -52,7 +56,8 @@ ScriptMgr::ScriptMgr() {
   // Build default script flags
   string flags = absl::GetFlag(FLAGS_default_lua_flags);
 
-  static_assert(ScriptParams{}.atomic && !ScriptParams{}.undeclared_keys);
+  static_assert(ScriptParams{}.atomic && !ScriptParams{}.undeclared_keys &&
```
romange, line 59: it's a weird assert, what does it check?

src/server/script_mgr.cc
```diff
@@ -287,6 +293,22 @@ io::Result<string, GenericError> ScriptMgr::Insert(string_view body, Interpreter
       params.undeclared_keys = true;
     }
   }
+
+  const char* kFloatAsIntShas[] = {
+      "8c4dafdf9b6b7bcf511a0d1ec0518bed9260e16d",  // django-cacheops
```
romange: do we have multiple scripts? I saw you changed only one.

romange:
I think it's an ok solution. The core problem is around the following incompatibility against valkey:

```
redis-cli EVAL "local data = cjson.decode(ARGV[1]); return 'foo_'..data.id" 0 '{"id": 101}'
```

and I think we can solve it directly in `cjson.decode`

romange:
specifically, I think you can do a fix in `json_process_value` in lua_cjson.c and do the similar check of `floor(token->value.number)` to decide either to push an integer or the number. This way cjson.decode will be fully compatible on integers.

src/server/script_mgr.cc
dranikpg: the default values of parameters, but yeah its a little weird 😅

src/core/interpreter.cc
```diff
@@ -807,6 +812,57 @@ auto Interpreter::RunFunction(string_view sha, std::string* error) -> RunResult
...
+if type(__dfly__) == "table" and __dfly__.orig_tostring then
+  tostring = __dfly__.orig_tostring
+  cjson.decode = __dfly__.orig_cjson_decode
+  __dfly__.orig_tostring = nil
+  __dfly__.orig_cjson_decode = nil
+end
+)";
+
+  const char* code = enable ? kEnableLegacyFloat : kDisableLegacyFloat;
+  RunSafe(lua_, code, enable ? "@enable_legacy_float" : "@disable_legacy_float");
```
dranikpg: lets store the state and run it only if it changes

dranikpg: because this will decrease performance even for the normal path

src/server/script_mgr.h
```diff
@@ -26,11 +26,13 @@ class ScriptMgr {
   struct ScriptParams {
     bool atomic = true;            // Whether script must run atomically.
     bool undeclared_keys = false;  // Whether script accesses undeclared keys.
+    bool float_as_int = false;     // Whether to return floats as integers.
 
     // Return GenericError if some flag was invalid.
     // Valid flags are:
     // - allow-undeclared-keys -> undeclared_keys=true
     // - disable-atomicity     -> atomic=false
+    // - legacy-float          -> float_as_int=true
```
dranikpg, line 35: maybe it would've been easier if it was a global flag, as I assume that if someones uses it, he'll use it for all scripts... but this is cleaner of course this way

src/server/script_mgr.cc
vyavdoshenko: I reused the same list of django-cacheops mentioned above. Only one of them has the same hash as the master branch. I guess the hashes can be used from previous versions, and my assumption is that we have to make these scripts work.

src/server/script_mgr.h
vyavdoshenko: I think it's better to use it per script

src/core/interpreter.cc
vyavdoshenko: I changed the implementation. Please take a look.

src/redis/lua/cjson/lua_cjson.c
```diff
@@ -1226,9 +1226,21 @@ static void json_process_value(lua_State *l, json_parse_t *json,
...
         lua_pushlstring(l, token->value.string, token->string_len);
         break;;
-    case T_NUMBER:
-        lua_pushnumber(l, token->value.number);
-        break;;
+    case T_NUMBER: {
+        double num = token->value.number;
+        double intpart;
+        /* Check if legacy float mode is enabled via global variable */
+        lua_getglobal(l, "__dfly_legacy_float__");
```
romange: I agree that now it's much cleaner. But I must ask @dranikpg  @vyavdoshenko  why not use lua_pushinteger unconditionally if the number is integer ?

romange: in other words, why do we need `__dfly_legacy_float__` mode at all?

vyavdoshenko: The flag preserves user choice. Scripts migrated from legacy Lua 4.1 can opt into legacy behavior, while new scripts keep Lua 5.4 semantics. Making it unconditional removes that flexibility.

If you insist on making it unconditionally, I will fix it, but in my opinion, it is better to save float behavior for future scripts.

romange: but when this flexibility is useful? lua will implicitly cast any integer to double if needed. Where do we need to return double even though the underlying value is integer?
I think we should be aligned with `RedisTranslator::OnDouble` code - which I am sure was written for the same reason. My preference is to use unconditional casting to integers if possible. @dranikpg WDYT?

dranikpg: I assume, if nobody type asserts the type in code, it has close to no real implications. Python's default json lib does the same for example:

Image

---

> I think we should be aligned with RedisTranslator::OnDouble code

This is something I don't understand and didn't intervene with, because thats unconditional, without checking the floating part at all

romange: what do you mean? we do check it there: `fractpart = modf(d, &intpart);`

romange:
```
 if (abs(fractpart) < kConvertEps && intpart < double(std::numeric_limits<lua_Integer>::max()) &&
      intpart > std::numeric_limits<lua_Integer>::min())
    lua_pushinteger(lua_, static_cast<lua_Integer>(d));
  else
    lua_pushnumber(lua_, d);
```

romange: btw, your PR shows why we added this:
https://github.com/dragonflydb/dragonfly/pull/922/files#diff-f70ba23140c4a2bf68aeb2b085ec3141759d51317d90150f9c328bb9b2fc6907

it's exactly the same reason: ZSCORE returned 1.0 instead of 1 which messed some lua scripts.

romange: bottom line -  i am more and more confident - let's do it unconditional

dranikpg: Ahh, I confused it with `EvalSerializer::OnDouble` which was also changed in this PR

vyavdoshenko: Fixed to uncondition behavior

src/server/main_service.cc
```diff
@@ -378,7 +378,8 @@ class InterpreterReplier : public RedisReplyBuilder {
 // Serialized result of script invocation to Redis protocol
 class EvalSerializer : public ObjectExplorer {
  public:
-  explicit EvalSerializer(RedisReplyBuilder* rb) : rb_(rb) {
+  explicit EvalSerializer(RedisReplyBuilder* rb, bool float_as_int = false)
```
romange: do we need the default argument here? are there cases where we do not pass the flag?

vyavdoshenko: No, we always pass the flag explicitly. I'll remove the default argument.

PR #6150, opened 2025-12-02
fix(tiering): missing comma in dashboard json

goochs:
Super small fix, #6086 malformed the grafana dashboard which prevented importing
image


dranikpg:
@goochs Thanks for finding and fixing it! 🔥

Issue #6131, opened 2025-11-28
Dragonfly latency spikes during full sync replication


jojoxhsieh:
**Describe the bug**
When setting up a DragonflyDB replica to sync from a primary instance with 32 million keys, we observe significant latency spikes (p99 response time jumps from 4ms to 50ms) on the primary during the full sync process. This issue persists even under moderate load conditions (around 132K ops/sec). The latency spikes are detrimental to our application's performance and user experience.

**To Reproduce**
1. Start a DragonflyDB primary instance with below command:
   ```
   docker run --rm --name dfly0 -d -p 16379:6379 --ulimit memlock=-1 --cpus=64 docker.dragonflydb.io/dragonflydb/dragonfly:v1.35.1 --cache_mode=true --maxmemory=64g --point_in_time_snapshot=false --background_snapshotting=false --dbfilename=
   ```
2. Populate the primary with 32 million keys using a data population script.
3. Start a DragonflyDB replica instance with below command:
   ```
   docker run --rm --name dfly1 -d -p 26379:6379 --ulimit memlock=-1 --cpus=64 docker.dragonflydb.io/dragonflydb/dragonfly:v1.35.1 --cache_mode=true --maxmemory=64g --point_in_time_snapshot=false --background_snapshotting=false --dbfilename=
   ```
4. Use `memtier_benchmark` to generate a load of around 244K ops/sec on the primary as below:
   ```
   docker run --rm -d -v /$(pwd):/home redislabs/memtier_benchmark -s <primary_ip> -p 16379 --data-size=256  --command="MGET __key__ __key__ __key__ __key__ __key__" --command-ratio=10 --command="SET __key__ __data__" --command-ratio=1 --test-time=60 --rate-limiting=660 --json-out-file=/home/memtier_out.json
   ```
5. Configure the replica to sync from the primary:
    ```
    redis-cli -h <replica_ip> -p 26379 replicaof <primary_ip> 16379
    ```
6. Monitor the p99 latency on the primary using `memtier_benchmark` output.

**Expected behavior**
The p99 latency on the primary should remain stable or show minimal increase during the replica's full sync process, ideally staying below 10ms.

**Screenshots**
<img width="718" height="352" alt="Image" src="https://github.com/user-attachments/assets/75e2fe8f-46be-4753-80c3-e5b63ff0ebab" />
[...]

PR #6154, opened 2025-12-03
chore: add a tool that creates charts based on memtier metrics data

romange:
The data should be generated with `memtier_benchmark .. --json-out-file memtier_out.json` command. Usage:  `./plot_memtier_latency.py memtier_out.json`

### DIFF
--- a/tools/plot_memtier_latency.py
+++ b/tools/plot_memtier_latency.py
@@ -1,0 +1,565 @@
+#!/usr/bin/env python3
+"""
+Script to read memtier_benchmark JSON output and generate interactive latency charts.
+
+The script generates interactive HTML charts (using Plotly) where you can:
+- Click on legend items to show/hide time series
+- Zoom in/out and pan
+- Hover over data points for detailed information
+
+To generate the JSON file, run memtier_benchmark with the --json-out-file option:
+
+    memtier_benchmark --server <host> --port <port> \\
+        --json-out-file memtier_out.json \\
+        [other options...]
+
+Example:
+    memtier_benchmark --json-out-file memtier_out.json \\
+        --clients 25 --threads 4 --test-time 120 \\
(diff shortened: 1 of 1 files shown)


romange:
Example chart:
[latency_chart.html](https://github.com/user-attachments/files/23902379/latency_chart.html)

Helps with investigation of #6131

tools/plot_memtier_latency.py
```diff
@@ -0,0 +1,565 @@
...
+import matplotlib.pyplot as plt
+import numpy as np
+from pathlib import Path
+import webbrowser
+import tempfile
+import os
+
+# Try to import plotly for interactive charts
+try:
+    import plotly.graph_objects as go
```
vyavdoshenko, line 40: nit: Why is this here? In my opinion, it's better to declare all imports at the head of the file.

tools/plot_memtier_latency.py
```diff
@@ -0,0 +1,565 @@
...
+        "p50": p50_latencies,
+        "p99": p99_latencies,
+        "p99.9": p99_9_latencies,
+        "min": min_latencies,
+        "max": max_latencies,
+        "ops_per_sec": ops_per_sec,
+    }
+
+
+def plot_latency_chart_interactive(data, output_file="latency_chart.html", open_browser=True):
```
vyavdoshenko, line 109: nit: This function is not complex, but very long.

tools/plot_memtier_latency.py
romange: because the code works without plotly as well. it's optional, hence this local import.

tools/plot_memtier_latency.py
romange: I wrote this script using claude code in 5min by iteratively requesting the desired behavior

PR #6156, opened 2025-12-03
feat(search): Simple multi-reader multi-writer mutex for hnsw index

mkaruza:
Implementation of multi-reader multi-writer mutex that supports concurrent writes or reads but not a mix of reads and writes in same time.

Unit test for testing MRMWMutex class.

### DIFF
--- a/src/core/search/mrmw_mutex.h
+++ b/src/core/search/mrmw_mutex.h
@@ -23,12 +23,10 @@ class MRMWMutex {
 
     // If we have any active_runners we need to check lock mode
     if (active_runners_) {
-      while (lock_mode_ != mode) {
-        auto& waiters = GetWaiters(mode);
-        waiters++;
-        GetCondVar(mode).wait(lk);
-        waiters--;
-      }
+      auto& waiters = GetWaiters(mode);
+      waiters++;
+      GetCondVar(mode).wait(lk, [&] { return lock_mode_ == mode; });
+      waiters--;
     } else {
       // No active runners so just update to requested lock mode
       lock_mode_ = mode;
--- a/src/core/search/mrmw_mutex_test.cc
+++ b/src/core/search/mrmw_mutex_test.cc
@@ -7,8 +7,12 @@
 #include <random>
 #include <thread>
 
+#include "absl/flags/flag.h"
 #include "base/gtest.h"
 #include "base/logging.h"
+#include "util/fibers/pool.h"
+
+ABSL_FLAG(bool, force_epoll, false, "If true, uses epoll api instead iouring to run tests");
 
 namespace dfly::search {
 
@@ -18,7 +22,7 @@ namespace {
 void ReadTask(MRMWMutex* mutex, std::atomic<size_t>& read_count, size_t sleep_time) {
   read_count.fetch_add(1, std::memory_order_relaxed);
   MRMWMutexLock lock(mutex, MRMWMutex::LockMode::kReadLock);
-  std::this_thread::sleep_for(std::chrono::milliseconds(sleep_time));
+  util::ThisFiber::SleepFor(std::chrono::milliseconds(sleep_time));
   read_count.fetch_sub(1, std::memory_order_relaxed);
 }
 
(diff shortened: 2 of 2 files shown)


src/core/search/mrmw_mutex.h
```diff
@@ -0,0 +1,91 @@
...
+  void Lock(LockMode mode) {
+    std::unique_lock lk(mutex_);
+
+    // If we have any active_runners we need to check lock mode
+    if (active_runners_) {
+      while (lock_mode_ != mode) {
+        auto& waiters = GetWaiters(mode);
+        waiters++;
+        GetCondVar(mode).wait(lk);
+        waiters--;
```
romange:
```suggestion
     waiters++;
     GetCondVar(mode).await(lk, [&] { return lock_mode_ == mode;});
     waiters--;
```

src/core/search/mrmw_mutex.h
```diff
@@ -0,0 +1,91 @@
...
+
+  inline util::fb2::CondVar& GetCondVar(LockMode target_mode) {
+    return target_mode == LockMode::kReadLock ? reader_cond_var_ : writer_cond_var_;
+  };
+
+  static inline LockMode GetInverseMode(LockMode mode) {
+    return mode == LockMode::kReadLock ? LockMode::kWriteLock : LockMode::kReadLock;
+  }
+
+  util::fb2::Mutex mutex_;
```
romange, line 61: nit: you also contend on mutex_, though here maybe absl::SpinLock would fit better as you take it for a small period of time. you could use `CondVarAny` to be able to use locks with a different type.

romange: not the most important thing though.

src/core/search/mrmw_mutex.h
```diff
@@ -0,0 +1,91 @@
+// Copyright 2025, DragonflyDB authors.  All rights reserved.
+// See LICENSE for licensing terms.
+//
+
+#include "base/logging.h"
+#include "util/fibers/synchronization.h"
+
+namespace dfly::search {
+
+// Simple implementation of multi-Reader multi-Writer Mutex
```
romange, line 10: nice, now I understand why ReadWrite lock does not work here. you want multi-writers

romange (dismissed):

src/core/search/mrmw_mutex.h
```diff
@@ -0,0 +1,91 @@
...
+// MRMWMutex supports concurrent reads or concurrent writes but not a mix of
+// concurrent reads and writes at the same time.
+
+class MRMWMutex {
```
kostasrim, line 14: What's wrong with ` util::fb2::SharedMutex` ? It supports both READER/WRITER mode and you can easily reuse `std::shared_lock` and `std::unique_lock`. Bonus points you don't have to implement any of this.

src/core/search/mrmw_mutex_test.cc
```diff
@@ -0,0 +1,180 @@
...
+#include "base/logging.h"
+
+namespace dfly::search {
+
+namespace {
+
+// Helper function to simulate reading operation
+void ReadTask(MRMWMutex* mutex, std::atomic<size_t>& read_count, size_t sleep_time) {
+  read_count.fetch_add(1, std::memory_order_relaxed);
+  MRMWMutexLock lock(mutex, MRMWMutex::LockMode::kReadLock);
```
kostasrim, line 24: I would really like to know how this works and does not crash. You use fiber synchronization primitives in a non proactor context (nothing is setup in the test). Something seems wrong here 🤔

romange: fibers work in the main thread as well, see fibers_test for example

kostasrim: Interesting

src/core/search/mrmw_mutex.h
kostasrim: Why do we want the multi writer part though ? Do we have lots of concurrent writes or ?

src/core/search/mrmw_mutex_test.cc
mkaruza: Modified test to use fibers

src/core/search/mrmw_mutex.h
mkaruza: Discussed with @kostasrim

PR #6158, opened 2025-12-03
ci: use OIDC for AWS authentication instead of access keys

AdiCloudbit-DevOps:
## Summary

Replace AWS IAM user access keys with OIDC (OpenID Connect) federation for AWS authentication in CI workflows.

### Composite Actions
- `.github/actions/regression-tests/action.yml` - Made AWS credentials optional, added session token support
- `.github/actions/repeat/action.yml` - Made AWS credentials optional, added session token support

### Workflows
- `.github/workflows/regression-tests.yml` - Use `role-to-assume` instead of access keys
- `.github/workflows/epoll-regression-tests.yml` - Use `role-to-assume` instead of access keys  
- `.github/workflows/repeat-tests.yml` - Use `role-to-assume` instead of access keys
- `.github/workflows/docker-dev-release.yml` - Use `role-to-assume` instead of access keys

### DIFF
--- a/.github/workflows/epoll-regression-tests.yml
+++ b/.github/workflows/epoll-regression-tests.yml
@@ -51,7 +51,7 @@ jobs:
           ls -l ..
 
       - name: Authenticate to AWS
-        uses: aws-actions/configure-aws-credentials@v4
+        uses: aws-actions/configure-aws-credentials@v5
         with:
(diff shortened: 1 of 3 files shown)


kostasrim (dismissed):
lgtm 

I would check if we use anymore aws-access-key-id and remove it completely -- no need to be done now IMO

Also, I would bump up from v4 to v5 as per the copilot suggestions

Issue #6109, opened 2025-11-25
`test_install_package_fedora` and `test_install_package_ubuntu` fail in CI


abhijat:
https://github.com/dragonflydb/dragonfly/actions/runs/19638856133/job/56236547245#step:6:1158

abhijat:
package installation fails for ARM

abhijat:
it is still failing on ubuntu https://github.com/dragonflydb/dragonfly/actions/runs/19663957686/job/56316295748

will try on arm image in ec2

abhijat:
The repository on arm64 does not find the package, even though it is in the bucket. 

The release file for the repository is downloaded but not the packages file. Perhaps the setup of the repo for arm is not correct.

abhijat:
It started to work now on the test arm machine, it seems there was some caching going on with the static website after the last action run a few minutes ago which uploaded the package. 

After I reran the action the package is now visible and installable in the arm test machine.

abhijat:
will keep this open for a couple of days to observe

abhijat:
https://github.com/dragonflydb/dragonfly/actions/runs/19668836159

still failing

abhijat:
On a freshly allocated arm machine in AWS the steps in the test pass as expected.

abhijat:
Running the test itself (which uses docker via testcontainer etc) on an ARM EC2 machine also passes, after setting up docker, python deps etc

abhijat:
note - the previous checks were done with EC2 VM -> docker. In CI there is an extra layer, EC2 VM -> base container -> ubuntu-latest

Perhaps the extra layer causes some issues.

abhijat:
it is no longer failing, seems to just have been an issue with updating the bucket

abhijat:
no longer fails, the tests seem to use the updated repo

abhijat:
https://github.com/dragonflydb/dragonfly/actions/runs/19767462474

started failing again

abhijat:
```
docker.errors.APIError: 500 Server Error for http+docker://localhost/v1.48/images/create?tag=latest&fromImage=fedora: Internal Server Error ("toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit"
```

PR #6161, opened 2025-12-04
tests: Use AWS public docker registry

abhijat:
The default docker registry heavily throttles pull requests. Switch to AWS docker registry to avoid throttling which causes tests to fail.

FIXES https://github.com/dragonflydb/dragonfly/issues/6109


tests/dragonfly/package_install_test.py
```diff
@@ -21,7 +21,7 @@ async def test_install_package_on_fedora():
 
 
 async def test_install_package_on_ubuntu():
-    with DockerContainer(image="ubuntu:latest", tty=True) as ubuntu:
+    with DockerContainer(image="public.ecr.aws/docker/library/ubuntu:noble", tty=True) as ubuntu:
```
abhijat, line 24: our apt repo is only set up for noble

PR #6167, opened 2025-12-05
chore: use BackedArguments for implementing a pipeline message

romange:
Remove duplicate code for collapsing multiple blobs, similarly to https://github.com/dragonflydb/dragonfly/pull/6171

### DIFF
--- a/src/common/backed_args.h
+++ b/src/common/backed_args.h
@@ -18,7 +18,15 @@ class BackedArguments {
   BackedArguments() {
   }
 
-  template <typename I> BackedArguments(I begin, I end);
+  // Construct the arguments from iterator range.
+  // TODO: In general we could get away without the len argument,
+  // but that would require fixing base::it::CompoundIterator to support subtraction.
+  // Similarly, I wish that CompoundIterator supported the -> operator.
+  template <typename I> BackedArguments(I begin, I end, size_t len) {
+    Assign(begin, end, len);
+  }
+
+  template <typename I> void Assign(I begin, I end, size_t len);
 
   size_t HeapMemory() const {
     size_t s1 = lens_.capacity() <= kLenCap ? 0 : lens_.capacity() * sizeof(uint32_t);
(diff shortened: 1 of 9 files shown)


src/common/backed_args.h
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  size_t StorageCapacity() const {
+    return args.capacity() + storage.capacity();
+  }
+
+  // The capacity is chosen so that we allocate a fully utilized (256 bytes) block.
+  using StorageType = absl::InlinedVector<char, kReqStorageSize>;
+
+  absl::InlinedVector<std::string_view, 5> args;
```
dranikpg: You can use a `InlinedVector<uint32_t> sizes` to save 75% memory. Sure, it requires decoding time, but you loose it now as well by fixing references in moves (we move at least twice)

src/common/backed_args.h
```diff
@@ -0,0 +1,57 @@
...
+  BackedArguments(BackedArguments&& o) noexcept {
+    *this = std::move(o);
+  }
+
+  BackedArguments& operator=(BackedArguments&& o) noexcept {
+    args = std::move(o.args);
+    storage = std::move(o.storage);
+    if (!args.empty()) {
+      const char* p = args.front().data();
+      bool is_inline = (p >= o.storage.begin() && p <= o.storage.end());
```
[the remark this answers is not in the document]
dranikpg: it should be `[&o.storage, &o.storage + sizeof(storage)]`

src/common/backed_args.h
```diff
@@ -0,0 +1,57 @@
...
+      if (is_inline) {  // fix references
+        char* next = storage.data();
+        for (auto& arg : args) {
+          arg = std::string_view{next, arg.size()};
+          next += (arg.size() + 1);  //  + \0
```
dranikpg: I was actually wondering if we ever rely on it? StoredCmd is very similar to this class and it does never use null termination

romange: we had some issues around this, yes

src/common/backed_args.h
romange: I am having trouble to use it in `Connection::SquashPipeline` as we need to push into  `vector<ArgSlice>`.
if we use vec<uint32_t> it means we need to introduce another vec<CmdArgVec> to store string_views.

do you have other ideas  @dranikpg ?

dranikpg: You can change the interface. Sadly, the current path isn't optimal as well, because we use StoredCmd there. This approach will actually be even better, as you can turn it into one more easily

dranikpg: The squasher works with stored cmds, so it will be better. Either way, once you push the async dispatch through, it won't be relevant anymore, but it will take some time I assume

src/common/backed_args.h
dranikpg: Strange... because all exec/squashing code work without those for more than a year

romange: I think we had some lua related issues but I do not remember the details.

src/facade/dragonfly_connection.cc
```diff
@@ -1755,25 +1728,15 @@ void Connection::AsyncFiber() {
...
+  if (ptr = GetFromPipelinePool(); !ptr) {
     // We must construct in place here, since there is a slice that uses memory locations
-    ptr = make_unique<PipelineMessage>(args.size(), backed_sz);
+    ptr = make_unique<PipelineMessage>();
   }
 
-  ptr->SetArgs(args);
+  auto map = [](const RespExpr& expr) { return expr.GetView(); };
+  auto range = base::it::Transform(map, base::it::Range(args.begin(), args.end()));
+  ptr->Assign(range.begin(), range.end(), args.size());
```
dranikpg, line 1742: Ohh I feel honoured for someone using my abstracations 👑

romange: it took me some time to learn how to use this algebra

PR #6169, opened 2025-12-06
chore: introduce ParsedArguments class

romange:
Currently we use ArgSlice, CmdArgList everywhere where we want a span on string_views. As a result, it's very hard to understand the requirements from the passed classes.

This PR introduces a new ParsedArguments class and reduces the API surface for the passed object. Make sure to unwind ParsedArguments to ArgSlice only when we start executing the command.

We will clean-up redundant aliases in the futures.

### DIFF
--- a/src/server/main_service.cc
+++ b/src/server/main_service.cc
@@ -1423,7 +1423,7 @@ std::optional<ErrorReply> Service::VerifyCommandState(const CommandId* cid, CmdA
   return VerifyConnectionAclStatus(cid, &dfly_cntx, "has no ACL permissions", tail_args);
 }
 
-DispatchResult Service::DispatchCommand(ArgSlice args, SinkReplyBuilder* builder,
+DispatchResult Service::DispatchCommand(facade::ParsedArgs args, SinkReplyBuilder* builder,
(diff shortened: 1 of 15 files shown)


src/server/main_service.cc
```diff
@@ -1423,16 +1423,16 @@ std::optional<ErrorReply> Service::VerifyCommandState(const CommandId* cid, CmdA
...
   DCHECK(!args.empty());
   DCHECK_NE(0u, shard_set->size()) << "Init was not called";
 
   absl::Cleanup clear_last_error([builder]() { builder->ConsumeLastError(); });
   ServerState& etl = *ServerState::tlocal();
 
-  string cmd = absl::AsciiStrToUpper(args[0]);
-  const auto [cid, args_no_cmd] = registry_.FindExtended(cmd, args.subspan(1));
+  string cmd = absl::AsciiStrToUpper(args.Front());
+  const auto [cid, args_no_cmd] = registry_.FindExtended(cmd, args.Tail());
```
dranikpg, line 1435: once we have them packed, we can refactor registry interface to pick front() one by one

PR #6170, opened 2025-12-06
chore: introduce common directory

romange:
We introduce a new directory that will contain low-level common utilities and classes that can be used by any other components: facade, core and server. We move StringOrView and HeapSize there.

### DIFF
--- a/src/common/heap_size.h
+++ b/src/common/heap_size.h
@@ -25,7 +25,7 @@
 #include <type_traits>
 #include <vector>
 
-namespace dfly {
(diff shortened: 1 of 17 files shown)


romange:
@abhijat  I think you were looking where to put common utilities in dragonfly space. this is the place

abhijat:
> @abhijat I think you were looking where to put common utilities in dragonfly space. this is the place

Is this going to be a cmake library in future? like a root level lib that can be linked against without pulling in many dependencies or creating circular references.

PR #6171, opened 2025-12-07
chore: add ability for ParsedArgs to reference BackedArguments

romange:
Reimplement StoredCmd using ParsedArgs and BackedArguments.
This reduces code duplication as well as paves the path to use BackedArguments during pipeline squashing.

Needed to adjust test_lua_auto_async test as we had accounting bug in StoredCmd::UsedMemory function

### DIFF
--- a/.github/actions/regression-tests/action.yml
+++ b/.github/actions/regression-tests/action.yml
@@ -31,7 +31,6 @@ inputs:
   epoll:
     required: false
     type: string
-
 runs:
   using: "composite"
   # bring back timeouts once composite actions start supporting them
@@ -41,7 +40,14 @@ runs:
       shell: bash
       run: |
         cd ${GITHUB_WORKSPACE}/tests/dragonfly/valkey_search
-        ./sync-valkey-search-tests.sh 90124dc91756b24cb2e58e5c4eea5b8d53004ea6 # main branch revision
+        # main branch revision
+        ./sync-valkey-search-tests.sh 90124dc91756b24cb2e58e5c4eea5b8d53004ea6
+    - name: Prepare Step
+      shell: bash
+      run: |
+        DOCKER_VERSION="26.1.4"
+        curl -fsSL "https://download.docker.com/linux/static/stable/x86_64/docker-${DOCKER_VERSION}.tgz" | tar xz --strip 1 docker/docker -C ./docker
+        registry="public.ecr.aws/docker/library" && for image in fedora:latest ubuntu:noble; do ./docker pull "$registry/$image"; done
     - name: Run PyTests
       id: main
       shell: bash
--- a/src/common/backed_args.h
+++ b/src/common/backed_args.h
@@ -12,11 +12,8 @@ namespace cmn {
 
 class BackedArguments {
   constexpr static size_t kLenCap = 5;
-#ifdef ABSL_HAVE_ADDRESS_SANITIZER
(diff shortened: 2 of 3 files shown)


src/facade/facade_types.h
```diff
@@ -45,31 +46,88 @@ class ParsedArgs {
  public:
   ParsedArgs() = default;
 
-  explicit ParsedArgs(ArgSlice args) : args_(args) {  // NOLINT google-explicit-constructor
+  // References backed arguments. The object must outlive this ParsedArgs.
+  ParsedArgs(const cmn::BackedArguments& bargs)  // NOLINT google-explicit-constructor
```
dranikpg, line 50: Please add a comment somewhere in your next PR, because one can get lost now with all those types

dranikpg: that this is a variant of <BackedArgs, ArgList> and we want to call DispathcCOmmand with both types

src/server/conn_context.h
```diff
@@ -28,43 +27,47 @@ struct FlowInfo;
...
-    }
-  };
+  const CommandId* cid_;     // underlying command
+  facade::ParsedArgs args_;  // arguments
 
-  std::variant<OwnStorage, CmdArgList> args_;  // args storage
-  facade::ReplyMode reply_mode_;               // reply mode
+  // TODO: we could optimize the storage further by introducing StoredCmdCollection and
+  // keep the backing storage there. Then this class will only use shallow copies.
+  std::unique_ptr<cmn::BackedArguments> backed_;
```
dranikpg, line 63: it seems like you can just steal the BackedArguments form the queue without much hesitation

src/server/conn_context.cc
```diff
@@ -34,88 +34,21 @@ static void SendSubscriptionChangedResponse(string_view action, std::optional<st
...
-
-  for (unsigned i = 0; i < args.size(); i++) {
-    // Assume tightly packed list.
-    DCHECK(i + 1 == args.size() || args[i].data() + args[i].size() == args[i + 1].data());
-    own_storage.sizes[i] = args[i].size();
-  }
+StoredCmd::StoredCmd(const CommandId* cid, facade::ArgSlice args, facade::ReplyMode mode)
+    : cid_{cid}, args_{args}, reply_mode_{mode} {
+  backed_ = std::make_unique<cmn::BackedArguments>(args.begin(), args.end());
+  args_ = facade::ParsedArgs{*backed_};
```
dranikpg, line 40: Elegant we re-used the type here 👍🏻

dranikpg (dismissed):

src/facade/facade_types.h
romange: actually we do not want, it's something I want to get rid of and it's a temporary state which will be fixed in the process.
I am planning to move to the memory model where arguments are always allocated on heap so we could dispatch asynchronously.

src/server/conn_context.h
romange: what do you mean? what queue?

src/facade/facade_types.h
dranikpg: ok, then you have to expose the core type of heap allocated arguments from stored cmd to run multi exc

src/server/conn_context.h
dranikpg: dispatch queue what in the future will be just ao object passed around

PR #6172, opened 2025-12-07
chore: rename WRITE to JOURNALED

romange:
We marked commands as WRITE, but in fact we marked only "journable" commands as write, EXEC, PUBLISH, EVAL(SHA) where not marked as such even though they mutate the db state.

this pr renames the attribute to better express its meaning.

### DIFF
--- a/src/server/command_registry.h
+++ b/src/server/command_registry.h
@@ -28,10 +28,10 @@ namespace CO {
 
 enum CommandOpt : uint32_t {
   READONLY = 1U << 0,
-  FAST = 1U << 1,  // Unused?
-  WRITE = 1U << 2,
-  LOADING = 1U << 3,  // Command allowed during LOADING state.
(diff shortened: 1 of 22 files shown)


src/server/command_registry.h
```diff
@@ -27,10 +27,11 @@ namespace CO {
 
 enum CommandOpt : uint32_t {
   READONLY = 1U << 0,
-  FAST = 1U << 1,  // Unused?
-  WRITE = 1U << 2,
-  LOADING = 1U << 3,  // Command allowed during LOADING state.
-  DENYOOM = 1U << 4,  // use-memory in redis.
+  FAST = 1U << 1,       // Unused?
+  WRITE = 1U << 2,      // Deprecated in favor of JOURNALED.
+  JOURNALED = 1U << 2,  // Command is logged to AOF / Journal.
```
dranikpg, line 32: You can just replace WRITE everywhere in this PR

PR #6175, opened 2025-12-08
tests: Skip package installer tests on CI

abhijat:
Package installer tests routinely fail due to throttling on docker registry.

These tests also do not make sense when run as part of CI suite, they should only be run when new releases are created. Here these tests are skipped in CI. In the next PR they will be wired to run only as part of the action which sets up the package repository.

### DIFF
--- a/tests/dragonfly/package_install_test.py
+++ b/tests/dragonfly/package_install_test.py
@@ -3,6 +3,11 @@ import platform
 import pytest
 from testcontainers.core.container import DockerContainer
 
+pytest.skip(
+    "These tests will only be run for the action that generates the package repository",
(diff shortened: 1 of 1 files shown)


abhijat:
The module is skipped now on running pytest

```
SKIPPED [1] tests/dragonfly/package_install_test.py:6: These tests will only be run for the action that generates the package repository
```

romange:
@abhijat  lets wait with this. I tried to fix it here:
https://github.com/dragonflydb/dragonfly/pull/6171/files#diff-84a1dffad3636502d8db6b1e152f90fe0ab18249563aaba59862894f3cf8cb12

let's see if it helps

PR #6176, opened 2025-12-08
chore: use BackedArguments for memcache_parser

romange:
Following https://github.com/dragonflydb/dragonfly/pull/6171 and https://github.com/dragonflydb/dragonfly/pull/6169 we implement Memcache::Command using BackedArguments.

### DIFF
--- a/src/common/backed_args.h
+++ b/src/common/backed_args.h
@@ -20,6 +20,12 @@ class BackedArguments {
 
   class iterator {
    public:
+    using iterator_category = std::forward_iterator_tag;
+    using value_type = std::string_view;
+    using difference_type = std::ptrdiff_t;
+    using pointer = const std::string_view*;
+    using reference = std::string_view;
+
     iterator(const BackedArguments* ba, size_t index) : ba_(ba), index_(index) {
     }
 
@@ -30,7 +36,7 @@ class BackedArguments {
     }
 
     bool operator==(const iterator& other) const {
-      return index_ == other.index_;
+      return index_ == other.index_ && ba_ == other.ba_;
     }
 
     bool operator!=(const iterator& other) const {
--- a/src/facade/memcache_parser.cc
+++ b/src/facade/memcache_parser.cc
@@ -61,25 +61,22 @@ MP::CmdType From(string_view token) {
   return MP::INVALID;
 }
 
-void MaterializeArgs(absl::Span<std::string_view> args, MP::Command* dest) {
-  dest->Assign(args.begin(), args.end(), args.size());
-}
-
 MP::Result ParseStore(ArgSlice tokens, MP::Command* res) {
   const size_t num_tokens = tokens.size();
-  unsigned opt_pos = 3;
+  unsigned opt_pos = 4;
   if (res->type == MP::CAS) {
     if (num_tokens <= opt_pos)
       return MP::PARSE_ERROR;
     ++opt_pos;
   }
 
+  // tokens[0] is key
   uint32_t flags;
-  if (!absl::SimpleAtoi(tokens[0], &flags) || !absl::SimpleAtoi(tokens[1], &res->expire_ts) ||
-      !absl::SimpleAtoi(tokens[2], &res->bytes_len))
+  if (!absl::SimpleAtoi(tokens[1], &flags) || !absl::SimpleAtoi(tokens[2], &res->expire_ts) ||
+      !absl::SimpleAtoi(tokens[3], &res->bytes_len))
     return MP::BAD_INT;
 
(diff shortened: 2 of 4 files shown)


src/facade/memcache_parser_test.cc
```diff
@@ -18,82 +18,87 @@ namespace facade {
 
 class MCParserTest : public testing::Test {
  protected:
+  MemcacheParser::Result Parse(string_view input) {
```
romange, line 21: Despite many changes, the tests here are left intact. most changes are around Parse called being replaced with this auxillary function.

src/common/backed_args.h
```diff
@@ -18,6 +18,35 @@ class BackedArguments {
...
+    }
+
+    iterator& operator++() {
+      offset_ += ba_->elem_capacity(index_);
+      ++index_;
+      return *this;
+    }
+
+    bool operator==(const iterator& other) const {
+      return index_ == other.index_;
```
BorysTheDev: add dcheck(ba_ == other.ba_)

src/common/backed_args.h
```diff
@@ -18,6 +18,35 @@ class BackedArguments {
...
+    }
+
+    std::string_view operator*() const {
+      return ba_->at(index_, offset_);
+    }
+
+   private:
+    const BackedArguments* ba_;
+    size_t index_;
+    size_t offset_ = 0;
```
BorysTheDev, line 53: As I understand, offset we can get from index, so I don't see any point in storing it here

src/facade/memcache_parser.cc
```diff
@@ -321,20 +331,27 @@ auto MP::Parse(string_view str, uint32_t* consumed, Command* cmd) -> Result {
...
+    tmp_args_.push_back(tokens_view[0]);
     tokens_view.remove_prefix(1);
-    return ParseStore(tokens_view, cmd);
+    auto res = ParseStore(tokens_view, cmd);
+    if (res != MP::OK) {
+      return res;
+    }
+    MaterializeArgs(absl::MakeSpan(tmp_args_), cmd);
+
+    return MP::OK;
```
BorysTheDev: nit
```
    auto res = ParseStore(tokens_view, cmd);
    if (res == MP::OK) {
       MaterializeArgs(absl::MakeSpan(tmp_args_), cmd);
    }
    return res;
```

src/common/backed_args.h
romange: you need O(N) to compute offset from lens
`offset = Sum{len_[k] |  k <= index }`

romange:
augment review

src/common/backed_args.h
BorysTheDev: just for curiosity if instead of len we will store offset directly and calculate len like next - prev. It should solve the complexity

src/facade/memcache_parser.cc
```diff
@@ -126,16 +132,17 @@ MP::Result ParseValueless(ArgSlice tokens, MP::Command* res) {
   }
 
   while (key_pos < num_tokens) {
-    res->keys_ext.push_back(tokens[key_pos++]);
+    args->push_back(tokens[key_pos++]);
   }
 
   if (res->type >= MP::DELETE) {  // write commands
-    if (!res->keys_ext.empty() && res->keys_ext.back() == "noreply") {
+    if (args->size() > 1 && args->back() == "noreply") {
```
BorysTheDev, line 139: should this comparison be case insensitive? args->back() == "noreply"

src/common/backed_args.h
romange: interesting idea, let me think about it.

src/facade/memcache_parser.cc
romange: 1. it's not a regression 
2. based on https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L209 it should be lowcase.

src/facade/memcache_parser.cc
```diff
@@ -321,15 +331,13 @@ auto MP::Parse(string_view str, uint32_t* consumed, Command* cmd) -> Result {
 
   ArgSlice tokens_view{tokens};
   tokens_view.remove_prefix(1);
+  cmd->clear();
 
   if (cmd->type <= CAS) {                                    // Store command
     if (tokens_view.size() < 4 || tokens[0].size() > 250) {  // key length limit
```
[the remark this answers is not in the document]
romange: will double check in the next PR.

src/common/backed_args.h
BorysTheDev: for 3 elements with len  5 4 3 you will have the next offset array [5, 9, 12] to get len `return id > 0 ? offset[id] - offset[id-1] : offset[0]; `

PR #6178, opened 2025-12-08
chore(rax): add move ctor to rax tree, cleanup defragmap 

abhijat:
Adds move ctors to rax tree map. To avoid copying internal state of the rax iterato piece by piece, a new seek is done from the source iterator's key.

Using the move iterator, the `DefragmentMap` implementation is cleaned up as discussed in previous PR touching this code: https://github.com/dragonflydb/dragonfly/pull/6144

### DIFF
--- a/src/core/search/rax_tree.h
+++ b/src/core/search/rax_tree.h
@@ -66,11 +66,8 @@ template <typename V> struct RaxTreeMap {
     explicit SeekIterator(rax* tree) : SeekIterator(tree, "^", std::string_view{nullptr, 0}) {
     }
 
-    SeekIterator(SeekIterator&& other) noexcept {
-      if (::detail::CopyIteratorState(it_, other.it_))
-        operator++();
-      if (other.IsValid())
-        other.InvalidateIterator();
+    SeekIterator(SeekIterator&& other) noexcept : it_{} {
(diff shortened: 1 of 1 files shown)


abhijat:
there is a crash in `raxNext` when defragmenting a rax tree, probably related to the new move ctor added for the iterator.

eta: should be fixed now by pretending to move using seek ~~, but needs more tests for safety~~ tests added

src/core/search/rax_tree.h
```diff
@@ -40,10 +66,28 @@ template <typename V> struct RaxTreeMap {
     explicit SeekIterator(rax* tree) : SeekIterator(tree, "^", std::string_view{nullptr, 0}) {
     }
 
-    /* Remove copy/move constructors to avoid double iterator invalidation */
-    SeekIterator(SeekIterator&&) = delete;
+    SeekIterator(SeekIterator&& other) noexcept {
+      if (::detail::CopyIteratorState(it_, other.it_))
+        operator++();
+      if (other.IsValid())
+        other.InvalidateIterator();
```
dranikpg: this = move(other)

abhijat: nice, this is much shorter. changed.

PR #6179, opened 2025-12-08
Add coordinator connection reusing

BorysTheDev:
related to #6009

Added coordinator connection reusing

### DIFF
--- a/tests/dragonfly/cluster_test.py
+++ b/tests/dragonfly/cluster_test.py
@@ -3489,7 +3489,7 @@ async def test_SearchRequestDistribution(df_factory: DflyInstanceFactory):
 
     cclient = instances[0].cluster_client()
 
-    docs_num = 3
+    docs_num = 2
     for i in range(0, docs_num):
         assert await cclient.execute_command("HSET", f"s{i}", "title", f"test {i}") == 1
 


src/server/take_resp_expr.h
```diff
@@ -0,0 +1,58 @@
...
+#pragma once
+
+#include <variant>
+
+#include "facade/resp_expr.h"
+
+namespace dfly {
+
+class TakeRespExpr {
+ public:
```
dranikpg: 1. Take is a verb, its better to be Take**n** or Owned
2. Is it temporary? Why a new class? Why in /server?

RespExpr isn't much used besides the connection code (and tests), maybe you can modify the original one to have an owned buffer?

src/server/protocol_client.cc
```diff
@@ -314,6 +314,67 @@ io::Result<ProtocolClient::ReadRespRes> ProtocolClient::ReadRespReply(uint32_t t
...
 
+io::Result<TakeRespExpr::Vec> ProtocolClient::TakeRespReply(uint32_t timeout) {
+  auto prev_timeout = sock_->timeout();
+  sock_->set_timeout(timeout);
+  auto res = TakeRespReply(nullptr, false);
+  sock_->set_timeout(prev_timeout);
+  return res;
+}
+
+io::Result<TakeRespExpr::Vec> ProtocolClient::TakeRespReply(base::IoBuf* buffer, bool copy_msg) {
```
dranikpg: its the same as ReadRespExpr except some small details. If this code is not temporary, please keep it as one function. Futhermore, if you extend RespExpr to be owning, you can pass it just as a parameter without any templates

src/server/take_resp_expr.h
BorysTheDev: I've added it here because of protocol client

BorysTheDev: for further developement we need better command parsing approach and I think it will be more code, maybe I will create a separate directory and move this code there.

src/server/owned_resp_expr.h
```diff
@@ -0,0 +1,58 @@
...
+class OwnedRespExpr {
+ public:
+  using Buffer = std::vector<uint8_t>;
+
+  using Type = facade::RespExpr::Type;
+
+  using Vec = std::vector<OwnedRespExpr>;
+  Type type;
+
+  std::variant<int64_t, double, Buffer, Vec> u;
```
dranikpg, line 21: I still don't understand why you can't extedn RespExpr with an OwnedBuffer type and then adapt `TakeRespReply` to conditionally use either of those

dranikpg (approved):
I'm not going to force my opinion on you any further but it seems like the separate type is redundant

src/server/owned_resp_expr.h
BorysTheDev: will do in one of the next PRs

PR #6180, opened 2025-12-08
CI: download docker for correct architecture

abhijat:
Download the correct docker release for the runner.

### DIFF
--- a/tests/dragonfly/package_install_test.py
+++ b/tests/dragonfly/package_install_test.py
@@ -14,8 +14,10 @@ async def test_install_package_on_fedora():
     with DockerContainer(image="ghcr.io/romange/fedora:30", tty=True) as fedora:
         check(
             fedora,
-            "dnf config-manager addrepo --from-repofile=https://packages.dragonflydb.io/dragonfly.repo",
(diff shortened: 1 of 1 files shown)


abhijat:
download seems to work as expected in https://github.com/dragonflydb/dragonfly/actions/runs/20032246854

romange:
Btw, you can actually use `https://github.com/romange/container-foundry/pkgs/container/fedora/590755041?tag=30` to test fedora install as we want to support fedora 30 or newer.

abhijat:
I updated the registry and the fedora version. Test run ~~https://github.com/dragonflydb/dragonfly/actions/runs/20050213806~~

https://github.com/dragonflydb/dragonfly/actions/runs/20051590189

abhijat:
the image pull part succeeds now in test run

romange (commented):
you updated the registry here but not in the test itself.

abhijat:
```
dnf config-manager: error: unrecognized arguments: --from-repofile
```

The test doesn't work with `fedora:30`. It looks like the dnf command has changed since that release.

changed to download file manually

abhijat:
looks ok now, the tests are passing

abhijat:
augment review

Issue #6184, opened 2025-12-08
HINCRBY removes hash field TTL


daraber:
**Describe the bug**
The `HINCRBY` command does not preserve the TTL property of a hash field. When an increment operation is performed on a field with an existing TTL, the TTL is removed.

**To Reproduce**
See code snippet below.

**Expected behavior**
The TTL of the hash field should remain unchanged after the `HINCRBY` operation.

**Environment**
 - OS: `6.6.87.2-microsoft-standard-WSL2`
 - Kernel: `Linux DESKTOP-3MCQMPO 6.6.87.2-microsoft-standard-WSL2 #1 SMP PREEMPT_DYNAMIC x86_64 x86_64 x86_64 GNU/Linux`
 - Container: Docker
 - Dragonfly Version: `1.27.1` & `1.35.1` tested

**Reproducible Code Snippet**
```
HSETEX myhash 100 field1 0
> 1
FIELDTTL myhash field1
> 99
HINCRBY myhash field1 1
> 1
FIELDTTL myhash field1
> -1
```

PR #6186, opened 2025-12-09
hset_family: Preserve ttl on increment command

abhijat:
There are two issues which are handled here. First the `StringMap::AddOrUpdate`, which is not directly related to the original bug. 

If it is passed the combination of arguments:

`AddOrUpdate("field", "value", UINT32_MAX, true)` ie the ttl is max value and keepttl is true, then it runs into a bug:

* new entry is created with no space for ttl (because of the max value)
* if previous ttl is found, it is written into the new entry as if there is space. the ttl bit is also not set on the new entry.

There is no apparent corruption of the value, but the ttl is not registered even if copied into the memory because of the bit not being set.

The second issue is that `keepttl` is not passed to `StringMap::AddOrUpdate` so the existing value is not copied over.

In `StringMap::AddOrUpdate` an additional lookup is performed before creating the new entry, then the ttl from lookup is used to create the new entry. The `keepttl` flag is passed during `OpIncrBy`. This additional lookup is expensive, but only performed when `keepttl` is true.

fixes https://github.com/dragonflydb/dragonfly/issues/6184


kostasrim:
I was about to sent a PR 😄 

p.s. it seems we are also incosistent with redis/valkey on how we parse the command. See: https://github.com/dragonflydb/dragonfly/issues/6187

abhijat:
> I was about to sent a PR 😄
> 
> p.s. it seems we are also incosistent with redis/valkey on how we parse the command. See: #6187

I realized there is actually an inconsistency in this PR, it still works because of the way `StringMap::AddOrUpdate` works but it is wrong:

* get expiry time from string map for field - this is absolute time
* pass it as ttl to StringMap::AddOrUpdate - wrong, we need ttl seconds not absolute
* StringMap::AddOrUpdate then works things out correctly because it just copies ttl from previous entry, overwriting the absolute time incorrectly passed in

The expiry time should be converted to ttl.

abhijat:
Simply using expiry time - `StringMap::Now()` to calculate TTL might also introduce bugs, eg if the expiry time has passed by the time this calculation is done. In that case what should `HINCRBY` do? It i supposed to add the field if missing. Should it then add the field with no expiry?

abhijat:
> In that case what should `HINCRBY` do? It is supposed to add the field if missing. Should it then add the field with no expiry?

After reviewing the existing behavior, the right thing to do IMO is to add with the remaining TTL left however small

* if the field expired while searching in string map, we will not find the existing TTL, and the TTL defaults to no-expiry (max value). this is consistent with the contract that hincrby creates key if not existing
* if the field expired after finding existing value in string map, thus copying its TTL, and before creating the new value in the map (highly unlikely but can happen) then the TTL we computed when we found the existing value will be very small. The newly incremented key will quickly be expired

src/server/hset_family.cc
```diff
@@ -121,8 +137,8 @@ struct HMapWrap {
     return VisitMut(ov);
   }
 
-  void AddOrUpdate(std::string_view key, std::string_view value) {
-    Overloaded ov{[&](StringMap* sm) { sm->AddOrUpdate(key, value); },
+  void AddOrUpdate(std::string_view key, std::string_view value, uint32_t ttl_sec) {
+    Overloaded ov{[&](StringMap* sm) { sm->AddOrUpdate(key, value, ttl_sec, true); },
```
abhijat: `ttl_sec` is ultimately ignored in `AddOrUpdate` because `keepttl` is true, but leaving this to the default UINT32_MAX results in the new field being incorrectly initialized, without the ttl bit set, resulting in the TTL being lost instead of copied from old entry. 

In fact sending anything other than UINT32_MAX will trigger the correct code path here.

abhijat: fixed in `StringMap::AddOrUpdate`

src/core/string_map.cc
```diff
@@ -62,20 +62,35 @@ StringMap::~StringMap() {
...
-    }
+    if (!prev_has_ttl)
+      return ttl_sec;
+
+    const uint32_t prev_expire = ObjExpireTime(prev_entry);
+    const uint32_t now = time_now();
+    // FindInternal will have removed an expired entry, so prev_expire must be in the future. The
+    // ttl_sec is added here defensively for a corner case where the expiry happens between
+    // FindInternal and this comparison, to behave the same way as if an entry was not found or had
+    // no TTL.
```
abhijat: just realized `time_now()` returns a time which was set much earlier and will not change, so this is not really required, as expiry will use the same time as whatever `time_now()` returns here.

src/core/string_map.cc
```diff
@@ -62,20 +62,34 @@ StringMap::~StringMap() {
...
+
+    const uint32_t prev_expire = ObjExpireTime(prev_entry);
+    const uint32_t now = time_now();
+    // FindInternal will have removed an expired entry, so prev_expire must be in the future. The 0
+    // is added here defensively for a corner case where the expiry happens between FindInternal
+    // and this comparison, to behave the same way as if an entry was not found or had no TTL.
+    return prev_expire > now ? (prev_expire - now) : ttl_sec;
+  };
+
+  auto [newkey, sdsval_tag] = CreateEntry(field, value, time_now(), compute_ttl());
```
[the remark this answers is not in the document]
abhijat: fixed

src/core/string_map.cc
```diff
@@ -62,23 +67,30 @@ StringMap::~StringMap() {
...
-  auto [newkey, sdsval_tag] = CreateEntry(field, value, time_now(), ttl_sec);
-
-  // Replace the whole entry.
-  if (sds prev_entry = (sds)AddOrReplaceObj(newkey, sdsval_tag & kValTtlBit); prev_entry) {
-    const bool prev_has_ttl =
-        absl::little_endian::Load64(prev_entry + sdslen(prev_entry) + 1) & kValTtlBit;
-    if (keepttl && prev_has_ttl) {
-      SdsUpdateExpireTime(newkey, ObjExpireTime(prev_entry), 8);
-    }
+  const uint32_t computed_ttl = ComputeTtl(field, ttl_sec, keepttl);
```
romange, line 70: please add a test in string_map_test as well as you fixed the bug here.

abhijat: added a test with a few checks related to these changes

src/core/string_map_test.cc
```diff
@@ -245,4 +245,33 @@ TEST_F(StringMapTest, ExpiryChangesSize) {
...
+  sm_->AddOrUpdate("field", "value", 100);
+  auto k = sm_->Find("field");
+  EXPECT_TRUE(k.HasExpiry());
+  EXPECT_EQ(k.ExpiryTime(), 100);
+
+  // ttl is copied from prev. if max value is supplied
+  sm_->AddOrUpdate("field", "value", UINT32_MAX, true);
+  k = sm_->Find("field");
+  EXPECT_TRUE(k.HasExpiry());
+  EXPECT_EQ(k.ExpiryTime(), 100);
```
abhijat, line 258: this failed without the fix, the new object had no expiry

PR #6188, opened 2025-12-09
fix(search): Fix numeric tree finalization

dranikpg:
During finalization the tree doesn't put entries with the same value inside a single block which breaks the logical invariant

### DIFF
--- a/src/core/search/range_tree.cc
+++ b/src/core/search/range_tree.cc
@@ -181,33 +181,29 @@ void RangeTree::FinalizeInitialization() {
   DCHECK_EQ(entries_.size(), 1u);
 
   auto& block = entries_.begin()->second;
-  const size_t total_size = block.Size();
-
-  std::vector<Entry> entries(total_size);
-  size_t index = 0;
-  for (const auto& [id, value] : block) {
-    entries[index++] = {id, value};
-  }
+  std::vector<Entry> entries(block.Size());
+  std::copy(block.begin(), block.end(), entries.begin());  // avoid std::distance call
 
   enable_splitting_ = true;
   block.Clear();
+  block.max_seen = -std::numeric_limits<double>::infinity();  // reset right bound
 
   std::sort(entries.begin(), entries.end(),
             [](const auto& a, const auto& b) { return a.second < b.second; });
 
-  for (size_t b = 0; b < entries.size(); b += max_range_block_size_) {
-    RangeBlock* range_block;
-    if (b) {
-      range_block = &CreateEmptyBlock(entries[b].second)->second;
-    } else {
-      range_block = &block;
-    }
+  // Add sorted elements in batches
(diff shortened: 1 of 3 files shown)


dranikpg:
@romange, you've been the only one to review the last numeric tree PR, so I assume this goes out to you as well 😅

src/core/search/range_tree.cc
```diff
@@ -181,33 +181,29 @@ void RangeTree::FinalizeInitialization() {
...
   auto& block = entries_.begin()->second;
-  const size_t total_size = block.Size();
-
-  std::vector<Entry> entries(total_size);
-  size_t index = 0;
-  for (const auto& [id, value] : block) {
-    entries[index++] = {id, value};
-  }
+  std::vector<Entry> entries(block.Size());
+  std::copy(block.begin(), block.end(), entries.begin());  // avoid std::distance call
```
romange, line 185: just curious what `avoid std::distance call` means in this context?

src/core/search/range_tree.h
```diff
@@ -78,6 +78,7 @@ class RangeTree {
   // Returns all blocks in the tree.
   absl::InlinedVector<const RangeBlock*, 5> GetAllBlocks() const;
 
+  // Build tree ouf of a single block after replication
```
romange, line 81: ouf

src/core/search/range_tree_test.cc
```diff
@@ -491,10 +491,21 @@ TEST_F(RangeTreeTest, FinalizeInitialization) {
   tree.FinalizeInitialization();
 
   result = tree.RangeBlocks(10.0, 40.0);
-  EXPECT_THAT(
-      result,
-      BlocksAre(
-          {{{1, 10.0}}, {{2, 20.0}, {3, 20.0}, {5, 20.0}}, {{4, 30.0}, {6, 30.0}}, {{7, 40.0}}}));
+  EXPECT_THAT(result, BlocksAre({{{1, 10.0}, {2, 20.0}, {3, 20.0}, {5, 20.0}},
```
romange, line 494: the differrence is that the first block now contains 1, 2,3,5 ?

src/core/search/range_tree.cc
dranikpg: The most idiomatic and cleanest way is to initialize `entries {block.begin(), block.end()}`, however it resizes with `n = distance(begin, end)` which is a linear operation, so it's easier to resize first and then assign

src/core/search/range_tree_test.cc
dranikpg: It was a tricky test, because as soon as you change the 1->2 (see few lines above), the order breaks with the current code. 1 is kind of a special value. 

The division as before would've been actually better, but its hard to achieve with a simple algorithm.

Issue #6182, opened 2025-12-08
crash: "Unexpected override" in Renamer::DeserializeDest with BRPOPLPUSH + RENAME + LPUSH


vyavdoshenko:
Python script to reproduce:

```python
#!/usr/bin/env python3
import asyncio, random
from redis import asyncio as aioredis

PORT = 6379
NUM_KEYS = 10
ITERATIONS = 2000
WORKERS = 5

async def main():
    r = aioredis.Redis(port=PORT)
    for i in range(NUM_KEYS):
        await r.delete(f'k{i}')
        await r.rpush(f'k{i}', 'a', 'b', 'c')
    await r.aclose()

    async def worker(ops):
        c = aioredis.Redis(port=PORT)
        for _ in range(ITERATIONS):
            try:
                op = random.choice(ops)
                k1 = f'k{random.randint(0, NUM_KEYS-1)}'
                k2 = f'k{random.randint(0, NUM_KEYS-1)}'
                if op == 'brpoplpush':
                    await c.brpoplpush(k1, k2, timeout=0.001)
                elif op == 'lpush':
                    await c.lpush(k1, 'x')
                elif op == 'rename':
                    await c.rename(k1, k2)
            except:
                pass
        await c.aclose()

    tasks = [worker(['brpoplpush']) for _ in range(WORKERS)]
    tasks += [worker(['lpush']) for _ in range(WORKERS)]
    tasks += [worker(['rename']) for _ in range(WORKERS)]
    await asyncio.gather(*tasks)

    try:
        c = aioredis.Redis(port=PORT)
        await c.ping()
        print("Server alive")
        await c.aclose()
    except:
        print("CRASHED")

if __name__ == "__main__":
    asyncio.run(main())
```

Crash log:
```
F20251208 18:45:39.154712 32753 generic_family.cc:504] Unexpected override for key k5 0
*** Check failure stack trace: ***
    @     0x5b2b71bc04ef  google::LogMessage::Fail()
    @     0x5b2b71bc0435  google::LogMessage::SendToLog()
    @     0x5b2b71bbfbe8  google::LogMessage::Flush()
    @     0x5b2b71bbf963  google::LogMessage::~LogMessage()
    @     0x5b2b70e53938  dfly::(anonymous namespace)::Renamer::DeserializeDest()
[...]

PR #6190, opened 2025-12-09
[WIP] fix: RENAME crash with concurrent blocking commands

vyavdoshenko:
Fix crash in `Renamer::DeserializeDest` when RENAME runs concurrently with blocking commands like BRPOPLPUSH.

The bug: RENAME caches destination key existence in phase 1 (`dest_found_`),  but between phases, another transaction can create/delete the key. Using stale cached value in phase 2 causes a crash when trying to delete a non-existent key.

The fix: Check current key state via `dest_res.it.is_done()` instead of relying on stale `dest_found_` from phase 1.

Fixes: https://github.com/dragonflydb/dragonfly/issues/6182

### DIFF
--- a/src/server/generic_family.cc
+++ b/src/server/generic_family.cc
@@ -477,7 +477,10 @@ OpStatus Renamer::DeserializeDest(Transaction* t, EngineShard* shard) {
   auto& db_slice = t->GetDbSlice(shard->shard_id());
   auto dest_res = db_slice.FindMutable(op_args.db_cntx, dest_key_);
 
-  if (dest_found_) {
+  // Use the current state of the key, not the stale dest_found_ from phase 1.
+  // Another transaction might have created/deleted the key between phases.
+  const bool dest_exists = !dest_res.it.is_done();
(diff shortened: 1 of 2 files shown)


romange:
@vyavdoshenko  it should not happen. Another transaction should not be able to delete the key that is being changed by rename as we lock these keys for mutations. I suggest that you sync with @dranikpg  to discuss this.

dranikpg:
Its something we know, BRPOPLPUSH is the only command that has two hops and a write phase after unblocking and it doesn't respect transactions

romange:
Can you provide (simulate) the possible order of operations that causes such bug?

dranikpg:
Any order, once BRPOPLPUSH extracted a value, it will do a second hop and push it. While doing so, it doesn't check the continuation transacton, it just executes in PollExecution. You can for example start a script reading from `DEST` and then unnblock a `BRPOPLPUSH SCR DEST` and the script will see it, even though it was supposed to be atomic

PR #6194, opened 2025-12-10
chore(server): Don't decode size for empty blob

mkaruza:


dranikpg (approved):
I'll follow up on this, something else might be off 🤔

PR #6195, opened 2025-12-11
fix: macOS test

vyavdoshenko:
Fix broken build: https://github.com/dragonflydb/dragonfly/actions/runs/20123806667/job/57749266839

### DIFF
--- a/src/core/page_usage_stats_test.cc
+++ b/src/core/page_usage_stats_test.cc
@@ -376,6 +376,7 @@ TEST_F(PageUsageStatsTest, TagIndexDefragResumeWithChanges) {
 
   PageUsage p_small_quota{CollectPageStats::NO, 0.1, 10};
   p_small_quota.SetForceReallocate(true);
+  util::ThisFiber::SleepFor(10us);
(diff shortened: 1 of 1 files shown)


abhijat (approved):
thanks, missed adding a sleep here

src/core/page_usage_stats_test.cc
```diff
@@ -376,6 +376,7 @@ TEST_F(PageUsageStatsTest, TagIndexDefragResumeWithChanges) {
 
   PageUsage p_small_quota{CollectPageStats::NO, 0.1, 10};
   p_small_quota.SetForceReallocate(true);
+  util::ThisFiber::SleepFor(10us);
```
romange, line 379: maybe 10ms? :)
10us seems too small

vyavdoshenko: No, that's enough

vyavdoshenko: Fixed build:
https://github.com/dragonflydb/dragonfly/actions/runs/20127382098

Issue #6196, opened 2025-12-11
clean up Service::InvokeCmd


romange:
The goal is to be able to allow asynchronous invocations of a command. Right now everything is a mess and it's very hard to analyse. I will perform refactoring and clean ups with the logic I understand, but I need help with the logic I am less familiar with.

1. VerifyCommandExecution should not be there. for multi/exec txs, move all verification checks to multi and similarly to how we fail the command if it has wrong arguments already during multi, we should fail the command if it fails VerifyCommandExecution. The exec should fail with EXECABORT if any of the commands inside multi fail the verification.
2. Similarly with OOM checks - not clear to me why it is in Service::InvokeCmd and not inside DispatchCommand together with other checks - and inside multi to prevent the whole tx to execute if any of the commands should be rejected. @BorysTheDev  - who recently touched this code to reject slot migrations. Why OOM is handled here and not in DispatchCommand?
3. tracking code is triggered even if tracking is disabled - rewrite the code in such way that it will be dependent on a single `tracking` boolean - I do not want to see unrelated code called on the common path:
```
 tx->SetTrackingCallback({});`
....
 `if ((!tx && cid_name != "MULTI") || (tx && !tx->IsMulti())) {
```

romange:
Also, what's up with ` if (cmd_cntx->cid->name() == "REPLCONF" && absl::EqualsIgnoreCase(ArgS(tail_args, 0), "ACK")) {` 
inside InvokeCmd?  Why do we need it and why there?

dranikpg:
> Similarly with OOM checks - not clear to me why it is in Service::InvokeCmd and not inside DispatchCommand together with other checks - and inside multi to prevent the whole tx to execute if any of the commands should be rejected.

And what if the multi tx is too large? We currently throw OOM even inside a multi/script/etc.  All commands go through InvokeCmd, but not through DispatchCommand, so that's why its checked there

dranikpg:
> VerifyCommandExecution should not be there. for multi/exec txs, move all verification checks to multi and similarly to how we fail the command if it has wrong arguments already during multi, we should fail the command if it fails VerifyCommandExecution. The exec should fail with EXECABORT if any of the commands inside multi fail the verification.

Its high level verification of what we can tell without running the command. If you add a command after multi, it can get rejected, so we have to perform the filtering

dranikpg:
[...]

PR #6199, opened 2025-12-11
chore: to move some of the command-specific state into CommandContext

romange:
Main change is around command latency measurement, where we concentrate the logic in one central place: CommandContext::RecordLatency.

Related to #6196

This PR is first step to extend the responsibility of CommandContext so that all the context related to specific command execution would only be held there. Specifically here we delete a few fields stored in ConnectionContext that in fact were relevant only for the currently executed command. We move these fields to CommandContext.

### DIFF
--- a/src/server/command_registry.cc
+++ b/src/server/command_registry.cc
@@ -41,6 +41,7 @@ using namespace facade;
 
 using absl::AsciiStrToUpper;
 using absl::GetFlag;
+using absl::StrCat;
 using absl::StrSplit;
 
 namespace {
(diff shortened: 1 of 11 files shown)


src/server/command_registry.cc
```diff
@@ -128,16 +129,56 @@ constexpr int32_t kLatencyHistogramPrecision = 2;
 
 }  // namespace
 
+void CommandContext::RecordLatency(facade::ArgSlice tail_args) const {
+  DCHECK_GT(start_time_ns, 0u);
```
romange, line 133: factored out from InvokeCmd

src/server/main_service.cc
```diff
@@ -1694,29 +1691,7 @@ DispatchResult Service::InvokeCmd(const CommandId* cid, CmdArgList tail_args,
     cntx->conn_state.tracking_info_.IncrementSequenceNumber();
   }
 
-  // TODO: we should probably discard more commands here,
```
romange, line 1697: moved to CommandContext

src/server/main_service.cc
```diff
@@ -1292,15 +1292,15 @@ bool ShouldDenyOnOOM(const CommandId* cid, uint64_t curr_time_ns) {
   return false;
 }
 
-optional<ErrorReply> Service::VerifyCommandExecution(const ConnectionContext* cntx,
+optional<ErrorReply> Service::VerifyCommandExecution(const CommandContext& cmd_cntx,
                                                      CmdArgList tail_args) {
-  DCHECK(cntx->conn_state.cmd_start_time_ns);
-  if (ShouldDenyOnOOM(cntx->cid, cntx->conn_state.cmd_start_time_ns)) {
+  DCHECK(cmd_cntx.start_time_ns != 0);
```
abhijat: nit: please use `DCHECK_NE`

src/server/command_registry.cc
```diff
@@ -128,16 +129,56 @@ constexpr int32_t kLatencyHistogramPrecision = 2;
...
                      int8_t last_key, std::optional<uint32_t> acl_categories)
     : facade::CommandId(name, ImplicitCategories(mask), arity, first_key, last_key,
                         acl_categories.value_or(ImplicitAclCategories(mask))) {
   implicit_acl_ = !acl_categories.has_value();
-  hdr_histogram* hist = nullptr;
-  const int init_result = hdr_init(kLatencyHistogramMinValue, kLatencyHistogramMaxValue,
-                                   kLatencyHistogramPrecision, &hist);
-  CHECK_EQ(init_result, 0) << "failed to initialize histogram for command " << name;
-  latency_histogram_ = hist;
+  bool is_latency_tracked = GetFlag(FLAGS_latency_tracking);
```
abhijat, line 174: nit: `is_latency_tracked` should be static and const, to avoid reading the flag value every time. maybe this is acceptable now as we only do this once per command

romange: yes, this is why I removed it. no need to cache it now.

src/server/command_registry.h
```diff
@@ -70,14 +70,25 @@ enum class MultiControlKind : uint8_t {
...
+                 ConnectionContext* cntx)
+      : cid(_cid), tx(_tx), rb(_rb), conn_cntx(cntx) {
   }
 
+  const CommandId* cid;
   Transaction* tx;
   facade::SinkReplyBuilder* rb;
   ConnectionContext* conn_cntx;
+
+  uint64_t start_time_ns = 0;
```
dranikpg, line 86: its actually a variable that exists only in the scope of a function, its just inconvenient to pass it to VerifyOOM (if we shouldn't update that part at as well...)

src/server/command_registry.cc
```diff
@@ -128,16 +129,56 @@ constexpr int32_t kLatencyHistogramPrecision = 2;
 
 }  // namespace
 
+void CommandContext::RecordLatency(facade::ArgSlice tail_args) const {
+  DCHECK_GT(start_time_ns, 0u);
+  int64_t after = absl::GetCurrentTimeNanos();
```
dranikpg, line 134: Sometimes I forget we call `GetCurrentTimeNanos` twice per command 😢

romange: right. we should replace it with CycleClock

src/server/command_registry.h
romange: this variable can be in scope of a function because we execute commands synchronously. if we execute them asynchronously then we need to keep start_time in command context to be able to measure latency once the command returns.

PR #6203, opened 2025-12-11
fix(server): Fix EXPIRE bugs, support non-exclusive options

dranikpg:
Bugs: 
1. When the relative expire time is negative, we delete the value _while ignoring_ all flags that might not allow updating expiration
2. LT/GT were checked only for existing expiries, whereas the valkey docs say:
> A non-volatile key is treated as an infinite TTL for the purpose of GT and LT

Update: Allow NX together with LT/GT. It works as such that if LT or GT is matched, NX is ignored, otherwise it is applied... Basically the options are not exclusive, but just saitfying any of them will update the expiry

### DIFF
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -1061,7 +1061,7 @@ OpResult<int64_t> DbSlice::UpdateExpire(const Context& cntx, Iterator prime_it,
   DCHECK(params.IsDefined());
   DCHECK(IsValid(prime_it));
 
-  if (params.persist) {
+  if (params.persist) {  // Persist means remove expiry
     RemoveExpire(cntx.db_index, prime_it);
     return kPersistValue;
   }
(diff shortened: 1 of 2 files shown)


src/server/db_slice.cc
```diff
@@ -1060,7 +1060,7 @@ OpResult<int64_t> DbSlice::UpdateExpire(const Context& cntx, Iterator prime_it,
   constexpr uint64_t kPersistValue = 0;
   DCHECK(params.IsDefined());
   DCHECK(IsValid(prime_it));
-  // If this need to persist, then only set persist value and return
```
romange, line 1063: why removed?

tests/fakeredis/test/test_mixins/test_generic_commands.py
```diff
@@ -323,10 +317,11 @@ def test_expire_should_throw_error(r: redis.Redis):
...
         r.expire("foo", 1, nx=True, xx=True)
-    with pytest.raises(ResponseError):
-        r.expire("foo", 1, nx=True, gt=True)
-    with pytest.raises(ResponseError):
-        r.expire("foo", 1, nx=True, lt=True)
+    # We support NX with LT/GT
+    # with pytest.raises(ResponseError):
+    #    r.expire("foo", 1, nx=True, gt=True)
+    # with pytest.raises(ResponseError):
+    #    r.expire("foo", 1, nx=True, lt=True)
```
BorysTheDev: maybe remove?

romange (approved):

Issue #6181, opened 2025-12-08
dragonfly crashes on ft.info


vyavdoshenko:
How to reproduce (bash script):
```bash
HOST=localhost
PORT=6379
INDEX=test_idx
ITERATIONS=5000

create_drop_worker() {
    for i in $(seq 1 $ITERATIONS); do
        redis-cli -h $HOST -p $PORT FT.CREATE $INDEX ON HASH PREFIX 1 doc: SCHEMA title TEXT 2>/dev/null
        redis-cli -h $HOST -p $PORT FT.DROPINDEX $INDEX 2>/dev/null
    done
}

info_worker() {
    for i in $(seq 1 $ITERATIONS); do
        redis-cli -h $HOST -p $PORT FT.INFO $INDEX 2>/dev/null
    done
}

for i in {1..5}; do
    create_drop_worker &
    info_worker &
done

wait

redis-cli -h $HOST -p $PORT PING || echo "SERVER CRASHED!"
```

Crash log:
```
F20251208 17:12:39.918457 164343 search_family.cc:1405] Check failed: num_notfound == 0u || num_notfound == shard_set->size() 
*** Check failure stack trace: ***
    @     0x6253350844ef  google::LogMessage::Fail()
    @     0x625335084435  google::LogMessage::SendToLog()
    @     0x625335083be8  google::LogMessage::Flush()
    @     0x625335087ce0  google::LogMessageFatal::~LogMessageFatal()
    @     0x62533489273d  dfly::SearchFamily::FtInfo()
    @     0x6253342fa922  _ZN3fu27abi_4006detail10invocation6invokeIRKPFvN4absl12lts_202505124SpanIKSt17basic_string_viewIcSt11char_traitsIcEEEERKN4dfly14CommandContextEEJSC_SG_EEEDTclcl7forwardIT_Efp_Espcl7forwardIT0_Efp0_EEEOSL_DpOSM_
    @     0x6253342f9ea9  fu2::abi_400::detail::type_erasure::invocation_table::function_trait<>::internal_invoker<>::invoke()
    @     0x625334a3314f  _ZNK3fu27abi_4006detail12type_erasure6tables6vtableINS1_8propertyILb0ELb0EJKFvN4absl12lts_202505124SpanIKSt17basic_string_viewIcSt11char_traitsIcEEEERKN4dfly14CommandContextEEEEEE6invokeILm0EJPKNS2_13data_accessorERKmSE_SI_EEEDcDpOT0_
[...]

PR #6204, opened 2025-12-11
fix(search): remove invalid DCHECK in FT.INFO for concurrent index operations

vyavdoshenko:
Fixes: https://github.com/dragonflydb/dragonfly/issues/6181

Fix crash in `FT.INFO` when called concurrently with `FT.CREATE`/`FT.DROPINDEX`

### DIFF
--- a/src/server/search/search_family.cc
+++ b/src/server/search/search_family.cc
@@ -2235,8 +2235,10 @@ void SearchFamily::Register(CommandRegistry* registry) {
       << CI{"FT.ALTER", CO::JOURNALED | CO::GLOBAL_TRANS, -3, 0, 0, acl::FT_SEARCH}.HFUNC(FtAlter)
       << CI{"FT.DROPINDEX", CO::JOURNALED | CO::GLOBAL_TRANS, -2, 0, 0, acl::FT_SEARCH}.HFUNC(
              FtDropIndex)
-      << CI{"FT.INFO", CO::GLOBAL_TRANS | CO::NO_AUTOJOURNAL, -2, 0, 0, acl::FT_SEARCH}.HFUNC(
-             FtInfo)
+      << CI{"FT.INFO", CO::NO_KEY_TRANSACTIONAL | CO::NO_KEY_TX_SPAN_ALL | CO::NO_AUTOJOURNAL,
+            -2,        0,
+            0,         acl::FT_SEARCH}
+             .HFUNC(FtInfo)
       << CI{"FT.CONFIG", CO::ADMIN | CO::LOADING | CO::DANGEROUS, -3, 0, 0, acl::FT_SEARCH}.HFUNC(
              FtConfig)
       // Underscore same as in RediSearch because it's "temporary" (long time already)


src/server/search/search_family.cc
```diff
@@ -1405,9 +1405,11 @@ void SearchFamily::FtInfo(CmdArgList args, const CommandContext& cmd_cntx) {
     return OpStatus::OK;
   });
 
-  DCHECK(num_notfound == 0u || num_notfound == shard_set->size());
   auto* rb = static_cast<RedisReplyBuilder*>(cmd_cntx.rb);
 
+  // If index is not found on any shard, return error.
```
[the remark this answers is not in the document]
vyavdoshenko: Fixed

src/server/search/search_family.cc
```diff
@@ -1405,9 +1405,11 @@ void SearchFamily::FtInfo(CmdArgList args, const CommandContext& cmd_cntx) {
...
   });
 
-  DCHECK(num_notfound == 0u || num_notfound == shard_set->size());
   auto* rb = static_cast<RedisReplyBuilder*>(cmd_cntx.rb);
 
+  // If index is not found on at least one shard, return error.
+  // This can happen during concurrent FT.CREATE/FT.DROPINDEX operations
+  // where some shards may have the index while others don't.
   if (num_notfound > 0u)
     return rb->SendError(IndexNotFoundMsg(idx_name));
```
dranikpg, line 1414: It's a bug, transactional commands should only see consistent state. The issue is somewhere else:

* The command is marked as CO::IDEMPOTENT
* This implies that it can be called multiple times safely and each invocation on a shard should override it's most recent data
* which is actually the case for the data **except for counting errors**
* so here INFO is called on some shards, the transaction framework notices that data was overwritten and needs to be re-read for consistency and calls it again - the data is found, but the errors from the first invocation are persisted

So instead we should count how many entries were ** eventually** found, most likely just by counting how many entries in the vector are non-empty

tests/dragonfly/search_test.py
```diff
@@ -636,3 +636,59 @@ async def test_synonym_persistence(df_server):
...
     assert (await idx.search(Query("car"))).total == 2
+
+
+@dfly_args({"proactor_threads": 4})
+async def test_ft_info_concurrent_create_drop(df_server):
+    """
+    Test that FT.INFO doesn't crash when called concurrently with FT.CREATE/FT.DROPINDEX.
+    The bug was a DCHECK failure when some shards have the index while others don't.
+    """
+    ITERATIONS = 500
```
dranikpg, line 647: It didn't reproduce it for me, only the shell script (at operation x hundred on debug). To reproduce it, you likely have to increase the iterations number, but then this will take more time for a thing that will likely never break - its not a good runtime vs possible **new** bugs found investment

src/server/search/search_family.cc
dranikpg: If you remove the IDEMPOTENT flag the issue will disappear, but it will also reduce performance, so its better to make it safe for the callback to be called multiple times

romange:
I do not know if performance is important here, let's remove idempotent.

Roman Gershman
CTO
---

dranikpg:
> I do not know if performance is important here, let's remove idempotent.

Because...? And if you remove it for Ft.Info, why not for Ft.Search which have the same schematics? Its just a few lines of code that need to be updated

romange:
Ok, ok. Search is called thousands of times, ft.info not. But whatever
works for you

Roman Gershman
CTO
---

tests/dragonfly/search_test.py
vyavdoshenko: I reproduce the crash locally with this test. If you insist, I can remove this test, but in my opinion, it is still useful because it can fail on the CI side. Personally, I execute the flaky test 1000 times to reproduce it locally, and if we still have an issue, it will be exposed.

src/server/search/search_family.cc
vyavdoshenko: I tried the approach you suggested (counting empty entries in the final state instead of using atomic counter inside the callback), but the DCHECK still failed.

- FT.INFO and concurrent FT.CREATE/FT.DROPINDEX operations:
- FT.INFO was using kReadOnlyMask (not a global transaction)
- FT.CREATE/FT.DROPINDEX are GLOBAL_TRANS
- During FT.INFO's ScheduleSingleHop, a concurrent FT.DROPINDEX can execute between shard callbacks

Result: some shards see the index, some don't → DCHECK fails

The fix: changed FT.INFO to CO::GLOBAL_TRANS | CO::NO_AUTOJOURNAL. I kept the count_if approach (instead of atomic counter) since it's cleaner and avoids potential issues if IDEMPOTENT is ever added back.

@dranikpg 
Maybe I didn't get the idea. Please add additional clarification.

vyavdoshenko: Briefly speaking, I made FT.INFO global transaction, because I replaced kReadOnlyMask (that incudes IDEMPOTENT) with GLOBAL_TRANS.
From my undestanding, FT.INFO can't be broken with other global transactions like FT.DROP any longer.

dranikpg: lets do the following:
1. keep the array for checks
2. remove IDEMPOTENT for now

dranikpg: because the correct fix would rely on this: https://github.com/dragonflydb/dragonfly/pull/6215

which is broken 😢

vyavdoshenko: Done

PR #6206, opened 2025-12-12
Pass CommandContext as a mutable pointer to all the commands.

romange:
Sometimes commands need to change the command state, hence the change.


romange:
augment review

abhijat (approved):
looks like mostly mechanical changes to move from ref to ptr and remove the `const_cast` from previous pr

PR #6211, opened 2025-12-13
chore: move some SendError calls to CommandContext

romange:
It won't be possible to use reply builder directly if we want asynchronous execution of the commands. The logic that will decide how to order reply will be done via CommandContext. For now we just move some of the SendError calls in some of the files to use CommandContext.

### DIFF
--- a/src/facade/reply_builder.h
+++ b/src/facade/reply_builder.h
@@ -308,4 +308,11 @@ class RedisReplyBuilder : public RedisReplyBuilderBase {
   void SendEmptyArray();
 };
 
+#define RETURN_ON_PARSE_ERROR(parser, rb)       \
+  do {                                          \
+    if (auto err = (parser).TakeError(); err) { \
+      return (rb)->SendError(err.MakeReply());  \
+    }                                           \
+  } while (0)
+
 }  // namespace facade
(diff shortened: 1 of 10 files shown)


src/facade/reply_builder.h
```diff
@@ -308,4 +308,11 @@ class RedisReplyBuilder : public RedisReplyBuilderBase {
   void SendEmptyArray();
 };
 
+#define RETURN_ON_PARSE_ERROR(parser, rb)       \
```
vyavdoshenko, line 311: nit:
Consider this:
```
template<typename T>
bool HandleParseError(CmdArgParser& parser, T* ctx) {
  if (parser.HasError()) {
    ctx->SendError(parser.TakeError().MakeReply());
    return true;
  }
  return false;
}
```

Using: `if HandleParseError(...) return;`

vyavdoshenko (approved):
lgtm

src/facade/reply_builder.h
romange: yeah, the reason I used macro is:
1. reply builder is unaware of  `CmdArgParser` and `CmdArgParser` is unaware of reply builder.
2. I want wrap the action of return `if HandleParseError(...) return;` as well.
3. Sometimes we use reply builder (for now) and sometimes CommandContext that currently resides in server code.

PR #6212, opened 2025-12-14
fix: regression tests

romange:
1. Disable flaky test_SearchRequestDistribution
2. Free disk space for hosted runners.

### DIFF
--- a/.github/workflows/ioloop-v2-regtests.yml
+++ b/.github/workflows/ioloop-v2-regtests.yml
@@ -21,6 +21,8 @@ jobs:
       options: --security-opt seccomp=unconfined --sysctl "net.ipv6.conf.all.disable_ipv6=0"
       volumes:
         - /var/crash:/var/crash
+        - /:/hostroot
+        - /mnt:/mnt
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/actions/regression-tests/action.yml
```diff
@@ -62,6 +62,19 @@ runs:
...
+    - name: Free disk space
+      if: contains(runner.labels, 'self-hosted') == false
+      shell: bash
+      run: |
+        echo "===================Before freeing up space ============================================"
+        df -h
+        rm -rf /hostroot/usr/share/dotnet
+        rm -rf /hostroot/usr/local/share/boost
+        rm -rf /hostroot/usr/local/lib/android
+        rm -rf /hostroot/opt/ghc
```
abhijat, line 74: These paths are not mounted in `ioloop-v2-regtests` but this step is still run. Does  `/hostroot` need to be mounted in that ~~action~~ workflow?

romange: we should mount them, indeed. this mounts the host root disk, so we could free up some space.

PR #6214, opened 2025-12-14
chore(server): polish CommandContext data structure

romange:
To allow async execution, the i/o module (facade) should pass CommandContext to the server module. Currently this bridge is absent.

Before we can set up such bridge, we need to polish CommandContext.

This PR:
1. Moves CommandContext to conn_context.h
2. Hides `conn_cntx_` behind accessor server_conn_cntx().
3. Change the that accessed raw conn_cntx member to use server_conn_cntx.
3. Unrelated: moves server flag `journal_emulated` out of facade.

No functional changes.

### DIFF
--- a/src/server/command_registry.h
+++ b/src/server/command_registry.h
@@ -16,14 +16,8 @@
 #include "facade/command_id.h"
 #include "facade/facade_types.h"
 
-namespace facade {
-class SinkReplyBuilder;
-}  // namespace facade
-
 namespace dfly {
 
-class ConnectionContext;
-class Transaction;
 namespace CO {
 
 enum CommandOpt : uint32_t {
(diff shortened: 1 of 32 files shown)


src/server/command_registry.h
```diff
@@ -71,28 +65,7 @@ enum class MultiControlKind : uint8_t {
 using CmdCallStats = std::pair<uint64_t, uint64_t>;
 
 class CommandId;
-
-struct CommandContext {
-  CommandContext(const CommandId* _cid, Transaction* _tx, facade::SinkReplyBuilder* _rb,
```
romange, line 76: Moved to conn_context.h

romange:
augment review

src/server/conn_context.h
```diff
@@ -346,4 +347,41 @@ class ConnectionContext : public facade::ConnectionContext {
...
+      : cid(_cid), tx(_tx), rb_(rb), conn_cntx_(cntx) {
+  }
+
+  const CommandId* cid;
+  Transaction* tx;
+
+  uint64_t start_time_ns = 0;
+
+  // number of commands in the last exec body.
+  unsigned exec_body_len = 0;
```
vyavdoshenko, line 363: Do you plan to move it to the private section?

src/server/conn_context.h
```diff
@@ -346,4 +347,41 @@ class ConnectionContext : public facade::ConnectionContext {
                                             bool to_reply);
 };
 
+class CommandContext {
+ public:
+  CommandContext(const CommandId* _cid, Transaction* _tx, facade::SinkReplyBuilder* rb,
+                 ConnectionContext* cntx)
+      : cid(_cid), tx(_tx), rb_(rb), conn_cntx_(cntx) {
```
vyavdoshenko, line 354: I see inconsistent naming.  In my opinion, it would be consistent to use an underscore suffix for all members.

src/server/conn_context.h
romange: this was public before as well.

src/server/conn_context.h
romange: you are right, we keep private members with underscores, public without.

romange: i just did not want to do even more changes. trying to do them in non-huge batches.

PR #6218, opened 2025-12-14
chore: Split CompactObj into Key&Value

dranikpg:
Relevant to #6137 

Keys and Values don't have much in common actually except storing strings. Flags and more complicated objects are exlusive to either of those. 

This PR sets the ground for splitting them into two types that inherit CompactObj. It fixes a lot of type usage where we juggle around types and implicitly convert CompactObj into PrimeValue and back

### DIFF
--- a/src/server/generic_family.cc
+++ b/src/server/generic_family.cc
@@ -319,9 +319,9 @@ OpResult<string> DumpToString(string_view key, const PrimeValue& pv, const OpArg
       return OpStatus::IO_ERROR;
 
     // TODO: allow saving string directly without proxy object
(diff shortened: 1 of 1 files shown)


src/core/compact_object.h
```diff
@@ -566,6 +548,41 @@ inline bool CompactObj::operator==(std::string_view sv) const {
...
+  }
+
+  explicit CompactKey(std::string_view str) : CompactObj{str, true} {
+  }
+
+  void SetString(std::string_view str) {
+    CompactObj::SetString(str, true);
+  }
+
+  CompactKey AsRef() const {
```
BorysTheDev, line 562: why is it AsRef if we do memcpy

PR #6219, opened 2025-12-15
chore: Introduce ParsedCommand

romange:
1. change MP::Command to reference to BackedArguments
   Before the change MP::Command owned BackedArguments.
   However it was short sighted as we want a unified handling of arguments for both Memcached and RESP
   protocols.

2. Introduce a unified object ParsedCommand that requests will be parsed into. Currently we focus on
   memcached flow. The objects will be created/destroyed by ServiceInterface but will be filled by
   facade. Currently we just use new/delete for them. MP::Command is an optional field inside ParsedCommand.
   ParsedCommand is the union of all requirements for both protocol and will hold the command state for
   both facade and server state machines.

### DIFF
--- a/src/facade/parsed_command.h
+++ b/src/facade/parsed_command.h
@@ -46,6 +46,9 @@ class ParsedCommand : public cmn::BackedArguments {
     return rb_;
   }
 
+  ConnectionContext* conn_cntx() const {
+    return conn_cntx_;
+  }
   MemcacheParser::Command* mc_command() const {
     return mc_cmd_.get();
   }


src/facade/dragonfly_connection.h
```diff
@@ -116,7 +116,7 @@ class Connection : public util::Connection {
   using PipelineMessagePtr = std::unique_ptr<PipelineMessage>;
   using PubMessagePtr = std::unique_ptr<PubMessage>;
 
-  using MCPipelineMessagePtr = std::unique_ptr<MCPipelineMessage>;
+  using MCPipelineMessagePtr = MCPipelineMessage*;
```
romange, line 119: get rid of unique_ptr as we will deallocate via ServiceInterface.

src/common/backed_args.h
```diff
@@ -116,6 +116,11 @@ class BackedArguments {
     return std::string_view{storage_.data() + offset, elem_len(index)};
   }
 
+  char* data(uint32_t index) {
+    uint32_t offset = offsets_[index];
+    return storage_.data() + offset;
+  }
```
dranikpg, line 122: you can use `data()` in `at()` or generally juse use `at().data()`

src/facade/parsed_command.h
```diff
@@ -0,0 +1,59 @@
...
+
+  void CreateMemcacheCommand() {
+    mc_cmd_ = std::make_unique<MemcacheParser::Command>();
+    mc_cmd_->backed_args = this;
+  }
+
+  void Init(SinkReplyBuilder* rb, ConnectionContext* conn_cntx) {
+    rb_ = rb;
+    conn_cntx_ = conn_cntx;
+  }
```
dranikpg, line 43: why do you use it instead of constructor? To have a pool with recycling? Then why the constructor?

dranikpg: Even copilot commented about three step intialization :)

dranikpg (approved):

src/facade/parsed_command.h
romange: I will check it in the next PR

PR #6220, opened 2025-12-15
fix: load search index from old snapshots with TAG SORTABLE SEPARATOR order

vyavdoshenko:
Fixes loading of search indices from old snapshots where TAG fields were serialized as "TAG SORTABLE SEPARATOR x" instead of the expected "TAG SEPARATOR x SORTABLE".

The serialization order was changed in #5423 (BLOCKSIZE option), which broke backwards compatibility - snapshots created before that change fail to load with the error: "ERR Field type , is not supported"

Solution: Reorder arguments during RDB load when we detect the old "SORTABLE SEPARATOR" pattern, swapping them to "SEPARATOR x SORTABLE` before executing FT.CREATE.


src/server/rdb_test.cc
```diff
@@ -69,7 +69,12 @@ inline const uint8_t* to_byte(const void* s) {
 }
 
 io::FileSource RdbTest::GetSource(string name) {
-  string rdb_file = base::ProgramRunfile("testdata/" + name);
+  string rdb_file;
+  if (name[0] == '/') {
+    rdb_file = name;  // Absolute path
+  } else {
```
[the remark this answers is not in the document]
vyavdoshenko: fixed

src/server/rdb_test.cc
```diff
@@ -719,6 +724,31 @@ TEST_F(RdbTest, RestoreSearchIndexNameStartingWithColon) {
...
 
+// Test loading old snapshot where TAG fields were serialized as
+// "TAG SORTABLE SEPARATOR x" instead of "TAG SEPARATOR x SORTABLE"
+TEST_F(RdbTest, LoadOldSnapshotTagSortableSeparator) {
+  // Extract the tar archive to a temp directory
+  string tar_file = base::ProgramRunfile("testdata/tag_sortable.tar.gz");
+  char temp_dir[] = "/tmp/tag_sortable_test_XXXXXX";
+  ASSERT_NE(mkdtemp(temp_dir), nullptr);
+  string extract_cmd = absl::StrCat("tar -xf ", tar_file, " -C ", temp_dir);
+  ASSERT_EQ(system(extract_cmd.c_str()), 0);
```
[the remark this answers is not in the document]
vyavdoshenko: fixed

src/server/rdb_load.cc
```diff
@@ -2862,6 +2862,19 @@ void LoadSearchCommandFromAux(Service* service, string&& def, string_view comman
   CmdArgVec arg_vec;
   facade::RespExpr::VecToArgList(resp_vec, &arg_vec);
 
+  // Fix for backwards compatibility with old snapshots where TAG fields were
+  // serialized as "TAG SORTABLE SEPARATOR x" but parser expects
+  // "TAG SEPARATOR x SORTABLE". Reorder arguments if needed.
```
dranikpg: but itsn't it possible to just fix the parser?

dranikpg: the code structure is such because we first parse type specific params and then general ones (noindex, sortable)

dranikpg: I don't have a strong opinion about this fix, lets let @romange  decide

vyavdoshenko: It can be fixed in the parser.

src/server/rdb_test.cc
```diff
@@ -719,6 +726,30 @@ TEST_F(RdbTest, RestoreSearchIndexNameStartingWithColon) {
...
 }
 
+// Test loading old snapshot where TAG fields were serialized as
+// "TAG SORTABLE SEPARATOR x" instead of "TAG SEPARATOR x SORTABLE"
+TEST_F(RdbTest, LoadOldSnapshotTagSortableSeparator) {
+  // Extract the tar archive to a temp directory
+  string tar_file = base::ProgramRunfile("testdata/tag_sortable.tar.gz");
+  char temp_dir[] = "/tmp/tag_sortable_test_XXXXXX";
+  ASSERT_NE(mkdtemp(temp_dir), nullptr);
+  string extract_cmd = absl::StrCat("tar -xf '", tar_file, "' -C '", temp_dir, "'");
```
romange: i think it's an overkill to use tar. why not just add the original files?

romange: i mean, the dfs files.

src/server/rdb_load.cc
```diff
@@ -2862,6 +2862,20 @@ void LoadSearchCommandFromAux(Service* service, string&& def, string_view comman
   CmdArgVec arg_vec;
   facade::RespExpr::VecToArgList(resp_vec, &arg_vec);
 
+  // Temporary migration fix for backwards compatibility with old snapshots where TAG fields were
+  // serialized as "TAG SORTABLE SEPARATOR x" but parser expects "TAG SEPARATOR x SORTABLE".
+  // Reorder arguments if needed.
+  // TODO: Remove this workaround after sufficient time has passed to migrate the snapshots.
```
romange: nit, set a specific date, Apr 2026.

romange (approved):

src/server/rdb_load.cc
vyavdoshenko: fixed

PR #6223, opened 2025-12-15
chore: further clean ups around ParsedCommand.

romange:
1. Refactored captured payload so it could be used to fill the reply from shard threads.
2. Removed `memcache_flag` from connection state as it's been redundant now, when we have per command state.
3. Devised the plan for transaction changes around the SET command.


romange:
@dranikpg  most of the changes because of me factoring out Payload types into a separate namespace.

The functional change is around removal of memcache_flag as we now pass the command context which we can read directly.

romange:
augment review

romange:
@dranikpg  ping

PR #6230, opened 2025-12-16
chore: simplify memcache pipelining flow

romange:
Before - memcache piggybacked on the RESP flow for pipelining.
It was overkill and overcomplicated as memcache does not perform command squashing,
and the only actual optimization that helped with increasing throughput for memcache pipelining was
the fact we batched replies via reply_builder_->SetBatch interface.

This PR rewrites the IO flow for memcache, by introducing an intrusive queue of parsed commands.
The first step is to parse one or many commands and the second
step is to execute them one by one. We enable batching of replies as long as we have more commands in queue.

The throughput of the new approach is similar to what we had before but we do not use dispatch fiber for this,
and we perform pipelining in batches of limited size.

The flow looks more and more like what we need for fully asynchronous dispatching.
Once `Connection::ExecuteMCBatch` won't block on DispatchMC, the throughput of a single connection will skyrocket.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -412,9 +412,6 @@ size_t Connection::MessageHandle::UsedMemory() const {
     size_t operator()(const InvalidationMessage& msg) {
       return 0;
     }
-    size_t operator()(const MCPipelineMessagePtr& msg) {
-      return sizeof(MCPipelineMessage) + msg->HeapMemory();
-    }
   };
 
   return sizeof(MessageHandle) + visit(MessageSize{}, this->handle);
@@ -422,9 +419,7 @@ size_t Connection::MessageHandle::UsedMemory() const {
 
 bool Connection::MessageHandle::IsReplying() const {
   return IsPubMsg() || holds_alternative<MonitorMessage>(handle) ||
-         holds_alternative<PipelineMessagePtr>(handle) ||
-         (holds_alternative<MCPipelineMessagePtr>(handle) &&
-          !get<MCPipelineMessagePtr>(handle)->mc_command()->no_reply);
+         holds_alternative<PipelineMessagePtr>(handle);
 }
 
 struct Connection::AsyncOperations {
@@ -434,7 +429,6 @@ struct Connection::AsyncOperations {
 
   void operator()(const PubMessage& msg);
   void operator()(PipelineMessage& msg);
-  void operator()(MCPipelineMessagePtr msg);
   void operator()(const MonitorMessage& msg);
   void operator()(const AclUpdateMessage& msg);
   void operator()(const MigrationRequestMessage& msg);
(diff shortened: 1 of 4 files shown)


romange:
augment review

romange:
augment review

tests/dragonfly/pymemcached_test.py
```diff
@@ -125,6 +125,34 @@ def test_length_in_set_command(df_server: DflyInstance, memcached_client: MCClie
...
+    while len(response) < expected_len:
+        data = client.recv(1024)
+        if not data:
+            break
+        response += data
+    return response
+
+
+@dfly_args(DEFAULT_ARGS)
+def test_error_in_pipeline(df_server: DflyInstance, memcached_client: MCClient):
```
romange, line 139: that did not behave correctly before

src/facade/dragonfly_connection.cc
```diff
@@ -2065,13 +2036,97 @@ bool Connection::IsReplySizeOverLimit() const {
...
+
+void Connection::ExecuteMCBatch(bool has_more) {
+  MCReplyBuilder* builder = static_cast<MCReplyBuilder*>(reply_builder_.get());
+
+  bool is_pipeline = parsed_cmd_q_len_ > 1;
+
+  // Execute sequentially all parsed commands.
+  while (parsed_head_) {
+    bool batch_mode = has_more || (parsed_head_->next != nullptr);
+    reply_builder_->SetBatchMode(batch_mode);
```
dranikpg, line 2074: Please add a Todo to fix noreply mode

dranikpg: You actually removed a condition connected to it when checking msg.IsReplying()

src/facade/dragonfly_connection.cc
```diff
@@ -2065,13 +2036,97 @@ bool Connection::IsReplySizeOverLimit() const {
...
   parsed_cmd_->Init(reply_builder_.get(), cc_.get());
   if (protocol_ == Protocol::MEMCACHE)
     parsed_cmd_->CreateMemcacheCommand();
 }
 
+void Connection::EnqueueParsedCommand() {
+  parsed_cmd_->next = nullptr;
+  parsed_cmd_->parsed_cycle = base::CycleClock::Now();
+  if (parsed_head_ == nullptr) {
+    parsed_head_ = parsed_cmd_;
```
dranikpg, line 2116: Maybe use boost intrusive list in the future?

dranikpg (approved):

src/facade/dragonfly_connection.cc
dranikpg: Because it won't cause a write -> it wont flush

src/facade/dragonfly_connection.cc
romange: do you think it will contribute to clarity or to code conciseness?

src/facade/dragonfly_connection.cc
romange: what's the use-case what is the expected behavior?
I specifically tested it on `test_noreply_pipeline` test and it passed.

PR #6236, opened 2025-12-17
chore: further replace SendError calls with CommandContext

romange:
The goal is that all reply calls will go through CommandContext. Right now I am transitioning SendError calls.


vyavdoshenko (approved):
lgtm

PR #6240, opened 2025-12-18
feat(json): Improve json defragmentation

abhijat:
Instead of doing a serialize-deserialize roundtrip for the full object, walk the node as a tree, only defragment specific nodes as needed, and use knowledge about the node storage to defragment.

The actions in the changeset rely on existing move ctor and move-assignment operator within jsoncons library to perform reallocations and cleanup.

A couple of benchmarks are added to compare the new approach with serialization based defrag, and a benchmark for more realistic defragment where not every single node is fragmented.

Note that the new defragmentation is not hooked up to the compacted object code to make the pr small and easier to review. Once these changes are finalized and merged the next PR will use this new code path. The serialization based defragmentation can then be cleaned up.

The original thought was to add this code in our `jsoncons` fork, but after some exploration this codebase seems the right place for it, as defragmentation and related tools (`PageUsage` etc) are all in this code base.

Note for reviewer: despite the large size the bulk of this PR is unit tests and benchmarks.

### DIFF
--- a/src/core/page_usage_stats_test.cc
+++ b/src/core/page_usage_stats_test.cc
@@ -7,6 +7,8 @@
 #include <absl/flags/reflection.h>
 #include <gmock/gmock-matchers.h>
 
+#include <random>
+
 #include "base/gtest.h"
 #include "base/logging.h"
 #include "core/compact_object.h"
@@ -29,6 +31,84 @@ ABSL_DECLARE_FLAG(bool, experimental_flat_json);
 
 using namespace dfly;
 using namespace std::chrono_literals;
+
+namespace {
+
+std::string GenerateTestJSON(size_t num_objects) {
+  std::string data = R"({"contents":[)";
+  for (size_t i = 0; i < num_objects; ++i) {
+    const auto si = std::to_string(i);
+    data += R"({"id":)" + si + R"(,"class":"v___)" + si + R"(","value":)" + si + R"(})";
+    if (i < num_objects - 1) {
+      data += ",";
+    }
+  }
+  data += R"(], "data": "some", "count": 1, "checked": false})";
+  return data;
+}
+
+// Helper to defragment only if a randomly generated value is less than preset probability. For
+// benchmarking realistic situations, where some nodes are fragmented and others are not
+class SelectiveDefragment : public PageUsage {
+ public:
+  explicit SelectiveDefragment(const double fragmentation_probability)
+      : PageUsage(CollectPageStats::NO, 0), frag_prob_{fragmentation_probability} {
+  }
+
+  bool IsPageForObjectUnderUtilized(void*) override {
+    return dist_(rng_) < frag_prob_;
+  }
+
+ private:
+  double frag_prob_;
+  std::mt19937 rng_{99};
+  std::uniform_real_distribution<double> dist_{0.0, 1.0};
+};
+
+struct MemStats {
+  size_t total_reserved{0};
+  size_t total_committed{0};
+  size_t total_used{0};
+  size_t total_wasted{0};
+  size_t num_pages{0};
+};
+
+MemStats LogMemStats(const mi_heap_t* heap) {
+  MemStats stats;
+  mi_heap_visit_blocks(
+      heap, false,
+      [](const mi_heap_t* /*h*/, const mi_heap_area_t* area, void* /*block*/, size_t block_size,
+         void* arg) {
+        const size_t committed = area->committed;
+        const size_t used = area->used * block_size;
+
+        const auto s = static_cast<MemStats*>(arg);
+        s->num_pages++;
+        s->total_committed += committed;
+        s->total_reserved += area->reserved;
+        s->total_used += used;
+        s->total_wasted += committed - used;
+
+        return true;
+      },
+      &stats);
+
+  LOG(INFO) << "Pages: " << stats.num_pages;
+  LOG(INFO) << "Reserved : " << stats.total_reserved << " bytes";
+  LOG(INFO) << "Committed: " << stats.total_committed << " bytes";
+  LOG(INFO) << "Used: " << stats.total_used << " bytes";
+  LOG(INFO) << "Wasted: " << stats.total_wasted << " bytes";
+  if (stats.total_committed) {
+    LOG(INFO) << "Wasted%: "
+              << static_cast<double>(stats.total_wasted) / stats.total_committed * 100.0;
+    LOG(INFO) << "Utilization%: "
+              << static_cast<double>(stats.total_used) / stats.total_committed * 100.0;
+  }
+
+  return stats;
+}
+
+}  // namespace
 
 class PageUsageStatsTest : public ::testing::Test {
  protected:
@@ -194,15 +274,7 @@ TEST_F(PageUsageStatsTest, JSONCons) {
   // still fail. This is because freeing the compact object code path takes the wrong branch based
   // on encoding. The flat encoding was tested manually adjusting this same test with changed
   // encoding.
-  std::string data = R"({"contents":[)";
-  for (size_t i = 0; i < 1000; ++i) {
-    const auto si = std::to_string(i);
-    data += R"({"id":)" + si + R"(,"class":"v___)" + si + R"("})";
-    if (i < 999) {
-      data += ",";
-    }
-  }
-  data += R"(], "data": "some", "count": 1, "checked": false})";
+  std::string data = GenerateTestJSON(1000);
 
   auto* mr = static_cast<MiMemoryResource*>(CompactObj::memory_resource());
   size_t before = mr->used();
@@ -252,7 +324,7 @@ TEST_F(PageUsageStatsTest, JsonDefragNested) {
   p.SetForceReallocate(true);
 
   Defragment(parsed.value(), &p);
-  EXPECT_EQ(parsed->get_allocator<>(), old_alloc);
+  EXPECT_EQ(parsed->at("a").at("b").at("c").at("d").as_string_view(), "value");
 }
 
 TEST_F(PageUsageStatsTest, JsonDefragRemainsInSameHeap) {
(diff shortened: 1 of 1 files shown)


abhijat:
with google benchmark's comparison script it shows a three times speedup, but in reality it might be better, because in comparison benchmark we assume all nodes always need defragmentation, which is not going to be realistic

```
./build-opt/_deps/benchmark-src/tools/compare.py filters ./build-dbg/page_usage_stats_test BM_JSONDefragmentSerializeOld BM_JSONDefragmentTreeWalkNew
RUNNING: ./build-dbg/page_usage_stats_test --benchmark_filter=BM_JSONDefragmentSerializeOld --benchmark_out=/tmp/tmplq2khe4h
2025-12-18T15:05:31+05:30
Running ./build-dbg/page_usage_stats_test
Run on (32 X 1919.57 MHz CPU s)
CPU Caches:
  L1 Data 48 KiB (x16)
  L1 Instruction 32 KiB (x16)
  L2 Unified 2048 KiB (x16)
  L3 Unified 36864 KiB (x1)
Load Average: 0.57, 1.56, 1.83
***WARNING*** Library was built as DEBUG. Timings may be affected.
-----------------------------------------------------------------------------------------------
Benchmark                                                     Time             CPU   Iterations
-----------------------------------------------------------------------------------------------
BM_JSONDefragmentSerializeOld/objects_per_json:100       319710 ns       318920 ns         2223
BM_JSONDefragmentSerializeOld/objects_per_json:125       405072 ns       403909 ns         1771
BM_JSONDefragmentSerializeOld/objects_per_json:625      2047592 ns      2041246 ns          344
BM_JSONDefragmentSerializeOld/objects_per_json:3125    10622879 ns     10590007 ns           66
BM_JSONDefragmentSerializeOld/objects_per_json:10000   33663689 ns     33573157 ns           20
RUNNING: ./build-dbg/page_usage_stats_test --benchmark_filter=BM_JSONDefragmentTreeWalkNew --benchmark_out=/tmp/tmp0cwzzoh8
2025-12-18T15:05:39+05:30
Running ./build-dbg/page_usage_stats_test
Run on (32 X 5700 MHz CPU s)
CPU Caches:
  L1 Data 48 KiB (x16)
  L1 Instruction 32 KiB (x16)
  L2 Unified 2048 KiB (x16)
  L3 Unified 36864 KiB (x1)
Load Average: 0.64, 1.54, 1.82
***WARNING*** Library was built as DEBUG. Timings may be affected.
----------------------------------------------------------------------------------------------
Benchmark                                                    Time             CPU   Iterations
----------------------------------------------------------------------------------------------
BM_JSONDefragmentTreeWalkNew/objects_per_json:100       109263 ns       108945 ns         6332
BM_JSONDefragmentTreeWalkNew/objects_per_json:125       136242 ns       135851 ns         5152
BM_JSONDefragmentTreeWalkNew/objects_per_json:625       658140 ns       656566 ns         1052
BM_JSONDefragmentTreeWalkNew/objects_per_json:3125     3281013 ns      3271557 ns          217
BM_JSONDefragmentTreeWalkNew/objects_per_json:10000   10257647 ns     10229715 ns           68
Comparing BM_JSONDefragmentSerializeOld to BM_JSONDefragmentTreeWalkNew (from ./build-dbg/page_usage_stats_test)
Benchmark                                                                                                 Time             CPU      Time Old      Time New       CPU Old       CPU New
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
[BM_JSONDefragmentSerializeOld vs. BM_JSONDefragmentTreeWalkNew]/objects_per_json:100                  -0.6582         -0.6584        319710        109263        318920        108945
[BM_JSONDefragmentSerializeOld vs. BM_JSONDefragmentTreeWalkNew]/objects_per_json:125                  -0.6637         -0.6637        405072        136242        403909        135851
[BM_JSONDefragmentSerializeOld vs. BM_JSONDefragmentTreeWalkNew]/objects_per_json:625                  -0.6786         -0.6784       2047592        658140       2041246        656566
[BM_JSONDefragmentSerializeOld vs. BM_JSONDefragmentTreeWalkNew]/objects_per_json:3125                 -0.6911         -0.6911      10622879       3281013      10590007       3271557
[BM_JSONDefragmentSerializeOld vs. BM_JSONDefragmentTreeWalkNew]/objects_per_json:10000                -0.6953         -0.6953      33663689      10257647      33573157      10229715
OVERALL_GEOMEAN                                                                                        -0.6777         -0.6777             0             0             0             0
```

src/core/json/json_object.cc
```diff
@@ -62,4 +138,45 @@ JsonType DeepCopyJSON(const JsonType* j) {
   return std::move(deserialized.value());
 }
 
+void Defragment(JsonType& j, PageUsage* page_usage) {
+  // stack-based traversal inspired from jsoncons::basic_json::compute_memory_size
+  std::stack<JsonType*> stack;
+  stack.push(&j);
+
+  while (!stack.empty()) {
+    JsonType* current = stack.top();
```
abhijat, line 147: Once we have a cycle quota based system it might be worth it to check here and bail out if we ran over limit so we can partially defragment while not hogging the cpu.

src/core/json/json_object.cc
```diff
@@ -62,4 +138,45 @@ JsonType DeepCopyJSON(const JsonType* j) {
   return std::move(deserialized.value());
 }
 
+void Defragment(JsonType& j, PageUsage* page_usage) {
```
abhijat, line 141: These new functions can also live somewhere else, eg compact object or similar, then we would not have link `dfly_page_usage` with json library. I found it more logical to place it here with other json code.

src/core/page_usage_stats_test.cc
```diff
@@ -426,3 +564,138 @@ TEST_F(PageUsageStatsTest, DefragmentIndexWithNonDefragmentableFields) {
...
+    if (j.has_value()) {
+      Defragment(j.value(), &p);
+    }
+  }
+
+  mi_heap_collect(m_.heap(), true);
+  const auto after = LogMemStats(m_.heap());
+
+  EXPECT_LT(after.total_wasted, before.total_wasted);
+  EXPECT_LT(after.total_committed, before.total_committed);
```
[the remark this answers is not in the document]
abhijat: This is the core assertion of the test. The value must go down because pages are freed due to defragmentation.

src/core/json/json_object.cc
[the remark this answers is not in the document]
abhijat: It is safe to assume page usage is not null, the defragmentation call chain begins with a valid page usage object.

src/core/page_usage_stats_test.cc
```diff
@@ -426,3 +564,138 @@ TEST_F(PageUsageStatsTest, DefragmentIndexWithNonDefragmentableFields) {
...
+    static MiMemoryResource m{tlh};
+    InitTLStatelessAllocMR(&m);
+    CompactObj::InitThreadLocal(&m);
+    initialized = true;
+  }
+}
+
+}  // namespace
+
+void BM_JSONDefragSerialize(benchmark::State& state) {
```
abhijat, line 626: this and the next comparison focused benchmark can be removed in the next pr once the serialization based defrag is cleanup up.

src/core/json/json_object.cc
```diff
@@ -62,4 +138,45 @@ JsonType DeepCopyJSON(const JsonType* j) {
...
+    switch (storage_kind) {
+      case json_storage_kind::byte_str:
+        DefragmentByteString(*current, page_usage);
+        break;
+      case json_storage_kind::long_str:
+        DefragmentLongString(*current, page_usage);
+        break;
+      case json_storage_kind::object: {
+        DefragmentJsonObject(*current, page_usage);
+        auto& object = current->cast<JsonType::object_storage>().value();
```
abhijat, line 160: We already did this cast in the defragment function. So this is a waste, but only if we did not defragment. If we did defragment, then the object cannot be reused, it has probably changed. 

An optimization can be done here to avoid doing two casts for non-fragmented pages, and just reuse the object ref. from the method. But it did not yield much improvement in the benchmark with 50% fragmentation probability, so I am leaving it as it is.

src/core/json/json_object.cc
```diff
@@ -42,6 +45,79 @@ std::optional<T> ParseWithDecoder(std::string_view input, json_decoder<T>&& deco
...
+  JsonType::string_view_type svt{str_storage.data(), str_storage.length()};
+  j = JsonType(svt, j.tag(), str_storage.get_allocator());
+}
+
+// Allocates a new json object of type json_object_arg, with fresh memory allocation for its
+// contained vector of key value pairs. Then moves members from j to this new object. Finally j is
+// swapped with the new object.
+void DefragmentJsonObject(JsonType& j, PageUsage* page_usage) {
+  auto& object = j.cast<JsonType::object_storage>().value();
+  if (object.empty() || !page_usage->IsPageForObjectUnderUtilized(&*object.begin()))
```
romange, line 83: this condition is not accurate. at  least comment about it.
you check that the array storage should be defragmented. and assuming it should not, 
you stop early and do not try defragmenting sub-objects.

src/core/json/json_object.cc
```diff
@@ -42,6 +45,79 @@ std::optional<T> ParseWithDecoder(std::string_view input, json_decoder<T>&& deco
...
+  JsonType new_node{json_object_arg, j.tag(), object.get_allocator()};
+  new_node.reserve(object.size());
+
+  for (auto& member : object) {
+    // The member values are JsonType themselves, they just wrap pointers to actual storage.
+    // Their move invokes the move ctor in jsoncons, which will move the value wrappers to new_node,
+    // and leave the original in `j` holding references to `null_storage` type, see
+    // `uninitialized_move_a` in jsoncons. The member key (a string) is not moved but copied into
+    // new_node members.
+    new_node.try_emplace(member.key(), std::move(member.value()));
```
romange, line 96: is key/value ever defragmented with such flow?

src/core/json/json_object.cc
romange: ok, I see now how you use stack below to reach sub-members. Have you checked your code somehow?  I suspect it is buggy and you enqueue pointers to objects that are moved somewhere else

src/core/page_usage_stats_test.cc
```diff
@@ -30,6 +32,84 @@ ABSL_DECLARE_FLAG(bool, experimental_flat_json);
 using namespace dfly;
 using namespace std::chrono_literals;
 
+namespace {
+
+std::string GenerateTestJSON(size_t num_objects) {
+  std::string data = R"({"contents":[)";
+  for (size_t i = 0; i < num_objects; ++i) {
+    const auto si = std::to_string(i);
```
romange, line 40: this string is too short to trigger de-fragmentation. 
I would like a confirmation from you that your code reaches all the code paths including recursive defragmentation you implemented. if you have a invalid data access bug  and the test has not caught it - then please improve the test first to uncover it.

src/core/json/json_object.cc
abhijat: Defragmenting and accessing child values are done as two separate operations to avoid using stale, moved values. 

I cast again to safely obtain current storage from array or object after defragmentation is complete, the object and array both are tested in some unit tests in this PR.

src/core/page_usage_stats_test.cc
abhijat: Large string is tested in `JsonDefragRemainsInSameHeap`, in that test we check that the pointers moved and the new pointers are in the same heap etc. 

To ensure that defragmentation actually happened, in all tests I use `PageUsage` class with a special flag set as below:
```
  p.SetForceReallocate(true);
```

 this `force_reallocate_` flag forces reallocation irrespective of the fragmentation state (it actually does not check page usage, it just returns true for underutilized), this is done to make sure that the defragmentation code runs even if the page is well utilized.

abhijat: I was running through a debugger to provide confirmation like you asked for when the auto merge kicked in and merged it. I will update if I find an issue but I'm pretty sure that the reasoning is correct. 

In either case its not hooked in yet so if I do find an issue I will correct it before the next pr.

abhijat: I tried a test input with debugger and force reallocation turned on, it seems to work as expected, I added some prints to show the data states:

```
input: [{"a":{"b":"true"},"array":[1,2,3,4]}]
arr moved from: [null]
arr moved to:   [{"a":{"b":"true"},"array":[1,2,3,4]}]
 after defrag json arr member: {"a":{"b":"true"},"array":[1,2,3,4]}
obj moved from: {"a":null,"array":null}
obj moved to:   {"a":{"b":"true"},"array":[1,2,3,4]}
 after defrag json obj member: {"b":"true"}
 after defrag json obj member: [1,2,3,4]
arr moved from: [1,2,3,4]
arr moved to:   [1,2,3,4]
 after defrag json arr member: 1
 after defrag json arr member: 2
 after defrag json arr member: 3
 after defrag json arr member: 4
   trivial: 4
   trivial: 3
   trivial: 2
   trivial: 1
obj moved from: {"b":"true"}
obj moved to:   {"b":"true"}
 after defrag json obj member: "true"
   trivial: "true"
```

with the following test data
```
diff --git a/src/core/json/json_object.cc b/src/core/json/json_object.cc
index 848f6e0d..cfcbb041 100644
--- a/src/core/json/json_object.cc
+++ b/src/core/json/json_object.cc
@@ -101,0 +102,7 @@ void DefragmentJsonObject(JsonType& j, PageUsage* page_usage) {
+
+  std::cout << "obj moved from: ";
+  new_node.dump(std::cout);
+  std::cout << "\n";
+  std::cout << "obj moved to:   ";
+  j.dump(std::cout);
+  std::cout << "\n";
@@ -118,0 +126,6 @@ void DefragmentJsonArray(JsonType& j, PageUsage* page_usage) {
+  std::cout << "arr moved from: ";
+  new_node.dump(std::cout);
+  std::cout << '\n';
+  std::cout << "arr moved to:   ";
+  j.dump(std::cout);
+  std::cout << '\n';
@@ -145,0 +159,3 @@ void Defragment(JsonType& j, PageUsage* page_usage) {
+  std::cout << "input: ";
+  j.dump(std::cout);
+  std::cout << "\n";
@@ -161,0 +178,3 @@ void Defragment(JsonType& j, PageUsage* page_usage) {
+          std::cout << " after defrag json obj member: ";
+          member.value().dump(std::cout);
+          std::cout << "\n";
@@ -169,0 +189,3 @@ void Defragment(JsonType& j, PageUsage* page_usage) {
+          std::cout << " after defrag json arr member: ";
+          member.dump(std::cout);
+          std::cout << "\n";
@@ -174,0 +197,3 @@ void Defragment(JsonType& j, PageUsage* page_usage) {
+        std::cout << "   trivial: ";
+        current->dump(std::cout);
+        std::cout << "\n";
diff --git a/src/core/page_usage_stats_test.cc b/src/core/page_usage_stats_test.cc
index 7846c827..be773fae 100644
--- a/src/core/page_usage_stats_test.cc
+++ b/src/core/page_usage_stats_test.cc
@@ -319 +319,2 @@ TEST_F(PageUsageStatsTest, JsonDefragNested) {
-  constexpr auto data = R"({"a":{"b":{"c":{"d":"value"}}}})";
+  // constexpr auto data = R"({"a":{"b":{"c":{"d":"value"}}}})";
+  constexpr auto data = R"([{"a":{"b":"true"},"array":[1,2,3,4]}])";
@@ -327 +328 @@ TEST_F(PageUsageStatsTest, JsonDefragNested) {
-  EXPECT_EQ(parsed->at("a").at("b").at("c").at("d").as_string_view(), "value");
+  // EXPECT_EQ(parsed->at("a").at("b").at("c").at("d").as_string_view(), "value");
```

src/core/json/json_object.cc
abhijat: Looking at the json object api I found it easier to move the object to a new one and swap rather than try to modify the existing object, as the move ctors do exactly what we need to achieve defragmentation. As the values are just wrapper around actual data this is not very expensive and we handle the wrapped data later when visiting the child nodes in the stack.

PR #6241, opened 2025-12-18
chore: minor fixes around async execution

romange:
1. Introduced StoredReply to reply payload so that we could reply STORED/SKIPPED in MC protocol.
2. reorganized code in facade.
3. ReplyGuard does not crash now if async commands did not reply.

### DIFF
--- a/src/facade/facade.cc
+++ b/src/facade/facade.cc
@@ -219,20 +219,57 @@ std::string AbslUnparseFlag(const MemoryBytesFlag& flag) {
   return strings::HumanReadableNumBytes(flag.value);
 }
 
-void ParsedCommand::SendError(std::string_view str, std::string_view type) const {
-  rb_->SendError(str, type);
+void ParsedCommand::ResetForReuse() {
+  reply_direct_ = true;
+  reply_payload_ = std::monostate{};
+  dispatch_async_ = false;
+
+  offsets_.clear();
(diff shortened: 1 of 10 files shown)


src/facade/facade.cc
```diff
@@ -219,22 +219,59 @@ std::string AbslUnparseFlag(const MemoryBytesFlag& flag) {
...
-  if (status == OpStatus::OK)
-    return rb_->SendSimpleString("OK");
-  rb_->SendError(StatusToMsg(status));
+void ParsedCommand::SendError(std::string_view str, std::string_view type) {
+  if (reply_direct_) {
+    rb_->SendError(str, type);
+  } else {
+    reply_payload_ = payload::make_error(str, type);
+  }
+}
```
dranikpg, line 240: I know your not immediately clean writing all the code, but your interfaces are not fully planned ahead.

You're letting complexity escape - let rb be a capturing reply builder if we want to capture the reply. But would be introducing something better than the capturing reply builder so we write to buffers directly without small nested allocations

romange: I agree with the first point - I do not plan interfaces ahead because I am still in the "POC mode".

I did not understand your suggestion though. But in any case I suggest to wait for full end-to-end flow for a working Memcache SET to see the full picture. I will be open to suggestions on how it is possible to improve.

PR #6242, opened 2025-12-18
chore: reduce compilation dependencies between server and facade

romange:
1. Factor out ArgRange into common/arg_range header
2. Factor out ReplyMode into reply_mode.h
3. Remove unneeded includes.

This helps with avoiding recompilation of the entire server code when I change a random thing in facade.


vyavdoshenko (approved):
lgtm

Issue #6200, opened 2025-12-11
error during teardown of `test_SearchRequestDistribution`


abhijat:
this test often fails due to timeout recently

https://github.com/dragonflydb/dragonfly/actions/runs/20132447584/job/57777046375#step:6:2593

```
E           subprocess.TimeoutExpired: Command '['/__w/dragonfly/dragonfly/build/dragonfly', '--proactor_threads=4', '--cluster_mode=yes', '--cluster_search=yes', '--port=30169', '--admin_port=30170', '--vmodule=coordinator=2,search_family=3,redis_parser=3', '--dbfilename=', '--noversion_check', '--maxmemory=8G', '--jsonpathv2', '--omit_basic_usage', '--latency_tracking', '--log_dir=/tmp/dragonfly_logs/test_SearchRequestDistribution_df_factory0_', '--serialization_max_chunk_size=300000', '--fiber_safety_margin=4096', '--force_epoll=true', '--num_shards=3']' timed out after 120 seconds
```

dragonfly fails to stop. there is also list of fibers printed https://github.com/dragonflydb/dragonfly/actions/runs/20132447584/job/57777046375#step:6:1438

abhijat:
https://github.com/dragonflydb/dragonfly/actions/runs/20113169619
https://github.com/dragonflydb/dragonfly/actions/runs/20117341444
https://github.com/dragonflydb/dragonfly/actions/runs/20120970208
https://github.com/dragonflydb/dragonfly/actions/runs/20123624250
https://github.com/dragonflydb/dragonfly/actions/runs/20127624767
https://github.com/dragonflydb/dragonfly/actions/runs/20132447584

abhijat:
does not always happen on ARM (although majority happens there), there are a couple of x86 errors too.

abhijat:
https://github.com/dragonflydb/dragonfly/actions/runs/20142644534/job/57814287393#step:6:2523
https://github.com/dragonflydb/dragonfly/actions/runs/20147320915/job/57831160224#step:6:2541
https://github.com/dragonflydb/dragonfly/actions/runs/20151545853/job/57845251966#step:6:2540
https://github.com/dragonflydb/dragonfly/actions/runs/20155289026/job/57856333436#step:6:2549

abhijat:
https://github.com/dragonflydb/dragonfly/actions/runs/20170789507
https://github.com/dragonflydb/dragonfly/actions/runs/20175534633
[...]

PR #6244, opened 2025-12-18
fix: cross-shard search tests

BorysTheDev:
fixes: #6200

- added RESPParser into ProtocolClient
- fix deadlock in the Coordinator
- removed OwnedRespExpr

### DIFF
--- a/src/server/cluster/coordinator.cc
+++ b/src/server/cluster/coordinator.cc
@@ -88,13 +88,14 @@ class Coordinator::CrossShardClient : public ProtocolClient {
     {
       std::lock_guard lk(send_mu_);
       send_queue_.push(req);
+      ready_to_send_ = true;
     }
     {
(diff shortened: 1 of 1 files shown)


src/server/cluster/coordinator.cc
```diff
@@ -134,6 +140,7 @@ class Coordinator::CrossShardClient : public ProtocolClient {
         resp_queue_.front()->Exec(*resp);
         resp_queue_.pop();
       }
+      ready_to_resp_ = false;
```
dranikpg, line 144: I havent looked at the code deeply, but I see its an atomic and you have to be careful with those if they are used by multiplt threads. What if it becase true at the same instant due to send? You have to either hold resp_mu_ while setting ready=true above or do some test_and_set magic

dranikpg (approved):

PR #6248, opened 2025-12-18
chore: switch SendError calls in zset_family

romange:

### DIFF
--- a/src/server/zset_family.cc
+++ b/src/server/zset_family.cc
@@ -618,10 +618,6 @@ bool ParseLexBound(string_view src, ZSetFamily::LexBound* bound) {
   return true;
 }
 
-void SendAtLeastOneKeyError(string_view cmd, SinkReplyBuilder* builder) {
-  builder->SendError(absl::StrCat("at least 1 input key is needed for ", cmd));
-}
-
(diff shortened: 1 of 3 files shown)


dranikpg:
What was the rationale of separate SendError call on the context? Just less code?

src/server/zset_family.cc
```diff
@@ -2730,8 +2724,7 @@ void CmdZRandMember(CmdArgList args, CommandContext* cmd_cntx) {
   if (parser.HasNext())
     return rb->SendError(absl::StrCat("Unsupported option:", string_view(parser.Next())));
```
abhijat, line 2725: This should use `cmd_cntx->SendError` also.

abhijat:
There are a few places in `zset_family` where there is still `rb->SendError` although command context is available.

src/server/zset_family.cc
romange: i will fix in subsequent PRs.

romange:
> What was the rationale of separate SendError call on the context? Just less code?

Eventually, I would like to reduce the API surface of ReplyBuilder and move most of the high level logic to ParsedCommand.

PR #6251, opened 2025-12-19
chore: refactor ParsedCommand implementation into a separate file

romange:
1. Move the existing implementation from facade.cc
2. Every function that fills reply_payload notifies the connection fiber.
3. Introduce CheckDoneAndMarkHead/NotifyReplied that make sure that the connection never misses the notifications about ready replies.
4. Implemented SendPayload function that sends the payload to reply_builder.

### DIFF
--- a/src/facade/parsed_command.cc
+++ b/src/facade/parsed_command.cc
@@ -8,57 +8,11 @@
 #include "facade/conn_context.h"
 #include "facade/dragonfly_connection.h"
 #include "facade/reply_builder.h"
+#include "facade/reply_capture.h"
 
 namespace facade {
 
 using namespace std;
-
-namespace {
-
-struct PayloadVisitor {
-  void operator()(monostate) {
-  }
-
-  void operator()(long v) {
-    rb->SendLong(v);
-  }
-
-  void operator()(double v) {
-    LOG(FATAL) << "Double replies are not supported in MC protocol";
-  }
-
-  void operator()(const payload::SimpleString& ss) {
-    rb->SendSimpleString(ss);
-  }
-
-  void operator()(const payload::BulkString& bs) {
-    LOG(FATAL) << "TBD";
-  }
-
-  void operator()(payload::Null) {
-    LOG(FATAL) << "TBD";
-  }
-
-  void operator()(const payload::Error& err) {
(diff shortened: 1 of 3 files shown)


src/facade/parsed_command.cc
```diff
@@ -0,0 +1,167 @@
...
+    rb->SendLong(v);
+  }
+
+  void operator()(double v) {
+    LOG(FATAL) << "Double replies are not supported in MC protocol";
+  }
+
+  void operator()(const payload::SimpleString& ss) {
+    rb->SendSimpleString(ss);
+  }
```
dranikpg: You are reinventing something here, there is `CapturingReplyBuilder::Apply(Payload&& pl, RedisReplyBuilder* rb)`

src/facade/parsed_command.cc
```diff
@@ -0,0 +1,167 @@
...
+}
+
+void ParsedCommand::SendSimpleString(std::string_view str) {
+  if (reply_direct_) {
+    rb_->SendSimpleString(str);
+  } else {
+    reply_payload_ = payload::SimpleString{std::string(str)};
+    NotifyReplied();
+  }
+}
```
dranikpg, line 78: same here, this needs to be reviewed

romange: what do you mean by same here?

src/facade/parsed_command.cc
romange: I do not use `RedisReplyBuilder` here, I use MCReplyBuilder.
I can try and relax the interface in `CapturingReplyBuilder::Apply` and see if it works

src/facade/parsed_command.cc
romange: what needs to be reviewed

src/facade/parsed_command.cc
romange: @dranikpg  adjusted `CapturingReplyBuilder` to use SinkReplyBuilder but I think it's more fragile as some of its methofs are resp specific

romange: I agee I will have to solve this problem of differrent protocols sooner or later - we need to have a payload that covers the super set of concepts from both protocols

dranikpg (approved):

PR #6255, opened 2025-12-19
chore(server): Remove more family header definitions

dranikpg:
Remove more header command definitions


dranikpg:
Next step is to get rid of headers completely

PR #6258, opened 2025-12-20
chore: an issue of destruction of in-flight parsed commands

romange:
Also, make ParsedCommand polymorthic to be able to deallocate it easier.
In addition, expose GetSize method that returns the real object size for
ParsedCommand derivatives.

Finally, rename async_reply to is_deferred_reply_ and  reply_direct_ to allow_async_execution_, which imho has more clarity regarding their meanings.

### DIFF
--- a/src/facade/parsed_command.h
+++ b/src/facade/parsed_command.h
@@ -32,6 +32,12 @@ class ParsedCommand : public cmn::BackedArguments {
   ParsedCommand() = default;
 
  public:
+  virtual ~ParsedCommand() = default;
+
+  virtual size_t GetSize() const {
+    return sizeof(ParsedCommand);
+  }
+
   // time when the message was parsed as reported by CycleClock::Now()
   uint64_t parsed_cycle = 0;
   ParsedCommand* next = nullptr;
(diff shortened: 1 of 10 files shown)


romange:
Sorry, I updated the PR, PTAL - I renamed some variables in ParsedCommand as before their names were a bit confusing

src/facade/parsed_command.h
```diff
@@ -123,22 +139,24 @@ class ParsedCommand : public cmn::BackedArguments {
...
+    DELETE_INTENT = 1 << 2,
   };
   std::atomic_uint8_t state_{0};
 
-  // whether the reply should be sent directly, or captured for later sending
-  bool reply_direct_ = true;
-
   // whether the command can be dispatched asynchronously.
-  bool dispatch_async_ = false;
+  bool allow_async_execution_ = false;
```
dranikpg, line 147: allow... 🤔  Its not like you have a choice, once it started being async, it keeps so until everyhing finished, no?

src/facade/parsed_command.h
```diff
@@ -58,31 +64,30 @@ class ParsedCommand : public cmn::BackedArguments {
...
+  bool AsyncExecutionAllowed() const {
+    return allow_async_execution_;
   }
 
-  void set_dispatch_async(bool async) {
-    dispatch_async_ = async;
+  // Marks this command as executed asynchronously with deferred reply.
+  // Precondition: AsyncExecutionAllowed() must be true.
+  void SetDeferredReply() {
+    is_deferred_reply_ = true;
```
dranikpg, line 86: Is a command ever async and not directly reply?

src/facade/parsed_command.h
romange: Yes, I need to allow selective async execution per command.
For example, I am working to run memcache SET asynchronously, so inside DispatchMC we ill call AllowAsyncExecution. It does not mean that it will necessary run asynchronously. for example, if we have parse error then we reply directly.

so to summarise, deferred reply implies that async execution was enabled but enabled async execution does not necessary mean there will be deferred  reply @dranikpg

src/facade/parsed_command.h
romange: see above. allow_async_execution_ only allows. 
`is_deferred_reply_` can still be false at the end even if `allow_async_execution_` was true.

PR #6259, opened 2025-12-20
chore: simplify memcache_parser_test

romange:
1. Optimize the code, extend interfaces around BackedArguments which allows remove redundant tmp_args_.
2. Split the main parsing function to two smaller functions.

### DIFF
--- a/src/common/backed_args.h
+++ b/src/common/backed_args.h
@@ -139,6 +139,7 @@ class BackedArguments {
   }
 
   std::string_view back() const {
+    assert(size() > 0);
     return at(size() - 1);
   }
 


src/common/backed_args.h
```diff
@@ -138,13 +138,31 @@ class BackedArguments {
...
   // Reserves space for additional argument of given length at the end.
   void PushArg(size_t len) {
     size_t old_size = storage_.size();
     offsets_.push_back(old_size);
     storage_.resize(old_size + len + 1);
   }
 
+  void PushArg(std::string_view arg) {
+    PushArg(arg.size());
+    char* dest = storage_.data() + offsets_.back();
```
vyavdoshenko, line 155: what if offsets_ is empty?

src/common/backed_args.h
```diff
@@ -138,13 +138,31 @@ class BackedArguments {
     storage_.clear();
   }
 
+  std::string_view back() const {
```
vyavdoshenko, line 141: This method looks dangerous. I suggest add a condition and return an empty string if container is empty or add DCHECK at least

src/facade/memcache_parser.h
```diff
@@ -123,7 +123,7 @@ class MemcacheParser {
   }
 
   size_t UsedMemory() const {
-    return tmp_args_.capacity() * sizeof(std::string_view);
+    return 0;
```
vyavdoshenko, line 126: It looks strange. Is it correct change?

src/common/backed_args.h
romange: @vyavdoshenko  it can not be empty as we call `PushArg(arg.size());` at line 153

src/facade/memcache_parser.h
romange: yes, we got rid of the only heap usage we had. The next pr will change it to something else

src/common/backed_args.h
romange: done

PR #6260, opened 2025-12-20
chore: MemcacheParser now consumes all the input.

romange:
Before, the parser did not consume the input when it returned INPUT_PENDING. Now it caches the partial input internally and consumes everything.

### DIFF
--- a/src/facade/memcache_parser.h
+++ b/src/facade/memcache_parser.h
@@ -123,11 +123,12 @@ class MemcacheParser {
   }
 
   size_t UsedMemory() const {
-    return 0;
+    return tmp_buf_.capacity();
   }
 
   void Reset() {
     val_len_to_read_ = 0;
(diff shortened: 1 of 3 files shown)


src/facade/memcache_parser.h
```diff
@@ -123,11 +123,12 @@ class MemcacheParser {
   }
 
   size_t UsedMemory() const {
-    return 0;
+    return tmp_buf_.capacity();
   }
 
   void Reset() {
     val_len_to_read_ = 0;
+    tmp_buf_.clear();
```
vyavdoshenko, line 131: clear() does't change the capacity. Consider `tmp_buf_.shrink_to_fit()`. This can adjust the memory after big commands.

romange:
The clear is for correctness, I would like to reuse the buffer for next
parsing operations
Roman Gershman
CTO
---

PR #6262, opened 2025-12-22
chore(server): Use cycle based quota for defragmentation

abhijat:
* Switches defragmentation limits in engine shard defrag task from iteration count to cycle based quota, capped to 150 microseconds per run. This change will enable adding index defragmentation as a next step which works on cycle based quota.
* The quota class is extracted from page usage class.

The limit of 150 usec per defragmentation task was arrived at from a manual test in which the fragmentation check was forced to always return true, so every checked object is reallocated. Then the db is populated with 

```
  Run({"DEBUG", "POPULATE", "10000", "key", "256"});
```

Defragmentation (`shard->DoDefrag(CollectPageStats::NO, 0)`) is then run several times to calculate average run per task. On average one `DoDefrag` run with the limit of 40 iterations took around 600 usec, in practice it would be lower because not all objects will be reallocated.

In the next PR when defragmentation for search indices is added the quota will need to be split between compact objects and search indices.

### DIFF
--- a/src/core/page_usage/page_usage_stats.cc
+++ b/src/core/page_usage/page_usage_stats.cc
@@ -43,8 +43,7 @@ HllBufferPtr InitHllPtr() {
 }  // namespace
 
 CycleQuota::CycleQuota(const uint64_t quota_usec)
-    : quota_cycles_{quota_usec == kMaxQuota ? kMaxQuota : base::CycleClock::FromUsec(quota_usec)} {
-  Arm();
+    : CycleQuota(base::CycleClock::FromUsec(quota_usec), true) {
 }
 
 void CycleQuota::Arm() {
@@ -59,6 +58,14 @@ bool CycleQuota::Depleted() const {
 
 uint64_t CycleQuota::UsedCycles() const {
   return base::CycleClock::Now() - start_cycles_;
+}
+
+CycleQuota CycleQuota::Unlimited() {
+  return CycleQuota(kMaxQuota, true);
+}
+
+CycleQuota::CycleQuota(const uint64_t quota_cycles, bool /*tag*/) : quota_cycles_{quota_cycles} {
+  Arm();
 }
 
 void CollectedPageStats::Merge(CollectedPageStats&& other, uint16_t shard_id) {
--- a/src/core/page_usage/page_usage_stats.h
+++ b/src/core/page_usage/page_usage_stats.h
@@ -32,14 +32,14 @@ class CycleQuota {
 
   uint64_t UsedCycles() const;
 
+  static CycleQuota Unlimited();
+
  private:
+  explicit CycleQuota(uint64_t quota_cycles, bool /*tag*/);
+
   uint64_t quota_cycles_;
   uint64_t start_cycles_{0};
 };
-
-inline CycleQuota UnlimitedQuota() {
-  return CycleQuota(CycleQuota::kMaxQuota);
-}
 
 enum class CollectPageStats : uint8_t { YES, NO };
 
@@ -66,7 +66,8 @@ struct CollectedPageStats {
 
 class PageUsage {
  public:
-  PageUsage(CollectPageStats collect_stats, float threshold, CycleQuota quota = UnlimitedQuota());
+  PageUsage(CollectPageStats collect_stats, float threshold,
+            CycleQuota quota = CycleQuota::Unlimited());
 
   virtual ~PageUsage() = default;
 


src/server/dragonfly_test.cc
```diff
@@ -824,6 +824,39 @@ TEST_F(DefragDflyEngineTest, TestDefragOption) {
...
+      PageUsage page_usage{CollectPageStats::NO, 0, CycleQuota{CycleQuota::kDefaultDefragQuota}};
+      page_usage.SetForceReallocate(true);
+
+      shard->DoDefrag(page_usage);
+      cursor_states.push_back(shard->GetDefragCursor());
+      if (cursor_states.back() == 0)
+        break;
+    }
+
+    EXPECT_TRUE(cursor_states.size() > 1 && cursor_states.size() < max_attempts);
```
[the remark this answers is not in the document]
abhijat: changed this assertion a little to make it clearer

abhijat:
augment review

src/server/dragonfly_test.cc
```diff
@@ -824,6 +824,40 @@ TEST_F(DefragDflyEngineTest, TestDefragOption) {
...
+  Run({"DEBUG", "POPULATE", "5000", "key", "256"});
+
+  // Try to run defrag at least this many times and stop early if cursor reaches the end (winds
+  // back to 0)
+  constexpr auto max_attempts = 100;
+
+  std::vector<uint64_t> cursor_states;
+  cursor_states.reserve(max_attempts);
+
+  shard_set->pool()->AwaitFiberOnAll([&](unsigned, ProactorBase*) {
```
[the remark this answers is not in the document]
abhijat: In this case we have a single thread, so I moved the assertions out of the lambda to make it read a little clearer.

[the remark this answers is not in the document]
abhijat: changed

abhijat:
augment review

src/core/compact_object_test.cc
```diff
@@ -63,7 +63,7 @@ std::vector<void*> AllocateForTest(int size, std::size_t allocate_size, int fact
 }
 
 bool HasUnderutilizedMemory(const std::vector<void*>& ptrs, float ratio) {
-  PageUsage page_usage{CollectPageStats::NO, ratio};
+  PageUsage page_usage{CollectPageStats::NO, ratio, UnlimitedQuota()};
```
dranikpg: you could've had this last parameter with a default value of unlimited quota

src/core/page_usage/page_usage_stats.cc
```diff
@@ -42,6 +42,25 @@ HllBufferPtr InitHllPtr() {
 
 }  // namespace
 
+CycleQuota::CycleQuota(const uint64_t quota_usec)
+    : quota_cycles_{quota_usec == kMaxQuota ? kMaxQuota : base::CycleClock::FromUsec(quota_usec)} {
+  Arm();
+}
```
dranikpg, line 47: using a constructor with a sentinel value (or just default) would've dont it as well

dranikpg: or optional

src/core/page_usage/page_usage_stats.cc
```diff
@@ -42,6 +42,25 @@ HllBufferPtr InitHllPtr() {
...
+
+void CycleQuota::Arm() {
+  start_cycles_ = base::CycleClock::Now();
+}
+
+bool CycleQuota::Depleted() const {
+  if (quota_cycles_ == kMaxQuota)
+    return false;
+  return base::CycleClock::Now() - start_cycles_ >= quota_cycles_;
```
dranikpg: return UsedCycles() >= quota

src/server/engine_shard.h
```diff
@@ -206,11 +206,13 @@ class EngineShard {
...
   void FinalizeMulti(Transaction* tx);
 
-  // Scan the shard with the cursor and apply defragmentation for database entries. An optional
-  // threshold can be passed, which will be used to determine if defragmentation should be
-  // performed.
-  // Returns true if defragmentation was performed.
-  std::optional<CollectedPageStats> DoDefrag(CollectPageStats collect_page_stats, float threshold);
+  // Scan the shard with the cursor and apply defragmentation for database entries.
+  // Returns collected page stats if defragmentation was performed.
+  std::optional<CollectedPageStats> DoDefrag(PageUsage& page_usage);
```
dranikpg: Please take it by pointer (out styleguide)

dranikpg:
150 is a good value, likely there won't be any need to lower it. If this task runs as an on idle task (without sleeps), it should be also enough and there won't be any need to increase it

src/server/engine_shard.h
abhijat: fixed

src/core/page_usage/page_usage_stats.cc
abhijat: changed

src/core/page_usage/page_usage_stats.cc
abhijat: This constructor was a bit messy, as it mixes usec and cycles because of the same underlying type, I cleaned up the constructors a bit so one takes a usec quota and the other one takes cycles, hopefully its a little cleaner now although there is more code.

src/core/compact_object_test.cc
abhijat: changed

abhijat:
augment review

PR #6267, opened 2025-12-22
chore: move memcache reply rendering into MCRender

romange:
We need to store deferred replies in the reply_payload, so we need to reuse the same logic of rendering replies inside ParsedCommand. In order to preserve the separation of responsibilities we introduce a helper class MCRender that helps rendering replies for memcache commands.
The class is being used for writing replies directly into socket via ReplyBuilder, or for caching them in reply_payload.

### DIFF
--- a/src/facade/parsed_command.cc
+++ b/src/facade/parsed_command.cc
@@ -14,25 +14,25 @@ namespace facade {
 
 using namespace std;
 
-string MCRender::SendNotFound() const {
+string MCRender::RenderNotFound() const {
   if (flags_.no_reply)
     return {};
   return flags_.meta ? "NF" : "NOT_FOUND";
 }
 
-string MCRender::SendGetEnd() const {
+string MCRender::RenderGetEnd() const {
   if (flags_.no_reply || flags_.meta)
     return {};
   return "END";
 }
 
-string MCRender::SendMiss() const {
+string MCRender::RenderMiss() const {
   if (flags_.no_reply || !flags_.meta)
     return {};
   return "EN";
 }
 
-string MCRender::SendDeleted() const {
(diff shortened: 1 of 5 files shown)


src/facade/reply_payload.h
```diff
@@ -23,8 +23,10 @@ struct StoredReply {
   bool ok;  // true for SendStored, false for SendSetSkipped
 };
 
+struct NoReply {};  // for no reply mode.
```
dranikpg: why not use monostate which is already there?

src/facade/parsed_command.cc
```diff
@@ -71,9 +95,10 @@ void ParsedCommand::SendStored(bool ok) {
 
 void ParsedCommand::SendSimpleString(std::string_view str) {
   if (!is_deferred_reply_) {
-    rb_->SendSimpleString(str);
+    if (!str.empty())  // empty string means no-reply
```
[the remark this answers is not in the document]
dranikpg: He has a point

src/facade/reply_payload.h
romange: i thought about this and decided against it as I can not distinguish between "reply not filled" vs "reply not needed"

src/facade/parsed_command.cc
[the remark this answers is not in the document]
romange: I've searched for all `SendSimpleString` calls and I did not find any calls with possibly empty strings. 
In general `SendSimpleString` is used to send a plain, ascii status or value.

src/facade/reply_payload.h
romange: ok, thought it a bit more and decided to use monostate after all. 

PTAL

PR #6272, opened 2025-12-22
Add alloc size to OAHSet

BorysTheDev:
fixes: #6249

- Added allocated memory calculation
- fixed a couple of bugs in OAHSet (incorrect shifting and incorrect size)
- make StringSet benchmarks more relevant about memory consumption

### DIFF
--- a/src/core/oah_set.h
+++ b/src/core/oah_set.h
@@ -243,6 +243,12 @@ class OAHSet {  // Open Addressing Hash Set
         assert(!res.IsVector());
         --size_;
         obj_alloc_used_ -= res.AllocSize();
+        if (bucket.IsVector()) {
+          if (bucket.AsVector().Empty()) {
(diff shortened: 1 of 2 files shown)


src/core/oah_set.h
```diff
@@ -326,6 +345,13 @@ class OAHSet {  // Open Addressing Hash Set
     size_t prev_size = 1ul << prev_capacity_log;
     // we should prevent moving elements before current possition to avoid double processing
     constexpr size_t mix_size = 2 << kShiftLog;
+
```
dranikpg, line 348: probably was meant to be max_size

dranikpg (approved):

src/core/oah_set.h
BorysTheDev: no, it is mix_size, because the elements that were already moved can be mixed with the elements that aren't yet processed. It is a dangerous zone

PR #6273, opened 2025-12-22
chore: get rid of command specific flags in MCReplyBuilder

romange:
Before - we passed memcache command specific flags to MCReplyBuilder to remember the response configuration. However, this does not work with asynchronous dispatch, as a command can be replied at a differrent time from when these flags were set. We need to use flags kept inside ParsedCommand to have consistent behavior.

Also, we can not render memcache replies via CapturingReplyBuilder as it is stateless with respect to these flags, hence we move rendering of StoredReply response to ParsedCommand.

The other response that can not be rendered using CapturingReplyBuilder visitor is MGET response, as it requires keys and memcache specific flags to render the response correctly. This logic will be added in the subsequent PR.

### DIFF
--- a/src/facade/parsed_command.cc
+++ b/src/facade/parsed_command.cc
@@ -14,6 +14,15 @@ namespace facade {
 
 using namespace std;
 
+namespace {
+
+inline void SendSimpleDirect(std::string_view str, SinkReplyBuilder* rb) {
+  if (!str.empty())  // empty string means no-reply
+    rb->SendSimpleString(str);
+}
+
+}  // namespace
+
 string MCRender::RenderNotFound() const {
   if (flags_.no_reply)
     return {};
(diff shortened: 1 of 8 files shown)


src/facade/parsed_command.cc
```diff
@@ -105,7 +122,11 @@ void ParsedCommand::SendSimpleString(std::string_view str) {
 
 bool ParsedCommand::SendPayload() {
   if (is_deferred_reply_) {
-    CapturingReplyBuilder::Apply(std::move(reply_payload_), rb_);
+    if (const auto* stored = get_if<payload::StoredReply>(&reply_payload_); stored) {
```
romange, line 125: I have to exempt some types here because we need additional context to render them properly for MC @dranikpg .

This will be true for all commands under MC.

Issue #6235, opened 2025-12-17
Improve defragmentation of json


abhijat:
Current approach does a serialize + deserialize round trip. We should instead do a tree traversal and only change the nodes we need to change.

PR #6276, opened 2025-12-23
chore(json): Use new json defragmentation

abhijat:
A new flag is introduced:

`disable_json_defragmentation` - false by default. if true, it will skip defragmentation of all JSON objects

The defragmentation process now use tree walk instead of string conversion. 

Some helper functions which are no longer needed are cleaned up.

FIXES https://github.com/dragonflydb/dragonfly/issues/6235

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -1617,33 +1617,21 @@ MemoryResource* CompactObj::memory_resource() {
 }
 
 bool CompactObj::JsonConsT::DefragIfNeeded(PageUsage* page_usage) {
-  if (ShouldDefragment(page_usage)) {
-    const MiMemoryResource* mr = static_cast<MiMemoryResource*>(memory_resource());
+  const MiMemoryResource* mr = static_cast<MiMemoryResource*>(memory_resource());
 
-    const int64_t before = static_cast<int64_t>(mr->used());
-    DCHECK_GE(before, 0) << "Memory usage is more than int64_t max value";
+  const int64_t before = static_cast<int64_t>(mr->used());
+  DCHECK_GE(before, 0) << "Memory usage is more than int64_t max value";
 
-    Defragment(*json_ptr, page_usage);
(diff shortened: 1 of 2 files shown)


src/core/compact_object.cc
```diff
@@ -1610,35 +1617,21 @@ MemoryResource* CompactObj::memory_resource() {
 }
 
 bool CompactObj::JsonConsT::DefragIfNeeded(PageUsage* page_usage) {
-  if (ShouldDefragment(page_usage)) {
```
abhijat, line 1613: this check is no longer needed as we check every node while visiting it.

src/core/json/json_object.cc
```diff
@@ -150,21 +147,21 @@ void Defragment(JsonType& j, PageUsage* page_usage) {
...
+        did_defragment |= DefragmentJsonObject(*current, page_usage);
         auto& object = current->cast<JsonType::object_storage>().value();
         for (auto& member : object) {
           stack.push(&member.value());
         }
         break;
       }
       case json_storage_kind::array: {
-        DefragmentJsonArray(*current, page_usage);
+        did_defragment |= DefragmentJsonArray(*current, page_usage);
```
abhijat, line 164: As a future optimization, if we did not defragment, we can avoid the next cast, both here and in object storage. Since we did not defragment, the underlying storage is unchanged and safe to access, so we can reuse it.

If in the majority of cases there is no defragmentation, this saves significant time spent casting.

src/core/compact_object.cc
```diff
@@ -1087,6 +1088,9 @@ string_view CompactObj::GetSlice(string* scratch) const {
 }
 
 bool CompactObj::DefragIfNeeded(PageUsage* page_usage) {
+  static const bool disable_json_defragmentation =
```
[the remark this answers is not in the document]
abhijat: this is an acceptable and frequently used pattern

PR #6278, opened 2025-12-23
chore: swap ParsedEntry::CmdData with BackedArguments

romange:
Before - we used a custom implementation for buffer backed arguments.
Now - we switch to standard cmn::BackedArguments, which provides all the functionality
out of the box. Bonus points - change the interfaces to be able to reuse already existing objects
that allocate heap memory.

### DIFF
--- a/src/server/journal/executor.cc
+++ b/src/server/journal/executor.cc
@@ -24,19 +24,11 @@ namespace {
 template <typename... Ts> journal::ParsedEntry::CmdData BuildFromParts(Ts... parts) {
   vector<string> raw_parts{absl::StrCat(std::forward<Ts>(parts))...};
 
-  auto cmd_str = accumulate(raw_parts.begin(), raw_parts.end(), std::string{});
-  auto buf = make_unique<uint8_t[]>(cmd_str.size());
-  memcpy(buf.get(), cmd_str.data(), cmd_str.size());
-
-  CmdArgVec slice_parts;
-  size_t start = 0;
-  for (const auto& part : raw_parts) {
-    slice_parts.emplace_back(reinterpret_cast<char*>(buf.get()) + start, part.size());
-    start += part.size();
-  }
-
-  return {std::move(buf), std::move(slice_parts), cmd_str.size()};
+  journal::ParsedEntry::CmdData res;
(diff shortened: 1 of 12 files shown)


romange:
augment review

src/server/journal/executor.cc
```diff
@@ -24,19 +24,11 @@ namespace {
...
-  size_t start = 0;
-  for (const auto& part : raw_parts) {
-    slice_parts.emplace_back(reinterpret_cast<char*>(buf.get()) + start, part.size());
-    start += part.size();
-  }
-
-  return {std::move(buf), std::move(slice_parts), cmd_str.size()};
+  journal::ParsedEntry::CmdData res;
+  res.Assign(raw_parts.begin(), raw_parts.end(), raw_parts.size());
+  return res;
```
BorysTheDev, line 29: I like how short it is now

src/server/journal/tx_executor.cc
```diff
@@ -68,52 +68,50 @@ void TransactionData::AddEntry(journal::ParsedEntry&& entry) {
...
-  if (absl::EqualsIgnoreCase(ToSV(args[0]), "FLUSHDB"sv) ||
-      absl::EqualsIgnoreCase(ToSV(args[0]), "FLUSHALL"sv) ||
-      (absl::EqualsIgnoreCase(ToSV(args[0]), "DFLYCLUSTER"sv) &&
-       absl::EqualsIgnoreCase(ToSV(args[1]), "FLUSHSLOTS"sv))) {
+  string_view front = command.Front();
+
+  if (absl::EqualsIgnoreCase(front, "FLUSHDB"sv) || absl::EqualsIgnoreCase(front, "FLUSHALL"sv))
+    return true;
+
+  if (command.size() > 1 && absl::EqualsIgnoreCase(front, "DFLYCLUSTER"sv) &&
```
BorysTheDev, line 80: you don't need "command.size() > 1" check, because DFLYCLUSTER always come with at least one more argument

romange: when when the command is parsed and validated - yes, but here it comes from an external source and maybe it is  a security hole

PR #6282, opened 2025-12-24
chore: deferred mget

romange:
This does not implement async dispatching for MC/mget, just merely supports
replies in the correct order for cases where the previous command was dispatched
asynchronously. We meed to implement it for every MC command to preserve the correctness.

### DIFF
--- a/src/facade/parsed_command.cc
+++ b/src/facade/parsed_command.cc
@@ -120,10 +120,75 @@ void ParsedCommand::SendSimpleString(std::string_view str) {
   }
 }
 
+void ParsedCommand::SendMiss(unsigned index) {
+  if (is_deferred_reply_) {
+    if (holds_alternative<std::monostate>(reply_payload_)) {
+      // Lazily allocate the array on first use. Already initialized to not found,
+      // so we do not need to do anything else here.
+      auto arr = std::make_unique<facade::payload::SingleGetReply[]>(size());
+      reply_payload_ = std::move(arr);
+      // We do not notify here as there will be more entries to fill.
+    }
+  } else {
+    // Direct reply for either MC or Redis.
+    if (mc_cmd_ != nullptr)
+      rb_->SendSimpleString(MCRender{mc_cmd_->cmd_flags}.RenderMiss());
+    else
+      static_cast<RedisReplyBuilder*>(rb_)->SendNull();
+  }
+}
+
+void ParsedCommand::SendValue(unsigned index, std::string_view value, uint64_t mc_ver,
+                              uint32_t mc_flag) {
+  if (is_deferred_reply_) {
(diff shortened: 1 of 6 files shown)


src/facade/parsed_command.cc
```diff
@@ -120,10 +120,75 @@ void ParsedCommand::SendSimpleString(std::string_view str) {
...
+}
+
+void ParsedCommand::SendValue(unsigned index, std::string_view value, uint64_t mc_ver,
+                              uint32_t mc_flag) {
+  if (is_deferred_reply_) {
+    if (holds_alternative<std::monostate>(reply_payload_)) {
+      // Lazily allocate the array on first use.
+      auto arr = std::make_unique<facade::payload::SingleGetReply[]>(size());
+      reply_payload_ = std::move(arr);
+    }
```
dranikpg, line 148: lazily allocate and used from multiple threads don't go in a single sentence

src/facade/parsed_command.cc
```diff
@@ -174,15 +239,44 @@ void ParsedCommand::NotifyReplied() {
...
   }
 }
 
+void ParsedCommand::SendDirect(const payload::MGetReply& pl) {
+  RedisReplyBuilder* rrb = mc_cmd_ ? nullptr : static_cast<RedisReplyBuilder*>(rb_);
+  if (rrb) {
+    rrb->StartArray(size());
+  }
+
+  for (size_t i = 0; i < size(); ++i) {
```
dranikpg, line 259: sink reply builder needs a scope to avoid copies

src/facade/reply_payload.h
```diff
@@ -20,11 +20,22 @@ struct CollectionPayload;
...
-  bool ok;  // true for SendStored, false for SendSetSkipped
+  bool ok = false;  // true for SendStored, false for SendSetSkipped
 };
 
+struct SingleGetReply {
+  std::string value;
+  uint64_t mc_ver = 0;
+  uint32_t mc_flag = 0;
+  bool is_found = false;
+};
```
dranikpg, line 31: you need to rewrite mget then, because it allocates a buffer per thread

dranikpg: it will be a double copy

src/facade/parsed_command.cc
dranikpg: about the future: and what if it's not size()? what if its from a hmget reply? You've cut yourself off of collection information and now you can't get it from anywhere... and now you need a protocol on top?

src/facade/parsed_command.cc
dranikpg: I see, its because you cut yourself off from StartArray infromation

dranikpg (commented):
You are creating your own mixed protocol that if/else hotfixes the magic beneath

PR #6283, opened 2025-12-24
chore: change DispatchCommand interface

romange:
This PR refactors the DispatchCommand interface to consolidate command execution context. The main change simplifies the interface from taking separate SinkReplyBuilder* and ConnectionContext* parameters to accepting a single ParsedCommand* parameter that encapsulates both.

The DispatchCommand method signature is updated throughout the codebase to use ParsedCommand* instead of separate builder and context parameters
A CommandContext class (which extends ParsedCommand) is now used to carry command state, removing the temporary cmnd_ctx field from ConnectionContext
Helper functions like VecToArgList are refactored to populate BackedArguments directly,

### DIFF
--- a/src/common/backed_args.h
+++ b/src/common/backed_args.h
@@ -93,6 +93,11 @@ class BackedArguments {
     return s1 + s2;
   }
 
+  void SwapArgs(cmn::BackedArguments& other) {
+    offsets_.swap(other.offsets_);
+    storage_.swap(other.storage_);
+  }
+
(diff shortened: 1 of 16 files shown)


dranikpg (approved):

dranikpg:
If it wasn't for the facade division, we could split all the dispatch/verify/etc methods to be on a ParsedCommand object instead of generic service based ones (but we also need access to global state 🤔), either way it's just a vision

PR #6285, opened 2025-12-25
feat(search): Use stateless allocator for strings

abhijat:
The strings in sort indices are changed to use stateless allocator instead of polymorphic one. The same change is also made to vectors. 

The stateless allocator benefits from empty base optimization, no space needs to be reserved for it within strings because it is an empty class. The stl classes use a member called `_Alloc_hider` or similar which inherits from the provided allocator:

https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/bits/basic_string.h#L196-L197
https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/bits/stl_vector.h#L139-L140

Since our allocator has no state, it occupies no space inside this hider.

A static assert is added to confirm this, another one is added to make sure the stateless allocator remains stateless.

FIXES https://github.com/dragonflydb/dragonfly/issues/2313

### DIFF
--- a/src/core/detail/stateless_allocator.h
+++ b/src/core/detail/stateless_allocator.h
@@ -13,19 +13,12 @@ namespace detail {
 inline thread_local PMR_NS::memory_resource* tl_mr = nullptr;
 }
 
-template <typename T> class StatelessAllocator {
+template <typename T, typename Impl> class StatelessAllocatorBase {
  public:
   using value_type = T;
   using size_type = std::size_t;
   using difference_type = std::ptrdiff_t;
   using is_always_equal = std::true_type;
-
-  template <typename U> StatelessAllocator(const StatelessAllocator<U>&) noexcept {  // NOLINT
-  }
-
-  StatelessAllocator() noexcept {
-    assert(detail::tl_mr != nullptr);
-  };
 
   template <typename U, typename... _Args> void construct(U* __p, _Args&&... __args) {
     ::new (static_cast<void*>(__p)) U(std::forward<_Args>(__args)...);
@@ -33,15 +26,26 @@ template <typename T> class StatelessAllocator {
 
   static value_type* allocate(size_type n) {
     static_assert(
-        std::is_empty_v<StatelessAllocator>,
+        std::is_empty_v<Impl>,
         "StatelessAllocator must not contain state, so it can use empty base optimization");
 
-    void* ptr = detail::tl_mr->allocate(n * sizeof(value_type), alignof(value_type));
+    void* ptr = Impl::resource()->allocate(n * sizeof(value_type), alignof(value_type));
     return static_cast<value_type*>(ptr);
   }
 
   static void deallocate(value_type* ptr, size_type n) noexcept {
-    detail::tl_mr->deallocate(ptr, n * sizeof(value_type), alignof(value_type));
+    Impl::resource()->deallocate(ptr, n * sizeof(value_type), alignof(value_type));
+  }
+};
+
+template <typename T>
+class StatelessAllocator : public StatelessAllocatorBase<T, StatelessAllocator<T>> {
+ public:
+  StatelessAllocator() noexcept {
+    assert(detail::tl_mr != nullptr);
+  }
+
+  template <typename U> StatelessAllocator(const StatelessAllocator<U>&) noexcept {  // NOLINT
   }
 
   static PMR_NS::memory_resource* resource() {
--- a/src/core/search/stateless_allocator.h
+++ b/src/core/search/stateless_allocator.h
@@ -6,6 +6,7 @@
 #include <cassert>
 
 #include "base/pmr/memory_resource.h"
+#include "core/detail/stateless_allocator.h"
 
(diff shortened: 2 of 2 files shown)


src/core/search/sort_indices.h
```diff
@@ -4,23 +4,18 @@
...
-#include "core/search/compressed_sorted_set.h"
 
 namespace dfly::search {
 
-template <typename T> struct SimpleValueSortIndex : public BaseSortIndex {
+using StatelessString = std::basic_string<char, std::char_traits<char>, StatelessAllocator<char>>;
+static_assert(sizeof(StatelessString) == sizeof(std::string));
+
+template <typename T> using StatelessVector = std::vector<T, StatelessAllocator<T>>;
+static_assert(sizeof(StatelessVector<StatelessString>) == sizeof(std::vector<std::string>));
```
abhijat: I can't find any guarantees that EBO is always supported. These checks might fail in some esoteric setups. IMO it is better to keep these and investigate the build failure rather than silently have the program use much more RAM.

abhijat:
This approach actually breaks the `used` counter in the resource used for search indices. `ShardDocIndices` has its own `local_mr_` and thus its own used bytes tracking.

If we switch to stateless allocator specifically for sorted index, then we end up using the memory resource which that allocator uses, which is the one derived from engine shard `mi_resource_`. That has its own used bytes counter.

So the count is removed from the indices object and added to the engine shard resource which differs from previous behavior where the indices related usage contained sort indices used bytes.

abhijat:
> This approach actually breaks the `used` counter in the resource used for search indices. `ShardDocIndices` has its own `local_mr_` and thus its own used bytes tracking.
> 
> If we switch to stateless allocator specifically for sorted index, then we end up using the memory resource which that allocator uses, which is the one derived from engine shard `mi_resource_`. That has its own used bytes counter.
> 
> So the count is removed from the indices object and added to the engine shard resource which differs from previous behavior where the indices related usage contained sort indices used bytes.

I don't have a clean looking solution for this, for now I added a stateless search allocator, tied to its own memory resource. The shard local indices root object initializes this allocator with its own memory resource.

I think this pattern might be repeated where needed, one stateless allocator is only ever backed by one memory resource. It might be worth thinking of a cleaner way to express this.

ETA: crtp is a good fix for this, the derived classes can use it to make the base class call static method correctly

abhijat:
> I don't have a clean looking solution for this, for now I added a stateless search allocator, tied to its own memory resource. The shard local indices root object initializes this allocator with its own memory resource.

After investigating a bit more, this seems the only solution. The stateless allocator is statelesss only in name. In reality its type tightly coupled to exactly one static lifetime memory resource (eg `tl_mr`). One stateless allocator type cannot somehow use two resources selectively in different objects instantiated from that type. So one allocator type per static resource is necessary.

src/server/search/doc_index.cc
```diff
@@ -733,6 +734,7 @@ io::Result<StringVec, ErrorReply> ShardDocIndex::GetTagVals(string_view field) c
 }
 
 ShardDocIndices::ShardDocIndices() : local_mr_{ServerState::tlocal()->data_heap()} {
+  InitTLSearchMR(&local_mr_);
```
abhijat, line 737: I wanted to add a counterpart call in destructor, where this is set to nullptr, so the stateless allocator does not accidentally use address of a memory resource which has been destroyed. 

However this doesn't work. Setting this to nullptr causes the destruction of the members of `ShardDocIndices` to fail because they still need to deallocate. 

If it needs to be done, it should be done outside of `ShardDocIndices`.

src/core/detail/stateless_allocator.h
```diff
@@ -13,31 +13,39 @@ namespace detail {
...
 
   static value_type* allocate(size_type n) {
-    void* ptr = detail::tl_mr->allocate(n * sizeof(value_type), alignof(value_type));
+    static_assert(
+        std::is_empty_v<Impl>,
+        "StatelessAllocator must not contain state, so it can use empty base optimization");
+
+    void* ptr = Impl::resource()->allocate(n * sizeof(value_type), alignof(value_type));
     return static_cast<value_type*>(ptr);
   }
```
dranikpg, line 34: nice 👍🏻

src/core/search/stateless_allocator.h
```diff
@@ -0,0 +1,47 @@
...
+
+namespace dfly {
+
+namespace detail {
+inline thread_local PMR_NS::memory_resource* search_tl_mr = nullptr;
+}
+
+template <typename T>
+class StatelessSearchAllocator : public StatelessAllocatorBase<T, StatelessSearchAllocator<T>> {
+ public:
```
dranikpg, line 19: you could templatize it by a functor that returns a pointer to the thread local variable to avoid having seaprate structs

abhijat: yeah I saw some examples of similar approach calling it policy structs, I will explore this and create a follow up if it looks good.

PR #6286, opened 2025-12-25
chore: implement support for deferred mget

romange:
This does not implement async dispatching for MC/mget, just merely supports
replies in the correct order for cases where the previous command was dispatched
asynchronously. We need to implement it for every MC command to preserve the correctness.

While testing GET/GAT, I learned that GAT was broken for meta protocol so I added a few fixes
to be able to test it.

### DIFF
--- a/src/server/string_family.cc
+++ b/src/server/string_family.cc
@@ -1404,7 +1404,7 @@ void MGetGeneric(CmdArgList args, CommandContext* cmnd_cntx,
   }
 
   fb2::BlockingCounter tiering_bc{0};  // Count of pending tiered reads
-  AggregateError tiering_err;          // Fist tiering error
+  AggregateError tiering_err;          // First tiering error
 
   unique_ptr<MGetResponse[]> mget_resp(new MGetResponse[shard_set->size()]);
 


src/server/string_family.cc
```diff
@@ -1420,39 +1420,48 @@ void MGetGeneric(CmdArgList args, CommandContext* cmnd_cntx,
...
+
+  size_t arg_len = args.size();
+
+  unique_ptr<optional<GetResp>[]> mget_results(new optional<GetResp>[arg_len]);
+  ReorderShardResults(absl::MakeSpan(mget_resp.get(), shard_set->size()), cmnd_cntx->tx,
+                      absl::MakeSpan(mget_results.get(), arg_len));
+
+  if (cmnd_cntx->mc_command()) {
+    // We can not access `args` inside the lambda below if it runs asynchronously.
+    // Access to arguments must be done via cmnd_cntx.
```
dranikpg, line 1433: why can't you just capture it by value? if there's two ways to do a thing it should be simpler

dranikpg: you capture just the span, the backing storage for the strings is kept alive for the whole command lifecycle

dranikpg (approved):

src/server/string_family.cc
romange: because currently  `args` are not necessary span around `cmnd_cntx`.  That's something that we need to fix in the long run but for now args can reference other argument storages. We have many places where we invoke commands, so args that we pass there wrap different backed argument than those stored inside `cmnd_cntx`

romange: for MC we should have args wrapping `cmnd_cntx` but why should I capture both args and `cmnd_cntx` if I can capture only a single variable?

PR #6287, opened 2025-12-25
chore: remove Stored and Deleted interfaces from ParsedCommand

romange:
As disscussed internally, we remove protocol related logic from ParsedCommand and move it back to appropriate commands. This PR handles mc commands: INCR,DEL and SET,

Also, reformatting memcache pytests to be part of a single class.

### DIFF
--- a/src/facade/reply_payload.h
+++ b/src/facade/reply_payload.h
@@ -8,9 +8,13 @@
 #include <string>
 #include <variant>
 
+#include "base/function2.hpp"
 #include "facade/facade_types.h"
 
-namespace facade::payload {
+namespace facade {
+
+class SinkReplyBuilder;
+namespace payload {
 
 // SendError (msg, type)
 using Error = std::unique_ptr<std::pair<std::string, std::string>>;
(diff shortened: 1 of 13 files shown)


tests/dragonfly/pymemcached_test.py
```diff
@@ -12,119 +13,6 @@
 DEFAULT_ARGS = {"memcached_port": 11212, "proactor_threads": 4}
 
 
-# Generic basic tests
```
romange, line 15: there are no  changes in this file besides moving most tests under the same class to use the same process

src/facade/reply_payload.h
```diff
@@ -19,12 +23,16 @@ using Null = std::nullptr_t;  // SendNull or SendNullArray
...
 struct SimpleString : public std::string {};  // SendSimpleString
 struct BulkString : public std::string {};    // SendBulkString
-struct StoredReply {
-  bool ok;  // true for SendStored, false for SendSetSkipped
-};
+
+using ReplyFunction =
+    fu2::function_base<true /*owns*/, false /*non-copyable*/, fu2::capacity_fixed<16, 8>,
+                       false /* non-throwing*/, true /* strong exceptions guarantees*/,
+                       void(SinkReplyBuilder*)>;
```
dranikpg, line 30: CollectionPayload is quite large, maybe you can give it more stack space

src/server/string_family.cc
```diff
@@ -1023,7 +1024,13 @@ void CmdSet(CmdArgList args, CommandContext* cmnd_cntx) {
           GenericFamily::OpDel(tx->GetOpArgs(es), args, false);
           return OpStatus::OK;
         });
-        return cmnd_cntx->SendStored(true);
+        if (cmnd_cntx->mc_command() != nullptr) {
+          cmnd_cntx->SendSimpleString(
+              MCRender{cmnd_cntx->mc_command()->cmd_flags}.RenderStored(true));
+        } else {
+          cmnd_cntx->SendOk();
+        }
```
dranikpg, line 1032: Looks wird because there is no top-level memcache interface anymore 😓

src/facade/reply_payload.h
romange: we store std::unique_ptr<CollectionPayload> which is just 8 bytes

PR #6288, opened 2025-12-26
chore: simplify GAT command arguments

romange:
Before - Dragonfly GAT command used `<TS> key1 .. keyn ` format to pass expiry timestamp information further.
Now, we have memcache context as part of the ParsedCommand structure so we can access expire_ts directly.

This change moves the logic of parsing the correct timestamp to the parser and removes the first <TS> argument from GAT. This will help with consolidating GAT and MGET code down the road.

This change is possible as we do not need to replicate GAT as is, as it is translated to PEXPIREAT or DEL commands.

### DIFF
--- a/src/server/dragonfly_test.cc
+++ b/src/server/dragonfly_test.cc
@@ -386,16 +386,12 @@ TEST_F(DflyEngineTest, Memcache) {
   EXPECT_THAT(resp, ElementsAre("VALUE key 1 3 0", "bar", "VALUE key2 2 8 0", "bar2val2", "END"));
 
   EXPECT_THAT(RunMC(MP::SET, "foo", "bar"), ElementsAre("STORED"));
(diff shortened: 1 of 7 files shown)


src/server/dragonfly_test.cc
```diff
@@ -386,16 +386,12 @@ TEST_F(DflyEngineTest, Memcache) {
   EXPECT_THAT(resp, ElementsAre("VALUE key 1 3 0", "bar", "VALUE key2 2 8 0", "bar2val2", "END"));
 
   EXPECT_THAT(RunMC(MP::SET, "foo", "bar"), ElementsAre("STORED"));
-  // The value here is 1 month + 1 second, memcache treats an expiry of greater than 1 month as
```
romange, line 389: I removed this test case as now we always pass the absolute timestamp in this test

vyavdoshenko (approved):
lgtm

PR #5791, opened 2025-09-10
feat: Add eviction based on rss memory usage

BagritsevichStepan:
Pr includes two other PRs, my #4991 and @mkaruza #5775

PR #6289, opened 2025-12-26
chore: fix build error

romange:
Fix compile error in AWS specific code that was introduced by #6280 Also bring a check that was mistakenly deleted by #5791


romange:
regresssion tests running: https://github.com/dragonflydb/dragonfly/actions/runs/20520429837

Issue #6228, opened 2025-12-16
support Magento and Cm_Cache_Backend_Redis


romange:
1. see https://github.com/colinmollenhour/Cm_Cache_Backend_Redis/  (similar to django-cacheops)
2. https://github.com/magento/magento2/  that relies on it

## Motivation

it's quite popular framework from Adobe on php and it can not use cluster mode, so obviously Dragonfly can help there with scaling issues.

## Goal - to ensure we work seamlessly with magento.
I suspect that Cm_Cache_Backend_Redis requires some lua tunings. 
We should adjust Dragonfly SHAs library accordingly and also send a PR to the library 

Also, send a PR for README to both repos explaining how to setup with Dragonfly. 

### lua problems
It could be that Cm_Cache_Backend_Redis scripts can run in non-atomic mode. That something we need to evaluate internally and decide what adjustments needs to be done. I saw 
https://github.com/colinmollenhour/Cm_Cache_Backend_Redis/blob/ba4aed6768797e4f9c0556c6c7371629701c2956/Cm/Cache/Backend/Redis.php#L856  is where they upload the scripts, and they are awfully similar to django-cacheops.

### More steps
We should open an issue in our documentation repo to add tutorials on how to configure Dragonfly with magento and django-cacheops. https://www.dragonflydb.io/docs/integrations

vyavdoshenko:
https://github.com/colinmollenhour/Cm_Cache_Backend_Redis/pull/186

vyavdoshenko:
https://github.com/dragonflydb/documentation/pull/507

https://github.com/AdobeDocs/commerce-operations.en/pull/167

PR #6291, opened 2025-12-26
fix: CONFIG GET returns numeric bytes for memory configs

vyavdoshenko:
CONFIG GET for `maxmemory` value now returns raw numeric bytes instead of a human-readable format.

Before: `CONFIG GET maxmemory` -> `"23.96GiB"`
After:  `CONFIG GET maxmemory` -> `"25716744192"`

This matches Valkey behavior, where CONFIG GET returns numeric strings for the `maxmemory` value. CONFIG SET still accepts human-readable format (e.g., "1GB", "512MB").

This fix enables compatibility with `Cm_Cache_Backend_Redis`, which parses maxmemory as a numeric value:
https://github.com/colinmollenhour/Cm_Cache_Backend_Redis/

Relevant to #6228

This PR depends on: https://github.com/romange/helio/pull/523


src/server/config_registry.cc
```diff
@@ -77,7 +79,19 @@ absl::CommandLineFlag* ConfigRegistry::GetFlag(std::string_view config_name) {
 
 optional<string> ConfigRegistry::Get(string_view config_name) {
   absl::CommandLineFlag* flag = GetFlag(config_name);
-  return flag ? flag->CurrentValue() : optional<string>();
+  if (!flag) {
+    return nullopt;
+  }
+
+  // For MemoryBytesFlag, return numeric bytes for compatibility.
+  if (flag->IsOfType<facade::MemoryBytesFlag>()) {
```
romange, line 87: if you are already touching it, please alias `facade::MemoryBytesFlag` with `strings::MemoryBytesFlag` from strings/human_readable.h

vyavdoshenko: Fixed

PR #6302, opened 2025-12-30
feat(server): Allow SELECT inside global/non-atomic exec/eval

dranikpg:
Allow SELECT in global/non-atomic multi transactions


dranikpg:
I did it here in a different way than in the v1.34-branch, keeping verification code only to the select command implementation

PR #6304, opened 2025-12-31
chore: refresh release settings for dragonfly

romange:
1. Allow override key settings via environment
2. Flip WITH_SIMSIMD to ON by default.


romange:
passed at https://github.com/dragonflydb/dragonfly/actions/runs/20619815514

vyavdoshenko (approved):

Issue #6162, opened 2025-12-04
dragonfly crases during renaming


vyavdoshenko:
```
F20251204 14:18:36.832938 64394 generic_family.cc:419] Error during rename command, deserialize_status: 5 del_status: 0
*** Check failure stack trace: ***
    @     0x5749f2e50629  google::LogMessage::Fail()
    @     0x5749f2e5056f  google::LogMessage::SendToLog()
    @     0x5749f2e4fd22  google::LogMessage::Flush()
    @     0x5749f2e4fa9d  google::LogMessage::~LogMessage()
    @     0x5749f20e1c34  dfly::(anonymous namespace)::Renamer::FinalizeRename()
    @     0x5749f20e16ff  dfly::(anonymous namespace)::Renamer::Rename()
    @     0x5749f20e7950  dfly::(anonymous namespace)::RenameGeneric()
    @     0x5749f20ed5cb  dfly::GenericFamily::Rename()
    @     0x5749f20c5dac  _ZN3fu27abi_4006detail10invocation6invokeIRKPFvN4absl12lts_202505124SpanIKSt17basic_string_viewIcSt11char_traitsIcEEEERKN4dfly14CommandContextEEJSC_SG_EEEDTclcl7forwardIT_Efp_Espcl7forwardIT0_Efp0_EEEOSL_DpOSM_
    @     0x5749f20c5333  fu2::abi_400::detail::type_erasure::invocation_table::function_trait<>::internal_invoker<>::invoke()
    @     0x5749f27fe181  _ZNK3fu27abi_4006detail12type_erasure6tables6vtableINS1_8propertyILb0ELb0EJKFvN4absl12lts_202505124SpanIKSt17basic_string_viewIcSt11char_traitsIcEEEERKN4dfly14CommandContextEEEEEE6invokeILm0EJPKNS2_13data_accessorERKmSE_SI_EEEDcDpOT0_
    @     0x5749f27fe21a  _ZN3fu27abi_4006detail12type_erasure7erasureILb1ENS1_6configILb1ELb1ENS_16capacity_defaultEEENS1_8propertyILb0ELb0EJKFvN4absl12lts_202505124SpanIKSt17basic_string_viewIcSt11char_traitsIcEEEERKN4dfly14CommandContextEEEEEE6invokeILm0ERKSN_JSG_SK_EEEDcOT0_DpOT1_
[...]

PR #6306, opened 2025-12-31
fix: Handle expired hash fields during RENAME without crashing

vyavdoshenko:
- Fix RENAME crash when all hash fields expire during deserialization
- Handle `value_expired` error gracefully in `RdbRestoreValue::Add`

Solution:
1. In `RdbRestoreValue::Add`: return `SKIPPED` instead of `INVALID_VALUE` when `value_expired` occurs
2. In `DeserializeDest`: treat `SKIPPED` as success (similar to key expiry handling)

After fix: RENAME deletes the source key but doesn't create the destination key when all fields have expired (nothing to copy).

Relate to: #6162

### DIFF
--- a/src/server/generic_family.cc
+++ b/src/server/generic_family.cc
@@ -229,6 +229,10 @@ OpResult<DbSlice::ItAndUpdater> RdbRestoreValue::Add(string_view key, string_vie
     config.reserve = pending_read_.reserve;
 
     if (auto ec = FromOpaque(*opaque_res, config, &pv); ec) {
+      // Handle value_expired gracefully - all fields expired during deserialize
+      if (ec.value() == rdb::errc::value_expired) {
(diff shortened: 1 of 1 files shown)


src/server/generic_family_test.cc
```diff
@@ -1452,4 +1452,13 @@ TEST_F(GenericFamilyTest, CopyKeyExists) {
   EXPECT_EQ(Run({"get", "destination"}), "value1");
 }
 
+TEST_F(GenericFamilyTest, HashFieldExpiryDuringDeserialize) {
```
romange, line 1455: did this test crash before the fix?

vyavdoshenko: Yes

romange (approved):

PR #6308, opened 2026-01-01
chore: fix release action

romange:
Change the release image to 20-gcc14 that has a modern build chain. Disable SIMSIMD for fedora.


romange:
Tested here: https://github.com/dragonflydb/dragonfly/actions/runs/20640242087

PR #6313, opened 2026-01-02
fix: handle control messages during SquashPipeline iteration

vyavdoshenko:
When SquashPipeline() iterates over dispatch_q_ using index-based access, control messages (like AclUpdateMessage or CheckpointMessage) can be inserted at the front of the queue by another fiber. This shifts all indices, causing dispatch_q_[i] to access a control message instead of the expected PipelineMessagePtr, resulting in a CHECK failure crash. The fix: Skip control messages when iterating in get_next_fn.
Since this is a race condition, a deterministic test required injecting control messages at the exact moment during SquashPipeline iteration. The test uses a debug-only hook (DEBUG SQUASH_HOOK) that activates before sending pipelined commands, injects a control message on the first get_next_fn call, and verifies the server handles this correctly instead of crashing.
The test hook code (#ifndef NDEBUG) can be removed if preferred - its only purpose is reproducing this specific use case.

Related to: #6162

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -124,14 +124,6 @@ using nonstd::make_unexpected;
 
 namespace facade {
 
-#ifndef NDEBUG
-thread_local bool tl_squash_test_hook_active = false;
-
-void ActivateSquashTestHook() {
-  tl_squash_test_hook_active = true;
-}
-#endif
-
 namespace {
 
 void SendProtocolError(RedisParser::Result pres, SinkReplyBuilder* builder) {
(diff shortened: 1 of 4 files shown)


romange:
Nice catch @vyavdoshenko !

dranikpg (approved):
Good catch, but the testing might be a little overkill 😅

vyavdoshenko:
I can remove the testing part. It was needed to reproduce the original issue.

dranikpg:
I think it might be reproducible by sending pipeline traffic for squashing and doing lots of ACL updates for example

dranikpg:
But not sure if its worth trying to reproduce with outside behaviour

dranikpg (approved):

src/facade/dragonfly_connection.cc
```diff
@@ -1459,8 +1459,14 @@ void Connection::SquashPipeline() {
   uint64_t start = CycleClock::Now();
 
   // We use indexes as iterators are invalidated when pushing into the queue.
-  auto get_next_fn = [i = 0, this]() mutable -> ParsedArgs {
-    const auto& elem = dispatch_q_[i++];
+  // Control messages may be inserted at the front during iteration, so we skip them.
+  auto get_next_fn = [i = size_t{0}, this]() mutable -> ParsedArgs {
```
kostasrim, line 1463: In general, we should not skip the processing of control messages as these are important for:

* acl changes -> if we changed the acl of a user yet we move past the message on the queue we will run a command which we should have never run -> security breach
* checkpoint messages -> takeover flow -> if we keep ignoring them because of squashing we will timeout the takeover

Anyway, I haven't followed the full story -- just thinking out loud here

vyavdoshenko: This is a fix for the crash. The inserted control message will be processed anyway. This case just prevents a crash when the dragonfly processes the messages in a loop, and the control message is inserted in the head unexpectedly.

Issue #6309, opened 2026-01-02
test_reply_guard_oom


BorysTheDev:
we have a fiber stack overflow for test_reply_guard_oom
https://github.com/dragonflydb/dragonfly/actions/runs/20654539833/job/59304938719

```
44219➜    @          0x18fa908         32  google::GetStackTrace()

44219➜    @          0x18f1f34      37392  google::LogMessage::RecordCrashReason()

44219➜    @          0x18f1d08         32  google::LogMessage::SendToLog()

44219➜    @          0x18f1730        160  google::LogMessage::Flush()

44219➜    @          0x18f4cc8         64  google::LogMessageFatal::~LogMessageFatal()

44219➜    @          0x14a0734         32  util::fb2::detail::FiberInterface::CheckStackMargin()

44219➜    @          0x14a03d4        288  util::fb2::detail::FiberInterface::Terminate()

44219➜    @          0x14b7194        496  util::fb2::detail::WorkerFiberImpl<>::run_()

44219➜    @          0x14b60d8         80  util::fb2::detail::WorkerFiberImpl<>::WorkerFiberImpl<>()::{lambda()#1}::operator()()

44219➜    @          0x14bf76c         48  std::__invoke_impl<>()

44219➜    @          0x14be1d0         64  std::__invoke<>()

44219➜    @          0x14bc390         64  std::invoke<>()

44219➜    @          0x14bb160         48  boost::context::detail::fiber_record<>::run()

44219➜    @          0x14b92dc         80  boost::context::detail::fiber_entry<>()
```

PR #6315, opened 2026-01-02
fix: increase connection_dispatch fiber stack size to prevent overflow

vyavdoshenko:
Increased connection_dispatch fiber stack size from 64KB to 128KB to prevent stack overflow during command dispatch.

Closes: #6309


romange:
the overflow happens only in DEBUG mode, so I would not rush increasing the stack size for release. historically we increased `kFiberDefaultStackSize`  in dfly_main for DEBUG

vyavdoshenko:
> the overflow happens only in DEBUG mode, so I would not rush increasing the stack size for release. historically we increased `kFiberDefaultStackSize` in dfly_main for DEBUG

Maybe it's worth removing this change and executing flaky test in release mode only?

dranikpg:
With a few thousand clients it can be around half a gigabyte

Issue #6314, opened 2026-01-02
Reply handlers and transaction waits


dranikpg:

dranikpg:
### Callback return

From takeaway (1): A command handler can return a variant<ErrorReply, Replier>
From takeaway (2): Can be done by a single void callback

Then the sync flow will look as follows:
```c++
variant<ErrorReply, Callback> res = Cmd();
if (res == error)
  Send(error.error());
else
  tx->Wait();
  res.callback(reply_builder);
```

Async:
```c++
variant<ErrorReply, Callback> res = Cmd();
if (res == error)
  Send(error.error())
else
  cmd_cntx->blocker = &tx->run_barrier_
  cmd_cntx->replier = res.callback();
```

```c++
// handle next head
head_ready = false;
waiter = []{
  head_ready = true;
  single_ec_.notify();
};

head->blocker->Subscribe(&waiter);
single_ec_.await(io_event or head_ready);

if (head_ready)
  head->replier(reply_builder);
```

So the command handler will be like
```c++
AsyncHandlerReply CmdGet2(CmdArgList args, CommandContext* cmnd_cntx) {
  string_view key = args[0];
  if (key == "BAD")
    return ErrorReply{"Bad key"};

  auto strres = make_unique<OpResult<StringResult>>();
  auto cb = [key, strres = strres.get()](Transaction* tx, EngineShard* es) {
    auto it_res = tx->GetDbSlice(es->shard_id()).FindReadOnly(tx->GetDbContext(), key, OBJ_STRING);
    if (!it_res.ok())
      *strres = it_res.status();
    else
      *strres = ReadString(tx->GetDbIndex(), key, (*it_res)->second, es);
    return OpStatus::OK;
  };

  cmnd_cntx->tx->Execute(cb, true, true);
  return [strres = std::move(strres)](SinkReplyBuilder* rb) {
    GetReplies{rb}.Send(std::move(*strres));
  };
}
```

dranikpg:
### Coroutines

A coroutine is similar to a functor with a state machine

```c++
void MyCoroutine() {
  // phase 1
  co_await some_event()
  // phase 2
}

struct MyCoroutine {
  void call() {
    switch(phase) {
      case 1:
        // do phase 1
        return
      case 2:
        // do phase 2
        return
     }
  }
 
  local variables...;
  int phase;
}
```

[...]

PR #6317, opened 2026-01-03
Coroutine POC

dranikpg:
Related to #6314

### DIFF
--- a/src/server/CMakeLists.txt
+++ b/src/server/CMakeLists.txt
@@ -55,7 +55,7 @@ if (WITH_EXTENSION_CMDS)
 endif()
 
 add_library(dfly_transaction db_slice.cc blocking_controller.cc
-            command_registry.cc  cluster_support.cc
+            command_registry.cc  async_cmd.cc cluster_support.cc
             journal/cmd_serializer.cc journal/tx_executor.cc namespaces.cc
             common.cc journal/journal.cc journal/types.cc journal/journal_slice.cc
             server_state.cc table.cc  transaction.cc tx_base.cc
--- a/src/server/async_cmd.cc
+++ b/src/server/async_cmd.cc
@@ -1,0 +1,37 @@
+#include "server/async_cmd.h"
+
+#include "server/transaction.h"
+
+namespace dfly::cmd {
+
+namespace detail {
+
+facade::SinkReplyBuilder* ReplyEventWaiter::await_resume() const noexcept {
+  return cmnd_cntx->rb();
+}
+
+bool LastHopWaiter::await_ready() const noexcept {
+  cmnd_cntx->tx->Execute(callback, true, true);
+
+  if (cmnd_cntx->IsDeferredReply())
+    return false;
+
+  // Sync execution
+  cmnd_cntx->tx->Blocker()->Wait();
+  return true;
+}
+
+void LastHopWaiter::await_suspend(std::coroutine_handle<> handle) const noexcept {
+  cmnd_cntx->Resolve(cmnd_cntx->tx->Blocker(), handle);
+}
+
+facade::OpStatus LastHopWaiter::await_resume() const noexcept {
+  return cmnd_cntx->tx->GetLocalResult();
+}
+
+}  // namespace detail
+
+void Task::Promise::return_value(facade::ErrorReply&& err) noexcept {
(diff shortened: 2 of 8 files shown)


src/server/string_family.cc
```diff
@@ -1406,32 +1352,35 @@ void MGetGeneric(CmdArgList args, CommandContext* cmnd_cntx,
...
+  shared_ptr<optional<GetResp>[]> mget_results(new optional<GetResp>[arg_len]);
   ReorderShardResults(absl::MakeSpan(mget_resp.get(), shard_set->size()), cmnd_cntx->tx,
                       absl::MakeSpan(mget_results.get(), arg_len));
 
   if (cmnd_cntx->mc_command()) {
     // We can not access `args` inside the lambda below if it runs asynchronously.
     // Access to arguments must be done via cmnd_cntx.
-    auto replier = [mget_results = std::move(mget_results), cmnd_cntx](SinkReplyBuilder* builder) {
+    boost::intrusive_ptr<Transaction> keepalive{cmnd_cntx->tx};
+    auto replier = [mget_results, mget_resp, cmnd_cntx, keepalive](SinkReplyBuilder* builder) {
```
dranikpg: issues:
1. Use a move-only functor so we can use unique_ptr
2. Automatically implement a keepalive mechanism for the transaction

src/facade/parsed_command.h
```diff
@@ -112,78 +111,42 @@ class ParsedCommand : public cmn::BackedArguments {
...
-  bool MarkForDestruction() {
-    if (!is_deferred_reply_)
-      return true;
-    uint8_t prev_state = state_.fetch_or(DELETE_INTENT, std::memory_order_acq_rel);
-
-    // If the reply is already done, we can destroy it now.
-    return (prev_state & ASYNC_REPLY_DONE) != 0;
-  }
+  bool IsReady() const;  // If deferred is ready
+  bool OnCompletion(util::fb2::detail::Waiter* waiter);
```
romange, line 135: what does it return?

src/facade/dragonfly_connection.cc
```diff
@@ -2364,10 +2345,15 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
...
       // we rely on the kernel to keep feeding us data until we multishot is disabled.
       DoReadOnRecv(FiberSocketBase::RecvNotification{true});
-      io_event_.await([this]() {
-        return io_buf_.InputLen() > 0 || (parsed_head_ && parsed_head_->PollHeadForCompletion()) ||
-               io_ec_;
-      });
+
+      // Update head state and subscribe to updates if its not ready
+      if (parsed_head_ && parsed_head_ != parsed_to_execute_ && !head_ready)
+        head_ready |= parsed_head_->OnCompletion(&head_waiter);
```
romange, line 2351: Please write explanation here what exactly we do here.  It's gentle, so I prefer we have comments than going through low-level code to understand

src/facade/parsed_command.cc
```diff
@@ -93,75 +90,50 @@ void ParsedCommand::SendSimpleString(std::string_view str) {
...
+bool ParsedCommand::OnCompletion(util::fb2::detail::Waiter* waiter) {
+  dfly::Overloaded ov{[](const payload::Payload& pl) {
+                        DCHECK(pl.index() > 0);
+                        return true;
+                      },
+                      [waiter](const AsyncTask& task) {
+                        task.blocker->OnCompletion(waiter);
+                        return false;
+                      }};
+  return visit(ov, reply_);
```
romange, line 126: how the access to `reply_` is synchronized between threads? 
before, with `state_` one a single thread could access `reply_payload`.
now I see OnCompletion is called from the io thread but you also write into reply_ from the shard thread

dranikpg: I never write into reply from the shard thread, its either assigned to the return value of the handler or set to the callback that will reply

romange: Ok, I will need to see SET command I guess - how it propagates the Stored reply from the shard thread.

dranikpg: it won't propagate from the shard thread, it will work via a replier

dranikpg:
Already implemented as non POC

PR #6319, opened 2026-01-05
CI: sync valkey tests

abhijat:
Sync valkey tests so repeat-tests match the rest of the tests in CI.

### DIFF
--- a/.github/actions/repeat/action.yml
+++ b/.github/actions/repeat/action.yml
@@ -39,6 +39,12 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: Sync valkey-search tests
+      shell: bash
+      run: |
+        cd ${GITHUB_WORKSPACE}/tests/dragonfly/valkey_search
+        # main branch revision
+        ./sync-valkey-search-tests.sh 90124dc91756b24cb2e58e5c4eea5b8d53004ea6
     - name: Repeat pytests
       id: main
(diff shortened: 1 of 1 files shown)


romange:
another thing we could do is to remove valkey tests from this action

.github/actions/repeat/action.yml
```diff
@@ -39,6 +39,12 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: Sync valkey-search tests
+      shell: bash
+      run: |
+        cd ${GITHUB_WORKSPACE}/tests/dragonfly/valkey_search
+        # main branch revision
+        ./sync-valkey-search-tests.sh 90124dc91756b24cb2e58e5c4eea5b8d53004ea6
```
vyavdoshenko, line 47: This is a copy-paste. Please look at the others' actions and extract this step in common. Reuse this step in both places. It will help us in the future, in revision, when changes will be made in one place but skipped in another.

abhijat: Makes sense, I'll do a follow up PR, right now this workflow is broken so I will add it as is

abhijat: created https://github.com/dragonflydb/dragonfly/pull/6333 - I wasn't able to remove as much duplication as I imagined, but now there is one composite action which is reused wherever sync is required.

PR #6321, opened 2026-01-05
feat(server): Support squashing of global eval with single shard

dranikpg:
Global EVAL with a single shard instance can now be squashed

Refactored single shard evalation to use proper squashing to keep execution mode information.
This PR adds support for squashing global EVAL commands when running on a single shard instance, optimizing the execution path to avoid unnecessary hops.

## Key changes:

1. Refactored PrepareMultiForScheduleSingleHop to PrepareSingleSquash with support for both LOCK_AHEAD and GLOBAL modes
2. Updated CanRunSingleShardMulti to allow single-shard global commands to be squashed
3. Added test coverage for the single-shard multi squash optimization

### DIFF
--- a/src/server/main_service.cc
+++ b/src/server/main_service.cc
@@ -2327,7 +2327,7 @@ void Service::EvalInternal(CmdArgList args, const EvalArgs& eval_args, Interpret
     });
 
     // Migration only makes sense if there are distinct shards
-    if (sid.has_value() && real_sid != ss->thread_index()) {
+    if (sid.has_value() && *sid != ss->thread_index()) {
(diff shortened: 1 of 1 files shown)


src/server/main_service.cc
```diff
@@ -2318,10 +2326,11 @@ void Service::EvalInternal(CmdArgList args, const EvalArgs& eval_args, Interpret
       return OpStatus::OK;
     });
 
-    if (*sid != ss->thread_index()) {
+    // Migration only makes sense if there are distinct shards
+    if (sid.has_value() && real_sid != ss->thread_index()) {
```
romange:
```suggestion
    if (sid.has_value() && *sid != ss->thread_index()) {
```

PR #6324, opened 2026-01-05
chore: Reused parsed messages for memcache workloads

romange:
Also setup batch reply for multiple messages.
When running `./dfly_bench  -n 5000000 -p 11211 --nogreet --qps=0  -d 128 --key_maximum=20000000   --ratio=1:0  --pipeline=30 -c 3  -P memcache_text` I get:

1572K qps before the change, and 2168K qps after the change. It's still a bit worse (10% higher latency and 10% lower throughtput) than running the same benchmark with RESP SET protocol with dragonfly running `--experimental_io_loop_v2=false`.

`experimental_io_loop_v2` has a worse performance in general for both modes but it is a must for using async dispatching, as we can not block on Socket::Recv unconditionally.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1184,7 +1184,11 @@ Connection::ParserStatus Connection::ParseRedis(unsigned max_busy_cycles) {
     service_->DispatchCommand(ParsedArgs{*parsed_cmd_}, parsed_cmd_);
   };
 
-  auto dispatch_async = [this]() -> MessageHandle { return {FromArgs(tmp_parse_args_)}; };
+  auto dispatch_async = [this]() -> MessageHandle {
+    PipelineMessagePtr ptr = GetFromPoolOrCreate();
+    FillBackedArgs(tmp_parse_args_, ptr.get());
+    return {std::move(ptr)};
+  };
 
   io::Bytes read_buffer = io_buf_.InputBuffer();
   // Keep track of total bytes consumed/parsed. The do/while{} loop below preempts,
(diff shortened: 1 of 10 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -2174,14 +2197,13 @@ void Connection::ReleaseParsedCommand(ParsedCommand* cmd, bool is_pipelined) {
...
-  // Cache a single command for immediate reuse, otherwise free it.
-  // TODO: we can cache parsed commands similarly to pipeline_req_pool_.
-  // In fact we should unify both approaches.
   if (parsed_cmd_ == nullptr) {
     parsed_cmd_ = cmd;
     parsed_cmd_->ResetForReuse();
   } else {
-    delete cmd;
+    // TODO: limit pipeline_cmd_cache_bytes.
+    stats_->pipeline_cmd_cache_bytes += UsedMemoryInternal(*cmd);
```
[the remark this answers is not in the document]
romange: it's wrong

src/facade/dragonfly_connection.cc
```diff
@@ -1669,20 +1672,6 @@ void Connection::AsyncFiber() {
   qbp.pipeline_cnd.notify_all();
 }
 
-Connection::PipelineMessagePtr Connection::FromArgs(const RespVec& args) {
-  PipelineMessagePtr ptr = GetFromPipelinePool();
```
dranikpg, line 1673: You decided to avoid this helper function because it has only two usages?

romange: no, it had only one usage.

PR #6331, opened 2026-01-06
refactor: add checks to investigate the race condition in slots migration

BorysTheDev:
we have failed [test](https://github.com/dragonflydb/dragonfly/actions/runs/20643100710/job/59277432436)  with a stack trace that could be explained only by a race condition. This PR should add checks and logs to prove the problem

### DIFF
--- a/src/server/cluster/outgoing_slot_migration.cc
+++ b/src/server/cluster/outgoing_slot_migration.cc
@@ -45,8 +45,7 @@ class OutgoingMigration::SliceSlotMigration : private ProtocolClient {
     // it should already be unregistered, this cancel was added to avoid race condition that we
     // possibly have.
     bool unregistered = streamer_.Cancel();
-    DCHECK(!unregistered);
-    LOG_IF(WARNING, unregistered)
+    LOG_IF(DFATAL, unregistered)
         << "Streamer was not unregistered properly. Check code for race conditions.";
     exec_st_.JoinErrorHandler();
   }


src/server/journal/streamer.cc
```diff
@@ -155,14 +155,19 @@ void JournalStreamer::Start(util::FiberSocketBase* dest) {
...
 
-void JournalStreamer::Cancel() {
+bool JournalStreamer::Cancel() {
   VLOG(1) << "JournalStreamer::Cancel " << cntx_->IsCancelled();
   waker_.notifyAll();
+  bool res = false;
   if (journal_cb_id_) {
-    journal_->UnregisterOnChange(journal_cb_id_);
+    auto cb_id = journal_cb_id_;
+    journal_cb_id_ = 0;  // Reset to prevent double unregistration in another fiber
```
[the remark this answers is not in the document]
BorysTheDev: we don't have another thread

vyavdoshenko (approved):
lgtm

src/server/cluster/outgoing_slot_migration.cc
```diff
@@ -42,6 +42,12 @@ class OutgoingMigration::SliceSlotMigration : private ProtocolClient {
 
   ~SliceSlotMigration() {
     CloseSocket();
+    // it should already be unregistered, this cancel was added to avoid race condition that we
+    // possibly have.
+    bool unregistered = streamer_.Cancel();
+    DCHECK(!unregistered);
+    LOG_IF(WARNING, unregistered)
```
romange: we usually merge such pattern into `LOG_IF(DFATAL, unregistered) << ` 
instead of dchecking and logging

src/server/journal/streamer.cc
dranikpg, line 164: its because we can block below, right 🤔

dranikpg: btw a c++ function I like is

```c++
if (auto old_id = exchange(journal_cb_id_, 0); old_id) {
...
}
```

dranikpg (approved):

src/server/journal/streamer.cc
BorysTheDev: Will update next time, thx for example

PR #6333, opened 2026-01-06
CI: Extract common step for syncing valkey tests into action

abhijat:
Extract duplicate code to single action

### DIFF
--- a/.github/actions/sync-valkey-tests/action.yml
+++ b/.github/actions/sync-valkey-tests/action.yml
@@ -1,0 +1,12 @@
+name: Sync valkey-search tests
+description: "Synchronizes valkey-search tests using a fixed revision"
+
+runs:
+  using: composite
+  steps:
+    - name: Sync valkey-search tests
+      shell: bash
+      run: |
+        cd ${GITHUB_WORKSPACE}/tests/dragonfly/valkey_search
+        # main branch revision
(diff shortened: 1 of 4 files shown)


.github/actions/sync-valkey-tests/action.yml
```diff
@@ -0,0 +1,12 @@
...
+
+runs:
+  using: composite
+  steps:
+    - name: Sync valkey-search tests
+      shell: bash
+      run: |
+        cd ${GITHUB_WORKSPACE}/tests/dragonfly/valkey_search
+        # main branch revision
+        ./sync-valkey-search-tests.sh 90124dc91756b24cb2e58e5c4eea5b8d53004ea6
```
kostasrim, line 12: I don't know what these tests are or how long they take to execute but If they can regress for some reason I would consider using `timeout` here just in case.  We do the same on reg test action when we run the pytests. 

The reason is that when using actions, setting a timeout on the yaml file itself won't apply to the action so each action must be "responsible" and "bounded"

Still waiting for gh to fix this.

p.s. ignore me if my comment does not apply and these tests are fast

abhijat: These are actually just commands to sync the test code using git + sed, the tests are run elsewhere like other python tests via pytest etc where needed (eg in regression tests workflow). 

So these commands are pretty fast, they only clone the repo and patch some files using sed

Issue #6226, opened 2025-12-16
Simplify ACL updates


romange:
Current status:
 
1. VerifyConnectionAclStatus is called for each command during the EXEC phase  (see also #6196).
2. But then we use `AclUpdateMessage` to asynchronously update the ACL AFTER the inflight commands finish. 

The whole flow is over-complicated. What should be done imho:
1. Move the ACL checks for MULTI/EXEC into the enqueue phase (similarly to how we test for valid argument, know command etc).
2. At the exec phase either the whole transaction fails with EXECABORT or it is executed and we do not need to check for ACL.
3. AclUpdateMessage should be deleted. We should update the the connection context directly from `AclFamily::StreamUpdatesToAllProactorConnections` and the acl state should be moved from facade into dfly::ConnectionContext - we should not have any acl related variables inside facade.

kostasrim:
> At the exec phase either the whole transaction fails with EXECABORT or it is executed and we do not need to check for ACL.

That's actually a bug on our end also. Consider the following:

acl setuser kostas on >pass +@all ~*
auth kostas pass
multi
set foo bar
set bar foo
**admin** runs acl setuser kostas resetkeys ~foo 

dragonfly returns:

1) OK
2) (error) NOPERM kostas ACL rules changed between the MULTI and EXEC

Valkey:

rejects the multi/exec block

The whole block should have been rejected instead and this is an `atomicity bug`. 

P.s we should still check the list twice: 1) when the user sends the commands 2) when the exec is called (to cover the above as well)

romange:
I do not understand why we need to check twice though. According to your example, once - during enqueue phase should be enough.

kostasrim:
@romange if you enqueue a command on key `foo`, revoke the acl rule `foo` from user (via some other client) and then sent exec we will execute this without an acl violation. 

For multi/exec blocks commands are ACL checked twice:

* once when you sent them 
* again once the client sents the final exec

In-between, the ACL's can change at any time

romange:
@kostasrim  and this is perfectly fine. Management commands and in general any type of control commands in Dragonfly (or  many other systems) are eventually consistent. Nobody expects that a ACL update affects the in-flight operation that started the execution, so this complexity is unnecessary. Specifically for you example, the operator observes "EXEC" as if it happens "before" he pressed enter.

kostasrim:
we discussed this internally. we opted in for eventual consistency and we will do 1 check only when we enqueue

Issue #6299, opened 2025-12-29
Simplify DispatchTracker functionality


romange:
## Background

DispatchTracker is used for several scenarios, where the most important one is DFLY TAKEOVER, see https://github.com/dragonflydb/dragonfly/pull/6142 for more context.

The high level goal of DispatchTracker is to make sure that the connection is not in middle of Dispatching, i.e. either before cmd dispatch call or after it. Why? Sometimes we want to switch states in such way that other commands won't interfere. Currently  DispatchTracker is implemented by issuing CheckPoint message (see SendCheckpoint) which enqueues a message to the back of the dispatch queue.  Once it is processed, it decreases blocking counter. The waiter (issuer) waits for the counter to reach 0 at which point it knows that all destinations processed the message and it is safe to do whatever it needs to do.

So instead of waiting for currently running command to finish, it waits for all the already enqueued commands to run. The dispatch queue can hold hundreds of thousands of even millions of commands which can take a long time. Also, the state after processing of Checkpoint message is not well defined. Nothing prevents from connections to continue dispatching commands after Checkpoint was processed. The issue must update the state of destination connections before Checkpoint is processed by them so by the time it is processed there will be a guarantee that the connection state is changed, all commands after Checkpoint are handled according to the new state.

1. We continuously have problems with TAKEOVER and dispatch tracker. See https://github.com/dragonflydb/dragonfly/issues/6114
2. Processing Checkpoint message is slow: it is enqueued to the end of the queue because it piggybacks on dispatch_q generic processing. dispatch_q is an async message queue between another fiber and the connection as well as the pipeline queue of the commands.

## Suggestion
1. Analyze each one of the use-cases that involve DispatchTracker and see if DispatchTracker is what we really need. Suppose DispatchTracker is indeed the simplest method for making ensuring that currently running commands finish execution.
2. Then separate pipeline processing of RESP commands from other messages. Move pipelining to parsed_head_/parsed_tail_/parsed_to_execute_ intrusive queue  that is already used by memcache. This way dispatch_q will only be used by other messages.
3. Process dispatch_q messages regularly at times when a dispatched command finished running but without waiting for all pipelined commands to finish. 
4. In other words `dispatch_q` will generally have higher priority though it will be processed after all the currently dispatched commands are finished.

** Somewhat related  to https://github.com/dragonflydb/dragonfly/issues/6226 which should deprecate `AclUpdateMessage`

dranikpg:
[...]

PR #6340, opened 2026-01-07
refactor(acl): from facade to server

kostasrim:
Simplifies ACL implementation by eliminating asynchronous update messages and relocating ACL state from the facade I/O layer to the server layer where it belongs.

**ACL State Migration**
- Moved `authed_username`, `acl_commands`, `keys`, `pub_sub`, `acl_db_idx` from `facade::ConnectionContext` to `dfly::ConnectionContext`
- Removed `facade/conn_context.cc` (constructor inlined)

**Eliminated Async Update Messages**
- Deleted `AclUpdateMessage` struct and `SendAclUpdateAsync()`
- `StreamUpdatesToAllProactorConnections()` now updates connection contexts directly via callback
- Removed message variant overhead and related dispatch logic

**Simplified MULTI/EXEC ACL Verification**
- ACL checks now occur only during command enqueue phase
- Removed redundant verification during EXEC phase
- In-flight transactions no longer affected by concurrent ACL changes

resolves: #6226

### DIFF
--- a/src/server/conn_context.cc
+++ b/src/server/conn_context.cc
@@ -53,7 +53,6 @@ std::string StoredCmd::FirstArg() const {
 
 ConnectionContext::ConnectionContext(facade::Connection* owner, acl::UserCredentials cred)
     : facade::ConnectionContext(owner) {
-  skip_acl_validation = false;
   if (owner) {
     skip_acl_validation = owner->IsPrivileged();
     has_main_or_memcache_listener = owner->IsMainOrMemcache();


kostasrim:
@copilot write me a summary of the changes so I can include it in the PR description.

src/server/main_service.cc
```diff
@@ -1282,8 +1282,7 @@ optional<ErrorReply> Service::VerifyCommandExecution(const CommandContext& cmd_c
     return facade::ErrorReply{OpStatus::OUT_OF_MEMORY};
   }
```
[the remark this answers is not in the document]
kostasrim: overkill, discussed on the issue as well.

src/server/acl/acl_family.cc
```diff
@@ -117,8 +117,13 @@ void AclFamily::StreamUpdatesToAllProactorConnections(const std::string& user,
...
     if (!connection->IsHttp() && connection->cntx()) {
-      connection->SendAclUpdateAsync(facade::Connection::AclUpdateMessage{
-          user, update_commands, update_keys, update_pub_sub, db});
+      auto* cntx = static_cast<dfly::ConnectionContext*>(connection->cntx());
+      if (user == cntx->authed_username) {
+        cntx->acl_commands = update_commands;
+        cntx->keys = update_keys;
+        cntx->pub_sub = update_pub_sub;
+        cntx->acl_db_idx = db;
+      }
```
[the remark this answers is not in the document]
kostasrim: It does not violate the shared nothing architecture and the comment is nonsense. 

We traverse connections atomically and in a thread safe manner because TraverConnections uses AwaitFiberOnAll.

tests/dragonfly/acl_family_test.py
```diff
@@ -212,13 +202,9 @@ async def test_acl_cat_commands_multi_exec_squash(df_factory):
...
     assert res == "OK"
 
-    res = await client.execute_command("AUTH myuser kk")
-    assert res == "OK"
-
-    await client.execute_command("MULTI")
-    assert res == "OK"
-    for x in range(33):
-        await client.execute_command(f"SET x{x} {x}")
+    await client.execute_command("AUTH myuser kk")
```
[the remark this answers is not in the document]
kostasrim: it's fine, I don't add assertions when they are not needed -- auth logic is checked elsewere

tests/dragonfly/acl_family_test.py
```diff
@@ -300,6 +287,7 @@ async def test_acl_del_user_while_running_lua_script(df_server):
 
 
 @pytest.mark.asyncio
+@pytest.mark.skip("Check TODO in the body below")
```
kostasrim, line 290: @romange we break LUA scripts because now the changes done to the ACL's of a user are `immediate` (we don't place an acl update message in the dispatch queue). So the problem here is:

1. script running
2. acl's change
3. script fails because of (2)

There is a solution to this using a `stab` context. The idea here is that we copy the ACL's checks into the dummy context and the script uses that for its entire execution. THat way we decoupled the ACL's used before and after the script started running. However, ConnectionContext data type is `non trivial` (has a connection state) and therefore using a stab context is not straightforward

romange: Please leave it as is.  I "feel" that it is related to https://github.com/dragonflydb/dragonfly/issues/6299

but I need to think how to formalize the problem.

src/server/conn_context.h
```diff
@@ -328,6 +328,17 @@ class ConnectionContext : public facade::ConnectionContext {
   // The related connection is bound to main listener or serves the memcached protocol
   bool has_main_or_memcache_listener = false;
 
+  // ACL's
+  // TODO fix inherit actual values from default
+  std::string authed_username{"default"};
+  std::vector<uint64_t> acl_commands;
```
romange, line 344: please comment heavily these fields. 
I lack the understanding how `acl_commands` are used  for example for check for ACL permissions.

src/server/conn_context.h
```diff
@@ -328,6 +328,17 @@ class ConnectionContext : public facade::ConnectionContext {
...
+  // keys
+  dfly::acl::AclKeys keys{{}, true};
```
romange: why do you need to specify c'tor parameters explicitly? 
I not make the default c'tor have such semantics?

src/server/conn_context.h
```diff
@@ -328,6 +328,17 @@ class ConnectionContext : public facade::ConnectionContext {
...
+  // pub/sub
+  dfly::acl::AclPubSub pub_sub{{}, true};
+  // db index, std::numeric_limits<size_t>::max for ALL db's
+  size_t acl_db_idx = 0;
```
romange: is it a whitelisted index?

tests/dragonfly/acl_family_test.py
```diff
@@ -234,7 +221,7 @@ async def test_acl_cat_commands_multi_exec_squash(df_factory):
     denied = False
     while not denied and time.time() - start < 10:
         try:
-            await client.execute_command(f"SET x{x} {x}")
+            await client.execute_command(f"SET x bar")
             asyncio.sleep(0.1)
```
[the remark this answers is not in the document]
romange: please fix

src/server/conn_context.h
kostasrim: We don't. These are all initialized in ctor anyway. I removed the defaults

src/server/conn_context.h
kostasrim: I agree it's confusing. It's not whitelisted -- it's initialized in the contructor. 

I removed the default here.

tests/dragonfly/acl_family_test.py
kostasrim: Sounds good. Thinking out loud here: acl's should be immutable once a script starts running so if we somehow had a copy of them in the "lua context" we would effectively decouple them . That way, changes in ConnectionContext do not affect or observed by the "lua context/script" that is currently running

src/server/conn_context.h
```diff
@@ -328,6 +328,31 @@ class ConnectionContext : public facade::ConnectionContext {
...
+  std::vector<uint64_t> acl_commands;
+
+  // Keyspace. Each key referenced in a command must match (any) of the rules (globs).
+  dfly::acl::AclKeys keys;
+
+  // Pub/sub channels. Each channel referenced in a command must match (any) of the rules (globs).
+  dfly::acl::AclPubSub pub_sub;
+
+  // db index, std::numeric_limits<size_t>::max for ALL db's. Dragonfly specific extension.
+  size_t acl_db_idx;
```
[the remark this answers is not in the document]
kostasrim: @romange we have a protest by copilot 🤣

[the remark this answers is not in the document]
kostasrim: well, that field is not used when the context does not define it but I agree, default initializing struct members is a good habit. moved

src/server/conn_context.h
romange: so what `acl_commands[0]` represents? an allow mask of all the acl categories?
what is the index of this array?

kostasrim: > so what acl_commands[0] represents?  what is the index of this array?

by order they get registered by `Service::RegisterCommands()`. So 
[0] should be all commands of main service. bit 0 maps to QUIT, 1 to MULTI, 2 to WATCH etc..
[1] server_family, 
[2] generic family, 
[3] list family and so on. 

>  an allow mask of all the acl categories?

When a user is granted a category, internally the command bits are set to reflect to that set (their respective bits are set to 1)

p.s. the mappings are handled automatically through indexers build in acl family

src/server/conn_context.h
```diff
@@ -328,6 +328,31 @@ class ConnectionContext : public facade::ConnectionContext {
...
+  // The following variables represent the ACL rules of the context.
+  // Each command, before run, is authorized against those rules by
+  // IsUserAllowedToInvokeCmd(and variants) in validator.cc
+
+  // Username
+  std::string authed_username{"default"};
+
+  // Each list entry is a bitfield for a command family. Each bit in the bitfield maps to
+  // a single command from that family. The whole table encodes the full access list of commands
+  // for the user. For more info, see acl_commands_def.h and CommandRegistry::operator<<
```
romange:
```suggestion
// Each entry in the list is a bitfield representing a specific command family, 
// where each bit corresponds to an individual command within that family. 
// Together, these entries encode the user's full ACL to commands. 
// The index 'i' in 'acl_commands[i]' refers to the command family based on 
// its registration order at runtime. For more details, see acl_commands_def.h.
```

src/server/conn_context.h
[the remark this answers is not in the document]
romange: I agree - why not initialize it to `std::numeric_limits<size_t>::max ` by default?

src/server/conn_context.h
```diff
@@ -328,6 +328,31 @@ class ConnectionContext : public facade::ConnectionContext {
...
+  dfly::acl::AclKeys keys;
+
+  // Pub/sub channels. Each channel referenced in a command must match (any) of the rules (globs).
+  dfly::acl::AclPubSub pub_sub;
+
+  // db index, std::numeric_limits<size_t>::max for ALL db's. Dragonfly specific extension.
+  size_t acl_db_idx;
+
+  // Skip ACL validation, used by internal commands and commands run on admin port
+  bool skip_acl_validation;
```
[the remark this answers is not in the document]
romange: I am confused, why not initiailize it to `false`?
I did not say not to use initializers - I only commented about a non-POD class that I beleive should have a reasonable default state.

romange:
seems that my rewrite was poorly formatted

romange (approved):

PR #6349, opened 2026-01-08
chore(cmake): Update jsoncons to 1.5.0

abhijat:
Refreshes jsoncons to 1.5.0. The following steps were performed:

* jsoncons fork: 
    - start with tag 1.5.0
    - apply dragonfly related changes on top
    - squash some closely related commits for cleanliness
    - change unit tests so all tests pass
    - create final tag for use in dragonfly
* dragonfly:
    - use jsoncons tag
    - make API changes where required (mainly wrapped json path)
    - fix failing tests (`json_test`)

### DIFF
--- a/src/core/json/json_test.cc
+++ b/src/core/json/json_test.cc
@@ -72,16 +72,20 @@ TEST_F(JsonTest, Query) {
 }
 
 TEST_F(JsonTest, Errors) {
-  auto cb = [](json_errc err, const ser_context& contexts) { return false; };
+  auto cb = [](json_errc, const ser_context&) { return false; };
 
   json_decoder<json> decoder;
   basic_json_parser<char> parser(basic_json_decode_options<char>{}, cb);
 
   std::string_view input{"\000bla"};
   parser.update(input.data(), input.size());
-  parser.parse_some(decoder);
-  parser.finish_parse(decoder);
-  parser.check_done();
+
+  std::error_code ec;
+  parser.parse_some(decoder, ec);
+
+  EXPECT_TRUE(ec);
+
+  EXPECT_EQ(ec, json_errc::unexpected_eof);
   EXPECT_FALSE(decoder.is_valid());
 }
 


src/server/detail/wrapped_json_path.h
```diff
@@ -271,7 +271,7 @@ OpResult<JsonCallbackResult<std::optional<T>>> WrappedJsonPath::ExecuteMutateCal
     custom_functions<JsonType> funcs = custom_functions<JsonType>();
 
     std::error_code ec;
-    static_resources<ValueType, Reference> static_resources(funcs);
+    static_resources static_resources(funcs);
     Evaluator e;
 
     JsonSelector expr = e.compile(static_resources, path_.view(), ec);
```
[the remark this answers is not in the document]
abhijat: fixed

src/core/json/json_test.cc
```diff
@@ -72,16 +72,20 @@ TEST_F(JsonTest, Query) {
...
   basic_json_parser<char> parser(basic_json_decode_options<char>{}, cb);
 
   std::string_view input{"\000bla"};
   parser.update(input.data(), input.size());
-  parser.parse_some(decoder);
-  parser.finish_parse(decoder);
-  parser.check_done();
+
+  std::error_code ec;
+  parser.parse_some(decoder, ec);
```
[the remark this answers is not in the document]
abhijat: `finish_parse` is still valid, this change was made because this test checks failure, and it is much cleaner to use the `error_code` variant of the call and then examine `ec` than catch failures.

The main reason this test changed is that `jsoncons` now does slightly different handling of `eof` than it did before.

src/server/detail/wrapped_json_path.h
```diff
@@ -271,16 +271,16 @@ OpResult<JsonCallbackResult<std::optional<T>>> WrappedJsonPath::ExecuteMutateCal
...
 
-    JsonSelector expr = e.compile(static_resources, path_.view(), ec);
+    JsonSelector expr = e.compile(static_res, path_.view(), ec);
     if (ec) {
       VLOG(1) << "Failed to mutate json with error: " << ec.message();
       return OpStatus::SYNTAX_ERR;
     }
 
-    dynamic_resources<ValueType, Reference> resources;
+    eval_context<ValueType, Reference> resources;
```
abhijat, line 283: Looking at git history in jsoncons this class was renamed.

src/server/detail/wrapped_json_path.h
```diff
@@ -271,16 +271,16 @@ OpResult<JsonCallbackResult<std::optional<T>>> WrappedJsonPath::ExecuteMutateCal
     custom_functions<JsonType> funcs = custom_functions<JsonType>();
 
     std::error_code ec;
-    static_resources<ValueType, Reference> static_resources(funcs);
+    static_resources static_res(funcs);
```
abhijat, line 274: This class now only takes one template parameter, which can be inferred.

src/core/json/json_test.cc
abhijat, line 84: `parse_some` now handles eof on parsing start early and does not call the supplied callback for that case. so this test needed adjustment

PR #6351, opened 2026-01-08
chore(server): Updates around DispatchMC/DispatchCommand

dranikpg:
1. Introduce `AsyncPreference` enum to control async dispatching
2. Return DispatchResult from `DispatchMC`
3. Refactor DispatchMC to use string_views instead of buffer copying as arguments are immutable now
4. Refactor DispatchCommand to factor out three large functions and reduce its size 

(1) and (2) are an extraction from #6317

### DIFF
--- a/src/server/main_service.cc
+++ b/src/server/main_service.cc
@@ -1485,7 +1485,7 @@ DispatchResult Service::DispatchCommand(facade::ParsedArgs args, facade::ParsedC
   ArgSlice tail_args = args_no_cmd.ToSlice(&tmp_vec);
 
   // Block on CLIENT PAUSE if needed
-  if (auto* conn = cmnd_cntx->conn(); conn /* replica context doesn't have an owner */) {
+  if (auto* conn = cmd_cntx->conn(); conn /* replica context doesn't have an owner */) {
     if (VLOG_IS_ON(2)) {
       LOG(INFO) << "Got (" << conn->GetClientId() << "): " << (under_script ? "LUA " : "")
                 << cid->name() << " " << tail_args << " in dbid=" << dfly_cntx->conn_state.db_index;
@@ -1514,7 +1514,7 @@ DispatchResult Service::DispatchCommand(facade::ParsedArgs args, facade::ParsedC
       }
     }
     DCHECK(!err->status);
-    cmnd_cntx->SendError(*err);
+    cmd_cntx->SendError(*err);
     return DispatchResult::ERROR;
   }
 
@@ -1526,7 +1526,7 @@ DispatchResult Service::DispatchCommand(facade::ParsedArgs args, facade::ParsedC
   bool is_trans_cmd = cid->MultiControlKind() == CO::MultiControlKind::EXEC;
   if (dfly_cntx->conn_state.exec_info.IsCollecting() && !is_trans_cmd) {
     StoreInMultiBlock(dfly_cntx, cid, tail_args);
-    cmnd_cntx->SendSimpleString("QUEUED");
+    cmd_cntx->SendSimpleString("QUEUED");
     return DispatchResult::OK;
   }
 
@@ -1541,7 +1541,7 @@ DispatchResult Service::DispatchCommand(facade::ParsedArgs args, facade::ParsedC
   intrusive_ptr<Transaction> dist_trans;
   if (auto err = PrepareTransaction(&dist_trans, dispatching_in_multi, cid, tail_args, dfly_cntx);
       err) {
-    cmnd_cntx->SendError(*err);
+    cmd_cntx->SendError(*err);
     return DispatchResult::ERROR;
   }
 


dranikpg:
> Refactored DispatchMC implementation with **modern** C++ patterns and proper return values

Copilot tries to be friendsss with me 😆

src/server/main_service.cc
```diff
@@ -1793,110 +1802,98 @@ DispatchManyResult Service::DispatchManyCommands(std::function<facade::ParsedArg
...
-        args.emplace_back(*it);
-      }
-    }
-  } else {  // write commands.
-    if (store_opt[0]) {
-      args.emplace_back(store_opt, strlen(store_opt));
+    if (cmd.expire_ts && cmd_name == "SET") {
+      args.emplace_back("EXAT");
+      absl::numbers_internal::FastIntToBuffer(cmd.expire_ts, buffer);
+      args.emplace_back(buffer);
```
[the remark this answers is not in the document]
dranikpg: Not true, because the buffer is used once either for INCR/DECR or once for expiry which is only possible with SET command

src/facade/dragonfly_connection.cc
```diff
@@ -501,7 +501,7 @@ void Connection::AsyncOperations::operator()(ParsedCommand& cmd) {
   DVLOG(2) << "Dispatching pipeline: " << cmd.Front();
 
   ++self->local_stats_.cmds;
-  self->service_->DispatchCommand(ParsedArgs{cmd}, &cmd);
+  self->service_->DispatchCommand(ParsedArgs{cmd}, &cmd, facade::AsyncPreference::ONLY_SYNC);
```
romange, line 503: why do you need this preference? 

1. why would you want to prefer in facade async/sync?
2. How the contract looks like? can DispatchMC still run async if the preference is async?

romange: I saw your comments. but today we can run `DispatchMC` asynchronously and it is decided inside (for SET command). why facade gets to decide?

dranikpg: > why do you need this preference?

Not all callers of DispatchCommand() support async execution immediately (think of journal executor, http api, rdb_load, etc.). They won't be able to handle an async command immediately, so they will requrest only synchrous execution

> why would you want to prefer in facade async/sync?

First, the reply above. Second, you need a way to tell that there are already pending commands in the queue and you only want to run async commands (ONLY_ASYNC). If there are none, you can run any command, but we always will prefer async I guess

> why facade gets to decide?

Because its easiest for it to manage its own command queue (like dragonfly_connection does now) instead of inventing yet another protocl where it will expose the queue to server level and it will decide on execution mode

src/server/main_service.cc
```diff
@@ -881,6 +881,63 @@ string_view CommandOptName(CO::CommandOpt opt, bool enabled) {
   return "";
 }
 
+// Check CLIENT PAUSE state and block if needed
+void CheckPauseState(facade::Connection* conn, ConnectionContext* dfly_cntx, const CommandId* cid) {
```
romange, line 892: is it factored out or you changed the code as well?

src/server/main_service.cc
```diff
@@ -881,6 +881,63 @@ string_view CommandOptName(CO::CommandOpt opt, bool enabled) {
...
+    is_write |= cid->name() == "EXEC" && dfly_cntx->conn_state.exec_info.is_write;
+
+    dfly_cntx->paused = true;
+    etl.AwaitPauseState(is_write);
+    dfly_cntx->paused = false;
+  }
+}
+
+// Prepare transaction for DispatchCommand
+optional<facade::ErrorReply> PrepareTransaction(boost::intrusive_ptr<Transaction>* tx,
```
romange, line 906: same q

src/server/main_service.cc
```diff
@@ -1793,110 +1804,98 @@ DispatchManyResult Service::DispatchManyCommands(std::function<facade::ParsedArg
...
 
-  if (MemcacheParser::IsStoreCmd(cmd.type)) {
-    args.emplace_back(value);
-
-    if (store_opt[0]) {
-      args.emplace_back(store_opt, strlen(store_opt));
-    }
+    if (is_store)
+      args.emplace_back(cmd.value());
+    if (!store_opt.empty())
```
romange: nit: inside is_store block

src/server/main_service.cc
```diff
@@ -1793,110 +1804,98 @@ DispatchManyResult Service::DispatchManyCommands(std::function<facade::ParsedArg
...
-  absl::InlinedVector<string_view, 8> args;
-  args.emplace_back(cmd_name, strlen(cmd_name));
+  absl::InlinedVector<string_view, 8> args = {cmd_name};
 
-  if (!cmd.backed_args->empty()) {
-    args.emplace_back(cmd.key());
-  }
+  bool is_store = MemcacheParser::IsStoreCmd(cmd.type);
+  bool is_read = !is_store && cmd.type < MemcacheParser::QUIT;
+  if (is_store || !is_read) {
```
romange: by definition `is_store` is subset of `!is_read`.

dranikpg: True, simplified that

src/server/main_service.cc
dranikpg: No changes here

src/server/main_service.cc
dranikpg: Small simplification here, I united two `InitByArgs` calls into a common code path

dranikpg:
I only rebased

PR #6353, opened 2026-01-08
chore: show slave_repl_offset on replicas when master link is down

kostasrim:
The `INFO replication` command was not reporting `slave_repl_offset` when the master link was down, even after full sync completion. Additionally,  after a node is taken over it waits for the rest of the replicas to catch up before it shuts down. That flow however also increments the lsn because it sents PING commands to other replicas and now if any of those wish to psync to a new node they won't because they will try to "sync from the future"

**Changes:**
- Modified replication info output to include `slave_repl_offset` when `full_sync_completed && !master_link_established`
- Previously only shown when `full_sync_done` was true and master link was up
- add flag to allow preserving journal offsets after `replicaof no one`
- fix bug on journal offsets after takeover
- add tests to cover all the above

This ensures the replication offset remains visible during temporary master disconnections after initial sync completion.

addresses #4183

### DIFF
--- a/src/server/replica.h
+++ b/src/server/replica.h
@@ -187,7 +187,8 @@ class Replica : ProtocolClient {
 
   // When replica starts full sync it is set to false and true when it completes the full sync.
   // Disconnects do not reset this, so this variable is still true if the master
-  // is not connected and the state_mask_ is cleared. Furthermore, on reconnects that enter full sync
+  // is not connected and the state_mask_ is cleared.
+  // Furthermore, on reconnects that enter full sync
   // again this variable is set to false until full sync completes.
   // Therefore, we have a consistent view of the replica:
   // 1. True. Replica passed full sync even if master disconnects. In fact, once a
--- a/tests/dragonfly/replication_test.py
+++ b/tests/dragonfly/replication_test.py
@@ -3724,7 +3724,7 @@ async def test_repl_offset(df_factory):
     # 1 repl flow per proactor.
     proactors = 2
     # if `replicaof no one` on `c_replica2` does not preserve the journal offsets,
-    # then the assertion below shall fail. In that case, replicas full sync first
+    # then the assertion below shall fail. In that case, replicas perform a full sync first
     # and as there are no journal changes the slave offsets are 2 (1 per shard).
     assert info["slave_repl_offset"] > proactors
     assert info["psync_successes"] == 1


kostasrim:
@copilot write me a short description for the PR

src/server/replica.cc
```diff
@@ -661,6 +662,8 @@ error_code Replica::InitiateDflySync(std::optional<LastMasterSyncData> last_mast
     if (!exec_st_.IsRunning())
       return exec_st_.GetError();
 
+    full_sync_completed_ = true;
+
```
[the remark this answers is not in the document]
kostasrim: it will never perform a partial sync unless it first does a full sync. what described here is not the intent

tests/dragonfly/replication_test.py
```diff
@@ -3666,3 +3666,46 @@ async def test_replica_of_self(async_client):
...
+    # 4183
+    info = await c_replica2.info("replication")
+    assert info["slave_repl_offset"] > 0
+    assert info["master_link_status"] == "down"
+    assert "OK" == await c_replica2.execute_command("replicaof no one")
+    await c_replica3.execute_command(f"REPLICAOF localhost {replica2.port}")
+    await wait_for_replicas_state(c_replica3)
+
+    info = await c_replica3.info("replication")
+    assert info["slave_repl_offset"] > 3
```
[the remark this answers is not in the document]
kostasrim: added a variable

tests/dragonfly/replication_test.py
```diff
@@ -3666,3 +3666,69 @@ async def test_replica_of_self(async_client):
...
+    # Partial sync here
+    await c_replica3.execute_command(f"REPLICAOF localhost {replica2.port}")
+    # Full sync here
+    await c_replica1.execute_command(f"REPLICAOF localhost {replica2.port}")
+
+    await check_all_replicas_finished([c_replica1, c_replica3], c_replica2)
+
+    info = await c_replica3.info("replication")
+    # 1 repl flow per proactor.
+    proactors = 3
```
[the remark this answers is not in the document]
kostasrim: interesting! I was aware but I didn't want to explain this in the comment and I relaxed it

src/server/dflycmd.cc
```diff
@@ -80,11 +80,14 @@ bool ToSyncId(string_view str, uint32_t* num) {
 }
 
 bool WaitReplicaFlowToCatchup(absl::Time end_time, const DflyCmd::ReplicaInfo* replica,
-                              EngineShard* shard) {
+                              EngineShard* shard, bool with_ping = false) {
```
romange: please do not use default argument unless it's 100 places to update.

src/server/replica.h
```diff
@@ -132,6 +132,7 @@ class Replica : ProtocolClient {
     uint64_t repl_offset_sum;
     size_t psync_attempts;
     size_t psync_successes;
+    bool full_sync_completed;
```
romange: why `full_sync_done` above is not good?

src/server/replica.h
```diff
@@ -179,6 +180,8 @@ class Replica : ProtocolClient {
   size_t repl_offs_ = 0, ack_offs_ = 0;
   unsigned state_mask_ = 0;  // see State enum above.
 
+  bool full_sync_completed_ = false;
```
romange: why `(state_mask_ & R_SYNC_OK);` is not good?

kostasrim: We clear the `state_mask_` on disconnects and if master is down we won't print the offsets (if ofc we reached full sync and we got disconnected afterwards)

src/server/dflycmd.cc
romange: did you update anything?

kostasrim: yes, and I pushed the changed. I rerequested review in case you had more comments (I was kinda hopping to apply them all at once)

src/server/replica.h
romange: you did not answer here.

romange: Given that we have both, it's important to comment on how they differ as it's a bit confusing.
`full_sync_done` is false if replication was broken but full sync was completed?
when `full_sync_completed` is reset and when it's set? 
Lets add comments to both fields to understand the differrences

src/server/replica.h
romange: so let's add a comment here, explaining when `full_sync_completed_` is reset and set, and how it differs from state_mask_ and refer from the field above to here.

src/server/replica.cc
```diff
@@ -627,6 +627,7 @@ error_code Replica::InitiateDflySync(std::optional<LastMasterSyncData> last_mast
 
       DVLOG(1) << "Calling Flush on all slots " << this;
 
+      full_sync_completed_ = false;
```
romange: are you sure it's the best place to reset it?
specifically we call `last_journal_LSNs_.reset()` above. should they be updated together?

src/server/replica.h
kostasrim: > you did not answer here.

I did below. you asked the exact same question twice 😄

kostasrim: > Given that we have both, it's important to comment on how they differ as it's a bit confusing. `full_sync_done` is false if replication was broken but full sync was completed? when `full_sync_completed` is reset and when it's set? Lets add comments to both fields to understand the differrences

sounds good!

kostasrim: p.s. it's 3 variables doing very similar things. Still `R_SYNC_OK` alongside the state_mask is cleared on disconnects but it's confusing to have 3 variables around this. I am working on some more psync improvements and I will try to clear this as well.

For now I leave a coment

src/server/replica.cc
kostasrim: Yes positive it should be here. If we attempt to start full sync (which means we flush the datastore before) this value should be false. If the full sync succeeds, we flip the bool and it's set to true. 

On disconnects: 
* if it happens before we finish full sync, this is still false -> we won't print the offsets on disconnects
* if it happens after we finished full sync, this is true -> we will print the offsets (externally the datastore is in stable sync even if it's disconnected)

src/server/replica.h
kostasrim: I did a small rename. IMO I think we should just fetch the state instead of a bool, at least we squash the two variables (full_sync_done && full_sync_in_progress) into one. 

I will follow up on this separately, these 3 bools give me the ICK

src/server/replica.h
```diff
@@ -179,6 +185,16 @@ class Replica : ProtocolClient {
...
   unsigned state_mask_ = 0;  // see State enum above.
 
+  // When replica starts full sync it is set to false and true when it completes the full sync.
+  // Disconnects do not reset this, so this variable is still true if the master
+  // is not connected and the state_ is cleared. Furthermore, on reconnects that enter full sync
+  // again this variable is set to false until full sync completes.
+  // Therefore, we have a consistent view of the replica:
+  // 1. True. Replica passed full sync even if master disconnects. In fact, once a
+  // node reached stable, the deltas from journal are the only missing items.
+  // 2. False. Replica has not passed full sync or a disconnect started full sync again.
```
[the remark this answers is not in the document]
kostasrim: nope that's not the case. it means that we skipped full sync not that completed partial sync.

src/server/dflycmd.cc
```diff
@@ -80,11 +80,14 @@ bool ToSyncId(string_view str, uint32_t* num) {
 }
 
 bool WaitReplicaFlowToCatchup(absl::Time end_time, const DflyCmd::ReplicaInfo* replica,
-                              EngineShard* shard) {
+                              EngineShard* shard, bool with_ping) {
   // We don't want any writes to the journal after we send the `PING`,
   // and expirations could ruin that.
```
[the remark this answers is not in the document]
kostasrim: It's clear IMO. The comment is though made me think what happens if expire elements between the two steps. Then parital sync will fail because we try to sync from the future. Corner case, I will investigate at some point in the future (non critical)

Opened  #6416

src/server/dflycmd.cc
```diff
@@ -80,11 +80,14 @@ bool ToSyncId(string_view str, uint32_t* num) {
...
   namespaces->GetDefaultNamespace().GetDbSlice(shard->shard_id()).SetExpireAllowed(false);
-  shard->journal()->RecordEntry(0, journal::Op::PING, 0, 0, nullopt, {});
+
+  if (with_ping) {
+    shard->journal()->RecordEntry(0, journal::Op::PING, 0, 0, nullopt, {});
```
romange, line 92: Can you add a comment why PING is needed in this context?

src/server/dflycmd.cc
```diff
@@ -549,7 +552,7 @@ void DflyCmd::TakeOver(CmdArgList args, CommandContext* cmd_cntx) {
   if (*status == OpStatus::OK) {
     dfly::SharedLock lk{replica_ptr->shared_mu};
     auto cb = [replica_ptr = replica_ptr, end_time, &catchup_success](EngineShard* shard) {
-      if (!WaitReplicaFlowToCatchup(end_time, replica_ptr.get(), shard)) {
+      if (!WaitReplicaFlowToCatchup(end_time, replica_ptr.get(), shard, true)) {
```
romange, line 559: similarly add a comment here why  ping is needed here and comment below why it **should be avoided** below

src/server/replica.h
```diff
@@ -132,6 +132,12 @@ class Replica : ProtocolClient {
     uint64_t repl_offset_sum;
     size_t psync_attempts;
     size_t psync_successes;
+    // We can't rely on full_sync_done or full_sync_in_progress because
+    // on disconnects the replica state mask is cleared. We use this variable
+    // to track if the replica reached full sync. When master disconnects,
+    // we use this variable to print the journal offsets in info command even
+    // when the link is down.
```
romange:
```suggestion
    // when the link is down. It's reset whenever a full sync is initiated again.
```

src/server/server_family.cc
```diff
@@ -146,6 +146,8 @@ ABSL_FLAG(string, availability_zone, "",
           "server availability zone, used by clients to read from local-zone replicas");
 
 ABSL_FLAG(bool, keep_legacy_memory_metrics, true, "legacy metrics format");
+ABSL_FLAG(bool, replicaof_no_one_start_journal, false,
```
romange: Can you explain about this flag? does it break compatibility? affects performance?
in other words, what's the reason we need a flag in the first place?

romange:
LGTM + minor comments

src/server/server_family.cc
kostasrim: My rationale behind the flag was to support 2 use cases for `replicaof no one`:

* One when after `replicaof no one` we reset the journal offsets to 0. If this node is not shut down and becomes a master again, its journal offsets will start from 0. If for some reason we attach a replica, it will never partially sync.
* One to keep the `shard lsn's` even after `replicaof no one`. So for example, old master has 1 proactor with journal offset at 100. After `replicaof no one` this node starts at offset `100` so if I make any other node a replica of this one, it will partial sync. Also the INFO command will show offsets `consistent` with the `rest of the nodes in the setup`.

No performance difference whatsoever, it just controls if the journal resumes from the last master LSN's or resets to 0. This is the default behaviour on `TAKEOVER` but `REPLICAOF NO ONE` is not takeover flow and blindly assuming that offsets should start from the last master would be a mistake.

romange: but we should configure this flag in production? it's either `true` or `false`, and  you give here two choices but we will have to choose one and stick with it.

kostasrim: I flipped it to true. We can set it in our production to `false` for a few weeks before we slowly migrate

src/server/replica.h
```diff
@@ -179,6 +185,17 @@ class Replica : ProtocolClient {
   size_t repl_offs_ = 0, ack_offs_ = 0;
   unsigned state_mask_ = 0;  // see State enum above.
 
+  // When replica starts full sync it is set to false and true when it completes the full sync.
+  // Disconnects do not reset this, so this variable is still true if the master
+  // is not connected and the state_mask_ is cleared.
+  // Furthermore, on reconnects that enter full sync
+  // again this variable is set to false until full sync completes.
```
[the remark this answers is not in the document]
kostasrim: it's clear.

PR #6358, opened 2026-01-09
fix(server): Set error type when returning -NOGROUP error

### LINKED ISSUE #6357: Specify type when handling -NOGROUP errors
When we are sending back -NOGROUP error we don't specify error type. 

Result is that `INFO ERRORSTATS` is filled with  unique string, for each error,  instead of aggregating nogroup error.

mkaruza:
Specify error type when NOGROUP error is returned. Currently every time we return NOGROUP error it will be treated as unique error type. Introduce type that will aggregate all different errors under one type.

Fixes #6357

### DIFF
--- a/src/server/stream_family.cc
+++ b/src/server/stream_family.cc
@@ -222,13 +222,16 @@ string StreamIdRepr(const streamID& id) {
   return absl::StrCat(id.ms, "-", id.seq);
 };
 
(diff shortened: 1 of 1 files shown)


dranikpg:
Please change `NoGroupOrKey` to return `facace::ErrorReply` that includes the proper error type

dranikpg (dismissed):

PR #6359, opened 2026-01-09
fix: memory calculation for OAHSet

BorysTheDev:
I have fixed a couple of bugs in the memory calculations for OAHSet


src/core/oah_entry.cc
```diff
@@ -147,9 +147,9 @@ void OAHEntry::SetExpiry(uint32_t at_sec) {
 // TODO refactor, because it's inefficient
 std::optional<uint32_t> OAHEntry::Find(std::string_view str, uint64_t ext_hash,
                                        uint32_t capacity_log, uint32_t shift_log,
-                                       uint32_t* set_size, uint32_t time_now) {
+                                       uint32_t* set_size, size_t* alloc_used, uint32_t time_now) {
```
dranikpg, line 150: having a more extensive return value that reports deltas might be nicer than juggling around wiht pointers

BorysTheDev: I will move this function into OAHSet, just didn't have time for nice refactoring

PR #6361, opened 2026-01-09
fix(set): SPOP random member even if it contains null bytes

### LINKED ISSUE #6355: SPOP fails to remove elements and returns truncated binary data
I have encountered a severe issue where the SPOP command in DragonflyDB fails to behave atomically. Specifically:

    It returns an element, but does not remove it from the Set (SCARD remains unchanged).

    The returned binary data (Python Pickle object) is truncated, returning only the first few bytes instead of the full payload.

This behavior makes SPOP act like a broken version of SRANDMEMBER.
Steps to Reproduce

The following Python script demonstrates both the failure to remove the item and the data corruption (truncation):
Python

```
r.delete('test')
pipe = r.pipeline()
for i in range(0,10):
    obj={'cur':'sdisox','data':np.array(list(range(0,100))),'i':i}
    pipe.sadd("test", pickle.dumps(obj, protocol=4))
pipe.execute()

print(r.scard('test'))
popped_data=r.spop('test')
print(f"Returned data length: {len(popped_data)} bytes")
print(r.scard('test'))
```

Expected Behavior

    SPOP should return the full 800+ bytes of the pickled object.

    SCARD should return 9 after the operation.

Actual Behavior

    SPOP returns only 5 bytes (e.g., b'\x80\x04\x95\xcb\x03').

    SCARD still returns 10.

    Subsequent SPOP calls continue to return the same 5-byte truncated fragment without clearing the set.

mkaruza:
If member contains null bytes we will not read complete member identifier and it will not be removed from set.

Fixes #6355

### DIFF
--- a/src/server/set_family.cc
+++ b/src/server/set_family.cc
@@ -279,8 +279,10 @@ StringVec RandMemberStrSetPicky(StringSet* strset, size_t count) {
   picks.reserve(count);
 
   size_t tries = 0;
-  while (picks.size() < count && tries++ < count * 2)
-    picks.insert(picks.end(), string{*strset->GetRandomMember()});
+  while (picks.size() < count && tries++ < count * 2) {
+    auto member = *strset->GetRandomMember();
+    picks.insert(picks.end(), {member, sdslen(member)});
+  }
 
   if constexpr (is_same_v<StringVec, C>)
     return picks;
--- a/tests/dragonfly/set_test.py
+++ b/tests/dragonfly/set_test.py
@@ -23,3 +23,23 @@ async def test_sscan_regression(df_factory: DflyInstanceFactory):
     # Takes 3 seconds
     res = await client.execute_command("SLOWLOG GET 100")
     assert res == []
+
+
+@pytest.mark.asyncio
+async def test_spop_with_null_byte_members(df_factory: DflyInstanceFactory):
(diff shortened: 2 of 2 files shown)


tests/dragonfly/set_test.py
```diff
@@ -23,3 +23,23 @@ async def test_sscan_regression(df_factory: DflyInstanceFactory):
     # Takes 3 seconds
     res = await client.execute_command("SLOWLOG GET 100")
     assert res == []
+
+
+@pytest.mark.asyncio
+async def test_spop_with_null_byte_members(df_factory: DflyInstanceFactory):
```
dranikpg, line 29: could've been a c++ test

src/server/set_family.cc
```diff
@@ -279,8 +279,10 @@ StringVec RandMemberStrSetPicky(StringSet* strset, size_t count) {
   picks.reserve(count);
 
   size_t tries = 0;
-  while (picks.size() < count && tries++ < count * 2)
-    picks.insert(picks.end(), string{*strset->GetRandomMember()});
+  while (picks.size() < count && tries++ < count * 2) {
+    auto member = *strset->GetRandomMember();
+    picks.insert(picks.end(), {member, sdslen(member)});
+  }
```
dranikpg, line 285: good catch

dranikpg (approved):

Issue #6265, opened 2025-12-22
Replace RedisParser with hiredis for tests


BorysTheDev:
we want to use all client code with hiredis instead of RedisParser. The first step is tests, and then we can do the same for replication and migration

BorysTheDev:
fixed by: https://github.com/dragonflydb/dragonfly/pull/6588

PR #6362, opened 2026-01-09
test: use hiredis for reply_builder_test

BorysTheDev:
related to: https://github.com/dragonflydb/dragonfly/issues/6265
Replaced RedisParser with Hiredis in reply_builder_test

### DIFF
--- a/src/facade/resp_parser.h
+++ b/src/facade/resp_parser.h
@@ -175,8 +175,10 @@ template <class T> std::optional<T> RESPObj::As() const {
   }
   if constexpr (std::is_constructible_v<T, std::string_view>) {
     if (reply_->type == REDIS_REPLY_STRING || reply_->type == REDIS_REPLY_ERROR ||
-        reply_->type == REDIS_REPLY_STATUS || reply_->type == REDIS_REPLY_NIL) {
+        reply_->type == REDIS_REPLY_STATUS) {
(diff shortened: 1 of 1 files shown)


src/facade/reply_builder_test.cc
```diff
@@ -489,9 +449,9 @@ TEST_F(RedisReplyBuilderTest, NullBulkString) {
   ASSERT_TRUE(NoErrors());
   ASSERT_EQ(str(), kNullBulkString);
   auto parsing_output = Parse();
-  ASSERT_TRUE(parsing_output.Verify(SinkSize()));
-  ASSERT_TRUE(parsing_output.IsNull());
-  ASSERT_THAT(parsing_output.args, ElementsAre(ArgType(RespExpr::NIL)));
+
+  ASSERT_EQ(parsing_output->GetType(), RESPObj::Type::NIL);
```
romange: You should introduce utilities that translate new  abstractions to RespExpr and keep the tests mostly intact

BorysTheDev: I've already done it, why do we need RespExpr at all?

romange: so that you won't change testing code across 100 tests.

vyavdoshenko (commented):
lgtm

PR #6363, opened 2026-01-09
chore(server): Simplify InvokeCmd

dranikpg:
We use `last_error` from reply builder just to log it. And we call `ConsumeLastError()` everywhere just because we're afraid to trigger our own DCHECK that checks for the error to be consumed when we start the command. But we don't have to assert anything - we can just throw it away. 

The inconvenience comes from the fact that we use SendError for both command errors and state errors from generic code in DispatchCommand, so those conflict.

Refactor some code in main_service to make it simpler

### DIFF
--- a/src/server/main_service.cc
+++ b/src/server/main_service.cc
@@ -1627,6 +1627,7 @@ DispatchResult Service::InvokeCmd(CmdArgList tail_args, CommandContext* cmd_cntx
 
   ServerState::tlocal()->RecordCmd(cntx->has_main_or_memcache_listener);
   TrackIfNeeded(cmd_cntx);
+  auto* tx = cmd_cntx->tx;
 
 #ifndef NDEBUG
   // Verifies that we reply to the client when needed.
@@ -1652,7 +1653,6 @@ DispatchResult Service::InvokeCmd(CmdArgList tail_args, CommandContext* cmd_cntx
     }
   }
 
-  auto* tx = cmd_cntx->tx;
   if ((!tx && cid->name() != "MULTI") || (tx && !tx->IsMulti())) {
     // Each time we execute a command we need to increase the sequence number in
     // order to properly track clients when OPTIN is used.


src/server/main_service.cc
```diff
@@ -881,6 +871,41 @@ string_view CommandOptName(CO::CommandOpt opt, bool enabled) {
...
+
+  // TODO: There is a bug here that we track all arguments instead of tracking only keys.
+  auto& db_slice = slice_args.GetDbSlice();
+  for (auto key : args)
+    db_slice.TrackKey(conn_ref, key);
+
+  return OpStatus::OK;
+}
+
+void TrackIfNeeded(CommandContext* cmnd_cntx) {
```
dranikpg, line 891: Moved without changes

src/server/main_service.cc
```diff
@@ -313,28 +313,18 @@ std::string MakeMonitorMessage(const ConnectionContext* cntx, const CommandId* c
...
-  }
-  VLOG(2) << "Thread " << ProactorBase::me()->GetPoolIndex() << " sending monitor message '" << msg
-          << "' for " << monitors.size();
-
-  for (auto monitor_conn : monitors) {
-    monitor_conn->SendMonitorMessageAsync(msg);
-  }
-}
-
 void DispatchMonitor(ConnectionContext* cntx, const CommandId* cid, CmdArgList tail_args) {
```
dranikpg, line 316: Simplified

romange:
would not be it simpler if we move "last_error" to CommandContext?
This way, there is nothing to consume as it becomes a command state (which btw where it belongs) ?

For async dispatch this won't work I suppose.

dranikpg:
> would not be it simpler if we move "last_error" to CommandContext?
> This way, there is nothing to consume as it becomes a command state (which btw where it belongs) ?
> For async dispatch this won't work I suppose.

Maybe, except for the where it belongs part 🙂 What surely isn't wrong is reducing usage of SendError calls and making it a value where it can be one. At the end such things are easier to handle

PR #6364, opened 2026-01-10
feat(server): Async handler, first try

dranikpg:
Main problem around async execution is keeping alive:
1. The transaction
2. The handler (because Transaction takes it by reference)
3. The value that the handler writes into

### DIFF
--- a/src/server/generic_family.cc
+++ b/src/server/generic_family.cc
@@ -1080,33 +1080,42 @@ io::Result<int32_t, string> ParseExpireOptionsOrReply(const CmdArgList args) {
   return flags;
 }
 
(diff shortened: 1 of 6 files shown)


src/server/generic_family.cc
```diff
@@ -1080,33 +1080,57 @@ io::Result<int32_t, string> ParseExpireOptionsOrReply(const CmdArgList args) {
   return flags;
 }
 
-void DeleteGeneric(CmdArgList args, CommandContext* cmd_cntx, bool async) {
-  atomic_uint32_t result{0};
+// Keeps shard callback lambda and value in one struct
```
romange, line 1083:
```suggestion
// Packages a lambda and value in one struct. 
// Used to keep a shard callback together with its data dependencies on heap.
```

PR #6384, opened 2026-01-13
chore: encapsulate tx object within CommandContext

romange:
1. Just syntactic changes. Added an accessor  CommandContext::tx() and moved the pointer to private scope.

2. Renamed cmnd_cntx to cmd_cntx everywhere as sometimes it was former and sometimes the latter. This increases consistency across the codebase.


romange:
how is your comment related to this PR? :)

romange:
btw, my goal here is potentially to reuse transactions as I saw it their allocations/deallocation as a somewhat high on prof map.

dranikpg:
> how is your comment related to this PR? :)

I assume you do preparation or refactoring for async commands, as you change the whole code base. Obviously you don't need it for current sync commands, but likely we won't use it in final async too

PR #6386, opened 2026-01-13
chore: move CommandContext cid to private scope

romange:
No functional changes.


src/server/command_registry.cc
```diff
@@ -129,38 +129,6 @@ constexpr int32_t kLatencyHistogramPrecision = 2;
 
 }  // namespace
 
-void CommandContext::RecordLatency(facade::ArgSlice tail_args) const {
```
romange, line 132: moved to conn_context where CommandContext is defined.

PR #6392, opened 2026-01-14
chore: update AGENTS.md

romange:
Make it more concise and not specific to some agent. Create a softlink for CLAUDE.md

This seems to work better on my experience with claude cli.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -31,7 +31,8 @@
 
 1. ✅ **Read Before Edit** - Always read files before modifying
 2. ✅ **Use Correct Build Commands** - See [Quick Command Reference](#quick-command-reference) below
-3. ✅ **Test After Changes** - Run tests immediately: `ctest -V -L DFLY`
+3. ✅ **Test After Changes** - Build and run a relevant unit test -
+   `ninja <unit_test> && ./unit_test`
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -1,96 +1,106 @@
...
-2. **Test After Changes** - Run tests immediately: `ctest -V -L DFLY` / `pytest dragonfly/`
-3. **Init Submodules** - `git submodule update --init --recursive`
-4. **Format Code** - `pre-commit run --files <files>`
-5. **Follow Architecture** - See [Universal Patterns](#universal-architecture-patterns) below
+**MANDATORY - Always Follow This Order:**
 
----
+1. ✅ **Read Before Edit** - Always read files before modifying
+2. ✅ **Use Correct Build Commands** - See [Quick Command Reference](#quick-command-reference) below
+3. ✅ **Test After Changes** - Run tests immediately: `ctest -V -L DFLY`
```
vyavdoshenko: It looks like overkill.
In my opinion, it's better to give an instruction to run proper tests.

vyavdoshenko (dismissed):

AGENTS.md
romange: I agree I missed that. Initially it wrote to run pytests as well and I removed it :)

PR #6393, opened 2026-01-14
chore: Update CommandContext c'tor

romange:
Reduce it to set fields that usually do not change through the lifecycle of the CommandContext and reuse the same CommandContext instance in batch scenarios.

The change is done to reflect the actual state changes we make on this object.

### DIFF
--- a/src/server/debugcmd.cc
+++ b/src/server/debugcmd.cc
@@ -1588,8 +1588,8 @@ void DebugCmd::Segments(CmdArgList args, CommandContext* cmd_cntx) {
 }
 
 void DebugCmd::DoPopulateBatch(const PopulateOptions& options, const PopulateBatch& batch) {
-  boost::intrusive_ptr<Transaction> local_tx =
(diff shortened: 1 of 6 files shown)


src/server/debugcmd.cc
```diff
@@ -1629,8 +1631,7 @@ void DebugCmd::DoPopulateBatch(const PopulateOptions& options, const PopulateBat
       stub_tx->MultiSwitchCmd(cid);
       crb.SetReplyMode(ReplyMode::NONE);
       stub_tx->InitByArgs(cntx_->ns, cntx_->conn_state.db_index, args_span);
-
-      CommandContext cmd_cntx{cid, stub_tx.get(), &crb, cntx_};
+      cmd_cntx.UpdateCid(cid);
```
dranikpg, line 1634: what was the costly part of creating it?

romange: it's not costly at all. I would like to move (possibly) the tx object ownership to CommandContext and this c'tor made things complicated for me.

PR #6397, opened 2026-01-15
chore: Clean up DispatchCommand a bit more.

romange:
reduce number of arguments in PrepareTransaction and move more logic there. Number of lines of code stayed the same because I added more DCHECKS to the code.

### DIFF
--- a/src/server/main_service.cc
+++ b/src/server/main_service.cc
@@ -907,9 +907,9 @@ void CheckPauseState(facade::Connection* conn, ConnectionContext* dfly_cntx, con
 // Return value:
 //   first  - newly created top-level transaction (or nullptr if none).
 //   second - result: overall status of preparation.
-pair<intrusive_ptr<Transaction>, DispatchResult> PrepareTransaction(const CommandId* cid,
-                                                                    ArgSlice tail_args,
-                                                                    CommandContext* cmd_ctx) {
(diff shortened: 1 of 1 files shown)


src/server/main_service.cc
```diff
@@ -902,39 +902,50 @@ void CheckPauseState(facade::Connection* conn, ConnectionContext* dfly_cntx, con
...
   if (init) {
-    auto* tx = dfly_cntx->transaction;
-    OpStatus status = tx->InitByArgs(dfly_cntx->ns, dfly_cntx->conn_state.db_index, tail_args);
-    if (status != OpStatus::OK)
-      return facade::ErrorReply{status};
+    DCHECK(cmd_ctx->tx());
+    if (auto st =
+            cmd_ctx->tx()->InitByArgs(dfly_cntx->ns, dfly_cntx->conn_state.db_index, tail_args);
+        st != OpStatus::OK) {
+      cmd_ctx->SendError(StatusToMsg(st));
```
dranikpg: Nooo 😭 😭 😭 😭  The whole reason I created this was to de-couple reply logic from command logic and have the function be isolted. Single responsibility.

dranikpg: I support returning the new transaction by value, but lets please also return the error as a value

dranikpg (dismissed):

Issue #6345, opened 2026-01-08
v1.36.0 cannot find S3 snapshots on S3-compatible endpoint (works on v1.35.1)


Ttsartsidze:
### Summary
After upgrading DragonflyDB from v1.35.1 to v1.36.0, snapshot auto-load from S3 stopped working.
With the same S3 bucket/prefix and credentials, v1.35.1 successfully loads the latest snapshot, while v1.36.0 reports "No snapshot found".

### Environment
- DragonflyDB:
  - works: v1.35.1
  - fails: v1.36.0 (df-v1.36.0-65c71bd7cfe37befb119217045c0678393e0904d)
- OS: Linux 5.14.x
- Deployment: Kubernetes
- S3: S3-compatible endpoint (NOT AWS)
  - endpoint: s3.example.internal
  - region: us-east-1
  - https: true
  - credentials: provided via environment variables

### Configuration / Flags
--s3_endpoint=s3.example.internal
--dir=s3://example-bucket/example-prefix/
--snapshot_cron=0 0 * * *
--s3_sign_payload=false
--proactor_threads=2

### Steps to Reproduce
1. Ensure a snapshot exists under:
   s3://example-bucket/example-prefix/
   (e.g. dump-2026-01-08T06:15:36-summary.dfs)
2. Start DragonflyDB v1.35.1 with the flags above -> snapshot loads successfully.
3. Start DragonflyDB v1.36.0 with the SAME flags above -> snapshot is not found.

### Expected Behavior
v1.36.0 should discover and load the latest snapshot from the configured S3 bucket/prefix, same as v1.35.1.

### Actual Behavior
v1.36.0 prints "Load snapshot: No snapshot found" even though snapshots exist and v1.35.1 loads them.

#### v1.35.1 (works)
I... Load snapshot: Searching for snapshot in S3 path: s3://example-bucket/example-prefix
I... Loading s3://example-bucket/example-prefix/dump-2026-01-08T06:15:36-summary.dfs

#### v1.36.0 (fails)
I... Creating AWS S3 client; region=us-east-1; https=true; endpoint=s3.example.internal
I... aws: loaded credentials; provider=environment
I... Load snapshot: Searching for snapshot in S3 path: s3://example-bucket/example-prefix/
W... Load snapshot: No snapshot found

### Additional Notes
- Adding/removing trailing slash in --dir does not change the behavior on v1.36.0.
- Works reliably on v1.35.1 with the same configuration.

[...]

PR #6400, opened 2026-01-16
fix: s3 snapshot saving

BorysTheDev:
fixes: #6345

The problem was in the incorrect escaping of the {} symbols during the snapshot path processing
Also, added fix for the case when the bucket contains some directory in the path

### DIFF
--- a/src/server/detail/snapshot_storage.cc
+++ b/src/server/detail/snapshot_storage.cc
@@ -664,7 +664,8 @@ AwsS3SnapshotStorage::ListObjects(std::string_view bucket_name, std::string_view
     Aws::S3::Model::ListObjectsV2Request request;
     request.SetBucket(std::string(bucket_name));
     if (!prefix.empty()) {
-      // Ensure prefix ends with '/' to treat it as a directory-like namespace and avoid matching objects with similar prefix names.
+      // Ensure prefix ends with '/' to treat it as a directory-like namespace and avoid
+      // matching objects with similar prefix names.
       if (prefix.back() == '/') {
         request.SetPrefix(std::string(prefix));
       } else {


src/server/detail/snapshot_storage.cc
```diff
@@ -63,7 +63,8 @@ const int kRdbWriteFlags = O_CREAT | O_WRONLY | O_TRUNC | O_CLOEXEC | O_DIRECT;
 
 std::string EscapeRegex(string_view input) {
   // List of regex special characters that need escaping
-  constexpr std::string_view chars{"\\.^$|?*+()[]{}"};
+  // We don't escape "{}" since we use them for our own placeholders.
+  constexpr std::string_view chars{"\\.^$|?*+()[]"};
```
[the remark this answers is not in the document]
BorysTheDev: @copilot add pytest in the current branch to the snapshot_test.py file. As an example,  use "test_s3_snapshot" You should cover "{Y}" "{m}" "{d}" "{timestamp}" patterns in the snapshot name. Also, add a comment in the test about the default s3 bucket path 'os.environ["DRAGONFLY_S3_BUCKET"]', that we use as default, and where it can be changed for tests on github.

BorysTheDev:
https://github.com/dragonflydb/dragonfly/actions/runs/21134922471

tests/dragonfly/snapshot_test.py
```diff
@@ -408,6 +408,47 @@ async def test_s3_snapshot(async_client, tmp_dir):
...
+        start_capture = await DebugPopulateSeeder.capture(async_client)
+        instance.stop()
+
+        new_instance = df_factory.create()
+        new_instance.start()
+        new_async_client = new_instance.client()
+
+        await wait_available_async(new_async_client)
+
+        assert await DebugPopulateSeeder.capture(new_async_client) == start_capture
```
romange, line 443: What do you test here?

src/server/detail/snapshot_storage.cc
```diff
@@ -662,8 +663,13 @@ AwsS3SnapshotStorage::ListObjects(std::string_view bucket_name, std::string_view
   do {
     Aws::S3::Model::ListObjectsV2Request request;
     request.SetBucket(std::string(bucket_name));
-    if (!prefix.empty())
-      request.SetPrefix(std::string(prefix));
+    if (!prefix.empty()) {
+      if (prefix.back() == '/') {
+        request.SetPrefix(std::string(prefix));
+      } else {
+        request.SetPrefix(std::string(prefix) + '/');
```
romange, line 672: Add a comment here as it is required for minio?

tests/dragonfly/snapshot_test.py
BorysTheDev: I test that after DF restart I can load the same snapshot

src/server/detail/snapshot_storage.cc
BorysTheDev: It's a requirement AWS API, I just found out this when I wrote the test

src/server/detail/snapshot_storage.cc
```diff
@@ -662,8 +663,14 @@ AwsS3SnapshotStorage::ListObjects(std::string_view bucket_name, std::string_view
   do {
     Aws::S3::Model::ListObjectsV2Request request;
     request.SetBucket(std::string(bucket_name));
-    if (!prefix.empty())
-      request.SetPrefix(std::string(prefix));
+    if (!prefix.empty()) {
+      // according to S3 docs, prefix should end with '/' to list objects in a directory
```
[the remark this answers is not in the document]
romange: +1 to this comment

src/server/detail/snapshot_storage.cc
romange: i believe it's not required for s3, is it?

tests/dragonfly/snapshot_test.py
```diff
@@ -408,6 +408,49 @@ async def test_s3_snapshot(async_client, tmp_dir):
...
+    # we can load the snapshot from s3 correctly.
+    try:
+        instance = df_factory.create()
+        instance.start()
+        async_client = instance.client()
+        seeder = DebugPopulateSeeder(key_target=10_000)
+        await seeder.run(async_client)
+
+        # Force a snapshot now rather than waiting for cron so reload is deterministic
+
```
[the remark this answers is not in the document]
romange: yeah, maybe rephrase

src/server/detail/snapshot_storage.cc
BorysTheDev: https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/model/ListObjectsV2Request.html#prefix()

romange (approved):

PR #6407, opened 2026-01-18
chore: tx clean-ups

romange:
1. Remove unused declarations
2. Remove redundant object copy when using Range transformations.
3. Increase provided bufffers size

### DIFF
--- a/src/server/tx_base.h
+++ b/src/server/tx_base.h
@@ -60,7 +60,7 @@ struct KeyIndex {
   }
 
   auto Range(const cmn::ArgSlice& args) const {
-    return base::it::Transform([&args](unsigned idx) { return args[idx]; }, Range());
+    return base::it::Transform([args](unsigned idx) { return args[idx]; }, Range());
   }
 
  public:


src/server/tx_base.h
```diff
@@ -59,8 +59,8 @@ struct KeyIndex {
     return base::it::Range(*this, KeyIndex{end, end, step, std::nullopt});
   }
 
-  auto Range(cmn::ArgRange args) const {
-    return base::it::Transform([args](unsigned idx) { return args[idx]; }, Range());
+  auto Range(const cmn::ArgSlice& args) const {
+    return base::it::Transform([&args](unsigned idx) { return args[idx]; }, Range());
```
dranikpg: arg slice is a span - this is like passing a string_view by reference, or a pointer by pointer - dangerous because those types are always passed by value (so their actual lifetime is limited) and its just an unnecessary indirection

romange: I agree but I saw a minor hotspot around these calls inside transactional code. I will revert the lambda capture for now.

dranikpg: the base iterator can also be a bit heavy

PR #6417, opened 2026-01-19
chore: wrap list operations within ListWrapper

romange:
This is done to make the list logic neutral with respect to any encoding we would implement inside ListWrapper. Currently only QList is used but in the future, we could provide listpack or tiered storage.

The motivation - we can use listpack directly for tiny lists bypassing qlist layers.
When qlist is created for larger lists - we can also make it larger and have more metdata in it and it won't matter in relative terms. Specifically, we can add a state that will track tiering.

### DIFF
--- a/src/core/qlist.h
+++ b/src/core/qlist.h
@@ -39,21 +39,24 @@ class QList {
    * items). recompress: 1 bit, bool, true if node is temporary decompressed for usage.
    * attempted_compress: 1 bit, boolean, used for verifying during testing.
    * dont_compress: 1 bit, boolean, used for preventing compression of entry.
-   * extra: 9 bits, free for future use; pads out the remainder of 32 bits
    * */
 
   struct Node {
     Node* prev;
     Node* next;
     unsigned char* entry;
-    size_t sz;                           /* entry size in bytes */
-    unsigned int count : 16;             /* count of items in listpack */
-    unsigned int encoding : 2;           /* RAW==1 or LZF==2 */
-    unsigned int container : 2;          /* PLAIN==1 or PACKED==2 */
-    unsigned int recompress : 1;         /* was this node previous compressed? */
-    unsigned int attempted_compress : 1; /* node can't compress; too small */
-    unsigned int dont_compress : 1;      /* prevent compression of entry that will be used later */
-    unsigned int extra : 25;             /* more bits to steal for future usage */
+    size_t sz : 48;    /* entry size in bytes */
+    size_t count : 16; /* count of items in listpack */
+
+    uint16_t encoding : 2;           /* RAW==1 or LZF==2 */
+    uint16_t container : 2;          /* PLAIN==1 or PACKED==2 */
+    uint16_t recompress : 1;         /* was this node previous compressed? */
+    uint16_t attempted_compress : 1; /* node can't compress; too small */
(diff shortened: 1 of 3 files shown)


src/server/list_family.cc
```diff
@@ -187,7 +284,7 @@ OpResult<string> OpMoveSingleShard(const OpArgs& op_args, string_view src, strin
 
   QList* destql_v2 = CreateOrGet(op_args, dest, dest_res.is_new, &dest_res.it->second);
 
-  val = srcql_v2->Pop(ToWhere(src_dir));
+  val = srcql_v2.Pop(ToWhere(src_dir));
   destql_v2->Push(val, ToWhere(dest_dir));
```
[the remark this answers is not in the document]
romange: it's fine, I will handle creation flows in the subsequent PRs

src/core/qlist.h
```diff
@@ -39,21 +39,24 @@ class QList {
...
     unsigned char* entry;
-    size_t sz;                           /* entry size in bytes */
-    unsigned int count : 16;             /* count of items in listpack */
-    unsigned int encoding : 2;           /* RAW==1 or LZF==2 */
-    unsigned int container : 2;          /* PLAIN==1 or PACKED==2 */
-    unsigned int recompress : 1;         /* was this node previous compressed? */
-    unsigned int attempted_compress : 1; /* node can't compress; too small */
-    unsigned int dont_compress : 1;      /* prevent compression of entry that will be used later */
-    unsigned int extra : 25;             /* more bits to steal for future usage */
+    size_t sz : 48;    /* entry size in bytes */
```
romange, line 48: I did not increase the node size, just reordered the bits to have more of them freed.

src/server/list_family.cc
```diff
@@ -69,16 +69,114 @@ using time_point = Transaction::time_point;
 
 namespace {
 
-QList* GetQLV2(const PrimeValue& mv) {
-  return (QList*)mv.RObjPtr();
+struct ListWrapper {
+  unsigned encoding = 0;
+  union {
+    QList* ql = nullptr;
+    uint8_t* listpack;
```
romange, line 76: listpack and encoding are currently not used in this PR

src/server/list_family.cc
```diff
@@ -69,16 +69,114 @@ using time_point = Transaction::time_point;
...
+ListWrapper GetLW(const PrimeValue& mv) {
+  return ListWrapper{.encoding = mv.Encoding(), .ql = static_cast<QList*>(mv.RObjPtr())};
 }
 
+vector<uint32_t> ListWrapper::Pos(string_view element, int rank, uint32_t count, uint32_t max_len,
+                                  QList::Where where) const {
+  DCHECK_GT(rank, 0);
+  vector<uint32_t> matches;
+
+  auto it = ql->GetIterator(where);
```
romange, line 131: moved from below

src/server/list_family.cc
```diff
@@ -69,16 +69,114 @@ using time_point = Transaction::time_point;
...
+      }
+    }
+    index++;
+  }
+  return matches;
+}
+
+unsigned ListWrapper::Remove(string_view elem, unsigned count, QList::Where where) {
+  unsigned removed = 0;
+  int64_t ival;
```
romange, line 151: moved from below

PR #6423, opened 2026-01-20
refactor(facade): decouple pipeline traffic from administrative command queue (phase 1)

glevkovich:
This phase 1 PR refactors the connection dispatch logic to separate
high-volume pipeline traffic from administrative control messages.

Major Changes:
********************
- Queue Separation: Introduced parsed_cmd_q_ (linked list) for pipeline
commands, keeping dispatch_q_ (deque) for Admin/Control messages.

- Dispatch Policy Update: Admin messages (in dispatch_q_) generally take
absolute priority over pipeline commands. Exception: Migration Requests.
 If a migration request arrives while parsed_cmd_q_ is not empty,
 the dispatch fiber now temporarily
prioritizes the pipeline. This ensures the pipeline drains completely
before migration starts, preventing a potential livelock where the
migration blocks the IO loop but cannot execute because the pipeline is
busy.

Stats & Accounting:
********************
-Fixed memory accounting drift in ReleaseParsedCommand regarding
MessageHandle overhead.

-Included pending pipeline bytes in GetMemoryUsage for accurate RSS
reporting.

-Fixed latency calculation to account for time spent in the dispatch
queue.

-Safety: Added pointer safety checks in AsyncFiber and SquashPipeline
to prevent Use-After-Free scenarios when queues empty out.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1486,7 +1486,7 @@ void Connection::SquashPipeline() {
       stats_->pipelined_wait_latency += CycleClock::ToUsec(start - current->parsed_cycle);
     }
 
-    ReleaseParsedCommand(current, true /* is_pipelined */);
+    ReleaseParsedCommand(current, result.account_in_stats /* is_pipelined */);
     current = next;
   }
   parsed_head_ = current;
@@ -1637,14 +1637,19 @@ void Connection::AsyncFiber() {
       MessageHandle msg;
 
       // If the front message is a Migration Request, but we still have pipeline data
-      // (parsed_head_), we must block the migration and run the pipeline first.
-      bool is_blocked_migration{
+      // (parsed_head_), we must block the migration and process the pipeline messages first.
+      bool is_migration_req =
           !dispatch_q_.empty() &&
-          std::holds_alternative<MigrationRequestMessage>(dispatch_q_.front().handle) &&
-          (parsed_head_ != nullptr)};
+          std::holds_alternative<MigrationRequestMessage>(dispatch_q_.front().handle);
 
-      // Process Admin Queue - only if not blocked
-      if (!dispatch_q_.empty() && !is_blocked_migration) {
+      // Check if we are blocked:
+      // Blocked = It IS a migration AND we have pipeline work AND we are in Redis mode.
+      bool is_blocked =
+          is_migration_req && (parsed_head_ != nullptr) && (protocol_ == Protocol::REDIS);
+
+      /// Process Admin Queue
+      // Run if: Queue is not empty and we are not blocked.
+      if (!dispatch_q_.empty() && !is_blocked) {
         msg = std::move(dispatch_q_.front());
         dispatch_q_.pop_front();
       } else {


src/facade/dragonfly_connection.cc
```diff
@@ -1600,27 +1622,56 @@ void Connection::AsyncFiber() {
...
+            << "Pointers diverged before dispatch decision.";
+        auto* cmd{parsed_to_execute_};
+        parsed_to_execute_ = cmd->next;
+        parsed_head_ = parsed_to_execute_;
+
+        // Wrap manually for uniform execution below
+        msg = MessageHandle{PipelineMessagePtr{cmd}};
+      }
 
       stats_->pipelined_wait_latency += CycleClock::ToUsec(CycleClock::Now() - msg.dispatch_cycle);
```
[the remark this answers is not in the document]
glevkovich: This is known issue, I am going to deal with it later.

src/facade/dragonfly_connection.cc
```diff
@@ -2140,42 +2207,50 @@ ParsedCommand* Connection::CreateParsedCommand() {
...
   --parsed_cmd_q_len_;
   if (is_pipelined) {
     stats_->pipelined_cmd_cnt++;
     stats_->pipelined_cmd_latency += CycleClock::ToUsec(CycleClock::Now() - cmd->parsed_cycle);
   }
 
+  if (protocol_ == Protocol::REDIS) {
+    DCHECK_GT(pending_pipeline_cmd_cnt_, 0u);
+    pending_pipeline_cmd_cnt_--;
+    pending_pipeline_bytes_ -= used_mem;
```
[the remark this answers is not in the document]
glevkovich: fixed

src/facade/dragonfly_connection.cc
```diff
@@ -1600,27 +1622,56 @@ void Connection::AsyncFiber() {
...
+      } else {
+        // Process Pipeline: we reach here if:
+        // 1. dispatch_q is empty (Normal pipeline processing)
+        // 2. OR dispatch_q has a migration but is blocked (We must clear pipeline first)
+        DCHECK(parsed_head_) << "Logic Error: Fiber woke up but no work found";
+        DCHECK_EQ(parsed_head_, parsed_to_execute_)
+            << "Pointers diverged before dispatch decision.";
+        auto* cmd{parsed_to_execute_};
+        parsed_to_execute_ = cmd->next;
+        parsed_head_ = parsed_to_execute_;
```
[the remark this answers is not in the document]
glevkovich: fixed

src/facade/dragonfly_connection.cc
```diff
@@ -2140,42 +2207,50 @@ ParsedCommand* Connection::CreateParsedCommand() {
...
   parsed_cmd_q_len_++;
 
-  // We do not recreate parsed_cmd_ here, to support non-pipelined usage pattern,
-  // where we will reuse the same parsed_cmd_ object again after we execute it.
-  parsed_cmd_ = nullptr;  // ownership transferred
+  if (protocol_ == Protocol::MEMCACHE) {
+    stats_->dispatch_queue_bytes += cmd->UsedMemory();
+    stats_->dispatch_queue_entries++;
+  }
 }
```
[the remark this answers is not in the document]
glevkovich: This one doesnt have unittests due to the complexity of introducing new ones. Anyway, this is not going to be done as part of the task.

src/facade/dragonfly_connection.cc
```diff
@@ -1429,26 +1431,24 @@ bool Connection::ShouldEndAsyncFiber(const MessageHandle& msg) {
...
+  // This lambda advances a temporary pointer exec_it to feed the execution engine.
+  // We do not modify parsed_to_execute_ yet, in case execution throws/fails.
+  auto exec_it{parsed_to_execute_};
+  auto get_next_fn = [&exec_it]() mutable -> facade::ParsedCommand& {
+    DCHECK(exec_it) << "exec_it should not be nullptr";
+    facade::ParsedCommand* cmd = exec_it;
+
+    // Note: We don't need to check for nullptr here because DispatchManyCommands
+    // is guaranteed to call this exactly 'pipeline_count' times.
+    exec_it = exec_it->next;
```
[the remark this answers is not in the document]
glevkovich: done

src/facade/facade_types.h
```diff
@@ -141,10 +141,13 @@ inline std::string_view ArgS(ArgSlice args, size_t i) {
...
 
 struct ConnectionStats {
-  size_t read_buf_capacity = 0;                // total capacity of input buffers
-  uint64_t dispatch_queue_entries = 0;         // total number of dispatch queue entries
-  size_t dispatch_queue_bytes = 0;             // total size of all dispatch queue entries
+  size_t read_buf_capacity = 0;         // total capacity of input buffers
+  uint64_t dispatch_queue_entries = 0;  // total count of all pending
+                                        // messages (data + administrative messages)
+  size_t dispatch_queue_bytes = 0;  // total size of memory used by all pending messages (pipeline
+                                    // data queue + dispatch queue admin messages)
```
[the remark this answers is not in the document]
glevkovich: done

src/facade/dragonfly_connection.cc
```diff
@@ -1475,34 +1477,42 @@ void Connection::SquashPipeline() {
...
+        CycleClock::ToUsec(CycleClock::Now() - flush_start_cycle_cnt);
   }
-  auto it = dispatch_q_.begin();
-  while (it->IsCheckPoint())  // Skip newly received checkpoint messages
-    ++it;
 
-  for (auto rit = it; rit != it + dispatched; ++rit) {
+  for (size_t i{}; (i < num_dispatched_cmds) && parsed_to_execute_; ++i) {
+    parsed_to_execute_ = parsed_to_execute_->next;
+  }
```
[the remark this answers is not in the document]
glevkovich: claim is incorrect, but I'm going to simplify code.

src/facade/dragonfly_connection.cc
```diff
@@ -1600,27 +1622,56 @@ void Connection::AsyncFiber() {
...
+        // 2. OR dispatch_q has a migration but is blocked (We must clear pipeline first)
+        DCHECK(parsed_head_) << "Logic Error: Fiber woke up but no work found";
+        DCHECK_EQ(parsed_head_, parsed_to_execute_)
+            << "Pointers diverged before dispatch decision.";
+        auto* cmd{parsed_to_execute_};
+        parsed_to_execute_ = cmd->next;
+        parsed_head_ = parsed_to_execute_;
+
+        // Wrap manually for uniform execution below
+        msg = MessageHandle{PipelineMessagePtr{cmd}};
```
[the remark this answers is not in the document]
glevkovich: known issue. will be handled later.

src/facade/dragonfly_connection.cc
```diff
@@ -1511,6 +1521,21 @@ void Connection::ClearPipelinedMessages() {
...
+    auto* curr{parsed_head_};
+    parsed_head_ = parsed_head_->next;
+    ReleaseParsedCommand(curr, false);
+  }
+  parsed_head_ = nullptr;
+  parsed_tail_ = nullptr;
+  parsed_to_execute_ = nullptr;
+  pending_pipeline_cmd_cnt_ = 0;
+  pending_pipeline_bytes_ = 0;
+  parsed_cmd_q_len_ = 0;
```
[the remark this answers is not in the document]
glevkovich: done

src/facade/dragonfly_connection.cc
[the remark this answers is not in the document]
glevkovich: incorrect claim. dropping.

src/facade/dragonfly_connection.cc
```diff
@@ -1600,27 +1622,56 @@ void Connection::AsyncFiber() {
...
+          (parsed_head_ != nullptr)};
+
+      // Process Admin Queue - only if not blocked
+      if (!dispatch_q_.empty() && !is_blocked_migration) {
+        msg = std::move(dispatch_q_.front());
+        dispatch_q_.pop_front();
+      } else {
+        // Process Pipeline: we reach here if:
+        // 1. dispatch_q is empty (Normal pipeline processing)
+        // 2. OR dispatch_q has a migration but is blocked (We must clear pipeline first)
```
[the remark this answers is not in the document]
glevkovich: drop...this change can cause infinite logical loop.

src/facade/dragonfly_connection.cc
```diff
@@ -1812,17 +1868,24 @@ void Connection::SendAsync(MessageHandle msg) {
...
+  // 2. Pipeline messages go to the parsed command queue.
+  // 3. All other messages go to the back of the dispatch_q.
   if (msg.IsCheckPoint()) {
     auto it = dispatch_q_.begin();
     while (it < dispatch_q_.end() && it->IsCheckPoint())
       ++it;
     dispatch_q_.insert(it, std::move(msg));
+  } else if (msg.IsPipelineMsg()) {
+    auto& pmsg = std::get<PipelineMessagePtr>(msg.handle);
+    EnqueueParsedCommand(pmsg.release());
```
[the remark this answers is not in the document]
glevkovich: fixed already

src/facade/dragonfly_connection.cc
```diff
@@ -2140,42 +2207,50 @@ ParsedCommand* Connection::CreateParsedCommand() {
...
 }
 
 void Connection::ReleaseParsedCommand(ParsedCommand* cmd, bool is_pipelined) {
   size_t used_mem = cmd->UsedMemory();
   DCHECK_GE(stats_->dispatch_queue_bytes, used_mem);
+  DCHECK_GT(stats_->dispatch_queue_entries, 0u);
   DCHECK_GT(parsed_cmd_q_len_, 0u);
   stats_->dispatch_queue_bytes -= used_mem;
+  --stats_->dispatch_queue_entries;
   --parsed_cmd_q_len_;
```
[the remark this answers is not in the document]

src/facade/dragonfly_connection.cc
```diff
@@ -1600,27 +1623,63 @@ void Connection::AsyncFiber() {
...
+            << "Pointers diverged before dispatch decision.";
+        auto* cmd{parsed_to_execute_};
+        parsed_to_execute_ = cmd->next;
+        parsed_head_ = parsed_to_execute_;
+        if (parsed_head_ == nullptr) {
+          parsed_tail_ = nullptr;
+        }
+
+        // Wrap manually for uniform execution below
+        msg = MessageHandle{PipelineMessagePtr{cmd}};
```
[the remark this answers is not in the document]
glevkovich: fixed

src/facade/facade_types.h
```diff
@@ -142,9 +142,12 @@ inline std::string_view ArgS(ArgSlice args, size_t i) {
 
 struct ConnectionStats {
   size_t read_buf_capacity = 0;                // total capacity of input buffers
-  uint64_t dispatch_queue_entries = 0;         // total number of dispatch queue entries
-  size_t dispatch_queue_bytes = 0;             // total size of all dispatch queue entries
+  uint64_t dispatch_queue_entries = 0;         // total count of all pending
+                                               // messages (data + administrative messages)
+  size_t dispatch_queue_bytes = 0;             // total size of memory used by all pending
+                                               // messages (pipeline data queue + dispatch queue
```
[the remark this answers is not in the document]
glevkovich: fixed

src/facade/dragonfly_connection.cc
```diff
@@ -1429,26 +1431,24 @@ bool Connection::ShouldEndAsyncFiber(const MessageHandle& msg) {
 }
 
 void Connection::SquashPipeline() {
-  DCHECK_EQ(dispatch_q_.size(), pending_pipeline_cmd_cnt_);
+  DCHECK_EQ(CommandsPendingCount(), pending_pipeline_cmd_cnt_);
```
romange: use verbs for functions, i.e. GetCommandsPendingCount

src/facade/dragonfly_connection.cc
```diff
@@ -845,8 +845,9 @@ pair<string, string> Connection::GetClientInfoBeforeAfterTid() const {
   }
   string after;
   absl::StrAppend(&after, " irqmatch=", int(cpu == my_cpu_id));
-  if (dispatch_q_.size()) {
-    absl::StrAppend(&after, " pipeline=", dispatch_q_.size());
+  size_t cmd_pending_count{CommandsPendingCount()};
+  if (cmd_pending_count > 0) {
+    absl::StrAppend(&after, " pipeline=", cmd_pending_count);
```
romange: if CommandsPendingCount is total number of pending commands, then why use it for pipeline here?
should not it be `parsed_cmd_q_len_` ?

src/facade/dragonfly_connection.cc
```diff
@@ -1108,16 +1109,17 @@ void Connection::DispatchSingle(bool has_more, absl::FunctionRef<void()> invoke_
   bool optimize_for_async = has_more;
   QueueBackpressure& qbp = GetQueueBackpressure();
   if (optimize_for_async &&
-      qbp.IsPipelineBufferOverLimit(stats_->dispatch_queue_bytes, dispatch_q_.size())) {
+      qbp.IsPipelineBufferOverLimit(stats_->dispatch_queue_bytes, CommandsPendingCount())) {
```
romange: same, why account dispatch_q here if we move pipeline commands to parsed_q ?

src/facade/dragonfly_connection.cc
```diff
@@ -1429,26 +1431,24 @@ bool Connection::ShouldEndAsyncFiber(const MessageHandle& msg) {
...
-    const auto& elem = dispatch_q_[ctrl_offset + i++];
-    CHECK(holds_alternative<PipelineMessagePtr>(elem.handle));
-    const auto& pmsg = get<PipelineMessagePtr>(elem.handle);
-
-    return *pmsg;
+  // Define a "Feeder" Lambda
+  // This lambda advances a temporary pointer exec_cmd_ptr to feed the execution engine.
+  // We do not modify parsed_to_execute_ yet, in case execution throws/fails.
+  auto exec_cmd_ptr{parsed_to_execute_};
+  auto get_next_fn = [&exec_cmd_ptr]() mutable -> facade::ParsedCommand& {
```
romange: why returning a reference?

src/facade/dragonfly_connection.cc
```diff
@@ -1577,19 +1600,19 @@ void Connection::AsyncFiber() {
     // wants to.
     // As an optimization, we only yield if the fiber was not suspended since the last dispatch.
     uint64_t cur_epoch = fb2::FiberSwitchEpoch();
-    if (dispatch_q_.size() == 1 && cur_epoch == prev_epoch) {
+    if ((CommandsPendingCount() == 1) && (cur_epoch == prev_epoch)) {
```
romange: i think it's wrong. we care only about pipelined commands, not dispatch_q commands.

src/facade/dragonfly_connection.cc
```diff
@@ -2140,42 +2221,63 @@ ParsedCommand* Connection::CreateParsedCommand() {
...
+  if (protocol_ == Protocol::MEMCACHE) {
+    stats_->dispatch_queue_bytes += cmd->UsedMemory();
+    stats_->dispatch_queue_entries++;
+  }
 }
 
 void Connection::ReleaseParsedCommand(ParsedCommand* cmd, bool is_pipelined) {
   size_t used_mem = cmd->UsedMemory();
+
+  // Redis stats (via SendAsync) include the MessageHandle overhead.
```
romange, line 2260: not sure I understand. do we use MessageHandle with ParsedCommand now?

src/facade/dragonfly_connection.cc
```diff
@@ -1600,27 +1623,66 @@ void Connection::AsyncFiber() {
...
+    bool are_all_plain_cmds = pending_pipeline_cmd_cnt_ == CommandsPendingCount();
     if (squashing_enabled && threshold_reached && are_all_plain_cmds && !skip_next_squashing_ &&
         !IsReplySizeOverLimit()) {
       SquashPipeline();
     } else {
-      MessageHandle msg = std::move(dispatch_q_.front());
-      dispatch_q_.pop_front();
+      MessageHandle msg;
+
+      // If the front message is a Migration Request, but we still have pipeline data
```
romange, line 1639: why we must do that?

romange:
I am confused - do we have pipeline messages in dispatch_q or not? Why do we still check for `IsPipelineMsg` in the code?

src/facade/dragonfly_connection.cc
glevkovich: done

src/facade/dragonfly_connection.cc
glevkovich: My initial intention was to strictly preserve the legacy behavior (where 'pipeline' displayed the total pending count) for Phase 1 and apply your suggestion in the second PR. The current code in main follows this exact logic.

I will take your suggested approach: where it doesn't add unnecessary complexity, I will apply these behavioral improvements now. If a specific change gets too complex, I'll defer it to the second PR.

src/facade/dragonfly_connection.cc
glevkovich: See my comment here:
https://github.com/dragonflydb/dragonfly/pull/6423#discussion_r2711348176
I applied a more accurate change using pending_pipeline_bytes_ and pending_pipeline_cmd_cnt_

src/facade/dragonfly_connection.cc
glevkovich: I see it as "safer". I'll make it a pointer if you think that the style doesn't fit.

src/facade/dragonfly_connection.cc
glevkovich: See my comment here:
https://github.com/dragonflydb/dragonfly/pull/6423#discussion_r2711348176
I applied a change using parsed_cmd_q_len_

src/facade/dragonfly_connection.cc
glevkovich: This is temporary, until I finalise the 2nd PR. Currently:

- admin messages go to dispatch_q.
- pipeline messages got to Parsed Commands queue. 
But the wrapping logic around still has PipelineMessagePtr which hasn't been removed from the variant.  fully separating the logic will create a larger change that doesn't' fit into a single PR and adds complexity. Since this logic is using  MessageHandle extensively, I temporarily wrap the pipeline messages with MessageHandle and adjust statistics accordingly.

This is a temporary bridge for Phase 1. SendAsync still accounts for sizeof(MessageHandle) when tracking memory usage for the queue. To ensure dispatch_queue_bytes returns exactly to zero when the queue is empty, I must account for that overhead here during release. I will remove this dependency in Phase 2.

src/facade/dragonfly_connection.cc
glevkovich: With the current dequeuing policy, we  prioritize popping from dispatch_q first. Before this change, a migration message was sometimes pushed back into the dispatch_q in ShouldEndAsyncFiber() whenever there was pending work. Because it was a single queue, we were guaranteed it would be served only after the preceding messages were processed.

Now that we have two separate queues, we must implement a form of priority inversion for migrations. If we didn't, the migration request would keep popping from the high-priority dispatch_q, seeing a non-empty pipeline, and pushing itself back indefinitely - creating an infinite logical loop that prevents the pipeline from ever draining.

glevkovich:
> I am confused - do we have pipeline messages in dispatch_q or not? Why do we still check for `IsPipelineMsg` in the code?

@romange , The answer is no - pipeline messages do not go to dispatch queue.
You can see the queing rules in SendAsync, under "// Queueing rules:" and the code below uses EnqueueParsedCommand to push pipeline messages into the Parsed commands queue.

I'm aware and agree that wrapping ParsedCommand in a MessageHandle is somewhat "artificial" as you've commented, but this was an aware decision made to split the work into two parts. The full separation hasn't been completed yet. In the 2nd phase PR, I plan to remoeve PipelineMessagePtr from the variant, which will force me to change the logic and address this concern. The memory accounting is a side effect of this decision. Using IsPipelineMsg or PipelineMessagePtr in the code is still temporary.

src/facade/dragonfly_connection.cc
```diff
@@ -1430,26 +1432,24 @@ bool Connection::ShouldEndAsyncFiber(const MessageHandle& msg) {
...
-    }
-    const auto& elem = dispatch_q_[ctrl_offset + i++];
-    CHECK(holds_alternative<PipelineMessagePtr>(elem.handle));
-    const auto& pmsg = get<PipelineMessagePtr>(elem.handle);
-
-    return *pmsg;
+  // Define a "Feeder" Lambda
+  // This lambda advances a temporary pointer exec_cmd_ptr to feed the execution engine.
+  // We do not modify parsed_to_execute_ yet, in case execution throws/fails.
+  auto exec_cmd_ptr{parsed_to_execute_};
```
kostasrim, line 1444: Before my refactor `ControlMessages` was a larger set of types. Now it's only `Checkpoint` messages which are only sent:

* On shutdown, via DispatchTracker (see dfly listener)
* On Pause command (see Pause)
* On TakeOver flow (see dfly::Takeover)
* On cluster config change (see cluster fam) 

IMO it's much simpler now. If you want to keep a two level queue, I would advice to:

* dispatch_q -> only for checkpoint messages -> you can call it checkpoint_q
* the other ParsedHead...Tail which will contain the rest of the messages. (Plz see my other comments for the rationale)

src/facade/dragonfly_connection.cc
```diff
@@ -1813,17 +1884,27 @@ void Connection::SendAsync(MessageHandle msg) {
...
     auto it = dispatch_q_.begin();
     while (it < dispatch_q_.end() && it->IsCheckPoint())
       ++it;
     dispatch_q_.insert(it, std::move(msg));
+  } else if (msg.IsPipelineMsg()) {
+    auto& pmsg = std::get<PipelineMessagePtr>(msg.handle);
+    // Transfer the dispatch time from the Handle to the Command
+    // so we calculate latency from the moment SendAsync was called.
+    pmsg->parsed_cycle = msg.dispatch_cycle;
+    EnqueueParsedCommand(pmsg.release());
```
kostasrim, line 1912: You can just `parsed_command_ = pms.release()`and that way you don't need to change the signature of `EnqueueParsedCommand`.

src/facade/dragonfly_connection.cc
```diff
@@ -2120,42 +2205,63 @@ ParsedCommand* Connection::CreateParsedCommand() {
   return res;
 }
 
-void Connection::EnqueueParsedCommand() {
-  parsed_cmd_->next = nullptr;
-  parsed_cmd_->parsed_cycle = base::CycleClock::Now();
+void Connection::EnqueueParsedCommand(ParsedCommand* cmd) {
```
kostasrim, line 2227: It's weird to pass the state of the object (parsed_command_) as argument to its member function + the additional refactoring changes. Consider:

1. For the place that you have msgptr, set parsed_command_. 
2. revert the argument here

src/facade/dragonfly_connection.cc
```diff
@@ -2120,42 +2205,63 @@ ParsedCommand* Connection::CreateParsedCommand() {
...
+      parsed_to_execute_ = cmd;
     }
-    // We have a pipelined command
-    local_stats_.dispatch_entries_added++;
   }
-  parsed_tail_ = parsed_cmd_;
-  stats_->dispatch_queue_bytes += parsed_cmd_->UsedMemory();
+  parsed_tail_ = cmd;
 
+  local_stats_.dispatch_entries_added++;
```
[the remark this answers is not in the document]

src/facade/dragonfly_connection.h
```diff
@@ -411,6 +418,16 @@ class Connection : public util::Connection {
...
   unsigned parsed_cmd_q_len_ = 0;
 
+  // Returns true if there are any commands pending in the parsed command queue or dispatch queue.
+  bool IsCommandsPending() const {
+    return parsed_head_ || !dispatch_q_.empty();
+  }
+
+  // Returns total count of commands pending in the parsed command queue and dispatch queue.
+  size_t GetPendingCommandCount() const {
+    return static_cast<size_t>(parsed_cmd_q_len_) + dispatch_q_.size();
```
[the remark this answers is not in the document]
glevkovich: done

src/facade/dragonfly_connection.cc
```diff
@@ -1601,27 +1624,66 @@ void Connection::AsyncFiber() {
...
+        dispatch_q_.pop_front();
+      } else {
+        // Process Pipeline: we reach here if:
+        // 1. dispatch_q is empty (Normal pipeline processing)
+        // 2. OR dispatch_q has a migration but is blocked (We must clear pipeline first)
+        DCHECK(parsed_head_) << "Logic Error: Fiber woke up but no work found. "
+                             << "Context: dispatch_q_.empty()=" << dispatch_q_.empty()
+                             << ", IsCommandsPending()=" << IsCommandsPending()
+                             << ", parsed_head_=" << parsed_head_
+                             << ", parsed_to_execute_=" << parsed_to_execute_;
```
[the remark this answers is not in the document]
glevkovich: done

src/facade/dragonfly_connection.cc
kostasrim: I do not understand this either. First, `MigrationRequest` is always `pushed_back` at the dispatch queue right ? 

Second, I am troubled by the `behavioral` change of `separating the two queues`. In this PR:

* Only PipelineMsgs -> `ParsedHead..Tail`
* All other messages  (Checkpoints, PubSub, Monitor etc)-> dispatch queue

well my *concern* here is that we change the `order that we process the messages`. 

### Before:

Checkpoints messages where placed at the front of the queue (preserving relative order to other similar messages at the front). Any other queue item was placed at the back. 

### Now

You place pipeline messages together. This breaks relative order to: pub sub, monitor and migration messages. In other words what's the reason for doing this ? Do you know that we don't break existing correct behaviour that relies on the relative order of those message types ?


IMO the simplest first step would be:

1. -> All messages go to parsed head..tail -> preserving relative order
2. -> Only checkpoints gets placed in dispatch queue (which will removed if we get rid of the checkpoint mechanism alltogether)


Is there something I am missing that you both discussed and I am not aware of ?

kostasrim: And also my other problem is that this PR is a `refactor` PR so there should be not `behavioral` changes and yet we are here arguing about dispatch policies and priorities 🤔

src/facade/dragonfly_connection.h
```diff
@@ -364,8 +364,13 @@ class Connection : public util::Connection {
   };
 
   ParsedCommand* CreateParsedCommand();
-  void EnqueueParsedCommand();
+  void EnqueueParsedCommand(ParsedCommand* cmd);
+
+  // Releases the command memory back to the pool.
+  // Set is_pipelined=true if the command was executed (to update latency/throughput stats).
+  // Set is_pipelined=false if the command is being dropped/cleaned up without execution.
```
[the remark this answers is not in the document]
glevkovich: done

src/facade/dragonfly_connection.cc
```diff
@@ -1601,27 +1624,66 @@ void Connection::AsyncFiber() {
...
+        }
+
+        // Wrap manually for uniform execution below
+        msg = MessageHandle{PipelineMessagePtr{cmd}};
+        // Transfer timestamp so wait_latency is calculated
+        // from the time it entered the queue, not time it was popped.
+        msg.dispatch_cycle = cmd->parsed_cycle;
+      }
 
       stats_->pipelined_wait_latency += CycleClock::ToUsec(CycleClock::Now() - msg.dispatch_cycle);
```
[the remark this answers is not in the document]
glevkovich: moved the line into the else above, which handles a pipeline message

src/facade/dragonfly_connection.cc
```diff
@@ -1601,27 +1624,66 @@ void Connection::AsyncFiber() {
...
+        // Process Pipeline: we reach here if:
+        // 1. dispatch_q is empty (Normal pipeline processing)
+        // 2. OR dispatch_q has a migration but is blocked (We must clear pipeline first)
+        DCHECK(parsed_head_) << "Logic Error: Fiber woke up but no work found. "
+                             << "Context: dispatch_q_.empty()=" << dispatch_q_.empty()
+                             << ", IsCommandsPending()=" << IsCommandsPending()
+                             << ", parsed_head_=" << parsed_head_
+                             << ", parsed_to_execute_=" << parsed_to_execute_;
+        DCHECK_EQ(parsed_head_, parsed_to_execute_)
+            << "Pointers diverged before dispatch decision.";
```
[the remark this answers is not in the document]
glevkovich: fixxd the wait condition so now it can't happen for memcached.

src/facade/dragonfly_connection.cc
glevkovich: I don't think it's a correct , since parsed_cmd_ holds the  the buffer for the next command to be read at this point: 

If you look at DispatchSingle, we use std::exchange in dispatch_async to detach the fully parsed command and immediately reset parsed_cmd_ to a fresh object for the next read loop. If EnqueueParsedCommand relied on parsed_cmd_, it would queue the empty buffer instead of the populated command.

another reason, is because EnqueueParsedCommand is reached via SendAsync, which is called from fucntions like SendCheckpoint or SendInvalidationMessageAsync where parsed_cmd_ is irrelevant or currently in use by the IO loop. We must pass the specific command pointer explicitly.

src/facade/dragonfly_connection.cc
glevkovich: I am currently using this approach as a temporary bridge for Phase 1 to wrap the ParsedCommand. Based on the feedback provided in the [other comment ](https://github.com/dragonflydb/dragonfly/pull/6423#discussion_r2712288605 )and the planned architectural cleanup for Phase 2, I expect to revert this signature then.
Implementing a more comprehensive solution now would require significant additional refactoring, which would bloat this PR and make it much harder to review. This is exactly what I was trying to avoid.

glevkovich:
>I do not understand this either. First, MigrationRequest is always pushed_back at the dispatch queue right ?
..
> well my concern here is that we change the order that we process the messages.
https://github.com/dragonflydb/dragonfly/pull/6423#discussion_r2712360966
@kostasrim 
I totally agree with "well my concern here is that we change the order that we process the messages." and I've mentioned it in our meetings. Even with my phase 1 change, the order of queue/dequeue has already changed.
I believe a solution is directly related to [this comment](https://github.com/dragonflydb/dragonfly/pull/6423#discussion_r2712288605) since if we manage all  messages else than checkpoint messages  in a single queue we can preserve exactly the same behaviour. So there is no reason in my opinion to open here a new discussion . I think we should focus on your suggestion to have a checkpoint_q,  as this solution, if accepted, will preserve the same behaviour.

romange:
@kostasrim  I asked @glevkovich  to rattle the cage and change both the implementation and the behaviour. 
1. We need to make sure that Checkpoints still processed ASAP, i.e. they should take precedence over pipeline processing, so they should break the loop of processing/enqueuing the parsed_queue.
2. PubSub / Pipeline - the behavior is implementation dependent as we should not have pipelines flowing with subscribed connections. A (possible) behaviour could be - like (1).
3. Monitor / Pipeline - similarly, we should not have such case, as monitor messages should flow to a passive client. We can also default to (1).

So to summarise: we should break from a pipelining loop as soon as we see dispatch_q not empty, imho. Maybe I am missing something but this is the goal of this task - to find what do we miss.  @glevkovich  please pay attention to FakeRedis tests - these are pretty good to fix the "spec" behavior and should pass, imho.

kostasrim:
> So to summarise: we should break from a pipelining loop as soon as we see dispatch_q not empty, imho. Maybe I am missing something but this is the goal of this task - to find what do we miss. @glevkovich please pay attention to FakeRedis tests - these are pretty good to fix the "spec" behavior and should pass, imho.

ok ok, I now understand. You do want to push all other messages to a different queue (the dispatch queue) and keep pipeline msgs explicit and self contained within the head..tail list. Gotcha. 

> PubSub / Pipeline - the behavior is implementation dependent as we should not have pipelines flowing with subscribed connections. A (possible) behaviour could be - like (1).

Is this really the case ? For `resp3` a connection can `send pipelines` while being `subscribed to a channel`.  My concern here is even for single connections we break the relative ordering of operations:

client -> pipeline of published messaged -> subscribe 

Now subscribe breaks the pipeline. Imagine all of this arrive to the socket in one go and as we parse those we divide them:

head..tail -> all pipelined publish commands 
dispatch_q_ -> SUBSCRIBE command

Now the question how do you execute ? if you drain the head..tail first and then call `SUBSCRIBE` you keep the `order of operations from the client` but your `new policy is to drain the dispatch q instead`. So what you end up running is:

```
SUBSCRIBE -> publish 1, publish2 publish N
```

And now client received `msg 1, 2 N` in `OUT OF ORDER`. And then what this means is, that effectively the series of commands from the connection got scrambled. 

This does't seem like eventual consistency but out of order execution of commands from a single connection 

Am I missing something ?

romange:
TLDR: I do not know - that's up to Gil to investigate

In more detail, dispatch_q does not contain SUBSCRIBE command coming from the same socket. It contains Publish message notifications coming from OTHER clients, meaning we might interleave publish notifications and pipeline processing in undefined order. It's true that RESP3 allows to handle publish notifications in parallel to processing the commands and it's a good point but I do not see how it can break and specifically in your example, SUBSCRIBE should come before pipelined commands to set the client into the subscribed state and start receiving the notifications. So the order will be: client -> SUBSCRIBE -> pipeline burst. And then dispatch_q will have "Publish" notification that will be handled and passed on to the socket "out of bound" in the middle of processing of pipelining batch. RESP3 is designed to handle that and it should not affect correctness but can change our implementation detail.

src/facade/dragonfly_connection.cc
```diff
@@ -1813,17 +1884,27 @@ void Connection::SendAsync(MessageHandle msg) {
     pending_pipeline_bytes_ += used_mem;
```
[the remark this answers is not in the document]
glevkovich: ignore..after revert to check a failing test.

src/facade/dragonfly_connection.cc
```diff
@@ -846,8 +846,8 @@ pair<string, string> Connection::GetClientInfoBeforeAfterTid() const {
   }
   string after;
   absl::StrAppend(&after, " irqmatch=", int(cpu == my_cpu_id));
-  if (dispatch_q_.size()) {
-    absl::StrAppend(&after, " pipeline=", dispatch_q_.size());
+  if (parsed_cmd_q_len_ > 0) {
```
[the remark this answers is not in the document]
glevkovich: ignore..after revert to check a failing test.

src/facade/dragonfly_connection.cc
```diff
@@ -1927,7 +2008,10 @@ Connection::MemoryUsage Connection::GetMemoryUsage() const {
     mem += UsedMemoryInternal(*parsed_cmd_);
   }
 
-  // We add a hardcoded 9k value to accomodate for the part of the Fiber stack that is in use.
+  // Account for the pipelined commands waiting in the linked list
+  mem += pending_pipeline_bytes_;
```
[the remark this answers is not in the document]
glevkovich: ignore..after revert to check a failing test.

src/facade/dragonfly_connection.cc
```diff
@@ -1632,15 +1694,20 @@ void Connection::AsyncFiber() {
...
-      if (dispatch_q_.empty() && is_replying &&
+      if (!IsCommandsPending() && is_replying &&
           (replies_recorded_before == reply_builder_->RepliesRecorded())) {
         reply_builder_->Flush();
       }
-      RecycleMessage(std::move(msg));
+      if (msg.IsPipelineMsg()) {
+        auto& ptr_ref = std::get<PipelineMessagePtr>(msg.handle);
+        ReleaseParsedCommand(ptr_ref.release(), true);
+      } else {
```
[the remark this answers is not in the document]
glevkovich: ignore..after revert to check a failing test.

src/facade/dragonfly_connection.cc
```diff
@@ -2120,42 +2205,63 @@ ParsedCommand* Connection::CreateParsedCommand() {
...
     }
-    // We have a pipelined command
-    local_stats_.dispatch_entries_added++;
   }
-  parsed_tail_ = parsed_cmd_;
-  stats_->dispatch_queue_bytes += parsed_cmd_->UsedMemory();
+  parsed_tail_ = cmd;
 
+  local_stats_.dispatch_entries_added++;
   parsed_cmd_q_len_++;
```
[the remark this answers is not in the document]
glevkovich: ignore..after revert to check a failing test.

src/facade/dragonfly_connection.cc
```diff
@@ -1430,26 +1432,24 @@ bool Connection::ShouldEndAsyncFiber(const MessageHandle& msg) {
...
-    const auto& elem = dispatch_q_[ctrl_offset + i++];
-    CHECK(holds_alternative<PipelineMessagePtr>(elem.handle));
-    const auto& pmsg = get<PipelineMessagePtr>(elem.handle);
-
-    return *pmsg;
+  // Define a "Feeder" Lambda
+  // This lambda advances a temporary pointer exec_cmd_ptr to feed the execution engine.
+  // We do not modify parsed_to_execute_ yet, in case execution throws/fails.
+  auto exec_cmd_ptr{parsed_to_execute_};
+  auto get_next_fn = [&exec_cmd_ptr]() mutable -> facade::ParsedArgs {
```
romange:
```suggestion
  auto get_next_fn = [&exec_cmd_ptr]() mutable -> ParsedArgs {
```

src/facade/dragonfly_connection.cc
```diff
@@ -1430,26 +1432,24 @@ bool Connection::ShouldEndAsyncFiber(const MessageHandle& msg) {
...
+    DCHECK(exec_cmd_ptr) << "exec_cmd_ptr should not be nullptr";
```
romange:
```suggestion
    DCHECK(exec_cmd_ptr);
```

src/facade/dragonfly_connection.cc
```diff
@@ -1476,34 +1478,39 @@ void Connection::SquashPipeline() {
...
+    stats_->pipeline_dispatch_flush_usec +=
+        CycleClock::ToUsec(CycleClock::Now() - flush_start_cycle_cnt);
   }
-  auto it = dispatch_q_.begin();
-  while (it->IsCheckPoint())  // Skip newly received checkpoint messages
-    ++it;
 
-  for (auto rit = it; rit != it + dispatched; ++rit) {
+  auto* current{parsed_head_};
+  for (size_t i{}; (i < num_dispatched_cmds) && current; ++i) {
```
romange: regarding `size_t i{}`  - I personally find such initialisation less readable. I guess it's fine in generic code but I do not see much value for using this with numbers. unrelred - we also do not use `auto` for simple, short types as it hurts readability in many cases

src/facade/dragonfly_connection.cc
```diff
@@ -1476,34 +1478,39 @@ void Connection::SquashPipeline() {
...
-  dispatch_q_.erase(it, it + dispatched);
+    ReleaseParsedCommand(current, true /* is_pipelined */);
+    current = next;
+  }
+  parsed_head_ = current;
+  if (!parsed_head_) {
+    parsed_tail_ = nullptr;
+  }
+  parsed_to_execute_ = parsed_head_;
+  DCHECK_EQ(parsed_head_, parsed_to_execute_) << "SquashPipeline: Pointers diverged.";
```
romange: we usually do not add checks that verify a trivial outcome of a line above.

src/facade/dragonfly_connection.cc
```diff
@@ -1565,8 +1590,23 @@ void Connection::AsyncFiber() {
...
+        return true;
+
+      // If we are currently executing a synchronous dispatch (e.g. inside IoLoop),
+      // we must wait until it finishes to avoid race conditions.
+      if (cc_->sync_dispatch)
+        return false;
+
+      // For Memcache, we ONLY wake up for Admin messages (dispatch_q_).
+      // The pipeline (parsed_head_) is processed synchronously by the memcache loop.
+      // For Redis, we wake up for both.
```
romange:
```suggestion
      // For Memcache, we ONLY wake up for Admin messages (dispatch_q_) as we process
      // parsed_head_  in the connection fiber. For RESP, we wake up for both queues.
```

glevkovich:
**PR Summary & Phase 1 Rationale**

* **Transitional State:** `MessageHandle::handle` still holds `PipelineMessagePtr`. This design choice was made to minimize the scope of code changes in this PR. Removing `PipelineMessagePtr` entirely requires a complete separation of logic within `AsyncFiber`, which would significantly increase review complexity. That separation is being prepared for Phase 2.
* **Queuing Policy:**
1. **Admin/Control:** All admin messages are pushed into `dispatch_q_`. Checkpoint messages continue to be placed at the front (behavior unchanged).
2. **Data:** Pipeline messages are pushed into the separate "Parsed Commands Queue" (linked list).


* **Prioritization:** This split effectively changes the prioritization model from 2 levels to 3:
1. Checkpoint Messages (Highest)
2. Other Admin Messages
3. Pipeline/Data Messages (Lowest)


* **Priority Inversion:** There is a specific exception for Migration messages. Due to how they are re-queued in certain scenarios, the logic strictly drains the pipeline queue first to prevent deadlocks.
* **Temporary Implementation details:** Since `AsyncFiber` still shares execution logic for all message types in this phase, `PipelineMessagePtr` remains wrapped in `MessageHandle` with adjusted memory statistics. This is a temporary bridge; Phase 2 will introduce complete separation and remove this wrapper.

romange:
> **PR Summary & Phase 1 Rationale**
> 
> * **Transitional State:** `MessageHandle::handle` still holds `PipelineMessagePtr`. This design choice was made to minimize the scope of code changes in this PR. Removing `PipelineMessagePtr` entirely requires a complete separation of logic within `AsyncFiber`, which would significantly increase review complexity. That separation is being prepared for Phase 2.
> * **Queuing Policy:**
> 
> 1. **Admin/Control:** All admin messages are pushed into `dispatch_q_`. Checkpoint messages continue to be placed at the front (behavior unchanged).
> 2. **Data:** Pipeline messages are pushed into the separate "Parsed Commands Queue" (linked list).
> 
> * **Prioritization:** This split effectively changes the prioritization model from 2 levels to 3:
> 
> 1. Checkpoint Messages (Highest)
> 2. Other Admin Messages
> 3. Pipeline/Data Messages (Lowest)
> 
> * **Priority Inversion:** There is a specific exception for Migration messages. Due to how they are re-queued in certain scenarios, the logic strictly drains the pipeline queue first to prevent deadlocks.
> * **Temporary Implementation details:** Since `AsyncFiber` still shares execution logic for all message types in this phase, `PipelineMessagePtr` remains wrapped in `MessageHandle` with adjusted memory statistics. This is a temporary bridge; Phase 2 will introduce complete separation and remove this wrapper.

Thanks, looks good. Can you tell what will contain phase 2?

glevkovich:
> > **PR Summary & Phase 1 Rationale**
> > 
> > * **Transitional State:** `MessageHandle::handle` still holds `PipelineMessagePtr`. This design choice was made to minimize the scope of code changes in this PR. Removing `PipelineMessagePtr` entirely requires a complete separation of logic within `AsyncFiber`, which would significantly increase review complexity. That separation is being prepared for Phase 2.
> > * **Queuing Policy:**
> > 
> > 
> > 1. **Admin/Control:** All admin messages are pushed into `dispatch_q_`. Checkpoint messages continue to be placed at the front (behavior unchanged).
> > 2. **Data:** Pipeline messages are pushed into the separate "Parsed Commands Queue" (linked list).
> > 
> > 
> > * **Prioritization:** This split effectively changes the prioritization model from 2 levels to 3:
> > 
> > 
> > 1. Checkpoint Messages (Highest)
> > 2. Other Admin Messages
> > 3. Pipeline/Data Messages (Lowest)
> > 
> > 
> > * **Priority Inversion:** There is a specific exception for Migration messages. Due to how they are re-queued in certain scenarios, the logic strictly drains the pipeline queue first to prevent deadlocks.
> > * **Temporary Implementation details:** Since `AsyncFiber` still shares execution logic for all message types in this phase, `PipelineMessagePtr` remains wrapped in `MessageHandle` with adjusted memory statistics. This is a temporary bridge; Phase 2 will introduce complete separation and remove this wrapper.
> 
> Thanks, looks good. Can you tell what will contain phase 2?

**Phase 2: Finalizing the Refactoring Task**

The goal of Phase 2 is to achieve a clean separation of concerns within `AsyncFiber` and `Connection` logic, removing the temporary "artificial code" introduced in Phase 1.

* **Remove PipelineMessagePtr from MessageHandle:** This is the core change. `MessageHandle` will be strictly reserved for Admin/Control messages (like PubSub, Monitor, Checkpoints). Pipeline commands will no longer be wrapped in a variant, eliminating unnecessary overhead.
* **Decouple Execution Paths in AsyncFiber:** Currently, `AsyncFiber` treats all requests uniformly as `MessageHandle` objects. In Phase 2, the execution loop will have two distinct, optimized paths:
1. **Admin Path:** Continues to process `MessageHandle` objects via the existing `dispatch_q_`.
2. **Data Path:** Processes `ParsedCommand` objects directly from the pipeline queue without the overhead of variant unwrapping.


* **Eliminate SendAsync for Pipelines:** Since pipeline messages will no longer be wrapped in `MessageHandle`, they cannot pass through `SendAsync`. We will introduce a dedicated,  function (e.g., `EnqueueParsedCommand`) to handle the queuing and notification for pipeline traffic directly.
* **Refactor Statistics:** With the removal of the wrapper, memory accounting will be simplified. We will remove the manual adjustments currently required to account for the `MessageHandle` size in pipeline statistics, making the tracking code cleaner and more accurate.

src/facade/dragonfly_connection.cc
glevkovich: We decided not to go with such separation but keep the original design.

src/facade/dragonfly_connection.cc
glevkovich: done

src/facade/dragonfly_connection.cc
glevkovich: done

src/facade/dragonfly_connection.cc
glevkovich: OK, i'll try to keep the assignment where possible in my next PRs,  to keep the code style uniform.

src/facade/dragonfly_connection.cc
glevkovich: removed

src/facade/dragonfly_connection.cc
glevkovich: done

src/facade/dragonfly_connection.h
```diff
@@ -409,7 +418,17 @@ class Connection : public util::Connection {
...
+
+  // Returns true if there are any commands pending in the parsed command queue or dispatch queue.
+  bool IsCommandsPending() const {
+    return parsed_head_ || !dispatch_q_.empty();
+  }
+
+  // Returns total count of commands pending in the parsed command queue and dispatch queue.
+  size_t GetPendingCommandCount() const {
+    return parsed_cmd_q_len_ + dispatch_q_.size();
+  }
```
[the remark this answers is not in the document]
glevkovich: applied renames

src/facade/dragonfly_connection.cc
```diff
@@ -1578,19 +1618,19 @@ void Connection::AsyncFiber() {
...
     uint64_t cur_epoch = fb2::FiberSwitchEpoch();
-    if (dispatch_q_.size() == 1 && cur_epoch == prev_epoch) {
+    if ((parsed_cmd_q_len_ == 1) && (cur_epoch == prev_epoch)) {
       if (pipeline_wait_batch_usec > 0) {
         ThisFiber::SleepFor(chrono::microseconds(pipeline_wait_batch_usec));
       } else {
         ThisFiber::Yield();
       }
-      DVLOG(2) << "After yielding to producer, dispatch_q_.size()=" << dispatch_q_.size();
+      DVLOG(2) << "After yielding to producer, parsed_cmd_q_len_=" << parsed_cmd_q_len_;
```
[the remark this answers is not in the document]

src/facade/dragonfly_connection.cc
```diff
@@ -1813,17 +1901,27 @@ void Connection::SendAsync(MessageHandle msg) {
...
+    EnqueueParsedCommand(pmsg.release());
   } else {
     dispatch_q_.push_back(std::move(msg));
   }
 
   // Don't notify if a sync dispatch is in progress, it will wake after finishing.
-  if (dispatch_q_.size() == 1 && !cc_->sync_dispatch) {
+  if (GetPendingCommandCount() == 1 && !cc_->sync_dispatch) {
     cnd_.notify_one();
   }
```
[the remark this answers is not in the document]

src/facade/dragonfly_connection.cc
```diff
@@ -1512,6 +1518,25 @@ void Connection::ClearPipelinedMessages() {
...
+  }
+  parsed_head_ = nullptr;
+  parsed_tail_ = nullptr;
+  parsed_to_execute_ = nullptr;
+
+  DCHECK_EQ(pending_pipeline_cmd_cnt_, 0u);
+  DCHECK_EQ(parsed_cmd_q_len_, 0u);
+  DCHECK_EQ(pending_pipeline_bytes_, 0u);
+  pending_pipeline_cmd_cnt_ = 0;
+  pending_pipeline_bytes_ = 0;
```
kostasrim: Do we really need any of these resets here ? Isn't it the case that: `ClearPipelinedMessages` is only called at the end of ConnectionFlow which means that the connection is closing soon and the object will be deallocated ? Why do I need to reset those members if I am about to deallocate the relevant object ? Is there something I missed here?

p.s. The DCHECKS are harmless but the rest of assigments seems redundant.

src/facade/dragonfly_connection.h
```diff
@@ -376,8 +383,10 @@ class Connection : public util::Connection {
   util::fb2::EventCount io_event_;
   std::optional<WaitEvent> current_wait_;
 
-  uint64_t pending_pipeline_cmd_cnt_ = 0;  // how many queued Redis async commands in dispatch_q
-  size_t pending_pipeline_bytes_ = 0;      // how many bytes of the queued Redis async commands
+  uint64_t pending_pipeline_cmd_cnt_ =
+      0;  // Count of queued Redis async commands in Parsed Commands queue
+  size_t pending_pipeline_bytes_ =
+      0;  // Total byte size of queued Redis async commands in Parsed Commands queue
```
kostasrim:
```suggestion
  // Count of queued Redis async commands in Parsed Commands queue
  uint64_t pending_pipeline_cmd_cnt_ = 0;
  // Total byte size of queued Redis async commands in Parsed  Commands queue
  size_t pending_pipeline_bytes_ = 0;
```

src/facade/dragonfly_connection.cc
```diff
@@ -1430,26 +1432,24 @@ bool Connection::ShouldEndAsyncFiber(const MessageHandle& msg) {
...
+  // We do not modify parsed_to_execute_ yet, in case execution throws/fails.
+  auto exec_cmd_ptr{parsed_to_execute_};
+  auto get_next_fn = [&exec_cmd_ptr]() mutable -> ParsedArgs {
+    DCHECK(exec_cmd_ptr);
+    facade::ParsedCommand* cmd = exec_cmd_ptr;
+
+    // Note: We don't need to check for nullptr here because DispatchManyCommands
+    // is guaranteed to call this exactly 'pipeline_count' times.
+    exec_cmd_ptr = exec_cmd_ptr->next;
+    return ParsedArgs{*cmd};
```
kostasrim:
```suggestion
    DCHECK(exec_cmd_ptr);
    return ParsedArgs{*std::exchange(exec_cmd_ptr, exec_cmd_ptr->next)}
```

kostasrim: In general `std::exchange` is a very useful utility function :)

src/facade/dragonfly_connection.cc
```diff
@@ -1601,27 +1639,66 @@ void Connection::AsyncFiber() {
...
+
+      // If the front message is a Migration Request, but we still have pipeline data
+      // (parsed_head_), we must block the migration and run the pipeline first.
+      bool is_blocked_migration{
+          !dispatch_q_.empty() &&
+          std::holds_alternative<MigrationRequestMessage>(dispatch_q_.front().handle) &&
+          (parsed_head_ != nullptr)};
+
+      // Process Admin Queue - only if not blocked
+      if (!dispatch_q_.empty() && !is_blocked_migration) {
```
kostasrim:
```suggestion
      if (!is_blocked_migration) {
```

kostasrim: redundant because the former check is included (is the first condition) in `is_blocked_migration`.

Out of curiosity (and not that I really care), what's the rationale to prefer `{}` over `=` for initialization ?

src/facade/dragonfly_connection.cc
```diff
@@ -1813,17 +1899,35 @@ void Connection::SendAsync(MessageHandle msg) {
...
-  if (dispatch_q_.size() == 1 && !cc_->sync_dispatch) {
+  // Pipeline path: Notify if this is the ONLY work pending.
+  // If other work exists, the fiber is already awake/scheduled
+  //
+  // Admin/Control Path:
+  // - For Memcache: Must notify if dispatch_q_ became non-empty (0 -> 1).
+  // - RESP: Also safe to notify if dispatch_q_ became non-empty,
+  // even if parsed_q has items (redundant notify is cheap/harmless).
+  bool should_notify =
+      msg.IsPipelineMsg() ? (GetPendingMessageCount() == 1) : (dispatch_q_.size() == 1);
```
kostasrim, line 1925: Since pipeline messages are resp only, the former check is an `implicit check for resp protocol` right ? (the answer should be yes -- jsut clarifying)

kostasrim: Also, maybe I am missing something here but why is the check memcache here is needed ? Isn't it the case that `memcache` now uses exclusively the connection fiber to dispatch commands async ? Judging from the implementation of `ParseMemcache` that should be the case. It's only ParseRedis that is currently using the `AsyncFb`. 

Am I missing something ?

src/facade/dragonfly_connection.cc
```diff
@@ -1813,17 +1899,35 @@ void Connection::SendAsync(MessageHandle msg) {
...
   if (msg.IsCheckPoint()) {
     auto it = dispatch_q_.begin();
     while (it < dispatch_q_.end() && it->IsCheckPoint())
       ++it;
     dispatch_q_.insert(it, std::move(msg));
+  } else if (msg.IsPipelineMsg()) {
+    auto& pmsg = std::get<PipelineMessagePtr>(msg.handle);
+    // Transfer the dispatch time from the Handle to the Command
+    // so we calculate latency from the moment SendAsync was called.
+    pmsg->parsed_cycle = msg.dispatch_cycle;
```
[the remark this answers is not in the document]
glevkovich: I added a comment in header, I still use it for queue latency, we do not want to waste extra 8 bytes to keep 2 different fields.

src/facade/dragonfly_connection.cc
glevkovich: I will remove the 3 lines here, nulling 2 pointers is a must. and keep the DCHECKs. thanks.

src/facade/dragonfly_connection.h
glevkovich: Thanks, didn't see it with the autoformat.

src/facade/dragonfly_connection.cc
glevkovich: done

src/facade/dragonfly_connection.cc
glevkovich: > Since pipeline messages are resp only, the former check is an implicit check for resp protocol right ? (the answer should be yes -- jsut clarifying)

Correct. Pipeline messages are specific to RESP, so GetPendingMessageCount() > 0 while dispatch_q_ is empty implies a Redis pipeline. For Memcached, the data commands are indeed processed synchronously in the connection fiber, but we must still check dispatch_q_ specifically.

> Also, maybe I am missing something here but why is the check memcache here is needed ? Isn't it the case that memcache now uses exclusively the connection fiber to dispatch commands async ? Judging from the implementation of ParseMemcache that should be the case. It's only ParseRedis that is currently using the AsyncFb.

While ParseMemcache handles data commands synchronously, Memcached connections still rely on the AsyncFiber for the Control Path. Administrative events (like CLIENT KILL), CheckpointMessages for snapshots, and MigrationRequestMessages are all pushed to dispatch_q_. If we don't notify when dispatch_q_ transitions from 0 to 1, the AsyncFiber for a Memcached connection could sleep indefinitely while an admin message sits in the queue. MemCacche user SendAsync , and SendAsync calls also:

```
  if (!cc_->conn_closing) {
    LaunchAsyncFiberIfNeeded();
  }
```

src/facade/dragonfly_connection.cc
glevkovich: his code is confusing and im going to refactor it to have dispatch_q_.empty()  just once. So the discussion doesn't matter  anymore.

in short: uniform init is safer. more compiler warnings/errors.
Here is a more official explanation: 
image

src/facade/dragonfly_connection.cc
```diff
@@ -1601,27 +1630,66 @@ void Connection::AsyncFiber() {
...
       SquashPipeline();
     } else {
-      MessageHandle msg = std::move(dispatch_q_.front());
-      dispatch_q_.pop_front();
+      MessageHandle msg;
+
+      // If the front message is a Migration Request, but we still have pipeline data
+      // (parsed_head_), we must block the migration and run the pipeline first.
+      bool is_blocked_migration{
+          !dispatch_q_.empty() &&
```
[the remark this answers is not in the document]
glevkovich: fixed

src/facade/dragonfly_connection.cc
```diff
@@ -1476,34 +1473,38 @@ void Connection::SquashPipeline() {
...
-      stats_->pipelined_wait_latency += CycleClock::ToUsec(start - rit->dispatch_cycle);
-    } else {
-      rit->dispatch_cycle = 0;  // Reset dispatch cycle to avoid accounting inside RecycleMessage
+      stats_->pipelined_wait_latency += CycleClock::ToUsec(start - current->parsed_cycle);
     }
-    RecycleMessage(std::move(*rit));
-  }
 
-  dispatch_q_.erase(it, it + dispatched);
+    ReleaseParsedCommand(current, true /* is_pipelined */);
```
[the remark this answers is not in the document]
glevkovich: done

src/facade/dragonfly_connection.cc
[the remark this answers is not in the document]
glevkovich: done

PR #6424, opened 2026-01-20
chore: support listpack with lists

romange:
The code handles listpack encoding but it's effectively disabled as the code that creates
the data-structure creates only qlist.

The next step would be to create listpack as well as implement promotion to qlist when needed.

### DIFF
--- a/src/server/list_family.cc
+++ b/src/server/list_family.cc
@@ -5,10 +5,12 @@ extern "C" {
 #include "redis/sds.h"
 }
 
+#include <absl/functional/overload.h>
 #include <absl/strings/numbers.h>
 
 #include "base/flags.h"
 #include "base/logging.h"
+#include "core/detail/listpack.h"
 #include "core/qlist.h"
 #include "facade/cmd_arg_parser.h"
 #include "server/acl/acl_commands_def.h"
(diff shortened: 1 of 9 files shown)


src/server/list_family.cc
```diff
@@ -77,63 +78,203 @@ struct ListWrapper {
...
+
+      if (match) {
+        int where = (insert_opt == QList::BEFORE) ? LP_BEFORE : LP_AFTER;
+        listpack =
+            lpInsertString(listpack, (unsigned char*)elem.data(), elem.size(), p, where, nullptr);
+        return true;
+      }
+      p = lpNext(listpack, p);
+    }
+    return false;
```
dranikpg: The absence of any abstractions makes it unnecessarily verbose. This is why I created ListpackWrap when I worked with listpack for hashes

romange: i forgot about ListpackWrap, I will use it

romange:
augment review

src/server/list_family.cc
```diff
@@ -63,79 +65,109 @@ ABSL_FLAG(int32_t, list_compress_depth, 0, "Compress depth of the list. Default
...
-struct ListWrapper {
-  unsigned encoding = 0;
-  union {
-    QList* ql = nullptr;
-    uint8_t* listpack;
-  };
+class ListWrapper {
+  using LP = detail::ListPack;
+
+  std::variant<QList*, LP> impl_;
```
dranikpg, line 79: 👍🏻

src/server/list_family.cc
```diff
@@ -63,79 +65,109 @@ ABSL_FLAG(int32_t, list_compress_depth, 0, "Compress depth of the list. Default
...
+
+ public:
+  template <typename T> explicit ListWrapper(T&& t) : impl_(std::forward<T>(t)) {
+  }
 
   size_t Size() const {
-    return ql->Size();
+    return visit(
+        Overload{[](QList* ql) { return ql->Size(); }, [](const LP& lp) { return lp.Size(); }},
+        impl_);
```
dranikpg, line 88: I have this hack in hset_family to handle equal interfaces but different reference/pointer types with a single function

 ```c++
template <typename F> decltype(auto) VisitRef(F f) const {  // Cast T* to T&
    return std::visit(Overloaded{[&f](auto* s) { return f(*s); }, f}, impl_);
 }
```

romange: i wondered what is it.

PR #6426, opened 2026-01-20
chore: factor out ContainerEntry and QList::Entry into a common class

romange:


src/core/collection_entry.h
```diff
@@ -0,0 +1,88 @@
...
+
+  // Assumes value is not null.
+  std::string_view view() const {
+    return {value_, length_};
+  }
+
+  // compatibility method
+  std::string to_string() const {
+    return ToString();
+  }
```
dranikpg, line 59: maybe its easier to fix usages than providing multiple functions with different naming 😅

PR #6427, opened 2026-01-20
docs: add async-tiering.md

romange:

### DIFF
--- a/docs/async-tiering.md
+++ b/docs/async-tiering.md
@@ -67,7 +67,9 @@ sequenceDiagram
 
 The coordinator fiber schedules a command on a shard thread. The command performs initial work, issues an asynchronous read, and returns a `ResultFuture` to the coordinator. The coordinator waits for fulfillment before replying. This parallelism hides most I/O latency (assuming non-saturated SSDs).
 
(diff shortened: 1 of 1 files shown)


docs/async-tiering.md
```diff
@@ -0,0 +1,163 @@
...
+    Coordinator->>Shard: Get
+    Shard->>Disk: IO_Read
+    Shard-->>Coordinator: ResultFuture
+    Disk-->>Shard: ReadCallback
+    Shard-->>Coordinator: ResultFulfilled
+```
+
+The coordinator fiber schedules a command on a shard thread. The command performs initial work, issues an asynchronous read, and returns a `ResultFuture` to the coordinator. The coordinator waits for fulfillment before replying. This parallelism hides most I/O latency (assuming non-saturated SSDs).
+
+For complex operations like `APPEND`, a post-read handler runs on the shard thread. Since in-place disk modification isn't supported, `APPEND` becomes an IO-READ followed by a handler that modifies the value in memory. The updated value is then returned to the coordinator.
```
dranikpg: A few details are missing:

1. complex = read + modify
2. After a modification is performed, the value is uploaded to memory and deleted from disk. Every new access will see the value in memory.
3. Multiple async operations are executed atomically consecutively and only a single read is issued. This is crucial for correctness, so that partial results can never be observed.

dranikpg (approved):
Pushed my own commit with extended description of the operation loop after a disk read

romange:
thank you, sir!

PR #6430, opened 2026-01-21
fix: static assert on macos

vyavdoshenko:
This static_assert is flaky on macOS:
https://github.com/dragonflydb/dragonfly/pull/6239
https://github.com/dragonflydb/dragonfly/pull/6293

I suggest leaving it for Linux only.


vyavdoshenko:
Broken build: https://github.com/dragonflydb/dragonfly/actions/runs/21199318567/job/60981615773

vyavdoshenko:
The check of this change: https://github.com/dragonflydb/dragonfly/actions/runs/21206189482

PR #6432, opened 2026-01-21
chore: add ListPack class that wraps c listpack

romange:
Its APIs are aligned with actual actions needed by list_family.


dranikpg:
By providing an iterator you can implement many operations on top of it

romange:
I agree but it's too much work for me and I do not want to block on it.

PR #6434, opened 2026-01-21
feat(server): add TLS cert info (CN, issue/expiry dates) to INFO server

### LINKED ISSUE #3292: Include TLS information into INFO command output
Include certification information such as name, expiration date, issue date

BenjaminGoehry:
Closes #3292


BenjaminGoehry:
image

moredure:
@BenjaminGoehry I am not completely sure have you checked does dragonfly supports multiple certs per single server, if so maybe worth to output more information, probably as sort of list , have no idea how it's better to be represented in this case.

BenjaminGoehry:
@moredure  I checked whether Dragonfly supports multiple TLS certificates. According to the Listener class, only one certificate is loaded into the SSL_CTX. Additionally, the TLS configuration flags only allow specifying one certificate and key pair.
Based on this, it appears that Dragonfly does not currently support multiple certificates per server.

BenjaminGoehry:
@moredure Just following up. Does the TLS output look good to you? Happy to make any adjustments needed! 🙏

moredure:
> @moredure Just following up. Does the TLS output look good to you? Happy to make any adjustments needed! 🙏

looks good, might need to check how it would look like with wildcard/ multiple domains.

src/facade/dragonfly_listener.cc
```diff
@@ -221,9 +271,11 @@ bool Listener::ReconfigureTLS() {
     if (!ctx) {
       return false;
     }
+    cert_info_ = GetCertInfoFromCtx(ctx);
     ctx_ = ctx;
   } else {
     ctx_ = nullptr;
+    cert_info_ = TLSCertificateInfo{};
```
romange: probably redundant.

BenjaminGoehry: Thanks! I removed it 🚀

romange:
this does not build

PR #6437, opened 2026-01-21
feat(search): Build vector index by using data pointer for HSET objects

mkaruza:
For hash set objects don't construct temporary vector and pass pointer to data during index adding. If pointer is null (i.e. field doesn't exists) use special null_pointer_ (created in hnsw index) to use as data. For JSON objects we still need to copy vector.


romange:
augmentcode review

PR #6440, opened 2026-01-21
chore: harden listpack checks in t_stream code

romange:


romange:
I do not think we will need this PR

Issue #6435, opened 2026-01-21
`hll_family_test` aborts during CI


abhijat:
The failure log varies but is closely associated to listing stats during deadlock

```
E20260121 12:35:22.708484 19521 test_utils.cc:298] TxLocks for shard 1
E20260121 12:35:22.708540 19521 test_utils.cc:306] Transaction for shard 1
F20260121 12:35:22.708611 19521 init.cc:28] [/__w/dragonfly/dragonfly/build/_deps/abseil_cpp-src/absl/container/inlined_vector.h:373]: assert(false && "i < size()") failed!
```

Assert failure sometimes during raw hash set access

abhijat:
https://github.com/dragonflydb/dragonfly/actions/runs/21208595701/job/61011078104?pr=6432#step:9:22008

abhijat:
https://github.com/dragonflydb/dragonfly/actions/runs/21130314222/job/60759698981?pr=6368
https://github.com/dragonflydb/dragonfly/actions/runs/21086766895/job/60751865780

always seems to fail for:

ubuntu-dev:24, Debug, clang++, clang

abhijat:
can't reproduce this locally (docker container, same as ci ubuntu-24 + clang,debug). https://github.com/dragonflydb/dragonfly/pull/6444 for temporary debugging

abhijat:
```
F20260122 06:21:11.755468 11815 init.cc:28] [/__w/dragonfly/dragonfly/build/_deps/abseil_cpp-src/absl/container/inlined_vector.h:373]: assert(false && "i < size()") failed!
*** Check failure stack trace: ***
    @     0x5633a16b3a69  google::LogMessage::Fail()
    @     0x5633a16ae226  google::LogMessage::SendToLog()
    @     0x5633a16b27e1  google::LogMessage::Flush()
    @     0x5633a16bfcfb  google::LogMessageFatal::~LogMessageFatal()
    @     0x5633a168cb01  __assert_fail
    @     0x56339d87df00  _ZZNK4absl12lts_2025051213InlinedVectorIN4dfly11Transaction12PerShardDataELm4ESaIS4_EEixEmENKUlvE_clEv
    @     0x56339d87dc67  absl::lts_20250512::InlinedVector<>::operator[]()
    @     0x5633a01878a3  dfly::Transaction::IsActive()
[...]

PR #6446, opened 2026-01-22
tests: Check if tx is initialized before accessing its state

abhijat:
A transaction may be created (non null) but not fully initialized yet. The watchdog timer in test-utils can find a tx in such a place, resulting in a crash due to the asserts in the tx member variables, specifically the inlined vector that holds shard data.

Also adds a lock while accessing connections within test utils, since another assert was seen recently when accessing that field.

FIXES https://github.com/dragonflydb/dragonfly/issues/6435

### DIFF
--- a/src/server/test_utils.cc
+++ b/src/server/test_utils.cc
@@ -304,9 +304,11 @@ void BaseFamilyTest::ResetService() {
           }
 
           LOG(ERROR) << "Transaction for shard " << es->shard_id();
+          std::unique_lock conn_lck{mu_};
           for (auto& conn : connections_) {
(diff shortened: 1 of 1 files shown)


src/server/transaction.h
```diff
@@ -392,6 +392,10 @@ class Transaction {
            " local_res=" + std::to_string(int(local_result_));
   }
 
+  bool IsInitialized() const {
```
romange: you should be able to check it with ` bool IsScheduled() `

abhijat: I have fixed it and tests are passing

PR #6448, opened 2026-01-22
Add hnsw serialization

BorysTheDev:
Added HNSW nodes serialization code for every HASH and JSON key

For every key (hash, json) I check presence in the indexes, if the index is hnsw and contains the key, I take the internal hnsw data (links) and store it together with the key.

### DIFF
--- a/src/server/snapshot.cc
+++ b/src/server/snapshot.cc
@@ -15,6 +15,8 @@
 #include "server/journal/journal.h"
 #include "server/rdb_extensions.h"
 #include "server/rdb_save.h"
+#include "server/search/doc_index.h"
+#include "server/search/global_hnsw_index.h"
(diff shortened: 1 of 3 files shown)


src/server/snapshot.cc
```diff
@@ -295,7 +297,74 @@ void SliceSnapshot::SerializeEntry(DbIndex db_indx, const PrimeKey& pk, const Pr
     CHECK(res);
     ++type_freq_map_[*res];
   }
+
+#ifdef WITH_SEARCH
```
romange, line 301: nit: to avoid multiple #ifdefs  
you can keep the function definition and add ifdef in its body

BorysTheDev:
we decided to proceed with alternative implementation

PR #6454, opened 2026-01-23
feat(search): Omit key defrag is vector data is used as reference

mkaruza:
If we don't copy vector data and use it as reference in vector index we should skip defragmentation of key. Change functions definitions so that PrimeValue object can be modified when needed.


dranikpg:
It's never cleared indeed, but I assume that it's fine for now as df would be used exclusively as a vector db. The nicest (but cumbersome) solution would be to a have a flag in json/string map.

Issue #4713, opened 2025-03-06
remove extra includes


BorysTheDev:
We need to check our code with "Include what you use" tool and remove extra headers, this should reduce compilation time

dranikpg:
```
Start initialization project ...
Start analyze sources ...
Start report results ...
Most impact files:
1 : "absl/types/span.h" impact on 94 files
2 : "absl/container/inlined_vector.h" impact on 88 files
3 : "base/iterator.h" impact on 79 files
4 : "common/arg_range.h" impact on 78 files
5 : "absl/container/flat_hash_set.h" impact on 75 files
6 : "absl/container/flat_hash_map.h" impact on 73 files
7 : "facade/op_status.h" impact on 70 files
8 : "common/backed_args.h" impact on 69 files
9 : "strings/human_readable.h" impact on 68 files
10 : "facade/facade_types.h" impact on 67 files
11 : "base/pmr/memory_resource.h" impact on 66 files
12 : "sys/types.h" impact on 63 files
13 : "redis/sds.h" impact on 57 files
14 : "util/fibers/synchronization.h" impact on 53 files
15 : "core/detail/stateless_allocator.h" impact on 52 files
16 : "util/fibers/fibers.h" impact on 51 files
17 : "server/tx_base.h" impact on 49 files
18 : "server/common.h" impact on 46 files
19 : "mimalloc.h" impact on 43 files
20 : "core/json/detail/interned_blob.h" impact on 43 files
21 : "jsoncons/json.hpp" impact on 42 files
22 : "jsoncons_ext/jsonpath/jsonpath.hpp" impact on 42 files
23 : "core/json/detail/interned_string.h" impact on 42 files
24 : "core/json/json_object.h" impact on 41 files
25 : "absl/base/internal/endian.h" impact on 39 files
26 : "common/string_or_view.h" impact on 39 files
27 : "core/mi_memory_resource.h" impact on 38 files
28 : "redis/redis_aux.h" impact on 38 files
29 : "base/sse2neon.h" impact on 37 files
30 : "base/sse2rvv.h" impact on 37 files
31 : "boost/intrusive/list_hook.hpp" impact on 37 files
32 : "emmintrin.h" impact on 37 files
33 : "tmmintrin.h" impact on 37 files
34 : "vecintrin.h" impact on 37 files
35 : "core/small_string.h" impact on 37 files
36 : "absl/random/random.h" impact on 36 files
37 : "base/logging.h" impact on 36 files
[...]

PR #6455, opened 2026-01-23
chore(server): Try to simplify build

dranikpg:
We have long include chains, for example:
 * db_slice.h includes conn_context.h 
 * which includes acl_commands_def.h 
 * which includes command_registry.h (and conn_context.h to be two-sided)
 * and other includes down the line along the way 

So everything including just db_slice was including a lot of other dependencies as well, including dependencies from the `dragonfly_lib` target

1. Remove backref include of "conn_context.h"  from "db_slice.h" as they're from different targets and it's an unnecessary include, use pointer to atomic_bool* instead of ExecInfo*
2. Remove some uneccessary includes from acl_commands_def, tests, etc
3. Add CMakeListst.txt to journal and cluster just with the sources list for now

Aids #4713

### DIFF
--- a/src/server/CMakeLists.txt
+++ b/src/server/CMakeLists.txt
@@ -28,14 +28,14 @@ cxx_link(dfly_bench dfly_parser_lib fibers2 absl::random_random redis_lib)
 
 # Include journal sources (not separate target for now)
 add_subdirectory(journal)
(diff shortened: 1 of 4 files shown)


dranikpg:
augment review

dranikpg:
augment review

dranikpg:
augment review

PR #6368, opened 2026-01-11
[WIP]: use refcounted interned strings

abhijat:
WIP

PR #6456, opened 2026-01-23
feat(json): Adds a ref-counted blob

abhijat:
A new class is introduced which stores a string along with its reference count and size.

A blob contains one char * allocated on the mimalloc heap via stateless allocator. The blob contains a string, its size and reference count in the same allocated block of memory. 

The size and ref count are stored right before the char data. The char data end with a nul character so as to be able to use the char* directly as a string. This is important for `jsoncons` which expects such an API. 

There are custom hash and eq operators which allow for heterogenous lookup, ie do not require conversion to blob type before hashing/comparison, and only consider the string content for checks, ignoring ref. count. 

An interned pool is introduced which is simply a flat set of blob pointers. The blob itself does not control its reference count. In the next PR a string class will be added, which will control refcount and itself store a pointer to a blob.

---

Note: for a broader context, see the spike PR https://github.com/dragonflydb/dragonfly/pull/6368 from which this one is derived.

### DIFF
--- a/src/core/json/detail/interned_blob.cc
+++ b/src/core/json/detail/interned_blob.cc
@@ -14,7 +14,7 @@ constexpr size_t kHeaderSize = sizeof(uint32_t) * 2;
 
 namespace dfly::detail {
 
-BlobPtr MakeBlobPtr(std::string_view sv) {
+InternedBlobHandle InternedBlobHandle::Create(std::string_view sv) {
   constexpr uint32_t ref_count = 1;
   DCHECK_LE(sv.size(), std::numeric_limits<uint32_t>::max());
 
@@ -30,7 +30,7 @@ BlobPtr MakeBlobPtr(std::string_view sv) {
 
   // null terminate so jsoncons can directly access the char* as string
   blob[kHeaderSize + str_len] = '\0';
-  return blob + kHeaderSize;
+  return InternedBlobHandle{blob + kHeaderSize};
 }
 
 uint32_t InternedBlobHandle::Size() const {
--- a/src/core/json/detail/interned_blob.h
+++ b/src/core/json/detail/interned_blob.h
@@ -17,16 +17,13 @@ namespace dfly::detail {
 //     ^-8      ^- 4         ^blob_
 using BlobPtr = char*;
 
-BlobPtr MakeBlobPtr(std::string_view sv);
-
 // A lightweight handle around a blob pointer, used to wrap the blob data when storing it in hashset
 // and also within interned strings. Does not handle lifetime of the data. Only provides convenience
 // methods to change state inside the blob and "view" style methods to access the string inside the
 // blob. Multiple handles can point to the same blob.
 class InternedBlobHandle {
  public:
-  explicit InternedBlobHandle(BlobPtr blob) : blob_{blob} {
-  }
+  static InternedBlobHandle Create(std::string_view sv);
 
   uint32_t Size() const;
 
@@ -61,7 +58,10 @@ class InternedBlobHandle {
   operator std::string_view() const;  // NOLINT (non-explicit operator for easier comparisons)
 
  private:
-  BlobPtr blob_;
+  explicit InternedBlobHandle(BlobPtr blob) : blob_{blob} {
+  }
+
+  BlobPtr blob_{nullptr};
 };
 
 struct BlobHash {
--- a/src/core/json/interned_blob_test.cc
+++ b/src/core/json/interned_blob_test.cc
@@ -37,16 +37,12 @@ using detail::InternedBlobHandle;
 TEST_F(InternedBlobTest, MemoryUsage) {
   const auto* mr = MemoryResource();
   const auto usage_before = mr->used();
-  {
-    BlobPtr p = detail::MakeBlobPtr("1234567");
-    InternedBlobHandle blob{p};
-    const auto usage_after = mr->used();
-    const auto expected_delta = blob.MemUsed();
-    EXPECT_EQ(usage_before + expected_delta, usage_after);
-    InternedBlobHandle::Destroy(blob);
-  }
+  InternedBlobHandle blob = InternedBlobHandle::Create("1234567");
   const auto usage_after = mr->used();
-  EXPECT_EQ(usage_before, usage_after);
+  const auto expected_delta = blob.MemUsed();
+  EXPECT_EQ(usage_before + expected_delta, usage_after);
+  InternedBlobHandle::Destroy(blob);
+  EXPECT_EQ(usage_before, mr->used());
 }
 
 void CheckBlob(InternedBlobHandle& blob, std::string_view expected, uint32_t ref_cnt = 1) {
@@ -56,66 +52,35 @@ void CheckBlob(InternedBlobHandle& blob, std::string_view expected, uint32_t ref
 }
 
 TEST_F(InternedBlobTest, Ctors) {
-  {
-    BlobPtr p = detail::MakeBlobPtr("");
-    InternedBlobHandle blob{p};
-    EXPECT_EQ(blob.Size(), 0);
-    InternedBlobHandle::Destroy(blob);
-  }
-
-  {
-    BlobPtr p = detail::MakeBlobPtr("foobar");
-    InternedBlobHandle src{p};
-    InternedBlobHandle dest{src};
-    CheckBlob(dest, "foobar");
-    CheckBlob(src, "foobar");
-    InternedBlobHandle::Destroy(dest);
-  }
-
-  std::string data(100000, 'x');
-  BlobPtr p = detail::MakeBlobPtr(data);
-  InternedBlobHandle blob{p};
-  EXPECT_EQ(blob.Size(), data.length());
+  auto blob = InternedBlobHandle::Create("");
+  EXPECT_EQ(blob.Size(), 0);
   InternedBlobHandle::Destroy(blob);
+
+  InternedBlobHandle src = InternedBlobHandle::Create("foobar");
+  InternedBlobHandle dest{src};
+  CheckBlob(dest, "foobar");
+  CheckBlob(src, "foobar");
+  InternedBlobHandle::Destroy(dest);
 }
 
 TEST_F(InternedBlobTest, Comparison) {
-  BlobPtr p = detail::MakeBlobPtr("foobar");
-  InternedBlobHandle blob{p};
-  const detail::BlobEq blob_eq;
+  auto blob = InternedBlobHandle::Create("foobar");
+  constexpr detail::BlobEq blob_eq;
 
   EXPECT_TRUE(blob_eq(blob, "foobar"));
   EXPECT_TRUE(blob_eq("foobar", blob));
 
-  InternedBlobHandle second{p};
+  InternedBlobHandle second = blob;
   second.SetRefCount(2000);
 
   EXPECT_TRUE(blob_eq(blob, second));
   InternedBlobHandle::Destroy(blob);
 }
 
-TEST_F(InternedBlobTest, Accessors) {
-  BlobPtr p = detail::MakeBlobPtr("1234567");
-  InternedBlobHandle blob{p};
-  EXPECT_EQ(blob.Size(), 7);
-  EXPECT_STREQ(blob.Data(), "1234567");
-  EXPECT_EQ(blob, "1234567"sv);
-  InternedBlobHandle::Destroy(blob);
-}
-
 TEST_F(InternedBlobTest, RefCounts) {
-  BlobPtr p = detail::MakeBlobPtr("1234567");
-  InternedBlobHandle blob{p};
+  auto blob = InternedBlobHandle::Create("1234567");
   EXPECT_EQ(blob.RefCount(), 1);
-  blob.IncrRefCount();
-  blob.IncrRefCount();
-  blob.IncrRefCount();
-  EXPECT_EQ(blob.RefCount(), 4);
-  blob.DecrRefCount();
-  blob.DecrRefCount();
-  blob.DecrRefCount();
-  blob.DecrRefCount();
-  EXPECT_EQ(blob.RefCount(), 0);
+  blob.SetRefCount(0);
   EXPECT_DEBUG_DEATH(blob.DecrRefCount(), "Attempt to decrease zero refcount");
   blob.SetRefCount(std::numeric_limits<uint32_t>::max());
   EXPECT_DEBUG_DEATH(blob.IncrRefCount(), "Attempt to increase max refcount");
@@ -124,8 +89,7 @@ TEST_F(InternedBlobTest, RefCounts) {
 
 TEST_F(InternedBlobTest, Pool) {
   detail::InternedBlobPool pool{};
-  BlobPtr p = detail::MakeBlobPtr("foo");
-  InternedBlobHandle b1(p);
+  InternedBlobHandle b1 = InternedBlobHandle::Create("foo");
   pool.emplace(b1);
 
   // search by string view
@@ -133,80 +97,6 @@ TEST_F(InternedBlobTest, Pool) {
 
   // increment the refcount. The blob is still found because the hasher only looks at the string
   b1.IncrRefCount();
-  b1.IncrRefCount();
-  b1.IncrRefCount();
-
   EXPECT_TRUE(pool.contains("foo"));
   InternedBlobHandle::Destroy(b1);
 }
-
-void BM_Getters(benchmark::State& state) {
-  InitTLStatelessAllocMR(MemoryResource());
-  BlobPtr p = detail::MakeBlobPtr("foobar");
-  InternedBlobHandle b{p};
-  for (auto _ : state) {
-    benchmark::DoNotOptimize(b.Size());
-    benchmark::DoNotOptimize(b.RefCount());
-    benchmark::DoNotOptimize(std::string_view{b});
-  }
-  InternedBlobHandle::Destroy(b);
-}
-
-BENCHMARK(BM_Getters);
-
-void BM_Setters(benchmark::State& state) {
-  InitTLStatelessAllocMR(MemoryResource());
-  BlobPtr p = detail::MakeBlobPtr("foobar");
-  InternedBlobHandle b{p};
-  for (auto _ : state) {
-    b.IncrRefCount();
-    benchmark::ClobberMemory();
-    b.DecrRefCount();
-    benchmark::ClobberMemory();
-  }
-  InternedBlobHandle::Destroy(b);
-}
-
-BENCHMARK(BM_Setters);
-
-void BM_Hash(benchmark::State& state) {
-  InitTLStatelessAllocMR(MemoryResource());
-  BlobPtr p = detail::MakeBlobPtr("typical_keys");
-  InternedBlobHandle b{p};
-  detail::BlobHash hasher;
-  for (auto _ : state) {
-    benchmark::DoNotOptimize(hasher(b));
-  }
-  InternedBlobHandle::Destroy(b);
-}
-BENCHMARK(BM_Hash);
-
-void BM_PoolLookup(benchmark::State& state) {
-  InitTLStatelessAllocMR(MemoryResource());
-  detail::InternedBlobPool pool;
-  BlobPtr ptr = detail::MakeBlobPtr("foobar!!!");
-  InternedBlobHandle blob(ptr);
-  pool.insert(blob);
-
-  for (auto _ : state) {
-    benchmark::DoNotOptimize(pool.find("foobar!!!"));
-  }
-
-  InternedBlobHandle::Destroy(blob);
-}
-BENCHMARK(BM_PoolLookup);
-
-void BM_LookupMiss(benchmark::State& state) {
-  InitTLStatelessAllocMR(MemoryResource());
-  detail::InternedBlobPool pool;
-  BlobPtr ptr = detail::MakeBlobPtr("foobar!!!");
-  InternedBlobHandle blob(ptr);
-  pool.insert(blob);
-
-  for (auto _ : state) {
-    benchmark::DoNotOptimize(pool.find("!!!foobar"));
-  }
-
-  InternedBlobHandle::Destroy(blob);
-}
-BENCHMARK(BM_LookupMiss);


abhijat:
benchmark numbers from release build look decent
```
  ┌────────────────────┬─────────┐
  │     Operation      │  Time   │
  ├────────────────────┼─────────┤
  │ Accessors          │ ~1.0 ns │
  ├────────────────────┼─────────┤
  │ Refcount setters   │ ~1.8 ns │
  ├────────────────────┼─────────┤
  │ Hash               │ ~2.8 ns │
  ├────────────────────┼─────────┤
  │ Pool lookup (hit)  │ ~2.8 ns │
  ├────────────────────┼─────────┤
  │ Pool lookup (miss) │ ~2.8 ns │
  └────────────────────┴─────────┘
```

src/core/json/detail/interned_blob.cc
```diff
@@ -0,0 +1,126 @@
...
+void InternedBlob::Destroy() {
+  if (blob_) {
+    const size_t to_destroy = kHeaderSize + Size() + 1;
+    StatelessAllocator<char>{}.deallocate(blob_ - kHeaderSize, to_destroy);
+    blob_ = nullptr;
+  }
+}
+
+size_t BlobHash::operator()(const InternedBlob* b) const {
+  return std::hash<std::string_view>{}(b->View());
```
abhijat: played with replacing these with `absl::Hash` but benchmark numbers dont really change much.

src/server/interned_blob_test.cc
```diff
@@ -0,0 +1,190 @@
...
+BENCHMARK(BM_Hash);
+
+void BM_PoolLookup(benchmark::State& state) {
+  InitTLStatelessAllocMR(MemoryResource());
+  detail::InternedBlobPool pool;
+  auto blob = std::make_unique<InternedBlob>("foobar!!!");
+  pool.insert(blob.get());
+
+  for (auto _ : state) {
+    pool.find("foobar!!!");
```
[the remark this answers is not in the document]
abhijat: fixed

src/core/json/interned_blob_test.cc
```diff
@@ -0,0 +1,190 @@
...
+  // increment the refcount. The blob is still found because the hasher only looks at the string
+  b1->IncrRefCount();
+  b1->IncrRefCount();
+  b1->IncrRefCount();
+
+  EXPECT_TRUE(pool.contains("foo"));
+}
+
+void BM_Getters(benchmark::State& state) {
+  InitTLStatelessAllocMR(MemoryResource());
```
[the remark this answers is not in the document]
abhijat: I dont think this is a problem, the benchmarks are run separately

src/core/json/detail/interned_blob.cc
```diff
@@ -0,0 +1,126 @@
...
+constexpr size_t int_size = sizeof(uint32_t);
+}
+
+namespace dfly::detail {
+
+InternedBlob::InternedBlob(const std::string_view sv) {
+  auto alloc = StatelessAllocator<char>{};
+
+  constexpr uint32_t ref_count = 1;
+  const uint32_t str_len = sv.size();
```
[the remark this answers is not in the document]
abhijat: added a check

src/core/json/detail/interned_blob.cc
```diff
@@ -0,0 +1,126 @@
...
+  return {blob_, Size()};
+}
+
+const char* InternedBlob::Data() const {
+  return blob_;
+}
+
+void InternedBlob::IncrRefCount() {  // NOLINT (cannot be made const)
+  const uint32_t ref_count = RefCount();
+  DCHECK_LT(ref_count, std::numeric_limits<uint32_t>::max()) << "Attempt to increase max refcount";
```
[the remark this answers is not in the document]
abhijat: I originally had ` CHECK` but I realised it is unrealistic for an overflow to occur in real life (4 bliion references for keys). And it eats into performance in release mode for no real benefit

src/core/detail/stateless_allocator.h
```diff
@@ -67,4 +67,8 @@ inline void InitTLStatelessAllocMR(PMR_NS::memory_resource* mr) {
   detail::tl_mr = mr;
 }
 
+inline void CleanupStatelessAllocMR() {
```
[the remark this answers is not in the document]
abhijat: This will be part of the next PR. I do not want to touch engine shard in this PR.

src/core/json/detail/interned_blob.h
```diff
@@ -0,0 +1,88 @@
...
+};
+
+// This pool holds blob pointers and is used by InternedString to manage string access. It would be
+// nice to keep this on the mimalloc heap by using StatelessAllocator. However, JSON memory usage is
+// estimated by comparing mimalloc usage before and after creating an object. If we keep this pool
+// on mimalloc, it can introduce variations such as resizing of its internal store when adding a new
+// object. This results in non-deterministic memory usage, which introduces incorrectness in tests
+// and the memory usage command. To keep memory estimation per object accurate, the pool is
+// allocated on the default heap.
+using InternedBlobPool = absl::flat_hash_set<InternedBlob*, BlobHash, BlobEq>;
```
romange: I am curious, why do you need to hash by `InternedBlob*` and not by `InternedBlob` ? 
in other words, why do we need two allocations to create a string?

src/core/json/interned_blob_test.cc
```diff
@@ -0,0 +1,190 @@
+// Copyright 2025, DragonflyDB authors.  All rights reserved.
```
romange, line 1: why this test is located here and not under src/core/ ?

src/core/json/detail/interned_blob.cc
```diff
@@ -0,0 +1,128 @@
...
+#include "core/detail/stateless_allocator.h"
+
+namespace {
+constexpr size_t kUint32Size = sizeof(uint32_t);
+}
+
+namespace dfly::detail {
+
+InternedBlob::InternedBlob(const std::string_view sv) {
+  auto alloc = StatelessAllocator<char>{};
```
romange: i find this notation unnecessary complicated. why not use `StatelessAllocator<char> alloc` ?

src/core/json/detail/interned_blob.cc
```diff
@@ -0,0 +1,128 @@
...
+  blob_[kHeaderSize + str_len] = '\0';
+  blob_ += kHeaderSize;
+}
+
+InternedBlob::~InternedBlob() {
+  Destroy();
+}
+
+InternedBlob::InternedBlob(InternedBlob&& other) noexcept : blob_(other.blob_) {
+  other.blob_ = nullptr;
```
romange: please move one-liners to a header file.

src/core/json/detail/interned_blob.h
```diff
@@ -0,0 +1,88 @@
...
+  std::string_view View() const;
+
+  // Returns nul terminated string
+  const char* Data() const;
+
+  // Increment ref count, asserts if count grows over type max limit
+  void IncrRefCount();
+
+  // Decrement ref count, asserts if count falls below 0
+  void DecrRefCount();
```
romange, line 48: why do you need `IncrRefCount, SetRefCount, DecrRefCount` functions being exposed? this is not the correct way to design your class, imho. Instead - either allow copy c'tors or introduce an function "Clone()" that creates a copy of the string with increased refcount.

src/core/json/detail/interned_blob.h
```diff
@@ -0,0 +1,88 @@
...
+
+// This pool holds blob pointers and is used by InternedString to manage string access. It would be
+// nice to keep this on the mimalloc heap by using StatelessAllocator. However, JSON memory usage is
+// estimated by comparing mimalloc usage before and after creating an object. If we keep this pool
+// on mimalloc, it can introduce variations such as resizing of its internal store when adding a new
+// object. This results in non-deterministic memory usage, which introduces incorrectness in tests
+// and the memory usage command. To keep memory estimation per object accurate, the pool is
+// allocated on the default heap.
+using InternedBlobPool = absl::flat_hash_set<InternedBlob*, BlobHash, BlobEq>;
+static_assert(sizeof(InternedBlob) == sizeof(char*));
```
romange: Also, I would expect to see some sort of manager that actually implements:
`InternedBlob CreateOrGet(std::string_view str)` functionality. I see the class, I see the hash_set but I do not see the code that manages unique strings.

src/core/json/interned_blob_test.cc
abhijat: this test needs `MiMemoryResource` which is not linked to the json cmake target. I thought of moving the mem. resource to json but decided on keeping the test here, instead of moving code around to serve the test.

src/core/json/detail/interned_blob.cc
abhijat: Yeah this slipped in while changing lines around, will fix.

src/core/json/detail/interned_blob.h
abhijat: for both your comment https://github.com/dragonflydb/dragonfly/pull/6456#discussion_r2731844364 and this one, I designed it so the blob is lower layer and a string wraps blobs and uses ref counting api to actually deduplicate the blobs. This way the blob class looks simpler and the string class also ends up being simpler, it can just focus on the `string` api that jsoncons expects, and the blob doesnt have to expose those methods (of which there are a lot). I plan to have the second pr introduce the string class. I did not add it here to keep loc low for easier review, but you can see the code here:

https://github.com/dragonflydb/dragonfly/pull/6368/changes/7a91c5969e3a6070f36c1905da4da9fd7fb8f91c

I can write a one pager to explain the design but its pretty straightforward in the end:

* blob - handles storage
* string - handles ref-count, decides when to create blob, when to destroy
* pool - stores blob pointers, this way on adding/removing items to pool we dont need to worry about iterator invalidation because the pool only stores pointers rather than blob directly - so the string doesnt need to store iterator into the pool and worry about invalidation

romange: I am sorry I do not understand why do you need iterators and worry about invalidations ? 
You can always create a InternedBlob from its pointer `blob_`. Its pointer has the state and `InternedBlob` is just a wrapper. So you do not need an access to specific `InternedBlob`, you just need an access to its variable.

Yes, please provide a design so I can see the whole picture.

src/core/json/interned_blob_test.cc
romange: it's fine to link test with more libs - we do it all the time. the dependency is added to the test, not to the `json` lib.

src/core/json/detail/interned_blob.h
romange: pls do not share corporate links in public repo

src/core/json/detail/interned_blob.h
```diff
@@ -0,0 +1,95 @@
...
+  // Decrement ref count, asserts if count falls below 0
+  void DecrRefCount();
+
+  void SetRefCount(uint32_t ref_count);
+
+  // Returns bytes used, including string, header and trailing byte
+  size_t MemUsed() const;
+
+  // Convenience method to deallocate storage. Not for use in destructor.
+  void Destroy();
```
romange: it should be a static method accepting `InternedBlobHandle` - not the data member.

src/core/json/detail/interned_blob.cc
```diff
@@ -0,0 +1,105 @@
...
+size_t BlobHash::operator()(const InternedBlobHandle& b) const {
+  return std::hash<std::string_view>{}(b.View());
+}
+
+size_t BlobHash::operator()(std::string_view sv) const {
+  return std::hash<std::string_view>{}(sv);
+}
+
+bool BlobEq::operator()(const InternedBlobHandle& a, const InternedBlobHandle& b) const {
+  return a.View() == b.View();
```
romange: that's a really bad implementation for a unique interned string. You can just compare pointers.

Also, please move all oneliners to header file. No need for so many boilerplate.

src/core/json/detail/interned_blob.cc
```diff
@@ -0,0 +1,105 @@
...
+
+void InternedBlobHandle::Destroy() {
+  if (blob_) {
+    const size_t to_destroy = kHeaderSize + Size() + 1;
+    StatelessAllocator<char>{}.deallocate(blob_ - kHeaderSize, to_destroy);
+    blob_ = nullptr;
+  }
+}
+
+size_t BlobHash::operator()(const InternedBlobHandle& b) const {
```
romange: create a casting operator `string_view operator() const` instead of `.View()` and then you won't need this override. string_view will suffice, imho.

src/core/json/detail/interned_blob.h
abhijat: I initially thought of making this static or even a free function like `DestroyBlob(Handle)`. 

The reason I made it an instance method is I want to keep the pointer (`blob_`) private and still use `Size()` The other option is to use `handle.Data()` to get the pointer to deallocate, but that gives a const pointer, and then I had to use const cast which I wanted to avoid. 

I agree that destroying is not really tied to the handle instance, but its just a convenience so that string can use this. I will explore making it a static method.

src/core/json/interned_blob_test.cc
abhijat: I will move this

src/core/json/detail/interned_blob.h
romange: you can use a static method inside class. it then has access to its private data

abhijat:
briefly converting to draft to run benchmarks, will check if there are changes in perf with the refactors

src/core/json/detail/interned_blob.h
abhijat: changed

src/core/json/interned_blob_test.cc
abhijat: moved to `core/json`

abhijat:
eliminating pointer chasing/indirection makes things much faster in pool interactions:

```
  ┌───────────────┬─────────┬─────────┬─────────────┐
  │   Benchmark   │ Before  │  After  │   Change    │
  ├───────────────┼─────────┼─────────┼─────────────┤
  │ BM_Getters    │ 1.11 ns │ 1.15 ns │ ~same       │
  ├───────────────┼─────────┼─────────┼─────────────┤
  │ BM_Setters    │ 1.86 ns │ 1.87 ns │ ~same       │
  ├───────────────┼─────────┼─────────┼─────────────┤
  │ BM_Hash       │ 3.15 ns │ 2.81 ns │ 11% faster  │
  ├───────────────┼─────────┼─────────┼─────────────┤
  │ BM_PoolLookup │ 2.82 ns │ 1.28 ns │ 2.2x faster │
  ├───────────────┼─────────┼─────────┼─────────────┤
  │ BM_LookupMiss │ 2.79 ns │ 1.15 ns │ 2.4x faster │
  └───────────────┴─────────┴─────────┴─────────────┘
```

src/core/json/detail/interned_blob.cc
abhijat: added operator and removed View

src/core/json/detail/interned_blob.cc
abhijat: fixed

src/core/json/detail/interned_blob.h
```diff
@@ -0,0 +1,95 @@
...
+
+// Layout is: 4 bytes size, 4 bytes refcount, char data, followed by nul-char.
+// The trailing nul-char is required because jsoncons needs to access c_str/data without a
+// size. The blob_ itself points directly to the data, so that callers do not have to perform
+// pointer arithmetic for c_str() and data() calls:
+//     [size:4] [refcount:4] [string] [\0]
+//     ^-8      ^- 4         ^blob_
+using BlobPtr = char*;
+
+BlobPtr MakeBlobPtr(std::string_view sv);
```
romange: nit: not sure why you need exposing BlobPtr, it's not clear from this PR.
you can add `static InternedBlobHandle Create(string_view sv)` that allocates and returns InternedBlobHandle. As an added bonus you can delete `explicit InternedBlobHandle(BlobPtr blob)` as it's not needed and rely on implicit c'tors instead.

src/core/json/detail/interned_blob.h
```diff
@@ -0,0 +1,95 @@
...
+  // Returns bytes used, including string, header and trailing byte
+  size_t MemUsed() const;
+
+  // Convenience method to deallocate storage. Not for use in destructor.
+  static void Destroy(InternedBlobHandle& handle);
+
+  operator std::string_view() const;  // NOLINT (non-explicit operator for easier comparisons)
+
+ private:
+  BlobPtr blob_;
```
romange:
```suggestion
  BlobPtr blob_ = nullptr;
```

src/core/json/interned_blob_test.cc
```diff
@@ -0,0 +1,212 @@
...
+};
+
+using detail::BlobPtr;
+using detail::InternedBlobHandle;
+
+TEST_F(InternedBlobTest, MemoryUsage) {
+  const auto* mr = MemoryResource();
+  const auto usage_before = mr->used();
+  {
+    BlobPtr p = detail::MakeBlobPtr("1234567");
```
romange: if your follow my suggestion you will reduce the boilerplate code in the tests as well

src/core/json/interned_blob_test.cc
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  std::string data(100000, 'x');
+  BlobPtr p = detail::MakeBlobPtr(data);
+  InternedBlobHandle blob{p};
+  EXPECT_EQ(blob.Size(), data.length());
+  InternedBlobHandle::Destroy(blob);
+}
+
+TEST_F(InternedBlobTest, Comparison) {
```
romange, line 66: i think tests are bloated for such a simple class. No need to check trivial features and you can combine many of the checks in the same test.

src/core/json/interned_blob_test.cc
```diff
@@ -0,0 +1,212 @@
...
+  // increment the refcount. The blob is still found because the hasher only looks at the string
+  b1.IncrRefCount();
+  b1.IncrRefCount();
+  b1.IncrRefCount();
+
+  EXPECT_TRUE(pool.contains("foo"));
+  InternedBlobHandle::Destroy(b1);
+}
+
+void BM_Getters(benchmark::State& state) {
```
romange: these "trivial" benchmarks are redundant, imho.

abhijat: I meant to remove them and not check them in, it was just a way to try to optimize but I think there is not much to optimize here, I will delete them now.

src/core/json/detail/interned_blob.h
abhijat: My plan was so that:

* InternedString (next PR) creates blob ptr
* It creates a handle around that and stores as its own field (or increase ref count)
* Another handle goes into the pool

But it doesn't need to create the blob ptr, it can just create the handle directly and place a copy in the pool.

I will change it

src/core/json/detail/interned_blob.h
abhijat: fixed

src/core/json/interned_blob_test.cc
abhijat: cleaned up some of the tests and the benchmark code.

src/core/json/interned_blob_test.cc
abhijat: removed

src/core/json/interned_blob_test.cc
abhijat: Changed the tests to use static method

src/core/json/detail/interned_blob.h
abhijat: I kept the explicit ctor but made it private, if I use implicit ctor I have to set the blob on the handle after construction, the private ctor is only used in the static factory style method.

src/core/json/detail/interned_blob.cc
abhijat: moved

src/core/json/detail/interned_blob.cc
abhijat: fixed

src/core/json/detail/interned_blob.cc
```diff
@@ -0,0 +1,85 @@
...
+  uint32_t ref_count;
+  std::memcpy(&ref_count, blob_ - kUint32Size, kUint32Size);
+  return ref_count;
+}
+
+void InternedBlobHandle::IncrRefCount() {  // NOLINT - non-const, mutates via ptr
+  const uint32_t ref_count = RefCount();
+  DCHECK_LT(ref_count, std::numeric_limits<uint32_t>::max()) << "Attempt to increase max refcount";
+  const uint32_t updated_count = ref_count + 1;
+  std::memcpy(blob_ - kUint32Size, &updated_count, kUint32Size);
```
dranikpg, line 54: `SetRefCount(RefCount() + 1)`

src/core/json/detail/interned_blob.cc
```diff
@@ -0,0 +1,85 @@
...
+}
+
+void InternedBlobHandle::DecrRefCount() {  // NOLINT - non-const, mutates via ptr
+  const uint32_t ref_count = RefCount();
+  DCHECK_GE(ref_count, 1ul) << "Attempt to decrease zero refcount";
+  const uint32_t updated_count = ref_count - 1;
+  std::memcpy(blob_ - kUint32Size, &updated_count, kUint32Size);
```
dranikpg, line 61: Same here, the dcheck for >= 0 is quite useful

src/core/json/detail/interned_blob.cc
```diff
@@ -0,0 +1,85 @@
...
+namespace dfly::detail {
+
+InternedBlobHandle InternedBlobHandle::Create(std::string_view sv) {
+  constexpr uint32_t ref_count = 1;
+  DCHECK_LE(sv.size(), std::numeric_limits<uint32_t>::max());
+
+  const uint32_t str_len = sv.size();
+
+  // We need +1 byte for \0 because jsoncons expects c_str() and data() style accessors on keys
+  BlobPtr blob = StatelessAllocator<char>{}.allocate(kHeaderSize + str_len + 1);
```
dranikpg, line 24: maybe if we aligned by 8 bytes, we could use the integers without memcpy

src/core/json/detail/interned_blob.cc
```diff
@@ -0,0 +1,85 @@
...
+  return blob_ ? Size() + kHeaderSize + 1 : 0;
+}
+
+void InternedBlobHandle::Destroy(InternedBlobHandle& handle) {
+  if (handle.blob_) {
+    const size_t to_destroy = kHeaderSize + handle.Size() + 1;
+    StatelessAllocator<char>{}.deallocate(handle.blob_ - kHeaderSize, to_destroy);
+    handle.blob_ = nullptr;
+  }
+}
```
dranikpg, line 78: could be just a member function

dranikpg: I see there was a discussion above and it changed from one to a static function 😆

romange: yes, I find it weird to have a non-owning wrapper that destroy itself.

src/core/json/detail/interned_blob.cc
romange: it's fine. Look at godbolt with `memcpy(,,4)` - it's a single load.

src/core/json/detail/interned_blob.cc
abhijat: I added `SetRefCount` later for tests, so didn't think of using it here, in fact I'm thinking of removing `SetRefCount` altogether. If not I will change it to this style in next PR

abhijat:
follow up PR is https://github.com/dragonflydb/dragonfly/pull/6497 which adds string

PR #6461, opened 2026-01-23
Add index serialization

BorysTheDev:
Add HNSW index serialization. this is alternative approach for https://github.com/dragonflydb/dragonfly/pull/6448
The serialization process is simple, iterating through all index entries and saving them using 0 thread
The index loading isn't implemented, so the data is skipped.
I will add globalId savingfor hashes and JSON in the separate PR

### DIFF
--- a/src/server/snapshot.cc
+++ b/src/server/snapshot.cc
@@ -164,7 +164,7 @@ void SliceSnapshot::FinalizeJournalStream(bool cancel) {
 void SliceSnapshot::SerializeGlobalHnswIndices() {
 #ifdef WITH_SEARCH
   // Serialize HNSW global indices for shard 0 only
-  if (db_slice_->shard_owner()->shard_id() != 0 || SaveMode() == RDB) {
+  if (db_slice_->shard_owner()->shard_id() != 0 || SaveMode() == dfly::SaveMode::RDB) {
     return;
   }
 


src/server/snapshot.cc
```diff
@@ -159,6 +161,45 @@ void SliceSnapshot::FinalizeJournalStream(bool cancel) {
...
+    constexpr size_t kBatchSize = 1000;
+    for (size_t i = 0; i < node_count; i += kBatchSize) {
+      size_t batch_end = std::min(i + kBatchSize, node_count);
+      auto nodes = index->GetNodesRange(i, batch_end);
+      for (const auto& node : nodes) {
+        tmp_buf.resize(node.TotalSize());
+        if (auto ec = serializer_->SaveHNSWEntry(node, absl::MakeSpan(tmp_buf)); ec)
+          break;
+      }
+      ThrottleIfNeeded();
```
dranikpg, line 197: I know it's just a call of `PushSerialied()` but it seems like this is the more appropriate call

src/server/rdb_save.cc
```diff
@@ -784,6 +785,37 @@ error_code RdbSerializer::SendJournalOffset(uint64_t journal_offset) {
   return WriteRaw(buf);
 }
 
+error_code RdbSerializer::SaveHNSWEntry(const search::HnswNodeData& node,
+                                        absl::Span<uint8_t> tmp_buf) {
+  // Binary format using little-endian encoding for efficiency:
+  // - internal_id: 4 bytes (uint32_t)
+  // - global_id: 8 bytes (uint64_t)
+  // - level: 4 bytes (int)
+  // - for each level (0 to level): links_num (4 bytes) + links (4 bytes each)
```
dranikpg, line 794: links = internal_id of link node

src/server/search/global_hnsw_index.h
```diff
@@ -33,6 +33,11 @@ class GlobalHnswIndexRegistry {
 
   bool Exist(std::string_view index_name, std::string_view field_name) const;
 
+  absl::flat_hash_map<std::string, std::shared_ptr<search::HnswVectorIndex>> GetAll() const {
+    std::shared_lock<std::shared_mutex> lock(registry_mutex_);
+    return indices_;
+  }
```
dranikpg, line 39: maybe you need to return them together with the mutex to prevent modifications during serialization?

BorysTheDev: we don't consider modification for this time

Issue #6431, opened 2026-01-21
Deadlock in `JournalStreamer::Cancel` (WaitForInflightToComplete) upon Abrupt Disconnection


boomballa:
**Describe the bug**
The Dragonfly server experiences a deadlock and eventually crashes (resulting in a watchdog abort) when a replica client that is actively sending `REPLCONF ACK` heartbeats disconnects abruptly (e.g., via `kill -9` or TCP RST).

The issue is located in `JournalStreamer::Cancel()`, specifically in the `WaitForInflightToComplete()` loop. It appears to be a race condition where `in_flight_bytes_` fails to returning to zero during the connection cleanup phase if the underlying connection enters an error state while both Read (ACK) and Write (Journal) operations are pending.

**To Reproduce**
1.  Connect a custom replica client to Dragonfly.
    *   *Context: We are developing a high-performance synchronization tool called **[df2redis](https://github.com/boomballa/df2redis)** (simulating a Dragonfly replica).*
2.  Enter the **Stable Sync** phase (streaming journal data).
3.  Ensure the client sends **`REPLCONF ACK <lsn>`** periodically and frequently (e.g., every 1 second). **Bidirectional traffic is critical**—if the client only reads, the issue is hard to reproduce.
4.  Kill the client process abruptly using `SIGKILL` (`kill -9 <pid>`) to force a `TCP RST`.
5.  **Result**: Dragonfly hangs, stops processing commands on that shard, and is eventually killed by the Watchdog.

**Expected behavior**
Dragonfly should handle the abrupt disconnection gracefully. `JournalStreamer::Cancel` should time out or force-clean the `in_flight_bytes_` counter if the connection is dead, rather than waiting indefinitely for a callback that may never come.

**Environment (please complete the following information):**
 - OS: Debian GNU/Linux 11
 - Kernel: Linux 5.10.0-30-amd64 #1 SMP Debian 5.10.218-1 (2024-06-01) x86_64 GNU/Linux
 - Containerized?: Bare Metal
 - Dragonfly Version: v1.36.0

**Root Cause Analysis**
The deadlock occurs in `src/server/journal/streamer.cc` during the cancellation sequence.

### 1. The Deadlock Loop
In `JournalStreamer::Cancel()`, the code waits for all pending async writes to complete. This logic blocks the main thread/fiber indefinitely if the condition is never met.

**File**: `src/server/journal/streamer.cc`

```cpp
bool JournalStreamer::Cancel() {
[...]

PR #6463, opened 2026-01-24
fix(server): ignore SIGHUP to prevent termination on session closing

boomballa:
## Description
This PR changes the [dragonfly](cci:1://file:///Users/boomballa/Desktop/Dev/AwesomeProject/df2redis/scripts/reproduce_dragonfly_crash.sh:86:0-108:1) server behavior to explicitly ignore `SIGHUP` signals.
Previously, [dragonfly](cci:1://file:///Users/boomballa/Desktop/Dev/AwesomeProject/df2redis/scripts/reproduce_dragonfly_crash.sh:86:0-108:1) did not handle `SIGHUP`, causing the process to terminate immediately if the parent shell session ended (e.g., SSH session timeout on a jump host, or simply closing the terminal while the process is running in the background without `nohup`).
This behavior led to abrupt terminations which triggered the deadlock described in #6431 (Deadlock in JournalStreamer::Cancel upon Abrupt Disconnection).
## Changes
- Added `signal(SIGHUP, SIG_IGN);` in [src/server/dfly_main.cc](cci:7://file:///Users/boomballa/Desktop/Dev/AwesomeProject/df2redis/dragonfly/src/server/dfly_main.cc:0:0-0:0) to ignore the hangup signal.
  - This brings the behavior in line with standard daemon practices, ensuring the server survives parent session termination.
## Verification
**Manual Test:**
1. Built the binary with the fix.
2. Launched dragonfly in background: `./dragonfly-x86_64 ... &`.
3. Sent SIGHUP signal: `kill -HUP <pid>`.
4. Verified that the process **did not terminate** and continued serving requests.
Before this fix, the same `kill -HUP` command caused the process to exit immediately.
## Related Issues
Fixes the root cause of the crash reported in #6431.

### DIFF
--- a/src/server/dfly_main.cc
+++ b/src/server/dfly_main.cc
@@ -906,6 +906,9 @@ Usage: dragonfly [FLAGS]
   sigemptyset(&act.sa_mask);
   sigaction(SIGILL, &act, nullptr);
 
+  // Ignore SIGHUP to prevent termination when the parent shell exits
+  signal(SIGHUP, SIG_IGN);
+
   if (GetFlag(FLAGS_port) == 0u) {
     string usock = GetFlag(FLAGS_unixsocket);
     if (usock.length() == 0u) {


romange:
please sign your commit.

dranikpg:
I was suprised Redis does the same. If it runs as a deamon its best to run it with a proper parent process

romange:
To me it's also weird - i would assume users run backends via systemd or nohup at least. Why would anyone run them directly from the shell?

boomballa:
Hi @romange ,
Thanks for the quick review! I've updated the PR with `sigaction` as suggested.

I checked the Redis source code and noticed that **Redis explicitly ignores SIGHUP** in `server.c:2770`, so it survives session disconnects even without daemonization.

Redis users often rely on `daemonize yes` in config files to ensure process stability and detach from the shell. 
However, **Dragonfly does not appear to support a `daemonize` flag**. This leaves users (especially on bare-metal via jump hosts) without a built-in mechanism to detach safely, making the process vulnerable to SIGHUP.

By explicitly ignoring SIGHUP in the code, we provide that missing layer of stability, aligning with what Redis users expect from a production-grade server.

Summary of changes in this push:
1.  Addressed the linter suggestion: Replaced `signal()` with `sigaction()`.
2.  Commit is signed-off.

Ready for another look!

romange:
Thanks @boomballa . The PR looks great and is approved. I am just trying to understand  your **production** methodology. I understand what jump host (or bastion is). I would still expect that there will be a script on the bare server that is run either systemd or some other background task management system and the eventual  process won't be dependent on the connecting shell. It's not the case here?  Are you guys using some sort of ansible approach where the orchestrator launches processes directly via ssh?

romange:
@boomballa  your commit is still not signed. Please rebase, squash and sign a single commit. While you are at it, I actually prefer the previous `signal` one-liner and not this version.

image

boomballa:
Hi @romange. Done. I've reverted to the `signal` one-liner as you preferred and squashed everything into a single signed commit.

boomballa:
> Thanks @boomballa . The PR looks great and is approved. I am just trying to understand your **production** methodology. I understand what jump host (or bastion is). I would still expect that there will be a script on the bare server that is run either systemd or some other background task management system and the eventual process won't be dependent on the connecting shell. It's not the case here? Are you guys using some sort of ansible approach where the orchestrator launches processes directly via ssh?

@romange  Thanks for the approval! You are spot on about production best practices for cloud/container environments.
To clarify our specific methodology:

In our setup, we maximize bare-metal resource utilization by deploying multiple database instances per physical server. While we do use Systemd/Docker for designated single-purpose resources (like cloud VMs), for these massive bare-metal clusters, the workflow is different.

I am a DBA (not using centralized SRE tools like Ansible/Puppet here), and I frequently need to SSH into these servers via a Jump Host using my personal credentials to manually launch, tune, or maintain specific instances. 
Using Systemd for every dynamic, multi-instance deployment on a shared physical server can be cumbersome and less flexible for ad-hoc DBA operations.

In this context, Jump Host session timeouts are an unavoidable reality. Having the process survive these disconnects (like Redis does) acts as a critical safety net for us DBAs during manual operations, and also significantly improves Dragonfly’s overall stability and robustness against environmental disruptions.

romange:
image


You still do not have a signed commit.you should use something like this:
```
git config commit.gpgsign true
git config user.email <yourmail>
git config user.signingkey <yoursignkey>   
```

see https://codex.so/gpg-verification-github for more details.

boomballa:
Hi @romange .
Sorry for the extra noise regarding the GPG signature! I've set up my GPG key following your instructions and force-pushed. 
The commit is now properly "Verified".

Thank you so much for your patience and for guiding me through the contribution process—I learned a lot today!

PR #6464, opened 2026-01-24
chore: reduce includes

dranikpg:
Weekend cleanup 🧹 

1. Make server/journal its own target
2. Reduce facade/facade_types to move stats structs into facade/facade_stats.h
3. Reduce includes in common.h and move GenRandomHex to separate header/cc
4. Fix long chain of absl/other includes that is needed with smaller common.h

### DIFF
--- a/src/core/detail/gen_utils.h
+++ b/src/core/detail/gen_utils.h
@@ -9,7 +9,7 @@
 
 #include <string>
 
-namespace dfly::detail {
+namespace dfly {
 
 inline std::string GetRandomHex(absl::InsecureBitGen& gen, size_t len, size_t len_deviation = 0) {
   static_assert(std::is_same<uint64_t, decltype(gen())>::value);
@@ -37,4 +37,4 @@ inline std::string GetRandomHex(absl::InsecureBitGen& gen, size_t len, size_t le
   return res;
 }
 
-}  // namespace dfly::detail
+}  // namespace dfly
--- a/src/server/cluster/cluster_family_test.cc
+++ b/src/server/cluster/cluster_family_test.cc
@@ -438,7 +438,7 @@ TEST_F(ClusterFamilyTest, ClusterConfigFullMultipleInstances) {
 
   absl::InsecureBitGen eng;
   while (true) {
-    string random_key = detail::GetRandomHex(eng, 40);
+    string random_key = GetRandomHex(eng, 40);
     SlotId slot = KeySlot(random_key);
     if (slot > 10'000) {
       continue;
@@ -450,7 +450,7 @@ TEST_F(ClusterFamilyTest, ClusterConfigFullMultipleInstances) {
   }
 
   while (true) {
-    string random_key = detail::GetRandomHex(eng, 40);
+    string random_key = GetRandomHex(eng, 40);
     SlotId slot = KeySlot(random_key);
     if (slot <= 10'000) {
       continue;
--- a/src/server/debugcmd.cc
+++ b/src/server/debugcmd.cc
@@ -88,7 +88,7 @@ struct ValueCompressInfo {
 
 std::string GenerateValue(size_t val_size, bool random_value, absl::InsecureBitGen* gen) {
   if (random_value) {
-    return detail::GetRandomHex(*gen, val_size);
+    return GetRandomHex(*gen, val_size);
   } else {
     return string(val_size, 'x');
   }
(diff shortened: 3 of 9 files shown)


dranikpg:
augment review

src/facade/facade_stats.h
```diff
@@ -0,0 +1,104 @@
...
+#pragma once
+
+#include <absl/container/flat_hash_map.h>
+
+#include <atomic>
+#include <cstdint>
+
+namespace facade {
+
+struct ConnectionStats {
```
dranikpg, line 14: all of this is moved from `facade_types.h`

src/server/journal/CMakeLists.txt
```diff
@@ -1,5 +1,8 @@
-SET(DF_JOURNAL_SRCS
-    journal/cmd_serializer.cc journal/tx_executor.cc namespaces.cc
-    journal/journal.cc journal/types.cc journal/journal_slice.cc
-    journal/serializer.cc journal/executor.cc journal/streamer.cc
-    PARENT_SCOPE)
+add_library(dfly_journal
+    cmd_serializer.cc tx_executor.cc
+    journal.cc types.cc journal_slice.cc
+    serializer.cc executor.cc streamer.cc)
+target_link_libraries(dfly_journal dfly_transaction dragonfly_lib dfly_facade base io)
```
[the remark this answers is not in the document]
dranikpg: Sadly everyhing depends on everything

src/facade/facade_stats.h
```diff
@@ -0,0 +1,104 @@
...
+  uint32_t num_blocked_clients = 0;
+
+  // number of times the connection yielded due to max_busy_read_usec limit
+  uint32_t num_read_yields = 0;
+  uint64_t num_migrations = 0;
+  uint64_t num_recv_provided_calls = 0;
+
+  // Number of times the tls connection was closed by the time we started reading from it.
+  uint64_t tls_accept_disconnects = 0;  // number of TLS socket disconnects during the handshake
+                                        //
```
kostasrim, line 52:
```suggestion
  // number of TLS socket disconnects during the handshake
  uint64_t tls_accept_disconnects = 0;
```

src/facade/facade_stats.h
```diff
@@ -0,0 +1,104 @@
...
+
+  ConnectionStats& operator+=(const ConnectionStats& o);
+};
+
+struct ReplyStats {
+  struct SendStats {
+    int64_t count = 0;
+    int64_t total_duration = 0;
+
+    SendStats& operator+=(const SendStats& other) {
```
kostasrim, line 72: Not applicable here but n general, we also have the spaceship operator now :)

src/facade/resp_validator.cc
```diff
@@ -6,6 +6,7 @@
 
 #include <cstdint>
 #include <fstream>
+#include <iostream>
```
kostasrim, line 9: 🤔

src/server/journal/CMakeLists.txt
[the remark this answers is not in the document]
kostasrim: IMHO we should take care of that... I opened an issue: https://github.com/dragonflydb/dragonfly/issues/6479

BorysTheDev:
Honestly, I hate common.h it's just a trash bin

src/server/journal/CMakeLists.txt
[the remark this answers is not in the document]
kostasrim: Also my `concern` here is that we `add a new circular dependency`:

before: journal sources are part of `dfly_transaction` that links with `dfly_lib` -> so no circular deps
now: separate lib (dflyjournal) that links both with `dfly_transaction` but and vice versa -> new circular dep

This change is suspicious unless I am missing something with the linking order.

src/facade/resp_validator.cc
dranikpg: It uses cout, but cout doesn't include it. It failed to build after I cleaned up common.h

src/core/detail/gen_utils.h
```diff
@@ -0,0 +1,40 @@
...
+//
+
+#pragma once
+
+#include <absl/random/random.h>
+#include <absl/strings/str_cat.h>
+
+#include <string>
+
+namespace dfly::detail {
```
BorysTheDev: it is strange to show detail namespace outside, maybe we don't need it? In most cases this namespace is considered for internal usage

dranikpg: Fixed

dranikpg:
* Moved journal back to source files
* Removed detail namespace from GetRandomHex

kostasrim (approved):

PR #6465, opened 2026-01-24
chore(tiering): polish type system for better clarity and safety

romange:
1. Rename EntryId to PendingId and move OwnedEntryId to private scope.
2. Pass DiskSegment directly to the most generic Read method, removing PrimeValue dependency.
3. Update NotifyStashed/NotifyFetched signatures to use const OwnedEntryId& as EntryId did not save any allocations.
4. The only functional change: changing the signature of one of the Read functions to pass string_view instead of string  which saves a copy and allocation.

### DIFF
--- a/src/server/string_family.cc
+++ b/src/server/string_family.cc
@@ -131,7 +131,9 @@ OpResult<TResultOrT<size_t>> OpStrLen(const OpArgs& op_args, string_view key) {
   // TODO(vlad): Omit decoding string to just query it's length
   if (const auto& co = it_res.value()->second; co.IsExternal()) {
     TieredStorage::TResult<size_t> fut;
-    auto cb = [fut](io::Result<string> s) mutable { fut.Resolve(s.transform(&string::size)); };
+    auto cb = [fut](io::Result<string_view> s) mutable {
+      fut.Resolve(s.transform(&string_view::size));
+    };
     op_args.shard->tiered_storage()->Read(op_args.db_cntx.db_index, key, co, std::move(cb));
     return {std::move(fut)};
   } else {
@@ -172,7 +174,7 @@ OpResult<TResultOrT<size_t>> OpSetRange(const OpArgs& op_args, string_view key, 
 
 OpResult<StringResult> OpGetRange(const OpArgs& op_args, string_view key, int32_t start,
                                   int32_t end) {
-  auto read = [start, end](std::string_view slice) mutable -> string_view {
+  auto read_cb = [start, end](std::string_view slice) mutable -> string {
     int32_t strlen = slice.size();
     if (strlen == 0)
       return "";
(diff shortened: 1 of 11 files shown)


src/server/tiering/common.h
```diff
@@ -41,6 +35,10 @@ struct DiskSegment {
     return offset == other.offset && length == other.length;
   }
 
+  DiskSegment ContainingPages() const {
```
romange, line 38: just moved

src/server/string_family.cc
```diff
@@ -172,7 +174,7 @@ OpResult<TResultOrT<size_t>> OpSetRange(const OpArgs& op_args, string_view key,
 
 OpResult<StringResult> OpGetRange(const OpArgs& op_args, string_view key, int32_t start,
                                   int32_t end) {
-  auto read = [start, end](std::string_view slice) mutable -> string_view {
+  auto read_cb = [start, end](std::string_view slice) mutable -> string {
```
romange, line 177: we always wrapped `read_cb` within string, so lets just return it.

src/server/string_family.cc
```diff
@@ -210,12 +212,14 @@ OpResult<StringResult> OpGetRange(const OpArgs& op_args, string_view key, int32_
     fb2::Future<io::Result<std::string>> fut;
     op_args.shard->tiered_storage()->Read(
         op_args.db_cntx.db_index, key, co,
-        [read, fut](const io::Result<std::string>& s) mutable { fut.Resolve(string{read(*s)}); });
+        [read_cb, fut](const io::Result<std::string_view>& s) mutable {
```
romange, line 215: when this callback runs in the shard thread, the backing string storage still exists.

src/server/tiered_storage.cc
```diff
@@ -396,28 +399,30 @@ void TieredStorage::Close() {
...
+  Read(dbid, key, value, std::move(cb));
   return fut;
 }
 
 void TieredStorage::Read(DbIndex dbid, std::string_view key, const PrimeValue& value,
-                         std::function<void(io::Result<std::string>)> readf) {
+                         std::function<void(io::Result<std::string_view>)> readf) {
   auto cb = [readf = std::move(readf)](io::Result<tiering::StringDecoder*> res) mutable {
-    readf(res.transform([](auto* d) { return string{d->Read()}; }));
+    readf(res.transform([](tiering::StringDecoder* d) { return d->GetView(); }));
```
romange, line 423: this is the only change - @dranikpg  PTAL. 
It seemed to me that creating of string here is unneeder as it is never moved and only copied from.
so let's just pass the view

dranikpg: I agree. We should actually move those functons to string_family, so only the generic decoder read interface remains for tiered storage. With multiple types, it's all it should be concerned about

PR #6469, opened 2026-01-25
chore(tiering): refactor Read/Modify to free functions

romange:
Move TieredStorage::Read/Modify methods to standalone ReadTiered/ModifyTiered functions that take explicit TieredStorage* parameter for cleaner API design.

Add additional ReadTiered and ReadTieredString overloads that simplify tiered reads in several places across the codebase.

### DIFF
--- a/src/server/generic_family.cc
+++ b/src/server/generic_family.cc
@@ -320,7 +320,8 @@ OpResult<string> DumpToString(string_view key, const PrimeValue& pv, const OpArg
   io::StringSink sink;
   if (pv.IsExternal() && !pv.IsCool()) {
     // TODO: consider moving blocking point to coordinator to avoid stalling shard queue
-    auto res = op_args.shard->tiered_storage()->Read(op_args.db_cntx.db_index, key, pv).Get();
+    auto res =
+        ReadTieredString(op_args.db_cntx.db_index, key, pv, op_args.shard->tiered_storage()).Get();
     if (!res.has_value())
       return OpStatus::IO_ERROR;
 
@@ -1138,7 +1139,7 @@ void GenericFamily::Delex(CmdArgList args, CommandContext* cmd_cntx) {
   string_view key = ArgS(args, 0);
 
   // Parse optional condition
-  enum class Condition { NONE, IFEQ, IFNE, IFDEQ, IFDNE };
+  enum class Condition : uint8_t { NONE, IFEQ, IFNE, IFDEQ, IFDNE };
   Condition cond = Condition::NONE;
   string_view compare_value;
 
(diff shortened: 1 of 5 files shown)


src/server/tiered_storage.cc
```diff
@@ -742,4 +701,34 @@ detail::TieredColdRecord* TieredStorage::PopCool() {
...
+                                       std::function<T(std::string*)> modf, TieredStorage* ts) {
+  DCHECK(value.IsExternal());
+  DCHECK_EQ(value.ObjType(), OBJ_STRING);
+
+  util::fb2::Future<io::Result<T>> future;
+
+  auto cb = [future, modf = std::move(modf)](io::Result<tiering::StringDecoder*> res) mutable {
+    future.Resolve(res.transform([&modf](auto* d) { return modf(d->Write()); }));
+  };
+  ts->Read(dbid, key, value.GetExternalSlice(), tiering::StringDecoder{value}, std::move(cb));
```
romange, line 746: `ModifyTiered` is simpler now, without double transformations.


Image

src/server/tiered_storage.cc
```diff
@@ -742,4 +701,34 @@ detail::TieredColdRecord* TieredStorage::PopCool() {
...
+TieredStorage::TResult<T> ModifyTiered(DbIndex dbid, std::string_view key, const PrimeValue& value,
+                                       std::function<T(std::string*)> modf, TieredStorage* ts) {
+  DCHECK(value.IsExternal());
+  DCHECK_EQ(value.ObjType(), OBJ_STRING);
+
+  util::fb2::Future<io::Result<T>> future;
+
+  auto cb = [future, modf = std::move(modf)](io::Result<tiering::StringDecoder*> res) mutable {
+    future.Resolve(res.transform([&modf](auto* d) { return modf(d->Write()); }));
+  };
```
dranikpg, line 745: Actually you can keep it just inside string_family

src/server/tiered_storage.cc
```diff
@@ -742,4 +701,34 @@ detail::TieredColdRecord* TieredStorage::PopCool() {
   return &res;
 }
 
+void ReadTiered(DbIndex dbid, std::string_view key, const PrimeValue& value,
+                function<void(io::Result<string_view>)> readf, TieredStorage* ts) {
+  auto cb = [readf = std::move(readf)](io::Result<tiering::StringDecoder*> res) mutable {
+    readf(res.transform([](tiering::StringDecoder* d) { return d->GetView(); }));
+  };
+  ts->Read(dbid, key, value.GetExternalSlice(), tiering::StringDecoder{value}, std::move(cb));
+}
```
dranikpg, line 733: The only thing I don't like about this and previous structures is how deeply wrapped in possibly allocating std::functions our callback becomes

src/server/tiered_storage.cc
romange: i was not sure if `tiering::StringDecoder` is considered a private class or not.

src/server/tiered_storage.cc
romange: what about start using https://github.com/abseil/abseil-cpp/blob/master/absl/functional/any_invocable.h#L60 ?

romange: it's like std::move_only_function

dranikpg: that's not a problem, the only solution is to use fu2:: with a larger small-object size

dranikpg: if it shows up in benchmarks we can think avout optikizing it

src/server/generic_family.cc
```diff
@@ -1173,8 +1174,18 @@ void GenericFamily::Delex(CmdArgList args, CommandContext* cmd_cntx) {
...
 
+  auto compare_str = [&](string_view val) {
+    bool is_digest = (cond == Condition::IFDEQ || cond == Condition::IFDNE);
+
+    if (is_digest) {
+      string dig = XXH3_Digest(val);
+      return (dig == compare_value) == (cond == Condition::IFDEQ);
+    }
+    return (val == compare_value) == (cond == Condition::IFEQ);
+  };
```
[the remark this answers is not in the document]
dranikpg: Yep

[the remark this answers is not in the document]
romange: why it is a bug? we wait for both `cb` and disk reads to finish before we exit the function.

dranikpg:
Oh no, I just glitched for a second and merged the wrong PR 😭

src/server/generic_family.cc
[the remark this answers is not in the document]
dranikpg: True, for some reason I though this is inside a shard function behcase the top was collapsed

PR #6475, opened 2026-01-26
feat: global_id mapping serialization

BorysTheDev:
Part of HNSW index replication task: added save/load for hash/ json globalId mapping

### DIFF
--- a/src/server/search/doc_index.h
+++ b/src/server/search/doc_index.h
@@ -347,7 +347,7 @@ class ShardDocIndex {
   DocKeyIndex key_index_;
   Synonyms synonyms_;
 
-  // Mapping from key to global_id received from master during replication.
+  // Mapping from key to master global_id received from master during replication.
   // Used to restore vector index relationships after RDB load.
   absl::flat_hash_map<std::string, GlobalDocId> master_doc_ids_;
 };
@@ -380,11 +380,10 @@ class ShardDocIndices {
   void AddDoc(std::string_view key, const DbContext& db_cnt, const PrimeValue& pv);
   void RemoveDoc(std::string_view key, const DbContext& db_cnt, const PrimeValue& pv);
 
-  // Callback type for ForEachGlobalDocId: (index_name, global_id) -> void
+  // Callback type for ForEachGlobalDocId: (index_name, global_id)
   using GlobalIdCallback = std::function<void(std::string_view, search::GlobalDocId)>;
 
   // Invoke callback for each (index_name, global_id) pair for a key across all matching indices.
-  // Avoids memory allocation compared to returning a vector.
   void ForEachGlobalDocId(std::string_view key, const DbContext& db_cntx, const PrimeValue& pv,
                           GlobalIdCallback cb) const;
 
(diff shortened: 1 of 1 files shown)


src/server/rdb_load.cc
```diff
@@ -2719,6 +2733,15 @@ void RdbLoader::CreateObjectOnShard(const DbContext& db_cntx, const Item* item,
...
   }
 
+  // Store master doc_id mappings for search indices
+  if (!item->global_ids.empty()) {
+    if (auto* search_indices = db_slice->shard_owner()->search_indices(); search_indices) {
+      for (const auto& [index_name, global_id] : item->global_ids) {
+        search_indices->SetMasterDocId(index_name, item->key, global_id);
+      }
+    }
+  }
```
dranikpg, line 2743: What if the indices haven't been created yet. They're loaded from an aux field on shard 0 snapshot

src/server/rdb_save.h
```diff
@@ -249,6 +249,11 @@ class RdbSerializer : public SerializerBase {
   io::Result<uint8_t> SaveEntry(const PrimeKey& pk, const PrimeValue& pv, uint64_t expire_ms,
                                 uint32_t mc_flags, DbIndex dbid);
 
+  // Write a single global_id entry for search-indexed keys.
+  // Format: RDB_OPCODE_GLOBAL_ID + index_name (string) + global_id (8 bytes).
```
dranikpg, line 253: imagine how wasteful it's to send the index name each time

src/server/rdb_load.cc
```diff
@@ -2075,6 +2079,16 @@ error_code RdbLoader::Load(io::Source* src) {
...
     }
 
+    if (type == RDB_OPCODE_GLOBAL_ID) {
+      /* GLOBAL_ID: search index global document id (index_name + global_id) */
+      string index_name;
+      SET_OR_RETURN(FetchGenericString(), index_name);
+      uint64_t global_id;
+      SET_OR_RETURN(FetchInt<uint64_t>(), global_id);
+      settings.global_ids.emplace_back(std::move(index_name), global_id);
+      continue; /* Read next opcode. */
```
dranikpg, line 2089: I don't understand it from a consistency point of view. If we plan to support writes in the future we have to plan ahead. I'll message in a private group

src/server/rdb_save.h
BorysTheDev: this is why I wanted to have global_id for all indexes

src/server/rdb_load.cc
BorysTheDev: In the next PR I will create index automatically if we have such fields

src/server/rdb_load.cc
BorysTheDev: I plan to implement the same mechanism as we use for replication.

src/server/rdb_load.cc
dranikpg: Then it is better to add some proper utilities for loading indices rather than abusing the indices themself

BorysTheDev: didn't get your point

Issue #6396, opened 2026-01-15
DFLYCLUSTER GETSLOTINFO Ranges


tnxbutno:
**Did you search GitHub Issues and GitHub Discussions First?**
Yes

**Is your feature request related to a problem? Please describe.**
Currently, the DFLYCLUSTER GETSLOTINFO command accepts either a single slot or an explicit list of slots:
`DFLYCLUSTER GETSLOTINFO [SLOTS slot [slot ...]]`
When working with large clusters or analyzing slot distribution, it is often necessary to inspect statistics for a continuous range of slots (for example, slots 1–100 or 5000–8000).
At the moment, this requires manually expanding the range into a long list of slot IDs, which is inconvenient and inefficient for both users and tooling.

**Describe the solution you'd like**
Extend the DFLYCLUSTER GETSLOTINFO command to support specifying slot ranges, in addition to individual slots or lists.
Examples (syntax is flexible and only illustrative):
```
DFLYCLUSTER GETSLOTINFO SLOTS 1-100
DFLYCLUSTER GETSLOTINFO SLOTS 1-100 200-300
DFLYCLUSTER GETSLOTINFO SLOTS 1-100 250 400-450
```
The command would return the same per-slot statistics as today (slot id, key count, memory usage, read requests, write requests), but for all slots within the specified ranges.

This would significantly improve usability.

**Describe alternatives you've considered**
Manually expanding slot ranges into a full list of slot IDs.

PR #6485, opened 2026-01-27
feat(cluster): add range syntax support to DFLYCLUSTER GETSLOTINFO

vyavdoshenko:
Add support for range notation in DFLYCLUSTER GETSLOTINFO command.

Before: DFLYCLUSTER GETSLOTINFO SLOTS 0 1 2 3 4 5
After: DFLYCLUSTER GETSLOTINFO SLOTS 0-5

Supports mixed syntax (0-10 50 100-150) and reversed ranges (5-2 → 2-5).

Fixes: #6396

### DIFF
--- a/src/server/cluster/cluster_family.cc
+++ b/src/server/cluster/cluster_family.cc
@@ -639,15 +639,51 @@ void ClusterFamily::DflyClusterGetSlotInfo(CmdArgList args, CommandContext* cmd_
   parser.ExpectTag("SLOTS");
   auto* rb = static_cast<RedisReplyBuilder*>(cmd_cntx->rb());
 
-  vector<std::pair<SlotId, SlotStats>> slots_stats;
-  do {
-    auto sid = parser.Next<uint32_t>();
-    if (sid > kMaxSlotNum)
-      return cmd_cntx->SendError("Invalid slot id");
-    slots_stats.emplace_back(sid, SlotStats{});
-  } while (parser.HasNext());
-
   RETURN_ON_PARSE_ERROR(parser, cmd_cntx);
+
+  vector<std::pair<SlotId, SlotStats>> slots_stats;
+  while (parser.HasNext()) {
+    auto arg = parser.Next<std::string_view>();
+    // Check if argument contains a dash for range notation (e.g., "1-100")
+    size_t dash_pos = arg.find('-');
+    if (dash_pos != std::string_view::npos && dash_pos > 0) {
+      // Parse as range: start-end
+      std::string_view start_str = arg.substr(0, dash_pos);
+      std::string_view end_str = arg.substr(dash_pos + 1);
+
+      uint32_t start_slot, end_slot;
+      if (!absl::SimpleAtoi(start_str, &start_slot) || !absl::SimpleAtoi(end_str, &end_slot)) {
+        return cmd_cntx->SendError("Invalid slot range format");
+      }
+
+      if (start_slot > kMaxSlotNum || end_slot > kMaxSlotNum) {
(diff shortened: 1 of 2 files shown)


src/server/cluster/cluster_family.cc
```diff
@@ -639,15 +639,51 @@ void ClusterFamily::DflyClusterGetSlotInfo(CmdArgList args, CommandContext* cmd_
...
+      if (!absl::SimpleAtoi(start_str, &start_slot) || !absl::SimpleAtoi(end_str, &end_slot)) {
+        return cmd_cntx->SendError("Invalid slot range format");
+      }
 
-  RETURN_ON_PARSE_ERROR(parser, cmd_cntx);
+      if (start_slot > kMaxSlotNum || end_slot > kMaxSlotNum) {
+        return cmd_cntx->SendError("Invalid slot id");
+      }
+
+      // Swap if range is specified in reverse order (e.g., "100-0")
```
romange, line 663: i think it's redundant

src/server/cluster/cluster_family.cc
```diff
@@ -639,15 +639,51 @@ void ClusterFamily::DflyClusterGetSlotInfo(CmdArgList args, CommandContext* cmd_
...
-    auto sid = parser.Next<uint32_t>();
-    if (sid > kMaxSlotNum)
-      return cmd_cntx->SendError("Invalid slot id");
-    slots_stats.emplace_back(sid, SlotStats{});
-  } while (parser.HasNext());
+  while (parser.HasNext()) {
+    auto arg = parser.Next<std::string_view>();
+    // Check if argument contains a dash for range notation (e.g., "1-100")
+    size_t dash_pos = arg.find('-');
+    if (dash_pos != std::string_view::npos && dash_pos > 0) {
```
romange, line 649:
```suggestion
    if (dash_pos != std::string_view::npos) {
```

vyavdoshenko: I can't agree with you. The dash_pos > 0 check correctly treats -1 as a negative number (-> "value is not an integer") rather than a malformed range with empty start (-> "Invalid slot range format"). A dash at position 0 is a minus sign, not a range separator.

src/server/cluster/cluster_family.cc
vyavdoshenko: Why? I have to return an error in the opposite way. This is more reliable.

BorysTheDev:
@romange  I would prefer a more REDIS natural approach like `DFLYCLUSTER GETSLOTINFO SLOTS [RANGE start end] [single_slot]`

or

`DFLYCLUSTER GETSLOTINFO SLOT_RANGES [start end] ...` where start end can be the same slot if we need to provide 1 slot and then we can drop the old comand

romange:
why does it matter? it's Dragonfly management command

BorysTheDev:
> why does it matter? it's Dragonfly management command

Yes. but consistency is important for understanding the code. And our parser can parse it with less code. It will be 2 lines of code instead of 20

romange:
The proposed syntax is more user-friendly, and in my view, that advantage outweighs LOC arguments. We design APIs for humans; if machine efficiency were the priority, we’d be using RPC.

src/server/cluster/cluster_family.cc
romange: well, I am adept of "strict" apis. why would someone reverse the range? when using in python `[10, 1]` do we expect python do some "automatic" stuff and reverse ranges? auto conversions are confusing, imho.
(no need to fix this now).

Issue #6409, opened 2026-01-19
Test replication_test.py::test_replication_all fails (on epoll, list pack gets corrupted in replication)


glevkovich:
list pack gets corrupted in replication.

- build (ubuntu-dev:20-gcc14, Epoll, Debug, ubuntu-latest)
- job: https://github.com/dragonflydb/dragonfly/actions/runs/21124356087/job/60742577555
- call stack:
[call_stack.txt](https://github.com/user-attachments/files/24708257/call_stack.txt)
- Logs:  [logs.zip](https://github.com/user-attachments/files/24708271/logs.zip)

PR #6488, opened 2026-01-28
chore: add more invariant checks for streams (#6459)

romange:
1. Check rax validity in more places in stream_family.cc
2. Cover streamTrim function and check lp validity  following the assert-fail in the tests.
3. Move raxStop to bottop of StreamAppendItem function though it's not clear that raxStop can
   actually cause the bug as ri.key is always 16 bytes and is never allocated on heap.

Addresses https://github.com/dragonflydb/dragonfly/issues/6409 though not necessary fixes it.

### DIFF
--- a/src/server/stream_family.cc
+++ b/src/server/stream_family.cc
@@ -353,9 +353,9 @@ void StreamNextID(uint64_t now_ms, const streamID* last_id, streamID* new_id) {
 
 /* Convert the specified stream entry ID as a 128 bit big endian number, so
  * that the IDs can be sorted lexicographically. */
-inline void StreamEncodeID(uint8_t* buf, streamID* id) {
-  absl::big_endian::Store64(buf, id->ms);
-  absl::big_endian::Store64(buf + 8, id->seq);
+inline void StreamEncodeID(uint8_t* buf, const streamID& id) {
+  absl::big_endian::Store64(buf, id.ms);
+  absl::big_endian::Store64(buf + 8, id.seq);
 }
 
 /* Adds a new item into the stream 's' having the specified number of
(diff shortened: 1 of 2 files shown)


romange:
augment review

src/server/stream_family.cc
```diff
@@ -353,9 +353,9 @@ void StreamNextID(uint64_t now_ms, const streamID* last_id, streamID* new_id) {
 
 /* Convert the specified stream entry ID as a 128 bit big endian number, so
  * that the IDs can be sorted lexicographically. */
-inline void StreamEncodeID(uint8_t* buf, streamID* id) {
-  absl::big_endian::Store64(buf, id->ms);
-  absl::big_endian::Store64(buf + 8, id->seq);
+inline void StreamEncodeID(uint8_t* buf, const streamID& id) {
```
romange, line 356: just hardened the interface

src/server/stream_family.cc
```diff
@@ -431,19 +431,20 @@ int StreamAppendItem(stream* s, CmdArgList fields, uint64_t now_ms, streamID* ad
   size_t lp_bytes = 0;      /* Total bytes in the tail listpack. */
   unsigned char* lp = NULL; /* Tail listpack pointer. */
 
+  /* We have to add the key into the radix tree in lexicographic order,
```
romange, line 434: moved from below so I could copy to `rax_key` at line 446

src/server/stream_family.cc
```diff
@@ -431,19 +431,20 @@ int StreamAppendItem(stream* s, CmdArgList fields, uint64_t now_ms, streamID* ad
...
+
   if (!raxEOF(&ri)) {
     /* Get a reference to the tail node listpack. */
     lp = (uint8_t*)ri.data;
     lp_bytes = lpBytes(lp);
     CHECK_GT(lp_bytes, 0U);
+    DCHECK(ri.key_len == sizeof(rax_key));
+    memcpy(rax_key, ri.key, sizeof(rax_key));
   }
-  raxStop(&ri);
```
romange, line 440: moved to the bottom of this function.

src/server/stream_family.cc
```diff
@@ -615,14 +620,19 @@ int StreamAppendItem(stream* s, CmdArgList fields, uint64_t now_ms, streamID* ad
   lp = lpAppendInteger(lp, lp_count);
 
   /* Insert back into the tree in order to update the listpack pointer. */
-  if (ri.data != lp) {
```
romange, line 618: it's not the correctness fix, it's a performance fix. in cases when rax is empty, we would call raxInsert here second time as `ri.data` would be NULL. Now we actually check if `lp` was reallocated due to insertions.

kostasrim (approved):

PR #6497, opened 2026-01-29
feat(json): Adds a reference counted string class

abhijat:
* Wraps the blob introduced in https://github.com/dragonflydb/dragonfly/pull/6456
* Handles reference count and lifetimes
* Exposes a string api for jsoncons

The code is not hooked into jsoncons yet, it will be done in a small, final PR, kept separate because a lot of tests have to be adjusted at that point.

### DIFF
--- a/src/core/json/detail/interned_string.cc
+++ b/src/core/json/detail/interned_string.cc
@@ -5,21 +5,8 @@
 
 namespace dfly::detail {
 
-InternedString& InternedString::operator=(const InternedString& other) {
-  if (this != &other) {
-    Release();
-    entry_ = other.entry_;
-    Acquire();
-  }
-  return *this;
-}
-
-InternedString& InternedString::operator=(InternedString&& other) noexcept {
-  if (this != &other) {
-    Release();
-    entry_ = other.entry_;
-    other.entry_ = {};
-  }
+InternedString& InternedString::operator=(InternedString other) {
+  swap(other);
   return *this;
 }
 
--- a/src/core/json/detail/interned_string.h
+++ b/src/core/json/detail/interned_string.h
@@ -41,8 +41,7 @@ class InternedString {
     other.entry_ = {};
   }
 
-  InternedString& operator=(const InternedString& other);
-  InternedString& operator=(InternedString&& other) noexcept;
+  InternedString& operator=(InternedString other);
 
   ~InternedString() {
     Release();


src/core/json/detail/interned_string.cc
```diff
@@ -0,0 +1,91 @@
...
+int InternedString::compare(const InternedString& other) const {
+  return std::string_view{*this}.compare(std::string_view{other});
+}
+
+int InternedString::compare(std::string_view other) const {
+  return std::string_view{*this}.compare(other);
+}
+
+void InternedString::shrink_to_fit() {  // NOLINT (must be non-const to align with jsoncons usage)
+  DCHECK(entry_);
```
abhijat: This could just as well be an empty method. Initially added this `DCHECK` here to make sure jsoncons does not end up shrinking an empty string through some code path, but that hasn't happened in any tests.

abhijat: removed it, it doesnt seem to add any value

src/core/json/detail/interned_blob.cc
```diff
@@ -15,6 +15,10 @@ constexpr size_t kHeaderSize = sizeof(uint32_t) * 2;
 namespace dfly::detail {
 
 InternedBlobHandle InternedBlobHandle::Create(std::string_view sv) {
+  if (sv.empty()) {
+    return InternedBlobHandle{nullptr};
```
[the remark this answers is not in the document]
abhijat: the comment ended up being misleading. `data()/c_str()` in  `InternedString` do the right thing though, will fix the comment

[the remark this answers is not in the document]
abhijat: fixed

src/core/json/detail/interned_string.h
```diff
@@ -0,0 +1,127 @@
...
+
+  [[nodiscard]] int compare(const InternedString& other) const;
+  [[nodiscard]] int compare(std::string_view) const;
+
+  // lex. comparison
+  auto operator<=>(const InternedString& other) const {
+    return std::string_view{*this} <=> std::string_view{other};
+  }
+
+  bool operator==(const InternedString& other) const = default;
```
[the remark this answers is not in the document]
abhijat: comparisons would only be for the same thread's blob pool when comparing interned strings.

src/core/json/detail/interned_blob.cc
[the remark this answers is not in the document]
abhijat: this is handled in the string class, it always returns a null terminated string

src/core/json/detail/interned_blob.h
```diff
@@ -23,14 +23,16 @@ using BlobPtr = char*;
...
 class InternedBlobHandle {
  public:
-  static InternedBlobHandle Create(std::string_view sv);
+  InternedBlobHandle() = default;
 
-  uint32_t Size() const;
+  [[nodiscard]] static InternedBlobHandle Create(std::string_view sv);
 
-  uint32_t RefCount() const;
+  [[nodiscard]] uint32_t Size() const;
```
romange: I do not see value in adding `[[nodiscard]] ` to all the accessors.  the only function that matters imho is `Create` as it can potentially cause a leak

src/core/json/detail/interned_string.cc
```diff
@@ -0,0 +1,86 @@
...
+  if (this != &other) {
+    Release();
+    entry_ = other.entry_;
+    other.entry_ = {};
+  }
+  return *this;
+}
+
+int InternedString::compare(const InternedString& other) const {
+  return std::string_view{*this}.compare(other);
```
romange: move one-liners  to header files

abhijat: moved

src/core/json/detail/interned_blob.h
abhijat: removed

src/core/json/detail/interned_string.cc
```diff
@@ -0,0 +1,78 @@
+// Copyright 2025, DragonflyDB authors.  All rights reserved.
```
romange: btw, please change the date to 2026 in all the touched files.

src/core/json/detail/interned_string.cc
```diff
@@ -0,0 +1,78 @@
...
+}
+
+InternedString& InternedString::operator=(InternedString&& other) noexcept {
+  if (this != &other) {
+    Release();
+    entry_ = other.entry_;
+    other.entry_ = {};
+  }
+  return *this;
+}
```
dranikpg: could've been copy&swap as you already have swap defined

abhijat:
```
Merging is blocked
Commits must have verified signatures.
```

I missed signature in a commit. Will have to rebase

src/core/json/detail/interned_string.cc
abhijat: fixed

src/core/json/detail/interned_string.cc
abhijat: Changed to copy-and-swap. I originally added swap because jsoncons didn't compile without it, but it seems to have an added bonus here.

src/core/json/detail/interned_blob.cc
```diff
@@ -78,8 +79,7 @@ void InternedBlobHandle::Destroy(InternedBlobHandle& handle) {
 }
 
 InternedBlobHandle::operator std::string_view() const {
-  DCHECK(blob_) << "Attempt to convert empty blob to string_view";
-  return {blob_, Size()};
+  return blob_ ? std::string_view{blob_, Size()} : "";
```
romange, line 82: i could be in header file.

romange (approved):

src/core/json/detail/interned_blob.cc
abhijat: addressed in https://github.com/dragonflydb/dragonfly/pull/6511

PR #6500, opened 2026-01-29
feat(hnswlib): Use __builtin_prefetch

mkaruza:
Use __builtin_prefetch instead of _mm_prefetch so we can benefit from prefetching memory on ARM also.

### DIFF
--- a/src/core/search/hnsw_alg.h
+++ b/src/core/search/hnsw_alg.h
@@ -302,12 +302,9 @@ template <typename dist_t> class HierarchicalNSW : public hnswlib::AlgorithmInte
         tableint candidate_id = *(datal + j);
         //                    if (candidate_id == 0) continue;
 
-        // Request prefetching of max 32 next vectors memory locations
-        if ((j % 32) == 0) {
-          size_t builtin_prefetch_limit = std::min(size - j, (size_t)32);
-          for (size_t k = 0; k + 1 < builtin_prefetch_limit; k++) {
-            __builtin_prefetch(getDataByInternalId(*(datal + j + k + 1)), 0, 3);
-          }
+        // Request prefetching next vector data memory
+        if (j + 1 < size) {
+          __builtin_prefetch(getDataByInternalId(*(datal + j + 1)), 0, 3);
         }
 
         if (visited_array[candidate_id] == visited_array_tag)
@@ -411,12 +408,9 @@ template <typename dist_t> class HierarchicalNSW : public hnswlib::AlgorithmInte
         int candidate_id = *(data + j);
         //                    if (candidate_id == 0) continue;
 
-        // Request prefetching of max 32 next vectors
-        if ((j % 32) == 0) {
-          size_t builtin_prefetch_limit = std::min(size - j, (size_t)32);
-          for (size_t k = 0; k + 1 < builtin_prefetch_limit; k++) {
-            __builtin_prefetch(getDataByInternalId(*(data + j + k + 1)), 0, 3);
-          }
+        // Request prefetching next vector data memory
+        if (j + 1 < size) {
+          __builtin_prefetch(getDataByInternalId(*(data + j + 1)), 0, 3);
         }
 
         if (!(visited_array[candidate_id] == visited_array_tag)) {


src/core/search/hnsw_alg.h
```diff
@@ -292,22 +292,23 @@ template <typename dist_t> class HierarchicalNSW : public hnswlib::AlgorithmInte
...
         tableint candidate_id = *(datal + j);
-//                    if (candidate_id == 0) continue;
-#ifdef USE_SSE
-        if (j + 1 < size) {
-          _mm_prefetch((char*)(visited_array + *(datal + j + 1)), _MM_HINT_T0);
-          _mm_prefetch(getDataByInternalId(*(datal + j + 1)), _MM_HINT_T0);
+        //                    if (candidate_id == 0) continue;
+
+        // Request prefetching of max 32 next vectors
+        if (!(j % 32)) {
```
romange:
```suggestion
        if (j % 32 == 0) {
```

src/core/search/hnsw_alg.h
```diff
@@ -292,22 +292,23 @@ template <typename dist_t> class HierarchicalNSW : public hnswlib::AlgorithmInte
...
+          size_t builtin_prefetch_limit = std::min(size - j, (size_t)32);
+          for (size_t k = 0; k + 1 < builtin_prefetch_limit; k++) {
+            __builtin_prefetch(getDataByInternalId(*(datal + j + k + 1)), 0, 3);
```
romange: say `char* vecPtr = ...` points to 768 long vector of size 768x4bytes.
this vector belongs to candidate  `*(datal + j + k)`.

1. is `getDataByInternalId(*(datal + j + k + 1))` a pointer to that vector, i.e. `vecPtr`  ? or  it is an address of vecPtr - `&vecPtr` ?
2. In case we prefetch the vector data (i.e. vecPtr), __builtin_prefetch loads only a single cache line imho. and not 3kb.

in any case please comment in more detail what do we prefetch here

src/core/search/hnsw_alg.h
```diff
@@ -401,25 +401,23 @@ template <typename dist_t> class HierarchicalNSW : public hnswlib::AlgorithmInte
...
       }
 
-#ifdef USE_SSE
-      _mm_prefetch((char*)(visited_array + *(data + 1)), _MM_HINT_T0);
-      _mm_prefetch((char*)(visited_array + *(data + 1) + 64), _MM_HINT_T0);
-      _mm_prefetch(data_level0_memory_ + (*(data + 1)) * size_data_per_element_ + offsetData_,
-                   _MM_HINT_T0);
-      _mm_prefetch((char*)(data + 2), _MM_HINT_T0);
-#endif
+      __builtin_prefetch((char*)(visited_array + *(data + 1)), 0, 3);
```
romange, line 402: same here - can you please explain what detail do we prefetch here?

src/core/search/hnsw_alg.h
```diff
@@ -401,25 +401,23 @@ template <typename dist_t> class HierarchicalNSW : public hnswlib::AlgorithmInte
...
-#ifdef USE_SSE
-        if (j + 1 < size) {
-          _mm_prefetch((char*)(visited_array + *(data + j + 1)), _MM_HINT_T0);
-          _mm_prefetch(
-              data_level0_memory_ + (*(data + j + 1)) * size_data_per_element_ + offsetData_,
-              _MM_HINT_T0);  ////////////
+        //                    if (candidate_id == 0) continue;
+
+        // Request prefetching of max 32 next vectors
+        if (!(j % 32)) {
```
romange:
```suggestion
        if (j % 32 == 0) {
```

romange (commented):
LGTM

src/core/search/hnsw_alg.h
mkaruza: It will point to memory address of data. Idea is to prefetch explicitly data to highest cache and expected that larger memory area is loaded in some upper layers for faster retrieval.

src/core/search/hnsw_alg.h
mkaruza: This is something that is prefetch already in hnswlib, but this is list of already visited nodes so we don't search them again.

romange (dismissed):

PR #6505, opened 2026-01-29
feat: send index definition for every thread

BorysTheDev:
Send index definitions in every replica thread

### DIFF
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -3012,7 +3012,7 @@ void RdbLoader::LoadSearchSynonymsFromAux(string&& def) {
   pending_synonym_cmds_.push_back(std::move(def));
 }
 
-void RdbLoader::PerformPostLoad(Service* service) {
+void RdbLoader::PerformPostLoad(Service* service, bool is_error) {
   const CommandId* cmd = service->FindCmd("FT.CREATE");
   if (cmd == nullptr)  // On MacOS we don't include search so FT.CREATE won't exist.
     return;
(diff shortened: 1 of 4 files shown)


dranikpg (commented):
The main reason why this is safe is because header serialization happens under a global lock

dranikpg:
and once you settle on this format it will be difficult to change

src/server/rdb_load.cc
```diff
@@ -2990,6 +3017,12 @@ void RdbLoader::PerformPostLoad(Service* service) {
   if (cmd == nullptr)  // On MacOS we don't include search so FT.CREATE won't exist.
     return;
 
+  // Clear the created indices tracking set for next load
+  {
+    std::lock_guard lk(search_index_mu_);
+    created_search_indices_.clear();
+  }
+
```
dranikpg, line 3025: if we return on an error it doesn't clear

src/server/rdb_save.cc
```diff
@@ -1409,10 +1409,89 @@ error_code RdbSaver::Impl::FlushSerializer() {
...
+
+}  // namespace
+
+RdbSaver::GlobalData RdbSaver::GetGlobalData(const Service* service, bool is_summary) {
   StringVec script_bodies, search_indices, search_synonyms;
+  size_t table_mem_result = 0;
 
+  if (!is_summary) {
+    shard_set->RunBriefInParallel([&](EngineShard* shard) {
+      if (shard->shard_id() == 0)
```
dranikpg, line 1487: `shard_set->pool()->at(0)->DispatchBrief(Func &&f)`

src/server/rdb_load.h
```diff
@@ -273,7 +275,7 @@ class RdbLoader : protected RdbLoaderBase {
 
   // Performs post load procedures while still remaining in global LOADING state.
   // Called once immediately after loading the snapshot / full sync succeeded from the coordinator.
-  static void PerformPostLoad(Service* service);
+  static void PerformPostLoad(Service* service, bool is_error = false);
```
dranikpg, line 278: Doesn't feel nice, but maybe that's the simplest way 😅

dranikpg (approved):

Issue #6344, opened 2026-01-08
EPIC: json memory optimization


romange:
1.  Our [jsoncons](https://github.com/dragonflydb/jsoncons) fork is 200 commits behind https://github.com/danielaparker/jsoncons/releases/tag/v1.5.0. Similarly to how our fork has https://github.com/dragonflydb/jsoncons/tree/Dragonfly.178  branch, lets create a branch on top v1.5.0 called Dragonfly.1.5.0 - apply the necessary changes (maybe not all of them are needed) and create a refreshed branch that can be used by Dragonfly
2. refcounted strings.  This is the main thing. The motivation - LLM/agentic session memory use-case.
   When using many json objects of similar schema, there are lots of strings that repeat ourselves. We should improve this efficiency. Attaching the script that reproduces this problem `scale_test.py --users 10000 --memories 50`

[scale_test.py](https://github.com/user-attachments/files/24487433/scale_test.py)

  
## refcounted strings

The inspiration for design is taken from https://github.com/Diggsey/ijson which also uses unique refcnt strings underneath.

1. jsoncons allows using a custom `key_type` by specifying it in the policy
https://github.com/danielaparker/jsoncons/blob/2d50086165e9ca2c0d5e0a069eec717cd2673aa2/doc/ref/corelib/basic_json.md?plain=1#L91
2. We can create an immutable string class that logically  has "size, refcnt, blob" members.
3. The class will represent an immutable string. and will be aliases to key_type.
4. Underneath its blobs will be managed by a thread-local singleton blob manager that will use a hashtable to 
return deduplicated strings. In one of my iterations I created a simplified version of such functionality:
https://github.com/romange/gaia/blob/master/strings/unique_strings.h
 but it's not exactly the same as it lack reference counting. But it's good as a reference 
5. In (2) I mentioned that this string has logically 3 distinct fields like `size, refcnt, blob` but in fact `size`, `refcnt` should reside in the same allocated blob (unlike std::string that has meta fields together with the pointer).
So physical representation of the blob should be `<refcnt, size, char data[]> `.  In a way this blob extends `sds` string as it already has `size` as part of its internal structure.  

At the end - each thread will manage its own pool of unique strings. The improvement can be replicated to HMAP as well as we mature this enhancement

romange:
[...]

PR #6511, opened 2026-01-30
feat(json): Use reference counted strings for json keys

abhijat:
follow up to https://github.com/dragonflydb/dragonfly/pull/6497, fnal PR that hooks in ref-counted strings to jsoncons for use as keys.

Adds interned string pool cleanup before destroying shard local memory resource in engine shard, to ensure no interned strings remain when the memory resource is destroyed.

Some tests are adjusted because the json memory usage pattern has changed.

---

Some basic comparisons using attached scale test script (attached, with random payloads):
[scale_test.py](https://github.com/user-attachments/files/25035477/scale_test.py)

### without ref-counted strings

```
python scale_test.py --users 10000 --memories 500
Starting scale test with:
  Users: 10000
  Memories per user: 500
  Workers: 4
  Redis: redis://localhost:6379
----------------------------------------
Generating user IDs...

[WRITE PHASE] Starting 4 workers...
[WRITE PHASE] Completed in 184.37 seconds
  Throughput: 54.24 users/sec
  Total Memories: 5000000
  Memory Throughput: 27119.93 memories/sec
```

without

memory used is 6.25 GiBs

### with ref-counted strings

```
python scale_test.py --users 10000 --memories 500
Starting scale test with:
  Users: 10000
  Memories per user: 500
  Workers: 4
  Redis: redis://localhost:6379
----------------------------------------
Generating user IDs...

[WRITE PHASE] Starting 4 workers...
[WRITE PHASE] Completed in 203.48 seconds
  Throughput: 49.14 users/sec
  Total Memories: 5000000
  Memory Throughput: 24572.33 memories/sec
```

with

memory used is 4.5 GiBs, so there is close to 28% reduction in memory use, but slightly slower than STL strings. To quantify the slowness

```
Time overhead: (203.48 - 184.37) / 184.37 = 10.4% slower
Throughput reduction: (27119.93 - 24572.33) / 27119.93 = 9.4% lower throughput
```

More comparisons are in https://github.com/dragonflydb/dragonfly/issues/6344 - the implementation was slightly different but the numbers look the same.

---

Performance with the following command:

```
debug populate 200000 k 100 rand type json elements 100
```

### Without ref-count strings

debug-populate-without

both mem used and rss are around 4 GiB

### With ref-count strings

debug-populate-with

This is interesting:

* memory used is lower, 3.5 GiB
* RSS gap is larger - probably the hash set overhead - around 300 MiB gap
* After flushall the gap does not diminish - its still a 300 MiB gap. The hash set does not release memory

I'm not sure how ref-counted strings saved memory with random json, the only explanation might be repeated keys over such a large space of random strings. We might need stats here.

We also need to be able release the hash set memory back to the OS. It's unexpected behavior if flushall doesn't release all memory immediately -- update below

---

With `flushall` shrinking the pool now, the memory usage goes down as expected:

image

### DIFF
--- a/src/server/engine_shard.cc
+++ b/src/server/engine_shard.cc
@@ -553,9 +553,9 @@ void EngineShard::DestroyThreadLocal() {
   shard_->Shutdown();
 
   detail::InternedString::ResetPool();
+  shard_->~EngineShard();
   CleanupStatelessAllocMR();
 
-  shard_->~EngineShard();
   mi_free(shard_);
   shard_ = nullptr;
   CompactObj::InitThreadLocal(nullptr);


src/server/engine_shard.cc
```diff
@@ -556,6 +556,9 @@ void EngineShard::DestroyThreadLocal() {
   mi_free(shard_);
   shard_ = nullptr;
   CompactObj::InitThreadLocal(nullptr);
+  detail::InternedString::ResetPool();
```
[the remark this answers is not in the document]
abhijat: fixed

src/core/json/detail/interned_blob.cc
```diff
@@ -67,7 +68,7 @@ void InternedBlobHandle::DecrRefCount() {  // NOLINT - non-const, mutates via pt
 }
 
 size_t InternedBlobHandle::MemUsed() const {
-  return blob_ ? Size() + kHeaderSize + 1 : 0;
+  return blob_ ? mi_good_size(Size() + kHeaderSize + 1) : 0;
```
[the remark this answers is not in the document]
abhijat: fixed

src/server/json_family_memory_test.cc
```diff
@@ -155,13 +158,15 @@ TEST_F(JsonFamilyMemoryTest, SimpleDel) {
...
   resp = Run({"JSON.GET", "j1"});
-  EXPECT_EQ(resp, json_after_del);
+  EXPECT_EQ(resp, R"({"b":" "})");
   resp = GetJsonMemoryUsageFromDb("j1");
 
   /* We still expect the initial size here, because after deletion we do not call shrink_to_fit on
      the JSON object. As a result, the memory will not be deallocated. Check
      JsonFamilyMemoryTest::JsonConsDelTest for example. */
-  EXPECT_THAT(resp, IntArg(start_size));
+  detail::InternedString dropped("a");
```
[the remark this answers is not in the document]
abhijat: fixed

tests/dragonfly/memory_test.py
```diff
@@ -1,10 +1,10 @@
...
-import asyncio
-from redis import asyncio as aioredis
 from .utility import *
 import logging
+import logging
+
 from . import dfly_args
-from .instance import DflyInstance, DflyInstanceFactory
+from .instance import DflyInstanceFactory
+from .utility import *
```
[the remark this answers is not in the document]
abhijat: this is just moved around, was already there before, if it has to be fixed it should be in a separate PR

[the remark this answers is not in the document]
abhijat: fixed

src/core/json/json_object.h
```diff
@@ -23,7 +24,12 @@ namespace dfly {
 class PageUsage;
 
 using TmpJson = jsoncons::json;
-using JsonType = jsoncons::basic_json<char, jsoncons::sorted_policy, StatelessAllocator<char>>;
+
+struct InternedStringPolicy : jsoncons::sorted_policy {
+  template <typename, typename, typename> using member_key = detail::InternedString;
+};
+
+using JsonType = jsoncons::basic_json<char, InternedStringPolicy, StatelessAllocator<char>>;
```
abhijat, line 32: @romange should there be a configuration setting for this, to allow disabling ref counted strings for json? Or will this be the default.

romange: default. I would like to see some numbers in terms of RSS memory usage - before this change and after with

1. the script a gave you - that shows the improvement.
2. some other script  or "DEBUG POPULATE ...  RAND type JSON ..." with unique keys where each key is unique

Please add both comparisons to the PR description

tests/dragonfly/memory_test.py
```diff
@@ -83,11 +82,11 @@ async def check_memory():
     p.info("memory")
 
     info = (await p.execute())[-1]
-    assert info["used_memory"] < 2 * 1_000_000  # Table memory
-    assert info["used_memory_rss"] < min_rss / 10  # RSS must have been freed
-
-
-#
+    assert info["used_memory"] < 4 * 1_000_000  # Table memory
+    # interned strings used by jsoncons are stored in a pool. The pool is not cleared on flushall
```
romange: hmm why actually? can we flush it?

abhijat: yes, we can flush it, I have added an API for it (`ResetPool`), but would need to decide which commands to call it within, eg flushdb.

dranikpg: Its not per db so it can only be with flushall, and we also have to be careful with tasks that copy table pointers and continue iterating. For example, freeing memory is async, so the pointers will be stale/corrupted if we flush immediately

romange: when I think about it - suppose we have 1M json objects in db0 and 5 json objects in db1.
flushdb will delete 1M strings from the pool and will leave there at most 5 strings.

Therefore, we do not need to do anything special in flushall/flushdb - they ONLY thing we need to do is to make sure that the set capacity is adequate to its expected size. i.e. if it becomes empty during the deletion - shrink to fit. If its load factor is less than `alpha < 1` - shrink to fit. I think flat_hash_set already has parameters that control the latter.

src/core/json/json_object.h
abhijat: I added some comparisons. The script performance comparison is predictable, mostly it is as my earlier tests. 

The `debug populate` comparison is also added. Noticeable things from that comparison:

* memory usage is lower in debug populate with random strings - I suspect collisions in random string keys caused interning to work - we should add stats as follow up to confirm
* RSS vs used gap is wider - this is the hashset, as we have more interned strings the hashset size (allocated on default heap) causes widening of the gap.
* RSS is not released on flushall - need to fix this as follow up probably

dranikpg: Is there currently any small string optimization in jsoncons? would losing it be noticable?

abhijat: There is SSO in jsoncons (short_string_storage) but afaics its only for values, not keys, which are just STL strings. 

We will lose the SSO benefit in the STL string that we used for keys with jsoncons.

dranikpg: True, it uses std::basic_string for keys by default. I would assume that in many real-world workloads keys would be short

abhijat: I think the tradeoffs can be summarized as below:

* lots of short, unrepeated keys: worst case - overhead of allocation but no deduplication benefits
* mix of short, long mostly unrepeated keys: still a little worse than STL, some overhead of allocation for short keys, not much deduplication
* short or long keys but with repetition: best case - benefit of deduplication overrides extra allocation

I'm hoping that due to the structure of large json objects repeated keys result in scenario 3 more often than 1 or 2, but making this tunable with fallback to STL string (which is just a single line - only need to switch one type definition) might be useful after some real world usage and stats

romange:
So the action items from this PR:
1. we should shrink the pool
2. we should expose stats

Are you planning to implement those here or in the following PR? We must have both before a release.

abhijat:
> So the action items from this PR:
> 
> 1. we should shrink the pool
> 2. we should expose stats
> 
> Are you planning to implement those here or in the following PR? We must have both before a release.

I would prefer to add both in follow ups, first the immediate one for shrink and then stats. 

But I'm open either way, if you strongly believe we should add here I can do that. Although for stats I would really prefer a follow up.

romange:
Sure, lets add the shrinkage here and the stats later

src/core/json/json_object.h
dranikpg: On the other hand we also have the benefit that an interned string is now just 8 bytes in size

abhijat:
> Sure, lets add the shrinkage here and the stats later

I have added pool shrinking call per shard to `flushall` code path. 

This enables removing the changes from `memory_test.py`. I manually checked that after `debug populate` followed by `flushall` the 300MiB gap is now no longer seen. RSS is back to normal levels after db is cleared. 

Updated the PR description.

abhijat:
augment review

src/server/json_family_memory_test.cc
```diff
@@ -130,7 +131,9 @@ TEST_F(JsonFamilyMemoryTest, JsonConsDelTest) {
   size_t memory_usage_after_erase = GetMemoryUsage() - start;
 
   EXPECT_GT(memory_usage_before_erase, memory_usage_after_erase);
-  EXPECT_EQ(memory_usage_after_erase, GetJsonMemoryUsageFromString(R"({"b":" "})"));
+  // b is interned, parsing it again will just reuse the same object and not use extra memory. to
+  // force a realistic comparison use a new character.
+  EXPECT_EQ(memory_usage_after_erase, GetJsonMemoryUsageFromString(R"({"x":" "})"));
```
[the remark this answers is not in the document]
abhijat: Not part of this PR

src/server/engine_shard.cc
```diff
@@ -552,10 +552,14 @@ void EngineShard::DestroyThreadLocal() {
 
   shard_->Shutdown();
 
+  detail::InternedString::ResetPool();
+  CleanupStatelessAllocMR();
```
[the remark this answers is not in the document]
abhijat: I'm not sure this is going to be an issue, but all json objects do use the stateless allocator. So its safer to clean up after the destructor runs. Fixed

src/server/db_slice.cc
```diff
@@ -962,10 +962,11 @@ util::fb2::Fiber DbSlice::FlushDbIndexes(const std::vector<DbIndex>& indexes) {
   LOG_IF(DFATAL, !fetched_items_.empty())
       << "Some operation might bumped up items outside of a transaction";
 
-  auto cb = [indexes, flush_db_arr = std::move(flush_db_arr)]() mutable {
+  auto cb = [flush_db_arr = std::move(flush_db_arr)]() mutable {
     flush_db_arr.clear();
     ServerState::tlocal()->DecommitMemory(ServerState::kDataHeap | ServerState::kBackingHeap |
                                           ServerState::kGlibcmalloc);
+    detail::InternedString::ShrinkPool();
```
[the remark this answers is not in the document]
abhijat: this is an interesting find. I think the present approach is fine for now and kind of unavoidable but I will try to think about fixing it.

abhijat:
augment review

src/server/db_slice.cc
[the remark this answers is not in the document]
abhijat: this is acceptable for now. I dont want to change order of existing code.

dranikpg: Flush is asynchronous, by the time you get to this line the db can be populated with new json items already. What you can only do is call `shrink_to_fit` but there's no such method, so you can constuct only a new pool

abhijat: I moved this to `InternedString::Release()`

src/server/engine_shard.cc
```diff
@@ -552,10 +552,14 @@ void EngineShard::DestroyThreadLocal() {
 
   shard_->Shutdown();
 
+  detail::InternedString::ResetPool();
```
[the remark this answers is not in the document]
abhijat: This is a very contrived scenario, I think the current approach is fine.

src/core/json/detail/interned_blob.cc
```diff
@@ -67,7 +68,7 @@ void InternedBlobHandle::DecrRefCount() {  // NOLINT - non-const, mutates via pt
 }
 
 size_t InternedBlobHandle::MemUsed() const {
-  return blob_ ? Size() + kHeaderSize + 1 : 0;
+  return blob_ ? mi_usable_size(blob_ - kHeaderSize) : 0;
```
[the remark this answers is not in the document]
abhijat: we're closely tied to mimalloc anyway. we need to see for eg. 16 bytes instead of 10 where a single char string takes 10 bytes, so this approach is correct.

tests/dragonfly/memory_test.py
```diff
@@ -1,10 +1,6 @@
-import pytest
-import asyncio
-from redis import asyncio as aioredis
-from .utility import *
-import logging
 from . import dfly_args
-from .instance import DflyInstance, DflyInstanceFactory
+from .instance import DflyInstanceFactory
+from .utility import *
```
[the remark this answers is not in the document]
abhijat: fixed

src/core/json/detail/interned_string.cc
```diff
@@ -3,6 +3,10 @@
 
 #include "core/json/detail/interned_string.h"
 
+namespace {
+constexpr auto kLoadFactorToShrinkPool = 0.01;
```
romange: I think it's too low. it catches flushdb scenario but it will be over provisioned during the regular operations

romange: i suggest setting it to 0.2 at least.

src/core/json/detail/interned_string.cc
```diff
@@ -48,8 +52,14 @@ void InternedString::Release() {
...
   if (entry_.RefCount() == 0) {
-    GetPoolRef().erase(entry_);
+    InternedBlobPool& pool_ref = GetPoolRef();
+    pool_ref.erase(entry_);
     InternedBlobHandle::Destroy(entry_);
+
+    if (const auto load_factor = pool_ref.load_factor();
+        ABSL_PREDICT_FALSE(load_factor > 0 && load_factor < kLoadFactorToShrinkPool)) {
+      InternedBlobPool(pool_ref).swap(pool_ref);
+    }
```
dranikpg, line 62: And if we use a higher values please verify it doesn't doom loop on growing or default initialization with a minimun size

dranikpg: Please also add a comment what we do, because it's not obvious immediately 🙂  Something like swap with copy to shrink

src/core/json/detail/interned_string.cc
abhijat: changed

src/core/json/detail/interned_string.cc
abhijat: I will verify this

romange (approved):

src/core/json/detail/interned_string.cc
abhijat: I will add a test for this code path and a comment in a follow up PR

abhijat: https://github.com/dragonflydb/dragonfly/pull/6544

I checked that there is not a possibility of a degenerate case where the pool keeps shrinking and growing on a small number of operations. 

After the swap the new pool has 0.8 load factor, so the next insert will trigger a resize but then load factor falls to something like 0.4, and then we have a large margin before the pool has to shrink again (`0.4 -> 0.2`).

Issue #5169, opened 2025-05-23
test_preempt_in_atomic_section_of_heartbeat


kostasrim:
https://github.com/dragonflydb/dragonfly/actions/runs/15193742986/job/42732700591

```
F20250522 09:48:10.333969 26191 db_slice.cc:1190] Check failed: shard_owner()->shard_lock()->Check(IntentLock::Mode::EXCLUSIVE) 0x4d0eb7  dfly::DashTable<>::Traverse<>()::{lambda()#2}::operator()()
0x4d56a6  dfly::detail::Segment<>::TraverseLogicalBucket<>()::{lambda()#1}::operator()<>()
0x4d5734  dfly::detail::Segment<>::Bucket::ForEachSlotImpl<>()
0x4d41f6  dfly::detail::Segment<>::Bucket::ForEachSlot<>()
0x4d2eef  dfly::detail::Segment<>::TraverseLogicalBucket<>()
```

kostasrim:
I think I understand what is happening. It's a subtle corner case (however I did not yet made this fail locally so I can't really confirm). 

```
  Transaction::Guard::~Guard() {                                                                         
    auto cb = [&](Transaction* t, EngineShard* shard) {                                                  
      namespaces->GetDefaultNamespace().GetDbSlice(shard->shard_id()).SetExpireAllowed(true);            
      return OpStatus::OK;                                                                               
    };                                                                                                   
    tx->Execute(cb, true);                                                                               
    tx->Refurbish();  
```

We call `SetExpireAllowed(true)` but we have not yet released the `shard lock`. How? We preempt during the journal write but before we finalize the transaction (see RunInShard). What I think the problematic flow is:

We call `RunInShard()` which calls `RunCallback()` which calls `LogAutoJournalToShard`. At this point, we called `SetExpireAllowed(true)` but we have `NOT` yet released the `shard lock`. Scan kicks in, and boom, `ExpireIfNeeded` sees that we can `expire` but fails on the `DCHECK` because the shard lock is not yet released. 

Any preemption point, between these two steps, is a potential violation of the contract in `ExpireIfNeeded`. 

FYI @adiholden

kostasrim:
p.s. I will take care of this on monday

kostasrim:
[...]

PR #6518, opened 2026-02-01
fix: QUIT pipeline flush

### LINKED ISSUE #6516: v1.36.0 breaks many tests in BullMQ
**Describe the bug**
This new version has introduced some regression that breaks many many tests in BullMQ. I cannot really see a pattern here yet, maybe you can point it out easier than us. But is seems like it is closing connections when
we do some operations maybe in LUA scripts or something...

**To Reproduce**
Try to run BullMQ test suite with dragonflyDB
https://github.com/taskforcesh/bullmq/actions/runs/21482055900/job/61880800894?pr=3654

**Expected behavior**
All tests should pass as they normally have been doing.

**Screenshots**

 - OS:  Ubuntu
  24.04.3
  LTS
 - Kernel: # Command: `uname -a`
 - Containerized?:Github Actions
 - Dragonfly Version: [e.g. 1.36.0]

**Reproducible Code Snippet**
```
```

**Additional context**

vyavdoshenko:
Summary:
- Fix the QUIT command in the pipeline, closing the socket before replies are flushed
- Fix key expiration being completely disabled during GLOBAL-mode Lua scripts (EVAL with allow-undeclared-keys)
- Rewrite BullMQ CI workflow to build from source

Changes:
- main_service.cc: Remove ShutdownSelfBlocking() from QUIT handler to avoid closing socket mid-pipeline
- dragonfly_connection.cc: Move socket shutdown to AsyncFiber after dispatch loop exits and replies are flushed
- connection_test.py: Add test_quit_in_pipeline regression test
- bullmq-tests.yml: Build Dragonfly from source instead of using the weekly container image

Success test run: https://github.com/dragonflydb/dragonfly/actions/runs/21563003843/job/62129895726

Partially Fixes #6516

### DIFF
--- a/.github/workflows/bullmq-tests.yml
+++ b/.github/workflows/bullmq-tests.yml
@@ -5,6 +5,9 @@ on:
 permissions:
   contents: read
 
+env:
+  NODE_VERSION: "22.12.0"
+
 jobs:
   build:
     runs-on: ubuntu-latest
@@ -37,10 +40,10 @@ jobs:
 
       - name: Install Node.js
         run: |
-          wget -q https://unofficial-builds.nodejs.org/download/release/v22.12.0/node-v22.12.0-linux-x64-glibc-217.tar.xz
-          tar -xf node-v22.12.0-linux-x64-glibc-217.tar.xz
-          cp -r node-v22.12.0-linux-x64-glibc-217/* /usr/local/
-          apt-get update && apt-get install -y jq
+          wget -q https://unofficial-builds.nodejs.org/download/release/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-x64-glibc-217.tar.xz
+          tar -xf node-v${NODE_VERSION}-linux-x64-glibc-217.tar.xz
+          cp -r node-v${NODE_VERSION}-linux-x64-glibc-217/* /usr/local/
+          apt-get update && apt-get install -y jq redis-tools
           npm install -g yarn
           node --version
           yarn --version
@@ -54,12 +57,11 @@ jobs:
             --default_lua_flags=allow-undeclared-keys \
             --dbfilename= \
             --port 6379 &
-          sleep 3
-          redis-cli -p 6379 PING
+          timeout 15s bash -c 'until redis-cli -p 6379 PING 2>/dev/null | grep -q PONG; do sleep 0.1; done'
 
       - name: Build BullMQ
         run: |
-          mkdir -p /tmp/bullmq && cd /tmp/bullmq
+          cd ${GITHUB_WORKSPACE}
           git clone https://github.com/dragonflydb/bullmq
           cd bullmq
           yarn install
@@ -67,7 +69,7 @@ jobs:
 
       - name: Run BullMQ tests
         run: |
-          cd /tmp/bullmq/bullmq
+          cd ${GITHUB_WORKSPACE}/bullmq
           BULLMQ_TEST_PREFIX={b} yarn test
 
       - name: Upload logs on failure
--- a/tests/dragonfly/connection_test.py
+++ b/tests/dragonfly/connection_test.py
@@ -1572,56 +1572,23 @@ async def test_quit_in_pipeline(df_server: DflyInstance):
     """
     Regression test: when QUIT is pipelined together with other commands
     (e.g. DEL DEL ... DEL QUIT), the server must flush all preceding replies
-    before closing the connection. Previously QUIT's ShutdownSelfBlocking()
-    closed the socket before SquashPipeline could flush the buffered DEL
-    responses, causing the client to see "Connection is closed".
+    before closing the connection.
 
-    This reproduces the BullMQ removeAllQueueData() pattern: scanStream +
-    pipeline DEL + client.quit() where pipeline.exec() is not awaited.
+    Reproduces the BullMQ removeAllQueueData() pattern.
     """
-
-    def resp_encode(*args):
-        parts = [f"*{len(args)}\r\n"]
-        for a in args:
-            a = str(a)
-            parts.append(f"${len(a)}\r\n{a}\r\n")
-        return "".join(parts)
-
     NUM_KEYS = 9
-    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-    s.connect(("127.0.0.1", df_server.port))
-    s.settimeout(2)
+    client = df_server.client()
 
     # Setup: create NUM_KEYS keys
-    setup = ""
     for i in range(NUM_KEYS):
-        setup += resp_encode("SET", f"{{b}}:pqt:k{i}", "v")
-    s.sendall(setup.encode())
+        await client.set(f"{{b}}:pqt:k{i}", "v")
 
-    # Read all +OK responses
-    data = b""
-    while data.count(b"+OK") < NUM_KEYS:
-        data += s.recv(4096)
-
-    # Send DEL for all keys + QUIT in one TCP write (pipeline)
-    pipeline = ""
+    # Send DEL for all keys + QUIT in one pipeline
+    pipe = client.pipeline(transaction=False)
     for i in range(NUM_KEYS):
-        pipeline += resp_encode("DEL", f"{{b}}:pqt:k{i}")
-    pipeline += resp_encode("QUIT")
-    s.sendall(pipeline.encode())
+        pipe.delete(f"{{b}}:pqt:k{i}")
+    pipe.execute_command("QUIT")
+    res = await pipe.execute()
 
-    # Read all responses: NUM_KEYS ":1\r\n" for DELs + "+OK\r\n" for QUIT
-    data = b""
-    try:
-        while True:
-            chunk = s.recv(4096)
-            if not chunk:
-                break
-            data += chunk
-    except socket.timeout:
-        pass
-
-    s.close()
-
-    assert data.count(b":1\r\n") == NUM_KEYS, f"Expected {NUM_KEYS} DEL replies, got: {data!r}"
-    assert b"+OK\r\n" in data, f"Expected QUIT OK reply, got: {data!r}"
+    assert res[:NUM_KEYS] == [1] * NUM_KEYS, f"Expected {NUM_KEYS} DEL replies, got: {res}"
+    assert res[NUM_KEYS] in (b"OK", True), f"Expected QUIT OK reply, got: {res[NUM_KEYS]}"


src/facade/dragonfly_connection.cc
```diff
@@ -1725,6 +1725,14 @@ void Connection::AsyncFiber() {
   }
 
   DCHECK(cc_->conn_closing || reply_builder_->GetError());
+
+  // If a command (e.g. QUIT) called CloseConnection(), the reply builder has an error
+  // but the socket is still open. Shut it down now that all replies have been flushed,
+  // which also wakes up the IoLoop blocked on socket reads.
+  if (reply_builder_->GetError() && !cc_->conn_closing) {
+    ShutdownSelfBlocking();
```
romange: It's not a good fix  but you found a problem. I will think how to fix it better!
cc'ing @dranikpg  - as well.
we do not want to "catch" quit condition based on these symptoms for sure.

src/server/db_slice.cc
```diff
@@ -1292,9 +1292,11 @@ DbSlice::PrimeItAndExp DbSlice::ExpireIfNeeded(const Context& cntx, PrimeIterato
...
   int64_t expire_time = ExpireTime(expire_it->second);
 
-  // Never do expiration on replica or if expiration is disabled or global lock was taken.
-  if (int64_t(cntx.time_now_ms) < expire_time || owner_->IsReplica() || !expire_allowed_ ||
-      !shard_owner()->shard_lock()->Check(IntentLock::Mode::EXCLUSIVE)) {
+  // Never do expiration on replica or if expiration is disabled.
+  // Note: we no longer skip expiration when a global (EXCLUSIVE) shard lock is held,
+  // because sub-commands within a global transaction (e.g. EVAL in GLOBAL mode) execute
+  // sequentially and won't hold iterators across calls, so deletion is safe.
+  if (int64_t(cntx.time_now_ms) < expire_time || owner_->IsReplica() || !expire_allowed_) {
```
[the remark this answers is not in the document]
romange: can you please explain what are you fixing here?

tests/dragonfly/connection_test.py
```diff
@@ -1565,3 +1565,63 @@ async def test_issue_6165_squash_invalid_syntax(async_client):
     res = await pip.execute(raise_on_error=False)
     assert res[0] == True  # SET key1
     assert isinstance(res[1], aioredis.ResponseError)  # INVALID SYNTAX
+
+
+@dfly_args({"proactor_threads": "2", "pipeline_squash": 1})
+async def test_quit_in_pipeline(df_server: DflyInstance):
+    """
+    Regression test: when QUIT is pipelined together with other commands
```
romange, line 1573: the test can be simplified. why do you need SET for issuing DELs? is it important to the test?
in general, lets minimize it to whatever is necessary to reproduce the bug.

Also, please consider using `async_client` with `pipeline(Transaction=False)` interface - see `run_pipeline_mode` for example

src/server/db_slice.cc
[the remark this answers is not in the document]
vyavdoshenko: Dragonfly doesn't remove the keys in Lua scripts because of EXCLUSIVE condition. BullMQ tests check the expired keys during the run, and these tests fail. From my understanding, it's safe to remove the expired keys even during EVAL execution.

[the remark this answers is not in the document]
dranikpg: I missed when this check was added. This is bad design. We already have `Transaction::Guard` that bans expiry by explicitly disabling it. We should see in the git history for what case this was added and fix it instead

src/facade/dragonfly_connection.cc
dranikpg: Having a control message or a safe shutdown function that sets flags and wakes up would probably be the easiest

tests/dragonfly/connection_test.py
dranikpg: Yes, we use sockets only if we're not okay with how the client sends/receives data. It seems like this test can easily be changed to just redis-py usage

src/server/db_slice.cc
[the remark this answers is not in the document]
dranikpg: Given that nothing from the tests failed we don't know what it was added for 🙂 But it can be some tricky corner case for migrations or any of the new code we added last year

.github/workflows/bullmq-tests.yml
```diff
@@ -10,62 +9,65 @@ jobs:
...
+          cd ${GITHUB_WORKSPACE}/build && ninja dragonfly
+
+      - name: Install Node.js
         run: |
           wget -q https://unofficial-builds.nodejs.org/download/release/v22.12.0/node-v22.12.0-linux-x64-glibc-217.tar.xz
           tar -xf node-v22.12.0-linux-x64-glibc-217.tar.xz
-          sudo cp -r node-v22.12.0-linux-x64-glibc-217/* /usr/local/
-          sudo apt install yarn jq
+          cp -r node-v22.12.0-linux-x64-glibc-217/* /usr/local/
+          apt-get update && apt-get install -y jq
```
glevkovich: nit:
versions repeat 3 times, its less maintainable and readable, and error prune. I suggest to use environment variables::

```
NODE_VER="v22.12.0"
NODE_DIST="node-${NODE_VER}-linux-x64-glibc-217"
          
wget -q https://unofficial-builds.nodejs.org/download/release/${NODE_VER}/${NODE_DIST}.tar.xz
tar -xf ${NODE_DIST}.tar.xz
cp -r ${NODE_DIST}/* /usr/local/
```

.github/workflows/bullmq-tests.yml
```diff
@@ -10,62 +9,65 @@ jobs:
...
+      - name: Start Dragonfly
+        run: |
+          ${GITHUB_WORKSPACE}/build/dragonfly \
+            --alsologtostderr \
+            --cluster_mode=emulated \
+            --lock_on_hashtags \
+            --default_lua_flags=allow-undeclared-keys \
+            --dbfilename= \
+            --port 6379 &
+          sleep 3
```
glevkovich: nit: consider a safer / faster approach:
`timeout 15s bash -c 'until redis-cli -p 6379 PING; do sleep 0.1; done'`

.github/workflows/bullmq-tests.yml
```diff
@@ -10,62 +9,65 @@ jobs:
...
         run: |
-          mkdir -p $GITHUB_WORKSPACE/../bullmq
-          cd $GITHUB_WORKSPACE/../bullmq
-          # TODO: Use BullMQ latest release instead of our fork:
-          # DOWNLOAD_URL=$(curl -s https://api.github.com/repos/taskforcesh/bullmq/releases/latest | jq -r '.tarball_url')
-          # echo "Downloading latest BullMQ release from ${DOWNLOAD_URL}"
-          # wget -O bullmq.tar.gz ${DOWNLOAD_URL}
-          # tar -zxvf bullmq.tar.gz
-          # mv taskforcesh-bullmq-* bullmq
+          mkdir -p /tmp/bullmq && cd /tmp/bullmq
```
glevkovich: Are you intentionally using /tmp/ ? As far as i know Guthub actions knows how to handle disk IO better, with better disk performance when using $GITHUB_WORKSPACE. I suggest tohave the repo in $GITHUB_WORKSPACE/bullmq .

src/server/db_slice.cc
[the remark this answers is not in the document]
romange: > Dragonfly doesn't remove the keys in Lua scripts because of EXCLUSIVE condition. BullMQ tests check the expired keys during the run, and these tests fail. From my understanding, it's safe to remove the expired keys even during EVAL execution.

We need a unit test for this. In dragonfly_test or multi_test  - that verifies the correct behavior. I suggest taking it out of this PR and do it separately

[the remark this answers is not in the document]
dranikpg: https://github.com/dragonflydb/dragonfly/pull/4967
https://github.com/dragonflydb/dragonfly/pull/5151
https://github.com/dragonflydb/dragonfly/issues/5169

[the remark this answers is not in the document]
kostasrim: @dranikpg bad design or not we just added a new bug here. 

`Heartbeat` will not `skip` expirations on keys that are locked. So for multi hop transactions this will break the state. 

In other words, this change should break `ExpireIfNeeded` + heartbeat unless I am missing something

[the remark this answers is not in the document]
kostasrim: oh hold that for for the global lock (shard lock), just a second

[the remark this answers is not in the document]
kostasrim: Ok now I remember: 

The bug:
entries were expired at the time of transitioning from full sync to stable sync leading to journal changes not propagated to replica.
The fix:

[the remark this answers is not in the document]
kostasrim: This change will break data parity for replication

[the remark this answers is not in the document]
kostasrim: > Given that nothing from the tests failed we don't know what it was added for 🙂 But it can be some tricky corner case for migrations or any of the new code we added last year

I know why cause I helped with that investigation and I now remember. In short, we break data parity in replication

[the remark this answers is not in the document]
kostasrim: and a small rant to myself -- ALWAYS leave a comment for cases like that. Otherwise we all question the rationale and some corner cases which do not have unit tests are hard to verify :/ 

always always leave a comment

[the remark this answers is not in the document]
dranikpg: But don't we already skip the heartbeat if a global lock is acquired?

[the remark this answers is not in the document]
kostasrim: hmm actually this is also true 🤔 (and I think I added that check) + all the changes that made heartbeat non-preemptive (i.e. `atomic`).

I will take a look again a little later but it seems you are right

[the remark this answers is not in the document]
vyavdoshenko: @dranikpg @kostasrim
I suggest to continue discussion here: https://github.com/dragonflydb/dragonfly/pull/6524
I will remove this change here.

.github/workflows/bullmq-tests.yml
romange: Also, please consider using the github action that installs node https://github.com/actions/setup-node

.github/workflows/bullmq-tests.yml
```diff
@@ -10,62 +9,65 @@ jobs:
...
-#          cmake .. -DCMAKE_BUILD_TYPE=Release -GNinja
-#          ninja dragonfly
-#          ./dragonfly --alsologtostderr --cluster_mode=emulated --lock_on_hashtags --dbfilename= &
+      - name: Start Dragonfly
+        run: |
+          ${GITHUB_WORKSPACE}/build/dragonfly \
+            --alsologtostderr \
+            --cluster_mode=emulated \
+            --lock_on_hashtags \
+            --default_lua_flags=allow-undeclared-keys \
```
romange, line 57: do we need `default_lua_flags` here?

.github/workflows/bullmq-tests.yml
```diff
@@ -10,62 +9,65 @@ jobs:
...
           cd bullmq
-          pwd
           yarn install
           yarn build
 
       - name: Run BullMQ tests
         run: |
-          cd $GITHUB_WORKSPACE/../bullmq/bullmq
+          cd /tmp/bullmq/bullmq
           BULLMQ_TEST_PREFIX={b} yarn test
```
romange, line 73: if some tests fail but many pass, consider using `--testPathIgnorePatterns` or equivalent to establish a green base line. this will also give us clarity what tests need to be investigated

tests/dragonfly/connection_test.py
```diff
@@ -1565,3 +1565,63 @@ async def test_issue_6165_squash_invalid_syntax(async_client):
...
+    pipeline = ""
+    for i in range(NUM_KEYS):
+        pipeline += resp_encode("DEL", f"{{b}}:pqt:k{i}")
+    pipeline += resp_encode("QUIT")
+    s.sendall(pipeline.encode())
+
+    # Read all responses: NUM_KEYS ":1\r\n" for DELs + "+OK\r\n" for QUIT
+    data = b""
+    try:
+        while True:
```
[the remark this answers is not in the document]
vyavdoshenko: Not actual, test simpified

tests/dragonfly/connection_test.py
vyavdoshenko: Test simplified. Please take a look.

.github/workflows/bullmq-tests.yml
vyavdoshenko: Fixed

.github/workflows/bullmq-tests.yml
vyavdoshenko: Fixed

.github/workflows/bullmq-tests.yml
vyavdoshenko: @glevkovich - Fixed.

@romange
actions/setup-node doesn't work in our container — Node 20+ requires glibc >= 2.28, but ubuntu-dev:20-gcc14 has glibc 2.17. That's why we use unofficial builds with the glibc-217 suffix.

.github/workflows/bullmq-tests.yml
vyavdoshenko: `--default_lua_flags=allow-undeclared-keys` is needed because some BullMQ tests (4 tests) intentionally access keys not passed via KEYS[] (e.g., the Telemetry Flows error-handling test does redis.call('GET', 'invalidQueue:invalidParentId') where that key is not in KEYS[]). Without this flag, Dragonfly rejects undeclared key access.

.github/workflows/bullmq-tests.yml
vyavdoshenko: Why? I would like to merge this PR and https://github.com/dragonflydb/dragonfly/pull/6524 PR. All tests must be green after these PRs.

romange: we do not run dragonfly with bullmq with `--default_lua_flags=allow-undeclared-keys` in production.
so once we identify the problematic tests we can check with bullmq maintainers what we can do.

.github/workflows/bullmq-tests.yml
romange: see below. we can  maybe fix the tests by respecting `BULLMQ_TEST_PREFIX` or we can exclude them. in any case `allow-undeclared-keys` may hide real problems. we can keep it as is and follow up later.

romange (approved):

PR #6519, opened 2026-02-02
feat: add serialization for shard DocId mapping

BorysTheDev:
added serialization for DocID key_name mapping per shard. Further step index recreation during loading

### DIFF
--- a/src/server/snapshot.cc
+++ b/src/server/snapshot.cc
@@ -183,7 +183,7 @@ void SliceSnapshot::SerializeIndexMapping(
     if (auto ec = serializer_->SaveLen(doc_id); ec)
       return;
   }
-  ThrottleIfNeeded();
+  PushSerialized(false);
 }
 
 void SliceSnapshot::SerializeIndexMappings() {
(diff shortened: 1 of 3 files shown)


src/server/snapshot.cc
```diff
@@ -161,6 +163,54 @@ void SliceSnapshot::FinalizeJournalStream(bool cancel) {
...
+    }
+
+    auto mappings = index->SerializeKeyIndex();
+    if (mappings.empty()) {
+      continue;
+    }
+
+    SerializeIndexMapping(shard_id, index_name, mappings);
+  }
+#endif
```
dranikpg: We don't need it for regular indices

src/server/snapshot.cc
```diff
@@ -161,6 +163,54 @@ void SliceSnapshot::FinalizeJournalStream(bool cancel) {
...
+  if (auto ec = serializer_->SaveLen(mappings.size()); ec)
+    return;
+
+  for (const auto& [key, doc_id] : mappings) {
+    if (auto ec = serializer_->SaveString(key); ec)
+      return;
+    if (auto ec = serializer_->SaveLen(doc_id); ec)
+      return;
+  }
+  ThrottleIfNeeded();
```
dranikpg: PushSerialized()

PR #6524, opened 2026-02-02
fix: allow lazy expiry during global EVAL transactions

vyavdoshenko:
Remove the redundant `shard_lock()->Check(EXCLUSIVE)` guard from `ExpireIfNeeded` that prevented lazy key expiration during global transactions (EVAL with `allow-undeclared-keys`). This caused expired keys to be returned as still alive from Lua scripts, breaking BullMQ and similar libraries that rely on TTL semantics inside EVAL.

- **db_slice.cc**: Remove `shard_lock()->Check(EXCLUSIVE)` from `ExpireIfNeeded`. The `expire_allowed_` flag remains as the mechanism for controlling expiry during replication transitions.
- **multi_test.cc**: Add `EvalGlobalLazyExpire` C++ unit test that verifies expired keys return nil when accessed via global-mode EVAL.

Related to #6516


src/server/db_slice.cc
```diff
@@ -1292,9 +1292,8 @@ DbSlice::PrimeItAndExp DbSlice::ExpireIfNeeded(const Context& cntx, PrimeIterato
   // TODO: to employ multi-generation update of expire-base and the underlying values.
   int64_t expire_time = ExpireTime(expire_it->second);
 
-  // Never do expiration on replica or if expiration is disabled or global lock was taken.
-  if (int64_t(cntx.time_now_ms) < expire_time || owner_->IsReplica() || !expire_allowed_ ||
-      !shard_owner()->shard_lock()->Check(IntentLock::Mode::EXCLUSIVE)) {
+  // Never do expiration on replica or if expiration is disabled.
+  if (int64_t(cntx.time_now_ms) < expire_time || owner_->IsReplica() || !expire_allowed_) {
```
kostasrim, line 1296: I was wrong in respect to my comment. Heartbeat is fully atomic and we skip it if shard lock is active

kostasrim: @dranikpg +1 you were right

src/server/multi_test.cc
```diff
@@ -1474,4 +1474,25 @@ TEST_F(MultiTest, StoredCmdBytesMetric) {
   ASSERT_EQ(GetMetrics().coordinator_stats.stored_cmd_bytes, 0);
 }
 
+// Verify that lazy expiration works inside EVAL running in global mode
+// (allow-undeclared-keys). Previously, the shard_lock()->Check(EXCLUSIVE) guard
+// in ExpireIfNeeded prevented any lazy expiry while a global transaction held
+// the shard lock, causing expired keys to be returned as if they were still alive.
+TEST_F(MultiEvalTest, EvalGlobalLazyExpire) {
```
kostasrim: In order for this to work you need to set: default_lua_flags=allow-undeclared-keys for the transaction run in 'global` mode.

Also, why do we need to loop 50 times? 

p.s. you did that with the replication test ;)

src/server/multi_test.cc
```diff
@@ -1474,4 +1474,25 @@ TEST_F(MultiTest, StoredCmdBytesMetric) {
...
+  for (int i = 0; i < 50; i++) {
+    string key = absl::StrCat("key", i);
+
+    // Use a large TTL so real wall-clock time cannot expire it before EVAL.
```
romange: this is incorrect. we control the clock in tests and it only advances with `AdvanceTime`

src/server/db_slice.cc
dranikpg: Image

src/server/multi_test.cc
dranikpg: One iteration with a script that is actually global (starts with `--!df flags=allow-undeclared-keys`, see examples in multi_test)

tests/dragonfly/replication_test.py
```diff
@@ -3904,3 +3904,68 @@ async def compare_group_info(stream_key, expected_pending, expected_entries_read
...
+    await c_replica.execute_command(f"REPLICAOF localhost {master.port}")
+
+    n_keys = 500
+
+    # Set keys with 1ms TTL via EVAL. Each EVAL holds the global shard lock
+    # during execution, so heartbeat cannot expire the key between SET and PEXPIRE.
+    set_script = "redis.call('SET', KEYS[1], ARGV[1]); redis.call('PEXPIRE', KEYS[1], 1)"
+    pipe = c_master.pipeline(transaction=False)
+    for i in range(n_keys):
+        pipe.eval(set_script, 1, f"expkey:{i}", f"val:{i}")
```
dranikpg: I don't really understand the logic. Heartbeat can't expire between SET and PEXPIRE because there is no expiry. It can expire between different scripts however. What is more, you can just achieve the same with `SET <key> <value> PX ...`

romange:
@vyavdoshenko  I am just putting it here for context: https://github.com/dragonflydb/dragonfly/pull/5151

I think it's important to discuss why it was added in the first place.

dranikpg:
> @vyavdoshenko I am just putting it here for context: https://github.com/dragonflydb/dragonfly/pull/5151
> I think it's important to discuss why it was added in the first place.

I left comments in the other PR. It was before heartbeat was atomic and it was also decided to guard inside db_slice code instead of fixing heartbeat

tests/dragonfly/replication_test.py
vyavdoshenko: Not relevant.
I removed this test, I guess cpp test is enough.

src/server/multi_test.cc
vyavdoshenko: Fixed

src/server/multi_test.cc
vyavdoshenko: Fixed.
Please take a look again.

PR #6528, opened 2026-02-03
fix(server): Fix concurrent access to now_streamed_ (#6486) [backport]

abhijat:
Fix concurrent access to now_streamed_ from CreateObjectOnShard

(cherry picked from commit 387a5f2d3951fc5568c90e62cecf90d38ddfab9e)

### DIFF
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -2646,7 +2646,7 @@ void RdbLoader::CreateObjectOnShard(const DbContext& db_cntx, const Item* item, 
 
   if (item->load_config.streamed) {
     std::unique_lock lk{now_streamed_mu_};
-    if (!now_streamed_.contains(item->key))
+    if (const auto it = now_streamed_.find(item->key); it == now_streamed_.end())
       now_streamed_.emplace(item->key, make_unique<PrimeValue>(std::move(pv)));
 
     if (!item->load_config.finalize)


abhijat:
the failure is because the fakeredis CI setup is behind I think:

```
/__w/dragonfly/dragonfly/src/server/rdb_load.cc:2649:24: error: 'class std::unordered_map<std::__cxx11::basic_string<char>, std::unique_ptr<dfly::CompactValue> >' has no member named 'contains'
 2649 |     if (!now_streamed_.contains(item->key))
```

not at C++20. Not a problem with the PR and we should certainly not change the code to make it build with the workflow

not sure why it passed here though https://github.com/dragonflydb/dragonfly/actions/runs/21624235965/job/62320266966?pr=6527

abhijat:
> the failure is because the fakeredis CI setup is behind I think:
> 
> ```
> /__w/dragonfly/dragonfly/src/server/rdb_load.cc:2649:24: error: 'class std::unordered_map<std::__cxx11::basic_string<char>, std::unique_ptr<dfly::CompactValue> >' has no member named 'contains'
>  2649 |     if (!now_streamed_.contains(item->key))
> ```
> 
> not at C++20. Not a problem with the PR and we should certainly not change the code to make it build with the workflow
> 
> not sure why it passed here though https://github.com/dragonflydb/dragonfly/actions/runs/21624235965/job/62320266966?pr=6527

trying to fix here https://github.com/dragonflydb/dragonfly/pull/6529

dranikpg:
Ah, actually yes, we shouldn't rely on c++ 20 in older branches. Maybe it's easiest to fix it to use contains and just merge it

src/server/rdb_load.cc
```diff
@@ -2638,12 +2641,12 @@ void RdbLoader::CreateObjectOnShard(const DbContext& db_cntx, const Item* item,
...
     LOG(ERROR) << "Could not load value for key '" << item->key << "' in DB " << db_ind;
     stop_early_ = true;
-    now_streamed_.clear();
     return;
   }
 
   if (item->load_config.streamed) {
-    if (now_streamed_.find(item->key) == now_streamed_.end())
+    std::unique_lock lk{now_streamed_mu_};
+    if (const auto it = now_streamed_.find(item->key); it != now_streamed_.end())
```
dranikpg: Ah, wait, you mixed up the condition here. Just leave it as it was before I replaced it with contains

abhijat: sorry, yeah fixed

PR #6538, opened 2026-02-04
fix(facade): fix migration stats accounting crash

glevkovich:
Previously, connection migration only transferred connection counts and IO buffer capacities between threads. It failed to account for pending items in the dispatch queue (Control Path) and pipeline queue (Data Path).

This caused the old thread to retain "ghost" usage stats (leaks) while the new thread failed to account for the migrated items. Processing these inherited items on the new thread caused statistical drift, eventually triggering underflow crashes (DCHECK failures) when thread-local counters reached zero.

This commit fixes the issue by:
- Adding local variables to track dispatch queue bytes within the Connection object (dispatch_q_bytes_, dispatch_q_subscriber_bytes_).
- Updating `Increase/DecreaseConnStats` to fully transfer all stateful statistics (Dispatch, Pipeline, and Subscriber bytes) during migration.
- Ensuring local and global stats remain in sync during normal operations via `UpdateDispatchStats`.
- Using `absl::Cleanup` in `ProcessAdminMessage` to guarantee stats are updated on all return paths.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1022,6 +1022,9 @@ io::Result<bool> Connection::CheckForHttpProto() {
 void Connection::ConnectionFlow() {
   DCHECK(reply_builder_);
 
+  // Register the new connection with the thread-local statistics.
+  // At this point (connection birth), local queue stats/luggage are 0,
+  // so only connection counts and buffer capacities are incremented.
   IncreaseConnStats();
   ++stats_->conn_received_cnt;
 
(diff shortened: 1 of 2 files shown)


vyavdoshenko (commented):
lgtm, but I always try to add a crash test that reproduces the issue.
Could you add the test that crashes on the main branch and green with fixes?

glevkovich:
> lgtm, but I always try to add a crash test that reproduces the issue. Could you add the test that crashes on the main branch and green with fixes?

@vyavdoshenko but there is already one or even few that reproduce it, I ran it on my laptop before the change - crashes few times. Then I ran it 1500 times after my fix - no crash.
I don't see a reason why adding new test will help?

vyavdoshenko:
> > lgtm, but I always try to add a crash test that reproduces the issue. Could you add the test that crashes on the main branch and green with fixes?
> 
> @vyavdoshenko but there is already one or even few that reproduce it, I ran it on my laptop before the change - crashes few times. Then I ran it 1500 times after my fix - no crash. I don't see a reason why adding new test will help?

OK

PR #6539, opened 2026-02-04
fix(search): Avoid second traversal without vector fields

dranikpg:
and also refactor it a little to call AddDocToGlobalVectorIndex from RebuildGlobalVEctorIndices

### DIFF
--- a/src/server/search/doc_index.cc
+++ b/src/server/search/doc_index.cc
@@ -117,7 +117,7 @@ std::pair<std::vector<FieldReference>, std::vector<bool>> GetBasicFields(
   return {std::move(basic_fields), std::move(is_sortable_field)};
 }
 
-auto IndexableVectorFields(const search::Schema& schema) {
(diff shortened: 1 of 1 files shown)


dranikpg:
augment review

src/server/search/doc_index.cc
```diff
@@ -116,6 +117,12 @@ std::pair<std::vector<FieldReference>, std::vector<bool>> GetBasicFields(
   return {std::move(basic_fields), std::move(is_sortable_field)};
 }
 
+auto IndexableVectorFields(const search::Schema& schema) {
```
romange: do not replace pod types like bool with `auto`

romange: ah sorry it's not bool

romange: IndexableVectorFields -> `GetIndexableVectorFieldsRange`

mkaruza (approved):

PR #6542, opened 2026-02-04
feat(tiering): SSD tiering support during slot migration

### LINKED ISSUE #5616: Add SSD tiering support in cluster slot migration
In the CmdSerializer class, we need to add logic for strings that aren't in memory.
Current state:
```
    if (pv.IsCool()) {
      pv.GetCool().record->value.GetString(&str);
    }
    LOG(FATAL) << "External string not supported yet";
```

mkaruza:
Tiered entries are requested for fetching and collected in container
that is processed after each bucket scan is done. Same logic is used
during snapshotting.
    
Closes #5616

### DIFF
--- a/tests/dragonfly/cluster_test.py
+++ b/tests/dragonfly/cluster_test.py
@@ -3749,19 +3749,29 @@ async def test_cluster_migration_with_tiering_and_deletes(df_factory: DflyInstan
     await push_config(json.dumps(generate_config(nodes)), [node.admin_client for node in nodes])
 
     # Delete 50k keys during migration to create mutations and verify that they are applied correctly
-    delete_num = 50_000
+    delete_expected_num = 50_000
+    delete_succeded = 0
+
+    # Indicator that migration is done and we can stop deleting keys
+    migration_done = False
 
     async def delete_job():
-        keys_to_delete = [f"key:{i}" for i in range(0, delete_num)]
-        for key in keys_to_delete:
+        nonlocal delete_succeded
+        for i in range(delete_expected_num):
+            if migration_done:
+                break
             try:
-                await nodes[0].client.delete(key)
+                await nodes[0].client.delete(f"key:{i}")
+                delete_succeded += 1
             except Exception as e:
-                logging.info(f"Error deleting key {key}: {e}")
+                pass
 
-    asyncio.create_task(delete_job())
+    delete_task = asyncio.create_task(delete_job())
 
     await wait_for_status(nodes[0].admin_client, nodes[1].id, "FINISHED", 300)
+    migration_done = True
+
+    await delete_task
 
     nodes[0].migrations = []
     nodes[0].slots = []
@@ -3780,4 +3790,4 @@ async def test_cluster_migration_with_tiering_and_deletes(df_factory: DflyInstan
 
     # Verify that mutations are applied on the target node after migration
     info = await nodes[1].client.info("keyspace")
-    assert info["db0"]["keys"] == keys - delete_num
+    assert info["db0"]["keys"] == keys - delete_succeded


src/server/journal/cmd_serializer.cc
```diff
@@ -118,6 +124,30 @@ size_t CmdSerializer::SerializeEntry(string_view key, const PrimeKey& pk, const
...
 
+size_t CmdSerializer::SerializeDelayedEntries(bool force) {
+  if (!force && delayed_entries_.size() < 32)
+    return 0;
+
+  size_t serialized = 0;
+  if (!delayed_entries_.empty()) {
+    do {
+      TieredDelayedEntry entry = std::move(delayed_entries_.back());
+      delayed_entries_.pop_back();
```
dranikpg: Some comments wouldn't hurt, at least referring to the same mechanism inside snapshot.cc because now it's not evident at all to someone who reads it for the first time why some things are the way they are

src/server/journal/streamer.cc
```diff
@@ -487,6 +489,9 @@ void RestoreStreamer::Run() {
       stats_.buckets_loop += WriteBucket(it, expire_table);
     });
 
+    // Serialize any pending delayed entries.
+    cmd_serializer_->SerializeDelayedEntries(false);
```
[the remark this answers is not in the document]
dranikpg: It seems to me that the snapshot code is wrong as well. We need to serialize the delayed entries before we serialize and new journal change.

tests/dragonfly/cluster_test.py
```diff
@@ -3640,3 +3640,66 @@ async def test_remove_docs_on_cluster_migration(df_factory):
...
+            logging.info(f"Tiered entries: {info['tiered_entries']}")
+            assert info["tiered_entries"] >= 10_000
+
+    nodes[0].migrations.append(
+        MigrationInfo("127.0.0.1", instances[1].port, [(0, 16383)], nodes[1].id)
+    )
+
+    await push_config(json.dumps(generate_config(nodes)), [node.admin_client for node in nodes])
+
+    await wait_for_status(nodes[0].admin_client, nodes[1].id, "FINISHED", 300)
```
dranikpg, line 3687: Please add some traffic and consistency checks

src/server/journal/streamer.cc
```diff
@@ -604,16 +623,39 @@ bool RestoreStreamer::WriteBucket(PrimeTable::bucket_iterator it, const ExpireTa
...
+    // for force-flushing their delayed entries
+    if (track_tiered_keys) {
+      for (it.AdvanceIfNotOccupied(); !it.is_done(); ++it) {
+        const auto& pv = it->second;
+        string_view key = it->first.GetSlice(&key_buffer);
+        if (ShouldWrite(key) && pv.IsExternal() && !pv.IsCool()) {
+          tiered_keys.emplace(key);
+        }
+      }
+    }
```
dranikpg, line 647: its a bit wasteful to go over the whole bucket and decode the key, grouping delayed_entries by something like the segment id would be enough

dranikpg: because this will be actionable only the first time a bucket is accessed after delayed entries were populated

mkaruza: @dranikpg 

Yeah i agree, we have multiple way how to make this more efficient.

IMO, i think that key name should be passed  somehow to this callback. With this we will forcefully serialize only one delayed key and not whole bucket.

dranikpg: I think that the "slow" path - the conflciting case is very rare in general, passing the specific key won't likely imporve it much

tests/dragonfly/cluster_test.py
```diff
@@ -3640,3 +3640,154 @@ async def test_remove_docs_on_cluster_migration(df_factory):
...
+    await asyncio.sleep(5)  # wait for tiered deletions to finish
+
+    info = await nodes[0].client.info("keyspace")
+    assert info["db0"]["keys"] == 0
+
+
+@pytest.mark.exclude_epoll
+@pytest.mark.opt_only
+@dfly_args({"cluster_mode": "yes"})
+async def test_cluster_migration_with_tiering_and_deletes(df_factory: DflyInstanceFactory):
```
dranikpg, line 3711: 1. Maybe one test is enough. Unless it deletes 99% of the values, it should test regular transfer
2. Append is really the best command beacuse it is stateful and order dependant, you can issue multiple appends, but not multiple deletes
3. You can still use consistency checks  with captures

PR #6543, opened 2026-02-04
fix: search and dfly lib circular dep

### LINKED ISSUE #6479: remove circular dependencies
See comment from https://github.com/dragonflydb/dragonfly/pull/6464/changes#r2726657385

I will write a plan/direction once I get a chance

kostasrim:
There is `circular dependency` between dragonfly search server and dragonfly lib as both link against each other.

I understand the rationale of wanting to separate search as separate library. It's also important to switch off/on search for a cmake build. What we did here however is not correct and in fact, `search` is a `part` of dragonfly lib. If you think otherwise, read `search_family.cc`. This unit belongs to dragonfly lib.

Therefore this PR:

* removes search library all together and moves it as part of dragonfly lib
* allow to optionally compile search (to respect previous use cases)
* keeps original folder hierarchy same (all search units go to server/search/*)

resolves #6479

### DIFF
--- a/src/server/CMakeLists.txt
+++ b/src/server/CMakeLists.txt
@@ -43,6 +43,10 @@ cxx_link(dfly_transaction dfly_core strings_lib TRDP::fast_float TRDP::hdr_histo
 
 # Include search module
 add_subdirectory(search)
+if(NOT DEFINED DF_SEARCH_SRCS)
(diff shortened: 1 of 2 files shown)


dranikpg (approved):
Best would be to have search_family in lib and the rest as a separate target

kostasrim:
> Best would be to have search_family in lib and the rest as a separate target

I do not have a strong opinion about this IMO 🤷 We can always follow up

PR #6551, opened 2026-02-05
test: add search with replication for 1.36

BorysTheDev:
Make sure we can replicate all indices (both search and vector) via rebuild.
We can run vector search together with geo radius and category tags

PR was tested on v1.36-branch

### DIFF
--- a/tests/dragonfly/search_test.py
+++ b/tests/dragonfly/search_test.py
@@ -948,52 +948,67 @@ async def test_vector_search_with_geo_and_tags(async_client: aioredis.Redis):
             f"'category', '{cat}'" not in result_str and f"b'category', b'{cat}'" not in result_str
         )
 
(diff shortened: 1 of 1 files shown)


BorysTheDev:
@copilot review and fix problems

romange:
@BorysTheDev  why does it crash on CI?

BorysTheDev:
> @BorysTheDev why does it crash on CI?

I know this crash on main, but I've checked 1.36 it should work fine, I will check one more time tomorrow

romange:
If it should work on main as well, the PR should also go to main at some point.

BorysTheDev:
@romange For now these tests are failed only for GEO search because of DCHECK for unsorted results. We lack this functionality. Should I comment out this part of the tests?

romange:
Yes, and pls open the issue

PR #6553, opened 2026-02-05
fix: hnsw global serialization and add flag for hnsw save\load

BorysTheDev:
fixes crash in hnsw serialization
add flag save_vector_index to use hnsw serialization

### DIFF
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -55,7 +55,7 @@ extern "C" {
 ABSL_DECLARE_FLAG(int32_t, list_max_listpack_size);
 ABSL_DECLARE_FLAG(int32_t, list_compress_depth);
 ABSL_DECLARE_FLAG(uint32_t, dbnum);
-ABSL_DECLARE_FLAG(bool, save_vector_index);
+ABSL_FLAG(bool, deserialize_hnsw_index, false, "Deserialize HNSW vector index graph structure");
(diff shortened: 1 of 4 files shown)


dranikpg (dismissed):
maybe `serialize_hnsw_index` would be a better name

src/server/rdb_load.cc
```diff
@@ -2267,6 +2268,10 @@ error_code RdbLoader::Load(io::Source* src) {
         pim.mappings.emplace_back(std::move(key), static_cast<search::DocId>(doc_id));
       }
 
+      if (!GetFlag(FLAGS_save_vector_index)) {
```
romange: can you please explain the change in this file?

romange: In other words, why do we need a flag in the loader?

BorysTheDev: we ignore index data if the flag is false. But we still need to read it.

BorysTheDev: I've added this flag for the case if index rebuilding will fail for some reason

romange: I think saving and loading are differrent use-cases. And sometimes we want to load and rebuild and then serialize it or the opposite. So we need two flags with different names: `serialize_hnsw_index` and `ignore_hnsw_index_on_load` or something like that

BorysTheDev: you can do this with the same flag, too my mind, it's superfluous, but ok I will add it

PR #6569, opened 2026-02-09
chore: supress maint_notifications client errors

romange:
Summary: Reduces warning-log noise by suppressing expected failures for CLIENT maint_notifications (and -BUSYGROUP) via a centralized ShouldLogError() helper.
Changes: Cleans up several deprecated ABSL_RETIRED_FLAG declarations that are no longer needed.

### DIFF
--- a/src/server/dflycmd.cc
+++ b/src/server/dflycmd.cc
@@ -37,8 +37,6 @@
 #include "util/fibers/synchronization.h"
 using namespace std;
 
-ABSL_RETIRED_FLAG(uint32_t, allow_partial_sync_with_lsn_diff, 0,
-                  "Do partial sync in case lsn diff is less than the given threshold");
 ABSL_DECLARE_FLAG(bool, info_replication_valkey_compatible);
 ABSL_DECLARE_FLAG(uint32_t, replication_timeout);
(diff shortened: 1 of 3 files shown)


romange:
augment review

romange:
> lgtm, but I had the same question as copilot about the wide ranging suppression.

Where do I have wide ranging suppression?

abhijat:
> Where do I have wide ranging suppression?

I think so because the code now ignores whatever is after `maint_notifications`. I searched for this string in code but couldn't find it, maybe it's safe to ignore everything after that word. 

If its the only expected string after `client` then the condition should be stricter (tail size == x/nothing after the last string). But maybe its overkill. 

```
!absl::EqualsIgnoreCase(tail_args.front(), "maint_notifications")
```

romange:
Sorry I do not understand what you are saying :)

The check filters out all the errors coming from "clirent maint_notifications ... " subcommand. but it will pass all other "client" subcommand errors

romange:
We do not support this command but it adds a noise to our warning logs.

abhijat:
> The check filters out all the errors coming from "clirent maint_notifications ... " subcommand

Its fine then, I misunderstood the scope, I thought it was filtering only for specific:

`client maint_notifications`

but if it intends to ignores all errors starting at that phrase, then its good.

Issue #6573, opened 2026-02-09
Dragonfly crashes in transaction.cc


vyavdoshenko:
[traffic.tar.gz](https://github.com/user-attachments/files/25183129/traffic.tar.gz)

`./traffic-replay run traffic/traffic*`

```
F20260209 14:43:00.470770 25590 transaction.cc:358] Check failed: full_args_.size() % key_index.step == 0u (1 vs. 0) [key value,key,8WEifoy5499iKMXbH2]
*** Check failure stack trace: ***
    @     0xaaaae56a85d0  google::LogMessage::Fail()
    @     0xaaaae56a84d8  google::LogMessage::SendToLog()
    @     0xaaaae56a7cd8  google::LogMessage::Flush()
    @     0xaaaae56ab9e0  google::LogMessageFatal::~LogMessageFatal()
    @     0xaaaae51458c4  dfly::Transaction::InitByKeys()
    @     0xaaaae5146498  dfly::Transaction::InitByArgs()
    @     0xaaaae4d10394  dfly::(anonymous namespace)::PrepareTransaction()
    @     0xaaaae4d15adc  dfly::Service::DispatchCommand()
    @     0xaaaae53d7524  _ZZN6facade10Connection10ParseRedisEjENKUlvE_clEv
    @     0xaaaae53e70d0  _ZSt13__invoke_implIvRKZN6facade10Connection10ParseRedisEjEUlvE_JEET_St14__invoke_otherOT0_DpOT1_
    @     0xaaaae53e6420  _ZSt8__invokeIRKZN6facade10Connection10ParseRedisEjEUlvE_JEENSt15__invoke_resultIT_JDpT0_EE4typeEOS6_DpOS7_
    @     0xaaaae53e5664  _ZSt6invokeIRKZN6facade10Connection10ParseRedisEjEUlvE_JEENSt13invoke_resultIT_JDpT0_EE4typeEOS6_DpOS7_
    @     0xaaaae53e457c  _ZN4absl12lts_2025051219functional_internal12InvokeObjectIZN6facade10Connection10ParseRedisEjEUlvE_vJEEET0_NS1_7VoidPtrEDpNS1_8ForwardTIT1_E4typeE
    @     0xaaaae53eb6f4  absl::lts_20250512::FunctionRef<>::operator()()
    @     0xaaaae53d73c8  facade::Connection::DispatchSingle()
    @     0xaaaae53d7ae0  facade::Connection::ParseRedis()
[...]

PR #6574, opened 2026-02-09
fix: validate interleaved-keys argument count by stored step

vyavdoshenko:
MSETNX with an odd number of arguments bypassed argument validation and crashed in `Transaction::InitByKeys()` with a DCHECK failure: `full_args_.size() % key_index.step == 0u`.

The validation checked command names (`MSET`, `JSON.MSET`) but missed `MSETNX`, and any `--command_alias` clone would also bypass the check since `Clone()` changes `name()`.

Store the interleave step in `CommandId` instead of deriving it from the command name at runtime:
- Add `interleave_step_` field, propagated through `Clone()`
- Register MSET/MSETNX with step=2, JSON.MSET with step=3
- Use stored step in `Validate()`, `DetermineKeys()`, and COMMAND output
- Add regression test `MSetNxOddArgs`

Fixes: https://github.com/dragonflydb/dragonfly/issues/6573

### DIFF
--- a/src/server/command_registry.cc
+++ b/src/server/command_registry.cc
@@ -153,6 +153,11 @@ CommandId::CommandId(const char* name, uint32_t mask, int8_t arity, int8_t first
   else if (base::_in(name_, {"SPUBLISH", "SSUBSCRIBE", "SUNSUBSCRIBE"}))
     kind_pubsub_ = CO::PubSubKind::SHARDED;
   can_be_monitored_ = (opt_mask_ & CO::ADMIN) == 0 && name_ != "EXEC";
+
+  if (base::_in(name_, {"MSET", "MSETNX"}))
+    interleave_step_ = 2;
+  else if (name_ == "JSON.MSET")
+    interleave_step_ = 3;
 }
 
 CommandId::~CommandId() {
--- a/src/server/command_registry.h
+++ b/src/server/command_registry.h
@@ -34,8 +34,7 @@ enum CommandOpt : uint32_t {
   ADMIN = 1U << 7,  // implies NOSCRIPT,
   NOSCRIPT = 1U << 8,
   BLOCKING = 1U << 9,
-  HIDDEN = 1U << 10,            // does not show in COMMAND command output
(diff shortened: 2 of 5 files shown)


src/server/command_registry.cc
```diff
@@ -206,7 +206,7 @@ optional<facade::ErrorReply> CommandId::Validate(CmdArgList tail_args) const {
 
   if ((opt_mask() & CO::INTERLEAVED_KEYS)) {
     if ((name() == "JSON.MSET" && tail_args.size() % 3 != 0) ||
-        (name() == "MSET" && tail_args.size() % 2 != 0))
+        ((name() == "MSET" || name() == "MSETNX") && tail_args.size() % 2 != 0))
```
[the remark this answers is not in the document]
vyavdoshenko: Fixed

src/server/command_registry.cc
```diff
@@ -204,10 +205,8 @@ optional<facade::ErrorReply> CommandId::Validate(CmdArgList tail_args) const {
     return facade::ErrorReply{prefix + facade::WrongNumArgsError(name()), kSyntaxErrType};
   }
 
-  if ((opt_mask() & CO::INTERLEAVED_KEYS)) {
-    if ((name() == "JSON.MSET" && tail_args.size() % 3 != 0) ||
-        (name() == "MSET" && tail_args.size() % 2 != 0))
-      return facade::ErrorReply{facade::WrongNumArgsError(name()), kSyntaxErrType};
+  if (interleave_step_ && tail_args.size() % interleave_step_ != 0) {
```
romange, line 213: Ok, if you want to introduce `interleave_step_`, let's do this in less intrusive way.
Set it inside `CommandId` constructor as it can be derived from the name, and then you do not need `SetInterleavedStep`

If you introduce `interleave_step_` maybe you can erase the `INTERLEAVED_KEYS` mask as it's redundant.

vyavdoshenko: Fixed

dranikpg (approved):

PR #5597, opened 2025-08-01
feat(client): support client info command 

guozhihao-224:
#4860 support client info command

PR #6576, opened 2026-02-09
feat(server): add CLIENT INFO command

### LINKED ISSUE #4860: CLIENT INFO is not supported
We should support CLIENT INFO subcommand

romange:
- Add CLIENT INFO subcommand (Redis-compatible)
- Returns info about current connection (same format as CLIENT LIST)
- Add test in connection_test.py

Closes #4860

Supersedes #5597 (rebased on current main with updated API)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,8 +22,15 @@ endif
 # all the release builds are performed by gcc.
 LINKER_FLAGS += -static-libstdc++ -static-libgcc
 
+# Optional ASAN support: make ASAN=1 release
+ifdef ASAN
+SANITIZE_COMPILE_FLAGS = -fsanitize=address -Wno-maybe-uninitialized
+SANITIZE_LINK_FLAGS = -fsanitize=address
+endif
+
 HELIO_FLAGS = -DHELIO_RELEASE_FLAGS="-g" \
-			  -DCMAKE_EXE_LINKER_FLAGS="$(LINKER_FLAGS)" \
+			  -DCMAKE_CXX_FLAGS="$(SANITIZE_COMPILE_FLAGS)" \
+			  -DCMAKE_EXE_LINKER_FLAGS="$(LINKER_FLAGS) $(SANITIZE_LINK_FLAGS)" \
               -DBoost_USE_STATIC_LIBS=$(HELIO_USE_STATIC_LIBS) \
               -DOPENSSL_USE_STATIC_LIBS=$(HELIO_OPENSSL_USE_STATIC_LIBS) \
               -DENABLE_GIT_VERSION=$(HELIO_ENABLE_GIT_VERSION) \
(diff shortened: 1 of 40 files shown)


romange:
@kostasrim  PTAL

.vscode/settings.json
```diff
@@ -0,0 +1,70 @@
+{
```
kostasrim, line 1: Your .vscode settings are specific to you and only and IMO they should not be a part of dragonfly. 

I also just realized that there is `.vscode` folder. Should this really be checked in ?

And it also has bugs:

```
cat .vscode/c_cpp_properties.json 
{
  "configurations": [
    {
      "name": "Linux",
      "includePath": [
        "${default}"
      ],
      "cStandard": "c17",
      "cppStandard": "c++17",
      "intelliSenseMode": "${default}",
      "compileCommands": "${workspaceFolder}/build-dbg/compile_commands.json",
      "configurationProvider": "ms-vscode.cmake-tools"
    }
  ],
  "version": 4
}
```

This one says C++17 -- we updated to C++20

src/server/server_family.cc
```diff
@@ -391,6 +391,19 @@ void ClientGetName(CmdArgList args, CommandContext* cmd_cntx) {
...
 }
 
+void ClientInfo(CmdArgList args, CommandContext* cmd_cntx) {
+  if (!args.empty()) {
+    return cmd_cntx->SendError(facade::kSyntaxErr);
+  }
+  auto* conn = cmd_cntx->conn();
+  string info = conn->GetClientInfo();
+
+  // redis-py (5expects these fields. We append dummy values to keep the output parsable.
```
kostasrim, line 401: Yeah this is mostly dummy fields anyway. See also `GetClientInfo()`

```
   910   // The following are dummy fields and users should not rely on those unless                                                                                                                                                         
   911   // we decide to implement them.                                                                                                                                                                                                     
   912   // This is only done because the redis pyclient parser for the field "client-info"                                                                                                                                                  
   913   // for the command ACL LOG hardcodes the expected values. This behaviour does not                                                                                                                                                   
   914   // conform to the actual expected values, since it's missing half of them.                                                                                                                                                          
   915   // That is, even for redis-server, issuing an ACL LOG command via redis-cli and the pyclient                                                                                                                                        
   916   // will return different results! For example, the fields:                                                                                                                                                                          
   917   // addr=127.0.0.1:57275                                                                                                                                                                                                             
   918   // laddr=127.0.0.1:6379                                                                                                                                                                                                             
   919   // are missing from the pyclient.                                                                                                                                                                                                   
   920                                                                                                                                                                                                                                       
   921   absl::StrAppend(&before, " qbuf=0 ", "qbuf-free=0 ", "obl=0 ", "argv-mem=0 ");                                                                                                                                                      
   922   absl::StrAppend(&before, "oll=0 ", "omem=0 ", "tot-mem=0 ", "multi=0 ");                                                                                                                                                            
   923   absl::StrAppend(&before, "psub=0 ", "sub=0");  
```

PR #6587, opened 2026-02-10
refactor: store hnsw metadata in separate AUX field

BorysTheDev:
removed extra fields from HNSW metadata serialization
store HNSW metadata in a separate AUX field


BorysTheDev:
> Wouldn't it be just much simpler to send the full index definiton everywhere? Why do we need this json/etc?

we need it only in 0 thread. Anyway we can discuss improvements later

PR #6590, opened 2026-02-10
build: add optional ASAN support for release builds

vyavdoshenko:
Add optional AddressSanitizer support for release builds via `make ASAN=1 release`.

Passes `-fsanitize=address` to both compiler and linker via `CMAKE_CXX_FLAGS` and `CMAKE_EXE_LINKER_FLAGS` (using `HELIO_RELEASE_FLAGS` doesn't work because `internal.cmake` wraps it in a generator expression that breaks with multiple flags).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,8 +22,15 @@ endif
 # all the release builds are performed by gcc.
 LINKER_FLAGS += -static-libstdc++ -static-libgcc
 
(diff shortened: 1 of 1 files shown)


kostasrim:
+1 if we crash on update any time this should pinpoint the memory corruption

PR #6595, opened 2026-02-10
chore: simplify snapshotting serialization callback

romange:
Refactor RdbSerializer to internalize the flush threshold check and pass serialized data directly to the callback. This eliminates the circular dependency where the callback had to invoke 'FlushToSink' on the serializer.

- RdbSerializer::FlushFun now accepts 'std::string' directly.
- SliceSnapshot callback receives data push-style, removing the need for manual flushing.
- Unified data handling logic in 'SliceSnapshot::HandleFlushData'.

### DIFF
--- a/src/server/snapshot.cc
+++ b/src/server/snapshot.cc
@@ -430,7 +430,7 @@ void SliceSnapshot::HandleFlushData(std::string data) {
   seq_cond_.notify_all();
 
   if (!use_background_mode_) {
-    // serializer_->Flush can be quite slow for large values or due compression, therefore
+    // serializer_->Flush can be quite slow for large values or due to compression, therefore
     // we counter-balance CPU over-usage by sleeping.
     // We measure running_cycles before the preemption points, because they reset the counter.
     uint64_t sleep_usec = (running_cycles * 1000'000 / base::CycleClock::Frequency()) / 2;


romange:
augment review

romange:
augment review

src/server/rdb_save.cc
```diff
@@ -1775,9 +1781,11 @@ size_t RdbSerializer::GetTempBufferSize() const {
   return SerializerBase::GetTempBufferSize() + tmp_str_.size();
 }
 
-void RdbSerializer::FlushIfNeeded(SerializerBase::FlushState flush_state) {
-  if (flush_fun_) {
-    flush_fun_(SerializedLen(), flush_state);
+void RdbSerializer::PushToConsumerIfNeeded(SerializerBase::FlushState flush_state) {
```
kostasrim, line 1784: Yeah this is a good one, will setup for an intermediate layer for the future. But plz, do not add any spool or anything and let's discuss the snapshot issue first 😄

src/server/snapshot.cc
```diff
@@ -396,15 +396,11 @@ void SliceSnapshot::SerializeEntry(DbIndex db_indx, const PrimeKey& pk, const Pr
   }
 }
 
-size_t SliceSnapshot::FlushSerialized(SerializerBase::FlushState flush_state) {
-  io::StringFile sfile;
```
kostasrim, line 400: Too many times I wondered what the use case of StringFile is when there was not actual write to a file (merely an IO interface writting to a plain in-memory string!)

Issue #6602, opened 2026-02-11
Dragonfly. crashes on fuzzer input


vyavdoshenko:
`nc localhost 6379 < crash.txt`

[crash.txt](https://github.com/user-attachments/files/25231219/crash.txt)

```
F20260211 12:12:28.048491 52468 script_mgr.cc:68] Check failed: sha.size() == size() (9 vs. 40) 
*** Check failure stack trace: ***
    @     0x5555574c86d5  google::LogMessage::Fail()
    @     0x5555574c861b  google::LogMessage::SendToLog()
    @     0x5555574c7dce  google::LogMessage::Flush()
    @     0x5555574cbd7a  google::LogMessageFatal::~LogMessageFatal()
    @     0x555556d082ae  dfly::ScriptMgr::ScriptKey::ScriptKey()
    @     0x555556d090fc  dfly::ScriptMgr::ConfigCmd()
    @     0x555556d08863  dfly::ScriptMgr::Run()
    @     0x5555568c15d8  dfly::ServerFamily::Script()
    @     0x55555692c87e  _ZZN4dfly11HandlerFuncEPNS_12ServerFamilyEMS0_FvN4absl12lts_202505124SpanIKSt17basic_string_viewIcSt11char_traitsIcEEEEPNS_14CommandContextEEENKUlSA_SC_E_clESA_SC_
    @     0x555556960e0d  _ZN3fu27abi_4006detail10invocation6invokeIRKZN4dfly11HandlerFuncEPNS4_12ServerFamilyEMS5_FvN4absl12lts_202505124SpanIKSt17basic_string_viewIcSt11char_traitsIcEEEEPNS4_14CommandContextEEEUlSF_SH_E_JSF_SH_EEEDTclcl7forwardIT_Efp_Espcl7forwardIT0_Efp0_EEEOSN_DpOSO_
    @     0x55555695e05a  _ZN3fu27abi_4006detail12type_erasure16invocation_table14function_traitIKFvN4absl12lts_202505124SpanIKSt17basic_string_viewIcSt11char_traitsIcEEEEPN4dfly14CommandContextEEE16internal_invokerINS2_3boxILb1EZNSE_11HandlerFuncEPNSE_12ServerFamilyEMSL_FvSD_SG_EEUlSD_SG_E_SaISP_EEELb0EE6invokeEPKNS2_13data_accessorEmSD_SG_
[...]

PR #6603, opened 2026-02-11
fix: reject SCRIPT FLAGS with invalid SHA length instead of crashing

vyavdoshenko:
Validate SHA length in `SCRIPT FLAGS` before constructing `ScriptKey`.
Previously, a non-40-character SHA triggered a `DCHECK` crash.

Fixes: #6602

### DIFF
--- a/src/server/multi_test.cc
+++ b/src/server/multi_test.cc
@@ -893,6 +893,10 @@ TEST_F(MultiTest, ScriptFlagsCommand) {
   }
 }
 
+TEST_F(MultiTest, ScriptFlagsInvalidSha) {
+  EXPECT_THAT(Run({"script", "flags", "short", "allow-undeclared-keys"}), ErrArg(""));
+}
+
 TEST_F(MultiTest, ScriptFlagsEmbedded) {
   const char* s1 = R"(
   --!df flags=allow-undeclared-keys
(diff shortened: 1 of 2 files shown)


src/server/multi_test.cc
```diff
@@ -893,6 +893,10 @@ TEST_F(MultiTest, ScriptFlagsCommand) {
   }
 }
 
+TEST_F(MultiTest, ScriptFlagsInvalidSha) {
+  EXPECT_THAT(Run({"script", "flags", "short", "allow-undeclared-keys"}), ErrArg(""));
```
kostasrim, line 897: TIP for future PR's: you can use a single string: `Run("script flags short allow-undeclared-keys")`

src/server/script_mgr.cc
```diff
@@ -171,8 +171,13 @@ void ScriptMgr::LoadCmd(CmdArgList args, Transaction* tx, SinkReplyBuilder* buil
 }
 
 void ScriptMgr::ConfigCmd(CmdArgList args, Transaction* tx, SinkReplyBuilder* builder) {
+  string_view sha = ArgS(args, 1);
+  if (sha.size() != ScriptKey{}.size()) {
```
kostasrim, line 175: nice catch! we assumed sha size 😄

kostasrim (approved):

PR #6606, opened 2026-02-11
feat: add HNSW index deserialization

BorysTheDev:
Added deserialization for HNSW index.
Added a test for serialization/deserialization scenario


src/core/search/hnsw_index.cc
```diff
@@ -283,4 +438,32 @@ std::vector<HnswNodeData> HnswVectorIndex::GetNodesRange(size_t start, size_t en
...
+void HnswVectorIndex::RestoreFromNodes(const std::vector<HnswNodeData>& nodes,
+                                       const HnswIndexMetadata& metadata) {
+  adapter_->RestoreFromNodes(nodes, metadata);
+}
+
+bool HnswVectorIndex::UpdateVectorData(GlobalDocId id, const DocumentAccessor& doc,
+                                       std::string_view field) {
+  auto vector_ptr = doc.GetVector(field, dim_);
+  if (!vector_ptr) {
+    // Document doesn't have the vector field - mark node as deleted to prevent
```
mkaruza, line 450: Nullopt is returned if vector fields is not matched. For documents that doesn't have vector fields we check against nullptr

Issue #6614, opened 2026-02-12
Dragonfly crashes in dragonfly_connection.cc


vyavdoshenko:
[crash-000000.tar.gz](https://github.com/user-attachments/files/25271112/crash-000000.tar.gz)

Start dragonfly:
```
./dragonfly --dbfilename= --logtostderr --proactor_threads=1 --dbfilename="" --max_bulk_len=1048576 2>&1
```

```
tar xzf crash-000000.tar.gz
cd crash-000000
python3 replay_crash.py crashes 000000
```

```
F20260212 21:07:48.394249 193178 dragonfly_connection.cc:1917] Check failed: cc_ 
*** Check failure stack trace: ***
    @     0x589392b62245  google::LogMessage::Fail()
    @     0x589392b6218b  google::LogMessage::SendToLog()
    @     0x589392b6193e  google::LogMessage::Flush()
    @     0x589392b658ea  google::LogMessageFatal::~LogMessageFatal()
    @     0x589392897d2f  facade::Connection::SendAsync()
    @     0x589392897509  facade::Connection::SendMonitorMessageAsync()
    @     0x58939223a583  _ZZN4dfly12_GLOBAL__N_115DispatchMonitorEPNS_17ConnectionContextEPKNS_9CommandIdEN4absl12lts_202505124SpanIKSt17basic_string_viewIcSt11char_traitsIcEEEEENKUljPN4util3fb212ProactorBaseEE_clEjSI_
    @     0x589392254430  _ZZN4util12ProactorPool13DispatchBriefIZN4dfly12_GLOBAL__N_115DispatchMonitorEPNS2_17ConnectionContextEPKNS2_9CommandIdEN4absl12lts_202505124SpanIKSt17basic_string_viewIcSt11char_traitsIcEEEEEUljPNS_3fb212ProactorBaseEE_Li0EEEvOT_ENUlvE_clEv
[...]

PR #6615, opened 2026-02-12
fix: make ChangeMonitor idempotent to prevent double-add in MULTI/EXEC

vyavdoshenko:
Make `ChangeMonitor` idempotent — skip if already in the desired state.

MONITOR queued multiple times inside MULTI/EXEC adds the connection to the monitors list more than once. `Remove` only erases the first occurrence, leaving a dangling pointer that crashes in `SendMonitorMessageAsync` after `cc_.reset()`.

Fixes: https://github.com/dragonflydb/dragonfly/issues/6614

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1879,10 +1879,6 @@ void Connection::SendPubMessageAsync(PubMessage msg) {
 }
 
 void Connection::SendMonitorMessageAsync(string msg) {
-  // DispatchMonitor uses async DispatchBrief, so by the time this callback
-  // runs the connection may already be closing.
-  if (!cc_ || !listener())
-    return;
   SendAsync({MonitorMessage{std::move(msg)}});
 }
 
--- a/src/server/conn_context.cc
+++ b/src/server/conn_context.cc
@@ -69,6 +69,10 @@ ConnectionContext::ConnectionContext(facade::Connection* owner, acl::UserCredent
 }
 
 void ConnectionContext::ChangeMonitor(bool start) {
+  // Ensure idempotency: MONITOR may be queued multiple times inside MULTI/EXEC.
+  if (start == monitor)
+    return;
+
   // This will either remove or register a new connection
   // at the "top level" thread --> ServerState context
   // note that we are registering/removing this connection to the thread at which at run
--- a/src/server/dragonfly_test.cc
+++ b/src/server/dragonfly_test.cc
@@ -1066,30 +1066,4 @@ TEST_F(DflyCommandAliasTest, AliasesShareHistogramPtr) {
   EXPECT_EQ(command_histograms.at("ping"), command_histograms.at("___ping"));
 }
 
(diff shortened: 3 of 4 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -1879,6 +1879,10 @@ void Connection::SendPubMessageAsync(PubMessage msg) {
 }
 
 void Connection::SendMonitorMessageAsync(string msg) {
+  // DispatchMonitor uses async DispatchBrief, so by the time this callback
```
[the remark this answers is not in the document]
romange: are you sure this test catches something? usually unit tests can not cover facade bugs and we use pytests for this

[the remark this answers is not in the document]
vyavdoshenko: Yes, this test catches exactly the same crash. I always check the test before changes.
btw, it's easy to check on main:
```
git checkout bobik/connection_crash -- src/server/dragonfly_test.cc
```

src/facade/dragonfly_connection.cc
```diff
@@ -1879,6 +1879,10 @@ void Connection::SendPubMessageAsync(PubMessage msg) {
...
+  // runs the connection may already be closing.
+  if (!cc_ || !listener())
+    return;
```
dranikpg: This fixes the symptoms, but doesn't explain the bug. That is not supposed to happend and likely changing the connection code broke it. `ConnectionFlow` calls `service_->OnConnectionClose` which calls `DeactivateMonitoring` which removes the connection from the monitor repo - and waits for it to complete on all threads. Only then it calls cc_.reset()

src/facade/dragonfly_connection.cc
```diff
@@ -1879,6 +1879,10 @@ void Connection::SendPubMessageAsync(PubMessage msg) {
 }
 
 void Connection::SendMonitorMessageAsync(string msg) {
+  // DispatchMonitor uses async DispatchBrief, so by the time this callback
+  // runs the connection may already be closing.
+  if (!cc_ || !listener())
```
romange: why `listener()` check is needed? connections always have a listener

vyavdoshenko: The test uncovered that listener() returns null and crashes later.

romange: it's because our c++ testing framework does not fully cover facade code.

src/facade/dragonfly_connection.cc
romange: +1 to @dranikpg .

src/facade/dragonfly_connection.cc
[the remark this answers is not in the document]
vyavdoshenko: Test was replaced with python test

src/facade/dragonfly_connection.cc
vyavdoshenko: Fixed.

src/facade/dragonfly_connection.cc
vyavdoshenko: I got it, this test was replaced with python test

PR #6616, opened 2026-02-13
make migration more stable

BorysTheDev:
To finalize the migration on the master, we do the following:
1) PAUSE
2) send LSN for every shard connection 
3) send ACK for main connection and wait for OK to finalize the migration

If we send LSN and after LSN we send any command via journal (pause prevents new command, but we can have some that haven't finished yet), it prevents finalization. 
So to make migration more stable, we can wait until all data is transferred.

**Summary:** Makes slot migration finalization more stable by draining any in-flight journal writes before emitting the finalize (LSN) marker.
**Why:** Avoids the incoming side seeing new journal traffic after the finalize marker, which can prevent/undo finalization.

### DIFF
--- a/src/server/journal/streamer.cc
+++ b/src/server/journal/streamer.cc
@@ -166,7 +166,7 @@ bool JournalStreamer::Cancel() {
     res = true;
   }
   StopStalledDataWriterFiber();
-  WaitForInflightToComplete();
+  WaitForInflightToComplete(false);
   return res;
 }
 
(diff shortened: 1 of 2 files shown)


romange:
Is it possible to reproduce in a test?

Roman Gershman
CTO
---

BorysTheDev:
@romange It reproduces on our tests with my flags for ARM

src/server/journal/streamer.h
```diff
@@ -61,7 +59,7 @@ class JournalStreamer : public journal::JournalConsumerInterface {
     return cntx_->IsRunning();
   }
 
-  void WaitForInflightToComplete();
+  void WaitForInflightToComplete(bool with_timeout = false);
```
romange: do not use default args to save changing few lines of code.

romange:
Please expand the PR description as it's a bit hard to understand the scenario and how "with_timeout" solves it.

src/server/journal/streamer.cc
```diff
@@ -513,6 +518,10 @@ void RestoreStreamer::SendFinalize(long attempt) {
           << ", throttle usec on db update: " << stats_.throttle_usec_on_db_update
           << ", iter_skips: " << stats_.iter_skips;
 
+  // Drain all pending journal data before sending the finalize marker.
+  // At this point client pause is active, so no new entries can arrive.
+  WaitForInflightToComplete(true);
```
romange, line 530: to the best of my knowledge journal updates are serialized. So why this barrier is needed? The pr description is not clear

BorysTheDev: I've updated description

romange: much better. Btw, when I submit a PR, I look at augment/copilot summaries and often enrich my original PR description.

PR #6617, opened 2026-02-13
feat(dash): implement Segment merge algorithm

kostasrim:
This PR implements Segment merging for dash table. It does not implement Directory merging (the inverse of Grow() operation). Interestingly the dash table does not mention segment merging but the original paper for extendible hashing does: https://dl.acm.org/doi/pdf/10.1145/320083.320092 

When a segment splits:

* One segments becomes two buddies
* Local depth increases by 1
* Items redistribute across the new segments

When buddies merge:
* Two buddies become one segment
* Depth decreases by 1
* Items get merged into one segment

The algorithm in pseudocode:

```
for segment in segments:
   Buddy = FindBuddy(segment)    
   if(!buddy) continue
   size = buddy->elements() + segment->elements()
   if(size > threshold * segment->capacity()) {     
     continue
   }
   Merge(segment, buddy)
}

Merge:
  Move all elements from buddy to segment
  Update buddy to point to segment
  Deallocate buddy
  Decrease depth
```

* add command debug table_gc [threshold]
* add test
* implement Merge() — inverse of Split() in dash table

First steps to  #7

### DIFF
--- a/src/core/dash.h
+++ b/src/core/dash.h
@@ -182,6 +182,59 @@ class DashTable : public detail::DashTableBase {
     return segment_[segment_id];
   }
 
+  unsigned FindBuddyId(unsigned segment_id) {
+    auto* seg = GetSegment(segment_id);
+    uint8_t depth = seg->local_depth();
+
+    assert(depth != 0);
+
+    const size_t bit_pos = global_depth_ - depth;
+    const size_t buddy_idx = segment_id ^ (1 << bit_pos);
+    assert(buddy_idx < segment_.size());
+
+    auto* buddy = GetSegment(buddy_idx);
+    // For that case, buddy is as if it were the segment did not split at depth - 1
+    if (buddy->local_depth() != depth) {
+      return segment_id;
+    }
+
+    return buddy_idx;
+  }
+
+  // Merges two segments
+  void Merge(unsigned keep_id, unsigned buddy_id) {
+    auto* keep = GetSegment(keep_id);
+    auto* buddy = GetSegment(buddy_id);
+    assert((keep->local_depth() == buddy->local_depth()));
+
+    // Move all items from buddy to keep
+    buddy->TraverseAll([&](const auto& it) {
+      uint64_t hash = DoHash(buddy->Key(it.index, it.slot));
+      auto& src_bucket = buddy->GetBucket(it.index);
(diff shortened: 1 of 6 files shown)


kostasrim:
Not fully ready, this is a sketch :)

kostasrim:
@romange feel free if you wish to look or comment -- I still probably need to polish more + add more tests

romange:
We never write a data-structure without tests. The first PR should be only around dash tablei, i.e. under src/core only and include tests in dash_test.

romange:
What I mean is -  you should narrow the scope to dash code only in this PR. Server code should be out of scope.

src/core/dash.h
```diff
@@ -182,6 +182,59 @@ class DashTable : public detail::DashTableBase {
...
+  void Merge(unsigned keep_id, unsigned buddy_id) {
+    auto* keep = GetSegment(keep_id);
+    auto* buddy = GetSegment(buddy_id);
+    assert((keep->local_depth() == buddy->local_depth()));
+
+    // Move all items from buddy to keep
+    buddy->TraverseAll([&](const auto& it) {
+      uint64_t hash = DoHash(buddy->Key(it.index, it.slot));
+      auto& src_bucket = buddy->GetBucket(it.index);
+      keep->InsertUniq(std::move(src_bucket.key[it.slot]), std::move(src_bucket.value[it.slot]),
```
dranikpg, line 214: I haven't checked, but if segment merging is compatible with snapshotting, it should make sure both segments were serialized or bump up their versions, otherwise it could lead to missed entries

romange: we won't merge during snapshotting. There is no reason to do this operation during this time.

kostasrim: Merging should be stable just like split is. IMO eventually I would assume we won't even have a command to trigger this flow, the original paper triggers it per delete if I recall correctly (checks if a segment and it's buddy is underutilized and does the merge).

Your comment is very valid Vlad -- just not a priority for now

kostasrim:
> We never write a data-structure without tests. The first PR should be only around dash tablei, i.e. under src/core only and include tests in dash_test.

Yes it's incomplete -- I am working on it, this was the prototyped I wrapped on Friday 😄

PR #6621, opened 2026-02-13
fix: string corruption during rdb load

romange:
When ToSV called we can not reference previous string_view blobs returned by it.
We broke this rule with sets and this PR fixes it.

1. We explicitly pass a scratch buffer now so it will be visible when we might reuse the same buffer several times.
2. Adding the test exposing the bug

### DIFF
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -220,6 +220,8 @@ class RdbLoaderBase::OpaqueObjLoader {
   }
 
  private:
+  using ScratchBuf = base::PODArray<char>;
+
   void CreateSet(const LoadTrace* ltrace);
   void CreateHMap(const LoadTrace* ltrace);
   void CreateList(const LoadTrace* ltrace);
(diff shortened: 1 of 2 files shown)


romange:
augment review

src/server/rdb_load.cc
```diff
@@ -779,23 +779,25 @@ void RdbLoaderBase::OpaqueObjLoader::CreateStream(const LoadTrace* ltrace) {
...
-    cg_id.ms = cg.ms;
-    cg_id.seq = cg.seq;
-
-    uint64_t entries_read = cg.entries_read;
-    if (rdb_type_ == RDB_TYPE_STREAM_LISTPACKS) {
-      entries_read = streamEstimateDistanceFromFirstEverEntry(s, &cg_id);
-    }
+    streamCG* cgroup = nullptr;
+    {
+      string_view cgname = ToSV(cg.name, &buf1_);
```
romange, line 784: just indentation.

PR #6622, opened 2026-02-14
chore: journal clean-ups

romange:
1. Add a helper bash script that runs master/replica locally.
2. Refactor: Simplify journal interface by moving methods to free functions

Convert journal consumer methods from member functions to free functions
in the journal namespace, reducing coupling and simplifying the API.

Key changes:
- Move GetLsn(), RecordEntry(), SetFlushMode() to journal:: namespace
- Rename RegisterOnChange/UnregisterOnChange to RegisterConsumer/UnregisterConsumer
- Rename JournalFlushGuard to DisableFlushGuard for clarity
- Update DisableFlushGuard to use bool flag instead of Journal* pointer
- Remove journal parameter from LogJournalOnShard (uses thread-local state)

This makes the journal API more consistent and reduces the need to pass
journal pointers throughout the codebase.

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -3799,9 +3799,6 @@ void ServerFamily::ReplicaOfNoOne(SinkReplyBuilder* builder) {
       // We enable it on all threads, even non-shard threads as we mirror the I/O flows.
       //
       // TODO(roman): it seems wrong to me that journal LSN is tracked per io thread.
-      // What happens if the replica has less threads than the master, for example.
-      // What happens when the second replica fails over to the first one and both
-      // have less shards than I/O threads? Seems all sort of trouble to me.
       shard_set->pool()->AwaitBrief([this, repl_ptr](auto index, auto*) {
         auto flow_map = repl_ptr->GetFlowMapAtIndex(index);
         size_t rec_executed = repl_ptr->GetRecCountExecutedPerShard(flow_map);


dranikpg (dismissed):

src/server/server_family.cc
```diff
@@ -3796,7 +3796,13 @@ void ServerFamily::ReplicaOfNoOne(SinkReplyBuilder* builder) {
     auto repl_ptr = replica_;
     if (absl::GetFlag(FLAGS_replicaof_no_one_start_journal)) {
       // Start journal and keep offsets.
-      shard_set->pool()->AwaitFiberOnAll([this, repl_ptr](auto index, auto*) {
+      // We enable it on all threads, even non-shard threads as we mirror the I/O flows.
```
romange, line 3799: @kostasrim  please review -  I have gaps around this part.

src/server/server_family.cc
```diff
@@ -3796,7 +3796,13 @@ void ServerFamily::ReplicaOfNoOne(SinkReplyBuilder* builder) {
...
+      //
+      // TODO(roman): it seems wrong to me that journal LSN is tracked per io thread.
```
kostasrim, line 3801: What was the reason of this change can you give some background ? Where did we crash ? 

It's fine in principle to start the journal on all threads -- the journal for IO threads that are *not* shard threads will be empty.

We always replicate with `X` flows == `number of shards on master`. 

Lastly, your comment is invalid:

> // What happens if the replica has less threads than the master, for example.

It's fine, we will `full sync` instead of doing a `partial` sync. We have tests for all of these cases

 > / What happens when the second replica fails over to the first one and both
      // have less shards than I/O threads? 

It's ok again, `full sync` instead of `partial`

kostasrim: https://github.com/dragonflydb/dragonfly/commit/c4d797f9c371c2ea9a1258bff27e7c2f6f61da6d#diff-8174e8eae32841a10b583a676b1b8ec068e7b3f2aa3cf61d2e952553ab343875R583

src/server/server_family.cc
kostasrim: https://github.com/dragonflydb/dragonfly/issues/6623 is this what you are trying to fix ?

src/server/server_family.cc
romange: The reason - clarity and readability. There is no issue. I want to remove Journal wrapper and eventually keep journal_slice as it's redundant and does not contribute to anything. I want to remove the journal object as it does not have any real state and just complicates APIs.

Also I would like to add comments explaining the threading model of journal_slice.

Is it the correct to say that journal_slice is activated per **shard** thread on master
but it's per **I/O thread** on replica and is used to mirror changes on replica frontend coming from the master?

kostasrim: > Is it the correct to say that journal_slice is activated per shard thread on master
but it's per I/O thread on replica and is used to mirror changes on replica frontend coming from the master?

No. We do not start the journal on the replica unless we start the takeover as we do below. The node that is taking over shall always `start the journal on all of its shard threads` as we do below (check `Partition` in replica.cc)

To clarify all cases: for master with `m` shards and replica with `r` shards there are the following cases:

### Case 1: `m < r` 

master has more shards. 
For `m = 2` and `r = 1` the partition map for `r` will be of size `1`. There will be `2 flows` from master to replica. After the takeover lsn_of_each_master_flow == `sum(mapped_flows_lsn's)`. So if lsn of flow a is 10 and lsn of flow b is 20, then `after takeover` the journal starts on `one shard thread`(since `r=1`) at `lsn 30`. 

### Case 2: `m > r`. 
For example `m=1` and `r=2`.  Parition map of `r` will be of size `2`. However, there is `one flow` from master to `replica`. After takeover, we will have 2 journals (1 per shard thread since `r=2`). Imagine lsn for flow of master was 22, now we start one journal at 22 and another at 1 (since new master has more shards).

### Case 3: `m == r`
This is the easiest one, each shard in `r` contains `1 flow`. After takeover, new master has `r` journals all initialized with their respective master lsn. So for example for `m=1`, `r=1`. lsn_of_each_master_flow == journal offset of newly promoted master

src/server/server_family.cc
```diff
@@ -3796,7 +3796,13 @@ void ServerFamily::ReplicaOfNoOne(SinkReplyBuilder* builder) {
...
+      // What happens if the replica has less threads than the master, for example.
+      // What happens when the second replica fails over to the first one and both
+      // have less shards than I/O threads? Seems all sort of trouble to me.
```
kostasrim:
```suggestion
```

kostasrim: This is not needed, semantics are clear on how the new elected node start its journal. Partial sync is also "shard map" aware so it won't crash or attempt the wrong thing.

If something is unclear -- I am happy to jump on a call or write here and explain

src/server/server_family.cc
romange: I am also asking about something else. Suppose servers have 8 threads: 0..3 are shard threads, 4..7 are i/o threads.
On a master we always run journaling on shard threads only (0-3), but here we  initialise a journal on all threads. So it's not clear to me what journal on thread 4 is managing. Moreover, what will its state will be once the replica takes over. Ideally, I would like to write this explanation here

romange: And BTW your cases titles are opposite of what you wrote.

kostasrim: > And BTW your cases titles are opposite of what you wrote.

uh sorry for that, I moved it around because I was trying to add split it :/ I messed it up

> I am also asking about something else. Suppose servers have 8 threads: 0..3 are shard threads, 4..7 are i/o threads.
On a master we always run journaling on shard threads only (0-3), but here we initialise a journal on all threads. So it's not clear to me what journal on thread 4 is managing. Moreover, what will its state will be once the replica takes over. Ideally, I would like to write this explanation here

It should be on `shard threads` only. https://github.com/dragonflydb/dragonfly/pull/6652/changes 

Tests pass.

PR #6636, opened 2026-02-16
fix(facade): Fix missing DFLY_USE_SSL defines

dranikpg:
disabling SSL breaks the build


BorysTheDev:
@dranikpg is  DF_USE_SSL a typo?

dranikpg:
Ah, yes, this was a typo

PR #6642, opened 2026-02-16
chore: add logs to rdb_load

kostasrim:
* add logs to rdb load when an invalid rdb type is read
* fix: journal should only start on shard threads yet we started it in all threads in takeover and no-one flow

### DIFF
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -2292,7 +2292,7 @@ error_code RdbLoader::Load(io::Source* src) {
     if (!rdbIsObjectTypeDF(type)) {
       LOG(ERROR) << "Unrecognized rdb object type: " << type;
       LOG(ERROR) << "Last iteration: ";
-      LOG(ERROR) << "key loaded: " << last_key_loaded_;
+      LOG(ERROR) << "key loaded: " << absl::CHexEscape(last_key_loaded_);
       LOG(ERROR) << "pending_read_.remaining: " << pending_read_.remaining
                  << "\npending_read_.reserve: " << pending_read_.reserve;
       // In case we encounter an error, it might worth peeking the InputBuffer()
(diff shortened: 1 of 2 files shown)


src/server/rdb_load.cc
```diff
@@ -545,6 +545,7 @@ void RdbLoaderBase::OpaqueObjLoader::CreateList(const LoadTrace* ltrace) {
   QList* qlv2 = nullptr;
   if (config_.append) {
     if (pv_->ObjType() != OBJ_LIST) {
+      LOG(ERROR) << "List operation on wrong object type. Expected list got: " << pv_->ObjType();
```
kostasrim: In the snapshot they are not using lists at all. I highly doubt it's related to this flow

src/server/rdb_load.cc
```diff
@@ -2284,6 +2285,12 @@ error_code RdbLoader::Load(io::Source* src) {
     }
 
     if (!rdbIsObjectTypeDF(type)) {
+      LOG(ERROR) << "Unrecognized rdb object type " << type;
```
kostasrim: The second place we return `invalid_rdb_type` is when we call `EnsureObjEncoding`. But we also `LOG an error message` which we did not see in the regression/logs. So I highly doubt it's that path that returned `invalid_rdb_type`.

src/server/rdb_load.cc
```diff
@@ -2284,6 +2285,12 @@ error_code RdbLoader::Load(io::Source* src) {
...
+      // Ok let's check the key:
+      auto bytes = mem_buf_->InputBuffer();
+      // Obj dump - bytes
+      std::string_view out{reinterpret_cast<const char*>(bytes.data()), bytes.size()};
```
kostasrim: I would like to know the buffer contents, I suspect we fail here -- I am not sure what we can do more since we `Load` the key value after we the object type

PR #6645, opened 2026-02-16
fix(server): Wait for index load on replication

dranikpg:
I'm afraid not waiting until index construction after replication will be unexpected for operations like takeover that should happen transparently, so block until it's done

### DIFF
--- a/src/server/search/doc_index.cc
+++ b/src/server/search/doc_index.cc
@@ -904,7 +904,7 @@ void ShardDocIndices::BlockUntilConstructionEnd() {
       indexing |= ptr->GetInfo().indexing;
 
     if (indexing)
-      util::ThisFiber::SleepFor(1ms);
+      util::ThisFiber::SleepFor(5ms);
   } while (indexing);
 }
 


src/server/rdb_load.cc
```diff
@@ -3068,16 +3070,24 @@ void RdbLoader::PerformPostLoad(Service* service, bool is_error) {
...
-                   DbContext{&namespaces->GetDefaultNamespace(), 0, GetCurrentTimeMs()}};
-    es->search_indices()->RebuildAllIndices(op_args);
-  });
+  if (had_indices) {
+    shard_set->RunBriefInParallel([](EngineShard* es) {
+      OpArgs op_args{es, nullptr,
+                     DbContext{&namespaces->GetDefaultNamespace(), 0, GetCurrentTimeMs()}};
+      es->search_indices()->RebuildAllIndices(op_args);
+    });
+  }
```
dranikpg: One issue in general is that index building is not cancelable

dranikpg:
augment review

src/server/search/doc_index.cc
```diff
@@ -798,6 +799,18 @@ void ShardDocIndices::RebuildAllIndices(const OpArgs& op_args) {
     ptr->Rebuild(op_args, &local_mr_);
 }
 
+void ShardDocIndices::BlockUntilConstructionEnd() {
```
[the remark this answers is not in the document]
dranikpg: Cancel() clears the builder, so GetInfo().indexing will become false. The bigger paint point is that we don't have build cancelling for failed replication attempts/stopping snapshot loading, so

src/server/search/doc_index.h
```diff
@@ -388,6 +388,9 @@ class ShardDocIndices {
   // Rebuild all indices
   void RebuildAllIndices(const OpArgs& op_args);
 
+  // Block until construction of all indices finishes
+  void BlockUntilConstructionEnd();
```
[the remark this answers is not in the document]
dranikpg: True!!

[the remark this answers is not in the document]
romange: what happens if we have a large index and sigint is sent. Will index building break?

src/server/rdb_load.cc
```diff
@@ -3078,6 +3079,10 @@ void RdbLoader::PerformPostLoad(Service* service, bool is_error) {
   for (auto& syn_cmd : synonym_cmds) {
     LoadSearchCommandFromAux(service, std::move(syn_cmd), "FT.SYNUPDATE", "synonym definition");
   }
+
+  // Wait until index building ends
+  shard_set->RunBlockingInParallel(
+      [](EngineShard* es) { es->search_indices()->BlockUntilConstructionEnd(); });
```
BorysTheDev, line 3195: are you sure that we want to block here?

BorysTheDev: we can't switch on stable_sync until finish this method

src/server/search/doc_index.h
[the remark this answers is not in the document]
dranikpg: Yes, because SearchFamily::Shutdown calls DropAllIndices which makes this function finish

src/server/rdb_load.cc
dranikpg: Yes, because once you switch to full sync you can take over. I assume this process is supposed to be transparent. If you don't block here, you indices won't be filled with data / in sync, so ft.search calls will suddenly return empty results after takeover

dranikpg: in the future we can start the index building earlier and this critical section will be smaller

PR #6647, opened 2026-02-16
Load hnsw index

BorysTheDev:
Summary: This PR adds a restore path to load serialized HNSW vector index graphs from Dragonfly snapshots for cases when the number of shards is the same during serialization and deserializationinstead of always rebuilding them.

Changes:

- RDB load now parses RDB_OPCODE_VECTOR_INDEX and conditionally restores global HNSW graph nodes when FLAGS_deserialize_hnsw_index is enabled and shard counts match.
- Loads and applies per-shard key→DocId mappings (RDB_OPCODE_SHARD_DOC_INDEX) so restored HNSW graph GlobalDocIds stay aligned with rebuilt search indices.
- Extends search index rebuild pipeline to preserve restored DocIds and repopulate vector payloads via UpdateVectorData (separating regular index rebuild from vector restoration).
- Captures HNSW metadata via AUX fields and threads it into the restore flow; adds helper methods to restore/skip vector index data.
- Updates replica loading to set snapshot shard-count, and expands tests to validate DF save/load HNSW restoration across multiple document counts.

Technical Notes: Restoration is gated by FLAGS_deserialize_hnsw_index and requires identical shard topology because GlobalDocId encodes shard id; otherwise the loader skips serialized graph data and rebuilds indices from documents.

### DIFF
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -2230,7 +2230,9 @@ error_code RdbLoader::Load(io::Source* src) {
       uint64_t elements_number;
       SET_OR_RETURN(LoadLen(nullptr), elements_number);
 
-      // Only restore if the flag is enabled and shard count matches (GlobalDocId encodes shard_id)
+      // We can keep the same internal ids if shards count is the same and we can reuse the
+      // data-structures as is, otherwise we must rebuild from scratch
+      // TODO add ability to restore index with different shard count by remapping internal ids
       const bool should_restore = deserialize_hnsw_index_ && shard_count_ == shard_set->size();
 
       if (should_restore) {


src/server/rdb_load.h
```diff
@@ -341,34 +342,44 @@ class RdbLoader : protected RdbLoaderBase {
...
     std::string index_name;
     std::vector<std::pair<std::string, search::DocId>> mappings;
   };
-  std::vector<PendingIndexMapping> pending_index_mappings_;
+
+  // Get and clear pending index mappings collected from all RdbLoader instances
+  static absl::flat_hash_map<uint32_t, std::vector<PendingIndexMapping>> TakePendingIndexMappings();
+
+  Service* service_;
+  static std::vector<std::string> pending_synonym_cmds_;
```
romange, line 365: I know you have not added it but why it is static @BorysTheDev

src/server/search/index_builder.cc
```diff
@@ -51,8 +55,24 @@ void IndexBuilder::CursorLoop(dfly::DbTable* table, DbContext db_cntx) {
...
+        auto accessor = GetAccessor(db_cntx, pv);
+        if (!index_->indices_->Add(*doc_id, *accessor)) {
+          LOG(WARNING) << "Failed to restore index entry for key: " << key
+                       << ", removing from key index";
+          index_->key_index_.Remove(*doc_id);
+        }
+      }
+    } else {
       index_->AddDoc(key, db_cntx, pv);
+    }
```
dranikpg, line 76: TODO: make it a paramater of SharDocIndex::AddDoc()

src/server/rdb_load.cc
```diff
@@ -2228,34 +2230,30 @@ error_code RdbLoader::Load(io::Source* src) {
...
-
-        for (uint32_t lvl = 0; lvl <= level; ++lvl) {
-          uint32_t links_num;
-          SET_OR_RETURN(FetchInt<uint32_t>(), links_num);
-          for (uint32_t i = 0; i < links_num; ++i) {
-            [[maybe_unused]] uint32_t link;
-            SET_OR_RETURN(FetchInt<uint32_t>(), link);
-          }
-        }
+      // Only restore if the flag is enabled and shard count matches (GlobalDocId encodes shard_id)
```
romange: what do we do if `should_restore` is false? remind me what do we restore if it is true?

src/server/search/index_builder.cc
```diff
@@ -70,6 +86,13 @@ void IndexBuilder::VectorLoop(dfly::DbTable* table, DbContext db_cntx) {
   if (!any_vector || !state_.IsRunning())
     return;
 
+  // If any HNSW index was restored from RDB, use UpdateVectorData instead of Add.
+  if (is_restored_) {
+    OpArgs op_args{EngineShard::tlocal(), nullptr, db_cntx};
+    index_->RestoreGlobalVectorIndices(index_->base_->name, op_args);
```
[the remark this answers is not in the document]
dranikpg: True, TODO: Add support for concurrent modifications

src/server/rdb_load.cc
```diff
@@ -2271,15 +2269,21 @@ error_code RdbLoader::Load(io::Source* src) {
         pim.mappings.emplace_back(std::move(key), static_cast<search::DocId>(doc_id));
       }
 
-      if (!GetFlag(FLAGS_deserialize_hnsw_index)) {
+      // Only store mappings if deserialization is enabled AND shard count matches.
+      // With different shard counts, keys are distributed differently so the mappings are invalid.
+      if (!deserialize_hnsw_index_ || shard_count_ == 0 || shard_set == nullptr ||
```
romange: use `should_restore` here?

src/server/rdb_load.cc
```diff
@@ -2927,7 +2931,7 @@ void RdbLoader::LoadScriptFromAux(string&& body) {
 namespace {
 
 void LoadSearchCommandFromAux(Service* service, string&& def, string_view command_name,
-                              string_view error_context) {
+                              string_view error_context, bool ignore_dispatch_errors = false) {
```
romange: why do we need to ignore errors ?

dranikpg (dismissed):
I think it would be simplest to create a test afterwards where we replicate a hnsw index under load and see how it goes

src/server/rdb_load.cc
```diff
@@ -2976,48 +2980,42 @@ void LoadSearchCommandFromAux(Service* service, string&& def, string_view comman
...
-  {
-    std::lock_guard lk(search_index_mu_);
-    auto [it, inserted] = created_search_indices_.insert(index_name);
-    if (!inserted) {
-      VLOG(1) << "Index creation already in progress or completed, skipping: " << index_name;
-      return;
-    }
-  }
+absl::flat_hash_map<uint32_t, std::vector<RdbLoader::PendingIndexMapping>>
+RdbLoader::TakePendingIndexMappings() {
```
romange, line 3012: i am not happy we use static storage for these things

src/server/rdb_load.h
BorysTheDev: I think for simplicity, because we send them only in 0 threads or via summary for snapshot.

src/server/rdb_load.cc
BorysTheDev: if should_restore is false we ignore HNSW index data and do rebuild. if should_restore is true we read HNSW index data, including bindings from internal HNSW links to the old doc_ids, after that we restore index and do rebindings to the new doc_ids on replica

src/server/rdb_load.cc
BorysTheDev: we try to create index from every thread and only one succeeds, others return an error. At first I've implemented a separate mechanism to avoid double index attempt creation, but it looks ugly, so I've decided we can ignore index creation error instead, because we can't have 2 indexes with the same name

src/server/rdb_load.cc
BorysTheDev: Me too, but let's have it as is for now. I will create a task for refactoring.

src/server/search/index_builder.cc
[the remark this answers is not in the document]
BorysTheDev: Not now

src/server/rdb_load.cc
BorysTheDev: It is another Opcode so I need to move this code outside. Let's leave as is for now.

src/server/rdb_load.cc
romange: So I think it's not good as it will hide an error from the first call that should succeed.  check if index exists first and then try creating it. One idea - add to FT.CREATE a modifier like `NX` .

src/server/rdb_load.cc
romange: so please expand the comment in the code and explain the thought behind.  i.e. explain 
that we can keep the same internal ids if shard count is the same and we can reuse the data-structures as is and otherwise we must rebuild from scratch or something like that.
What your comment currently does is translating to english the c++ expression `shard_count_ == shard_set->size()`.

BorysTheDev: It's a limitation for the first iteration for the same number of shards, in the second iteration I will remove this condition with shards.

src/server/rdb_load.cc
romange: can `shard_set` be null?

src/server/rdb_load.cc
BorysTheDev: the first idea with checking needs synchronization that I want to avoid, I will consider the idea with NX

src/server/rdb_load.cc
```diff
@@ -3040,12 +3038,94 @@ void RdbLoader::LoadHnswIndexMetadataFromAux(string&& def) {
...
+      std::lock_guard lk(search_index_mu_);
+      for (const auto& phm : pending_hnsw_metadata_) {
+        if (phm.index_name == index_name && phm.field_name == field_name) {
+          metadata = phm.metadata;
+          break;
+        }
+      }
+    }
+
+    CHECK(metadata);
```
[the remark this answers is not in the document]
romange: why did you resolve it? It's not true?

src/server/rdb_load.cc
```diff
@@ -3056,25 +3136,44 @@ void RdbLoader::PerformPostLoad(Service* service, bool is_error) {
...
   std::vector<std::string> synonym_cmds = TakePendingSynonymCommands();
+  auto index_mappings = TakePendingIndexMappings();
+
   if (is_error)
     return;
 
-  // Start index building for all indices
-  // TODO: don't build all indices concurrently or limit cumulative budget
-  shard_set->RunBriefInParallel([](EngineShard* es) {
+  if (shard_set != nullptr && !index_mappings.empty()) {
```
romange: can `shard_set` be null here?

BorysTheDev: no, I will remove it

src/server/rdb_load.cc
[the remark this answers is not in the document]
BorysTheDev: Because I want to crash if we don't have metadata.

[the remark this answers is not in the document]
BorysTheDev: in this code branch, we shouldn't have this situation

src/server/rdb_load.h
```diff
@@ -341,34 +342,44 @@ class RdbLoader : protected RdbLoaderBase {
...
+  static base::SpinLock search_index_mu_;  // created_search_indices_, pending_index_mappings_
+  static absl::flat_hash_map<uint32_t, std::vector<PendingIndexMapping>> pending_index_mappings_;
+  static absl::flat_hash_set<std::string> created_search_indices_;
```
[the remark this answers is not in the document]
BorysTheDev: not important, will be refactored later

src/server/rdb_load.cc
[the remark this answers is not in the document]
romange: will it crash if we replicate from older versions?

PR #6652, opened 2026-02-17
[Do not review -- testing] chore: start journal only on shard threads

kostasrim:


kostasrim:
Random failure unit failure. Opened https://github.com/dragonflydb/dragonfly/issues/6653

Issue #6646, opened 2026-02-16
test_redis_om failing with AssertionError and missing index warnings


glevkovich:
The test tests/dragonfly/search_test.py::test_redis_om is failing in CI:
https://github.com/dragonflydb/dragonfly/actions/runs/22062286515/job/63745244026
Note: another test failed on this run. This issue is only for test_redis_om.

Error: AssertionError: assert equals failed. The actual result returned from the query TestCar.find().all() does not match the expected list of cars (missing "Mercedes" in the reported diff). Screenshot attached for the assertion error: 
<img width="1506" height="398" alt="Image" src="https://github.com/user-attachments/assets/d7eb6c89-2b84-4d69-93c7-f6b61abe2b2e" />

The server logs show the following warning during the test execution:
`W20260216 13:18:39.948372 27513 main_service.cc:1666]  FT.INFO :dragonfly.search_test.TestCar:index failed with reason: Index with name ':dragonfly.search_test.TestCar:index' not found`

Replica logs attached (with both tests that fail, open test_redis_om_df_factory0_)

[logs (1).zip](https://github.com/user-attachments/files/25343685/logs.1.zip)

romange:
@BorysTheDev  could be due to async index creation?

BorysTheDev:
@romange I think @dranikpg knows better, but I can try to check

romange:
oh, I just noticed it's tiering tests. @dranikpg  please check

BorysTheDev:
@romange  it is search test and I think I know the problem, and it looks like you are right, it is connected with the async index that @dranikpg added. I will try to fix

PR #6655, opened 2026-02-17
fix: test_redis_om

BorysTheDev:
fixes: https://github.com/dragonflydb/dragonfly/issues/6646

Because of async index creation, we should wait a little to make the test correct


tests/dragonfly/search_test.py
```diff
@@ -576,6 +576,15 @@ def make_car(producer, description, speed):
 
     redis_om.Migrator().run()
 
+    # Wait for async indexing of existing documents to complete
+    for index in client.execute_command("FT._LIST"):
+        index_name = index.decode() if isinstance(index, bytes) else index
```
dranikpg: decode_responses=True inside Client()

dranikpg (dismissed):

PR #6656, opened 2026-02-17
chore(server): Extend script slowlog stats

dranikpg:
Include
1. Transaction shard number
2. Transaction mode
3. Number of nested slowlog calls

inside slowlog info for EVAL

### DIFF
--- a/src/server/server_family_test.cc
+++ b/src/server/server_family_test.cc
@@ -222,6 +222,7 @@ TEST_F(ServerFamilyTest, SlowLogMinusOneDisabled) {
 
 // Test how slowlog captures additional information about heavy commands
 TEST_F(ServerFamilyTest, SlowLogExecEval) {
+  Run({"config", "set", "slowlog_max_len", "20"});
   Run({"config", "set", "slowlog_log_slower_than", "0"});
 
   // Run EXEC


dranikpg:
augment review

src/server/main_service.cc
```diff
@@ -2329,6 +2333,7 @@ void Service::EvalInternal(CmdArgList args, const EvalArgs& eval_args, Interpret
       }
     } else {
       scheduled = StartMulti(conn_cntx, script_mode, eval_args.keys);
+      sinfo->stats.tx_shards = tx->GetUniqueShardCnt();
```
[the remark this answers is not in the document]
dranikpg: Yes, but it almost impossible to get to that multi info except introducing fields inside MultiData

src/server/main_service.cc
```diff
@@ -2348,6 +2353,8 @@ void Service::EvalInternal(CmdArgList args, const EvalArgs& eval_args, Interpret
       tx->UnlockMulti();
   }
 
+  sinfo->stats.tx_mode = script_mode;
```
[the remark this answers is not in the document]
dranikpg: Only if "Multi mode conflict when running eval in multi transaction"

src/server/main_service.cc
```diff
@@ -2329,6 +2333,7 @@ void Service::EvalInternal(CmdArgList args, const EvalArgs& eval_args, Interpret
       }
     } else {
       scheduled = StartMulti(conn_cntx, script_mode, eval_args.keys);
+      sinfo->stats.tx_shards = tx->GetUniqueShardCnt();
     }
```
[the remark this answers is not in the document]
dranikpg: it will be a wrong number anyways and it requires storing more stats for multi transactions, see previous comment

src/server/conn_context.cc
```diff
@@ -46,8 +48,11 @@ vector<string> FormatEvalSlowlog(const ConnectionState& state) {
 
   const auto& sinfo = *state.script_info;
   return {
-      sinfo.stats.sha,
+      string{sinfo.stats.sha, sizeof(sinfo.stats.sha)},
       absl::StrCat("num_cmds: ", sinfo.stats.num_commands),
+      absl::StrCat("slow_cmds: ", sinfo.stats.slow_commands.load(memory_order_relaxed)),
```
romange, line 53: Please add comments explaining new fields here. Also
we will need to update https://www.dragonflydb.io/docs/command-reference/server-management/slowlog-get

I had a good experience asking copilot just look at the code and update the page but you need comments here so it will understand better the meaning of new fields

romange: actually @copilot please add concise comments.

romange: ah, it does not work because you ~~are stubborn~~  you fork

dranikpg: > Please add comments explaining new fields here

They are already explained where they're defined, inside conn_context.h

dranikpg: I always put it there, so you can hover of click in the IDE instead of searching for usages

PR #6662, opened 2026-02-18
fix(search): Fix HNSW index creation failure, logs, drop

dranikpg:
1. Fix printing of index loading errors by removing newline from definition output
2. Add cleanup to failures in FT.CREATE for vector indices. If (for some unknown reason) we fail to create index N, we have already created indices 0....N and never delete them, so any retries will fail
3. Drop global vector indices with Clear(). It is simpler and possibly more correct, however it might cover some problems

### DIFF
--- a/src/server/search/search_family.cc
+++ b/src/server/search/search_family.cc
@@ -1213,7 +1213,7 @@ vector<SearchResult> SearchGlobalHnswIndex(
   return results;
 }
 
-// Try creating globla hnsw indices for given fields and return true on success
+// Try creating global hnsw indices for given fields and return true on success
 bool CreateHnswIndices(std::string_view idx_name, const DocIndex& index) {
   std::vector<std::string> created_vector_indices;
   for (const auto& [field_ident, field_info] : index.schema.fields) {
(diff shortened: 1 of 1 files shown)


dranikpg:
augment review

src/server/search/search_family.cc
```diff
@@ -1214,6 +1213,31 @@ vector<SearchResult> SearchGlobalHnswIndex(
...
+      continue;
+
+    bool success = GlobalHnswIndexRegistry::Instance().Create(idx_name, field_info.short_name,
+                                                              vparams, index.type);
+    if (!success) {
+      // Clean created indices
+      for (const auto& cfname : created_vector_indices)
+        GlobalHnswIndexRegistry::Instance().Remove(idx_name, cfname);
+      return false;
+    }
```
BorysTheDev, line 1232: can we have a race conditionbetween create and remove?

dranikpg: No, because Ft.Create is a global transaction. And in between the first hop and concluding (last hop), no commands can run

BorysTheDev (dismissed):

dranikpg:
fixed formatting

src/server/search/doc_index.cc
```diff
@@ -773,18 +773,10 @@ unique_ptr<ShardDocIndex> ShardDocIndices::DropIndex(string_view name) {
...
-          !(field.flags & search::SchemaField::NOINDEX)) {
-        GlobalHnswIndexRegistry::Instance().Remove(it->first, field.short_name);
-      }
-    }
-    DropIndexCache(*it->second);
-  }
+  for (const auto& [_, idx] : indices_)
+    DropIndexCache(*idx);
   indices_.clear();
+  GlobalHnswIndexRegistry::Instance().Reset();
```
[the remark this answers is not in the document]
dranikpg: Yes, but `Instance().Remove(name)` _also clears the index across all shards at once_, so its an issue in either configuration

PR #6664, opened 2026-02-18
fix: backward compatibility for replication regarding HNSW index seri…

BorysTheDev:
fixes: #6661

Summary: This PR improves replication backward compatibility around search/HNSW index serialization by version-gating newer RDB extensions.

Changes:

- Introduced DflyVersion::VER6 and bumped CURRENT_VER to reflect new search/HNSW replication features.
- Propagated the negotiated replica version into replication full-sync by setting the version on RdbSaver from the per-flow replication context.
- Added RdbSaver::SetDflyVersion() and threaded the version into SliceSnapshot so snapshot serialization can conditionally include/exclude new opcodes.
- Version-gated emission of search index AUX fields (search-index, hnsw-index-metadata) to avoid confusing older replicas.
- Version-gated HNSW-related snapshot opcodes (RDB_OPCODE_VECTOR_INDEX, RDB_OPCODE_SHARD_DOC_INDEX) so older replicas don’t receive unknown records.
- Adjusted per-shard vs summary behavior for search index definitions to preserve older expectations (definitions from summary-only for older replicas).
- Added a replication integration test that verifies a new master with an HNSW vector index can replicate successfully to a v1.35 replica, including KNN querying on the replica.

### DIFF
--- a/src/server/rdb_save.cc
+++ b/src/server/rdb_save.cc
@@ -1117,7 +1117,7 @@ class RdbSaver::Impl final : public SliceSnapshot::SnapshotDataConsumerInterface
   // We pass K=sz to say how many producers are pushing data in order to maintain
   // correct closing semantics - channel is closing when K producers marked it as closed.
   Impl(bool align_writes, unsigned producers_len, CompressionMode compression_mode,
-       SaveMode save_mode, io::Sink* sink);
+       SaveMode save_mode, io::Sink* sink, DflyVersion replica_dfly_version);
 
   ~Impl();
 
@@ -1139,10 +1139,6 @@ class RdbSaver::Impl final : public SliceSnapshot::SnapshotDataConsumerInterface
   error_code SaveAuxFieldStrStr(string_view key, string_view val);
 
   void CancelInShard(EngineShard* shard);
-
-  void SetDflyVersion(DflyVersion version) {
-    dfly_version_ = version;
-  }
 
   size_t GetTotalBuffersSize() const;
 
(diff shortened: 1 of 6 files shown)


src/server/rdb_save.cc
```diff
@@ -1182,6 +1186,7 @@ class RdbSaver::Impl final : public SliceSnapshot::SnapshotDataConsumerInterface
   // make snapshot size smaller and opreation faster.
   CompressionMode compression_mode_;
   SaveMode save_mode_;
+  DflyVersion dfly_version_ = DflyVersion::CURRENT_VER;
```
dranikpg: This deserves a comment that this is the replica version because I understood in only once I looked all the files

dranikpg: Or even a better name

src/server/snapshot.h
```diff
@@ -75,6 +74,12 @@ class SliceSnapshot : public journal::JournalConsumerInterface {
   // In journal streaming mode it needs to be stopped by either Stop or Cancel.
   enum class SnapshotFlush : uint8_t { kAllow, kDisallow };
 
+  // Set the replica version for conditional serialization of new features.
+  // Must be called before Start.
+  void SetDflyVersion(DflyVersion version) {
+    dfly_version_ = version;
+  }
+
```
dranikpg: > Must be called before Start.

Then it can be part of the constructor or start

src/server/rdb_save.cc
dranikpg: also if you pass it as a parameter until the snapshot, you don't need to keep it as a field

src/server/rdb_save.h
```diff
@@ -128,6 +129,14 @@ class RdbSaver {
     return save_mode_;
   }
 
+  // Set the replica version for conditional serialization of new features.
+  // Must be called before SaveHeader() and StartSnapshotInShard().
+  void SetDflyVersion(DflyVersion version);
+
+  DflyVersion GetDflyVersion() const {
+    return dfly_version_;
+  }
```
dranikpg: Set can be a parameter, Get is never used

dranikpg (approved):
I understand you want to fix it quickly, but its not very elegant

PR #6670, opened 2026-02-18
refactor: move stats into one file to reduce the number of includes in headers

BorysTheDev:
Summary: This PR refactors server statistics types to reduce header dependencies and compile-time includes.

Changes:

- Moves stats structs (e.g., SlotStats, DbTableStats, DbStats, SliceEvents, ShardStats) into a new centralized server/stats.h
- Moves the corresponding operator+= and helper implementations into server/stats.cc
- Replaces EngineShard::Stats with an alias to ShardStats and updates aggregated metrics to use the shared type
- Reduces header includes via forward declarations (notably in namespaces.h and server_family.h) and adds an out-of-line Namespace destructor
- Moves shard_count() implementations for Service and ServerFamily out of headers to further cut includes
- Updates many .cc files to explicitly include db_slice.h/engine_shard_set.h/config_registry.h as needed after the refactor
- Technical Notes: stats.cc now validates the object-type count via a static_assert against Redis’ OBJ_TYPE_MAX to keep the new per-type memory arrays in sync.


dranikpg:
I don't really understand what the benefit of moving out DbSlice::Stats if for example, if we just include the whole stats header in db_slice.h 🤔

BorysTheDev:
> I don't really understand what the benefit of moving out DbSlice::Stats if for example, if we just include the whole stats header in db_slice.h 🤔

We don't need to include db_slice.h when we need only stats

kostasrim:
> > I don't really understand what the benefit of moving out DbSlice::Stats if for example, if we just include the whole stats header in db_slice.h 🤔
> 
> We don't need to include db_slice.h when we need only stats

I don't see a `single` change in this PR that includes `only stats.h` or any other change that justifies its existence one way or another.

I also do not like or want a `big header/pile of stats`. I rather have each stat logically related where it belongs: it's translation unit (slightly bad wording here -- headers are not actual TU). 

To make it worse, imagine if we put on all stat related declarations from `src/server/*.h` to `src/server/stats`

BorysTheDev:
> > > I don't really understand what the benefit of moving out DbSlice::Stats if for example, if we just include the whole stats header in db_slice.h 🤔
> > 
> > 
> > We don't need to include db_slice.h when we need only stats
> 
> I don't see a `single` change in this PR that includes `only stats.h` or any other change that justifies its existence one way or another.
> 
> I also do not like or want a `big header/pile of stats`. I rather have each stat logically related where it belongs: it's translation unit (slightly bad wording here -- headers are not actual TU).
> 
> To make it worse, imagine if we put on all stat related declarations from `src/server/*.h` to `src/server/stats`

src/server/server_family.h is included in several places, and now it depends only on stats. but not on the engine_shard and db_slice

kostasrim:
> luded in several places, and now it depends only on stats. but not on the engine_shard

It doesn't matter where it is included. It matters how dependencies propagate:

server/dflycmd.cc:36:#include "server/server_family.h"   -> includes `db_slice` anyway
server/conn_context.cc:15:#include "server/server_family.h" -> include `engine_shard_set` which includes `engine_shard`
server/memory_cmd.cc:25:#include "server/server_family.h"   -> same reason as above
server/cluster/cluster_family.cc:33:#include "server/server_family.h"
server/cluster/outgoing_slot_migration.cc:22:#include "server/server_family.h"
server/dfly_main.cc:55:#include "server/server_family.h"  

One way or another you end up including `engine_shard and db_slice` so most of the boilerplate declarations are there.
Moreover:

1. These are translation units, so including X in Y.cc does not propagate X's declarations since X is not part of Y's header.
2. You don't save any CPU cycles with these transitive includes unless you understand `the exact dependency graph` you are trying to break or you have heavy template programming. Neither apply here.

PR #6674, opened 2026-02-18
refactor(mc): async MGET/GAT

glevkovich:
This refactor migrates MGET and GAT to SimpleContext, enabling async
support as part of the final push for IO loop v2 for Memcached.

* Migrated from synchronous BlockingCounter to SimpleContext for
parallel shard execution and deferred replies.

* Distinguish soft errors (KEY_NOTFOUND, WRONG_TYPE) to return nil,
while capturing hard errors (OOM) via atomic status storage.

* Fixed SendValue parameter order to correctly propagate CAS and TTL
according to the Memcached protocol requirements.

* Optimized flag reading by caching deduplication settings in the
Prepare() phase to reduce per-shard overhead.

### DIFF
--- a/src/server/string_family.cc
+++ b/src/server/string_family.cc
@@ -1462,7 +1462,11 @@ void MGetCmd::Reply(SinkReplyBuilder* rb) {
       // Note: Missed keys are silent.
       if (res.has_value()) {
         string value;
-        if (holds_alternative<string>(res->val)) {
+
+        if (mc_cmd_flags_.meta && !mc_cmd_flags_.return_value) {
+          // If it's a Meta command and the client didn't ask for the value,
+          // we don't need to materialize the string or wait for disk I/O.
+        } else if (holds_alternative<string>(res->val)) {
           value = get<string>(std::move(res->val));
         } else {
           auto io_res = get<TieredRes>(std::move(res->val)).Get();


src/server/string_family.cc
```diff
@@ -1446,116 +1443,123 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
-void MGetGeneric(CmdArgList args, CommandContext* cmd_cntx,
-                 const DbSlice::ExpireParams* gat_params) {
-  DCHECK_GE(args.size(), 1U);
+OpStatus MGetCmd::operator()(const ShardArgs& args, const OpArgs& op_args) const {
+  for (auto it = args.begin(); it != args.end(); ++it) {
+    size_t i = it.index();
 
-  MemcacheCmdFlags cmd_flags;
+    // skip the 'expiry' argument for GAT.
+    if (is_gat_ && (i == 0)) {
```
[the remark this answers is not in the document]
glevkovich: fixed

src/server/string_family.cc
```diff
@@ -1446,116 +1443,123 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
-    // Access to arguments must be done via cmd_cntx.
-    auto replier = [mget_results = std::move(mget_results), cmd_cntx](SinkReplyBuilder* builder) {
-      auto* mc_builder = static_cast<MCReplyBuilder*>(builder);
-      facade::MCRender mc_render{cmd_cntx->mc_command()->cmd_flags};
-      for (size_t i = 0; i < cmd_cntx->size(); ++i) {
-        const auto& entry = mget_results[i];
-        if (entry) {
-          mc_builder->SendValue(cmd_cntx->mc_command()->cmd_flags, cmd_cntx->at(i), entry->value, 0,
-                                entry->mc_flag, entry->ttl_sec);
+      // Note: Missed keys are silent.
```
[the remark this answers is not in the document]
glevkovich: fixed

src/server/string_family.cc
```diff
@@ -1446,116 +1443,123 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
         } else {
-          mc_builder->SendSimpleString(mc_render.RenderMiss());
+          auto iores = get<TieredRes>(std::move(*res)).Get();
+          if (!iores)
+            continue;  // skip on tiering error
+          value = std::move(*iores);
         }
+
+        string_view current_key = args_[i];
+        mc_rb->SendValue(mc_cmd_flags_, current_key, value, 0, 0, 0);
```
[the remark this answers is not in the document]
glevkovich: fixed

src/server/string_family.cc
```diff
@@ -1446,116 +1443,123 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
-                    cmd_cntx](SinkReplyBuilder* builder) {
-      auto* redis_builder = static_cast<RedisReplyBuilder*>(builder);
-      redis_builder->StartArray(arg_len);
-      for (size_t i = 0; i < arg_len; ++i) {
-        const auto& entry = mget_results[i];
-        if (entry) {
-          redis_builder->SendBulkString(entry->value);
+    }
+
+    mc_rb->SendSimpleString("END");
```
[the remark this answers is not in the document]
glevkovich: already fixed

src/server/string_family.cc
```diff
@@ -1446,116 +1443,123 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
+      if (res.has_value()) {
+        if (holds_alternative<string>(*res)) {
+          redis_rb->SendBulkString(get<string>(std::move(*res)));
         } else {
-          redis_builder->SendNull();
+          auto iores = get<TieredRes>(std::move(*res)).Get();
+          if (iores) {
+            redis_rb->SendBulkString(*iores);
+          } else {
+            redis_rb->SendNull();  // Tiering read failed
```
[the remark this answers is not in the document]
glevkovich: fixed

src/server/string_family.cc
```diff
@@ -1446,116 +1443,123 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
+      return facade::ErrorReply{"GAT is a memcache-only command"};
     }
+    expire_params_ = DbSlice::ExpireParams{
+        .value = mc_cmd->expire_ts, .absolute = true, .persist = (mc_cmd->expire_ts == 0)};
+  }
+  if (is_mc_) {
+    mc_cmd_flags_ = mc_cmd->cmd_flags;
   }
+  // Reserve space for all keys before parallel execution starts
+  results_.resize(args.size());
```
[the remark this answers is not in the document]
glevkovich: fixed

src/server/string_family.cc
```diff
@@ -1446,116 +1443,123 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
-  DCHECK_GE(args.size(), 1U);
+OpStatus MGetCmd::operator()(const ShardArgs& args, const OpArgs& op_args) const {
+  for (auto it = args.begin(); it != args.end(); ++it) {
+    size_t i = it.index();
 
-  MemcacheCmdFlags cmd_flags;
+    // skip the 'expiry' argument for GAT.
+    if (is_gat_ && (i == 0)) {
+      continue;
+    }
```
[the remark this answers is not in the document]
glevkovich: fixed already

src/server/string_family.cc
```diff
@@ -1446,116 +1443,123 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
+    // GAT has an expiry argument at index 0, so keys start at 1.
+    // MGET keys start at 0.
+    size_t start_idx = is_gat_ ? 1 : 0;
 
-  auto cb = [&](Transaction* t, EngineShard* shard) {
-    mget_resp[shard->shard_id()] =
-        OpMGet(tiering_bc, &tiering_err, cmd_flags, t, shard, gat_params);
-    return OpStatus::OK;
-  };
+    MCRender mc_render{mc_cmd_flags_};
```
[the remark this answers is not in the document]
glevkovich: used now.

src/server/string_family.cc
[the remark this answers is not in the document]
glevkovich: fixed already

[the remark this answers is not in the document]
glevkovich: fixed already

src/server/string_family.cc
```diff
@@ -1446,116 +1443,123 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
-      auto* redis_builder = static_cast<RedisReplyBuilder*>(builder);
-      redis_builder->StartArray(arg_len);
-      for (size_t i = 0; i < arg_len; ++i) {
-        const auto& entry = mget_results[i];
-        if (entry) {
-          redis_builder->SendBulkString(entry->value);
+    }
+
+    mc_rb->SendSimpleString("END");
+  } else {  // MGET
```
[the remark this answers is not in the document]
glevkovich: fixed

src/server/string_family.cc
```diff
@@ -1446,116 +1443,123 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
-  if (cmd_cntx->mc_command()) {
-    cmd_flags = cmd_cntx->mc_command()->cmd_flags;
+    if (result) {
+      results_[i] = std::move(result.value());
+    } else {
+      if (result.status() == OpStatus::KEY_NOTFOUND) {
+        results_[i] = std::nullopt;
+      } else {  // system error (e.g OUT_OF_MEMORY)
+        return result.status();
+      }
```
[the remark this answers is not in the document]
glevkovich: fixed

src/server/string_family.cc
```diff
@@ -1446,116 +1443,123 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
+        if (holds_alternative<string>(*res)) {
+          redis_rb->SendBulkString(get<string>(std::move(*res)));
         } else {
-          redis_builder->SendNull();
+          auto iores = get<TieredRes>(std::move(*res)).Get();
+          if (iores) {
+            redis_rb->SendBulkString(*iores);
+          } else {
+            redis_rb->SendNull();  // Tiering read failed
+          }
```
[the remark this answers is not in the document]
glevkovich: drop

src/server/string_family.cc
```diff
@@ -793,26 +793,23 @@ OpResult<DbSlice::Iterator> FindKeyAndSetExpiry(const GetAndTouchParams& params)
...
+      return res.status();
+    pv = &res.value()->second;
+  } else {  // MGET
+    auto res = op_args.GetDbSlice().FindReadOnly(op_args.db_cntx, key, OBJ_STRING);
+    if (!res)
+      return OpStatus::KEY_NOTFOUND;
+    pv = &res.value()->second;
   }
+
+  return ReadString(op_args.db_cntx.db_index, key, *pv, op_args.shard);
```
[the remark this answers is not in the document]
glevkovich: keeping optimizations foer later.

src/server/string_family.cc
```diff
@@ -793,26 +793,23 @@ OpResult<DbSlice::Iterator> FindKeyAndSetExpiry(const GetAndTouchParams& params)
...
+
+  if (expire_params) {  // GAT
+    auto res = FindKeyAndSetExpiry({op_args.tx, op_args.shard, *expire_params, key});
+    if (!res)
+      return res.status();
+    pv = &res.value()->second;
+  } else {  // MGET
+    auto res = op_args.GetDbSlice().FindReadOnly(op_args.db_cntx, key, OBJ_STRING);
+    if (!res)
+      return OpStatus::KEY_NOTFOUND;
```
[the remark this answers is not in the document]
glevkovich: fixed

src/server/string_family.cc
```diff
@@ -1446,116 +1355,166 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
+            if (!miss_resp.empty()) {
+              mc_rb->SendSimpleString(miss_resp);
+            }
+            continue;
+          }
+          value = std::move(*io_res);
+        }
+
+        string_view current_key = args_[i];
+        mc_rb->SendValue(mc_cmd_flags_, current_key, value, res->meta.mc_flag, res->meta.mc_ver, 0);
```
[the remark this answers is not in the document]
glevkovich: fixed already

src/server/string_family.cc
```diff
@@ -1446,116 +1340,177 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
+
+        // If the original result is already fetched and is a string (RAM hit),
+        // we copy it to avoid a second database lookup.
+        // If it's a Tiered Future (Disk hit), we fall through to fetch it again because
+        // sharing or cloning futures is complex and slower than a second lookup.
+        if (results_[orig_idx].has_value()) {
+          if (auto* val_str = std::get_if<string>(&results_[orig_idx]->val)) {
+            results_[i] = Result{*val_str, results_[orig_idx]->meta};
+            continue;
+          }
```
dranikpg: That's not true. A second lookup will at least construct a new future _and_ run all the fetching logic

dranikpg:
```c++
if (size_t orig_idx = map_it->second; !inserted && results_[orig_idx])
  results_[i] = results[orid_idx];
```

src/server/string_family.cc
```diff
@@ -1446,116 +1340,177 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
-
-  MemcacheCmdFlags cmd_flags;
+OpStatus MGetCmd::operator()(const ShardArgs& args, const OpArgs& op_args) const {
+  // Only use a map if:
+  // 1) dedup_enabled_ is on AND
+  // 2)we have potential duplicates (>1 key) to save heap
+  //  allocations. If args.Size() is 0 or 1, duplicates are impossible, so we save the heap
+  //  allocation. AND
+  // 3) command is MGET
+  bool dedup = dedup_enabled_ && (args.Size() > 1) && !is_gat_;
```
dranikpg, line 1405: nit: I think the code is really self explanatory

src/server/string_family.cc
```diff
@@ -1446,116 +1340,177 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
+  absl::flat_hash_map<string_view, size_t> key_first_idx;
```
dranikpg, line 1408: missing an important comment from the original code why it can't be thread local

src/server/string_family.cc
```diff
@@ -1446,116 +1340,177 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
+    if (result) {
+      results_[i] = Result{std::move(result.value()), meta};
+    } else {
+      auto status = result.status();
+      if ((status == OpStatus::KEY_NOTFOUND) || (status == OpStatus::WRONG_TYPE)) {  // soft errors
+        results_[i] = std::nullopt;
+      } else {  // hard errors (e.g OUT_OF_MEMORY)
+        status_code_.store(status, std::memory_order_relaxed);
+        return status;
+      }
```
dranikpg: we have this bad convention that a handler can either return OK, OUT_OF_MEMORY or throw an std::bad_alloc. And almost exclusively we use (1) and (3). 

if (result = OpMget())
  results[i] = {result}

should be enough

src/server/string_family.cc
```diff
@@ -1446,116 +1340,177 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
+}
 
-  fb2::BlockingCounter tiering_bc{0};  // Count of pending tiered reads
-  AggregateError tiering_err;          // First tiering error
-
-  unique_ptr<MGetResponse[]> mget_resp(new MGetResponse[shard_set->size()]);
+void MGetCmd::Reply(SinkReplyBuilder* rb) {
+  if (OpStatus status = status_code_.load(std::memory_order_relaxed); status != OpStatus::OK) {
+    return rb->SendError(status);
+  }
```
dranikpg: And then you tx->LocalResultPtr() here, I'll improve that part later. `status_code_` is not a substitute because it dones't catch std::bad_alloc

src/server/string_family.cc
```diff
@@ -1446,116 +1340,177 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
+            // Treat disk failure as a miss
+            redis_rb->SendNull();
+          }
         }
+      } else {
+        redis_rb->SendNull();  // Key not found (KEY_NOTFOUND)
       }
-    };
-    cmd_cntx->ReplyWith(std::move(replier));
+    }
```
dranikpg: all of this is just GetReplies{rb}.Send(std::move(res))

src/server/string_family.cc
```diff
@@ -1446,116 +1340,177 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
+          redis_rb->SendBulkString(get<string>(std::move(res->val)));
         } else {
-          redis_builder->SendNull();
+          auto io_res = get<TieredRes>(std::move(res->val)).Get();
+          if (io_res) {
+            redis_rb->SendBulkString(*io_res);
+          } else {
+            // We can't send error, else we violate RESP protocol
+            // Treat disk failure as a miss
+            redis_rb->SendNull();
```
dranikpg: you can add a parameter to GetReplies to send null on io error

src/server/string_family.cc
```diff
@@ -793,26 +660,53 @@ OpResult<DbSlice::Iterator> FindKeyAndSetExpiry(const GetAndTouchParams& params)
...
-    return CollectKeys(std::move(wait_bc), err, cmd_flags, t, shard, std::move(find_op));
+struct MGetMetadata {
+  uint32_t mc_flag = 0;
+  uint64_t mc_ver = 0;
+  uint32_t ttl_sec = 0;
+};
+
+OpResult<StringResult> OpMGet(const OpArgs& op_args, string_view key,
+                              optional<DbSlice::ExpireParams> expire_params,
+                              MGetMetadata* metadata = nullptr) {
```
dranikpg: Using StringResult is of course nicer, however comparing to previous mget we now:
1. have separate strings instead of a single char[] buffer for all values
2. create a new future for every tiered read - previously we used a callback that wrote directly to the buffer and decremented a single latch for all futures

dranikpg:
This changes how MGET works and now uses separate StringResults instead of a single buffer (and a single counter for tiering). We should either measure if it has any impact or ask @romange if he's okay with it. It won't have any impact for a single key, but MGET in redis is often used to fetch many values at once.

src/server/string_family.cc
glevkovich: fixed, thanks !

src/server/string_family.cc
glevkovich: adding "  // We can not make it thread-local because we may preempt during the Find loop due to
  // replication of expiry events."
thanks.

src/server/string_family.cc
glevkovich: OK, I think this comment made things clear to me, Thanks for that.
I removed status_code_ and made the above simpler.

src/server/string_family.cc
glevkovich: I already removed status_code_  based one previous comment.
Ok, so I'm not going to deal with tx->LocalResultPtr() (exception handling?) for this PR

src/server/string_family.cc
glevkovich: Using GetReplies without changing it isn't a good idea...
I looked into using GetReplies{rb}.Send(std::move(res->val)), but I don't think we can use it here for MGET without causing a bug.

In GetReplies::Send(TResultOrT<T>&&), if the tiered storage future fails, it calls Send(iores.error().message()), which routes to rb->SendBulkString().

If we use that for MGET, a disk I/O error will result in the string "Input/output error" being returned to the client as the actual value of the key, rather than a RESP Null. My current manual block safely catches the I/O error and returns SendNull() to respect the MGET array protocol.

src/server/string_family.cc
glevkovich: ok with this clarification and my previous comment, I'll make the change on GetReplies to adapt it for my case.
Thanks.

src/server/string_family.cc
glevkovich: I will answer this together with https://github.com/dragonflydb/dragonfly/pull/6674#issuecomment-3925660069

glevkovich:
> This changes how MGET works and now uses separate StringResults instead of a single buffer (and a single counter for tiering). We should either measure if it has any impact or ask @romange if he's okay with it. It won't have any impact for a single key, but MGET in redis is often used to fetch many values at once.

@dranikpg 
I agree with this observation,  actually Copilot and other AI review tools flagged this exact same trade-off to me while I was writing it. I was fully aware of that..
I decided to proceed with this way for a few main reasons:

1)
Typical Workloads: from my understanding, in practice, most MGET queries are for a small batch of keys (e.g., 2-20), not thousands. mimalloc is very fast at handling these small string allocations, so the overhead for the strings and the extra futures should be negligible for standard use cases.

2)
PR Scope: The primary goal of this PR is the architectural refactor, so I didn't want to deal with performance optimization at this stage.

3)
I found out that adapting the old code to the new refactoring added significant complexity. In addition, the previous implementation had its own drawbacks - it was much longer, harder to maintain, and suffered from inefficiencies like extra memory copies. The new approach is significantly simpler and reduces those copies, which should help offset the overhead of the individual string and future allocations.
The new StringResult approach drastically simplifies the execution flow. If this ever shows up as a bottleneck in profiling later, we can definitely optimise it with a custom memory pool or buffer, but I think the massive win in readability and maintainability is worth the trade-off here.
Please keep in mind that I haven't finished the  task yet. My next step is to run benchmarks and compare the results against the existing implementation (pre-merge). I’ll follow up with those numbers soon so we can see the actual impact and decide if further optimizations are necessary.

src/server/string_family.cc
```diff
@@ -793,26 +660,51 @@ OpResult<DbSlice::Iterator> FindKeyAndSetExpiry(const GetAndTouchParams& params)
...
-      return db_slice.FindReadOnly(t->GetDbContext(), key, OBJ_STRING);
-    };
-    return CollectKeys(std::move(wait_bc), err, cmd_flags, t, shard, std::move(find_op));
+struct MGetMetadata {
+  uint32_t mc_flag = 0;
+  uint32_t ttl_sec = 0;
+};
+
+OpResult<StringResult> OpMGet(const OpArgs& op_args, string_view key,
+                              optional<DbSlice::ExpireParams> expire_params,
```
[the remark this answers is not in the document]
glevkovich: fixed

src/server/string_family.cc
```diff
@@ -793,26 +660,51 @@ OpResult<DbSlice::Iterator> FindKeyAndSetExpiry(const GetAndTouchParams& params)
...
+      metadata->mc_flag = db_slice.GetMCFlag(db_index, it->first);
+    }
+    if (it->first.HasExpire()) {  // TTL
+      auto exp_it = db_slice.GetDBTable(db_index)->expire.Find(it->first);
+      if (exp_it != db_slice.GetDBTable(db_index)->expire.end()) {
+        int64_t expire_time_ms = db_slice.ExpireTime(exp_it->second);
+        int64_t ttl_ms = expire_time_ms - op_args.db_cntx.time_now_ms;
+        metadata->ttl_sec = ttl_ms > 0 ? static_cast<uint32_t>((ttl_ms + 999) / 1000) : 0;
+      }
+    }
```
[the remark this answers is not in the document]
glevkovich: fixed

src/server/string_family.cc
romange: I think it's better to have comments that explain the context behind than not to have. I am talking about (2) at least.

glevkovich: @dranikpg  I agree that 1 and 3 are WHAT but 2 is WHY and that's why I preferred putting here a complete explanation) +  I don't think that 100% of the readers will understand it from the code, so I would prefer keeping it.

src/server/string_family.cc
```diff
@@ -653,7 +520,8 @@ OpResult<TResultOrT<size_t>> OpExtend(const OpArgs& op_args, std::string_view ke
 
 // Helper for building replies for strings
 struct GetReplies {
-  GetReplies(SinkReplyBuilder* rb) : rb{static_cast<RedisReplyBuilder*>(rb)} {
+  GetReplies(SinkReplyBuilder* rb, bool null_on_io_error = false)
+      : rb{static_cast<RedisReplyBuilder*>(rb)}, null_on_io_error_{null_on_io_error} {
     DCHECK(dynamic_cast<RedisReplyBuilder*>(rb));
```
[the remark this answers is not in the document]
glevkovich: fixed

src/server/string_family.cc
```diff
@@ -1446,116 +1340,161 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
-  if (cmd_cntx->mc_command()) {
-    cmd_flags = cmd_cntx->mc_command()->cmd_flags;
+    MGetMetadata meta;
+    if (auto result = OpMGet(op_args, key, expire_params_, is_mc_ ? &meta : nullptr, want_ttl)) {
+      results_[i] = Result{std::move(result.value()), meta};
+    }
+    // If result is falsy (KEY_NOTFOUND, WRONG_TYPE), results_[i] remains nullopt (miss).
+    // Hard errors like OUT_OF_MEMORY throw std::bad_alloc per Dragonfly convention.
   }
+  return OpStatus::OK;
```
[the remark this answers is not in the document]
glevkovich: dropping...not true

src/server/string_family.cc
```diff
@@ -1446,114 +1345,169 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
+          // original was a RAM hit. Safely copy the string. skip OpMGet
+          results_[i] = results_[orig_idx];
+          continue;
         }
+
+        // Original key is a pending disk read (Future). In fb2::Future is single-consumer, we
+        // cannot copy it without risking a hang or empty result during consumption in Reply().
+        // While a two-pass alias-index solution could avoid redundant I/O, it adds significant
+        // structural complexity to the Reply logic. Given this is a rare edge case, refetching a
+        // fresh Future is the safest and most maintainable approach at this stage.
```
dranikpg, line 1439: Ah, true, that's inconvenient. It doesn't have to be this way, but it would require fixing helio as well, so its cumbersome for some edge case

src/server/string_family.cc
```diff
@@ -1446,114 +1345,169 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
+    // This is because a single tiered-storage read failure shouldn't abort the entire MGET batch.
+    GetReplies replies{rb, /*null_on_io_error=*/true};
+
+    for (auto& res : results_) {
+      if (res.has_value()) {
+        replies.Send(std::move(res->val));
+      } else {
+        redis_rb->SendNull();  // Key not found (KEY_NOTFOUND)
+      }
+    }
```
dranikpg, line 1508: GetReplies knows how to send an optional

src/server/string_family.cc
```diff
@@ -1446,114 +1345,169 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
+        std::string miss_resp = mc_render.RenderMiss();
+        if (!miss_resp.empty()) {
+          mc_rb->SendSimpleString(miss_resp);
+        }
+      }
+    }
+    std::string end_resp = mc_render.RenderGetEnd();
+    if (!end_resp.empty()) {
+      mc_rb->SendSimpleString(end_resp);
+    }
```
dranikpg, line 1499: That makes me remember that we don't do this for other commands that can be noreply. This is generally a flag for mutation commands and not intended to use for gets. `MCRender` was introduced by Roman because he wanted to abolish reply builders, but we haven't chosen that path, so it's in some kind of weird state now

dranikpg (approved):
Some minor code/noreply comments

src/server/string_family.cc
glevkovich: I'm going to open a task for this, assign to myself for now. My understanding, is to have a shared future. But there are other solutions on dragonfly only, which makes it more complex. I think shared future is a nice solution.

src/server/string_family.cc
glevkovich: changed, thank you.
```
  for (auto& res : results_) {
      replies.Send(res ? std::make_optional(std::move(res->val)) : std::nullopt);
    }
```

src/server/string_family.cc
glevkovich: I see what you mean about the weird state. There are other flags like return_ttl, return_cas, and base64 in that struct,  and MCRender is currently the only thing that knows how to format the response correctly. I'll leave it as is for now but I agree refactoring of the reply building is needed and then we can totally remove the MCRender.

src/server/string_family.cc
```diff
@@ -1446,114 +1345,164 @@ void CmdIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
...
+      // Note: Missed keys are silent.
+      if (res.has_value()) {
+        string value;
+        if (holds_alternative<string>(res->val)) {
+          value = get<string>(std::move(res->val));
+        } else {
+          auto io_res = get<TieredRes>(std::move(res->val)).Get();
+          if (!io_res) {
+            std::string miss_resp = mc_render.RenderMiss();
+            if (!miss_resp.empty()) {
```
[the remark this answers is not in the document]
glevkovich: fixed

PR #6675, opened 2026-02-18
test: mark heavy skipped tests as large instead of skip

vyavdoshenko:
Replace `@pytest.mark.skip` with `@pytest.mark.large` for tests that were skipped solely because they require significant resources (RAM, time), not because they are broken.

- `test_tiered_entries_throttle` (snapshot_test.py)
- `test_big_huge_streaming_restart` (replication_test.py)
- `test_cluster_memory_consumption_migration` (cluster_test.py) — 15G RAM, 3 nodes
- `test_sbf_chunked_replication_over_4gb` (replication_test.py) — ~12GiB RAM

These tests will now run on CI-LARGE runners via the heavy-tests workflow instead of being permanently skipped.

### DIFF
--- a/tests/dragonfly/snapshot_test.py
+++ b/tests/dragonfly/snapshot_test.py
@@ -594,9 +594,8 @@ async def test_bgsave_and_save(async_client: aioredis.Redis):
 @dfly_args(
     {
         **BASIC_ARGS,
-        "proactor_threads": 4,
         "dbfilename": "tiered-entries",
-        "tiered_prefix": "tiering-test-backing",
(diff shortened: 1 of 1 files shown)


tests/dragonfly/cluster_test.py
dranikpg, line 2679: I doubt this test actually does anything useful

dranikpg: it doesn't measure memory consumption, maybe intended as a manual test

dranikpg:
I changed a parameter of a test to speed it up

dranikpg:
I'll check the tiering test for failures

PR #6676, opened 2026-02-18
feat(dash): segment merging

kostasrim:
This PR implements a simple Merge function for two buddy segments in DashTable. It does not implement directory merging and it's *only specific* to two segments. It does not provide a high level loop (in dash table) that attempts to merge all segments for example. 

When a segment splits:

* One segments becomes two buddies
* Local depth increases by 1
* Items redistribute across the new segments

When buddies merge:
* Two buddies become one segment
* Depth decreases by 1
* Items get merged into one segment

```
Merge:
  Move all elements from buddy to segment
  Update buddy to point to segment
  Deallocate buddy
  Decrease depth
```

It's important to note that the call to `InsertUniq` should not fail if `Merge` is called with its `preconditions` met.

In summary:

* add Merge and FindBuddy to DashTable
* add unit test

Helps with #7

### DIFF
--- a/src/core/dash.h
+++ b/src/core/dash.h
@@ -191,7 +191,9 @@ class DashTable : public detail::DashTableBase {
     auto* seg = GetSegment(segment_id);
     uint8_t depth = seg->local_depth();
 
-    assert(depth != 1);
+    if (depth <= 1) {
+      return segment_id;
+    }
 
     const size_t bit_pos = global_depth_ - depth;
     const size_t buddy_idx = segment_id ^ (1u << bit_pos);


src/core/dash.h
```diff
@@ -182,6 +182,74 @@ class DashTable : public detail::DashTableBase {
...
+
+    return buddy_idx;
+  }
+
+  // API Contract:
+  //  Do not call this function unless the combined
+  //  size of both segments is less than x * segment_capacity. With x:
+  //  0 < x < 0.25. Under uniform hash distribution insertions
+  //  during merge will never fail as long as x remains under 25%.
+  //  What is more, this 25% is not universal. For non default bucket size (14 slots)
```
kostasrim: In other words we do not need to handle any special case like `SplitBug`

src/core/dash.h
```diff
@@ -182,6 +182,74 @@ class DashTable : public detail::DashTableBase {
...
+
+    // Free buddy segment
+    PMR_NS::polymorphic_allocator<SegmentType> pa(segment_.get_allocator());
+    using alloc_traits = std::allocator_traits<decltype(pa)>;
+    alloc_traits::destroy(pa, buddy);
+    alloc_traits::deallocate(pa, buddy, 1);
+
+    // Decrement unique segment counter
+    --unique_segments_;
+    bucket_count_ -= keep->num_buckets();
```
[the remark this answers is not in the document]
kostasrim: same number doesn't matter

src/core/dash.h
```diff
@@ -182,6 +182,74 @@ class DashTable : public detail::DashTableBase {
     return segment_[segment_id];
   }
 
+  unsigned FindBuddyId(unsigned segment_id) {
+    auto* seg = GetSegment(segment_id);
+    uint8_t depth = seg->local_depth();
+
```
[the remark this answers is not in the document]
kostasrim: We start at depth 1 and we should not go below that.

src/core/dash.h
```diff
@@ -182,6 +182,74 @@ class DashTable : public detail::DashTableBase {
...
+  //  during merge will never fail as long as x remains under 25%.
+  //  What is more, this 25% is not universal. For non default bucket size (14 slots)
+  //  this number needs to be calibrated.
+  void Merge(unsigned keep_id, unsigned buddy_id) {
+    auto* keep = GetSegment(keep_id);
+    auto* buddy = GetSegment(buddy_id);
+
+    assert((keep->local_depth() == buddy->local_depth()));
+    assert((keep->SlowSize() + buddy->SlowSize() < (0.25 * buddy->capacity())));
+    assert(keep->local_depth() != 1);
```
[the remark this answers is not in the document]
kostasrim: nonsense. If we merge two level 1 segments we end up with a level 0 which is `not allowed` and violates the basic property that `gdepth >= 1`

src/core/dash_test.cc
```diff
@@ -373,6 +373,71 @@ TEST_F(DashTest, Split) {
...
+      LOG(INFO) << "Merging segments " << seg_id << " (depth=" << local_depth
+                << ", size=" << seg->SlowSize() << ") and " << buddy_id
+                << " (depth=" << buddy->local_depth() << ", size=" << buddy->SlowSize()
+                << "), combined=" << combined_size;
+
+      dt_.Merge(seg_id, buddy_id);
+    }
+  }
+
+  EXPECT_EQ(dt_.unique_segments(), 2);
```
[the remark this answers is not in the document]
kostasrim: If my grandma had wheels instead of hair she would have been a hot wheels car instead of a human being.

Split() or Merge() won't change, neither xxxhash we use.

src/core/dash_test.cc
```diff
@@ -373,6 +373,71 @@ TEST_F(DashTest, Split) {
...
+
+    if (combined_size <= safe_threshold) {
+      LOG(INFO) << "Merging segments " << seg_id << " (depth=" << local_depth
+                << ", size=" << seg->SlowSize() << ") and " << buddy_id
+                << " (depth=" << buddy->local_depth() << ", size=" << buddy->SlowSize()
+                << "), combined=" << combined_size;
+
+      dt_.Merge(seg_id, buddy_id);
+    }
+  }
```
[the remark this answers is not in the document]
kostasrim: Nonsense for the most part.

> The test iterates through all directory entries and may attempt to merge the same buddy pair multiple times or in different orders.

No it does not. FindBuddy adjusts the depth which finds the CORRECT buddy at each iteration.

> e. If Merge has bugs in directory pointer updates, the test might access deallocated memory or merge segments incorrectly

This will crash....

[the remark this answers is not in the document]
kostasrim: > Consider adding a set to track which unique segments have been processed

This part is actually ok, I had LOG's but I guess these were not enough so I will the assertions also.

src/core/dash.h
```diff
@@ -182,6 +182,74 @@ class DashTable : public detail::DashTableBase {
...
+    buddy->TraverseAll([&](const auto& it) {
+      uint64_t hash = DoHash(buddy->Key(it.index, it.slot));
+      auto& src_bucket = buddy->GetBucket(it.index);
+      [[maybe_unused]] auto res =
+          keep->InsertUniq(std::move(src_bucket.key[it.slot]), std::move(src_bucket.value[it.slot]),
+                           hash, false, [](auto&&...) {});
+      // Because of preconditions
+      assert(res.found());
+    });
+
```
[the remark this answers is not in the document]
kostasrim: bs, we destruct the segment which clears it anyway

src/core/dash.h
```diff
@@ -182,6 +182,78 @@ class DashTable : public detail::DashTableBase {
...
+      return segment_id;
+    }
+
+    return buddy_idx;
+  }
+
+  // API Contract:
+  //  Do not call this function unless the combined
+  //  size of both segments is less than x * segment_capacity. With x:
+  //  0 < x < 0.25. Under uniform hash distribution insertions
```
romange: why `x < 0.25` condition is part of the contract? 
I think it is possible to fail the merge even with x < 0.25, and it is possible to succeed with `x > 0.25`.

If indeed my statement is correct, the real contract should be is that Merge can fail and return false, in which case the moved items returned back to their segments

src/core/dash.h
romange: not clear to me why.

src/core/dash.h
```diff
@@ -182,6 +182,78 @@ class DashTable : public detail::DashTableBase {
...
+  //  during merge will never fail as long as x remains under 25%.
+  //  What is more, this 25% is not universal. For non default bucket size (14 slots)
+  //  this number needs to be calibrated.
+  void Merge(unsigned keep_id, unsigned buddy_id) {
```
romange: Please add the comment what this function does, where it moves items from

src/core/dash.h
kostasrim: Precisely to avoid the rollback you suggest below 😄 

> If indeed my statement is correct, the real contract should be is that Merge can fail and return false, in which case the moved items returned back to their segments

That's the thing -- I don't want to ping pong items and I want to guarantee statistically that this can not happen. I am proactive instead of reactive in this context. 

Of x < 0.25 then at least `75%` of the `merged` segment will be empty. For our bucket/slot sizes we should be in the "statistacally impossible" case for insertion to fail (assuming the worst possible case of insertion). 

I do not blindly trust the math of copilot, and that's why the assertions + preconditions. I can verify this again or run the math on my own

src/core/dash.h
kostasrim: +1

src/core/dash.h
romange: "guarantee statistically"  is oxymoron. You can still merge segments that have less than 25% itemsm but from correctness perspective I find it problematic that we "ignore" possible cases. 

It's like saying "using spinlocks is not required, because concurrent access to a single integer is improbable"

src/core/dash.h
```diff
@@ -182,6 +182,78 @@ class DashTable : public detail::DashTableBase {
...
+  // If there is no buddy for segment_id return segment_id
+  // Otherwise, return buddy_id.
+  unsigned FindBuddyId(unsigned segment_id) {
+    auto* seg = GetSegment(segment_id);
+    uint8_t depth = seg->local_depth();
+
+    assert(depth != 1);
+
+    const size_t bit_pos = global_depth_ - depth;
+    const size_t buddy_idx = segment_id ^ (1 << bit_pos);
```
dranikpg: Please add a comment who a `buddy` is, i.e. the other _leaf_ segment with the same parent, computed by flipping `depth` bit

src/core/dash.h
```diff
@@ -182,6 +182,78 @@ class DashTable : public detail::DashTableBase {
     return segment_[segment_id];
   }
 
+  // If there is no buddy for segment_id return segment_id
+  // Otherwise, return buddy_id.
+  unsigned FindBuddyId(unsigned segment_id) {
```
dranikpg: Please add a short comment on who a buddy is - a segment with the last real adress bit flipped

src/core/dash.h
```diff
@@ -182,6 +182,78 @@ class DashTable : public detail::DashTableBase {
...
+    const size_t buddy_idx = segment_id ^ (1 << bit_pos);
+    assert(buddy_idx < segment_.size());
+
+    auto* buddy = GetSegment(buddy_idx);
+    // If buddy has different local depth, then the segment
+    // at buddy_idx is not actually a buddy. We return back
+    // the segment_id to denote that.
+    if (buddy->local_depth() != depth) {
+      return segment_id;
+    }
```
dranikpg: this comment explains the code, but not the causes - adjacent subtree might be deeper

src/core/dash.h
dranikpg: isn't there only one option for buddy_id and that is `FindBuddyId(keep_id)` so it doesn't have to be a parameter?

src/core/dash.h
```diff
@@ -182,6 +182,78 @@ class DashTable : public detail::DashTableBase {
...
+  // API Contract:
+  //  Do not call this function unless the combined
+  //  size of both segments is less than x * segment_capacity. With x:
+  //  0 < x < 0.25. Under uniform hash distribution insertions
+  //  during merge will never fail as long as x remains under 25%.
+  //  What is more, this 25% is not universal. For non default bucket size (14 slots)
+  //  this number needs to be calibrated.
+  void Merge(unsigned keep_id, unsigned buddy_id) {
+    auto* keep = GetSegment(keep_id);
+    auto* buddy = GetSegment(buddy_id);
```
dranikpg, line 221: and this can become 
```c++
auto* keep = GetSegment(keep_id);
auto* buddy = GetSegment(FindBuddyId(keep_id));

if (keep == buddy)
  return;
```

src/core/dash.h
dranikpg: I get Kosta's idea, its lucrative if you can guarantee it - 25% is a good number and you don't need to write the rollback code or compute it ahead

src/core/dash.h
kostasrim: yes but the idea is to return false/true if the merge fails (because of rollback) not because of improper use of the API (let's say because there is no buddy). I don;'t have a strong preference

src/core/dash.h
kostasrim: see my other comment also, I would like to avoid returning false on violated preconditions and keep true/false for when merge actualyl fails and we need to rollbackj

src/core/dash_test.cc
```diff
@@ -373,6 +373,162 @@ TEST_F(DashTest, Split) {
...
+
+  EXPECT_EQ(dt_.size(), total_size_before);
+
+  // After rollback, src and buddy pointers should still be valid
+  for (auto key : keep_keys) {
+    uint64_t hash = dt_.DoHash(key);
+    auto it = src->FindIt(hash, EqTo(key));
+    EXPECT_TRUE(it.found());
+    // Bucket layout in keep segment changes during failed merge rollback.
+    // InsertUniq can displace existing items in the keep segment (via MoveToOther)
```
kostasrim: @romange during merge an insertion might `displace` items. During rollbacks, the `buddy` is reconstructed with each key/value being placed in the exact same location before the merge.

This is not true however for the `keep` segment if it had to `displace` elements during `merge insertions`.

It should not be an actual problem but hten we can't say the operation is exactly the inverse. The data is the same, the elements though of one are now displaced 

WDYT?

src/core/dash.h
```diff
@@ -182,6 +182,130 @@ class DashTable : public detail::DashTableBase {
...
+      if (!res.found()) {
+        should_rollback = true;
+        return;
+      }
+
+      moved.emplace_back((it.index << 8) | it.slot, (res.index << 8) | res.slot, hash);
+    });
+
+    if (should_rollback) {
+      for (const auto& rec : moved) {
```
romange: why do you need the "moved" to rollback? why not do something like calling `SegmentType::Split`.
i.e. `buddy->Split(keep)`

kostasrim: because `buddy->Split()` does extra work which should not be part of a "reverse operation"

* updates bucket version
* applies eviction policy on_move

If we fail `merge` it means we should `revert back to the same state` as before. If we bump up the version, or apply eviction policy we won't have segment(before rollback)==segment(after  rollback).

This is true for the source bucket as well. Insertions during merge can cause `displacements`. This means that the source bucket after a rollback will have the same data but might have a different layout (e,g, some of its elements got displaced). 

Also thinking about this, we should `bump up` the bucket version after a succesful merge. I guess this is not important for now as we are not planning to run this during snapshoting

romange: But now you pay a significant price for happy path as you need to keep everything just in case. Even if not split, why not do what split does, i.e go over all the items and move if their hash should not be there?

As you said - position argument does not hold here as we change positions if merge succeeds

src/core/dash.h
```diff
@@ -182,6 +182,107 @@ class DashTable : public detail::DashTableBase {
...
+      return false;
+    }
+
+    // Same as Split()
+    uint32_t buddy_chunk_size = 1u << (global_depth_ - buddy->local_depth());
+    uint32_t buddy_start = buddy_id & ~(buddy_chunk_size - 1u);
+    for (size_t i = buddy_start; i < buddy_start + buddy_chunk_size; ++i) {
+      segment_[i] = keep;
+    }
+
```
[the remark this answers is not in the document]
kostasrim: I also added a test

src/core/dash_test.cc
```diff
@@ -373,6 +373,143 @@ TEST_F(DashTest, Split) {
...
+    if (inserted) {
+      all_keys.push_back(i);
+    }
+  }
+
+  EXPECT_GE(dt_.depth(), 2);
+
+  unsigned sid = 0;
+  size_t buddy_id = dt_.FindBuddyId(sid);
+  EXPECT_NE(buddy_id, sid);
```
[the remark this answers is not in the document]
kostasrim: yeah and the assertion will trigger. it's controlled

src/core/dash_test.cc
```diff
@@ -373,6 +373,143 @@ TEST_F(DashTest, Split) {
...
+    EXPECT_TRUE(it.found());
+  }
+
+  for (auto key : buddy_keys) {
+    uint64_t hash = dt_.DoHash(key);
+    auto it = buddy->FindIt(hash, EqTo(key));
+    EXPECT_TRUE(it.found());
+  }
+
+  EXPECT_FALSE(merge_succeeded);
```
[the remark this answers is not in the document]
kostasrim: yeah it's controlled, it won't flake

src/core/dash.h
```diff
@@ -182,6 +182,112 @@ class DashTable : public detail::DashTableBase {
...
+    if (should_rollback) {
+      auto hash_fn = [this](const auto& k) { return policy_.HashFn(k); };
+      keep->Split(hash_fn, buddy,
+                  [&](uint32_t segment_from, detail::PhysicalBid from, uint32_t segment_to,
+                      detail::PhysicalBid to) {
+                    // no-op
+                  });
+
+      return false;
```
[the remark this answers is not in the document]
kostasrim: That's should not happen, we should call Merge() in a canonical way. I added an assertion

src/core/dash.h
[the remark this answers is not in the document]
kostasrim: It confused me so much. It should be canonical, and I added an assertion

src/core/dash.h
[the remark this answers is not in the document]
kostasrim: I also had to properly use a bottom up merge since I broke merge symmetry. But IMO it's so much simpler algorithm -- and conceptually -- we iterate from the leafs and we repeat that either until we reach to depth=1 with 2 leafs or if we can't further merge

src/core/dash.h
```diff
@@ -182,6 +182,104 @@ class DashTable : public detail::DashTableBase {
...
+  bool Merge(unsigned keep_id, unsigned buddy_id) {
+    auto* keep = GetSegment(keep_id);
+    auto* buddy = GetSegment(buddy_id);
+
+    assert((keep->local_depth() == buddy->local_depth()));
+    // assert((keep->SlowSize() + buddy->SlowSize() < (0.25 * buddy->capacity())));
+    assert(keep->local_depth() != 1);
+    assert(keep != buddy);
+    assert(keep_id < buddy_id);  // Callers must iterate low to high to ensure correct orientation
+
```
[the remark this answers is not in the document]
kostasrim: No I won't normalize. Final merging will be a bottom up algorithm starting from the leaves and exiting early if it can't merge any further.

I am completely allergic to the idea of symmetric `Merge()`. It overcomplicates the same simple leaf only algorithm.

src/core/dash.h
kostasrim: @romange I ended up using `Split` as suggested -- PTAL

src/core/dash.h
```diff
@@ -182,6 +182,104 @@ class DashTable : public detail::DashTableBase {
...
+    uint8_t depth = seg->local_depth();
+
+    assert(depth != 1);
+
+    const size_t bit_pos = global_depth_ - depth;
+    const size_t buddy_idx = segment_id ^ (1u << bit_pos);
+    assert(buddy_idx < segment_.size());
+
+    auto* buddy = GetSegment(buddy_idx);
+    // There is no adjacent subtree of different depth
```
romange:
```suggestion
    // There is no adjacent subtree of the same depth
```

romange:
augment review

src/core/dash.h
```diff
@@ -182,6 +182,104 @@ class DashTable : public detail::DashTableBase {
...
+  // - If there is no buddy for segment_id return segment_id.
+  //   Otherwise, return buddy_id.
+  // - A buddy is a sibling segment that was created from the
+  //   same parent during split and can be merged back together.
+  //   It's the adjacent subtree of the same depth.
+  unsigned FindBuddyId(unsigned segment_id) {
+    auto* seg = GetSegment(segment_id);
+    uint8_t depth = seg->local_depth();
+
+    assert(depth != 1);
```
romange:
```suggestion
    if (depth <= 1)
       return segment_id;
```

src/core/dash.h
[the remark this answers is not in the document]
romange: even though it does not matter right now, lets write more robust code and move these two lines before buddy deallocation and use `buddy->bucket_count()`. we had an idea to have dynamic buckets in the past.

romange:
I added more tests @kostasrim .

romange:
Looks good overall. 
1. your merge comment is mostly about the threshold, the whole paragraph can be shortened given that we support rollbacks.
2. There is no mention of versioning. we should comment that Merge() can only run when there are no active snapshots.

kostasrim:
> Looks good overall. 


done -- I still kept the 25% comment but reduced greatly

PR #6679, opened 2026-02-19
refactor: clean includes

BorysTheDev:
Summary: Refactors header dependencies across the facade/server code to tighten include boundaries and reduce compile coupling.

Changes:

- Replaces several transitive includes with forward declarations (e.g., HTTP listener/stats types in facade, shard/slice/controller types in server).
- Reworks replication-related types so `Replica`/`ServerFamily` can depend on a lighter-weight “types” header instead of full replica definitions.
- Adds direct includes in a number of `.cc` units (e.g., `db_slice.h`, `common.h`, `blocking_controller.h`, `absl/strings/ascii.h`) and removes unused standard/absl headers.


src/server/replica_types.h
```diff
@@ -0,0 +1,41 @@
...
+#pragma once
+
+#include <string>
+#include <vector>
+
+#include "server/common_types.h"
+
+namespace dfly {
+
+struct ReplicaSummary {
```
kostasrim, line 14: plz revert this back to `Replica.h`

kostasrim (changes requested):

src/server/replica_types.h
BorysTheDev: Why?

kostasrim: As I explained iternally this is not how `src/server` is. We put stats next to the logical unit. 

This change has zero effect in compile time and it breaks `encapsulation`

dranikpg: I think this is actually fine

BorysTheDev: replica.h is a huge header, and we don't need to propagate dependency on it to a lot of other files via src/server/server_family.h if we can include only a lightweight summary.
The first layer of propagation:
- src/server/cluster/cluster_family.cc
- src/server/cluster/outgoing_slot_migration.cc
- src/server/conn_context.cc
- src/server/dfly_main.cc
- src/server/dflycmd.cc
- src/server/main_service.h     
    - propagates dependencies further to more than 50 files 
- src/server/memory_cmd.cc
- src/server/server_family_test.cc
- src/server/server_family.cc
- src/server/server_family.cc.orig

kostasrim: Again this is `parellizable` it doesn't matter

BorysTheDev: There is no sense include replica in files that don't need it. And no encapsulation is broken, because everything was public. and there is no sense compiling and wasting resources on the thing that you don't need

Issue #6667, opened 2026-02-18
Refactor rdb_load


BorysTheDev:
We need to refactor rdb_load to remove the next static fields:

-   static std::vector<std::string> pending_synonym_cmds_;
-   static base::SpinLock search_index_mu_;  // created_search_indices_, pending_index_mappings_
-   static absl::flat_hash_map<uint32_t, std::vector<PendingIndexMapping>> pending_index_mappings_;
-   static absl::flat_hash_set<std::string> created_search_indices_;
-   static std::vector<PendingHnswMetadata> pending_hnsw_metadata_;

kostasrim:
plz provide some context for all to understand 😄

PR #6684, opened 2026-02-19
refactor: remove static fields from rdb_load

BorysTheDev:
fixes: #6667
Summary: This PR refactors RDB loading to eliminate shared static state in RdbLoader by introducing a per-load-session RdbLoadContext.

Changes:

- Adds RdbLoadContext (rdb_load_context.{h,cc}) to collect pending search-related data (synonyms, key→DocId mappings, HNSW metadata) across multiple loader instances.
- Moves LoadSearchCommandFromAux and post-load search rebuild logic from RdbLoader statics into the shared context.
- Updates RdbLoader to accept/store a RdbLoadContext* and routes pending state through it instead of static containers + spinlock.
- Updates snapshot loading (ServerFamily::Load/LoadRdb) to create a shared context and call PerformPostLoad once after all shard loads complete.
- Updates Dragonfly replication flows (Replica/DflyShardReplica) to share the same context across flows and invoke PerformPostLoad at the end of full sync.
- Adjusts tests to construct RdbLoadContext when instantiating RdbLoader.
- Technical Notes: The new context uses an internal mutex for thread-safe accumulation and provides a single consolidation point for rebuilding/restoring search indices after load.

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -1369,7 +1369,7 @@ std::optional<fb2::Future<GenericError>> ServerFamily::Load(const std::string& p
   load_fibers.reserve(paths.size());
 
   LoadOptions load_opts;
-  auto load_context = std::make_shared<RdbLoadContext>();
+  auto load_context = std::make_unique<RdbLoadContext>();
   if (absl::EndsWith(path, "summary.dfs")) {
     // we read summary first to get snapshot_id and load data correctly
     error_code load_ec = pool.GetNextProactor()->Await(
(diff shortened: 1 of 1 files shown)


dranikpg (approved):

src/server/server_family.cc
```diff
@@ -1368,10 +1368,11 @@ std::optional<fb2::Future<GenericError>> ServerFamily::Load(const std::string& p
   load_fibers.reserve(paths.size());
 
   LoadOptions load_opts;
+  auto load_context = std::make_shared<RdbLoadContext>();
```
romange: seems that the real owner of `load_context` is `load_join_func` and `load_context` can be a unique_ptr.

src/server/server_family.cc
```diff
@@ -1392,8 +1393,9 @@ std::optional<fb2::Future<GenericError>> ServerFamily::Load(const std::string& p
       proactor = pool.GetNextProactor();
     }
 
-    auto load_func = [file, existing_keys, load_opts, aggregated_result, this]() mutable {
-      error_code load_ec = LoadRdb(file, existing_keys, &load_opts);
+    auto load_func = [file, existing_keys, load_opts, aggregated_result, load_context,
```
romange: and here you can pass `load_ctx_ptr = load_context.get`

src/server/replica.h
```diff
@@ -190,7 +190,8 @@ class RdbLoader;
 class DflyShardReplica : public ProtocolClient {
  public:
   DflyShardReplica(ServerContext server_context, MasterContext master_context, uint32_t flow_id,
-                   Service* service, std::shared_ptr<MultiShardExecution> multi_shard_exe);
+                   Service* service, std::shared_ptr<MultiShardExecution> multi_shard_exe,
+                   class RdbLoadContext* load_context);
```
romange, line 194: is it a c++20 feature?

BorysTheDev: no

PR #6685, opened 2026-02-19
chore: add pipeline_blocking_commands_total counter

### LINKED ISSUE #6669: Add metrics for pipelines that contain blocking commands (blpop, etc)
This would help us understand if a pipeline is stack on a blocked command.

kostasrim:
Add a counter "pipeline_blocking_commands_total" that tracks the total number of blocking commands encountered during pipelined execution. Expose this to prometheous metrics

* add metric

resolves #6669

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -1628,7 +1628,7 @@ void PrintPrometheusMetrics(uint64_t uptime, const Metrics& m, DflyCmd* dfly_cmd
   AppendMetricWithoutLabels("pipeline_queue_wait_duration_seconds", "",
                             conn_stats.pipelined_wait_latency * 1e-6, MetricType::COUNTER,
                             &resp->body());
-  AppendMetricWithoutLabels("pipelined_blocking_commands_total", "",
+  AppendMetricWithoutLabels("pipeline_blocking_commands_total", "",
                             m.coordinator_stats.blocking_commands_in_pipelines, MetricType::COUNTER,
                             &resp->body());
 


src/facade/dragonfly_connection.cc
```diff
@@ -1531,7 +1533,7 @@ void Connection::SquashPipeline() {
       stats_->pipelined_wait_latency += CycleClock::ToUsec(start - current->parsed_cycle);
     }
 
-    ReleaseParsedCommand(current, result.account_in_stats /* is_pipelined */);
+    ReleaseParsedCommand(current, true);
```
kostasrim: This was a bug. `result.account_in_stats` is set to `true` only if `bool account_in_stats = total_usec > squash_stats_latency_lower_limit_cached;` The signature however is: `ReleaseParsedCommand(ParsedCommand* cmd, bool is_pipelined)`. So we `missed` a bunch of pipeline stats.

src/facade/dragonfly_connection.cc
```diff
@@ -1487,9 +1487,11 @@ void Connection::SquashPipeline() {
   // This lambda advances a temporary pointer exec_cmd_ptr to feed the execution engine.
   // We do not modify parsed_to_execute_ yet, in case execution throws/fails.
   auto exec_cmd_ptr{parsed_to_execute_};
-  auto get_next_fn = [&exec_cmd_ptr]() mutable -> ParsedArgs {
+  auto get_next_fn = [&exec_cmd_ptr]() mutable -> std::pair<ParsedArgs, bool*> {
```
kostasrim: I am not the biggest fan of this but I see no other simpler way 🤷

kostasrim:
I can also add a test I guess 🤷

src/server/main_service.cc
```diff
@@ -1748,10 +1753,15 @@ DispatchManyResult Service::DispatchManyCommands(std::function<facade::ParsedArg
...
   for (unsigned i = 0; i < count; i++) {
-    ParsedArgs args = arg_gen();
+    auto [args, is_blocking_ptr] = arg_gen();
     string cmd = absl::AsciiStrToUpper(args.Front());
     const auto [cid, tail_args] = registry_.FindExtended(cmd, args.Tail());
 
+    // Mark if this is a blocking command for latency tracking
+    if (is_blocking_ptr && cid && cid->IsBlocking()) {
+      *is_blocking_ptr = true;
+    }
```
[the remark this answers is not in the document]
kostasrim: nope, when I Release it I reset the state

romange:
I do not think that latency is the main issue here.  It's rather a symptom.

The core issue is the EXISTENCE of blocking commands being sent via pipelining mode. The metric purpose is to identify such cases so we could notify users of misuse of Dragonfly.

src/facade/parsed_command.cc
```diff
@@ -50,6 +50,7 @@ string MCRender::RenderDeleted() const {
 void ParsedCommand::ResetForReuse() {
   is_deferred_reply_ = false;
   reply_ = std::monostate{};
+  is_blocking_cmd = false;
```
romange: I think the logic should the the other way around. facade::ConnectionContext already has `async_dispatch` flag. You should just add a counter to ServerState::Stats that is increased if `async_dispatch` and `is_blocking()`. facade should not know what blocking command is.

So lets just count such events.

src/facade/ok_main.cc
```diff
@@ -33,11 +33,11 @@ class OkService : public ServiceInterface {
     return DispatchResult::OK;
   }
 
-  DispatchManyResult DispatchManyCommands(std::function<ParsedArgs()> arg_gen, unsigned count,
-                                          SinkReplyBuilder* builder,
+  DispatchManyResult DispatchManyCommands(std::function<std::pair<ParsedArgs, bool*>()> arg_gen,
+                                          unsigned count, SinkReplyBuilder* builder,
```
romange: the change should be much simpler.

src/facade/dragonfly_connection.cc
romange: I suspect it was NOT a bug. `result.account_in_stats` is set only from the pipeline flow.

kostasrim: I am confused. This code is part of `SquashPipeline` so we are on pipeline flow. The variable `account_in_stats` is only set however, when the pipeline latency was higher than the flag's value. 

Now if you look on `ReleaseParsedCommand` 

```
  2360   if (is_pipelined) {                                                                                                                                                                                                                
  2361     stats_->pipelined_cmd_cnt++;                                                                                                                                                                                                     
  2362     uint64_t latency_usec = CycleClock::ToUsec(CycleClock::Now() - cmd->parsed_cycle);                                                                                                                                               
  2363     stats_->pipelined_cmd_latency += latency_usec;                                                                                                                                                                                   
  2364                                                                                                                                                                                                                                      
  2365     if (cmd->is_blocking_cmd) {                                                                                                                                                                                                      
  2366       stats_->pipelined_blocking_cmd_latency += latency_usec;                                                                                                                                                                        
  2367     }                                                                                                                                                                                                                                
  2368   }                                                                                                                                                                                                                                  
  2369     
```

So we only want to count slow pipeline commands only ? Same for latency ?

On previous versions of dragonfly (e.g. v1.33.0 with `RecycleMsg`), we run this code unconditionally so `we changed our metrics around pipelines`. Was this intended ? If yes,  I revert, if no then this is an actual bug on what pipeline metrics we track (slow or all)

romange: yes, this was intended. This is how I measure percentiles by filtering out samples less than some threshold.

src/server/test_utils.cc
```diff
@@ -497,10 +497,10 @@ void BaseFamilyTest::RunMany(const std::vector<std::vector<std::string>>& cmds)
   for (size_t i = 0; i < cmds.size(); ++i) {
     backed_args_vec[i] = cmn::BackedArguments(cmds[i].begin(), cmds[i].end(), cmds[i].size());
   }
-  auto next_fn = [it = backed_args_vec.begin()]() mutable {
+  auto next_fn = [it = backed_args_vec.begin()]() mutable -> std::pair<ParsedArgs, bool*> {
     ParsedArgs args(*it);
     ++it;
-    return args;
+    return {args, nullptr};
```
romange: not needed imho

kostasrim:
@romange should we warn via `LOG` or just expose the metric ?

src/facade/ok_main.cc
```diff
@@ -32,7 +32,6 @@ class OkService : public ServiceInterface {
     cmd->rb()->SendOk();
     return DispatchResult::OK;
   }
-
```
romange, line 35: pls revert

src/facade/dragonfly_connection.cc
```diff
@@ -1544,7 +1544,7 @@ void Connection::SquashPipeline() {
       conn_stats.pipelined_wait_latency += CycleClock::ToUsec(start - current->parsed_cycle);
     }
 
-    ReleaseParsedCommand(current, result.account_in_stats /* is_pipelined */);
+    ReleaseParsedCommand(current, result.account_in_stats);
```
romange: pls revert

src/server/server_family.cc
```diff
@@ -3178,6 +3178,7 @@ string ServerFamily::FormatInfoMetrics(const Metrics& m, std::string_view sectio
     append("total_pipelined_commands", conn_stats.pipelined_cmd_cnt);
     append("pipeline_throttle_total", conn_stats.pipeline_throttle_count);
     append("pipelined_latency_usec", conn_stats.pipelined_cmd_latency);
+    append("pipelined_blocking_cmd_total", m.coordinator_stats.blocking_commands_in_pipelines);
```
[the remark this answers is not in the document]
romange: I actually prefer we add to prometheus metrics rather in stats as it's simpler to track and monitor.
I switched to adding new things to /metrics only - it's more scalable.

src/facade/dragonfly_connection.cc
kostasrim: I will revert but `is_pipelined` is missleading in this context. The command is pipelined anyway on this flow. It's a matter of if it's slow or not (based on the flag)

src/server/server_family.cc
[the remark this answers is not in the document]
kostasrim: done, removed the other one

kostasrim:
> Much simpler now. Did you test it manually at least?

copilot did :)

PR #6691, opened 2026-02-20
fix(tiering): Track all keys mutation callback for cluster migration

### LINKED ISSUE #6683: test_cluster_migration_with_tiering_and_deletes fails with off-by-one key count mismatch
The failure occurred on my PR (#6674) but is completely unrelated to the changes (my PR only touches `MGET`/`GAT` logic).

**CI Link:** [https://github.com/dragonflydb/dragonfly/actions/runs/22180622257/job/64140668595?pr=6674](https://github.com/dragonflydb/dragonfly/actions/runs/22180622257/job/64140668595?pr=6674)

**Failure Output:**
The test fails on the final assertion where the destination node's key count is off by exactly one key:

```python
>       assert info["db0"]["keys"] == keys - delete_succeded
E       assert equals failed
E         962938  962937

```
Right before the assertion fails, the replica throws an I/O error on the migration connection.

From `dragonfly.WARNING` (Process 22804):
```text
W20260219 12:06:25.729902 22804 execution_state.cc:93] ReportError: Input/output error

```
Logs attached.

[large-tests-arm-logs.zip](https://github.com/user-attachments/files/25415626/large-tests-arm-logs.zip)

mkaruza:
During OnDbChange callback we need to track all keys in bucket when
tiering is used. When key is external and we have already request it to be
fetched from tiered storage it can happen that this externalization
request is already processed - removing external flag but not yet
serialized. When mutation is executed we cannot relay that key is marked
as external and need to check all keys in bucket to be first serialized.
    
Fixes #6683

### DIFF
--- a/src/server/journal/streamer.cc
+++ b/src/server/journal/streamer.cc
@@ -602,6 +602,11 @@ bool RestoreStreamer::WriteBucket(PrimeTable::bucket_iterator it, const ExpireTa
   // Only track tiered keys when needed and flush delayed entries
   // 1. When we have tiered storage
   // 2. We're called from a OnDbChange callback
+  //
+  // We need to track all keys in bucket with tiering. Even if they are not set as external. There
+  // is situation when we request externalization of key and key is read - marking it as not
+  // external but not yet flushed. When OnDbChange callback is called we need to flush it and than
+  // write journal changes - so we cannot realy on IsExternal flag and need to track all keys.
   const bool track_tiered_keys =
       on_db_change_cb && EngineShard::tlocal()->tiered_storage() != nullptr;
 
@@ -622,7 +627,7 @@ bool RestoreStreamer::WriteBucket(PrimeTable::bucket_iterator it, const ExpireTa
           expire = db_slice_->ExpireTime(eit->second);
         }
         // Track tiered keys that will need delayed entry flushing
-        if (track_tiered_keys && pv.IsExternal() && !pv.IsCool()) {
+        if (track_tiered_keys) {
           tiered_keys.emplace(key);
         }
         WriteEntry(key, it->first, pv, expire);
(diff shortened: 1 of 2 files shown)


src/server/journal/streamer.cc
```diff
@@ -602,6 +602,11 @@ bool RestoreStreamer::WriteBucket(PrimeTable::bucket_iterator it, const ExpireTa
   // Only track tiered keys when needed and flush delayed entries
   // 1. When we have tiered storage
   // 2. We're called from a OnDbChange callback
+  //
+  // We need to track all keys in bucket with tiering. Even if they are not set as external. There
+  // is situation when we request externalization of key and key is read - marking it as not
+  // external but not yet flushed. When OnDbChange callback is called we need to flush it and than
+  // write journal changes - so we cannot realy on IsExternal flag and need to track all keys.
```
dranikpg, line 609: I don't understand 😅 

> Even if they are not set as external. There
>  // is situation when we request externalization of key and key is read - marking it as not
> // external but not yet flushed

if its not external that its a handled like a regular value? how come reordering can happen?

mkaruza: @dranikpg 

We request fetching external and add it to container to be serialized later. 
External is materialized and key external flag is removed. 

DELETE is executed, OnCbChange is called. We iterate bucket that is already serialized and collect keys that are external and needs to be serialized. Since this key is not anymore external and it's not serialized we write delete and on next iteration of cluster migration we will write key.

dranikpg: Ok, it was just phrased not very clearly 🙂

dranikpg (approved):
Its inconvenient now that we have to check all entries

mkaruza:
> Its inconvenient now that we have to check all entries

I added todo comment in code that this needs to be improved - this PR is so we have fix issue and have correct results.

PR #6692, opened 2026-02-20
refactor: move methods from OAHEntry to OAHSet

BorysTheDev:


src/core/oah_set_test.cc
```diff
@@ -112,54 +104,6 @@ TEST_F(OAHSetTest, OAHEntryTest) {
   EXPECT_EQ(test.Pop().Key(), "123456789");
 }
 
-TEST_F(OAHSetTest, HashCheckTest) {
```
vyavdoshenko, line 115: Why was this removed?

BorysTheDev: because this test was for the removed functionality, this functionality now in OAHSet, and it is cover by another tests

PR #6693, opened 2026-02-20
[WIP, DRAFT] ci: add shared build infrastructure with S3 artifact caching

vyavdoshenko:
Add new CI infrastructure that builds on LARGE runners and shares artifacts via S3. Runs alongside existing `ci.yml` for validation — no changes to current CI.

Changes:
- **`.github/actions/ensure-build/`** — composite action: check S3 for existing artifact, build if miss, upload
- **`.github/actions/fetch-build/`** — composite action: download artifact from S3, extract to `/build`
- **`.github/workflows/ci-shared-build.yml`** — new workflow (2 build configs: x86 Debug + ARM Release, 3 test jobs)
- **`.github/actions/regression-tests/`** — add optional `dragonfly-path` input (backward-compatible)

How it works:
1. Build jobs run on `CI-LARGE-86`/`CI-LARGE-ARM`, use `cmake -B /build` (fixed path for consistent CTestTestfile.cmake)
2. Artifacts packaged with `tar --dereference | zstd` and uploaded to S3 (`ci-builds/{sha}/{build-id}/`)
3. Test jobs download from S3, extract to `/build`, run ctest/pytest as usual
4. Concurrency groups with `cancel-in-progress: false` queue builds for the same SHA — second build skips if artifact exists

Prerequisites:
- S3 bucket with lifecycle rule (5-day TTL on `ci-builds/` prefix)
- `CI_BUILD_BUCKET` secret in repo settings

### DIFF
--- a/.github/workflows/bullmq-tests.yml
+++ b/.github/workflows/bullmq-tests.yml
@@ -72,7 +72,7 @@ jobs:
       - name: Run BullMQ tests
         run: |
           cd ${GITHUB_WORKSPACE}/bullmq
-          BULLMQ_TEST_PREFIX={b} yarn test --grep "handle errors.*for flows|Flows - addBulk.*handle errors|removes deduplication key|rejects with missing key for job message|emits waiting when a job has been added|gets all workers for this queue only" --invert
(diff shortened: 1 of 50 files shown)


romange:
Note that using s3 for caching will increase our egress costs on AWS. Also, not sure how much time you will save there.

vyavdoshenko:
> Note that using s3 for caching will increase our egress costs on AWS. Also, not sure how much time you will save there.

How much time we can save, we will measure when we have a couple of first builds. Let's try it out.
We have to find a place to store the artifacts. If you think that S3 is not a good place, please give me another direction to consider. It was the first idea for me to reuse AWS S3 for this task as well.

PR #6706, opened 2026-02-23
feat(server): First coroutine for Del

dranikpg:
Use coroutine for DEL command

### DIFF
--- a/src/server/cmd_support.cc
+++ b/src/server/cmd_support.cc
@@ -4,6 +4,8 @@
 
 #include "server/cmd_support.h"
 
+#include <absl/cleanup/cleanup.h>
+
 #include "base/logging.h"
 
 namespace dfly::cmd {
@@ -54,6 +56,27 @@ BlockResult HopCoordinator::SingleHop(CommandContext* cmd_cntx, Transaction::Run
   // Schedule single hop and return blocker
   tx_keepalive_->SingleHopAsync(cb);
   return tx_keepalive_->Blocker();
+}
+
+bool SingleHopWaiter::await_ready() noexcept {
+  return (blocker = SingleHop(cmd_cntx, callback)) == nullptr;
+}
+
+void SingleHopWaiter::await_suspend(std::coroutine_handle<> handle) const noexcept {
+  // TODO: functor calling resume is double indirection and wasted space for cleaner
+  absl::Cleanup cleaner = [handle] { handle.destroy(); };
+  cmd_cntx->Resolve(blocker, [handle, cleaner = std::move(cleaner)](auto* rb) mutable {
+    handle.resume();
+    std::move(cleaner).Cancel();
+  });
+}
+
+facade::OpStatus SingleHopWaiter::await_resume() const noexcept {
+  return *cmd_cntx->tx()->LocalResultPtr();
+}
+
+void CmdR::Coro::return_value(facade::ErrorReply&& err) const noexcept {
+  cmd_cntx->SendError(err);
 }
 
 }  // namespace dfly::cmd
--- a/src/server/cmd_support.h
+++ b/src/server/cmd_support.h
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <coroutine>
 #include <variant>
 
 #include "facade/error.h"
@@ -19,6 +20,12 @@ using BlockResult = util::fb2::EmbeddedBlockingCounter*;
 
 // No execution was performed, the command is ready to reply
 struct JustReplySentinel {};
+
+// Request to perform a single hop
+using SingleHopSentinel = Transaction::RunnableType;
+SingleHopSentinel SingleHop(auto&& f) {
+  return f;
+}
 
 // Handler for dispatching hops, must be part of a context
 struct HopCoordinator {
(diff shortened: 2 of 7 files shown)


dranikpg:
augment review

src/server/cmd_support.cc
```diff
@@ -56,4 +56,21 @@ BlockResult HopCoordinator::SingleHop(CommandContext* cmd_cntx, Transaction::Run
...
 }
 
+bool SingleHopWaiter::await_ready() noexcept {
+  return (blocker = SingleHop(cmd_cntx, callback)) == nullptr;
+}
+
+void SingleHopWaiter::await_suspend(std::coroutine_handle<> handle) const noexcept {
+  // TODO: functor calling resume is double indirection
+  cmd_cntx->Resolve(blocker, [handle](auto* rb) { handle.resume(); });
+}
```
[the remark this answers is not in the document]
dranikpg: I don't want to do it now because its inefficient but lets do it now

src/server/cmd_support.h
```diff
@@ -100,4 +107,52 @@ struct SimpleContext : public AsyncContextInterface, private HopCoordinator {
...
+  Transaction::RunnableType callback;
+};
+
+// Underlying driver (promise) of async
+struct CmdR::Coro {
+  Coro(facade::CmdArgList arg, CommandContext* cmd_cntx) : cmd_cntx{cmd_cntx} {
+  }
+
+  auto await_transform(SingleHopSentinel callback) {
+    return SingleHopWaiter{{}, cmd_cntx, nullptr, callback};
```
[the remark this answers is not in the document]
dranikpg: we support only one suspension point for now

src/server/cmd_support.cc
```diff
@@ -56,4 +56,21 @@ BlockResult HopCoordinator::SingleHop(CommandContext* cmd_cntx, Transaction::Run
   return tx_keepalive_->Blocker();
 }
 
+bool SingleHopWaiter::await_ready() noexcept {
+  return (blocker = SingleHop(cmd_cntx, callback)) == nullptr;
+}
+
+void SingleHopWaiter::await_suspend(std::coroutine_handle<> handle) const noexcept {
+  // TODO: functor calling resume is double indirection
```
romange: what do you mean?

src/server/cmd_support.h
```diff
@@ -100,4 +107,52 @@ struct SimpleContext : public AsyncContextInterface, private HopCoordinator {
...
+  void await_suspend(std::coroutine_handle<> handle) const noexcept;
+  facade::OpStatus await_resume() const noexcept;
+
+  CommandContext* cmd_cntx;
+  BlockResult blocker;
+  Transaction::RunnableType callback;
+};
+
+// Underlying driver (promise) of async
+struct CmdR::Coro {
```
romange, line 127: where do we use this fixture?

src/server/cmd_support.h
```diff
@@ -20,6 +21,12 @@ using BlockResult = util::fb2::EmbeddedBlockingCounter*;
 // No execution was performed, the command is ready to reply
 struct JustReplySentinel {};
 
+// Request to perform a single hop
+using SingleHopSentinel = Transaction::RunnableType;
```
romange, line 25: 1. why do we need SingleHop function? 
2. `SingleHopSentinel` is function ref. is it safe to return an rvalue as SingleHopSentinel?

src/server/cmd_support.h
```diff
@@ -20,6 +21,12 @@ using BlockResult = util::fb2::EmbeddedBlockingCounter*;
...
+SingleHopSentinel SingleHop(auto&& f) {
```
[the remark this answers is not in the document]
romange: that

src/server/cmd_support.cc
dranikpg: Instead of storing a functor, we can store directly the coroutine handle, so it will be resumed. I just didn't do this to keep the diff smaller

src/server/cmd_support.h
dranikpg: Transparently, it is the type managing the coroutine and it is exported via promise_type inside CmdR

src/server/cmd_support.h
```diff
@@ -100,4 +107,52 @@ struct SimpleContext : public AsyncContextInterface, private HopCoordinator {
 // Example: ASYNC_CMD(Get) { ... };
 #define ASYNC_CMD(Name) struct Cmd##Name : public ::dfly::cmd::SimpleContext<Cmd##Name>
 
+// Return type of async command
+struct CmdR {
+  struct Coro;
+  using promise_type = Coro;
```
romange, line 113: ah I see. this is  used by convention.

src/server/cmd_support.h
romange: got it.

src/server/cmd_support.h
dranikpg: I want to make hops explicit. Of course I can make it so that `co_await [](){}` will be a hop, but it will be much more transparent. Also we will need to use some kind of utility type for multi hops - we can change it later

src/server/cmd_support.h
```diff
@@ -100,4 +107,52 @@ struct SimpleContext : public AsyncContextInterface, private HopCoordinator {
...
+  }
+
+  // Return error
+  void return_value(facade::ErrorReply&& err) const noexcept;
+
+  // Conclude command without any error
+  void return_value(std::nullopt_t) const noexcept {
+  }
+
+  // Blank default implmenetations
```
romange:
```suggestion
  // Blank default implementations
```

src/server/cmd_support.h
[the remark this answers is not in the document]
dranikpg: `co_await cmd::SingleHop([&](...){...});`

C++ guarantees that the object is kept alive until the end of the statement, so the temporary functor will only drop once this statement unblocks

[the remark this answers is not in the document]
romange: ok, but:
```
auto coro = cmd::SingleHop([&](...){...});
co_await coro;
```

is UB?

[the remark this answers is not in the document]
romange: ?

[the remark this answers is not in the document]
dranikpg: This sadly is, I will change the inteface co that await_transform expects an r-value and you have no choice

PR #6720, opened 2026-02-24
chore: Skip KNN search if prefilter is empty

### LINKED ISSUE #6712: Don't execute KNN search if there is no prefilter docs found
If prefiltering results phase is empty - no documents matching - we should skip KNN search.

mkaruza:
Closes #6712


vyavdoshenko (dismissed):
I would add additional tests

vyavdoshenko (approved):

dranikpg (approved):

PR #6722, opened 2026-02-24
test: add test for replication HNSW index

BorysTheDev:
Add HNSW search index replication test with network disruptions

Changes:

- Add HnswSearchSeeder with methods for index creation, document seeding, mixed write traffic (insert/update/delete), continuous KNN search queries, and cross-instance verification
- Add test_hnsw_search_replication_with_network_disruptions that creates an HNSW index, runs concurrent traffic and search queries, replicates through a proxy, drops the connection at a random moment (0-10s), then verifies dbsize and KNN search consistency between master and replica


BorysTheDev:
> Looks good, but for extensive testing you need more than 200 documents 🙂

completely agree, but it is for the future

PR #6727, opened 2026-02-24
fix(rdb_load): Fix dry run with shard optimization

dranikpg:
We have this short cut with inline runs with `EngineShard::tlocal()->id  == dest_id` that ignores the dry run flag. I wanted to do it first in CreateObjectOnShard, but we actually don't even have to populate the item, so I moved it further upstream

### DIFF
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -2798,13 +2798,7 @@ void RdbLoader::LoadItemsBuffer(DbIndex db_ind, const ItemsBuf& ib) {
 
   DCHECK(!db_slice.IsCacheMode());
 
-  bool dry_run = absl::GetFlag(FLAGS_rdb_load_dry_run);
-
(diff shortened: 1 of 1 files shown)


src/server/rdb_load.cc
```diff
@@ -2854,6 +2849,9 @@ error_code RdbLoader::LoadKeyValPair(int type, ObjSettings* settings) {
       continue;
     }
 
+    if (dry_run)
```
romange, line 2852: I am not sure if it's correct. I do not remember the logic there but what about 
```
// If the key can be discarded, we must still continue to read the
// object from the RDB so we can read the next key.
```
at above?

dranikpg: The code below only pushes the item and resets `pending_read_.reserve = 0;`, there is no read or IO logic there

PR #6733, opened 2026-02-24
fix: restore HNSW index during replication with different shard counts

vyavdoshenko:
- When master and replica have different shard counts, the HNSW vector index graphs were skipped during replication and rebuilt from scratch. This change remaps `global_id`s during loading, so the serialized HNSW graph can be restored directly, avoiding the expensive rebuild.

Changes:
- Extract `LoadVectorIndexNodes` helper from `RestoreVectorIndex` so nodes can be loaded without immediate restoration
- When shard counts differ, defer HNSW nodes into `PendingHnswNodes` and always store key mappings regardless of shard count
- `RemapForDifferentShardCount` redistributes key mappings to replica shards, assigns new sequential doc_ids, builds an `old_global_id → new_global_id` remap table, and restores the HNSW graph with remapped ids
- Per-index failure tracking via `MarkHnswRestoreFailed` — a failed index falls back to full rebuild without affecting other indices
- Validate `internal_id` bounds before `RestoreFromNodes` to prevent CHECK-crash on corrupted data
- Strengthen replication test: exact KNN result comparison between master and replica, log assertions for restore vs rebuild path

Fixes: #6709

### DIFF
--- a/.github/workflows/cov.yml
+++ b/.github/workflows/cov.yml
@@ -16,17 +16,19 @@ jobs:
       matrix:
         include:
           - container: "ubuntu-dev:20-gcc14"
-            build-type: Release
+            build-type: Debug
             compiler: {cxx: g++, c: gcc}
             cxx_flags: "-fprofile-arcs -ftest-coverage"
-    timeout-minutes: 60
+    timeout-minutes: 300
     container:
       image: ghcr.io/romange/${{ matrix.container }}
+      options: --security-opt seccomp=unconfined --sysctl "net.ipv6.conf.all.disable_ipv6=0"
+      volumes:
+        - /:/hostroot
+        - /mnt:/mnt
       credentials:
         username: ${{ github.repository_owner }}
         password: ${{ secrets.GITHUB_TOKEN }}
-      volumes:
-        - /:/hostroot
 
     steps:
       - uses: actions/checkout@v6
@@ -71,23 +73,38 @@ jobs:
             -L
           pwd
           cd build && pwd
-      - name: Build & Test Coverage
+      - name: Build
         run: |
           cd $GITHUB_WORKSPACE/build
-          ninja
+          echo "-----------------------------"
+          ninja src/all
 
-          # Be permissive with errors, we hit compiler bugs from time to time.
-          ninja test || true
+      - name: Run C++ Unit Tests
+        run: |
+          cd $GITHUB_WORKSPACE/build
+          ctest -V -L DFLY
 
+      - name: Run Python Integration Tests
+        run: |
+          cd $GITHUB_WORKSPACE/build
           export DRAGONFLY_PATH=`pwd`/dragonfly
-          pytest ../tests/dragonfly/ || true
+          pytest ../tests/dragonfly/ --durations=10 --timeout=300 --color=yes --log-cli-level=INFO
 
+      - name: Generate Coverage Report
+        run: |
+          cd $GITHUB_WORKSPACE/build
           lcov -c -d . -o main_coverage.info
           lcov --remove main_coverage.info -o main_coverage.info '/usr/*' '*/_deps/*' '*/third_party/*'
           genhtml main_coverage.info --ignore-errors source --output-directory covout -p $GITHUB_WORKSPACE
           ls ./
           echo ls covout
           ls covout/
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v4
+        with:
+          files: build/main_coverage.info
+          fail_ci_if_error: true
+          token: ${{ secrets.CODECOV_TOKEN }}
       - name: Upload coverage
         uses: actions/upload-artifact@v6
         with:
--- a/src/core/search/hnsw_alg.h
+++ b/src/core/search/hnsw_alg.h
@@ -1404,6 +1404,40 @@ template <typename dist_t> class HierarchicalNSW : public hnswlib::AlgorithmInte
     return result;
   }
 
+  // Brute-force KNN search over a pre-filtered set of label IDs.
+  // Computes distances for all provided IDs and returns the top-k closest, ordered by distance.
+  std::priority_queue<std::pair<dist_t, labeltype>> subsetKnnSearch(
+      const void* query_data, size_t k, const std::vector<labeltype>& ids) const {
+    std::priority_queue<std::pair<dist_t, labeltype>> result;
+
+    if (cur_element_count == 0 || ids.empty() || k == 0)
+      return result;
+
+    for (const auto& label : ids) {
+      auto it = label_lookup_.find(label);
+
+      if (it == label_lookup_.end()) {
+        continue;
+      }
+
+      tableint internal_id = it->second;
+
+      if (isMarkedDeleted(internal_id)) {
+        continue;
+      }
+
+      dist_t dist = fstdistfunc_(query_data, getDataByInternalId(internal_id), dist_func_param_);
+      if (result.size() < k) {
+        result.emplace(dist, label);
+      } else if (dist < result.top().first) {
+        result.pop();
+        result.emplace(dist, label);
+      }
+    }
+
+    return result;
+  }
+
   std::vector<std::pair<dist_t, labeltype>> searchStopConditionClosest(
       const void* query_data, BaseSearchStopCondition& stop_condition,
       BaseFilterFunctor* isIdAllowed = nullptr) const {
--- a/src/core/search/hnsw_index.cc
+++ b/src/core/search/hnsw_index.cc
@@ -144,6 +144,14 @@ struct HnswlibAdapter {
     return QueueToVec(world_.searchKnn(target, k, &filter));
   }
 
+  // Brute-force KNN search over a specific subset of documents.
+  // Computes distances for all provided document IDs and returns the k nearest neighbors.
+  vector<pair<float, GlobalDocId>> SubsetKnn(float* target, size_t k,
+                                             const vector<GlobalDocId>& docs) {
+    MRMWMutexLock lock(&mrmw_mutex_, MRMWMutex::LockMode::kReadLock);
+    return QueueToVec(world_.subsetKnnSearch(target, k, docs));
+  }
+
   HnswIndexMetadata GetMetadata() const {
     MRMWMutexLock lock(&mrmw_mutex_, MRMWMutex::LockMode::kReadLock);
     HnswIndexMetadata metadata;
@@ -541,6 +549,11 @@ std::vector<std::pair<float, GlobalDocId>> HnswVectorIndex::Knn(
     float* target, size_t k, std::optional<size_t> ef,
     const std::vector<GlobalDocId>& allowed) const {
   return adapter_->Knn(target, k, ef, allowed);
+}
+
+std::vector<std::pair<float, GlobalDocId>> HnswVectorIndex::SubsetKnn(
+    float* target, size_t k, const std::vector<GlobalDocId>& docs) const {
+  return adapter_->SubsetKnn(target, k, docs);
 }
 
 void HnswVectorIndex::Remove(GlobalDocId id, const DocumentAccessor& doc, string_view field) {
--- a/src/core/search/hnsw_index.h
+++ b/src/core/search/hnsw_index.h
@@ -58,6 +58,8 @@ class HnswVectorIndex {
                                                  std::optional<size_t> ef) const;
   std::vector<std::pair<float, GlobalDocId>> Knn(float* target, size_t k, std::optional<size_t> ef,
                                                  const std::vector<GlobalDocId>& allowed) const;
+  std::vector<std::pair<float, GlobalDocId>> SubsetKnn(float* target, size_t k,
+                                                       const std::vector<GlobalDocId>& docs) const;
 
   // Get metadata for serialization
   HnswIndexMetadata GetMetadata() const;
--- a/src/core/search/search_test.cc
+++ b/src/core/search/search_test.cc
@@ -15,6 +15,7 @@
 #include <mimalloc.h>
 
 #include <algorithm>
+#include <cmath>
 #include <memory_resource>
 #include <random>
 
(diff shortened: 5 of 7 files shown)


src/server/rdb_load.cc
```diff
@@ -2237,21 +2237,32 @@ error_code RdbLoader::Load(io::Source* src) {
       uint64_t elements_number;
       SET_OR_RETURN(LoadLen(nullptr), elements_number);
 
-      // We can keep the same internal ids if shards count is the same and we can reuse the
-      // data-structures as is, otherwise we must rebuild from scratch
-      // TODO add ability to restore index with different shard count by remapping internal ids
-      const bool should_restore = deserialize_hnsw_index_ && shard_count_ == shard_set->size();
-
-      if (should_restore) {
+      if (!deserialize_hnsw_index_ || shard_count_ == 0) {
```
BorysTheDev: I think shard_count_ can not be 0 at all

vyavdoshenko: This is a guard, I guess it's better to have it.

src/server/rdb_load.cc
```diff
@@ -2997,14 +3000,49 @@ error_code RdbLoader::RestoreVectorIndex(string_view index_key, string_view inde
...
+  return {};
+}
+
+error_code RdbLoader::RestoreVectorIndex(string_view index_key, string_view index_name,
+                                         string_view field_name, uint64_t elements_number) {
+  // Look up the HNSW index in the global registry. It should exist from FT.CREATE in aux.
+  auto hnsw_index = GlobalHnswIndexRegistry::Instance().Get(index_name, field_name);
+  if (!hnsw_index) {
+    LOG(ERROR) << "HNSW index not found for restoration: " << index_key
+               << ". Index will be rebuilt from scratch.";
```
BorysTheDev: if index isn't exist it couldn't be rebuilt

src/server/rdb_load.cc
```diff
@@ -2997,14 +3000,49 @@ error_code RdbLoader::RestoreVectorIndex(string_view index_key, string_view inde
...
+
   if (!nodes.empty()) {
-    // Look up metadata from pending_hnsw_metadata (loaded from AUX field)
     auto metadata = load_context_->FindHnswMetadata(index_name, field_name);
+    if (!metadata) {
+      LOG(ERROR) << "HNSW metadata not found for restoration: " << index_key
+                 << ". Index will be rebuilt from scratch.";
+      load_context_->MarkHnswRestoreFailed(std::string{index_name});
+      return {};
+    }
```
BorysTheDev: why do you need it? It is AI recommendation, but I don't think we need it

src/server/rdb_load.cc
```diff
@@ -2997,14 +3000,49 @@ error_code RdbLoader::RestoreVectorIndex(string_view index_key, string_view inde
...
+    bool valid = true;
+    for (const auto& node : nodes) {
+      if (node.internal_id >= metadata->cur_element_count) {
+        LOG(ERROR) << "HNSW node internal_id " << node.internal_id << " >= cur_element_count "
+                   << metadata->cur_element_count << " for " << index_key
+                   << ". Index will be rebuilt from scratch.";
+        valid = false;
+        break;
+      }
+    }
```
BorysTheDev: why we need to validate our data

src/server/rdb_load.cc
```diff
@@ -2618,6 +2628,9 @@ error_code RdbLoader::HandleAux() {
     uint32_t shard_count;
     if (absl::SimpleAtoi(auxval, &shard_count)) {
       shard_count_ = shard_count;
+      if (load_context_) {
```
BorysTheDev: load_context_ always exists

src/server/rdb_load_context.cc
```diff
@@ -98,6 +102,21 @@ void RdbLoadContext::AddPendingHnswMetadata(PendingHnswMetadata metadata) {
   pending_hnsw_metadata_.emplace_back(std::move(metadata));
 }
 
+void RdbLoadContext::AddPendingHnswNodes(PendingHnswNodes nodes) {
+  util::fb2::LockGuard<util::fb2::Mutex> lk(mu_);
+  pending_hnsw_nodes_.emplace_back(std::move(nodes));
+}
+
+void RdbLoadContext::SetMasterShardCount(uint32_t count) {
+  util::fb2::LockGuard<util::fb2::Mutex> lk(mu_);
```
BorysTheDev: Why do we need a mutex here

src/server/rdb_load_context.cc
```diff
@@ -124,27 +143,205 @@ RdbLoadContext::TakePendingIndexMappings() {
   return result;
 }
 
+std::vector<PendingHnswNodes> RdbLoadContext::TakePendingHnswNodes() {
+  util::fb2::LockGuard<util::fb2::Mutex> lk(mu_);
+  std::vector<PendingHnswNodes> result;
+  std::swap(result, pending_hnsw_nodes_);
+  return result;
```
BorysTheDev: for vector I believe you can do return std::move(pending_hnsw_nodes_)

src/server/rdb_load_context.cc
```diff
@@ -124,27 +143,205 @@ RdbLoadContext::TakePendingIndexMappings() {
...
+  absl::flat_hash_map<uint32_t, absl::flat_hash_map<std::string, size_t>> shard_index_pos;
+
+  for (auto& [master_shard_id, pim_vec] : index_mappings) {
+    for (auto& pim : pim_vec) {
+      auto& remap = remap_tables[pim.index_name];
+      auto& counters = doc_id_counters[pim.index_name];
+
+      for (const auto& [key, old_doc_id] : pim.mappings) {
+        search::GlobalDocId old_gid =
+            search::CreateGlobalDocId(static_cast<search::ShardId>(master_shard_id), old_doc_id);
```
BorysTheDev: we shouldn't have static_cast, it is better to fix the type

src/server/rdb_load_context.cc
```diff
@@ -124,27 +143,205 @@ RdbLoadContext::TakePendingIndexMappings() {
...
+        break;
+      }
+    }
+
+    if (!found_metadata) {
+      LOG(ERROR) << "HNSW metadata not found for deferred restoration: " << pn.index_name << ":"
+                 << pn.field_name << ". Will rebuild from scratch.";
+      failed_remap_indices.insert(pn.index_name);
+      continue;
+    }
```
BorysTheDev: it should be impossible

src/server/rdb_load.cc
vyavdoshenko: Fixed

src/server/rdb_load.cc
vyavdoshenko: Fixed

src/server/rdb_load.cc
vyavdoshenko: Fixed

src/server/rdb_load.cc
vyavdoshenko: Fixed

src/server/rdb_load.cc
vyavdoshenko: Fixed

src/server/rdb_load_context.cc
vyavdoshenko: Fixed

src/server/rdb_load_context.cc
vyavdoshenko: Fixed

src/server/rdb_load_context.cc
vyavdoshenko: Fixed

src/server/rdb_load_context.cc
vyavdoshenko: Fixed

src/server/rdb_load_context.h
```diff
@@ -48,6 +55,13 @@ class RdbLoadContext {
   void AddPendingSynonymCommand(std::string cmd);
   void AddPendingIndexMapping(uint32_t shard_id, PendingIndexMapping mapping);
   void AddPendingHnswMetadata(PendingHnswMetadata metadata);
+  void AddPendingHnswNodes(PendingHnswNodes nodes);
+  void SetMasterShardCount(uint32_t count);
+
+  // Mark a specific index's HNSW restoration as failed. Its key mappings are removed in
+  // PerformPostLoad (after all loaders finish), avoiding a race with concurrent
+  // AddPendingIndexMapping calls. Other indices' mappings are preserved.
+  void MarkHnswRestoreFailed(std::string index_name);
```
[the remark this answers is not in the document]
vyavdoshenko: Fixed

src/server/rdb_load.cc
```diff
@@ -2237,21 +2237,33 @@ error_code RdbLoader::Load(io::Source* src) {
...
+
+        if (shard_count_ == shard_set->size()) {
+          // Same shard count: restore directly.
+          RETURN_ON_ERR(RestoreVectorIndex(index_key, index_name, field_name, elements_number));
+        } else {
+          // Different shard count: load nodes and defer restoration.
+          // Global_ids will be remapped in PerformPostLoad after all key mappings are collected.
+          PendingHnswNodes pending;
+          pending.index_name = index_name;
+          pending.field_name = field_name;
```
BorysTheDev: nit: PendingHnswNodes pending{std::move(index_name), std::move(field_name), {}};

src/server/rdb_load_context.cc
```diff
@@ -124,39 +143,193 @@ RdbLoadContext::TakePendingIndexMappings() {
...
+  absl::flat_hash_map<std::string, absl::flat_hash_map<uint32_t, search::DocId>> doc_id_counters;
+
+  for (const auto& [master_shard_id, pim_vec] : index_mappings) {
+    for (const auto& pim : pim_vec) {
+      auto& shard_remap = remap_table[pim.index_name];
+      auto& counters = doc_id_counters[pim.index_name];
+
+      for (const auto& [key, old_doc_id] : pim.mappings) {
+        ShardId new_shard_id = Shard(key, new_shard_count);
+        search::DocId new_doc_id = counters[new_shard_id]++;
```
BorysTheDev: I'm not sure regarding the external counters for new_doc_id, maybe we need to use counter that is inside shardDocIndex

src/server/rdb_load.cc
vyavdoshenko: Fixed

src/server/rdb_load_context.cc
vyavdoshenko: Comment was added

BorysTheDev: Please check that we can not have a parallel simple index build process, because these IDs are used not only for HNSW. If possible, it is better to move id generation to the index, it will be easier to support the code in the future

src/server/rdb_load_context.cc
```diff
@@ -124,39 +143,197 @@ RdbLoadContext::TakePendingIndexMappings() {
...
+        // Counter starts at 0 for each (index, shard) — equivalent to DocKeyIndex::Add() on a
+        // fresh index (free_ids_ empty → id = last_id_++). DocKeyIndex::Restore() is later called
+        // with these exact (key, doc_id) pairs, so the key_index stays consistent with the
+        // global_ids stored in the remapped HNSW graph.
+        search::DocId new_doc_id = counters[new_shard_id]++;
+        search::GlobalDocId new_gid = search::CreateGlobalDocId(new_shard_id, new_doc_id);
+
+        auto& vec = shard_remap[master_shard_id];
+        if (old_doc_id >= vec.size()) {
+          vec.resize(old_doc_id + 1, kInvalidRemapGid);
```
BorysTheDev: could it be that every iteration old_doc_id >= vec.size() ? in this case the complexity is quite crazy

src/server/rdb_load_context.cc
```diff
@@ -124,39 +143,197 @@ RdbLoadContext::TakePendingIndexMappings() {
...
-      for (const auto& pim : it->second) {
-        if (auto* index = es->search_indices()->GetIndex(pim.index_name); index) {
-          index->RestoreKeyIndex(pim.mappings);
-          VLOG(1) << "Restored " << pim.mappings.size() << " key mappings for index "
-                  << pim.index_name << " on shard " << es->shard_id();
+  // When shard counts differ, remap HNSW global_ids and redistribute key mappings on-the-fly.
+  bool shard_count_differs = master_shards != 0 && master_shards != shard_set->size();
+
+  if (shard_count_differs && !index_mappings.empty()) {
+    auto remap_table =
```
BorysTheDev: Instead of a global table, can we create per-shard tables, and drop items when we use them, in this case we will decrease memory consumption

src/server/rdb_load_context.cc
vyavdoshenko: No, it doesn't happen.

src/server/rdb_load_context.cc
vyavdoshenko: Fixed

src/server/rdb_load_context.cc
vyavdoshenko: Fixed

src/server/rdb_load_context.h
```diff
@@ -59,12 +68,28 @@ class RdbLoadContext {
  private:
   std::vector<std::string> TakePendingSynonymCommands();
   absl::flat_hash_map<uint32_t, std::vector<PendingIndexMapping>> TakePendingIndexMappings();
+  std::vector<PendingHnswNodes> TakePendingHnswNodes();
+
+  // Pre-distributed key mappings: target_shard_id -> index_name -> [(key, new_doc_id)].
+  using PerShardMappings = absl::flat_hash_map<
```
BorysTheDev: We have a fixed number of shards, so we use a vector or an array for such cases

src/server/rdb_load_context.cc
```diff
@@ -124,40 +143,209 @@ RdbLoadContext::TakePendingIndexMappings() {
...
+      const auto& remap_vec = shard_it->second;
+
+      for (const auto& [key, old_doc_id] : pim.mappings) {
+        if (old_doc_id >= remap_vec.size())
+          continue;
+        search::GlobalDocId new_gid = remap_vec[old_doc_id];
+        if (new_gid == kInvalidRemapGid)
+          continue;
+        auto [new_shard_id, new_doc_id] = search::DecomposeGlobalDocId(new_gid);
+        per_shard[new_shard_id][pim.index_name].emplace_back(key, new_doc_id);
```
BorysTheDev: doc_ids is a sequence, I think we don't need to store it directly and can use simple vector as we do for shard_index ionstead of set

src/server/rdb_load_context.cc
```diff
@@ -124,40 +143,209 @@ RdbLoadContext::TakePendingIndexMappings() {
...
+  PerShardMappings per_shard;
+  for (const auto& [master_shard_id, pim_vec] : index_mappings) {
+    for (const auto& pim : pim_vec) {
+      auto idx_it = remap_table.find(pim.index_name);
+      if (idx_it == remap_table.end())
+        continue;
+      auto shard_it = idx_it->second.find(master_shard_id);
+      if (shard_it == idx_it->second.end())
+        continue;
+      const auto& remap_vec = shard_it->second;
```
BorysTheDev: let's create a small function to make code readable

src/server/rdb_load_context.h
vyavdoshenko: Fixed

src/server/rdb_load_context.cc
vyavdoshenko: Fixed

src/server/rdb_load_context.cc
vyavdoshenko: Done

Issue #6234, opened 2025-12-17
Fix missing counts in page usage stats / manual defragmentation cmd output


abhijat:
Since we moved from bloom filter to HLL in page usage stats the counts seem to be off, for example page reserved for malloc should be 1 but is shown as zero (unless something changed within mimalloc).

This is not a big issue but might be hiding some problems with reporting of results, it should be debugged and fixed.

PR #6737, opened 2026-02-25
core: Fix page usage accounting

abhijat:
Previously the count was updated in exclusive branches ie update count for flag A, or B or C etc

But there can be multiple bits set on the same flag. So the code is updated to count using AND branches instead of OR

FIXES https://github.com/dragonflydb/dragonfly/issues/6234

### DIFF
--- a/.github/actions/fuzzing/action.yml
+++ b/.github/actions/fuzzing/action.yml
@@ -6,6 +6,11 @@ inputs:
     description: "Fuzzing mode: 'smoke' (stop on first crash) or 'long' (collect all crashes)"
     required: true
     type: string
+  target:
+    description: "Fuzz target: 'resp' or 'memcache'"
+    required: false
+    default: 'resp'
+    type: string
   duration-minutes:
     description: "Fuzzing duration in minutes"
     required: true
@@ -15,7 +20,7 @@ inputs:
     required: true
     type: string
   extra-seeds-dir:
-    description: "Directory with additional .resp seed files (initial fuzzer inputs) to merge into the corpus"
+    description: "Directory with additional seed files (initial fuzzer inputs) to merge into the corpus"
     required: false
     default: ''
   focus-commands:
@@ -59,12 +64,13 @@ runs:
       if: ${{ inputs.extra-seeds-dir != '' }}
       run: |
         EXTRA_DIR="${{ inputs.extra-seeds-dir }}"
-        SEEDS_DIR="fuzz/seeds/resp"
+        SEEDS_DIR="fuzz/seeds/${{ inputs.target }}"
 
-        if [ -d "$EXTRA_DIR" ] && [ -n "$(ls -A "$EXTRA_DIR"/*.resp 2>/dev/null)" ]; then
-          COUNT=$(ls "$EXTRA_DIR"/*.resp 2>/dev/null | wc -l)
+        # Copy only seed files, skip metadata like focus_commands.json
+        COUNT=$(find "$EXTRA_DIR" -maxdepth 1 -type f ! -name '*.json' 2>/dev/null | wc -l)
+        if [ "$COUNT" -gt 0 ]; then
           echo "Merging ${COUNT} targeted seeds into corpus"
-          cp "$EXTRA_DIR"/*.resp "$SEEDS_DIR/"
+          find "$EXTRA_DIR" -maxdepth 1 -type f ! -name '*.json' -exec cp -t "$SEEDS_DIR/" {} +
         else
           echo "No targeted seed files to merge"
         fi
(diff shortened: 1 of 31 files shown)


src/core/page_usage/page_usage_stats.cc
```diff
@@ -148,27 +148,29 @@ PageUsage::UniquePages::~UniquePages() {
   hdr_close(page_usage_hist);
 }
 
-void PageUsage::UniquePages::AddStat(mi_page_usage_stats_t stat) {
+void PageUsage::UniquePages::AddStat(mi_page_usage_stats_t stat) {  // NOLINT should not be const
   const auto data = reinterpret_cast<const unsigned char*>(&stat.page_address);
```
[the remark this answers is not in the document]
abhijat: pointless noise

src/core/page_usage/page_usage_stats.cc
```diff
@@ -148,27 +148,29 @@ PageUsage::UniquePages::~UniquePages() {
...
   constexpr size_t size = sizeof(stat.page_address);
   pfadd_dense(pages_scanned, data, size);
-  if (stat.flags == MI_DFLY_PAGE_BELOW_THRESHOLD) {
+
+  if (stat.flags & MI_DFLY_PAGE_BELOW_THRESHOLD) {
     pfadd_dense(pages_marked_for_realloc, data, size);
```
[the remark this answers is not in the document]
abhijat: `MI_DFLY_PAGE_BELOW_THRESHOLD` is exclusive, only set when no other bit is set.

src/core/page_usage/page_usage_stats.cc
```diff
@@ -148,27 +148,29 @@ PageUsage::UniquePages::~UniquePages() {
...
-      pfadd_dense(pages_above_threshold, data, size);
-      const double perc = static_cast<double>(stat.used) / static_cast<double>(stat.capacity);
-      hdr_record_value(page_usage_hist, perc * 100);
-    }
+  }
+  if (stat.flags & MI_DFLY_PAGE_FULL) {
+    pfadd_dense(pages_full, data, size);
+  }
+  if (stat.flags & MI_DFLY_HEAP_MISMATCH) {
+    pfadd_dense(pages_with_heap_mismatch, data, size);
```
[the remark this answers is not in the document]
abhijat: its overkill to test this, this is mostly just presentation logic

src/core/page_usage/page_usage_stats.cc
```diff
@@ -148,27 +148,29 @@ PageUsage::UniquePages::~UniquePages() {
...
+  }
+  if (stat.flags & MI_DFLY_PAGE_USED_FOR_MALLOC) {
+    pfadd_dense(pages_reserved_for_malloc, data, size);
+  }
+  if (stat.flags == 0) {
+    // No special flags means the page is above the threshold but not full - record usage for
+    // histogram. This allows tuning the threshold for future commands.
+    pfadd_dense(pages_above_threshold, data, size);
+    const double perc = static_cast<double>(stat.used) / static_cast<double>(stat.capacity);
```
glevkovich: nit:
same effect, and looks cleaner (implicit promotion):
`const double percentage = (100.0 * stat.used) / stat.capacity;`

src/core/page_usage/page_usage_stats.cc
glevkovich: nit:
not  a big deal, but you can consider:
```
auto record_page = [&](auto& counter) {
    pfadd_dense(counter, data, size);
  };
```
and then you save the repetitions:
```

if (stat.flags & MI_DFLY_PAGE_BELOW_THRESHOLD) {
    record_page(pages_marked_for_realloc);
  }
  if (stat.flags & MI_DFLY_PAGE_FULL) {
    record_page(pages_full);
  }
....
```

glevkovich (dismissed):
LGTM, left 2 small comments.

src/core/page_usage/page_usage_stats.cc
abhijat: fixed

src/core/page_usage/page_usage_stats.cc
abhijat: yeah this looks better, changed

abhijat:
> Please add a unit test with combined flags to cover the new logic.

Initially I thought it would be hard to test because mimalloc is kind of hard to predict in a test case so I skipped it, but now I realize for this case there is actually no need to check allocation, the stats can be fed directly (another old test actually does that already). 

I have added a test which doesn't touch mimalloc but just adds some dummy stats with mixed flags.

abhijat:
error looks to be unrelated but is a new crash

https://github.com/dragonflydb/dragonfly/issues/6742

PR #6740, opened 2026-02-25
fix: replication for data loaded by DFLY LOAD command

### LINKED ISSUE #6739: support replication for data loaded by DFLY LOAD command
Our replication mechanism is based on the journal updates. 

In most cases, every command set/update/delete key is added into the journal via Transaction::LogAutoJournalOnShard() call at the end of the transaction; in other cases, it is a direct RecordJournal() call in the command.

Problem: During loading data via the DFLY LOAD command, we write keys directly into the db_slice without journaling them, and because the replica is in the STABLE_SYNC phase, we track only journal updates.

Possible solutions:
- restart the replication process, it will trigger FULL SYNC, so all data in the  db_slices will be replicated
- add explicit recording data into the journal during RDB loading

BorysTheDev:
fixes: #6739


src/server/dflycmd.cc
```diff
@@ -667,6 +667,26 @@ void DflyCmd::Load(CmdArgList args, CommandContext* cmd_cntx) {
...
 
+  // After loading, cancel all active replicas to force them to reconnect with
+  // a full sync. This is necessary because the loaded data bypasses the journal,
+  // so replicas would not receive the new keys through partial replication.
+  // We advance the journal LSN and clear the ring buffer to ensure that reconnecting
+  // replicas cannot do a partial sync (their old LSN won't match the new one).
+  shard_set->RunBriefInParallel([](EngineShard* shard) {
+    if (shard->journal())
+      journal::StartInThreadAtLsn(journal::GetLsn() + 1);
+  });
```
kostasrim: This is dangerous. Assume you had a registered replica before `DFLY LOAD`.

You loaded the snapshot on line `661`. After that line `other threads` can `write to the journal` so your code here is actually a `race condition`. If the `callback` in `RunBriefInParallel` runs `first` in all `proactors` then everything is fine. However, if `other` commands run `before` the `callback` runs on the proactors you just `corrupted the lsn counting and state`.

tests/dragonfly/replication_test.py
```diff
@@ -4175,3 +4175,39 @@ async def test_hnsw_search_replication_with_network_disruptions(
...
+
+    await c_master.execute_command("SAVE", "DF", dbfilename)
+
+    await c_master.execute_command("FLUSHALL")
+
+    await c_replica.execute_command("REPLICAOF", "localhost", str(master.port))
+    await wait_available_async(c_replica)
+
+    await c_master.execute_command("DFLY", "LOAD", f"{dbfilename}-summary.dfs")
+
```
kostasrim: This test can be improved. The main issue is that it calls:

1. Debug populate
2. Save
3. Flushes the datastore (so everything is empty)
4. Calls replica of -> nothing gets replicated empty datastore
5. Calls DFLY LOAD -> loads the new snapshot

Wouldn't it be better if:

1. You actually had some data in both master/replica before you load a new snapshot
2. Stream data while trying to LOAD the snapshot via DFLY LOAD. This will cause the bug I explained in the other comment to come up

src/server/dflycmd.cc
BorysTheDev: The idea was to avoid a partial sync situation, and because we have a gap in lsn the replication should restart.
Anyway, I will try to apply your suggestion to the test

kostasrim: The idea to avoid partial sync is fine. The issue here is that we might restart the journal on the wrong lsn because other transactions can run between the Load() and the `RunBriefInParallel`

BorysTheDev: But it is global transaction

kostasrim: It is not a global transaction. It's not being scheduled. Just because a command has `CO::GLOBAL` it does not mean all paths actually schedule a global transaciton. In other words, `DFLY LOAD` is *not* transactional

src/server/dflycmd.cc
```diff
@@ -667,6 +667,26 @@ void DflyCmd::Load(CmdArgList args, CommandContext* cmd_cntx) {
...
   }
 
+  // After loading, cancel all active replicas to force them to reconnect with
+  // a full sync. This is necessary because the loaded data bypasses the journal,
+  // so replicas would not receive the new keys through partial replication.
+  // We advance the journal LSN and clear the ring buffer to ensure that reconnecting
+  // replicas cannot do a partial sync (their old LSN won't match the new one).
+  shard_set->RunBriefInParallel([](EngineShard* shard) {
+    if (shard->journal())
+      journal::StartInThreadAtLsn(journal::GetLsn() + 1);
```
romange: also not clear why advancing to next lsn. I would expect to reset it completely (start from 0, flush the buffer)

src/server/dflycmd.cc
```diff
@@ -667,6 +667,26 @@ void DflyCmd::Load(CmdArgList args, CommandContext* cmd_cntx) {
...
+  });
+
+  ReplicaInfoMap pending;
+  {
+    util::fb2::LockGuard lk(mu_);
+    pending = std::move(replica_infos_);
+  }
+
+  for (auto& [_, replica_ptr] : pending) {
+    replica_ptr->Cancel();
```
romange: I agree with Kostas - I would expect that Cancel does some per shard-thread reset underneath - this is where you should flush partial sync state as well. No need for multiple hops that are not atomic and add subtle bugs.

romange: Also, why not cancel before starting the load?

romange:
@BorysTheDev  what's the status of this PR?

BorysTheDev:
> @BorysTheDev what's the status of this PR?

This PR is incorrect. I have postponed this task, but I remember it.

BorysTheDev:
augment review

src/server/dflycmd.cc
```diff
@@ -997,7 +997,7 @@ bool DflyCmd::CheckReplicaStateOrReply(const ReplicaInfo& repl_info, SyncState e
   return true;
 }
 
-void DflyCmd::Shutdown() {
+void DflyCmd::CancelAllReplicas() {
```
kostasrim: Plz rename this back to Shutdown(). It `shuts down` the replica. Then the replica identifies that the master link is broken and the replica reconnects automatically. In other words, `Shutdown` is the right name here.

src/server/dflycmd.cc
```diff
@@ -1009,6 +1009,10 @@ void DflyCmd::Shutdown() {
   }
 }
 
+void DflyCmd::Shutdown() {
```
kostasrim: Remove this wrapper plz

src/server/server_family.cc
```diff
@@ -1445,6 +1445,17 @@ std::optional<fb2::Future<GenericError>> ServerFamily::Load(const std::string& p
...
+
+      // Cancel all replicas first (stops journal streamers), then reset the
+      // journal to invalidate old partial sync buffers. The loaded data bypasses
+      // the journal, so replicas in STABLE_SYNC would miss it.
+      // Both operations run while still in LOADING state, so no commands can
+      // execute and no journal writes can race with the reset.
+      dfly_cmd_->CancelAllReplicas();
+      shard_set->RunBriefInParallel([](EngineShard* shard) {
+        if (shard->journal())
+          journal::StartInThreadAtLsn(1);
```
kostasrim: This a bug. Even copilot found it and you resolved it 🤣 The correct fix here is to call `journal::ClearBuffer();` instead of reseting the lsn to 1.

tests/dragonfly/cluster_test.py
```diff
@@ -3835,3 +3835,62 @@ async def test_cluster_config_slot_overflow_doesnt_crash(df_factory: DflyInstanc
     # CONFIG must return an error (not crash), MYID must still work
     assert isinstance(results[0], Exception)
     assert results[1] == node_id
+
+
+@dfly_args({"cluster_mode": "yes"})
+async def test_cluster_snapshot_load_replication(df_factory: DflyInstanceFactory):
```
kostasrim: This test is redundant and it does not more logic than the replication one (you can ask copilot to verify what I write here). 

1. Remove this test.
2. Keep replication test only as it suffices for this use case

kostasrim:
We can reduce the lines of changes in this PR + small fixes

src/server/server_family.cc
BorysTheDev: there is no journal::ClearBuffer() in the code. and previously I did journal::StartInThreadAtLsn(shard_journal->GetLsn() + 1); and got comment to make it 1

src/server/server_family.cc
```diff
@@ -1445,6 +1445,13 @@ std::optional<fb2::Future<GenericError>> ServerFamily::Load(const std::string& p
     } else {
       load_context->PerformPostLoad(&service_);
       LOG(INFO) << "Load finished, num keys read: " << aggregated_result->keys_read;
+
+      // Loaded data bypasses the journal, so force replicas into full sync.
+      dfly_cmd_->Shutdown();
```
dranikpg: Shutdown is a tricky name, it just cancels all replicas. But if it ever does something non-restorable, this code will break

BorysTheDev: I wanted to have another name, @kostasrim doesn't like this idea

romange: I agree `DflyCmd::Shutdown` should be renamed to `BreakReplication` or `CancelReplicas` or something like that

kostasrim (approved):
💯
 
 🚢 🇮🇹

Issue #6738, opened 2026-02-25
Blocking commands in pipeline do not flush their replies


qlimenoque:
**Describe the bug**

Blocked `BRPOPLPUSH` clients are not unblocked when `LPUSH`/`RPUSH` happens inside a Lua script (`EVALSHA`/`EVAL`). They wait for the full timeout instead of waking immediately. Redis handles this correctly.

Discovered while running [n8n](https://github.com/n8n-io/n8n) in queue mode with Dragonfly as the queue backend. n8n uses [Bull](https://github.com/OptimalBits/bull) under the hood, which enqueues jobs via `EVALSHA` (Lua script doing `RPUSH` + `ZADD` + `PUBLISH`) and dequeues via `BRPOPLPUSH` with a 5s timeout. Workers are never woken by the Lua push - they only pick up jobs after the timeout expires, causing 15-50s webhook response times instead of ~1.5s.

Reproduces with both single and multiple blocked clients. Persists across v1.35.0 and v1.37.0, with 1 or 2 proactor threads.

Likely affects any queue library using blocking pops + Lua enqueue (Bull, BullMQ, Sidekiq, asynq, etc.)

There is a related [TODO in the test suite](https://github.com/dragonflydb/dragonfly/blob/main/src/server/list_family_test.cc#L904-L906) (`src/server/list_family_test.cc`)  ~~acknowledging a known gap~~: "TODO: there is a bug here. we do not wake the dest shard, when source is awaked which prevents the atomicity and causes the first bug as well." 

**To Reproduce**

The bug manifests under real workload conditions (multiple concurrent connections, replication, persistent state). A trivial `redis-cli` script may not reliably trigger it locally. The most reliable reproduction is with an actual Bull/BullMQ workload.

1. Deploy Dragonfly with `--default_lua_flags=allow-undeclared-keys` and `replicas: 2`
2. Run a Bull-based application (e.g. n8n in queue mode) with workers using `BRPOPLPUSH`
3. Trigger a job enqueue (which uses `EVALSHA` containing `RPUSH`)
4. Observe via `MONITOR` that blocked clients are not woken - they poll every N seconds (the BRPOPLPUSH timeout)

**Expected behavior**

Blocked clients wake immediately after the Lua script completes (~microseconds), same as Redis.

**Environment:**
- OS: Linux 6.6.113+ x86_64
- Kernel: 6.6.113+
- Containerized: Kubernetes (GKE, Dragonfly Operator)
[...]

PR #6743, opened 2026-02-26
fix(server): flush replies on encountering blocking command in pipeline

abhijat:
When a blocking command is encountered while processing a pipeline, any pending replies are flushed, so that the blocking command (which might wait for an event until timeout) does not block the replies from any previous commands which were lined up and now have data to report. 

The new test based on https://github.com/dragonflydb/dragonfly/commit/beded93520ec60978e1167bdcb9484a576059b6a passes with the change and fails without

FIXES https://github.com/dragonflydb/dragonfly/issues/6738

### DIFF
--- a/src/core/page_usage/page_usage_stats.cc
+++ b/src/core/page_usage/page_usage_stats.cc
@@ -148,27 +148,30 @@ PageUsage::UniquePages::~UniquePages() {
   hdr_close(page_usage_hist);
 }
 
-void PageUsage::UniquePages::AddStat(mi_page_usage_stats_t stat) {
+void PageUsage::UniquePages::AddStat(mi_page_usage_stats_t stat) {  // NOLINT should not be const
   const auto data = reinterpret_cast<const unsigned char*>(&stat.page_address);
-  constexpr size_t size = sizeof(stat.page_address);
-  pfadd_dense(pages_scanned, data, size);
-  if (stat.flags == MI_DFLY_PAGE_BELOW_THRESHOLD) {
-    pfadd_dense(pages_marked_for_realloc, data, size);
-  } else {
-    if (stat.flags & MI_DFLY_PAGE_FULL) {
-      pfadd_dense(pages_full, data, size);
-    } else if (stat.flags & MI_DFLY_HEAP_MISMATCH) {
-      pfadd_dense(pages_with_heap_mismatch, data, size);
-    } else if (stat.flags & MI_DFLY_PAGE_USED_FOR_MALLOC) {
-      pfadd_dense(pages_reserved_for_malloc, data, size);
-    } else {
-      // We record usage only for pages which have usage above the given threshold but which are not
-      // full. This allows tuning the threshold for future commands. This also excludes full pages,
-      // so the only pages here have: threshold < usage% < 100
-      pfadd_dense(pages_above_threshold, data, size);
-      const double perc = static_cast<double>(stat.used) / static_cast<double>(stat.capacity);
-      hdr_record_value(page_usage_hist, perc * 100);
-    }
+
+  auto record = [&data](HllBufferPtr ctr) { pfadd_dense(ctr, data, sizeof(stat.page_address)); };
+
+  record(pages_scanned);
+
+  if (stat.flags & MI_DFLY_PAGE_BELOW_THRESHOLD) {
+    record(pages_marked_for_realloc);
+  }
+  if (stat.flags & MI_DFLY_PAGE_FULL) {
+    record(pages_full);
+  }
+  if (stat.flags & MI_DFLY_HEAP_MISMATCH) {
+    record(pages_with_heap_mismatch);
+  }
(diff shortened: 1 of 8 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -838,6 +838,10 @@ void Connection::RegisterBreakHook(BreakerCb breaker_cb) {
   breaker_cb_ = std::move(breaker_cb);
 }
 
+void Connection::FlushReplies() {  // NOLINT must not be const due to flush side effect
+  reply_builder_->Flush();
```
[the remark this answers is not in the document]
abhijat: added dcheck

tests/dragonfly/connection_test.py
```diff
@@ -1651,3 +1647,61 @@ async def test_tls_partial_header_read(
...
+
+    # Wait for the first RESP response and measure latency
+    t0 = time.monotonic()
+    total_timeout = blpop_timeout * num_pipeline + 5
+    response = await asyncio.wait_for(conn_reader.readline(), timeout=total_timeout)
+    first_response_time = time.monotonic() - t0
+
+    assert response == b"*2\r\n", f"Expected BLPOP success, got {response!r}"
+
+    writer.close()
```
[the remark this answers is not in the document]
abhijat: fixed

src/server/main_service.cc
```diff
@@ -1514,6 +1514,9 @@ DispatchResult Service::DispatchCommand(facade::ParsedArgs args, facade::ParsedC
 
   if (dfly_cntx->async_dispatch && cid->IsBlocking()) {
     ++ServerState::tlocal()->stats.blocking_commands_in_pipelines;
+    if (auto* conn = cmd_cntx->conn()) {
```
romange: I think we always have `cmd_cntx->conn()` so this condition is redundant, imho

tests/dragonfly/connection_test.py
```diff
@@ -1651,3 +1647,63 @@ async def test_tls_partial_header_read(
...
+
+    # Build a raw RESP pipeline of N BLPOP commands, raw TCP to avoid
+    # client-library buffering that could mask the flush timing issue.
+    def encode_resp_command(*args):
+        encoded_args = [str(a).encode() for a in args]
+        header = f"*{len(encoded_args)}\r\n".encode()
+        body = b"".join(f"${len(a)}\r\n".encode() + a + b"\r\n" for a in encoded_args)
+        return header + body
+
+    pipeline_data = b"".join(
```
romange: as discussed internally, lets reproduce it as well on non-blocking command being blocked by the subsequent blocking command.

romange:
Some of the bullmq tests are currently failing and they are filtered out:
https://github.com/dragonflydb/dragonfly/actions/runs/22431986371/workflow#L75

Can you please also check if this fix remove some of the issues in those tests? Does not have to be part of this PR.

tests/dragonfly/connection_test.py
abhijat: I have added now two non blocking commands before the blocking ones in the pipeline

abhijat:
> Some of the bullmq tests are currently failing and they are filtered out: https://github.com/dragonflydb/dragonfly/actions/runs/22431986371/workflow#L75
> 
> Can you please also check if this fix remove some of the issues in those tests? Does not have to be part of this PR.

Interestingly, I tried the bullmq run with nothing filtered out (all grep-invert removed) 

```
NODE_OPTIONS="--no-experimental-strip-types" BULLMQ_TEST_PREFIX={b} yarn test
```

and I get just two failures, and those failures are with OR without this PR. So either the filtering is out of date, or this change does not affect those tests. I think both are true.

These are the only tests that fail with flushing/without flushing:

```
  662 passing (5m)
  3 pending
  2 failing

  1) Job Scheduler
       should create job schedulers with different cron patterns:
     AssertionError: expected [ { key: 'fifth', …(6) }, …(4) ] to deep include { key: 'second', name: 'second', …(5) }
      at Context.<anonymous> (tests/test_job_scheduler.ts:355:20)
      at processTicksAndRejections (node:internal/process/task_queues:105:5)

  2) repeat
       should get repeatable jobs with different cron pattern:

      AssertionError: expected 4 to deeply equal 5
      + expected - actual

      -4
      +5

      at Context.<anonymous> (tests/test_repeat.ts:247:25)
      at processTicksAndRejections (node:internal/process/task_queues:105:5)

```

ETA: this might have been my local env. will try this in CI instead https://github.com/dragonflydb/dragonfly/actions/runs/22477399190

abhijat:
bullmq run with these changes and no filters:

https://github.com/dragonflydb/dragonfly/actions/runs/22477399190/job/65107233819

~~everything passes. I think it might have been true even before this PR but I will create separate PR to remove the filters.~~ forgot to remove filters - will upddate

romange:
No need to remove filters in this PR. I think they fail sporadically so maybe we need to run them like 20 times with and without to see if there are differences

PR #6747, opened 2026-02-26
chore: upload coverage

kostasrim:
Uploads coverage similar to helio

### DIFF
--- a/.github/workflows/cov.yml
+++ b/.github/workflows/cov.yml
@@ -104,6 +104,7 @@ jobs:
         with:
           files: build/main_coverage.info
           fail_ci_if_error: true
+          token: ${{ secrets.CODECOV_TOKEN }}
       - name: Upload coverage
         uses: actions/upload-artifact@v6
         with:


.github/workflows/cov.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       matrix:
         include:
           - container: "ubuntu-dev:20-gcc14"
-            build-type: Release
+            build-type: Debug
```
romange, line 19: curious - why Debug? coverage does not work with Release?

kostasrim: It works but my two cents is that it should be less effective since the release optimizations remove a lot of the code

kostasrim: You have debug in helio so I opted for the same 🤷

kostasrim:
https://github.com/dragonflydb/dragonfly/actions/runs/22483886310 let's see!

romange (approved):

PR #6752, opened 2026-02-27
feat(server): Get and set byte at index position in string object

mkaruza:
Added get and set function that read single byte from index position in string object. Added pack and unpack byte functionality for ascii encoding.

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -1750,8 +1750,8 @@ bool CompactObj::StrEncoding::DecodeByte(std::string_view blob, size_t idx, uint
       break;
     case ASCII1_ENC:
     case ASCII2_ENC:
-      detail::ascii_unpack_byte(reinterpret_cast<const uint8_t*>(blob.data()), decoded_len, idx,
-                                dest);
+      *dest = detail::ascii_unpack_byte(reinterpret_cast<const uint8_t*>(blob.data()), decoded_len,
+                                        idx);
       break;
     case HUFFMAN_ENC: {
       std::string decoded_huff_string(decoded_len, 0);
--- a/src/core/compact_object_test.cc
+++ b/src/core/compact_object_test.cc
@@ -352,8 +352,7 @@ TEST_F(CompactObjectTest, AsciiPackByte) {
 
     // Verify initial pack/unpack round-trip at byte level.
     for (size_t i = 0; i < len; ++i) {
-      uint8_t got = 0;
-      detail::ascii_unpack_byte(packed.data(), len, i, &got);
+      uint8_t got = detail::ascii_unpack_byte(packed.data(), len, i);
       ASSERT_EQ(static_cast<uint8_t>(original[i]), got) << "len=" << len << " offset=" << i;
     }
 
@@ -366,16 +365,14 @@ TEST_F(CompactObjectTest, AsciiPackByte) {
       detail::ascii_pack_byte(modified.data(), len, i, new_val);
 
       // The modified byte should read back correctly.
-      uint8_t got = 0;
-      detail::ascii_unpack_byte(modified.data(), len, i, &got);
+      uint8_t got = detail::ascii_unpack_byte(modified.data(), len, i);
       EXPECT_EQ(new_val, got) << "len=" << len << " set offset=" << i;
 
       // All other bytes should be unchanged.
       for (size_t j = 0; j < len; ++j) {
         if (j == i)
           continue;
-        uint8_t other = 0;
-        detail::ascii_unpack_byte(modified.data(), len, j, &other);
+        uint8_t other = detail::ascii_unpack_byte(modified.data(), len, j);
         EXPECT_EQ(static_cast<uint8_t>(original[j]), other)
             << "len=" << len << " set offset=" << i << " check offset=" << j;
       }
@@ -390,8 +387,7 @@ TEST_F(CompactObjectTest, AsciiPackByte) {
         expected[i] = '\0';
       }
       for (size_t i = 0; i < len; ++i) {
-        uint8_t got = 0;
-        detail::ascii_unpack_byte(zeroed.data(), len, i, &got);
+        uint8_t got = detail::ascii_unpack_byte(zeroed.data(), len, i);
         EXPECT_EQ(0, got) << "len=" << len << " zero check offset=" << i;
       }
     }
@@ -403,8 +399,7 @@ TEST_F(CompactObjectTest, AsciiPackByte) {
         detail::ascii_pack_byte(maxed.data(), len, i, 0x7F);
       }
       for (size_t i = 0; i < len; ++i) {
-        uint8_t got = 0;
-        detail::ascii_unpack_byte(maxed.data(), len, i, &got);
+        uint8_t got = detail::ascii_unpack_byte(maxed.data(), len, i);
         EXPECT_EQ(0x7F, got) << "len=" << len << " max check offset=" << i;
       }
     }
--- a/src/core/detail/bitpacking.cc
+++ b/src/core/detail/bitpacking.cc
@@ -270,7 +270,7 @@ void ascii_unpack(const uint8_t* bin, size_t ascii_len, char* ascii) {
   }
 }
 
-void ascii_unpack_byte(const uint8_t* bin, size_t ascii_len, size_t idx, uint8_t* dest) {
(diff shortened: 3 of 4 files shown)


src/core/detail/bitpacking.cc
```diff
@@ -270,6 +270,52 @@ void ascii_unpack(const uint8_t* bin, size_t ascii_len, char* ascii) {
   }
 }
 
+void ascii_unpack_byte(const uint8_t* bin, size_t ascii_len, size_t idx, uint8_t* dest) {
+  if (idx >= ascii_len) {
```
romange: I do not like these things. The contract should be clear. In this case, add "idx must be within the ascii_len." in the header file.

src/core/detail/bitpacking.cc
```diff
@@ -270,6 +270,52 @@ void ascii_unpack(const uint8_t* bin, size_t ascii_len, char* ascii) {
...
+  }
+
+  // Unpack ascii group and return byte at idx
+  char buf[8];
+  ascii_unpack(bin + group * 7, 8, buf);
+  *dest = buf[idx_in_group];
+}
+
+void ascii_pack_byte(uint8_t* bin, size_t ascii_len, size_t idx, uint8_t val) {
+  if (idx >= ascii_len) {
```
romange: same

src/core/detail/bitpacking.h
```diff
@@ -21,6 +21,10 @@ bool validate_ascii_fast(const char* src, size_t len);
 void ascii_unpack(const uint8_t* bin, size_t ascii_len, char* ascii);
 void ascii_unpack_simd(const uint8_t* bin, size_t ascii_len, char* ascii);
 
+// pack/unpack byte at ascii encoded string
```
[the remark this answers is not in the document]
romange: +1 to that. The comments are not clear. I have not read copilot's but they should be more detailed.

src/core/compact_object.cc
```diff
@@ -1648,6 +1785,38 @@ size_t CompactObj::StrEncoding::Decode(std::string_view blob, char* dest) const
...
 }
 
+bool CompactObj::StrEncoding::DecodeByte(std::string_view blob, size_t idx, uint8_t* dest) const {
+  if (blob.empty()) {
+    *dest = 0;
+    return false;
+  }
+  size_t decoded_len = DecodedSize(blob);
+  if (idx >= decoded_len) {
+    LOG(INFO) << "Index out of bounds for string: " << idx << " >= " << decoded_len;
```
romange: again,  using `LOG(INFO)` for this is dangerous.

romange: what is the contract? can idx be  => decoded_len ?

src/core/compact_object.h
```diff
@@ -366,6 +368,10 @@ class CompactObj {
   }
 
   uint8_t GetFirstByte() const;
+  void GetByteAtIndex(size_t idx, uint8_t* res) const;
+  // Returns a pair of booleans: {success, in_place}. success is false if offset is out of bounds
+  // in_place is true if the byte was set without needing to rewrite the string.
```
romange, line 375: curious, why do you need `in_place` return value?
Also, what is precondition for idx? should `idx < Size()` or it is expected it to be out of bounds?

src/core/detail/bitpacking.h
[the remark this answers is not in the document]
mkaruza: Added

src/core/detail/bitpacking.cc
mkaruza: Added comment for ascii_unpack_byte and ascii_pack_byte.

src/core/compact_object.cc
mkaruza: Function returns false if index is out of bounds of decoded size.

src/core/compact_object.h
mkaruza: `in_place` is used to inform if string was set again (which can trigger different encoding, size change, ...) or is it just updated byte directly.

Both functions get/set return false if index is out of bounds.

src/core/detail/bitpacking.h
```diff
@@ -21,6 +21,13 @@ bool validate_ascii_fast(const char* src, size_t len);
 void ascii_unpack(const uint8_t* bin, size_t ascii_len, char* ascii);
 void ascii_unpack_simd(const uint8_t* bin, size_t ascii_len, char* ascii);
 
+// Access a single byte in a 7-bit ASCII-packed string without unpacking the entire buffer.
+// These helpers read/write the ASCII byte at logical position `idx` in the unpacked string
+// directly from/into the packed `bin` representation.
+// Validate that `idx` is within bounds of the unpacked string length.
```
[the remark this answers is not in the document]
romange: I am sorry I still do not understand the contract. If you pass idx and ascii_len why do you need to pass invalid `idx`?  why not have precondition that `idx` must be less than `ascii_len` ?

src/core/compact_object.cc
```diff
@@ -1648,6 +1736,35 @@ size_t CompactObj::StrEncoding::Decode(std::string_view blob, char* dest) const
...
+  if (idx >= decoded_len) {
+    return false;
+  }
+  switch (enc_) {
+    case NONE_ENC:
+      *dest = blob[idx];
+      break;
+    case ASCII1_ENC:
+    case ASCII2_ENC:
+      detail::ascii_unpack_byte(reinterpret_cast<const uint8_t*>(blob.data()), decoded_len, idx,
```
romange: you see - you define a contract and immediately breke it here because you ignore here the return value which is reasonable because you check `idx` before.
so DCHECK inside `ascii_unpack_byte` and make `ascii_unpack_byte` void.

src/core/compact_object_test.cc
```diff
@@ -721,6 +793,210 @@ TEST_F(CompactObjectTest, Huffman) {
...
 }
 
+TEST_F(CompactObjectTest, GetByteAtOffset) {
+  // Inline string (INLINE_TAG)
+  {
+    string s = "hello";
+    cobj_.SetString(s);
+    for (size_t i = 0; i < s.size(); ++i) {
+      uint8_t res = 0;
+      cobj_.GetByteAtIndex(i, &res);
```
romange: use ASSERT_TRUE/FALSE  everywhere you call GetByteAtIndex to check it's return value. do not ignore them

src/core/detail/bitpacking.h
```diff
@@ -21,6 +21,15 @@ bool validate_ascii_fast(const char* src, size_t len);
 void ascii_unpack(const uint8_t* bin, size_t ascii_len, char* ascii);
 void ascii_unpack_simd(const uint8_t* bin, size_t ascii_len, char* ascii);
 
+// Access a single byte in a 7-bit ASCII-packed string without unpacking the entire buffer.
+// These helpers read/write the ASCII byte at logical position `idx` in the unpacked string
+// directly from/into the packed `bin` representation.
+// It's up to caller to verify:
+// `1. idx` must be less that `ascii_len` to avoid out-of-bounds access.
+// `2. `val` must be less than 128 (7-bit ASCII) for packing.
```
romange: ### 
```suggestion
// `2. `ascii` must be less than 128 (7-bit ASCII) for packing.
```

src/core/detail/bitpacking.h
```diff
@@ -21,6 +21,15 @@ bool validate_ascii_fast(const char* src, size_t len);
 void ascii_unpack(const uint8_t* bin, size_t ascii_len, char* ascii);
 void ascii_unpack_simd(const uint8_t* bin, size_t ascii_len, char* ascii);
 
+// Access a single byte in a 7-bit ASCII-packed string without unpacking the entire buffer.
+// These helpers read/write the ASCII byte at logical position `idx` in the unpacked string
+// directly from/into the packed `bin` representation.
+// It's up to caller to verify:
+// `1. idx` must be less that `ascii_len` to avoid out-of-bounds access.
```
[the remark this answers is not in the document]
mkaruza: @copilot update typo mistake

src/core/detail/bitpacking.cc
```diff
@@ -270,6 +270,46 @@ void ascii_unpack(const uint8_t* bin, size_t ascii_len, char* ascii) {
   }
 }
 
+void ascii_unpack_byte(const uint8_t* bin, size_t ascii_len, size_t idx, uint8_t* dest) {
```
romange: @copilot change the interface to of `ascii_unpack_byte` to return `uint8_t` by value and remove `dest` argument

PR #6759, opened 2026-02-28
refactor: move stream alloc/free into compact_object and fix rdb_load cleanup

romange:
## Summary
Move `streamNew`, `freeStream`, and their internal helpers from `t_stream.c` into
`compact_object.cc` under the `dfly` namespace, removing the dependency on `stream.h`
for stream allocation/deallocation.

Fix an inverted cleanup condition in `rdb_load.cc` `CreateStream` — the stream was
freed when `append` was true (successfully handed off) instead of when false (locally owned).

## Changes
- Move `streamNew`, `freeStream`, `streamFreeConsumer`, `streamFreeCGVoid`, `lpFreeVoid` to `compact_object.cc`
- Declare `streamNew`/`freeStream` in `compact_object.h`
- Fix `!config_.append` cleanup guard in `rdb_load.cc`
- Correct `StreamDecrID` comment in `stream_family.cc`

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -1662,4 +1662,57 @@ StringOrView CompactObj::StrEncoding::Decode(std::string_view blob) const {
   return {};
 }
 
+/* Create a new stream data structure. */
+stream* streamNew() {
+  stream* s = (stream*)zmalloc(sizeof(stream));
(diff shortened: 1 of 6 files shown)


romange:
augment review

Issue #6764, opened 2026-03-01
Dragonfly crashes in hset_family.cc:973


vyavdoshenko:
[crash-000000.tar.gz](https://github.com/user-attachments/files/25658954/crash-000000.tar.gz)

```
F20260301 10:36:34.388500 96406 hset_family.cc:973] Check failed: lplen > 0 && lplen % 2 == 0
```

```
  1. Start dragonfly:
     ./build/dragonfly --port 6379 --logtostderr --proactor_threads 1 --dbfilename=""
  2. Extract and replay:
     tar xzf crash-000000.tar.gz
     cd crash-000000
     python3 replay_crash.py crashes 000000
```

PR #6772, opened 2026-03-01
fix: prevent empty hash on HINCRBYFLOAT with NaN/Inf

vyavdoshenko:
`OpIncrBy` called `AddOrFind` (which allocates an empty listpack for new keys) before validating the increment. When `IncrementValue` rejected a NaN/Inf increment, it returned early — leaving an orphaned empty hash in the DB. Any subsequent `HRANDFIELD` on that key hit the CHECK.

Fix:
Validate NaN/Inf in the increment value before touching the DB. For a new key `prev_val` is 0.0, so `0.0 + NaN/Inf` is always invalid — reject it upfront without calling `AddOrFind`.

Fixes: #6764

### DIFF
--- a/src/server/hset_family.cc
+++ b/src/server/hset_family.cc
@@ -786,6 +786,10 @@ void CmdHIncrByFloat(CmdArgList args, CommandContext* cmd_cntx) {
     return cmd_cntx->SendError(kInvalidFloatErr);
   }
 
+  if (isnan(dval) || isinf(dval)) {
+    return cmd_cntx->SendError(kNanOrInfDuringIncr);
+  }
+
   IncrByParam param{dval};
 
   auto cb = [&](Transaction* t, EngineShard* shard) {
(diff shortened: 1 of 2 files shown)


romange:
good catch but the fix does not correspond  to the pr description. we still call AddOrFind before argument santization. Can we check before we call AddOrFind?

vyavdoshenko:
> good catch but the fix does not correspond to the pr description. we still call AddOrFind before argument santization. Can we check before we call AddOrFind?

Fixed

src/server/hset_family.cc
```diff
@@ -296,6 +296,13 @@ OpStatus IncrementValue(optional<string_view> prev_val, IncrByParam* param) {
 }
 
 OpStatus OpIncrBy(const OpArgs& op_args, string_view key, string_view field, IncrByParam* param) {
+  if (holds_alternative<double>(*param)) {
```
romange: it's an ok solution the better one would be to sanitize `param` before passing it to  `OpIncrBy` inside CmdHIncrByFloat.

romange: and in general if you find yourself sanitizing a parameter at the beginning of the call, ask yourself why not do  it at the caller?

vyavdoshenko: Fixed

PR #6784, opened 2026-03-02
chore(tiering): pass FragmentRef instead of PrimeValue in tiered storage

romange:
Wrap PrimeValue references in a FragmentRef type to encapsulate tiering
metadata access. Update Delete, CancelStash, and ShouldStash to accept
FragmentRef instead of raw PrimeValue pointers.

This removes PrimeValue references from the external APIs for tiered storage.
Next step would be to support Node* in FragmentRef and provide support for listpacks/blobs
for qlist.

Also - rename `TieredColdRecord` to `TieredCoolRecord` for consistency. 
"Cool" state is where items reside both on disk and in memory and can be immediately evicted any moment.

### DIFF
--- a/src/server/tiered_storage.cc
+++ b/src/server/tiered_storage.cc
@@ -58,10 +58,9 @@ namespace dfly {
 
 using namespace std;
 using namespace util;
-
 using tiering::FragmentRef;
 using tiering::KeyRef;
-using tiering::TieredColdRecord;
+using tiering::TieredCoolRecord;
 
 namespace {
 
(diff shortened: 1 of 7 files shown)


src/server/tiered_storage.cc
```diff
@@ -434,20 +433,16 @@ std::optional<util::fb2::Future<bool>> TieredStorage::Stash(DbIndex dbid, string
   return {};
 }
 
-void TieredStorage::Delete(DbIndex dbid, PrimeValue* value) {
```
romange, line 437: there are only syntactic changes in this file. the functionality is kept the same.

Issue #6765, opened 2026-03-01
Dragonfly crashes in compact_object.cc:52


vyavdoshenko:
[crash-000001.tar.gz](https://github.com/user-attachments/files/25658972/crash-000001.tar.gz)
```
F20260301 10:37:23.222342 96924 compact_object.cc:52] Check failed: false Can't decrease 1840 from 2896
```

[crash-000003.tar.gz](https://github.com/user-attachments/files/25658973/crash-000003.tar.gz)
```
F20260301 10:38:39.701440 97963 compact_object.cc:52] Check failed: false Can't decrease 304 from 432
```

[crash-000004.tar.gz](https://github.com/user-attachments/files/25658974/crash-000004.tar.gz)
```
F20260301 10:39:31.892174 98498 compact_object.cc:52] Check failed: false Can't decrease 128 from 144
```

[crash-000007.tar.gz](https://github.com/user-attachments/files/25658975/crash-000007.tar.gz)
```
F20260301 10:41:25.072171 99820 compact_object.cc:52] Check failed: false Can't decrease 208 from 272
```

[crash-000008.tar.gz](https://github.com/user-attachments/files/25658976/crash-000008.tar.gz)
```
F20260301 10:42:03.472087 100327 compact_object.cc:52] Check failed: false Can't decrease 176 from 224
```

[crash-000009.tar.gz](https://github.com/user-attachments/files/25658977/crash-000009.tar.gz)
```
F20260301 10:42:27.809290 100632 compact_object.cc:52] Check failed: false Can't decrease 96 from 112
```

To reproduce:
```
  1. Start dragonfly:
     ./build/dragonfly --port 6379 --logtostderr --proactor_threads 1 --dbfilename=""
  2. Extract and replay:
     tar xzf crash-000012.tar.gz
     cd crash-000012
     python3 replay_crash.py crashes 000012
```

PR #6785, opened 2026-03-02
fix: json family memory tracking and orphans

### LINKED ISSUE #6766: Dragonfly crashes in json_family.cc:99
[crash-000002.tar.gz](https://github.com/user-attachments/files/25658984/crash-000002.tar.gz)

[crash-000005.tar.gz](https://github.com/user-attachments/files/25658985/crash-000005.tar.gz)

[crash-000010.tar.gz](https://github.com/user-attachments/files/25658986/crash-000010.tar.gz)

[crash-000011.tar.gz](https://github.com/user-attachments/files/25658987/crash-000011.tar.gz)

[crash-000012.tar.gz](https://github.com/user-attachments/files/25658988/crash-000012.tar.gz)

[crash-000014.tar.gz](https://github.com/user-attachments/files/25658989/crash-000014.tar.gz)

```
F20260301 10:37:59.960086 97473 json_family.cc:99] Check failed: GetPrimeValue().MallocUsed() != 0
```

To reproduce:
```
  1. Start dragonfly:
     ./build/dragonfly --port 6379 --logtostderr --proactor_threads 1 --dbfilename=""
  2. Extract and replay:
     tar xzf crash-000001.tar.gz
     cd crash-000001
     python3 replay_crash.py crashes 000001
```

kostasrim:
We violated the contract `ShardJsonFromString` in `OpMerge`:

`If you do memory tracking, make sure to initialize it before calling this method`

* do not allocate before setting up memory tracking
* add test
* add error/memory correction for InternedStrings

resolves  #6766 and #6765

### DIFF
--- a/src/server/json_family.cc
+++ b/src/server/json_family.cc
@@ -96,8 +96,7 @@ class JsonAutoUpdater {
     GetPrimeValue().SetJsonSize(diff);
 
     // Under any flow we must not end up with this special value.
-    // TODO: disable for now as it breaks with interned strings.
-    // DCHECK(GetPrimeValue().MallocUsed() != 0);
+    DCHECK(GetPrimeValue().MallocUsed() != 0);
   }
 
   void AddDocToIndexes() {


src/server/json_family.cc
```diff
@@ -502,12 +514,19 @@ OpStatus SetFullJson(const OpArgs& op_args, string_view key, string_view json_st
...
-  // constructing JsonAutoUpdater. This ensures start_size_ doesn't include the old
-  // string's memory, which would cause a negative diff in SetJsonSize() and crash
-  // in UpdateSize().
+  const bool is_new_key = it_res->is_new;
+
   if (type != OBJ_JSON) {
+    // TODO -- this is a performance regression. If json_str is large, we parse it here
+    // then again  later after we set up the JsonAutoUpdater. The issue here is that we
+    // need to deallocate before we create the updater but after we parse the string. Yet,
+    // parsing requires the updater to be created first(see its comments).
```
kostasrim: My concern here is that we parse twice. json heap allocates and this can easily penaltize long json literals. 

We can't however avoid this in a smart way, as storing the result of parsing requires for the memory tracker to be setup first but we can't because we need to first reset the object below. We can't order the operations to satisfy all the invariants :/ 

My main question is what's the reason of `ShardJsonFromString` requirements ? In other words, why does memory tracking break exactly when we allocate before (if that variable is passed by const & arround and by RAII we will exist after updater exits and therefore the acounting should be in principle correct as it cancels out)

src/server/json_family.cc
```diff
@@ -1634,16 +1657,24 @@ OpStatus OpMSet(const OpArgs& op_args, const ShardArgs& args) {
...
   auto cb = [&](std::optional<std::string_view> cur_path, JsonType* val) -> MutateCallbackResult<> {
     string_view strpath = cur_path ? *cur_path : string_view{};
 
     DVLOG(2) << "Handling " << strpath << " " << val->to_string();
+
+    // Parse JSON inside the callback, after JsonAutoUpdater has measured start_size_.
+    // This avoids memory tracking bugs and double parsing.
+    std::optional<JsonType> parsed_json = ShardJsonFromString(json_str);
+    if (!parsed_json) {
+      parse_error = OpStatus::INVALID_JSON;
```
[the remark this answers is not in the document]
kostasrim: Wow good catch!

src/server/json_family.cc
```diff
@@ -1634,16 +1657,24 @@ OpStatus OpMSet(const OpArgs& op_args, const ShardArgs& args) {
 // implemented yet.
 OpStatus OpMerge(const OpArgs& op_args, string_view key, string_view path,
                  const WrappedJsonPath& json_path, std::string_view json_str) {
-  std::optional<JsonType> parsed_json = ShardJsonFromString(json_str);
-  if (!parsed_json) {
-    VLOG(1) << "got invalid JSON string '" << json_str << "' cannot be saved";
-    return OpStatus::INVALID_JSON;
-  }
+  // To avoid double parsing we delegate it for later inside the callback.
```
romange: this change looks ok - tradeoff are reasonable. 
I would like to understand better the previous change

src/server/json_family_memory_test.cc
```diff
@@ -247,4 +247,23 @@ TEST_F(JsonFamilyMemoryTest, ShortKeyAccounting) {
   EXPECT_LE(std::llabs(actual - expected), 64);
 }
 
+TEST_F(JsonFamilyMemoryTest, MergeMemoryTrackingCrash) {
+  Run("JSON.SET key $ {\"x\":1}");
+
+  // Perform many merges that grow the JSON slightly each time.
+  // The fuzzer needed ~121 merges to trigger the crash, followed by a delete operation.
```
romange:
```suggestion
```

src/server/json_family.cc
kostasrim: Which one ?

src/server/json_family_memory_test.cc
```diff
@@ -247,4 +247,23 @@ TEST_F(JsonFamilyMemoryTest, ShortKeyAccounting) {
...
+  // We'll do 150 merges to ensure we hit the accounting drift.
+  for (int i = 0; i < 150; i++) {
+    auto resp = Run("JSON.MERGE key $ {\"y\":2}");
```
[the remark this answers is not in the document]
kostasrim: We dont need to loop at all. good catch!

src/server/json_family.cc
romange: `TODO -- this is a performance regression.` ....

kostasrim:
@romange PTAL

src/core/compact_object.cc
```diff
@@ -893,7 +893,14 @@ void CompactObj::SetJson(JsonType&& j) {
 void CompactObj::SetJsonSize(int64_t size) {
   if (taglen_ == JSON_TAG && JsonEnconding() == kEncodingJsonCons) {
     // JSON.SET or if mem hasn't changed from a JSON op then we just update.
-    u_.json_obj.cons.bytes_used = UpdateSize(u_.json_obj.cons.bytes_used, size);
+    int64_t result = static_cast<int64_t>(u_.json_obj.cons.bytes_used) + size;
+    if (result <= 0) {
+      LOG_EVERY_N(ERROR, 30) << "JSON size underflow: " << u_.json_obj.cons.bytes_used << " + "
```
romange: use `LOG_EVERY_T(ERROR, 10)` or something

src/core/compact_object.cc
```diff
@@ -893,7 +893,14 @@ void CompactObj::SetJson(JsonType&& j) {
...
+                             << size << " = " << result;
+      u_.json_obj.cons.bytes_used = (result < 0) ? static_cast<size_t>(-result) : 1;
```
romange: why 1?

src/server/json_family.cc
```diff
@@ -503,12 +515,19 @@ OpStatus SetFullJson(const OpArgs& op_args, string_view key, string_view json_st
...
   }
 
-  // For non-JSON types (i.e., strings), validate JSON and free the old value before
-  // constructing JsonAutoUpdater. This ensures start_size_ doesn't include the old
-  // string's memory, which would cause a negative diff in SetJsonSize() and crash
-  // in UpdateSize().
+  const bool is_new_key = it_res->is_new;
+
   if (type != OBJ_JSON) {
+    // TODO -- this is a performance regression. If json_str is large, we parse it here
```
romange: I am looking  at a diff. What exactly is performance regression? `DelMutable` ?

src/core/compact_object.cc
```diff
@@ -893,7 +893,14 @@ void CompactObj::SetJson(JsonType&& j) {
 void CompactObj::SetJsonSize(int64_t size) {
   if (taglen_ == JSON_TAG && JsonEnconding() == kEncodingJsonCons) {
     // JSON.SET or if mem hasn't changed from a JSON op then we just update.
-    u_.json_obj.cons.bytes_used = UpdateSize(u_.json_obj.cons.bytes_used, size);
+    int64_t result = static_cast<int64_t>(u_.json_obj.cons.bytes_used) + size;
+    if (result <= 0) {
```
romange: < 1

src/server/json_family.cc
romange: please rephrase, use `inefficiency`

src/server/json_family.cc
```diff
@@ -503,12 +515,19 @@ OpStatus SetFullJson(const OpArgs& op_args, string_view key, string_view json_st
...
+    // then again  later after we set up the JsonAutoUpdater. The issue here is that we
+    // need to deallocate before we create the updater but after we parse the string. Yet,
+    // parsing requires the updater to be created first(see its comments).
     if (!ShardJsonFromString(json_str)) {
+      if (is_new_key) {
+        // Delete the key if it's new
```
romange:
```suggestion
      if (is_new_key) {
        // Delete the key if it was created during this operation
```

src/server/json_family.cc
romange: // this is the case of the existing object with a differrent type being rewritten by json set.
// in that case we double check that json is valid before we erase the object and before 
// we create JsonAutoUpdater to account for json-related memory  changes.

kostasrim: Sorry I missed this, this is the `base` case. `auto it_res = op_args.GetDbSlice().AddOrFind(op_args.db_cntx, key, std::nullopt);` above will return `OBJ_STRING`. This if will trigger for `all new json`. 

So we double parse every time we create a new json 

p.s. this was already there. I left a comment will push soon

src/server/json_family.cc
romange: you need to uncomment this one too @kostasrim

PR #6794, opened 2026-03-03
fix(facade): improve connection setup and buffer capacity tracking

romange:
## Summary

Cleans up connection initialization and buffer tracking in `dragonfly_connection.cc`.

## Changes

- Move `TCP_NODELAY` setsockopt call from `HandleRequests` to `OnConnectionStart` for earlier setup
- Simplify `IsMainOrMemcache` to use the cached `protocol_` field instead of querying the listener
- Fix `DoReadOnRecv` to wrap `WriteAndCommit` in `UpdateIoBufCapacity` so buffer growth is tracked in stats
- Remove redundant null guard on `stats` in `UpdateIoBufCapacity` (caller always passes non-null)
- Remove stale TODO and TLS comments

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -159,7 +159,7 @@ void UpdateIoBufCapacity(const io::IoBuf& io_buf, ConnectionStats* stats,
   const size_t prev_capacity = io_buf.Capacity();
   f();
   const size_t capacity = io_buf.Capacity();
-  if (stats != nullptr && prev_capacity != capacity) {
+  if (prev_capacity != capacity) {
     VLOG(2) << "Grown io_buf to " << capacity;
     stats->read_buf_capacity += capacity - prev_capacity;
   }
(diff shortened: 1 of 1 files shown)


abhijat (dismissed):

src/facade/dragonfly_connection.cc
```diff
@@ -703,8 +703,13 @@ void Connection::OnPostMigrateThread() {
 void Connection::OnConnectionStart() {
   SetName(absl::StrCat(id_));
 
-  if (const Listener* lsnr = static_cast<Listener*>(listener()); lsnr) {
-    is_main_ = lsnr->IsMainInterface();
+  const Listener* lsnr = static_cast<Listener*>(listener());
```
glevkovich: I would add here:
DCHECK(lsnr);
since some unittests tests may not set ListenerInterface::listener_

romange: reverted and added the comment

src/facade/dragonfly_connection.cc
```diff
@@ -159,7 +159,7 @@ void UpdateIoBufCapacity(const io::IoBuf& io_buf, ConnectionStats* stats,
   const size_t prev_capacity = io_buf.Capacity();
   f();
   const size_t capacity = io_buf.Capacity();
-  if (stats != nullptr && prev_capacity != capacity) {
+  if (prev_capacity != capacity) {
```
glevkovich, line 162: since this is a public/global function, consider adding here a DCHECK(stats) in case some test decide to pass null stats.

romange: it's local function in the file.the whole point of it is to update stats.

PR #6804, opened 2026-03-04
refactor: deduplicate build steps into reusable composite action

romange:
Create .github/actions/builder to eliminate duplicated CMake/Ninja commands across 7 workflow files, reducing 94 lines to 39.

Hardcoded options: GCP=OFF, unwind=OFF, gperf=OFF, stacktraces=ON. Removed ccache and simplified to single target (ninja src/all).

### DIFF
--- a/.github/actions/builder/action.yml
+++ b/.github/actions/builder/action.yml
@@ -35,7 +35,12 @@ inputs:
   with-aws:
     description: "Build with AWS support"
     required: false
-    default: 'OFF'
+    default: 'ON'
+    type: string
+  targets:
+    description: "Build targets to compile"
+    required: false
+    default: 'src/all'
     type: string
 
 runs:
@@ -90,5 +95,5 @@ runs:
       shell: bash
       run: |
         cd ${GITHUB_WORKSPACE}/${{ inputs.build-dir }}
-        echo "Building target: src/all"
-        ninja src/all
+        echo "Building target: ${{ inputs.targets }}"
+        ninja ${{ inputs.targets }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,6 +110,7 @@ jobs:
           cxx-compiler: ${{matrix.compiler.cxx}}
           cxx-flags: ${{matrix.cxx_flags}}
           sanitizers: ${{matrix.sanitizers}}
+          with-aws: 'OFF'
 
       - name: PostFail
         if: failure()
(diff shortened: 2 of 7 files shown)


.github/actions/builder/action.yml
```diff
@@ -0,0 +1,94 @@
...
+        # Execute CMake
+        echo "Running: ${CMAKE_CMD[@]}"
+        "${CMAKE_CMD[@]}"
+
+    - name: Build
+      shell: bash
+      run: |
+        cd ${GITHUB_WORKSPACE}/${{ inputs.build-dir }}
+        echo "Building target: src/all"
+        ninja src/all
```
vyavdoshenko: It will always build all targets, even for regression tests.
I suggest parametrizing it and using either `src/all` or `dragonfly`.

.github/workflows/ci.yml
```diff
@@ -249,16 +222,11 @@ jobs:
...
-                -DCMAKE_CXX_FLAGS=-no-pie
-
-          cd ${GITHUB_WORKSPACE}/build && ninja dragonfly
-          pwd
-          ls -l ..
+      - name: Build Dragonfly
+        uses: ./.github/actions/builder
+        with:
+          build-type: Release
+          with-aws: 'ON'
```
vyavdoshenko: Is it worth flipping this flag to `ON` as the default and explicitly setting `OFF`?

.github/actions/builder/action.yml
```diff
@@ -0,0 +1,94 @@
...
+runs:
+  using: "composite"
+  steps:
+    - name: Configure CMake
+      shell: bash
+      run: |
+        # Set sanitizer flags
+        ASAN="OFF"
+        USAN="OFF"
+        if [ '${{ inputs.sanitizers }}' = 'Sanitizers' ]; then
```
vyavdoshenko, line 55: Better to use this combination for GitHub Actions.
```
env:
  SANITIZERS_INPUT: ${{ inputs.sanitizers }}
```

```
run: |
  if [ "$SANITIZERS_INPUT" = 'Sanitizers' ]; then
```

.github/actions/builder/action.yml
romange: I did not understand the comment. It build `src/all` - it's not enough?

vyavdoshenko: We had two different workflows. src/all for unit tests and dragonfly target only for regression tests. The second option is faster, but src/all will work for all builds, but the build will be slower.

.github/actions/builder/action.yml
romange: I guess it's matter of what works.  what's wrong with `if [ '${{ inputs.sanitizers }}' = 'Sanitizers' ];` ?

vyavdoshenko: Never mind. Your code should work.
I mentioned it only because this is a shell injection and suggested a more preferable way to GitHub Actions.

PR #6805, opened 2026-03-04
feat: add pending keys to HNSW index during full sync phase

BorysTheDev:
feat: add pending keys to HNSW index during full sync phase
   
    During replication full sync, journal events (adds, updates, deletes) can
    arrive at the replica while HNSW vector indices are being restored from RDB.
    Without buffering, these mutations are lost, causing index inconsistency.
    
    This change introduces pending_vector_updates_ to buffer document keys that
    are modified during the LOADING window (while is_restoring_vectors_ is true).
    After RestoreGlobalVectorIndices completes, the buffer is drained:
    - For updated documents: Remove old HNSW node and add new one
    - For new documents: Add to both regular and vector indices
    - For deleted documents: Remove HNSW node and key_index_ entry
    
    The buffering prevents journal events from being lost and ensures HNSW
    graph consistency between master and replica after full sync completes.

### DIFF
--- a/src/server/search/doc_index.cc
+++ b/src/server/search/doc_index.cc
@@ -338,7 +338,15 @@ void ShardDocIndex::Rebuild(const OpArgs& op_args, PMR_NS::memory_resource* mr, 
   // the existing DocIds to add documents to the regular indices.
   if (!is_restored) {
     key_index_ = DocKeyIndex{};
+    // Full rebuild handles all documents — discard any buffered state from LOADING.
+    is_restoring_vectors_ = false;
+    pending_vector_updates_.clear();
+  } else {
+    // Restored path: VectorLoop will call RestoreGlobalVectorIndices which drains
+    // the buffers. Until then, buffer any journal-driven mutations.
+    is_restoring_vectors_ = true;
   }
+
   indices_.emplace(base_->schema, base_->options, mr, &synonyms_);
 
   // Create builder and start indexing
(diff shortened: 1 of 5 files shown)


src/server/search/doc_index.cc
```diff
@@ -463,6 +474,13 @@ void ShardDocIndex::AddDocToGlobalVectorIndex(ShardDocIndex::DocId doc_id, const
 
 void ShardDocIndex::RemoveDocFromGlobalVectorIndex(ShardDocIndex::DocId doc_id,
                                                    const DbContext& db_cntx, const PrimeValue& pv) {
+  if (is_restoring_vectors_) {
+    // Buffer the key — will be re-applied after RestoreGlobalVectorIndices completes.
+    std::string_view key = key_index_.Get(doc_id);
+    pending_vector_updates_.emplace(key);
+    return;
```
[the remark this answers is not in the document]
vyavdoshenko: @BorysTheDev please take a look at this comment

src/server/search/doc_index.h
```diff
@@ -377,13 +377,22 @@ class ShardDocIndex {
                                                     const SearchParams::SortOption& sort,
                                                     const OpArgs& op_args) const;
 
+  // Remove a DocId from all HNSW indices for this index.
+  void RemoveFromAllHnswIndices(search::DocId doc_id);
```
vyavdoshenko, line 381: should be private

BorysTheDev:
> There's no test that exercises the scenario this PR fixes: documents mutated via journal during the full sync LOADING window ending up correctly reflected in HNSW after sync completes. Without it, a future refactor could silently regress.
> 
> A minimal test: start replication, pause after RDB snapshot is sent, inject journal ADD/DELETE events for vector-indexed keys, resume, verify HNSW search results match the master.

I didn't add tests because there are several bugs exist in code, but I don't want to add their fix into this PR. I will add them and tests in the next PR

src/server/search/doc_index.cc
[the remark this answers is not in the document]
BorysTheDev: I know about it, it will be in the next PRs

src/server/search/doc_index.h
BorysTheDev: I will do in the next PR

BorysTheDev: by the way it's already private, see above

PR #6807, opened 2026-03-04
fix(tiering): Serialize delayed entries from SerializeBucket

### LINKED ISSUE #6816: bug(tiering): data race between journaling changes and tiering changes during the replication
Note that the issue refers to the state of v1.36 or earlier. https://github.com/dragonflydb/dragonfly/pull/6610 tried to fix it but unsuccessfully and created other bugs that will be described separately.

## The problem
1. When we serialize a bucket and encounter offloaded (string) entitities, we issue an async read request, enqueue it into `delayed_entries_` collection and continue serializing normal entities. then, at some later time we harvest the realized futures in `delayed_entries_`  and send them to the replication channel. this is done to avoid high latencies due to blocking on disk and have fast replication process.

This however does not work with journal writes. Specifically, if we have an entity "key: external_value",
and we serialize it due to a APPEND call, for example,
the order will be:
a. serialize its bucket, omit "key" and enqueue its read request to  `delayed_entries_`.
b. serialize `append` journal change.
c. Later when read request is finalized, send its result to the replication channel and override the "strappend" journal change.

2. We have data race when `SliceSnapshot::ThrottleIfNeeded` that is called by differrent fibers (Heartbeat, for example)
     and then `ThrottleIfNeeded` calls PushSerialized that goes over delayed entries (gathered by `snapshot_fb_` ) and flushes them to stream. The problem is that if that path preempts, `snapshot_fb_` may resume, potentially encounter a huge value (hashmap or a huge string) start serializing that entry. Once it suspends in the middle, the original `PushSerialized` will resume and will corrupt the stream. The stream us usually protected by `big_value_mu_` but this interaction escapes it.

As I mentioned, #6610  tries to fix (1) but its implementation is not good as it corrupts the stream 
because of using a separate RdbSerializer instance. It flushes data from it while the original `serializer_ ` may still have leftover data in it not yet flushed. In addition it does not solve (2).

mkaruza:
We should serialize entries outside of PushSerialized to avoid
corrupting data. Idea is same as done in migration, we should serialize
delayed entries one they are resolved after each bucket iteration.
For mutations that could execute during serialization we must serialize
delayed entries forcefully to have correct order.

Cherry-picked #6824 
    
Closes #6816 #6687

### DIFF
--- a/src/server/snapshot.cc
+++ b/src/server/snapshot.cc
@@ -375,7 +375,8 @@ unsigned SliceSnapshot::SerializeBucket(DbIndex db_index, PrimeTable::bucket_ite
   unsigned result = 0;
 
   std::vector<TieredDelayEntryKey> bucket_tiered_keys;
-  const bool track_tiered_keys = push_tiered && EngineShard::tlocal()->tiered_storage() != nullptr;
+  const bool tiering_enabled = EngineShard::tlocal()->tiered_storage() != nullptr;
+  const bool track_tiered_keys = push_tiered && tiering_enabled;
 
   for (it.AdvanceIfNotOccupied(); !it.is_done(); ++it) {
     ++result;
@@ -387,11 +388,14 @@ unsigned SliceSnapshot::SerializeBucket(DbIndex db_index, PrimeTable::bucket_ite
     }
   }
 
-  // Push tracked tiered keys forcefully. If there are too many delayed entries
-  // accumulated we should also push them forcefully.
-  const size_t kMaxDelayedEntries = 512;
-  PushDelayedEntries(delayed_entries_.size() > kMaxDelayedEntries,
-                     track_tiered_keys ? &bucket_tiered_keys : nullptr);
+  if (tiering_enabled) {
+    // Push tracked tiered keys forcefully. If there are too many delayed entries
+    // accumulated we should also push them forcefully.
+    const size_t kMaxDelayedEntries = 512;
+    PushDelayedEntries(delayed_entries_.size() > kMaxDelayedEntries,
+                       track_tiered_keys ? &bucket_tiered_keys : nullptr);
+  }
+
   serialize_bucket_running_ = false;
   return result;
 }


mkaruza:
> It works, though its very straightforward. Avoiding blocking inside SerializeBucket was a primary goal - but now we do it

I agree that it's not most nicest place to block, but it should give correct behavior and ordering.

romange:
please cherry-pick the test

src/server/snapshot.cc
```diff
@@ -454,72 +464,62 @@ void SliceSnapshot::HandleFlushData(std::string data) {
 }
 
 size_t SliceSnapshot::FlushSerialized(RdbSerializer* serializer) {
```
romange: remove `RdbSerializer* serializer` argument

src/server/snapshot.h
```diff
@@ -139,8 +139,8 @@ class SliceSnapshot : public journal::JournalConsumerInterface {
   // Push regardless of buffer size if force is true.
   // Return true if pushed. Can block. Is called from the snapshot thread.
   bool PushSerialized(bool force);
-  void SerializeExternal(DbIndex db_index, PrimeKey key, const PrimeValue& pv, time_t expire_time,
-                         uint32_t mc_flags);
+  void SerializeExternal(DbIndex db_index, std::string_view key, const PrimeValue& pv,
```
romange: this change is not needed

src/server/snapshot.cc
```diff
@@ -355,12 +357,13 @@ bool SliceSnapshot::BucketSaveCb(DbIndex db_index, PrimeTable::bucket_iterator i
...
 
-  stats_.loop_serialized += SerializeBucket(db_index, it);
+  stats_.loop_serialized += SerializeBucket(db_index, it, false);
 
   return false;
 }
 
-unsigned SliceSnapshot::SerializeBucket(DbIndex db_index, PrimeTable::bucket_iterator it) {
+unsigned SliceSnapshot::SerializeBucket(DbIndex db_index, PrimeTable::bucket_iterator it,
+                                        bool from_cb) {
```
romange: rename it to `push_tiered`. never name arguments by from where they are passed. instead name them by what they do.

src/server/snapshot.cc
```diff
@@ -453,73 +464,71 @@ void SliceSnapshot::HandleFlushData(std::string data) {
...
+      continue;
+    }
 
-  // Atomic bucket serialization might have accumulated some delayed values.
-  // Because we can finally block in this function, we'll await and serialize them
-  thread_local LocalLatch delayed_flush_latch_;
-  while (!delayed_entries_.empty()) {
-    // After pop_front there is no indication of the operation ongoing, so we need a latch
-    std::unique_lock lk{delayed_flush_latch_};
+    // If tiered_keys filter is provided, only serialize matching keys
```
romange: there is an inefficiency here. If a future is resolved you won't erase it just because it's not part of the filter. I would first thing in the loop check `if (entry->value.IsResolved())` and handle this state.

src/server/snapshot.cc
```diff
@@ -453,73 +464,71 @@ void SliceSnapshot::HandleFlushData(std::string data) {
...
-      // TODO: to introduce RdbSerializer::SaveString that can accept a string value directly.
-      PrimeValue pv{*res};
-      delayed_serializer.SaveEntry(entry.key, pv, entry.expire, entry.mc_flags, entry.dbid);
-    } while (!delayed_entries_.empty());
+    // Serialize the entry and remove it from delayed_entries_
+    PrimeValue pv{*value};
+    io::Result<uint8_t> res =
+        serializer_->SaveEntry(entry->key, pv, entry->expire, entry->mc_flags, entry->dbid);
+    CHECK(res);
+    delayed_entries_.erase(it++);
```
romange: another inefficiency: you erase from this vector in `O(n^2)`complexity. It's unnecessary, 
you can erase by: 
```
*it = delayed_entries_.back();
delayed_entries_.pop_back();
``` in `O(1)`.

File: src/server/snapshot.h [hunk 6]
```diff
@@ -160,7 +164,11 @@ class SliceSnapshot : public journal::JournalConsumerInterface {
   DbIndex snapshot_db_index_ = 0;
 
   std::unique_ptr<RdbSerializer> serializer_;
-  std::deque<TieredDelayedEntry> delayed_entries_;  // collected during atomic bucket traversal
+
+  // Delayed entries that are waiting for tiered storage reads to complete before they can be
+  // serialized. Tuple <db_index, key> is used as a key to uniquely identify entry.
+  absl::flat_hash_map<std::pair<DbIndex, std::string>, std::unique_ptr<TieredDelayedEntry>>
```
romange: I now see that you changed `delayed_entries_` to a map but you never use this quality. Have you done it only for erasing in `O(1)` ? my comment still holds you can still do it with vector.

src/server/snapshot.cc
```diff
@@ -453,73 +464,71 @@ void SliceSnapshot::HandleFlushData(std::string data) {
...
-        cntx_->ReportError(make_error_code(errc::io_error),
-                           absl::StrCat("Failed to read ", entry.key.ToString()));
-        return false;
-      }
+    // Get the value from the future (blocks if not resolved and force=true)
+    auto value = entry->value.Get();
+    if (!value.has_value()) {
+      cntx_->ReportError(make_error_code(errc::io_error),
+                         absl::StrCat("Failed to read tiered key: ", entry->key.ToString()));
+      return;
```
romange, line 502: in case of I/O error - there is no need to return early - you can just skip serialization of this value but you can still process the collection.

src/server/snapshot.cc
```diff
@@ -315,6 +315,8 @@ void SliceSnapshot::IterateBucketsFb(bool send_full_sync_cut) {
     } while (snapshot_cursor_);
 
     DVLOG(2) << "after loop " << ThisFiber::GetName();
+    // We should now push all delayed entries
```
romange:
```suggestion
    // Wait for all the outstanding delayed entries and serialise them as well. 
```

src/server/snapshot.h
mkaruza: It is used to: 

1. Find entries in map when track_tiered_keys is used.
2. For removal, we can skip keys that are not yet resolved.

src/server/snapshot.cc
mkaruza: We are handling first filter keys than follow to serialize all keys that are resolved.

PR #6823, opened 2026-03-06
fix: finalize memory accounting before tiered stash during RDB load

romange:
During RDB load with tiering enabled, StashPrimeValue was called while the AutoUpdater was still alive. The async stash could complete before the AutoUpdater destructor, transforming PrimeValue to external and changing MallocUsed(). The AutoUpdater then computed a negative memory delta, crashing in AccountObjectMemory with:
"Check failed: obj_memory_usage + size >= 0"

Fix by calling AutoUpdater::Run() before StashPrimeValue to finalize memory accounting while PrimeValue still has its in-memory form.

### DIFF
--- a/tests/dragonfly/tiering_test.py
+++ b/tests/dragonfly/tiering_test.py
@@ -10,7 +10,6 @@ from . import dfly_args
 from .seeder import DebugPopulateSeeder, Seeder as SeederV2
 from .utility import (
     info_tick_timer,
-    tmp_file_name,
     wait_for_replicas_state,
     check_all_replicas_finished,
 )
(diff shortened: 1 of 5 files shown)


src/server/rdb_load.cc
```diff
@@ -2667,20 +2667,26 @@ void RdbLoader::CreateObjectOnShard(const DbContext& db_cntx, const Item* item,
...
 
   if (auto* ts = db_slice->shard_owner()->tiered_storage(); ts) {
-    StashPrimeValue(db_cntx.db_index, item->key, false, &res.it->second, ts);
+    // Finalize the AutoUpdater before stashing. The stash callback may complete
+    // (e.g. during the SleepFor yield below) and transform the PrimeValue to external,
+    // changing MallocUsed(). If the AutoUpdater ran after that, it would compute a
+    // bogus negative memory delta and crash in AccountObjectMemory.
+    auto it = updater.it;
+    updater.post_updater.Run();
+    StashPrimeValue(db_cntx.db_index, item->key, &it->second, ts, nullptr);
```
dranikpg, line 2689: nit: Passing the auto updater on as a value might be the nicest inteface

tests/dragonfly/tiering_test.py
```diff
@@ -179,3 +180,65 @@ async def fill_job():
...
+@pytest.mark.large
+@pytest.mark.opt_only
+async def test_rdb_load_with_tiering(df_factory: DflyInstanceFactory):
+    """
+    Regression test for RDB load with tiering. Verifies that loading a snapshot
+    into a tiered instance produces correct memory accounting (no underflow)
+    and preserves data integrity. Previously, the AutoUpdater's memory accounting
+    could race with the async stash callback, causing a crash in AccountObjectMemory.
+    """
+    dbfilename = f"dump_{tmp_file_name()}"
```
dranikpg: We already hava a test in snapshot_test.py

src/server/rdb_load.cc
romange: why?

tests/dragonfly/tiering_test.py
romange: but it did not find the bug. this one does. should I replace the previous test?

romange: moved there

src/server/rdb_load.cc
dranikpg: Because you pass the iterator there by value and it runs there on destruction

tests/dragonfly/tiering_test.py
dranikpg: I think its always better from a code size/running time perspective to have once test than two

romange: i agree but I am not sure what other tests cover. mine is not redundant. which one should I remove?

PR #6827, opened 2026-03-07
doc(serialization): expand shard serialization design with tiered analysis and roadmap

romange:
## Summary

- Restructure "Open Questions" (3 sections) into "Inefficiencies and
  Improvement Goals" (5 sections) with problem/goal/approach structure
- Add analysis of journal entries that bypass `OnDbChange` (expiry,
  eviction, control signals) and tiered delayed serialization semantics
- Add phased technical roadmap (Phases 0–4, items 1–17) for
  incremental `big_value_mu_` decoupling and removal

## Changes

- Define self-contained vs baseline-dependent journal entry terminology
- Add "Journal Entries Without `OnDbChange`" section with bypass-path
  table and deferred deletion queue proposal
- Add counter-example: byte-stream corruption from inline eviction
  without `ConsumeJournalChange` mutex
- Add "Delayed Serialization of tiered entities" section with
  `RestoreStreamer` note and two notions of bucket completion
- Clarify `CVCUponInsert` non-usage in non-PIT mode
- Restructure improvement goals into 5 sections:
  §1 bucket completion tracking, §2 shard-wide stall,
  §3 shared serializer buffer, §4 non-PIT journal traffic,
  §5 mutex-role-to-replacement summary table
- Add 17-item technical roadmap across 5 phases
- Update flowchart with expiry/eviction path
- Document `OnDbChange` blocking latency tradeoff and large-value
  lock-release stability (bucket versioning + `OnDbChange` blocking)

### DIFF
--- a/docs/shard-serialization.md
+++ b/docs/shard-serialization.md
@@ -672,10 +672,10 @@ See Roadmap items 3, 5, 6.
 ### 3. Shared serializer buffer and wire-format coupling
 
 **Problem.** `ConsumeJournalChange` and `SerializeBucket` write to the same `serializer_`
-buffer (the "buffer exclusivity" role from [§2](#2-shard-wide-stall-under-big_value_mu_)).
+buffer (the "buffer exclusivity" role from [§1](#1-shard-wide-stall-under-big_value_mu_)).
 Even with separate buffers, interleaved output from two serializers cannot be demuxed by the
 consumer without a framing protocol — a journal entry injected mid-RDB-entry produces an
-unparseable byte stream (see the [eviction counter-example](#2-shard-wide-stall-under-big_value_mu_)
+unparseable byte stream (see the [eviction counter-example](#1-shard-wide-stall-under-big_value_mu_)
 for a concrete scenario).
 
 **Goal.** Decouple journal and bucket serialization so they can produce data independently,


docs/shard-serialization.md
```diff
@@ -489,151 +591,440 @@ Even without I/O blocking, serializing large containers (listpack, quicklist, st
...
+  Each bucket is identified by a stable `BucketIdentity`. A bucket must remain in the
+  tracking map (`currently_serialized_: map<BucketIdentity, State>`) until all work completes; otherwise `version >= snapshot_version_` + absent-from-map would falsely read as `Covered`.
+  State encoding:
+
+  | State | Encoding | Meaning |
+  |-------|----------|---------|
+  | **NotVisited** | `version < snapshot_version_`, not in map | Traversal has not reached this bucket |
+  | **Serializing** | `version >= snapshot_version_`, in map as `Serializing` | Traversal is iterating this bucket |
+  | **DelayedPending** | `version >= snapshot_version_`, in map as `DelayedPending` | Iteration done, tiered entries still pending |
+  | **Covered** | `version >= snapshot_version_`, not in map | Baseline fully emitted |
```
dranikpg: Because serializing state is currently not visible as it happens under big_value_mu_, only DelayedPending is the real actionable value - we can store the delated entries explicitly instead of a state. And then we can actually omit this tracking map for buckets without delayed entries

docs/shard-serialization.md
```diff
@@ -489,151 +591,440 @@ Even without I/O blocking, serializing large containers (listpack, quicklist, st
...
+and non-preempting chunk production. See Roadmap items 6, 7, 8, 9.
+
+### 3. Shared serializer buffer and wire-format coupling
+
+**Problem.** `ConsumeJournalChange` and `SerializeBucket` write to the same `serializer_`
+buffer (the "buffer exclusivity" role from [§2](#2-shard-wide-stall-under-big_value_mu_)).
+Even with separate buffers, interleaved output from two serializers cannot be demuxed by the
+consumer without a framing protocol — a journal entry injected mid-RDB-entry produces an
+unparseable byte stream (see the [eviction counter-example](#2-shard-wide-stall-under-big_value_mu_)
+for a concrete scenario).
```
dranikpg, line 679: A tagged format is a good idea, but I still don't really understand how (except for huge values) the format is not partitionable. It seems that the outputs of Serialize() -> Buffer should be freely chainable

docs/shard-serialization.md
```diff
@@ -489,151 +591,440 @@ Even without I/O blocking, serializing large containers (listpack, quicklist, st
...
+
+**Goal.** In non-PIT mode, reduce journal traffic by skipping entries that are guaranteed to
+be covered by the traversal, without compromising eventual consistency.
+
+**Approach.** Use the bucket completion state machine (§1) to classify mutations:
+
+- **Self-contained entries** (`SET`, `DEL`, `EXPIRE`): skip for `NotVisited` buckets (traversal will see post-mutation value); emit for `Covered` buckets; emit conservatively for
+  `Serializing`/`DelayedPending`. Classification is by **emitted journal command form**, not
+  the user-facing command — commands like `JSON.SET` may be self-contained or not depending
+  on arguments and must be validated individually.
```
dranikpg, line 712: This is actually as an action more complicated than you would assume, because you have to differentiate between more than self-contained or baseline dependent entries. Imagine this case

```
DEL set-1 // skips
SINTERSTORE set-output set-1 set-2 set-3
```

It is of course safe to omit DEL, but if a SINTERSTORE with it appears, you now are out of sync.

Now each command that has _read-dependencies_ needs to perform versioning checks for them as well. We don't do that now with FindReadOnly. If you try not updating the bucket version, you will get in trouble with multiple snapshots occuring at once (replication + save)

docs/shard-serialization.md
```diff
@@ -489,151 +591,440 @@ Even without I/O blocking, serializing large containers (listpack, quicklist, st
...
+
+The previous subsections identify `big_value_mu_`'s three roles and the mechanisms that
+replace each:
+
+| Mutex role | Replacement | Source |
+|-----------|-------------|--------|
+| Journal ordering | Bucket completion state + deferred deletion queue | §1 |
+| Buffer exclusivity | Separate `RdbSerializer` per producer + tagged chunks | §3 |
+| Bucket atomicity (PIT) | Bucket versioning + `OnDbChange` blocking | §1, §2 |
+| Bucket atomicity (non-PIT) | Non-preempting chunk production | §2, §3 |
```
dranikpg, line 733: Fixing non-pit requires much efforts and as such its simpler just to not think about it if we implement (4)

docs/shard-serialization.md
romange: Indeed the problem is mainly around huge values and that's why we need the tagged streams or logical locks around the serializer.

docs/shard-serialization.md
romange: we do not (or at least we should not) emit SINTERSTORE as a journal. That's something that @BorysTheDev  did 6 months ago - hopefully covering all of such commands.
Instead we should should emit it as "DEL set-output", "SADD ...." or alternatively as "RESTORE overwrite" entry.

docs/shard-serialization.md
romange: i agree. this is why I tried to structure the document mostly around PIT and move all non-PIT optimizations to the end.

docs/shard-serialization.md
romange: but I agree this should be in the document.

romange: Also, there was never a question about journal reorderings - which are not allowed. The scope here is around RDB / journal reorderings.

romange: verified that `SINTERSTORE` indeed rewrites the journal.

docs/shard-serialization.md
```diff
@@ -455,185 +557,443 @@ from running if `SerializeBucket` preempts (e.g., during large value serializati
...
+  [ordering invariant](#ordering-invariant) (see PR #6824).
+- [Non-transaction journal entries](#journal-entries-without-ondbchange) (expiry, eviction)
+  bypass `OnDbChange` entirely. Since there is no bucket completion state to consult, `DEL`
+  entries can interleave mid-serialization of the deleted key's baseline.
+
+**Goal.** Make "baseline fully emitted" precise for every bucket — including tiered values —
+so that ordering decisions can be expressed through per-bucket state rather than shard-wide mutex exclusion.
+
+**Approach.**
+- Introduce a per snapshot instance/bucket state machine:
```
kostasrim, line 643: I suspect this will greatly simplify `Heartbeat` flow as well. Why ? Because, as is we rely on the latch not being taken (`WillBlockOnJournalWrite`) or otherwise heartbeat is skipped ("Stalled").

With what is being proposed here, `Heartbeat` does not need to stall at all because both `FreeMemWithEvictionStep` && `DeleteExpiredStep` can just ignore the buckets whose state is `Serializing` or `DelayedPending ` (Covered and NotVisited do not affect us)

romange: true, but we could skip these states easily. but what this document suggests - even better, imho. use delayed deletions queue for buckets in "Serializing" state to gather such deletions and then `SerializeBucket` could emit journal writes once it finishes serializing.

docs/shard-serialization.md
```diff
@@ -455,185 +557,443 @@ from running if `SerializeBucket` preempts (e.g., during large value serializati
...
+     `OnDbChange` → mutation → `ConsumeJournalChange` on the same fiber). The real gap is
+     **tiered delayed entries** (see
+     [Delayed Serialization](#delayed-serialization-of-tiered-entities)): a journal entry
+     can reach the buffer before the delayed baseline is drained.
+   - **`OnDbChange` blocking:** block (fiber-aware wait) when the bucket is `Serializing` or
+     `DelayedPending`; proceed on `NotVisited` (serialize now → `Covered`) or `Covered`
+     (baseline already emitted). Because `OnDbChange` → mutation → `RecordJournal` →
+     `ConsumeJournalChange` is sequential on the mutation fiber, blocking `OnDbChange`
+     guarantees baseline-before-journal for all transaction-driven mutations.
+   - **Deferred deletion queue** for
```
kostasrim, line 851: Why not make these fibers (e.g heartbeat) skip the buckets that are under serialization/not reached baseline ? That way, we don't need to store `pending deletions` at all. These flows already relied on the latch to not be active so now we can instead rely on the bucket state (just like we skip for example keys with locktags in FreeMemWithEvictionStepAtomic). I suspect the bonus is that if multiple buckets are being serialized (because write heavy workloads will trigger OnDbChange bucket serialziation) we won't need the extra storage from vector<string> at all

romange: we can go either way indeed. it's very rare that a key collides with a serialized bucket so indeed it does not matter.

PR #6849, opened 2026-03-09
chore: Fix clang compilation issues for ZSetFamily structs

mkaruza:
Add constructor and copy constructor for Lex and LexBound. Compilation with clang 18 fails because of these missing.

### DIFF
--- a/src/server/zset_family.h
+++ b/src/server/zset_family.h
@@ -32,6 +32,9 @@ class ZSetFamily {
   struct Bound {
     double val;
     bool is_open = false;
+    Bound() = default;
+    Bound(double v, bool open) : val(v), is_open(open) {
+    }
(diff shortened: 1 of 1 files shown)


src/server/zset_family.h
```diff
@@ -32,13 +32,19 @@ class ZSetFamily {
   struct Bound {
     double val;
     bool is_open = false;
+    Bound() = default;
```
kostasrim, line 35: Why aren't structs around the codebase don't fail to compile ? Since bound is POD, default constructor is auto generated and aggregate initialization {} should work.

mkaruza: I don't know, when i compile locally with clang i see failure because these are missing.

PR #6851, opened 2026-03-09
Revert "refactor(string): migrate SET, APPEND, PREPEND to async SimpleContext"

dranikpg:
1. Reverts dragonflydb/dragonfly#6580 that restructured sync code into struct based async command
2. Uses coroutine based asynchronicity (see second commit)

### DIFF
--- a/src/server/string_family.cc
+++ b/src/server/string_family.cc
@@ -587,9 +587,9 @@ cmd::CmdR ExtendGeneric(CmdArgList args, CommandContext* cmd_cntx) {
     OpResult<bool> result = co_await cmd::SingleHopT(cb);
     MCRender render(cmd_cntx->mc_command()->cmd_flags);
     if (result) {
-      cmd_cntx->SendSimpleString(render.RenderStored(result.value()));
+      cmd_cntx->rb()->SendSimpleString(render.RenderStored(result.value()));
     } else {
-      cmd_cntx->SendError(result.status());
+      cmd_cntx->rb()->SendError(result.status());
     }
   }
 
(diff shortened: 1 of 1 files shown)


dranikpg:
augment review

dranikpg:
I'm running 

```
taskset -c 3,4 ./dfly_bench -P memcache_text --greet=false --proactor_threads 2 -c 4 --ratio=1:0 --test_time=20 -p 11211 --qps=0
```

agains dragonfly with 2 threads on cores 1,2 and I see around 5% latency increase and around 2% qps decrease

when I run 

```
taskset -c 3,4 ./dfly_bench -P memcache_text --greet=false --proactor_threads 2 -c 4 --ratio=1:0 --test_time=20 -p 11211 --qps=0 --pipeline=10
```

I see no difference. I tried multiple times - it seems that allocations might have a visible (1-2%) trace, but I'll need to profile it for that

src/server/cmd_support.cc
```diff
@@ -65,17 +65,17 @@ bool SingleHopWaiter::await_ready() noexcept {
 void SingleHopWaiter::await_suspend(std::coroutine_handle<> handle) const noexcept {
   // TODO: functor calling resume is double indirection and wasted space for cleaner
   absl::Cleanup cleaner = [handle] { handle.destroy(); };
-  cmd_cntx->Resolve(blocker, [handle, cleaner = std::move(cleaner)](auto* rb) mutable {
-    handle.resume();
+  cmd_cntx->Resolve(blocker, [handle, ran = false, cleaner = std::move(cleaner)](auto* rb) mutable {
     std::move(cleaner).Cancel();
+    handle.resume();
```
romange, line 70: care to explain? what path does trigger this cleaner?

dranikpg: just dropping the parsed cmd object when closing the connection - we need to destroy the coroutine frame beucase coroutine_handle is just a non owning type

romange: is cleaner run when callback is dropped and not run?
please add comments there

Issue #6830, opened 2026-03-08
shard-serialization P1.3: introduce bucket-level completion state machine


romange:
# shard-serialization P1.3: SerializerBase class with bucket-level state machine

## Phase 1 — PIT and PIT+tiered foundation (Task 3 of 17)

Goal: Factor the duplicated `OnDbChange` / version-stamping pattern out of
`SliceSnapshot` and `RestoreStreamer` into a shared base class
(`SerializerBase`) that owns callback lifecycle and a per-bucket state machine
tracking each bucket through `NotVisited -> Serializing -> DelayedPending -> Covered`.

Design doc: [docs/shard-serialization.md — Technical Roadmap](https://github.com/dragonflydb/dragonfly/blob/main/docs/shard-serialization.md#technical-roadmap)

### Dependencies

- #6828 (P0.1: document invariants)
- #6829 (P0.2: observability)

### Design decisions

- **Non-PIT mode**: stamp bucket versions uniformly. Removes
  `use_snapshot_version_` branching — state machine works for both modes.
- **Passive tracking**: state tracking is observational (DCHECKs, stats) —
  no behavioral changes in the initial PRs.

---

## Class hierarchy

```
SerializerBase               (NEW base class)
   ^            ^
   |            |
SliceSnapshot   RestoreStreamer -> JournalStreamer -> JournalConsumerInterface
(also: JournalConsumerInterface)
```

`SerializerBase` does NOT inherit from `JournalConsumerInterface` — no diamond.

---

## BucketIdentity

`using BucketIdentity = uintptr_t` — the physical bucket's memory address,
unique across all databases/segments. Obtained via a new
`bucket_iterator::bucket_address()` accessor on `DashTable::Iterator`
(returns `&segment_[seg_id_]->GetBucket(bucket_id_)`).

Stable for the lifetime of a serialization: once version-stamped, mutations
that could trigger segment splits are blocked.

---

## Bucket state machine

Only transient states stored in map. `NotVisited` and `Covered` are implicit:

| State | Encoding |
|-------|----------|
| **NotVisited** | `version < snapshot_version_` AND absent from map |
| **Serializing** | present in map, `phase == Serializing` |
| **DelayedPending** | present in map, `phase == DelayedPending` |
| **Covered** | `version >= snapshot_version_` AND absent from map |

`BucketState` is a struct holding the phase enum and a
`vector<TieredDelayedEntry>` for tiered entries whose reads are still
[...]

PR #6854, opened 2026-03-10
chore: introduce SerializerBase

kostasrim:
* add SerilizerBase and bucket state machine
* add DashTable::Iterator::bucket_address()
* add state machine tests

resolves PR1 of #6830

### DIFF
--- a/.claude/skills/reproduce-fuzz-crash/SKILL.md
+++ b/.claude/skills/reproduce-fuzz-crash/SKILL.md
@@ -1,0 +1,134 @@
+---
+name: reproduce-fuzz-crash
+description: >
+  Reproduce AFL++ fuzz crashes from GitHub Actions. Use when user provides a
+  GitHub Actions fuzz run URL and wants to reproduce and analyze the crash locally.
+argument-hint: <github-actions-run-url>
+allowed-tools: Bash, Read, Grep, Glob, Write
+---
+
+# Reproduce Fuzz Crash
+
+Given a GitHub Actions fuzz run URL, download crash artifacts, triage them
+with `fuzz/triage_crashes.sh`, and produce a crash analysis report.
+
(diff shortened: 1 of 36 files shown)


src/server/serializer_base_test.cc
```diff
@@ -0,0 +1,88 @@
+// Copyright 2026, DragonflyDB authors.  All rights reserved.
+// See LICENSE for licensing terms.
+//
+
+#include "server/serializer_base.h"
```
romange, line 5: is it really useful? I do not want to add tests just for sake of it.

kostasrim: Yes and no. It goes over the simple changes in the state machine + the issue explicitly asked for that. I generally like e2e tests but it doesn't hurt to have a base for the state changes 🤷‍♂️

src/server/serializer_base_test.cc
```diff
@@ -0,0 +1,88 @@
...
+
+#include "base/gtest.h"
+#include "base/logging.h"
+#include "server/test_utils.h"
+
+namespace dfly {
+
+class SerializerBaseTest : public BaseFamilyTest, public SerializerBase {
```
romange, line 13: why `BaseFamilyTest` is needed?

src/server/serializer_base_test.cc
romange: the issue was generated by AI and maybe I missed it. 
if this test file won't provide real tests at the end - we will get rid of it.

src/server/serializer_base_test.cc
kostasrim: linker errors :( (you can see the previous run failures)

src/server/serializer_base_test.cc
kostasrim: I am not married to the tests -- whatever works 😄

PR #6855, opened 2026-03-10
fix: make SwitchState return previous state to prevent double DFLY LOAD

romange:
## Summary

`SwitchState(from, to)` previously returned the **new/current** state, making the return value ambiguous: `SwitchState(ACTIVE, LOADING)` returned `LOADING` both when the switch succeeded (state was `ACTIVE`) and when it was already `LOADING` (no switch performed). Callers checked `!= LOADING` to detect failure, but this passed in both cases — allowing a second `DFLY LOAD` to enter the load flow while one was already running.

## Fix

Change `SwitchState` to return the **previous** state (before the attempt):
- Switch succeeded → returns `from`
- Switch failed → returns the unchanged current state (≠ `from`)

Callers now check `prev_state != ACTIVE` to unambiguously detect that the switch did not happen.

`RequestLoadingState` is updated to allow piggybacking when already in `LOADING` state (`prev == ACTIVE || prev == LOADING`), preserving the ref-counted behavior for concurrent replica syncs.

## Changes

- `SwitchState`: save and return `prev` instead of `to`/`global_state_`
- `RequestLoadingState`: check `prev == ACTIVE || prev == LOADING`
- `ServerFamily::Load`: `prev_state != ACTIVE` (was `!= LOADING`)
- `ServerFamily::ReplicaOfInternal`: same
- `DflyCmd::TakeOver`: `prev_state != ACTIVE` (was `!= TAKEN_OVER`)
- Updated `SwitchState` doc comment in `main_service.h`

### DIFF
--- a/src/server/main_service.h
+++ b/src/server/main_service.h
@@ -65,9 +65,10 @@ class Service : public facade::ServiceInterface {
 
   facade::ErrorReply ReportUnknownCmd(std::string_view cmd_name) ABSL_LOCKS_EXCLUDED(mu_);
 
-  // Returns: the new state.
-  // if from equals the old state then the switch is performed "to" is returned.
-  // Otherwise, does not switch and returns the current state in the system.
+  // Attempts to switch global state from 'from' to 'to'.
+  // Returns the PREVIOUS global state (before the switch attempt).
+  // If from equals the previous state then the switch is performed and 'from' is returned.
+  // Otherwise, does not switch and returns the current (unchanged) state.
   // Upon switch, updates cached global state in threadlocal ServerState struct.
   GlobalState SwitchState(GlobalState from, GlobalState to) ABSL_LOCKS_EXCLUDED(mu_);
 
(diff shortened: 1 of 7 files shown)


src/server/main_service.h
```diff
@@ -65,9 +65,10 @@ class Service : public facade::ServiceInterface {
...
 
-  // Returns: the new state.
-  // if from equals the old state then the switch is performed "to" is returned.
-  // Otherwise, does not switch and returns the current state in the system.
+  // Attempts to switch global state from 'from' to 'to'.
+  // Returns the PREVIOUS global state (before the switch attempt).
+  // If from equals the previous state then the switch is performed and 'from' is returned.
+  // Otherwise, does not switch and returns the current (unchanged) state.
   // Upon switch, updates cached global state in threadlocal ServerState struct.
   GlobalState SwitchState(GlobalState from, GlobalState to) ABSL_LOCKS_EXCLUDED(mu_);
```
dranikpg, line 73: that's a really tricky return value, something like `pair<current state, bool switched>` woudl've done it as well (akin to insert() result)

romange: I also had doubts. At the end I saw I can preserve the old semantics with the new API so I chose this option.

romange: it's like CAS though

PR #6858, opened 2026-03-10
chore(server): Optimize coroutine code

dranikpg:
1. Optimize double indirection by waking up directly std::coroutine_handle from the connection loop
2. Allow avoiding allocations by using a 256 byte buffer for synchronous dispatches


dranikpg:
Interestingly (2) turned out to be primarily ineffective

PR #6859, opened 2026-03-10
Add Claude Code skill for reproducing fuzz crashes

romange:
## Summary
- Adds a `reproduce-fuzz-crash` Claude Code skill that automates downloading crash artifacts from GitHub Actions fuzz runs, replaying them locally, and producing structured crash reports
- Checks for existing binary before building, decodes crash inputs with both `strings` and `xxd`, reads source at crash location for better analysis

## Test plan
- [x] Tested against https://github.com/dragonflydb/dragonfly/actions/runs/22906484769?pr=6855 — crash reproduced successfully

### DIFF
--- a/.claude/skills/reproduce-fuzz-crash/SKILL.md
+++ b/.claude/skills/reproduce-fuzz-crash/SKILL.md
@@ -9,8 +9,8 @@ allowed-tools: Bash, Read, Grep, Glob, Write
 
 # Reproduce Fuzz Crash
 
-Given a GitHub Actions fuzz run URL, download crash artifacts, replay them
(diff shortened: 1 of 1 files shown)


romange:
> Fine with me, but I prefer to use the triage script approach. https://github.com/dragonflydb/dragonfly/blob/main/fuzz/triage_crashes.sh
> 
> It can reproduce and detect false positives for single dragonfly instances.



> Fine with me, but I prefer to use the triage script approach. https://github.com/dragonflydb/dragonfly/blob/main/fuzz/triage_crashes.sh
> 
> It can reproduce and detect false positives for single dragonfly instances.

fixed

PR #6860, opened 2026-03-10
chore: SliceSnapshot inherits from SerializerBase

kostasrim:
* use SerializerBase as base class from SliceSnapshot
* remove non-PIT code

resolves PR2 of https://github.com/dragonflydb/dragonfly/issues/6830

### DIFF
--- a/src/server/search/doc_index.cc
+++ b/src/server/search/doc_index.cc
@@ -695,6 +695,8 @@ bool ShardDocIndex::Matches(string_view key, unsigned obj_code) const {
 
 optional<ShardDocIndex::LoadedEntry> ShardDocIndex::LoadEntry(DocId id,
                                                               const OpArgs& op_args) const {
+  if (!key_index_.IsValid(id))
+    return std::nullopt;
   auto& db_slice = op_args.GetDbSlice();
   string_view key = key_index_.Get(id);
   auto it = db_slice.FindReadOnly(op_args.db_cntx, key, base_->GetObjCode());
--- a/src/server/search/search_family_test.cc
+++ b/src/server/search/search_family_test.cc
@@ -8,6 +8,7 @@
 #include <absl/strings/str_format.h>
 
 #include <algorithm>
+#include <atomic>
 #include <string_view>
 
 #include "base/gtest.h"
@@ -4526,6 +4527,89 @@ TEST_F(SearchFamilyTest, NumericIndexRejectsNonFiniteValues) {
   // Non-finite docs are not in the numeric index; only doc:5 should match the range query.
   auto resp = Run({"FT.SEARCH", "idx", "@val:[-inf +inf]"});
   EXPECT_THAT(resp, RespArray(ElementsAre(IntArg(1), "doc:5", _)));
+}
+
+class HnswRaceTest : public BaseFamilyTest {
+ protected:
+  HnswRaceTest() {
+    num_threads_ = 4;
+  }
+};
+
+TEST_F(HnswRaceTest, HnswKnnDeleteRaceCrash) {
+  constexpr int kDim = 128;
+  constexpr int kNumDocs = 5000;
+  // Only delete the K nearest docs to the query: these are guaranteed to appear
+  // in every KNN result, so every deletion is a potential crash trigger.
+  constexpr int kK = 200;
+
+  auto make_vec = [&](float seed) -> std::string {
+    std::string s(kDim * sizeof(float), '\0');
+    for (int j = 0; j < kDim; ++j) {
+      float v = seed + static_cast<float>(j) * 0.001f;
+      memcpy(s.data() + j * sizeof(float), &v, sizeof(float));
+    }
+    return s;
+  };
+
+  Run({"FT.CREATE",
+       "hnsw_race_idx",
+       "ON",
+       "HASH",
+       "PREFIX",
+       "1",
+       "doc:",
+       "SCHEMA",
+       "vec",
+       "VECTOR",
+       "HNSW",
+       "10",
+       "TYPE",
+       "FLOAT32",
+       "DIM",
+       absl::StrCat(kDim),
+       "DISTANCE_METRIC",
+       "L2",
+       "M",
+       "16",
+       "EF_CONSTRUCTION",
+       "200"});
+
+  for (int i = 0; i < kNumDocs; ++i)
+    Run({"HSET", absl::StrCat("doc:", i), "vec", make_vec(static_cast<float>(i))});
+
+  const std::string kQueryVec = make_vec(0.0f);
+  std::atomic<bool> done{false};
+
+  auto search_fiber = pp_->at(3)->LaunchFiber([&] {
+    for (int i = 0; i < 50 && !done.load(); ++i) {
+      auto resp = Run({"FT.SEARCH", "hnsw_race_idx",
+                       absl::StrCat("*=>[KNN ", kK, " @vec $vec EF_RUNTIME 50000]"), "PARAMS", "2",
+                       "vec", kQueryVec, "DIALECT", "2"});
+      EXPECT_NE(resp.type, RespExpr::ERROR);
+    }
+    done.store(true);
+  });
+
+  auto make_del_fiber = [&](int thread_idx) {
+    return pp_->at(thread_idx)->LaunchFiber([&, thread_idx] {
+      while (!done.load()) {
+        for (int i = 0; i < kK && !done.load(); ++i) {
+          Run({"DEL", absl::StrCat("doc:", i)});
+          Run({"HSET", absl::StrCat("doc:", i), "vec", make_vec(static_cast<float>(i))});
+        }
+      }
+    });
+  };
+
+  auto del0 = make_del_fiber(0);
+  auto del1 = make_del_fiber(1);
+  auto del2 = make_del_fiber(2);
+
+  search_fiber.Join();
+  del0.Join();
+  del1.Join();
+  del2.Join();
 }
 
 }  // namespace dfly
--- a/src/server/string_family.cc
+++ b/src/server/string_family.cc
@@ -720,15 +720,6 @@ cmd::CmdR ExtendGeneric(CmdArgList args, CommandContext* cmd_cntx) {
   co_return std::nullopt;
 }
 
-// Wrapper to call SetCmd::Set in ScheduleSingleHop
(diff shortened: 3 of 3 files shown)


src/server/snapshot.cc
```diff
@@ -279,33 +264,33 @@ void SliceSnapshot::IterateBucketsFb(bool send_full_sync_cut) {
...
       if (use_background_mode_) {
         // Yielding for background fibers has low overhead if the time slice isn't used up.
         // Do it after every bucket for maximum responsiveness.
         DCHECK(ThisFiber::Priority() == fb2::FiberPriority::BACKGROUND);
         ThisFiber::Yield();
+        PushDelayedEntries(false, nullptr);
         PushSerialized(false);
       } else {
+        PushDelayedEntries(false, nullptr);
         if (!PushSerialized(false)) {
```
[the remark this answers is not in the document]
kostasrim: This is not true. Tiering are `strings` so they won't preempt middle serialization and since we `PushDelayedEntries(false, nullptr)` we gurantee not to block on `unresolved tiered reads`

src/server/snapshot.cc
```diff
@@ -315,7 +300,7 @@ void SliceSnapshot::IterateBucketsFb(bool send_full_sync_cut) {
     } while (snapshot_cursor_);
 
     DVLOG(2) << "after loop " << ThisFiber::GetName();
-    // Wait for all the outstanding delayed entries and serialize them as well.
+    // Flush all remaining delayed tiered entries for this database.
     PushDelayedEntries(true, nullptr);
```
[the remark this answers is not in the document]
kostasrim: I didn't change this + we will get rid of big_value_mu_ soon anyway

src/server/snapshot.cc
```diff
@@ -490,128 +456,71 @@ bool SliceSnapshot::PushSerialized(bool force) {
...
+    auto res = entry->value.Get();  // blocks if not resolved and force=true
+    if (!res.has_value()) {
+      LOG(ERROR) << "Failed to read delayed entry for key " << entry->key.ToString();
       continue;
     }
-    serialize_entry(it++);
+
+    PrimeValue pv{*res};
+    serializer_->SaveEntry(entry->key, pv, entry->expire, entry->mc_flags, entry->dbid);
+
```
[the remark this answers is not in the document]
kostasrim: we should report an error to the context

src/server/snapshot.h
```diff
@@ -114,98 +114,77 @@ class SliceSnapshot : public journal::JournalConsumerInterface {
...
   std::unique_ptr<RdbSerializer> serializer_;
+  absl::flat_hash_map<std::string, std::unique_ptr<TieredDelayedEntry>> delayed_entries_;
 
-  // Delayed entries that are waiting for tiered storage reads to complete before they can be
-  // serialized.
-  absl::flat_hash_map<TieredDelayEntryKey, std::unique_ptr<TieredDelayedEntry>> delayed_entries_;
+  // Accumulates the tiered keys found in the most recent DoSerializeBucket call.
+  // DoSerializeBucket is always invoked under big_value_mu_, so writes here are safe.
+  // OnChange reads this after the base call to force-flush the relevant delayed entries.
+  absl::flat_hash_set<std::string> tiered_keys_on_change_;
```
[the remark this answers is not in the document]
kostasrim: I will turn a blind eye on this one on purpose. we will move to a per bucket pending list anyway

src/server/snapshot.h
```diff
@@ -114,98 +114,78 @@ class SliceSnapshot : public journal::JournalConsumerInterface {
...
   std::unique_ptr<RdbSerializer> serializer_;
-
-  // Delayed entries that are waiting for tiered storage reads to complete before they can be
-  // serialized.
   absl::flat_hash_map<TieredDelayEntryKey, std::unique_ptr<TieredDelayedEntry>> delayed_entries_;
 
+  // Accumulates the tiered keys found during DoSerializeBucket calls.
+  // OnChange reads this after the base call to force-flush the relevant delayed entries
+  // under big_value_mu_ before any journal mutation can interleave.
+  absl::flat_hash_set<TieredDelayEntryKey> tiered_keys_on_change_;
```
romange: why do you need to be a set type?

src/server/journal/cmd_serializer.cc
```diff
@@ -127,33 +127,43 @@ size_t CmdSerializer::SerializeEntry(string_view key, const PrimeKey& pk, const
...
 
     // If tiered_keys filter is provided, only serialize matching keys
-    // Compare the string key from the map with the keys in tiered_keys set
-    if (tiered_keys && !tiered_keys->contains(it->first)) {
+    if (tiered_keys && !tiered_keys->contains(key)) {
       ++it;
       continue;
     }
 
+    // Extract before blocking on the future so no delayed_entries_ iterator survives a yield.
```
romange: can you explain the comment?

src/server/serializer_base.cc
```diff
@@ -14,8 +14,23 @@ SerializerBase::SerializerBase(DbSlice* slice) : db_slice_(slice) {
 SerializerBase::~SerializerBase() {
 }
 
+// Ordering invariant (both modes):
```
romange: you moved this comment but it's not `OnDbChange` here.

src/server/snapshot.cc
```diff
@@ -491,131 +455,75 @@ bool SliceSnapshot::PushSerialized(bool force) {
...
 
-    // If we have serialized enough data we should push it to avoid building
-    // up a large blob in memory.
-    PushSerialized(false);
-  };
-
-  // When tiered_keys are provided, we should serialize the entries matching the keys.
-  if (bucket_tiered_keys) {
-    for (const auto& key : *bucket_tiered_keys) {
+  // First, directly extract entries matching the provided key filter (O(k)).
```
romange: why the previous code was not good?

src/server/snapshot.cc
```diff
@@ -401,6 +382,15 @@ unsigned SliceSnapshot::SerializeBucket(DbIndex db_index, PrimeTable::bucket_ite
   return result;
 }
 
+unsigned SliceSnapshot::DoSerializeBucket(DbIndex db_index, PrimeTable::bucket_iterator it) {
+  return SerializeBucket(db_index, it, false);
```
kostasrim, line 386: To preserve semantics for now

src/server/snapshot.cc
```diff
@@ -539,83 +529,21 @@ void SliceSnapshot::PushDelayedEntries(bool force,
...
-  uint8_t depth = db_slice_->GetTables(id).first->depth();
-
-  return id < snapshot_db_index_ ||
-         (id == snapshot_db_index_ &&
-          (cursor.bucket_id() < snapshot_cursor_.bucket_id() ||
-           (cursor.bucket_id() == snapshot_cursor_.bucket_id() &&
-            cursor.segment_id(depth) < snapshot_cursor_.segment_id(depth))));
-}
-
-void SliceSnapshot::OnMoved(DbIndex id, const DbSlice::MovedItemsVec& items) {
```
kostasrim, line 587: Ditching the non-pit paths

src/server/journal/cmd_serializer.cc
kostasrim: removed

src/server/snapshot.h
kostasrim: removed

src/server/snapshot.cc
kostasrim: reverted

kostasrim:
@romange should be clean now. Ignore the failure in the fuzzer it's completely unrelated

src/server/serializer_base.h
```diff
@@ -91,6 +91,12 @@ class SerializerBase {
   // Called while big_value_mu_ is held.
   virtual unsigned DoSerializeBucket(DbIndex db_index, PrimeTable::bucket_iterator it) = 0;
 
+  // Serialize a single bucket from the OnChange path.  Defaults to DoSerializeBucket().
+  // Subclasses can override when the callback path must preserve behavior that differs
+  // from the main traversal (for example, forcing delayed tiered entries for the bucket
+  // to be flushed before releasing big_value_mu_).
+  virtual unsigned DoSerializeBucketOnChange(DbIndex db_index, PrimeTable::bucket_iterator it);
```
romange, line 102: Do we really need two virtual functions here, with one calling another?
Also, looked at SliceSnapshot - they both calling the same function as well. So why this duplication is needed?

src/server/snapshot.cc
```diff
@@ -547,75 +538,14 @@ void SliceSnapshot::SerializeExternal(DbIndex db_index, PrimeKey pk, const Prime
...
-// PIT mode: enforced by serialize-before-mutate. OnDbChange serializes the bucket before the
-//   mutation commits; ConsumeJournalChange runs after the mutation on the same fiber, so the
-//   baseline is always first. big_value_mu_ prevents interleaving with the traversal fiber's
-//   SerializeBucket (which can preempt via consume_fun_).
-//
-// Non-PIT mode: OnDbChange only acquires big_value_mu_ as a barrier — no serialization. The
-//   mutex prevents journaling mutations from slipping in the middle of bucket serialization
-//   on the traversal fiber — see ConsumeJournalChange for details. OnMoved handles items
-//   displaced across the traversal cursor.
-void SliceSnapshot::OnDbChange(DbIndex db_index, const DbSlice::ChangeReq& req) {
```
romange, line 556: why these comments were deleted?

src/server/serializer_base.h
kostasrim: yes to preserve tiering semantics. One calls SerializeBucket(arg, arg true) and the other SerializeBucket(arg. arg, false). 

I will ditch this function once we bake in the per bucket behaviour

romange: Could be helpful to explain that it's temporary as it's confusing.

src/server/snapshot.cc
kostasrim: Uh I thought I replied :/ So I removed `onDbChange` since this was replaced by `OnChange()` in the SerializerBased.

I moved part of the comment to explain this when we register in the serializer base. 

Note that non-pit is removed here and the comment was slightly modified.

romange:
this needs to be resolved

dranikpg:
Rebased

dranikpg:
This was a (most likely) unrelated tiering failure, I just didn't restart the cancelled jobs

Issue #6856, opened 2026-03-10
Crash in cluster mode


vyavdoshenko:
[crash-000000.tar.gz](https://github.com/user-attachments/files/25871968/crash-000000.tar.gz)

Terminal 0:
```
./build/dragonfly --port 6380 --logtostderr --proactor_threads 1 --dbfilename="" \
    --cluster_mode=yes --cluster_node_id=0000000000000000000000000000000000000000
```

Terminal 1:
```
./build/dragonfly --port 6379 --logtostderr --proactor_threads 1 --dbfilename="" \
    --cluster_mode=yes --cluster_node_id=1111111111111111111111111111111111111111 --hz=0
```

Terminal 2:
```
tar xzf crash-000000.tar.gz
cd crash-000000
python3 replay_crash.py crashes 000000
```

Crash log:
```
F20260310 15:51:54.024091 36293 init.cc:140] Uncaught exception: bitset::set: __position (which is 16384) >= _Nb (which is 16384)
*** Check failure stack trace: ***
    @     0xaaaaacb60d90  google::LogMessage::Fail()
    @     0xaaaaacb60c98  google::LogMessage::SendToLog()
    @     0xaaaaacb60498  google::LogMessage::Flush()
    @     0xaaaaacb641a0  google::LogMessageFatal::~LogMessageFatal()
    @     0xaaaaababd124  _ZZN13MainInitGuardC4EPiPPPcjENKUlvE_clEv
    @     0xaaaaababd238  _ZZN13MainInitGuardC4EPiPPPcjENUlvE_4_FUNEv
    @     0xfffff780a570  (unknown)
    @     0xfffff780a570  (unknown)
*** SIGABRT received at time=1773150714 on cpu 0 ***
PC: @     0xfffff7537608  (unknown)  (unknown)
    @     0xaaaaacbc3a84         64  absl::lts_20250512::WriteFailureInfo()
    @     0xaaaaacbc3d9c         96  absl::lts_20250512::AbslFailureSignalHandler()
    @     0xfffff7ffa7d0       4752  (unknown)
    @     0xfffff74ecb3c         64  raise
    @     0xfffff74d7e00        192  abort
    @     0xaaaaacb6c2a8        208  google::DumpStackTraceAndExit()
    @     0xaaaaacb60d90         16  google::LogMessage::Fail()
    @     0xaaaaacb60c98        144  google::LogMessage::SendToLog()
[...]

PR #6861, opened 2026-03-10
fix(cluster): Validate slot IDs in DFLYCLUSTER FLUSHSLOTS to prevent out-of-bounds crash

vyavdoshenko:
- `DFLYCLUSTER FLUSHSLOTS` did not validate that slot IDs are within the valid range [0, 16383]
- Passing `end=16384` caused `SlotSet::Set()` to call `std::bitset<16384>::set(16384)`, which throws `std::out_of_range` and crashes the server
- Added bounds check

Fixes: #6856

### DIFF
--- a/src/server/cluster/cluster_family.cc
+++ b/src/server/cluster/cluster_family.cc
@@ -730,11 +730,13 @@ void ClusterFamily::DflyClusterFlushSlots(CmdArgList args, CommandContext* cmd_c
 
   CmdArgParser parser(args);
   do {
-    auto [slot_start, slot_end] = parser.Next<SlotId, SlotId>();
+    auto [slot_start, slot_end] = parser.Next<ParsedSlotId, ParsedSlotId>();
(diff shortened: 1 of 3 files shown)


src/server/cluster/cluster_family.cc
```diff
@@ -731,6 +731,12 @@ void ClusterFamily::DflyClusterFlushSlots(CmdArgList args, CommandContext* cmd_c
   CmdArgParser parser(args);
   do {
     auto [slot_start, slot_end] = parser.Next<SlotId, SlotId>();
+    if (slot_start > kMaxSlotNum || slot_end > kMaxSlotNum) {
+      return cmd_cntx->SendError("Invalid slot id");
+    }
+    if (slot_start > slot_end) {
+      return cmd_cntx->SendError("Invalid slot range");
+    }
```
dranikpg: if we use it in more places it would be cool to make a type that is natively usable by the parser, i.e. Next<SlotId> is already validated

PR #6863, opened 2026-03-11
test: make test_hnsw_search_replication_with_network_disruptions stable

BorysTheDev:
Summary: This PR re-enables and stabilizes the HNSW replication test under network disruptions by making result verification tolerant to approximate vector search.

Changes:

- Improves the TCP proxy to gracefully close the client connection when the remote endpoint can’t be reached
- Unskips test_hnsw_search_replication_with_network_disruptions and increases initial seeded documents
- Adds a doc_id TAG field to the HNSW test index/documents to enable deterministic filtered checks
- Updates HNSW verification to compare KNN outputs loosely and confirm suspected “missing” docs via filtered vector search

Technical Notes: The new verification avoids requiring identical KNN ordering between master/replica by using TAG-filtered queries as an index-presence check.

### DIFF
--- a/tests/dragonfly/proxy.py
+++ b/tests/dragonfly/proxy.py
@@ -18,6 +18,7 @@ class Proxy:
             )
         except OSError:
             writer.close()
+            await writer.wait_closed()
             return
 
         async def forward(reader, writer):


tests/dragonfly/replication_test.py
```diff
@@ -4122,8 +4122,6 @@ async def test_sbf_chunked_replication_over_4gb(df_factory: DflyInstanceFactory)
     assert await c_replica.execute_command("BF.EXISTS", "bf", "hello") == 1
 
 
-@pytest.mark.skip("HNSW index replication hasn't finished yet")
-@pytest.mark.large
```
vyavdoshenko, line 4126: How long does it take? Why did you remove it?

BorysTheDev: On my laptop 2-3 seconds. I want run it every time to be sure that everything works

PR #6864, opened 2026-03-11
chore: another way to track per object memory

romange:
Before: we estimated average object size by using global metrics like used_memory and table memory

Now as we have rather precise computation of object memory per db table, switch to use those stats.
Add warnings if both metrics have huge discrepancies (which may happen).

### DIFF
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -73,8 +73,6 @@ void AccountObjectMemory(string_view key, unsigned type, int64_t size, DbTable* 
     return;
 
   DbTableStats& stats = db->stats;
-  DCHECK_GE(static_cast<int64_t>(stats.obj_memory_usage) + size, 0)
-      << "Can't decrease " << size << " from " << stats.obj_memory_usage;
 
   stats.AddTypeMemoryUsage(type, size);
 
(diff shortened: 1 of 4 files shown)


romange:
augment review

src/server/db_slice.cc
```diff
@@ -146,27 +144,44 @@ bool PrimeEvictionPolicy::CanGrow(const PrimeTable& tbl) const {
...
+
+  size_t obj_memory_usage = db_slice_->GetDBTable(cntx_.db_index)->stats.obj_memory_usage;
+  size_t avg_obj_size = obj_memory_usage / tbl.size();
+
+  // Catch significant discrepancies in average object size estimation.
+  // Note that this may happen if for example, db0 hosts a lot of small keys,
+  // db1 hosts huge keys etc. The goal of this comparison is to detect these cases and
+  // confirm that discrepancy is justified. Once we gather empirical evidence,
+  // we can remove this check and drop `db_slice_->bytes_per_object()` computation entirely.
+  if (avg_obj_size * 20 < db_slice_->bytes_per_object() ||
```
BorysTheDev, line 163: what is 20?

BorysTheDev: Could you give it a name

romange: `The goal of this comparison is to detect these cases and confirm that discrepancy is justified`

PR #6438, opened 2026-01-21
feat: Add Count-Min Sketch (CMS) support

Hugoch:
This PR adds data structures and commands to support Redis-like Count-Min Sketch (CMS) with a multi-shard merge implementation.

Not a C++ expert here and very new to this codebase. PR was AI-assisted, I may have missed obvious stuff.

## Performances

I compared performances against Redis 8.4.0 on a AWS `m5d.24xlarge`:
- Initialization performances are way worse than Redis as it uses `calloc`to get kernel pre-zeroed pages.
- CMS.MERGE needs multiple shard hops to gather the sketches to merge, so lower performances are expected

```
=========================================================================================================
                           Server Comparison Results
=========================================================================================================
Server 1: Redis 8.4.0 | Server 2: Dragonfly

---------------------------------------------------------------------------------------------------------
                                Throughput (ops/sec)
---------------------------------------------------------------------------------------------------------
Command          Scenario                Redis        Dragonf           Diff      Ratio
---------------------------------------------------------------------------------------------------------
CMS.INITBYDIM    100x5                  19418.00      6835.32      -12582.68      0.35x
CMS.INITBYDIM    1000x7                 20034.71      5403.69      -14631.02      0.27x
CMS.INITBYDIM    10000x10               19868.41      1876.09      -17992.32      0.09x
CMS.INITBYPROB   err=0.01,prob=0.01     18168.57      5925.08      -12243.49      0.33x
CMS.INCRBY       single                 19796.87     11732.67       -8064.20      0.59x
CMS.INCRBY       batch_10               13150.28     10799.49       -2350.79      0.82x
CMS.INCRBY       batch_100               3094.24      2452.17        -642.07      0.79x
CMS.QUERY        single                 20439.92     14141.70       -6298.22      0.69x
CMS.QUERY        batch_10               15295.18      8379.16       -6916.02      0.55x
CMS.QUERY        batch_100               4566.28      4230.23        -336.05      0.93x
CMS.MERGE        2_sources               6482.11      2803.88       -3678.23      0.43x
CMS.MERGE        2_sources_weighted      6355.53      2950.29       -3405.24      0.46x
CMS.INFO         populated              19885.94     11942.68       -7943.26      0.60x

---------------------------------------------------------------------------------------------------------
                                P99 Latency (ms)
---------------------------------------------------------------------------------------------------------
Command          Scenario                Redis         Dragonf           Diff      Ratio
---------------------------------------------------------------------------------------------------------
[...]

PR #6867, opened 2026-03-11
chore(core): add CMS implementation with unit tests

romange:
## Summary
- Add Count-Min Sketch (CMS) data structure with PMR memory allocation
- Fix memory leaks in destructor and move-assignment operator
- Add `ErrorRateTag` tag-dispatched constructor for init-by-probability
- Add `total_count()` accessor for CMS.INFO compatibility
- Add 19 unit tests covering all CMS operations, inspired by fakeredis test cases

## Changes
- `src/core/cms.h` / `cms.cc`: CMS class with two constructors (width/depth and ErrorRateTag), proper PMR allocate/deallocate, move semantics
- `src/core/cms_test.cc`: Unit tests for init, IncrBy, Query, Merge (with weights), move semantics, dimension mismatch, total_count tracking
- `src/core/CMakeLists.txt`: Register cms.cc in dfly_core and cms_test

## Test plan
- [x] All 19 cms_test cases pass
- [x] Pre-commit formatting passes

### DIFF
--- a/src/core/cms.cc
+++ b/src/core/cms.cc
@@ -1,0 +1,118 @@
+// Copyright 2026, DragonflyDB authors.  All rights reserved.
+// See LICENSE for licensing terms.
+//
+
+#include "core/cms.h"
+
+#include <xxhash.h>
+
+#include <algorithm>
+#include <cmath>
+
+#include "base/logging.h"
+
+namespace dfly {
+namespace {
+
+uint32_t Offset(uint64_t h1, uint64_t h2, uint32_t row, uint32_t width) {
+  uint32_t idx = static_cast<uint32_t>((h1 + (row * h2)) % width);
+  return row * width + idx;
+}
+
+}  // namespace
+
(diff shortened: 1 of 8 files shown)


romange:
based on #6438

src/core/cms.cc
```diff
@@ -0,0 +1,118 @@
...
+    depth_ = other.depth_;
+    mr_ = other.mr_;
+    count_ = other.count_;
+    counters_ = other.counters_;
+    other.width_ = 0;
+    other.depth_ = 0;
+    other.count_ = 0;
+    other.counters_ = nullptr;
+  }
+  return *this;
```
BorysTheDev, line 71: I prefer doing `std::swap(std::tie(this.members,...), std::tie(other.members,...)); `
Pros: we don't need to check for this, and delete anything.
Cons: Deallocation will be done later for "this" previous state.

romange: nice trick. Both claude and me did not know it.

PR #6871, opened 2026-03-12
fix(regtests): Make Python 3.9 available

dranikpg:
Needs https://github.com/romange/container-foundry/pull/30 to update the container and have a more recent python version out of the box (3.10)


romange:
I fixed the error with python3.9 code.  I do not want to update now - we should just stop using ubuntu20 for tests

PR #6872, opened 2026-03-12
feat: add extract method into StringMap

BorysTheDev:
Summary: This PR introduces “extract/detach” APIs to remove entries from a StringMap (and its underlying DenseSet) while transferring ownership to the caller.

Changes:

- Added DenseSet::Detach and DenseSet::DetachInternal to unlink an element and return the raw object without calling ObjDelete.
- Extended StringMap with Extract() to remove a field and return its backing sds entry (key+value blob) without freeing it.
- Added AddOrUpdateAndExtract() to return the previous sds entry on updates, enabling callers to manage/free old values explicitly.
- Introduced StringMap::DeleteEntry() as the supported way to free extracted sds entries.
- Added unit tests covering extract, replace-with-extract, TTL behavior, and multiple extractions.
- Adjusted an HNSW replication disruption test to run under multiple master/replica thread-count combinations.

Technical Notes: The new APIs update DenseSet size/memory accounting when detaching, and they require callers to free returned entries explicitly via DeleteEntry().

### DIFF
--- a/src/core/dense_set.cc
+++ b/src/core/dense_set.cc
@@ -688,7 +688,7 @@ auto DenseSet::Find2(const void* ptr, uint32_t bid, uint32_t cookie)
   return {0, nullptr, nullptr};
 }
 
-void DenseSet::Delete(DensePtr* prev, DensePtr* ptr) {
+void* DenseSet::Delete(DensePtr* prev, DensePtr* ptr, bool detach) {
   void* obj = nullptr;
 
   if (ptr->IsObject()) {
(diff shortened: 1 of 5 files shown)


src/core/string_map.h
```diff
@@ -109,16 +109,29 @@ class StringMap : public DenseSet {
...
 
-  // otherwise updates its value and returns false.
+  // Adds a new field or updates its value. Returns true if added, false if updated.
   bool AddOrUpdate(std::string_view field, std::string_view value, uint32_t ttl_sec = UINT32_MAX,
                    bool keepttl = false);
 
+  // Like AddOrUpdate but on update returns the previous sds entry (key+value blob)
+  // instead of deleting it. Caller must free the returned entry via DeleteEntry().
+  // Returns nullptr if a new field was added.
+  sds AddOrUpdateAndExtract(std::string_view field, std::string_view value,
```
dranikpg: Maybe AddOrExchange like std::exchange

src/core/string_map_test.cc
```diff
@@ -274,4 +274,102 @@ TEST_F(StringMapTest, ExpiryWithMaxAndKeepTTL) {
...
+
+  // Verify the extracted entry has the old value
+  sds val = StringMap::GetValue(prev);
+  EXPECT_EQ(string_view(val, sdslen(val)), "old_value");
+
+  // Verify map now has the new value
+  EXPECT_STREQ(sm_->Find("f1")->second, "new_value");
+  EXPECT_EQ(sm_->UpperBoundSize(), 1u);
+
+  StringMap::DeleteEntry(prev);
```
dranikpg, line 325: Do you need the returned value to be strictly sds*, a simple wrapper wouldn't hurt. Even stl has a `node_handle` with key(), mapped() and destructor

src/core/dense_set.cc
```diff
@@ -721,6 +721,38 @@ void DenseSet::Delete(DensePtr* prev, DensePtr* ptr) {
...
 }
 
+void* DenseSet::Detach(DensePtr* prev, DensePtr* ptr) {
+  void* obj = nullptr;
+
+  if (ptr->IsObject()) {
+    obj = ptr->Raw();
+    ptr->Reset();
+    if (prev) {
+      DCHECK(prev->IsLink());
```
dranikpg: Isn't it absolutely the same as Delete? Why can't it be just a flag to delete?

PR #6878, opened 2026-03-13
chore(facade): Simplify parsed command replies

dranikpg:
Asynchronous command either return an error or proceed to direct replies, the middle layer of ParsedCommand for capturing replies is not used. Synchronous commands use it only as a passthrough layer to call reply builder functions.

_Furthermore_, with coroutines it is very easy to accidentally call a reply function on the context instead of the reply builder (source of multiple bugs while developing them, confuses AI). The less reply functions it has left, the better.

Gradually remove this layer to keep only essential SendError calls

### DIFF
--- a/src/facade/parsed_command.h
+++ b/src/facade/parsed_command.h
@@ -124,7 +124,6 @@ class ParsedCommand : public cmn::BackedArguments {
     SendSimpleString("OK");
   }
 
-  // TODO: remove
   void SendLong(long val);
   template <typename F> void ReplyWith(F&& func) {
     assert(!is_deferred_reply_);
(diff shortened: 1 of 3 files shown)


dranikpg:
augment review

dranikpg:
augment review

src/facade/parsed_command.h
```diff
@@ -126,19 +126,10 @@ class ParsedCommand : public cmn::BackedArguments {
 
   // TODO: remove
```
romange: Remove todo

romange:
Like Resp_SendLong  direct wrapper

dranikpg:
> Lgtm, maybe you can still have direct wrappers to have a nicercode without static cast in many places?

I was thinking of `redis_rb()` or `rrb()` function on CommandContext to get the casted replier

romange:
I prefer `redis_rb` then

dranikpg:
> I prefer redis_rb then

But I assume not here? We have literally dozens of such casts, it can be done separately

Issue #6802, opened 2026-03-04
Support `vector_range` operator for search


abhijat:
The semantic routing library called redisvl allows a user to make queries against a db such that there are some pre-defined routes. A route is a group of objects (strings/references), it has a name and a threshold of distance, outside of the threshold a query will not match the route.

The semantic router performs a query over multiple routes, and one route, whose references are closest to the query, wins. Currently we support most of the infrastructure to support semantic routing out of the box, except for two critical components.

This ticket is for implementing the `vector_range` operator, which the library sends as a query to dragonfly, so that semantic routing can work. Please see https://github.com/dragonflydb/dataplane-private/issues/176#issuecomment-3988480532 for details of the query sent.

```
"FT.AGGREGATE" "dragonfly-test-router" "@vector:[VECTOR_RANGE $distance_threshold $vector]=>{$YIELD_DISTANCE_AS: vector_distance}" "SCORER" "TFIDF" "DIALECT" "2" "GROUPBY" "1" "@route_name" "REDUCE" "AVG" "1" "vector_distance" "AS" "distance" "SORTBY" "1" "@distance" "MAX" "1" "FILTER" "(@route_name == 'technology' && @distance < 0.8) || (@route_name == 'sports' && @distance < 0.8) || (@route_name == 'cooking' && @distance < 0.8)" "params" "4" "vector" "..." "distance_threshold" "0.8"
```

vyavdoshenko:
Implementation plan:

1. Parser (src/core/search/parser.y)
   - Add VECTOR_RANGE token
   - Add grammar rule: @field:[VECTOR_RANGE $radius $vec]=>{$YIELD_DISTANCE_AS: alias}

2. AST (src/core/search/ast_expr.h)
   - Add AstVectorRangeNode

3. FLAT index range search (src/core/search/search.cc)
   - Linear scan: return all docs with dist(vec, doc_vec) <= radius
   - Attach distance as a field named by YIELD_DISTANCE_AS alias

4. HNSW range search (src/core/search/hnsw_index.h/cc)
   - Add RangeQuery() method
   - Implement own solution like searchRangeBottomLayer_WithTimeout + processCandidate_RangeSearch from RedisAI/VectorSimilarity
   - Key idea: dynamic_range shrinks as closer candidates are found

5. Global HNSW (src/server/search/global_hnsw_index.h/cc)
   - Add RangeQuery counterpart alongside existing Knn()

[...]

PR #6880, opened 2026-03-13
feat(search): implement VECTOR_RANGE operator for FLAT index

vyavdoshenko:
Implements the `vector_range` operator for full-text search queries, enabling semantic routing support.

Adds support for the query syntax:
`@field:[VECTOR_RANGE $radius $vec]=>{$YIELD_DISTANCE_AS: alias}`

HNSW range search, global index wiring, and `FT.AGGREGATE` pipeline integration are planned separately (items 4–6 of #6802).

Implements (partial items 1-3) #6802

### DIFF
--- a/src/core/search/search.cc
+++ b/src/core/search/search.cc
@@ -353,7 +353,7 @@ struct BasicSearch {
   }
 
   void SearchVectorRangeFlat(FlatVectorIndex* vec_index, const AstVectorRangeNode& node) {
-    auto all_docs = vec_index->GetAllDocsWithNonNullValues();
+    const auto& all_docs = indices_->GetAllDocs();
     auto [dim, sim] = vec_index->Info();
     for (DocId doc : all_docs) {
       float dist = VectorDistance(node.vec.first.get(), vec_index->Get(doc), dim, sim);
@@ -450,6 +450,7 @@ struct BasicSearch {
     // used by knn
     DCHECK(top_level || holds_alternative<AstKnnNode>(node.Variant()) ||
            holds_alternative<AstGeoNode>(node.Variant()) ||
+           holds_alternative<AstVectorRangeNode>(node.Variant()) ||
            visit([](auto* set) { return is_sorted(set->begin(), set->end()); }, result.Borrowed()));
 
     if (profile_builder_)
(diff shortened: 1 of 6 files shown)


src/core/search/search.cc
```diff
@@ -710,4 +759,13 @@ void SearchAlgorithm::EnableProfiling() {
   profiling_enabled_ = true;
 }
 
+bool SearchAlgorithm::IsVectorRangeQuery() const {
```
romange: nit: you do not need it. you use it in a test but you can check if `GetVectorRangeNode()`  is not null

src/core/search/parser.y
```diff
@@ -124,6 +130,28 @@ opt_ef_runtime:
...
+      auto alias = std::move($12);
+      auto vec_result = BytesToFtVectorSafe($6);
+      if (!vec_result) {
+        auto empty_vec = std::make_unique<float[]>(0);
+        $$ = AstVectorRangeNode(std::move(field), radius,
+                                {std::move(empty_vec), size_t{0}}, std::move(alias));
+      } else {
+        $$ = AstVectorRangeNode(std::move(field), radius, std::move(*vec_result),
+                                std::move(alias));
+      }
```
dranikpg, line 147: I see its done above like this as well. The idea is that is causes an error by having the wrong dimension, righ?

src/core/search/search.cc
```diff
@@ -351,6 +352,54 @@ struct BasicSearch {
     knn_distances_.resize(prefix_size);
   }
 
+  void SearchVectorRangeFlat(FlatVectorIndex* vec_index, const AstVectorRangeNode& node) {
+    auto all_docs = vec_index->GetAllDocsWithNonNullValues();
```
dranikpg: 1. `GetAllDocsWithNonNullValues` copies all ids, you can use `GetAllDocs` which returns a refence. Why allocate millions of elements if just a few will match
2. I just looked and it looks like the notion of "null" is broken for the vector index

> // Check if the vector is not zero (all elements are 0)
> // TODO: Valid vector can contain 0s, we should use a better approach

Maybe we can fix it somehow right here? Maybe store one more float than the dimension to mark nullability?

src/core/search/parser.y
vyavdoshenko: Same pattern as KNN above. When BytesToFtVectorSafe fails (e.g. non-binary string passed as vector), we store a size-0 placeholder. During search, vec.second == 0 is checked before the dimension check and returns empty results silently - no error is set.

src/core/search/search.cc
vyavdoshenko: Fixed point 1. For point 2, a new issue has been created. I suggest fixing it in a separate PR.
https://github.com/dragonflydb/dragonfly/issues/6884

src/core/search/search.cc

PR #6883, opened 2026-03-13
chore(common): Some C++20 simplifications

dranikpg:
Careful introduction of some c++20 in common folder


dranikpg:
augment review

src/common/heap_size.h
```diff
@@ -27,48 +28,43 @@
...
+requires requires {
+  typename T::is_stackonly;
+}
 size_t HeapSize(const T& t) {
   return 0;
 }
 
-template <typename T> size_t HeapSize(absl::Span<T>) {
+// Overload for trivial types we don't have to account for
+template <typename T> size_t HeapSize(const T& t) {
```
kostasrim, line 60: requires should work here as well :)

src/common/heap_size.h
```diff
@@ -27,48 +28,43 @@
...
+  static_assert(std::is_trivial_v<T> || std::is_same_v<std::string_view, T>);
   return 0;
 }
```
kostasrim, line 63:
```suggestion
template <typename T> size_t HeapSize(const T& t) 
requires (std::is_trivial_v<T> || std::is_same_v<std::string_view, T>) {
return 0;
}
```

src/common/heap_size.h
```diff
@@ -100,46 +96,27 @@ template <typename T, size_t N> size_t HeapSize(const absl::InlinedVector<T, N>&
...
   size_t size = 0;
-
-  if constexpr (!heap_size_detail::StackOnlyType<typename Container::value_type>()) {
-    for (const auto& e : c) {
-      size += HeapSize(e);
-    }
-  }
-
+  for (const auto& e : c)
+    size += HeapSize(e);
```
kostasrim, line 117: std::ranges::fold but it's c++ 23 :(

src/facade/resp_srv_parser_test.cc
```diff
@@ -42,6 +42,12 @@ class RespSrvParserTest : public testing::Test {
  protected:
   RespSrvParser::Result Parse(std::string_view str);
 
+  auto Vec() {
+    vector<string_view> out;
```
kostasrim, line 46: You can do something like   

return {std::ranges::begin(args_.view()), std::ranges::end(args_.view())} 

and avoid the auto return type 😄 

there is also ranges::to but again c++23

kostasrim:
p.s. I hate string concatenation with `+` because the `operator is not symmetrical`. In other words for two strings `a` + `b` != `b` + `a`. This breaks generic code that rely on symmetrical mathematical operations. 

IGNORE this message, the context you used `+` is perfectly fine

PR #6886, opened 2026-03-14
feat(server): Introduce tagged chunk header and tests

abhijat:
Introduces a tagged chunk header with two 4 bytes fields, a tag and a payload size. The header can be serialized/deserialized to an 8 byte array.

This PR is a precursor to sending the chunk to replicas prefixed into data blobs

follow up PR for reference https://github.com/dragonflydb/dragonfly/pull/6890

### DIFF
--- a/src/core/cms.cc
+++ b/src/core/cms.cc
@@ -8,8 +8,7 @@
 
 #include <algorithm>
 #include <cmath>
-
-#include "base/logging.h"
+#include <limits>
 
 namespace dfly {
 namespace {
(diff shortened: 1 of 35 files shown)


src/server/tagged_chunk.cc
```diff
@@ -0,0 +1,43 @@
...
+#include "base/endian.h"
+#include "base/logging.h"
+
+namespace dfly {
+
+using nonstd::make_unexpected;
+
+std::array<char, 8> TaggedChunkHeader::Serialize() const {
+  std::array<char, 8> output;
+  base::LE::StoreT(static_cast<uint32_t>(tag), output.data());
```
romange, line 16: I am sorry I do not understand what's all this.

We already have RDB utilities to serialize integers. Is it an AI slope?

why do we need 3 files with unit tests for this?

romange:
This PR is not needed. You do not write unit tests to cover an arbitrary trivial lines. This is absurd.

src/server/tagged_chunk.cc
abhijat: There's no AI in this PR

PR #6889, opened 2026-03-14
chore: update pytests for more modern redis-py client

romange:
## Summary

Update pytest suite to be compatible with modern redis-py (v7+) client library.

## Changes

- **cluster_test.py**: Handle `MovedError` (now a distinct exception type in redis-py v7) alongside `ResponseError`; use `endswith()` for MOVED assertion strings since the prefix format changed; split `CLUSTER HELP` into two args for proper command routing
- **connection_test.py**: Disable redis-py v7 automatic retries (`Retry(NoBackoff(), 0)`) in `test_tls_client_kill_preemption` so connection kills are observed immediately; fix race condition by polling for client disconnect before cancelling the seed task
- **replication_test.py**: Disable auto-retries in `test_replicaof_reject_on_load` so `BusyLoadingError` (which now inherits from `ConnectionError`) is raised instead of silently retried; add `socket_timeout` in `test_take_over_read_commands` and stop querying master after it goes down
- **search_test.py**: Handle module rename `indexDefinition` → `index_definition` across redis-py versions via try/except import
- **server_family_test.py**: Disable auto-retries in `test_client_kill`; use specific exception types (`ResponseError`, `ConnectionError`) instead of bare `Exception`
- **acl_family_test.py**: Accept `NoPermissionError` alongside `ConnectionError` for revoked pub/sub
- **conftest.py**: Suppress noisy "Unclosed ClusterNode" asyncio warnings from redis-py topology refreshes

### DIFF
--- a/tests/dragonfly/replication_test.py
+++ b/tests/dragonfly/replication_test.py
@@ -1363,7 +1363,7 @@ async def test_take_over_read_commands(df_factory, master_threads, replica_threa
     replica = df_factory.create(proactor_threads=replica_threads)
     df_factory.start_all([master, replica])
 
-    c_master = master.client()
+    c_master = master.client(socket_timeout=1, socket_connect_timeout=1)
     await c_master.execute_command("SET foo bar")
 
     c_replica = replica.client()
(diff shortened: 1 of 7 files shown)


tests/dragonfly/replication_test.py
```diff
@@ -1372,15 +1372,17 @@ async def test_take_over_read_commands(df_factory, master_threads, replica_threa
...
-                assert res == "OK"
-            except:
-                pass
+            if master_alive:
+                try:
+                    res = await c_master.execute_command("GET foo")
+                    assert res == "bar"
+                    res = await c_master.execute_command("CONFIG SET aclfile myfile")
+                    assert res == "OK"
+                except:
```
kostasrim, line 1384: can't we `break` out of the for-loop ?

PR #6890, opened 2026-03-14
[WIP] feat(server): Receiver logic for tagged chunks

abhijat:
Child PR of https://github.com/dragonflydb/dragonfly/pull/6886

* Introduces new version 7 for use as gate - the new version is not `CURRENT_VER` - that will only be done when the sender code changes. The receiver checks the version, if it is 7, tags are expected.
* The sender code is not changed yet, so tags are not sent within this PR.
* Adds new io source which can parse and strip off headers and tag
* The parsed tag is discarded for now. After future work in https://github.com/dragonflydb/dragonfly/blob/main/docs/shard-serialization.md the parsed tag can be used for demultiplexing.
* The rdb loader will optionally receive the prefix source wrapped in this new source, if the version check passes. The loader still reads the same blob as before, it is not concerned with header which has been stripped out before reaching the loader.

### DIFF
--- a/src/server/version.h
+++ b/src/server/version.h
@@ -42,6 +42,8 @@ enum class DflyVersion {
   // - hnsw-index-metadata AUX field
   VER6,
 
+  // Sends tagged chunks to replicas to differentiate between journal and baseline data.
+  VER7,
   // Always points to the latest version
   CURRENT_VER = VER6,
 };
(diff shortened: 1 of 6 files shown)


src/server/version.h
```diff
@@ -42,6 +42,8 @@ enum class DflyVersion {
   // - hnsw-index-metadata AUX field
   VER6,
 
+  // Sends tagged chunks to replicas to differentiate between journal and baseline data.
+  VER7,
   // Always points to the latest version
   CURRENT_VER = VER6,
```
abhijat, line 48: does not yet point to 7, so that master does not send v7 without tags.

PR #6892, opened 2026-03-14
fix(search): Fix null vector detection in FlatVectorIndex

### LINKED ISSUE #6884: Fix null vector detection in FlatVectorIndex
Problem:

`FlatVectorIndex::GetAllDocsWithNonNullValues` currently detects "null" (unset) vectors by checking if all float elements are zero:

```cpp
// TODO: Valid vector can contain 0s, we should use a better approach
for (size_t i = 0; i < dim_; ++i) {
  if (vec[i] != 0.0f) { is_zero_vector = false; break; }
}
```

This is incorrect - a valid vector [0.0, 0.0, ..., 0.0] is a legitimate embedding that would be silently skipped in @field:* queries and KNN searches. Also, Remove() is currently a no-op, so deleted documents are never actually removed from the index.

Solution:
Store dim + 1 floats per document in entries_, where the extra float at index dim acts as a null marker:

1.0 = valid vector (set on AddVector)
0.0 = null/removed (default after resize, set on Remove)
Changes needed (all in indices.cc):

FlatVectorIndex() - reserve capacity * (dim + 1)
AddVector - use stride dim + 1, set marker to 1.0
Remove - set marker to 0.0 (currently no-op)
GetAllDocsWithNonNullValues - check marker instead of all-zeros heuristic

Notes:
No API changes — Get(doc) still returns const float* to the vector data
No persistence impact — index is rebuilt on restart
Need to verify existing tests for @field:* queries still pass (some may rely on current behavior)

vyavdoshenko:
Fix two bugs in `FlatVectorIndex`:

1. Zero vector falsely treated as null - `GetAllDocsWithNonNullValues` skipped documents whose vector was `[0.0, 0.0, ..., 0.0]`, even though it's a valid embedding.
2. `Remove()` was a no-op - removed documents continued to appear in `@field:*` queries.

Root cause: null detection relied on an all-zeros heuristic with no way to distinguish "no vector set" from a valid zero vector.

Fix: each document now occupies `dim + 1` floats in `entries_`. The extra float acts as a presence marker (`1.0` = present, `0.0` = absent). `Remove()` clears the marker; `GetAllDocsWithNonNullValues` checks it instead of scanning all floats.

No API changes — `Get(doc)` still returns `const float*` to the vector data. Index is rebuilt on restart, so no persistence impact.

Fixes #6884

### DIFF
--- a/src/core/search/indices.cc
+++ b/src/core/search/indices.cc
@@ -523,59 +523,53 @@ bool BaseVectorIndex::Add(DocId id, const DocumentAccessor& doc, std::string_vie
   return true;
 }
 
+// Each document occupies (dim_ + 1) floats in entries_: dim_ floats for the vector data,
+// followed by one float as a presence marker (1.0 = present, 0.0 = absent/removed).
+// This avoids the previous heuristic of treating all-zero vectors as null.
+static constexpr float kPresent = 1.0f;
+static constexpr float kAbsent = 0.0f;
+
 FlatVectorIndex::FlatVectorIndex(const SchemaField::VectorParams& params,
                                  PMR_NS::memory_resource* mr)
     : BaseVectorIndex{params.dim, params.sim}, entries_{mr} {
   DCHECK(!params.use_hnsw);
-  entries_.reserve(params.capacity * params.dim);
+  entries_.reserve(params.capacity * (params.dim + 1));
 }
 
 void FlatVectorIndex::AddVector(DocId id, const void* vector) {
-  DCHECK_LE(id * dim_, entries_.size());
-  if (id * dim_ == entries_.size())
-    entries_.resize((id + 1) * dim_);
+  const size_t stride = dim_ + 1;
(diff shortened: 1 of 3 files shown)


src/core/search/indices.cc
```diff
@@ -523,59 +523,50 @@ bool BaseVectorIndex::Add(DocId id, const DocumentAccessor& doc, std::string_vie
...
 }
 
 void FlatVectorIndex::Remove(DocId id, const DocumentAccessor& doc, string_view field) {
-  // noop
+  const size_t stride = dim_ + 1;
+  if (id * stride + dim_ < entries_.size())
+    entries_[id * stride + dim_] = kAbsent;
 }
 
 const float* FlatVectorIndex::Get(DocId doc) const {
```
dranikpg, line 557: This needs to be updated too. Maybe it should return nullptr? Vector search and your last PR about vector range should know now how to skip them

vyavdoshenko: fixed

PR #6896, opened 2026-03-15
chore: add cms family

kostasrim:
Implements all CMS family functions in dragonfly.

* add cms family
* add unit tests
* small refactor cms.{h/cc}

This PR does not include rdb load/save and journal (write commands are not journaled atm) which will be added in the last PR (also cms.merge needs a journal rewrite)

Split from #6882
PR 2/3

### DIFF
--- a/src/server/cms_family.h
+++ b/src/server/cms_family.h
@@ -1,0 +1,13 @@
+// Copyright 2026, DragonflyDB authors.  All rights reserved.
+// See LICENSE for licensing terms.
+//
+
+#pragma once
+
+namespace dfly {
+
+class CommandRegistry;
+
+void RegisterCmsFamily(CommandRegistry*);
+
+}  // namespace dfly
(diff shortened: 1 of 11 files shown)


kostasrim:
fuzzer is expected to fail because of missing serialization. Leave it as if for now (failing), I have the other PR ready with rdb logic

src/server/cms_family.h
```diff
@@ -0,0 +1,13 @@
...
+
+#pragma once
+
+namespace dfly {
+
+class CommandRegistry;
+
+void RegisterCmsFamily(CommandRegistry*);
+
+}  // namespace dfly
```
dranikpg, line 13: Plase use command_families.h, I moved all those functions there to reduce file count

kostasrim: sure, I will sent a third PR soon and do the refactor there. Will ask for your review

PR #6897, opened 2026-03-15
chore: CMS serialization

kostasrim:
Allow serialization/deserialization of CMS types and add a unit test

Split from https://github.com/dragonflydb/dragonfly/pull/6882
PR 3/3

### DIFF
--- a/src/core/cms.cc
+++ b/src/core/cms.cc
@@ -119,10 +119,9 @@ void CMS::Reset() {
   count_ = 0;
 }
 
-void CMS::Load(int64_t total_incr_count, const int64_t* data, size_t size) {
-  DCHECK_EQ(size, NumCounters());
+void CMS::Load(int64_t total_incr_count, const int64_t* data) {
   count_ = total_incr_count;
-  std::copy_n(data, size, counters_);
+  std::copy_n(data, NumCounters(), counters_);
 }
 
 }  // namespace dfly
--- a/src/core/cms.h
+++ b/src/core/cms.h
@@ -52,7 +52,7 @@ class CMS {
   void Reset();
 
   // Load serialized counter state. data must have exactly NumCounters() elements.
-  void Load(int64_t total_incr_count, const int64_t* data, size_t size);
+  void Load(int64_t total_incr_count, const int64_t* data);
 
   // Accessors for CMS properties
   uint32_t width() const {
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -294,7 +294,8 @@ void RdbLoaderBase::OpaqueObjLoader::operator()(const RdbSBF& src) {
 
 void RdbLoaderBase::OpaqueObjLoader::operator()(const RdbCMS& src) {
   CMS* cms = CompactObj::AllocateMR<CMS>(src.width, src.depth, CompactObj::memory_resource());
-  cms->Load(src.count, src.counters.data(), src.counters.size());
+  DCHECK_EQ(src.counters.size(), cms->NumCounters());
+  cms->Load(src.total_incr_count, src.counters.data());
   pv_->SetCMS(cms);
 }
 
@@ -1833,16 +1834,11 @@ auto RdbLoaderBase::ReadSBF2() -> io::Result<OpaqueObj> {
 io::Result<RdbLoaderBase::OpaqueObj> RdbLoaderBase::ReadCMS() {
   RdbCMS res;
 
-  uint64_t width, depth, count, num_counters;
-  SET_OR_UNEXPECT(LoadLen(nullptr), width);
-  SET_OR_UNEXPECT(LoadLen(nullptr), depth);
-  SET_OR_UNEXPECT(LoadLen(nullptr), count);
-  SET_OR_UNEXPECT(LoadLen(nullptr), num_counters);
+  SET_OR_UNEXPECT(LoadLen(nullptr), res.width);
+  SET_OR_UNEXPECT(LoadLen(nullptr), res.depth);
+  SET_OR_UNEXPECT(LoadLen(nullptr), res.total_incr_count);
 
-  res.width = width;
-  res.depth = depth;
-  res.count = count;
-
+  const size_t num_counters = res.width * res.depth;
   res.counters.resize(num_counters);
   for (size_t i = 0; i < num_counters; ++i) {
     uint64_t raw;
--- a/src/server/rdb_save.cc
+++ b/src/server/rdb_save.cc
@@ -677,8 +677,6 @@ std::error_code RdbSerializer::SaveCMSObject(const PrimeValue& pv) {
   size_t num_counters = cms->NumCounters();
   const int64_t* data = cms->Data();
 
-  RETURN_ON_ERR(SaveLen(num_counters));
-
   // Serialize counters as little-endian 64-bit values
   std::vector<uint64_t> buf(num_counters);
   for (size_t i = 0; i < num_counters; ++i) {
--- a/src/server/cms_family.cc
+++ b/src/server/cms_family.cc
@@ -463,7 +463,7 @@ void CmdMerge(CmdArgList args, CommandContext* cmd_cntx) {
 
       for (const auto& cms_data : result.value()) {
         CMS temp_cms(cms_data.width, cms_data.depth, CompactObj::memory_resource());
-        temp_cms.Load(cms_data.count, cms_data.counters.data(), cms_data.counters.size());
+        temp_cms.Load(cms_data.count, cms_data.counters.data());
 
         if (!dest_cms->MergeFrom(temp_cms, merge_args.weights[cms_data.src_index])) {
           write_result = OpStatus::INVALID_VALUE;
--- a/src/server/rdb_load.h
+++ b/src/server/rdb_load.h
@@ -63,7 +63,7 @@ class RdbLoaderBase {
 
   struct RdbCMS {
     uint32_t width, depth;
-    int64_t count;
+    int64_t total_incr_count;
     std::vector<int64_t> counters;
   };
 


src/server/cms_family.h
```diff
@@ -1,13 +0,0 @@
-// Copyright 2026, DragonflyDB authors.  All rights reserved.
```
kostasrim, line 1: @dranikpg done

src/server/cms_family_test.cc
```diff
@@ -201,4 +201,21 @@ TEST_F(CmsFamilyTest, InfoAfterMerges) {
   EXPECT_THAT(resp, ErrArg("CMS: key does not exist"));
 }
 
+TEST_F(CmsFamilyTest, Serialization) {
```
romange: we usually add it into rdb_test.cc.

src/server/rdb_extensions.h
```diff
@@ -8,17 +8,18 @@ extern "C" {
 #include "redis/rdb.h"
 }
 
-//  Custom types: Range 30-35 is used by DF RDB types.
+//  Custom types: Range 30-39 is used by DF RDB types.
```
romange: why 39?

src/server/rdb_save.cc
```diff
@@ -660,6 +667,20 @@ std::error_code RdbSerializer::SaveSBFObject(const PrimeValue& pv) {
...
 
+std::error_code RdbSerializer::SaveCMSObject(const PrimeValue& pv) {
+  CMS* cms = pv.GetCMS();
+
+  RETURN_ON_ERR(SaveLen(cms->width()));
+  RETURN_ON_ERR(SaveLen(cms->depth()));
+  RETURN_ON_ERR(SaveLen(cms->total_count()));
+
+  size_t counter_bytes = cms->CounterBytes();
+  string_view counter_data(reinterpret_cast<const char*>(cms->Data()), counter_bytes);
```
romange: add a comment here that the code is not portable on big-endian systems.

src/server/rdb_load.cc
```diff
@@ -290,6 +292,14 @@ void RdbLoaderBase::OpaqueObjLoader::operator()(const RdbSBF& src) {
   pv_->SetSBF(sbf);
 }
 
+void RdbLoaderBase::OpaqueObjLoader::operator()(const RdbCMS& src) {
+  const auto* counters = reinterpret_cast<const int64_t*>(src.counter_data.data());
+  size_t counter_count = src.counter_data.size() / sizeof(int64_t);
+  CMS* cms = CompactObj::AllocateMR<CMS>(src.width, src.depth, src.count, counters, counter_count,
```
romange: I would prefer we do not have a c'tor that accepts every field and exposes the internal state and have a bit more elegant API for CMS object where we create it with:
`uint32_t width, uint32_t depth, PMR_NS::memory_resource* mr` c'tor and then use `Load()` method.

The comment above is correct: `const auto* counters = reinterpret_cast<const int64_t*>(src.counter_data.data());` is UB.
you should use `absl::little_endian::Load64` from the binary data and set counters in a loop.

why did you change `counters_`  to vector? It unnecessary increases the size of CMS and we do not use vector functionality.

kostasrim: @romange you had a `pointer + size` now you have a vector which is a `pointer + size`. Is there really a difference than one actual being a proper class ?

romange: I had `size()` that derived from the existing data members.

vector is 24 bytes. unique_ptr is 8 bytes. You replaced something that worked well with a subjectively "better" version, did not gain any additional capabilities and wasted 16 bytes.

romange: a pointer, not unique_ptr. actually a pmr vector is at least 32 bytes. so 24 bytes wasted

romange: Also, I wrote this code 2 days ago, so on a personal level you could just asked me instead of just rewriting it.

kostasrim: A few things:

> a pointer, not unique_ptr. actually a pmr vector is at least 32 bytes. so 24 bytes wasted

vector<T> = 32 bytes (8ptr + 8capacity + 8size + 8pmr ptr). 
I replaced a `unique_ptr` + `mr` = 16 bytes with 32bytes so the `actual overhead of what I did is 16 bytes`.

My mistake was that I did not notice the size can't be removed (width+height or count) and we ended up with 16 more bytes on the type. But IMO is it really bad/ a regression?

To waste 10mb of those 16 bytes chunks we need `700k+` CMS objects. Is this even a low hanging fruit ?

> Also, I wrote this code 2 days ago, so on a personal level you could just asked me instead of just rewriting it.
Had I suspected I would introduce a regression I would have pinged


I can revert back to RAII (although will check if I can get rid of widhg/height and bring back the type to the same size)

src/server/rdb_save.cc
dranikpg: I assume iterating and to calling absl::Store...() wouldn't be slower of more difficult, but portable

src/server/rdb_load.cc
```diff
@@ -290,6 +292,14 @@ void RdbLoaderBase::OpaqueObjLoader::operator()(const RdbSBF& src) {
   pv_->SetSBF(sbf);
 }
 
+void RdbLoaderBase::OpaqueObjLoader::operator()(const RdbCMS& src) {
+  const auto* counters = reinterpret_cast<const int64_t*>(src.counter_data.data());
```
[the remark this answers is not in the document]
dranikpg: That's formally also true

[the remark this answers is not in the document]
romange: not just formally. I actually had a real bug with that.  
https://www.romange.com/2017/09/26/how-to-serialize-integers-into-memory/

src/server/rdb_load.cc
romange: it's not a low hanging fruit if the situation is reversed. But  I do not understand this logic why you need to spend energy and time to change code from A to B, when `B` is "just a bit worse".

kostasrim: The rationale was that I don't have to argue about RAII. For a vector it's encapsulated into the type, for a raw pointer it is not.

"just a bit worse" is true when it comes to the added memory -- for the actual implementation simplicity, I would argue otherwise

Either way, I agree it was an overkill and I will revert

romange: you can use unique_ptr with custom deleter I guess.

src/core/cms.h
```diff
@@ -56,6 +51,9 @@ class CMS {
   // Reset all counters and total count to zero.
   void Reset();
 
+  // Load serialized counter state. data must have exactly NumCounters() elements.
+  void Load(int64_t total_count, const int64_t* data, size_t count);
```
romange: 1. It's not clear what total_count is.
2. if data must have exactly `NumCounters()` elements then why do you need "count"?

kostasrim: it's the counter related to incr operators. missleading name, will change it

src/server/rdb_load.cc
```diff
@@ -1819,6 +1830,31 @@ auto RdbLoaderBase::ReadSBF2() -> io::Result<OpaqueObj> {
...
+  res.depth = depth;
+  res.count = count;
+
+  res.counters.resize(num_counters);
+  for (size_t i = 0; i < num_counters; ++i) {
+    uint64_t raw;
+    auto ec = FetchBuf(sizeof(raw), &raw);
+    if (ec)
+      return make_unexpected(ec);
+    res.counters[i] = static_cast<int64_t>(base::LE::LoadT<uint64_t>(&raw));
```
kostasrim, line 1848: reading int by int is not great but not a deal breaker as is. it's alway possible to optimize if needed,

src/core/cms.cc
```diff
@@ -107,12 +115,14 @@ bool CMS::MergeFrom(const CMS& other, int64_t weight) {
...
 
 void CMS::Reset() {
-  std::fill(counters_.begin(), counters_.end(), 0);
+  std::fill_n(counters_, NumCounters(), 0);
   count_ = 0;
 }
 
-size_t CMS::MallocUsed() const {
-  return counters_.size() * sizeof(int64_t);
+void CMS::Load(int64_t total_incr_count, const int64_t* data, size_t size) {
```
romange: size is redundant here, as it should be NumCounters()

src/server/rdb_load.cc
```diff
@@ -290,6 +292,12 @@ void RdbLoaderBase::OpaqueObjLoader::operator()(const RdbSBF& src) {
   pv_->SetSBF(sbf);
 }
 
+void RdbLoaderBase::OpaqueObjLoader::operator()(const RdbCMS& src) {
+  CMS* cms = CompactObj::AllocateMR<CMS>(src.width, src.depth, CompactObj::memory_resource());
+  cms->Load(src.count, src.counters.data(), src.counters.size());
```
romange: pls DCHECK_EQ  `src.counters.size()` and `cms->NumCounters`

src/server/rdb_load.cc
```diff
@@ -1819,6 +1830,31 @@ auto RdbLoaderBase::ReadSBF2() -> io::Result<OpaqueObj> {
...
 }
 
+io::Result<RdbLoaderBase::OpaqueObj> RdbLoaderBase::ReadCMS() {
+  RdbCMS res;
+
+  uint64_t width, depth, count, num_counters;
+  SET_OR_UNEXPECT(LoadLen(nullptr), width);
+  SET_OR_UNEXPECT(LoadLen(nullptr), depth);
+  SET_OR_UNEXPECT(LoadLen(nullptr), count);
+  SET_OR_UNEXPECT(LoadLen(nullptr), num_counters);
```
romange: `num_counters` is redundant as you can derive it from width and depth.

src/server/rdb_load.cc
```diff
@@ -1819,6 +1830,31 @@ auto RdbLoaderBase::ReadSBF2() -> io::Result<OpaqueObj> {
...
+
+  res.width = width;
+  res.depth = depth;
+  res.count = count;
```
romange: rename count to total_count

src/server/rdb_load.cc
dranikpg: Yes, given that you copy when you serialize, you could do the same when loading - read the whole buffer at once. But afair FetchBuf goes through an internal buffer as well, so it shouldn't be a big difference

dranikpg: Also didn't know we have this LoadT thing, only knew about absl

kostasrim: > Also didn't know we have this LoadT thing, only knew about absl

yeah but IMO it's <T> screams meant to be used in template context where var<T> depends on a template parameter (yeah we are no in this context 😄 )

PR #6901, opened 2026-03-16
chore: add BullMQ/sidekiq tests, fix celery worker teardown timeout

romange:
- Fix celery_worker fixture teardown: add timeout to t.join() to prevent
  hanging when worker thread is blocked on socket.recv()
- Log queue memory usage in test_celery_push_jobs
- Change instance.py startup logging from debug to info

- Add test_sidekiq_push_jobs: pushes 2000 jobs in Sidekiq wire format
  (verified against sidekiq/lib/sidekiq/client.rb)
- Sidekiq uses plain LPUSH to queue:default, no Lua scripts
- Added commented out bullmq test - unfortunately it does not work with python3.8 that we currently run in CI.

### DIFF
--- a/tests/dragonfly/bull_sidekiq_test.py
+++ b/tests/dragonfly/bull_sidekiq_test.py
@@ -1,0 +1,89 @@
+import json
+import logging
+import time
+import uuid
+
+from redis import asyncio as aioredis
+
+# from bullmq import Queue
+# from . import dfly_args
+
+
+# BULLMQ_QUEUE_NAME = "{test_queue}"
+
+# @pytest.fixture
+# async def bullmq_queue(df_server):
+#     queue = Queue(BULLMQ_QUEUE_NAME, {"connection": {"host": "localhost", "port": df_server.port}})
+#     yield queue
+#     await queue.close()
+
+
+# @dfly_args({"lock_on_hashtags": True})
+# async def test_bullmq_push_jobs(async_client: aioredis.Redis, bullmq_queue: Queue):
+#     """Push 200 jobs and verify they are stored in Dragonfly."""
+#     for i in range(200):
+#         await bullmq_queue.add(
+#             "process_job",
+#             {"job_id": f"job{i}", "payload": f"data for job {i}"},
+#         )
+
+#     # BullMQ stores waiting jobs in a list key: bull:<queue_name>:wait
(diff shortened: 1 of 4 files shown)


tests/dragonfly/bull_sidekiq_test.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def _make_sidekiq_job(i: int) -> str:
+    """Generate a job payload matching the Sidekiq wire format.
+
+    Verified against sidekiq/lib/sidekiq/client.rb (atomic_push) and
+    sidekiq/lib/sidekiq/job_util.rb (normalize_item).
+    """
+    jid = uuid.uuid4().hex[:24]  # SecureRandom.hex(12)
+    now_ms = int(time.time() * 1000)  # Process.clock_gettime(CLOCK_REALTIME, :millisecond)
```
vyavdoshenko: Incorrect timestamp format in Sidekiq wire payload

sidekiq/lib/sidekiq/job_util.rb uses: float_second (Unix float), not :millisecond:


item["created_at"] ||= Process.clock_gettime(CLOCK_REALTIME, :float_second)
So created_at/enqueued_at should be time.time() (1.7e9), not int(time.time() * 1000) (1.7e12).
The variable name now_ms and the inline comment both reinforce the wrong assumption.

Suggested fix:

```
now = time.time()
```
...
```
"created_at": now,
"enqueued_at": now,
```

tests/dragonfly/requirements.txt
```diff
@@ -30,3 +30,4 @@ hiredis==2.4.0
 PyYAML>=6.0
 valkey>=6.0.2
 celery>=5.3.0
+# bullmq>=2.0.0
```
vyavdoshenko, line 33: nit: remove?

tests/dragonfly/bull_sidekiq_test.py
```diff
@@ -0,0 +1,89 @@
+import json
+import logging
+import time
+import uuid
+
+from redis import asyncio as aioredis
+
+# from bullmq import Queue
```
vyavdoshenko, line 8: Do you need this commented code below?

romange: I want to enable it once we fully switch to ubuntu:24. Will follow up once we stabilize our tests

PR #6902, opened 2026-03-16
feat: add distroless Docker image with built-in healthcheck

vyavdoshenko:
Adds `Dockerfile.distroless-dev` and `Dockerfile.distroless-prod` for a minimal shell-less runtime image based on `gcr.io/distroless/cc-debian12`, and a `/healthz` HTTP(s) endpoint to Dragonfly for container health checks.

Health checking uses [microcheck](https://github.com/tarampampam/microcheck) - two tiny static binaries (~75 KB each) copied into the image:
- `httpcheck` — default HEALTHCHECK for plain HTTP deployments
- `httpscheck` — available for TLS deployments (`https://localhost:6379/healthz`)

The prod image (`Dockerfile.distroless-prod`) is ready but requires a release that
includes the `/healthz` endpoint. Local strip experiment: **~70 MB** release image.

### DIFF
--- a/src/server/dfly_main.cc
+++ b/src/server/dfly_main.cc
@@ -21,10 +21,11 @@
 #include <mimalloc-new-delete.h>
 #endif
 
+#include <unistd.h>
+
 #ifdef __linux__
 #include <grp.h>
 #include <pwd.h>
-#include <unistd.h>
 
 #include <cerrno>
 #include <cstring>
@@ -225,19 +226,29 @@ template <typename... Args> unique_ptr<Listener> MakeListener(Args&&... args) {
 // Returns false on failure.
 bool DropPrivilegesToUser(const string& user_spec, const string& data_dir) {
   // Resolve user — accept username or numeric uid.
+  // For numeric UIDs not in /etc/passwd (common in distroless containers),
+  // fall back to using the raw UID with gid=uid and no supplementary groups.
   uid_t uid;
+  gid_t gid;
   struct passwd* pw = nullptr;
   if (absl::SimpleAtoi(user_spec, &uid)) {
-    pw = getpwuid(uid);
+    pw = getpwuid(uid);  // may be NULL for UIDs not in /etc/passwd
   } else {
     pw = getpwnam(user_spec.c_str());
+    if (!pw) {
+      LOG(ERROR) << "--user: unknown user '" << user_spec << "'";
+      return false;
+    }
   }
-  if (!pw) {
-    LOG(ERROR) << "--user: unknown user '" << user_spec << "'";
-    return false;
+
+  if (pw) {
+    uid = pw->pw_uid;
+    gid = pw->pw_gid;
+  } else {
+    // Numeric UID without passwd entry — use uid as gid.
+    gid = uid;
+    LOG(INFO) << "--user: uid " << uid << " not in /etc/passwd, using gid=" << gid;
   }
-  uid = pw->pw_uid;
-  gid_t gid = pw->pw_gid;
 
   // Recursively chown the data directory so the new user can access existing snapshots.
   // Only when --dir is explicitly set. Uses lchown to avoid following symlinks.
@@ -251,17 +262,26 @@ bool DropPrivilegesToUser(const string& user_spec, const string& data_dir) {
     do_chown(data_dir);
     std::error_code ec;
     fs::recursive_directory_iterator it(data_dir, ec);
-    if (!ec) {
-      for (const auto& entry : it)
-        do_chown(entry.path());
+    for (; !ec && it != fs::recursive_directory_iterator(); it.increment(ec)) {
+      do_chown(it->path());
+    }
+    if (ec) {
+      LOG(WARNING) << "Error iterating " << data_dir << ": " << ec.message();
     }
   }
 
   // Drop supplementary groups first, then gid, then uid.
   // Order is critical: after setuid we no longer have permission to call setgid.
-  if (initgroups(pw->pw_name, gid) != 0) {
-    LOG(ERROR) << "initgroups failed: " << strerror(errno);
-    return false;
+  if (pw) {
+    if (initgroups(pw->pw_name, gid) != 0) {
+      LOG(ERROR) << "initgroups failed: " << strerror(errno);
+      return false;
+    }
+  } else {
+    if (setgroups(0, nullptr) != 0) {
+      LOG(ERROR) << "setgroups failed: " << strerror(errno);
+      return false;
+    }
   }
 
   if (setgid(gid) != 0) {
--- a/tools/packaging/Dockerfile.distroless-prod
+++ b/tools/packaging/Dockerfile.distroless-prod
@@ -34,9 +34,12 @@ VOLUME /data
 WORKDIR /data
 
 # Health check against the /healthz HTTP endpoint (returns 200 when server is ACTIVE).
+# Requires primary_port_http_enabled=true (the default).
 # Default assumes plain HTTP on port 6379. Override at runtime if needed:
 #   TLS:          docker run --health-cmd '/usr/local/bin/httpscheck https://localhost:6379/healthz' ...
 #   Custom port:  docker run --health-cmd '/usr/local/bin/httpcheck http://localhost:6380/healthz' ...
+# Note: during snapshot loading /healthz returns 503. For large snapshots increase --start-period
+# (e.g. --health-start-period=5m) to avoid premature "unhealthy" status.
 HEALTHCHECK --start-period=30s --interval=15s --timeout=5s --retries=3 \
   CMD ["/usr/local/bin/httpcheck", "http://localhost:6379/healthz"]
 
--- a/tools/packaging/Dockerfile.distroless-dev
+++ b/tools/packaging/Dockerfile.distroless-dev
@@ -41,9 +41,12 @@ VOLUME /data
 WORKDIR /data
 
 # Health check against the /healthz HTTP endpoint (returns 200 when server is ACTIVE).
+# Requires primary_port_http_enabled=true (the default).
 # Default assumes plain HTTP on port 6379. Override at runtime if needed:
 #   TLS:          docker run --health-cmd '/usr/local/bin/httpscheck https://localhost:6379/healthz' ...
 #   Custom port:  docker run --health-cmd '/usr/local/bin/httpcheck http://localhost:6380/healthz' ...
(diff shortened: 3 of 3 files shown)


src/server/CMakeLists.txt
```diff
@@ -14,6 +14,10 @@ add_executable(dragonfly dfly_main.cc version_monitor.cc)
 add_custom_target(check_dfly WORKING_DIRECTORY .. COMMAND ctest -L DFLY)
 cxx_link(dragonfly base dragonfly_lib)
 
+# Minimal health-check helper for distroless containers (no dragonfly_lib dependency).
+add_executable(dragonfly-healthcheck dfly_healthcheck.cc)
```
romange: why do we need `/healthz` via HTTP if we have this?

romange:
please note I do not want to switch to distroless right away - but we can replace internal tools in our current image.

src/server/dfly_healthcheck.cc
```diff
@@ -0,0 +1,126 @@
+// Copyright 2024, DragonflyDB authors.  All rights reserved.
```
romange:
```
Commonly Used Minimal Health Checkers:
microcheck: A suite of tiny (~75 KB) C-compiled binaries including httpcheck, httpscheck, and portcheck. They offer configurable settings, such as response codes and timeout, and provide standard exit codes (0 for healthy, 1 for unhealthy).
dmikusa/tiny-health-checker: A very small binary written in Go designed to only make HTTP requests to localhost. It is specifically designed to be safe and difficult to misuse if a container is compromised.
cloudlena/healthcheck: An extension of the scratch Docker image that adds a built-in HTTP health check tool, adding only about 2.8MB to the image size.
```

src/server/CMakeLists.txt
vyavdoshenko: removed

src/server/dfly_healthcheck.cc
vyavdoshenko: We have to add an additional binary regardless of our new binary or a tiny external check.
All these external checks can check the HTTP request or just open a TCP port. In this case, our binary is preferable because it can check the health using the PING command. /healthz handler has already been removed.

romange: what about tls configured datastore?

romange: in general I do have preference to package an external binary if it does the same as we do not want to reinvent the wheel and their static tool chain is probably better than of C++.

vyavdoshenko: Fixed
I added httpcheck and httpscheck in the image, but httpcheck is used by default.

src/server/dfly_main.cc
```diff
@@ -203,6 +220,117 @@ template <typename... Args> unique_ptr<Listener> MakeListener(Args&&... args) {
   return res;
 }
 
+#ifdef __linux__
+// Parses --user flag value and drops privileges to the specified user/group.
+// Must be called after ports are bound (so we can bind privileged ports first)
+// and BEFORE service.Init() to avoid racing with background snapshot loading.
+// Returns false on failure.
+bool DropPrivilegesToUser(const string& user_spec, const string& data_dir) {
```
romange, line 227: this function is unnecessary complicated and long.
1.  Drop the user:group syntax entirely. The primary use case is distroless containers where you pass a username/uid. The group is virtually always the user's primary group. Supporting user:group adds ~20 lines of branching for an edge case nobody needs. 
2. Require pw to succeed — don't support numeric-uid-without-passwd-entry                                                                                                                                               
  The pw == nullptr after getpwuid() path (lines 264-271) handles a case where someone passes a numeric uid that doesn't exist in /etc/passwd. In a container, this is a misconfiguration, not something to paper over. Just fail early.
3. Drop the CWD fallback for chown. The CWD fallback (lines 276-282) is surprising behavior. If --dir isn't set, don't chown anything — the Dockerfile already sets WORKDIR /data and the volume is mounted correctly. This will remove chown root guards as well.

tools/packaging/Dockerfile.distroless-prod
```diff
@@ -0,0 +1,45 @@
...
+
+COPY --from=builder /build/dragonfly /usr/local/bin/dragonfly
+COPY --from=healthcheck-dl /usr/local/bin/httpcheck /usr/local/bin/httpcheck
+COPY --from=healthcheck-dl /usr/local/bin/httpscheck /usr/local/bin/httpscheck
+COPY --from=builder /build/syslibs/ /
+
+VOLUME /data
+WORKDIR /data
+
+# Use httpcheck for plain HTTP (default). For TLS deployments, override with:
```
romange: how tls deployments should override? 
does it break the existing behavior?

src/server/dfly_main.cc
```diff
@@ -92,6 +103,12 @@ ABSL_FLAG(bool, version_check, true,
           "If true, Will monitor for new releases on Dragonfly servers once a day.");
 
 ABSL_FLAG(uint16_t, tcp_backlog, 256, "TCP listen(2) backlog parameter.");
+#ifdef __linux__
+ABSL_FLAG(string, user, "",
```
romange, line 106: does it mean that distroless containers will have to run with root permissions?

tools/packaging/Dockerfile.distroless-prod
vyavdoshenko: TLS override: For TLS deployments, users override the Docker HEALTHCHECK at container start:
`docker run --health-cmd '/usr/local/bin/httpscheck https://localhost:6379/healthz' ...`

Both httpcheck and httpscheck are included in the image for this purpose.

Why this is an improvement over the current approach: The existing healthcheck.sh sends a Redis PING via netcat - it only checks that the port is open and something responds. For TLS ports, it relies on getting `-ERR Bad TLS header` back (plain text on a TLS port), which is more of a hack than a real health check. The /healthz endpoint actually checks GlobalState::ACTIVE, so it verifies the server is truly ready to serve requests. And httpscheck does a proper TLS handshake rather than relying on error responses from plain text on a TLS port.

src/server/dfly_main.cc
vyavdoshenko: Current state: The distroless image runs as root, same as the existing images before `entrypoint.sh` drops privileges. The `--user` flag allows dropping privileges at runtime after binding ports.

Distroless base ships a built-in nonroot user (uid 65534). We could add USER nonroot to the Dockerfile, but there are trade-offs:

- Volume mounts (-v /host/path:/data) may not be writable by uid 65534 on the host - root doesn't have this problem
- The /data directory ownership needs to be set up at build time (extra COPY --chown layer)
- The --user flag for runtime privilege dropping becomes less useful (can't chown as non-root)

For now, running as root in distroless is less risky than in ubuntu - there's no shell, no package manager, no writable system paths. Users who require non-root can use docker run --user 65534 or Kubernetes securityContext.runAsUser.

I can add `USER nonroot` if you prefer - let me know.

src/server/dfly_main.cc
vyavdoshenko: fixed

src/server/dfly_main.cc
```diff
@@ -203,6 +218,67 @@ template <typename... Args> unique_ptr<Listener> MakeListener(Args&&... args) {
...
+  if (!data_dir.empty()) {
+    namespace fs = std::filesystem;
+    auto do_chown = [&](const fs::path& p) {
+      if (lchown(p.c_str(), uid, gid) != 0) {
+        LOG(WARNING) << "lchown(" << p << ") failed: " << strerror(errno);
+      }
+    };
+    do_chown(data_dir);
+    std::error_code ec;
+    fs::recursive_directory_iterator it(data_dir, ec);
```
romange, line 264: I run claude /review on this, as I do not have much knowledge and it gave me good comments, I think:

1.  dfly_main.cc:230-238 — When a numeric UID is passed (e.g. --user=1000), getpwuid() is called. If it returns NULL (UID not in /etc/passwd), the function fails and the server exits. This is very common in       
  containers — distroless only has root and nobody in passwd. A user running --user=65534 (nobody) works, but --user=1000 fails even though it's a perfectly valid UID for setuid.   Consider falling back to using the raw UID with setgroups(0, nullptr) when getpwuid returns NULL, similar to the more complete version discussed in earlier review rounds. Without this, the --user flag is      
  significantly less useful in the very containers it's designed for.

2.  /healthz requires primary_port_http_enabled — silent failure

  The /healthz handler is registered via ConfigureHttpHandlers (main_service.cc:2797), which only runs when the HTTP console is active on the port. The HTTP listener is only created when
  primary_port_http_enabled=true (default) or on the admin port (dragonfly_listener.cc:162-163).
  worth documenting in the Dockerfile comments.

  3. HEALTHCHECK --start-period=30s is too short for large snapshots

  Dockerfile.distroless-dev:47 — During snapshot loading, gstate() == GlobalState::LOADING (common_types.h:29), so /healthz returns 503. After the 30s start period, Docker begins counting health failures. With
  --retries=3 --interval=15s, the container gets 75s total (30 + 3*15) before being marked unhealthy. Large snapshots can take minutes to load. Worth noting in the Dockerfile comment.

src/server/dfly_main.cc
```diff
@@ -22,7 +22,17 @@
 #endif
 
 #ifdef __linux__
+#include <grp.h>
+#include <pwd.h>
+#include <unistd.h>
```
romange: nit:  Since unistd.h is a POSIX header (not Linux-specific), it could stay outside the guard.

src/server/dfly_main.cc
romange, line 264: use error_code overload for iterating:
```
 std::error_code ec;                                                                                                                                                                                              
  fs::recursive_directory_iterator it(data_dir, ec);                                                                                                                                                               
  for (; !ec && it != fs::recursive_directory_iterator(); it.increment(ec)) {
    do_chown(it->path());                                                                                                                                                                                          
  }                                                                                                                                                                                                                
  if (ec) {                                                                                                                                                                                                        
    LOG(WARNING) << "Error iterating " << data_dir << ": " << ec.message();                                                                                                                                        
  }                                                                                

```

src/server/dfly_main.cc
vyavdoshenko: fixed

src/server/dfly_main.cc
vyavdoshenko: fixed

vyavdoshenko: 1. fixed
2. the comment was updated
3. the comment was updated

PR #6903, opened 2026-03-16
core: add dict_builder module with HLL compressibility estimation

romange:
## Summary
- Add standalone `dict_builder` module with HyperLogLog-based compressibility estimation using 6-byte d-mer hashing
- Implement FastCover-style dictionary training via iterative greedy segment selection
- Add tests validating compressibility detection, dictionary training, and ZSTD compression roundtrip

### DIFF
--- a/src/core/dict_builder.cc
+++ b/src/core/dict_builder.cc
@@ -173,11 +173,10 @@ double EstimateCompressibility(absl::Span<const std::pair<const uint8_t*, size_t
   }
 
   if (total_dmers == 0) {
-    return 0.0;
+    return 1.0;  // No d-mers - we consider it incompressible
   }
 
   double estimate = EstimateHllCardinality(registers.get());
-
   double ratio = estimate / static_cast<double>(total_dmers);
   return std::min(ratio, 1.0);
 }
--- a/src/core/dict_builder.h
+++ b/src/core/dict_builder.h
@@ -16,6 +16,7 @@ namespace dfly {
 // Estimates compressibility by counting unique 6-byte d-mers using HyperLogLog.
 // data_pieces: spans of raw data (e.g., one per QList node).
 // step: sampling stride (1 = every offset, higher = faster but less accurate).
+// Returns a value in [0, 1] where 0 means very compressible, and 1 means incompressible.
 double EstimateCompressibility(absl::Span<const std::pair<const uint8_t*, size_t>> data_pieces,
                                unsigned step);
 


src/core/dict_builder.h
```diff
@@ -0,0 +1,33 @@
...
+
+// Trains a compression dictionary from a collection of sample data.
+//
+// Arguments:
+//   data_pieces:  Input data sources (spans of bytes) to extract dictionary segments from.
+//   dict_size:    The maximum target size of the resulting dictionary in bytes.
+//   segment_size: The size of continuous byte segments chosen and appended per iteration.
+//
+// Returns a raw string containing the trained dictionary up to `dict_size` bytes.
+std::string TrainDictionary(absl::Span<const std::pair<const uint8_t*, size_t>> data_pieces,
```
dranikpg, line 31: A comment about what it applies, what the return value is (double 0-1 percent compressed size relative to total?), the dictionary return value would be useful

dranikpg: Ok, you can tell by the test, but still not obvious

src/core/dict_builder.cc
```diff
@@ -0,0 +1,221 @@
...
+  uint32_t freq_table_mask = freq_table_size - 1;
+
+  unique_ptr<uint16_t[]> freq(new uint16_t[freq_table_size]());
+  PopulateFrequencyTable(data_pieces, freq.get(), freq_table_mask);
+
+  std::string dictionary;
+  dictionary.reserve(dict_size);
+
+  while (dictionary.size() < dict_size) {
+    auto best = FindBestSegment(data_pieces, segment_size, freq.get(), freq_table_mask);
```
dranikpg, line 205: Is this hand written? It's interesting to see more or less good c++ code and then a unique_ptr array and pass by pointer

dranikpg (dismissed):

src/core/dict_builder.cc
romange: I wrote it to avoid stack blow-up. what would you change? pass a slice?

src/core/dict_builder.h
romange: fixed

romange:
augment review

romange:
augment review

src/core/dict_builder.cc
dranikpg: vector and pass as abls::Span

PR #6904, opened 2026-03-17
fix(test): Disable heartbeat during keys hash calculation in tiering replication test

### LINKED ISSUE #6915: ` dragonfly/tiering_test.py::test_replication` fails on comparing keys on master vs replica
` dragonfly/tiering_test.py::test_replication[df_factory0]`

run
https://github.com/dragonflydb/dragonfly/actions/runs/23143520155/job/67225580614#step:7:732

```
        if len(set(hashes)) != 1:
            for key in keys:
                key_master = await async_client.get(key)
                key_replica = await replica_client.get(key)
>               assert key_master == key_replica
E               AssertionError: assert equals failed
E                 '8d5a58ba23d9aac7ffe4a4b3438abf  '8d5a58ba23d9aac7ffe4a4b3438abf 
E                 021e42bbabc4a12a0c16d5763ada736  021e42bbabc4a12a0c16d5763ada736 
E                 a59e0289ec6432941701466b6a3e48c  a59e0289ec6432941701466b6a3e48c 
E                 f81afcf39dcdc2422ad418ab4e53e08  f81afcf39dcdc2422ad418ab4e53e08 
E                 e24420cec93c9d699810dfd02960dd2  e24420cec93c9d699810dfd02960dd2 
E                 0985ec7d38be6258e926f6234d348c7  0985ec7d38be6258e926f6234d348c7 
E                 5b1987aaa4d6a6b24c2d7aeb1e9e736  5b1987aaa4d6a6b24c2d7aeb1e9e736 
E                 87300e7b36039d93d640ea8b2157c40  87300e7b36039d93d640ea8b2157c40 
E                 4beafb8eda0123c3c7e308ee22326ab  4beafb8eda0123c3c7e308ee22326ab 
E                 59ebbd14067e92eaaecac4508a120a5  59ebbd14067e92eaaecac4508a120a5 
E                 5dee0c82e86d4336b394546cc1d014d  5dee0c82e86d4336b394546cc1d014d 
E                 62030137b7a33519f4719e2d0e909bb  62030137b7a33519f4719e2d0e909bb 
E                 3c47a776b792871f1ae961ced2e2869  3c47a776b792871f1ae961ced2e2869 
E                 a21d8edfce020790966d5efc495120a  a21d8edfce020790966d5efc495120a 
E                 d3a6d35b735a1dd4bdbbf9be326577e  d3a6d35b735a1dd4bdbbf9be326577e 
E                 :80::80::80:'                    :80:'

```

mkaruza:
Replication in cache mode can lead to inconsistent key comparisons between master and replica.

Disable heartbeat while we are calculating hash of keys.

Added a utility to compare all keys between master and replica.

Closes #6915

### DIFF
--- a/tests/dragonfly/tiering_test.py
+++ b/tests/dragonfly/tiering_test.py
@@ -150,9 +150,7 @@ async def test_tiered_replication_strings_with_append(df_factory: DflyInstanceFa
     fill_tasks = [asyncio.create_task(fill_job()) for _ in range(3)]
 
     # Start replication
-    logging.info(f"Starting replication {master.port} -> {replica.port}")
     await replica_client.replicaof("localhost", master.port)
-    logging.info("Waiting for replica to sync")
 
     # Wait for replication to finish
     try:


mkaruza:
> Could you plz provide some context in the PR description ?
> 
>     1. Why `cache_mode` does not work/removed ? Do we have a bug ? If so do we have an isue ?

It works, and i don't think we have issue. With cache mode we are updating datastore even after replication is done. When we calculate hash on both replica and master we can have differences there.

>     2. Why the changes to the key comparison ? Why hashes fail to compare properly ?

Dynamic system modify datastore, for example we can have key deleted on master while on replica it still exists.

kostasrim:
> > Could you plz provide some context in the PR description ?
> > ```
> > 1. Why `cache_mode` does not work/removed ? Do we have a bug ? If so do we have an isue ?
> > ```
> 
> It works, and i don't think we have issue. With cache mode we are updating datastore even after replication is done. When we calculate hash on both replica and master we can have differences there.

Maybe I am missing something here but we should not blindly delete or remove flags from tests and I suspect we have an actual bug here (I could be wrong if I am missing context so plz pushback if I am missing something). 

First of all, we have `cache_mode` tests with `replication` in `test_replication_all` without any issues. I don't see how they don't fail but this one does (in other words I suspect we have a bug with tiering)

Second, eventually all operations will stabilize (all in-flight commands will complete) and heartbeat won't evict/free anything. So if everything is stable why do hashes fail ? 🤔 

> 
> > ```
> > 2. Why the changes to the key comparison ? Why hashes fail to compare properly ?
> > ```
> 
> Dynamic system modify datastore, for example we can have key deleted on master while on replica it still exists.

Yes and eventually they should converge. In other words if the hashes don't much this is an actual bug and we are just hiding it out

dranikpg (approved):

tests/dragonfly/utility.py
```diff
@@ -49,6 +49,32 @@ def batch_fill_data(client, gen, batch_size=100):
         client.mset({k: v for k, v, in group})
 
 
+async def get_all_keys(client, BATCH_SIZE=5000):
```
romange, line 52: @mkaruza  I wish you let AI write these things. 

1. it can be written in more concise (pythonic) way and AI can teach you, imho
2. `get_all_keys` can blow up with large datasets. You can make it generator, even batch generator and then run below  ` for batch in get_all_keys(client, ..):`

PR #6906, opened 2026-03-17
[WIP] feat(server): Send tagged chunks

### LINKED ISSUE #6831: shard-serialization P1.4: extend replication wire protocol with tagged chunks
## Phase 1 — PIT and PIT+tiered foundation (Task 4 of 17)

**Goal:** Have the wire-format infrastructure ready before it is needed, so Phase 2 is not blocked on a protocol change.

**Design doc:** [docs/shard-serialization.md — Technical Roadmap](https://github.com/dragonflydb/dragonfly/blob/main/docs/shard-serialization.md#technical-roadmap)

### Dependencies

None — this is pure protocol work and can be developed independently of all other tasks.

### Description

Add a stream-tag framing layer to the replication protocol, gated behind a new \`DflyVersion\`. Each data chunk carries a tag identifying its logical stream (e.g., baseline-bucket vs. journal). The replica reassembles same-tag chunks before decoding.

#### Protocol design
- Each chunk in the replication stream gets a header: \`(stream_tag: uint32, payload_length: uint32, payload: bytes)\`.
- \`stream_tag\` identifies the logical stream. Reserved tags:
  - Tag 0: bucket baseline data (default / legacy).
  - Tag 1: journal entries.
  - Tags 2+: reserved for future use (e.g., per-bucket streams for large values).
- The replica maintains a per-tag reassembly buffer. Same-tag chunks are concatenated in order before decoding.

#### Version gating
- Gate behind a new \`DflyVersion\` value. Replicas that don't support tagged chunks receive the untagged legacy format.
- The RDB save path (backups) does NOT use tagged chunks — only the replication path.

#### Scope
- This task adds the framing layer and replica-side reassembly.
- Sending side initially wraps all output in tag-0 chunks (functionally equivalent to legacy).
- Actual multi-stream use (separate bucket and journal serializers) comes in task 7.

#### Key files
- \`src/server/snapshot.h\` / \`snapshot.cc\` — sending side.
- \`src/server/dfly_protocol.h\` or similar — version definitions.
- Replica-side deserialization (RDB loader / replication consumer).

### Acceptance criteria
- [ ] New \`DflyVersion\` defined and advertised.
- [ ] Tagged-chunk framing implemented on sender side.
- [ ] Replica reassembles tagged chunks correctly.
- [ ] Backward compatibility: old replicas still work (untagged fallback).
- [ ] Replication tests pass with the new protocol version.
- [ ] RDB save path unaffected.

abhijat:
* Master sends a tag in header to replica if:
  * the replica version is `>=` 7 and
  * the stream journal arg is true (to differentiate from rdb file save code path)
* A new IO source is passed to rdb loader. It reads and removes the header if present
* Right now the tag is only set to 0. It is not used by the replica, just parsed and validated.

FIXES https://github.com/dragonflydb/dragonfly/issues/6831

### DIFF
--- a/src/server/CMakeLists.txt
+++ b/src/server/CMakeLists.txt
@@ -96,7 +96,7 @@ add_library(dragonfly_lib
             ${DF_CLUSTER_SRCS}
             acl/user.cc acl/user_registry.cc acl/acl_family.cc
             acl/validator.cc
-            sharding.cc cmd_support.cc)
+            sharding.cc cmd_support.cc tagged_chunk.cc)
 
 if (DF_ENABLE_MEMORY_TRACKING)
   target_compile_definitions(dragonfly_lib PRIVATE DFLY_ENABLE_MEMORY_TRACKING)
@@ -159,6 +159,7 @@ helio_cxx_test(cluster/cluster_family_test dfly_test_lib LABELS DFLY)
 helio_cxx_test(acl/acl_family_test dfly_test_lib LABELS DFLY)
 helio_cxx_test(engine_shard_set_test dfly_test_lib LABELS DFLY)
 helio_cxx_test(serializer_base_test dfly_test_lib LABELS DFLY)
+helio_cxx_test(tagged_chunk_test dfly_test_lib LABELS DFLY)
 
 add_dependencies(check_dfly dragonfly_test json_family_test list_family_test
                  generic_family_test memcache_parser_test rdb_test journal_test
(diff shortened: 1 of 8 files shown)


abhijat:
tests time out https://github.com/dragonflydb/dragonfly/actions/runs/23187956874/job/67376217824 - will need to investigate if it is related to changes.

abhijat:
merged in main PR

PR #6913, opened 2026-03-17
perf(server): optimize OpStrLen for tiered storage

asherlaau:
## Overview

This PR optimizes the `STRLEN` command for tiered storage entries by avoiding unnecessary disk I/O.

When a value is offloaded but **idle** (no modifications currently in flight), the string length can be returned directly from the `CompactObj` metadata already present in memory.

### Optimization
- Updated `OpStrLen` in `string_family.cc` to return `co.Size()` immediately when `HasModificationPending` is false for the associated disk segment.

### Infrastructure
- Introduced `ReadIntent` enum (`READ`, `MODIFY`) in the `dfly::tiering` namespace to distinguish between immutable reads and read-modify-write operations.
- Implemented `HasModificationPending` in `OpManager` and `TieredStorage` to track active modification tasks.
- Updated `OpManager` logic to upgrade a pending `READ` intent to `MODIFY` when a mutable request arrives for the same segment.
- Updated `ModifyTiered` to explicitly signal `ReadIntent::MODIFY`.

## Verification

Added tests in `src/server/tiered_storage_test.cc` covering:

1. **Idle offloaded entries**  
   `STRLEN` returns the correct length with zero disk fetches (validated via `total_fetches` metrics).

2. **Pending modifications**  
   When `APPEND` is in-flight, `STRLEN` bypasses the optimization and performs a read to ensure correctness.

3. **Edge cases**  
   Non-existent keys and standard in-memory strings.

4. **Regression safety**  
   Verified existing `AppendStorm` and `Defrag` tests pass successfully.

### DIFF
--- a/src/server/tiered_storage_test.cc
+++ b/src/server/tiered_storage_test.cc
@@ -139,6 +139,56 @@ TEST_P(LatentCoolingTSTest, SimpleGetSet) {
   EXPECT_EQ(metrics.db_stats[0].tiered_used_bytes, 0);
 }
 
+TEST_P(LatentCoolingTSTest, StrLen) {
+  absl::FlagSaver saver;
+  SetFlag(&FLAGS_tiered_offload_threshold, 1.0f);  // force offloading
+  UpdateFromFlags();
+
+  // Edge case: Non-existent key
+  EXPECT_EQ(Run({"STRLEN", "nonexistent"}), 0);
+
+  const int kLen = 4000;
+  Run({"SET", "k1", BuildString(kLen)});
+
+  // Make sure it's stashed
+  ExpectConditionWithinTimeout([this] { return GetMetrics().tiered_stats.total_stashes == 1; });
+
+  auto metrics_before = GetMetrics();
+
+  // Perform STRLEN - should be optimized (no fetch)
+  auto resp = Run({"STRLEN", "k1"});
+  EXPECT_EQ(resp, kLen);
+
+  auto metrics_after = GetMetrics();
+  EXPECT_EQ(metrics_after.tiered_stats.total_fetches, metrics_before.tiered_stats.total_fetches);
+
+  // Now perform APPEND which should trigger a modification pending
+  Run({"APPEND", "k1", "suffix"});
+
+  // STRLEN now should still work but will fetch if modification IS pending
+  // NOTE: with experimental cooling enabled, the value might still be in memory (CoolQueue)
+  // so it might NOT trigger a disk fetch if it was just appended to.
+  resp = Run({"STRLEN", "k1"});
+  EXPECT_EQ(resp, kLen + 6);
+
+  // Verify fetch happened IF cooling is disabled (direct to disk)
+  metrics_after = GetMetrics();
+  LOG(INFO) << "GetParam: " << GetParam()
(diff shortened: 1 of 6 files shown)


src/server/tiering/op_manager.h
```diff
@@ -98,6 +102,7 @@ class OpManager {
     // We may have multiple callbacks for the same entry.
     absl::InlinedVector<ReadCallback, 1> read_cbs;
     std::unique_ptr<Decoder> decoder;
+    ReadIntent intent;
```
dranikpg: a bool probably would'be been enough as there is no other value except those two binary states

src/server/tiering/op_manager.cc
```diff
@@ -50,14 +50,27 @@ void OpManager::Close() {
...
+  auto it = pending_reads_.find(segment.ContainingPages().offset);
+  if (it == pending_reads_.end())
+    return false;
+
+  const auto& read_op = it->second;
+  for (const auto& ops : read_op.entry_ops) {
+    if (ops.segment.offset == segment.offset && ops.intent == ReadIntent::MODIFY)
+      return true;
+  }
+  return false;
```
dranikpg: as far as I remember the invariant is that we keep only a single entry for the offset - we use a vector (inlined) to reduce data structure complexity. Based on this

```c++
auto* ops_ptr = Find(segment.offset);
return ops_ptr ? ops_ptr->intent == MODIFY : false;
```

should be all

dranikpg (approved):
Thank you, this is quality wise a really good contribution. And the tests are decent as well. Left some small nits what could be possibly made simpler, but happy to accept it as such as well

src/server/tiering/op_manager.h
asherlaau: Thank you for the review! I replaced the enum with a bool (read_only) to simplify it.

src/server/tiering/op_manager.cc
asherlaau: I also updated the lookup to use ReadOp::Find(segment) instead of iterating. I hadn't noticed that helper earlier.
Since HasModificationPending is a const function, I added a const overload for Find as well.

dranikpg (dismissed):

dranikpg:
augment review

src/server/tiered_storage_test.cc
```diff
@@ -139,6 +139,56 @@ TEST_P(LatentCoolingTSTest, SimpleGetSet) {
...
+  auto metrics_before = GetMetrics();
+
+  // Perform STRLEN - should be optimized (no fetch)
+  auto resp = Run({"STRLEN", "k1"});
+  EXPECT_EQ(resp, kLen);
+
+  auto metrics_after = GetMetrics();
+  EXPECT_EQ(metrics_after.tiered_stats.total_fetches, metrics_before.tiered_stats.total_fetches);
+
+  // Now perform APPEND which should trigger a modification pending
```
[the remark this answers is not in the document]
asherlaau: You're right that APPEND may complete before STRLEN executes, so the pending
state may not always be observed.

The test is mainly intended to verify correct behavior in the presence of
modifications rather than strictly enforcing the timing of the pending state.

dranikpg:
@asherlaau 

One thing left before merging - please sign your commits so they are verified

PR #6920, opened 2026-03-18
feat(core): introduce TOPK data structure (Stage 1)

glevkovich:
- Introduce the core TOPK class to provide the backing data structure for the Redis Top-K command family.
- Add topk.cc to dfly_core and implement a thorough suite of unit tests (topk_test) covering all edge cases and math invariants.
- Separated this from the internal TopKeys tracker to provide PMR support and RDB serialization required for public Redis APIs.

### DIFF
--- a/src/core/topk.cc
+++ b/src/core/topk.cc
@@ -199,7 +199,7 @@ size_t TOPK::GetCounterIndex(std::string_view item, uint32_t row) const {
   return idx;
 }
 
-uint32_t TOPK::GetMinCount(std::string_view item) const {
+uint32_t TOPK::Count(std::string_view item) const {
   uint32_t min_count = std::numeric_limits<uint32_t>::max();
 
   for (uint32_t row = 0; row < depth_; ++row) {
@@ -211,6 +211,8 @@ uint32_t TOPK::GetMinCount(std::string_view item) const {
 }
 
 std::optional<std::string> TOPK::IncrementInternal(std::string_view item, uint32_t increment) {
+  uint32_t min_count = std::numeric_limits<uint32_t>::max();
+
   // Update counters using HeavyKeeper logic
   for (uint32_t row = 0; row < depth_; ++row) {
     size_t idx = GetCounterIndex(item, row);
@@ -226,11 +228,11 @@ std::optional<std::string> TOPK::IncrementInternal(std::string_view item, uint32
           std::min(static_cast<uint64_t>(counters_[idx]) + increment,
                    static_cast<uint64_t>(std::numeric_limits<uint32_t>::max())));
     }
-  }
 
-  // Count-Min Sketch property: The minimum counter across all rows is the
-  // most accurate, as it has suffered the fewest hash collisions.
-  uint32_t min_count = GetMinCount(item);
+    // Count-Min Sketch property: The minimum counter across all rows is the
+    // most accurate, as it has suffered the fewest hash collisions.
+    min_count = std::min(min_count, counters_[idx]);
+  }
 
   return UpdateHeap(item, min_count);
 }
@@ -262,8 +264,11 @@ std::vector<TOPK::TopKItem> TOPK::List() const {
 }
 
 std::optional<std::string> TOPK::UpdateHeap(std::string_view item, uint32_t new_count) {
-  // Fast path: O(K) cache-friendly linear scan over contiguous memory.
-  // For small K, this is faster than paying the overhead of a hash map lookup.
+  // Fast path: O(K) linear scan.
+  // For small K, this avoids hash map overhead. Short keys benefit from SSO
+  // (Small String Optimization), keeping memory contiguous and cache-friendly.
+  // TODO: Benchmark to find the crossover point where larger K OR long strings (SSO not applicable)
+  // justify re-introducing a hash map.
   for (size_t i = 0; i < min_heap_.size(); ++i) {
     if (min_heap_[i].key == item) {
       uint32_t old_count = min_heap_[i].count;
--- a/src/core/topk.h
+++ b/src/core/topk.h
@@ -111,9 +111,7 @@ class TOPK {
 
   // Estimates the frequency count for an item using the underlying sketch.
   // Returns the minimum counter value across all hash rows (Count-Min Sketch estimate).
-  [[nodiscard]] uint32_t Count(std::string_view item) const {
-    return GetMinCount(item);
-  }
+  [[nodiscard]] uint32_t Count(std::string_view item) const;
 
   // Retrieves the complete list of current Top-K high-frequency items.
   //
@@ -187,9 +185,6 @@ class TOPK {
 
   // Exponential decay logic
   [[nodiscard]] bool ShouldDecay(uint32_t current_count) const;
-
-  // Get the minimum count for an item across all hash table rows
-  [[nodiscard]] uint32_t GetMinCount(std::string_view item) const;
 
   // Updates the min-heap with the new count for the given item.
   // Returns the evicted item's key if the heap is at capacity and a new item displaces an existing


src/core/topk.h
```diff
@@ -0,0 +1,269 @@
...
+#pragma once
+
+#include <cstdint>
+#include <memory>
+#include <optional>
+#include <queue>
+#include <string>
+#include <string_view>
+#include <vector>
+
```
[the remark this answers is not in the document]
glevkovich: fixed

src/core/topk_test.cc
```diff
@@ -0,0 +1,729 @@
...
+// Counter Saturation (Overflow Prevention)
+// ---------------------------------------------------------------------------
+
+// IncrBy must saturate at UINT32_MAX rather than wrapping around to 0.
+// A wrap-around would trick the heap into evicting a top item — a correctness
+// and security issue (malicious TOPK.INCRBY with a huge increment).
+TEST_F(TOPKTest, CounterSaturationPreventsOverflow) {
+  const uint32_t max_val = numeric_limits<uint32_t>::max();
+  topk_.IncrBy("max_item", max_val);
+  EXPECT_EQ(topk_.Count({"max_item"})[0], max_val);
```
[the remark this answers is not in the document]
glevkovich: fixed

src/core/topk.cc
```diff
@@ -0,0 +1,440 @@
...
+
+  return data;
+}
+
+void TOPK::Deserialize(const SerializedData& data) {
+  DCHECK_EQ(data.counters.size(), static_cast<size_t>(width_) * depth_);
+  DCHECK_LE(data.heap_items.size(), k_);
+  DCHECK_EQ(data.k, k_);
+  DCHECK_EQ(data.width, width_);
+  DCHECK_EQ(data.depth, depth_);
```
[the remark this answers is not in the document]
glevkovich: fixed

src/core/topk.cc
```diff
@@ -0,0 +1,440 @@
+// Copyright 2026, DragonflyDB authors.  All rights reserved.
+// See LICENSE for licensing terms.
+//
+
+#include "core/topk.h"
+
+#include <xxhash.h>
+
+#include <algorithm>
+#include <cmath>
```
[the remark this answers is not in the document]
glevkovich: fixed

src/core/topk.h
```diff
@@ -0,0 +1,269 @@
...
+  // Updates the min-heap with the new count for the given item.
+  // Returns the evicted item's key if the heap is at capacity and a new item displaces an existing
+  // one. Otherwise, returns std::nullopt.
+  std::optional<std::string> UpdateHeap(std::string_view item, uint32_t new_count);
+
+  // Check if an item is in the Top-K heap
+  [[nodiscard]] bool IsInHeap(std::string_view item) const {
+    return item_to_hash_.contains(item);
+  }
```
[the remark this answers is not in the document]
glevkovich: this path is already optmized by absl. dropping.

src/core/topk.h
```diff
@@ -0,0 +1,269 @@
...
+
+  // Extracts the current structural state of the sketch for RDB persistence.
+  [[nodiscard]] SerializedData Serialize() const;
+
+  // Reconstructs the internal state of the sketch from a previously serialized dataset.
+  void Deserialize(const SerializedData& data);
+
+ private:
+  struct HeapItem {
+    std::string key;
```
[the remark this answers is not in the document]
glevkovich: will be done on next stage

src/core/top_keys.h
```diff
@@ -9,10 +9,20 @@
 #include <string>
 #include <string_view>
 #include <vector>
+
 #include "base/random.h"
 
 namespace dfly {
 
+// INTERNAL USE ONLY: This class is an optimized, O(1) probabilistic hot-key tracker designed
+// specifically to run on the database's hot path (e.g., tracking hot keys for DEBUG HTKEYS).
```
romange: HTKEYS -> `TOPK`

src/core/topk.cc
```diff
@@ -0,0 +1,441 @@
...
+#include "absl/random/distributions.h"
+#include "base/logging.h"
+#include "base/random.h"
+
+namespace dfly {
+
+namespace {
+
+const std::array<double, TOPK::kDecayLookupSize>& GetDefaultDecayTable() {
+  static const auto table = [] {
```
romange, line 23: can you please explain?


Image

src/core/topk.cc
```diff
@@ -0,0 +1,441 @@
...
+            [](const TopKItem& a, const TopKItem& b) { return a.count > b.count; });
+
+  return result;
+}
+
+std::optional<std::string> TOPK::UpdateHeap(std::string_view item, uint32_t new_count) {
+  // Fast path: O(1) hash lookup for membership, followed by a cache-friendly
+  // O(K) linear scan and O(log K) heap update.
+  auto it = item_to_hash_.find(item);
+  if (it != item_to_hash_.end()) {
```
romange: it should not happen?

src/core/topk.cc
```diff
@@ -0,0 +1,441 @@
...
+  uint32_t min_count = GetMinCount(item);
+
+  return UpdateHeap(item, min_count);
+}
+
+std::optional<std::string> TOPK::Add(std::string_view item) {
+  return IncrementInternal(item, 1);
+}
+
+std::vector<std::optional<std::string>> TOPK::AddMultiple(
```
romange: why this should be part of the  API?

src/core/topk.h
```diff
@@ -0,0 +1,276 @@
...
+  [[nodiscard]] bool IsInHeap(std::string_view item) const {
+    return item_to_hash_.contains(item);
+  }
+
+  // Hashes the item for a specific row and calculates its flattened 1D index
+  // within the counters_ array. Maps the 2D Count-Min Sketch grid (depth x width)
+  // into a single contiguous block of memory for better CPU cache locality.
+  size_t GetCounterIndex(std::string_view item, uint32_t row) const;
+
+  // Shared increment logic
```
romange, line 208: what does it return?

src/core/topk.h
```diff
@@ -0,0 +1,276 @@
...
+  //          contains either the evicted item resulting from that specific
+  //          insertion, or std::nullopt.
+  std::vector<std::optional<std::string>> AddMultiple(const std::vector<std::string_view>& items);
+
+  // Increments an item's estimated frequency by a specific, arbitrary amount.
+  //
+  // Precondition: 'increment' must be strictly greater than 0.
+  // Returns: The string of the evicted item if this operation caused a resident
+  //          item to be displaced from the Top-K min-heap, or std::nullopt.
+  std::optional<std::string> IncrBy(std::string_view item, uint32_t increment);
```
romange, line 105: do we need returning the evicted item as part of the TOPK API?

src/core/topk.h
```diff
@@ -0,0 +1,276 @@
...
+  std::vector<std::optional<std::string>> IncrByMultiple(
+      const std::vector<std::pair<std::string_view, uint32_t>>& items);
+
+  // Queries whether a batch of items currently resides in the Top-K min-heap.
+  //
+  // Input: A vector of strings (items) to check against the Top-K list.
+  // Output: A vector of integers mapping 1:1 to the input vector.
+  //         The i-th element of the output will be 1 if items[i] is
+  //         currently a Top-K high-frequency item, and 0 otherwise.
+  [[nodiscard]] std::vector<int> Query(const std::vector<std::string_view>& items) const;
```
romange: I think these APIs do not contribute much - I did not see any batch-related optimization that justifies this, so why?

src/core/top_keys.h
glevkovich: fixed

src/core/topk.cc
glevkovich: I wrote explanations in header. There are two key  choices here:

- Speed vs. Memory: The lookup table pre-calculates pow() results to keep the hot path O(1). I increased its size to prioritize execution speed over a small, 32KB memory trade-off.
- Shared Defaults: I assume that vast majority of users will use the default decay. If no one users it, we wasted 32KB of memory. We share this single table across all TOPK instances using the default decay. This means 1,000 keys only occupy the memory of one table. Custom (non-default) decays allocate their own instance-specific table (which we can optimize for deduplication later).
 
Also, why the lambda and thread safety?  This is thread-safe by default due to C++11 "magic statics" . the compiler ensures the table is built exactly once, even if multiple threads call this simultaneously, without needing manual locks.

src/core/topk.cc
glevkovich: Could you clarify your point regarding the membership check?I first used item_to_hash_ as a fast filter, assuming the O(1) lookup would be more efficient than a scan. However, I now realize this creates a 'double lookup' (hashing the string for the map, then performing a linear scan anyway to find the heap index).

Rethinking, perhaps I should remove the  outer find find() entirely.. given that K is typically small , maybe a single linear scan to benefit from better cache locality and avoid the CPU overhead of hashing the string twice.

glevkovich: This also simplifies code, reduce memory overhead since item_to_hash_ is removed , and simplify memory tracking for PMRs.

glevkovich: So aliminating this double lookup will actually improve performance. The decision to use a hash map for O(1) was not optimal , since K is typically small, the overhead of hashing and potential cache misses outweighs the benefits of the map.I will remove the item_to_hash_ and relay on a linear scan and cache locallity.

src/core/topk.cc
glevkovich: Im going to remove  AddMultiple/IncrByMultiple - they are leftover convenience wrappers that I did not remove. , The family layer (topk_family.cc) can simply loop over and call Add/IncrBy and stream replies directly with less intermidiate vectors allocated.

src/core/topk.h
glevkovich: It actually calls  UpdateHeap and returns it return value. I will update the comment above.

```
// Increments the item's count in the sketch and updates the min-heap.
  // Returns: The key of the item evicted from the Top-K list, or std::nullopt 
  //          if no eviction occurred.
  std::optional<std::string> IncrementInternal(std::string_view item, uint32_t increment);
```

src/core/topk.h
glevkovich: Yes, it is required. The Redis API  for both TOPK.ADD and TOPK.INCRBY dictates that the command must return "Array of Simple string reply - if an element was dropped from the TopK list, Nil reply otherwise.."

src/core/topk.h
glevkovich: i;m going to change both Query and Count to have a single input element.

src/core/topk.cc
```diff
@@ -0,0 +1,368 @@
+// Copyright 2026, DragonflyDB authors.  All rights reserved.
+// See LICENSE for licensing terms.
+//
+
+#include "core/topk.h"
+
+#include <xxhash.h>
+
+#include <algorithm>
+#include <cmath>
```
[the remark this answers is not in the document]
glevkovich: fixed

src/core/topk.h
```diff
@@ -0,0 +1,249 @@
...
+  // Returns: The string of the evicted item if this insertion caused a resident
+  //          item to be displaced from the Top-K min-heap, or std::nullopt
+  //          if no eviction occurred.
+  std::optional<std::string> Add(std::string_view item);
+
+  // Increments an item's estimated frequency by a specific, arbitrary amount.
+  //
+  // Precondition: 'increment' must be strictly greater than 0.
+  // Returns: The string of the evicted item if this operation caused a resident
+  //          item to be displaced from the Top-K min-heap, or std::nullopt.
```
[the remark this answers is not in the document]
glevkovich: fixed

src/core/topk.cc
romange: yes, I asked about the lambda hence the inception meme.
Can you explain the "magic statics" hack? why it is thread-safe? I was not aware of it.

src/core/topk.cc
```diff
@@ -0,0 +1,369 @@
...
+    } else {
+      counters_[idx] = static_cast<uint32_t>(
+          std::min(static_cast<uint64_t>(counters_[idx]) + increment,
+                   static_cast<uint64_t>(std::numeric_limits<uint32_t>::max())));
+    }
+  }
+
+  // Count-Min Sketch property: The minimum counter across all rows is the
+  // most accurate, as it has suffered the fewest hash collisions.
+  uint32_t min_count = GetMinCount(item);
```
romange: you do not need `GetMinCount` here, you can calculate it in the loop

src/core/topk.cc
```diff
@@ -0,0 +1,369 @@
...
+  // Sort by count (descending) for output
+  std::sort(result.begin(), result.end(),
+            [](const TopKItem& a, const TopKItem& b) { return a.count > b.count; });
+
+  return result;
+}
+
+std::optional<std::string> TOPK::UpdateHeap(std::string_view item, uint32_t new_count) {
+  // Fast path: O(K) cache-friendly linear scan over contiguous memory.
+  // For small K, this is faster than paying the overhead of a hash map lookup.
```
romange: I am not so sure and it's not contiguous memory.`min_heap_[i].key == item` is not contiguous. 

I any case I am fine to start like this and then we can follow up with benchmarking.

src/core/topk.cc
glevkovich: Thanks, I'll move it into the loop.

src/core/topk.cc
```diff
@@ -0,0 +1,369 @@
...
+  }
+}
+
+void TOPK::HeapifyDown(size_t index) {
+  DCHECK_LT(index, min_heap_.size());
+  // Restores the min-heap property by shifting the element at 'index' downward.
+  // Triggered in two cases:
+  // 1. Root replacement/removal: The minimum item is evicted/replaced and the new root must sink.
+  // 2. Count increase: An existing item's count grows (becomes heavier), sinking lower.
+  size_t size = min_heap_.size();
```
romange, line 166: would https://en.cppreference.com/w/cpp/algorithm/make_heap.html 

work here?

src/core/topk.cc
glevkovich: The std::string data pointers break contiguity for larger strings. (not for SSO) .I'll update the comment to be more accurate, and we should follow  up with benchmarking on later stages..

src/core/topk.cc
glevkovich: btw.i'm using make_heap on Deserialize().

And it will work here functionally-wise, but will be a performance disaster probably  - it throws the existing heap and build it from scratch. STL provided pop_heap() and push_heap() to add item at the end or remove from the root. but no API for updating item in the middle of the heap. We are updating arbitrary indices in the heap.

romange (dismissed):
lgtm

src/core/topk.cc
glevkovich: Adding here part of the message I wrote in the C++ channel:

- Before C++11, doing this concurrently was undefined behavior.
- Since C++11, block-scope static variables (often called "magic statics") are guaranteed to be thread-safe by the compiler.
- According to cppreference: "If multiple threads attempt to initialize the same static local variable concurrently, the initialization occurs exactly once." (See the 'Static local variables' section here: https://en.cppreference.com/w/cpp/language/storage_duration).
- Because of this compiler-level guarantee, my claim is that initializing this shared lookup table using a static lambda is perfectly thread-safe and avoids the overhead of manual locking.

PR #6922, opened 2026-03-18
chore: switch actions to ubuntu-dev:24

romange:
Keep build step in the release pipeline with ubuntu-dev:20.

Confirmed that both Heavy and Regression tests pass.


romange:
> worth kicking a release and then deleting it for testing.
> 
> I can do this instead if you like :)

did it already

PR #6923, opened 2026-03-18
chore(server): remove expire table

romange:
## Summary

Each expiring key previously required both a prime-table entry and a
separate `ExpireTable` entry (~26-35 extra bytes per key). Building on the
SDS_TTL_TAG CompactKey support from PR #6925, the TTL is now embedded
directly inside the key, making the per-shard `ExpireTable` redundant.

- Replace expire table insert/lookup/erase with `SetExpireTime` /
  `GetExpireTime` / `ClearExpireTime` on `CompactKey`
- Scan prime table for expiry instead of expire table
- Fix memory accounting in `AddExpire`/`RemoveExpire` for inline <->
  `SDS_TTL_TAG` key transitions (`inline_keys` counter + `OBJ_KEY` heap tracking)
- Add regression test for inline key accounting

## Benchmark

`debug populate 10000000 key 16 EXPIRE 300 1000`

| | Used memory |
|-|-------------|
| Before | 900 MB |
| After  | 665 MB |
------------------------------------------------------------------------------
**Next step:** clean up of  redundant types and arguments, simplify the interfaces (ExpireIterator, AsRef() etc).

### DIFF
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -57,15 +57,9 @@ using Payload = journal::Entry::Payload;
 namespace {
 
 constexpr auto kPrimeSegmentSize = PrimeTable::kSegBytes;
-constexpr auto kExpireSegmentSize = ExpireTable::kSegBytes;
-constexpr auto kExpireRegularSize = ExpireTable::kSegRegularBytes;
+
 // mi_malloc good size is 32768. i.e. we have malloc waste of 1.5%.
 static_assert(kPrimeSegmentSize <= 32304);
-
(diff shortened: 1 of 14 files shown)


romange:
augment review

romange:
It's indeed the only "regression" i found but we yet need to see if it is real. i.e. if the real use-cases have extremely low `volatile / total` ratio where background expiry will be inefficient.

In addition to background expiry but we check for expired items when updating existing keys.  I believe we may have there some room for optimizations if we see that background traversal is not efficient.

PR #6924, opened 2026-03-19
tests: Remove index=True kwarg from class

abhijat:
The class init seems to fail on passing `index=True`. Oddly the library home page in docs does not show the keyword is deprecated or removed but it does not seem to be accepted anymore.


abhijat:
this should be fixed by 

https://github.com/dragonflydb/dragonfly/pull/6922

which runs CI on ubuntu 24

PR #6925, opened 2026-03-19
feat(core): embed TTL directly in CompactKey via SDS_TTL_TAG

romange:
## Summary
- Add `SDS_TTL_TAG` to `CompactObj` that stores key string as SDS alongside an int64 TTL, removing the need for separate expiry table lookups
- Refactor encoded blob retrieval into `GetEncodedBlob()` to deduplicate logic across `GetString` and `GetByteAtIndex`
- Add `CompactKey::SetTtl`/`ClearTtl`/`GetTtl` API with full support for converting from all existing tag types (inline, INT, SMALL, ROBJ)

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -1446,17 +1446,24 @@ size_t CompactObj::MallocUsed(bool slow) const {
   return 0;
 }
 
-bool CompactObj::operator==(const CompactObj& o) const {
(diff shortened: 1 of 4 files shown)


src/core/compact_object.cc
dranikpg: See the TODO, this needs to be fixed

romange: good catch.

PR #6929, opened 2026-03-19
feat: backpressure offloading for ioloopv2 (memcache)

### LINKED ISSUE #6030: Offload connection backpressure to disk + test
Last part of #5952 

* Use the disk backed queue to offload and read data to/from disk
* Adjust the logic to ConnFb and AsyncFb to work with disk backed backpressure
* Test

kostasrim:
* add --pipeline_disk_offload_threshold (IoLoopV2 only) to trigger offloading once the parsed command queue exceeds a byte threshold.
* add per-connection disk offload state (disk_queue_, in-flight flags) and helpers to push raw io_buf_ bytes to disk and later restore them.
* add tests

resolves #6030


src/facade/dragonfly_connection.cc
```diff
@@ -2555,6 +2588,81 @@ void Connection::EnsureMemoryBudget(unsigned tid) {
...
+  //      reading.
+  //   2. ParseLoop and CheckIoBufCapacity both check disk_push_in_flight_ and skip any
+  //      operation that would re-parse or reallocate the buffer while the write is in flight.
+  // Calling ConsumeInput here would reset offs_/size_ to 0, making AppendBuffer() alias the
+  // same memory io_uring is still reading, and would allow Reserve() to free that memory.
+  disk_queue_->PushAsync(io_buf_.InputBuffer(), [this, to_offload](std::error_code ec) {
+    disk_push_in_flight_ = false;
+    if (ec) {
+      LOG(ERROR) << "Disk offload write failed: " << ec.message();
+      // Should we abort the connection ?
```
kostasrim: How should we handle failing writes or reads ? Shall we shut down the connection ? Shall we retry ?

I guess a first step would be to just abort the connection ? 

WDYT @romange

src/facade/dragonfly_connection.cc
```diff
@@ -2555,6 +2588,81 @@ void Connection::EnsureMemoryBudget(unsigned tid) {
...
+  if (!InitDiskQueueIfNeeded())
+    return;
+
+  if (!disk_queue_->HasEnoughBackingSpaceFor(io_buf_.InputLen()))
+    return;
+
+  const size_t to_offload = io_buf_.InputLen();
+  disk_push_in_flight_ = true;
+
+  // We intentionally defer ConsumeInput until the write completes rather than deep-copying
```
kostasrim: This or `copy by value`. 

I chose to `pin` the memory, effectively deferring to consume it until the `in-flight write` completes. 

The cost here is that we need to be careful to not empty or consume or resize the io buffer.

The reason I opted in for this approach is that I don't want `yet another level of copy` 

1. Copy for the callback so we don't have to pin memory (I did not want this)
2. Copy when you Write(). The kernel doesn't just write to disk, it just writes to the `page cache` and marks the `region as dirty` 
3. Finally the page cache gets written (flushed to disk)

I explicitly decided to remove (1) copy

@romange if you disagree object or anything else to add I am all eyes 😄

kostasrim: Ok after having a little bit better understanding of the side effects of that decision, the answer is that it messes tcp ordering on io_buf_  and fixing those cases offloads the backpressure back on the tcp layer (which beats the purpose of it). 

Using two buffers solves all the ordering issues and provides a good separation of concerns:

socket_read->io_buf->offloaded_to_disk
disk_read->io_buf_for_disk

This is a simple produer/consumer. socket reads go directly to disk and new parsed commands are first loaded/read from disk and then we fall back to io_buf_ read providing natural data flow and ordering.

kostasrim:
@romange plz look on my comments

src/facade/dragonfly_connection.h
```diff
@@ -358,12 +365,12 @@ class Connection : public util::Connection {
   // Returns true if one or more commands were parsed from the read buffer,
   // and false if no complete commands could be parsed (for example, when
   // parsing is pending more input).
-  bool ParseMCBatch();
```
romange: We do not pass mutable arge by reference. Why do you need to pass a buffer, it's not a data member?

src/facade/dragonfly_connection.h
```diff
@@ -407,11 +414,19 @@ class Connection : public util::Connection {
...
+  bool disk_push_in_flight_ = false;
+  bool disk_pop_in_flight_ = false;
+
   // how many bytes of the current request have been consumed
   size_t request_consumed_bytes_ = 0;
 
   util::FiberSocketBase::ProvidedBuffer recv_buf_;
-  io::IoBuf io_buf_;  // used in io loop and parsers
+  // parser input: fed exclusively from disk pops (or socket_buf_ when disk empty)
+  io::IoBuf io_buf_;
```
romange, line 470: Why do you need 2 instances?

docs/disk-offload-state-machine.md
```diff
@@ -0,0 +1,215 @@
...
+## 1. Watermark Flags
+
+Three thresholds form a three-level band:
+
+| Flag                                   | Default      | Role |
+|----------------------------------------|--------------|------|
+| `pipeline_disk_offload_threshold`      | 0 (disabled) | **Offload trigger** – `parsed_cmd_q_bytes_ >= threshold` → incoming bytes go to disk instead of `io_buf_`. |
+| `disk_backpressure_hysteresis_arm`     | x KB         | **High-water mark** – once `total_backing + queued >= arm`, `hysteresis_armed` is set, enabling the drain phase. |
+| `disk_backpressure_hysteresis_trigger` | y KB         | **Low-water mark** – while armed and `total_backing + queued < trigger`, `IsDraining()=true`. New socket reads are blocked until the queue fully drains to memory. |
+
```
dranikpg, line 14: I think Roman needs to give his opinion on this, but it really seems to me that there is a way to semi-automate those flags. Specifically the lower bound - maybe we should know whether the client is reading from the socket or just sending, that alone will give us information. If the goal of this feature is to avoid deadlocks with heavy pushing clients, we can differentiate those from a just slow instance that piles up commands

dranikpg: So if a client had many succesive writes without reads, it likely is pushing without reading, so we can expand the limit and not care about shrinking. But as soon as it starts reading (or when it is reading), we can just thorttle it to avoid growing unbounded and having a large pile. Otherwise it seems that with every client that writes faster than dragonfly can process we will fall back to the disk offloading, which is actually not needed.

PR #6930, opened 2026-03-19
chore: improve naming and get rid of references for mcflag

romange:
1. CompactObj references are fragile and will be removed. This Pr removes them from mcflag table.
2. The Ttl API in CompactKey was wrong - we keep the expire time, and not the ttl there.


romange:
augment review

PR #6936, opened 2026-03-20
fix(search): guard LoadEntry against freed DocIds in global HNSW KNN

vyavdoshenko:
- `SearchGlobalHnswIndex` calls `Knn()` holding the HNSW read lock, then dispatches `ScheduleSingleHop` to serialize results. A concurrent `DEL` on a shard thread calls `RemoveDocFromGlobalVectorIndex()`, which tries the HNSW write lock - fails (read lock held) - and defers the HNSW removal. `RemoveDoc()` then calls `key_index_.Remove(id)` immediately, adding `id` to `free_ids_`. When `ScheduleSingleHop` later calls `LoadEntry(id)` -> `key_index_.Get(id)`, the DCHECK fires: SIGABRT.
- Fix: `LoadEntry()` now checks `key_index_.IsValid(id)` before calling `Get()`, returning `nullopt` for stale IDs.

Changes:
- `src/server/search/doc_index.cc`: add `IsValid` guard in `LoadEntry()`
- `src/server/search/search_family_test.cc`: add `HnswRaceTest::HnswKnnDeleteRaceCrash` - a deterministic regression test that crashes without the fix and passes with it (search on a dedicated I/O thread, 3 delete fibers on the shard threads, DIM=128 / 5000 docs to widen the Knn() race window)

Fixes regression test: replication_test.py::test_hnsw_search_replication_with_network_disruptions

### DIFF
--- a/src/server/search/doc_index.cc
+++ b/src/server/search/doc_index.cc
@@ -695,6 +695,8 @@ bool ShardDocIndex::Matches(string_view key, unsigned obj_code) const {
 
 optional<ShardDocIndex::LoadedEntry> ShardDocIndex::LoadEntry(DocId id,
                                                               const OpArgs& op_args) const {
+  if (!key_index_.IsValid(id))
+    return std::nullopt;
   auto& db_slice = op_args.GetDbSlice();
   string_view key = key_index_.Get(id);
   auto it = db_slice.FindReadOnly(op_args.db_cntx, key, base_->GetObjCode());
--- a/src/server/search/search_family_test.cc
+++ b/src/server/search/search_family_test.cc
@@ -8,6 +8,7 @@
 #include <absl/strings/str_format.h>
 
 #include <algorithm>
+#include <atomic>
 #include <string_view>
 
 #include "base/gtest.h"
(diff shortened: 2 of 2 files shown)


vyavdoshenko:
> The two hop problem is broader in a sense - the entry can not only be deleted in that time, it can be replaced - so it will return wrong data

What do you suggest?

dranikpg:
I assume only using one more mutex or latch for such cases. I don't have a good idea and only the assumption that global data might not be integrable into a sharded system without issues

romange:
Fixes #6941

romange:
Lets focus on fixing the crashes first. ABA - is currently out of scope.

what about 5.2 here?
https://github.com/dragonflydb/dragonfly/issues/6941#issuecomment-4105981802

is it a valid concern? it says that `IdsOnly` check may lead to a similar crash.

dranikpg:
What is https://github.com/dragonflydb/dragonfly/pull/6939 for? This PR is perfectly fine for fixing the crash

dranikpg:
> ABA - is currently out of scope.

With high volume and many connections you could eventually hit ABA. It doesn't seem unlikely

romange:
> What is #6939 for? This PR is perfectly fine for fixing the crash

it's nothing. I asked copilot to explain the issue in more detail and this is how he outputs his answers.

vyavdoshenko:
Regarding 5.2 (btw, Copilot commented it as an issue, and I checked it separately).

The IdsOnly() path in ShardDocIndex::Search() is not reachable from the global HNSW pipeline. SearchGlobalHnswIndex never calls ShardDocIndex::Search() - it always goes directly through SerializeDocWithKey -> LoadEntry(), which is now guarded. The IdsOnly() branch is only exercised by shard-local text/numeric searches, where the shard lock is held throughout ScheduleSingleHop, so stale DocIds cannot appear. I confirmed this by writing a NOCONTENT variant of the race test — it never crashes without an IdsOnly guard, which is consistent with the code path analysis.

PR #6938, opened 2026-03-20
feat(search): support VECTOR_RANGE operator in FT.AGGREGATE with YIELD_DISTANCE_AS

### LINKED ISSUE #6802: Support `vector_range` operator for search

vyavdoshenko:
Implements VECTOR_RANGE support for FT.AGGREGATE, enabling semantic routing use cases where documents are retrieved by vector similarity and then grouped or reduced by metadata fields.

Fixes #6802

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -56,6 +56,7 @@ jobs:
           submodules: true
 
       - name: Configure AWS credentials
+        if: github.repository_owner == 'dragonflydb' && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
         uses: aws-actions/configure-aws-credentials@v5
         with:
           role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -228,8 +228,7 @@ jobs:
           targets: 'dragonfly'
 
       - name: Authenticate to AWS
-        # Runs if it's NOT a PR, OR if the PR originates from the same repository (not a fork)
-        if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
+        if: github.repository_owner == 'dragonflydb' && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
         uses: aws-actions/configure-aws-credentials@v5
         with:
           role-to-assume: ${{ secrets.AWS_CI_S3_ROLE_ARN }}
--- a/.github/workflows/docker-dev-release.yml
+++ b/.github/workflows/docker-dev-release.yml
@@ -53,6 +53,7 @@ jobs:
           credentials_json: ${{ secrets.GCP_SA_KEY }}
 
       - name: Configure AWS Credentials
+        if: github.repository_owner == 'dragonflydb' && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
         uses: aws-actions/configure-aws-credentials@v5
         with:
           role-to-assume: ${{ secrets.AWS_CI_S3_ROLE_ARN }}
--- a/.github/workflows/epoll-regression-tests.yml
+++ b/.github/workflows/epoll-regression-tests.yml
@@ -47,6 +47,7 @@ jobs:
           targets: 'dragonfly'
 
       - name: Authenticate to AWS
+        if: github.repository_owner == 'dragonflydb' && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
         uses: aws-actions/configure-aws-credentials@v5
         with:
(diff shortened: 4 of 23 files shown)


src/server/search/doc_index.cc
```diff
@@ -855,6 +856,16 @@ vector<SearchDocData> ShardDocIndex::SearchForAggregator(
...
 
+  // Build distance lookup for VECTOR_RANGE queries so YIELD_DISTANCE_AS alias is
+  // available as a field in the aggregation pipeline (GROUPBY/REDUCE etc).
+  absl::flat_hash_map<DocId, float> knn_score_map;
+  std::string score_alias;
+  if (auto* vr = search_algo->GetVectorRangeNode(); vr && !vr->score_alias.empty()) {
+    score_alias = vr->score_alias;
+    for (auto& [doc_id, dist] : search_results.knn_scores)
+      knn_score_map[doc_id] = dist;
+  }
```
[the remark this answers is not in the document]
vyavdoshenko: reserve was added

romange:
@dranikpg  ptal

src/server/search/doc_index.cc
```diff
@@ -855,6 +856,17 @@ vector<SearchDocData> ShardDocIndex::SearchForAggregator(
...
+  // Build distance lookup for VECTOR_RANGE queries so YIELD_DISTANCE_AS alias is
+  // available as a field in the aggregation pipeline (GROUPBY/REDUCE etc).
+  absl::flat_hash_map<DocId, float> knn_score_map;
+  std::string score_alias;
+  if (auto* vr = search_algo->GetVectorRangeNode(); vr && !vr->score_alias.empty()) {
+    score_alias = vr->score_alias;
+    knn_score_map.reserve(search_results.knn_scores.size());
+    for (auto& [doc_id, dist] : search_results.knn_scores)
+      knn_score_map[doc_id] = dist;
+  }
```
dranikpg: I don't understand this part, don't we call unconditionally `LoadHnswRangeDocsForAggregator` is we have a vector range query?

src/server/search/doc_index.cc
```diff
@@ -873,11 +885,50 @@ vector<SearchDocData> ShardDocIndex::SearchForAggregator(
...
+    SearchDocData loaded = accessor->Serialize(base_->schema, fields_to_load);
+    out.emplace_back(make_move_iterator(extracted_sort_indicies.begin()),
+                     make_move_iterator(extracted_sort_indicies.end()));
+    out.back().insert(make_move_iterator(loaded.begin()), make_move_iterator(loaded.end()));
+
+    if (!score_alias.empty())
+      out.back()[string{score_alias}] = static_cast<double>(dist);
+  }
+  return out;
+}
```
dranikpg, line 919: Maybe we can unite it somehow with the function above? They differ only in the source of ids. 

doc_index and search_family grew to multiple thousands lines of code, it seems to be getting really difficult to maintain with allt he edge cases

src/server/search/doc_index.cc
vyavdoshenko: LoadHnswRangeDocsForAggregator is only called for HNSW VECTOR_RANGE. The knn_score_map code in SearchForAggregator handles the FLAT path.

In CmdFtAggregate, we check GlobalHnswIndexRegistry::Exist() to decide the path:

- field is in the global HNSW registry -> LoadHnswRangeDocsForAggregator
- otherwise (FLAT) -> SearchForAggregator (which runs the per-shard search and gets distances from search_results.knn_scores)

FLAT vector indices live in per-shard FieldIndices and are searched inline via SearchAlgorithm::Search, which populates knn_scores_ in SearchVectorRangeFlat. HNSW indices live in GlobalHnswIndexRegistry and require a separate RangeQuery call outside the shard loop.

src/server/search/doc_index.cc
vyavdoshenko: fixed

PR #6943, opened 2026-03-22
chore(tiering): Move external storage metadata to fragment object

mkaruza:
Replace CompactObj::ExternalPtr inline union member with a pointer to a heap-allocated tiering::Fragment. Fragment now owns all external storage metadata (disk offset, serialized size, representation, cool record, Huffman first byte) that was previously packed into the 16-byte ExternalPtr struct.

1. Rename FragmentRef to Fragment
2. Replace ExternalPtr union member with tiering::Fragment* pointer
3. Store Fragments in TieredStorage::tiered_fragments_ (node_hash_map) keyed by monotonic IDs for pointer stability
4. Track pending stashes via EntryMap<size_t> keyed by (dbid, key) to be safe against DashTable entry relocations

### DIFF
--- a/src/server/tiered_storage.h
+++ b/src/server/tiered_storage.h
@@ -3,7 +3,7 @@
 //
 #pragma once
 
-#include <absl/container/flat_hash_map.h>
+#include <absl/container/node_hash_map.h>
 
 #include <boost/intrusive/list.hpp>
 #include <memory>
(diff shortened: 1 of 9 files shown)


src/server/tiered_storage.h
```diff
@@ -122,14 +124,20 @@ class TieredStorage : public TieredStorageBase {
     return stats_.cool_memory_used;
   }
 
+  // Create a Fragment for the given value and register it as pending under (dbid, key).
+  // Returns a stable pointer (node_hash_map guarantees pointer stability).
+  tiering::Fragment* AddFragment(DbIndex dbid, std::string_view key,
```
romange, line 129: I really do not understand why it is better. Before: 0 bytes to hold in-memory state. 
now: `O(N)` storage cost to hold the state. why?

src/server/tiered_storage.h
```diff
@@ -145,6 +153,18 @@ class TieredStorage : public TieredStorageBase {
   using CoolQueue = ::boost::intrusive::list<tiering::TieredCoolRecord>;
   CoolQueue cool_queue_;
 
+  // Fragment container — keyed by monotonic id, pointer-stable.
+  using TieredFragments = absl::node_hash_map<size_t, tiering::Fragment>;
+  size_t next_fragment_id_ = 0;
+  TieredFragments tiered_fragments_;
```
romange, line 159: It's a no go, imho.

mkaruza:
After discussion with @romange - implementation is not need. Closing PR.

PR #6944, opened 2026-03-22
feat(tiering): Add experimental list node offloading

mkaruza:
Extend tiered storage to offload individual QList nodes. Offloading is supported 
for RAW node objects.  Functionality is controlled  with flag
--tiered_experimental_list_support.

Key changes:
- Add ListNodeId (dbid, ql*, node*) as a new tiering identity type
  alongside KeyRef; introduce PendingId/ReadId variants so OpManager
  can handle both key-level and node-level I/O uniformly.
- Add tiering callbacks (offload_cb, onload_cb, delete_cb) to
  QList::TieringParams.  PromoteToQLIfNeeded now wires them up when
  tiered storage is available, using the new db_id_ member on
  ListWrapper to store DbIndex for usage.
- Update CancelStash / ReadInternal signatures to accept the generic
  PendingId / ReadId types instead of (dbid, key).
- Expose StashListNode / ReadTieredListNode free functions used by the
  ListWrapper callbacks.
- Add QList::Materialize that is used during RDB save to fetch back data in memory 
  from tiered storage.
- Added ListNodeTieringTest suite. Replication test for tiered list objects.

### DIFF
--- a/src/core/qlist.h
+++ b/src/core/qlist.h
@@ -55,10 +55,10 @@ class QList {
     Node* prev;
     Node* next;
 
-    union U {
+    union {
       unsigned char* entry;  // Pointer to the memory value of the node.
       size_t ext_offset;     // Offset in tiered storage
-    } u_;
+    };
 
     // For offloaded nodes, we store the offset of the value in colder storage
     // instead of a pointer to the data in memory.
@@ -90,7 +90,7 @@ class QList {
 
     void SetExternal(size_t offset, uint32_t sz);
     std::pair<size_t, size_t> GetExternalSlice() const {
-      return std::make_pair(size_t(u_.ext_offset), size_t(ext_size));
+      return std::make_pair(size_t(ext_offset), size_t(ext_size));
     }
   };
 
--- a/src/core/tiering_types.cc
+++ b/src/core/tiering_types.cc
@@ -62,10 +62,10 @@ auto FragmentRef::GetDescr(const CompactValue* pv) -> SerializationDescr {
 }
 
 auto FragmentRef::GetDescr(const QList::Node* node) -> SerializationDescr {
-  if (!node->u_.entry || node->sz == 0)
+  if (!node->entry || node->sz == 0)
     return {};
 
-  std::string_view entry_view{reinterpret_cast<const char*>(node->u_.entry), node->GetEntrySize()};
+  std::string_view entry_view{reinterpret_cast<const char*>(node->entry), node->GetEntrySize()};
   return {std::array<std::string_view, 2>{entry_view, {}}, CompactObj::ExternalRep::LIST_NODE};
 }
 
--- a/src/server/list_family.cc
+++ b/src/server/list_family.cc
@@ -104,8 +104,8 @@ auto NodeOnloadCb(TieredStorage* ts, DbIndex db_id, QList* ql) {
     // We are going to overwrite node external offset so first call delete.
     ts->Delete(db_id, node);
 
-    node->u_.entry = static_cast<unsigned char*>(zmalloc(res->size()));
-    memcpy(node->u_.entry, res->data(), res->size());
+    node->entry = static_cast<unsigned char*>(zmalloc(res->size()));
+    memcpy(node->entry, res->data(), res->size());
   };
 }
 
--- a/src/core/qlist.cc
+++ b/src/core/qlist.cc
@@ -179,7 +179,7 @@ bool NodeAllowMerge(const QList::Node* a, const QList::Node* b, const int fill) 
 // the owner over entry is passed to the node.
 QList::Node* CreateRAW(int container, uint8_t* entry, size_t sz) {
   QList::Node* node = (QList::Node*)zmalloc(sizeof(*node));
-  node->u_.entry = entry;
+  node->entry = entry;
   node->count = 1;
   node->sz = sz;
   node->next = node->prev = NULL;
@@ -224,8 +224,8 @@ QList::Node* CreateFromSV(int container, string_view value) {
 
 // Returns the relative increase in size.
 inline ssize_t NodeSetEntry(QList::Node* node, uint8_t* entry) {
-  node->u_.entry = entry;
-  size_t new_sz = lpBytes(node->u_.entry);
+  node->entry = entry;
+  size_t new_sz = lpBytes(node->entry);
   ssize_t diff = new_sz - node->sz;
   node->sz = new_sz;
   return diff;
@@ -235,7 +235,7 @@ inline ssize_t NodeSetEntry(QList::Node* node, uint8_t* entry) {
  * 'sz' is byte length of 'compressed' field.
  * 'compressed' is LZF data with total (compressed) length 'sz'
  * NOTE: uncompressed length is stored in quicklistNode->sz.
- * When quicklistNode->u_.entry is compressed, node->u_.entry points to a quicklistLZF */
+ * When quicklistnode->entry is compressed, node->entry points to a quicklistLZF */
 using quicklistLZF = struct quicklistLZF {
   size_t sz; /* LZF size in bytes*/
   char compressed[];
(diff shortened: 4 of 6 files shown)


mkaruza:
augment review

dranikpg:
wow this is a lot of code 😟

src/server/tiered_storage.h
```diff
@@ -301,6 +326,12 @@ inline void ReadTiered(DbIndex dbid, std::string_view key, const PrimeValue& val
...
 
+inline void ReadTieredListNode(DbIndex dbid, QList::Node* node, QList* ql,
+                               const tiering::DiskSegment& segment,
+                               std::function<void(io::Result<std::string_view>)> readf,
+                               TieredStorage* ts) {
+}
+
 inline TieredStorage::TResult<std::string> ReadTieredString(DbIndex dbid, std::string_view key,
                                                             const PrimeValue& value,
                                                             TieredStorage* ts) {
```
dranikpg, line 337: I remeber this want introduced to keep compatibility, but this can be part of list_family - my only other question would be replication then

src/server/tiered_storage.cc
```diff
@@ -448,26 +521,35 @@ void TieredStorage::Delete(DbIndex dbid, FragmentRef fragment_ref) {
...
+    } else if (auto bin = bins_->Delete(key->first, key->second); bin) {
+      op_manager_->CancelPending(*bin);
+    }
+  } else if (auto* key = std::get_if<tiering::ListNodeId>(&id); key) {
+    QList* ql = reinterpret_cast<QList*>(std::get<2>(*key));
+    ql->DecreaseNumOffloadedNodes();
+    op_manager_->CancelPending(id);
   }
-  fragment_ref.ClearStashPending();
+  fragment_ref.SetStashPending(false);
```
dranikpg: it seems like the logic shouldn't split on that and this should be handled by fragment ref internally

src/server/tiered_storage.cc
```diff
@@ -448,26 +521,35 @@ void TieredStorage::Delete(DbIndex dbid, FragmentRef fragment_ref) {
     DCHECK_EQ(hot.ObjType(), OBJ_STRING);
   }
   fragment_ref.ClearOffloaded();
-  op_manager_->DeleteOffloaded(dbid, segment);
+  // If fragment is not using small bins, we can directly mark the segment as free.
+  if (IsFragmentInSmallBins(fragment_ref)) {
+    op_manager_->MarkSegmentFree(dbid, segment);
+  } else {
+    op_manager_->DeleteOffloaded(dbid, segment);
+  }
```
dranikpg: But it already worked like this before, just via a longer call chain DeleteOffloaded -> NotifyDeleted returns true -> we call MarkAsFree. Is there any other reason other than optimization?

dranikpg: it just makes it more complicated

src/server/tiering/op_manager.h
```diff
@@ -66,6 +67,10 @@ class OpManager {
     return storage_.PrepareStash(length);
   }
 
+  void MarkSegmentFree(DiskSegment segment) {
+    storage_.MarkAsFree(segment.ContainingPages());
+  }
```
dranikpg: See my comment on how to keep the interface simpler

src/server/list_family.cc
```diff
@@ -90,16 +93,59 @@ class ListWrapper {
...
+                util::fb2::Future<io::Result<std::string>> node_entry;
+                ReadTieredListNode(
+                    db_id, node, ql, node->GetExternalSlice(),
+                    [node_entry](io::Result<std::string_view> res) mutable {
+                      node_entry.Resolve(
+                          res.transform([](std::string_view sv) { return std::string{sv}; }));
+                    },
+                    ts);
+
+                auto res = node_entry.Get();
```
dranikpg: just a style comment, you can make functions that return auto type and return a functor with state captured, this way you don't need this large differnetly indented block

romange:
Please rebase and resolve conflicts.

romange:
augment review

src/server/list_family.cc
```diff
@@ -90,16 +93,59 @@ class ListWrapper {
...
-          QList::TieringParams{.node_depth_threshold = GetFlag(FLAGS_list_tiering_threshold)});
+
+    if (GetFlag(FLAGS_list_tiering_threshold) > 0 && EngineShard::tlocal()->tiered_storage()) {
+      TieredStorage* ts = EngineShard::tlocal()->tiered_storage();
+      QList::TieringParams params{
+
+          .node_depth_threshold = GetFlag(FLAGS_list_tiering_threshold),
+
+          .offload_cb =
+              [ts, db_id = db_id_, ql](QList::Node* node) {
```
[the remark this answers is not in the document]
romange: while this is true, a TODO is enough, no need to fix it. Maybe it also breaks string tiering.

src/core/qlist.h
```diff
@@ -61,16 +62,30 @@ class QList {
...
     bool IsCompressed() const {
       return encoding != QUICKLIST_NODE_ENCODING_RAW;
     }
 
+    // Returns the size of entry data.
+    size_t GetEntrySize() const;
+
     size_t GetLZF(void** data) const;
+
+    struct __attribute__((__packed__)) ExternalRecord {
```
romange: you do not need this. You can repurpose  existing fields: `sz` for size, and have union of `entry` and `offset` for example.

src/server/tiering/common.h
```diff
@@ -48,10 +48,17 @@ struct DiskSegment {
 };
 
 using KeyRef = std::pair<uint16_t /* DbIndex */, std::string_view>;
+using ListNodeId = std::tuple<uint16_t /* DbIndex */, void* /* QList::Node */, void* /* QList */>;
```
romange: nit: make it hierarchical: DbIndex, QList*, QList::Node *

src/core/qlist.h
mkaruza: Modified.

mkaruza:
augment review

src/core/tiering_types.cc
```diff
@@ -27,12 +61,31 @@ auto FragmentRef::GetDescr(const CompactValue* pv) -> SerializationDescr {
...
+    return {};
+
+  std::string_view entry_view{reinterpret_cast<const char*>(node->u_.entry), node->GetEntrySize()};
+  return {std::array<std::string_view, 2>{entry_view, {}}, CompactObj::ExternalRep::LIST_NODE};
+}
+
+auto FragmentRef::GetSerializationDescr() const -> SerializationDescr {
+  return std::visit(Overloaded{[](CompactValue* pv) { return GetDescr(pv); },
+                               [](QList::Node* node) { return GetDescr(node); }},
+                    val_);
```
dranikpg: `visit([](auto* v) {return GetDescr(v); }, val_)` is enough

src/server/list_family.cc
dranikpg: 🙂🙂🙂

mkaruza: Sorry i missed this, comment is good suggestion.

dranikpg (dismissed):

src/core/qlist.h
```diff
@@ -53,7 +54,14 @@ class QList {
   struct Node {
     Node* prev;
     Node* next;
-    unsigned char* entry;
+
+    union U {
+      unsigned char* entry;  // Pointer to the memory value of the node.
+      size_t ext_offset;     // Offset in tiered storage
+    } u_;
```
romange: tip: to reduce the size of this change you can create an anonymous union:
```
union U {
      unsigned char* entry;  // Pointer to the memory value of the node.
      size_t ext_offset;     // Offset in tiered storage
    };
```

mkaruza: Good point. Changed.

romange:
> It's really hard to review such a large PR as if its "ready to be commited" code. I know it's very cumbersome to split all of this. I usually make different commits and there are lots of refactoring parts here (uniting db_index, key), extending ListWrapper, etc.
> 
> I will take a look a little later at the other parts. I'm not much in context of list tiering yet

I join the sentiment. There is no reason to review it as a whole. I should be split into 3 at least. Lets start with `src/core/` changes, first, then tiering and only then - families.

mkaruza:
@romange @dranikpg 

I have split this PR into 3:

#7021 extend QList::Node to support tiering.
#7022 extend TeieredStorage to handle QList::Node
#7023 Enables tiering of nodes and adds test

PR's needs to be merged in order. I'll close this once they are review/merged

mkaruza:
Closing: #7021 #7022 #7023 were merged

PR #6948, opened 2026-03-23
chore(server): Simplify SerializerBase interface

dranikpg:
Greatly simplifies SerializerBase interface

1. Simplify base class
  * Don't need to pass around snapshot_version if we can just store it
  * Simplify some of the functions, remove `BucketStateCountForTesting`
2. Unite OnChange bucker serailization and loop bucket serialization into a single ProcessBucket
  Those paths had more in common than expected - almost everything. This allows us to further to simplify the interface by hiding `MarkBucketSerializing` and other state control functions to be internal

### DIFF
--- a/src/server/serializer_base.cc
+++ b/src/server/serializer_base.cc
@@ -11,8 +11,7 @@
 
 namespace dfly {
 
-SerializerBase::SerializerBase(DbSlice* slice)
-    : db_slice_(slice), db_array_(slice ? slice->databases() : DbTableArray{}) {
+SerializerBase::SerializerBase(DbSlice* slice) : db_slice_(slice) {
   DCHECK(db_slice_);
 }
 
@@ -76,7 +75,7 @@ std::optional<BucketIdentity> SerializerBase::ShouldProcessBucket(PrimeTable::bu
     return std::nullopt;
   }
 
-  // Check if this bucker is currently being serialized
+  // Check if this bucket is currently being serialized
   if (bucket_states_.contains(it.bucket_address())) {
     ++stats_.change_during_serialization;
     return std::nullopt;
--- a/src/server/snapshot.h
+++ b/src/server/snapshot.h
@@ -99,8 +99,8 @@ class SliceSnapshot : public SerializerBase, public journal::JournalConsumerInte
   RdbSaver::SnapshotStats GetCurrentSnapshotProgress() const;
 
   // Journal listener
-  void ConsumeJournalChange(const journal::JournalChangeItem& item) override;
-  void ThrottleIfNeeded() override;
+  void ConsumeJournalChange(const journal::JournalChangeItem& item) final;
+  void ThrottleIfNeeded() final;
 
  private:
   [[maybe_unused]] void SerializeIndexMapping(


dranikpg:
augment review

dranikpg:
augment review

src/server/serializer_base.cc
```diff
@@ -18,45 +22,42 @@ SerializerBase::~SerializerBase() {
...
                                            std::vector<TieredDelayedEntry> delayed) {
   auto it = bucket_states_.find(bid);
-  DCHECK(it != bucket_states_.end() && it->second.phase == BucketPhase::kSerializing);
+  DCHECK(it != bucket_states_.end());
+  DCHECK(it->second.phase == BucketPhase::kSerializing);
 
   if (delayed.empty()) {
     // Serializing -> Covered
     bucket_states_.erase(it);
+    ++stats_.buckets_serialized;
```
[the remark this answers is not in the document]
dranikpg: Yes, we don't use that code now. It was created in "advance"

src/server/serializer_base.cc
```diff
@@ -5,11 +5,15 @@
...
 
 #include "base/logging.h"
+#include "server/common_types.h"
+#include "server/journal/journal.h"
+#include "server/synchronization.h"
 
 namespace dfly {
 
 SerializerBase::SerializerBase(DbSlice* slice)
     : db_slice_(slice), db_array_(slice ? slice->databases() : DbTableArray{}) {
```
romange: move db_array_ initialization under DCHECK.

src/server/serializer_base.cc
```diff
@@ -18,45 +22,42 @@ SerializerBase::~SerializerBase() {
...
 //   same fiber, so the baseline is serialized first. big_value_mu_ prevents this callback path
 //   from interleaving with the traversal fiber's bucket serialization, which may preempt while
 //   emitting large values.
-uint64_t SerializerBase::RegisterChangeListener() {
-  DCHECK(db_slice_);
-  db_array_ = db_slice_->databases();
-  auto cb = [this](DbIndex db_index, const DbSlice::ChangeReq& req) {
-    HandleChangeReq(db_index, req);
+void SerializerBase::RegisterChangeListener() {
+  db_array_ = db_slice_->databases();  // copy pointers to survive flush
```
romange, line 30: actually, maybe we do not need to initialize them in c'tor at all?

dranikpg: ah, wow, yes. This all comes from the fact that this class was made testable with a custom constructor

src/server/serializer_base.cc
```diff
@@ -68,47 +69,57 @@ void SerializerBase::CompleteBucketDelayed(BucketIdentity bid) {
...
+  // Check if bucket is invalid or was already serialized
   if (it.is_done() || it.GetVersion() >= snapshot_version_) {
     ++stats_.buckets_skipped;
-    return;
+    return std::nullopt;
   }
 
-  BucketIdentity bid = it.bucket_address();
-  if (bucket_states_.contains(bid)) {
+  // Check if this bucker is currently being serialized
```
BorysTheDev: Check if this bucket

src/server/snapshot.h
```diff
@@ -98,8 +99,8 @@ class SliceSnapshot : public SerializerBase, public journal::JournalConsumerInte
   RdbSaver::SnapshotStats GetCurrentSnapshotProgress() const;
 
   // Journal listener
-  void ConsumeJournalChange(const journal::JournalChangeItem& item);
-  void ThrottleIfNeeded();
+  void ConsumeJournalChange(const journal::JournalChangeItem& item) override;
```
BorysTheDev: maybe we need to start using `override final`?

dranikpg:
Implemented mentioned fixes

PR #6949, opened 2026-03-23
fix(search): cap GROUPBY nargs before reserve to prevent DoS crash

vyavdoshenko:
ParseAggregatorParams() passed the user-supplied GROUPBY nargs value directly to fields.reserve() without any bounds check.  An attacker could send a single FT.AGGREGATE command with GROUPBY 9999999999999 to trigger a ~320 TiB
allocation.  On memory-constrained systems (containers with RLIMIT_AS, Docker --memory, Kubernetes resources.limits) mimalloc calls abort() instead of throwing std::bad_alloc, crashing the server.

Fix: cap num_fields to parser->Tail().size() before calling reserve(), so the allocation is bounded by the number of tokens actually remaining in the command.

### DIFF
--- a/src/server/search/search_family.cc
+++ b/src/server/search/search_family.cc
@@ -524,7 +524,7 @@ ParseResult<AggregateParams::JoinParams> ParseAggregatorJoinParams(
   known_indexes->insert(join_params.index_alias);
 
   size_t num_fields = parser->Next<size_t>();
-  join_params.conditions.reserve(num_fields);
+  join_params.conditions.reserve(std::min(num_fields, parser->Tail().size()));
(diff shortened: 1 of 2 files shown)


romange:
pls resolve

dranikpg:
There are at least 99999 more ways to cause such behaviour, but yes, I think for potential typos or other stuff there should be a safety net around this

vyavdoshenko:
> There are at least 99999 more ways to cause such behaviour, but yes, I think for potential typos or other stuff there should be a safety net around this

This is not only a typo, but it is also a pain because of fuzzy testing.

PR #6950, opened 2026-03-23
feat(topk): add TOPK command family

glevkovich:
- Add RedisBloom-compatible TOPK commands and register the family in the main service. Include command tests and RDB save/load tests for the new type.

- Also make TOPK list ordering deterministic on equal counts and fix TOPK decay validation to run after reading the serialized value.


src/server/topk_family.cc
```diff
@@ -0,0 +1,440 @@
...
+    width = parser.Next<uint32_t>();
+    depth = parser.Next<uint32_t>();
+    decay = parser.Next<double>();
+    RETURN_ON_PARSE_ERROR(parser, rb);
+
+    if (width == 0 || depth == 0) {
+      return rb->SendError("width and depth must be greater than 0");
+    }
+
+    if (decay < 0.0 || decay > 1.0) {
```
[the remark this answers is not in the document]
glevkovich: fixed

src/server/topk_family.cc
```diff
@@ -0,0 +1,440 @@
...
+    return cmd_cntx->SendError(kKeyNotFoundErr);
+  }
+  if (result.status() == OpStatus::WRONG_TYPE) {
+    return cmd_cntx->SendError(kWrongTypeErr);
+  }
+
+  // Build array response: [k, <k>, width, <width>, depth, <depth>, decay, <decay>]
+  {
+    SinkReplyBuilder::ReplyScope scope(rb);
+    rb->StartArray(8);
```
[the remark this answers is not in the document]
glevkovich: this is incorrect. should return flat array.

src/server/topk_family.cc
```diff
@@ -0,0 +1,440 @@
...
+#define HFUNC(x) SetHandler(&TopkFamily::x)
+
+void RegisterTopkFamily(CommandRegistry* registry) {
+  registry->StartFamily(acl::BLOOM);
+  *registry << CI{"TOPK.RESERVE", CO::JOURNALED | CO::DENYOOM | CO::FAST, -3, 1, 1}.HFUNC(Reserve)
+            << CI{"TOPK.ADD", CO::JOURNALED | CO::DENYOOM | CO::FAST, -3, 1, 1}.HFUNC(Add)
+            << CI{"TOPK.INCRBY", CO::JOURNALED | CO::DENYOOM | CO::FAST, -4, 1, 1}.HFUNC(IncrBy)
+            << CI{"TOPK.QUERY", CO::READONLY | CO::FAST, -3, 1, 1}.HFUNC(Query)
+            << CI{"TOPK.COUNT", CO::READONLY | CO::FAST, -3, 1, 1}.HFUNC(Count)
+            << CI{"TOPK.LIST", CO::READONLY | CO::FAST, -2, 1, 1}.HFUNC(List)
```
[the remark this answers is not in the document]
glevkovich: fixed

src/server/topk_family.cc
```diff
@@ -0,0 +1,440 @@
...
+    return OpAdd(t->GetOpArgs(shard), key, items);
+  };
+
+  auto result = cmd_cntx->tx()->ScheduleSingleHopT(std::move(cb));
+  if (result.status() == OpStatus::KEY_NOTFOUND) {
+    return cmd_cntx->SendError(kKeyNotFoundErr);
+  }
+  if (result.status() == OpStatus::WRONG_TYPE) {
+    return cmd_cntx->SendError(kWrongTypeErr);
+  }
```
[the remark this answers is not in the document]
glevkovich: fixed

src/server/topk_family.cc
```diff
@@ -0,0 +1,429 @@
...
+    rb->SendBulkString("decay");
+    rb->SendDouble(result->decay);
+  }
+}
+
+using CI = CommandId;
+#define HFUNC(x) SetHandler(&TopkFamily::x)
+
+void RegisterTopkFamily(CommandRegistry* registry) {
+  registry->StartFamily(acl::BLOOM);
```
vyavdoshenko: TOPK commands are registered under `acl::BLOOM`, but TOPK is a different probabilistic data structure (Heavy Hitters / HeavyKeeper). This means `COMMAND DOCS`, ACL categories, and audit logs will incorrectly report all TOPK commands as Bloom filter operations.

`bloom_family.cc` and `cms_family.cc` already have separate ACL families. Please add `acl::TOPK` (following the same pattern) or, at a minimum, use `acl::CMS`, which is semantically closer.

src/server/topk_family.cc
```diff
@@ -0,0 +1,429 @@
...
+}  // namespace
+
+void TopkFamily::Reserve(CmdArgList args, CommandContext* cmd_cntx) {
+  CmdArgParser parser(args);
+  string_view key = parser.Next();
+  uint32_t k = parser.Next<uint32_t>();
+  auto* rb = static_cast<RedisReplyBuilder*>(cmd_cntx->rb());
+  RETURN_ON_PARSE_ERROR(parser, rb);
+
+  if (k == 0) {
```
vyavdoshenko, line 180: There is no upper bound on `k`. `TOPK.RESERVE key 4294967295 1 1 0.9` will reach `min_heap_.reserve(k_)` in the TOPK constructor (topk.cc:48), which attempts to allocate ~128 GB (4,294,967,295 × sizeof(HeapItem) ≈ 128 GB on x86-64).

The `CO::DENYOOM` flag is evaluated before the transaction callback runs, so it does not protect against this: by the time `SetTOPK()` constructs the object, the DENYOOM check has already passed. An uncaught `std::bad_alloc` here would crash the server fiber.

Consider how the competitors do it and add an explicit upper bound, e.g.:
```
  if (k > 50'000)
    return rb->SendError("k exceeds maximum allowed value");
```

src/server/topk_family.cc
```diff
@@ -0,0 +1,429 @@
...
+      return rb->SendError("decay must be between 0 and 1");
+    }
+  }
+
+  if (parser.HasNext()) {
+    return rb->SendError(kSyntaxErr);
+  }
+
+  // Guard against overflow: width * depth * sizeof(uint32_t) must fit in size_t.
+  if (width > SIZE_MAX / sizeof(uint32_t) / depth) {
```
vyavdoshenko, line 223: The overflow check guards against `width * depth * 4 > SIZE_MAX`, which is correct. However, it allows values that are technically within SIZE_MAX but still cause enormous allocations. For example:

  width = 4,294,967,295, depth = 1
  -> passes the check (4.29B < SIZE_MAX/4)
  -> allocates width * depth * 4 ≈ 16 GB for the counter array alone

Similar to the `k` issue: this happens inside the shard callback after the DENYOOM check, so maxmemory enforcement does not help.

Please add independent caps on `width` and `depth` (e.g., both <= 8192, matching common HeavyKeeper implementations), in addition to the product overflow check.

src/server/topk_family.cc
glevkovich: Thanks , fixed.

src/server/topk_family.h
```diff
@@ -0,0 +1,24 @@
...
+  static void Reserve(CmdArgList args, CommandContext* cmd_cntx);
+  static void Add(CmdArgList args, CommandContext* cmd_cntx);
+  static void IncrBy(CmdArgList args, CommandContext* cmd_cntx);
+  static void Query(CmdArgList args, CommandContext* cmd_cntx);
+  static void Count(CmdArgList args, CommandContext* cmd_cntx);
+  static void List(CmdArgList args, CommandContext* cmd_cntx);
+  static void Info(CmdArgList args, CommandContext* cmd_cntx);
+};
+
+}  // namespace dfly
```
dranikpg: We decided to ditch such a declarative header approach (that we had for historic reasons) because it literally has no real function. No state, no scope, no use from outside, etc

We have now a single command_families.h header where a family registers itself. I see you exetended it.

dranikpg: It will also be less handy with coroutines as we need to import the types and maybe wrap them in the future

src/server/topk_family.cc
glevkovich: Thank for the catch, fixed

src/server/topk_family.cc
glevkovich: fixed

src/server/topk_family.h
glevkovich: That makes sense, I'm goinf to delete this header.
Thanks.

Issue #6947, opened 2026-03-23
error during teardown in `TestSearchFTDropindexCME.test_ft_dropindex_fails_on_replica_cme`


abhijat:
https://github.com/dragonflydb/dragonfly/actions/runs/23420660301/job/68124834667#step:7:1931
build (ubuntu-dev:24, Epoll, Debug, self-hosted, linux, ARM64)
timeout during dragonfly process shutdown

fiber stack trace in comments

abhijat:
trace of fibers from test log
```
I20260323 04:24:19.387625 23696 scheduler.cc:534] ------------ Fiber main (suspended:120910ms) ------------
0x1843fe8  std::function<>::operator()()
0x1842948  util::fb2::detail::FiberInterface::ExecuteOnFiberStack()::{lambda()#1}::operator()()
0x1843c80  boost::context::detail::fiber_ontop<>()
0x18435a4  boost::context::fiber::resume_with<>()
0x1842160  util::fb2::detail::FiberInterface::SwitchTo()
0x183598c  util::fb2::detail::Scheduler::Preempt()
0x439d4c  util::fb2::detail::FiberInterface::Suspend()
0x1819ae8  util::fb2::ProactorBase::Run()
0x18a2c6c  util::ProactorPool::SetupProactors()::{lambda()#1}::operator()()
0x18a3fb4  std::__invoke_impl<>()
I20260323 04:24:19.434561 23696 scheduler.cc:534] ------------ Fiber _dispatch (ready:46ms) ------------
0x1843fe8  std::function<>::operator()()
0x1842948  util::fb2::detail::FiberInterface::ExecuteOnFiberStack()::{lambda()#1}::operator()()
0x1843c80  boost::context::detail::fiber_ontop<>()
0x18435a4  boost::context::fiber::resume_with<>()
0x1842160  util::fb2::detail::FiberInterface::SwitchTo()
0x1839958  util::fb2::detail::Scheduler::RunReadyFibersInternal()
0x1839378  util::fb2::detail::Scheduler::Run()
0x18e7260  util::fb2::EpollProactor::MainLoop()
0x181c094  util::fb2::ProactorDispatcher::Run()
0x183402c  util::fb2::detail::(anonymous namespace)::DispatcherImpl::Run()
0x1833b5c  util::fb2::detail::(anonymous namespace)::DispatcherImpl::DispatcherImpl()::{lambda()#1}::operator()()
[...]

PR #6956, opened 2026-03-23
fix: ft.dropindex deadlock

BorysTheDev:
fixes: #6947

Summary: Fixes a potential deadlock in FT.DROPINDEX when dropping an index that has an active vector builder fiber.

Changes:

- Defers per-shard ShardDocIndex destruction until after the transaction finishes, then destroys on shard proactor threads to avoid joining from the shard FiberQueue.
- Documents the constraint that IndexBuilder::Cancel() must not be called from the shard FiberQueue when vector work may be queued there.

We still have a problem with other commands that are interchanged with index building.

### DIFF
--- a/src/server/search/index_builder.h
+++ b/src/server/search/index_builder.h
@@ -26,8 +26,9 @@ struct IndexBuilder {
   // for HNSW indices (restored from RDB). This flag is passed from PerformPostLoad.
   void Start(const OpArgs& op_args, bool is_restored, std::function<void()> on_complete);
 
-  // Cancel building and wait for worker to finish. Safe to delete after
-  // TODO: Maybe implement nonblocking version?
+  // Cancel building and wait for worker to finish. Safe to delete after.
+  // WARNING: Must NOT be called from the shard's FiberQueue context if VectorLoop
+  // may have dispatched work to the same queue — that would deadlock.
   void Cancel();
 
   // Get fiber reference. Temporary to polyfill sync construction places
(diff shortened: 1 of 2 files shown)


romange:
@BorysTheDev  can you explain what are the problems we may have? what is the potential scenario?

BorysTheDev:
@romange the problem is described in the ticket https://github.com/dragonflydb/dragonfly/issues/6947#issuecomment-4110255590 
The current fix only fixes the deadlock with ft.dropindex. Index builder uses Fiber queue to build index, but we use this queue to execute any other commands and it is a problem, because we are blocked until we finish the index. Also, the expiration of the keys can be a problem because we do it in a separate fiber, but I need to investigate it more. Also if we decide to support more than 1 vector field in the document, more problems can appear.

romange:
ok, in that case why not call `RunBlockingInParallel` inside `VectorLoop`? In other words, if VectorLoop causes cyclic dependencies, it can run in an its own fiber.

BorysTheDev:
> ok, in that case why not call `RunBlockingInParallel` inside `VectorLoop`? In other words, if VectorLoop causes cyclic dependencies, it can run in an its own fiber.

In this case we don't control data presence and changes

romange:
I do not understand what does it mean. can you provide an example?

BorysTheDev:
> I do not understand what does it mean. can you provide an example?

if we want data consistency, we need to implement a similar replication mechanism, because if we iterate without controlling key changes, we can't guarantee that we insert all keys or we can have documents that were dropped

Issue #6960, opened 2026-03-24
Background vector indexing correctness


dranikpg:
1. Previosuly all indexing was synchronous under a transaction lock.

```c++
void FtCreate() {
  Hop({
    for (key : table->traverse())
      Add(key)
  })
}
```

2. I introduced IndexBuilder https://github.com/dragonflydb/dragonfly/pull/6504 that allows indexing in the background by iterating with Traverse in a separate fiber. It doesn't use transactional locking and relies on the fact that adding to the _classic_ index is a non-suspendable operation

```c++

Fiber({
  for (key : table->traverse()) {
    Add(key)
    ThisFiber::Sleep(x ms); // keeps dragonfly operational even when building a many gigabyte index
  }
);
```

3. However, access to a global vector index requires locking a mutex, breaking the assumption of (2).
  Currently, the global vector index is a regular std::mutex suspending full thread execution, making this operation safe. 

This issue prevents us from adopting a fiber-aware mutex for global hnsw indices

Suggested workarounds:
1. Use a "quiescence" concept where a given "period" is safe for performing modifications, during which indexing can't run. An instance will juggle between those two modes

PR #6962, opened 2026-03-24
fix(search): Simplify hnsw indexing

dranikpg:
```
  // Because order of acquiring mutexes for global vector indices is not determined, we must run
  // all accesses on a single thread through the shard queue to have a single linear order
```

This comment is wrong because the shard queue is not the only place where commands execute - we have L2 queue and inlined execution.

Nonetheless, this safeguard is not needed now because global hnsw index uses thread based mutex that suspends the whole thread, being actually "invisible" to something living in a fiber based world


dranikpg:
For the record #6960

Issue #6975, opened 2026-03-25
SIGSEGV in JournalXReadGroupIfNeeded


jnugh:
**Describe the bug**
After upgrading to v1.37.0 we get regular Segfaults on our Dragonfly cluster (managed by the Kubernetes operator). 

```
I20260325 08:31:13.927850     7 init.cc:127] dragonfly running in opt mode.
I20260325 08:31:13.927994     7 dfly_main.cc:924] Starting dragonfly df-v1.37.0-aa35ea84bfa6af5ad046d2a994bd040ad61fb3a9
                   .--::--.                   
   :+*=:          =@@@@@@@@=          :+*+:   
  %@@@@@@%*=.     =@@@@@@@@-     .=*%@@@@@@#  
  @@@@@@@@@@@@#+-. .%@@@@#. .-+#@@@@@@@@@@@%  
  -@@@@@@@@@@@@@@@@*:#@@#:*@@@@@@@@@@@@@@@@-  
    :+*********####-%@%%@%-####********++.    
   .%@@@@@@@@@@@@@%:@@@@@@:@@@@@@@@@@@@@@%    
   .@@@@@@@@%*+-:   =@@@@=  .:-+*%@@@@@@@%.   
     =*+-:           ###*          .:-+*=     
                     %@@%                     
                     *@@*                     
                     +@@=                     
                     :##:                     
                     :@@:                     
                      @@                      
                      ..                      
* Logs will be written to the first available of the following paths:
/tmp/dragonfly.*
./dragonfly.*
* For the available flags type dragonfly [--help | --helpfull]
* Documentation can be found at: https://www.dragonflydb.io/docs
I20260325 08:31:13.928297     7 dfly_main.cc:969] maxmemory has not been specified. Deciding myself....
I20260325 08:31:13.928318     7 dfly_main.cc:978] Found 1.00GiB available memory. Setting maxmemory to 819.20MiB
I20260325 08:31:13.929637    12 uring_proactor.cc:256] IORing with 1024 entries, allocated 98368 bytes, cq_entries is 2048
I20260325 08:31:13.929903     7 proactor_pool.cc:149] Running 1 io threads
I20260325 08:31:13.931135     7 dragonfly_listener.cc:150] SSL version: OpenSSL 3.0.17 1 Jul 2025
I20260325 08:31:13.931532     7 dfly_main.cc:327] Listening on admin socket any:9999
[...]

PR #6980, opened 2026-03-25
fix: null ptr deref in JournalXReadGroupIfNeeded for XREAD BLOCK with replication

romange:
## Summary

- `JournalXReadGroupIfNeeded` crashed with SIGSEGV when `XREAD BLOCK` unblocked while replication was active
- `sitem.group` is always `nullptr` for plain XREAD (only set for XREADGROUP), but the function dereferenced it unconditionally via `sitem.group->entries_read`
- Fix: add `|| !opts.read_group || records.empty()` to the early-return guard; remove now-unnecessary `journal_xgroup` lambda

## Root Cause

In `XReadBlock`'s `range_cb`, `JournalXReadGroupIfNeeded` was called without checking `opts->read_group`. With a replica connected (journal non-null), the function would proceed into the lambda and dereference a null `group` pointer.

Crash observed in production at: `JournalXReadGroupIfNeeded()::{lambda()#1}::operator<>()`

## Changes

- `src/server/stream_family.cc`: guard + refactor `JournalXReadGroupIfNeeded`
- `tests/dragonfly/replication_test.py`: regression test `test_xread_block_replication_crash`

Fixes https://github.com/dragonflydb/dragonfly/issues/6975

### DIFF
--- a/tests/dragonfly/replication_test.py
+++ b/tests/dragonfly/replication_test.py
@@ -4008,6 +4008,42 @@ async def test_xreadgroup_replication(df_factory):
 
 
 """
+Regression test for SIGSEGV when XREAD BLOCK unblocks with replication active.
+JournalXReadGroupIfNeeded accessed sitem.group->entries_read without checking that
+group is non-null (group is always null for XREAD, only set for XREADGROUP).
+"""
+
+
+@dfly_args({"proactor_threads": 2})
+async def test_xread_block_replication_crash_6975(df_factory):
+    master = df_factory.create()
+    replica = df_factory.create()
+
+    master.start()
+    replica.start()
+
+    c_master = master.client()
+    c_replica = replica.client()
+
+    await c_replica.execute_command(f"REPLICAOF localhost {master.port}")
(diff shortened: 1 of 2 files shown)


romange:
augment review

tests/dragonfly/replication_test.py
```diff
@@ -4007,6 +4007,42 @@ async def compare_group_info(stream_key, expected_pending, expected_entries_read
...
 
+"""
+Regression test for SIGSEGV when XREAD BLOCK unblocks with replication active.
+JournalXReadGroupIfNeeded accessed sitem.group->entries_read without checking that
+group is non-null (group is always null for XREAD, only set for XREADGROUP).
+"""
+
+
+@dfly_args({"proactor_threads": 2})
+async def test_xread_block_replication_crash_6975(df_factory):
```
dranikpg, line 4018: Instead of making one more test, maybe it could've been embedded into a stream replication test as a scenarion, if we have any of those (likely we don't)

romange: it's quite fast though

PR #6981, opened 2026-03-25
fix: JSON.NUMINCRBY negative result overflow

romange:
## Summary
Casting negative double results to `uint64_t` in `BinOpApply` caused `JSON.NUMINCRBY` to return huge positive values instead of negative numbers (e.g., `1 + (-2)` returned `18446744073709551615` instead of `-1`).

## Changes
- Cast integer arithmetic results to `int64_t` instead of `uint64_t`
- Add regression test for negative increment on integer JSON values


romange:
Fixed int overflow use-cases as well.

PR #6985, opened 2026-03-25
fix: proactively delete expired keys on replicas via replica_delete_expired flag

shahyash2609:
## Summary
- Replicas now proactively delete expired keys in `ExpireIfNeeded` instead of returning stale data
- Adds `--replica_delete_expired` flag (default: true) to control this behavior
- Expired keys on replicas correctly return nil/`-2` instead of stale values with TTL ~UINT64_MAX

## Problem
On replicas with `hz=0` (no background expiry sweep), `ExpireIfNeeded()` skipped expiry for replicas entirely — returning valid iterators for expired keys. This caused:
- **GET** returning stale values for expired keys
- **TTL** returning ~18446744073709550 (unsigned underflow from `UINT64_MAX`)
- Keys persisting indefinitely on replicas until master explicitly deleted them

## Fix
Replace the read-path hiding approach with actual deletion in `ExpireIfNeeded`:
- When `replica_delete_expired=true` (default), replicas proactively delete expired keys on access
- This correctly invalidates iterators so `InsertNew` always finds a clean slot (no replication stalls)
- `replica_delete_expired=false` preserves old behavior for consistency tests comparing master/replica data signatures

## Validated on VM (old vs new build)
**Old build** — 10/10 expired keys still visible on replica:
```
BUG! exptest-0 = value-0, TTL = 18446744073709550
...
BUG CONFIRMED: 10/10 expired keys still visible on replica!
```

**New build** — all expired keys correctly return nil:
```
OK: exptest-0 = nil, TTL = -2
...
FIX CONFIRMED: All expired keys correctly return nil on replica!
```

## Changes
- `src/server/db_slice.cc`: Add `replica_delete_expired` flag; when true, `ExpireIfNeeded` deletes expired keys on replicas
- `src/server/db_slice.h`: Declare the flag
- `tests/dragonfly/replication_test.py`: Add `test_expiry_on_replica` pytest
- `tests/dragonfly/instance.py`: Gate `replica_delete_expired` default behind version > 1.36 for old binary compat

### DIFF
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -46,6 +46,9 @@ ABSL_FLAG(std::string, notify_keyspace_events, "",
 
 ABSL_FLAG(bool, cluster_flush_decommit_memory, false, "Decommit memory after flushing slots");
 
+ABSL_FLAG(bool, replica_delete_expired, true,
+          "If true, replicas proactively delete expired keys on the read path.");
+
 namespace dfly {
 
 using namespace std;
@@ -1234,8 +1237,10 @@ PrimeIterator DbSlice::ExpireIfNeeded(const Context& cntx, PrimeIterator it) con
 
   int64_t expire_time = it->first.GetExpireTime();
 
-  // Never do expiration on replica or if expiration is disabled.
-  if (int64_t(cntx.time_now_ms) < expire_time || owner_->IsReplica() || !expire_allowed_) {
+  // Never do expiration if expiration is disabled, or on replicas unless replica_delete_expired
+  // is enabled (which allows replicas to proactively delete expired keys on the read path).
+  if (int64_t(cntx.time_now_ms) < expire_time || !expire_allowed_ ||
+      (owner_->IsReplica() && !absl::GetFlag(FLAGS_replica_delete_expired))) {
     return it;
   }
 
--- a/tests/dragonfly/instance.py
+++ b/tests/dragonfly/instance.py
@@ -425,6 +425,10 @@ class DflyInstanceFactory:
         vmod = "dragonfly_connection=1,db_slice=1,listener_interface=1,main_service=1,rdb_save=1,replica=1,cluster_family=1,engine_shard=1,dflycmd=1,snapshot=1,streamer=1"
         args.setdefault("vmodule", vmod)
         args.setdefault("jsonpathv2")
+        # Disable replica_delete_expired by default so consistency tests that compare master/replica
+        # data signatures are not broken by replicas proactively deleting expired keys.
+        if version > 1.37:
+            args.setdefault("replica_delete_expired", "false")
         if version > 1.27:
             args.setdefault("omit_basic_usage")
 
--- a/tests/dragonfly/replication_test.py
+++ b/tests/dragonfly/replication_test.py
@@ -863,6 +863,71 @@ async def test_expiry(df_factory: DflyInstanceFactory, n_keys=1000):
     assert all(v is None for v in res)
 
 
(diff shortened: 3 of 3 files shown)


romange:
Thanks, can you please provide a pytest (in replication_test.py) that reproduces this bug?

romange:
Thanks @shahyash2609 . Allow me few more days to review this PR.

src/server/db_slice.cc
```diff
@@ -605,6 +605,19 @@ auto DbSlice::FindInternal(const Context& cntx, string_view key, optional<unsign
       db.stats.events.misses += miss_weight;
       return OpStatus::KEY_NOTFOUND;
     }
+
+    // On replica, ExpireIfNeeded returns a valid iterator for expired keys (since replicas
```
romange: I understand there is an issue but I would like to suggest a differrent fix.
1. There is no point in keeping the data in memory and return "Not found" when looking up for it - it just does not make sense. So lets delete it in `ExpireIfNeeded` even for replica.
2. However, unfortunately, it will break our tests that rely on the fact they can capture data signatures on both master and replica to compare replication consistency. Therefore I suggest doing this under flag. So we will have a runtime flag, "replica_delete_expired" that can even be true by default. In our pytests  it should be false by default (grep for `args.setdefault` to see where we do that). Please keep the pytest you added, which of course should override the flag to "true" to demonstrate that replica deletes the entry. I expect the test to fail without the fix in C++ code.

shahyash2609: understood. Let me change the fix in that way!

shahyash2609:
> thank you @shahyash2609 . Did you run the test you wrote and saw it indeed fails on the current code?

Yes, I had. Though let me verify it once again once I do the suggested changes!

shahyash2609:
@romange, I have updated the PR by fixing the issue with the suggested approach. Please review the same.
Existing tests are passing along with the newly added one.
image
Thanks!

src/server/db_slice.cc
```diff
@@ -1234,8 +1237,10 @@ PrimeIterator DbSlice::ExpireIfNeeded(const Context& cntx, PrimeIterator it) con
 
   int64_t expire_time = it->first.GetExpireTime();
 
-  // Never do expiration on replica or if expiration is disabled.
-  if (int64_t(cntx.time_now_ms) < expire_time || owner_->IsReplica() || !expire_allowed_) {
+  // Never do expiration if expiration is disabled, or on replicas unless replica_delete_expired
+  // is enabled (which allows replicas to proactively delete expired keys on the read path).
+  if (int64_t(cntx.time_now_ms) < expire_time ||
+      (owner_->IsReplica() && !absl::GetFlag(FLAGS_replica_delete_expired)) || !expire_allowed_) {
```
romange:
```suggestion
  if (int64_t(cntx.time_now_ms) < expire_time ||  !expire_allowed_
      (owner_->IsReplica() && !absl::GetFlag(FLAGS_replica_delete_expired))) {
```

shahyash2609: Applied the suggestion!

shahyash2609:
@romange , whenever you get a chance, could you please approve this so we can resume the checks and tests?

romange (approved):

shahyash2609:
@romange all checks are passed! Can we merge this as I don't have the access!

romange:
@shahyash2609  your commits are not signed so can not be merged.  Please squash all the commits and sign the final commit with your gpg key.

shahyash2609:
@romange its done. Would need your approval to run the workflow again!

romange:
@shahyash2609 thanks for working on this! 🙏🏼

PR #6991, opened 2026-03-26
Use pcre2 regex & enable auto async

dranikpg:
Make auto async use pcre2 for regexes. PCRE2 uses limited stack space, whereas std::regex is not suited for the small fiber stacks that we have


dranikpg:
augment reviw

dranikpg:
augment review

Issue #6894, opened 2026-03-15
test_network_disconnect_during_migration


kostasrim:
This test has been failing for some time now and it's unstable/ 

https://github.com/dragonflydb/dragonfly/actions/runs/23104962128/job/67112586556

```
    
        ready = []
        try:
>           fd_event_list = self._selector.poll(timeout, max_ev)
E           Failed: Timeout (>300.0s) from pytest-timeout.

/usr/lib/python3.8/selectors.py:468: Failed
----------------------------- Captured stdout call -----------------------------
30147➜ 07:02:05.860965 28031 interpreter.cc:177] -CROSSSLOT Keys in request don't hash to the same slot

30147➜ 07:02:05.861032 28031 interpreter.cc:177] -CROSSSLOT Keys in request don't hash to the same slot

30147➜ 07:02:05.861099 28031 interpreter.cc:177] -CROSSSLOT Keys in request don't hash to the same slot

30147➜ 07:02:05.861166 28031 interpreter.cc:177] -CROSSSLOT Keys in request don't hash to the same slot

30147➜ 07:02:05.861233 28031 interpreter.cc:177] -CROSSSLOT Keys in request don't hash to the same slot

30147➜ 07:02:05.861300 28031 interpreter.cc:177] -CROSSSLOT Keys in request don't hash to the same slot

30147➜ 07:02:05.861366 28031 interpreter.cc:177] -CROSSSLOT Keys in request don't hash to the same slot

```

kostasrim:
https://github.com/dragonflydb/dragonfly/actions/runs/23589159131/job/68701436984

vyavdoshenko:
https://github.com/dragonflydb/dragonfly/actions/runs/23602847148/job/68737376918

romange:
Another one https://github.com/dragonflydb/dragonfly/actions/runs/23638990930/job/68854723005?pr=6992#step:13:1237

vyavdoshenko:
https://github.com/dragonflydb/dragonfly/actions/runs/23722775217/job/69100588559

kostasrim:
https://github.com/dragonflydb/dragonfly/actions/runs/23748179687/job/69182455884#step:11:272

PR #6998, opened 2026-03-27
fix: restart migration if can not reconnect

BorysTheDev:
fix: #6894

We have an infinite loop during migration finalization: it happens if we have already transferred all data and try to finalize the migration. When the proxy is off, we detect the main connection dropped only, but not for the flow connections, because there is no more data for them, so we try to reconnect the main connection endlessly. This fix restarts the migration if we can not reconnect the main connection during the finalization process. When we do Finish() call with error, the migration process restarts, and we can see the error in the status.
Restart the migration if we can not reconnect during finalization

### DIFF
--- a/src/server/cluster/outgoing_slot_migration.cc
+++ b/src/server/cluster/outgoing_slot_migration.cc
@@ -342,6 +342,7 @@ bool OutgoingMigration::FinalizeMigration(long attempt) {
       LOG(WARNING) << "Couldn't connect to " << cf_->MyID() << " : " << migration_info_.node_info.id
                    << " attempt " << attempt << ": " << ec.message()
                    << ", socket state: " + GetSocketInfo(Sock()->native_handle());
+      Finish(GenericError(ec, "Finalize reconnect failed"));
       return false;
     }
   }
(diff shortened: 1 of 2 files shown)


src/server/cluster/outgoing_slot_migration.cc
```diff
@@ -342,6 +342,7 @@ bool OutgoingMigration::FinalizeMigration(long attempt) {
       LOG(WARNING) << "Couldn't connect to " << cf_->MyID() << " : " << migration_info_.node_info.id
                    << " attempt " << attempt << ": " << ec.message()
                    << ", socket state: " + GetSocketInfo(Sock()->native_handle());
+      Finish(GenericError(ec, "Finalize reconnect failed"));
```
kostasrim, line 345: I do not understand this fix. How this does't get into an endless loop here ?

1. We fail to connect so we call `Finish` 
2. Within `Finish` we `state_` becomes `MigrationState::C_ERROR` 
3. should_cancel_flows=true., we return false

and now on the caller side:

313 while (GetState() != MigrationState::C_FINISHED && !FinalizeMigration(++attempt)) {
314   // Break loop and don't sleep in case of C_FATAL
315   if (GetState() == MigrationState::C_FATAL) {
316     break;
317   }


endless loop

Also:

Plz write and edit the description of this PR and provide `context`. What is the reason the test failed and what is this fix trying to do

PR #7005, opened 2026-03-28
fix: remove set if lazy expiration

BorysTheDev:
This PR ensures Redis-like semantics for sets that use per-member TTL (via lazy expiry): if lazy expiry removes all members, the set key itself should be deleted rather than persisting as an empty collection.

Changes:

- Added SetFamily::DeleteSetIfEmpty helper and invoked it from multiple set operations after lazy-expiry-capable iterations/lookups.
- Updated set commands (notably SCARD, SSCAN, SISMEMBER, SMISMEMBER, SINTER, SMOVE path) to trigger best-effort cleanup of now-empty sets.
- Added regression tests covering key deletion after lazy member expiry across several commands.

### DIFF
--- a/src/server/set_family_test.cc
+++ b/src/server/set_family_test.cc
@@ -760,4 +760,32 @@ TEST_F(SetFamilyTest, SMoveDeletesEmptySourceSet) {
   EXPECT_THAT(Run({"exists", "src"}), IntArg(0));
 }
 
+TEST_F(SetFamilyTest, FieldExpireDeletesEmptySet) {
+  TEST_current_time_ms = kMemberExpiryBase * 1000;
+
+  // Single member so FIELDEXPIRE touches its bucket and triggers lazy expiry.
+  Run({"saddex", "key", "1", "a"});
+  AdvanceTime(2000);
+
+  // FIELDEXPIRE on an already-expired member should clean up the empty set.
+  auto resp = Run({"fieldexpire", "key", "100", "a"});
+  // -2 means the field was not found (expired).
+  EXPECT_THAT(resp, IntArg(-2));
+  EXPECT_THAT(Run({"exists", "key"}), IntArg(0));
+}
+
+TEST_F(SetFamilyTest, FieldTtlDeletesEmptySet) {
+  TEST_current_time_ms = kMemberExpiryBase * 1000;
+
+  // Single member so FIELDTTL touches its bucket and triggers lazy expiry.
+  Run({"saddex", "key", "1", "a"});
+  AdvanceTime(2000);
+
+  // FIELDTTL on an already-expired member should clean up the empty set.
+  auto resp = Run({"fieldttl", "key", "a"});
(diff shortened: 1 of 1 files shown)


BorysTheDev:
augment review

tests/dragonfly/cluster_test.py
```diff
@@ -1480,9 +1480,7 @@ async def test_migration_with_key_ttl(df_factory):
     assert await nodes[1].client.execute_command("stick k_sticky") == 0
```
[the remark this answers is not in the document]
BorysTheDev: #7019
I've removed epoll skip to check is it works ok now or not

src/server/generic_family.cc
```diff
@@ -865,7 +865,11 @@ OpResult<vector<long>> OpFieldExpire(const OpArgs& op_args, string_view key, uin
 
   PrimeValue* pv = &it->second;
   if (pv->ObjType() == OBJ_SET) {
-    return SetFamily::SetFieldsExpireTime(op_args, ttl_sec, values, pv);
+    auto result = SetFamily::SetFieldsExpireTime(op_args, ttl_sec, values, pv);
+    // Finalize memory accounting before potential deletion.
+    auto_updater.Run();
+    SetFamily::DeleteSetIfEmpty(db_slice, op_args.db_cntx, key, *pv);
```
vyavdoshenko, line 871: Could you add test for this case?

src/server/generic_family.cc
```diff
@@ -887,6 +891,7 @@ OpResult<long> OpFieldTtl(Transaction* t, EngineShard* shard, string_view key, s
   int32_t res = -1;
   if (it->second.ObjType() == OBJ_SET) {
     res = SetFamily::FieldExpireTime(db_cntx, it->second, field);
+    SetFamily::DeleteSetIfEmpty(db_slice, db_cntx, key, it->second);
```
vyavdoshenko, line 894: And this as well

BorysTheDev:
augment review

src/server/set_family.cc
```diff
@@ -1599,6 +1599,24 @@ void CmdSAddEx(CmdArgList args, CommandContext* cmd_cntx) {
...
+
+  if (auto res = db_slice.FindMutable(db_cntx, key, OBJ_SET); res) {
+    db_slice.DelMutable(db_cntx, std::move(*res));
+    if (db_slice.shard_owner()->journal()) {
+      RecordDelete(db_cntx.db_index, key);
+    }
+    return true;
+  }
+  return false;
+}
```
BorysTheDev, line 1618: moved without changes

src/server/generic_family.cc
dranikpg, line 871: I see this pattern repeating a few times, and even `DeleteSetIfEmpty` was exported to support it. What if there was a single function that took the auto updater by value and did all of this

BorysTheDev: good  idea, we can return to it after making build stable

PR #7007, opened 2026-03-28
chore: detect hash field duplication during RDB load via HLL

romange:
## Summary
Adds HyperLogLog-based detection of duplicated hash field names during RDB load. When loading an RDB snapshot containing many hash objects, this detects whether field names are heavily reused across different hash keys.

### Changes
- **src/core/hll_estimator.h** — New shared header extracting HLL primitives (UpdateRegister, EstimateCardinality, MergeRegisters) from dict_builder.cc
- **src/core/dict_builder.cc** — Refactored to use hll_estimator.h instead of inline HLL code
- **src/server/rdb_load.h** — Added per-shard ShardHllState struct to RdbLoader
- **src/server/rdb_load.cc** — Tracks hash field names through the HLL in CreateObjectOnShard (for RDB_TYPE_HASH and RDB_TYPE_HASH_WITH_EXPIRY, skips ZIPLIST/LISTPACK), merges and logs in FinishLoad

### Design
- **Lock-free**: each shard thread writes to its own HLL register slot (per_shard_hll_[shard_id])
- **Merge at end**: FinishLoad() merges all per-shard registers (element-wise max) and estimates cardinality
- **LOG(INFO)** reports unique vs total field counts when total > 0
- Chunked append items are skipped to avoid double-counting
- Uses FNV-1a hash for field name strings

### DIFF
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -2405,8 +2405,8 @@ void RdbLoader::FinishLoad(absl::Time start_time, size_t* keys_loaded) {
   }
   if (total_fields > 0) {
     uint64_t unique_est = static_cast<uint64_t>(hll::EstimateCardinality(merged_regs.data()));
-    LOG(INFO) << "RDB load hash field stats: ~" << unique_est << " unique field names out of "
-              << total_fields << " total across all hashes";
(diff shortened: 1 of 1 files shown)


romange:
Superseded by #7009

PR #7011, opened 2026-03-29
chore(facade): improve IoLoopV2 readability and simplify recv flow

romange:
## Summary

The main functional change - I pulled socket reading from `DoReadOnRecv` for the "regular", 
epoll-notification flow outside from the callback. Socket pending read state is separated from io_buf_ state by introducing an explicit `pending_input_` boolean.

Everything else is non-functional readability improvements.

- Introduce `HasInFlightCommands()` and `HasCommandToExecute()` predicates for the parsed command queue, replacing raw pointer comparisons
- Move socket draining from `DoReadOnRecv` into `IoLoopV2` directly via `pending_input_` flag, removing duplicated `TryRecv` logic
- Simplify `ReplyBatch` loop by advancing `parsed_head_` directly instead of aliasing through a reference variable with `exchange()`

## Changes

- **`src/facade/dragonfly_connection.h`** — Add `HasInFlightCommands()`, `HasCommandToExecute()` inline helpers; add `pending_input_` bit field
- **`src/facade/dragonfly_connection.cc`** — Refactor `IoLoopV2` recv loop to drain socket inline; simplify `ReplyBatch`; replace raw pointer comparisons with named predicates; simplify `DoReadOnRecv` to just set `pending_input_`; move `migration_allowed_to_register_` init into `IoLoopV2`
- **`tests/dragonfly/pymemcached_test.py`** — Add `test_get_many` for multi-key get coverage

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -685,7 +685,7 @@ void Connection::OnPostMigrateThread() {
 
   if (ioloop_v2_ && socket_ && socket_->IsOpen() && migration_allowed_to_register_) {
     socket_->RegisterOnRecv([this](const FiberSocketBase::RecvNotification& n) {
-      DoReadOnRecv(n);
+      NotifyOnRecv(n);
       io_event_.notify();
     });
   }
@@ -2400,12 +2400,12 @@ bool Connection::ExecuteBatch() {
 
 bool Connection::ReplyBatch() {
   reply_builder_->SetBatchMode(true);
-  while (HasInFlightCommands() && parsed_head_->CanReply()) {
+  while (HasDispatchedCommands() && parsed_head_->CanReply()) {
     current_wait_.reset();  // we must free waiter before proceeding with other commands
     auto* cmd = parsed_head_;
     parsed_head_ = cmd->next;
     cmd->SendReply();
-    ReleaseParsedCommand(cmd, HasInFlightCommands() /* is_pipelined */);
+    ReleaseParsedCommand(cmd, HasDispatchedCommands() /* is_pipelined */);
     if (reply_builder_->GetError())
       return false;
   }
@@ -2588,7 +2588,7 @@ bool ConnectionRef::operator==(const ConnectionRef& other) const {
   return client_id_ == other.client_id_;
 }
 
-void Connection::DoReadOnRecv(const util::FiberSocketBase::RecvNotification& n) {
+void Connection::NotifyOnRecv(const util::FiberSocketBase::RecvNotification& n) {
   if (std::holds_alternative<std::error_code>(n.read_result)) {
     io_ec_ = std::get<std::error_code>(n.read_result);
     return;
@@ -2608,6 +2608,34 @@ void Connection::DoReadOnRecv(const util::FiberSocketBase::RecvNotification& n) 
                         [&]() { io_buf_.WriteAndCommit(buf.data(), buf.size()); });
   } else {
     LOG(FATAL) << "Should not reach here";
+  }
+}
+
+void Connection::ReadPendingInput() {
+  // Drain available socket data into io_buf_.
+  io::MutableBytes buf = io_buf_.AppendBuffer();
+  // A recv call can return fewer bytes than requested even if the
+  // socket buffer actually contains enough data to satisfy the full request.
+  while (!buf.empty()) {
+    io::Result<size_t> res = socket_->TryRecv(buf);
+    if (!res) {
+      auto ec = res.error();
+      // TryRecv is non-blocking: it returns EAGAIN/EWOULDBLOCK when nothing is ready.
+      if (ec == errc::resource_unavailable_try_again || ec == errc::operation_would_block)
+        pending_input_ = false;
+      else
+        io_ec_ = ec;
+      break;
+    }
+
+    if (*res == 0) {
+      io_ec_ = make_error_code(errc::connection_aborted);  // *res == 0, clean EOF
+      pending_input_ = false;
+      break;
+    }
+
+    io_buf_.CommitWrite(*res);
+    buf = io_buf_.AppendBuffer();
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/facade/dragonfly_connection.h
```diff
@@ -437,10 +437,23 @@ class Connection : public util::Connection {
...
   // Total bytes used by commands in parsed command queue
   size_t parsed_cmd_q_bytes_ = 0;
+
+  // Returns true if there are dispatched commands that haven't been replied yet.
+  bool HasInFlightCommands() const {
+    return parsed_head_ != parsed_to_execute_;
+  }
+
+  // Returns true if the head command is ready to execute (nothing in-flight ahead of it).
+  bool HasCommandToExecute() const {
```
glevkovich: I think that a more accurate name would be `IsHeadReadyToExecute()`. This is what is being checked here. It doesn't check if there is at least one command that can be executed, not necessarily the head:  it checks if the head command is the one currently waiting to be executed, ensuring no previously dispatched commands are blocking it.

src/facade/dragonfly_connection.h
```diff
@@ -437,10 +437,23 @@ class Connection : public util::Connection {
...
   ParsedCommand* parsed_to_execute_ = nullptr;
+
   // Total number of commands in parsed command queue
   size_t parsed_cmd_q_len_ = 0;
+
   // Total bytes used by commands in parsed command queue
   size_t parsed_cmd_q_bytes_ = 0;
+
+  // Returns true if there are dispatched commands that haven't been replied yet.
+  bool HasInFlightCommands() const {
```
glevkovich: nit: suggest to rename HasInFlightCommands() to `HasDispatchedCommands().`
The term "in-flight" introduces new terminology and implies a specific execution state we can't strictly guarantee (they could be running or already completed). HasDispatchedCommands() is more accurate and grounded in the code's existing logic: we know the commands have been dispatched, even if we don't know their current processing status (running or completed and pending reply).

src/facade/dragonfly_connection.cc
```diff
@@ -2712,33 +2682,46 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
...
+            // TryRecv is non-blocking: it returns EAGAIN/EWOULDBLOCK when nothing is ready.
+            if (ec == errc::resource_unavailable_try_again || ec == errc::operation_would_block)
+              pending_input_ = false;
+            else
+              io_ec_ = ec;
+          }
+          break;
+        }
+        io_buf_.CommitWrite(*res);
+        buf = io_buf_.AppendBuffer();
```
dranikpg: Please move such portions out of the functions, otherwise it becomes too large to be easily readable

dranikpg:
oh, I merged my ioloop pr and now it has a conflict

src/facade/dragonfly_connection.cc
```diff
@@ -2712,33 +2682,46 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
...
+    }
+
+    if (pending_input_) {
+      // Drain available socket data into io_buf_.
+      io::MutableBytes buf = io_buf_.AppendBuffer();
+      // A recv call can return fewer bytes than requested even if the
+      // socket buffer actually contains enough data to satisfy the full request.
+      while (!buf.empty()) {
+        io::Result<size_t> res = socket_->TryRecv(buf);
+        if (!res || *res == 0) {
```
glevkovich: I think you can simplify with a single layer of if/else (and 2 break):

```
io::Result<size_t> res = socket_->TryRecv(buf);
        
        if (!res) { //  / error
          auto ec = res.error();
          if (ec == errc::resource_unavailable_try_again || ec == errc::operation_would_block) {
            pending_input_ = false;
          } else {
            io_ec_ = ec;
          }
          break;
        } 
        else if (*res == 0) { // clean EOF
          io_ec_ = make_error_code(errc::connection_aborted);
          pending_input_ = false;
          break; 
        }
```

tests/dragonfly/pymemcached_test.py
```diff
@@ -88,6 +88,12 @@ async def test_noreply_pipeline(self, df_server: DflyInstance, memcached_client:
 
         assert False, "Pipelining not detected"
 
+    def test_get_many(self, memcached_client: MCClient):
```
glevkovich, line 91: AI generated:
The test_get_many method lacks coverage for edge cases such as empty keys list, non-existent keys, or mixed existing/non-existing keys. Add test cases to verify behavior with empty input and keys that don't exist in the cache.

and in more details:
test_get_many only tests the "happy path" (fetching 100 keys that all exist). In batch operations like Memcached's get (which supports multiple keys), it is standard practice to verify how the server handles missing keys. Memcached protocol specifically dictates that missing keys are simply omitted from the response, while existing ones are returned. Testing a mix of existing and non-existing keys ensures the parser and reply builder don't lose alignment or crash.

romange: the goal is not to pursue correctness of `get` - it was to get more coverage for my changes as i had a bug and this test helped to debug it so I decided to keep it. The goal of this test is to increase coverage of facade.

src/facade/dragonfly_connection.h
romange: I think I will push back on that as "head ready" is an implementation detail. What I want to know at high level is whether we have commands ready to dispatch. I will align it though with our jargon here and below

src/facade/dragonfly_connection.h
romange: ok

src/facade/dragonfly_connection.h
romange: Atually you are correct. We could execute other commands, so, if we have  `head-> ... -> to_execute-> tail` 
why do we call `ExecuteBatch` only if `head == to_execute` ? @dranikpg  what's the reason for that?

src/facade/dragonfly_connection.cc
romange: done

src/facade/dragonfly_connection.h
dranikpg: It seems there is no good answer for it, it can be called eariler if there is something to execute

romange: @glevkovich  please add a TODO near the calls to this function to explore why we do not execute commands that are added to the queue while HasDispatchedCommands is true. Seems like an arbitrary decision.

PR #7012, opened 2026-03-29
chore(facade): rename predicates, add IoBuf& params, non-functional improvements

romange:
## Summary
Rename parsed-command predicates for clarity:
- HasDispatchedCommands() -> HasInFlightCommands()
- HeadReadyToDispatch() -> HasCommandToExecute()

Generalize parse methods to accept an explicit IoBuf* parameter
(ParseRedis, ParseMCBatch, ParseRedisBatch, CheckIoBufCapacity),
decoupling them from io_buf_ so they can operate on any buffer.
All callers pass io_buf_ so behavior is unchanged.

Add ParseFromBuffer(IoBuf&): protocol-agnostic helper that dispatches
to ParseMCBatch or ParseRedisBatch based on the current protocol.

Finally, move to clearer and simpler ReadBufTracker instead of UpdateIoBufCapacity monstrosity.
I can critisize it - I added it.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1366,12 +1366,6 @@ Connection::ParserStatus Connection::ParseRedis(base::IoBuf& io_buf, unsigned ma
   return ERROR;
 }
 
-void Connection::ParseFromBuffer(base::IoBuf& buf) {
-  auto parse_func =
-      protocol_ == Protocol::MEMCACHE ? &Connection::ParseMCBatch : &Connection::ParseRedisBatch;
-  (this->*parse_func)(buf);
-}
-
 auto Connection::ParseLoop() -> ParserStatus {
   auto parse_func =
       protocol_ == Protocol::MEMCACHE ? &Connection::ParseMCBatch : &Connection::ParseRedisBatch;
@@ -1495,7 +1489,7 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoop() {
     }
 
     phase_ = PROCESS;
-    bool is_iobuf_full = io_buf_.AppendLen() == 0;
+    bool reached_capacity = io_buf_.AppendLen() == 0;
 
     if (redis_parser_) {
       parse_status = ParseRedis(io_buf_, max_busy_read_cycles_cached);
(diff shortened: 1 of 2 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -1353,13 +1366,19 @@ Connection::ParserStatus Connection::ParseRedis(unsigned max_busy_cycles, bool e
   return ERROR;
 }
 
+void Connection::ParseFromBuffer(base::IoBuf& buf) {
```
glevkovich: Can you clarify why are you silently dropping the bool returned by ParseMCBatch/ParseRedisBatch?
Also, function has no callers  (dead code for now) - do you plan to use it in future commits?

src/facade/dragonfly_connection.h
```diff
@@ -287,7 +287,7 @@ class Connection : public util::Connection {
   // Drains currently available bytes from socket into io_buf_ using non-blocking reads.
   void ReadPendingInput();
 
-  void CheckIoBufCapacity(bool is_iobuf_full);
+  void CheckIoBufCapacity(bool reached_capacity, base::IoBuf* buf);
```
glevkovich, line 290: 1) nit: personally, I feel like  is_iobuf_full is clearer, but I guess it's matter of taste. Even if you decide to keep the new name, there is inconsistency with the local variable naming: if you decided to use reached_capacity, I would suggest to rename local variables as well:
is_iobuf_full should be renamed to reached_capacity. See both in IoLoopV2() and IoLoop().

2) Also, I think it would be more consistent, to update this to take a reference (base::IoBuf& buf). The rest of the file uses references for these non-optional mutable buffers.

romange: what does it mean `is_iobuf_full` ? what does it mean a collection (vector) is full?

I am confused about the semantics of this as our collections are resizable and do not have "is_full" state. moreover, when we call `CheckIoBufCapacity`,  io_buf is not "full".

src/facade/dragonfly_connection.cc
romange: it's a mistake

src/facade/dragonfly_connection.h
glevkovich: > what does it mean `is_iobuf_full` ? what does it mean a collection (vector) is full?
> 
> I am confused about the semantics of this as our collections are resizable and do not have "is_full" state. moreover, when we call `CheckIoBufCapacity`, io_buf is not "full".

OK, make sense, still the suggestion to rename on the 1st part is still valid.
look for (2 places):
`bool is_iobuf_full = io_buf_.AppendLen() == 0;`

glevkovich (approved):
LGTM

Issue #7013, opened 2026-03-29
FT.INFO omits vector field parameters in attributes


vyavdoshenko:
`FT.INFO` returns vector fields without their parameters (`algorithm`, `dim`, `data_type`, `distance_metric`, `M`, `ef_construction`).

Reproduce:

```
FT.CREATE test ON HASH PREFIX 1 doc: SCHEMA text TEXT embedding VECTOR HNSW 10 TYPE FLOAT32 DIM 4 DISTANCE_METRIC COSINE M 16 EF_CONSTRUCTION 200
FT.INFO test
```

DragonflyDB df-dev — attributes section:

```
1. identifier
2. embedding
3. attribute
4. embedding
5. type
6. VECTOR
```

Competitor - attributes section:

```
1. identifier
2. embedding
3. attribute
4. embedding
5. type
6. VECTOR
7. algorithm
8. HNSW
9. data_type
10. FLOAT32
11. dim
12. 4
13. distance_metric
14. COSINE
15. M
16. 16
17. ef_construction
18. 200
```

Impact:

`redisvl` uses `FT.INFO` to reconstruct the index schema in `SearchIndex.from_existing()`. Without vector parameters, the vector field is silently dropped from the schema.

This breaks `langchain-redis` `RedisVectorStore` with `from_existing=True` — any `similarity_search()` call raises `KeyError: 'embedding'`.

vyavdoshenko:
FT.INFO: multiple missing field parameters and data loss in schema round-trip

Summary:

`FT.INFO` omits field-type-specific parameters for VECTOR, TAG, and TEXT fields. Beyond cosmetics, several of these parameters are also **silently dropped during `FT.CREATE` parsing**, meaning they cannot be recovered even if FT.INFO is fixed independently. Together, these issues break schema round-trips relied on by tools such as `redisvl` and `langchain-redis`.

---

Problem 1 - VECTOR: `TYPE` parameter is silently discarded at parse time

**Impact: data loss**

`ParseVectorParams` has no handler for the `TYPE` keyword — it falls through to `parser->Skip(2)` and the value is never stored:

```
FT.CREATE test ON HASH PREFIX 1 doc: SCHEMA
  embedding VECTOR HNSW 10 TYPE FLOAT32 DIM 4 DISTANCE_METRIC COSINE M 16 EF_CONSTRUCTION 200
```

[...]

PR #7014, opened 2026-03-29
fix: preserve M and EF_CONSTRUCTION in HNSW index restore command

vyavdoshenko:
`BuildRestoreCommand` serialized HNSW VECTOR fields with a hardcoded argument count of 6, covering only DIM, DISTANCE_METRIC, and INITIAL_CAP. M and EF_CONSTRUCTION were silently dropped.

On replication or RDB restore, the HNSW index was reconstructed with default values (M=16, EF_CONSTRUCTION=200) regardless of what the user had configured.

Fix: emit all 5 HNSW parameters (arg count 10) and separate the HNSW and FLAT branches since FLAT legitimately takes only 3 parameters.

Related to #7013

### DIFF
--- a/src/server/search/search_family_test.cc
+++ b/src/server/search/search_family_test.cc
@@ -17,6 +17,7 @@
 #include "facade/error.h"
 #include "facade/facade_test.h"
 #include "facade/resp_parser.h"
+#include "server/search/doc_index.h"
 #include "server/test_utils.h"
 
 using namespace testing;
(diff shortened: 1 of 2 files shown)


src/server/search/search_family_test.cc
```diff
@@ -4922,4 +4923,49 @@ TEST_F(SearchFamilyTest, FtAggregateFilterSemanticRouting) {
                                          IsMap("route_name", "sports", "avg_dist", "0.6")));
 }
 
+// Verify that BuildRestoreCommand round-trips all HNSW parameters.
+// Regression test: M and EF_CONSTRUCTION were previously omitted, causing
+// replication to silently reconstruct the index with wrong defaults.
+TEST(BuildRestoreCommandTest, HnswVectorPreservesAllParams) {
```
dranikpg, line 4929: we also have some pytest for that where we load the index back

vyavdoshenko: I think unit tests are more preferable.

PR #7018, opened 2026-03-30
feat(conn): add pipeline backpressure for IoLoopV2

glevkovich:
Park the IoLoopV2 fiber via io_event_.await() when the pipeline queue
exceeds limits, preventing busy-spinning. The fiber resumes when
memory is freed, commands are ready, the dispatch queue has items,
or the connection closes.

To prevent lost wakeups on thread-global limits, parked V2 connections
now register their io_event_ in a thread-local waiter list. This ensures
they reliably wake up when other connections free memory or when limits
are dynamically raised via CONFIG SET.

- ParseRedis breaks early if over limit, leaving bytes in io_buf_.
- ExecuteBatch and ReplyBatch notify io_event_ to wake the fiber.
- Global memory relief safely wakes parked V2 fibers, utilizing an O(1)
  swap-and-pop unregister on cleanup.
- Wakeups are optimized to avoid the "thundering herd" problem by only
  notifying waiters when thread-global pipeline budget is actually freed.
- Updated logs with (V1)/(V2) labels for easier debugging.
- Added tests for V2 correctness and safe disconnects.

Signed-off-by: Gil Levkovich <69595609+glevkovich@users.noreply.github.com>

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -2294,9 +2294,15 @@ bool Connection::IsReplySizeOverLimit() const {
 
 bool Connection::ParseRedisBatch() {
   QueueBackpressure& qbp = GetQueueBackpressure();
-  if (qbp.IsPipelineBufferOverLimit(GetLocalConnStats().pipeline_queue_bytes, parsed_cmd_q_len_)) {
+
+  // Only throttle parsing if this connection is actively contributing to the queue.
+  // Connections with parsed_cmd_q_len_ == 0 must always be allowed to parse so that
+  // administrative commands (CONFIG SET, etc.) can execute and relieve backpressure.
+  if ((parsed_cmd_q_len_ > 0) &&
+      qbp.IsPipelineBufferOverLimit(GetLocalConnStats().pipeline_queue_bytes, parsed_cmd_q_len_)) {
     // Signal ParseLoop to stop. IoLoopV2 will drain before resuming.
     DVLOG(2) << "Pipeline buffer over limit. Avoid parsing Redis batch.";
+    GetLocalConnStats().pipeline_throttle_count++;
     return false;
   }
   return ParseRedis(max_busy_read_cycles_cached, true) == ParserStatus::OK;
@@ -2810,10 +2816,23 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
     // Handle Parsed Commands Queue (Data Path)
     auto& conn_stats = GetLocalConnStats();
     QueueBackpressure& qbp = GetQueueBackpressure();
+
+    // Only parse data if we are under the memory limit (backpressure).
+    // Exception: If the queue is empty, we always parse to allow admin commands
+    // (like CONFIG SET) to run so they can fix the memory limits if needed.
     bool pre_over_limit =
+        (parsed_cmd_q_len_ > 0) &&
         qbp.IsPipelineBufferOverLimit(conn_stats.pipeline_queue_bytes, parsed_cmd_q_len_);
     if (io_buf_.InputLen() > 0 && !pre_over_limit) {  // Parse, execute and reply
+      size_t mem_before = conn_stats.pipeline_queue_bytes;
       parse_status = ParseLoop();
+
+      // Executing and replying to commands (in ParseLoop()) frees up memory. Because those internal
+      // functions only wake up this specific connection, we need to manually notify
+      // other connections on this thread that there is now room to resume.
+      if (conn_stats.pipeline_queue_bytes < mem_before) {
+        qbp.NotifyPipelineWaiters();
+      }
     } else {  // Execute and reply what we have, then wait if we are over the limit
       parse_status = NEED_MORE;
 
@@ -2833,7 +2852,10 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
       // await block (backpressure)
       // Re-check if pipeline buffer over limit after draining - ExecuteBatch/ReplyBatch may have
       // freed memory. If still over limit, sleep to prevent busy-spin.
+      // Only park if this connection is actively contributing (parsed_cmd_q_len_ > 0).
+      // Connections with an empty queue must stay in the read loop.
       bool post_over_limit =
+          (parsed_cmd_q_len_ > 0) &&
           qbp.IsPipelineBufferOverLimit(conn_stats.pipeline_queue_bytes, parsed_cmd_q_len_);
       if (post_over_limit) {
         conn_stats.pipeline_throttle_count++;
@@ -2845,27 +2867,25 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
             << ", Connection parsed commands queue size: " << parsed_cmd_q_len_
             << ", consider increasing pipeline_buffer_limit/pipeline_queue_limit";
 
-        // Subscribe to the global backpressure EventCount so that when another connection
-        // frees memory (or CONFIG SET raises limits), our backpressure_waiter callback fires
-        // io_event_.notify(), waking this fiber. The returned SubKey auto-unlinks in O(1)
-        // when it goes out of scope.
-        auto under_limit_pred = [this]() {
-          return !GetQueueBackpressure().IsPipelineBufferOverLimit(
-              GetLocalConnStats().pipeline_queue_bytes, parsed_cmd_q_len_);
-        };
-        auto sub_key = qbp.v2_pipeline_backpressure_ec.check_or_subscribe(under_limit_pred,
-                                                                          &backpressure_waiter);
+        // Subscribe persistently to the global backpressure EventCount so that when another
+        // connection frees memory (or CONFIG SET raises limits), our backpressure_waiter callback
+        // fires io_event_.notify(), waking this fiber. Must be persistent because
+        // io_event_.await()'s internal loop may re-sleep if the predicate is still false after the
+        // first notification. A one-shot subscription would be consumed on the first wake, leaving
+        // us "deaf" to future memory relief.
+        auto sub_key = qbp.v2_pipeline_backpressure_ec.subscribe_persistent(&backpressure_waiter);
 
-        if (sub_key) {  // we had to subscribe, so need to wait
-          io_event_.await([this, under_limit_pred]() {
-            bool cmd_ready = parsed_head_ && parsed_head_->CanReply();
-            // We wake up if:
-            // 1. Memory is freed (under_limit_pred()) or we can free it ourselves (cmd_ready).
-            // 2. Control-plane messages need processing (!dispatch_q_.empty()).
-            // 3. The connection is terminating (io_ec_).
-            return under_limit_pred() || cmd_ready || !dispatch_q_.empty() || io_ec_;
-          });
-        }
+        io_event_.await([this]() {
+          bool cmd_ready = parsed_head_ && parsed_head_->CanReply();
+          bool under_limit = !GetQueueBackpressure().IsPipelineBufferOverLimit(
+              GetLocalConnStats().pipeline_queue_bytes, parsed_cmd_q_len_);
+
+          // We wake up and exit the backpressure wait if:
+          // 1. Memory is freed (under_limit) or we can free it ourselves (cmd_ready).
+          // 2. Control-plane messages need processing (!dispatch_q_.empty()).
+          // 3. The connection is terminating (io_ec_).
+          return under_limit || cmd_ready || !dispatch_q_.empty() || io_ec_;
+        });
       }
     }
 
--- a/tests/dragonfly/connection_test.py
+++ b/tests/dragonfly/connection_test.py
@@ -1326,7 +1326,12 @@ async def test_pipeline_cache_size(df_server: DflyInstance):
 
 
 @dfly_multi_test_args(
-    {"proactor_threads": 4, "pipeline_queue_limit": 10, "pipeline_buffer_limit": "1024"},
+    {
+        "proactor_threads": 4,
+        "pipeline_queue_limit": 10,
+        "pipeline_buffer_limit": "1024",
+        "experimental_io_loop_v2": "false",
+    },
     {
         "proactor_threads": 4,
         "pipeline_queue_limit": 10,
@@ -1367,10 +1372,18 @@ async def test_pipeline_overlimit(df_factory: DflyInstanceFactory):
 
     await asyncio.sleep(2)
 
-    # Verify backpressure was exercised
+    # Verify backpressure was exercised.
+    # V1 increments pipeline_throttle_total reliably because DispatchSingle parks the fiber.
+    # V2's self-regulating ParseLoop (parse→execute→reply) drains the queue each iteration,
+    # so the throttle counter may not fire even though backpressure is applied internally.
     info = await client.info("stats")
-    assert int(info["pipeline_throttle_total"]) > 0, "Expected pipeline_throttle_total > 0"
+    is_v2 = server.args.get("experimental_io_loop_v2") == "true"
+    if not is_v2:
+        assert int(info["pipeline_throttle_total"]) > 0, "Expected pipeline_throttle_total > 0"
 
+    # Raise byte limit FIRST: if queue-limit is raised first, connections wake and
+    # immediately re-hit the byte ceiling, blocking the second CONFIG SET.
+    await client.config_set("pipeline_buffer_limit", 1024 * 1024 * 50)
     await client.config_set("pipeline_queue_limit", 10000)
     for task in pipeline_tasks:
         await task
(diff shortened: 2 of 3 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -2770,16 +2794,47 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
...
+        io_event_.await([this]() {
+          bool cmd_ready = parsed_head_ && parsed_head_->CanReply();
+          bool under_limit = !GetQueueBackpressure().IsPipelineBufferOverLimit(
+              GetLocalConnStats().pipeline_queue_bytes, parsed_cmd_q_len_);
+          // We wake up if:
+          // 1. Memory is freed (under_limit) or we can free it ourselves (cmd_ready).
+          // 2. Control-plane messages need processing (!dispatch_q_.empty()).
+          // 3. The connection is terminating (io_ec_).
+          return under_limit || cmd_ready || !dispatch_q_.empty() || io_ec_;
+        });
```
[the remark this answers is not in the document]
glevkovich: fixed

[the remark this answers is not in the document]
glevkovich: Added a registration and notification mechanism to fix this

src/facade/dragonfly_connection.cc
```diff
@@ -2770,16 +2794,47 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
...
+      if (post_over_limit) {
+        conn_stats.pipeline_throttle_count++;
+        LOG_EVERY_T(WARNING, 10)
+            << "Pipeline buffer over limit (V2)."
+            << ", Thread pipeline_queue_bytes: " << conn_stats.pipeline_queue_bytes
+            << ", Thread pipeline_queue_entries: " << conn_stats.pipeline_queue_entries
+            << ", Connection parsed_cmd_q_bytes_: " << parsed_cmd_q_bytes_
+            << ", Connection parsed commands queue size: " << parsed_cmd_q_len_
+            << ", consider increasing pipeline_buffer_limit/pipeline_queue_limit";
+        io_event_.await([this]() {
```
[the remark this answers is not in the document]
glevkovich: already fixed. dropping.

tests/dragonfly/connection_test.py
```diff
@@ -1325,29 +1325,121 @@ async def push_pipeline(bad_actor_client, size=1):
...
+    # Verify backpressure was exercised
+    info = await client.info("stats")
+    assert int(info["pipeline_throttle_total"]) > 0, "Expected pipeline_throttle_total > 0"
+
     await client.config_set("pipeline_queue_limit", 10000)
     for task in pipeline_tasks:
         await task
 
 
+@dfly_args({"proactor_threads": 4, "pipeline_queue_limit": 10, "experimental_io_loop_v2": "true"})
```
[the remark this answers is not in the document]
glevkovich: for now, validating manually...dropping this since we are going to enable it soon..

tests/dragonfly/connection_test.py
```diff
@@ -1325,29 +1325,121 @@ async def push_pipeline(bad_actor_client, size=1):
...
+
+
+@dfly_multi_test_args(
+    {"proactor_threads": 1, "pipeline_queue_limit": 5},
+    {"proactor_threads": 1, "pipeline_queue_limit": 5, "experimental_io_loop_v2": "true"},
+)
+async def test_pipeline_backpressure_disconnect(df_factory: DflyInstanceFactory):
+    """Verify a client stuck in backpressure can safely disconnect (io_ec_ trigger)."""
+    server = df_factory.create()
+    server.start()
```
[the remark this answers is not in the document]
glevkovich: this will be solved/fixed on later prs.

tests/dragonfly/connection_test.py
```diff
@@ -1325,29 +1325,121 @@ async def push_pipeline(bad_actor_client, size=1):
...
     await client.config_set("pipeline_queue_limit", 10000)
     for task in pipeline_tasks:
         await task
 
 
+@dfly_args({"proactor_threads": 4, "pipeline_queue_limit": 10, "experimental_io_loop_v2": "true"})
+async def test_pipeline_backpressure_v2_correctness(df_server: DflyInstance):
+    """Verify that V2 backpressure throttles without corrupting responses."""
+    client = df_server.client()
+    value = "v" * 512
```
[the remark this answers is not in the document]
glevkovich: this will be solved/fixed on later prs.

src/facade/dragonfly_connection.cc
```diff
@@ -361,6 +362,19 @@ struct QueueBackpressure {
   // Used together with pipeline_buffer_limit to limit the pipeline usage per thread.
   util::fb2::CondVarAny pipeline_cnd;
 
+  // V2 connections parked on backpressure register their io_event_ here so that
+  // global memory relief (from other connections or CONFIG SET) can wake them.
+  // Access is thread-local (per-proactor), no synchronization needed.
+  std::vector<util::fb2::EventCount*> v2_pipeline_waiters;
```
romange: if access is indeed thread-local, then using CondVarAny with NoOp lock is better, 
as EventCount is designed  for multi-threaded coordination.

src/facade/dragonfly_connection.cc
```diff
@@ -1632,7 +1655,7 @@ void Connection::ClearPipelinedMessages() {
   parsed_to_execute_ = nullptr;
 
   QueueBackpressure& qbp = GetQueueBackpressure();
-  qbp.pipeline_cnd.notify_all();
+  qbp.NotifyPipelineWaiters();
```
romange, line 1657: I am trying to understand how it worked before - we did not have `O(n)` array of fibers parked. 
so what changed now?

src/facade/dragonfly_connection.cc
```diff
@@ -2413,6 +2446,10 @@ bool Connection::ReplyBatch() {
   if (parsed_head_ == nullptr)
     parsed_tail_ = nullptr;
 
+  if (ioloop_v2_) {
```
romange, line 2457: please add comment why it's needed.

src/facade/dragonfly_connection.cc
```diff
@@ -2758,16 +2796,70 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
...
+            << ", consider increasing pipeline_buffer_limit/pipeline_queue_limit";
+
+        // Register this connection's io_event_ so that global memory relief (from other connections
+        // draining or CONFIG SET raising limits) can wake us up. Without this, we'd park
+        // indefinitely because io_event_ is per-connection but the predicate depends on
+        // thread-global state.
+        qbp.v2_pipeline_waiters.push_back(&io_event_);
+        absl::Cleanup v2_unregister = [&qbp, this] {
+          auto& w = qbp.v2_pipeline_waiters;
+          auto it = std::find(w.begin(), w.end(), &io_event_);
```
romange: i have a problem with this and also with `NotifyPipelineWaiters` that runs in `O(N)`.

I want to understand what has changed that now we need `O(n)` runtime for these operations.
Specifically how this works in V1?  I would guess we have a single `cndVar` that you block on. Why do we need to change the approach here?

src/facade/dragonfly_connection.cc
glevkovich: OK

src/facade/dragonfly_connection.cc
glevkovich: I will answer [here](https://github.com/dragonflydb/dragonfly/pull/7018#discussion_r3011354813) Since i think few comments are related to the same issue.

src/facade/dragonfly_connection.cc
glevkovich: I will answer [here](https://github.com/dragonflydb/dragonfly/pull/7018#discussion_r3011354813) Since i think few comments are related to the same issue.

src/facade/dragonfly_connection.cc
glevkovich: Answering here to 3 of your comments:

In V1:
- We have 2 fibers per connection, one is a producer (IoLoop) and one is a consumer (the one is the consumer that runs Connection::AsyncFiber()).
- If memory got too full, the IoLoop fiber could block on a thread-global CondVarAny (pipeline_cnd) in Dispatchsingle, while expecting AsyncFiber which was still awake, to eventually wake it up.
- Connections waiting on backpressure all park on the same thread-level pipeline_cnd (CondVarAny) . One notify_all() reaches them all because they share a single wait object.

in V2:
- There is only a single fiber now (no AsyncFiber). The single fiber uses io_event as a unified notification object for receive completions, connection shutdown, async message dispatch, and command completion.
- When parked on io_event_ we must park on all, we can;t park on 2 different objects at the same time, and that includes parking on backpressure relief events. If we park on pipeline_cnd_ we will be "deaf" to all other events.

Regarding your comments:
- "if access is indeed thread-local, then using CondVarAny with NoOp lock is better,
as EventCount is designed for multi-threaded coordination.": you are right but as I wrote, the reason I can;t use pipeline_cndd for V2 is because the fiber can only be parked on a single object at a time and must expect all possible events, and the code was written to support passing events (e.g socket receive completions) via this io_event_.

- "I am trying to understand how it worked before - we did not have O(n) array of fibers parked.
so what changed now?"  If you look at the implementation of the pipeline_cnd.notiftall(), it is also O(n), it's just in the internal implementation of CondVarAny:
```
bool WaitQueue::NotifyAll(FiberInterface* active) {
  bool notified = false;
  while (NotifyOne(active))
    notified = true;
  return notified;
}
```
To summarize: 

what actually change is the way we notify and how we organize members logically:
- V1 loop has all connections sharing one CondVarAny with a single notify_all() works.
- V2 connections each have their own io_event_, and  you need to explicitly reach each one in dragonfly_connection code. The vector is the minimal mechanism to do so.

Why we can't use a new or existing CondVarAny in my opinion?
Since it has the same fundamental issue - during the back-pressure wait, the fiber must remain parked on io_event_ because socket completions, shutdowns, and async dispatches all arrive exclusively via io_event_.notify(). A fiber can only wait in one place at a time, so we cannot park on both io_event_ and pipeline_cnd or any new CondVarAny  simultaneously.

Let me know if you'd prefer to explore a different architectural approach for V2's event multiplexing. I think that this might require expanding the scope of this PR to refactor how V2 waits for events.

dranikpg: I think the difference here is that _unregistering_ after waking up is now a O(n) operation due to the find, so if you wake up m connections at once, it will be O(m * n) list operations done asymptotically. The current helio waiters unlink in O(1) time. What can be done, is:

1. Making the vector and std::list manually like helio does internally
2. `EventCount` supports subscribe with a callback based waiter, support can also be extended to other primitives

dranikpg: (2) is already used to wait for command execution

glevkovich: Thanks for the advice, I think I will go with option 2 (new sharted EventCount), it's faster / simpler and no need to modify helio at this stage. I'm working on it.

src/facade/dragonfly_connection.cc
```diff
@@ -2271,7 +2293,13 @@ bool Connection::IsReplySizeOverLimit() const {
 }
 
 bool Connection::ParseRedisBatch() {
-  // TODO: Handle pipeline backpressure
+  QueueBackpressure& qbp = GetQueueBackpressure();
+  if (qbp.IsPipelineBufferOverLimit(tl_facade_stats->conn_stats.pipeline_queue_bytes,
+                                    parsed_cmd_q_len_)) {
+    // Signal ParseLoop to stop. IoLoopV2 will drain before resuming.
+    DVLOG(2) << "Pipeline buffer over limit. Avoid parsing Redis batch.";
+    return false;
```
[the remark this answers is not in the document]
glevkovich: fixed

tests/dragonfly/connection_test.py
```diff
@@ -1325,29 +1325,151 @@ async def push_pipeline(bad_actor_client, size=1):
...
-async def test_pipeline_overlimit(df_server: DflyInstance):
-    client = df_server.client()
+@dfly_multi_test_args(
+    {"proactor_threads": 4, "pipeline_queue_limit": 10, "pipeline_buffer_limit": "1024"},
+    {
+        "proactor_threads": 4,
+        "pipeline_queue_limit": 10,
+        "pipeline_buffer_limit": "1024",
+        "experimental_io_loop_v2": "true",
+    },
```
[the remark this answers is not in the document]
glevkovich: for now i trst it manually. the final change will be done later.

src/facade/dragonfly_connection.cc
```diff
@@ -361,6 +362,18 @@ struct QueueBackpressure {
...
+  // which in turn calls io_event_.notify() to wake the V2 fiber.
+  // Registration and unregistration are O(1) via intrusive linked list.
+  util::fb2::EventCount v2_pipeline_backpressure_ec;
+
+  // Notifies both V1 waiters (pipeline_cnd) and V2 subscribers (v2_pipeline_backpressure_ec).
+  void NotifyPipelineWaiters() {
+    pipeline_cnd.notify_all();
+    v2_pipeline_backpressure_ec.notifyAll();
+  }
+
```
dranikpg, line 376: todo: unite somewhen in the future, but maybe even not if we abolish v1 fast

romange:
Lets hold on with merging it until we release 1.38

src/facade/dragonfly_connection.cc
```diff
@@ -2758,16 +2807,66 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
...
+          io_event_.await([this, under_limit_pred]() {
+            bool cmd_ready = parsed_head_ && parsed_head_->CanReply();
+            // We wake up if:
+            // 1. Memory is freed (under_limit_pred()) or we can free it ourselves (cmd_ready).
+            // 2. Control-plane messages need processing (!dispatch_q_.empty()).
+            // 3. The connection is terminating (io_ec_).
+            return under_limit_pred() || cmd_ready || !dispatch_q_.empty() || io_ec_;
+          });
+        }
+      }
```
dranikpg: maybe we can unite it somehow in the future like I did with the with the command waiter

dranikpg: also this function is really big now, maybe we can move some parts out of it

glevkovich:
> Lets hold on with merging it until we release 1.38

This PR is currently on hold, I won't merge it till we release 1.38.

src/facade/dragonfly_connection.cc
```diff
@@ -2758,16 +2813,80 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
...
+            << ", Connection parsed commands queue size: " << parsed_cmd_q_len_
+            << ", consider increasing pipeline_buffer_limit/pipeline_queue_limit";
+
+        // Subscribe persistently to the global backpressure EventCount so that when another
+        // connection frees memory (or CONFIG SET raises limits), our backpressure_waiter callback
+        // fires io_event_.notify(), waking this fiber. Must be persistent because
+        // io_event_.await()'s internal loop may re-sleep if the predicate is still false after the
+        // first notification. A one-shot subscription would be consumed on the first wake, leaving
+        // us "deaf" to future memory relief.
+        auto sub_key = qbp.v2_pipeline_backpressure_ec.subscribe_persistent(&backpressure_waiter);
```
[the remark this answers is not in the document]
glevkovich: dropped. compiler is smart enough to recognise RAII

src/facade/dragonfly_connection.cc
glevkovich: Noted. I won't do it in this PR.

src/facade/dragonfly_connection.cc
```diff
@@ -2758,16 +2813,80 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
...
+        qbp.NotifyPipelineWaiters();
+      }
+
+      // await block (backpressure)
+      // Re-check if pipeline buffer over limit after draining - ExecuteBatch/ReplyBatch may have
+      // freed memory. If still over limit, sleep to prevent busy-spin.
+      // Only park if this connection is actively contributing (parsed_cmd_q_len_ > 0).
+      // Connections with an empty queue must stay in the read loop.
+      bool post_over_limit =
+          (parsed_cmd_q_len_ > 0) &&
```
dranikpg, line 2858: My comment still applies. IoLoopV2 function now becomes veeeery long, it doesn't even fit on my screen. Maybe we can refactor it, at least in the future?

src/facade/dragonfly_connection.cc
glevkovich: Noted, will consider add a task.

src/facade/dragonfly_connection.cc
glevkovich: Noted, sure, wrote also in the previous comment. I will add a task. For sure not part of this PR.

glevkovich:
@dranikpg  added 2 issues/tasks:
https://github.com/dragonflydb/dataplane-private/issues/209
https://github.com/dragonflydb/dataplane-private/issues/208

PR #7021, opened 2026-03-30
chore: Extend QList::Node to support tiered storage

mkaruza:
* Extend FragmentRef with QList::Node and update related functions to support the new type.
* Extend ListWrapper to accept DbIndex, enabling statistics updates from tiered storage.
* During RDB save explicitly materialize nodes before serialization.

### DIFF
--- a/src/core/qlist.cc
+++ b/src/core/qlist.cc
@@ -459,15 +459,12 @@ size_t QList::DefragIfNeeded(PageUsage* page_usage) {
   return reallocated;
 }
 
-void QList::SetTieringParams(const TieringParams& params) {
-  tiering_params_ = make_unique<TieringParams>(params);
-}
-
 QList::QList(int fill, int compress)
     : fill_(fill),
       dict_learning_failed_(0),
       dict_compress_failed_(0),
       dict_bulk_finished_(0),
+      tiering_enabled_(0),
       compress_(compress),
       bookmark_count_(0) {
 }
@@ -480,10 +477,10 @@ QList::QList(QList&& other) noexcept
       dict_learning_failed_(other.dict_learning_failed_),
       dict_compress_failed_(other.dict_compress_failed_),
       dict_bulk_finished_(other.dict_bulk_finished_),
+      tiering_enabled_(other.tiering_enabled_),
       compress_(other.compress_),
       bookmark_count_(other.bookmark_count_),
-      num_offloaded_nodes_(other.num_offloaded_nodes_),
-      tiering_params_(std::move(other.tiering_params_)) {
+      num_offloaded_nodes_(other.num_offloaded_nodes_) {
   other.head_ = nullptr;
   other.len_ = other.count_ = 0;
   other.num_offloaded_nodes_ = 0;
@@ -503,9 +500,9 @@ QList& QList::operator=(QList&& other) noexcept {
     dict_learning_failed_ = other.dict_learning_failed_;
     dict_compress_failed_ = other.dict_compress_failed_;
     dict_bulk_finished_ = other.dict_bulk_finished_;
+    tiering_enabled_ = other.tiering_enabled_;
     compress_ = other.compress_;
     bookmark_count_ = other.bookmark_count_;
-    tiering_params_ = std::move(other.tiering_params_);
     num_offloaded_nodes_ = other.num_offloaded_nodes_;
     other.head_ = nullptr;
     other.len_ = other.count_ = other.num_offloaded_nodes_ = 0;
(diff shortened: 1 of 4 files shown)


src/core/qlist.cc
```diff
@@ -1008,20 +1044,50 @@ void QList::CompressByDepth(Node* node) {
   if (!in_depth && node) {
     malloc_size_ += TryCompress(node);
   }
-  /* At this point, forward and reverse are one node beyond depth */
   malloc_size_ += TryCompress(forward);
   malloc_size_ += TryCompress(reverse);
 }
 
+void QList::Materialize(Node* node) {
```
romange, line 1044: why ` QList::Materialize` duplicates `QList::AccessForReads` ? 

either remove or make one call another.

src/server/list_family.cc
```diff
@@ -168,7 +176,10 @@ class ListWrapper {
   }
 
  public:
-  template <typename T> explicit ListWrapper(T t) : impl_(std::forward<T>(t)) {
+  // TODO: passing current dbid of objecte. It could happen that object is moved to
```
romange:
```suggestion
  // TODO: passing current dbid of object. It could happen that object is moved to
```

src/server/list_family.cc
```diff
@@ -168,7 +176,10 @@ class ListWrapper {
...
+  // another db so this dbid will be incorrect. Refactor to support moving objects between dbs.
+  template <typename T>
+  explicit ListWrapper(DbIndex dbid, T t) : db_id_(dbid), impl_(std::forward<T>(t)) {
```
romange, line 179: why do you need dbid here?

src/core/qlist.cc
```diff
@@ -417,6 +418,14 @@ size_t QList::Node::GetLZF(void** data) const {
   return lzf->sz;
 }
 
+void QList::Node::SetExternal(size_t offset, uint32_t size) {
+  DCHECK(entry);
```
romange: `DCHECK(!io_pending);` ?

src/core/qlist.cc
```diff
@@ -1008,20 +1044,50 @@ void QList::CompressByDepth(Node* node) {
...
   DCHECK(node);
   stats.total_node_reads++;
-  if (node->offloaded) {
-    DCHECK(tiering_params_);
-    stats.onload_requests++;
-    num_offloaded_nodes_--;
-    node->offloaded = 0;
+
+  if (tiering_params_ && (node->offloaded || node->io_pending)) {
+    // If the nodes is io_pending. Just call the delete callback to cancel stashing.
```
romange:
```suggestion
    // If the node is io_pending. Just call the delete callback to cancel stashing.
```

src/core/qlist.cc
```diff
@@ -1008,20 +1044,50 @@ void QList::CompressByDepth(Node* node) {
...
+    if (node->io_pending && tiering_params_->delete_cb) {
+      tiering_params_->delete_cb(node);
+    }
+    // If the nodes is offlodaded. Load the data back to node.
```
romange:
```suggestion
    // If the node is offlodaded. Load the data back to node.
```

src/server/list_family.cc
mkaruza: We are updating database statistic from tiered storage.

src/server/list_family.cc
```diff
@@ -93,16 +96,21 @@ class ListWrapper {
...
-      ql->SetTieringParams(
-          QList::TieringParams{.node_depth_threshold = GetFlag(FLAGS_list_tiering_threshold)});
+
+    if (GetFlag(FLAGS_list_tiering_threshold) > 0 && EngineShard::tlocal()->tiered_storage()) {
+      QList::TieringParams params{.node_depth_threshold = GetFlag(FLAGS_list_tiering_threshold),
+                                  .offload_cb = nullptr,
+                                  .onload_cb = nullptr,
+                                  .delete_cb = nullptr};
+
+      ql->SetTieringParams(params);
```
dranikpg: I remember this part from the bigger PR. I would assume in the future that the callbacks would always be defined (never null) as you currently have checks for that in your code

src/core/qlist_test.cc
```diff
@@ -413,10 +413,30 @@ TEST_F(QListTest, DefragmentListpackCompressed) {
 
 TEST_F(QListTest, Tiering) {
   QList::stats.offload_requests = 0;
-  ql_.SetTieringParams(QList::TieringParams{.node_depth_threshold = 1});
```
dranikpg, line 416: I missed this from earlier PRs, but any reason this is not just thread local? Will there be any list specific state?

dranikpg: Given that you also add three std::functions to each

dranikpg: thread local function pointers

src/server/list_family.cc
mkaruza: Yes, probably. I'll change that in following PRs

src/core/qlist_test.cc
mkaruza: Removed callbacks and structure.

PR #7022, opened 2026-03-30
chore: Extend TieredStorage to support QList::Node

mkaruza:
* Add ListNodeId (dbid, ql*, node*) as a new tiering identity type alongside KeyRef; 
   introduce PendingId/ReadId variants so OpManager can handle both key-level 
   and node-level I/O uniformly.
* Update CancelStash / ReadInternal signatures to accept the generic 
   PendingId / ReadId types instead of (dbid, key).
* Add flag `--tiered_experimental_list_support` that is used to enable list node
   tiered storage functionality.
* Generalize tiering::Decoder::Upload to accept a void* target and introduce
   a ListNodeDecoder type.

### DIFF
--- a/src/server/tiered_storage.cc
+++ b/src/server/tiered_storage.cc
@@ -160,9 +160,6 @@ class TieredStorage::ShardOpManager : public tiering::OpManager {
     stats_.total_cancels++;
     QList::Node* node = reinterpret_cast<QList::Node*>(std::get<2>(id));
     node->io_pending = 0;
-    // If stashing failed we need to decrease offloaded nodes count.
-    QList* ql = reinterpret_cast<QList*>(std::get<1>(id));
-    ql->IncrementNumOffloadedNodes(-1);
   }
 
   void CancelStash(tiering::KeyRef id, size_t size) {
@@ -176,8 +173,6 @@ class TieredStorage::ShardOpManager : public tiering::OpManager {
   }
 
   void CancelStash(tiering::ListNodeId id) {
-    QList* ql = reinterpret_cast<QList*>(std::get<1>(id));
-    ql->IncrementNumOffloadedNodes(-1);
     CancelPending(id);
   }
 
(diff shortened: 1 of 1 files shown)


src/server/tiering/common.h
```diff
@@ -48,10 +48,17 @@ struct DiskSegment {
...
 
-// Two separate keyspaces are provided - one for strings, one for numeric identifiers.
+// Separate keyspaces are provided:
+// 1. PrimeValues with KeyRef
+// 2. Small bins with numeric identifiers
+// 3. List nodes with ListNodeId
 // Ids can be used to track auxiliary values that don't map to real keys (like a page index).
 // Specifically, we track page indexes when serializing small-bin pages with multiple items.
-using PendingId = std::variant<uintptr_t, KeyRef>;
+using PendingId = std::variant<uintptr_t, KeyRef, ListNodeId>;
```
dranikpg, line 59: Missed who changed small bins id to uintptr_t - it's a very confusing type

src/server/tiered_storage.cc
```diff
@@ -293,10 +345,28 @@ bool TieredStorage::ShardOpManager::NotifyFetched(const OwnedEntryId& id,
...
+    ql->AdjustMallocSize(segment.length);
+    ql->IncrementNumOffloadedNodes(-1);
+
+    DbTableStats* stats = GetDbTableStats(std::get<0>(*key));
+    stats->AddTypeMemoryUsage(OBJ_LIST, segment.length);
+
+    // We return false here, because we don't want to delete the value from storage yet.
+    // It will be done in onload_cb callback.
+    return false;
+  }
```
dranikpg: Hm... this flow doesn't fit into our existing one much, I will think of how to change it, but don't want to hold you back for now

src/server/tiered_storage.cc
```diff
@@ -293,10 +333,21 @@ bool TieredStorage::ShardOpManager::NotifyFetched(const OwnedEntryId& id,
...
-    Defragment(segment, bdecoder->slice);
-    return true;  // delete
+  if (const auto* key = std::get_if<tiering::ListNodeId>(&id); key) {
+    DbIndex db_id = std::get<0>(*key);
+    QList::Node* node = reinterpret_cast<QList::Node*>(std::get<2>(*key));
+    ++stats_.total_uploads;
+    decoder->Upload(node);
+    RecordDeleted(node, segment.length, GetDbTableStats(db_id));
+    return true;
+  }
```
dranikpg, line 343: it's okay this way, maybe we'll unite it somewhen with the code below. You should respect the conditions from below

```
bool should_upload = metrics.modified;
  should_upload |= (ts_->UploadBudget() > int64_t(metrics.estimated_mem_usage)) &&
                   !SliceSnapshot::IsSnaphotInProgress();
```

later

PR #7023, opened 2026-03-30
feat(tiering): Experimental list node tiering

mkaruza:
Decoupline QList from TieredStorage implementation details via
TieringParams structure callbacks.

- Introduce TieringParams struct holding function callbacks and offload state
- Add Node::Upload() to restore an externally-loaded node back into memory
- Add QList::SetDbIndex() to handle db reassignment, materializing offloaded
  nodes when tiering is active
- Add StashListNode / ReadTieredListNode in TieredStorage
- Add tiered_storage_test.cc coverage for list node offload/reload lifecycle
- Add tiering_test.py integration tests for list tiering behavior

### DIFF
--- a/src/server/tiered_storage.h
+++ b/src/server/tiered_storage.h
@@ -199,6 +199,10 @@ inline TieredStorage::TResult<std::string> ReadTieredString(DbIndex dbid, std::s
       dbid, key, value, [](std::string_view val) { return std::string(val); }, ts);
 }
 
+TieredStorage::TResult<bool> ReadTieredListNode(DbIndex dbid, QList* ql, QList::Node* node,
+                                                const tiering::DiskSegment& segment,
+                                                TieredStorage* ts);
+
 // Reads offloaded value, and applies modifications on it and return generic result from callback.
 // Unlike with immutable Reads - the modified value will be uploaded back to memory.
 // This is handled by OpManager when modf completes.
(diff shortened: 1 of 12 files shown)


tests/dragonfly/tiering_test.py
```diff
@@ -247,3 +247,74 @@ async def test_tiered_replication_with_hashes(
...
+        proactor_threads=2,
+        maxmemory="512MB",
+        tiered_prefix="/tmp/tiered/backing_master_list",
+        tiered_offload_threshold="1.0",
+        tiered_storage_write_depth=1500,
+        tiered_experimental_cooling="false",
+        list_max_listpack_size=1,
+        list_tiering_threshold=2,
+        tiered_experimental_list_support="true",
+    )
```
dranikpg, line 266: I think there might be no reason to fully repeat the test and we can just mix it in the already existing test

src/server/tiered_storage.h
```diff
@@ -211,6 +215,11 @@ TieredStorage::TResult<T> ModifyTiered(DbIndex dbid, std::string_view key, const
 void StashPrimeValue(DbIndex dbid, std::string_view key, PrimeValue* pv, TieredStorage* ts,
                      BackPressureFuture* backpressure);
 
+// Stash list node if it meets criteria.
+// Returns true if stash was initiated, false otherwise.
+bool StashListNode(DbIndex dbid, QList* ql, QList::Node* node, TieredStorage* ts,
+                   BackPressureFuture* backpressure);
+
```
dranikpg, line 222: They're only used from list_family.cc - why not keep them there?

tests/dragonfly/tiering_test.py
dranikpg: The more tests, the more difficult it's to manage it all

mkaruza: I know, but i wanted to add test to replicate tiered list only - so if we get regression on list tiering we would likely see this test failing.

src/server/tiered_storage.h
mkaruza: It make sense that we group different `Stash*` helper functions together - just because of that

tests/dragonfly/tiering_test.py
dranikpg: can be a different set of parameters

dranikpg (approved):

PR #7024, opened 2026-03-30
chore: enable test_network_disconnect_during_migration

kostasrim:


BorysTheDev:
I've already enabled it in my next PR, so close it

PR #7025, opened 2026-03-30
chore: pending chunks for disk backed queue

kostasrim:
* PushAsync with pending list
* Add cancellation 
* Tests

### DIFF
--- a/src/facade/disk_backed_queue.cc
+++ b/src/facade/disk_backed_queue.cc
@@ -14,7 +14,9 @@
 
 #include <cerrno>
 #include <cstring>
+#include <deque>
 #include <string>
+#include <utility>
 
 #include "base/flags.h"
 #include "facade/facade_types.h"
@@ -22,6 +24,7 @@
 #include "util/fibers/uring_proactor.h"
 
 using facade::operator""_MB;
+using facade::operator""_KB;
 
 ABSL_FLAG(std::string, disk_backpressure_folder, "/tmp/",
           "Folder to store disk-backed connection backpressure");
@@ -30,16 +33,32 @@ ABSL_FLAG(size_t, disk_backpressure_file_max_bytes, 50_MB,
           "Maximum size of the backing file. When max size is reached, connection will "
           "stop offloading backpressure to disk and block on client read.");
 
+ABSL_FLAG(size_t, disk_backpressure_drain_threshold, 16_KB,
+          "When remaining bytes in the disk-backed queue drop below this "
+          "threshold, queue is in draining phase (backpressure drains). "
+          "Defaults to 2x the max chunk size (16KB).");
+
 namespace facade {
 
 struct DiskBackedQueue::Impl {
   std::unique_ptr<util::fb2::LinuxFile> file;
   size_t write_offset = 0;
-  size_t total_backing_bytes = 0;
+  size_t total_backing_bytes = 0;  // bytes actually on disk, available to pop
+  size_t queued_bytes = 0;         // bytes in write_queue not yet written to disk
   size_t next_read_offset = 0;
   // Tracks how far into the file holes have been punched (always 4096-aligned).
   size_t punch_offset = 0;
   size_t in_flight_callbacks = 0;
+  bool push_in_flight = false;
+  bool pop_in_flight = false;
+  // Set by Cancel() or on I/O error — no further async operations are submitted.
+  bool cancelled = false;
+
+  struct PendingWrite {
+    Chunk chunk;
+    AsyncPushCallback cb;
+  };
+  std::deque<PendingWrite> write_queue;
 
   // Punch holes over the aligned region we have fully read past so the OS can reclaim pages.
   void MaybePunchHole() {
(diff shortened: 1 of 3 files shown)


src/facade/disk_backed_queue.cc
```diff
@@ -30,16 +33,32 @@ ABSL_FLAG(size_t, disk_backpressure_file_max_bytes, 50_MB,
           "Maximum size of the backing file. When max size is reached, connection will "
           "stop offloading backpressure to disk and block on client read.");
 
+ABSL_FLAG(size_t, disk_backpressure_drain_threshold, 16_KB,
```
kostasrim, line 36: @romange I added this watermark flag to allow shifting disk backpressure back to the socket. This allows a smooth transition from disk to memory without ping ponging the disk when draining. That way socket reads are deferred while the last few bytes of the disk queue are being drained by the connection

src/facade/disk_backed_queue.h
```diff
@@ -17,26 +18,58 @@ namespace facade {
...
   void PopAsync(io::MutableBytes out, AsyncPopCallback cb);
 
-  // Check if backing file is empty, i.e. backing file has 0 bytes.
+  // Check if backing store (on-disk bytes) is empty.
   bool Empty() const;
 
+  // Cancel all pending (not-yet-submitted) writes and mark the queue as cancelled.
+  // In-flight write/read CQEs will complete with operation_canceled.
+  // Caller must wait for in-flight operations to complete before destroying the queue.
+  void Cancel();
```
kostasrim, line 71: I can introduce a blocking `Shutdown` but `Cancel()` + wait if in-flight is more than enough for the users of this class

src/facade/disk_backed_queue.cc
```diff
@@ -96,44 +159,69 @@ bool DiskBackedQueue::Empty() const {
...
+bool DiskBackedQueue::IsPushInFlight() const {
+  return impl_->push_in_flight;
+}
+
+bool DiskBackedQueue::HasEnoughBackingSpaceFor(size_t bytes) const {
+  return (bytes + impl_->total_backing_bytes + impl_->queued_bytes) < max_backing_size_;
+}
+
+void DiskBackedQueue::PushAsync(Chunk chunk, AsyncPushCallback cb) {
+  DCHECK(!chunk.data.empty());
```
kostasrim, line 189: I did not introduce a fiber similar to let's say the journal streamer because it's not really needed. AsyncWrites on pending entries can continue from the async cb. Also we don't need any of the extra complexity of the context

src/facade/disk_backed_queue.cc
```diff
@@ -96,44 +159,69 @@ bool DiskBackedQueue::Empty() const {
...
+
+  impl_->queued_bytes += chunk.data.size();
+  impl_->write_queue.push_back({std::move(chunk), std::move(cb)});
+  impl_->MaybeFlushQueue();
 }
 
 void DiskBackedQueue::PopAsync(io::MutableBytes out, AsyncPopCallback cb) {
+  DCHECK(!impl_->pop_in_flight);
```
kostasrim, line 197: @romange we have a pending write list on the write path. we could similarly do the same with read path. That way CPU can be busy parsing while the disk is reading the next chunk.

I leave it as is for now since performance is not the main goal here

src/facade/disk_backed_queue.cc
```diff
@@ -30,16 +33,32 @@ ABSL_FLAG(size_t, disk_backpressure_file_max_bytes, 50_MB,
...
   size_t punch_offset = 0;
   size_t in_flight_callbacks = 0;
+  bool push_in_flight = false;
+  bool pop_in_flight = false;
+  // Set by Cancel() or on I/O error — no further async operations are submitted.
+  bool cancelled = false;
+
+  struct PendingWrite {
+    Chunk chunk;
+    AsyncPushCallback cb;
```
kostasrim, line 59: We do not really need to store `cb` since it's stateless so maybe we can pass it during the construction of the queue and we don;t have to store it each time

kostasrim:
toy e2e https://github.com/dragonflydb/dragonfly/pull/6929

kostasrim:
@romange before I finalize anything let's make sure we agree on the direction:

1. No extra fiber, let the async cb progress the state machine --> simpler than streamer
2. Pending list for Push
3. I did not add a pending list for Pop but can be done (since disk reads are non-cpu) 
4. Simple bool cancellation without explicit Wait() -- caller *must* wait the current in-flight pop/push `before destroying the queue`

src/facade/disk_backed_queue.cc
romange: `When remaining bytes in the disk-backed queue drop below this` - what does it mean remaining bytes?

src/facade/disk_backed_queue.h
```diff
@@ -17,26 +18,58 @@ namespace facade {
...
 
+  // Returns true if the queue has any pending work: data on disk, chunks
+  // waiting to be written, a write or read in-flight.
+  bool IsActive() const;
+
+  // Watermark drain.
+  // Returns true when remaining bytes have dropped below the drain threshold.
+  // The caller should block new socket reads to allow a clean drain-to-memory
+  // transition (TCP backpressure builds during this window).
+  bool IsDraining() const;
```
romange, line 42: Is "draining" the right term here? 
what the queue actually does during this state?

PR #7027, opened 2026-03-30
feat(server): Serializer bucket states

### LINKED ISSUE #6830: shard-serialization P1.3: introduce bucket-level completion state machine

dranikpg:
Add bucket dependency counter to track "async" serialization operations (like big value streaming or tiered loads). Resolves #6830

Why did I choose just a latch over a `struct BucketState { vector<DelayedEntry> }`?

Because multiple fibers can serialize delayed entries, we need a mechanism to not serialize the same value concurrently (i.e. pop before wait as before). This means that we need a signalling mechanism to wait for the last popped item to finish serializing (as the vector can be empty), which is either a bool or a counter. And then we also need a wakeup primitive to unblock waiters -> we need just a latch or condvar.

What is more, we need some kind of latch or condvar for big value streaming. We can unite all that into a unified notion of "dependency"  and use a single latch.


dranikpg:
augment review

src/server/serializer_base.cc
```diff
@@ -106,45 +136,21 @@ void SerializerBase::UnregisterChangeListener() {
...
 bool SerializerBase::ProcessBucket(DbIndex db_index, PrimeTable::bucket_iterator it,
                                    bool on_update) {
   std::lock_guard guard(big_value_mu_);
 
-  // Check if this bucket should be serialized
-  if (!ShouldProcessBucket(it)) {
+  // Check if this bucket is stale
+  if (it.is_done() || it.GetVersion() >= snapshot_version_) {
+    stats_.buckets_skipped++;
+
```
dranikpg, line 155: here

src/server/CMakeLists.txt
```diff
@@ -165,7 +165,6 @@ helio_cxx_test(cluster/cluster_config_test dfly_test_lib LABELS DFLY)
 helio_cxx_test(cluster/cluster_family_test dfly_test_lib LABELS DFLY)
 helio_cxx_test(acl/acl_family_test dfly_test_lib LABELS DFLY)
 helio_cxx_test(engine_shard_set_test dfly_test_lib LABELS DFLY)
-helio_cxx_test(serializer_base_test dfly_test_lib LABELS DFLY)
```
romange, line 168: why is this deleted?

romange:
is it synced with latest?

src/server/CMakeLists.txt
dranikpg: because the test uses a bucket state machine that I completely replaced

dranikpg: I can add some other tests

dranikpg:
> is it synced with latest?

Yes

src/server/CMakeLists.txt
dranikpg: ah, my bad, did not delete the file

src/server/serializer_base.cc
```diff
@@ -44,44 +73,45 @@ void DelayedEntryHandler::ProcessDelayedEntries(bool force, BucketIdentity flush
...
     SerializeFetchedEntry(*entry, pv);
+
+    deps_.Decrement(it->first);
+    delayed_entries_.erase(it++);
+  };
+
+  // Flush all entries of bucket
+  if (flush_bucket) {
+    auto range = delayed_entries_.equal_range(flush_bucket);
+    for (auto it = range.first; it != range.second;) {
```
romange, line 97: nit: why not increment in the "for" loop, like a true professional?

Issue #4251, opened 2024-12-04
Potential bug in entrypoint script


applike-ss:
**Describe the bug**
When i start my dragonfly test container via the following command the io threads are correctly set:
```
# set proactor threads to 1
docker run -it --rm ghcr.io/dragonflydb/dragonfly:v1.25.4-ubuntu --proactor_threads=1
# > I20241204 09:13:18.126736     1 proactor_pool.cc:147] Running 1 io threads

# set proactor threads to 2
docker run -it --rm ghcr.io/dragonflydb/dragonfly:v1.25.4-ubuntu --proactor_threads=2
# > I20241204 09:07:52.068485     1 proactor_pool.cc:147] Running 2 io threads
```

When i now try to achieve the same in my k3d test kubernetes cluster, it does always set the proactor threads to the amount of threads available to the pod. This would be fine if i currently did have another issue where limiting the pod resources isn't working correctly.

snippet from the pod:
```
  containers:
    - name: dragonfly
      image: ghcr.io/dragonflydb/dragonfly:v1.25.4
      args:
        - '--proactor_threads=1'
```

**To Reproduce**
1. `k3d cluster create`
2. Install dragonfly operator as per documentation `kubectl apply -f https://raw.githubusercontent.com/dragonflydb/dragonfly-operator/main/manifests/dragonfly-operator.yaml`
3. Create dragonfly cluster via Dragonfly resource with args passed
4. See log message showing a different amount of proactor threads

Demo Resource:
```yaml
apiVersion: dragonflydb.io/v1alpha1
kind: Dragonfly
metadata:
  name: dragonfly-sample
  namespace: default
spec:
  image: ghcr.io/dragonflydb/dragonfly:v1.25.4
  args:
    - '--maxmemory=512M'
    - '--proactor_threads=1'
  replicas: 2
  resources:
    limits:
      cpu: 2
      memory: 768Mi
    requests:
      cpu: 2
      ephemeral-storage: 1Gi
      memory: 768Mi
```

**Expected behavior**
Setting proactor threads (potentially other flags are also affected?) should work the same, no matter whether using docker or k8s/containerd

**Screenshots**
-

**Environment (please complete the following information):**
 - OS: Darwin + Linux
 - Kernel: 6.10.14-linuxkit
 - Containerized?: Docker and Kubernetes
 - Dragonfly Version: 1.25.4

**Reproducible Code Snippet**
see above at "To reproduce"

**Additional context**
-

[...]

PR #7030, opened 2026-03-31
fix(proactor_threads): fix that proactor_threads is not respected in k8s env

starek4:
Fixing issue: https://github.com/dragonflydb/dragonfly/issues/4251

When a pod has CPU resource limits set, UpdateResourceLimitsIfInsideContainer derives a thread count from the cgroup CPU quota and passes it directly to ProactorPool. Because helio's ProactorPool constructor only checks FLAGS_proactor_threads when its pool_size argument is 0, a non-zero cgroup-derived value silently overrides whatever the user set via --proactor_threads (or DFLY_proactor_threads).                                                                                                                                                              
                                                                                                                                                                                                                                                                                         
This caused pods to start with a thread count dictated by the CPU limit rather than the explicitly configured value, often failing the maxmemory >= threads * 256MB check on startup.                                                                                                  
   
Changes:                                                                                                                                                                                                                                                                               
  - After cgroup detection, reset max_available_threads to 0 when --proactor_threads is explicitly set (non-zero), so ProactorPool uses the flag value instead
  - Handles both --proactor_threads CLI flag and DFLY_proactor_threads environment variable, since ParseFlagsFromEnv() runs before this check                                                                                                                                            
                                                                                                                                             
Behavior:                                                                                                                                                                                                                                                                              
  - No CPU limits → unchanged (helio auto-detects CPUs)                                                                                                                                                                                                                                  
  - CPU limits set, no --proactor_threads → unchanged (cgroup value used)                                                                                                                                                                                                                
  - CPU limits set + --proactor_threads=N → fixed (flag is now respected) 
  
Tests:                                                                                                                                                                                                                                                                                 
  - tests/dragonfly/proactor_threads_test.py — three new integration tests:                                                                                                                                                                                                              
    - test_proactor_threads_flag_is_respected → "verifies --proactor_threads=N results in exactly N threads"                                                                                                                                                                               
    - test_proactor_threads_env_var_is_respected → "verifies DFLY_proactor_threads=N env var (the Kubernetes env: equivalent) is respected the same way"                                                                                                                                   
    - test_proactor_threads_flag_overrides_env_var → "verifies CLI flag takes priority over the env var, consistent with existing WasPresentOnCommandLine behaviour"

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -187,6 +187,82 @@ jobs:
           name: regression_logs
           path: /tmp/failed/*
 
+      - name: Upload dragonfly binary for cgroup test
+        if: matrix.container == 'ubuntu-dev:24' && matrix.build-type == 'Release' && matrix.sanitizers == 'NoSanitizers' && matrix.compiler.cxx == 'g++'
+        uses: actions/upload-artifact@v7
+        with:
+          name: dragonfly-binary
+          path: ${{ github.workspace }}/build/dragonfly
+          retention-days: 1
+
+  cgroup-thread-detection:
+    runs-on: ubuntu-latest
+    needs: [build]
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: actions/download-artifact@v7
+        with:
+          name: dragonfly-binary
+          path: build
+
+      - name: Install redis-tools
+        run: sudo apt-get update && sudo apt-get install -y redis-tools
+
+      - name: Prepare cgroup test image
+        run: |
+          chmod +x build/dragonfly
+          # Verify the binary's runtime dependencies and install them in a local image.
+          # The binary is built in ubuntu-dev:24 and links dynamically against Boost,
+          # which is not present in the minimal ubuntu:24.04 image.
+          echo "Runtime deps of dragonfly binary:"
+          ldd build/dragonfly
+          docker build -t df-cgroup-test - <<'EOF'
+          FROM ubuntu:24.04
+          RUN apt-get update -qq && \
+              apt-get install -y --no-install-recommends libboost-context1.83.0 libssl3 && \
+              rm -rf /var/lib/apt/lists/*
+          EOF
+
+      - name: Test cgroup CPU auto-detection
+        run: |
+          wait_for_server() {
+            local port=$1
+            local cid=$2
+            for i in $(seq 1 30); do
+              sleep 1
+              redis-cli -p "$port" PING 2>/dev/null | grep -q PONG && return 0
+            done
+            echo "ERROR: dragonfly on port $port did not start within 30s"
+            docker logs "$cid" 2>&1 || true
+            docker rm -f "$cid" 2>/dev/null || true
+            return 1
+          }
+
+          # Scenario 1: --cpus=2, no flag -> should auto-detect 2 threads
+          CID=$(docker run --rm -d --cpus=2 --security-opt seccomp=unconfined \
+            -v "$PWD/build/dragonfly:/dragonfly" -p 16380:6379 df-cgroup-test \
+            /dragonfly --port 6379 --maxmemory 1G --dbfilename "" --noversion_check)
+
+          wait_for_server 16380 "$CID"
+
+          THREADS=$(redis-cli -p 16380 INFO server | grep -oP 'thread_count:\K\d+')
+          docker rm -f "$CID" 2>/dev/null || true
+          echo "Scenario 1: expected 2, got $THREADS"
+          [ "$THREADS" -eq 2 ]
+
+          # Scenario 2: --cpus=2 + --proactor_threads=4 -> flag must win
+          CID=$(docker run --rm -d --cpus=2 --security-opt seccomp=unconfined \
+            -v "$PWD/build/dragonfly:/dragonfly" -p 16381:6379 df-cgroup-test \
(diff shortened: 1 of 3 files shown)


starek4:
@dranikpg would you be so kind to fill in some reviewers? I am not so familiar with GitHub, and it seems I cannot add any after the PR is created. Thanks!

romange:
@starek4  thanks for the contribution. I am trying to understand - the tests are not related to the fix, correct?

They are just "fixing" the current behavior with cli args and the env vars?

romange:
Also, can I reproduce the problem locally with `docker/podman` or  I need k8s environment for that? 

(Sorry, not a k8s expert).

starek4:
@romange, you can definitely reproduce the issue locally (and also the fix), let me show you how:

1. Create a cgroup with a CPU limit (2CPUs, 200000us quota / 100000us period)
  ```
  sudo mkdir -p /sys/fs/cgroup/dragonfly-test
  echo "200000 100000" | sudo tee /sys/fs/cgroup/dragonfly-test/cpu.max
  ```

2. Check your machine has more than 2 threads (otherwise the fix is a no-op)
  `nproc`

3. Run dragonfly without the flag (should be limited to 2 threads by cgroup)
  ```
  sudo cgexec -g cpu:dragonfly-test ./build-dbg/dragonfly --alsologtostderr 2>&1 | grep -i "thread\|proactor"
  # Before and after fix: thread_count:2
  ```

4. Run dragonfly WITH --proactor_threads=6 (should override the cgroup limit)
  ```
  sudo cgexec -g cpu:dragonfly-test ./build-dbg/dragonfly --proactor_threads=6 --alsologtostderr
  # Before fix: thread_count:2
  # After fix: thread_count:6
  ```
5. Test env var path too
  ```
  sudo cgexec -g cpu:dragonfly-test env DFLY_proactor_threads=5 ./build-dbg/dragonfly --alsologtostderr
  # Before fix: thread_count:2
  # After fix: thread_count:5
  ```

6. Cleanup
  `sudo rmdir /sys/fs/cgroup/dragonfly-test`

And about the tests, you are right. It is only about passing the arguments.

romange:
Thanks! Let me few days to review it (currently on Passover vacation). Meanwhile can you pls edit your comment above and add what is the actual behavior before the fix for each of the cases you wrote down?

starek4:
Enjoy your vacation! I have added the before and after fix behaviour.

romange:
The fix is correct and minimal. When `FLAGS_proactor_threads > 0`, resetting `max_available_threads = 0` makes `ProactorPool` fall back to the flag value.

However, the tests don't validate the actual bug scenario. They run on a normal machine where `max_available_threads` is already 0 (no cgroup limits), so they exercise the path that was never broken.

### Suggested: integration test with real cgroup limits

The existing CI runs pytest inside Docker containers where `/sys/fs/cgroup` is read-only and Docker isn't available. Add a separate job that runs directly on the runner and uses `docker run --cpus=N` to verify dragonfly's behavior under real cgroup limits. Maybe something like this:

```yaml
cgroup-thread-detection:
  runs-on: ubuntu-latest
  needs: [build]
  steps:
    - uses: actions/checkout@v6

    # Reuse the binary built by the build job (requires uploading it as an artifact there)
    - uses: actions/download-artifact@v7
      with:
        name: dragonfly-binary
        path: build

    - name: Test cgroup CPU auto-detection
      run: |
        chmod +x build/dragonfly

        # Scenario 1: --cpus=2, no flag -> should auto-detect 2 threads
        CID=$(docker run --rm -d --cpus=2 --security-opt seccomp=unconfined \
          -v "$PWD/build/dragonfly:/dragonfly" -p 16380:6379 ubuntu:24.04 \
          /dragonfly --port 6379 --maxmemory 1G --dbfilename "" --noversion_check)
        #
        # Wait for PING to return PONG 
        #
        THREADS=$(redis-cli -p 16380 INFO server | grep -oP 'thread_count:\K\d+')
        docker rm -f "$CID"
        echo "Scenario 1: expected 2, got $THREADS"
        [ "$THREADS" -eq 2 ]

        # Scenario 2: --cpus=2 + --proactor_threads=4 -> flag must win
        CID=$(docker run --rm -d --cpus=2 --security-opt seccomp=unconfined \
          -v "$PWD/build/dragonfly:/dragonfly" -p 16381:6379 ubuntu:24.04 \
          /dragonfly --port 6379 --proactor_threads 4 --maxmemory 1G --dbfilename "" --noversion_check)
        #
        # Wait for PING to return PONG
        #
        THREADS=$(redis-cli -p 16381 INFO server | grep -oP 'thread_count:\K\d+')
        docker rm -f "$CID"
        echo "Scenario 2: expected 4, got $THREADS"
        [ "$THREADS" -eq 4 ]
```

### Minor suggestion

Add a `LOG(INFO)` when the flag overrides the cgroup value -- helps debugging in production:
```cpp
if (absl::GetFlag(FLAGS_proactor_threads) > 0) {
  LOG(INFO) << "Using proactor_threads=" << absl::GetFlag(FLAGS_proactor_threads)
            << " (overriding cgroup-derived " << max_available_threads << ")";
  max_available_threads = 0;
}
```

romange:
@starek4  you need to install redis-tools or valkey-tools to be able to ping the server

starek4:
@romange yep, no worries, will do it when I have time.

starek4:
@romange it should work now, tested on my forked actions

romange:
@starek4 please squash and sign your commits

starek4:
@romange done

romange:
thank you!

PR #7034, opened 2026-03-31
feat(rdb_saver): Flush each SBF chunk during serialization

mkaruza:
Call PushToConsumerIfNeeded after writing each chunk so the buffer is drained incrementally.

### DIFF
--- a/tests/dragonfly/replication_test.py
+++ b/tests/dragonfly/replication_test.py
@@ -4200,8 +4200,7 @@ async def test_sbf_chunked_replication(df_factory: DflyInstanceFactory):
     test_rng = random.Random(random_seeed)
 
     random_items = [f"item:{i}" for i in test_rng.sample(range(1_000_000), 20_000)]
-    for item in random_items:
-        await c_master.execute_command("BF.ADD", "bf", item)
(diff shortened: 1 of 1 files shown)


tests/dragonfly/replication_test.py
```diff
@@ -4169,6 +4170,69 @@ async def test_sbf_chunked_replication_over_4gb(df_factory: DflyInstanceFactory)
...
+    bf_size = await c_master.execute_command("MEMORY USAGE", "bf")
+    assert bf_size > 2**30, f"Bloom filter should be >1GB, got {bf_size}"
+
+    # Fix set to have reproducible test results and log the seed for debugging
+    random_seeed = random.getrandbits(64)
+    logging.info(f"Using random seed {random_seeed} for test reproducibility")
+    test_rng = random.Random(random_seeed)
+
+    random_items = [f"item:{i}" for i in test_rng.sample(range(1_000_000), 20_000)]
+    for item in random_items:
```
romange: it's gonna be slow. use pipeline for this. or [BF](https://www.dragonflydb.io/docs/command-reference/bloom-filter/bf.madd)

PR #7035, opened 2026-03-31
fix: prevent master connection closing during await_synced_all

BorysTheDev:
The bug was in await_synced(): it calls c_master.close() inside the loop body.  await_synced_all() passes the same c_master to await_synced() for each of the replicas.  After the first replica check succeeds, c_master is closed. For other replicas, c_master.set(key, "dummy") operates on a closed connection — the key either isn't set on the master, or the reconnection race means it doesn't propagate through the replication stream in time, causing the 30-second timeout.

### DIFF
--- a/tests/dragonfly/redis_replication_test.py
+++ b/tests/dragonfly/redis_replication_test.py
@@ -27,8 +27,6 @@ async def await_synced(c_master: aioredis.Redis, c_replica: aioredis.Redis, dbco
             await asyncio.sleep(1)
 
             timeout -= 1
-        await c_master.close()
-        await c_replica.close()
         assert timeout > 0, "Timeout while waiting for replica to sync"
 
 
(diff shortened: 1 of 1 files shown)


tests/dragonfly/redis_replication_test.py
```diff
@@ -27,8 +27,6 @@ async def await_synced(c_master: aioredis.Redis, c_replica: aioredis.Redis, dbco
             await asyncio.sleep(1)
 
             timeout -= 1
-        await c_master.close()
-        await c_replica.close()
```
kostasrim, line 31: Isn't that just a wasteful close ? As far as I remember and know, `execute_command()` or `set()` works even after calling `close()` because the library will automatically reconnect internally. You can try this as well with a simple 2 line test :)

BorysTheDev: I agree with you, Kostas, but Claude's code thinks this is the reason for the failed tests, so I want to prove it. Anyway, these changes are completely safe and for now I don't have other ideas. if it doesn't fix the problem I will get new logs and proceed investigation

kostasrim: Then `repeat` action would help here ? That way regression tests will stay green if it doesn't work

kostasrim: https://github.com/dragonflydb/dragonfly/actions/workflows/repeat-tests.yml

PR #7040, opened 2026-04-01
chore: use ranges and concepts

kostasrim:
* replace std::algorithm with ranges::algorithm
* use concepts instead of SFINAE

### DIFF
--- a/src/server/cluster/cluster_defs.h
+++ b/src/server/cluster/cluster_defs.h
@@ -24,13 +24,7 @@ struct SlotRange {
   SlotId start = 0;
   SlotId end = 0;
 
-  bool operator==(const SlotRange& r) const noexcept {
-    return start == r.start && end == r.end;
-  }
-
-  bool operator<(const SlotRange& r) const noexcept {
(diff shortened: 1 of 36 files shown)


kostasrim:
@dranikpg need to also extend some of our types so they can satisfy the `strictier` concepts for the range iterators <begin, Sentinel> such that we can do std::sort() instead of std::sort(c, std::less) for example (requires the full <> operators from the type)

src/server/cluster/cluster_defs.h
```diff
@@ -24,13 +24,7 @@ struct SlotRange {
...
   SlotId end = 0;
 
-  bool operator==(const SlotRange& r) const noexcept {
-    return start == r.start && end == r.end;
-  }
-
-  bool operator<(const SlotRange& r) const noexcept {
-    return start < r.start || (start == r.start && end < r.end);
-  }
+  std::strong_ordering operator<=>(const SlotRange&) const noexcept = default;
```
kostasrim, line 27: This is beautiful.

PR #7044, opened 2026-04-01
fix(search): emit complete field params in FT.INFO and BuildRestoreCommand

### LINKED ISSUE #7013: FT.INFO omits vector field parameters in attributes

vyavdoshenko:
Fix FT.INFO to return complete field parameters and fix BuildRestoreCommand serialization issues.

Fixes #7013

- Problem 1: Parse and store `TYPE` (FLOAT32/FLOAT64/BFLOAT16) for VECTOR fields
- Problem 2: Emit all VECTOR params in FT.INFO (algorithm, data_type, dim, distance_metric, M, ef_construction)
- Problem 3: Emit TAG params in FT.INFO (SEPARATOR, CASESENSITIVE)
- Problem 4: Emit TEXT param in FT.INFO (WITHSUFFIXTRIE)
- Problem 5: Already fixed before
- Problem 6: Return `stopwords_list` as a top-level field in FT.INFO when explicitly configured
- Problem 7: Intentionally skipped - `default_score` is deprecated and unused by client libraries

Additionally, fixes two pre-existing bugs in BuildRestoreCommand:
- TEXT `WITH_SUFFIXTRIE` was serialized with the wrong name (underscore), silently dropped on restore
- TAG `WITHSUFFIXTRIE` was not serialized at all, lost on restore/replication

### DIFF
--- a/src/server/search/search_family.cc
+++ b/src/server/search/search_family.cc
@@ -39,6 +39,8 @@
 #include "server/transaction.h"
 #include "src/core/overloaded.h"
 
+namespace rng = std::ranges;
+
 ABSL_FLAG(bool, search_reject_legacy_field, true, "FT.AGGREGATE: Reject legacy field names.");
 ABSL_FLAG(bool, cluster_search, false,
           "Enable search commands for cross-shard search. turned off by default for safety.");
@@ -213,6 +215,20 @@ ParsedSchemaField ParseVector(CmdArgParser* parser) {
   if (vector_params.dim == 0) {
     return CreateSyntaxError("Knn vector dimension cannot be zero"sv);
   }
+
+  // Validate that the initial allocation (capacity * (dim+1) floats) cannot
+  // overflow size_t or request an unreasonable amount of memory.  Without this
+  // check FlatVectorIndex::FlatVectorIndex() would throw std::bad_alloc,
+  // leaving a half-initialised index registered in ShardDocIndices.
+  static constexpr size_t kMaxFlatBufEntries = size_t{1} << 30;  // ~4 GiB of floats
+  if (vector_params.dim >= kMaxFlatBufEntries) {
+    return CreateSyntaxError("Vector index initial allocation is too large"sv);
+  }
+  size_t dim_plus1 = vector_params.dim + 1;
+  if (vector_params.capacity > kMaxFlatBufEntries / dim_plus1) {
+    return CreateSyntaxError("Vector index initial allocation is too large"sv);
+  }
+
   return std::make_pair(search::SchemaField::VECTOR, vector_params);
 }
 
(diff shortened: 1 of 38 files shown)


src/server/search/search_family.cc
```diff
@@ -111,6 +111,8 @@ search::SchemaField::VectorParams ParseVectorParams(CmdArgParser* parser) {
     } else if (parser->Check("INITIAL_CAP", &params.capacity)) {
     } else if (parser->Check("M", &params.hnsw_m)) {
     } else if (parser->Check("EF_CONSTRUCTION", &params.hnsw_ef_construction)) {
+    } else if (parser->Check("TYPE")) {
+      params.data_type = absl::AsciiStrToUpper(parser->Next<string_view>());
```
dranikpg, line 117: it's actually true, you will silently accept any datatype and act as if we support it, yet we support only float32

dranikpg: it's misleading at least

src/server/search/search_family.cc
```diff
@@ -1641,7 +1645,38 @@ void CmdFtInfo(CmdArgList args, CommandContext* cmd_cntx) {
...
-    if (field_info.type == search::SchemaField::NUMERIC) {
+    if (field_info.type == search::SchemaField::VECTOR) {
+      auto& vparams = std::get<search::SchemaField::VectorParams>(field_info.special_params);
+      info.emplace_back("algorithm");
+      info.emplace_back(vparams.use_hnsw ? "HNSW" : "FLAT");
+      info.emplace_back("data_type");
+      info.emplace_back(vparams.data_type);
+      info.emplace_back("dim");
+      info.emplace_back(std::to_string(vparams.dim));
+      info.emplace_back("distance_metric");
```
dranikpg, line 1681: ah, I remember unifying this with the code above, so the function returns vector<string> and then the two callsites decide to either concatenate it or to put it out as an array like here, but not needed here

src/server/search/search_family.cc
vyavdoshenko: Yes, but we need this parameter to be saved to have the ability to recreate the index later from outside the dragonfly

PR #7048, opened 2026-04-01
refactor: increase timeout for test_redis_replication_all

BorysTheDev:
run 50 times https://github.com/dragonflydb/dragonfly/actions/runs/23851349718/job/69532061242

### DIFF
--- a/tests/dragonfly/redis_replication_test.py
+++ b/tests/dragonfly/redis_replication_test.py
@@ -173,14 +173,14 @@ async def test_redis_replication_all(
     await stream_task
 
     # Check data after full sync
-    await await_synced_all(c_master, c_replicas, timeout=60)
+    await await_synced_all(c_master, c_replicas, timeout=120)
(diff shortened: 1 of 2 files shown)


tests/dragonfly/redis_replication_test.py
```diff
@@ -173,14 +173,14 @@ async def test_redis_replication_all(
     await stream_task
 
     # Check data after full sync
-    await await_synced_all(c_master, c_replicas, timeout=60)
+    await await_synced_all(c_master, c_replicas, timeout=120)
```
kostasrim: I wrote this internally. We stream 2k keys on 8, 1proactor replicas. That's 8k total keys streamlined. Debug build or not, even with 1 hardware thread (1 core) the load does not justify the 30 second timeout 

I suggest for us to add LOG's and see exactly what is being delayed and why when we wait in the `await_synced_all`

PR #7051, opened 2026-04-02
fix: dev container build failure on ARM runners

vyavdoshenko:
Add `apt-get update` before installing `redis-tools` in the Test Image step.
The ARM runner's package cache is stale, causing 404 errors when fetching `redis-tools` - this has been failing nightly builds for the past 5 days, preventing `:ubuntu` and `:alpine` manifest tags from being updated.


vyavdoshenko:
fixed build: https://github.com/dragonflydb/dragonfly/actions/runs/23891217620

PR #7055, opened 2026-04-02
feat(server): Make parallel delayed entry processing safe

dranikpg:
Make it possible to call ProcessDelayedEntries from multiple fibers. Currently this is impossible due to big_value_mu_ guarding all paths, but in the future we want to unlock this behaviour


dranikpg:
augment review

PR #7060, opened 2026-04-03
feat(server): Journal omits

dranikpg:
Experimental journal omits

Implements `DbSlice::IsOmittableWrite` that can skip the CallOnChangeCallbacks + version update and has to notify the command to omit the journal write. 

This is safe if we have one eventually consistent snapshot (replica, migration) and the bucket was not visited yet. It will be fully serialized once the iteration loop reaches it

image

### DIFF
--- a/src/server/db_slice.h
+++ b/src/server/db_slice.h
@@ -74,6 +74,9 @@ struct SliceEvents {
   // how many updates and insertions of keys between snapshot intervals
   size_t update = 0;
 
+  // how many journal omit optimizations were performed
+  size_t journal_omit = 0;
+
   uint64_t huff_encode_total = 0, huff_encode_success = 0;
 
   SliceEvents& operator+=(const SliceEvents& o);
@@ -99,6 +102,7 @@ class DbSlice {
     virtual void WaitForNoBucketBlocked() const {
     }
 
+    bool eventually_consistent_ = false;
     uint64_t snapshot_version_ = 0;
   };
 
@@ -266,6 +270,10 @@ class DbSlice {
     Iterator it;
     AutoUpdater post_updater;
     bool is_new = false;
+
+    // Set if DbContext::is_omittable_operation was set and the conditions were met.
+    // Means that the journal write should NOT be performed.
+    bool omitted_journal = false;
   };
 
   ItAndUpdater FindMutable(const Context& cntx, std::string_view key);
@@ -555,6 +563,9 @@ class DbSlice {
   void SendQueuedInvalidationMessagesAsync();
 
   void CreateDb(DbIndex index);
+
+  // Returns true if this write could be ignored during replication without losing consistency
+  bool IsOmittableWrite(const Context& cntx, ChangeReq req);
 
   enum class UpdateStatsMode : uint8_t {
     kReadStats,
--- a/src/server/journal/journal.cc
+++ b/src/server/journal/journal.cc
@@ -56,8 +56,8 @@ error_code Close() {
   return {};
 }
 
-bool HasRegisteredCallbacks() {
-  return journal_slice.HasRegisteredCallbacks();
+unsigned GetCallbackCount() {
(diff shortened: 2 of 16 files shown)


src/server/db_slice.h
```diff
@@ -631,6 +654,9 @@ class DbSlice {
   // Record whenever a key expired to DbTable::expired_keys_events_ for keyspace notifications
   bool expired_keys_events_recording_ = true;
 
+  // Provided for the next FindMutableInternal operation
+  MutationHints* mutation_hints_ = nullptr;
```
romange, line 658: what is the size of `MutationHints` ? why do you need to pass by pointer and not by value? 
looks fragile

src/server/journal/journal.cc
```diff
@@ -47,8 +47,8 @@ error_code Close() {
   return {};
 }
 
-bool HasRegisteredCallbacks() {
-  return journal_slice.HasRegisteredCallbacks();
+unsigned CallbackNumber() {
```
romange, line 50: CallbackNumber is not a great name.

src/server/db_slice.h
dranikpg: Because it is also used to get output variables

src/server/string_family.cc
```diff
@@ -836,10 +837,15 @@ OpStatus SetCmd::Set(const SetParams& params, string_view key, string_view value
     }
   }
 
+  DbSlice::MutationHints hints{.hint{.single_key = true, .support_omit = explicit_journal_}};
```
romange, line 840: can we do it more elegantly?
1. at least in this example you pass hints with `.single_key = true` and set `support_omit` then why passing them at all? why not pass a final boolean true if the caller supports omit 
2. the boolean could be added to `op_args_.db_cntx`, maybe so you won't need to inject this before the find call.
3. `skip_journal_`  can be returned by AddOrFind inside `op_res`.

dranikpg: I agree, let's move it to existing structures

dranikpg:
Will rebase to include #7253

dranikpg:
augment review

src/server/serializer_base.cc
```diff
@@ -138,9 +138,10 @@ SerializerBase::~SerializerBase() {
 //   same fiber, so the baseline is serialized first. big_value_mu_ prevents this callback path
 //   from interleaving with the traversal fiber's bucket serialization, which may preempt while
 //   emitting large values.
-void SerializerBase::RegisterChangeListener() {
+void SerializerBase::RegisterChangeListener(bool replica) {
```
romange: can you explain `replica` argument?

src/server/string_family.cc
```diff
@@ -935,7 +940,11 @@ void SetCmd::PostEdit(const SetParams& params, std::string_view key, std::string
   }
 
   if (explicit_journal_ && op_args_.shard->journal()) {
-    RecordJournal(params, key, value);
+    // Skipping a journal write breaks the LSN buffer, so clear it
```
romange: can you please expand? what do you mean it breaks the LSN buffer?

src/server/string_family.cc
```diff
@@ -935,7 +940,13 @@ void SetCmd::PostEdit(const SetParams& params, std::string_view key, std::string
   }
 
   if (explicit_journal_ && op_args_.shard->journal()) {
-    RecordJournal(params, key, value);
+    // A skipped journal write is not added to the journal ring buffer, so it goes unnoticed for
```
romange, line 943: I feel it's still hard to understand the exact scenario of when it happens as the main scenario of snapshotting is unrelated to psync.

up to you.

romange (approved):

romange:
@dranikpg  one thing I forgot to add - can we add a runtime flag to be able to revert the feature if needed?

dranikpg:
will add it

PR #7061, opened 2026-04-03
feat: preserve indexdata during serialization

BorysTheDev:
Summary: This PR makes HNSW (global) vector indices safe to serialize when vectors are stored as external pointers into HASH StringMap entries (copy_vector=false).

Changes:

- Extend HNSW remove APIs to report whether a remove ran synchronously or was deferred under MRMW read locks.
- Add DrainPendingOps() to block on the HNSW write lock and drain deferred ops after serialization.
- Allow index-removal callbacks to mutate PrimeValue so HASH field sds entries can be swapped/copied for pointer preservation.
- Preserve old StringMap sds entries for deferred HNSW removes, and clear preserved data once deferred ops have drained.
- Update HASH command paths (e.g. HSET/HDEL/HINCRBY) to pass modified field names for selective preservation.
- After serializing all HNSW graphs, drain pending ops and clear preserved field data across shards.
- Extend replication stress test to cover “external vector” mode (DIM=256) and validate hashtable encoding.

### DIFF
--- a/src/server/search/doc_index.cc
+++ b/src/server/search/doc_index.cc
@@ -525,7 +525,7 @@ bool HnswShardIndex::Add(search::GlobalDocId id, const BaseAccessor& doc) {
 
 void HnswShardIndex::Remove(search::GlobalDocId id, const BaseAccessor& doc, PrimeValue& pv,
                             absl::Span<const std::string_view> modified_fields,
-                            ExtractionCache* cache) {
+                            FieldExtractionCache* cache) {
   bool sync = global_index_->Remove(id, doc, field_ident_);
   if (sync) {
     // Write lock acquired, ProcessDeferred drained all pending ops for this field.
@@ -610,7 +610,7 @@ void ShardDocIndex::AddDocToGlobalVectorIndex(ShardDocIndex::DocId doc_id, const
 
 void ShardDocIndex::RemoveDocFromGlobalVectorIndex(
     ShardDocIndex::DocId doc_id, const DbContext& db_cntx, PrimeValue& pv,
-    absl::Span<const std::string_view> modified_fields, HnswShardIndex::ExtractionCache* cache) {
+    absl::Span<const std::string_view> modified_fields, FieldExtractionCache* cache) {
   if (is_restoring_vectors_) {
     std::string_view key = key_index_.Get(doc_id);
     pending_vector_updates_.emplace(key);
@@ -1222,7 +1222,7 @@ void ShardDocIndices::RemoveDoc(string_view key, const DbContext& db_cntx, Prime
 
   // Shared extraction cache: when multiple search indices reference the same hash field,
   // each sds entry is extracted once and shared via shared_ptr across all indices.
-  HnswShardIndex::ExtractionCache extraction_cache;
+  FieldExtractionCache extraction_cache;
 
   for (auto& [index_name, index] : indices_) {
     if (index->Matches(key, pv.ObjType())) {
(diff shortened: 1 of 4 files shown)


src/core/search/hnsw_index.cc
```diff
@@ -375,6 +377,14 @@ struct HnswlibAdapter {
   }
 
  public:
+  // Block until the write lock is acquired and all deferred ops are drained.
+  // Returns the held write lock so the caller can keep it alive.
+  std::unique_ptr<MRMWMutexLock> DrainPendingOps() {
+    auto lock = std::make_unique<MRMWMutexLock>(&mrmw_mutex_, MRMWMutex::LockMode::kWriteLock);
+    ProcessDeferred();
+    return lock;
+  }
```
dranikpg, line 386: unique ptr to forward declare it?

src/server/search/doc_index.h
```diff
@@ -401,6 +403,17 @@ class ShardDocIndex {
...
+  // Extracts and keeps alive the old sds entries so deferred HNSW removal
+  // can still dereference pointers into them. Only called when
+  // RemoveDocFromGlobalVectorIndex was deferred (couldn't acquire write lock).
+  // When modified_fields is empty, preserves all HNSW-indexed fields.
+  void PreserveFieldData(PrimeValue& pv, absl::Span<const std::string_view> modified_fields);
+
+  void ClearPreservedData() {
+    preserved_field_data_.clear();
+  }
+
```
dranikpg: I don't understand why generally all the hnsw complexity has to be pushed upwrads to the very top level of the search module. Why can't it all be handled by the hnsw index internally?

dranikpg: After all it needs to keep the data alive just for itself?

src/server/search/doc_index.cc
```diff
@@ -525,27 +620,20 @@ void ShardDocIndex::RemoveDocFromGlobalVectorIndex(ShardDocIndex::DocId doc_id,
...
 void ShardDocIndex::RemoveFromAllHnswIndices(search::DocId doc_id) {
   GlobalDocId global_id = search::CreateGlobalDocId(EngineShard::tlocal()->shard_id(), doc_id);
-  for (const auto& [field_ident, field_info] : GetIndexedHnswFields(base_->schema)) {
-    if (auto index = GlobalHnswIndexRegistry::Instance().Get(base_->name, field_info.short_name);
-        index) {
-      index->Remove(global_id);
-    }
+  for (auto& hnsw : hnsw_shard_indices_) {
+    hnsw.RemoveById(global_id);
   }
```
dranikpg, line 632: why not store the shard index inside the field indices? this way you don't need any special hnsw code - and if you need it as it's own type you can cast it back with static_cast or reinterpret_cast

src/server/search/doc_index.cc
```diff
@@ -1093,6 +1175,7 @@ void ShardDocIndices::DropIndexCache(const dfly::ShardDocIndex& shard_doc_index)
 
 void ShardDocIndices::RebuildAllIndices(const OpArgs& op_args, bool is_restored) {
   for (auto& [index_name, ptr] : indices_) {
+    ptr->InitHnswShardIndices();
```
dranikpg, line 1178: and I assume that you then also don't need those parts

src/server/search/doc_index.cc
BorysTheDev: I will make these changes as a separate PR

Issue #7062, opened 2026-04-03
FT.AGGREGATE missing APPLY, SCORER, ADDSCORES, and DIALECT clauses


vyavdoshenko:
`FT.AGGREGATE` is missing several clauses that are already implemented in `FT.SEARCH`. This blocks compatibility with redisvl `HybridQuery` (BM25 + vector hybrid search) used by Microsoft AutoGen Agent Framework and other AI/ML frameworks.

Missing clauses:

| Clause | Status in FT.SEARCH | Status in FT.AGGREGATE | Error |
|--------|:---:|:---:|---|
| `DIALECT 2` | Works | Missing | `Unknown clause: DIALECT` |
| `APPLY expr AS alias` | N/A | Missing | `Unknown clause: APPLY` |
| `SCORER BM25STD` | Works | Missing | `Unknown clause: SCORER` |
| `ADDSCORES` | N/A (WITHSCORES works) | Missing | `Unknown clause: SCORER` (parser stops earlier) |
| KNN in query string | Works | Parses but returns 0 results | — |

Reproduction:

```
FT.CREATE test_agg ON HASH PREFIX 1 test: SCHEMA title TEXT SORTABLE price NUMERIC SORTABLE vec VECTOR HNSW 6 TYPE FLOAT32 DIM 4 DISTANCE_METRIC COSINE

HSET test:1 title "introduction to machine learning" price 29.99 vec "\x00\x00\x80\x3f\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
HSET test:2 title "advanced deep learning techniques" price 49.99 vec "\x00\x00\x00\x00\x00\x00\x80\x3f\x00\x00\x00\x00\x00\x00\x00\x00"
HSET test:3 title "cooking with science and chemistry" price 19.99 vec "\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x80\x3f\x00\x00\x00\x00"
```

DIALECT:
```
FT.AGGREGATE test_agg "*" LOAD 1 @title LIMIT 0 3 DIALECT 2
-> (error) ERR Unknown clause: DIALECT
```

APPLY:
```
FT.AGGREGATE test_agg "*" LOAD 1 @price APPLY "@price * 1.2" AS taxed LIMIT 0 3
-> (error) ERR Unknown clause: APPLY
```

SCORER:
```
[...]

PR #7066, opened 2026-04-04
feat(search): add DIALECT, APPLY, and KNN support to FT.AGGREGATE

vyavdoshenko:
FT.AGGREGATE was missing several clauses that blocked compatibility with redisvl used by AI/ML frameworks like Microsoft AutoGen.

Changes:
- DIALECT clause: accepted and ignored (DF always behaves as dialect 2)
- APPLY expr AS alias: evaluate arithmetic expressions over the pipeline fields, reuses the existing FILTER expression parser/evaluator
- KNN in FT.AGGREGATE: pop KNN node and run global HNSW search, with prefilter support for filtered KNN queries

This enables redisvl queries like:
```
FT.AGGREGATE idx "*=>[KNN 3 @vec $vector AS dist]"
LOAD 1 @title
APPLY "(2 - @dist) / 2" AS similarity
SORTBY 2 @similarity DESC
DIALECT 2
PARAMS 2 vector <bytes>
```

Not included (requires BM25 scoring engine, tracked in #7062):
- SCORER / ADDSCORES / ~ (tilde optional match)

First part fixes of #7062

### DIFF
--- a/src/server/search/search_family.cc
+++ b/src/server/search/search_family.cc
@@ -1414,6 +1414,43 @@ std::shared_ptr<search::HnswVectorIndex> GetValidatedHnswRangeIndex(
   return hnsw_index;
 }
 
+// Group global (score, GlobalDocId) pairs into per-shard vectors of (local DocId, score).
+std::vector<std::vector<std::pair<search::DocId, float>>> GroupByShardId(
+    const std::vector<std::pair<float, search::GlobalDocId>>& global_results, ShardId shard_count) {
+  std::vector<std::vector<std::pair<search::DocId, float>>> shard_docs(shard_count);
+  for (const auto& [score, global_doc_id] : global_results) {
+    auto [shard_id, local_doc_id] = search::DecomposeGlobalDocId(global_doc_id);
+    shard_docs[shard_id].emplace_back(local_doc_id, score);
+  }
+  return shard_docs;
+}
+
+// Collect GlobalDocIds from sharded prefilter search results.
+std::vector<search::GlobalDocId> CollectPrefilterGlobalIds(
+    absl::Span<const SearchResult> prefilter_docs) {
+  std::vector<search::GlobalDocId> ids;
+  for (size_t shard_id = 0; shard_id < prefilter_docs.size(); shard_id++) {
+    for (const auto& doc : prefilter_docs[shard_id].docs) {
+      ids.push_back(search::CreateGlobalDocId(shard_id, doc.id));
+    }
+  }
+  return ids;
+}
+
+// Try to pop KNN node from search algorithm if HNSW index exists for the field.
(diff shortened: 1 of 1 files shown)


src/server/search/search_family.cc
```diff
@@ -2104,15 +2124,89 @@ void CmdFtAggregate(CmdArgList args, CommandContext* cmd_cntx) {
...
+                prefilter_docs[es->shard_id()] =
+                    index->Search(t->GetOpArgs(es), sp, &search_algo, true);
+              }
+              return OpStatus::OK;
+            },
+            false);
+      }
+
+      // Collect prefilter GlobalDocIds for HNSW search
+      std::optional<std::vector<search::GlobalDocId>> prefilter_global_ids;
```
romange: why this block and the one above are split? you can unite them under the same condition.

src/server/search/search_family.cc
```diff
@@ -2104,15 +2124,89 @@ void CmdFtAggregate(CmdArgList args, CommandContext* cmd_cntx) {
...
+            auto doc_id = doc.id;
+            ids.push_back(search::CreateGlobalDocId(shard_id, doc_id));
+          }
+        }
+        prefilter_global_ids = std::move(ids);
+      }
+
+      // Run global HNSW KNN search
+      std::vector<std::pair<float, search::GlobalDocId>> knn_results;
+      if (prefilter_global_ids) {
```
romange: similarly here. run `/simplify in claude` if it was written by it.

src/server/search/search_family.cc
```diff
@@ -2104,15 +2124,89 @@ void CmdFtAggregate(CmdArgList args, CommandContext* cmd_cntx) {
...
+
+      // Group results by shard and load doc data
+      const ShardId shard_size = shard_set->size();
+      std::vector<std::vector<std::pair<search::DocId, float>>> shard_docs(shard_size);
+      for (const auto& [score, global_doc_id] : knn_results) {
+        auto [shard_id, local_doc_id] = search::DecomposeGlobalDocId(global_doc_id);
+        shard_docs[shard_id].emplace_back(local_doc_id, score);
+      }
+
+      auto load_cb = [&](Transaction* t, EngineShard* es) {
```
romange: I see we have same lambdas in the existing code. Please make an effort to define shared functions only once - either define them as private member functions or define lambdas at the higher scope. Avoid code duplication, please

dranikpg: CmdFtAggregate is now 280 lines of code, it's scary

src/server/search/search_family.cc

src/server/search/search_family.cc

src/server/search/search_family.cc
vyavdoshenko: Extracted as much as I could

PR #7067, opened 2026-04-04
chore(core): extract HllEstimator class from dict_builder

romange:
## Summary

- Extract HyperLogLog cardinality estimation logic from `dict_builder.cc` into a reusable `HllEstimator` class in `src/core/hll_estimator.h`
- Replace raw register array and free functions with the new class in `EstimateCompressibility`

### DIFF
--- a/src/core/hll_estimator.h
+++ b/src/core/hll_estimator.h
@@ -1,0 +1,62 @@
+// Copyright 2026, DragonflyDB authors.  All rights reserved.
+// See LICENSE for licensing terms.
+//
+
+#pragma once
+
+#include <algorithm>
+#include <array>
+#include <bit>
+#include <cmath>
+#include <cstdint>
+
+namespace dfly {
+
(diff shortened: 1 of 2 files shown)


src/core/hll_estimator.h
```diff
@@ -0,0 +1,69 @@
...
+    // Small range correction
+    constexpr double kSmallRangeThreshold = 2.5 * kM;
+    if (estimate <= kSmallRangeThreshold && zero_registers > 0) {
+      estimate = kM * std::log(kM / zero_registers);
+    }
+    return estimate;
+  }
+
+  // Merge another estimator's registers into this one (element-wise max).
+  void Merge(const HllEstimator& other) {
```
vyavdoshenko: Do you need this method? This code doesn't execute at all.

romange: oops, no I do not need it.

PR #7068, opened 2026-04-04
fix(core): Update CVCUponInsert

dranikpg:
Don't skip buckets with versions that are less than snapshort version in CVCUponInsert. This is because we set the bucket version just when we start serializing. As a result, a modification operation doesn't wait for serializatin to finish before proceeding - because the version is already equal to snapshot version. This bug would occur if the global mutex would be absent - so this PR reintroduces function scope locking for it in OnChange.

It also introduces BucketSet as a utility to get rid of the callback

### DIFF
--- a/src/server/serializer_base.cc
+++ b/src/server/serializer_base.cc
@@ -202,7 +202,7 @@ void SerializerBase::OnChangeBlocking(DbIndex db_index, PrimeTable::bucket_itera
   ProcessBucket(db_index, it, true);
 }
 
-void SerializerBase::OnChange(DbIndex db_index, std::string_view key) {
+void SerializerBase::OnChangeBlocking(DbIndex db_index, std::string_view key) {
   // We must acquire the mutex ahead and process all buckets under the same lock.
   // This ensures that CVCUponInsert and the table insertion that invoked this callback
   // will be operating on the same state as all writes are linarly ordered by this mutex.


dranikpg:
augment review

dranikpg:
augment review

src/server/serializer_base.cc
```diff
@@ -153,9 +163,18 @@ bool SerializerBase::ProcessBucket(DbIndex db_index, PrimeTable::bucket_iterator
...
+
   stats_.keys_serialized += SerializeBucket(db_index, it, on_update);
+  stats_.buckets_on_change += unsigned(on_update);
+
   FinishBucketIteration(it.bucket_address());
 
+  // Assert the version is equal to a snapshot version so no other modifications have happened
+  DCHECK_GE(it.GetVersion(), snapshot_version_);
+  auto current_snapshots = db_slice_->SnapshotVersions();
+  DCHECK(std::ranges::find(current_snapshots, it.GetVersion()) != current_snapshots.end())
```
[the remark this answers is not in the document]
dranikpg: one thread owns it, we iterate non-preemptively

src/core/dash_internal.h
```diff
@@ -531,19 +532,17 @@ class Segment {
...
-  // Returns UINT16_MAX if segment is full. Otherwise, returns number of touched bucket ids (1 or 2)
-  // if the insertion would happen. The ids are put into bid array that should have at least 2
-  // spaces.
+  // Determines ids of buckets that would possibly change on insertion
+  // Returns nullopt if segment is full.
+  // Otherwise, writes buckets to `bid` and returns the count (1 or 2).
   template <bool UV = kUseVersion>
-  requires UV unsigned CVCOnInsert(uint64_t ver_threshold, Hash_t key_hash,
-                                   PhysicalBid bid[2]) const;
+  requires UV std::optional<uint8_t> CVCOnInsert(Hash_t key_hash, PhysicalBid bid[2])
```
romange: can it return 0?

romange: Using optional for such trivial cases in unjustified. I do not think that "special" result-codes must die just because we have optional: you still need to understand the contract. "nullopt" is a special case. You do not simplify the caller code. You have plenty of range to use special numbers (0, or 255) - so why optional?

src/server/db_slice.cc
```diff
@@ -734,7 +734,6 @@ OpResult<DbSlice::ItAndUpdater> DbSlice::AddOrFindInternal(const Context& cntx,
     events_.insertion_rejections++;
     return OpStatus::OUT_OF_MEMORY;
   }
-  CallMovedCallbacks(cntx.db_index, evp.moved_items());
```
romange, line 740: should we kill pytests that cover non-pit mode as well?

romange: and also the flag?

src/core/dash.h
```diff
@@ -610,6 +606,25 @@ class DashTable<_Key, _Value, Policy>::Iterator {
   void Seek2Occupied();
 };  // Iterator
 
+// Limited set of buckets on a single segment that can be turned into a iterator view
+template <typename _Key, typename _Value, typename Policy>
+struct DashTable<_Key, _Value, Policy>::BucketSet {
```
romange, line 610: ok, but please move the members to private scope and add undersores.

src/core/dash.h
```diff
@@ -610,6 +606,25 @@ class DashTable<_Key, _Value, Policy>::Iterator {
...
+  auto buckets() const {
+    bool is_all = limit > 2;
+    return std::views::iota(0u, limit) |
+           std::views::transform([=, ids = ids, owner = owner, seg_id = seg_id](uint8_t i) {
```
romange: `[=, ids = ids, owner = owner, seg_id = seg_id]` ? why not `[=]` ?

src/core/dash.h
```diff
@@ -378,20 +382,12 @@ class DashTable : public detail::DashTableBase {
...
-  // Note: the interface a bit hacky.
-  // The functions call cb on physical buckets with version smaller than ver_threshold that
-  // due to entry movements might update its version to version greater than ver_threshold.
-  //
+  // Capture Version Change. Determine buckets that can potentially be modified when inserting key.
   // These are not const functions because they send non-const iterators that allow
   // updating contents/versions of the passed iterators.
-  template <typename U, typename Cb>
-  void CVCUponInsert(uint64_t ver_threshold, const U& key, Cb&& cb);
+  template <typename U> BucketSet CVCUponInsert(const U& key);
```
romange, line 387: if you do that, you can then validate that InsertNew created the iterator belonging to this set.

src/core/dash.h
```diff
@@ -610,6 +606,25 @@ class DashTable<_Key, _Value, Policy>::Iterator {
   void Seek2Occupied();
 };  // Iterator
 
+// Limited set of buckets on a single segment that can be turned into a iterator view
+template <typename _Key, typename _Value, typename Policy>
+struct DashTable<_Key, _Value, Policy>::BucketSet {
+  auto buckets() const {
+    bool is_all = limit > 2;
```
romange:
```suggestion
    bool is_all = limit > ids.size();
```

src/core/dash.h
```diff
@@ -679,43 +694,33 @@ DashTable<_Key, _Value, Policy>::~DashTable() {
...
 
-  // Segment is full, we need to return the whole segment, because it can be split
+  // If the segment is full, we need to return the whole segment, because it can be split
   // and its entries can be reshuffled into different buckets.
-  for (uint8_t i = 0; i < target->num_buckets(); ++i) {
-    if (target->GetVersion(i) < ver_threshold && !target->GetBucket(i).IsEmpty()) {
-      cb(bucket_iterator{this, seg_id, i});
-    }
-  }
+  uint8_t bucket_limit = num_touched.value_or(target->num_buckets());
```
romange: so if you do that, why `target->CVCOnInsert(key_hash, bids);` can not return `num_buckets()` ? 
why creating a special case on the caller side and chaining in the interface?

src/server/serializer_base.cc
```diff
@@ -156,9 +163,18 @@ bool SerializerBase::ProcessBucket(DbIndex db_index, PrimeTable::bucket_iterator
...
   MarkBucketSerializing(it.bucket_address());
+
   stats_.keys_serialized += SerializeBucket(db_index, it, on_update);
+  stats_.buckets_on_change += unsigned(on_update);
+
   FinishBucketIteration(it.bucket_address());
 
+  // Assert the version is equal to a snapshot version so no other modifications have happened
+  DCHECK_GE(it.GetVersion(), snapshot_version_);
+  auto current_snapshots = db_slice_->SnapshotVersions();
```
romange, line 180: ok, what is that?  what scenario it guards against?

Please note that `db_slice_->SnapshotVersions();` is called and executed in release mode as well.

src/server/serializer_base.cc
```diff
@@ -156,9 +163,18 @@ bool SerializerBase::ProcessBucket(DbIndex db_index, PrimeTable::bucket_iterator
 
   it.SetVersion(snapshot_version_);
   MarkBucketSerializing(it.bucket_address());
+
   stats_.keys_serialized += SerializeBucket(db_index, it, on_update);
```
romange: please move DCHECK_EQ inside `SerializeBucket` to the loop like I did here:
https://github.com/dragonflydb/dragonfly/pull/7071/changes#diff-f67586dd571688e986aba8f508c6ee198e0adbfa01ebd362b55403bac5f22b84R237

src/server/db_slice.cc
dranikpg: removed

src/core/dash.h
dranikpg: I will do it later to reduce changes - instead of passing the key, we can just pass the BucketSet to OnChange and validate there later

src/core/dash.h
```diff
@@ -3,6 +3,9 @@
 //
 #pragma once
 
+#include <array>
+#include <optional>
```
romange: do not need optional now

src/core/dash_internal.h
```diff
@@ -11,6 +11,7 @@
 #include <cstdint>
 #include <cstdio>
 #include <cstring>
+#include <optional>
```
romange: no need

src/core/dash_internal.h
```diff
@@ -531,19 +532,17 @@ class Segment {
   Iterator InsertUniq(U&& key, V&& value, Hash_t key_hash, bool spread, OnMoveCb&& on_move_cb);
 
   // capture version change in case of insert.
-  // Returns ids of buckets whose version would cross ver_threshold upon insertion of key_hash
-  // into the segment.
-  // Returns UINT16_MAX if segment is full. Otherwise, returns number of touched bucket ids (1 or 2)
-  // if the insertion would happen. The ids are put into bid array that should have at least 2
-  // spaces.
+  // Determines ids of buckets that would possibly change on insertion
+  // Returns nullopt if segment is full.
```
romange: update comments as well

src/server/serializer_base.cc
```diff
@@ -5,6 +5,9 @@
 #include "server/serializer_base.h"
 
 #include <absl/strings/match.h>
+#include <absl/strings/str_join.h>
+
+#include <mutex>
```
romange: can you change your settings so that auto-format stops adding includes?

src/server/serializer_base.cc
```diff
@@ -172,16 +191,20 @@ void SerializerBase::OnChange(DbIndex db_index, PrimeTable::bucket_iterator it)
       !absl::StartsWith(active->name(), "SliceSnapshot")) {
     LOG(DFATAL) << "Unexpected fiber: " << active->name() << " on " << util::fb2::GetStacktrace();
   }
-  if (ProcessBucket(db_index, it, true))
-    ++stats_.buckets_on_change;
+
+  std::lock_guard guard(big_value_mu_);
```
romange: can be replaced with `ProcessIfNeeded` which I find a bad name as it always calls `ProcessBucketInternal` so there is no `if needed` part

romange:
The PR description is outdated now

PR #7070, opened 2026-04-05
feat(server): Implement tagged chunk loading

abhijat:
Changes to rdb loader to handle tagged chunks. A data stream read by the loader now looks like:

```
{chunk1:[stream-id]chunk-size][data...]}{untagged data...}{chunk2:[stream-id]chunk-size][data...]}
```

If the sender does not turn on tagging, then all the data is untagged.

where
* stream-id is 4 byte id, same for all chunks of same db+key
* chunk size is 4 byte size of data in chunk

Code changes:
  - parse `RDB_OPCODE_TAGGED_CHUNK` and track the active chunk payload budget
  - save and restore per-stream partial object state across continuation chunks
  - add loader-only tests that feed handcrafted tagged byte streams, including split SBF cases and interleaving

New state in loader:
* current chunk: stream id and payload size
* stream states: partially read objects which can be resumed when next chunk for them is seen

Some of the new tests can be removed once the saver PR is merged as they will become redundant. But for now these are necessary to test the loader code in isolation by feeding it artificial byte stream containing chunks.

### Example of loading one set split across multiple tagged chunks
  - First chunk:
      - chunk budget is set from the tagged payload size
      - call chain: `LoadKeyValPair -> ReadAndDispatchObject -> ReadObj -> ReadSet`
      - reading stops when either the set is complete or the chunk budget is exhausted
  - If the chunk ends first (budget exhausted):
      - the partially built PrimeValue is kept in `now_chunked_`
      - parser continuation state, including remaining items, is saved in `stream_states_`
  - Continuation chunk:
      - chunk budget is set again from the new tagged payload size
      - call chain: `LoadValueChunk -> ReadAndDispatchObject -> ReadObj -> ReadSet` 
      - parser state is restored from `stream_states_`
      - newly read elements are appended to the partial value in `now_chunked_`
  - This repeats until the object is fully read, at which point:
      - continuation state is removed
      - the final value is written to the DB


src/server/rdb_load.cc
```diff
@@ -2341,6 +2435,33 @@ error_code RdbLoader::Load(io::Source* src) {
...
+      // A known stream id means this is a continuation chunk containing payload only.
+      // Otherwise this is the first chunk of a new object, and the normal object-loading
+      // path below will read its type and key.
+      if (stream_states_.contains(current_chunk_state_->stream_id)) {
+        RETURN_ON_ERR(LoadValueChunk());
+        if (current_chunk_state_->remaining_payload_bytes != 0) {
+          LOG(ERROR) << "chunk fully consumed but payload bytes remain "
+                     << current_chunk_state_->remaining_payload_bytes;
+          return RdbError(errc::rdb_file_corrupted);
+        }
```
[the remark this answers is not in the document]
abhijat: fixed

src/server/rdb_load.cc
```diff
@@ -2354,6 +2475,16 @@ error_code RdbLoader::Load(io::Source* src) {
     ++keys_loaded;
     RETURN_ON_ERR(LoadKeyValPair(type, &settings));
     settings.Reset();
+
+    if (current_chunk_state_ && current_chunk_state_->remaining_payload_bytes != 0) {
+      LOG(ERROR) << "chunk fully consumed but payload bytes remain "
+                 << current_chunk_state_->remaining_payload_bytes;
+      return RdbError(errc::rdb_file_corrupted);
+    }
```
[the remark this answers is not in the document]
abhijat: fixed

abhijat:
augment review

src/server/rdb_load.cc
```diff
@@ -2155,14 +2243,20 @@ error_code RdbLoader::Load(io::Source* src) {
     }
 
     if (type == RDB_OPCODE_EOF) {
```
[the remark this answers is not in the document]
abhijat: fixed

src/server/rdb_load.cc
```diff
@@ -2341,6 +2435,33 @@ error_code RdbLoader::Load(io::Source* src) {
       continue;
     }
 
+    if (type == RDB_OPCODE_TAGGED_CHUNK) {
+      ActiveTaggedChunk state;
+      SET_OR_RETURN(FetchInt<uint32_t>(), state.stream_id);
+      SET_OR_RETURN(FetchInt<uint32_t>(), state.remaining_payload_bytes);
```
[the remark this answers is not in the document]
abhijat: fixed

src/server/rdb_load.cc
```diff
@@ -2879,81 +3038,145 @@ error_code RdbLoader::LoadKeyValPair(int type, ObjSettings* settings) {
...
+      LOG(ERROR) << "attempt to add first chunk for id " << current_chunk_state_->stream_id
+                 << " which already exists";
+      return RdbError(errc::rdb_file_corrupted);
     }
-    // Delete the item if we fail to load the key/val pair.
-    auto cleanup = absl::Cleanup([item] { delete item; });
+    stream_states_[current_chunk_state_->stream_id] = {
+        .db_index = cur_db_index_,
+        .type = type,
+        .pending_read = pending_read_,
```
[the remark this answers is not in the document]
abhijat: changed to move

src/server/rdb_load.cc
```diff
@@ -2527,6 +2677,14 @@ error_code RdbLoaderBase::HandleCompressedBlob(int op_type) {
   string res;
   SET_OR_RETURN(FetchGenericString(), res);
 
+  // Stop counting payload bytes on decompressed data. At this point the entire payload size must be
+  // consumed as it was the compressed blob. We switch to another buffer and must be able to read
+  // everything from it without any checks
+  if (current_chunk_state_ && current_chunk_state_->remaining_payload_bytes > 0) {
```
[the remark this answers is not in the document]
abhijat: I need to confirm this, as I recall a tagged chunk does not contain compressed data by itself, compression is done one level higher than chunking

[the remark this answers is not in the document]
abhijat: confirmed that this concern is invalid, compression is done after chunking, tagged chunks do not "contain" compressed data.

src/server/rdb_load.cc
```diff
@@ -2229,6 +2240,13 @@ error_code RdbLoader::Load(io::Source* src) {
     }
 
     if (type == RDB_OPCODE_EOF) {
+      if (current_chunk_state_.has_value() || !stream_states_.empty()) {
+        LOG(ERROR) << "eof seen while a previous chunk is not yet finished, stream id "
+                   << current_chunk_state_->stream_id << ", remaining bytes "
+                   << current_chunk_state_->remaining_payload_bytes
+                   << ", pending stream states: " << stream_states_.size();
+        return RdbError(errc::rdb_chunk_payload_remaining);
+      }
```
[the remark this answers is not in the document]
abhijat: fixed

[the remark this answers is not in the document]
abhijat: fixed

src/server/rdb_load.cc
```diff
@@ -2231,6 +2242,19 @@ error_code RdbLoader::Load(io::Source* src) {
...
+      if (current_chunk_state_.has_value() || !stream_states_.empty()) {
+        if (current_chunk_state_.has_value()) {
+          LOG(ERROR) << "eof seen while a previous chunk is not yet finished, stream id "
+                     << current_chunk_state_->stream_id << ", remaining bytes "
+                     << current_chunk_state_->remaining_payload_bytes
+                     << ", pending stream states: " << stream_states_.size();
+          return RdbError(errc::rdb_chunk_payload_remaining);
+        }
+
+        if (!stream_states_.empty())
```
romange: if must be `!stream_states_.empty()` so DCHECK

src/server/rdb_load.cc
```diff
@@ -2418,6 +2442,39 @@ error_code RdbLoader::Load(io::Source* src) {
       continue;
     }
 
+    if (type == RDB_OPCODE_TAGGED_CHUNK) {
+      ActiveTaggedChunk state;
+      SET_OR_RETURN(FetchInt<uint32_t>(), state.stream_id);
+      SET_OR_RETURN(FetchInt<uint32_t>(), state.remaining_payload_bytes);
```
romange, line 2381: I looked at `remaining_payload_bytes` field definition and still do not understand  - what `remaining_payload_bytes`  mean?  I guess in this context it's just "chunk size", correct?

src/server/rdb_load.cc
```diff
@@ -2434,6 +2491,16 @@ error_code RdbLoader::Load(io::Source* src) {
     VLOG(2) << "LoadKeyValPair key=" << last_key_loaded_ << " rdb_type=" << type
             << " db= " << cur_db_index_;
     settings.Reset();
+    if (!stop_early_.load(memory_order_relaxed) && current_chunk_state_ &&
```
romange: seems like a duplication of the check above. Can we reorder and consolidate the code so that this check will be done once?  `current_chunk_state_.reset();` as well? Maybe moving `LoadValueChunk()` to  `LoadKeyValPair` function will consolidate things?

src/server/rdb_load.cc
```diff
@@ -2629,6 +2696,16 @@ error_code RdbLoaderBase::HandleCompressedBlob(int op_type) {
   string res;
   SET_OR_RETURN(FetchGenericString(), res);
 
+  // Stop counting payload bytes on decompressed data. At this point the entire payload size must be
+  // consumed as it was the compressed blob. We switch to another buffer and must be able to read
+  // everything from it without any checks
+  if (current_chunk_state_ && current_chunk_state_->remaining_payload_bytes > 0) {
```
romange: again, same comment.

src/server/rdb_load.cc
```diff
@@ -2997,13 +3074,66 @@ error_code RdbLoader::LoadKeyValPair(int type, ObjSettings* settings) {
     SET_OR_RETURN(ReadAndDispatchObject(type, key, *settings, cur_db_index_), finalized);
   } while (!finalized && remaining_payload_bytes() > 0 && !stop_early_.load(memory_order_relaxed));
 
+  // If the first tagged chunk did not finish the object, save enough state to resume it when
+  // the next chunk with the same stream id arrives.
+  if (!finalized && current_chunk_state_) {
+    if (stream_states_.contains(current_chunk_state_->stream_id)) {
```
romange: when you use contains and then operator[], you perform find operation twice. `emplace` or `insert` performs AddOrFind operation that should cover both conditions here.

src/server/rdb_load.cc
abhijat: yes it works like below:

1. when we see `RDB_OPCODE_TAGGED_CHUNK` we parse chunk size, and store it in `current_chunk_state_`
2. thereafter each read method deducts number of bytes read from this size
3. `remaining_payload_bytes` at every read point is the number of bytes in chunk remaining, it is used for validation and to stop reading the current data structure when the number goes down to 0, and return back to main loop to read the next chunk or untagged data

src/server/rdb_load.cc
abhijat: changed to use `try_emplace` and the returned boolean for check

src/server/rdb_load.cc
abhijat: I added a helper to put the duplicate checks in one place, which brings down the duplication, but I also tried to reorder and it proved a little more difficult. 

Currently `type == RDB_OPCODE_TAGGED_CHUNK` code path has two responsibilities

* if it is the first chunk of a key-value pair then it just goes on to read more opcodes which might be in the chunk payload, since its the beginning of a new object. eventually the loop iterations will fall through and begin to read the key and then the first part of object.
* if its a continuation chunk then there are going to be no opcodes, just more data belonging to value, so we can load the value chunk right now

So if I try to make `LoadKeyValPair` have two code paths where it will handle chunks also, then for that the tagged chunk branch needs to drop down instead of continuing for the first chunk, which is incorrect, because there can be more opcodes after the tagged chunk envelope (inside the payload bytes). 

Right now we read them by continuing with the loop until all opcodes are read and we fall through to read the kv pair. 

There are ways around it I guess by keeping track through local variables but I think it might end up looking a lot more complicated than it is now.

src/server/rdb_load.cc
abhijat: this condition ended up redundant, removed it

abhijat:
fuzzer error due to a download failure from internet

https://github.com/dragonflydb/dragonfly/actions/runs/25048793584/job/73370858321

ETA: other tests fail for same artifact

```
CMake Error at benchmark-subbuild/benchmark-populate-prefix/src/benchmark-populate-stamp/download-benchmark-populate.cmake:170 (message):
  Each download failed!

    error: downloading 'https://github.com/google/benchmark/archive/v1.9.5.tar.gz' failed
          status_code: 22
          status_string: "HTTP response code said error"
          log:
          --- LOG BEGIN ---
          Host github.com:443 was resolved.
```

src/server/rdb_load.cc
```diff
@@ -2563,6 +2619,11 @@ error_code RdbLoaderBase::HandleCompressedBlob(int op_type) {
   string res;
   SET_OR_RETURN(FetchGenericString(), res);
 
+  // Stop counting payload bytes on decompressed data. At this point the entire payload size must be
+  // consumed as it was the compressed blob. We switch to another buffer and must be able to read
+  // everything from it without any checks
+  FinishCurrentChunk();
```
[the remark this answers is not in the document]
abhijat: fixed

src/server/rdb_load.cc
```diff
@@ -2231,6 +2248,18 @@ error_code RdbLoader::Load(io::Source* src) {
     }
 
     if (type == RDB_OPCODE_EOF) {
+      if (current_chunk_state_.has_value() || !stream_states_.empty()) {
```
romange: style: have two separate checks would be cleaner

src/server/rdb_load.cc
```diff
@@ -2176,6 +2187,12 @@ error_code RdbLoader::Load(io::Source* src) {
     GetCurrentDbSlice().IncrLoadInProgress();
   }
 
+  auto finalize_curr_chunk = [&] {
```
romange: I think the simplification was in the good direction. Still not sure why this lambda exists.
If you rename  `FinishCurrentChunk` to `FinalizeCurrentChunkIfNeeded`, you can push this check inside. You already do checks there. Then you do not need this lambda.

src/server/rdb_test.cc
```diff
@@ -1377,4 +1381,246 @@ TEST_F(MemBufControllerTest, TaggingDisabled) {
...
+  plain.push_back(RDB_TYPE_STRING);
+  AppendString(&plain, "plain_key");
+  AppendString(&plain, "plain_val");
+
+  // p2 of blob1
+  std::string second;
+  AppendLen(&second, second_split);
+  second.append(blob1.data() + first_split, second_split);
+
+  // p2 of blob1
```
romange: copy paste error

src/server/rdb_test.cc
```diff
@@ -1377,4 +1381,246 @@ TEST_F(MemBufControllerTest, TaggingDisabled) {
...
+    resp = Run({"BF.ADD", "bf_src", StrCat("item", i)});
+    EXPECT_THAT(resp, AnyOf(0, 1));
+  }
+
+  std::string first;
+  std::string blob1;
+
+  // split the blob of the second filter into three chunks. this exercises the loader path where we
+  // first try to load the incomplete filter, and return early before that finishes
+  constexpr size_t first_split = 17;
```
romange: naming style. Constants should be like `kFirstSplit`.

src/server/rdb_test.cc
```diff
@@ -1377,4 +1381,246 @@ TEST_F(MemBufControllerTest, TaggingDisabled) {
...
+  AppendString(&plain, "plain_val");
+
+  // p2 of blob1
+  std::string second;
+  AppendLen(&second, second_split);
+  second.append(blob1.data() + first_split, second_split);
+
+  // p2 of blob1
+  std::string third;
+  constexpr auto sum = first_split + second_split;
```
romange: `sum` is not a good name. `kPrefixConsumed` maybe?

src/server/rdb_load.cc
```diff
@@ -2096,6 +2096,17 @@ struct RdbLoader::ObjSettings {
   ObjSettings() = default;
 };
 
+// State needed to resume loading a chunked value after the first tagged chunk.
+// The first chunk carries the object type and key. Continuation chunks carry only payload,
+// so we keep the object type, key, db, settings, and pending read.
+struct RdbLoader::StreamState {
```
romange, line 2102: nit: move key to be a first field.

romange (commented):
LGTM, mostly style fixes.

FYI: some of them I got by asking claude to /review this PR.

src/server/rdb_load.cc
abhijat: fixed

src/server/rdb_test.cc
abhijat: fixed

src/server/rdb_test.cc
abhijat: changed

src/server/rdb_test.cc
abhijat: changed

src/server/rdb_load.cc
abhijat: moved to first field

src/server/rdb_load.cc
abhijat: I introduced a new helper, the reason I did not want to put `stop_early_` check into the current one is that `FinishCurrentChunk` is also called in `RdbLoaderBase::HandleCompressedBlob`

In case we mark the finish as success if `stop_early_` is true, then `HandleCompressedBlob` will go on to do decompression even if payload size is mismatched, so the `stop_early_` check is in `FinalizeCurrentChunkIfNeeded` in addition to the other checks and `HandleCompressedBlob` uses `FinishCurrentChunk`

abhijat:
afl failure https://github.com/dragonflydb/dragonfly/actions/runs/25095270375/job/73530454923?pr=7070 again seems due to infra issues, another test fails also due to absl download failure https://github.com/dragonflydb/dragonfly/actions/runs/25095270389/job/73530464092?pr=7070

romange:
https://mitchellh.com/writing/ghostty-leaving-github

src/server/rdb_load.cc
```diff
@@ -2931,13 +3000,68 @@ error_code RdbLoader::LoadKeyValPair(int type, ObjSettings* settings) {
...
+
+  // Restore the state from the previous chunk
+  pending_read_ = state.pending_read;
+
+  bool finalized = false;
+  do {
+    SET_OR_RETURN(ReadAndDispatchObject(state.type, state.key, state.settings, state.db_index),
+                  finalized);
+  } while (!finalized && current_chunk_state_->remaining_payload_bytes > 0 &&
+           !stop_early_.load(memory_order_relaxed));
```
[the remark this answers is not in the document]
abhijat: interesting, this might be a genuine bug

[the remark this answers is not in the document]
abhijat: it was a bug, db index is moved into item as suggested by the bot, and db context and slice created once per item now in `RdbLoader::LoadItemsBuffer` for the correct db index

romange:
@abhijat  is it possible to catch in rdb_test.cc ?

abhijat:
> @abhijat is it possible to catch in rdb_test.cc ?

yes I am trying to update the test to cover this

src/server/rdb_load.cc
```diff
@@ -2352,6 +2375,32 @@ error_code RdbLoader::Load(io::Source* src) {
...
+      if (state.stream_id == 0) {
+        LOG(ERROR) << "invalid stream id 0 with size " << state.remaining_payload_bytes;
+        return RdbError(errc::rdb_file_corrupted);
+      }
+
+      if (state.remaining_payload_bytes == 0) {
+        LOG(ERROR) << "invalid payload of zero size for stream id " << state.stream_id;
+        return RdbError(errc::rdb_file_corrupted);
+      }
+
```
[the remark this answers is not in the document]
abhijat: Im not sure if this is relevant

src/server/rdb_load.cc
```diff
@@ -2931,13 +3002,68 @@ error_code RdbLoader::LoadKeyValPair(int type, ObjSettings* settings) {
...
 
-  int delta_ms = (absl::GetCurrentTimeNanos() - start) / 1000'000;
+  // If the first tagged chunk did not finish the object, save enough state to resume it when
+  // the next chunk with the same stream id arrives.
+  if (!finalized && current_chunk_state_) {
+    StreamState stream_state{
+        .key = std::move(key),
+        .db_index = cur_db_index_,
+        .type = type,
+        .pending_read = pending_read_,
```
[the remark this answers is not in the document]
abhijat: fixed

abhijat:
> @abhijat is it possible to catch in rdb_test.cc ?

I added in existing `InterleavedLoad`

* create key and hash which is on shard 1, while test runs on shard 0 (because bug was in non inlined path)
* split hash into chunks, first chunk goes in db 0
* add selectdb opcode to switch db to 1 before second chunk 
* assert that hash is reassembled in db 0 

with the old code it fails with

```
/home/abhijat/dev/cc/dragonfly/src/server/rdb_test.cc:1516: Failure
Expected equality of these values:
  Run({"HGET", key, "f1"})
    Which is: nil
  "v1"
    Which is: 0x1ab9b90

/home/abhijat/dev/cc/dragonfly/src/server/rdb_test.cc:1517: Failure
Expected equality of these values:
  Run({"HGET", key, "f2"})
    Which is: nil
  "v2"
    Which is: 0x1ab9b96
```

ie key, f1=v1 and f2=v2 created on db=1 which was active when last chunk was received, and not db=0, which was supposed to hold the object

with the changes it passes

src/server/rdb_load.cc
```diff
@@ -2931,13 +3002,68 @@ error_code RdbLoader::LoadKeyValPair(int type, ObjSettings* settings) {
...
+  if (it == stream_states_.end()) {
+    LOG(ERROR) << "missing stream id " << current_chunk_state_->stream_id;
+    return RdbError(errc::rdb_file_corrupted);
+  }
+
+  StreamState& state = it->second;
+
+  // Restore the state from the previous chunk
+  pending_read_ = state.pending_read;
+
```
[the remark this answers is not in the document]
abhijat: changed to move, the restore is already done if not finalized so nothing changes there.

romange (approved):

PR #7071, opened 2026-04-05
chore: add replication stream tracing for double-apply bug

romange:
## Summary

Debugging infrastructure for the replication double-apply bug where LIST
keys end up with duplicated elements on replicas during full sync. 
This PR helps reproducing the inconsistency and provides additional DCHECKs to catch it earlier.

- Rename `SerializeBucket` -> `SerializeBucketLocked` and `OnChange` ->
  `OnChangeBlocking` to clarify these methods may preempt
- Move `DCHECK(bucket_version == snapshot_version_)` inside the per-entry
  loop in `SerializeBucketLocked` — with big value serialization the
  snapshot fiber can preempt mid-bucket, during which another fiber may
  insert a new key into the same bucket (changing the version)
- Add sanity DCHECK in `OnChangeBlocking(key)` that re-runs CVCUponInsert
  after preemptible callbacks to verify segment state consistency
- Add `VLOG(2)` trace in RDB load path for post-mortem analysis
- Add `compare_datasets` helper and `test_replication_list_double_apply`
  regression test targeting the bug with LIST-only seeder and small
  `serialization_max_chunk_size`
- Add `snapshot_test.cc` (C++ unit test skeleton for snapshot round-trip)

## Root cause (suspected)

`SerializeBucketLocked` iterates a bucket while holding `big_value_mu_`.
Big value serialization triggers a flush callback that can preempt the
fiber. During the preemption, `AddOrFindInternal` on the shard fiber
inserts a new key into the same bucket and stamps it with
`NextVersion()`. When serialization resumes, the iterator picks up the
newly inserted entry — so the key appears in both the RDB bucket data
and the journal stream (via `ConsumeJournalChange`). The replica applies
both, doubling additive operations like LPUSH.

I reproduce the crash reliably on my local machine with `taskset -c 1 pytest dragonfly/replication_test.py -s -k test_replication_list_double_apply --maxfail=1 --repeat=250`

### DIFF
--- a/tests/dragonfly/replication_test.py
+++ b/tests/dragonfly/replication_test.py
@@ -36,11 +36,111 @@ Test full replication pipeline. Test full sync with streaming changes and stable
 """
 
 
+async def compare_datasets(c_master, c_replica):
+    r_port = c_replica.connection_pool.connection_kwargs.get("port", "unknown")
+    hash_script = """
+    local type = ARGV[1]
+    local res = {}
+    for i, key in ipairs(KEYS) do
+        local hash = 0
+        if type == 'STRING' then
+            hash = dragonfly.ihash(0, false, 'GET', key)
+        elseif type == 'LIST' then
+            hash = dragonfly.ihash(0, false, 'LRANGE', key, 0, -1)
+        elseif type == 'SET' then
+            hash = dragonfly.ihash(0, true, 'SMEMBERS', key)
+        elseif type == 'ZSET' then
+            hash = dragonfly.ihash(0, false, 'ZRANGE', key, 0, -1, 'WITHSCORES')
+        elseif type == 'HASH' then
+            hash = dragonfly.ihash(0, true, 'HGETALL', key)
+        elseif type == 'JSON' then
+            hash = dragonfly.ihash(0, false, 'JSON.GET', key)
+        elseif type == 'STREAM' then
+            hash = dragonfly.ihash(0, false, 'XRANGE', key, '-', '+')
+        end
+        table.insert(res, hash)
+    end
+    return res
+    """
+    sha = await c_master.script_load(hash_script)
+    await c_replica.script_load(hash_script)
+
+    for t in SeederV2.DEFAULT_TYPES:
+        m_keys, r_keys = set(), set()
+
+        scan_type = "ReJSON-RL" if t == "JSON" else t
+        logging.info(f"Scanning keys for type {t}")
+        cursor = "0"
+        while True:
+            res = await c_master.execute_command("SCAN", cursor, "TYPE", scan_type, "COUNT", 5000)
+            cursor, keys = res[0], res[1]
+            m_keys.update(keys)
+            if int(cursor) == 0:
+                break
+
+        cursor = "0"
+        while True:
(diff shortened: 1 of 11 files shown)


tests/dragonfly/replication_test.py
```diff
@@ -156,6 +257,65 @@ async def check():
...
+removals (LPOP/RPOP) it removes one extra element.
+
+The race is most pronounced with background_snapshotting=False (normal priority snapshot
+fiber): the snapshot holds the CPU for a full time-slice, causing queued mutations to
+burst against a large batch of not-yet-visited buckets the moment it yields.
+serialization_max_chunk_size=500 forces frequent yields within that burst.
+"""
+
+
+async def test_replication_list_double_apply(df_factory: DflyInstanceFactory):
```
dranikpg, line 281: what's the functional difference form test_replication_all here? it's just a fixed set of parameters, all of which can be customized otherwise

tests/dragonfly/replication_test.py
```diff
@@ -156,6 +259,66 @@ async def check():
...
+  2. ConsumeJournalChange fires -> the same command also written as a journal blob
+
+The replica applies both: loads the bucket state, then re-executes the journal command
+on top. For additive operations (LPUSH/RPUSH) this doubles the list contents; for
+removals (LPOP/RPOP) it removes one extra element.
+
+The race is most pronounced with background_snapshotting=False (normal priority snapshot
+fiber): the snapshot holds the CPU for a full time-slice, causing queued mutations to
+burst against a large batch of not-yet-visited buckets the moment it yields.
+serialization_max_chunk_size=500 forces frequent yields within that burst.
```
dranikpg, line 277: this is something very generic and not descriptive of the bug at all

tests/dragonfly/replication_test.py
```diff
@@ -36,11 +36,111 @@
...
+        scan_type = "ReJSON-RL" if t == "JSON" else t
+        logging.info(f"Scanning keys for type {t}")
+        cursor = "0"
+        while True:
+            res = await c_master.execute_command("SCAN", cursor, "TYPE", scan_type, "COUNT", 5000)
+            cursor, keys = res[0], res[1]
+            m_keys.update(keys)
+            if int(cursor) == 0:
+                break
+
```
dranikpg, line 80: you can just modify the seeder script a little to work in batches - it's the identical code there

tests/dragonfly/replication_test.py
```diff
@@ -36,11 +36,111 @@
...
+                        k, 0, -1, withscores=True
+                    ), await c_replica.zrange(k, 0, -1, withscores=True)
+                elif t == "JSON":
+                    m_v, r_v = await c_master.execute_command(
+                        "JSON.GET", k
+                    ), await c_replica.execute_command("JSON.GET", k)
+                else:
+                    m_v, r_v = None, None
+                logging.error(f"[{t}] P{r_port}: Mismatch for key {k}: master={m_v}, replica={r_v}")
+
```
dranikpg, line 135: I understand that it's easier to ask the agent to write it - but we also have this part in SeederV1 - it gives you a diff like report of what is different

PR #7084, opened 2026-04-07
[v1.36 backport] fix(server): Save and restore cmd id on eval

abhijat:
This is a backport fix. A wider fix is already in https://github.com/dragonflydb/dragonfly/pull/6584

If the command id is not restored, a redis.call or similar can overwrite it and result in metrics being misattributed to the inner command.


abhijat:
more targeted pr https://github.com/dragonflydb/dragonfly/pull/7085 replaces this one

PR #7098, opened 2026-04-09
feat(bloom): Implement LOADCHUNK cmd for SBF

abhijat:
follow up PR after https://github.com/dragonflydb/dragonfly/pull/7092 is merged

implements LOADCHUNK command for SBF
creates a new SBF, then progressively loads data into its filters.

The command expects key,cursor,byte data

On cursor=1, the byte data is expected to be SBF metadata (fp, filter count etc). It is used to create a new empty SBF with all buffers preallocated and zeroed out.

Subsequent cursors will in the data at the specific positions in the buffers. The cursors and data are return values from previous SCANDUMP commands.

The filter should not be modified externally until the LOADCHUNK commands are finished.

### DIFF
--- a/src/core/bloom.cc
+++ b/src/core/bloom.cc
@@ -234,4 +234,102 @@ size_t SBF::MallocUsed() const {
   return res;
 }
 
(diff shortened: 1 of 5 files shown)


abhijat:
closing, new updated pr https://github.com/dragonflydb/dragonfly/pull/7169 resolves conflicts etc

PR #7104, opened 2026-04-09
feat(core): Introduce zero-alloc StreamNodeObj tagged pointer

mkaruza:
Replace raw ri.data casts throughout the stream code with StreamNodeObj,
a zero-alloc value type wrapping the void* stored in stream rax nodes.

The class uses a tag bit scheme (bit 52) to distinguish encodings,
laying the groundwork for future compressed or tiered node types without
changing the rax storage layout.

### DIFF
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -759,7 +759,6 @@ void RdbLoaderBase::OpaqueObjLoader::CreateStream(const LoadTrace* ltrace) {
     CHECK(lpFirst(lp) != NULL);
     uint8_t* copy_lp = (uint8_t*)zmalloc(data.size());
     ::memcpy(copy_lp, lp, data.size());
-    /* Insert the key in the radix tree. */
     int retval =
         raxTryInsert(s->rax, (unsigned char*)nodekey.data(), nodekey.size(), copy_lp, NULL);
     if (!retval) {
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -767,6 +767,7 @@ void RdbLoaderBase::OpaqueObjLoader::CreateStream(const LoadTrace* ltrace) {
     CHECK(lpFirst(lp) != NULL);
     uint8_t* copy_lp = (uint8_t*)zmalloc(data.size());
     ::memcpy(copy_lp, lp, data.size());
+    /* Insert the key in the radix tree. */
     int retval =
         raxTryInsert(s->rax, (unsigned char*)nodekey.data(), nodekey.size(), copy_lp, NULL);
     if (!retval) {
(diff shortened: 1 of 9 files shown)


romange:
It's one way to do it. I actually like how Redis did that with `quicklistLZF`. Basically re-purposing the pointer for a differrent struct instead of wrapping it. We could use pointer tagging - effectively creating polymorphism - we do that in dense_set and in many other places. listpack is a regular pointer so it's at least 8 bytes aligned. That's 3 bits unused. Also, linux memory model guarantees that 12 high bits are unused.   

Another, less intrusive option is to change listpack format and use listpack format itself as an envelop for compression: 
```
* +-----+--------+----------+-------+-------+-/-+-------+-------+--------+
   * |flags|entry-id|num-fields|field-1|value-1|...|field-N|value-N|lp-count|
   * +-----+--------+----------+-------+-------+-/-+-------+-------+--------+
   
````

The second option is less intrusive as otherwise you will need to change stream iterator and all the places that access data. 


Finally, when working on list compression, before I wrote anything - I wanted to persuade myself that it actually works. I wrote a python code that dumps list entities after they were filled with celery jobs into files, then used zstd program to create a dictionary of size 4-8KB and then I showed how I can compress significantly those files. 

I would love to see something similar that proves the viability of this direction with streams. We can use bullmq to fill up a stream entry, dump the stream, create a dictionary from the dump, and use the dictionary to compress other small queues, that have a single listpack node. The goal is to show that a having a global dictionary is a good approach for compressing each listpack node independently.

**Comment by romange** [comment]
regarding node wrapping - I think it's the worst option among:

1. using listpack as a runtime struct  (this is how they evolved the format so far).
2. using pointer tagging to distinguish between differrent formats
3. adding a constant header before listpack - which you effectively do with `StreamNode` wrapping: you can just allocate K  more bytes and shift listpack by K bytes and use first K bytes for the header. K=4 in your case.
4. Using `StreamNode` struct wrapping

but if you disagree we can discuss cons and pros.

File: src/server/rdb_load.cc [hunk 1]
```diff
@@ -5,6 +5,7 @@
 #include "server/rdb_load.h"
 
 #include "absl/strings/escaping.h"
+#include "core/stream_node.h"
```
romange: Seems that this file should not change.

src/server/stream_family.cc
```diff
@@ -909,10 +916,13 @@ int StreamAppendItem(stream* s, CmdArgList fields, uint64_t now_ms, streamID* ad
   uint8_t rax_key[16]; /* Key in the radix tree containing the listpack.*/
   streamID master_id;  /* ID of the master entry in the listpack. */
 
+  void* current_node = nullptr;  // Tagged rax value for the node we are appending to.
```
romange: Nit: define `StreamNodeObj current_node;`  instead of having void* here.

romange (dismissed):

romange (approved):

src/server/rdb_load.cc
```diff
@@ -759,7 +759,6 @@ void RdbLoaderBase::OpaqueObjLoader::CreateStream(const LoadTrace* ltrace) {
     CHECK(lpFirst(lp) != NULL);
     uint8_t* copy_lp = (uint8_t*)zmalloc(data.size());
     ::memcpy(copy_lp, lp, data.size());
-    /* Insert the key in the radix tree. */
```
romange, line 770: revert

mkaruza: reverted.

romange (approved):

PR #7120, opened 2026-04-11
fix: replica stuck during flushall

BorysTheDev:
Fixed 2 bugs:
Bug 1 (in MultiShardExecution): Race between CancelAllBlockingEntities() and InsertTxToSharedMap() —
  a new entry created after cancellation was never cancelled, causing block->Wait() / barrier.Wait() to hang
  forever.

Bug 2 (in ExecuteTx): After barrier/block waits, the cancellation checks used
  exec_st_.IsRunning() — the flow's own ExecutionState, which is never set to ERROR by the error handler. Should
  use cntx->IsRunning() — the Replica's ExecutionState, which IS set to ERROR when the connection breaks. This
  caused flows to continue executing global commands (like FLUSHALL via executor_->Execute()) even after
  cancellation, potentially blocking on the cross-shard transaction scheduling.

Changes:

- Track a cancellation flag in MultiShardExecution and immediately cancel any tx sync entry inserted after cancellation.
- Set the cancellation flag when CancelAllBlockingEntities() runs.
- In DflyShardReplica::ExecuteTx, use the replica’s ExecutionState (cntx) for post-wait cancellation checks instead of the flow-local state.
- Add a regression test that kills the master during FLUSHALL under load and verifies the replica can recover and resync.

### DIFF
--- a/src/server/journal/tx_executor.cc
+++ b/src/server/journal/tx_executor.cc
@@ -18,6 +18,12 @@ namespace dfly {
 bool MultiShardExecution::InsertTxToSharedMap(TxId txid, uint32_t shard_cnt) {
   std::unique_lock lk(map_mu);
   auto [it, was_insert] = tx_sync_execution.emplace(txid, shard_cnt);
+  // If CancelAllBlockingEntities() already ran, cancel the new entry immediately
+  // to prevent shard fibers from blocking on entries that will never complete.
+  if (cancelled_) {
+    it->second.barrier.Cancel();
+    it->second.block->Cancel();
+  }
   lk.unlock();
 
   VLOG(2) << "txid: " << txid << " unique_shard_cnt_: " << shard_cnt
@@ -42,6 +48,7 @@ void MultiShardExecution::Erase(TxId txid) {
 
 void MultiShardExecution::CancelAllBlockingEntities() {
   lock_guard lk{map_mu};
+  cancelled_ = true;
   for (auto& tx_data : tx_sync_execution) {
     tx_data.second.barrier.Cancel();
     tx_data.second.block->Cancel();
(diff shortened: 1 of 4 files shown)


src/server/journal/tx_executor.cc
```diff
@@ -18,6 +18,12 @@ namespace dfly {
 bool MultiShardExecution::InsertTxToSharedMap(TxId txid, uint32_t shard_cnt) {
   std::unique_lock lk(map_mu);
   auto [it, was_insert] = tx_sync_execution.emplace(txid, shard_cnt);
```
romange, line 20: why do we need to insert it and then cancel? can we just not insert it?

romange (approved):

src/server/journal/tx_executor.cc
BorysTheDev: As I understand it, we can't skip insert, because we have next line further in the code `it->second.block->Dec();` Other shard fibers for the same txid may have already inserted and are blocked on block->Wait()

src/server/journal/tx_executor.cc
```diff
@@ -42,6 +48,7 @@ void MultiShardExecution::Erase(TxId txid) {
 
 void MultiShardExecution::CancelAllBlockingEntities() {
   lock_guard lk{map_mu};
+  cancelled_ = true;
```
kostasrim, line 51: This flow is called by the registered error handler and the replication state machine might not necessarily exit. 

In other words, `cancelled_` will be `true` even for cases where the replication state machine won't exit (see main replication fiber). So if the state machine resets, `InsertTxToSharedMap` will *always* fail after the flow got cancelled for the first time.

kostasrim: -1 on my comment above. MultiShardExecution is `reset` every time in the main loop so the flag does not really stick (even though it was implied from the PR changes)

It does not `stick` and this is `safe`.

PR #7123, opened 2026-04-12
feat(helm): add cluster mode values for Dragonfly chart

### LINKED ISSUE #3861: No predefined configuration for cluster mode in helm chart
The default configuration didn't fit for the production purposes because on high payload it gets timeout. At this point we tried to create HPA that also didn't help because the instances doesn't balancing the requests and still got timeout. Next thing we wanted to try is cluster mode, but we didn't find a solution to configure cluster mode with helm chart because there is no values for that.

Helm chart values for scalable solution

**Additional context**
The values we're using right now:
`podSecurityContext:
  fsGroup: 2000

securityContext:
  capabilities:
    drop:
      - ALL
  readOnlyRootFilesystem: false
  runAsNonRoot: true
  runAsUser: 1000

serviceMonitor:
  enabled: true
  annotations:
    prometheus.io/path: "/metrics"
    prometheus.io/port: "4999"
    prometheus.io/scrape: "true"
passwordFromSecret:
  enable: true
  existingSecret:
    name: "dragonfly"
    key: "DRAGONFLY_PASSWORD"
tls:
  enabled: true
  existing_secret: "certificate-default"
extraObjects:
  - apiVersion: autoscaling/v2
    kind: HorizontalPodAutoscaler
    metadata:
      name: dragonfly-hpa
      namespace: dragonfly
    spec:
      scaleTargetRef:
        apiVersion: apps/v1
        kind: Deployment
        name: dragonfly
      minReplicas: 1
      maxReplicas: 5
      metrics:
        - type: Resource
          resource:
            name: memory
            target:
              type: Utilization
              averageUtilization: 60
        - type: Resource
          resource:
            name: cpu
            target:
              type: Utilization
              averageUtilization: 80
      behavior:
        scaleUp:
          stabilizationWindowSeconds: 30
          policies:
            - type: Percent
              value: 10
              periodSeconds: 60
        scaleDown:
          stabilizationWindowSeconds: 30
          policies:
            - type: Percent
              value: 10
              periodSeconds: 60
`

VedantMadane:
## Summary
Adds predefined Helm values for Redis-compatible cluster mode so users can enable Dragonfly cluster flags without hand-writing `extraArgs`.

## Changes
- `values.yaml`: new `cluster.mode` (maps to `--cluster_mode`) and `cluster.experimentalShardBySlot` (maps to `--experimental_cluster_shard_by_slot=true` when mode is set)
- `templates/_pod.tpl`: append the corresponding CLI args when `cluster.mode` is non-empty

## Notes
Valid `cluster.mode` values match the Dragonfly flag: `emulated`, `yes`, or empty to leave cluster mode disabled. See server `cluster_support.cc` for flag semantics.

Fixes https://github.com/dragonflydb/dragonfly/issues/3861

### DIFF
--- a/contrib/charts/dragonfly/templates/_pod.tpl
+++ b/contrib/charts/dragonfly/templates/_pod.tpl
@@ -86,6 +86,21 @@ containers:
     {{- with .Values.extraArgs }}
       {{- toYaml . | trim | nindent 6 }}
     {{- end }}
+    {{- with .Values.cluster }}
+    {{- $modeRaw := .mode | default "" }}
+    {{- $clusterMode := "" }}
+    {{- if kindIs "bool" $modeRaw }}
+      {{- if $modeRaw }}{{- $clusterMode = "yes" }}{{- end }}
+    {{- else }}
+      {{- $clusterMode = $modeRaw | toString }}
+    {{- end }}
+    {{- if ne $clusterMode "" }}
+      - "--cluster_mode={{ $clusterMode }}"
+    {{- end }}
+    {{- if and (ne $clusterMode "") .experimentalShardBySlot }}
+      - "--experimental_cluster_shard_by_slot=true"
+    {{- end }}
+    {{- end }}
     {{- if .Values.tls.enabled }}
       - "--tls"
       - "--tls_cert_file=/etc/dragonfly/tls/tls.crt"
--- a/contrib/charts/dragonfly/values.yaml
+++ b/contrib/charts/dragonfly/values.yaml
@@ -173,6 +173,16 @@ command: []
 # -- Extra arguments to pass to the dragonfly binary
 extraArgs: []
 
+cluster:
+  # -- Redis cluster protocol mode passed to Dragonfly as --cluster_mode.
+  # -- Accepts "emulated" or "yes". Unquoted `yes` (parsed as boolean by YAML)
+  # -- is also supported and treated as the string "yes".
+  # -- Leave empty to disable cluster mode.
+  mode: ""
+  # -- When cluster mode is enabled, pass --experimental_cluster_shard_by_slot=true
+  # -- so sharding follows slots instead of hash tags.
+  experimentalShardBySlot: false
+
 # -- Extra volumes to mount into the pods
 extraVolumes: []
 


romange:
augment review

contrib/charts/dragonfly/templates/_pod.tpl
```diff
@@ -145,4 +154,4 @@ volumes:
   {{- toYaml . | trim | nindent 2 }}
 {{- end }}
 {{- end }}
-{{- end }}
+{{- end }}
```
Abhra303: add back newline at the end of the file

contrib/charts/dragonfly/values.yaml
```diff
@@ -223,4 +232,4 @@ tolerations: []
 affinity: {}
 
 # -- Topology Spread Constraints for pod assignment
-topologySpreadConstraints: []
+topologySpreadConstraints: []
```
Abhra303: add newline back at the end of this file.

contrib/charts/dragonfly/values.yaml
```diff
@@ -173,6 +173,15 @@ command: []
 # -- Extra arguments to pass to the dragonfly binary
 extraArgs: []
 
+cluster:
+  # -- Redis cluster protocol mode passed to Dragonfly as --cluster_mode.
+  # -- Use "emulated" or "yes" (must be quoted in YAML to avoid boolean coercion).
```
Abhra303: If the user passes unquoted yes, the template won't validate this and and df will fail. I suggest to either validate the input and fail if the input is not string or support boolean yes as well. As you're adding a cluster specific fields, it makes sense to support boolean yes as well.

for boolean yes support :
```
  {{- with .Values.cluster }}                                                                                                                                                                              
  {{- $modeRaw := .mode | default "" }}
  {{- $clusterMode := "" }}
  {{- if kindIs "bool" $modeRaw }}
    {{- if $modeRaw }}{{- $clusterMode = "yes" }}{{- end }}                                                                                                                                                
  {{- else }}
    {{- $clusterMode = $modeRaw | toString }}                                                                                                                                                              
  {{- end }}                                                                                                                                                                                               
  {{- if ne $clusterMode "" }}
    - "--cluster_mode={{ $clusterMode }}"                                                                                                                                                                  
  {{- end }}  
```

Or, if you want to validate:

```
  {{- with .Values.cluster }}
  {{- $modeRaw := .mode | default "" }}                                                                                                                                                                    
  {{- if kindIs "bool" $modeRaw }}
    {{- fail "cluster.mode must be a quoted string — use '\"yes\"' or '\"emulated\"' (unquoted 'yes' is parsed as a YAML boolean)" }}
  {{- end }}                                                                                                                                                                                               
  {{- $clusterMode := $modeRaw | toString }}

```

Abhra303 (changes requested):
Thanks for the PR, see some comments.

Abhra303:
Also wondering we have `extraArgs` which should be enough to add cluster mode values. Why do you think it is not viable to set cluster mode via extraArgs?

contrib/charts/dragonfly/templates/_pod.tpl
VedantMadane: Done, restored the trailing newline.

contrib/charts/dragonfly/values.yaml
VedantMadane: Done.

contrib/charts/dragonfly/values.yaml
VedantMadane: Good call, went with the boolean coercion approach. If `.mode` is a bool (unquoted `yes` in YAML), it now maps to the string `"yes"`. Updated the values.yaml comment to document that behavior too.

Abhra303 (approved):
Looks good.

romange (approved):

romange:
@VedantMadane  please squash and sign your commits. See this for details:

https://github.com/dragonflydb/dragonfly/blob/31a8fd362a7cea0e1beca371a34cc1e7cdb69277/CONTRIBUTING.md#L35-L53

VedantMadane:
You raise a fair point about `extraArgs`. The difference is that dedicated `cluster.mode` values let the chart validate input (the boolean coercion fix), set sensible defaults for related settings (like replica count), and keep the values.yaml self-documenting for users who `helm show values` the chart. `extraArgs` would work but pushes validation onto the user.

That said, if you prefer the extraArgs-only approach, I am happy to simplify the PR to just the documentation changes. Let me know which direction you would like.

Regarding squash and sign-off: will do that now.

romange:
@VedantMadane  it's still not signed

romange:
@VedantMadane still not signed. You need a gpg verified signature.

VedantMadane:
@snejus key for my email address is added and the commit's verified now. cleaned the history too. ci should be good.

romange:
thanks!

PR #7124, opened 2026-04-12
feat(memory): add MEMORY DECOMMIT COOL subcommand

romange:
## Summary

Extend the MEMORY DECOMMIT command to support flushing the tiered storage cool queue via a new COOL option. This allows operators to explicitly convert cooled entries (in-memory copies of offloaded values) to fully disk-backed state, reclaiming their memory.

## Changes

- Added `MEMORY DECOMMIT COOL` subcommand that flushes the cool storage queue on all shards
- Calls `TieredStorage::ReclaimMemory(SIZE_MAX)` to drain cool_queue_ entries
- Each cooled entry is frozen to disk-only state and its TieredCoolRecord is freed from memory
- Updated HELP text to document the new option
- Added comprehensive test in tiered_storage_test.cc that verifies the command works correctly

## Test Plan

- Run tiered_storage_test suite: all tests pass including new MemoryDecommitCool test
- Test runs with both cooling enabled and disabled variants
- Verifies cold_storage_bytes metric drops to 0 after command execution

🤖 Generated with Claude Code

### DIFF
--- a/src/server/memory_cmd.cc
+++ b/src/server/memory_cmd.cc
@@ -7,8 +7,6 @@
 #include <absl/strings/ascii.h>
 #include <absl/strings/str_cat.h>
 
-#include <limits>
-
 #ifdef __linux__
 #include <malloc.h>
 #endif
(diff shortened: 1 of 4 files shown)


src/server/memory_cmd.cc
```diff
@@ -249,6 +253,16 @@ void MemoryCmd::Run(CmdArgList args) {
   }
 
   if (parser.Check("DECOMMIT")) {
+    if (parser.Check("COOL")) {
+      shard_set->pool()->AwaitFiberOnAll([](util::ProactorBase*) {
+        if (auto* shard = EngineShard::tlocal(); shard) {
+          if (auto* ts = shard->tiered_storage(); ts) {
+            ts->ReclaimMemory(std::numeric_limits<size_t>::max());
```
vyavdoshenko: nit:
Use SIZE_MAX instead of std::numeric_limits<size_t>::max() (avoids extra include)

vyavdoshenko (dismissed):

vyavdoshenko (dismissed):

vyavdoshenko (approved):

PR #7125, opened 2026-04-12
fix: prevent SIGSEGV during HNSW index restoration with external vectors

vyavdoshenko:
During a full replica sync, the HNSW graph structure is restored from RDB with all nodes marked DELETED. In borrowed-vector mode (dim >= 256), data pointers are left as nullptr. hnswlib's graph traversal computes distances even to DELETED nodes (entry point, neighbor exploration), causing nullptr dereference when concurrent writes or searches trigger addPoint()/searchKnn().

Root cause: the HNSW graph is global (shared across shards), but vector restoration runs per-shard. When one shard finishes and accepts new writes, addPoint() traverses nodes from other shards that still have nullptr pointers.

Changes:
- Set restored borrowed-vector nodes to point to a non-zero placeholder (1.0f) instead of nullptr so distance computation is safe during the restoration window. Uses 1.0f rather than zeros because CosineDistance() returns 0 for zero-norm vectors, which would bias graph traversal toward DELETED nodes.
- Make FieldIndices::Remove() tolerate docs not yet indexed by CursorLoop, fixing a DCHECK crash when journal events race with index rebuilding
- Skip duplicate FieldIndices::Add() in CursorLoop for docs already indexed by a concurrent journal event
- Allow HnswShardIndex::RemoveById() to defer when the MRMW lock is held, fixing a DCHECK crash during RestoreGlobalVectorIndices()
- Remove skip marker from test_hnsw_search_replication_with_network_disruptions
- Add minimal reproducer test_hnsw_external_vector_replication_crash

### DIFF
--- a/src/core/search/hnsw_index.cc
+++ b/src/core/search/hnsw_index.cc
@@ -436,11 +436,23 @@ struct HnswlibAdapter {
              world_.size_data_per_element_);
       world_.setExternalLabel(internal_id, node.global_id);
 
-      // In copy mode, zero the vector memory so distance computations don't use
-      // uninitialized data for nodes that are marked deleted.
       if (world_.copy_vector_) {
+        // In copy mode, zero the vector memory so distance computations don't use
+        // uninitialized data for nodes that are marked deleted.
         char* data_ptr = world_.data_vector_memory_ + internal_id * world_.data_size_;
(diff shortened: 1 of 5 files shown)


BorysTheDev:
@vyavdoshenko let me fix this issue, it is not the fix that I want to do. Because we should defer index processing for all operations.

PR #7126, opened 2026-04-12
feat: deffer HNSW index updates during restoration

BorysTheDev:
Fixes SIGSEGV during HNSW vector index replication when journal events arrive while the graph is being restored
   with uninitialized vector data.                                                                               
                                                                                                                 
  Problem                                                         

  During full sync replication, the HNSW graph is restored from the RDB stream with all nodes marked as deleted  
  and vector data uninitialized. Journal events embedded in the RDB stream (or arriving during LOADING) can
  trigger addPoint/searchBaseLayer which traverses these nodes and dereferences null pointers in borrowed-vector 
  mode — causing a crash.                                         

  The root cause has two parts:                                                                                  
  1. Timing gap: ShardDocIndex was created during LOADING with is_restoring_vectors_=false (default). The flag
  was only set to true by Rebuild(), which runs later in PerformPostLoad. Journal events arriving in between     
  bypassed the guard.                                                                                       
  2. Cross-shard traversal: The HNSW graph is global (shared across shards) but restoration runs per-shard. When 
  one shard completes and starts accepting writes, addPoint traverses nodes from other shards that still have
  null pointers.                                                                                                 
   
  Changes                                                                                                        
                                                                  
  - 3-state lifecycle enum (kProhibit/kRestoring/kBuilding) replaces bool is_restoring_vectors_ on ShardDocIndex.
   Default kProhibit blocks all HNSW ops from index creation, closing the timing gap.
  - stub_vector_ for deleted nodes in RestoreFromNodes: borrowed-mode nodes point to a safe 1.0f-filled buffer   
  instead of nullptr, preventing SIGSEGV during cross-shard graph traversal.                                     
  - Deferred drain in PerformPostLoad: pending vector updates are drained only after BlockUntilConstructionEnd
  ensures ALL shards completed, so addPoint never traverses uninitialized nodes.                                 
  - Field preservation during restoration: when journal HSETs modify documents during kRestoring, the old sds
  entries are preserved (via ExtractField) so HNSW pointers set by UpdateVectorData remain valid.                
  - Extracted MaybePreserveField: deduplicates the sds preservation logic between HnswShardIndex::Remove and
  RemoveDocFromGlobalVectorIndex.

### DIFF
--- a/src/core/search/hnsw_index.cc
+++ b/src/core/search/hnsw_index.cc
@@ -486,9 +486,15 @@ struct HnswlibAdapter {
       world_.cur_element_count.store(++restored_count);
 
       // Mark node as deleted until UpdateVectorData provides valid vector data.
-      // This prevents crashes from dereferencing uninitialised data pointers
-      // (especially in borrowed-vector mode).
       world_.markDeletedInternal(internal_id);
+
+      // In borrowed mode, deleted nodes are still traversed by addPoint.
+      // Point to stub_vector_ so distance computations don't dereference nullptr.
+      if (!copy_vector_) {
+        const char* safe_ptr = reinterpret_cast<const char*>(stub_vector_.data());
+        char* ptr_location = world_.getDataPtrByInternalId(internal_id);
+        memcpy(ptr_location, &safe_ptr, sizeof(void*));
+      }
     }
 
     // Set the metadata for the graph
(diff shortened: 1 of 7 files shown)


BorysTheDev:
augment review

BorysTheDev:
augment review

vyavdoshenko (changes requested):
Pull the test from this branch: bobik/hnsw_restoration_fix
Try it out: test_hnsw_external_vector_replication_crash

It crashes in the same place. I tried it on your branch. I strongly recommend adding this test as well.

vyavdoshenko (approved):
lgtm

PR #7130, opened 2026-04-13
refactor(rdb): use std::bit_cast (Fixes private #188)

glevkovich:
Replace reinterpret_cast/union type-punning with std::bit_cast in SaveBinaryDouble/FetchBinaryDouble.

The previous patterns violate C++ strict-aliasing rules, triggering undefined behavior that compilers may exploit at -O3 optimization levels.

std::bit_cast (C++20) provides safe, well-defined bit-level type reinterpretation.

### DIFF
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -1169,9 +1169,6 @@ io::Result<double> RdbLoaderBase::FetchBinaryDouble() {
   uint8_t buf[8];
   mem_buf_->ReadAndConsume(8, buf);
   uint64_t val = base::LE::LoadT<uint64_t>(buf);
-
-  // std::bit_cast is the C++20-sanctioned way to reinterpret the bits of a double.
-  // Using reinterpret_cast/union approach is an undefined behavior under strict aliasing.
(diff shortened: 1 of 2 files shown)


src/server/rdb_load.cc
```diff
@@ -1161,20 +1162,17 @@ auto RdbLoaderBase::FetchIntegerObject(int enctype) -> io::Result<string> {
...
 
   uint8_t buf[8];
   mem_buf_->ReadAndConsume(8, buf);
-  u.val = base::LE::LoadT<uint64_t>(buf);
-  return u.d;
+  uint64_t val = base::LE::LoadT<uint64_t>(buf);
+
+  // std::bit_cast is the C++20-sanctioned way to reinterpret the bits of a double.
+  // Using reinterpret_cast/union approach is an undefined behavior under strict aliasing.
+  return std::bit_cast<double>(val);
```
dranikpg: no need for a comment why you use the right way 🙂

dranikpg (dismissed):

abhijat (dismissed):

dranikpg (approved):

PR #7136, opened 2026-04-14
fix: memory calculation after GC

BorysTheDev:
Summary: Fixes Lua interpreter pool memory accounting so forced GC decisions reflect net memory change after GC.

Changes:

- In InterpreterManager::Return(), track a signed delta from TakeUsedBytes() and incorporate the post-RunGC() delta.
- Update tl_stats().used_bytes using the net delta and adjust VLOG output accordingly.
- Add GcAccountingAfterReturn test to ensure freed bytes are drained and repeated forced GCs don’t occur.

### DIFF
--- a/src/core/interpreter_test.cc
+++ b/src/core/interpreter_test.cc
@@ -9,6 +9,8 @@ extern "C" {
 #include <lua.h>
 }
 
+#include <absl/flags/flag.h>
+#include <absl/flags/reflection.h>
 #include <absl/strings/str_cat.h>
 #include <absl/strings/str_replace.h>
 #include <gmock/gmock.h>
@@ -22,6 +24,8 @@ extern "C" {
 extern "C" {
 #include "redis/zmalloc.h"
 }
+
+ABSL_DECLARE_FLAG(uint64_t, lua_mem_gc_threshold);
 
 namespace dfly {
 using namespace std;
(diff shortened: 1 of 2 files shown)


src/core/interpreter_test.cc
```diff
@@ -600,4 +604,38 @@ TEST_F(InterpreterTest, LuaGcStatistic) {
...
+            interpreter->AddFunction({sha_buf, strlen(sha_buf)}, script, &result));
+  ASSERT_EQ(Interpreter::RUN_OK, interpreter->RunFunction({sha_buf, strlen(sha_buf)}, &error_));
+
+  auto& stats = InterpreterManager::tl_stats();
+  uint64_t gc_before = stats.force_gc_calls;
+  im.Return(interpreter);
+  ASSERT_GT(stats.force_gc_calls, gc_before);
+
+  // Re-borrow: if Return() drained used_bytes_ after GC, residual is 0.
+  // Bug: without the fix, residual is ~-300KB (stale freed bytes never drained).
```
romange, line 633: never drained from where?

romange (approved):

src/core/interpreter_test.cc
BorysTheDev: from tl_stats().used_bytes, in previous line, description

dranikpg (approved):

PR #7137, opened 2026-04-14
fix: delete empty hash after field expiry in FIELDTTL

### LINKED ISSUE #7133: SIGABRT in rdb_save.cc:274
[fuzz-long-resp-crashes-79.zip](https://github.com/user-attachments/files/26697759/fuzz-long-resp-crashes-79.zip)

```
./fuzz/triage_crashes.sh ./build-dbg/dragonfly resp ./fuzz-long-resp-crashes-79.zip
```

```
F20260414 08:28:06.961647   765 rdb_save.cc:274] SaveEntry skipped empty PrimeValue with key: mylist with tag 19
```

vyavdoshenko:
FIELDTTL triggers lazy expiry inside StringMap::Find(), which removes expired fields.  When the last field expires, the hash becomes empty (Size()==0), but the key stays in the database.  A subsequent SAVE hits the DFATAL in SaveEntry (rdb_save.cc:274) because OBJ_HASH does not allow empty values, crashing the server.

The fix mirrors what already exists for sets (DeleteSetIfEmpty) and for HEXPIRE/HPERSIST: after FieldExpireTime() returns, check whether the StringMap is empty and delete the key.

Fixes #7133

### DIFF
--- a/src/server/generic_family.cc
+++ b/src/server/generic_family.cc
@@ -897,6 +897,7 @@ OpResult<long> OpFieldTtl(Transaction* t, EngineShard* shard, string_view key, s
   } else {
     DCHECK_EQ(OBJ_HASH, it->second.ObjType());
     res = HSetFamily::FieldExpireTime(db_cntx, it->second, field);
+    HSetFamily::DeleteIfEmpty(db_slice, db_cntx, key, it->second);
   }
   return res <= 0 ? res : int32_t(res - MemberTimeSeconds(db_cntx.time_now_ms));
 }
--- a/src/server/hset_family.cc
+++ b/src/server/hset_family.cc
@@ -31,6 +31,7 @@ extern "C" {
 #include "server/tiering/decoders.h"
 #include "server/tiering/serialized_map.h"
 #include "server/transaction.h"
+#include "server/tx_base.h"
 
 using namespace std;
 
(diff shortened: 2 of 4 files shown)


src/server/generic_family.cc
```diff
@@ -897,6 +898,13 @@ OpResult<long> OpFieldTtl(Transaction* t, EngineShard* shard, string_view key, s
   } else {
     DCHECK_EQ(OBJ_HASH, it->second.ObjType());
     res = HSetFamily::FieldExpireTime(db_cntx, it->second, field);
+    if (it->second.Encoding() == kEncodingStrMap2) {
```
romange: it should not be done like this this logic should be encapsulated within HSetFamily code.

I suggest expanding `FieldExpireTime` API to return a special result code that says - empty map.
Maybe this is why we have this bug again as we fixed it in the wrong place in the past.

vyavdoshenko: fixed

BorysTheDev (dismissed):

src/server/hset_family.cc
```diff
@@ -1227,6 +1227,21 @@ int32_t HSetFamily::FieldExpireTime(const DbContext& db_context, const PrimeValu
...
+bool HSetFamily::DeleteIfEmpty(DbSlice& db_slice, const DbContext& db_cntx, std::string_view key,
+                               const PrimeValue& pv) {
+  if (pv.Encoding() != kEncodingStrMap2)
+    return false;
+
+  if (auto* sm = static_cast<StringMap*>(pv.RObjPtr()); !sm->Empty())
+    return false;
+
+  if (auto res = db_slice.FindMutable(db_cntx, key, OBJ_HASH); res) {
+    db_slice.DelMutable(db_cntx, std::move(*res));
```
romange, line 1240: check out `SetFamily::DeleteSetIfEmpty`. you may need to call `RecordDelete` to propagate the deletion.

romange (changes requested):

src/server/hset_family.cc
vyavdoshenko: fixed

romange (approved):

PR #7143, opened 2026-04-14
feat(facade):  add connection migration support for IO loop V2

glevkovich:
This Pr includes 2 separate commits:

1) feat(facade): support V2 loop conn migration

Enable CLIENT MIGRATE for connections using the experimental IoLoopV2.

C++ core changes:
* Extract MaybeEnableRecvMultishot() helper and invoke it after
  migration in OnPostMigrateThread() to re-arm io_uring.
* Remove ioloop_v2_ early-return in RequestAsyncMigration and trigger
  io_event_.notify() so the V2 loop wakes up immediately.
* Lift DCHECK(!ioloop_v2_) in HandleMigrateRequest and safely guard
  the V1-only async_fb_.Join() call.
* Add migration_request_ to io_event_.await() predicates so the loop
  breaks out of read/backpressure park states.
* Handle MigrationRequestMessage in dispatch_q by breaking back to
  the top-of-loop migration check.

Test additions & infrastructure:
* Extend migration and pipeline batching tests with explicit V1/V2
  variants using @dfly_multi_test_args.
* Add test_migration_during_backpressure to verify migrating a
  connection while parked on pipeline backpressure.
* Add test_pipeline_global_memory_relief to verify CONFIG SET safely
  wakes parked V2 connections via NotifyPipelineWaiters.

2) Fix two bugs in ClientMigrate:

a. Dispatcher fiber sleep violation (AwaitBrief misuse)
   AwaitBrief executes callbacks directly on the dispatcher fiber,
   which must never sleep. ClientMigrate's callback calls
   TraverseConnectionsOnThread, which yields via SleepFor to
   prevent event-loop starvation. This caused "Should not preempt
   dispatcher" CHECK failures.

   Fixed by using AwaitFiberOnAll, which wraps the callback in a
   dedicated fiber allowed to sleep.

b. Data race on migrated counter
   AwaitFiberOnAll runs callbacks concurrently on all proactor
   threads. The bare `unsigned migrated = 0` with `++migrated` is
   a textbook data race (UB), even though only one thread should
   find the matching client ID.

   Fixed by using std::atomic<unsigned> with relaxed ordering.

Both bugs were uncovered by test_client_migrate_no_conn_leak,
which migrates 20 clients concurrently.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -697,6 +697,7 @@ void Connection::OnPostMigrateThread() {
   }
 
   if (ioloop_v2_ && socket_ && socket_->IsOpen() && migration_allowed_to_register_) {
+    MaybeEnableRecvMultishot();
     socket_->RegisterOnRecv([this](const FiberSocketBase::RecvNotification& n) {
       NotifyOnRecv(n);
       io_event_.notify();
@@ -1390,22 +1391,28 @@ void Connection::OnBreakCb(int32_t mask) {
 }
 
 void Connection::HandleMigrateRequest() {
-  if (cc_->conn_closing || !migration_request_)
+  if (!migration_request_)
     return;
 
-  DCHECK(!ioloop_v2_);
+  if (cc_->conn_closing) {
+    migration_request_ = nullptr;
+    return;
+  }
+
   ProactorBase* dest = migration_request_;
-  if (async_fb_.IsJoinable()) {
+
+  // V1 loop: we must coordinate with and shutdown the AsyncFiber before migration.
+  // V2 loop: is single-fiber and handles this by breaking the dispatch_q loop.
+  if (!ioloop_v2_ && async_fb_.IsJoinable()) {
     SendAsync({MigrationRequestMessage{}});
     async_fb_.Join();
   }
 
-  // We don't support migrating with subscriptions as it would require moving thread local
-  // handles. We can't check above, as the queue might have contained a subscribe request.
-
+  // We don't support migrating with subscriptions as it would require moving thread-local handles.
   if (cc_->subscriptions == 0) {
     // RegisterOnErrorCb might be called on POLLHUP and the join above is a preemption point.
     // So, it could be the case that after this fiber wakes up the connection might be closing.
+    // Re-check closing status after the potential V1 Join() or any preemption.
     if (cc_->conn_closing) {
       return;
     }
@@ -1413,15 +1420,17 @@ void Connection::HandleMigrateRequest() {
     tl_facade_stats->conn_stats.num_migrations++;
     migration_request_ = nullptr;
 
-    // We need to return early as the socket is closing and IoLoop will clean up.
-    // The reason that this is true is because of the following DCHECK
-    DCHECK(!async_fb_.IsJoinable());
+    // Verify that no background command processing is active.
+    // V1 loop: Join() ensures this.
+    // V2 loop: it is guaranteed by the single-fiber loop.
+    DCHECK(ioloop_v2_ || !async_fb_.IsJoinable());
 
-    // which can never trigger since we Joined on the async_fb_ above and we are
-    // atomic in respect to our proactor meaning that no other fiber will
-    // launch the DispatchFiber.
     std::ignore = !this->Migrate(dest);
   }
+
+  // Note: If cc_->subscriptions > 0, we skip the hop but leave migration_request_
+  // set. This defers the migration, retrying at the start of every subsequent
+  // loop iteration until all subscriptions are cleared.
 }
 
 io::Result<size_t> Connection::HandleRecvSocket() {
@@ -2178,13 +2187,18 @@ facade::ConnectionContext* Connection::cntx() {
 }
 
 void Connection::RequestAsyncMigration(util::fb2::ProactorBase* dest, bool force) {
-  if ((!force && !migration_enabled_) || cc_ == nullptr || ioloop_v2_) {
+  if ((!force && !migration_enabled_) || cc_ == nullptr) {
     return;
   }
 
   // Connections can migrate at most once.
   migration_enabled_ = false;
   migration_request_ = dest;
+
+  // Wake up the V2 loop so it can immediately process the migration request.
+  if (ioloop_v2_) {
+    io_event_.notify();
+  }
 }
 
 void Connection::StartTrafficLogging(string_view path) {
@@ -2722,9 +2736,24 @@ void Connection::CheckIoBufCapacity(bool is_iobuf_full) {
   }
 }
 
+void Connection::MaybeEnableRecvMultishot() {
+#ifdef __linux__
+  if (fb2::ProactorBase::me()->GetKind() == fb2::ProactorBase::Kind::IOURING) {
+    auto* up = static_cast<fb2::UringProactor*>(fb2::ProactorBase::me());
+    // Only enable if the buffer ring is configured and we aren't using TLS
+    if (up->BufRingEntrySize(kRecvSockGid) > 0 && !is_tls_) {
+      static_cast<fb2::UringSocket*>(socket_.get())->EnableRecvMultishot();
+      pending_input_ = false;
+    }
+  }
+#endif
+}
+
 variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
   auto* peer = socket_.get();
   recv_buf_.res_len = 0;
+
+  auto is_ready_to_migrate = [this]() { return migration_request_ && (cc_->subscriptions == 0); };
 
   // Don't proceed with RegisterOnRecv() if socket is closed (possible cancellation)
   if (!peer->IsOpen())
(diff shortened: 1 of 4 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -2900,11 +2931,18 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
       return std::exchange(io_ec_, {});
     }
 
+    if ((parse_status != OK) && (parse_status != NEED_MORE)) {
+      break;
+    }
+
+    // Migration requested: skip buffer bookkeeping, jump to HandleMigrateRequest().
+    if (migration_request_) {
```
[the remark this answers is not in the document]

src/server/server_family.cc
```diff
@@ -688,7 +688,7 @@ void ClientMigrate(CmdArgList args, absl::Span<facade::Listener*> listeners,
     }
   };
 
-  shard_set->pool()->AwaitBrief(cb_brief);
+  shard_set->pool()->AwaitFiberOnAll(search_and_migrate_cb);
```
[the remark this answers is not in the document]
glevkovich: fixed

tests/dragonfly/connection_test.py
```diff
@@ -1657,6 +1667,168 @@ async def test_client_migrate_no_conn_leak(df_server: DflyInstance):
...
+    migrant_info = next(c for c in clients if c["name"] == "migrant")
+    migrant_id = int(migrant_info["id"])
+
+    # Flood without reading to saturate TCP send buffer -> queue exceeds limit.
+    writer.write(b"GET x\r\n" * 1000)
+    await writer.drain()
+
+    # Concurrent flood clients for global memory pressure.
+    async def pipe_flood():
+        c = df_server.client()
```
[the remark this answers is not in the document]
glevkovich: fixed

tests/pytest.ini
```diff
@@ -5,6 +5,7 @@ log_date_format = %Y-%m-%d %H:%M:%S
 log_file_level=INFO
 log_cli = true
 asyncio_mode=auto
+asyncio_default_fixture_loop_scope=function
```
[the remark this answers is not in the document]
glevkovich: fixed. bumped pytest-asyncio>=0.24.0,<0.25.0 and pytest>=8.3,<9.0

tests/dragonfly/connection_test.py
```diff
@@ -1657,6 +1667,168 @@ async def test_client_migrate_no_conn_leak(df_server: DflyInstance):
...
+        "experimental_io_loop_v2": "false",
+    },
+    {
+        "proactor_threads": 4,
+        "pipeline_buffer_limit": "1024",
+        "pipeline_queue_limit": 10,
+        "experimental_io_loop_v2": "true",
+    },
+)
+async def test_pipeline_global_memory_relief(df_server: DflyInstance):
```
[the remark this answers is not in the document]
glevkovich: fixed

vyavdoshenko:
@glevkovich 
If this PR contains 3 irrelevant to each other changes, why don't you want to split it into 3 different PRs?
Just imagine if only one change is buggy, and it will be convenient to revert only this change. Also, it would be more convenient to review it separately because I, as a reviewer, would be sure that this particular diff is relevant to this task only.

glevkovich:
> @glevkovich If this PR contains 3 irrelevant to each other changes, why don't you want to split it into 3 different PRs? Just imagine if only one change is buggy, and it will be convenient to revert only this change. Also, it would be more convenient to review it separately because I, as a reviewer, would be sure that this particular diff is relevant to this task only.
 
@vyavdoshenko 
That’s a fair point. I didn't plan on doing three separate tasks, but I found these issues while working on the migration feature.
I will split this into two PRs:
- One for the Python requirements and pytest changes.
- One for the connection migration feature and the C++ bug fix.
- 
- I’m keeping the bug fix and the migration feature together because I found the bug using the new migration tests. It doesn’t make sense to write separate tests for the old logic just to replace them in the next PR.

Bottom line: I’ll move the Python/pytest stuff to its own PR now.

src/facade/dragonfly_connection.cc
```diff
@@ -1390,38 +1391,46 @@ void Connection::OnBreakCb(int32_t mask) {
...
+    migration_request_ = nullptr;
+    return;
+  }
+
   ProactorBase* dest = migration_request_;
-  if (async_fb_.IsJoinable()) {
+
+  // V1 loop: we must coordinate with and shutdown the AsyncFiber before migration.
+  // V2 loop: is single-fiber and handles this by breaking the dispatch_q loop.
+  if (!ioloop_v2_ && async_fb_.IsJoinable()) {
```
romange, line 1406: why checking for `!ioloop_v2_` here ? is not the second condition enough?

src/facade/dragonfly_connection.cc
```diff
@@ -1390,38 +1391,46 @@ void Connection::OnBreakCb(int32_t mask) {
...
     tl_facade_stats->conn_stats.num_migrations++;
     migration_request_ = nullptr;
 
-    // We need to return early as the socket is closing and IoLoop will clean up.
-    // The reason that this is true is because of the following DCHECK
-    DCHECK(!async_fb_.IsJoinable());
+    // Verify that no background command processing is active.
+    // V1 loop: Join() ensures this.
+    // V2 loop: it is guaranteed by the single-fiber loop.
+    DCHECK(ioloop_v2_ || !async_fb_.IsJoinable());
```
romange, line 1426: same q

src/facade/dragonfly_connection.cc
glevkovich: I agree that functionality-wise, the 2nd condition is enough. It will work since for v2, async_fb_ is never started and IsJoinable() will return false. But in terms of code clarity and design, v2 logic should not touch async_fb_. I think "Separation of Concerns" is important here, so I prefer keeping it that way. Also, if we decide for some reason to use async_fb_ in the future in v2 for another use case, it will be much harder to change the code here and understand the logic.

src/facade/dragonfly_connection.cc
glevkovich: answered [here](https://github.com/dragonflydb/dragonfly/pull/7143#discussion_r3087447315)

src/facade/dragonfly_connection.cc
```diff
@@ -2900,11 +2939,19 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
       return std::exchange(io_ec_, {});
     }
 
+    if ((parse_status != OK) && (parse_status != NEED_MORE)) {
+      break;
+    }
+
+    // Migration requested and actionable: skip buffer bookkeeping, jump to HandleMigrateRequest().
+    bool ready_to_migrate = migration_request_ && (cc_->subscriptions == 0);
```
romange: this has been duplicated 3 times. pls introduce an private function instead

src/server/server_family.cc
romange, line 691: can you explain where do we preempt?

romange (commented):
lgtm

src/facade/dragonfly_connection.cc
romange: ok, though when I read such expression I immediately think "wait - does it mean `ioloop_v2_ && async_fb_.IsJoinable()` is a thing? 

To pass the message it's an invalid state, I would add a `DCHECK(!ioloop_v2_ || !async_fb_.IsJoinable())` before this line.

src/facade/dragonfly_connection.cc
glevkovich: sure, I am going to use a lambda, all ocuurences are in Connection::IoLoopV2().

src/server/server_family.cc
glevkovich: The preemption point is inside `ListenerInterface::TraverseConnectionsOnThread`:
```
uint64_t gate = migrate_traversal_state_.fetch_add(1, memory_order_acquire);
while ((gate >> 32)) {
    ThisFiber::SleepFor(1us);  // <-- here
    gate = migrate_traversal_state_.load(memory_order_acquire);
}
```
inside the spin-wait while loop, SleepsFor yields the fiber. This happens when migrations are active on the same thread.
Found it via the new test test_client_migrate_no_conn_leak.
This  test migrates 20 clients at once, so it failed pretty fast with "Should not preempt dispatcher".

romange: maybe add an one-liner comment mentioning that `TraverseConnectionsOnThread` preempts

glevkovich: Sure, I'll add a comment at listener_interface.h where the function documentation. 
I'll have to upload a small PR in Helio for that.

```
  // traverses all client connection in current thread. cb must adhere to rules from
  // `TraverseConnections`. Specifically, cb should not fiber block so that the underlying
  // connection list won't change during the traversal.
  // Note: This function may preempt internally (e.g., waiting for active migrations to finish).
  Connection* TraverseConnectionsOnThread(TraverseCB cb, uint32_t limit, Connection* from);
```

romange (approved):

PR #7148, opened 2026-04-15
chore: split debug info and improve build time on 10%

BorysTheDev:
Summary: Enables split DWARF (-gsplit-dwarf) for Debug builds to reduce link-time debug info load and improve build time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,7 @@ cmake_minimum_required(VERSION 3.15 FATAL_ERROR)
 set(PROJECT_CONTACT romange@gmail.com)
 
 include(CheckCXXCompilerFlag)
+include(CheckCCompilerFlag)
 
 enable_testing()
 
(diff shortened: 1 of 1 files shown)


BorysTheDev:
augment review

romange (commented):
I do not want this to be resolved unconditionally. it affects our release pipeline and maybe other pieces that rely on debug information.

BorysTheDev:
> I do not want this to be resolved unconditionally. it affects our release pipeline and maybe other pieces that rely on debug information.

@romange, what if I do it only for the debug build, or can I add a flag, or both?

By the way I don't remove debug info, it just stay in the separate file.

romange:
Well, in our CI for example, if df server crashes - we get a nice stacktrace. How will it work then? I do not know.

BorysTheDev:
@romange I've created a separate fake PR to demonstrate stack trace on CI. I've added crash into the HSET command.
https://github.com/dragonflydb/dragonfly/pull/7151

romange:
ok, lets try this for debug builds. Still not sure how it works though. does glog library uses the separate file to symbolize the stack trace?

BorysTheDev:
> ok, lets try this for debug builds. Still not sure how it works though. does glog library uses the separate file to symbolize the stack trace?

This PR is already for debug only, there is no sense in such changes for release. compiler adds references to .DWO files into the executable file, so if we run it from the build directory everything should be OK, if we want to transfer executable file with debug info, we need to .DWP file that is archive from all .DWO files. As I understand we don't use debug build separately from build directory so no more changes needed,

romange (approved):

PR #7165, opened 2026-04-16
fix: clean up empty containers after DEBUG OBJHIST/UNIQ-STRS

vyavdoshenko:
DEBUG OBJHIST and DEBUG UNIQ-STRS traverse the entire DB and iterate hashes/sets via IterateMap/IterateSet. If time_now_ was set by a prior command, the iteration triggers lazy expiry. If all entries expired, the empty container must be deleted.

### DIFF
--- a/src/server/debugcmd.cc
+++ b/src/server/debugcmd.cc
@@ -561,7 +561,9 @@ IOStat& IOStat::operator-=(const IOStat& other) {
   return *this;
 }
 
-// Traverse over all entries on all databases, manage cpu time automatically
+// Traverse over all entries on all databases, manage cpu time automatically.
+// The callback receives (DbIndex, PrimeIterator) — the DbIndex identifies
+// which database is currently being iterated.
 template <typename F> void TraverseAllEntries(bool background, ConnectionContext* cntx, F&& f) {
   util::fb2::BlockingCounter bc{0};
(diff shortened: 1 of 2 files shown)


src/server/debugcmd.cc
```diff
@@ -1177,13 +1183,26 @@ void DebugCmd::TxAnalysis(CommandContext* cmd_cntx) {
...
+    // in the currently traversed DB (not the connection-selected one).
+    if (it->second.Size() == 0 && it->second.Encoding() == kEncodingStrMap2) {
+      auto& db_slice = cntx->ns->GetDbSlice(EngineShard::tlocal()->shard_id());
+      DbContext db_cntx{cntx->ns, dbid, GetCurrentTimeMs()};
+      string key;
+      it->first.GetString(&key);
+      if (obj_type == OBJ_SET)
+        SetFamily::DeleteSetIfEmpty(db_slice, db_cntx, key, it->second);
+      else if (obj_type == OBJ_HASH)
+        HSetFamily::DeleteIfEmpty(db_slice, db_cntx, key, it->second);
```
dranikpg, line 1204: we really need some kind of better solution that avoid duplicating it everywhere, but doesn't matter as its debugcmd code

dranikpg (approved):

PR #7166, opened 2026-04-16
fix: delete empty hash after FT.SEARCH lazy field expiry

vyavdoshenko:
FT.SEARCH loads documents via GetAccessor -> GetStringMap (which sets time_now_) -> Serialize (which iterates the StringMap). The iteration triggers lazy field expiry. If all fields expired, the hash must be deleted to prevent zombie keys that crash SAVE.

### DIFF
--- a/src/server/search/doc_index.cc
+++ b/src/server/search/doc_index.cc
@@ -20,6 +20,7 @@
 #include "server/db_slice.h"
 #include "server/engine_shard_set.h"
 #include "server/family_utils.h"
+#include "server/hset_family.h"
 #include "server/search/doc_accessors.h"
 #include "server/search/global_hnsw_index.h"
 #include "server/search/index_builder.h"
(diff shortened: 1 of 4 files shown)


src/server/search/doc_index.cc
```diff
@@ -945,6 +948,18 @@ SearchResult ShardDocIndex::Search(const OpArgs& op_args, const SearchParams& pa
...
+    // If all fields expired, delete the now-empty hash.
+    if (base_->GetObjCode() == OBJ_HASH) {
+      auto chk = op_args.GetDbSlice().FindReadOnly(op_args.db_cntx, key, OBJ_HASH);
+      if (chk && IsValid(*chk) && (*chk)->second.Size() == 0) {
+        // Copy key: DelMutable inside DeleteIfEmpty may trigger a doc-deletion
+        // callback that mutates key_index_ and invalidates our string_view.
+        string owned_key{key};
+        HSetFamily::DeleteIfEmpty(op_args.GetDbSlice(), op_args.db_cntx, owned_key, (*chk)->second);
+      }
+    }
```
dranikpg: Please let's not have datastructure specific code in search code. It can be hidden inthe accessor, maybe even the destructor if that will work

vyavdoshenko: moved to doc_accessor

dranikpg (approved):
Much cleaner

PR #7167, opened 2026-04-16
fix: disable lazy expiry during migration serialization

vyavdoshenko:
CmdSerializer::SerializeHash/SerializeSet use IterateMap/IterateSet which can trigger lazy field expiry if time_now_ was set by a prior command. Since serialization runs under bucket lock, DeleteIfEmpty is not possible. Instead, disable lazy expiry with set_time(0) before iteration (same pattern as rdb_save.cc).

### DIFF
--- a/src/server/journal/cmd_serializer.cc
+++ b/src/server/journal/cmd_serializer.cc
@@ -6,11 +6,9 @@
 
 #include "core/string_map.h"
 #include "core/string_set.h"
-#include "server/common.h"
 #include "server/container_utils.h"
 #include "server/db_slice.h"
 #include "server/engine_shard.h"
-#include "server/engine_shard_set.h"
 #include "server/journal/serializer.h"
 #include "server/rdb_save.h"
 #include "server/tiered_storage.h"
@@ -165,8 +163,10 @@ size_t CmdSerializer::SerializeSet(string_view key, const PrimeValue& pv) {
   // Disable lazy expiry during serialization (same as rdb_save.cc).
   // We are called under bucket lock so DeleteIfEmpty is not possible.
   StringSet* ss = nullptr;
+  uint32_t prev_time = 0;
   if (pv.Encoding() == kEncodingStrMap2) {
     ss = static_cast<StringSet*>(pv.RObjPtr());
+    prev_time = ss->time_now();
     ss->set_time(0);
   }
 
@@ -180,9 +180,9 @@ size_t CmdSerializer::SerializeSet(string_view key, const PrimeValue& pv) {
     return true;
   });
 
-  // Restore time so subsequent operations can trigger lazy expiry again.
+  // Restore previous time so subsequent operations can trigger lazy expiry.
   if (ss)
-    ss->set_time(MemberTimeSeconds(GetCurrentTimeMs()));
+    ss->set_time(prev_time);
 
   return commands;
 }
@@ -207,8 +207,10 @@ size_t CmdSerializer::SerializeZSet(string_view key, const PrimeValue& pv) {
 size_t CmdSerializer::SerializeHash(string_view key, const PrimeValue& pv) {
   // Disable lazy expiry during serialization (same as rdb_save.cc).
   StringMap* sm = nullptr;
+  uint32_t prev_time = 0;
   if (pv.Encoding() == kEncodingStrMap2) {
     sm = static_cast<StringMap*>(pv.RObjPtr());
+    prev_time = sm->time_now();
     sm->set_time(0);
   }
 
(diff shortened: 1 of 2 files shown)


romange:
good find. but:

1. Please expand on https://github.com/dragonflydb/dragonfly/issues/7133 and explain there the failure/problem.
2. Open another issue with the task to redesign the current behavior (i.e. describe this scenario again, state the problem, state the workaround - link this PR)
3. Add that issue to https://github.com/orgs/dragonflydb/projects/2/views/6. One of the improvements we thought to implement was to allow journal deletions for locked buckets by gathering delayed deletions. This will greatly reduce the complexity around these interactions and will also obsolete this PR. 

besides this LGTM

vyavdoshenko:
> good find. but:
> 
> 1. Please expand on [SIGABRT in rdb_save.cc:274 #7133](https://github.com/dragonflydb/dragonfly/issues/7133) and explain there the failure/problem.
> 2. Open another issue with the task to redesign the current behavior (i.e. describe this scenario again, state the problem, state the workaround - link this PR)
> 3. Add that issue to [DF-DF Replication (view)](https://github.com/orgs/dragonflydb/projects/2/views/6). One of the improvements we thought to implement was to allow journal deletions for locked buckets by gathering delayed deletions. This will greatly reduce the complexity around these interactions and will also obsolete this PR.
> 
> besides this LGTM

Done:

1. Expanded the scenario: https://github.com/dragonflydb/dragonfly/issues/7133#issuecomment-4267464266
2. Opened follow-up tracking issue for the redesign: https://github.com/dragonflydb/dragonfly/issues/7178
3. Added to https://github.com/orgs/dragonflydb/projects/2/views/6

src/server/journal/cmd_serializer.cc
```diff
@@ -199,6 +224,11 @@ size_t CmdSerializer::SerializeHash(string_view key, const PrimeValue& pv) {
         commands += aggregator.AddArg(v.ToString());
         return true;
       });
+
+  // Restore time so subsequent operations can trigger lazy expiry again.
+  if (sm)
+    sm->set_time(MemberTimeSeconds(GetCurrentTimeMs()));
```
BorysTheDev: it looks like a pattern, let's create a wrapper TimeFreezer or something like this, that will set time to zero and return previous time in destructor

src/server/journal/cmd_serializer.cc
```diff
@@ -84,7 +88,8 @@ size_t CmdSerializer::SerializeEntry(string_view key, const PrimeKey& pk, const
   // We send RESTORE commands objects we don't support breaking.
   bool use_restore_serialization = true;
   size_t commands = 1;
-  switch (pv.ObjType()) {
+  auto obj_type = pv.ObjType();
+  switch (obj_type) {
```
BorysTheDev: why do you need this change

src/server/journal/cmd_serializer.cc
```diff
@@ -167,6 +180,11 @@ size_t CmdSerializer::SerializeSet(string_view key, const PrimeValue& pv) {
     commands += aggregator.AddArg(ce.ToString());
     return true;
   });
+
+  // Restore time so subsequent operations can trigger lazy expiry again.
+  if (ss)
+    ss->set_time(MemberTimeSeconds(GetCurrentTimeMs()));
```
BorysTheDev: we don't need the current time, just previous value

src/server/journal/cmd_serializer.cc
vyavdoshenko: It seems I found all current issues like this. I think all this code should be eliminated when new serialization is implemented and this particular code refactored: https://github.com/dragonflydb/dragonfly/issues/7178

src/server/journal/cmd_serializer.cc
```diff
@@ -4,9 +4,13 @@
 
 #include "server/journal/cmd_serializer.h"
 
+#include "core/string_map.h"
+#include "core/string_set.h"
+#include "server/common.h"
 #include "server/container_utils.h"
 #include "server/db_slice.h"
 #include "server/engine_shard.h"
+#include "server/engine_shard_set.h"
```
BorysTheDev: do you really need it?

src/server/journal/cmd_serializer.cc
vyavdoshenko: fixed

src/server/journal/cmd_serializer.cc
vyavdoshenko: Done intentionally to mirror the existing pattern in rdb_save.cc:

    string_map->set_time(MemberTimeSeconds(GetCurrentTimeMs()));

That file resets to the current time rather than saving/restoring the previous value, and I kept this consistent across both serializers. Happy to switch to previous-value if you prefer - the only practical difference is that seconds-level time may have advanced during a long iteration.

Either way, this whole pattern goes away with #7178.

src/server/journal/cmd_serializer.cc
vyavdoshenko: Yes - needed for the `static_cast<StringSet*>(pv.RObjPtr())->set_time(0)` / `static_cast<StringMap*>(...)` calls further down. Without the full class definition clang errors with "member access into incomplete type". The headers are only forward-declared via container_utils.h.

BorysTheDev (dismissed):

BorysTheDev (approved):

dranikpg (approved):

PR #7169, opened 2026-04-16
feat(bloom): Implement LOADCHUNK cmd for SBF

abhijat:
implements LOADCHUNK command for SBF
creates a new SBF, then progressively loads data into its filters.

The command expects `key,cursor,byte data`

On cursor=1, the byte data is expected to be SBF metadata: version and growth factor. It is used to create a new empty (valid) SBF.

Subsequent cursors will add new filters and update the SBF state such as fp prob, max capacity etc.

While the SBF is in a partially loaded state (determined by `num_filters == 0`) most commands on it are rejected. The commands are unblocked after the first filter is loaded.

# Concurrent write risks

**The filter should not be modified with `BF.ADD` until the `LOADCHUNK` commands are finished.** 

When the SBF is created initially it has no filters. There are guards put into place in this PR to avoid insert or query at this stage. But once some chunks have been loaded and filters are non-zero those checks are relaxed.

If `BF.ADD` is run in parallel, the SBF state may be overwritten (if the `BF.ADD` triggers filter growth), and subsequent `LOADCHUNK` commands on that SBF will fail due to validation checks. Example scenario:

* SBF is created with `LOADCHUNK ... 1` with 0 filters
* `BF.ADD` executes, is blocked due to 0 filters validation
* next `LOADCHUNK` adds one filter
* many `BF.ADD` execute and write to filter, resulting in SBF growth.
* next `LOADCHUNK` runs. Either it will overwrite the new filter, or fail validation if size does not match
* The SBF now contains inconsistent data in filters, and `LOADCHUNK` commands may fail at validation

It is not possible to lock the SBF until the last chunk has been loaded, because there is no way to determine which chunk is the last on the server, only the client has that information in the current wire protocol.

### DIFF
--- a/src/core/bloom.cc
+++ b/src/core/bloom.cc
@@ -196,11 +196,11 @@ SBF& SBF::operator=(SBF&& src) noexcept {
   return *this;
 }
 
-void SBF::AddFilter(const std::string& blob, unsigned hash_cnt) {
+uint8_t* SBF::AllocateFilter(size_t alloc_size, unsigned hash_cnt) {
   PMR_NS::memory_resource* mr = filters_.get_allocator().resource();
-  uint8_t* ptr = (uint8_t*)mr->allocate(blob.size());
-  memcpy(ptr, blob.data(), blob.size());
-  filters_.emplace_back().Init(ptr, blob.size(), hash_cnt);
+  const auto ptr = static_cast<uint8_t*>(mr->allocate(alloc_size));
+  filters_.emplace_back().Init(ptr, alloc_size, hash_cnt);
+  return ptr;
 }
 
 bool SBF::Add(std::string_view str) {
@@ -387,51 +387,80 @@ void SBFDumpIterator::ResolveCursorToPos() {
 
 nonstd::expected<SBF*, SBFLoadResult> LoadSBFHeader(std::string_view header_data,
                                                     PMR_NS::memory_resource* mr) {
+  using enum SBFLoadResult;
+  using nonstd::make_unexpected;
+
   if (header_data.size() < kDumpHeaderSize)
-    return nonstd::make_unexpected(SBFLoadResult::kTruncatedInput);
+    return make_unexpected(kTruncatedInput);
+
+  if (header_data.size() > kDumpHeaderSize)
+    return make_unexpected(kBadInput);
 
   const char* ptr = header_data.data();
   if (const uint32_t version = absl::little_endian::Load32(ptr); version != kSbfDumpVersion)
-    return nonstd::make_unexpected(SBFLoadResult::kBadVersion);
+    return make_unexpected(kBadVersion);
 
   const double grow_factor = std::bit_cast<double>(absl::little_endian::Load64(ptr + 4));
+  if (!std::isfinite(grow_factor) || grow_factor < 1.0)
+    return make_unexpected(kBadInput);
+
   // Initialize everything to 0, later filters will overwrite these values
   return CompactObj::AllocateMR<SBF>(grow_factor, 0.0, 0UL, 0UL, 0UL, mr);
 }
 
 SBFLoadResult AddNewFilterToSBF(std::string_view data, SBF* sbf) {
+  using enum SBFLoadResult;
+
   if (data.size() < kDumpFilterMetaSize)
-    return SBFLoadResult::kTruncatedInput;
+    return kTruncatedInput;
 
   auto [hash_cnt, data_length, state] = SBFFilterMeta::Parse(data.data());
+  if (hash_cnt == 0)
+    return kBadInput;
+
+  if (hash_cnt > std::numeric_limits<uint8_t>::max())
+    return kBadInput;
+
+  if (data_length == 0 || !absl::has_single_bit(data_length))
+    return kBadInput;
+
+  // probability should be 0 to 1 (probably less than 1)
+  if (!std::isfinite(state.fp_prob) || state.fp_prob <= 0.0 || state.fp_prob >= 1.0)
+    return kBadInput;
+
+  if (state.max_capacity == 0 || state.current_size >= state.max_capacity)
+    return kBadInput;
+
   const size_t payload = data.size() - kDumpFilterMetaSize;
   if (payload > data_length)
-    return SBFLoadResult::kOutOfRange;
+    return kOutOfRange;
 
   sbf->ApplyStateUpdate(state);
 
   const uint32_t new_index = sbf->num_filters();
   // TODO validate variables against bloom invariants (power of two etc)
-  sbf->AddFilter(std::string(data_length, '\0'), hash_cnt);
+  auto* ptr = sbf->AllocateFilter(data_length, hash_cnt);
+  memset(ptr, 0, data_length);
 
   if (payload > 0)
     memcpy(sbf->filter_data(new_index), data.data() + kDumpFilterMetaSize, payload);
 
-  return SBFLoadResult::kOk;
+  return kOk;
 }
 
 SBFLoadResult LoadSBFChunk(int64_t cursor, std::string_view data, SBF* sbf) {
   DCHECK_NE(sbf, nullptr) << "Input ptr must be valid SBF";
 
-  // TODO on implementing LOADCHUNK there should be closer validation of the data fed into the SBF.
-  // This current implementation is mostly a test helper and proof that the SCANDUMP algorithm is
-  // actually loadable.
+  const int64_t write_pos = cursor - static_cast<int64_t>(data.size());
+  if (write_pos < 1)
+    return SBFLoadResult::kOutOfRange;
 
-  size_t global_offset = cursor - static_cast<int64_t>(data.size()) - 1;
-
+  size_t global_offset = write_pos - 1;
   for (uint32_t i = 0; i < sbf->num_filters(); ++i) {
     const size_t filter_span = kDumpFilterMetaSize + sbf->data(i).size();
     if (global_offset < filter_span) {
+      // we should never have a write position inside the header. The header is always fully
+      // written.
       if (global_offset < kDumpFilterMetaSize)
         return SBFLoadResult::kOutOfRange;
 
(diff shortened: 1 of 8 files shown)


abhijat:
~~fuzzer error https://github.com/dragonflydb/dragonfly/actions/runs/24554350655/job/71787122701?pr=7169~~

build failure

src/core/bloom.cc
```diff
@@ -387,51 +387,80 @@ void SBFDumpIterator::ResolveCursorToPos() {
...
+
+  if (data_length == 0 || !has_single_bit(data_length))
+    return kBadInput;
+
+  // probability should be 0 to 1 (probably less than 1)
+  if (!std::isfinite(state.fp_prob) || state.fp_prob <= 0.0 || state.fp_prob >= 1.0)
+    return kBadInput;
+
+  if (state.max_capacity == 0 || state.current_size >= state.max_capacity)
+    return kBadInput;
```
abhijat, line 432: these validations might seem like overkill but since we're loading user input directly I think its fair to be a little over protective.

src/server/bloom_family.cc
```diff
@@ -202,6 +202,65 @@ void CmdScanDump(CmdArgList args, CommandContext* cmd_cntx) {
...
+                                << " blob_size=" << blob.size()
+                                << " load_res=" << ToString(load_result.error());
+        return OpStatus::INVALID_VALUE;
+      }
+
+      auto op_res = db_slice.AddOrFind(op_args.db_cntx, key, OBJ_SBF);
+      if (!op_res)
+        return op_res.status();
+
+      op_res->it->second.SetSBF(load_result.value());
```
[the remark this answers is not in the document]
abhijat: I added a check to the commands where commands to an SBF with 0 filters is rejected. Unfortunately a cleaner fix is impossible because:

* we do not know if the filter is fully hydrated until the last chunk is supplied by the client
* there is no special flag the client passes for the last chunk. it just stops sending chunks
* we cannot change the protocol ie client behavior
* we want to support growing SBF while dumping, so we cannot assume num_filters up front.

The best fix is to reject commands for SBF with 0 filters, it guards explicitly against modifying such filters where load is in progress. To be a little extra cautious I also added this check for exists style commands, although they work fine even with 0 filters.

The issue that remains is mostly a consistency issue, for example the following sequence:

* one filter is loaded, more to be loaded by client loadchunk
* some BF.ADD comes in and modifies the filter

Now that BF.ADD change might be overwritten/lost. But there is no way to lock the filter while it is loading other than something like locking or transactions. So the client has to guarantee to not write until loadchunk finishes.

One more reason for clients to avoid `BF.ADD` running parallel to LOADCHUNK is, if `BF.ADD` triggers growth, which means the remaining chunks after the growth will be rejected by validation (because the new filter layout does not match the expected chunk layout), and the SBF ends up in a mixed state - some filters came from the dump and some from `BF.ADD`.

src/core/bloom.cc
```diff
@@ -387,51 +387,80 @@ void SBFDumpIterator::ResolveCursorToPos() {
...
+    return kTruncatedInput;
 
   auto [hash_cnt, data_length, state] = SBFFilterMeta::Parse(data.data());
+  if (hash_cnt == 0)
+    return kBadInput;
+
+  if (hash_cnt > std::numeric_limits<uint8_t>::max())
+    return kBadInput;
+
+  if (data_length == 0 || !has_single_bit(data_length))
```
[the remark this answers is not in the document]
abhijat: I removed it due to clangd warning, but I have added it back

src/server/bloom_family_test.cc
```diff
@@ -85,6 +85,51 @@ TEST_F(BloomFamilyTest, ScanDump) {
...
+  for (const auto& [crs, data] : chunks)
+    EXPECT_EQ(Run({"bf.loadchunk", "b2", std::to_string(crs), data}), "OK");
+
+  // Verify all items exist in the loaded copy
+  for (int i = 0; i < total_items; ++i)
+    EXPECT_THAT(Run({"bf.exists", "b2", absl::StrCat("item", i)}), IntArg(1));
+
+  // Verify some non-inserted items are absent
+  for (int i = total_items; i < total_items + 50; ++i)
+    EXPECT_THAT(Run({"bf.exists", "b2", absl::StrCat("item", i)}), IntArg(0));
```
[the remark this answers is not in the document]
abhijat: this check was overkill, I removed it

src/core/bloom.h
```diff
@@ -109,7 +111,7 @@ class SBF {
 
   SBF& operator=(SBF&& src) noexcept;
 
-  void AddFilter(const std::string& blob, unsigned hash_cnt);
+  uint8_t* AllocateFilter(size_t alloc_size, unsigned hash_cnt);
```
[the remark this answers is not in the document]
abhijat: updated

abhijat:
augment review

src/server/bloom_family.cc
```diff
@@ -36,6 +36,10 @@ struct SbfParams {
 using AddResult = absl::InlinedVector<OpResult<bool>, 4>;
 using ExistsResult = absl::InlinedVector<bool, 4>;
 
+bool IsBeingLoaded(const SBF* sbf) {
+  return sbf->num_filters() == 0;
+}
```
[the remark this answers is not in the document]
abhijat: There is no finalize step in the protocol. this is a known limitation of it. The check is against running commands on an invalid state that can cause crash, not a guarantee to protect data - which is not possible in the current protocol.

src/server/bloom_family.cc
```diff
@@ -202,6 +226,66 @@ void CmdScanDump(CmdArgList args, CommandContext* cmd_cntx) {
...
+        LOG_EVERY_T(WARNING, 10) << "BF.LOADCHUNK invalid header"
+                                 << " key=" << key << " cursor=" << cursor
+                                 << " blob_size=" << blob.size()
+                                 << " load_res=" << ToString(load_result.error());
+        return OpStatus::INVALID_VALUE;
+      }
+
+      auto op_res = db_slice.AddOrFind(op_args.db_cntx, key, OBJ_SBF);
+      if (!op_res)
+        return op_res.status();
```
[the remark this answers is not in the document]
abhijat: this is a good point, loadchunk is expected to overwrite existing key. I added code to remove from indices if the object is there and remove ttl. 

But there is no need to free anything, SetSBF does that via SetMeta

src/server/bloom_family.cc
```diff
@@ -202,6 +226,66 @@ void CmdScanDump(CmdArgList args, CommandContext* cmd_cntx) {
...
+      if (!load_result.has_value()) {
+        LOG_EVERY_T(WARNING, 10) << "BF.LOADCHUNK invalid header"
+                                 << " key=" << key << " cursor=" << cursor
+                                 << " blob_size=" << blob.size()
+                                 << " load_res=" << ToString(load_result.error());
+        return OpStatus::INVALID_VALUE;
+      }
+
+      auto op_res = db_slice.AddOrFind(op_args.db_cntx, key, OBJ_SBF);
+      if (!op_res)
```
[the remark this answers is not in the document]
abhijat: fixed

src/server/bloom_family.cc
```diff
@@ -202,6 +227,76 @@ void CmdScanDump(CmdArgList args, CommandContext* cmd_cntx) {
...
+  const int64_t cursor = parser.Next<int64_t>();
+  if (const auto err = parser.TakeError(); err)
+    return rb->SendError(err.MakeReply());
+
+  if (cursor <= 0)
+    return rb->SendError(kInvalidIntErr);
+
+  const std::string_view blob = parser.Next();
+
+  const auto cb = [&](Transaction* t, EngineShard* shard) -> OpStatus {
```
romange: I prefer that this function would be factored out into a standalone function `OpLoadChunk` like we do with other commands.

src/server/bloom_family.cc
```diff
@@ -202,6 +227,76 @@ void CmdScanDump(CmdArgList args, CommandContext* cmd_cntx) {
...
+    auto op_args = t->GetOpArgs(shard);
+    auto& db_slice = op_args.GetDbSlice();
+
+    if (cursor == 1) {
```
romange:
```suggestion
    if (cursor == 1) {   // Init phase
```

src/server/bloom_family.cc
```diff
@@ -202,6 +227,76 @@ void CmdScanDump(CmdArgList args, CommandContext* cmd_cntx) {
...
+      // LOADCHUNK overwrites existing key
+      if (!op_res->is_new) {
+        // existing key might not necessarily be SBF, it could be HASH/JSON, and indexed
+        RemoveKeyFromIndexesIfNeeded(key, op_args.db_cntx, op_res->it->second, op_args.shard);
+        db_slice.RemoveExpire(op_args.db_cntx.db_index, op_res->it);
+      }
+
+      op_res->it->second.SetSBF(load_result.value());
+      return OpStatus::OK;
+    }
```
romange:
```suggestion
    }  //  cursor == 1
```

src/server/bloom_family.cc
```diff
@@ -202,6 +227,76 @@ void CmdScanDump(CmdArgList args, CommandContext* cmd_cntx) {
...
+
+    auto op_res = db_slice.FindMutable(op_args.db_cntx, key, OBJ_SBF);
```
romange:
```suggestion

    // Continue loading chunks into not-yet-fully-loaded filter.
    auto op_res = db_slice.FindMutable(op_args.db_cntx, key, OBJ_SBF);
```

romange (commented):
lgtm

src/server/bloom_family.cc
abhijat: fixed

src/server/bloom_family.cc
abhijat: fixed

src/server/bloom_family.cc
abhijat: fixed

src/server/bloom_family.cc
abhijat: extracted function

src/server/bloom_family.cc
```diff
@@ -186,12 +254,17 @@ void CmdScanDump(CmdArgList args, CommandContext* cmd_cntx) {
...
+      return OpStatus::INVALID_VALUE;
+
     SBFDumpIterator it(*sbf, cursor);
     return it.Next();
   };
 
   OpResult<SBFChunk> res = cmd_cntx->tx()->ScheduleSingleHopT(std::move(cb));
   if (!res) {
+    if (res.status() == OpStatus::INVALID_VALUE)
+      return rb->SendError("ERR bloom filter load is incomplete");
```
romange: I would let it slip but since it's duplicated many times, I think it justifies a proper addition of a dedicated enum and adding the message into `StatusToMsg`

abhijat: yeah its a little messy, I will fix it now

abhijat: fixed with new enum and msg

romange (approved):

PR #7170, opened 2026-04-16
refactor: move ExecutionState out of ProtocolClient

kostasrim:
`ProtocolClient::exec_st_` is a shared field even though it's only meaningful to specific subclasses. For example, exec_st_ of each replica flow is ignored (and the parent/main replica connection exec_st_ is passed around instead). This is confusing on the API level were *some* users of the ProtocolClient require the exec_st_ (main repl connection) while others don't at all (replica flows).

This PR moves it down to the classes that actually use it -- Replica, OutgoingMigration, SliceSlotMigration, and CrossShardClient. 

Also updates DflyShardReplica::StartSyncFlow to pass the correct context pointer instead of the now-removed base field.

### DIFF
--- a/src/server/protocol_client.h
+++ b/src/server/protocol_client.h
@@ -12,7 +12,6 @@
 #include "facade/redis_parser.h"
 #include "facade/resp_parser.h"
 #include "io/io_buf.h"
-#include "server/execution_state.h"
 #include "server/version.h"
 #include "util/fiber_socket_base.h"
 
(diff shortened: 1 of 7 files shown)


src/server/protocol_client.h
BorysTheDev, line 15: drop include

BorysTheDev (approved):

dranikpg (approved):

src/server/protocol_client.h
kostasrim: uh, both ExecutionState and GenericError are part of the interface (we dropped the member but still the API uses it). Yes we can forward declare but then we include it back in the source so I ended up bringing it back again

abhijat (approved):

BorysTheDev (approved):

PR #7174, opened 2026-04-16
feat(pubsub): Replace RCU-based ChannelStore with ShardedHashMap

### LINKED ISSUE #7056: Pub/sub PUBLISH notifications silently lost under concurrent Celery workloads
### Description

When using Dragonfly as both the broker and result backend for Celery, `PUBLISH` notifications are silently lost under concurrent load. Celery's Redis result backend relies on pub/sub to notify callers when a task result is ready (`SET` + `PUBLISH`). Under sustained concurrent usage, some `PUBLISH` messages are never delivered to the subscriber, even though the corresponding `SET` (result storage) succeeds.

This causes `GroupResult.get(timeout=N)` to block until the timeout expires, despite all tasks having completed successfully.

This does not occur with Redis under identical conditions.

### Reproduction

Fully containerized reproduction (no local dependencies beyond Docker):

```
git clone https://github.com/ryan-d-campbell/df-pubsub.git
cd df-pubsub

# Test against Dragonfly (expect FAIL -- exit code 1)
docker compose --profile test-dragonfly up --build --abort-on-container-exit

# Test against Redis (expect PASS -- exit code 0)
docker compose --profile test-redis up --build --abort-on-container-exit
```

**What it does:**
- Starts Dragonfly v1.37.0 (8 proactor threads) and Redis 7-alpine
- Starts 2 Celery workers per backend (concurrency=8 each, prefork pool)
- Warms up worker child processes, then runs 30 rounds of 8 concurrent tasks
- Each task dispatches a group of 50 subtasks and calls `.get(timeout=30)` from within the worker (using `allow_join_result()`)
- On timeout, verifies whether results were stored via direct `GET` to distinguish pub/sub loss from worker overload

### Results

Configuration: `batch_size=50, rounds=30, get_timeout=30s, task_sleep=0.02s, concurrent=8, 2 workers x concurrency=8`

**Dragonfly v1.37.0:**
- 71/240 rounds failed (29.6%)
- 3,550 tasks completed and stored results, but `PUBLISH` was never delivered
- Failures start around round 10 and are sustained through round 30
- In every failure: `worker_incomplete=0` -- all subtasks finished, only pub/sub was lost

**Redis 7.4.8:**
- 240/240 rounds passed (0% failure)
- 0 pub/sub notifications lost

### Production impact

In production, this manifests as `SoftTimeLimitExceeded` errors on Celery workers using `group().get()`. Switching from Redis to Dragonfly introduced ~7,850 errors in 30 days for a workload that had 0 errors in 90 days on Redis.

The workaround is polling with `GroupResult.ready()` (direct `GET`) instead of `.get()` (pub/sub), which completely eliminates the issue.

### Environment

- Dragonfly v1.37.0, `--proactor_threads=8 --cluster_mode=emulated`
- Celery 5.6.3 with `redis` transport
- Python 3.11
- Docker (linux/arm64 and linux/amd64)

### Possibly related

- #3035 (dispatch queue deadlock / back-pressure)
- #924 (response race on subscription)

mkaruza:
Replace the RCU ChannelStore (single ChannelMap with a global
control_block mutex and per-thread channel_store_ pointers) with
ShardedHashMap-backed channels_ and patterns_.

channel_store / channel_store.h:
- Drop ControlBlock, static Destroy(), ApplyAndUnsubscribe(), and
  GetTargetMap()/Modify() helpers
- Add global channel_store pointer; ChannelStore now has a plain dtor
- ChannelStoreUpdater::Record() hashes the key to a shard bucket;
  Apply() calls ShardedHashMap::Mutate per shard, doing in-place RCU
  on existing SubscribeMap entries and upgrading to an exclusive lock
  only when a map slot must be inserted or removed
- FetchSubscribers/ListChannels/PatternCount use FindIf/ForEachShared/
  SizeApproximate instead of iterating the old ChannelMap directly
- UnsubscribeAfterClusterSlotMigration rewritten as ForEachShared
  collection + RemoveAllSubscribers, removing the old two-phase
  ApplyAndUnsubscribe path

server_state.h / server_state.cc:
- Remove channel_store_ member, channel_store() accessor,
  UpdateChannelStore(), and UnsubscribeSlotsAndUpdateChannelStore()

main_service.cc:
- Init allocates the global channel_store; Shutdown calls delete
  (moved after PreShutdown/Shutdown so inflight ops finish first)
- All Publish/PubsubChannels/PubsubPatterns/PubsubNumSub call sites
  updated to use global channel_store directly

db_slice.cc, cluster_family.cc:
- Use global channel_store instead of ServerState::tlocal()->channel_store()

Closes https://github.com/dragonflydb/dragonfly/issues/7056


src/server/channel_store.cc
```diff
@@ -59,68 +61,25 @@ bool ChannelStore::Subscriber::ByThreadId(const Subscriber& lhs, const unsigned
...
-  control_block.most_recent = this;
-}
-
-ChannelStore::ChannelStore(ChannelMap* channels, ChannelMap* patterns)
-    : channels_{channels}, patterns_{patterns} {
+void ChannelStore::Add(string_view channel, ConnectionContext* cntx, uint32_t thread_id,
+                       bool pattern) {
+  auto& map = pattern ? patterns_ : channels_;
+  // try_emplace_l holds the write lock for the duration of the lambda ensuring atomicity
+  map.try_emplace_l(
```
romange: can you explain the arguments here?
when this lambda is called? what about `SubscribeMap{}` object?

src/server/channel_store.h
```diff
@@ -72,102 +61,35 @@ class ChannelStore {
...
 
   // Subscribers for a single channel/pattern.
-  using SubscribeMap = absl::flat_hash_map<ConnectionContext*, ThreadId>;
-
-  // Wrapper around atomic pointer that allows copying and moving.
-  // Made to overcome restrictions of absl::flat_hash_map.
-  // Copy/Move don't need to be atomic with RCU.
-  struct UpdatablePointer {
-    UpdatablePointer(SubscribeMap* sm) : ptr{sm} {
+  using SubscribeMap = phmap::flat_hash_map<ConnectionContext*, ThreadId>;
```
romange: you removed all maps, I thought we will replace only the first level  map (channel->subscribers).
This change is more significant and I am not sure how well it handles contention with multiple publishers sending to the same channel.

romange:
Great that you provided tools to measure pub-sub performance. Could you share the details of how DF performs for 4 and 8 threads?

src/server/channel_store.cc
mkaruza: This function ensures atomicity for:

1. Key lookup
2. Inserting new subscriber to the existing SubscribeMap
2. Insert channel with initial subscriber

src/server/channel_store.h
mkaruza: There was conflict with using  SubscribeMap (absl::flat_hash_ma) and `new` ChannelMap ( phmap::parallel_flat_hash_map_m) so i had to use `phmap::flat_hash_map`

tools/pubsub/latency-test.py
```diff
@@ -0,0 +1,130 @@
...
+    print("Test 2: Concurrent publish (16 clients, 100 publishes each)")
+    print("=" * 60)
+    latencies = test_concurrent_publish(num_concurrent=16, publishes_per_client=100)
+    print(f"Publishes measured: {len(latencies)}")
+    print(f"Min latency:    {min(latencies):.2f} µs")
+    print(f"p50 latency:    {statistics.median(latencies):.2f} µs")
+    print(f"p99 latency:    {sorted(latencies)[int(len(latencies)*0.99)]:.2f} µs")
+    print(f"Max latency:    {max(latencies):.2f} µs")
+    print(f"Avg latency:    {statistics.mean(latencies):.2f} µs")
+    print(f"Throughput:     {len(latencies) / sum(latencies) * 1e6:.0f} ops/sec")
```
glevkovich: I think there are 2 issues things here:

1. Theoretical bug (low chance to happen, but it can happen): If latencies is empty, this line crashes with ZeroDivisionError. Can't happen with the current hardcoded params, but worth a guard.

2. Real math bug:` len / sum(latencies) `gives you 1 / average_latency - that's the throughput of a single thread. But 16 threads ran in parallel, so their latencies overlap in time., they are not running sequentially. Summing the latencies list adds up overlapping concurrent times, which severely underreports the actual throughput (making it look ~16x slower). Throughput should be calculated using the actual wall-clock time from the start to the end of the thread pool execution.
inside test_concurrent_publish:
```
    start_time = time.time()
    with ThreadPoolExecutor(max_workers=num_concurrent) as executor:
        futures = [executor.submit(publish_and_measure, i) for i in range(num_concurrent)]
        for f in futures:
            f.result()
    elapsed = time.time() - start_time

    pubsub.close()
    r.close()

    return latencies, elapsed
```

and in the bottom:

```
     latencies, elapsed = test_concurrent_publish(num_concurrent=16, publishes_per_client=100)
    
    # ... (existing print statements for min, p50, max, etc) ...
    
    if elapsed > 0:
        print(f"Throughput:     {len(latencies) / elapsed:.0f} ops/sec")
    else:
        print("Throughput:     N/A (elapsed time was 0)")
```

src/server/channel_store.h
```diff
@@ -72,102 +61,35 @@ class ChannelStore {
...
-   private:
-    std::atomic<SubscribeMap*> ptr;
   };
 
-  // SubscriberMaps for channels/patterns.
-  struct ChannelMap : absl::flat_hash_map<std::string, UpdatablePointer> {
-    void Add(std::string_view key, ConnectionContext* me, uint32_t thread_id);
-    void Remove(std::string_view key, ConnectionContext* me);
+  using ChannelMap =
+      phmap::parallel_flat_hash_map_m<std::string, SubscribeMap, StringViewHash, std::equal_to<>>;
```
glevkovich: If you refer to https://greg7mdp.github.io/parallel-hashmap/ the full template is:
```
template <class K, class V,
          class Hash      = phmap::priv::hash_default_hash<K>,
          class Eq        = phmap::priv::hash_default_eq<K>,
          class Allocator = phmap::priv::Allocator<std::pair<const K, V>>, // alias for std::allocator
          size_t N        = 4,                 // 2**N submaps
          class Mutex     = phmap::NullMutex>   // use std::mutex to enable internal locks
class parallel_flat_hash_map;
```

`phmap::parallel_flat_hash_map_m` is a convenience alias that hardcodes `std::mutex` as its per-submap lock. see: https://github.com/greg7mdp/parallel-hashmap/blob/master/parallel_hashmap/phmap_fwd_decl.h
and
https://github.com/greg7mdp/parallel-hashmap/blob/master/parallel_hashmap/phmap.h

Because Dragonfly uses cooperative fiber scheduling, if two fibers on the same thread contend on the same submap, the OS thread will deadlock (the holding fiber cannot be scheduled to release it).
In addition, `FetchSubscribers` (the hot path for `PUBLISH`) will serialize concurrent publishers hitting the same submap. 

Is suggest to use the base `parallel_flat_hash_map` template so you can inject Dragonfly's fiber-aware `SharedMutex` as the 7th parameter. This prevents deadlocks and allows concurrent `PUBLISH` reads:

```cpp

// ...

using ChannelMap = phmap::parallel_flat_hash_map<
    std::string, 
    SubscribeMap, 
    StringViewHash, 
    std::equal_to<>,
    phmap::priv::Allocator<std::pair<const std::string, SubscribeMap>>,
    4, // 2^4 = 16 submaps
    util::fb2::SharedMutex // Fiber-safe Reader/Writer lock
>;
```

using SharedMutex boost performance because PUBLISH is a read-only operation. A shared (reader-writer) lock allows multiple concurrent PUBLISH commands to read from the map simultaneously without blocking each other, avoiding a major throughput bottleneck.

src/server/channel_store.cc
```diff
@@ -166,14 +125,15 @@ unsigned ChannelStore::SendMessages(std::string_view channel, facade::ArgRange m
...
 
-  if (auto it = channels_->find(channel); it != channels_->end())
-    Fill(*it->second, string{}, &res);
+  // if_contains holds the per-submap read lock for the duration of the lambda.
+  channels_.if_contains(channel, [&](const auto& kv) { Fill(kv.second, string{}, &res); });
 
-  for (const auto& [pat, subs] : *patterns_) {
-    GlobMatcher matcher{pat, true};
+  // for_each holds each per-submap read lock while iterating it.
+  patterns_.for_each([&](const auto& kv) {
```
glevkovich: This is no longer a point-in-time snapshot, so it's possible to miss a subscriber entirely (you are locking channels and patterns separately).

Example (credit: asked Gemini to construct a concrete scenario):

1. Client X is subscribed to the pattern news.*
2. Publisher calls FetchSubscribers("news.sports")
3. Publisher locks and checks channels_ — Client X is not there. Unlocks.
4. Client X switches subscription: sends SUBSCRIBE news.sports then PUNSUBSCRIBE news.*
5. Server adds Client X to channels_["news.sports"] and removes Client X from patterns_["news.*"]
6. Publisher now locks and iterates patterns_  - Client X is already gone.

Result: Client X was subscribed the entire time (first via pattern, then via channel), but the publisher checked channels_ before they arrived and patterns_ after they left. **The notification is silently dropped.**

The practical likelihood is extremely low (nanosecond race window), and Pub/Sub has no delivery guarantees by design. But since the old RCU approach didn't have this race at all, this is a relaxation of the previous consistency model that should be explicitly documented with an inline comment in FetchSubscribers.

src/server/channel_store.cc
```diff
@@ -59,68 +61,25 @@ bool ChannelStore::Subscriber::ByThreadId(const Subscriber& lhs, const unsigned
...
-  for (auto* chan_map : {store->channels_, store->patterns_}) {
-    chan_map->DeleteAll();
-    delete chan_map;
-  }
-  delete control_block.most_recent;
+void ChannelStore::Remove(string_view channel, ConnectionContext* cntx, bool pattern) {
+  auto& map = pattern ? patterns_ : channels_;
+  // erase_if holds the write lock for its duration, removing cntx from the SubscribeMap
+  // and erasing the channel entry atomically if no subscribers remain.
+  map.erase_if(string{channel}, [&](auto& kv) {
```
glevkovich: This creates a temporary string (allocation) for every unsubscribe, which can be avoided.
fix:

```
  map.erase_if(channel, [&](auto& kv) {
    kv.second.erase(cntx);
    return kv.second.empty();
  });
```

src/server/channel_store.h
```diff
@@ -14,37 +15,20 @@
...
-// pointer inside ChannelMap.
-//
-// To prevent parallel (and thus overlapping) updates, a centralized ControlBlock is used.
-// Update operations are carried out by the ChannelStoreUpdater.
-//
-// A centralized ChannelStore, contrary to sharded storage, avoids contention on a single shard
-// thread for heavy throughput on a single channel and thus seamlessly scales on multiple threads
-// even with a small number of channels. In general, it has a slightly lower latency, due to the
-// fact that no hop is required to fetch the subscribers.
+// A single global instance is shared across all threads. Concurrency is
```
glevkovich: The comment mentions the map is "protected by its own std::mutex".

If you accept /  apply the fix for the deadlock I mentioned (suggested to use fb2::SharedMutex), please update this documentation to reflect that the map uses Dragonfly's fiber-aware SharedMutex. This will ensure future readers understand how the data structure safely handles concurrent fiber access.

src/server/channel_store.h
mkaruza: Yes! Changed.

src/server/channel_store.cc
mkaruza: There is trivial change and that is to introduce mute  in ChannelStore that will protect this. I'll take some time to check and think if there is something that we can do.
But, IMO, we should keep as it is - and update comment that this situation is possible.

src/core/sharded_hash_map.h
```diff
@@ -0,0 +1,194 @@
...
+//
+// Each shard contains an absl::flat_hash_map protected by two fiber-aware locks:
+//   - write_mu_ (Mutex): serializes writers. Only one writer can modify the shard at a time.
+//   - read_mu_ (SharedMutex): guards readers. Acquired in shared mode for reads (FindIf,
+//     ForEachShared, Size) and in exclusive mode when a writer needs to commit changes that
+//     must be visible atomically to readers.
+//
+// The two-lock design allows multiple concurrent readers on a shard while a single writer
+// prepares its mutation (holding only write_mu_). The writer then briefly acquires read_mu_
+// exclusively to publish the change, minimizing the window during which readers are blocked.
```
romange, line 29: @mkaruza  was it your idea?

My first thinking would be to use a single SharedMutex for everything but this two-phase approach is interesting

src/core/sharded_hash_map.h
```diff
@@ -0,0 +1,194 @@
...
+  //       auto lm = lock_readers();
+  //       lm.map[key] = new_value;  // now no reader sees a partial update
+  //   });
+  template <typename F> void Mutate(const K& key, F&& f) {
+    Shard& shard = shards_[Hash{}(key) % NUM_SHARDS];
+    std::unique_lock write_lock{shard.write_mu_};
+    std::forward<F>(f)(static_cast<const Map&>(shard.map_), [&shard]() -> LockedMap {
+      return {std::unique_lock<util::fb2::SharedMutex>{shard.read_mu_}, shard.map_};
+    });
+  }
```
dranikpg: why not call `Mutate(Hash{}(key) % NUM_SHARDS, std::forward(f))`?

src/core/sharded_hash_map.h
```diff
@@ -0,0 +1,194 @@
...
+    std::forward<F>(f)();
+  }
+
+  // Shard-index overload of WithReadExclusiveLock. Same semantics but addresses the shard
+  // directly by its index `sid` (0 <= sid < NUM_SHARDS).
+  template <typename F> void WithReadExclusiveLock(size_t sid, F&& f) {
+    DCHECK_LT(sid, NUM_SHARDS);
+    std::unique_lock l{shards_[sid].read_mu_};
+    std::forward<F>(f)();
+  }
```
dranikpg, line 194: For some reason you tried to design every function around callbacks. Why not just return the lock? Why not call the sid version from the key one?

src/core/sharded_hash_map.h
dranikpg: also it took me a minute to decipher that you are passing a lambda 🙂

src/core/sharded_hash_map.h
dranikpg: I missed the discussion in choosing something different

dranikpg: this is a pattern that appears sometimes,

with shards, you strictly also don't need locking. the problem with rcu was it's update time + modification heavy for base level table changes

however the pattern single writer + readers is common, there is something like left-right datastructures, that maintain the same datastructure in two copies and then just "flip" an atomic pointer for readers either to use the left or right

mkaruza:
@dranikpg Please look into #7196 and leave comment around ShardedHashMap that is PR that needs to merged first (i just rebased this PR on top of that one, so once that once will be merged this PR will contain only changes related to pubsub).

src/core/sharded_hash_map.h
mkaruza: @romange yes. Idea was to have write/read path controlled by different locks (so we don't block as much in read path) because RCU on second layer.

src/core/sharded_hash_map.h
mkaruza: Previously work was with around parallel_hashmap that used this callback based approach (so maybe just recency bias)

src/server/channel_store.cc
```diff
@@ -249,157 +229,91 @@ ChannelStoreUpdater::ChannelStoreUpdater(bool pattern, bool to_add, ConnectionCo
...
+            auto* old_sm = it->second.Get();
+            auto* new_sm = new ChannelStore::SubscribeMap{*old_sm};
+            new_sm->emplace(cntx_, thread_id_);
+            it->second.Set(new_sm);
+            freelist_[sid].push_back(old_sm);
+          }
+        } else {
+          if (it == m.end())
+            continue;
+          if (it->second->size() == 1) {
```
glevkovich, line 289: This is a good place to DCHECK (after the continue)
```
// A channel must never have zero subscribers.
DCHECK(!it->second->empty()); 
```

src/server/channel_store.cc
```diff
@@ -249,157 +229,91 @@ ChannelStoreUpdater::ChannelStoreUpdater(bool pattern, bool to_add, ConnectionCo
...
+            needs_map_change[i] = true;
```
glevkovich, line 293: add before this line DCHECK that you dropped from Modify(), and  I don't see it in the new code:

```
// If a channel is being deleted because the last subscriber is leaving, 
// make sure the one leaving is actually that last subscriber.
DCHECK(it->second->begin()->first == cntx_);
```

src/server/channel_store.cc
```diff
@@ -249,157 +229,91 @@ ChannelStoreUpdater::ChannelStoreUpdater(bool pattern, bool to_add, ConnectionCo
...
+            has_map_change = true;
+          } else {
+            auto* old_sm = it->second.Get();
+            auto* new_sm = new ChannelStore::SubscribeMap{*old_sm};
```
glevkovich, line 300: Add a DCHECK above this line:

```
// You cannot unsubscribe from a channel if you aren't subscribed to it in the first place.
DCHECK(old_sm->contains(cntx_));
```

src/server/channel_store.h
```diff
@@ -58,12 +67,14 @@ class ChannelStore {
     std::string pattern;  // non-empty if registered via psubscribe
   };
 
-  ChannelStore();
+  ChannelStore() = default;
+  ~ChannelStore();
```
glevkovich, line 71: I think it is safer to add an explicit delete for copy/move here. You don't want to copy or move ChannelMap objects (members of this class) since they hold mutexes, which cannot be moved or copied.

Currently, the compiler implicitly deletes these for you, but relying on that is "fragile". If someone changes ShardedHashMap internals later (for example, wrapping a mutex in a smart pointer), the class could silently become movable, which would break things.

```
ChannelStore(const ChannelStore&) = delete;
ChannelStore& operator=(const ChannelStore&) = delete;
```

src/server/channel_store.h
```diff
@@ -74,100 +85,81 @@ class ChannelStore {
...
+  // concurrently active. The old SubscribeMap is deleted via WithReadExclusiveLock
+  // after write_mu_ is released — that acquires read_mu_ exclusive, draining any
+  // in-flight readers before calling delete.
   struct UpdatablePointer {
-    UpdatablePointer(SubscribeMap* sm) : ptr{sm} {
+    explicit UpdatablePointer(SubscribeMap* sm) : ptr{sm} {
     }
 
     UpdatablePointer(const UpdatablePointer& other);
+    UpdatablePointer(UpdatablePointer&& other) noexcept;
```
glevkovich, line 108: Here we declare a copy constructor and a move constructor for UpdatablePointer, but no assignment operators.
Defining a copy constructor suppresses the implicit generation of the move-assignment operator. Relying on implicit behavior here is fragile and not recommended (the Rule of 5).

You should choose one of two paths to make your intent clear: either explicitly define the assignment operators, or explicitly delete them.

Option 1: Define them (if reassignment is allowed)
```
UpdatablePointer& operator=(const UpdatablePointer& other);
UpdatablePointer& operator=(UpdatablePointer&& other) noexcept;
```

Option 2: Delete them (if reassignment should be banned)
```
UpdatablePointer& operator=(const UpdatablePointer&) = delete;
UpdatablePointer& operator=(UpdatablePointer&&) = delete;
```

src/server/channel_store.h
```diff
@@ -74,100 +85,81 @@ class ChannelStore {
 
   using ChannelsSubMap =
```
glevkovich: I think after some changes and cleanup you did here, ChannelsSubMap should not be public anymore. i should now be private.

src/server/channel_store.cc
```diff
@@ -249,157 +229,91 @@ ChannelStoreUpdater::ChannelStoreUpdater(bool pattern, bool to_add, ConnectionCo
...
+      // Apply map changes under exclusive if needed.
+      if (has_map_change) {
+        auto locked_map = LockReaders();
+        for (size_t i = 0; i < shard_keys.size(); ++i) {
+          if (!needs_map_change[i]) {
+            continue;
+          }
+          std::string_view key = shard_keys[i];
+          if (to_add_) {
+            locked_map.map.emplace(std::string{key},
```
glevkovich, line 318: You are doing this emplace with: `new SubscribeMap`  which will leak if the key is already present (since UpdatablePointer has no dtor that will delete SubscribeMap). Currently prevented by this logic in conn_context.cc just before calling Apply:
```
  for (string_view channel : channels) {
    if (to_add && local_store.emplace(channel).second)
      csu.Record(channel);
    else if (!to_add && local_store.erase(channel) > 0)
      csu.Record(channel);

    if (to_reply)
      result[i++] = sinfo.SubscriptionCount();
  }

csu.Apply();
..
```

but I still think you should add a guard here before this line to make sure key is not there and not leak:
`DCHECK(!locked_map.map.contains(key)); `

src/server/channel_store.cc
```diff
@@ -182,59 +140,81 @@ vector<ChannelStore::Subscriber> ChannelStore::FetchSubscribers(string_view chan
...
+    channel_subs_map.emplace(channel, std::move(subs));
+
+  shard_set->pool()->AwaitFiberOnAll([&channel_subs_map](unsigned idx, util::ProactorBase*) {
+    channel_store->UnsubscribeConnectionsFromDeletedSlots(channel_subs_map, idx);
+  });
+}
+
+void ChannelStore::RemoveAllSubscribers(bool pattern, string_view channel) {
+  ChannelMap& map = pattern ? patterns_ : channels_;
+  string key{channel};
```
glevkovich: This is an unnecessary heap allocation. I think this will work:
```
map.Mutate(channel, [&](const auto& m, auto AcquireReadExclusiveLock) {
    auto it = m.find(channel);
```

src/server/channel_store.cc
```diff
@@ -79,71 +76,34 @@ const ChannelStore::SubscribeMap& ChannelStore::UpdatablePointer::operator*() co
...
-ChannelStore::ControlBlock ChannelStore::control_block;
-
-unsigned ChannelStore::SendMessages(std::string_view channel, facade::ArgRange messages,
+unsigned ChannelStore::SendMessages(string_view channel, facade::ArgRange messages,
                                     bool sharded) const {
   vector<Subscriber> subscribers = FetchSubscribers(channel);
   if (subscribers.empty())
     return 0;
 
-  // Make sure none of the threads publish buffer limits is reached. We don't reserve memory ahead
```
glevkovich, line 130: I am not sure why did you remove this 3 lines comment above last_thread?

src/server/channel_store.cc
```diff
@@ -79,71 +76,34 @@ const ChannelStore::SubscribeMap& ChannelStore::UpdatablePointer::operator*() co
...
     int sub_thread = sub.LastKnownThreadId();
     DCHECK_LE(last_thread, sub_thread);
-    if (last_thread == sub_thread)  // skip same thread
+    if (last_thread == sub_thread)
       continue;
-
     if (sub.IsExpired())
       continue;
-
-    // Make sure the connection thread has enough memory budget to accept the message.
```
glevkovich, line 144: I am not sure why did you removed the 3 lines comment above `facade::Connection::EnsureMemoryBudget(sub_thread);`

src/server/channel_store.h
```diff
@@ -4,44 +4,53 @@
...
+//     1. Acquire write_mu_ exclusively (Mutate) — serializes writers on this shard.
+//     2. Copy the SubscribeMap, modify the copy, atomically swap via UpdatablePointer::Set.
+//        Readers are NOT blocked; they may still read the old pointer.
+//     3. Release write_mu_ (Mutate returns).
+//     4. Acquire read_mu_ exclusively (WithReadExclusiveLock, separate call) — this drains
+//        any reader that loaded the old SubscribeMap pointer, then deletes it.
+//
+//   Key does not exist (first subscriber) / last subscriber removed (channel erased):
+//     Inside the Mutate callback, acquire read_mu_ exclusively via the
+//     AcquireReaderExclusiveLock callable — blocks all readers in the shard while
```
glevkovich, line 51: nit: naming inconsistency.
Here you use the name `AcquireReaderExclusiveLock`
but then in 2 other places you use different name, which I think is not  consistent:

1)
in `RemoveAllSubscribers` you use `AcquireReadExclusiveLock`:
```
map.Mutate(key, [&](const auto& m, auto AcquireReadExclusiveLock) {
    auto it = m.find(key);
    if (it == m.end())
      return;
    auto locked_map = AcquireReadExclusiveLock();
```
2)
in  Apply() you use LockReaders:
```
map.Mutate(ChannelStore::ChannelMap::ShardId{sid}, [&](const auto& m, auto LockReaders) {
    ...
    auto locked_map = LockReaders();
```

For someone new to the code this is really confusing, 3 names for the same thing.
consider giving one consistent name.

src/server/channel_store.cc
```diff
@@ -51,24 +49,23 @@ auto BuildSender(string_view channel, facade::ArgRange messages, bool sharded =
...
-  return lhs.LastKnownThreadId() < thread;
+ChannelStore::UpdatablePointer::UpdatablePointer(const UpdatablePointer& other) {
+  ptr.store(other.ptr.load(memory_order_acquire), memory_order_release);
 }
 
-ChannelStore::UpdatablePointer::UpdatablePointer(const UpdatablePointer& other) {
-  ptr.store(other.ptr.load(memory_order_relaxed), memory_order_relaxed);
+ChannelStore::UpdatablePointer::UpdatablePointer(UpdatablePointer&& other) noexcept {
+  ptr.store(other.ptr.load(memory_order_acquire), memory_order_release);
+  other.ptr.store(nullptr, memory_order_release);
```
glevkovich: This nulls a moved-from object that is immediately destroyed under exclusive write_mu_. No other thread can load at this point, so the release ordering synchronizes with nothing.
memory_order_relaxed suffices.

src/server/channel_store.cc
```diff
@@ -249,157 +229,91 @@ ChannelStoreUpdater::ChannelStoreUpdater(bool pattern, bool to_add, ConnectionCo
...
+          }
+          std::string_view key = shard_keys[i];
+          if (to_add_) {
+            locked_map.map.emplace(std::string{key},
+                                   ChannelStore::UpdatablePointer{
+                                       new ChannelStore::SubscribeMap{{cntx_, thread_id_}}});
+          } else {
+            auto it = locked_map.map.find(key);
+            if (it == locked_map.map.end()) {
+              continue;
```
glevkovich: You are silencing a bug here.

This is the second phase (exclusive lock). In the scan pass above we already confirmed the key exists in the map and its SubscribeMap has exactly one entry (if (it->second->size() == 1)). Both phases run inside the same Mutate call under the same exclusive write_mu_, so nothing can erase this key in between.

If find returns end() here, it means map corruption -  the silent continue would hide that, skip the delete, and leak the SubscribeMap. I suggest to replace with:

```
// must exist: confirmed in scan pass under same write_mu_
DCHECK_NE(it, locked_map.map.end());
delete it->second.Get();
locked_map.map.erase(it);
```

src/server/channel_store.cc
mkaruza: Ok, no harm to add check, but it's done to not check duplicated because it is used only from current path where we remove duplicate keys.

glevkovich (approved):
LGTM

PR #7177, opened 2026-04-17
fix: SHRINK build break from obj_type out of scope after merge

vyavdoshenko:
The merge of #7160 (SHRINK zombie cleanup) and #7173 (SHRINK memory accounting refactor) left `obj_type` used out of its scope, breaking the build on all platforms.

Root cause:
#7173 moved the type/encoding check into an inner block `{ ... }` where `obj_type` is declared. #7160's empty-container cleanup was kept outside that block, but still referenced `obj_type`, which no longer exists there.

Fix:
The mutable iterator `it_res` (already obtained a few lines above for the actual Shrink call) can be reused for deletion - no second `FindMutable` is needed. This also drops the redundant lookup.


BorysTheDev (approved):

kostasrim (approved):
daily build also broke and I was about to sent a PR. Thanks for fixing!

PR #7185, opened 2026-04-19
fix(server): Process empty buckets in snapshot

dranikpg:
1. Allow "done" (aka end) iterators to support GetVersion calls by keeping the dash pointer and adding a done_ field
2. Allow including empty buckets in TraverseBuckets
3. Stamp snapshot versions on empty buckets

Needed for #7060 

Currently, the comparison `bucket_version < snapsho_version` determines whether the snapshot serialization process has reached this specific bucket. This fact can be later used to make decisions and perform various optimizations (for example journal omits, see #7060).

_However_, empty buckets are currently fully ignored by all actors: the dashtable iteration functions, the iterator itself and the snapshot. During serialization, their versions are not updated - because there is nothing to serialize. However, just by looking at the bucket and it's not-updated version, we can't tell _if the snapshot loop has already reached it or will do so only in the future._

This PR allows them in all places, so they act as regular buckets and their versions are updated

### DIFF
--- a/src/core/dash.h
+++ b/src/core/dash.h
@@ -348,7 +348,7 @@ class DashTable : public detail::DashTableBase {
   // Unlike Traverse, TraverseBuckets calls cb once on bucket iterator and not on each entry in
   // bucket. TraverseBuckets is stable during table mutations. It guarantees traversing all buckets
   // that existed at the beginning of traversal.
-  template <typename Cb> Cursor TraverseBuckets(Cursor curs, Cb&& cb);
+  template <typename Cb> Cursor TraverseBuckets(Cursor curs, Cb&& cb, bool visit_empty = false);
 
   // Traverses over a single bucket in table and calls cb(iterator). The traverse order will be
   // segment by segment over physical backets.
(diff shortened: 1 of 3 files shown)


dranikpg:
augment review

dranikpg:
I haven't included migrations intentionally, I will first implement omits for snapshots fully and then transfer this knowledge to migrations

romange:
For sake of discussion, if the bucket is empty and serialization have reached it, we **must**  emit a journal change.
If it have not reached it, and i'ts empty,  we **may** emit a journal change, and then the subsequent traversal will omit it as its version will be greater than the snapshot version.

So in either case we if emit a journal change when encountering an empty bucket in `OnDbChange ` will behave correctly even if we do not update the bucket version. Am I wrong?

dranikpg:
> For sake of discussion, if the bucket is empty and serialization have reached it, we must emit a journal change.
If it have not reached it, and i'ts empty, we may emit a journal change, and then the subsequent traversal will omit it as its version will be greater than the snapshot version.

Yes, we _may_ emit a journal change. And it might look harmless because there is no difference when to serialize this single value.

But we actually _don't want to do it_. It is because it increases the "attack" surface - any subsequent write to that bucket now has to be journaled. Compare that with omitting this bucket - no writes will be journaled.

romange (approved):

PR #7190, opened 2026-04-20
fix: OOB in BITCOUNT and some refactoring

BorysTheDev:
I've run a libc++ hardening build, found a bug in BITCOUNT. Also did some small refactoring to have a correct build.

Fixes an out-of-bounds read in BITCOUNT (BIT mode) discovered under libc++ hardening, and does small refactoring to keep builds correct.

Changes:

- Refactors bit-range counting to use an inclusive end index and avoid reading past the last byte.
- Adds a regression test for ranges that end on the last bit / a byte boundary.
- Moves several operator<< implementations out of headers and replaces <ostream> with <iosfwd> where possible.
- Adjusts a Linux-only static_assert to skip libc++ and qualifies ::bind in socket tests.

### DIFF
--- a/src/core/dragonfly_core.cc
+++ b/src/core/dragonfly_core.cc
@@ -4,10 +4,16 @@
 
 #include <absl/base/macros.h>
 
+#include <ostream>
+
 #include "base/logging.h"
 #include "core/intent_lock.h"
 
 namespace dfly {
+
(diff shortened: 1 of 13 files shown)


kostasrim:
@BorysTheDev fake redis tests are failing

dranikpg (approved):

PR #7192, opened 2026-04-20
feat: DEBUG TRAFFIC per-listener recording and memcache replay

vyavdoshenko:
Extend `DEBUG TRAFFIC` to record any combination of RESP, memcache, and admin listeners (previously only the main RESP interface) and make `tools/replay` faithfully reproduce memcache traffic via raw wire commands.

Changes:
- New command syntax:
  - `DEBUG TRAFFIC START <path> [LISTENER <resp|memcache|admin|all> ...]`
  - `DEBUG TRAFFIC STOP`
  Replaces `DEBUG TRAFFIC <path>|STOP`. No backwards-compat shim.
- `Connection::ListenerType { RESP=1, MEMCACHE=2, ADMIN=3 }`, set once in `OnConnectionStart` from the listener role + protocol.
- `TrafficLogger` gains a listener bitmask; the hot-path gate in `ParseRedis` no longer hard-codes `IsMain()`.
- Memcache parse path now records commands. `LogMemcacheTraffic` serializes `backed_args` plus the scalar `Command` members that live outside them (`flags`, `expire_ts`, `delta`, `cas_unique`, GAT expire) so the replayer has enough context to reproduce the exact wire command.
- `MemcacheParser::CmdName(CmdType)` exposes the inverse of the parser's command map.
- A second `START` while logging is active now returns an error; the active recording is left untouched (previously it was silently aborted).

File format v3:
- Version byte 2 -> 3.
- Per-record `listener_type` packed into bits 8..15 of the former `has_more` field. Binary layout is unchanged; v2 files remain readable by v3 readers (treated as RESP-only).

tools/replay:
- Parser accepts both v2 and v3 files.
- Per-record dispatch selects the backend by listener type:
  - `-host` for RESP records
  - `-mc-host` for memcache records
  - `-admin-host` (falls back to `-host`) for admin-listener records

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -177,8 +177,10 @@ struct TrafficLogger {
   // Also, makes sure that LogTraffic are executed atomically.
   fb2::Mutex mutex;
   unique_ptr<io::WriteFile> log_file;
-  // Bit mask over Connection::ListenerType values, only matching listeners are logged.
-  uint32_t listener_mask = 0;
+  // Listener type that this thread's file is recording. Only connections with a
+  // matching `listener_type_` produce records; others are skipped on the hot path.
+  // Set once when the file is opened, cleared in ResetLocked().
+  Connection::ListenerType listener_type = Connection::ListenerType::RESP;
 
   void ResetLocked();
   // Returns true if Write succeeded, false if it failed and the recording should be aborted.
@@ -191,7 +193,7 @@ void TrafficLogger::ResetLocked() {
     std::ignore = log_file->Close();
     log_file.reset();
   }
-  listener_mask = 0;
+  listener_type = Connection::ListenerType::RESP;
 }
 
 // Returns true if Write succeeded, false if it failed and the recording should be aborted.
@@ -223,9 +225,10 @@ thread_local uint32 pipeline_wait_batch_usec = absl::GetFlag(FLAGS_pipeline_wait
 
 // Opens the per-thread traffic log file. Distinguishes three outcomes so the caller
 // can report an accurate error to the user (was the logger already running, or did
-// we fail to open a file). `listener_mask` is only committed after the file is
+// we fail to open a file). `listener_type` is only committed after the file is
 // successfully opened so the logger's state stays consistent on failure.
-Connection::StartTrafficResult OpenTrafficLogger(string_view base_path, uint32_t listener_mask) {
+Connection::StartTrafficResult OpenTrafficLogger(string_view base_path,
+                                                 Connection::ListenerType listener_type) {
   using Res = Connection::StartTrafficResult;
   unique_lock lk{tl_traffic_logger.mutex};
   if (tl_traffic_logger.log_file)
@@ -241,24 +244,22 @@ Connection::StartTrafficResult OpenTrafficLogger(string_view base_path, uint32_t
     return Res::kOpenFailed;
   }
   tl_traffic_logger.log_file = unique_ptr<io::WriteFile>{file.value()};
-  tl_traffic_logger.listener_mask = listener_mask;
+  tl_traffic_logger.listener_type = listener_type;
 #else
   LOG(WARNING) << "Traffic logger is only supported on Linux";
   return Res::kOpenFailed;
 #endif
 
-  // File header: version byte.
-  // v3 added per-record listener type packed into the upper bits (8..15) of the former
-  // `has_more` (now `flags`) field. v2 files are forwards-compatible for v3 readers:
-  // the upper bits are zero and get mapped to RESP. v2 readers cannot parse v3 files.
-  uint8_t version[1] = {3};
-  std::ignore = tl_traffic_logger.log_file->Write(version);
+  // File header: version byte (v3), followed by a single byte carrying the listener
+  // type for the whole file. Every record in the file belongs to this listener.
+  uint8_t header[2] = {3, static_cast<uint8_t>(listener_type)};
+  std::ignore = tl_traffic_logger.log_file->Write(header);
   return Res::kStarted;
 }
 
 // Writes a single record. `parts[0]` is the command name, following entries are its arguments.
 void LogTrafficParts(uint32_t id, bool has_more, uint32_t db_index,
-                     Connection::ListenerType listener_type, absl::Span<const string_view> parts) {
+                     absl::Span<const string_view> parts) {
   if (parts.empty())
     return;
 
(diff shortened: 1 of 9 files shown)


romange (commented):
wow, hold your horses, cowboy :)

that's not what we want, imho.

We want to record ONE listener at any time. But currently we can not record anything but the main one.

so the suggestion was to extend the `DEBUG TRAFFIC <some new options> ` to cover more listeners BUT we do not need to listen on multiple listeners at the same time. 
So I expect 0 changes in replay tool and I think the whole change is going to be much simpler. 

sorry if it was not clear

vyavdoshenko:
@romange 

Quick context on the scope: the original issue explicitly asked for memcache listener recording ("monitor traffic on memcached listener") and mentioned multiple selected listeners in the body. I built against that.

The replay-tool changes are unavoidable for memcache support, not for multi-listener: memcache uses a text protocol distinct from RESP, so tools/replay cannot reproduce recorded memcache commands through the existing redis.Client path - the raw TCP memcache client, per-record routing, and -mc-host flag all exist specifically to make memcache recordings replayable. Without them, recording a memcache listener would be write-only, and the feature would be half-done.

Admin support is genuinely tiny - two lines in OnConnectionStart plus one -admin-host flag + routing branch in replay. Dropping it is easy if you prefer; keeping it doesn't change the PR size meaningfully.

Happy to simplify in either direction:

- keep multi-listener in a single recording (current state) - matches original issue literal text.
- restrict to one listener per DEBUG TRAFFIC START- drops the bitmask / LISTENER all / mixed files. Still needs all the memcache replay work.
Either way, the memcache recording + replay pair comes as a unit.

If you prefer strictly one-listener-per-file, I can fold listener_type into the file header and drop per-record tagging - it would be cleaner. Say the word.

romange:
Ok, understand.

1. Lets add mc support in replay
2. Lets support multiple listeners including admin
3. Lets restrict to a single listeners per recording. I do not think mixing multiple formats, listeners in a single fileset brings any value.

src/facade/dragonfly_connection.cc
```diff
@@ -230,21 +241,29 @@ void OpenTrafficLogger(string_view base_path) {
...
+  return Res::kStarted;
 }
 
-void LogTraffic(uint32_t id, bool has_more, const cmn::BackedArguments& args,
-                ServiceInterface::ContextInfo ci) {
-  string_view cmd = args.Front();
+// Writes a single record. `parts[0]` is the command name, following entries are its arguments.
+void LogTrafficParts(uint32_t id, bool has_more, uint32_t db_index,
+                     absl::Span<const string_view> parts) {
+  if (parts.empty())
```
romange: `DCHECK(!parts.empty());` ?

src/facade/dragonfly_connection.h
```diff
@@ -191,6 +191,19 @@ class Connection : public util::Connection {
   // This method returns true for customer facing listeners.
   bool IsMainOrMemcache() const;
 
+  // Classification of the listener this connection belongs to, used as a filter for the
+  // traffic logger (DEBUG TRAFFIC) and persisted in the recorded file format.
+  // The numeric values are part of the on-disk format, do not change them.
+  enum class ListenerType : uint8_t {
+    RESP = 1,      // main RESP listener (TCP and unix-socket)
```
romange:
```suggestion
    MAIN_RESP = 1,      // main RESP listener (TCP and unix-socket)
```

as admin is also resp

src/server/debugcmd.cc
```diff
@@ -1051,25 +1053,103 @@ void DebugCmd::Exec(CommandContext* cmd_cntx) {
...
+  // Syntax:
+  //   DEBUG TRAFFIC STOP
+  //   DEBUG TRAFFIC START <path> LISTENER <resp|memcache|admin>
+  // A recording captures exactly one listener type; mixing protocols in one file is
+  // intentionally not supported.
+  if (args.empty()) {
+    return cmd_cntx->SendError(facade::kSyntaxErr);
+  }
+
+  string first = absl::AsciiStrToUpper(ArgS(args, 0));
```
romange: consider using `CmdArgParser` if you changing the parsing code.

src/server/debugcmd.cc
```diff
@@ -1051,25 +1053,103 @@ void DebugCmd::Exec(CommandContext* cmd_cntx) {
...
+    return cmd_cntx->SendError(facade::kSyntaxErr);
+  }
+
+  string path(ArgS(args, 1));
+  string keyword = absl::AsciiStrToUpper(ArgS(args, 2));
+  if (keyword != "LISTENER") {
+    return cmd_cntx->SendError(facade::kSyntaxErr);
+  }
+
+  Connection::ListenerType listener_type;
```
romange: `CmdArgParser` ahs a convenient MapNext function  that maps arguments to enums

src/server/debugcmd.cc
```diff
@@ -1051,25 +1053,103 @@ void DebugCmd::Exec(CommandContext* cmd_cntx) {
...
+        }
+      });
+
+  unsigned started = started_new.load(std::memory_order_relaxed);
+  unsigned refused = already_logging.load(std::memory_order_relaxed);
+  unsigned failed = open_failed.load(std::memory_order_relaxed);
+
+  // Failures are always terminal: if any thread failed to open its file the recording
+  // would be incomplete. Roll back any thread that did start so we end up fully stopped.
+  if (failed > 0) {
```
romange: I find this code too verbose.
you can open `if (started >0)` block first and  both failed and refuse cases in the same block with calling `StopTrafficLogging` only once.

tools/replay/main.go
```diff
@@ -13,8 +13,10 @@ import (
 	"github.com/pterm/pterm"
 )
 
-var fHost = flag.String("host", "127.0.0.1:6379", "Redis host")
-var fCompareHost = flag.String("compare-host", "", "Redis host to compare with")
+var fHost = flag.String("host", "127.0.0.1:6379", "RESP host for main-listener replay")
```
romange: why do you need a dedicated named flag for the host/port pair?

src/facade/dragonfly_connection.cc
vyavdoshenko: done

src/facade/dragonfly_connection.h
vyavdoshenko: fixed

src/server/debugcmd.cc
vyavdoshenko: fixed

src/server/debugcmd.cc
vyavdoshenko: fixed

src/server/debugcmd.cc
vyavdoshenko: fixed

tools/replay/main.go
vyavdoshenko: simplified

src/facade/dragonfly_connection.h
```diff
@@ -191,6 +191,21 @@ class Connection : public util::Connection {
   // This method returns true for customer facing listeners.
   bool IsMainOrMemcache() const;
 
+  // Classification of the listener this connection belongs to, used as a filter for the
+  // traffic logger (DEBUG TRAFFIC) and persisted in the recorded file format.
```
romange, line 195: nit: I am sure you noticed that we also have  IsPrivileged, IsMain and IsMainOrMemcache that carry similar semantic meaning. I do not see why we must limit `ListenerType` interface to only debug traffic - we could get rid of some of these methods and make them wrappers around this enum.

tools/replay/main.go
```diff
@@ -13,8 +13,13 @@ import (
 	"github.com/pterm/pterm"
 )
 
-var fHost = flag.String("host", "127.0.0.1:6379", "Redis host")
-var fCompareHost = flag.String("compare-host", "", "Redis host to compare with")
+var fHost = flag.String("host", "127.0.0.1:6379",
+	"host:port of the replay target. By default interpreted as RESP (both main and admin listener "+
+		"records go here); pass -memcache to interpret it as a memcached server")
+var fMemcache = flag.Bool("memcache", false,
+	"if true, -host is a memcached server and only memcache-listener files are replayed. "+
```
romange: is not it something you put into the traffic files? 
why do you need  a special "fMemcache" flag?

romange (dismissed):

src/facade/dragonfly_connection.h
vyavdoshenko: agree, worth a follow-up; in this PR I'd prefer not to change semantics of IsMain() for UDS

tools/replay/main.go
vyavdoshenko: fixed

BorysTheDev (approved):

PR #7196, opened 2026-04-21
feat(core): Add ShardedHashMap — sharded concurrent hash map

mkaruza:
Introduce ShardedHashMap<K, V, NUM_SHARDS, HASH, EQ>, a general-purpose thread-safe hash map sharded across independent shards for high concurrency.

Each shard holds an absl::flat_hash_map guarded by two fiber-aware locks:
- write_mu_ (fb2::Mutex): serializes writers within a shard.
- read_mu_ (fb2::SharedMutex): taken shared for reads, exclusively only for map changes

The two-lock design keeps readers unblocked while a writer prepares its mutation; read_mu_ is acquired exclusively only at the moment of commit, minimizing reader stall time. Each shard is aligned to a CPU cache line (alignas(64)) to prevent false sharing under concurrent access.

Provides: FindIf, ForEachShared, ForEachExclusive, Mutate (by key and shard-id), ShardOf, WithReadExclusiveLock, SizeApproximate.

Add sharded_hash_map_test with 15 tests covering single-threaded correctness and concurrent reader/writer fiber scenarios.


src/core/sharded_hash_map.h
```diff
@@ -0,0 +1,172 @@
...
+
+  // Looks up `key` under a shared read lock on its shard. If found, invokes f(const V&)
+  // with the mapped value while still holding the lock, then returns true.
+  // Returns false if the key is not present. The callback must not modify the value.
+  //
+  // The template parameter Q allows heterogeneous lookup — any type hashable via
+  // absl::Hash<Q> and comparable against K can be used (e.g., std::string_view for
+  // std::string keys).
+  template <typename Q, typename F> bool FindIf(const Q& key, F&& f) const {
+    const Shard& shard = shards_[absl::Hash<Q>{}(key) % NUM_SHARDS];
```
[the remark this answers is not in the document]
mkaruza: Added HASH template parameter.

src/core/sharded_hash_map.h
```diff
@@ -0,0 +1,172 @@
...
+  // Typical usage pattern:
+  //   map.Mutate(key, [&](auto& m, auto lock_readers) {
+  //       /* prepare data without blocking readers */
+  //       auto reader_lock = lock_readers();
+  //       m[key] = new_value;  // now no reader sees a partial update
+  //   });
+  template <typename F> void Mutate(const K& key, F&& f) {
+    Shard& shard = shards_[absl::Hash<K>{}(key) % NUM_SHARDS];
+    std::unique_lock write_lock{shard.write_mu_};
+    std::forward<F>(f)(shard.map_, [&shard]() { return std::unique_lock{shard.read_mu_}; });
```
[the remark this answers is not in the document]
mkaruza: Returning mutable shard with lock

src/core/sharded_hash_map.h
```diff
@@ -0,0 +1,194 @@
...
+
+ public:
+  static constexpr size_t kNumShards = NUM_SHARDS;
+
+  // Returned by the AcquireReaderLock callable passed to Mutate(). Holds an exclusive lock on
+  // read_mu_ for the duration of its lifetime and exposes a mutable reference to the shard
+  // map. Mutations must be performed through LockedMap::map to guarantee that no reader
+  // observes a partial update.
+  struct LockedMap {
+    std::unique_lock<util::fb2::SharedMutex> lock;
```
glevkovich, line 73: you are using std::unique_lock but you do not include it. Some toolchains add include for <mutex> when <shared_mutex> is included but this is not gurenteed.
Recommended: add #include <mutex> with the existing <shared_mutex>.

src/core/sharded_hash_map.h
```diff
@@ -0,0 +1,194 @@
...
+
+ private:
+  // Aligned to cache line.
+  struct alignas(64) Shard {
+    util::fb2::Mutex write_mu_;
+    mutable util::fb2::SharedMutex read_mu_;
+    absl::flat_hash_map<K, V> map_;
+  };
+
+  std::array<Shard, NUM_SHARDS> shards_;
```
glevkovich, line 215: you are using std::array, but do not include it on the top.
Recommended:   `#include <array>`

src/core/sharded_hash_map.h
```diff
@@ -0,0 +1,194 @@
...
+    }
+    return total;
+  }
+
+ private:
+  // Aligned to cache line.
+  struct alignas(64) Shard {
+    util::fb2::Mutex write_mu_;
+    mutable util::fb2::SharedMutex read_mu_;
+    absl::flat_hash_map<K, V> map_;
```
glevkovich: you allow building the template with a template parameter typename Hash = absl::Hash<K>, but you never pass it here. If someone will pass a costume non-default Hash (e.g  MyKeyHash) , I think it won't compile.
I suggest to fix  it to:
` absl::flat_hash_map<K, V, Hash> map_`

Important Note: Consider adding a 5th Eq parameter if you intend to support heterogeneous lookups, that means if we want to compare std::string vs std::string_view, we will be able to find the correct shard but for search in the map compile will fail or silently fall back to  full scan since the equality comparator doesn't support the heterogeneous comparison.

src/core/sharded_hash_map.h
```diff
@@ -0,0 +1,194 @@
...
+  //     on read_mu_ and a mutable Map& reference. Mutations must go through LockedMap::map
+  //     only — this ensures no reader observes a partial update.
+  //
+  // Typical usage pattern:
+  //   map.Mutate(key, [&](const auto& m, auto lock_readers) {
+  //       /* optionally inspect m (const) without blocking readers */
+  //       auto lm = lock_readers();
+  //       lm.map[key] = new_value;  // now no reader sees a partial update
+  //   });
+  template <typename F> void Mutate(const K& key, F&& f) {
```
glevkovich: This comment is for both Mutate(This one and the one in line 137):
According to this documentation I can write this code for example:

```
map_.Mutate(key, [&](const auto& m, auto lock_readers) {
    auto lm1 = lock_readers();   // acquires read_mu_ exclusively
    lm1.map[key] = 1;
    auto lm2 = lock_readers();   // tries to acquire read_mu_ again ->  deadlock
    lm2.map[key] = 2;
});
```
But this is still OK I think?:
```
map_.Mutate(key, [&](const auto& m, auto lock_readers) {
    {
        auto lm1 = lock_readers();   // acquires read_mu_ exclusively
        lm1.map[key] = 1;
    }
    auto lm2 = lock_readers();   // tries to acquire read_mu_ again ->  deadlock
    lm2.map[key] = 2;
});
```
So I think you need to make it clearer in both function documentation and add:
`"Do not hold multiple LockedMap instances simultaneously within the callback - read_mu_ is non-recursive, so acquiring it twice will deadlock. Calling lock_readers() more than once is safe only if the previous LockedMap has gone out of scope first."`

src/core/sharded_hash_map.h
```diff
@@ -0,0 +1,194 @@
...
+    Shard& shard = shards_[Hash{}(key) % NUM_SHARDS];
```
glevkovich: you have multiple places using this expression:
`Hash{}(key) % NUM_SHARDS`

This is one of them.
You can replace them all to use the common function `ShardOf`. for example,here:
`Shard& shard = shards_[ShardOf(key)];`

src/core/sharded_hash_map.h
```diff
@@ -0,0 +1,194 @@
...
+  template <typename F> void ForEachShared(F&& f) const {
+    for (const Shard& shard : shards_) {
+      std::shared_lock read_lock(shard.read_mu_);
+      for (const auto& [k, v] : shard.map_) {
+        f(k, v);
+      }
+    }
+  }
+
+  // Iterates over all entries with full exclusive access, invoking f(K&, V&) for each.
```
glevkovich: nit:
f(K&, V&) 
should be
f(const K&, V&)   // K is not mutable

src/core/sharded_hash_map_test.cc
```diff
@@ -0,0 +1,258 @@
+// Copyright 2026, DragonflyDB authors.  All rights reserved.
```
glevkovich, line 1: Geenral comment about coverage:
There is no test that checks for heterogeneous lookup (If for example, I pass string_view or a C-string).
I looked at all calls to FindIf and you always pass std::string().

Example test (passing a string_view):

```
TEST_F(ShardedHashMapTest, HeterogeneousLookup) {
  map_.Mutate(string("hello"), [](const auto& m, auto lr) {
    auto lm = lr();
    lm.map["hello"] = 7;
  });
  string_view sv = "hello";
  bool found = map_.FindIf(sv, [](const int& v) { EXPECT_EQ(v, 7); });
  EXPECT_TRUE(found);
}
```

src/core/sharded_hash_map.h
```diff
@@ -0,0 +1,194 @@
...
+// Re-entrancy: callbacks passed to FindIf, ForEachShared, ForEachExclusive, and
+// WithReadExclusiveLock are invoked while one or more shard locks are held. Calling any
+// ShardedHashMap method that would re-acquire the same lock on the same shard from within
+// a callback will deadlock.
+//
+template <typename K, typename V, size_t NUM_SHARDS = 32, typename Hash = absl::Hash<K>>
+class ShardedHashMap {
+  static_assert(NUM_SHARDS > 0, "NUM_SHARDS must be greater than 0");
+  using Map = absl::flat_hash_map<K, V>;
+
```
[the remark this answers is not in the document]
glevkovich: I actually commented about it in the other place (internal map).
see:
https://github.com/dragonflydb/dragonfly/pull/7196/changes#r3130061876

src/core/sharded_hash_map.h
```diff
@@ -0,0 +1,206 @@
...
+    std::unique_lock<util::fb2::SharedMutex> lock;
+    InternalMap& map;
+  };
+
+  // Looks up `key` under a shared read lock on its shard. If found, invokes f(const V&)
+  // with the mapped value while still holding the lock, then returns true.
+  // Returns false if the key is not present. The callback must not modify the value.
+  //
+  // The template parameter Q allows heterogeneous lookup — any type hashable via
+  // absl::Hash<Q> and comparable against K can be used (e.g., std::string_view for
```
glevkovich: nit: this is not fully accurate, only if the default is used.
change:
"any type hashable via absl::Hash<Q>" 
to
"any type hashable via Hash" since user may supply it's costume hash which != absl::Hash<Q>

src/core/sharded_hash_map.h
```diff
@@ -0,0 +1,206 @@
...
+  // non-recursive, so acquiring it twice will deadlock. Calling lock_readers() more than once
+  // is safe only if the previous LockedMap has gone out of scope first.
+  //
+  // Typical usage pattern:
+  //   map.Mutate(key, [&](const auto& m, auto lock_readers) {
+  //       /* optionally inspect m (const) without blocking readers */
+  //       auto lm = lock_readers();
+  //       lm.map[key] = new_value;  // now no reader sees a partial update
+  //   });
+  template <typename F> void Mutate(const K& key, F&& f) {
```
glevkovich: I f I initiate ShardedHashMap with K = size_t, we will get compilation failure with ambiguity. Since we will get two Mutate functions:
```
void Mutate(const size_t& key, ...)
void Mutate(size_t sid, ...)
```
now I call `Mutate(size_t{5}, ...) `and you get compile error.

Same apply for `WithReadExclusiveLock`.

I think you can open a separate task for it:
1.  make the shard-index overload take a distinct tag type, e.g. Mutate(ShardId{5}, ...)
2.  or by using std::enable_if ( to disable the key overload when K  is convertible to size_t)

src/core/sharded_hash_map.h
```diff
@@ -0,0 +1,206 @@
...
+  template <typename Q> size_t ShardOf(const Q& key) const {
+    return Hash{}(key) % NUM_SHARDS;
+  }
+
+  // Acquires read_mu_ exclusively on the shard that owns `key`, blocking all concurrent
+  // readers (FindIf, ForEachShared, Size) on that shard, then invokes f(). The write_mu_
+  // is NOT acquired, so this does not serialize against other writers. Use this when you
+  // need to perform an external side-effect that must not race with readers of this shard
+  // but the map itself is not being modified.
+  template <typename F> void WithReadExclusiveLock(const K& key, F&& f) {
```
glevkovich: I saw that you added heterogeneous lookup to FindIf,  but I think we should apply this same logic to Mutate and WithReadExclusiveLock.

Right now, both functions take const K& key. If K is std::string and a caller passes (for example) a std::string_view, it forces an unnecessary string heap allocation just to compute ShardOf(key).

Suggestion: We can fix this by templating them on Q exactly like FindIf, this allows  to find the right shard and acquire the lock with zero allocations!

```
template <typename Q, typename F> 
void Mutate(const Q& key, F&& f) { ... }

template <typename Q, typename F> 
void WithReadExclusiveLock(const Q& key, F&& f) { ... }
```

glevkovich (dismissed):
Approved with 3 comments.

glevkovich (approved):
LGTM

glevkovich (approved):
LGTM

PR #7197, opened 2026-04-21
chore: add logs to detect OOM during replication restart process

BorysTheDev:
Summary: Adds INFO-level RSS logging around the replication restart/flush (“Drakarys”) flow to help diagnose OOM conditions.
Changes: Logs RSS at RDB FULLSYNC end, before/after per-shard decommit, and around post-load search index rebuild dispatch.

### DIFF
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -5,6 +5,7 @@
 #include "server/db_slice.h"
 
 #include "core/dense_set.h"
+#include "strings/human_readable.h"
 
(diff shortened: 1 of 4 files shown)


dranikpg (approved):
maybe vlog(1) instead? I don't think we want to see it everyhwere

BorysTheDev:
> maybe vlog(1) instead? I don't think we want to see it everyhwere

I missed your comment, but we will not see vlog without activation

PR #7205, opened 2026-04-22
feat: add memory consumption for HNSW

### LINKED ISSUE #7110: HNSW memory tracking
It looks like we don't track HNSW memory for now, so we need to include HNSW memory usage in dragonfly_memory_used_bytes metric and check that it is also included in search_used class of dragonfly_memory_by_class_bytes

BorysTheDev:
fixes: #7110

Summary: This PR adds approximate memory accounting for HNSW vector indices so their footprint is reflected in Dragonfly’s memory metrics.

Changes:

- Adds a cached, lock-free memory_size_ counter and memorySize() accessor in HierarchicalNSW, updated on construction and resizeIndex()
- Exposes per-index usage via HnswVectorIndex::GetMemoryUsage()
- Aggregates all registered HNSW index memory via GlobalHnswIndexRegistry::GetTotalMemoryUsage()
- Plumbs the aggregate into ServerFamily::GetMetrics() (behind WITH_SEARCH) for both search_used and overall heap gauges
- Adds a unit test (MemoryTrackingHnsw) to validate HNSW creation increases both metrics and that heap usage is at least the search usage delta

Technical Notes: The accounting is intentionally approximate and uses relaxed atomics to avoid locking in metrics paths.

### DIFF
--- a/src/core/search/hnsw_alg.h
+++ b/src/core/search/hnsw_alg.h
@@ -716,11 +716,15 @@ template <typename dist_t> class HierarchicalNSW : public hnswlib::AlgorithmInte
   // Recomputes memory_size_ from the current allocation-defining fields.
   // Must be called whenever max_elements_ changes.
   void updateMemorySize() {
-    size_t per_element = size_data_per_element_ + sizeof(char*) + sizeof(int);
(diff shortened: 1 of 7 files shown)


BorysTheDev:
augment review

dranikpg (approved):

PR #7210, opened 2026-04-23
feat: add stricter cluster config validation

### LINKED ISSUE #7195: improve cluster config validation
We need to check next:

- All slots are presented in the config
- migration slots should correspond to owned slots
- no slots intersections between shards

BorysTheDev:
fixes: #7195

Summary: This PR tightens cluster configuration validation, focusing on slot-migration entries, to reject malformed or inconsistent configs earlier.

Changes:

- Collects all shard master IDs during IsConfigValid and validates migrations against them.
- Rejects migrations that target the source master (self-migration) or a node ID that is not a shard master in the provided config.
- Validates each migration slot range is in-bounds/ordered (IsValid()) and is fully contained within the source shard’s owned slot ranges.
- Rejects overlapping migration slot ranges within a shard (across and within migration entries).
- Adds helpers: SlotRange::Overlaps and SlotRanges::Contains(SlotRange) to support the new checks.
- Adds a parameterized test suite covering valid/invalid migration configurations and updates existing config-comparison/health tests to conform to the stricter rules.

Technical Notes: Validation now enforces stronger invariants on migration metadata (target identity and range correctness), preventing invalid configs from being accepted and later failing during migration execution.

### DIFF
--- a/src/server/cluster/cluster_config.cc
+++ b/src/server/cluster/cluster_config.cc
@@ -53,6 +53,11 @@ bool IsConfigValid(const ClusterShardInfos& new_config) {
     return false;
   }
 
+  absl::flat_hash_set<string_view> master_ids;
+  for (const auto& shard : new_config) {
+    master_ids.insert(shard.master.id);
+  }
+
   for (const auto& shard : new_config) {
     for (const auto& slot_range : shard.slot_ranges) {
       if (slot_range.start > slot_range.end) {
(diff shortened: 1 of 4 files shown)


src/server/cluster/cluster_config.cc
```diff
@@ -77,6 +82,55 @@ bool IsConfigValid(const ClusterShardInfos& new_config) {
...
+          LOG(ERROR) << "Invalid cluster config: migration range " << slot_range.ToString()
+                     << " is not owned by shard master=" << shard.master.id;
+          return false;
+        }
+        for (const auto& prev : seen_migration_ranges) {
+          if (slot_range.Overlaps(prev)) {
+            LOG(ERROR) << "Invalid cluster config: overlapping migration ranges "
+                       << slot_range.ToString() << " and " << prev.ToString()
+                       << " in shard master=" << shard.master.id;
+            return false;
```
dranikpg, line 128: It was always quite hard to work with cluster configurations because you don't get the error returned to the client

dranikpg (approved):

src/server/cluster/cluster_config.cc
BorysTheDev: According to the design, the cluster config command can't return an error

BorysTheDev: if one of the node return error, it is unclear what to do with the other nodes

PR #7211, opened 2026-04-23
fix(server): Don't use item queue for inlined load

dranikpg:
The item queue showed up on the profile because it uses atomic which can become "cache dirty" if do have some inter-thread intercaction. Avoid atomics at all for inlined execution by using a local item struct


dranikpg:
augment review

abhijat (approved):

PR #7215, opened 2026-04-24
refactor: drop extra metadata for HNSW replication

BorysTheDev:
refactors HNSW vector-index persistence for RDB/replication by persisting less AUX metadata and deriving more state from the serialized node set.

fix potential OOB reads when a concurrent Add raised maxlevel between capture and node serialization

### DIFF
--- a/src/core/search/hnsw_index.cc
+++ b/src/core/search/hnsw_index.cc
@@ -132,35 +132,13 @@ struct HnswlibAdapter {
   HnswIndexMetadata GetMetadata() const {
     MRMWMutexLock lock(&mrmw_mutex_, MRMWMutex::LockMode::kReadLock);
     HnswIndexMetadata metadata;
-    metadata.max_elements = world_.max_elements_;
-    metadata.cur_element_count = world_.cur_element_count.load();
-    metadata.maxlevel = world_.maxlevel_;
     metadata.enterpoint_node = world_.enterpoint_node_;
     return metadata;
   }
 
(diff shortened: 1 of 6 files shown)


BorysTheDev:
augment review

BorysTheDev:
augment review

src/core/search/hnsw_index.cc
```diff
@@ -132,35 +132,13 @@ struct HnswlibAdapter {
...
-    // Runtime check for release builds to prevent silent corruption
-    if (world_.cur_element_count.load() != 0) {
-      LOG(ERROR) << "SetMetadata called on non-empty HNSW index with "
-                 << world_.cur_element_count.load() << " elements, ignoring";
-      return;
-    }
-
-    // Pre-allocate capacity based on expected element count, but don't set cur_element_count.
-    // cur_element_count will be set by RestoreFromNodes when the actual nodes are restored.
-    if (world_.max_elements_ < metadata.cur_element_count) {
```
romange, line 160: If we decide this is a hint, a lower bound - this is still useful imho and I do not understand why we should delete it. I saw huge performance benefits from preallocating huge sets, tables compared to just grow on the fly.

BorysTheDev: We don't grow on the fly, this field is useless now. because we need to check all ids and found max_id to prevent race condition. we do it independently on this field. So it is no performance impact

romange (approved):

PR #7224, opened 2026-04-26
test(connection): add a fix for FIN_WAIT_2 leak bug with BLPOP

romange:
Closing a client socket while the connection's reader fiber is parked on
    a blocking command (BLPOP) does not always wake the fiber to observe EOF.
    
 Under sustained concurrent open+close churn, a fraction of FDs linger on
 the server in CLOSE_WAIT (client in FIN_WAIT_2) until the command times
 out or the client is CLIENT KILL'd, matching the reaper-driven leak
 reported in production. valkey closes its half immediately
 on FIN even mid-BLPOP.
    
 The bug is fixed in helio by https://github.com/romange/helio/pull/580

### DIFF
--- a/tests/dragonfly/connection_test.py
+++ b/tests/dragonfly/connection_test.py
@@ -2136,3 +2136,57 @@ async def test_pubsub_pipeline_starvation(df_server: DflyInstance):
         await flood_task
         writer.close()
         await writer.wait_closed()
+
+
+async def test_blocking_command_close_eof(df_server: DflyInstance):
+    """Server must drop a connection that is parked on a blocking command
+    when the client closes its socket.
+
+    Without this handling, the connection's reader fiber is suspended waiting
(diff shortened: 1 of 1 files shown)


kostasrim (dismissed):
lgtm

tests/dragonfly/connection_test.py
```diff
@@ -2136,3 +2136,57 @@ async def flood():
...
+    a single batch is handled correctly. Plain `redis-server` closes its half
+    immediately on FIN even mid-BLPOP.
+    """
+    client = df_server.client()
+    baseline = int((await client.info("clients"))["connected_clients"])
+
+    workers = 8
+    conns_per_cycle = 25
+    cycles = 4
+    park_seconds = 0.2
```
kostasrim:
```suggestion
```

kostasrim (approved):

PR #7225, opened 2026-04-26
fix: avoid O(n) scans in NotifyWatchQueue

kostasrim:
Each queue maps to a single key. If the first iteration of `IsValid()` is false, then there is no reason to iterate over the rest of the O(n-1) items of the queue.

This avalanches under heavy load for sidekiq workloads as these heavily rely on blocking pop's from multiple connections (thousands).

### DIFF
--- a/src/server/blocking_controller.cc
+++ b/src/server/blocking_controller.cc
@@ -9,6 +9,7 @@
 #include <boost/smart_ptr/intrusive_ptr.hpp>
 
 #include "base/logging.h"
+#include "server/db_slice.h"
 #include "server/engine_shard_set.h"
 #include "server/namespaces.h"
 #include "server/transaction.h"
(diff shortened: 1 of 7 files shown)


src/server/blocking_controller.cc
```diff
@@ -227,13 +227,18 @@ void BlockingController::NotifyWatchQueue(std::string_view key, WatchQueue* wq,
   auto& queue = wq->items;
   ShardId sid = owner_->shard_id();
 
+  // key_ready_checker is key-level (same result for all items in this queue).
+  // Check once upfront to avoid an O(N) scan when the key no longer exists.
+  if (!queue.front().key_ready_checker(owner_, context, key))
```
kostasrim: Uh this is required for streams, I will split the paths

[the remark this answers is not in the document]
kostasrim: Very good! I did not see this comment before I sent a fix but very good!

BorysTheDev (dismissed):

BorysTheDev:
In my test, core usage was decreased by 4 times, and perf shows absoluteny another top. So I think the fix is correct

romange:
I did not understand the scenario as the reproduction example is not part of this PR. Why NotifyPending is being called on empty key? how this can happen @BorysTheDev  and @kostasrim ?  Can we not add an empty key in the first place?

BorysTheDev:
@romange Blocking commands can be run on the key that doesn't exist, when we have a queue of blocking commands on such a key. When we do LPUSH for example, only one should be unblocked, and key is empty after that, so we don't need to try unblock others.
Reproduction scenario is as follows: sustain 10000 blocking read commands in queue, and do 50 write commands per second.

src/server/list_family.cc
```diff
@@ -1016,9 +1019,12 @@ OpResult<string> BPopPusher::RunPair(time_point tp, Transaction* tx, ConnectionC
...
                               std::string_view key) -> KeyReadyResult {
-    return context.GetDbSlice(owner->shard_id()).FindReadOnly(context, key, OBJ_LIST).ok()
-               ? KeyReadyResult::kReady
-               : KeyReadyResult::kKeyNotFound;
+    auto res = context.GetDbSlice(owner->shard_id()).FindReadOnly(context, key, OBJ_LIST);
+    if (res.ok())
+      return KeyReadyResult::kReady;
+    if (res.status() == OpStatus::WRONG_TYPE)
+      return KeyReadyResult::kNotReady;
+    return KeyReadyResult::kKeyNotFound;
```
BorysTheDev: it looks the same, maybe we should move key_checker to one of the shared header?

romange (approved):

PR #7234, opened 2026-04-28
feat(facade): bound IoLoopV2 dispatch_q_ quota to prevent starvation

glevkovich:
Previously, IoLoopV2 drained dispatch_q_ with an unbounded while loop. Under a PubSub flood, this trapped the fiber in the control path, starving pipelined commands (e.g GET/SET) and causing client timeouts.

Key changes:
- Bounded dispatch: process at most FLAGS_async_dispatch_quota messages per iteration; if the quota is hit, fall through to the data path so pipeline commands get a turn. Mirrors V1's async_dispatch_quota / prefer_pipeline_execution mechanism in AsyncFiber.
- Deferred flush: the quota-hit path falls through to ParseLoop, which reaches the idle-await flush, coalescing PubSub and command replies into a single sendmsg syscall.
- Batched backpressure: pubsub_ec.notifyAll() is now called once per quota chunk instead of once per message.
- Testing: parameterized test_pubsub_pipeline_starvation for both V1 and V2 to prevent regressions.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1551,6 +1551,34 @@ void Connection::HandleMigrateRequest() {
   // loop iteration until all subscriptions are cleared.
 }
 
+bool Connection::ProcessControlMessages(uint32_t quota) {
+  uint32_t dispatched = 0;
+
+  while (!dispatch_q_.empty()) {
+    // If Quota reached: stop draining the dispatch queue and fall through to the data path.
+    if ((quota > 0) && (dispatched >= quota)) {
+      LOG_EVERY_T(INFO, 1) << "[" << id_ << "] V2 dispatch_q_ quota reached (" << dispatched << "/"
+                           << quota << "), falling through to data path";
+      return true;
+    }
+
+    auto msg = std::move(dispatch_q_.front());
+    dispatch_q_.pop_front();
+    UpdateDispatchStats(msg, false /* subtract */);
+    dispatched++;
+
+    // If a MigrationRequestMessage arrives via the dispatch queue, stop processing
+    // and let the loop iterate back to HandleMigrateRequest() at the top.
+    if (std::holds_alternative<MigrationRequestMessage>(msg.handle)) {
+      break;
+    }
+
+    std::visit(AsyncOperations{reply_builder_.get(), this}, msg.handle);
+  }
+
+  return false;
+}
+
 io::Result<size_t> Connection::HandleRecvSocket() {
   phase_ = READ_SOCKET;
   auto& conn_stats = tl_facade_stats->conn_stats;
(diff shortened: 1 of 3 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -2992,12 +2994,29 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
...
+    // - This mirrors V1's async_dispatch_quota / prefer_pipeline_execution mechanism in AsyncFiber.
     if (!dispatch_q_.empty()) {
+      uint32_t dispatched{};
+      bool quota_reached = false;
+
       while (!dispatch_q_.empty()) {
+        // Quota reached: stop draining the dispatch queue and fall through to the data path.
+        if ((async_dispatch_quota > 0) && (dispatched >= async_dispatch_quota)) {
+          quota_reached = true;
+          VLOG(1) << "[" << id_ << "] V2 dispatch_q_ quota reached (" << dispatched << "/"
```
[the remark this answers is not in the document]

tests/dragonfly/connection_test.py
```diff
@@ -715,9 +718,14 @@ async def measure(aw):
...
     # V1: aggressive squashing across dual fibers (<=2).
-    # V2: single-fiber loop flushes based on OS scheduling and batch boundaries (<=12).
+    # V2: conditional flushing defers flush when pending_input_ or io_buf_ has data,
+    # but fast synchronous commands (INCR) don't yield, so io_uring completions aren't processed
+    # mid-batch. Task 2 (Epoch Yield) will fix this by yielding before flush. Until then, V2
+    # flushes based on TCP segment boundaries (<=12).
     pipe_limit = 12 if is_v2 else 2
-    assert await measure(p.execute()) <= pipe_limit
+    pipe_flushes = await measure(p.execute())
+    logging.warning("pipe_flushes=%d pipe_limit=%d is_v2=%s", pipe_flushes, pipe_limit, is_v2)
```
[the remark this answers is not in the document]
glevkovich: fixed

tests/dragonfly/connection_test.py
```diff
@@ -729,6 +737,46 @@ async def measure(aw):
...
+    When a pipeline arrives fragmented (separate TCP segments), the server must flush
+    completed replies before sleeping on the socket. If maybe_flush() before the main
+    await fails to flush, the reply stays trapped in memory and the client hangs.
+    """
+    reader, writer = await asyncio.open_connection("127.0.0.1", df_server.port)
+
+    # Send an isolated command and wait for the server to process and sleep.
+    writer.write(b"PING\r\n")
+    await writer.drain()
+    await asyncio.sleep(0.1)
```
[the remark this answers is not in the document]
glevkovich: fixed

tests/dragonfly/connection_test.py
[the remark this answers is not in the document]
glevkovich: duplicate, fixed. dropping.

src/facade/dragonfly_connection.cc
```diff
@@ -3006,15 +3025,19 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
...
-
-      // Note: No flush needed here: the `continue` below re-enters the loop, which either
-      // hits the data path (ParseLoop flushes via ReplyBatch) or the idle-await block
-      // (Flush 1), which always flushes before sleeping.
+      }  // while
 
-      // TODO: Properly handle backpressure
       GetQueueBackpressure().pubsub_ec.notifyAll();
-      continue;
+
```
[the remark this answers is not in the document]
glevkovich: this is wrong. dropping.

src/facade/dragonfly_connection.cc
```diff
@@ -2992,12 +2994,29 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
...
 
-    // Temporary: Handle dispatch queue items (Control Path) one by one blocking command execution
+    // Handle dispatch queue items (Control Path) with a bounded quota to prevent
+    // starvation of the data path:
+    // - Under a PubSub flood, dispatch_q_ can accumulate thousands of messages.
+    // - Without a quota, the fiber would drain them all before parsing any new data from the
+    //   socket, starving GET/SET and other pipeline commands.
+    // - This mirrors V1's async_dispatch_quota / prefer_pipeline_execution mechanism in AsyncFiber.
     if (!dispatch_q_.empty()) {
+      uint32_t dispatched{};
```
romange: nit: personally I find this syntax confusing compared to explicit `{0}` or ` = 0` .

src/facade/dragonfly_connection.cc
```diff
@@ -3006,15 +3025,19 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
...
-      continue;
+
+      // Starvation prevention: if quota was hit, fall through to the data path so
+      // pipelined commands get a turn. Otherwise `continue` back to the top.
+      //
+      // No flush here: both paths reach a flush before sleeping via the idle-await block
+      // at the top of the loop, allowing PubSub and command replies to be coalesced into
+      // one sendmsg syscall.
+      if (!quota_reached) {
+        continue;
```
romange, line 3043: I do not understand the flow here.
if quota was not reached why go back and not fall through? I see continue existed before but I still do not understand what it does.

src/facade/dragonfly_connection.cc
glevkovich: Sure, I'll change it then.

src/facade/dragonfly_connection.cc
glevkovich: The continue is not just optimization, it  is needed for correctness and to keep low latency. I'll explain:

**Reason 1:**
When we are done processing here the dispatch_q, it took time, in  some cases even long time, and during that time we might have got new client messages waiting in the socket (e.g GET/SET command). Sicne we havn't called ReadPendingInput (worse-case scenario), the io_buf_ is totally empty. If we just fall through to the data path, it checks `io_buf_.InputLen() `and thinks that there is no data from the client  and skips parsing entirely (It makes a decision based on wrong info).

When we jump to the top, these are the only areas in the hot-path where we flush and read. (ignore the special case flashes further down). So we read and pull more data:
```
if (pending_input_) {
  ReadPendingInput();
}
```
Now, when the loop eventually reaches the data path, it is working with fresh, up-to-date network data.

**Reason 2:**
When we process PubSub messages, their replies accumulate in reply_builder_. Flush() only happens at the top of the loop (in the idle-await block). If we fall through, we traverse the entire data path section doing nothing useful, then loop back to the top to flush. The continue skips that dead code and reaches the flush immediately - one fewer loop iteration, lower latency.


**In short:** **We use continue to guarantee low latency and fresh socket reads when the queue is naturally empty**. We only use the fall-through as an emergency case (when quota_reached is true) to force the data path to run so it doesn't get starved by a never-ending flood of PubSub messages.

romange: Ok, then the comment above focuses only on "quota_reached" path, but it's not clear why continue is needed in the first place. I would add additional comment around "continue" to explain why it's needed. Maybe the code structured differently would provide a more natural flow but I might be wrong too and comment at least will close the gap.

src/facade/dragonfly_connection.cc
```diff
@@ -2992,12 +2994,29 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
...
-    // Temporary: Handle dispatch queue items (Control Path) one by one blocking command execution
+    // Handle dispatch queue items (Control Path) with a bounded quota to prevent
+    // starvation of the data path:
+    // - Under a PubSub flood, dispatch_q_ can accumulate thousands of messages.
+    // - Without a quota, the fiber would drain them all before parsing any new data from the
+    //   socket, starving GET/SET and other pipeline commands.
+    // - This mirrors V1's async_dispatch_quota / prefer_pipeline_execution mechanism in AsyncFiber.
     if (!dispatch_q_.empty()) {
+      uint32_t dispatched{};
+      bool quota_reached = false;
```
romange: another suggestion - consider extracting this inner while loop into a helper function, aka:
`bool quota_reached = ProcessControlCommands(async_dispatch_quota);`

glevkovich: thanks, done.

src/facade/dragonfly_connection.cc
glevkovich: ok i will.

dranikpg (approved):

PR #7235, opened 2026-04-28
fix: memory track before preemption in OpAdd

### LINKED ISSUE #7232: SIGABRT in StreamMemTracker::UpdateStreamSize during replication_test.py::test_bug_5221 under heavy memory pressure
The regression test test_bug_5221 in tests/dragonfly/replication_test.py is failing on CI. The Python test fails with a ConnectionError because the underlying Dragonfly master process is crashing with exit code -6 (SIGABRT).

**To Reproduce**
The crash occurs under aggressive memory and eviction settings while receiving a high volume of writes via SeederV2.
The Dragonfly master in the test is initialized with:

```
--proactor_threads=1 
--cache_mode=true 
--maxmemory=256mb 
--enable_heartbeat_eviction=true 
--eviction_memory_budget_threshold=0.9
```
Then, SeederV2(key_target=22000, data_size=1000) is run, which executes XADD commands heavily.

**Stack Trace**
The C++ crash log indicates the process is aborting during an XADD operation, specifically inside StreamMemTracker::UpdateStreamSize:

```
@           0xda2404  dfly::StreamMemTracker::UpdateStreamSize()
@           0xda59c4  dfly::(anonymous namespace)::OpAdd()::{lambda()#1}::operator()()
@           0xdbf734  absl::lts_20250512::cleanup_internal::Storage<...>::InvokeCallback()
@           0xdba09c  absl::lts_20250512::Cleanup<...>::~Cleanup()
@           0xda62f0  dfly::(anonymous namespace)::OpAdd()
@           0xdb2d4c  dfly::CmdXAdd()::{lambda()#1}::operator()()
@           0xdbd448  dfly::Transaction::ScheduleSingleHopT<...>()::{lambda()#1}::operator()()
```
**Expected Behavior**
The server should gracefully reject commands (e.g., return an OOM error) or successfully evict older data without crashing the entire process with a SIGABRT.

Possible Immediate Workaround for CI
To unblock the CI pipeline, the test's memory bounds can be relaxed slightly to avoid the immediate panic:

```
async def test_bug_5221(df_factory):
    master = df_factory.create(
        proactor_threads=1,
        cache_mode="true",
        maxmemory="512mb",           # Increased from 256mb
        enable_heartbeat_eviction="true",
        eviction_memory_budget_threshold=0.8, # Decreased from 0.9
    )
```
However, the underlying segmentation/abort issue in StreamMemTracker handling still needs to be resolved by the core team.

Links and logs:
https://github.com/dragonflydb/dragonfly/actions/runs/25027165583/job/73300746134
replcia logs: 
[logs.zip](https://github.com/user-attachments/files/27152185/logs.zip)

kostasrim:
`OpAdd` rewrites to the journal directly via `RecordJournal` which can preempt. On preemption, the heap size might change leading to an improper update to memory tracker after the preemption. 

The fix is simple, run the memory tracker update before the write to the journal.

Fixes  #7232

### DIFF
--- a/src/server/stream_family.cc
+++ b/src/server/stream_family.cc
@@ -1212,6 +1212,11 @@ OpResult<streamID> OpAdd(const OpArgs& op_args, string_view key, const AddOpts& 
             << " during the XADD command";
   }
 
+  // Must update before RecordJournal, which can preempt and cause evictions that change
+  // zmalloc_used_memory_tl, corrupting the diff computed by UpdateStreamSize.
+  mem_tracker.UpdateStreamSize(it->second);
+  std::move(on_exit).Cancel();
+
   if (op_args.shard->journal()) {
     std::string result_id_as_string = StreamsIdToString(result_id);
     const bool stream_is_empty = stream_inst->length == 0;
--- a/src/server/stream_family.cc
+++ b/src/server/stream_family.cc
@@ -1178,7 +1178,6 @@ OpResult<streamID> OpAdd(const OpArgs& op_args, string_view key, const AddOpts& 
   auto& it = add_res.it;
 
   StreamMemTracker mem_tracker;
-  absl::Cleanup on_exit([it, &mem_tracker]() mutable { mem_tracker.UpdateStreamSize(it->second); });
 
   if (add_res.is_new) {
     stream* s = streamNew();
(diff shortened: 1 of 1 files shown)


src/server/stream_family.cc
```diff
@@ -1212,6 +1212,11 @@ OpResult<streamID> OpAdd(const OpArgs& op_args, string_view key, const AddOpts&
             << " during the XADD command";
   }
 
+  // Must update before RecordJournal, which can preempt and cause evictions that change
+  // zmalloc_used_memory_tl, corrupting the diff computed by UpdateStreamSize.
+  mem_tracker.UpdateStreamSize(it->second);
```
romange, line 1217: nit: i find `on_exit` borderline useless in this function as it is only needed for lines 1201-1206.

romange (dismissed):

src/server/stream_family.cc
kostasrim: +1, I will take care of it

romange (approved):

PR #7236, opened 2026-04-28
Add --tls_reload_interval_secs for automatic TLS cert hot reload

oliverlii:
## Summary

Production TLS deployments rotate certificates periodically (e.g. via cert-manager, Vault, or ACME). Today Dragonfly requires a manual `CONFIG SET tls_cert_file` / `tls_key_file` / `tls true` sequence via redis-cli to pick up renewed certs. This does not scale across large fleets.

## Changes

- Add `--tls_reload_interval_secs` flag (minimum 60, default 0 = disabled)
- Add `TlsReloadScheduling()` background fiber that periodically `stat()`s the configured cert, key, and CA cert files; when any mtime changes, calls `ReconfigureTLS()` on every listener
- Zero downtime: `SSL_CTX` is refcounted, existing connections keep their sessions, only new handshakes use updated certs
- Fiber exits immediately when TLS is not enabled or interval is 0
- Wire up Init/Shutdown lifecycle (`tls_reload_fb_`, `tls_reload_done_`)

## Testing

- Integration test `test_tls_hot_reload`: starts server with TLS, overwrites cert/key files on disk with certs from a different CA, waits for reload timer, verifies old CA rejected, new CA accepted, and existing connection survives

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -159,6 +159,9 @@ ABSL_FLAG(string, availability_zone, "",
           "server availability zone, used by clients to read from local-zone replicas");
 
 ABSL_FLAG(bool, keep_legacy_memory_metrics, true, "legacy metrics format");
+ABSL_FLAG(uint32_t, tls_reload_interval_secs, 0,
+          "If non-zero, periodically checks whether the TLS cert/key files on disk have changed "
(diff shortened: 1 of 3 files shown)


romange:
@oliverlii please fix the checks if you wish that submitted

oliverlii:
Hi @romange , I've updated the pr with the fix

PR #7246, opened 2026-04-30
feat: integrate OAHSet

BorysTheDev:
Summary: This PR integrates OAHSet as an optional dense backing store for kEncodingStrMap2 SETs behind a new --use_oah_set flag.

Changes:

- Adds g_use_oah_set (startup snapshot) plus VisitSet/Key helpers to dispatch between StringSet and OAHSet
- Updates SET command paths and generic iteration utilities to work with either dense-set implementation
- Extends async deletion, defrag, size/memory accounting, and SHRINK logic to support OAHSet-backed SETs
- Updates RDB save/load and journal serialization to iterate/serialize dense sets via the new dispatch helpers
- Renames OAHSet memory counters to ObjMallocUsed/SetMallocUsed and adjusts unit tests accordingly

the results of use_oah_set=true are in fake PR https://github.com/dragonflydb/dragonfly/pull/7247

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -30,6 +30,7 @@ extern "C" {
 #include "core/cms.h"
 #include "core/detail/bitpacking.h"
 #include "core/huff_coder.h"
+#include "core/oah_set.h"
 #include "core/page_usage/page_usage_stats.h"
 #include "core/qlist.h"
 #include "core/sorted_map.h"
@@ -63,10 +64,12 @@ size_t UpdateSize(size_t size, int64_t update) {
 
 inline void FreeObjSet(unsigned encoding, void* ptr, MemoryResource* mr) {
   switch (encoding) {
-    case kEncodingStrMap2: {
-      CompactObj::DeleteMR<StringSet>(ptr);
+    case kEncodingStrMap2:
+      VisitSet(ptr, [](auto* ss) {
+        using T = std::remove_pointer_t<decltype(ss)>;
+        CompactObj::DeleteMR<T>(ss);
+      });
       break;
-    }
 
     case kEncodingIntSet:
       zfree((void*)ptr);
@@ -87,10 +90,10 @@ void FreeList(unsigned encoding, void* ptr, MemoryResource* mr) {
 
 size_t MallocUsedSet(unsigned encoding, void* ptr) {
   switch (encoding) {
-    case kEncodingStrMap2: {
-      StringSet* ss = (StringSet*)ptr;
-      return ss->ObjMallocUsed() + ss->SetMallocUsed() + zmalloc_usable_size(ptr);
-    }
+    case kEncodingStrMap2:
+      return VisitSet(ptr, [ptr](auto* ss) {
+        return ss->ObjMallocUsed() + ss->SetMallocUsed() + zmalloc_usable_size(ptr);
+      });
     case kEncodingIntSet:
       return intsetBlobLen((intset*)ptr);
   }
@@ -277,12 +280,10 @@ pair<void*, bool> DefragSortedMap(detail::SortedMap* sm, PageUsage* page_usage) 
   return {sm, reallocated};
 }
 
-pair<void*, bool> DefragStrSet(StringSet* ss, PageUsage* page_usage) {
+template <typename Set> pair<void*, bool> DefragDenseSet(Set* ss, PageUsage* page_usage) {
   bool realloced = false;
-
   for (auto it = ss->begin(); it != ss->end(); ++it)
     realloced |= it.ReallocIfNeeded(page_usage);
-
   return {ss, realloced};
 }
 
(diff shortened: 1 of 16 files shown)


BorysTheDev:
augment review

BorysTheDev:
augment review

src/server/zset_family.cc
```diff
@@ -741,7 +742,8 @@ ScoredMap ScoreMapFromSet(const PrimeValue& pv, double weight, const DbContext&
   // Enable lazy member expiry before iterating dense sets so expired members
   // do not pollute the result (and so the caller can detect an emptied set).
   if (pv.Encoding() == kEncodingStrMap2) {
-    static_cast<StringSet*>(pv.RObjPtr())->set_time(MemberTimeSeconds(db_cntx.time_now_ms));
+    uint32_t t = MemberTimeSeconds(db_cntx.time_now_ms);
```
romange: Before we do such integration we should reduce leaking of StringSet implementation across codebase.
the goal is to consolidate our APIs as much as possible under CompactObject API.

for example, this can be exposed as `CompactObject::SetMemberTime(seconds)`

romange (commented):
Before we do this, please reduce code duplication around `set_time` calls across the codebase.  I really want to consolidate the logic around StringSet as much as possible.

BorysTheDev:
augment review

src/core/compact_object.cc
```diff
@@ -868,29 +864,33 @@ void CompactObj::InitRobj(CompactObjType type, unsigned encoding, void* obj) {
...
-// (kEncodingStrMap2), otherwise nullptr. Encoding() is a safe accessor that
-// returns kEncodingStrMap2 only when taglen_ == ROBJ_TAG.
-DenseSet* AsDenseSetOrNull(const CompactObj& obj) {
-  if (obj.Encoding() != kEncodingStrMap2)
-    return nullptr;
-  return static_cast<DenseSet*>(obj.RObjPtr());
+// Dispatches `fn` over the dense set/map backing this CompactObj for
+// kEncodingStrMap2. OBJ_HASH always uses StringMap (a DenseSet subclass);
+// OBJ_SET uses StringSet by default and OAHSet under --use_oah_set.
+template <typename Fn> auto DispatchSetOrMap(const CompactObj& obj, Fn&& fn) {
```
romange, line 743: why it is better?!!

src/server/db_slice.cc
```diff
@@ -254,35 +255,48 @@ unsigned PrimeEvictionPolicy::Evict(const PrimeTable::HotBuckets& eb, PrimeTable
...
+// OAHSet exposes it as Capacity() (BucketCount() omits displacement slots).
+template <typename Set> uint32_t ClearStepEnd(Set* s) {
+  if constexpr (std::is_same_v<Set, OAHSet>)
+    return s->Capacity();
+  else
+    return s->BucketCount();
+}
+
+template <typename Set> void AsyncDeleter::EnqueDeletion(uint32_t next, Set* ds) {
+  auto step = +[](ClearNode* n) {
```
romange, line 296: what does `+` sign means here?

romange, line 296: why step is a lambda function that you pass to ClearNode at runtime? 

Seems a very complicated way to define a member function of ClearNode

src/server/test_utils.cc
```diff
@@ -215,6 +217,7 @@ void BaseFamilyTest::SetUpTestSuite() {
 
 void BaseFamilyTest::SetUp() {
   max_memory_limit = INT_MAX;
+  g_use_oah_set = absl::GetFlag(FLAGS_use_oah_set);
```
romange, line 220: maybe just set to to True?

src/server/set_family.cc
```diff
@@ -12,10 +12,13 @@ extern "C" {
 #include "redis/util.h"  // for string2ll
 }
 
+#include <absl/flags/flag.h>
```
romange: use "base/flags"

src/server/db_slice.cc
BorysTheDev: it forces the lambda to decay into a plain function pointer

src/server/test_utils.cc
BorysTheDev: but we will lose coverage for denseSet, but in production, we use denseSet. Is it OK?

src/core/compact_object.cc
BorysTheDev: Previously, it was only DenseSet, and now it is OAHSet too. and it also covers the difference for StringSet and StringMap. If you want, I can do less changes and make something like 
`template <class Set_t> Set_t* AsSetOrNull(const CompactObj& obj);` but in this case I need to write several if else for different types

src/server/db_slice.cc
BorysTheDev: I wanted to have a universal code for OAHSet and StringSet. An alternative is to write more templates or if-else to process different types. we check OAHSet or StringSet in runtime so anyway we have this overhead. This lambda gets its type from template <typename Set> void AsyncDeleter::EnqueDeletion. and we use a pointer on function to avoid extra if else.

src/server/test_utils.cc
romange: ok, lets leave it for now

romange (dismissed):

src/server/db_slice.cc
romange: I think it's possible to do the same with template member function.

BorysTheDev:
augment review

romange (approved):

PR #7250, opened 2026-05-01
fix(search): use global stats for shard-independent text scoring

vyavdoshenko:
BM25STD/TFIDF/TFIDF.DOCNORM rankings shifted with `--proactor_threads` because each shard scored docs using its **local** IDF and avgdl. With the same dataset, top-K and scores diverged across shard counts. This PR makes scoring use stats merged across all shards so ranking is shard-count-independent.

Approach:
DFS-style two-phase query:
1. **Phase 1**: each shard walks the AST and reports per-(field, term) doc counts and per-field `total_docs_len` / `num_docs`.
2. Coordinator merges into `GlobalScoringStats` (single instance, all shards).
3. **Phase 2**: each shard runs the actual search, scoring with the merged counts so IDF/avgdl are identical regardless of shard count.

Single-shard skips phase 1 (local == global). KNN/HNSW skips it (vector distance, not text scoring).

jaccard@10 identical across shard counts confirms shard-independence.

### DIFF
--- a/src/core/search/indices.h
+++ b/src/core/search/indices.h
@@ -133,6 +133,22 @@ template <typename C> struct BaseStringIndex : public BaseIndex {
     return field_num_docs_;
   }
 
+  // Sum of per-doc field lengths. Pair with GetFieldNumDocs() to aggregate
+  // avg doc len across shards without going through the lossy ratio.
+  size_t GetFieldTotalDocsLen() const {
+    return field_total_docs_len_;
+  }
+
+  // Schema canonical identifier of this field. Set by FieldIndices after
+  // construction; empty when the index is built outside that path.
+  std::string_view field_ident() const {
+    return field_ident_;
+  }
+
+  void set_field_ident(std::string_view ident) {
+    field_ident_ = ident;
+  }
+
  protected:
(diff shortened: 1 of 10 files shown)


vyavdoshenko:
augment review

vyavdoshenko:
@copilot resolve the merge conflicts in this pull request

vyavdoshenko:
@copilot+claude-opus-4.6 resolve the merge conflicts in this pull request

dranikpg (approved):
The implementation itself in terms of code and structure looks ok to me. 

I am worried about performance. How to we compare to R and V now for scored queries with the two-hop approach? I see it happens explicitly only with a scorer supplied, but I don't know if any redis libraries supply it just so, as redis always does scoring (as far as I know)

vyavdoshenko:
> The implementation itself in terms of code and structure looks ok to me.
> 
> I am worried about performance. How to we compare to R and V now for scored queries with the two-hop approach? I see it happens explicitly only with a scorer supplied, but I don't know if any redis libraries supply it just so, as redis always does scoring (as far as I know)

The performance is good enough against R. I will update the benchmark soon. I didn't test it against V because they still don't have enough implementation.

PR #7251, opened 2026-05-01
fix: handle incompressible data in HuffmanCheckTask

shitaoxai:
HuffmanCheckTask crashes the process with SIGABRT when key data is incompressible (e.g. random hashes, UUIDs, encrypted blobs):

  F huff_coder.cc:108] Check failed: !HUF_isError(size)

The task runs as an idle task on shard 0 once key memory exceeds 50 MB. It builds a Huffman frequency table from sampled key data, then unconditionally calls HuffmanEncoder::Export() to serialize the table for logging. When the data is high-entropy (compression ratio = 1.0), HUF_writeCTable_wksp returns an error for the degenerate table and the CHECK kills the process. Since the task re-triggers on every restart after snapshot recovery, this results in a crash loop.

Note: --compression_mode does not prevent this. That flag controls snapshot serialization compression (LZ4/ZSTD), not the background HuffmanCheckTask.

Fix:
- Export() returns std::optional<std::string> instead of crashing via CHECK. Logs a warning on failure.
- HuffmanCheckTask::Run() skips Export() when compression ratio >= 1.0, since there is no useful table to log.
- All callers (debugcmd.cc, tests) updated to handle the optional.

### DIFF
--- a/src/server/engine_shard.cc
+++ b/src/server/engine_shard.cc
@@ -190,11 +190,15 @@ int32_t HuffmanCheckTask::Run(DbSlice* db_slice) {
   string error_msg;
   if (huff_enc.Build(hist_.data(), kMaxSymbol, &error_msg)) {
     size_t compressed_size = huff_enc.EstimateCompressedSize(hist_.data(), kMaxSymbol);
+    double ratio = double(compressed_size) / total_freq;
     LOG(INFO) << "Huffman table built, reducing character count from " << total_freq << " to "
-              << compressed_size << ", compression ratio " << double(compressed_size) / total_freq;
-    string bintable = huff_enc.Export();
-    LOG(INFO) << "Huffman binary table: " << absl::Base64Escape(bintable);
-    db_slice->shard_owner()->stats().huffman_tables_built++;
+              << compressed_size << ", compression ratio " << ratio;
+    if (ratio < 1.0) {
+      if (auto bintable = huff_enc.Export()) {
+        LOG(INFO) << "Huffman binary table: " << absl::Base64Escape(*bintable);
+        db_slice->shard_owner()->stats().huffman_tables_built++;
+      }
+    }
   } else {
     LOG(WARNING) << "Huffman build failed: " << error_msg;
   }
--- a/src/core/compact_object_test.cc
+++ b/src/core/compact_object_test.cc
@@ -894,10 +894,11 @@ static void BuildEncoderAB(HuffmanEncoder* encoder) {
 TEST_F(CompactObjectTest, Huffman) {
   HuffmanEncoder encoder;
   BuildEncoderAB(&encoder);
-  string bindata = encoder.Export();
+  auto bindata = encoder.Export();
+  ASSERT_TRUE(bindata.has_value());
 
   for (CompactObj::HuffmanDomain domain : {CompactObj::HUFF_KEYS, CompactObj::HUFF_STRING_VALUES}) {
-    ASSERT_TRUE(CompactObj::InitHuffmanThreadLocal(domain, bindata));
+    ASSERT_TRUE(CompactObj::InitHuffmanThreadLocal(domain, *bindata));
     for (unsigned i = 30; i < 2048; i += 10) {
       string data(i, 'a');
 
--- a/src/core/dfly_core_test.cc
+++ b/src/core/dfly_core_test.cc
@@ -355,8 +355,9 @@ TEST_F(HuffCoderTest, Decode) {
   EXPECT_EQ(2, encoder_.GetNBits('a'));
   EXPECT_EQ(3, encoder_.GetNBits('b'));
 
-  string bindata = encoder_.Export();
-  ASSERT_TRUE(decoder_.Load(bindata, &error_msg_)) << error_msg_;
+  auto bindata = encoder_.Export();
+  ASSERT_TRUE(bindata.has_value());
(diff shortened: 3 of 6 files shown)


vyavdoshenko:
Hi @shitaoxai 
Thanks for your contribution. Could you please share a scenario of how we can reproduce this sigabrt? I mean, how can I crash the Dragonfly using redis-cli?
Also, make sure all your commits are properly signed to make further merges possible.
https://github.com/dragonflydb/dragonfly/blob/main/CONTRIBUTING.md

shitaoxai:
Thanks for your review @vyavdoshenko. Signed the PR. 



Also to reproduce the issue, you can run this in 1.38.1:

```
for p in $(seq 1 10); do
  (for i in $(seq 1 10000); do printf '*3\r\n$3\r\nSET\r\n$512\r\n'; head -c 512 /dev/urandom; printf '\r\n$1\r\nx\r\n'; done) | redis-cli --pipe &
done
```

And once it cross 50MB

```
redis-cli INFO memory | grep used_memory_human
used_memory_human:54.67MiB
```

it will shortly crash with this error:

```
F20260501 17:35:20.707816    12 huff_coder.cc:108] Check failed: !HUF_isError(size)
*** Check failure stack trace: ***
    @          0x14d026a  google::LogMessage::SendToLog()
    @          0x14c7f0f  google::LogMessage::Flush()
    @          0x14c8bd9  google::LogMessageFatal::~LogMessageFatal()
    @           0x9f57ff  dfly::HuffmanEncoder::Export()
    @           0x6efcf4  dfly::(anonymous namespace)::HuffmanCheckTask::Run()
    @          0x1090eca  util::fb2::ProactorBase::RunOnIdleTasks()
    @          0x10d9e2b  util::fb2::UringProactor::MainLoop()
    @          0x109a54e  _ZN5boost7context6detail11fiber_entryINS1_12fiber_recordINS0_5fiberENS0_21basic_fixedsize_stackINS0_12stack_traitsEEEZN4util3fb26detail12_GLOBAL__N_114DispatcherImplC4ERKNS0_12preallocatedEOS7_PNSA_9SchedulerEEUlOS4_E_EEEEvNS1_10transfer_tE
    @          0x10f3cdf  make_fcontext
*** SIGABRT received at time=1777656920 on cpu 0 ***
PC: @     0x7187eeeaa9fc  (unknown)  pthread_kill
```

Let me know if this works.

vyavdoshenko:
@shitaoxai 
Yes, your example is reproducible. However, your commit is still unverified. Please make sure you have passed all required steps.
https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

shitaoxai:
Thanks! @vyavdoshenko 

Should be verified now.  Thanks for the detailed instructions.

vyavdoshenko:
@shitaoxai 
In my opinion, there are two issues:
- There is no regression test as you mentioned above. I would like to see the exact same scenario as you described.
- This fix looks like a fix of a symptom, not a root cause.

@romange, could you please take a look? Is this fix acceptable?

romange:
@shitaoxai  thanks you for the contribution. As you can reproduce it easily, it should be rather easy to add a test to dfly_core_test.cc that covers this scenario. I probably misunderstood HUF_xxx API, so I would like to make sure we catch the error as early as possible. The unit test will help me with this.

shitaoxai:
@vyavdoshenko agree that this is fix of a symptom not the root cause

@romange i added a regression test to reproduce the issue using huff lib (IncompressibleUniformHistogram) 
the issue happens when it attempt to compress uniformly distributed keys

src/server/engine_shard.cc
```diff
@@ -190,10 +190,14 @@ int32_t HuffmanCheckTask::Run(DbSlice* db_slice) {
...
-              << compressed_size << ", compression ratio " << double(compressed_size) / total_freq;
-    string bintable = huff_enc.Export();
-    LOG(INFO) << "Huffman binary table: " << absl::Base64Escape(bintable);
+              << compressed_size << ", compression ratio " << ratio;
+    if (ratio < 1.0) {
+      if (auto bintable = huff_enc.Export()) {
+        LOG(INFO) << "Huffman binary table: " << absl::Base64Escape(*bintable);
+      }
+    }
     db_slice->shard_owner()->stats().huffman_tables_built++;
```
romange: Please move this increment into the inner block above together with `LOG(INFO)`

romange (commented):
LGTM, with a minor comment.

shitaoxai:
> LGTM, with a minor comment.

Thanks! Done

romange (approved):

shitaoxai:
@romange can you help me merge this PR?

i am not authorized to merge

```Merging is blocked
You're not authorized to push to this branch. Visit https://docs.github.com/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches for more information.
```

vyavdoshenko:
> @romange can you help me merge this PR?
> 
> i am not authorized to merge
> 
> ```
> You're not authorized to push to this branch. Visit https://docs.github.com/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches for more information.
> ```

Don't worry. We will merge it after the CI builds finish.
Thanks for your contribution.

shitaoxai:
@vyavdoshenko 

Got it thanks! 

BTW great infrastructure!

PR #7254, opened 2026-05-03
server: defrag 2

abhijat:


abhijat:
afl error is from build failure, not fuzzer

https://github.com/dragonflydb/dragonfly/actions/runs/25446794970/job/74652993045

abhijat:
real AFL crash
https://github.com/dragonflydb/dragonfly/actions/runs/25483772185/job/74774075578?pr=7254

E: cannot reproduce, ran 1000 times in loop but did not crash

PR #7257, opened 2026-05-03
fix: single-field hash should use listpack even with large values

### LINKED ISSUE #7249: a hashmap with a single field should be listpack
even if the value is large.

currently we do not do that but for a single field map - it's better to have a listpack.
we can still have a sanity check on a key length so it won't be too long.

romange:
Single-field hashes now stay as listpack even when the value exceeds
`max_map_field_len`, fixing issue #7249.

Previously `IsGoodForListpack` rejected any field or value larger than
`max_map_field_len` (64 bytes), causing a one-field hash to be
immediately promoted to a dense_set on creation. A single field is
always efficient to look up or mutate in a contiguous listpack blob.

The bypass is limited to listpacks that currently have 0 or 1
field-value pair (`lpLength <= 2`) so that the normal size-based
promotion still fires when fields are added one at a time to a
growing hash.

## Changes
- `IsGoodForListpack`: return `true` for a single field-value pair
  (`args.size() == 2`, field name < 4 KiB) when the listpack holds at
  most one existing entry
- `HSetFamilyTest.SingleFieldLargeValueRemainsListpack`: verifies that
  a one-field hash with a 200-byte value stays as `listpack`, and that
  a two-field hash with the same large value converts to `dense_set`

Fixes #7249

### DIFF
--- a/src/server/hset_family.cc
+++ b/src/server/hset_family.cc
@@ -46,7 +46,28 @@ using IncrByParam = std::variant<double, int64_t>;
 using OptStr = std::optional<std::string>;
 enum GetAllMode : uint8_t { FIELDS = 1, VALUES = 2 };
 
+// TODO: replace all the listpack code with our detail::Listpack wrapper.
 bool IsGoodForListpack(CmdArgList args, const uint8_t* lp) {
+  DCHECK_GE(args.size(), 2u);
+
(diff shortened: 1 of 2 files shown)


romange:
This will also automatically help ssd-tiering to serialize hashmaps with 16KB single len values. 
In fact I would like to test it.

dranikpg (approved):

PR #7269, opened 2026-05-06
Add memtier-compatible latency JSON to dfly_bench

romange:
## Summary

- Add `--json_out_file` flag to `dfly_bench` that writes a memtier-compatible latency JSON report on completion
- Track GET and SET latencies separately in aggregate and in 1-second time-series buckets
- Output schema matches `memtier_benchmark` so existing plotting scripts work without modification

## Changes

- `BenchOp` enum tags each in-flight request as GET or SET
- `SampleStats` struct pairs a `Histogram` with a count per operation type
- `ClientStats` gains `op_stats[2]` and a `time_series` map bucketed by elapsed seconds from the benchmark epoch
- `benchmark_start_ns` is captured once and threaded through `TLocalClient::Start` → `Driver::Run` so all fibers share the same epoch
- `WriteLatencyJsonReport` serialises aggregate and per-second sections in the memtier JSON schema
- Unsupported combinations (`--command`, `--noreply`, `--connect_only` with `--json_out_file`) are rejected at startup with `CHECK`

### DIFF
--- a/src/server/dfly_bench.cc
+++ b/src/server/dfly_bench.cc
@@ -13,8 +13,11 @@ extern "C" {
 #include <absl/strings/str_format.h>
 #include <absl/strings/str_split.h>
 
+#include <array>
 #include <boost/icl/interval_set.hpp>
 #include <csignal>
+#include <fstream>
+#include <map>
 #include <queue>
 #include <shared_mutex>
 #include <tuple>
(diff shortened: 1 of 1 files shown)


src/server/dfly_bench.cc
```diff
@@ -1189,6 +1436,14 @@ int main(int argc, char* argv[]) {
   }
 
   string dist = GetFlag(FLAGS_key_dist);
+  string json_out_file = GetFlag(FLAGS_json_out_file);
+
+  if (!json_out_file.empty()) {
+    CHECK(GetFlag(FLAGS_command).empty())
+        << "--json_out_file is supported only with the built-in GET/SET workload";
+    CHECK(!GetFlag(FLAGS_noreply)) << "--json_out_file is not supported together with --noreply";
+    CHECK(!GetFlag(FLAGS_connect_only)) << "--json_out_file requires sending benchmark commands";
```
vyavdoshenko, line 1445: I guess `LOG(ERROR) << "...";` would be better.

vyavdoshenko (approved):
lgtm

PR #7306, opened 2026-05-14
feat: integrate SIMD into OAHSet

BorysTheDev:
Added a wrapper for the vector compiler extension that was integrated into the OAHSet::Add() method.
Add a regression test ensuring re-adding keys never creates duplicates across rehashes and vector overflows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,6 +120,7 @@ include(FetchContent)
 FetchContent_Declare(
   eve
   URL https://github.com/jfalcou/eve/archive/refs/tags/v2023.02.15.tar.gz
+  URL_HASH SHA256=7a5fb59c0e6ef3bef3e8b36d62e138d31e7f2a9f1bdfe95a8e96512b207f84c5
 )
 FetchContent_GetProperties(eve)
 if (NOT eve_POPULATED)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -109,6 +109,28 @@ endif()
 
 include(third_party)
 include(internal)
+# helio's internal.cmake sets CMAKE_CXX_STANDARD=17 for its own legacy targets;
+# dragonfly itself uses C++20 (operator<=>, std::contiguous_iterator, concepts),
+# so restore the C++20 standard before configuring our own subdirectories.
+set(CMAKE_CXX_STANDARD 20)
+
+# Fetch the EVE (Expressive Vector Engine) SIMD library used by OAHSet.
+# Mirrors the FetchContent pattern in helio/cmake/third_party.cmake.
+include(FetchContent)
+FetchContent_Declare(
+  eve
+  URL https://github.com/jfalcou/eve/archive/refs/tags/v2023.02.15.tar.gz
+)
+FetchContent_GetProperties(eve)
+if (NOT eve_POPULATED)
+  FetchContent_Populate(eve)
+  set(EVE_BUILD_TEST OFF CACHE BOOL "")
+  set(EVE_BUILD_DOCUMENTATION OFF CACHE BOOL "")
+  set(EVE_BUILD_BENCHMARKS OFF CACHE BOOL "")
+  set(EVE_BUILD_INTEGRATION OFF CACHE BOOL "")
+  set(EVE_BUILD_RANDOM OFF CACHE BOOL "")
+  add_subdirectory(${eve_SOURCE_DIR} ${eve_BINARY_DIR})
+endif()
 
 include_directories(src)
 include_directories(helio)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -109,29 +109,6 @@ endif()
 
 include(third_party)
 include(internal)
-# helio's internal.cmake sets CMAKE_CXX_STANDARD=17 for its own legacy targets;
-# dragonfly itself uses C++20 (operator<=>, std::contiguous_iterator, concepts),
-# so restore the C++20 standard before configuring our own subdirectories.
-set(CMAKE_CXX_STANDARD 20)
-
-# Fetch the EVE (Expressive Vector Engine) SIMD library used by OAHSet.
-# Mirrors the FetchContent pattern in helio/cmake/third_party.cmake.
-include(FetchContent)
-FetchContent_Declare(
-  eve
-  URL https://github.com/jfalcou/eve/archive/refs/tags/v2023.02.15.tar.gz
-  URL_HASH SHA256=7a5fb59c0e6ef3bef3e8b36d62e138d31e7f2a9f1bdfe95a8e96512b207f84c5
-)
-FetchContent_GetProperties(eve)
-if (NOT eve_POPULATED)
-  FetchContent_Populate(eve)
-  set(EVE_BUILD_TEST OFF CACHE BOOL "")
-  set(EVE_BUILD_DOCUMENTATION OFF CACHE BOOL "")
-  set(EVE_BUILD_BENCHMARKS OFF CACHE BOOL "")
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -109,6 +109,29 @@ endif()
 
 include(third_party)
 include(internal)
+# helio's internal.cmake sets CMAKE_CXX_STANDARD=17 for its own legacy targets;
+# dragonfly itself uses C++20 (operator<=>, std::contiguous_iterator, concepts),
+# so restore the C++20 standard before configuring our own subdirectories.
+set(CMAKE_CXX_STANDARD 20)
```
romange: And I would like to avoid this as well. Or at least to take it slow at my own pace.

CMakeLists.txt
```diff
@@ -109,6 +109,29 @@ endif()
...
+
+# Fetch the EVE (Expressive Vector Engine) SIMD library used by OAHSet.
+# Mirrors the FetchContent pattern in helio/cmake/third_party.cmake.
+include(FetchContent)
+FetchContent_Declare(
+  eve
+  URL https://github.com/jfalcou/eve/archive/refs/tags/v2023.02.15.tar.gz
```
romange: I looked at it and I would like to avoid it.

1. It's a research project as they say themselves.
2. We already have `src/core/sse_port.h` which includes low-level instrinics for both x64 and arm64.
   I know that this is not the same but is it really hard to implement a few high level routines that you use via instrinics directly?

src/core/oah_set.h
```diff
@@ -8,7 +8,9 @@
 #include <absl/random/random.h>
 #include <absl/types/span.h>
 
+#include <bit>
 #include <concepts>
+#include <eve/eve.hpp>
```
romange: lets take more conservative approach and implement whatever semantics you need via self-contained code with low-level intrinsics.

CMakeLists.txt
BorysTheDev: it is not hard, I even can do quite similar optimizations without intrinsic using compiler optimization logic and some compiler builtin functions. It will take some time and maybe be not so efficient, but quite OK.

romange: so the whole range of compiler instrinsics + sse/port include is to your disposal.

BorysTheDev:
augment review

src/core/simd_op.h
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  SimdOp operator==(T value) const {  // NOLINT
+    return Vec(v_ == (Vec{} + value));
+  }
+
+  // The architecture branches exist because no portable formulation lowers to
+  // a single movemask instruction — every plain vector-extension version we
+  // tried measured ~5% slower on OAHSet's hot path.
+  BitsType ToBits() const {
```
romange: it's not clear to me what it does.
if it only extracts LSB bits from the words, it's not a good name. Moreover it should have explanation here, and not in the class description.
The comments about the implementation should be in the function body, not before its declaration.

you mention LSB above but https://doc.rust-lang.org/stable/core/arch/x86_64/fn._mm256_movemask_pd.html

says it takes the MSB.

src/core/simd_op.h
```diff
@@ -0,0 +1,115 @@
...
+
+  using Vec __attribute__((vector_size(sizeof(T) * N))) = T;
+
+ public:
+  using BitsType = std::uint32_t;
+  static constexpr std::size_t kLanes = N;
+
+  SimdOp() = default;
+
+  // Broadcasting via `Vec{} + value` lowers to vpbroadcast on AVX2 / dup on
```
romange: here as well. you talk about implementation but please explain what Broadcast does.

romange (commented):
lgtm with minor comments

src/core/simd_op.h
BorysTheDev: Broadcast means to set one value to several elements in an array; it is standard SIMD terminology

src/core/simd_op.h
BorysTheDev: LSB is for a bitmask, because we store one bit for every element from 0 to 31. I will update comments

src/core/simd_op.h
BorysTheDev: renamed to Fill

src/core/simd_op.h
BorysTheDev: renamed and updated comments

src/core/simd_op.h
```diff
@@ -0,0 +1,119 @@
...
+  // NEON; a per-lane scalar loop pessimizes to vpinsrq + vperm2i128.
+  static SimdOp Fill(T value) {
+    return Vec{} + value;
+  }
+
+  static SimdOp Load(const T* ptr) {
+    Vec v;
+    std::memcpy(&v, ptr, sizeof(Vec));
+    return v;
+  }
```
dranikpg, line 47: nit: you can use std::span or comment that it must be exactly the size of vec

src/core/simd_op.h
```diff
@@ -0,0 +1,119 @@
...
+    return Vec(v_ == o.v_);
+  }
+
+  SimdOp operator==(T value) const {  // NOLINT
+    return Vec(v_ == (Vec{} + value));
+  }
+
+  // Packs the most-significant bit of every lane into a uint32_t bitmask
+  // (LSB = lane 0). For the output of `operator==` (lanes are all-ones or
+  // all-zeros) this is equivalent to "bit i set iff lane i is non-zero".
```
dranikpg, line 75: the "most-significant" part is a bit misleading, because in "bit i set iff lane i is non-zero" there is no most significant, but I understand that it's the same

src/core/oah_set.h
```diff
@@ -136,33 +142,108 @@ class OAHSet {  // Open Addressing Hash Set
...
+      if (e.IsVector())
+        continue;
+      if (e.Key() != str) {  // after rehash, the pointer can miss hash so we need to set it for
+                             // better performance
+        if (e.GetHash() != ext_hash) {
+          e.SetExtHash(CalcExtHash(Hash(e.Key()), capacity_log_));
+        }
+        e.ExpireIfNeeded(time_now_, &size_, &obj_alloc_used_);
+        continue;
+      }
```
dranikpg, line 227: looks really clean when simd parts are hidden 👍🏻

dranikpg (dismissed):

src/core/oah_set.h
BorysTheDev: yes the algorithm is a little tricky because is built on bit magic, but it is a standard approach for SIMD so I think it should be clear

src/core/oah_set.h
```diff
@@ -248,24 +329,45 @@ class OAHSet {  // Open Addressing Hash Set
...
-        entries_[bucket_id] = std::move(e);
-        return;
+  // Walk the vector at ext_bid (vectors live only at the extension point)
+  // looking for str. Returns nullptr if not found. Returns a pointer if found:
+  // caller checks .Empty() to distinguish "live match" (return false from Add)
+  // from "match-but-expired" (slot is now vacant; reuse it for the new entry).
+  //
+  // Vectors have power-of-2 capacity with minimum 2, so we sweep in 2-lane
+  // SIMD strides.
+  OAHEntry* ProbeExtensionVector(uint32_t ext_bid, std::string_view str, uint64_t ext_hash) {
```
romange, line 339: I do not think that having those huge functions in header file helps with performance.

romange (approved):

src/core/oah_set.h
BorysTheDev: I will refactor it step by step

PR #7314, opened 2026-05-16
fix(tests): Transient 502 network errors crash test suite due to missing retries in urllib (#7238)

### LINKED ISSUE #7238: Transient 502 network errors crash test suite due to missing retries in urllib downloads
**CI failure:**
https://github.com/dragonflydb/dragonfly/actions/runs/25063061032/job/73422425244
https://github.com/dragonflydb/dragonfly/actions/runs/25101810957/job/73552721191 (similar)

### The Problem
The Dragonfly CI pipeline is experiencing transient crashes during test execution because multiple test setup scripts rely on Python's `urllib.request.urlretrieve` to download external binaries. 

Currently, these network calls have **no retry logic**. If the external server (GitHub, MinIO, etc.) experiences a momentary hiccup or proxy issue, it returns an `HTTP Error 502: Bad Gateway`, which instantly crashes the entire test suite.

We have observed this exact failure pattern across two separate incidents today:

#### Incident 1: GitHub Release Download (`replication_test.py`)
The test `test_replicate_search_index_to_old_replica` attempts to download an older version of Dragonfly directly from GitHub Releases.
* **Failure:** `urllib.error.HTTPError: HTTP Error 502: Bad Gateway or Proxy Error`
* **File:** `tests/dragonfly/replication_test.py:2665` (inside `download_dragonfly_release`)
* **Job Link:** [Action Run 25063061032](https://github.com/dragonflydb/dragonfly/actions/runs/25063061032/job/73422425244)

#### Incident 2: MinIO Binary Download (`conftest.py`)
The `pytest_configure` setup attempts to download the MinIO test binary from `dl.min.io` before running S3 snapshot tests.
* **Failure:** `urllib.error.HTTPError: HTTP Error 502: Bad Gateway or Proxy Error`
* **File:** `tests/dragonfly/conftest.py:57` (inside `_download_minio_binary`)
* **Job Link:** [Action Run 25101810957](https://github.com/dragonflydb/dragonfly/actions/runs/25101810957/job/73552721191)

### Why merge these into one ticket?
Both incidents share the exact same root cause: Fragile `urllib.request.urlretrieve` calls lacking network resiliency. Rather than playing whack-a-mole with individual test files, we should handle this as a systemic CI stability issue.

###  Proposed Solution
Instead of patching `urlretrieve` in multiple places, we should create a shared, robust download helper (e.g., in `tests/dragonfly/utility.py`) that includes a retry loop with exponential backoff. 

Example Implementation:
```python
import time
import urllib.request
import logging

def download_with_retries(url, dest, max_attempts=3):
    for attempt in range(1, max_attempts + 1):
        try:
            urllib.request.urlretrieve(url, dest)
            return
        except Exception as e:
            if attempt == max_attempts:
                logging.error(f"Download failed after {max_attempts} attempts: {url}")
                raise
            logging.warning(f"Download failed ({e}). Retrying...")
            time.sleep(2 + 2 * attempt) # Backoff

### PR DESCRIPTION
**Pull request by paarth-k2002**
Fixes a test suite flake where transient 502 Bad Gateway errors caused urllib downloads to fail immediately instead of retrying with exponential back-off.

Changes:
- Replaces urllib.request.urlretrieve with requests.Session and urllib3.Retry to handle transient 5xx errors and network failures with exponential backoff.

Closes #7238

### DIFF
--- a/tests/dragonfly/utility.py
+++ b/tests/dragonfly/utility.py
@@ -18,6 +18,10 @@ import fakeredis
 from typing import Iterable, Union
 from enum import Enum
 import re
+from shutil import copyfileobj
+from requests import Session
+from requests.adapters import HTTPAdapter
+from urllib3.util.retry import Retry
 
 
 def tmp_file_name():
@@ -37,6 +41,25 @@ def chunked(n, iterable):
 def eprint(*args, **kwargs):
     """Print to stderr"""
     print(*args, file=sys.stderr, **kwargs)
+
+
+def download_with_retries(url, dest, max_retries: int = 5) -> None:
+    """Download a file from url to dest with exponential backoff retries."""
+    adapter = HTTPAdapter(
+        max_retries=Retry(
+            total=max_retries,
+            backoff_factor=1,
+            status_forcelist=[500, 502, 503, 504],
+            raise_on_status=True,
+        )
+    )
+    with Session() as session:
+        session.mount("https://", adapter)
+        with session.get(url, stream=True) as r:
+            r.raise_for_status()
+            r.raw.decode_content = True
+            with open(dest, "wb") as f:
+                copyfileobj(r.raw, f)
 
 
 def gen_test_data(n, start=0, seed=None):
--- a/tests/dragonfly/conftest.py
+++ b/tests/dragonfly/conftest.py
@@ -25,7 +25,13 @@ from redis import asyncio as aioredis
 
 from . import PortPicker
 from .instance import DflyInstance, DflyParams, DflyInstanceFactory, RedisServer
-from .utility import DflySeederFactory, gen_ca_cert, gen_certificate, skip_if_not_in_github
+from .utility import (
+    DflySeederFactory,
+    gen_ca_cert,
+    gen_certificate,
+    skip_if_not_in_github,
+    download_with_retries,
+)
 
 logging.getLogger("asyncio").setLevel(logging.WARNING)
 # Suppress "Unclosed ClusterNode" warnings from redis-py topology refreshes (not actionable)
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: tests/dragonfly/utility.py [hunk 1]
```diff
@@ -39,6 +40,43 @@ def eprint(*args, **kwargs):
     print(*args, file=sys.stderr, **kwargs)
 
 
+def download_with_retries(url, dest, max_attempts: int = 5) -> None:
```
romange: thanks but it's not needed. If there exists a solution that we can import - we should use it.
For example,

```py
  from requests.adapters import HTTPAdapter
  from urllib3.util.retry import Retry
  import requests

  session = requests.Session()
  session.mount("https://", HTTPAdapter(max_retries=Retry(total=5, backoff_factor=1, status_forcelist=[500,502,503,504])))
  session.get(url)

```

paarth-k2002: Ah yes. Better to use HTTPAdapter and Retry directly to handle this without rolling a custom loop. I'll refactor the download logic to use this declarative approach, remove the redundant test suite, and update the PR shortly. Thanks!

tests/dragonfly/utility.py
```diff
@@ -39,6 +43,25 @@ def eprint(*args, **kwargs):
...
```
romange: rename max_attempts to max_retries

romange (commented):
Please update your commit and PR descriptions based on the latest changes

tests/dragonfly/utility.py
paarth-k2002: Done. Update the code and the PR descriptionsin accordance to the latest changes.

romange (approved):

romange:
Thanks!

PR #7315, opened 2026-05-17
fix(ci): install libheif codec plugins for HEIC/AV1 support in Mastodon tests

vyavdoshenko:
The Mastodon ruby test suite was failing daily with `heif: Unsupported feature: Unsupported codec (4.3000)` when processing a HEIC fixture file that uses AV1 encoding. On Ubuntu 24.04, libheif uses a plugin architecture - codec support is loaded dynamically from separate packages that are not automatically pulled in via `libvips42`.

https://github.com/dragonflydb/dragonfly/actions/runs/25984156368/job/76378291227
https://github.com/dragonflydb/dragonfly/actions/runs/25986392914/job/76384405378

### DIFF
--- a/.github/workflows/mastodon-ruby-tests.yml
+++ b/.github/workflows/mastodon-ruby-tests.yml
@@ -59,7 +59,7 @@ jobs:
       - name: Install pre-requisites
         run: |
           sudo apt update
-          sudo apt install -y libicu-dev libidn11-dev libvips42 ffmpeg imagemagick libpam-dev
+          sudo apt install -y libicu-dev libidn11-dev libvips42 libheif-plugin-libde265 libheif-plugin-dav1d ffmpeg imagemagick libpam-dev
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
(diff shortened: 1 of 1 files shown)


romange:
@vyavdoshenko  thanks! Do you know why it started failing now?

vyavdoshenko:
> @vyavdoshenko thanks! Do you know why it started failing now?

Most likely, Mastodon recently added a HEIC media attachment test that uses an AV1-encoded fixture file - the workflow always checks out the latest Mastodon main, so a new spec immediately becomes our problem.

vyavdoshenko:
Fixed build: https://github.com/dragonflydb/dragonfly/actions/runs/25986392914

romange:
@vyavdoshenko  can you please fix the mastodon version to some predefined tag? We should not chase their main.

vyavdoshenko:
> @vyavdoshenko can you please fix the mastodon version to some predefined tag? We should not chase their main.

done

romange (approved):

PR #7316, opened 2026-05-17
fix(server): Fix iterator invalidation

### LINKED ISSUE #7282: ` dragonfly/snapshot_test.py::test_s3_snapshot[df_factory0]` EOF during read-response on client
```
            # save to S3 + flush + load from S3 (with local --dir)
>           await async_client.execute_command("SAVE", "DF", s3_path, "snapshot")
```

error during client read response

```
        if isinstance(response, ResponseError):
>           raise response from None
E           redis.exceptions.ResponseError: end of stream
```

https://github.com/dragonflydb/dragonfly/actions/runs/25628455376/job/75227645646#step:7:7913

dranikpg:
Fixes #7282 

Since recently multiple "Flush change to earlier callback" calls can be running at the same time

https://github.com/dragonflydb/dragonfly/blob/82178599f330e9d470fd6c92a79412502865f30b/src/server/serializer_base.cc#L174-L177

It can happen that one such call finishes earlier than the others, unblocking some kind of write that caused it and allowing the write to delete the key. The other "flush" call proceeds with an invalid iterator

### DIFF
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -1378,7 +1378,10 @@ void DbSlice::FlushChangeToEarlierCallbacks(DbIndex db_ind, Iterator it, uint64_
     // does not change during the serialization, therefore we allow at most one serializer
     // reading the bucket at the same time.
     if (bucket_version <= cb_version) {
-      ccb->second(db_ind, ChangeReq{it.GetInnerIt()});
+      // Key might have been deleted because another FlushChangeToEarlierCallbacks finished
+      // and allowed a mutation to pass while this one was suspended
+      if (auto inner_it = it.GetInnerIt(); !inner_it.is_done())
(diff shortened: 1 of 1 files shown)


romange (approved):

romange:
@dranikpg how it can happen? `SerializerBase::ProcessBucket` calls `FlushChangeToEarlierCallbacks` only in traversal flow. So are we talking about two snapshots running concurrently on the same bucket?

What is the possible scenario that causes this bug?

src/server/db_slice.cc
romange, line 1355: as a side note `FlushChangeToEarlierCallbacks` is a bad name for a function that writes values

dranikpg:
> So are we talking about two snapshots running concurrently on the same bucket?

Yes

PR #7319, opened 2026-05-17
feat(search): exact phrase queries with slop and BM25 scoring

### LINKED ISSUE #7294: `FT.SEARCH`: implement exact phrase queries (quoted multi-word terms)
Problem:

`FT.SEARCH` does not support **exact phrase queries** - quoted multi-word terms that require the words to appear in the same field in the specified order at adjacent (or near-adjacent) positions.

The query parser accepts the syntax `"word1 word2"`, but the engine returns zero documents regardless of the corpus contents. Reference implementations (e.g. Redis Stack `redisearch`) honor the adjacency constraint and return the correct subset.

This breaks common RAG / IR workflows that combine free-text AND with phrase-pinned subspans:

```
@title:(machine learning) "fully convolutional network"
```

The phrase portion always returns the empty set, dragging the whole intersection to zero.

Reproduction:
Same corpus on both backends:

```
FT.CREATE idx SCHEMA t TEXT NOSTEM

HSET p:1 t "machine learning algorithm"   # "machine learning" adjacent
HSET p:2 t "learning machine works"       # words present, reversed order
HSET p:3 t "machine learning"             # "machine learning" adjacent
```

Expected (reference behavior):
```
> FT.SEARCH idx "machine learning" NOCONTENT
1) (integer) 3
2) "p:1"  3) "p:2"  4) "p:3"          # AND-match, all three

> FT.SEARCH idx "\"machine learning\"" NOCONTENT
1) (integer) 2
2) "p:1"  3) "p:3"                    # phrase: adjacency enforced; p:2 dropped
```

Dragonfly (current):
```
> FT.SEARCH idx "machine learning" NOCONTENT
1) (integer) 3
2) "p:1"  3) "p:2"  4) "p:3"          # AND-match works

> FT.SEARCH idx "\"machine learning\"" NOCONTENT
1) (integer) 0                        # phrase silently returns nothing
```

The result is consistent even with `NOSTEM`, so the gap is unrelated to stemming - it is a missing capability in the inverted index and query executor.

Real-world impact:

Frameworks that rely on phrase pinning:

- `redisvl.TextQuery` - exposes a `phrase` filter that emits quoted terms.
- `langchain` `RetrievalQAWithSourcesChain` over a vector + lexical store.
- Hybrid BM25 + KNN aggregators that include exact-phrase intents.

In all of these, the phrase clause is intersected with other conditions, so a silent zero result collapses the whole query.

vyavdoshenko:
Quoted multi-word terms in `FT.SEARCH` now match documents where the constituent tokens appear at adjacent positions (in order) within the same field - restoring parity for RAG / IR workflows that combine free-text AND with phrase-pinned subspans.

```
> FT.CREATE idx SCHEMA t TEXT NOSTEM
> HSET p:1 t "machine learning algorithm"
> HSET p:2 t "learning machine works"
> HSET p:3 t "machine learning"

> FT.SEARCH idx "machine learning"        # bare terms (AND)
1) 3) p:1, p:2, p:3

> FT.SEARCH idx "\"machine learning\""    # quoted phrase (adjacency)
1) 2) p:1, p:3
```

Behavior:
| Surface | After |
|---|---|
| `"a b"` | matches docs where `a` and `b` are adjacent in order in the same field |
| `"a b c"~N` | up to `N` intervening tokens between consecutive phrase terms (in order); `~0` ≡ adjacency |
| `@field:"a b"` | phrase restricted to a single TEXT field |
| Phrase combined with AND/OR/NOT | works as a regular sub-expression |
| Stemming (`TEXT` default) | bare terms still stem-expand; phrases match raw forms only |
| Stopwords inside the phrase | dropped at tokenize-time without advancing position |
| Single-token phrase `"foo"` | equivalent to bare `foo` but without stem expansion |
| Tag content `@tag:{"foo bar"}` | unchanged - quoted strings in `{...}` are literal tags |
| BM25 scoring | phrase-matched docs score as if constituent terms had matched independently |
| `FT.CREATE idx NOOFFSETS ...` | opt-out: skips position storage (saves memory, breaks phrase) |
| `FT.INFO`'s `index_options` | reports `NOOFFSETS` when set |

Closes #7294

### DIFF
--- a/src/core/search/indices.cc
+++ b/src/core/search/indices.cc
@@ -636,9 +636,10 @@ void TextIndex::Tokenize(std::string_view value, uint32_t* pos_counter,
 
 std::vector<std::string> TextIndex::TokenizePhraseQuery(std::string_view phrase) const {
   std::vector<std::string> out;
+  const StopWords* sw = stopwords_;
   for (std::string_view word : una::views::word_only::utf8(phrase)) {
     std::string lc = una::cases::to_lowercase_utf8(word);
-    if (stopwords_ && stopwords_->contains(lc))
+    if (sw && sw->contains(lc))
       continue;
     out.push_back(std::move(lc));
   }
--- a/src/core/search/indices.cc
+++ b/src/core/search/indices.cc
@@ -48,43 +48,46 @@ string ToLower(string_view word) {
   return IsAllAscii(word) ? absl::AsciiStrToLower(word) : una::cases::to_lowercase_utf8(word);
 }
 
-// Get all words from text as matched by the ICU library, counting term frequencies.
-// When stemming is on, both the raw and stemmed forms are emitted so wildcards
-// can match unstemmed tokens and synonym hits stay additive.
-absl::flat_hash_map<std::string, uint32_t> TokenizeWords(std::string_view text,
-                                                         const TextIndex::StopWords& stopwords,
-                                                         const Synonyms* synonyms,
-                                                         Stemmer* stemmer) {
-  absl::flat_hash_map<std::string, uint32_t> words;
+// Tokenize text into `out`, advancing *pos_counter per non-stopword token. Raw + stem +
+// synonym entries share the same position. Stopwords don't advance the counter.
+void TokenizeWords(std::string_view text, const TextIndex::StopWords& stopwords,
+                   const Synonyms* synonyms, Stemmer* stemmer, uint32_t* pos_counter,
+                   absl::flat_hash_map<std::string, TermInfo>* out) {
+  auto emit = [&](std::string key, uint32_t pos) {
+    auto& info = (*out)[std::move(key)];
+    info.freq++;
+    info.positions.push_back(pos);
+  };
   for (std::string_view word : una::views::word_only::utf8(text)) {
     std::string word_lc = una::cases::to_lowercase_utf8(word);
     if (stopwords.contains(word_lc))
       continue;
+    uint32_t pos = ++(*pos_counter);
     if (synonyms) {
       if (auto group_id = synonyms->GetGroupToken(word_lc); group_id)
-        words[*group_id]++;
+        emit(*group_id, pos);
     }
     if (stemmer) {
       std::string stem = stemmer->Stem(word_lc);
       if (stem != word_lc)
-        words[std::move(stem)]++;
+        emit(std::move(stem), pos);
     }
-    words[std::move(word_lc)]++;
+    emit(std::move(word_lc), pos);
   }
-  return words;
 }
 
-// Split taglist, remove duplicates and convert all to lowercase. Freq is always 1 for tags.
-absl::flat_hash_map<string, uint32_t> NormalizeTags(string_view taglist, bool case_sensitive,
-                                                    char separator) {
+// Split taglist, remove duplicates and convert all to lowercase. Freq is always 1 for tags;
+// positions are not tracked (phrase queries don't apply to tag fields).
+void NormalizeTags(string_view taglist, bool case_sensitive, char separator,
+                   absl::flat_hash_map<std::string, TermInfo>* out) {
   // Splitting utf8 by ascii character is safe
-  absl::flat_hash_map<string, uint32_t> tags;
   for (string_view tag : absl::StrSplit(taglist, separator, absl::SkipEmpty())) {
     string_view str = absl::StripAsciiWhitespace(tag);
     std::string s = case_sensitive ? string{str} : ToLower(str);
-    tags.emplace(std::move(s), 1);
+    auto& info = (*out)[std::move(s)];
+    if (info.freq == 0)
+      info.freq = 1;  // dedup: tag presence is binary
   }
-  return tags;
 }
 
 // Iterate over all suffixes of all words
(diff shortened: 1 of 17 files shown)


src/core/search/indices.cc
```diff
@@ -617,8 +629,20 @@ std::optional<DocumentAccessor::StringList> TextIndex::GetStrings(const Document
...
+void TextIndex::Tokenize(std::string_view value, uint32_t* pos_counter,
+                         absl::flat_hash_map<std::string, TermInfo>* out) const {
+  TokenizeWords(value, *stopwords_, synonyms_, stemmer_, pos_counter, out);
+}
+
+std::vector<std::string> TextIndex::TokenizePhraseQuery(std::string_view phrase) const {
+  std::vector<std::string> out;
+  for (std::string_view word : una::views::word_only::utf8(phrase)) {
+    std::string lc = una::cases::to_lowercase_utf8(word);
+    if (stopwords_ && stopwords_->contains(lc))
```
BorysTheDev: it looks like stopwords_ check should be moved out of the cycle

src/core/search/lexer.lex
```diff
@@ -84,8 +84,8 @@ astrsk_ch  \*
 [0-9]{1,9}                          return Parser::make_UINT32(str(), loc());
 [+-]?(([0-9]*[.])?[0-9]+|inf)       return Parser::make_DOUBLE(str(), loc());
 
-{dq}([^"]|{esc_seq})*{dq}           return make_StringLit(matched_view(1, 1), loc());
-{sq}([^']|{esc_seq})*{sq}           return make_StringLit(matched_view(1, 1), loc());
+{dq}([^"]|{esc_seq})*{dq}(~[0-9]+)?   return make_PhraseTok(str(), loc());
```
BorysTheDev, line 89: what '~' mean?

src/core/search/search.cc
```diff
@@ -303,6 +304,129 @@ struct BasicSearch {
...
+            return true;
+        return false;
+      };
+
+      vector<DocId> matches;
+      while (!any_at_end()) {
+        DocId max_id = *iters[0];
+        for (size_t i = 1; i < iters.size(); ++i)
+          if (*iters[i] > max_id)
+            max_id = *iters[i];
```
BorysTheDev: max_id = *iters[i] > max_id ? *iters[i]  : max_id;

src/core/search/search.cc
```diff
@@ -303,6 +304,129 @@ struct BasicSearch {
...
+            ++it;
+        }
+      }
+      return IndexResult{std::move(matches)};
+    };
+
+    if (!active_field.empty()) {
+      if (auto* index = GetIndex<TextIndex>(active_field); index)
+        return eval_in(index);
+      return IndexResult{};
```
BorysTheDev: auto* index = GetIndex<TextIndex>(active_field);
return index ? eval_in(index) : IndexResult{};

src/core/search/search.cc
```diff
@@ -303,6 +304,129 @@ struct BasicSearch {
...
+      if (CanExtendPhrase(positions, 1, start, slop))
+        return true;
+    }
+    return false;
+  }
+
+  // Quoted phrase: AND-intersect posting lists of raw tokens, then verify adjacency by positions.
+  // Stems are not consulted — phrase semantics require the exact surface form. Stopwords inside
+  // the phrase are dropped (matching the index-side tokenizer, which doesn't advance pos for them).
+  IndexResult Search(const AstPhraseNode& node, string_view active_field) {
```
BorysTheDev: very complex method, consider to split it to several or find out how to make it more readable

src/core/search/indices.cc
vyavdoshenko: fixed

src/core/search/lexer.lex
vyavdoshenko: comment was added

src/core/search/search.cc

src/core/search/search.cc

src/core/search/search.cc

src/core/search/lexer.lex
BorysTheDev: maybe it will be better to make as a separate token?

src/core/search/search.cc
```diff
@@ -303,6 +304,149 @@ struct BasicSearch {
...
+        if (iters[i] == lists[i]->end())
+          return true;
+      return false;
+    };
+
+    vector<DocId> matches;
+    while (!any_at_end()) {
+      DocId max_id = *iters[0];
+      for (size_t i = 1; i < iters.size(); ++i)
+        max_id = std::max(max_id, *iters[i]);
```
BorysTheDev, line 391: looks like std::max_element

BorysTheDev (approved):

src/core/search/lexer.lex
vyavdoshenko: This is possible, but it will be more complex. More code in lexer harder to read. I prefer to leave this approach.

src/core/search/search.cc
vyavdoshenko: Yes, you are right. We can replace three lines with two lines. I'd like to leave it for now.

PR #7330, opened 2026-05-18
fix(test): Disable cooling for pytest with tiering and cluster

### LINKED ISSUE #7318: dragonfly/cluster_test.py::test_cluster_migration_with_tiering fails
https://github.com/dragonflydb/dragonfly/actions/runs/25990834881/job/76396541138

mkaruza:
On some runs in CI we can see that test can fail because we couldn't add more keys. Disable tiering cooling layer to offload all keys to disk instead of keeping some of them in memory.

Closes #7318

### DIFF
--- a/tests/dragonfly/cluster_test.py
+++ b/tests/dragonfly/cluster_test.py
@@ -3717,6 +3717,7 @@ async def test_cluster_migration_with_tiering(df_factory):
             proactor_threads=2,
             tiered_prefix="/tmp/tiered/cluster_node",
             tiered_offload_threshold="0.2",
+            tiered_experimental_cooling="False",
             maxmemory="512MB",
         ),
         df_factory.create(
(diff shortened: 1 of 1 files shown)


dranikpg:
> On some runs in CI we can see that test can fail because we couldn't add more keys. Disable tiering cooling layer to offload all keys to disk instead of keeping some of them in memory.

I don't really understand - we failed to add more keys because of OOM? It should try to retire cold entries before placement of a new value. If this doesn't happen, it's too weak

mkaruza:
> It should try to retire cold entries before placement of a new value. If this doesn't happen, it's too weak



> > On some runs in CI we can see that test can fail because we couldn't add more keys. Disable tiering cooling layer to offload all keys to disk instead of keeping some of them in memory.
> 
> I don't really understand - we failed to add more keys because of OOM? It should try to retire cold entries before placement of a new value. If this doesn't happen, it's too weak

I'm seeing this after/during `DEBUG POPULATE` command.

```
I20260517 12:37:45.724974 18055 db_slice.cc:172] Can't grow, free_items 784013, obj_bytes: 418 vs 350  mem_available: 128479981
W20260517 12:37:45.725061 18055 db_slice.cc:749] AddOrFind: InsertNew failed, budget: 59214584 reclaimed: 0 offset: 69265397
...
I20260517 12:37:47.983084 18055 engine_shard.cc:1045] Entries count 999939 obj_memory: 340619976, bytes_per_obj: 340
```

Problem is is preparation, not what it is testing. We can also lower the number of keys inserted to mitigate this.

dranikpg (approved):
I don't under

PR #7333, opened 2026-05-18
fix: migration restart flush race

### LINKED ISSUE #7188: test_migration_restart test fail
https://github.com/dragonflydb/dragonfly/actions/runs/24622859828

```
        nodes[0].migrations = []
        nodes[0].slots = [(0, final_migration_range[0] - 1), (final_migration_range[1] + 1, 16383)]
        nodes[1].slots = [final_migration_range]
    
>       assert await seeder.compare(capture, nodes[0].instance.port)
E       assert False

```

kostasrim:
### The bug

When a new cluster config is sent a stale migration that is already in-flight will be cancelled.  To avoid stale data for cancelled migrations we `flush` the removed slot range `before` any of the new migration (given in the new cluster config) kick in. 

The `before` part however was `racy` and it allowed the following sequence of execution: 

1. the previous migration (call it `A`) is cancelled 
2. the new migrations (call it `B`) is setup
3. data for new migration `B` is moving in
4. flushed newly send data

This could also been seen in the LOGS where we flushed `after` the `connection for the new migration passes the Init stage` (which means it is already streaming data)

Should resolve #7188

### DIFF
--- a/src/server/cluster/cluster_family.cc
+++ b/src/server/cluster/cluster_family.cc
@@ -593,7 +593,8 @@ void ClusterFamily::DflyClusterConfig(CmdArgList args, CommandContext* cmd_cntx)
 
     new_config = new_config->CloneWithChanges(enable_slots, disable_slots);
 
-    StartNewSlotMigrations(*new_config);
+    // Capture prev config before SetCurrent so StartNewSlotMigrations can diff correctly.
+    auto prev_config = ClusterConfig::Current();
 
     SlotSet before =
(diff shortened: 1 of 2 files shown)


kostasrim:
https://github.com/dragonflydb/dragonfly/actions/runs/26035984107/job/76558656646 

Run them for 2 hours straight until the timed out. Should be fixed

vyavdoshenko (approved):

BorysTheDev (approved):

PR #7335, opened 2026-05-18
refactor: add method CompactObject::SetMemberTime

BorysTheDev:
Summary: Refactors per-member lazy expiry time handling for DenseSet-backed containers by adding dedicated helpers on CompactObj.

Changes:

- Added CompactObj::SetMemberTime(), MemberTime(), and HasMemberExpiration() with an internal DenseSet accessor.
- Updated server call sites (set/hash iteration, shrink, expiry queries) to use the new methods instead of casting to StringSet/StringMap.
- Simplified journal command serialization by saving/restoring member-time through MemberTime()/SetMemberTime().
- Adjusted RDB save logic to use HasMemberExpiration() and to disable lazy expiry via SetMemberTime(0) during serialization.
- Added unit tests covering set/map member-time behavior and ensuring calls are safe on non-kEncodingStrMap2 objects.

Technical Notes: The new APIs are safe to call unconditionally (no-op unless encoding is kEncodingStrMap2), which reduces repetition and centralizes the DenseSet/time semantics.

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -867,6 +867,32 @@ void CompactObj::InitRobj(CompactObjType type, unsigned encoding, void* obj) {
   u_.r_obj.Init(type, encoding, obj);
 }
 
+namespace {
+// Returns the underlying DenseSet if this CompactObj wraps a StringSet/StringMap
+// (kEncodingStrMap2), otherwise nullptr. Encoding() is a safe accessor that
+// returns kEncodingStrMap2 only when taglen_ == ROBJ_TAG.
+DenseSet* AsDenseSetOrNull(const CompactObj& obj) {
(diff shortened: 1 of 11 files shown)


src/core/compact_object.cc
```diff
@@ -867,6 +867,32 @@ void CompactObj::InitRobj(CompactObjType type, unsigned encoding, void* obj) {
   u_.r_obj.Init(type, encoding, obj);
 }
 
+namespace {
```
romange, line 870: I think we already have `namespace {`  there

romange (approved):

src/core/compact_object.cc
BorysTheDev: yes 400 lines above, so I think it is better to have this one too

PR #7338, opened 2026-05-18
fix: prevent PII leaks in FailedCommandToString error logs

vyavdoshenko:
When a command fails (e.g. due to OOM), `FailedCommandToString` logged all arguments and the full error reason without any size limits. This caused sensitive customer data to be written to plain-text logs, as confirmed in before where EVAL scripts with large BullMQ payloads dumped PII (IP addresses, user data) into Loki on OOM events.

Changes:
- Truncate arguments to 31 max and 128 bytes each for all commands (mirrors slowlog behavior)
- For EVAL/EVALSHA/EVAL_RO/EVALSHA_RO: log only script/SHA and numkeys, drop KEYS and ARGV entirely to prevent user payload exposure
- For EVAL error reasons: strip the Lua error part from "Error running script (call to <sha>): <lua_error>", keeping only the structured prefix; fall back to 256-byte truncation for non-standard formats
- AUTH and ACL SETUSER retain existing behavior (no args logged)

### DIFF
--- a/src/server/main_service.cc
+++ b/src/server/main_service.cc
@@ -629,16 +629,51 @@ string ConnectionLogContext(const facade::Connection* conn) {
 
 string FailedCommandToString(std::string_view command, facade::CmdArgList args,
                              std::string_view reason) {
+  constexpr size_t kMaxArgCount = 31;
+  constexpr size_t kMaxArgLength = 128;
+  constexpr size_t kMaxReasonLength = 256;
+
   string result;
   absl::StrAppend(&result, " ", command);
 
-  if (command != "AUTH" && command != "ACL SETUSER") {
-    for (auto arg : args) {
+  auto AppendArg = [&](string_view arg) {
+    if (arg.size() > kMaxArgLength) {
+      absl::StrAppend(&result, " ", absl::CHexEscape(arg.substr(0, kMaxArgLength)), "... (",
+                      arg.size() - kMaxArgLength, " more bytes)");
+    } else {
       absl::StrAppend(&result, " ", absl::CHexEscape(arg));
+    }
+  };
+
+  const bool is_eval =
+      command == "EVAL" || command == "EVALSHA" || command == "EVAL_RO" || command == "EVALSHA_RO";
+
+  if (command == "AUTH" || command == "ACL SETUSER") {
+    // skip all args to protect passwords
+  } else if (is_eval) {
+    // log only script/SHA and numkeys, skip KEYS and ARGV to avoid PII leaks
+    for (size_t i = 0; i < std::min(args.size(), size_t{2}); ++i) {
+      AppendArg(facade::ArgS(args, i));
+    }
+  } else {
+    size_t num_args = std::min(args.size(), kMaxArgCount);
+    for (size_t i = 0; i < num_args; ++i) {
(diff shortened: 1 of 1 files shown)


src/server/main_service.cc
```diff
@@ -629,16 +629,51 @@ string ConnectionLogContext(const facade::Connection* conn) {
...
+    } else {
       absl::StrAppend(&result, " ", absl::CHexEscape(arg));
     }
+  };
+
+  const bool is_eval =
+      command == "EVAL" || command == "EVALSHA" || command == "EVAL_RO" || command == "EVALSHA_RO";
+
+  if (command == "AUTH" || command == "ACL SETUSER") {
+    // skip all args to protect passwords
```
dranikpg: maybe even startswith(ACL)

src/server/main_service.cc
```diff
@@ -629,16 +629,51 @@ string ConnectionLogContext(const facade::Connection* conn) {
...
+      AppendArg(facade::ArgS(args, i));
+    }
+  } else {
+    size_t num_args = std::min(args.size(), kMaxArgCount);
+    for (size_t i = 0; i < num_args; ++i) {
+      AppendArg(facade::ArgS(args, i));
+    }
+    if (args.size() > kMaxArgCount) {
+      absl::StrAppend(&result, " ... (", args.size() - kMaxArgCount, " more arguments)");
+    }
```
dranikpg, line 664: Just a question: I don't understand where the boundary is. Why is logging command arguments even partially ok, but a script not?

dranikpg (dismissed):

src/server/main_service.cc
```diff
@@ -629,16 +629,51 @@ string ConnectionLogContext(const facade::Connection* conn) {
...
+  auto AppendArg = [&](string_view arg) {
+    if (arg.size() > kMaxArgLength) {
+      absl::StrAppend(&result, " ", absl::CHexEscape(arg.substr(0, kMaxArgLength)), "... (",
+                      arg.size() - kMaxArgLength, " more bytes)");
+    } else {
       absl::StrAppend(&result, " ", absl::CHexEscape(arg));
     }
+  };
+
+  const bool is_eval =
```
romange: nit: just `absl::StartsWith(command, "EVAL")`

romange (dismissed):

src/server/main_service.cc
vyavdoshenko: The 128-byte cap protects against scale (megabyte dumps). The EVAL exclusion addresses a confirmed structural leak - the incident showed BullMQ passes full JSON payloads as ARGV, where even 128 bytes can expose an IP or email. Different problems, different fixes.

src/server/main_service.cc
vyavdoshenko: fixed

src/server/main_service.cc
vyavdoshenko: fixed

src/server/main_service.cc
romange: Yes, it's not a hermetic fix - we will have to balance this for now until we see we do not need these arguments anymore

dranikpg (approved):

PR #7348, opened 2026-05-19
feat: info replication is lock free

### LINKED ISSUE #7320: lock free master side "info replication"
Similarly to `experimental_replicaof_v2` we should avoid locking dflycmd related data-structures on reads (info and metrics)

BorysTheDev:
fixes: #7320

The main idea of the changes is that we read/write all the data only in the threads the data belongs to, in such a way we avoid mutex lock

### Changes
- Add a per-proactor thread-local snapshot of `replica_infos_` (`tl_replica_infos`) for lock-free reads.
- Add `UpdateReplicaInfoCacheLocked()`: copies `replica_infos_` into a fresh `shared_ptr` and publishes it to every proactor via `DispatchBrief` (caller holds  `mu_`). Called from every mutator — `CreateSyncSession`, `StopReplication`,  `BreakStalledFlowsInShard`, `CancelReplicas`.
- Make `replica_state_` a `std::atomic<SyncState>` so readers load lock-free while  writers still serialize under `shared_mu_`.
- `GetReplicasRoleInfo()` drops `mu_` and the per-replica try-lock, reading the snapshot.
- `GetReplicationMemoryStats()` switches from `RunBlockingInParallel` under `mu_` to  lock-free `RunBriefInParallel` over the snapshot.
- Rename `ReplicationLagsLocked()` → `ReplicationLags(snapshot)`;

Write paths keep their `mu_`/`shared_mu_` discipline; `last_acked_lsn`/`saver`/`streamer` stay non-atomic, safe via the same-shard owner invariant.

### DIFF
--- a/src/server/dflycmd.cc
+++ b/src/server/dflycmd.cc
@@ -70,6 +70,9 @@ const char kIdNotFound[] = "syncid not found";
 const char kInvalidSyncId[] = "bad sync id";
 const char kInvalidState[] = "invalid state";
 
+// Per-proactor view of replica_infos_. Published by UpdateReplicaInfoState.
+thread_local std::shared_ptr<const DflyCmd::ReplicaInfoMap> tl_replica_infos;
+
 bool ToSyncId(string_view str, uint32_t* num) {
   if (!absl::StartsWith(str, "SYNC"))
     return false;
@@ -117,14 +120,14 @@ bool WaitReplicaFlowToCatchup(absl::Time end_time, const DflyCmd::ReplicaInfo* r
 
 void DflyCmd::ReplicaInfo::Cancel() {
   util::fb2::LockGuard lk{shared_mu_};
-  if (replica_state_ == SyncState::CANCELLED) {
+  if (replica_state_.load(std::memory_order_relaxed) == SyncState::CANCELLED) {
     return;
   }
 
   LOG(INFO) << "Disconnecting from replica " << address_ << ":" << listening_port_;
 
   // Update state and cancel context.
-  replica_state_ = SyncState::CANCELLED;
+  replica_state_.store(SyncState::CANCELLED, std::memory_order_relaxed);
   exec_st_.ReportCancelError();
   // Wait for tasks to finish.
   shard_set->RunBlockingInParallel([this](EngineShard* shard) {
@@ -782,6 +785,7 @@ auto DflyCmd::CreateSyncSession(ConnectionState* state) -> std::pair<uint32_t, u
   auto [it, inserted] = replica_infos_.emplace(sync_id, std::move(replica_ptr));
   CHECK(inserted);
 
+  UpdateReplicaInfoState();
   return {it->first, flow_count};
 }
 
@@ -814,6 +818,7 @@ void DflyCmd::StopReplication(uint32_t sync_id) {
 
   util::fb2::LockGuard lk(mu_);
   replica_infos_.erase(sync_id);
+  UpdateReplicaInfoState();
 }
 
 // Because we need to annotate unique_lock
@@ -852,6 +857,9 @@ void DflyCmd::BreakStalledFlowsInShard() {
 
   for (auto sync_id : deleted)
     replica_infos_.erase(sync_id);
+
+  if (!deleted.empty())
+    UpdateReplicaInfoState();
 }
 
 shared_ptr<DflyCmd::ReplicaInfo> DflyCmd::GetReplicaInfo(uint32_t sync_id) {
@@ -865,27 +873,17 @@ shared_ptr<DflyCmd::ReplicaInfo> DflyCmd::GetReplicaInfo(uint32_t sync_id) {
 
 std::vector<ReplicaRoleInfo> DflyCmd::GetReplicasRoleInfo() const {
   std::vector<ReplicaRoleInfo> vec;
-  util::fb2::LockGuard lk(mu_);
+  auto replica_infos = tl_replica_infos;
+  if (!replica_infos)
+    return vec;
 
-  vec.reserve(replica_infos_.size());
-  map replication_lags = ReplicationLagsLocked();
+  vec.reserve(replica_infos->size());
+  std::map<uint32_t, LSN> replication_lags = ReplicationLags(*replica_infos);
 
-  for (const auto& [id, info] : replica_infos_) {
-    LSN lag = replication_lags[id];
-    SyncState state = SyncState::PREPARATION;
-
-    // If the replica state being updated, its lag is undefined,
-    // the same applies of course if its state is not STABLE_SYNC.
-    shared_lock lk(info->GetMutex(), try_to_lock);
-    if (lk.owns_lock()) {
-      state = info->GetReplicaState();
-      // If the replica is not in stable sync, its lag is undefined, so we set it to 0.
-      if (state != SyncState::STABLE_SYNC) {
-        lag = 0;
-      }
-    } else {
-      lag = 0;
-    }
+  for (const auto& [id, info] : *replica_infos) {
+    SyncState state = info->GetReplicaState();
+    // Lag is undefined unless the replica is in stable sync.
+    LSN lag = (state == SyncState::STABLE_SYNC) ? replication_lags[id] : 0;
     vec.push_back(ReplicaRoleInfo{std::string{info->GetId()}, info->GetAddress(),
                                   info->GetListeningPort(), SyncStateName(state), lag});
   }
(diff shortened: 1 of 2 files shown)


src/server/dflycmd.h
```diff
@@ -40,14 +40,16 @@ struct FlowInfo {
...
+  // ever reads or writes these pointers, so no synchronization is needed.
   std::unique_ptr<RdbSaver> saver;            // Saver for full sync phase.
   std::unique_ptr<JournalStreamer> streamer;  // Streamer for stable sync phase
   std::string eof_token;
 
   DflyVersion version = DflyVersion::VER1;
 
   std::optional<LSN> start_partial_sync_at;
-  uint64_t last_acked_lsn = 0;
+  std::atomic<uint64_t> last_acked_lsn = 0;
```
romange: why it is atomic? which threads access it now?

src/server/server_family.cc
```diff
@@ -4238,7 +4238,7 @@ void ServerFamily::ReplConf(CmdArgList args, CommandContext* cmd_cntx) {
         return;
       }
       VLOG(2) << "Received client ACK=" << ack;
-      cntx->master_repl_flow->last_acked_lsn = ack;
+      cntx->master_repl_flow->last_acked_lsn.store(ack, std::memory_order_relaxed);
```
kostasrim: you break the consistency model of lsn's. `memory_order_relaxed` can backfire here and can `undercount`

BorysTheDev: I didn't change the consistency model, only allowed to read without a mutex.

src/server/dflycmd.h
BorysTheDev: I allow to read this var without lock for INFO command, this is why it is atomic now, everything else is the same. so all write operations under the lock.

BorysTheDev: Maybe I don't need this change, I will try to improve the code

kostasrim (commented):
I would suggest to approach this differently. Notice that `info` can work on `stale data`. There are only two flows:

1. GetReplicasRoleInfo
2. GetReplicationMemoryStats

and two locks:

1. mu_ (replica_infos)_
2. shared_mu_ / per replica 

You can get rid of the lock for `mu_` simply by keeping a thread local `replica_info` (just like you did here!). This solves 50% of your problems. 

Now notice how `shared_mu_` is used. For `GetReplicasRoleInfo` we use `try to lock`. So this adheres to the principle "stale data". Therefore, that path is fine.

What's left is `GetReplicationMemoryStats` which acquires a shared lock of the shared_mu_. To avoid this we can apply the same principle here: skip the path if the mutex is locked (cc @romange if you agree)

That way you get correctness with minimal size changes (and the path is lock free)

src/server/dflycmd.h
kostasrim: we don't need it. I wrote what we can do and if something don't make sense I can explain tomorrow in daily :)

romange:
> Now notice how `shared_mu_` is used. For `GetReplicasRoleInfo` we use `try to lock`. So this adheres to the principle "stale data". Therefore, that path is fine.

Hmm, I do not think so. It's not that `GetReplicasRoleInfo` returns the stale role in case the data is locked. if it skips the data is missing. It's fine to return eventually consistent reads but it's absolutely not fine to return sporadic nils.

BorysTheDev:
@romange @kostasrim I have another idea, let me implement it and then we can discuss

kostasrim:
> issing. It's fine to return eventually consistent reads but it's absolutely not fine to return sporadic nils.

yeah I understand, especially for memory I guess (we expose those on charts in grafana as well). Well then just dispatch and do the access only on a single thread (centralized single thread access). No mutex involved

src/server/dflycmd.cc
```diff
@@ -116,14 +119,14 @@ bool WaitReplicaFlowToCatchup(absl::Time end_time, const DflyCmd::ReplicaInfo* r
 
 void DflyCmd::ReplicaInfo::Cancel() {
   util::fb2::LockGuard lk{shared_mu};
-  if (replica_state == SyncState::CANCELLED) {
+  if (replica_state.load(std::memory_order_relaxed) == SyncState::CANCELLED) {
```
kostasrim: My only concern/question here is can `replica_state` here be stale ? 

So imagine two threads both trigger `BreakStalledFlows()`. Flow 1 grabs the mutex, and performs a `relaxed store` on `replica_state`.  Flow 1 completes, and flow 2 wakes up, acquires the mutex and `relax reads` the replica_state. Can it read a `stale value` ?

I think it won't read a stale value because the`shared_mu` above will act as a implicit barrier between the two threads (acquire release semantics of the spinlock underneath)

Could you plz double check that other places are safe as well so we don't end up getting any surprises around memory ordering ?

src/server/dflycmd.cc
```diff
@@ -939,21 +949,28 @@ pair<uint32_t, shared_ptr<DflyCmd::ReplicaInfo>> DflyCmd::GetReplicaInfoOrReply(
...
+    for (const auto& info : replicas) {
       const ReplicaInfo* replica = info.second.get();
-      if (shard->journal()) {
-        int64_t lag = journal::GetLsn() - replica->flows[shard->shard_id()].last_acked_lsn;
-        lags[info.first] = lag;
-      }
+      if (replica->replica_state.load(std::memory_order_relaxed) != SyncState::STABLE_SYNC)
+        continue;
+      LSN acked = replica->flows[shard->shard_id()].last_acked_lsn;
+      lags[info.first] = cur_lsn > acked ? cur_lsn - acked : 0;
```
kostasrim: How can it ever be that `cur_lsn < acked` unless we come from the future ? Am I missing something ?

kostasrim (commented):
lgtm, minor comments

src/server/dflycmd.cc
BorysTheDev: discussed

BorysTheDev: I've checked code and everytyhing should be correct

src/server/dflycmd.h
```diff
@@ -115,13 +121,17 @@ class DflyCmd {
     // Transition into cancelled state, run cleanup.
     void Cancel();
 
-    SyncState replica_state;  // always guarded by shared_mu
+    // Transitions still serialized under shared_mu; atomic so readers can load lock-free.
+    std::atomic<SyncState> replica_state;
     ExecutionState exec_st;
 
+    // id is written by the REPLCONF CLIENT-ID handler on owner_thread; readers from
```
romange: if we will allow writing id in "REPLCONF CLIENT-ID" handler only once (i.e. at most once for the lifetime of `ReplicaInfo` then we can avoid `AwaitBrief` by introducing an atomic_bool guard:
```
// saver
if (id_guard.load() == false) {
  id = ...
  id_guard.store(true, release);
}

// loader
if (id_guard.load(acquire)) {
   string local_id = replica_info->id;   // safe to access as it never changes.
}

```

src/server/dflycmd.h
```diff
@@ -115,13 +121,17 @@ class DflyCmd {
     // Transition into cancelled state, run cleanup.
     void Cancel();
 
-    SyncState replica_state;  // always guarded by shared_mu
+    // Transitions still serialized under shared_mu; atomic so readers can load lock-free.
+    std::atomic<SyncState> replica_state;
```
romange: I suggest making it a class and move variables to the private scope.
otherwise it's super hard to track which variables are thread-safe.

`Methods` are simpler to track and also some of them could hide the complexities of multi-threading access by exposing the exact semantics needed to operate on `ReplicaInfo` object. 

I suggest you first create a refactoring PR that hardens the interface around the object and then send this change

BorysTheDev:
augment review

romange:
```
ReplicaInfo::id - captures the writer's proactor as owner_thread at
construction; reader hops via AwaitBrief to read id on the same thread. Same pattern as Replica::GetSummary.
```

I think the commit and PR descriptions are not up to date. There is no `AwaitBrief` in the diff.

src/server/dflycmd.cc
```diff
@@ -70,6 +70,9 @@ const char kIdNotFound[] = "syncid not found";
 const char kInvalidSyncId[] = "bad sync id";
 const char kInvalidState[] = "invalid state";
 
+// Per-proactor view of replica_infos_. Published by UpdateReplicaInfoState.
```
romange:
```suggestion
// Per-proactor cached view of replica_infos_. 
// Every mutation of centralized `replica_infos_` warrants subsequent call to UpdateReplicaInfoState.
```

src/server/dflycmd.cc
```diff
@@ -814,6 +818,7 @@ void DflyCmd::StopReplication(uint32_t sync_id) {
 
   util::fb2::LockGuard lk(mu_);
   replica_infos_.erase(sync_id);
+  UpdateReplicaInfoState();
```
romange: I think `UpdateReplicaInfoCacheLocked` is a better name

src/server/dflycmd.cc
```diff
@@ -937,21 +940,27 @@ pair<uint32_t, shared_ptr<DflyCmd::ReplicaInfo>> DflyCmd::GetReplicaInfoOrReply(
   return {sync_id, sync_it->second};
 }
 
-std::map<uint32_t, LSN> DflyCmd::ReplicationLagsLocked() const {
-  DCHECK(!mu_.try_lock());  // expects to be under global lock
-  if (replica_infos_.empty())
+std::map<uint32_t, LSN> DflyCmd::ReplicationLags(const ReplicaInfoMap& replicas) const {
```
romange: rename `replicas` to `replicas_local_cache`

src/server/dflycmd.cc
```diff
@@ -937,21 +940,27 @@ pair<uint32_t, shared_ptr<DflyCmd::ReplicaInfo>> DflyCmd::GetReplicaInfoOrReply(
...
 
   // In each shard we calculate a map of replica id to replication lag in the shard.
+  // Only compute lag for replicas currently in STABLE_SYNC — for any other state the
+  // lag is undefined (e.g. last_acked_lsn is still 0 mid-full-sync). Replicas that
+  // transition out of (or into) STABLE_SYNC between this scan and the outer reader
+  // are naturally handled: missing entries default to lag=0, and the outer reader
+  // re-checks state before emitting the lag.
   std::vector<std::map<uint32_t, LSN>> shard_lags(shard_set->size());
-  shard_set->RunBriefInParallel([&shard_lags, this](EngineShard* shard) {
+  shard_set->RunBriefInParallel([&shard_lags, &replicas](EngineShard* shard) {
```
romange: it took me a while to understand why you need `RunBriefInParallel` if you operate on `replicas`.

The only reason is to read `journal::GetLsn()`, correct? Worth adding a comment.

src/server/dflycmd.cc
```diff
@@ -965,6 +974,12 @@ std::map<uint32_t, LSN> DflyCmd::ReplicationLagsLocked() const {
   return rv;
 }
 
+void DflyCmd::UpdateReplicaInfoState() {
+  auto replica_infos = std::make_shared<const ReplicaInfoMap>(replica_infos_);
```
romange, line 983: it's a bit confusing to me.
1. `auto replica_infos = std::make_shared<const ReplicaInfoMap>(replica_infos_);` is a deep copy of `replica_infos_` so it must be at least under read lock, correct?
2. `shard_set->pool()->AwaitFiberOnAll(` does not need a lock, as `replica_infos` but we still hold it, correct?

Moreover, you do not need `AwaitFiberOnAll` as this lambda is a cpu-only, atomic function.
Also, you do not need to `Await` - only to dispatch.

> Note, you actually need dispatching under lock as otherwise you may find yourself with inconsistent caches across fibers if two threads in parallell will call this function. But extending the lock for Await is still not needed, imho.

src/server/dflycmd.cc
```diff
@@ -865,27 +873,17 @@ shared_ptr<DflyCmd::ReplicaInfo> DflyCmd::GetReplicaInfo(uint32_t sync_id) {
...
-    } else {
-      lag = 0;
-    }
+  vec.reserve(replica_infos->size());
+  std::map<uint32_t, LSN> replication_lags = ReplicationLags(*replica_infos);
+
+  for (const auto& [id, info] : *replica_infos) {
+    SyncState state = info->GetReplicaState();
+    // Lag is undefined unless the replica is in stable sync.
+    LSN lag = (state == SyncState::STABLE_SYNC) ? replication_lags[id] : 0;
```
romange: nit: ReplicationLags already skips non `STABLE_SYNC` states. we can just call "read" `replication_lags[id]`.

romange (commented):
lgtm

src/server/dflycmd.cc
BorysTheDev: @romange I didn't get your question 1, we have `void UpdateReplicaInfoState() ABSL_EXCLUSIVE_LOCKS_REQUIRED(mu_);` that gives us a guarantee that this method calls under lock.
2. I think we need lock to prevent the race of replica_infos_ updating.
regarding dispatch I agree

romange (approved):

PR #7361, opened 2026-05-20
core: split LargeString out of RobjWrapper

romange:
## Summary

First step of decoupling `CompactObj` from the monolithic `detail::RobjWrapper`. The raw large-string case (formerly `OBJ_STRING` stored inside `RobjWrapper`) moves into a dedicated `detail::LargeString` type with its own `LARGE_STR_TAG`.

All churn is internal to `src/core/compact_object.{h,cc}` — verified that no external caller (`src/server/*_family.cc`, `rdb_*`, `container_utils`, `src/server/tiering/*`, `src/server/search/*`) ever used `RObjPtr`/`InitRobj` for `OBJ_STRING`. The `InitRobj`/`RObjPtr`/`SetRObjPtr` external API is unchanged.

## Changes

- **New `detail::LargeString`** — `{ void* ptr; uint64_t sz:56; uint64_t reserved:8 }` (16 bytes, packed). Inline `Size`, `AsView`, and `Equal` in the header; out-of-line `MallocUsed`, `HashCode`, `SetString`, `ReserveString`, `AppendString`, `Free`, `DefragIfNeeded`; private `ReallocateString`.
- **New `LARGE_STR_TAG` (= 26)** in `TagEnum` and a `large_str` member in `CompactObj::U`.
- **`CompactObj` rerouted** — `Size`, `HashCode`, `ObjType`, `Encoding`, `ReserveString`, `AppendString`, `GetSlice`, `GetString`, `GetEncodedBlob`, `GetFirstByte`, `SetByteAtIndex` (raw + ASCII-encoded), `Free`, `MallocUsed`, `DefragIfNeeded`, `HasAllocated`, `CmpNonInline`, `CmpEncoded`, `EncodeString`, `Materialize`, `GetRawString`, `SetExpireTime`.
- **`RobjWrapper` slimmed** — removed `OBJ_STRING` branches in `Free`/`Size`/`MallocUsed`/`DefragIfNeeded`; removed `HashCode` and both `Equal` overloads (string-only with `LOG(FATAL)` otherwise); removed `SetString`/`ReserveString`/`AppendString`/`AsView`/`ReallocateString`/`MakeInnerRoom`/`InnerObjMallocUsed` helpers. `RobjWrapper` now covers only collection types.
- **`LargeString::SetString` fully replaces contents on every call.** The original `RobjWrapper::SetString` guarded the entire body with `if (s.size() > sz_)`, so shorter/equal-length replacements silently no-op'd. Not currently reachable (every caller goes through `SetMeta` first which zeroes the union), but the contract was fragile. The carved-out version drops the guard. Added `LargeStringSetReplacesContents` regression test covering shorter / equal-length / longer / empty replacements.
- **No doubling/preallocation logic carried over.** Every current caller hits `SetString`/`ReserveString` with `cur_cap == 0` and `sz == 0`, so `MakeInnerRoom`'s growth heuristic and old-content preservation were dead code. Re-adding the doubling later, if a `SetString`-on-populated-buffer caller appears, is trivial.

## Why

- Eliminates dead branches: `RobjWrapper::HashCode` and both `Equal` overloads were string-only with `LOG(FATAL)` otherwise — string-only methods masquerading as polymorphic.
- The 4-bit `encoding_` field in `RobjWrapper` was dead weight for `OBJ_STRING` (always `OBJ_ENCODING_RAW`; `CompactObj::encoding_` already owns ASCII/Huffman encoding).
- Sets up a follow-up PR that replaces `ROBJ_TAG` with per-collection tags (`LIST_TAG`/`SET_TAG`/`HASH_TAG`/`ZSET_TAG`/`STREAM_TAG`) and drops the `type_` field from `RobjWrapper` — much cleaner once strings are out of the way.

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -408,9 +408,6 @@ size_t RobjWrapper::MallocUsed(bool slow) const {
     return 0;
 
   switch (type_) {
-    case OBJ_STRING:
-      CHECK_EQ(OBJ_ENCODING_RAW, encoding_);
-      return InnerObjMallocUsed();
     case OBJ_LIST:
       if (encoding_ == kEncodingListPack) {
         return zmalloc_usable_size(inner_obj_);
@@ -434,9 +431,6 @@ size_t RobjWrapper::MallocUsed(bool slow) const {
 
 size_t RobjWrapper::Size() const {
   switch (type_) {
-    case OBJ_STRING:
-      DCHECK_EQ(OBJ_ENCODING_RAW, encoding_);
-      return sz_;
     case OBJ_LIST:
       if (encoding_ == kEncodingListPack) {
         return lpLength((uint8_t*)inner_obj_);
@@ -495,11 +489,6 @@ void RobjWrapper::Free(MemoryResource* mr) {
   DVLOG(1) << "RobjWrapper::Free " << inner_obj_;
 
   switch (type_) {
-    case OBJ_STRING:
-      DVLOG(2) << "Freeing string object";
-      DCHECK_EQ(OBJ_ENCODING_RAW, encoding_);
-      mr->deallocate(inner_obj_, 0, 8);  // we do not keep the allocated size.
-      break;
     case OBJ_LIST:
       FreeList(encoding_, inner_obj_, mr);
       break;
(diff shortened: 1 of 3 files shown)


dranikpg (approved):
This is a great refactoring PR, RObjWrapper is hard to justify or explain

romange:
The justification is in its name - its Redis legacy stuff :)

PR #7366, opened 2026-05-20
perf: zero-copy GET for raw large strings (MVP, read-only)

romange:
> [!IMPORTANT]
> **MVP — read-only correctness only.** Ships the zero-copy code path but the lifetime guarantee is provided only by an externally-enforced read-only workload (no `SET`/`APPEND`/`DEL`/`EXPIRE`/defrag during the read window). The point of merging this is to **measure the gain** and decide whether the full Copy-on-Write follow-up is worth building.

## Summary

Removes the per-GET `pv.ToString()` allocation for raw (`NONE_ENC`) large string values. The shard-side now passes a borrowed view from `CompactObj`'s underlying storage straight to `SinkReplyBuilder`. The existing `WriteRef` + `ReplyScope` infrastructure already provides iovec-based zero-copy on the wire side (`reply_builder.cc:129-134`, early-flush at `cc:193` when `total_size_*2 >= kMaxBufferSize`); the only missing piece was avoiding the shard-side `std::string` allocation.

## Why MVP-first

Full Copy-on-Write — refcount, `PendingRead`, `MPSCIntrusiveQueue`, `PerformDeletionAtomic` interception, defrag pinning, MGET restructure, encoded-string fraction analysis — is substantial work that pays off only if the underlying perf gain is real. This change is small enough that we can land it, run a benchmark, and decide.

If the read-only number is compelling, the follow-up work then knows what it's paying for. If it isn't, we revert and move on.

## Changes

- **`src/core/compact_object.{h,cc}`** — new `CompactObj::TryGetRawView()` returning `optional<string_view>` for `taglen_==LARGE_STR_TAG && encoding_==NONE_ENC && !IsExternal()`; `nullopt` otherwise. Lifetime contract documented on the declaration.
- **`src/server/string_family.cc`**:
  - `StringResult` reshaped from `TResultOrT<string>` to a 3-variant `variant<string, BorrowedString, TResult<string>>` where `BorrowedString` is a strong-typed tag wrapping `string_view`. The tag avoids implicit-conversion ambiguity (`std::string` is convertible to both `std::string_view` and to the variant — using a strong type breaks the tie).
  - New `ReadStringBorrow` helper: returns the borrowed alternative for raw large strings, falls back to `ReadString` otherwise.
  - `GetReplies::Send(StringResult&&)` overload dispatches the 3 alternatives (`std::string`, `BorrowedString`, tiering future). The generic `Send(TResultOrT<T>&&)` stays for `T = size_t` and other 2-variant uses.
  - Explicit `Send(const std::string&)` overload added to disambiguate calls (was previously selected via implicit `string -> string_view` conversion, which became ambiguous with the new variant overload).
  - `CmdGet` switches to `ReadStringBorrow`. `GETDEL`/`GETEX`/`GETSET` are unchanged (mutate the key, can't borrow).
- **`src/server/string_family_test.cc`** — new `GetLargeRawBorrowed` test at 1 KiB, 4 KiB, 16 KiB random binary values, covering below and above the `SinkReplyBuilder` early-flush threshold (`kMaxBufferSize/2 = 4 KiB`).

## Lifetime / safety

The borrowed view's contract is documented at three layers:

- `CompactObj::TryGetRawView` declaration in `compact_object.h`.
- `ReadStringBorrow` definition in `string_family.cc`.
- Inline comment inside `CmdGet`.

The view is valid only as long as the underlying storage is not mutated, freed, or relocated. Concretely, during the window between `SendBulkString(borrowed_view)` and the completion of the next `Flush()`, none of the following may execute on the same shard fiber:

- `SET` overwriting the same key — frees the old buffer inline via `LargeString::SetString`.
- `APPEND` / `SETRANGE` — `string_family.cc:232-238` reads + `SetString`.
- `DEL` / TTL expiry / eviction — goes through `PerformDeletionAtomic`.
- `LargeString::DefragIfNeeded` — runs in the shard's idle task, can `ReallocateString()` a `LARGE_STR_TAG` buffer in place.

The current MVP relies on the benchmark / caller workload being read-only and defrag-disabled. Mixing this code path with concurrent mutations is undefined behavior. **Do not enable this in a workload that mutates keys until the full CoW follow-up lands.**

## Out-of-scope follow-ups (if benchmark warrants)

- `SET`/`APPEND` bypass `PerformDeletionAtomic` (they free the prior buffer inline via `LargeString::SetString`) — the CoW path must intercept inside `LargeString::SetString` or route mutating commands through `PerformDeletionAtomic`.
- Defrag pin awareness.
- MGET restructure (today `CollectKeys` allocates a per-shard storage buffer; zero-copy would skip it and carry borrowed views).
- Cross-thread free coordination (the IO thread drops the last ref; mimalloc supports cross-thread deallocation, but the overhead needs to be measured).
- Encoded-string fraction in production-shaped data — caps the upside of the full work.

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -930,6 +930,14 @@ void CompactObj::AppendString(std::string_view str) {
   u_.large_str.AppendString(str, tl.local_mr);
 }
 
+std::optional<std::string_view> CompactObj::TryGetRawView() const {
+  if (encoding_ != NONE_ENC)
+    return std::nullopt;
+  if (taglen_ != LARGE_STR_TAG)
+    return std::nullopt;
+  return u_.large_str.AsView();
+}
+
 string_view CompactObj::GetSlice(string* scratch) const {
   CHECK(!IsExternal());
 
--- a/src/core/compact_object.h
+++ b/src/core/compact_object.h
@@ -232,6 +232,19 @@ class CompactObj {
 
   std::string_view GetSlice(std::string* scratch) const;
 
+  // Read-only fast path. Returns a borrowed view into the underlying storage
+  // iff this CompactObj holds a raw (NONE_ENC) large string in memory — i.e.
+  // taglen_==LARGE_STR_TAG, encoding_==NONE_ENC, !IsExternal(). Returns
+  // std::nullopt otherwise (inline / int / small / encoded / external /
+  // collection / etc), in which case the caller must use the materializing
+  // path (GetSlice/GetString/ToString).
+  //
+  // The returned view is valid only as long as this CompactObj's storage is
+  // not mutated, freed, or relocated (defrag). The caller is responsible for
+  // that lifetime — see facade::SinkReplyBuilder::ReplyScope for the reply
+  // builder's contract.
+  std::optional<std::string_view> TryGetRawView() const;
+
   std::string ToString() const {
     std::string res;
     GetString(&res);
--- a/src/facade/reply_builder.h
+++ b/src/facade/reply_builder.h
@@ -188,6 +188,16 @@ class RedisReplyBuilderBase : public SinkReplyBuilder {
   void SendSimpleString(std::string_view str) override;
   virtual void SendBulkString(std::string_view str);  // RESP: Blob String
 
+  // Like SendBulkString, but the caller guarantees that the underlying bytes
+  // remain valid until the next reply flush — typically because they're
+  // borrowed from a stable in-memory source (e.g., a CompactObj raw payload
+  // under the read-only invariant). The default implementation forwards to
+  // SendBulkString; CapturingReplyBuilder overrides this to record the view
+  // itself, preserving zero-copy through the squashing capture/replay path.
+  virtual void SendBulkStringBorrowed(std::string_view str) {
+    SendBulkString(str);
+  }
+
   void SendLong(long val) override;
   virtual void SendDouble(double val);  // RESP: Number
 
--- a/src/facade/reply_capture.cc
+++ b/src/facade/reply_capture.cc
@@ -54,6 +54,15 @@ void CapturingReplyBuilder::SendBulkString(std::string_view str) {
   Capture(BulkString{string{str}});
 }
 
+// Borrowed variant: the caller guarantees the underlying bytes outlive this
+// captured reply (until Apply runs against the real sink and the iovec is
+// flushed). We store the view directly — no allocation, no copy — so the
+// zero-copy borrowed-view path from CmdGet survives squashing.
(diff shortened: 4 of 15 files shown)


romange:
## Benchmark: zero-copy GET impact on CPU and squashing reply memory

**Setup**
- Instance: `c6gn.4xlarge` (16 vCPUs)
- Workload: 64 KB binary string values, read-only
- Client: `memtier_benchmark --test-time=10 --ratio 0:1 --key-maximum=1000 --distinct-client-seed --hide-histogram --pipeline=100`

**Results**

| Metric | `--get_zero_copy=false` | `--get_zero_copy=true` | Delta |
|---|---:|---:|---:|
| Top CPU (server) | 950% | 330% | **−65%** (~2.9× lower) |
| Peak `commands_squashing_replies_bytes` | ~1 GB | ~700 KB | **~1500× lower** |
| Throughput (Ops/sec) | _unchanged_ | _unchanged_ | — |
| Latency (avg / p50 / p99) | _unchanged_ | _unchanged_ | — |

**Why throughput and latency are unchanged**

Both runs are **network/wire-bound at the 64 KB × pipeline=100 workload**, not CPU-bound. The wire is saturated at the same byte rate regardless of how much CPU is spent producing those bytes, so throughput and latency converge. The optimization shows up as **server-side headroom** instead: the same workload now runs at ~330% CPU instead of ~950%, freeing 6+ vCPUs for additional concurrent clients, other commands, replication, snapshotting, etc.

In other words, on a network-bottlenecked workload the gain is **capacity**, not **speed** — measured in CPU and memory, not Ops/sec. To observe a throughput gain, the workload has to be CPU-bound (smaller values, faster network, more concurrent clients, or all of the above).

**Takeaways**

- The zero-copy path eliminates the per-GET `pv.ToString()` allocation+memcpy (~64 KB each) plus the squasher's `CapturingReplyBuilder` copy. Under pipelined load this is the dominant CPU cost on the GET path.
- `commands_squashing_replies_bytes` collapses from ~1 GB to ~700 KB because captured borrowed views report ~40 B per entry (variant overhead only) instead of `40 B + value_size`. This effectively removes the per-batch reply-buffer footprint as a memory cost.
- Caveat: this is the read-only-invariant MVP. Production rollout still needs the CoW work (refcounting + `PerformDeletionAtomic` interception + defrag pinning). These numbers help justify that follow-up investment.

romange:
For 64KB+ strings the CPU overhead is much higher probably due to how mimalloc allocates large buffers. With buffers smaller than 64KB, the cpu overhead is 15-20% higher with main compared to zero-copy. Zero copy CPU grows on the other hand compared to 64KB+ buffers from 330% to 450-500% CPU.

PR #7370, opened 2026-05-20
fix(lua): Harden sandbox by protecting rawset, setmetatable, and getmetatable

mkaruza:
Override rawset, setmetatable, and getmetatable to block access to _G and
all protected global tables (string, table, math, etc.). Attach guard
metatables to all global library tables, preventing scripts from accessing
or replacing their metatables.

### DIFF
--- a/src/core/interpreter.cc
+++ b/src/core/interpreter.cc
@@ -348,6 +348,75 @@ void LoadLibrary(lua_State* lua, const char* libname, lua_CFunction luafunc) {
   lua_call(lua, 1, 0);
 }
 
+// Returns true if the table at stack index idx is _G, its metatable, or any global table value.
+bool IsGlobalTableOrMetatable(lua_State* lua, int idx) {
+  lua_pushglobaltable(lua);
+
+  // Check if idx is _G itself.
+  if (lua_rawequal(lua, idx, -1)) {
+    lua_pop(lua, 1);
+    return true;
+  }
+
+  // Check if metatable of _G.
+  bool has_mt = lua_getmetatable(lua, -1);
+  if (has_mt) {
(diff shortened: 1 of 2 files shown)


src/core/interpreter.cc
```diff
@@ -348,6 +348,75 @@ void LoadLibrary(lua_State* lua, const char* libname, lua_CFunction luafunc) {
...
+  // Check if idx is any table stored directly as a global value.
+  lua_pushglobaltable(lua);
+  lua_pushnil(lua);
+  while (lua_next(lua, -2) != 0) {
+    if (lua_type(lua, -1) == LUA_TTABLE && lua_rawequal(lua, idx, -1)) {
+      lua_pop(lua, 3);
+      return true;
+    }
+    lua_pop(lua, 1);
+  }
```
vyavdoshenko, line 383: Performance concern - IsGlobalTableOrMetatable is O(N) per call

Every rawset and setmetatable walks the entire _G table to check if arg-1 is one of its values. Scripts that call rawset in a loop will pay this cost on every call.

Suggestion: cache the protected-table set in the Lua registry once during InitLua, then do an O(1) lua_rawget lookup. This also makes the protected set explicit and auditable rather than "whatever happens to be in _G at call time."

vyavdoshenko (commented):
lgtm

src/core/interpreter.cc
mkaruza: @vyavdoshenko  Agree that this is O(N). But practically in the same time set keys is  limited to roughly ~35-40 keys. Since we don't allow modification of  G_ this set never should grow - so cache is not really worth it.

vyavdoshenko (approved):

PR #7371, opened 2026-05-20
fix(server): Enable journal omits only under flag

dranikpg:
Extends #7060 with a flag


dranikpg:
I don't want to call GetFlag each time, so I cache it. It's not mutable but I assume that's ok?

romange (approved):

src/server/db_slice.cc
romange, line 1981: what happens with two replicas? 
why `change_cb_ == 1` and `journal::GetCallbackCount() == 1` ?

dranikpg: we first must guarantee that they are all eventually consistent (i.e. no snapshot) and we must guaratnee that the condition is true for all replicas (we can't omit if in-between). We could generalize it though if it's worth the complexity

PR #7374, opened 2026-05-20
Remove SerializedMap, store listpack directly in tiered value

dranikpg:
Remove serialized map because listpack itself is also just fine, previously I planned to extend this format with something more sophisticated


dranikpg:
augment review

mkaruza (approved):
Sure!

PR #7377, opened 2026-05-21
feat(server): Support CLIENT LIST [TYPE ...] [ID ...] filters

### LINKED ISSUE #5050: add per connections stats in client list
Similarly to valkey, `client list` should show:  tot-net-in, tot-net-out, tot-cmds

see https://github.com/valkey-io/valkey/pull/327

### LINKED ISSUE #7373: CLIENT LIST does not accept TYPE or ID filters (rejects all arguments with syntax error)
Dragonfly's CLIENT LIST implementation rejects every argument with "syntax error", so all filtered forms of the command are unusable. Redis supports two optional filters per the command spec:

    CLIENT LIST [TYPE normal | master | replica | pubsub]
                [ID client-id [client-id ...]]

Against Redis 8.4.2, all of the following work; against Dragonfly, all of them fail with "syntax error":

    CLIENT LIST TYPE normal
    CLIENT LIST TYPE master
    CLIENT LIST TYPE replica
    CLIENT LIST TYPE pubsub
    CLIENT LIST ID 1
    CLIENT LIST ID 1 2 3 4

Only the no-argument form (CLIENT LIST) is currently accepted by Dragonfly.

Reproduction

    # Redis 8.4.2 - works
    redis-cli -p 6379 CLIENT LIST TYPE pubsub
    redis-cli -p 6379 CLIENT LIST ID 1 2 3

    # Dragonfly - both return "ERR syntax error"
    redis-cli -p 6380 CLIENT LIST TYPE pubsub
    redis-cli -p 6380 CLIENT LIST ID 1 2 3

Expected behavior (full spec)

1.  No arguments
    Return all connected clients. Already works today; keep as-is.

2.  TYPE filter
    CLIENT LIST TYPE normal     - all regular client connections
    CLIENT LIST TYPE master     - master link connection (replication source)
    CLIENT LIST TYPE replica    - replica connections currently attached
    CLIENT LIST TYPE pubsub     - clients with at least one active subscription
                                  (SUBSCRIBE, PSUBSCRIBE, or SSUBSCRIBE)

    Unknown type must return:  ERR Unknown client type 'xxx'
    Missing type value must return: ERR syntax error

3.  ID filter
    CLIENT LIST ID 1 2 3        - return only clients whose connection id is
                                  in the given set. Order in the reply matches
                                  the natural connection enumeration, not the
                                  order of ids supplied.
    Non-integer id must return: ERR Invalid client ID
    Missing id value must return: ERR syntax error

4.  Combining TYPE and ID
    In Redis 8.4.2 the two filters are mutually exclusive at parse time:
    "CLIENT LIST TYPE normal ID 1" returns "ERR syntax error".

5.  Argument order
    The optional clauses must come after the LIST subcommand. Anything
    else stays a syntax error so we do not over-accept.

6.  Output format
    Unchanged: one client per line, fields separated by spaces, terminated
    with newline, sent as a verbatim/bulk string - same as the current
    no-arg implementation.

Motivation

Found while running CrewAI (https://github.com/crewAIInc/crewAI) against Dragonfly. CrewAI uses portalocker.RedisLock for distributed locks (crewai_core.lock_store). portalocker's stale-subscriber recovery path calls connection.client_list('pubsub') to find and kill orphaned lock holders that did not respond to a ping. On Dragonfly this call fails with "syntax error" and the recovery loop has to wait out the full timeout instead of cleaning up the stale subscription quickly.

Outside CrewAI: CLIENT LIST TYPE/ID is part of the standard Redis command surface used by monitoring tools, ops dashboards, and any library that needs to enumerate connections of a specific class (RedisInsight, prometheus exporters, etc.).

Scope of this issue

This is a request to close the gap completely, matching Redis behavior across all the cases above:

  - parse TYPE with the four valid values and the correct error for unknown types
  - parse ID with one or more integers and the correct error for non-integer ids
  - reject combining TYPE and ID at parse time
  - reject any other tokens with a syntax error
  - preserve the existing zero-argument behavior and output format

Acceptance criteria

The following commands, run against Dragonfly, produce results that in shape (set of clients returned, error message text where applicable):

    CLIENT LIST
    CLIENT LIST TYPE normal
    CLIENT LIST TYPE master
    CLIENT LIST TYPE replica
    CLIENT LIST TYPE pubsub
    CLIENT LIST ID 1
    CLIENT LIST ID 1 2 3
    CLIENT LIST TYPE bogus            -> ERR Unknown client type 'bogus'
    CLIENT LIST TYPE                  -> ERR syntax error
    CLIENT LIST ID                    -> ERR syntax error
    CLIENT LIST ID abc                -> ERR Invalid client ID
    CLIENT LIST TYPE normal ID 1      -> ERR syntax error
    CLIENT LIST FOO                   -> ERR syntax error

A new unit test (alongside the existing server_family_test cases) covers each of the above invocations on a server that has at least one normal client, one pubsub subscriber, and one replica connection attached.

vyavdoshenko:
Implements the documented filter grammar:

    CLIENT LIST
    CLIENT LIST TYPE normal|master|replica|slave|pubsub
    CLIENT LIST ID <id> [<id> ...]

Until now, Dragonfly has rejected every argument to the CLIENT LIST with `syntax error`, which broke tooling and libraries (e.g., portalocker.RedisLock) that probe the pubsub-subscribed client set.

The outbound replication socket on a replica is not a facade::Connection on any listener, so it would not appear in CLIENT LIST. This PR reserves a real client ID for each Replica via the same monotonic allocator used by Connection, and formats a matching entry via a new Replica::GetClientInfo() run on the replica proactor.

Changes
- Parser: TYPE / ID grammar; error texts match the spec (`Unknown client type`,
  `Invalid client ID`, `syntax error`). TYPE and ID are mutually exclusive.
- Classifier: replica > pubsub > normal for listener-attached connections.
  kMaster is added separately via the synthetic master-link path.
- facade::Connection::NextClientId() exposes the monotonic id allocator so the
  outbound replica->master link can hold a real, collision-free id.
- Replica reserves client_id_ and creation_time_ at construction (both const).
- Replica::GetClientInfo() formats a CLIENT LIST line for the master link
  (flags=M, real addr/laddr/fd/age/idle, replication phase, byte/record
  counters derived from existing replication state).
- ServerFamily::GetMasterLinkClientInfo() and IsMasterLinkClientId() expose
  the synthetic rows and the id-membership predicate under replicaof_mu_.
- CLIENT KILL ID <master-link-id> now returns
  `Cannot kill master link; use REPLICAOF NO ONE` instead of silently no-oping.
- Fixed an adjacent bug in Replica::GetCurrentPhase() where R_SYNCING was
  shadowed by R_SYNC_OK, making FULL_SYNC_IN_PROGRESS unreachable. The same
  state is already reported correctly through INFO replication, so the two
  surfaces now agree.
- Suppresses the trailing newline on empty CLIENT LIST replies (the old
  always-append broke redis-py's parser on filtered queries that returned
  zero rows).

Tests
- C++: ServerFamilyTest.ClientListAccepted, ClientListRejected (parser table).
- Python: connection_test.py::test_client_list_filters covers TYPE pubsub,
  TYPE normal, TYPE replica/slave aliasing, ID filter hit + miss, empty reply.
- Python: replication_test.py::test_client_list_replication_types covers
  pre-REPLICAOF empty, TYPE replica on master, TYPE master on replica with
  real socket-derived fields, id stability across calls, ID filter match,
  CLIENT KILL ID rejection, and post-REPLICAOF NO ONE teardown.

Fixes #7373
Fixes #5050


src/server/replica.cc
```diff
@@ -1331,17 +1337,60 @@ std::string Replica::GetSyncId() const {
...
+    }
+
+    uint64_t bytes_in = (repl_offs_ >= initial_repl_offs_) ? repl_offs_ - initial_repl_offs_ : 0;
+
+    return StrCat("id=", client_id_, " addr=", addr, " laddr=", laddr, " fd=", fd,
+                  " name=", " tid=", proactor_->GetPoolIndex(),
+                  " irqmatch=0 age=", now - creation_time_, " idle=", idle_sec,
+                  " tot-cmds=", records_applied, " tot-net-in=", bytes_in,
+                  " tot-read-calls=0 tot-dispatches=0 db=0 flags=M phase=",
+                  absl::AsciiStrToLower(GetCurrentPhase()), " lib-name= lib-ver=");
```
dranikpg: it's a little bit hacky that you put it together here - if the format in dragonfly_connection changes, you have to change it here as well, and I doubt anyone will do it. Some common abstraction for both to format or aggregate those things would be helpfull

dranikpg: also I'm not sure that phase can be used like this with that ambiguity

src/server/server_family.cc
```diff
@@ -424,30 +424,121 @@ void ClientInfo(CmdArgList args, CommandContext* cmd_cntx) {
...
+      filter.type = ClientListType::kReplica;
+    else if (val == "PUBSUB")
+      filter.type = ClientListType::kPubsub;
+    else {
+      cmd_cntx->SendError(StrCat("Unknown client type '", ArgS(args, 1), "'"));
+      return nullopt;
+    }
+    return filter;
+  }
+
```
dranikpg: You can use our argument parser here, it will fit nicely and reduce code

src/server/replica.cc
vyavdoshenko: phase-replica field added to avoid ambiguity

vyavdoshenko: fixed. struct and method were added

dranikpg (approved):

PR #7378, opened 2026-05-21
facade: Trim parsed args if capacity much higher than size

abhijat:
On parsing new command, if capacity of backed argument storage is > 2x the size, then trim command to avoid holding onto over sized capacity for connection life time, the storage had grown too high in the past but is not required to be that large any more. 

Also do this only over for storage size over threshold of 256 KiB so as not to churn on small commands (and only correct large allocations).

Slightly related change added to this PR removes use of `clear` to deallocate, as newer versions of absl seem to not deallocate on clear in inlined vector, which will silently break expectations.

https://github.com/abseil/abseil-cpp/blob/master/absl/container/inlined_vector.h#L817-L823

### DIFF
--- a/src/common/backed_args.h
+++ b/src/common/backed_args.h
@@ -16,6 +16,8 @@ class BackedArguments {
   constexpr static size_t kLenCap = 5;
   constexpr static size_t kStorageCap = 88;
 
+  constexpr static size_t kShrinkFloor = 64 << 10;  // 256 KiB
+
  public:
   using value_type = std::string_view;
 
@@ -93,6 +95,13 @@ class BackedArguments {
     // Clear the contents without deallocating memory. clear() deallocates inlined_vector.
     offsets_.resize(0);
     storage_.resize(0);
+  }
+
+  void MaybeShrink() {
+    if (storage_.capacity() > kShrinkFloor && storage_.size() < storage_.capacity() / 2) {
+      storage_.shrink_to_fit();
+      offsets_.shrink_to_fit();
+    }
   }
 
   std::string_view back() const {
--- a/src/facade/parsed_command.cc
+++ b/src/facade/parsed_command.cc
@@ -52,10 +52,11 @@ void ParsedCommand::ResetForReuse() {
   is_deferred_reply_ = false;
   reply_ = std::monostate{};
 
-  offsets_.clear();
+  offsets_.resize(0);
+  offsets_.shrink_to_fit();
   if (HeapMemory() > 1024) {
-    storage_.clear();  // also deallocates the heap.
-    offsets_.shrink_to_fit();
+    storage_.resize(0);
+    storage_.shrink_to_fit();
   }
   ReuseInternal();
 }
--- a/src/facade/resp_srv_parser.cc
+++ b/src/facade/resp_srv_parser.cc
@@ -90,6 +90,7 @@ auto RespSrvParser::Parse(Buffer str, uint32_t* consumed, cmn::BackedArguments* 
     return resultc.first;
   }
 
+  args->MaybeShrink();
   return resultc.first;
 }
 
--- a/src/facade/resp_srv_parser_test.cc
+++ b/src/facade/resp_srv_parser_test.cc
@@ -211,4 +211,28 @@ TEST_F(RespSrvParserTest, InlineReset) {
   EXPECT_EQ(13, consumed_);
 }
 
+static string SetCmd(size_t val_size) {
+  const string val(val_size, 'x');
+  return absl::StrCat("*3\r\n$3\r\nSET\r\n$3\r\nfoo\r\n$", val_size, "\r\n", val, "\r\n");
+}
+
+TEST_F(RespSrvParserTest, ResetForParseChecksFloor) {
+  ASSERT_EQ(RespSrvParser::OK, Parse(SetCmd(10 * 1024)));
+  const size_t grown = args_.HeapMemory();
+  EXPECT_GE(grown, 10 * 1024);
+
+  ASSERT_EQ(RespSrvParser::OK, Parse("*1\r\n$4\r\nPING\r\n"));
+  ASSERT_EQ(RespSrvParser::OK, Parse("*1\r\n$4\r\nPING\r\n"));
+  // not reclaimed if below floor
+  EXPECT_EQ(grown, args_.HeapMemory());
+}
+
+TEST_F(RespSrvParserTest, ResetForParseShrinksAfterSingleCommand) {
+  ASSERT_EQ(RespSrvParser::OK, Parse(SetCmd(512 * 1024)));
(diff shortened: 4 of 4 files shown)


src/facade/resp_srv_parser_test.cc
```diff
@@ -211,4 +211,29 @@ TEST_F(RespSrvParserTest, InlineReset) {
...
+  return absl::StrCat("*3\r\n$3\r\nSET\r\n$3\r\nfoo\r\n$", val_size, "\r\n", val, "\r\n");
+}
+
+TEST_F(RespSrvParserTest, ResetForParseShrinksOversized) {
+  ASSERT_EQ(RespSrvParser::OK, Parse(SetCmd(512 * 1024)));
+  EXPECT_GE(args_.HeapMemory(), 512 * 1024);
+
+  ASSERT_EQ(RespSrvParser::OK, Parse("*1\r\n$4\r\nPING\r\n"));
+  ASSERT_EQ(RespSrvParser::OK, Parse("*2\r\n$4\r\nECHO\r\n$100\r\n" + string(100, 'x') + "\r\n"));
+  EXPECT_TRUE(args_.HeapMemory() >= 100 && args_.HeapMemory() < 200);
```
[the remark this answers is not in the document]
abhijat: changed to 1 kb

src/common/backed_args.h
```diff
@@ -95,6 +97,17 @@ class BackedArguments {
     storage_.resize(0);
   }
 
+  void ResetForParse() {
+    const bool oversized =
+        storage_.capacity() > kShrinkFloor && storage_.size() < storage_.capacity() / 2;
```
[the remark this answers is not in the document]
abhijat: changed to shrink after cmd is parsed

abhijat:
augment review

src/common/backed_args.h
```diff
@@ -95,6 +97,13 @@ class BackedArguments {
     storage_.resize(0);
   }
 
+  void MaybeShrink() {
+    if (storage_.capacity() > kShrinkFloor && storage_.size() < storage_.capacity() / 2) {
```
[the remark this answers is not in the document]
abhijat: out of scope for this change

abhijat:
I am wary of this change having performance implications, sometimes holding onto capacity might be useful, but trimming unused capacity over a 256 KiB threshold should be the correct thing here IMO considering that size has fallen way below capacity.

src/facade/parsed_command.cc
```diff
@@ -52,10 +52,11 @@ void ParsedCommand::ResetForReuse() {
...
 
-  offsets_.clear();
+  offsets_.resize(0);
+  offsets_.shrink_to_fit();
   if (HeapMemory() > 1024) {
-    storage_.clear();  // also deallocates the heap.
-    offsets_.shrink_to_fit();
+    storage_.resize(0);
+    storage_.shrink_to_fit();
   }
```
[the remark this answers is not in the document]
abhijat: this is unchanged behavior, previous version was:

```c++
  offsets_.clear();
  if (HeapMemory() > 1024) {
    storage_.clear();  // also deallocates the heap.
    offsets_.shrink_to_fit();
  }
  ReuseInternal();
}
```

`clear()` dropped memory too. if anything the `clear()` followed by `shrink_to_fit()` might have been redundant earlier.

romange:
I am not against it but would like to understand more. Specifically we also have a thread local pool of parsed commands that we reuse and I want to understand if this is related to the blow up.

romange:
Good catch regarding clear()

src/facade/resp_srv_parser.cc
```diff
@@ -90,6 +90,7 @@ auto RespSrvParser::Parse(Buffer str, uint32_t* consumed, cmn::BackedArguments*
     return resultc.first;
   }
 
+  args->MaybeShrink();
```
romange, line 93: What is the scenario that requires `MaybeShrink call here in addition to `ResetForReuse` fix?

src/common/backed_args.h
```diff
@@ -16,6 +16,8 @@ class BackedArguments {
   constexpr static size_t kLenCap = 5;
   constexpr static size_t kStorageCap = 88;
 
+  constexpr static size_t kShrinkFloor = 256u << 10;  // 256 KiB
```
romange: it's too big. I would reduce it to 64KB

abhijat:
> I am not against it but would like to understand more. Specifically we also have a thread local pool of parsed commands that we reuse and I want to understand if this is related to the blow up.

So far I strongly suspect that the pool does not cause this memory usage but there is only partial evidence to support it:

* lifetime of pool parsed commands is well defined and short, they can get released on `Connection::ReleaseParsedCommand` and are reset before acquiring in `Connection::ParseRedis` for async path. They can go back to pool and live for a long time, but they are reset when reused which will drop their heap allocation.
* I see mostly the pool is used in async path, the metrics where the issue was seen have very low pipeline metrics over the last week:
  - `pipeline_cmd_cache_bytes` is small
  - `pipeline_queue_length` near 0
  - `pipeline_queue_bytes` near 0

It leaves the sync path which uses the `parsed_cmd_` field directly instead of pool, the field is never reset or shrunk directly, and large `connections.direct_bytes` makes me suspect it is the field and not the pool items. In the `Connection::ParseRedis` loop we keep reusing `parsed_cmd_` but it is never reset completely.

src/facade/resp_srv_parser.cc
abhijat: This is the main fix IMO, `ResetForReuse` is never called on the `parsed_cmd_` field of the connection, only on pool objects. This change forces it to be called on `parsed_cmd_` on successful parse

src/common/backed_args.h
abhijat: I set it so high as I got 191 KiB average from the traffic-replay analyze command run on some captured traffic, but yes 64 KiB sounds more reasonable, I will change it.

abhijat: changed

romange:
pls fix tests and after the merge - cherry-pick into v1.38-branch.

romange (approved):

abhijat:
> pls fix tests and after the merge - cherry-pick into v1.38-branch.

created https://github.com/dragonflydb/dragonfly/pull/7386 for cherry-pick

src/common/backed_args.h
```diff
@@ -16,6 +16,8 @@ class BackedArguments {
   constexpr static size_t kLenCap = 5;
   constexpr static size_t kStorageCap = 88;
 
+  constexpr static size_t kShrinkFloor = 64 << 10;  // 256 KiB
```
romange, line 19: you did not update the comment, nor you did not update the PR description

abhijat: missed updating the comment. created https://github.com/dragonflydb/dragonfly/pull/7402 to fix

Issue #7187, opened 2026-04-20
test_big_huge_streaming_restart fails


mkaruza:
https://github.com/dragonflydb/dragonfly/actions/runs/24592826553
https://github.com/dragonflydb/dragonfly/actions/runs/24617535906

```
[31](https://github.com/dragonflydb/dragonfly/actions/runs/24592826553/job/71917014980#step:7:1338)
FAILED 😰  dragonfly/replication_test.py::test_big_huge_streaming_restart[df_factory0] - TimeoutError
```

glevkovich:
Happened again today;
https://github.com/dragonflydb/dragonfly/actions/runs/25038798464/job/73336665141
Logs:

[logs-CI-LARGE-86 (1).zip](https://github.com/user-attachments/files/27158007/logs-CI-LARGE-86.1.zip)

glevkovich:
Failed again here:
https://github.com/dragonflydb/dragonfly/actions/runs/25141695547/job/73692697673

[logs-CI-LARGE-86.zip](https://github.com/user-attachments/files/27330007/logs-CI-LARGE-86.zip)

glevkovich:
failed again here:
https://github.com/dragonflydb/dragonfly/actions/runs/25165895422

[logs-CI-LARGE-86 (1).zip](https://github.com/user-attachments/files/27330015/logs-CI-LARGE-86.1.zip)

glevkovich:
again: https://github.com/dragonflydb/dragonfly/actions/runs/25205905566

abhijat:
https://github.com/dragonflydb/dragonfly/actions/runs/25656211640
https://github.com/dragonflydb/dragonfly/actions/runs/25496623766

abhijat:
only fails on x86, always passes on ARM

abhijat:
I ran the repeat test action https://github.com/dragonflydb/dragonfly/actions/workflows/repeat-tests.yml after changing its matrix (to match large tests - ci large 86 runner) in https://github.com/dragonflydb/dragonfly/pull/7308 to bisect this issue. I suspect it is related to 32f0034474a8f5367d3cb1aedba1ac818172f26c

failure counts show up as below  (counts, SHAs and window suggested by claude based on its analysis of failures using gh command etc)

```
  ┌─────────┬──────────────────┬───────────────┐
  │  Month  │ Heavy Tests runs │ Test failures │
  ├─────────┼──────────────────┼───────────────┤
[...]

PR #7384, opened 2026-05-22
Increase test_big_huge_streaming_restart timeout

dranikpg:
It is indeed somehow connected to #7187, it just makes it slower. I will investigate further, but lets just stop the failures for now


glevkovich (approved):
LGTM!

PR #7385, opened 2026-05-22
feat(server): Implement COMMAND LIST (#5466)

paarth-k2002:
- Implements COMMAND LIST with FILTERBY MODULE (empty array, no module system), FILTERBY ACLCAT (case-insensitive, strips leading @), and FILTERBY PATTERN (GlobMatcher, case-insensitive). Unknown filter types return kSyntaxErr.
- Adds tests for LIST subcommand in server_family_test.cc.
- Updates COMMAND HELP strings to reflect new subcommand signatures.

### DIFF
--- a/src/server/main_service.cc
+++ b/src/server/main_service.cc
@@ -5,7 +5,7 @@
 #include "server/main_service.h"
 
 #include "absl/strings/str_split.h"
-#include "facade/resp_expr.h"
+#include "core/glob_matcher.h"
 #include "util/fibers/detail/fiber_interface.h"
 #include "util/fibers/proactor_base.h"
 #include "util/fibers/synchronization.h"
@@ -28,7 +28,7 @@ extern "C" {
 #include <xxhash.h>
 
 #include <csignal>
-#include <filesystem>
+#include <optional>
 
 #include "base/cycle_clock.h"
 #include "base/flag_utils.h"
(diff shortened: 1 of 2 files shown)


paarth-k2002:
augment review

src/server/server_family_test.cc
```diff
@@ -747,4 +747,95 @@ TEST_F(ServerFamilyTest, MemoryParserErrorHandling) {
   EXPECT_THAT(Run({"MEMORY", "DEFRAGMENT", "not-a-float"}), ErrArg("not a valid float"));
 }
 
+TEST_F(ServerFamilyTest, CommandInfo) {
+  // CASE: Single valid command
+  auto resp = Run({"command", "info", "get"});
+  // Validating size as 7 instead of 1 because Run() flattens a single top-level response array.
+  ASSERT_EQ(resp.GetVec().size(), 7);
```
paarth-k2002: Have added the comment on why this validation is being done with size equaling 7 instead of 1.

src/server/server_family_test.cc
```diff
@@ -747,4 +747,95 @@ TEST_F(ServerFamilyTest, MemoryParserErrorHandling) {
...
+  resp = Run({"command", "info", "get", "set"});
+  ASSERT_EQ(resp.GetVec().size(), 2);
+  ASSERT_EQ(resp.GetVec()[0].type, RespExpr::ARRAY);
+  ASSERT_EQ(resp.GetVec()[0].GetVec().size(), 7);
+  ASSERT_EQ(resp.GetVec()[1].type, RespExpr::ARRAY);
+  ASSERT_EQ(resp.GetVec()[1].GetVec().size(), 7);
+
+  // CASE: Unknown command
+  resp = Run({"command", "info", "nonExistentCommand"});
+  EXPECT_EQ(resp.type, RespExpr::NIL);
```
paarth-k2002: Same reason as the flagged issue above - working of Run().

paarth-k2002:
PR created for fixing the issue #5466 (INFO extension + LIST support).
Also attaching PR #7380 as it fixes the flagged issue by the review bot in the PR. Changes needed in some newly added tests if that PR merges before this one :)

src/server/main_service.cc
```diff
@@ -2767,38 +2767,126 @@ void Service::Command(CmdArgList args, CommandContext* cmd_cntx) {
...
+      if (!absl::EqualsIgnoreCase(ArgS(args, 1), "FILTERBY")) {
+        return rb->SendError(kSyntaxErr, kSyntaxErrType);
+      }
+      filter_type = ArgS(args, 2);
+      filter_value = ArgS(args, 3);
+    }
+
+    std::vector<string_view> matching_cmds;
+
+    if (args.size() == 4) {
```
romange: can you please try simplifiying the code in this function?

for example, why `if (args.size() == 4) {` needed twice?

similarly, can you make that  `registry_.Traverse` is call once in this function and just pass to it the already prepared predicate filter based on the parsing? This will also separate parsing and the execution stages .

paarth-k2002: Have pushed the updated changes. Updated the function for LIST for making it a bit more simpler and combined repeated checks and value usages.

paarth-k2002: Hi @romange, just a quick reminder about this PR. Please let me know if any changes or fixes are needed, and I’ll get right on them. Thanks.

romange:
hello @paarth-k2002  please resolve the conflicts before I review.

paarth-k2002:
Hey @romange,

The test file conflicts have been resolved, and the branch has been rebased with the latest main. Ready for review.

src/server/server_family_test.cc
```diff
@@ -793,4 +793,132 @@ TEST_F(ServerFamilyTest, InfoReplicationMemoryOnlyInMemorySection) {
...
+  EXPECT_EQ(resp.GetVec()[0].GetVec().size(), 7);
+  EXPECT_EQ(resp.GetVec()[1].type, RespExpr::NIL);
+
+  // CASE: Duplicate args
+  resp = Run({"command", "info", "get", "GET"});
+  EXPECT_EQ(resp.GetVec().size(), 2);
+  EXPECT_EQ(resp.GetVec()[0].type, RespExpr::ARRAY);
+  EXPECT_EQ(resp.GetVec()[1].type, RespExpr::ARRAY);
+
+  // CASE: No args
```
romange: I think it's wrong and in valkey `COMMAND INFO` with zero args returns all commands

src/server/main_service.cc
```diff
@@ -2744,38 +2744,117 @@ void Service::Command(CmdArgList args, CommandContext* cmd_cntx) {
...
+          return rb->SendEmptyArray();
+
+        const string_view cat = absl::StripPrefix(filter_value, "@");
+        const auto& rev_table = acl_family_.GetRevTable();
+        auto it = std::find_if(rev_table.begin(), rev_table.end(), [cat](string_view rev_cat) {
+          return absl::EqualsIgnoreCase(rev_cat, cat);
+        });
+
+        auto dist = std::distance(rev_table.begin(), it);
+        if (it == rev_table.end() || dist >= 32)
```
romange: where this constant 32 is coming from?

romange (commented):
small comments, but looks good

src/server/server_family_test.cc
paarth-k2002: Validated on Redis and Valkey and they indeed return the same info for COMMAND and COMMAND INFO with no args. Will correct the behavior and update the PR.

src/server/main_service.cc
paarth-k2002: The 32 is the number of bits in the uint32_t ACL category bitmask. It was an extra defensive check in case the table ever grows beyond 32 entries, preventing undefined behavior from (1u << cat_bit) if that happens. Since the `it == rev_table.end()` check already catches all real invalid categories today, I can remove this guard entirely if you prefer it cleaner. Or replace the fixed number with something like `sizeof(acl::AclCat) * CHAR_BIT` to make it self-documenting for the current changes. Let me know which one would you rather have.

paarth-k2002:
Updated the zero-arg case of COMMAND INFO and updated the ACL CAT comparison. The two commits can be squashed and merged if the PR is fine, else please update if any changes needed. Thanks.

vyavdoshenko:
@paarth-k2002 
Please resolve conflicts; I will review this PR.

paarth-k2002:
Hello @vyavdoshenko

I'll resolve the conflicts and push the changes soon today. Thanks.

paarth-k2002:
@vyavdoshenko 
Updated the PR. The INFO command was already updated so the changes are for the introduction of the LIST command here.

vyavdoshenko:
**Parsing and execution are still interleaved.** The `LIST` branch has two `Traverse` calls with duplicated collect/send loops, eight hand-written `SendError(kSyntaxErr)` sites and a discarded `TakeError()`. `CmdArgParser` is meant to be used as `ExpectTag("FILTERBY")` -> `MapNext(...)` -> `Next()` -> `if (!parser.Finalize()) return rb->SendError(parser.TakeError().MakeReply())`; all those errors already map to `kSyntaxErr`. Build one predicate from the parsed filter, then do a single `Traverse` and a single reply loop for all `LIST` variants.

**Look the category up in `cat_table_`, not the reverse table.** The reverse table is bit->name and `GetRevTable()` is a test-only accessor; name->mask is what `cat_table_` is for and what `ACL CAT` uses. Today `ACL CAT search` returns the FT.* commands while `COMMAND LIST FILTERBY ACLCAT search` returns an empty array, and `_RESERVED` is accepted as a category. A small `AclFamily::CategoryMask(string_view)` helper fixes both and removes `cat_bit` / `1u << cat_bit` altogether.

**The `sizeof(acl::AclCat) * CHAR_BIT` guard is dead.** The reverse table has exactly 32 entries, so `it != end()` already bounds `dist`. And `sizeof(AclCat)` is 4 only by accident - a `1ULL << 32` enumerator would widen the enum to 8 bytes while `acl_categories()` stays `uint32_t`. The real "32" is the width of `acl_categories()`. Drop it (moot after 2), along with the redundant `filter_value.empty()` branch.

PR #7387, opened 2026-05-22
perf(server): reduce ProactorBase::GetMonotonicTimeNs callers

romange:
Step toward deprecating `ProactorBase::GetMonotonicTimeNs` by removing its callers across the replica/server paths.

- **`ProtocolClient::last_io_time_`** — only consumed by `master_last_io_sec` (seconds resolution), so switch to `time(nullptr)` via a protected static `TimeSec()` helper. Storage drops from nanoseconds to seconds and `Summary` loses the `/ 1e9` divide.
- **`engine_shard::RunFPeriodically`** — use `absl::GetCurrentTimeNanos()` for the heartbeat watchdog timing.
- **`EngineShard::CacheStats`** — drop the 1ms throttle (and `last_mem_params_.updated_at` field) that was the only reason to read the monotonic clock here; `CacheStats` is already invoked from a periodic task.
- **`server_family.h`** — refresh stale comment to reference `CycleClock`.

### DIFF
--- a/src/server/engine_shard.cc
+++ b/src/server/engine_shard.cc
@@ -511,12 +511,12 @@ static void RunFPeriodically(std::function<void()> f, std::chrono::milliseconds 
       return;
     }
 
-    int64_t now_ms = fb2::ProactorBase::GetMonotonicTimeNs() / 1000000;
+    int64_t now_ms = absl::GetCurrentTimeNanos() / 1000000;
     if (now_ms - 5 * period_ms.count() > last_heartbeat_ms) {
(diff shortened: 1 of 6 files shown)


dranikpg (approved):
I remember that time queries often show up in profiles (IIRC vdso) which comes mainly from absl::GetCurrentTimeNanos, so the monotinic time caching was worth it

romange:
Now the caching itself shows as well. I will replace it with cycle clock
when needed.

Roman Gershman
CTO
---

PR #7389, opened 2026-05-23
chore(facade): Simplfy & split ioloopv2()

dranikpg:
The IoLoopV2() function became so huge, it needs to be split finally


src/facade/dragonfly_connection.cc
```diff
@@ -2902,12 +2901,184 @@ void Connection::MaybeEnableRecvMultishot() {
...
+           (parsed_head_ && parsed_head_->CanReply()) || !dispatch_q_.empty() || io_ec_ ||
+           IsReadyToMigrate();
+  });
+
+  return {};
+}
+
+// Drains the dispatch queue (control path): processes admin/pubsub messages.
+// Returns true if the caller should re-enter the main loop (i.e. `continue`).
+bool Connection::ProcessDispatchQueue() {
```
glevkovich: function should probably return void, since it always return true

src/facade/dragonfly_connection.cc
```diff
@@ -2902,12 +2901,184 @@ void Connection::MaybeEnableRecvMultishot() {
...
+
+// Parses available input, executes queued commands, and sends replies while
+// respecting pipeline memory backpressure. Parks the fiber when over the limit.
+// Returns the resulting ParserStatus, or an error_code on write failure.
+io::Result<Connection::ParserStatus> Connection::ParseAndExecuteCommands(
+    bool reached_capacity, util::fb2::detail::Waiter* backpressure_waiter) {
+  auto& conn_stats = GetLocalConnStats();
+  QueueBackpressure& qbp = GetQueueBackpressure();
+
+  ParserStatus parse_status;
```
glevkovich: I would initialise to OK.

src/facade/dragonfly_connection.cc
```diff
@@ -2902,12 +2901,184 @@ void Connection::MaybeEnableRecvMultishot() {
 #endif
 }
 
-variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
+// Flushes pending replies and suspends the fiber until new input, command completion,
```
glevkovich: This is redundant as these comments already exist in the header file. Having them in both places increases the risk that they will fall out of sync over time, creating ambiguity.

I suggest removing the documentation comments from the .cc implementation file for all four new methods (FlushAndAwaitInput, ProcessDispatchQueue, ParseAndExecuteCommands, and ParkOnBackpressure) since they repeat the header definitions verbatim.

src/facade/dragonfly_connection.cc
```diff
@@ -2902,12 +2901,184 @@ void Connection::MaybeEnableRecvMultishot() {
...
+    CheckIoBufCapacity(reached_capacity, &io_buf_);
+  }
+
+  return parse_status;
+}
+
+// Parks the fiber until pipeline memory usage drops below the configured limit.
+// Subscribes to the global backpressure EventCount, flushes replies, and awaits relief.
+// Returns a non-zero error_code if the flush failed.
+error_code Connection::ParkOnBackpressure(util::fb2::detail::Waiter* backpressure_waiter) {
```
glevkovich: nit: Consider renaming to `AwaitBackpressureRelief` for consistency with `FlushAndAwaitInput` - both functions suspend the fiber waiting for a condition, so sharing the "Await" verb makes the pattern more uniform.

glevkovich (dismissed):
@dranikpg approved, but  I left  4 comments.

src/facade/dragonfly_connection.cc
dranikpg: Sorry, didn't notice this (asked claude to refactor)

src/facade/dragonfly_connection.cc
dranikpg: Agree

glevkovich (dismissed):
LGTM!

glevkovich:
@dranikpg can we wait with this one till I merge https://github.com/dragonflydb/dragonfly/pull/7437? Then you rebase over it? Its impossible for me to understand the changes, without you rebasing on top of  #pr 7347 after it is merged, there were just too many changes.

PR #7391, opened 2026-05-23
feat(server): Yield fiber periodically during long container iterations

mkaruza:
Add a configurable yield mechanism to container iteration loops so that
long-running commands do  not monopolize the shard fiber for their entire duration.

A new flag `container_iteration_yield_interval_usec` (default: 500µs) controls
the yield interval. When  the running time of the current fiber exceeds the
threshold, `YieldLongIteration()` yields to the scheduler, allowing queued 
commands to be serviced before the scan resumes. Setting the 
flag to 0 disables yielding entirely.

We disable yielding during serialization to not preempt and we don't
yield if encoding is ListPack.

### DIFF
--- a/src/server/journal/cmd_serializer.cc
+++ b/src/server/journal/cmd_serializer.cc
@@ -170,10 +170,13 @@ size_t CmdSerializer::SerializeSet(string_view key, const PrimeValue& pv) {
       max_serialization_buffer_size_);
 
   size_t commands = 0;
-  container_utils::IterateSet(pv, [&](container_utils::ContainerEntry ce) {
-    commands += aggregator.AddArg(ce.ToString());
-    return true;
-  });
+  container_utils::IterateSet(
+      pv,
+      [&](container_utils::ContainerEntry ce) {
+        commands += aggregator.AddArg(ce.ToString());
+        return true;
+      },
+      /*allow_yield=*/false);
 
   // Restore previous time so subsequent operations can trigger lazy expiry.
   pv.SetMemberTime(prev_time);
@@ -194,7 +197,7 @@ size_t CmdSerializer::SerializeZSet(string_view key, const PrimeValue& pv) {
         commands += aggregator.AddArg(ce.ToString());
         return true;
       },
-      /*start=*/0, /*end=*/SIZE_MAX, /*reverse=*/false, /*use_score=*/true);
+      /*start=*/0, /*end=*/SIZE_MAX, /*reverse=*/false, /*use_score=*/true, /*allow_yield=*/false);
   return commands;
 }
 
@@ -209,11 +212,13 @@ size_t CmdSerializer::SerializeHash(string_view key, const PrimeValue& pv) {
 
   size_t commands = 0;
   container_utils::IterateMap(
-      pv, [&](container_utils::ContainerEntry k, container_utils::ContainerEntry v) {
+      pv,
+      [&](container_utils::ContainerEntry k, container_utils::ContainerEntry v) {
         aggregator.AddArg(k.ToString(), CommandAggregator::CommitMode::kNoCommit);
         commands += aggregator.AddArg(v.ToString());
         return true;
-      });
+      },
+      /*allow_yield=*/false);
 
   // Restore previous time so subsequent operations can trigger lazy expiry.
   pv.SetMemberTime(prev_time);
@@ -227,10 +232,13 @@ size_t CmdSerializer::SerializeList(string_view key, const PrimeValue& pv) {
       max_serialization_buffer_size_);
 
   size_t commands = 0;
-  container_utils::IterateList(pv, [&](container_utils::ContainerEntry ce) {
-    commands += aggregator.AddArg(ce.ToString());
-    return true;
-  });
+  container_utils::IterateList(
+      pv,
+      [&](container_utils::ContainerEntry ce) {
+        commands += aggregator.AddArg(ce.ToString());
+        return true;
+      },
+      /*start=*/0, /*end=*/SIZE_MAX, /*allow_yield=*/false);
   return commands;
 }
 
--- a/tests/dragonfly/generic_test.py
+++ b/tests/dragonfly/generic_test.py
@@ -1,12 +1,11 @@
 import logging
-import threading
 import pytest
 import redis
 import asyncio
 from redis import asyncio as aioredis
 
 from . import dfly_multi_test_args, dfly_args
-from .instance import DflyInstance, DflyInstanceFactory, DflyStartException
+from .instance import DflyInstance, DflyStartException
 from .utility import batch_fill_data, gen_test_data, EnvironCntx
 from .seeder import DebugPopulateSeeder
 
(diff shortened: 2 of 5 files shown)


src/server/journal/cmd_serializer.cc
```diff
@@ -170,10 +170,13 @@ size_t CmdSerializer::SerializeSet(string_view key, const PrimeValue& pv) {
...
-    commands += aggregator.AddArg(ce.ToString());
-    return true;
-  });
+  container_utils::IterateSet(
+      pv,
+      [&](container_utils::ContainerEntry ce) {
+        commands += aggregator.AddArg(ce.ToString());
+        return true;
+      },
+      /*allow_yield=*/false);
```
romange, line 179: why we do not allow yielding here? is it assumed to be an atomic cpu execution?

src/server/journal/cmd_serializer.cc
```diff
@@ -194,7 +197,7 @@ size_t CmdSerializer::SerializeZSet(string_view key, const PrimeValue& pv) {
         commands += aggregator.AddArg(ce.ToString());
         return true;
       },
-      /*start=*/0, /*end=*/SIZE_MAX, /*reverse=*/false, /*use_score=*/true);
+      /*start=*/0, /*end=*/SIZE_MAX, /*reverse=*/false, /*use_score=*/true, /*allow_yield=*/false);
```
romange, line 200: same question

tests/dragonfly/generic_test.py
```diff
@@ -344,6 +345,121 @@ async def test_key_bump_ups(df_factory):
         break
 
 
+@pytest.mark.large
```
romange: clearly it's vibe coded. Can you try and simplify the test - it's too long.
also we have internal metrics in dragonfly like `fiber_longrun_usec`, `fiber_longrun_cnt`, `fiber_switch_delay_usec`, `fiber_switch_cnt` that can be tested directly without sending GET requests to prove responsiveness.


In general, I am not against vibe coding but please apply critical thinking when doing it. I do not want that AI will be the one setting our quality levels.

src/server/journal/cmd_serializer.cc
mkaruza: There are from serialization path so preempting is disabled.

tests/dragonfly/generic_test.py
mkaruza: Python test with this metrics is too noisy so i have removed test and added c++ unit test that measure preemption count before/after long iteration.

mkaruza:
### Benchmark - [bench.sh](https://github.com/user-attachments/files/28256519/bench_container_iteration.sh): 

1. Dragonfly launched with 2 threads, 1 shard:

```
taskset -c 12,13 ./dragonfly --conn_use_incoming_cpu --dbfilename= --port 6380 \
  --logtostderr --proactor_threads 2 --num_shards 1 --cluster_mode=emulated
```

2. 5 parallel workers run continuous heavy operations (ZRANGE / ZRANGESTORE / ZDIFFSTORE) while dfly_bench measures GET/SET throughput with fixed parameters:

```
--qps 0, --ratio 1:10, 16 connections
```

Running as 

```
IP=<ip> PORT=<port> ZSET_SIZES="100000 1000000" PIPELINES="1 100" ./bench.sh
```

### Results:

**1. ZSET: 100k members**

| Operation | Pipeline | QPS (yield=0) | QPS (yield=500µs) | ΔQPS | p99 ms (yield=0) | p99 ms (yield=500µs) | Δp99 |
|-----------|----------|---------------|-------------------|------|------------------|----------------------|------|
| ZRANGE | 1 | 196,125 | 194,102 | −1.0% | 9.11 | 6.66 | −27% |
| ZRANGE | 100 | 986,907 | 968,178 | −1.9% | 48.07 | 32.90 |**−32%** |
| ZRANGESTORE | 1 | 12,678 | 34,082 | **+169%** | 117.27 | 57.07 | **−51%** |
| ZRANGESTORE | 100 | 852,116 | 896,362 | +5% | 53.96 | 48.79 | −10% |
| ZDIFFSTORE | 1 | 18,444 | 39,807 | **+116%** | 103.88 | 44.16 | **−57%** |
| ZDIFFSTORE | 100 | 864,416 | 897,633 | +4% | 59.24 | 49.98 | −16% |

**2. ZSET: 1M members**

| Operation | Pipeline | QPS (yield=0) | QPS (yield=500µs) | ΔQPS | p99 ms (yield=0) | p99 ms (yield=500µs) | Δp99 |
|-----------|----------|---------------|-------------------|------|------------------|----------------------|------|
| ZRANGE | 1 | 189,732 | 187,093 | −1.4% | 3.11 | 5.51 | +77%|
| ZRANGE | 100 | 887,691 | 888,016 | ~0% | 50.56 | 33.99 | **−33%** |
| ZRANGESTORE | 1 | 9,412 | 19,671 | **+109%** | 578.15 | 338.89 | **−41%** |
| ZRANGESTORE | 100 | 179,550 | 893,061 | **+397%** | 3,249.71 | 29.42 | **−99%** |
| ZDIFFSTORE | 1 | 12,095 | 24,018 | **+99%** | 613.42 | 430.12 | **−30%** |
| ZDIFFSTORE | 100 | 520,805 | 857,561 | **+65%** | 1,185.60 | 36.16 | **−97%** |

**3. Heavy operation iterations — ZSET: 100k members**

| Operation | Pipeline | Iterations (yield=0) | Iterations (yield=500µs) | Δ |
|-----------|----------|---------------------|--------------------------|---|
| ZRANGE | 1 | 938 | 936 | −0.2% |
| ZRANGE | 100 | 7 | 9 | +28.6% |
| ZRANGESTORE | 1 | 989 | 865 | −12.5% |
| ZRANGESTORE | 100 | 60 | 6 | **−90.0%** |
| ZDIFFSTORE | 1 | 717 | 605 | −15.6% |
| ZDIFFSTORE | 100 | 26 | 6 | **−76.9%** |

**4. Heavy operation iterations — ZSET: 1M members**

| Operation | Pipeline | Iterations (yield=0) | Iterations (yield=500µs) | Δ |
|-----------|----------|---------------------|--------------------------|---|
| ZRANGE | 1 | 54 | 34 | **−37.0%** |
| ZRANGE | 100 | 9 | 5 | **−44.4%** |
| ZRANGESTORE | 1 | 75 | 57 | −24.0% |
| ZRANGESTORE | 100 | 68 | 2 | **−97.1%** |
| ZDIFFSTORE | 1 | 58 | 18 | **−69.0%** |
| ZDIFFSTORE | 100 | 27 | 1 | **−96.3%** |

### Summary

Without yielding, CPU-intensive container iterations monopolize the shard fiber for their entire duration, blocking all other commands from executing. This becomes critical at scale — a single ZRANGESTORE or ZDIFFSTORE over 1M elements can hold the shard for hundreds of milliseconds, causing severe p99 latency spikes for unrelated GET/SET traffic on the same shard.

The impact grows with container size, driven by different factors (number of elements, size of each element).

**The one exception is ZRANGE (pipeline=1, 1M members) where `p99` slightly worsens from `3.11ms → 5.51ms (+77%)`.** ZRANGE is read-only where bottleneck is serializing 1M results back to the client, not shard iteration, so yielding adds per-element overhead without relieving any real contention.

Result from count of `heavy operation` illustrates the trade-off. The iteration count reflects how many times each heavy operation completed within the `30s` window. Enabling yielding deliberately slows down heavy operations — each one pauses every 500µs to let GET/SET commands run — so fewer iterations finish in the same time. This is the intended trade-off: heavy operations yield CPU time to unblock waiting commands.

**Same results should be expected with other types**

romange (approved):

romange (approved):

PR #7396, opened 2026-05-24
fix(facade): Cancel commands on connection disconnect (loopv2)

dranikpg:
With ioloopv2, all "in flight" commands are coroutines that suspended on the `tx->Blocker()` (aka `run_barrier`). When a connection disconnects, we can:
1. Try to cancel all such commands as they pass the conditions of #7399 
2. If it is not possible to cancel them, add them to the orphan list and wait for their completion while letting the connection forget about them and progress further with shutdown

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -13,6 +13,7 @@
 #include <absl/time/time.h>
 
 #include <algorithm>
+#include <boost/intrusive/list.hpp>
 #include <numeric>
 #include <variant>
 
@@ -548,6 +549,47 @@ class PipelineCacheSizeTracker {
 };
 
 thread_local PipelineCacheSizeTracker tl_pipe_cache_sz_tracker;
+
+// An orphaned command is a command whose execution was aborted due to a client disconnect,
+// but which cannot be safely cancelled (due to transactional framework limitations),
+// so we wait for its execution to finish and keep it in a separate list.
+// TODO: Explore more safe ways to cancel a scheduled transaction atomically
+struct OrphanedCommand : public boost::intrusive::list_base_hook<
+                             boost::intrusive::link_mode<boost::intrusive::auto_unlink>> {
+  explicit OrphanedCommand(ParsedCommand* c)
+      : cmd(c),
+        callback(std::bind_front(&OrphanedCommand::OnFinish, this)),
+        waiter(util::fb2::detail::Waiter::Callback(callback)),
+        sub_key(cmd->Blocker()->OnCompletion(&waiter)) {
+  }
+
+  void OnFinish() {
+    sub_key->Drop();
+    sub_key.reset();
+    unlink();
+    delete this;
+  }
+
+  // Takes ownership of cmd: registers a self-deleting waiter, or deletes immediately if done.
+  static void Adopt(ParsedCommand* cmd);
+
+  std::unique_ptr<ParsedCommand> cmd;  // owned by this struct
+
+  std::function<void()> callback;
+  util::fb2::detail::Waiter waiter;
+  std::optional<util::fb2::EventCount::SubKey> sub_key;
+};
+
+using OrphanedList =
+    boost::intrusive::list<OrphanedCommand, boost::intrusive::constant_time_size<false>>;
+thread_local OrphanedList tl_orphaned_commands;
+
+void OrphanedCommand::Adopt(ParsedCommand* cmd) {
+  if (auto* orphan = new OrphanedCommand(cmd); orphan->sub_key)
+    tl_orphaned_commands.push_back(*orphan);
+  else
+    delete orphan;  // already completed
+}
 
 size_t Connection::MessageHandle::UsedMemory() const {
   struct MessageSize {
@@ -1810,8 +1852,20 @@ void Connection::ClearPipelinedMessages() {
     auto* curr{parsed_head_};
     parsed_head_ = parsed_head_->next;
 
-    // Wait for the in-flight async commands processing by consumer to finish before recycling.
+    // For in-flight async commands that haven't completed yet:
     if (curr->IsDeferredReply() && !curr->CanReply()) {
+      if (ioloop_v2_) {
+        // Multi-shard commands are atomic — orphan instead of blocking.
+        size_t used_mem = curr->UsedMemory();
+        curr->next = nullptr;
+        VLOG(1) << "Orphaning in-flight command from closed connection " << id_;
+        OrphanedCommand::Adopt(curr);
+        parsed_cmd_q_len_--;
+        parsed_cmd_q_bytes_ -= used_mem;
+        tl_facade_stats->conn_stats.pipeline_queue_entries--;
+        tl_facade_stats->conn_stats.pipeline_queue_bytes -= used_mem;
+        continue;
+      }
       curr->Blocker()->Wait();
     }
 
@@ -2157,6 +2211,17 @@ Connection::WeakRef Connection::Borrow() {
 
 void Connection::ShutdownThreadLocal() {
   pipeline_req_pool_.clear();
+
+  // Wait at most 10ms for orphaned commands to finish
+  for (size_t i = 0; i < 10 && !tl_orphaned_commands.empty(); i++)
+    util::ThisFiber::SleepFor(1ms);
+
+  // Force delete orphaned commands then
+  if (!tl_orphaned_commands.empty()) {
+    size_t cmds = std::distance(tl_orphaned_commands.begin(), tl_orphaned_commands.end());
+    LOG(WARNING) << "Shutting down with " << cmds << " orphaned commands still pending";
+    tl_orphaned_commands.clear_and_dispose([](OrphanedCommand* o) { delete o; });
+  }
 }
 
 bool Connection::IsCurrentlyDispatching() const {
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -565,7 +565,6 @@ struct OrphanedCommand : public boost::intrusive::list_base_hook<
 
   void OnFinish() {
     sub_key->Drop();
-    sub_key.reset();
     unlink();
     delete this;
   }
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1854,16 +1854,11 @@ void Connection::ClearPipelinedMessages() {
 
     // For in-flight async commands that haven't completed yet:
     if (curr->IsDeferredReply() && !curr->CanReply()) {
+      // For V2 loop we can disown the commands instead of waiting for them
       if (ioloop_v2_) {
-        // Multi-shard commands are atomic — orphan instead of blocking.
-        size_t used_mem = curr->UsedMemory();
+        UnaccountParsedCommand(curr);
         curr->next = nullptr;
-        VLOG(1) << "Orphaning in-flight command from closed connection " << id_;
         OrphanedCommand::Adopt(curr);
-        parsed_cmd_q_len_--;
-        parsed_cmd_q_bytes_ -= used_mem;
-        tl_facade_stats->conn_stats.pipeline_queue_entries--;
-        tl_facade_stats->conn_stats.pipeline_queue_bytes -= used_mem;
         continue;
       }
       curr->Blocker()->Wait();
@@ -2780,18 +2775,8 @@ void Connection::EnqueueParsedCommand(ParsedCommand* cmd) {
 }
 
 void Connection::ReleaseParsedCommand(ParsedCommand* cmd, bool is_pipelined) {
-  size_t used_mem = cmd->UsedMemory();
   auto& conn_stats = tl_facade_stats->conn_stats;
-
-  DCHECK_GT(parsed_cmd_q_len_, 0u);
-  DCHECK_GE(parsed_cmd_q_bytes_, used_mem);
-  DCHECK_GT(conn_stats.pipeline_queue_entries, 0u);
-  DCHECK_GE(conn_stats.pipeline_queue_bytes, used_mem);
-  parsed_cmd_q_len_--;
-  parsed_cmd_q_bytes_ -= used_mem;
-
-  conn_stats.pipeline_queue_entries--;
-  conn_stats.pipeline_queue_bytes -= used_mem;
+  UnaccountParsedCommand(cmd);
 
   if (is_pipelined) {
     conn_stats.pipelined_cmd_cnt++;
@@ -2821,6 +2806,20 @@ void Connection::ReleaseParsedCommand(ParsedCommand* cmd, bool is_pipelined) {
       delete cmd;
     }
   }
+}
+
+void Connection::UnaccountParsedCommand(ParsedCommand* cmd) {
+  size_t used_mem = cmd->UsedMemory();
+  auto& conn_stats = tl_facade_stats->conn_stats;
+
+  DCHECK_GT(parsed_cmd_q_len_, 0u);
+  DCHECK_GE(parsed_cmd_q_bytes_, used_mem);
+  DCHECK_GT(conn_stats.pipeline_queue_entries, 0u);
+  DCHECK_GE(conn_stats.pipeline_queue_bytes, used_mem);
+  parsed_cmd_q_len_--;
+  parsed_cmd_q_bytes_ -= used_mem;
+  conn_stats.pipeline_queue_entries--;
+  conn_stats.pipeline_queue_bytes -= used_mem;
 }
 
 void Connection::DestroyParsedQueue() {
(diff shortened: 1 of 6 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -2111,6 +2159,17 @@ Connection::WeakRef Connection::Borrow() {
...
+  // Wait at most 10ms for orphaned commands to finish
+  for (size_t i = 0; i < 10 && !tl_orphaned_commands.empty(); i++)
+    util::ThisFiber::SleepFor(1ms);
+
+  // Force delete orphaned commands then
+  if (!tl_orphaned_commands.empty()) {
+    size_t cmds = std::distance(tl_orphaned_commands.begin(), tl_orphaned_commands.end());
+    LOG(WARNING) << "Shutting down with " << cmds << " orphaned commands still pending";
+    tl_orphaned_commands.clear_and_dispose([](OrphanedCommand* o) { delete o; });
+  }
```
dranikpg, line 2221: I don't really know what to do there. One option is to move it and place it after we have shut down the shard set.

dranikpg: but there is no real need to deallocate memory if we shut down the process, so I can just remove that forceful part? or both parts even?

glevkovich:
@dranikpg  some general comment minor about your commits:
I see you have 3 commits in this PR but the author name is different in each ( ("Vlad" vs "Vladislav Oleshko"). 
I assume you are working from different machines/vms.

I suggest to update your global Git config on all of your machines to keep them uniform:

```
git config --global user.name "Vladislav Oleshko"
git config --global user.email "vlad@dragonflydb.io"
```

src/facade/dragonfly_connection.cc
```diff
@@ -546,6 +547,47 @@ class PipelineCacheSizeTracker {
 
 thread_local PipelineCacheSizeTracker tl_pipe_cache_sz_tracker;
 
+// An orphaned command is a command whose execution was aborted due to a client disconnect,
+// but which cannot be safely cancelled (due to transactional framework limitations),
+// so we wait for its execution to finish and keep it in a separate list.
+// TODO: Explore more safe ways to cancel a scheduled transaction atomically
```
glevkovich, line 556: nit: "explore safer ways"

romange:
I am having a hard time to understand the particular problem you are solving. Specifically, how OrphanedCommand that you introduce help you to overcome whatever limitations we have deeper in the stack. 
Can you please expand in the PR description what is the problem and how it is solved?

src/facade/dragonfly_connection.cc
```diff
@@ -2111,6 +2159,17 @@ Connection::WeakRef Connection::Borrow() {
 
 void Connection::ShutdownThreadLocal() {
   pipeline_req_pool_.clear();
+
+  // Wait at most 10ms for orphaned commands to finish
+  for (size_t i = 0; i < 10 && !tl_orphaned_commands.empty(); i++)
+    util::ThisFiber::SleepFor(1ms);
+
+  // Force delete orphaned commands then
+  if (!tl_orphaned_commands.empty()) {
```
glevkovich, line 2217: I think you should  add an explicit destructor to OrphanedCommand that drops the SubKey.

struct OrphanedCommand .... {
...
...
~OrphanedCommand() {
    if (sub_key) {
      sub_key->Drop();
    }
  }
...
...
};
This forced-delete will be clearer and more robust if `OrphanedCommand` explicitly drops its SubKey in the destructor before the memory is freed. This makes the lifecycle clearer, acts as defense-in-depth, and ensures the teardown is explicitly safe.

With this addition, code is safer than before, but not yet 100% safe in my opinion, we have the delete this in OnFinish. and you need to deal with;
`// TODO: Explore more safe ways to cancel a scheduled transaction atomically`

src/facade/dragonfly_connection.cc
```diff
@@ -546,6 +547,47 @@ class PipelineCacheSizeTracker {
...
+        callback(std::bind_front(&OrphanedCommand::OnFinish, this)),
+        waiter(util::fb2::detail::Waiter::Callback(callback)),
+        sub_key(cmd->Blocker()->OnCompletion(&waiter)) {
+  }
+
+  void OnFinish() {
+    sub_key->Drop();
+    sub_key.reset();
+    unlink();
+    delete this;
```
glevkovich: This is risky IMHO. Im not claiming its a bug, I'm claiming its a very sensitive code so you should re-consider or research more into it ( I did not).
I asked the AI to draw me the usual call stack ,and explain it:


Let's trace the call stack precisely:
```
Waiter::invoke_callback()
  → std::function::operator()()    ← executing callback member of OrphanedCommand
    → OrphanedCommand::OnFinish()
      → sub_key->Drop()
      → sub_key.reset()
      → delete this                ← destroys `callback` (the std::function) while it's on the call stack
    ← OnFinish returns
  ← std::function::operator()() returns   ← but std::function is now freed!
← Waiter does... what here?
```

**The risk:** After `OnFinish` returns, execution goes back up into the Waiter/EventCount code. Does that code touch the `callback` or `waiter` objects again? We don't know — that depends on helio internals we can't see (the submodule isn't populated in this worktree).

**The mitigating argument:** `Drop()` is called *before* `delete this`, which deregisters the waiter from EventCount under its spinlock. If the EventCount's notification code does nothing after invoking the callback (just releases the spinlock), then `delete this` is safe.

**But we can't prove that without reading helio.** That's exactly the claim — it's not a guarantee of a crash, it's a "this is fragile because it depends on assumptions about the caller."

In practice `delete this` from a `std::function` callback is always a code smell. It works *if* the caller never touches the callable object after the call returns — which is true for simple function dispatch, but not guaranteed for every implementation.

The claim is valid as a **"this is risky and depends on helio internals"** concern, not a guaranteed crash. Worth noting in review but not a blocker.

src/facade/dragonfly_connection.cc
```diff
@@ -2111,6 +2159,17 @@ Connection::WeakRef Connection::Borrow() {
...
+
+  // Wait at most 10ms for orphaned commands to finish
+  for (size_t i = 0; i < 10 && !tl_orphaned_commands.empty(); i++)
+    util::ThisFiber::SleepFor(1ms);
+
+  // Force delete orphaned commands then
+  if (!tl_orphaned_commands.empty()) {
+    size_t cmds = std::distance(tl_orphaned_commands.begin(), tl_orphaned_commands.end());
+    LOG(WARNING) << "Shutting down with " << cmds << " orphaned commands still pending";
+    tl_orphaned_commands.clear_and_dispose([](OrphanedCommand* o) { delete o; });
```
glevkovich, line 2220: This one Poped up later, so adding it as a seperate comment:

Forcefully deleting orphans at shutdown: when we delete o in clear_and_dispose;, we destroy the underlying ParsedCommand and its coroutine. If that coroutine is still suspended mid-flight on a shard thread, destroying it will corrupt the engine's state. This is (another) strong reason to avoid forcefully deleting these commands while they are still pending.

src/facade/dragonfly_connection.cc
```diff
@@ -546,6 +547,47 @@ class PipelineCacheSizeTracker {
 
 thread_local PipelineCacheSizeTracker tl_pipe_cache_sz_tracker;
 
+// An orphaned command is a command whose execution was aborted due to a client disconnect,
+// but which cannot be safely cancelled (due to transactional framework limitations),
+// so we wait for its execution to finish and keep it in a separate list.
+// TODO: Explore more safe ways to cancel a scheduled transaction atomically
+struct OrphanedCommand : public boost::intrusive::list_base_hook<
+                             boost::intrusive::link_mode<boost::intrusive::auto_unlink>> {
+  explicit OrphanedCommand(ParsedCommand* c)
```
glevkovich, line 559: Even if you dont accept https://github.com/dragonflydb/dragonfly/pull/7396/changes#r3312375820, I think that we should anyhow in the general sense have an explicit destructor for OrphanedCommand: because of the complex lifecycle, the struct needs an explicit destructor to ensure sub_key->Drop() is safely called if the object is destroyed abruptly.

src/facade/dragonfly_connection.cc
```diff
@@ -546,6 +547,47 @@ class PipelineCacheSizeTracker {
...
+
+using OrphanedList =
+    boost::intrusive::list<OrphanedCommand, boost::intrusive::constant_time_size<false>>;
+thread_local OrphanedList tl_orphaned_commands;
+
+void OrphanedCommand::Adopt(ParsedCommand* cmd) {
+  if (auto* orphan = new OrphanedCommand(cmd); orphan->sub_key)
+    tl_orphaned_commands.push_back(*orphan);
+  else
+    delete orphan;  // already completed
```
glevkovich: Add DCHECK:
`DCHECK(orphan->cmd->Blocker()->IsCompleted());`

src/facade/dragonfly_connection.cc
```diff
@@ -546,6 +547,47 @@ class PipelineCacheSizeTracker {
...
+    sub_key->Drop();
+    sub_key.reset();
+    unlink();
+    delete this;
+  }
+
+  // Takes ownership of cmd: registers a self-deleting waiter, or deletes immediately if done.
+  static void Adopt(ParsedCommand* cmd);
+
+  std::unique_ptr<ParsedCommand> cmd;  // owned by this struct
```
glevkovich, line 577: Nit: consider adding a comment why this can be the owner instead of the pool `pipeline_req_pool_` . This make sense since connection is already gone, but for other readers it can help.

glevkovich:
@dranikpg , 
I paused my review here because this
`// TODO: Explore more safe ways to cancel a scheduled transaction atomically`
feels like an unresolved correctness issue rather than a future "polish/TODO" item.

The current solution force-deleted orphaned commands during shutdown, but you don't propose any safe way to cancel the scheduled transactions atomically. So this leaves a potential crash risk in during shutdown, even after 10ms wait. I think that fixing this might require major code changes, so I prefer to wait and see before adding more comments.

Also, I read @romange  comemnt , I think that one thing that might help is to expand the PR description a bit?  
Focus on outlining the intended behaviour when orphaned commands haven't completed by the time shutdown runs, whether that's considered:

- temporary limitation you plan to fix later
- an acceptable data loss
- or exactly what the TODO is meant to address.

dranikpg:
> I am having a hard time to understand the particular problem you are solving. Specifically, how OrphanedCommand that you introduce help you to overcome whatever limitations we have deeper in the stack.
Can you please expand in the PR description what is the problem and how it is solved?

There is no stack for in-flight async commands, there is a coroutine that cleans up it's memory when deleted

> feels like an unresolved correctness issue rather than a future "polish/TODO" item.

This is why there is https://github.com/dragonflydb/dragonfly/pull/7399. But not all transactions are or will be cancellable. I will just add a TryCancel check here

src/facade/dragonfly_connection.cc
dranikpg: Just deleting OrphanedCommand currently will fail - and this is why I want to guarantee that it was only deleted with a placed SubKey when OnFinish was called

glevkovich: OK, so  if the only safe delete according to your design is through OnFinish(),  then implementing the "TODO" is the only real fix.

glevkovich:
@dranikpg  - A general comment for PR reviews. Could you please acknowledge review comments as you address them (or after reading one, if no-op)? Either a 👍, a quick reply, or using GitHub's 'Resolve conversation' button works great. Without that, it’s really hard to track what's been fixed versus what might have been missed when new commits are pushed. I just want to make sure we don't accidentally drop anything (and from my experience this happens allot).

Examples of unacknowledged comments:
https://github.com/dragonflydb/dragonfly/pull/7396#discussion_r3312583269
https://github.com/dragonflydb/dragonfly/pull/7396#discussion_r3312534761
https://github.com/dragonflydb/dragonfly/pull/7396#discussion_r3312475421
https://github.com/dragonflydb/dragonfly/pull/7396#discussion_r3312469885
https://github.com/dragonflydb/dragonfly/pull/7396#discussion_r3312420600
https://github.com/dragonflydb/dragonfly/pull/7396#discussion_r3312375820
https://github.com/dragonflydb/dragonfly/pull/7396#discussion_r3311643496

src/facade/dragonfly_connection.cc
```diff
@@ -1810,9 +1851,19 @@ void Connection::ClearPipelinedMessages() {
     auto* curr{parsed_head_};
     parsed_head_ = parsed_head_->next;
 
-    // Wait for the in-flight async commands processing by consumer to finish before recycling.
+    // For in-flight async commands that haven't completed yet:
     if (curr->IsDeferredReply() && !curr->CanReply()) {
-      curr->Blocker()->Wait();
+      if (!ioloop_v2_) {
+        curr->Blocker()->Wait();
+        continue;
```
glevkovich: You are skipping ReleaseParsedCommand: curr is never freed (leak?) and parsed_cmd_q_len_ , parsed_cmd_q_bytes_ are never updated.
I couldn't find ReleaseParsedCommand being called anywhere during an orphan command lifecycle.
 Am I missing something?

src/facade/dragonfly_connection.cc
```diff
@@ -1810,9 +1851,19 @@ void Connection::ClearPipelinedMessages() {
...
+      }
+
+      // Try cancelling the command, if not - let it finish and add it to the orphan list
+      if (!curr->TryCancel()) {
+        UnaccountParsedCommand(curr);
```
glevkovich, line 1853: ReleaseParsedCommand already calls UnaccountParsedCommand.
We get here double accounting for  parsed_cmd_q_len_ , parsed_cmd_q_bytes_ . parsed_cmd_q_len_ will underflow.

Also since the code continues and ownership transferred, ReleaseParsedCommand will free and later will get double-free.

src/server/conn_context.cc
```diff
@@ -301,6 +301,10 @@ bool ConnectionState::ClientTracking::ShouldTrackKeys() const {
   return option_ == OPTIN ? match : !match;
 }
 
+bool CommandContext::TryCancel() {
```
glevkovich, line 304: This is  a segfault risk which with in current code logic never happens - but if a deferred command lacks a transaction context,it will l crash.  Also, tx_ == nullptr initial state is nullptr. I suggest to return:

`tx_ && tx_->CancelScheduledTx();`

Anyway, this check cost nothing.

tests/dragonfly/connection_test.py
```diff
@@ -2165,19 +2165,18 @@ async def flood():
...
+    """With ioloopv2, commands from RST-closed connections must not be dropped.
+
+    Clogger floods MULTI/GET <1 MB key>/EXEC on the Redis port without reading,
+    blocking Send() mid-EXEC while holding the shard lock.  Ghost connections on
+    the memcache port (ioloopv2) send SET on the same key and RST-close while
+    blocked.  The orphan mechanism keeps commands alive until they complete.
     """
     import struct
 
+    mc_port = 11311
```
glevkovich: nit: The memcached port is defined twice here - once in @dfly_args and once here. consider dropping the local mc_port variable and just using` int(df_server["memcached_port"])` or something similar.

src/facade/dragonfly_connection.cc
```diff
@@ -549,6 +550,46 @@ class PipelineCacheSizeTracker {
...
+
+using OrphanedList =
+    boost::intrusive::list<OrphanedCommand, boost::intrusive::constant_time_size<false>>;
+thread_local OrphanedList tl_orphaned_commands;
+
+void OrphanedCommand::Adopt(ParsedCommand* cmd) {
+  if (auto* orphan = new OrphanedCommand(cmd); orphan->sub_key)
+    tl_orphaned_commands.push_back(*orphan);
+  else
+    delete orphan;  // already completed
```
glevkovich: Commenting again this one - it looks like it might have been missed in your latest push, unless you disagree. DCHECK should be added before the delete.
I think it's an important safeguard.


```
void OrphanedCommand::Adopt(ParsedCommand* cmd) {
  if (auto* orphan = new OrphanedCommand(cmd); orphan->sub_key) {
    tl_orphaned_commands.push_back(*orphan);
  } else {
    DCHECK(orphan->cmd->Blocker()->IsCompleted()); // <-- Added here
    delete orphan;  // already completed
  }
}
```

src/facade/dragonfly_connection.cc
dranikpg: True, continue is in the wrong place

src/facade/dragonfly_connection.cc
dranikpg: You are allowed to delete yourself from a member, this happens here as well. I added a Dispatch call which brings it out of scope, but otherwise there is no choice other than adding more queues that delayed delete stuff

src/facade/dragonfly_connection.cc
dranikpg: added

src/facade/dragonfly_connection.cc
dranikpg: I know, this is why I want a discussion about what to do. There are three options:
1. Breaking atomicity with finishing a command half-way
2. Deadlocking or locking for a very long time
3. Corruption via deletion of the underlying command

PR #7399, opened 2026-05-24
fix(server): Cancel transactions

dranikpg:
Try to cancel scheduled + armed transactions safely. We can do this for readonly commands and mutable commands that haven't yet started executing on any shard (otherwise they already touched the data and we can't interrupt them).

### DIFF
--- a/src/server/test_utils.cc
+++ b/src/server/test_utils.cc
@@ -747,9 +747,9 @@ auto BaseFamilyTest::AddFindConn(Protocol proto, std::string_view id) -> TestCon
   return it->second.get();
 }
 
-Transaction* BaseFamilyTest::GetTransaction(string_view id) {
+Transaction* BaseFamilyTest::GetTransaction(string_view conn_id) {
   unique_lock lk(mu_);
-  auto it = connections_.find(id);
+  auto it = connections_.find(conn_id);
   if (it == connections_.end())
     return nullptr;
   return it->second->cmd_cntx()->transaction;
--- a/src/server/test_utils.cc
+++ b/src/server/test_utils.cc
@@ -747,6 +747,14 @@ auto BaseFamilyTest::AddFindConn(Protocol proto, std::string_view id) -> TestCon
   return it->second.get();
 }
 
+Transaction* BaseFamilyTest::GetTransaction(string_view id) {
+  unique_lock lk(mu_);
+  auto it = connections_.find(id);
+  if (it == connections_.end())
+    return nullptr;
+  return it->second->cmd_cntx()->transaction;
+}
+
 vector<string> BaseFamilyTest::StrArray(const RespExpr& expr) {
   CHECK(expr.type == RespExpr::ARRAY || expr.type == RespExpr::NIL_ARRAY);
   if (expr.type == RespExpr::NIL_ARRAY)
--- a/src/server/test_utils.h
+++ b/src/server/test_utils.h
@@ -136,6 +136,7 @@ class BaseFamilyTest : public ::testing::Test {
   }
 
   TestConnWrapper* AddFindConn(Protocol proto, std::string_view id);
+  Transaction* GetTransaction(std::string_view id);
   static std::vector<std::string> StrArray(const RespExpr& expr);
 
   Metrics GetMetrics() const {
--- a/src/server/test_utils.h
+++ b/src/server/test_utils.h
@@ -136,7 +136,7 @@ class BaseFamilyTest : public ::testing::Test {
   }
 
   TestConnWrapper* AddFindConn(Protocol proto, std::string_view id);
-  Transaction* GetTransaction(std::string_view id);
+  Transaction* GetTransaction(std::string_view conn_id);
   static std::vector<std::string> StrArray(const RespExpr& expr);
 
   Metrics GetMetrics() const {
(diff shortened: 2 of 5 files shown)


dranikpg:
augment review

dranikpg:
augment review

src/server/transaction.cc
```diff
@@ -1038,6 +1038,60 @@ void Transaction::Conclude() {
...
+
+  // For mutable transctions we can cancel only if all shards can be stopped.
+  // Reads do not cause any inconsistencies if stopped mid-read, so any condition is ok
+  unsigned disarmed_cnt = disarmed_shards.count();
+  if (disarmed_cnt == unique_shard_cnt_ || is_readonly) {
+    // Cancel the transaction
+    auto is_active = [this](uint32_t i) { return IsActive(i); };
+    shard_set->RunBriefInParallel([this](EngineShard* shard) { CancelShardCb(shard); }, is_active);
+    coordinator_state_ = (coordinator_state_ & ~COORD_SCHED) | COORD_CANCELLED;
+    cb_ptr_.reset();
```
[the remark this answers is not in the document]
dranikpg: good catch actually

src/server/test_utils.h
```diff
@@ -136,6 +136,7 @@ class BaseFamilyTest : public ::testing::Test {
   }
 
   TestConnWrapper* AddFindConn(Protocol proto, std::string_view id);
+  Transaction* GetTransaction(std::string_view id);
```
romange:
```suggestion
  Transaction* GetTransaction(std::string_view conn_id);
```

src/server/test_utils.cc
```diff
@@ -747,6 +747,14 @@ auto BaseFamilyTest::AddFindConn(Protocol proto, std::string_view id) -> TestCon
   return it->second.get();
 }
 
+Transaction* BaseFamilyTest::GetTransaction(string_view id) {
```
romange:
```suggestion
Transaction* BaseFamilyTest::GetTransaction(string_view conn_id) {
```

src/server/test_utils.cc
```diff
@@ -747,6 +747,14 @@ auto BaseFamilyTest::AddFindConn(Protocol proto, std::string_view id) -> TestCon
...
+  unique_lock lk(mu_);
+  auto it = connections_.find(id);
```
romange:
```suggestion
  auto it = connections_.find(conn_id);
```

src/server/transaction.cc
```diff
@@ -1038,6 +1038,66 @@ void Transaction::Conclude() {
   Execute(std::move(cb), true);
 }
 
+bool Transaction::CancelScheduledTx() {
+  DCHECK(coordinator_state_ & COORD_SCHED);
+  bool is_readonly = cid_->IsReadOnly();
+
+  // First, try to clear all is_armed flags
+  std::bitset<1024> disarmed_shards;
+  IterateActiveShards([&](auto& sd, ShardId i) {
```
romange, line 1051: I feel we can have all sorts of bugs for multi-shard transactions, mutating transactions, global transactions.
I know you tried to catch most cases but since it's a transaction code, I suggest we limit it for now for `is_readonly` only AND single-shard AND !GLobal() and leave the TODO to extend to more cases.

src/server/transaction.cc
```diff
@@ -1038,6 +1038,66 @@ void Transaction::Conclude() {
   Execute(std::move(cb), true);
 }
 
+bool Transaction::CancelScheduledTx() {
+  DCHECK(coordinator_state_ & COORD_SCHED);
```
romange, line 1046: Are you sure we won't race with ScheduleInternal call that sets `COORD_SCHED` bit at the end of the call?

dranikpg: it's all called from the coordinator thread, so there can't be racy operations on those coordinator-local variables

src/server/transaction.cc
dranikpg: Every shard is indepdendent. If we correctly handle the case for one shard (scheduled in the most generic way), then we likely handle it correctly also for two. Added a !IsGlobal check at the top to not even try for those unsafe commands

src/server/transaction.cc
romange: I am not talking about cross thread race. ScheduleInternal is a preemptable operation.

Suppose it is being blocked on waiting for its shard callbacks. The connection is blocked on Schedule.

Meanwhile another fiber on the same thread calls CancelScheduledTx. Since ScheduleInternal sets `COORD_SCHED` at the end, CancelScheduledTx will dcheck fail here.

dranikpg: Ah, yes, the comment mentions that it's only for commands that have already scheduled. For V2 loop commands we know that they already have scheduled, so we can call it safely. Otherwise it's really diffcult because you need to factor in every suspension point and not break the Schedule function on its own

romange: How can you know for V2 loop that the commands have been scheduled? How you do you recognize such transactions externally?

romange:
```
Must be called from the coordinator thread after DispatchHop() but before run_barrier_.Wait().
```

how you ensure this in V2?

dranikpg: I will call it only in V2 loop for now and because V2 loop has only _one fiber_ and yields via coroutines in a very specific state, all enqueued commands must be in this state. V2 is just a single fiber with full control over the await(i.e. suspension) points

romange: Ok, I trust you understand what you are doing

romange (approved):

PR #7406, opened 2026-05-25
feat(compact_object): extend Huffman encoding up to 16KB with varint header

romange:
## Summary

- Raises `kMaxHuffLen` from 288 bytes → 16 KB so larger strings benefit from Huffman compression
- Switches the on-blob size-delta header to a varint: 1 byte for delta ≤ 127, 2 bytes (15-bit delta) otherwise
- Reshuffles `ExternalPtr` bitfields to cache both header bytes (still 16 bytes total)
- Hoists `kMaxHuffLen` to a public `CompactObj` constant and reuses it in `debug compression` histogram sampling (bumped from 512 → 16 KB per string) so the trained table reflects what `EncodeString` will see
- Adds a `HuffmanVarintHeader` test that sweeps input lengths across the 1-byte/2-byte boundary and verifies both forms are exercised; existing `Huffman` test extended through 16 KB

The encoded format is in-memory only — RDB, replication and tiered storage go through `GetSlice`/`GetString`, so persisted data is unaffected.

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -52,6 +52,47 @@ namespace {
 constexpr XXH64_hash_t kHashSeed = 24061983;
 constexpr size_t kAlignSize = 8u;
 
+// Maximum size of the HUFFMAN_ENC varint size-delta header, in bytes.
+constexpr unsigned kMaxHuffHeaderSize = 2;
+
+// HUFFMAN_ENC varint size-delta header layout:
+//   * (b0 & 0x80) == 0 : 1-byte header; delta = b0 (0..127).
+//   * (b0 & 0x80) != 0 : 2-byte header; delta = ((b0 & 0x7F) << 8) | b1 (0..32767),
+//                        i.e. 7 high bits in byte 0 and 8 low bits in byte 1.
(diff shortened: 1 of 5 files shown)


src/core/compact_object.cc
```diff
@@ -1582,30 +1638,47 @@ void CompactObj::EncodeString(string_view str) {
...
+        if (delta < 128) {
+          tl.tmp_buf[1] = static_cast<uint8_t>(delta);
+          header_len = 1;
+        } else {
+          tl.tmp_buf[0] = static_cast<uint8_t>(0x80 | ((delta >> 8) & 0x7F));
+          tl.tmp_buf[1] = static_cast<uint8_t>(delta & 0xFF);
+          header_len = 2;
+        }
+        size_t encoded_size = header_len + dest_len;
         // we accept huffman encoding only if it is:
```
dranikpg: I made so much efforts trying to keep this function simple, because all the duplicated endocing and decoding code is really difficult to manage and read

romange: I will simplify it.

dranikpg (approved):

PR #7407, opened 2026-05-25
chore(server): Make listpack bounds configurable

dranikpg:
Make listpack bounds configurable (AFAIR it applies only to hashes)


mkaruza (approved):
Approved. Just curious what should be use case to have this ?

dranikpg:
> Approved. Just curious what should be use case to have this ?

Change settings to allow listpacks for tiering of hashes

PR #7412, opened 2026-05-26
core: zero-copy GET primitives

romange:
Foundational `src/core` + `docs` changes for the zero-copy GET feature.
Adds the data-structure primitives; reply-builder integration and
`CmdGet` wiring ship as follow-ups. No new callers in this PR.

See `docs/zero_copy_get.md` for the full design, lifetimes, and race
analysis.

## Changes

- `detail::LargeString`: new `read_pending : 1` bit; `SetString`/`Free`
  defer deallocation via a shared `ReleasePtr` helper when the bit is
  set (CoW). `SetString` now requires `!s.empty()`; `DefragIfNeeded`
  bails when pinned; `AppendString` `DCHECK`s `!read_pending`.
- `detail::PendingRead { void* ptr; atomic<uint32> refcnt; bool orphaned; UnpinRead(); }`.
- `CompactObj::TryGetRaw() → optional<RawBorrow>` — sole public entry
  for borrowers. Returns `{encoded, decoded_size, encoding, pin}` for
  `NONE`/`ASCII1`/`ASCII2` large strings, `nullopt` otherwise. Stamps
  `read_pending` and registers the pin internally.
- `CompactObj::DrainPendingReads()` — reaps `refcnt==0` entries from
  the thread-local pin map; called from shard heartbeat.
- Thread-local pin map (`absl::flat_hash_map<void*, PendingRead*>` on
  the existing `TL` struct in `compact_object.cc`). No MPSC queue —
  drain iterates the map to avoid a double-push race on re-pin.

## Tests (`compact_object_test.cc`)

`TryGetRaw_NulloptForNonLargeStr`, `TryGetRaw_NoneEncLargeString`,
`TryGetRaw_AsciiLargeString`, `TryGetRaw_PinRefcountAndDrain`,
`TryGetRaw_CowOnMutation`, `TryGetRaw_DefragSkipsWhenPinned`.

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -569,7 +569,9 @@ namespace detail {
 // map lookup either finds the active pin or falls through to deallocate.
 // Precondition: ls->ptr != nullptr.
 static void ReleasePtr(LargeString* ls, LargeString::MemoryResource* mr) {
-  DCHECK(ls->ptr != nullptr);
+  if (!ls->ptr)
+    return;
+
   bool orphaned = ls->read_pending && tl.pin_map.Orphan(ls->ptr);
   if (!orphaned)
     mr->deallocate(ls->ptr, 0, kAlignSize);
@@ -593,8 +595,6 @@ void LargeString::SetString(string_view s, MemoryResource* mr) {
   // (b) read_pending is set (in-place overwrite would corrupt pinned readers;
   // ReleasePtr hands the old buffer to its PendingRead).
   if (s.size() > zmalloc_size(ptr) || read_pending) {
-    if (ptr)
-      ReleasePtr(this, mr);
     ptr = mr->allocate(s.size(), kAlignSize);
   }
   memcpy(ptr, s.data(), s.size());
@@ -617,8 +617,6 @@ void LargeString::AppendString(string_view s, MemoryResource* mr) {
 }
 
 void LargeString::Free(MemoryResource* mr) {
-  if (!ptr)
-    return;
   ReleasePtr(this, mr);
   sz = 0;
 }
@@ -669,7 +667,9 @@ void CompactObj::DrainPendingReads() {
 }
 
 void CompactObj::BorrowedString::Unpin() noexcept {
-  DCHECK(pin_ != nullptr);
+  if (!pin_)
+    return;
+
   static_cast<PendingRead*>(pin_)->UnpinRead();
   pin_ = nullptr;
 }
@@ -1146,15 +1146,10 @@ std::optional<CompactObj::BorrowedString> CompactObj::TryBorrow() const {
     return std::nullopt;
   if (encoding_ != NONE_ENC && encoding_ != ASCII1_ENC && encoding_ != ASCII2_ENC)
     return std::nullopt;
-  auto view = u_.large_str.AsView();
-  size_t decoded_size;
-  if (encoding_ == NONE_ENC) {
-    decoded_size = view.size();
-  } else {
-    // ASCII1_ENC / ASCII2_ENC: derive decoded size from packed length and
-    // first byte (StrEncoding::DecodedSize knows the rounding semantics).
-    decoded_size = GetStrEncoding().DecodedSize(view.size(), *(uint8_t*)view.data());
-  }
+
+  string_view view = u_.large_str.AsView();
+  size_t decoded_size = Size();
+
   // Register the pin and stamp read_pending. The bit is bookkeeping; safe to
   // mutate via const_cast.
   PendingRead* pin = tl.pin_map.RegisterPin(view.data());
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -4,10 +4,10 @@
 
 #include "core/compact_object.h"
 
+#include <atomic>
+
 // #define XXH_INLINE_ALL
 #include <xxhash.h>
-
-#include <array>
 
 #include "core/stream_node.h"
 
@@ -20,6 +20,7 @@ extern "C" {
 #include "redis/util.h"
 #include "redis/zmalloc.h"  // for non-string objects.
 }
+#include <absl/container/flat_hash_map.h>
 #include <absl/strings/str_cat.h>
 #include <absl/strings/strip.h>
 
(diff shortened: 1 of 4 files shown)


romange:
## Benchmark: zero-copy GET impact on CPU and squashing reply memory

**Setup**
- Instance: `c6gn.4xlarge` (16 vCPUs)
- Workload: 64 KB binary string values, read-only
- Client: `memtier_benchmark --test-time=10 --ratio 0:1 --key-maximum=1000 --distinct-client-seed --hide-histogram --pipeline=100`

**Results**

| Metric | `--get_zero_copy=false` | `--get_zero_copy=true` | Delta |
|---|---:|---:|---:|
| Top CPU (server) | 950% | 330% | **−65%** (~2.9× lower) |
| Peak `commands_squashing_replies_bytes` | ~1 GB | ~700 KB | **~1500× lower** |
| Throughput (Ops/sec) | _unchanged_ | _unchanged_ | — |
| Latency (avg / p50 / p99) | _unchanged_ | _unchanged_ | — |

**Why throughput and latency are unchanged**

Both runs are **network/wire-bound at the 64 KB × pipeline=100 workload**, not CPU-bound. The wire is saturated at the same byte rate regardless of how much CPU is spent producing those bytes, so throughput and latency converge. The optimization shows up as **server-side headroom** instead: the same workload now runs at ~330% CPU instead of ~950%, freeing 6+ vCPUs for additional concurrent clients, other commands, replication, snapshotting, etc.

In other words, on a network-bottlenecked workload the gain is **capacity**, not **speed** — measured in CPU and memory, not Ops/sec. To observe a throughput gain, the workload has to be CPU-bound (smaller values, faster network, more concurrent clients, or all of the above).

**Takeaways**

- The zero-copy path eliminates the per-GET `pv.ToString()` allocation+memcpy (~64 KB each) plus the squasher's `CapturingReplyBuilder` copy. Under pipelined load this is the dominant CPU cost on the GET path.
- `commands_squashing_replies_bytes` collapses from ~1 GB to ~700 KB because captured borrowed views report ~40 B per entry (variant overhead only) instead of `40 B + value_size`. This effectively removes the per-batch reply-buffer footprint as a memory cost.

src/core/compact_object.h
```diff
@@ -450,6 +527,11 @@ class CompactObj {
   static Stats GetStatsThreadLocal();
   static void InitThreadLocal(MemoryResource* mr);
 
+  // Iterate the thread-local pin map and reap entries with refcnt==0:
+  // free orphaned buffers, erase the map slot. Typically called from
+  // EngineShard::Heartbeat.
+  static void DrainPendingReads();
```
BorysTheDev, line 530: I don't see usage of this method in the code, only in the tests

src/core/compact_object.h
```diff
@@ -242,6 +248,77 @@ class CompactObj {
...
 
+  // Borrowed view of the underlying LargeString storage, including the
+  // encoding metadata needed to decode it and the registered read pin.
+  // Returned by TryBorrow().
+  //
+  // For NONE_ENC: `encoded` is the user-visible bytes; the reply can stream
+  // them directly. For ASCII1/ASCII2_ENC: `encoded` is the packed source
+  // (`encoded.size() < decoded_size`) and the reply must decode in chunks.
+  //
+  struct BorrowedString {
```
BorysTheDev, line 259: maybe move to a separate file?

src/core/compact_object.h
```diff
@@ -242,6 +248,77 @@ class CompactObj {
...
+    BorrowedString(BorrowedString&& o) noexcept {
+      MoveFrom(std::move(o));
+    }
+
+    BorrowedString& operator=(BorrowedString&& o) noexcept {
+      if (this == &o)
+        return *this;
+      if (pin_)
+        Unpin();
+      MoveFrom(std::move(o));
```
BorysTheDev, line 269: you can use std::swap and get all this behaviour with one line

src/core/compact_object.h
```diff
@@ -242,6 +248,77 @@ class CompactObj {
...
+      return *this;
+    }
+
+    ~BorrowedString() noexcept {
+      if (pin_)
+        Unpin();
```
BorysTheDev: I would prefer to have if(pin_) inside Unpin method instead of DCHECK

src/core/compact_object.cc
```diff
@@ -1027,6 +1141,28 @@ void CompactObj::AppendString(std::string_view str) {
...
+    return std::nullopt;
+  auto view = u_.large_str.AsView();
+  size_t decoded_size;
+  if (encoding_ == NONE_ENC) {
+    decoded_size = view.size();
+  } else {
+    // ASCII1_ENC / ASCII2_ENC: derive decoded size from packed length and
+    // first byte (StrEncoding::DecodedSize knows the rounding semantics).
+    decoded_size = GetStrEncoding().DecodedSize(view.size(), *(uint8_t*)view.data());
+  }
```
BorysTheDev: can we make a function to get the size, it looks like common logic

src/core/compact_object.cc
```diff
@@ -521,20 +609,24 @@ void LargeString::ReserveString(size_t size, MemoryResource* mr) {
...
+  // In-place append would corrupt pinned readers. Today's only caller
+  // (rdb_load) never produces pinned values.
+  DCHECK(!read_pending);
   memcpy(reinterpret_cast<uint8_t*>(ptr) + sz, s.data(), s.size());
   sz += s.size();
 }
 
 void LargeString::Free(MemoryResource* mr) {
   if (!ptr)
     return;
```
BorysTheDev: I prefer to move this check inside ReleasePtr to have the similar guarantees to delete() method and we can drop DCHECK

src/core/compact_object.cc
```diff
@@ -420,13 +421,85 @@ struct Huffman {
...
+  }
+  return it->second;
+}
+
+bool PinnedMap::Orphan(const void* ptr) {
+  auto it = pin_map_.find(ptr);
+  if (it == pin_map_.end())
+    return false;
+  it->second->orphaned = true;
+  return true;
```
BorysTheDev, line 475: return pin_map_.contains(ptr) ? (it->second->orphaned = true) : false;

BorysTheDev (dismissed):

src/core/compact_object.h
romange: yes, it's TBD

src/core/compact_object.h
romange: we can do it afterwards.

src/core/compact_object.h
romange: remind me how?

src/core/compact_object.cc
romange: `it` is defined via find.

src/core/compact_object.h
BorysTheDev: sorry not one line, you have several fields, so you just need to swap all fields; Or you can do it manually with std::tie
```
auto tmp = tuple{field1, ..., fieldn};
std::tie(field1, ..., fieldn) = std::tie(o.field1, ..., 0.fieldn);
std::tie(o.field1, ..., 0.fieldn) = tmp;
```

src/core/compact_object.cc
romange: ok, removed.

src/core/compact_object.h
BorysTheDev: or better implement swap method and call it without any checks, all destruction will be done in destructor autmatically for temporary object

src/core/compact_object.cc
BorysTheDev: but looks simpler and only one line

src/core/compact_object.h
romange: I think current approach is good enough :)

src/core/compact_object.cc
romange: I prefer having a simpler code than relying on random side effect of `it->second->orphaned = true`  expression.

BorysTheDev (approved):

PR #7415, opened 2026-05-26
refactor: ReplicaInfo into class

BorysTheDev:
related to: #7348

• Convert ReplicaInfo from struct to class with encapsulation
• Replace direct member access with getter/setter methods
• Add thread-safe atomic access for version and id fields

### DIFF
--- a/src/server/dflycmd.h
+++ b/src/server/dflycmd.h
@@ -116,18 +116,18 @@ class DflyCmd {
     }
 
     // Immutable after construction; safe to read without locking.
-    const std::string& Address() const {
+    const std::string& GetAddress() const {
       return address_;
     }
-    uint32_t ListeningPort() const {
+    uint32_t GetListeningPort() const {
       return listening_port_;
     }
 
     // Returns the replica ID, or an empty view if SetId has not been called.
     // Thread-safe: id is written at most once via SetId, with release/acquire
     // ordering enforced by id_set_.
-    std::string_view Id() const {
-      if (id_set_.load(std::memory_order_acquire)) {
+    std::string_view GetId() const {
+      if (id_set_.load(std::memory_order_relaxed)) {
         return id_;
       }
       return {};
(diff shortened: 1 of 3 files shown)


src/server/dflycmd.h
```diff
@@ -102,32 +104,115 @@ class DflyCmd {
...
+    }
+    uint32_t ListeningPort() const {
+      return listening_port_;
+    }
+
+    // Returns the replica ID, or an empty view if SetId has not been called.
+    // Thread-safe: id is written at most once via SetId, with release/acquire
+    // ordering enforced by id_set_.
+    std::string_view Id() const {
+      if (id_set_.load(std::memory_order_acquire)) {
```
kostasrim: relaxed memory ordering suffices here and you don't need acquire/release semantics. 

The reason is that once a thread observes a value, let's call it X of an atomic variable V, a subsequent load on that variable is *guaranteed* to be X or any other value that came after X in the modification order of variable V. And because `id_set_` is only set once, it is eventual consistent.

We had a tiny race here in the past which you just fixed, but even then an empty id was acceptable during start.

src/server/dflycmd.h
```diff
@@ -102,32 +104,115 @@ class DflyCmd {
...
+    bool AllFlowsConnected() const {
+      return std::ranges::all_of(flows_, [](const FlowInfo& flow) { return flow.conn != nullptr; });
+    }
+
+    // Returns the per-replica mutex. Callers acquire it via util::fb2::LockGuard
+    // (exclusive), dfly::SharedLock (shared), or std::shared_lock(..., std::try_to_lock)
+    // (try-shared). Returned by reference rather than a factory-returned scoped lock
+    // because clang's -Wthread-safety-analysis cannot track ownership transfer
+    // through a factory function's return value.
+    util::fb2::SharedMutex& GetMutex() {
```
kostasrim, line 194: Are there any operations we could cover via a function here such that we avoid leaking the internal Mutex() to external callers ? I understand for some cases we can't but I was wandering if there are operations that can hide away the manual complexity of locking the mutex outside the scope of the class you just added.

kostasrim (commented):
lgtm small comments

src/server/dflycmd.h
romange: yeah, we could just return fb2::LockGuard or fb2::SharedLock by value.

In any case it's a preparation PR, maybe the next one will reduce the friction?

src/server/dflycmd.h
```diff
@@ -102,32 +104,115 @@ class DflyCmd {
...
-          flows{flow_count} {
+        : replica_state_{SyncState::PREPARATION},
+          exec_st_{std::move(err_handler)},
+          address_{std::move(address)},
+          listening_port_(listening_port),
+          flows_{flow_count} {
+    }
+
+    // Immutable after construction; safe to read without locking.
+    const std::string& Address() const {
```
romange: naming conventions: either use GetFoo() or only for trivial accessors like here you can name `address() const { return address_;}`.

if your function does something more than just returning - it should be `GetFoo`

src/server/dflycmd.h
```diff
@@ -102,32 +104,115 @@ class DflyCmd {
...
+      return address_;
+    }
+    uint32_t ListeningPort() const {
```
romange: same: `listening_port()` or `GetListeningPort`.

src/server/dflycmd.h
BorysTheDev: We can not return the lock, as I wrote in the comment clang thread analysis doesn't allow. It was the first thing that I have tried

romange (approved):

PR #7417, opened 2026-05-26
chore: Add python tests for concurrent QList node materialization

### LINKED ISSUE #7403: List tiering crash
```
0x0000555555b7772d in _mi_page_malloc_zero (heap=heap@entry=0x5e9081a0400, page=page@entry=0x5e90a000180, size=size@entry=3322, zero=zero@entry=false)
    at /home/dranikpg/dragonfly/build-opt/third_party/mimalloc2/src/alloc.c:44
44        page->free = mi_block_next(page, block);
(gdb) bt
#0  0x0000555555b7772d in _mi_page_malloc_zero (heap=heap@entry=0x5e9081a0400, page=page@entry=0x5e90a000180, size=size@entry=3322, zero=zero@entry=false)
    at /home/dranikpg/dragonfly/build-opt/third_party/mimalloc2/src/alloc.c:44
#1  0x0000555555b8bc50 in _mi_malloc_generic (heap=0x5e9081a0400, size=size@entry=3322, zero=zero@entry=false, huge_alignment=huge_alignment@entry=0)
    at /home/dranikpg/dragonfly/build-opt/third_party/mimalloc2/src/page.c:1041
#2  0x0000555555b787b1 in _mi_heap_malloc_zero_ex (zero=false, huge_alignment=0, heap=<optimized out>, size=3322)
    at /home/dranikpg/dragonfly/build-opt/third_party/mimalloc2/src/alloc.c:184
#3  _mi_heap_malloc_zero (heap=<optimized out>, size=3322, zero=false) at /home/dranikpg/dragonfly/build-opt/third_party/mimalloc2/src/alloc.c:197
#4  mi_heap_malloc (heap=<optimized out>, size=3322) at /home/dranikpg/dragonfly/build-opt/third_party/mimalloc2/src/alloc.c:201
#5  _mi_heap_realloc_zero (heap=<optimized out>, p=0x5e9084b0c00, newsize=3322, zero=false) at /home/dranikpg/dragonfly/build-opt/third_party/mimalloc2/src/alloc.c:269
#6  mi_heap_realloc (heap=<optimized out>, p=p@entry=0x5e9084b0c00, newsize=newsize@entry=3322) at /home/dranikpg/dragonfly/build-opt/third_party/mimalloc2/src/alloc.c:290
#7  0x0000555555b769da in zrealloc_usable (ptr=ptr@entry=0x5e9084b0c00, size=size@entry=3322, usable=usable@entry=0x5e909bd20a0)
    at /home/dranikpg/dragonfly/src/redis/zmalloc_mi.c:81
#8  0x0000555555b76a2e in zrealloc (ptr=ptr@entry=0x5e9084b0c00, size=size@entry=3322) at /home/dranikpg/dragonfly/src/redis/zmalloc_mi.c:54
#9  0x0000555555b6364b in lpInsert (lp=<optimized out>, elestr=elestr@entry=0x0, eleint=<optimized out>, eleint@entry=0x0, size=size@entry=0, p=<optimized out>, 
    where=<optimized out>, where@entry=2, newp=0x0) at /home/dranikpg/dragonfly/src/redis/listpack.c:935
#10 0x0000555555b6407d in lpDelete (lp=<optimized out>, p=<optimized out>, newp=<optimized out>) at /home/dranikpg/dragonfly/src/redis/listpack.c:1060
#11 0x0000555555a29052 in dfly::QList::DelPackedIndex (this=this@entry=0x5e908460140, node=node@entry=0x5e908460aa0, 
    p=p@entry=0x5e9084b18f9 "\377\357{\"body\": \"W1siYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFh"...) at /home/dranikpg/dragonfly/src/core/qlist.cc:1246
#12 0x0000555555a292b4 in dfly::QList::Pop[abi:cxx11](dfly::QList::Where) (this=0x5e908460140, where=where@entry=dfly::QList::TAIL)
    at /home/dranikpg/dragonfly/src/core/qlist.cc:624
#13 0x0000555555919400 in operator()<dfly::QList> (__closure=<optimized out>, list=...) at /home/dranikpg/dragonfly/src/server/list_family.cc:230
#14 operator()<dfly::QList> (__closure=<optimized out>, s=<optimized out>) at /home/dranikpg/dragonfly/src/server/list_family.cc:129
#15 std::__invoke_impl<std::__cxx11::basic_string<char>, absl::lts_20250512::Overload<dfly::(anonymous namespace)::ListWrapper::VisitMut<dfly::(anonymous namespace)::ListWrapper::Pop(dfly::QList::Where)::<lambda(auto:99&)> >(dfly::(anonymous namespace)::ListWrapper::Pop(dfly::QList::Where)::<lambda(auto:99&)>)::<lambda(auto:97*)>, dfly::(anonymous namespace)::ListWrapper::Pop(dfly::QList::Where)::<lambda(auto:99&)> >, dfly::QList*&> (__f=...) at /usr/include/c++/15.2.1/bits/invoke.h:63
#16 std::__invoke<absl::lts_20250512::Overload<dfly::(anonymous namespace)::ListWrapper::VisitMut<dfly::(anonymous namespace)::ListWrapper::Pop(dfly::QList::Where)::<lambda(auto:99&)> >(dfly::(anonymous namespace)::ListWrapper::Pop(dfly::QList::Where)::<lambda(auto:99&)>)::<lambda(auto:97*)>, dfly::(anonymous namespace)::ListWrapper::Pop(dfly::QList::Where)::<lambda(auto:99&)> >, dfly::QList*&> (__fn=...) at /usr/include/c++/15.2.1/bits/invoke.h:99
#17 std::__detail::__variant::__gen_vtable_impl<std::__detail::__variant::_Multi_array<std::__detail::__variant::__deduce_visit_result<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > (*)(absl::lts_20250512::Overload<dfly::(anonymous namespace)::ListWrapper::VisitMut<dfly::(anonymous namespace)::ListWrapper::Pop(dfly::QList::Where)::<lambda(auto:99&)> >(dfly::(anonymous namespace)::ListWrapper::Pop(dfly::QList::Where)::<lambda(auto:99&)>)::<lambda(auto:97*)>, dfly::(anonymous namespace)::ListWrapper::Pop(dfly::QList::Where)::<lambda(auto:99&)> >&&, std::variant<dfly::QList*, dfly::detail::ListPack>&)>, std::integer_sequence<long unsigned int, 0> >::__visit_invoke (
    __visitor=..., __vars#0=std::variant [index 0] = {...}) at /usr/include/c++/15.2.1/variant:1057
#18 std::__do_visit<std::__detail::__variant::__deduce_visit_result<std::__cxx11::basic_string<char> >, absl::lts_20250512::Overload<dfly::(anonymous namespace)::ListWrapper::VisitMut<dfly::(anonymous namespace)::ListWrapper::Pop(dfly::QList::Where)::<lambda(auto:99&)> >(dfly::(anonymous namespace)::ListWrapper::Pop(dfly::QList::Where)::<lambda(auto:99&)>)::<lambda(auto:97*)>, dfly::(anonymous namespace)::ListWrapper::Pop(dfly::QList::Where)::<lambda(auto:99&)> >, std::variant<dfly::QList*, dfly::detail::ListPack>&> (__visitor=...)
    at /usr/include/c++/15.2.1/variant:1891
#19 std::visit<absl::lts_20250512::Overload<dfly::(anonymous namespace)::ListWrapper::VisitMut<dfly::(anonymous namespace)::ListWrapper::Pop(dfly::QList::Where)::<lambda(auto:99&)> >(dfly::(anonymous namespace)::ListWrapper::Pop(dfly::QList::Where)::<lambda(auto:99&)>)::<lambda(auto:97*)>, dfly::(anonymous namespace)::ListWrapper::Pop(dfly::QList::Where)::<lambda(auto:99&)> >, std::variant<dfly::QList*, dfly::detail::ListPack>&> (__visitor=...) at /usr/include/c++/15.2.1/variant:1956
#20 dfly::(anonymous namespace)::ListWrapper::VisitMut<dfly::(anonymous namespace)::ListWrapper::Pop(dfly::QList::Where)::<lambda(auto:99&)> > (this=0x5e909bd22f0, f=...)
    at /home/dranikpg/dragonfly/src/server/list_family.cc:129
#21 dfly::(anonymous namespace)::ListWrapper::Pop (this=0x5e909bd22f0, where=dfly::QList::TAIL) at /home/dranikpg/dragonfly/src/server/list_family.cc:230
#22 dfly::(anonymous namespace)::OpBPop (t=0x5e909f51b00, shard=0x5e9081c0200, key=..., dir=dfly::(anonymous namespace)::ListDir::RIGHT)
    at /home/dranikpg/dragonfly/src/server/list_family.cc:403
```

[adder.py](https://github.com/user-attachments/files/28211032/adder.py)
[benchmark.sh](https://github.com/user-attachments/files/28211031/benchmark.sh)
[tasks.py](https://github.com/user-attachments/files/28211033/tasks.py)

```
python3 adder.py
./benchmark.sh -w 8 -c 2
```

mkaruza:
Multiple fibers racing to access the same offloaded node must not
cause double-free or corruption. Cover three concurrent access patterns:
LINDEX (read race), BRPOP (concurrent drain), and LREM/LTRIM/LINSERT/LSET
(concurrent mutation with LREM correctness invariant).

Closes #7403

### DIFF
--- a/tests/dragonfly/tiering_test.py
+++ b/tests/dragonfly/tiering_test.py
@@ -319,3 +319,170 @@ async def test_tiered_replication_with_lists(df_factory: DflyInstanceFactory):
         *(SeederV2.capture(c, types=["LIST"]) for c in [master_client, replica_client])
     )
     assert len(set(hashes)) == 1, "Inconsistency detected between master and replica lists."
+
+
+LIST_TIERING_ARGS = {
+    "proactor_threads": 1,
+    "tiered_prefix": "/tmp/list_tiering",
+    "tiered_offload_threshold": "1.0",
+    "tiered_experimental_list_support": "true",
+    "tiered_experimental_cooling": "false",
+    "list_max_listpack_size": 1,
+    "list_tiering_threshold": 2,
+    "maxmemory": "256M",
+}
+
+
+@pytest.mark.large
+@pytest.mark.exclude_epoll
+@pytest.mark.opt_only
+async def test_concurrent_tiered_list_blocking_command(df_factory: DflyInstanceFactory):
+    """
+    Multiple concurrent BRPOP consumers drain a tiered list simultaneously.
+
+    Exercises concurrent consumption from tiered QList nodes. Multiple fibers may
+    attempt to materialize and drain the same node concurrently, so synchronization
+    must ensure correct coordination without corruption or crashes.
+    """
+    instance = df_factory.create(**LIST_TIERING_ARGS)
+    instance.start()
+
+    NUM_CONSUMERS = 32
+    NUM_ELEMENTS = 2000
+    KEY = "lst"
+    VALUE = "x" * 3000
+
+    client = instance.client()
+
+    pipe = client.pipeline(transaction=False)
+    for _ in range(NUM_ELEMENTS):
+        pipe.rpush(KEY, VALUE)
+    await pipe.execute()
+
+    async for info, breaker in info_tick_timer(client, section="TIERED", timeout=30):
+        with breaker:
+            assert info["tiered_total_stashes"] > 0
+
+    async def worker_drain(c):
+        results = []
+        while True:
+            r = await c.brpop([KEY], timeout=5)
+            if r is None:
+                break
+            results.append(r[1])
+        return results
+
+    consumer_clients = [instance.client() for _ in range(NUM_CONSUMERS)]
+    all_results = await asyncio.gather(*(worker_drain(c) for c in consumer_clients))
+
+    total = sum(len(r) for r in all_results)
+    assert total == NUM_ELEMENTS, f"Expected {NUM_ELEMENTS} elements, got {total}"
+    assert all(v == VALUE for r in all_results for v in r)
+
+    for c in consumer_clients:
+        await c.aclose()
+
+
+@pytest.mark.large
+@pytest.mark.exclude_epoll
+@pytest.mark.opt_only
+async def test_concurrent_tiered_list_materialization(df_factory: DflyInstanceFactory):
+    """
+    Multiple concurrent read calls on the same tiered list key.
+    Exercises concurrent materialization. All readers race to access
+    the same tiered node while it is being loaded into memory.
+    Synchronization ensures the node is materialized only once and shared
+    across concurrent readers without corruption or crashes.
+    """
+    instance = df_factory.create(**LIST_TIERING_ARGS)
+    instance.start()
+
+    NUM_READERS = 32
+    NUM_ELEMENTS = 200
+    KEY = "lst"
+    VALUE = "x" * 10_000  # large value so I/O takes longer
+    TARGET_IDX = NUM_ELEMENTS // 2
+
+    client = instance.client()
+
+    pipe = client.pipeline(transaction=False)
+    for _ in range(NUM_ELEMENTS):
+        pipe.rpush(KEY, VALUE)
+    await pipe.execute()
+
+    async for info, breaker in info_tick_timer(client, section="TIERED", timeout=30):
+        with breaker:
+            assert info["tiered_total_stashes"] > 0
+
+    reader_clients = [instance.client() for _ in range(NUM_READERS)]
+    reads = [asyncio.create_task(c.lindex(KEY, TARGET_IDX)) for c in reader_clients]
+
+    results = await asyncio.gather(*reads)
+
+    for i, result in enumerate(results):
+        assert result == VALUE, f"Reader {i} got wrong result"
+
+    for c in reader_clients:
+        await c.aclose()
+
+
+@pytest.mark.large
+@pytest.mark.exclude_epoll
+@pytest.mark.opt_only
+@pytest.mark.parametrize("command", ["lrem", "ltrim", "linsert", "lset"])
+async def test_concurrent_tiered_list_mutation(df_factory: DflyInstanceFactory, command: str):
+    """
+    Concurrent mutation of a tiered list via LREM, LTRIM, LINSERT, or LSET.
+    LREM also check correctness invariant: removed + remaining == original length.
+    """
+    instance = df_factory.create(**LIST_TIERING_ARGS)
+    instance.start()
+
+    NUM_WORKERS = 32
+    NUM_ELEMENTS = 500
+    KEY = "lst"
+    VALUE = "x" * 3000  # large value forces each element into its own plain node
+    PIVOT = "pivot"  # used only by linsert; placed deep past the in-memory threshold
+    TARGET_IDX = NUM_ELEMENTS // 2  # used only by lset; deep in offloaded territory
+
+    client = instance.client()
+
+    pipe = client.pipeline(transaction=False)
+    if command == "linsert":
+        for _ in range(NUM_ELEMENTS // 2):
+            pipe.rpush(KEY, VALUE)
+        pipe.rpush(KEY, PIVOT)
+        for _ in range(NUM_ELEMENTS // 2 - 1):
+            pipe.rpush(KEY, VALUE)
+    else:
+        for _ in range(NUM_ELEMENTS):
+            pipe.rpush(KEY, VALUE)
+    await pipe.execute()
+
+    async for info, breaker in info_tick_timer(client, section="TIERED", timeout=30):
+        with breaker:
+            assert info["tiered_total_stashes"] > 0
+
+    async def worker(c):
+        if command == "lrem":
+            return await c.lrem(KEY, 0, VALUE)
+        elif command == "ltrim":
+            return await c.ltrim(KEY, 1, -1)
+        elif command == "linsert":
+            return await c.linsert(KEY, "BEFORE", PIVOT, "new")
+        elif command == "lset":
+            return await c.lset(KEY, TARGET_IDX, "new")
+
+    worker_clients = [instance.client() for _ in range(NUM_WORKERS)]
+    tasks = [asyncio.create_task(worker(c)) for c in worker_clients]
+    results = await asyncio.gather(*tasks)
+
+    if command == "lrem":
+        total_removed = sum(results)
+        remaining = await client.llen(KEY)
+        assert (
+            total_removed + remaining == NUM_ELEMENTS
+        ), f"Invariant broken: removed={total_removed} remaining={remaining} total={NUM_ELEMENTS}"
+
+    for c in worker_clients:
+        await c.aclose()


dranikpg:
> Blocking list operations can wake multiple fibers simultaneously on the same shard

There are in general no transactional semantics for those if they suspend. But what about inlined scheduling? Why is this not a problem anymore? Does it never happen with non-blocking operations with inlined scheduling? 

Also, even if it really is only the case for blocking commands, maybe a BLPOP can steal a value of a suspended LPOP?

To me it looks like all operations should be guarded

romange:
Please verify that https://github.com/dragonflydb/dragonfly/pull/7459 fixes the issue

romange:
@mkaruza  we decided to go with this PR for 1.39.

Can you please either resolve or adress the comments if they are relevant?

romange:
It should be reverted post-release.

src/server/list_family.cc
```diff
@@ -94,11 +94,15 @@ void OffloadListNode(QList* ql, QList::Node* node) {
 void LoadListNode(QList* ql, QList::Node* node) {
   TieredStorage* ts = EngineShard::tlocal()->tiered_storage();
   DCHECK(ts);
+  node->load_pending = 1;
   QList::stats.onload_requests++;
   auto res = ReadTieredListNode(ql->GetDbIndex(), ql, node, node->GetExternalSlice(), ts).Get();
   if (!res) {
     LOG(WARNING) << "Failed to load list node from tiered storage: " << res.error().message();
+  } else {
+    node->offloaded = 0;
```
liujianyeey-oss: `fd5ad09` correctly stopped clearing `offloaded` on the IO-error path (qodo #1 from the bot review). The success path here still has a subtle hole: `res.ok()` does not imply that `Upload(node)` ran.

`OpManager::ProcessRead` (src/server/tiering/op_manager.cc:170-187) invokes the read_cb first (resolves the future to `OK`) and only then calls `NotifyFetched`. Inside `NotifyFetched` (src/server/tiered_storage.cc:362-372), `should_upload` is gated on `UploadBudget() > metrics.estimated_mem_usage`; per the comment on `UploadBudget()` (tiered_storage.h:115) the budget is explicitly allowed to go negative under memory pressure. When that gate is false, `Upload(node)` is skipped — but the future is already resolved with OK, so this line still sets `offloaded = 0`.

`SetExternal` (src/core/qlist.cc:422-428) calls `zfree(entry)` without nulling, so `entry` is a freed-but-non-null pointer in this window. The subsequent state is `offloaded=0, io_pending=0, entry=<dangling>`, and `Materialize` (qlist.cc:1049-1071) short-circuits at the first `if` without re-loading; the bottom DCHECK passes because `entry != nullptr`. Any caller that then dereferences `entry` (`LRANGE` / `LINDEX` / `LPOP` / `BLPOP` / `BLMOVE` / …) reads freed heap memory.

Trigger is deterministic under memory pressure, not a race. The mimalloc backtrace in #7403 sits on the same call surface.

Two ways to close this without changing the IO machinery:

1. Gate the `offloaded = 0` assignment on `node->entry != nullptr` (i.e. only clear when Upload actually ran), and on the Upload-skipped success path keep `offloaded = 1` so a subsequent `Materialize()` retries the load.
2. Null `node->entry` in `SetExternal` after `zfree`. That converts this from a silent UAF to a deterministic null-deref at a defined location (debug DCHECK / release segfault), and the existing `DCHECK(node->offloaded || node->entry != nullptr)` would then fire on the bad state instead of silently passing. Cheap defensive change.

Either patch is local. A regression test that forces `UploadBudget` negative (small `--tiered_upload_threshold`, push past it, then `LRANGE` on a tiered node) reproduces it without contention.

src/server/list_family.cc
```diff
@@ -445,15 +532,63 @@ OpResult<string> OpMoveSingleShard(const OpArgs& op_args, string_view src, strin
...
+    // In tiered mode, only a first fiber is allowed to perform the operation on
+    // a given QList node at a time. If another fiber is already active, this one
+    // returns KEY_NOTFOUND to avoid concurrent processing of the same node.
+    const bool tiered_src_dest =
+        IsTieringEnabled() && srcql_v2.IsQList() && srcql_v2.IsQListNodeTiered(ToWhere(src_dir));
+
+    if (tiered_src_dest) {
+      auto tiered_it = tl_tiered_list_blocking_active.find(src_it->second.RObjPtr());
+      if (tiered_it != tl_tiered_list_blocking_active.end() && tiered_it->second > 0) {
+        return OpStatus::KEY_NOTFOUND;
```
liujianyeey-oss: The busy-counter contract is consistent inside `OpMoveSingleShard`, but the downstream mapping in `MoveGeneric` (line ~1013-1016) turns this `KEY_NOTFOUND` into `rb->SendNull()` — the wire reply for *source list does not exist*. The non-blocking client has no way to distinguish *another fiber is mid-IO on this key* from *the key is truly absent*.

Concrete interleaving (single shard, tiered key `A` with offloaded head):

1. Fiber A: `BRPOP A 10` enters `OpBPop`, increments `tl_tiered_list_blocking_active[A]` to 1, yields inside `Materialize()` waiting on tiered IO.
2. Fiber B: `LMOVE A A LEFT RIGHT` (non-blocking) on the same shard enters `OpMoveSingleShard`, `tiered_src_dest = true`, finds counter > 0 here, returns `KEY_NOTFOUND`.
3. `MoveGeneric` maps that to `SendNull`. Fiber B's client sees nil while `A` is non-empty.

Two options:

1. In this `src==dest` tiered branch, `await` on `node_load_ec` (or on a `tl_tiered_list_blocking_active`-reaches-0 notification) and re-check, rather than returning `KEY_NOTFOUND` immediately. Symmetrical with how `OpBPop` already waits.
2. Introduce a distinct status (e.g. `TRANSIENT_BUSY`) so `MoveGeneric` can `SendError` or retry rather than send nil. Cleaner if other call paths grow the same need.

Worth a regression test under `tests/dragonfly/tiering_test.py`: a `BLPOP` consumer holding the counter while a sibling non-blocking `LMOVE A A` runs — current behavior is silent nil.

src/server/list_family.cc
```diff
@@ -445,15 +532,63 @@ OpResult<string> OpMoveSingleShard(const OpArgs& op_args, string_view src, strin
...
+        tl_tiered_list_blocking_active.erase(tiered_it);
+      }
+    }
+
     return val;
   }
 
+  // If tiered, this path should follow the same tiered-mode handling logic as OpBPop
+  // to prevent incorrect behavior under concurrent execution.
+  const bool tiered_src_dest = IsTieringEnabled() && srcql_v2.IsQList();
```
liujianyeey-oss: This `src!=dest` gate uses only `IsTieringEnabled() && IsQList()`, while the matching `src==dest` gate at line 541 also requires `IsQListNodeTiered(ToWhere(src_dir))`. The counter is incremented on the next line for any QList in tiering mode — even when no tiered IO is actually in flight on `src`'s head/tail.

The comment on line 574 already calls out the intended invariant: *"tiered_src_dest is always true inside the loop body (we only enter when node is tiered)"*. That holds inside the `while` at line 578 because that loop tests `IsQListNodeTiered`, but the counter increment at line 571 runs before the loop is entered, so the invariant doesn't carry back to the gate.

Concrete cross-collision (single shard, tiering enabled, key `A` with head currently in memory):

1. Fiber X: `LMOVE A B`. Line 569 gate true (QList + tiering on), line 571 increments `counter[A] = 1`. The `while` body at line 578 doesn't execute because `A`'s head is not tiered. X proceeds with the move.
2. Background offload tick offloads `A`'s head.
3. Fiber Y: `LMOVE A A` (same shard, concurrent). Line 541 strong gate is now true (`IsQListNodeTiered` flipped true between X and Y). Line 545 sees `counter[A] > 0`, returns `KEY_NOTFOUND`. `MoveGeneric` sends nil.

Y's nil response is caused by X's increment on a non-tiered axis. Same wire-level surprise as F2 (sibling finding), but via a different path.

Fix is to make this gate match the `src==dest` form:

```cpp
const bool tiered_src_dest = IsTieringEnabled() && srcql_v2.IsQList() &&
                             srcql_v2.IsQListNodeTiered(ToWhere(src_dir));
```

That keeps the counter restricted to fibers actually doing tiered IO, which is what the line 574 comment expects.

liujianyeey-oss (commented):
Read through the tiered list_family changes. Three issues inline. All distinct from qodo's set, which `fd5ad09` already handles.

romange:
We can close it, I think

mkaruza:
> We can close it, I think

PR adds test around list tiering and this protection mechanism around materialization. With current approach there would be no penalty.

src/core/qlist.h
```diff
@@ -73,7 +74,8 @@ class QList {
     uint16_t dont_compress : 1;      /* prevent compression of entry that will be used later */
     uint16_t offloaded : 1;          /* node is offloaded to colder storage */
     uint16_t io_pending : 1;         /* node has pending io operation */
-    uint16_t reserved1 : 7;          /* reserved for future use */
+    uint16_t load_pending : 1;       /* node is being loaded from tiered storage */
```
romange: do we still need it? we have a logic now that ensures no RunCallback can run at the same time.

src/server/list_family.cc
```diff
@@ -94,11 +94,15 @@ void OffloadListNode(QList* ql, QList::Node* node) {
 void LoadListNode(QList* ql, QList::Node* node) {
   TieredStorage* ts = EngineShard::tlocal()->tiered_storage();
   DCHECK(ts);
+  node->load_pending = 1;
```
romange: I expect that when we block on ReadTieredListNode, no other callback will run until we finish.

src/core/qlist.h
mkaruza: I'll revert this changes. We don't need to use any synchronization mechanism.

src/server/list_family.cc
mkaruza: It's, i'll revert changes.

romange (approved):

PR #7420, opened 2026-05-26
feat(search): implement FT.HYBRID command with LINEAR/RRF score fusion

### LINKED ISSUE #7379: Implement FT.HYBRID
## Summary

Dragonfly does not implement `FT.HYBRID`. It is the only RediSearch surface needed by `redisvl >= 0.18` that is still missing - everything else (`FT.CREATE`, `FT.SEARCH` KNN + prefilter, `FT.AGGREGATE` GROUPBY, `FT.ALTER`, `FT.TAGVALS`, `FT.PROFILE`) already works.

## Repro

```bash
# Setup identical index + one document on both servers
for port in 6379 6380; do
  redis-cli -p $port FLUSHALL
  redis-cli -p $port FT.CREATE idx ON HASH PREFIX 1 d: SCHEMA \
    content TEXT \
    embedding VECTOR FLAT 6 TYPE FLOAT32 DIM 4 DISTANCE_METRIC COSINE
  python3 -c 'import struct,sys; sys.stdout.buffer.write(struct.pack("4f", 1, 0, 0, 0))' \
    | redis-cli -p $port -x HSET d:1 content "agent reasoning" embedding
done

# Hybrid query — vector via $vec param, linear combiner
python3 -c 'import struct,sys; sys.stdout.buffer.write(struct.pack("4f", 1, 0, 0, 0))' \
  | redis-cli -p 6380 -x FT.HYBRID idx \
      SEARCH "agent" \
      VSIM "@embedding" "\$vec" \
      COMBINE LINEAR 4 ALPHA 0.7 BETA 0.3 \
      LIMIT 0 5 \
      PARAMS 2 vec
```

Output on 6380 (Dragonfly):

```
ERR unknown command `FT.HYBRID`
```

The same query against an upstream-compatible server returns a `HybridResult` envelope with combined-score-sorted hits.

## Impact

`redisvl 0.18` is the current default on PyPI. Any project that pins `redisvl >= X` without an upper bound and runs `pip install` today gets it. Confirmed affected (live calls in current main branches):

- **Agno** - [`vectordb/redis/redisdb.py:475`](https://github.com/agno-agi/agno/blob/main/libs/agno/agno/vectordb/redis/redisdb.py) calls `HybridQuery` for `SearchType.hybrid`. Pins `redisvl>=0.12.1` with no upper bound, so fresh installs grab 0.18+.
- **Microsoft Agent Framework** - `agent-framework-redis.RedisContextProvider` uses the same `redisvl` hybrid path (`AggregateHybridQuery`).

## What clients send

`redisvl.HybridQuery` → `redis-py SearchCommands.hybrid_search()` -> wire shape:

```
FT.HYBRID <index>
  SEARCH "<text query>"
  VSIM @<vec_field> $<vec_param>
  COMBINE LINEAR 4 ALPHA <a> BETA <b>     # or RRF
  LIMIT <off> <count>
  PARAMS 2 <vec_param> <binary-vector>
```

The default combiner from `redisvl` is `LINEAR` with the user's `linear_alpha` (Agno passes `vector_score_weight`, default 0.7).

## Minimum viable implementation

To unblock Agno + Microsoft Agent Framework, support the `LINEAR` combiner only:

1. Parse `FT.HYBRID <idx> SEARCH "<text>" VSIM @<field> $<param> COMBINE LINEAR 4 ALPHA <a> BETA <b> LIMIT <o> <c> PARAMS ...`
2. Execute the text query and the KNN query through existing search paths.
3. Normalize both score sets to [0, 1].
4. Combine: `score = a * vector_score + b * text_score`.
5. Sort by combined score, apply `LIMIT`, return as `HybridResult` envelope.

`RRF` and other combiners can come later — production users hit `LINEAR` first.

## References

- `redisvl` call site: `redisvl/index/index.py:1166` (`hybrid_query()`)
- `redis-py` method: `redis.commands.search.commands.SearchCommands.hybrid_search` (added 7.1.0)
- Agno call site: `libs/agno/agno/vectordb/redis/redisdb.py:468-486`

vyavdoshenko:
Adds the `FT.HYBRID` command - a single-query hybrid search that combines full-text (BM25) with vector similarity (FLAT or HNSW) and fuses the two result sets using either LINEAR (weighted sum) or RRF (Reciprocal Rank Fusion) score combination.

Also exposes the corresponding `FT.PROFILE HYBRID` form with full per-phase timing and a properly nested iterator tree (the existing `FT.PROFILE SEARCH` tree renderer was broken - fixed in the same change).

Changes:

- **`FT.HYBRID idx SEARCH query VSIM @field $param [KNN|RANGE ...] [FILTER ...] [COMBINE LINEAR|RRF ...] [LOAD ...] [LIMIT off n] [PARAMS ...]`**
  - LINEAR: min-max normalised text + KNN scores, weighted by `ALPHA`/`BETA`.
  - RRF: rank-based fusion with `CONSTANT` (default 60) and optional `WINDOW`.
  - Supports `YIELD_SCORE_AS` aliases for text/vector/combined scores.
  - Supports `LOAD *` and `LOAD N @field …`. With `LOAD`, `__key`/`__score` are suppressed in the per-doc map (matching Redis behaviour).
- **`FT.PROFILE idx HYBRID [LIMITED] QUERY …`** - same query shape, returns `[search_result, profile_info]`. `LIMITED` collapses each event's `children` array into its count (also applied to `FT.PROFILE SEARCH`).
- **Profile tree fix** - `FT.PROFILE SEARCH/HYBRID` now nests child iterators under a `children` array on the parent event (previously emitted as siblings, breaking the declared collection sizes); typo `procecssed` -> `processed`.
- **Multi-shard correctness** - text-scoring stats collected with a dedicated pre-hop, schema validation captured via `std::call_once`, FLAT KNN and HNSW filter time measured separately from text time so the `phases` breakdown is accurate per path.
- **HNSW with FILTER** - pre-flight dim/param validation avoids scheduling a multi-shard text+filter hop on a bad query vector.

New helpers:
- `search::VectorSimilarityToString` / `search::DistanceToSimilarity` in `core/search/vector_utils.h` — single source of truth for metric -> string and metric-aware distance -> similarity ([0, 1]) conversion. L2 -> `1/(1+d²)`,
  IP/COSINE -> `(2-d)/2`.
- `ShardDocIndex::base()` — direct const-ref to the underlying `DocIndex` to avoid the `GetInfo()` copy when only the schema is needed (used by hybrid + 4 existing call sites).

Fixes: #7379


src/server/search/search_family.cc
```diff
@@ -1554,6 +1555,795 @@ std::pair<std::unique_ptr<search::AstNode>, search::AstKnnNode*> TryPopHnswKnnNo
...
+    auto scorer_fn = ParseScorer(parser);
+    if (!scorer_fn)
+      return CreateSyntaxError(absl::StrCat("No such scorer: ", parser->Peek()));
+    p.scorer = *scorer_fn;
+  }
+  if (parser->Check("YIELD_SCORE_AS"))
+    p.yield_text_score_as = string{parser->Next()};
+
+  if (!parser->Check("VSIM"))
+    return CreateSyntaxError("expected VSIM keyword"sv);
```
BorysTheDev: Check CmdArgParser functionality, you will find more convenient way I think

BorysTheDev: you can use Check("arg", &field);
you can use Apply(a lot of parameters or lambdas)

BorysTheDev: you can set error into parser and do error processing in one place

src/server/search/search_family.cc
```diff
@@ -1602,138 +1602,117 @@ struct HybridDocEntry {
...
+      p.scorer = *scorer_fn;
+  };
 
-  if (!parser->Check("VSIM"))
-    return CreateSyntaxError("expected VSIM keyword"sv);
+  if (!parser->Check("SEARCH")) {
+    parser->ReportCustom("expected SEARCH keyword");
+    return take_error();
+  }
+  p.text_query = string{parser->Next()};
```
BorysTheDev: `parser->Next<string>()`

src/server/search/search_family.cc
```diff
@@ -1602,138 +1602,117 @@ struct HybridDocEntry {
...
+    else
+      p.scorer = *scorer_fn;
+  };
 
-  if (!parser->Check("VSIM"))
-    return CreateSyntaxError("expected VSIM keyword"sv);
+  if (!parser->Check("SEARCH")) {
+    parser->ReportCustom("expected SEARCH keyword");
+    return take_error();
+  }
```
BorysTheDev: This check looks like parser->ExpectTag() with custom error, I suggest extending CmdArgParser functionality. because the parser is designed in a way that we shouldn't check errors until we have parsed all fields, even with errors.

src/server/search/search_family.cc
```diff
@@ -1602,138 +1602,117 @@ struct HybridDocEntry {
...
+  parser->Apply(Tag("SCORER", parse_scorer), Tag("YIELD_SCORE_AS", &p.yield_text_score_as));
 
-  string_view field = parser->Next();
-  if (!absl::StartsWith(field, "@"))
-    return CreateSyntaxError("VSIM field must start with @"sv);
-  p.vsim_field = string{field.substr(1)};
+  if (!parser->Check("VSIM")) {
+    parser->ReportCustom("expected VSIM keyword");
+    return take_error();
+  }
```
BorysTheDev: I suggest to add ExpectTag("tag", "error") to have ability use Apply

src/server/search/search_family.cc
```diff
@@ -1602,138 +1602,117 @@ struct HybridDocEntry {
...
+  if (string_view field = parser->Next(); absl::StartsWith(field, "@")) {
+    p.vsim_field = string{field.substr(1)};
+  } else {
+    parser->ReportCustom("VSIM field must start with @");
+    return take_error();
+  }
```
BorysTheDev: Consider to add ExpectStartsWith("str", "Error") into CmdArgParser::Apply() functionality

src/server/search/search_family.cc
```diff
@@ -1602,138 +1602,117 @@ struct HybridDocEntry {
...
+    const size_t n = sub->Next<size_t>();
+    std::vector<FieldReference> fields;
+    fields.reserve(n);
+    for (size_t i = 0; i < n; i++) {
+      string_view f = sub->Next();
+      if (absl::StartsWith(f, "@"))
+        f.remove_prefix(1);
+      string_view alias;
+      sub->Check("AS", &alias);
+      fields.emplace_back(f, alias);
```
BorysTheDev, line 1677: what if Check fails?

vyavdoshenko: Check("AS", &alias) returns false if AS isn't there - alias stays default-empty, meaning "no alias". Same pattern as ParseLoadOrReturnFields at search_family.cc:474. Check's internal bounds guard prevents reading past the end, so no garbage either way.

src/server/search/search_family.cc

src/server/search/search_family.cc

src/server/search/search_family.cc

src/server/search/search_family.cc

BorysTheDev (approved):

PR #7423, opened 2026-05-26
fix(ci): correct docker health check condition

fallintoplace:
## Summary
- fix the Docker test action's health-check loop so it actually waits for `healthy`

## Why
The shell test in `.github/actions/test-docker/action.yml` was malformed, so the loop succeeded immediately instead of comparing the container health status.

## Validation
- `bash -lc 'status=starting; if [ "$status"=="healthy" ]; then echo malformed:true; else echo malformed:false; fi; if [ "$status" = "healthy" ]; then echo fixed:true; else echo fixed:false; fi'`
- `pre-commit run --files .github/actions/test-docker/action.yml`

### DIFF
--- a/.github/actions/test-docker/action.yml
+++ b/.github/actions/test-docker/action.yml
@@ -20,6 +20,6 @@ runs:
 
         # docker run with port-forwarding
         docker run  --name test -d -p 6379:6379 ${{inputs.image_id}}
-        until [ "`docker inspect -f {{.State.Health.Status}} test`"=="healthy" ]; do
+        until [ "$(docker inspect -f '{{.State.Health.Status}}' test)" = "healthy" ]; do
           sleep 0.1;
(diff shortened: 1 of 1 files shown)


vyavdoshenko:
@fallintoplace 
Thanks for your effort.

vyavdoshenko (approved):

vyavdoshenko:
@fallintoplace 
Your commit must be signed. Please do it.

fallintoplace:
@vyavdoshenko Thank you for your attention. I will check later today.

vyavdoshenko:
@fallintoplace 
This commit should be verified in the same way as #7422. Please sign it, and will be able to merge it after it.

fallintoplace:
@vyavdoshenko Thank you. Please have a look when you have time.

PR #7426, opened 2026-05-26
perf: zero-copy GET for large strings with CoW

romange:
## Summary

Wire the zero-copy GET path end-to-end on top of the `cmn::BorrowedString` infrastructure landed in #7433.

For non-tiered string values, `CmdGet` now calls `pv.TryBorrow()` (gated by `--get_zero_copy`, default on) instead of `pv.ToString()`. A `cmn::BorrowedString` flows through the existing `StringResult` variant; the reply builder dispatches on encoding:

- **Raw (`NONE_ENC`)**: `WriteRef` pushes a pointer iovec into the source — `writev` reads the borrowed bytes directly, no copy.
- **ASCII1/2 packed**: `WriteDecodedAscii` chunk-decodes the packed source into the reply builder's scratch, one alignment unit at a time — no fully decoded payload ever materialises.

The borrow's destructor releases the pin once its bytes are in the kernel. Capture/replay (MULTI/EXEC squashing) preserves the path: the `cmn::BorrowedString` is moved into a `payload::Payload` variant alternative and back out at replay, with the same Flush-then-release flow on the real sink.

## What's new

- `string_family.cc`: `ReadStringBorrow`, the `cmn::BorrowedString` alternative in `StringResult`, and the dispatch in `Send(StringResult&&)`.
- `EngineShard::Heartbeat` calls `CompactObj::DrainPendingReads()` to reap pins whose refcnt reached zero between drains.
- New `--get_zero_copy` flag (default true) toggles the borrow path.
- New stat `borrowed_strings_sent_total` (in INFO `stats`) — incremented in `SendBulkStringBorrowed`, covers direct sink + capture/replay.

## Tests

6 new cases in `string_family_test.cc`:
- `GetLargeRawBorrowed` / `GetLargeRawBorrowedSquashed` — NONE_ENC, direct sink + MULTI/EXEC capture.
- `GetLargeAsciiBorrowedChunked` / `GetLargeAsciiBorrowedChunkedSquashed` — ASCII1/2 chunked decode, both paths.
- `PendingReadPinOrphanDrain` — writer racing reader → orphan → drain.
- `PendingReadDrainNonOrphaned` — clean release without orphan transition.

## Drive-by fixes

- `reply_builder.cc::write_piece(string_view, char*)`: skip `memcpy` when the view is empty — silences a UBSan null-pointer-to-`memcpy` warning that fired in CI on unrelated tests.
- `tiered_storage_test::ThrottleClients`: dropped a flaky `clients_throttled > N/10` heuristic; the stronger `total_clients_throttled == N` assertion remains.
- `tests/dragonfly/connection_test.py::test_multi_exec_phantom_connections`: pinned `get_zero_copy=false` for this regression test. The borrow path makes the clogger's EXEC callbacks complete fast enough that ghost EXECs no longer queue behind it, shrinking the repro window below detection. The underlying `Transaction::Execute` io_uring/scheduling path that the test characterises is unchanged.

## Notes

- Builds on the merged #7433 infrastructure (`cmn::BorrowedString`, `BorrowedStringOps`, `CompactObj::TryBorrow`, `PinnedMap`, single `SendBulkStringBorrowed` virtual).
- `ReplyStats` grows by one `uint64_t` for the new counter (`sizeof` static_assert bumped 80 → 88).

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -446,6 +446,14 @@ class PinnedMap {
   absl::flat_hash_map<const void*, PendingRead*> pin_map_;
 
  public:
+  ~PinnedMap() {
+    // Thread shutdown: free any PendingRead entries that EngineShard::Heartbeat
+    // never had a chance to drain. Orphan buffers are best-effort — the
+    // thread's mimalloc heap is going away anyway.
+    for (auto& [_, pin] : pin_map_)
+      delete pin;
+  }
+
   PendingRead* RegisterPin(const void* ptr);
   // Returns true if `ptr` was pinned.
   bool Orphan(const void* ptr);
--- a/src/facade/facade.cc
+++ b/src/facade/facade.cc
@@ -64,7 +64,7 @@ ReplyStats::ReplyStats(ReplyStats&& other) noexcept {
 }
 
 ReplyStats& ReplyStats::operator+=(const ReplyStats& o) {
-  static_assert(sizeof(ReplyStats) == 80u + kSanitizerOverhead);
+  static_assert(sizeof(ReplyStats) == 88u + kSanitizerOverhead);
   ADD(io_write_cnt);
   ADD(io_write_bytes);
 
@@ -73,6 +73,7 @@ ReplyStats& ReplyStats::operator+=(const ReplyStats& o) {
   }
 
   ADD(script_error_count);
+  ADD(borrowed_string_sent_cnt);
 
   send_stats += o.send_stats;
   squashing_current_reply_size.fetch_add(o.squashing_current_reply_size.load(memory_order_relaxed),
@@ -83,7 +84,7 @@ ReplyStats& ReplyStats::operator+=(const ReplyStats& o) {
 #undef ADD
 
 ReplyStats& ReplyStats::operator=(const ReplyStats& o) {
-  static_assert(sizeof(ReplyStats) == 80u + kSanitizerOverhead);
+  static_assert(sizeof(ReplyStats) == 88u + kSanitizerOverhead);
 
   if (this == &o) {
     return *this;
@@ -94,6 +95,7 @@ ReplyStats& ReplyStats::operator=(const ReplyStats& o) {
   io_write_bytes = o.io_write_bytes;
   err_count = o.err_count;
   script_error_count = o.script_error_count;
+  borrowed_string_sent_cnt = o.borrowed_string_sent_cnt;
   squashing_current_reply_size.store(o.squashing_current_reply_size.load(memory_order_relaxed),
                                      memory_order_relaxed);
   return *this;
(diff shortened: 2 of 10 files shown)


src/core/compact_object.cc
```diff
@@ -446,6 +446,14 @@ class PinnedMap {
   absl::flat_hash_map<const void*, PendingRead*> pin_map_;
 
  public:
+  ~PinnedMap() {
+    // Thread shutdown: free any PendingRead entries that EngineShard::Heartbeat
+    // never had a chance to drain. Orphan buffers are best-effort — the
+    // thread's mimalloc heap is going away anyway.
+    for (auto& [_, pin] : pin_map_)
+      delete pin;
+  }
```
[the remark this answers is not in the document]
romange: `PinnedMap` being thread-local destructs during thread destruction.

romange:
augment review

src/server/string_family.cc
```diff
@@ -678,11 +707,34 @@ struct GetReplies {
...
+      // of the borrow (and associated pin) and parks the pin until all the writes complete.
+      rb->SendBulkStringBorrowed(std::move(get<cmn::BorrowedString>(res)));
+      return;
+    }
+    auto fut = get<TieredStorage::TResult<std::string>>(std::move(res));
+    io::Result<std::string> iores = fut.Get();
+    if (iores.has_value())
+      Send(*iores);
+    else
+      Send(iores.error().message());
```
BorysTheDev, line 724: nit: `Send(iores.has_value() ? *iores :  iores.error().message());`

src/facade/reply_builder.cc
```diff
@@ -64,7 +64,7 @@ template <typename T> enable_if_t<is_integral_v<T>, char*> write_piece(T num, ch
 }
 
 char* write_piece(string_view str, char* dest) {
-  return (char*)memcpy(dest, str.data(), str.size()) + str.size();
+  return str.empty() ? dest : (char*)memcpy(dest, str.data(), str.size()) + str.size();
```
BorysTheDev, line 67: you can use `return (char*)memcpy(dest, str.data(), str.size()) + str.size()` you don't need check for empty

src/facade/facade.cc
```diff
@@ -83,7 +84,7 @@ ReplyStats& ReplyStats::operator+=(const ReplyStats& o) {
 #undef ADD
 
 ReplyStats& ReplyStats::operator=(const ReplyStats& o) {
-  static_assert(sizeof(ReplyStats) == 80u + kSanitizerOverhead);
+  static_assert(sizeof(ReplyStats) == 88u + kSanitizerOverhead);
```
BorysTheDev, line 87: you can write static_assert only once in a class, instead of in every operator

BorysTheDev (approved):

src/facade/reply_builder.cc
romange: I must - clang with ASAN complains because `memcpy` declared as it can not accept nullptr even with 0 size.

src/facade/facade.cc
romange: The reason I do that is to fix each method when I add variables. This is how I know where to add as compilation breaks.

PR #7427, opened 2026-05-27
fix(test): Remove stale keyspace assertion in _run_tiering_migration

mkaruza:
After a full slot migration, originating node may still have a non-zero key count. 
Remove the assertion and keep only the meaningful check on destination node.

### DIFF
--- a/tests/dragonfly/cluster_test.py
+++ b/tests/dragonfly/cluster_test.py
@@ -3792,9 +3792,6 @@ async def _run_tiering_migration(
         with breaker:
             assert info["tiered_entries"] == 0
 
-    info = await nodes[0].client.info("keyspace")
-    assert info["db0"]["keys"] == 0
-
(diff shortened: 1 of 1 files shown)


romange:
@mkaruza  why the source node will still have non-zero count?

mkaruza:
> @mkaruza why the source node will still have non-zero count?

AFAIK, flushing of slots will be triggered after migration FINISHED with executing this code.

```
    nodes[0].migrations = []
    nodes[0].slots = []
    nodes[1].slots = [(0, 16383)]
    logging.debug("finalize migration")
```

This triggers deletion of keys via dispatched fiber `FlushSlotsFb`. 

We are waiting only that tiered entries are removed, not that key count is zero.

But this check is not necessary to show that migration with tiered entries is correct.

romange (approved):

PR #7428, opened 2026-05-27
chore: tls socket NEED_WRITE on recv while WRITE_IN_PROGRESS

kostasrim:
We had a test that used protocol renegotiation to exercise the same path. However it was unreliable and reproduced very rarely. 1.3 connections use KeyUpdates to exercise the same path (in fact it's the **only** case the `Recv might NEED_WRITE`)


kostasrim:
moved to helio

PR #7433, opened 2026-05-27
perf: zero-copy GET infrastructure (common, core, facade)

romange:
## Summary

Foundation for zero-copy GET on large strings. Adds the move-only `BorrowedString` handle, the pin registry, the reply-builder integration, and the capture/replay path — without the server-side GET wiring (that lives on top in `perf/zero-copy-cow`).

## Design

A `BorrowedString` ties the lifetime of borrowed bytes to a refcount-bearing pin on the source `LargeString`. The reply builder owns the borrow for the duration of `writev`; the capturing builder owns it for the duration of squash/MULTI-EXEC replay. Concurrent mutation of the same key orphans the buffer to the pin and lets the original reader finish; the owning shard reaps orphaned buffers on `Heartbeat` via `DrainPendingReads`.

The pin handle is opaque to everything outside `core/`. `cmn::BorrowedStringOps` is the single seam: `Release` decrements the refcount, `DecodeChunk` unpacks ASCII chunks. The concrete impl is file-local in `compact_object.cc` and is the only place that knows `PendingRead` exists or that decoding means `detail::ascii_unpack`.

## What's new

**common** — `cmn::BorrowedString`, `cmn::BorrowedStringOps`.

**core** — `CompactObj::TryBorrow()`, `PinnedMap` / `PendingRead`, `CompactObjBorrowOps` (registered in `InitThreadLocal`), `DrainPendingReads()` called from `EngineShard::Heartbeat`. `LargeString::SetString` / `Free` orphan into the pin map when `read_pending` is set; `DefragIfNeeded` skips while pinned.

**facade** — single `SendBulkStringBorrowed(cmn::BorrowedString)` virtual on `RedisReplyBuilderBase` (raw → `WriteRef`; packed → `WriteDecodedAscii`). `ParkBorrow` extends pin lifetime across `writev`; `post_send_borrows_` is cleared at the end of `Send()`. `CapturingReplyBuilder` override moves the borrow into a `payload::BorrowedString` variant; replay moves it back into the real sink. No global unpin fn, no `void*` pin machinery.

**docs** — `docs/zero_copy_get.md` updated to describe the implemented design end-to-end.

## Notes

- `dfly_facade` does not link `dfly_core` — the bitpacking primitive is reached only via the registered `BorrowedStringOps`, so the facade stays decoupled.
- `payload::Payload` size is preserved (`sizeof == 40`); the new alternative is `std::unique_ptr<cmn::BorrowedString>`.
- HTTP `CaptureVisitor` handles the new payload (raw bytes JSON-escaped directly, packed materialized once via `DecodeChunk`).
- Server-side `string_family.cc` integration is out of scope for this PR.

### DIFF
--- a/src/common/borrowed_string.h
+++ b/src/common/borrowed_string.h
@@ -1,0 +1,144 @@
+// Copyright 2026, DragonflyDB authors.  All rights reserved.
+// See LICENSE for licensing terms.
+//
+
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+#include <span>
+#include <string_view>
+#include <utility>
+
+namespace cmn {
+
+// Move-only owning handle for a borrowed bulk string. The bytes in `encoded`
+// remain valid as long as this object is alive — the destructor releases the
+// underlying pin via BorrowedStringOps::Get()->Release().
+//
+// For NONE_ENC (encoding == 0): `encoded` is the user-visible bytes;
+// `decoded_size` is unused (caller treats it as `encoded.size()`).
+// For ASCII1/ASCII2_ENC: `encoded` is the packed source; the consumer
+// decodes chunk-by-chunk via BorrowedStringOps::Get()->DecodeChunk().
+
+class BorrowedStringOps;
+
+class BorrowedString {
+  friend class BorrowedStringOps;
+
+ public:
+  // BorrowedString() noexcept = default;
+  BorrowedString(std::string_view encoded, size_t decoded_size, uint8_t encoding,
+                 void* pin) noexcept
+      : encoded_(encoded), pin_(pin), encoding_(encoding) {
+  }
+
+  BorrowedString(const BorrowedString&) = delete;
+  BorrowedString& operator=(const BorrowedString&) = delete;
+
+  BorrowedString(BorrowedString&& o) noexcept {
+    MoveFrom(std::move(o));
+  }
+
+  BorrowedString& operator=(BorrowedString&& o) noexcept {
+    if (this != &o) {
+      Unpin();
+      MoveFrom(std::move(o));
+    }
+    return *this;
+  }
+
+  ~BorrowedString() noexcept {
+    Unpin();
+  }
+
+  void* pin() const noexcept {
+    return pin_;
+  }
+
+  bool IsEncoded() const noexcept {
+    return encoding_ != 0;
+  }
+
+  std::string_view view() const noexcept {
+    return encoded_;
+  }
+
+  uint8_t encoding() const noexcept {
+    return encoding_;
+  }
+
+ private:
+  // Release the pin if held. Delegates to BorrowedStringOps::Get()->Release().
+  void Unpin() noexcept;
+
+  void MoveFrom(BorrowedString&& o) noexcept {
+    encoded_ = o.encoded_;
+    encoding_ = o.encoding_;
+    pin_ = std::exchange(o.pin_, nullptr);
+  }
+
+  std::string_view encoded_;
+  void* pin_ = nullptr;
+  uint8_t encoding_ = 0;
+};
+
+// Abstract interface registered once (by CompactObj) to provide pin lifecycle
+// and ASCII decode operations without coupling common/ or facade/ to core/.
+// Designed to be a singleton accessed via BorrowedStringOps::Get().
+class BorrowedStringOps {
+ public:
+  virtual ~BorrowedStringOps() = default;
+
+  // Consume the borrow: extract pin via bs.ReleasePin() and decrement its
+  // refcount. Called by BorrowedString::Unpin() / destructor.
+  void Release(BorrowedString&& bs) noexcept {
+    ReleaseInternal(bs);
+    bs.pin_ = nullptr;  // ensure pin is released even if ReleaseInternal throws
+  }
+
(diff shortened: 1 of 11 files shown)


src/core/compact_object.h
```diff
@@ -248,73 +250,26 @@ class CompactObj {
...
+  // Read-only fast path. Returns a cmn::BorrowedString iff this CompactObj
+  // holds a string value that can be borrowed, otherwise std::nullopt
+  // (caller uses GetSlice/GetString/ToString). The borrowed bytes are valid
+  // until the pin is released.
   //
   // For NONE_ENC: `encoded` is the user-visible bytes; the reply can stream
   // them directly. For ASCII1/ASCII2_ENC: `encoded` is the packed source
   // (`encoded.size() < decoded_size`) and the reply must decode in chunks.
   //
-  struct BorrowedString {
```
romange, line 259: moved to a separate header.

romange:
@BorysTheDev @dranikpg  it may look like a big PR but a big part of it is to move BorrowedString to another header and introducing a singleton for backward binding to compact object logic. The only new thing is in facade where we serialize BorrowedString or pass it further.

src/facade/reply_builder.cc
```diff
@@ -127,6 +127,38 @@ template <typename... Ts> void SinkReplyBuilder::WritePieces(Ts&&... pieces) {
...
+  // Huge strings benefit from max scratch capacity. Going via Flush() (rather
+  // than EnsureCapacity directly) drains any pending iovecs first — otherwise
+  // a reallocate-grow would dangling-pointer the prefix iovec written by the
+  // caller into our own scratch.
+  if (buffer_.Capacity() < kMaxBufferSize)
+    Flush(kMaxBufferSize);
+
+  while (src_offset < src_total) {
+    // Ensure buffer_ has some room.
+    if (buffer_.AppendLen() < 64)
```
BorysTheDev, line 144: why 64

src/server/http_api.cc
```diff
@@ -119,6 +120,22 @@ struct CaptureVisitor {
...
+    // encodings the bytes are already user-visible; for packed encodings we
+    // materialize once via the registered decode op (chunked decoding only
+    // matters for iovec sinks).
+    if (!bs.IsEncoded()) {
+      absl::StrAppend(&str, JsonEscape(bs.view()));
+      return;
+    }
+    auto* ops = cmn::BorrowedStringOps::Get();
+
+    std::string decoded(ops->DecodedSize(bs), '\0');
```
BorysTheDev: looks like extra work to fill zeroes

BorysTheDev (dismissed):

src/common/borrowed_string.h
```diff
@@ -0,0 +1,143 @@
...
+  // Decode a chunk of source [src_offset, src_offset + src_len) from
+  // bs.view() into `dest` (up to `dest_len` decoded bytes). The impl picks
+  // how much to actually consume/produce — subject to its own alignment
+  // rules — and reports both via the returned struct. Caller advances its
+  // source cursor by `src_consumed` and the destination cursor by
+  // `dec_written`. Caller must provide enough scratch for the impl to make
+  // progress — a buffer big enough for one alignment unit is always
+  // sufficient. Only called for non-raw (packed) encodings.
+  virtual DecodeResult DecodeChunk(const BorrowedString& bs, size_t src_len, size_t src_offset,
+                                   size_t dest_len, char* dest) noexcept = 0;
```
dranikpg: std::span dest

src/common/borrowed_string.h
```diff
@@ -0,0 +1,143 @@
...
+    return s_ops_;
+  }
+
+ protected:
+  virtual void ReleaseInternal(BorrowedString& bs) noexcept = 0;
+
+ private:
+  inline static BorrowedStringOps* s_ops_ = nullptr;
+};
+
```
dranikpg, line 138: Kinda cool but three function pointers would've done the job as well 😅

src/facade/reply_capture.cc
```diff
@@ -54,6 +54,13 @@ void CapturingReplyBuilder::SendBulkString(std::string_view str) {
   Capture(BulkString{string{str}});
 }
 
+// Capture the borrow into the payload, extending the pin's lifetime until
+// replay moves it into the real sink where it is parked across the writev.
+void CapturingReplyBuilder::SendBulkStringBorrowed(cmn::BorrowedString bs) {
+  SKIP_LESS(ReplyMode::FULL);
+  Capture(std::move(bs));
+}
+
```
dranikpg, line 63: we should look into having an optimized version of the CRB that writes directly to a buffer isntead of operating with those partial values, then you also don't need a special type for this

dranikpg (dismissed):

src/common/borrowed_string.h
romange: Yes, but having 3 function pointers and this class is the same ADT to my mind. What am I missing?

src/facade/reply_builder.cc
romange: arbitrary minimum space we need to decode.

src/common/borrowed_string.h
dranikpg: less code

dranikpg (approved):

PR #7437, opened 2026-05-28
perf(facade): defer pipeline flushes in IoLoopV2

glevkovich:
In IoLoopV2, pipelined sync-only commands (e.g., PUBLISH) previously
triggered immediate sendmsg syscalls via FinishScope(). This change
defers all flushing until the fiber is truly idle, collapsing
multiple syscalls into one per pipeline.

Changes:
- ExecuteBatch: Set batch_mode=true on the head command when more
  commands follow to suppress FinishScope() flushes mid-pipeline.
- IoLoopV2: Extract should_wake predicate and guard the flush block
  with !should_wake() to skip flushing when commands are queued.
- ExecuteBatch/ReplyBatch: Use absl::Cleanup RAII guards to ensure
  SetBatchMode(false) on all return paths (fixes early-return bug
  where GetError() could leak batched_=true).
- reply_builder.h: Add IsBatchMode() getter to support the DCHECK.

Benchmark context: Pinned client, 2 proactor threads, laptop, mean of
5 runs with all other apps closed.

Pubsub (Sync-only) Results:
- Pubsub RPS (p=10): 121K -> 357K (+195%)
- Pubsub RPS (p=100): 132K -> 556K (+321%)
- Pubsub syscalls (p=10): 171K -> 29K (-83%)
- Pubsub syscalls (p=100): 157K -> 6.9K (-96%)

Throughput / Pipeline Results:
- Neutral (~0% change). Async-capable commands (SET/GET) already
  defer replies to buffers, meaning they naturally bypassed this
  syscall bottleneck. RPS remains highly stable (e.g., p=10 at
  ~403K RPS, p=100 at ~502K RPS) with zero regressions across
  fragmentation and throughput tests.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1570,6 +1570,10 @@ void Connection::HandleMigrateRequest() {
 }
 
 bool Connection::ProcessControlMessages(uint32_t quota) {
+  // Invariant: batched_ must be false on entry.
+  // PubSub replies flush immediately via FinishScope() only when batched_ is false.
+  // ReplyBatch() and ExecuteBatch() both reset it via absl::Cleanup guards on all return paths.
+  DCHECK(!reply_builder_->IsBatchMode());
   uint32_t dispatched = 0;
 
   while (!dispatch_q_.empty()) {
@@ -2561,6 +2565,10 @@ bool Connection::ParseMCBatch(base::IoBuf& io_buf) {
 }
 
 bool Connection::ExecuteBatch() {
+  // Invariant: batched_ must be false on entry.
+  // Both ReplyBatch() and ExecuteBatch() reset it via absl::Cleanup guards on all return paths.
+  DCHECK(!reply_builder_->IsBatchMode());
+  absl::Cleanup batch_guard = [this] { reply_builder_->SetBatchMode(false); };
   auto& conn_stats = tl_facade_stats->conn_stats;
   auto advance_head = [this]() -> ParsedCommand* {
     auto* cmd = parsed_head_;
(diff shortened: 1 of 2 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -3028,7 +3036,24 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
...
                (parsed_head_ && parsed_head_->CanReply()) || !dispatch_q_.empty() || io_ec_ ||
                is_ready_to_migrate();
-      });
+      };
+
+      // Only flush and park if the fiber is truly idle. When synchronous commands
+      // (e.g. PUBLISH) are pipelined, ExecuteBatch processes one at a time and loops
+      // back here with HasCommandToExecute() == true. Skipping the flush lets the
+      // entire pipeline execute in-memory before a single sendmsg at the end.
+      if (!should_wake()) {
```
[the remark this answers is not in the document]
glevkovich: Dropping this, reasons:

- Batching is off for PubSub: When the server processes PubSub messages (ProcessControlMessages), a specific setting called batched_ is set to false. I am adding a DCHECK for this.
- The Auto-Flush: Because batching is false, the code that builds the message (SendBulkStrArr) automatically sends it to the network the exact millisecond it is finished.
- The Result: Every single PubSub message is sent out immediately. Nothing piles up or accumulates in memory. If you added the bot's extra Flush() command, it would be a "no-op" (a useless action) because it would just be trying to flush an already-empty pipe.

Comment: I have a future task: "Reply Batching in ProcessControlMessages", but for now there is not batching.

src/facade/dragonfly_connection.cc
```diff
@@ -2561,6 +2561,9 @@ bool Connection::ParseMCBatch(base::IoBuf& io_buf) {
 }
 
 bool Connection::ExecuteBatch() {
+  // Invariant: batched_ must be false on entry.
+  // ReplyBatch() always resets it to false before returning (or we exit before re-entering here).
+  DCHECK(!reply_builder_->IsBatchMode());
```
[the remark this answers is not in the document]
glevkovich: fixed

src/facade/dragonfly_connection.cc
```diff
@@ -2595,16 +2603,28 @@ bool Connection::ExecuteBatch() {
...
+    // V2: batch the head command's reply when more commands are queued behind it.
+    // Without this, sync-only commands (PUBLISH, SET, INCR) that can't run async
+    // become head one at a time, and FinishScope() flushes each reply immediately
+    // (batched_ is false). This causes one sendmsg syscall per command instead of
+    // one for the entire pipeline. IoLoopV2's idle-await block handles the final flush.
+    //
+    // FinishScope() is triggered by ReplyScope's RAII destructor inside every Send*() method
+    // of the reply builder. When the command handler calls SendXxx(), ReplyScope is created;
+    // on return, ~ReplyScope() calls FinishScope() which flushes unless batched_ is true.
+    reply_builder_->SetBatchMode(ioloop_v2_ && is_head && (cmd->next != nullptr));
```
dranikpg: The comment is helpful, but too long comments sometimes reduce redability of code because functions become too large.  You won't understand from this comment how command replies work fully, but if you understand how they work, you don't need this comment. It's pretty evident what batching does and what for

dranikpg (dismissed):

src/facade/dragonfly_connection.cc
romange: My 2 cents: maybe it can be shortened but I disagree with the premise. Sometimes "recaps" are helpful to understand non-trivial interactions. I know it's subjective and some things can look trivial to more folks more experienced with the codebase but in general, my feeling is that our code is biased towards "no comments" than "too verbose".

romange: I for example, had to remember again and again the differrence between batching and scoping when working with reply builder.

dranikpg: >  differrence between batching and scoping

This is why I document the objects themselves 🙂 You can open them and read one or two sentences about them.  But somewhere in code is not the place I'll search for explanaitions

glevkovich: I will shorten it.

dranikpg (approved):

PR #7439, opened 2026-05-28
fix: Break big value flush on socket error

### LINKED ISSUE #7187: test_big_huge_streaming_restart fails

dranikpg:
When replication is cancelled or errors, there is no error path for big values, their serialization continues unstopped. Each failed socket write comes with sleep throttling, so the snapshot is stuck writing a big value and sleeping. This PR short circuits the socket write, so no sleep occurs after and it becomes a cpu-only operation. 

The best solution would be of course to add cancelation to RdbSerializer. But this solution is already better except for the CPU hogging 

Fixes #7187


abhijat (dismissed):
lgtm

dranikpg:
Had to merge with conflicts

PR #7441, opened 2026-05-28
chore(server): Transaction inlining comments

dranikpg:
Add comments to transaction inlining flow

### DIFF
--- a/src/server/transaction.h
+++ b/src/server/transaction.h
@@ -546,7 +546,13 @@ class Transaction {
   // Should be called immediately after the last hop.
   void LogAutoJournalOnShard(EngineShard* shard, RunnableResult shard_result);
 
-  // Whether the callback can be run directly on this thread without dispatching on the shard queue
+  // Whether the callback can be run directly on this fiber without dispatching on the shard queue.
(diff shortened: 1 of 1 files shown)


src/server/transaction.h
```diff
@@ -546,7 +546,13 @@ class Transaction {
   // Should be called immediately after the last hop.
   void LogAutoJournalOnShard(EngineShard* shard, RunnableResult shard_result);
 
-  // Whether the callback can be run directly on this thread without dispatching on the shard queue
+  // Whether the callback can be run directly on this fiber without dispatching on the shard queue.
+  // It checks internally that there are no possible suspension points.
+  //
+  // We do not support suspendable shard callbacks. Because all locks are acquried as intent locks,
```
romange:
```suggestion
  // We do not support suspendable shard callbacks. Because all locks are acquired as intent locks,
```

romange (approved):

romange:
Omitting inline transactions based on the intent locks from inline path won't help as well, as inline tx could suspend before a contending transaction from shard queue starts.

romange:
@dranikpg  I have an idea on how to solve this systematically.

PR #7446, opened 2026-05-29
tests: Fix test_denyoom_commands by removing RSS factor

### LINKED ISSUE #7279: `test_denyoom_commands[df_factory0]` parameters rejected (weak test)
```
E       AssertionError: Weak testcase: too little used memory
E       assert 262492296 > 268435456
```

https://github.com/dragonflydb/dragonfly/actions/runs/25368737302/job/74386346444#step:7:1246

abhijat:
Currently this test does the following:

* set up max memory
* use debug populate to trigger fill
* assert used mem > max mem
* proceed with firing off cmds to see them get denied

This second step is broken. debug populate issues commands to fill the data set, and RSS climbs such that rss / used > 1.25 at which point the RSS leg of the deny clause kicks in. In short

`used < 256m < 256m * 1.25 < rss`

which causes deny to kick in.

Then the debug populate partially blocked, and the test thinks it did not produce enough data (used < max). Since this test does not intend to test rss limits, it should keep that limit off and only test the used mem leg of the deny condition.

FIXES https://github.com/dragonflydb/dragonfly/issues/7279

### DIFF
--- a/tests/dragonfly/generic_test.py
+++ b/tests/dragonfly/generic_test.py
@@ -232,7 +232,9 @@ async def test_reply_guard_oom(df_factory, df_seeder_factory):
 
 @pytest.mark.asyncio
 async def test_denyoom_commands(df_factory):
-    df_server = df_factory.create(proactor_threads=1, maxmemory="256mb", oom_deny_commands="get")
+    df_server = df_factory.create(
+        proactor_threads=1, maxmemory="256mb", oom_deny_commands="get", rss_oom_deny_ratio=-1
+    )
     df_server.start()
     client = df_server.client()
     await client.execute_command("DEBUG POPULATE 7000 size 44000")


abhijat:
It is suspicious that this test started failing more recently. It raises the question whether we are using more RSS than we used to for the same data set.

romange:
@abhijat  yes, it's very suspicious. Can you please find the commit causing the degradation?

abhijat:
@romange since the test failure and this fix is unrelated to the memory increase (increased RSS just showed up here due to test behavior), I think this fix can go ahead as it is?

romange:
@abhijat  can we introduce a test that exposes memory usage change between LEGACY_GLOG settings?

I am ok with fixing the test but it uncovered something else by chance so I want it to be analyzed.

tests/dragonfly/generic_test.py
```diff
@@ -232,7 +232,9 @@ async def test_reply_guard_oom(df_factory, df_seeder_factory):
 
 @pytest.mark.asyncio
 async def test_denyoom_commands(df_factory):
-    df_server = df_factory.create(proactor_threads=1, maxmemory="256mb", oom_deny_commands="get")
+    df_server = df_factory.create(
+        proactor_threads=1, maxmemory="256mb", oom_deny_commands="get", rss_oom_deny_ratio=-1
```
romange, line 236: I do not understand the fix. if `rss_oom_deny_ratio=-1` then why `oom_deny_commands` will be denied?

abhijat: It is because the condition in our code is (src/server/main_service.cc:1291)

```c++
if (memory_stats.used_mem > limit ||
        (etl.rss_oom_deny_ratio > 0 && memory_stats.rss_mem > (limit * etl.rss_oom_deny_ratio))) 
```

The test is supposed to check the the first branch of the predicate `memory_stats.used_mem > limit` behavior 

Instead of reaching test body the debug populate is blocked by the rss because of the second branch and then it fails as weak test. For this test the rss related denial is irrelevant, we should only test used mem based denial. 

For rss specific branch we already have `memory_test.test_rss_oom_ratio`

abhijat:
> @abhijat can we introduce a test that exposes memory usage change between LEGACY_GLOG settings?
> 
> I am ok with fixing the test but it uncovered something else by chance so I want it to be analyzed.

sure I will add it

romange (approved):

abhijat:
follow up issue for new test https://github.com/dragonflydb/dragonfly/issues/7471

PR #7449, opened 2026-05-29
chore(core): Re-enable container iterations yielding

mkaruza:
Re-enable `container_iteration_yield_interval_usec` and set back to 500us.

Verify that a container iteration command returns a consistent result even
when it yields and concurrent writes are issued at the same time.

The test runs LRANGE on a large list while other fibers continuously perform
RPUSH operations and verifies that each LRANGE result matches the list state
observed when the command started.

### DIFF
--- a/src/server/container_utils.cc
+++ b/src/server/container_utils.cc
@@ -24,7 +24,7 @@ extern "C" {
 #include "redis/util.h"
 }
 
-ABSL_FLAG(uint32_t, container_iteration_yield_interval_usec, 0,
+ABSL_FLAG(uint32_t, container_iteration_yield_interval_usec, 500,
           "Yield the fiber every N microseconds during container iteration. "
           "0 disables yielding.");
 
--- a/src/server/generic_family_test.cc
+++ b/src/server/generic_family_test.cc
@@ -2053,8 +2053,6 @@ TEST_F(GenericFamilyTest, RmDeletesMatchingKeys) {
 // behavior is expected for other containers (SET, HASH, LIST) with non
 // listpack encodings.
 TEST_F(GenericFamilyTest, ContainerIterationYields) {
-  GTEST_SKIP() << "Until --container_iteration_yield_interval_usec is re-enabled";
-
   // Build a large sorted set that will be encoded as SKIPLIST.
   constexpr int N = 500'000;
   for (int start = 0; start < N; start += 1'000) {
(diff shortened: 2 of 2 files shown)


src/server/container_utils.cc
```diff
@@ -148,8 +149,9 @@ OpResult<ShardFFResult> FindFirstNonEmpty(Transaction* trans, int req_obj_type)
 void YieldLongIteration(bool allow_yield) {
   static const uint32_t interval_usec =
       absl::GetFlag(FLAGS_container_iteration_yield_interval_usec);
-  if (allow_yield && interval_usec > 0 &&
-      base::CycleClock::ToUsec(util::ThisFiber::GetRunningTimeCycles()) > interval_usec) {
+  uint64_t running_usec = base::CycleClock::ToUsec(util::ThisFiber::GetRunningTimeCycles());
+  if (allow_yield && !ServerState::tlocal()->IsInlinedTxRunning() && interval_usec > 0 &&
```
kostasrim: This is a point in time check that can be violated at any point. Can this be a problem ?

kostasrim:
@mkaruza 

>> If a command runs as an inlined transaction, the shard callback executes directly
on the coordinator fiber instead of being enqueued in the txq.
A mid-iteration yield from the reader would allow the shard-dispatch fiber to run
PollExecution(), potentially executing a conflicting write on the same key while the

What I do not understand here (and don't remember) is why the head (or OOO) is allowed to execute ? inline transactions still acquire intent locks right (it should be because I patched that 2 years ago) ? That probably means we don't have a mechanism in place to reject running the tx (head or OOO) for those cases ?

Also 

"[Suppress yield from container iteration if inlined txn](https://github.com/dragonflydb/dragonfly/pull/7449#top)#7449"

A lot of transactions are inline and this is not a pessimistic optimization that happens rarely. In practice, this means that inline transactions are orthogonal to what you implemented. So maybe, a better solution than suppressing yields is actually figuring out if it's possible to not run those tx's during PollExecution since they are already running inline

kostasrim:
> @mkaruza
> 
> > > If a command runs as an inlined transaction, the shard callback executes directly
> > > on the coordinator fiber instead of being enqueued in the txq.
> > > A mid-iteration yield from the reader would allow the shard-dispatch fiber to run
> > > PollExecution(), potentially executing a conflicting write on the same key while the
> 
> What I do not understand here (and don't remember) is why the head (or OOO) is allowed to execute ? inline transactions still acquire intent locks right (it should be because I patched that 2 years ago) ? That probably means we don't have a mechanism in place to reject running the tx (head or OOO) for those cases ?
> 
> Also
> 
> "[Suppress yield from container iteration if inlined txn](https://github.com/dragonflydb/dragonfly/pull/7449#top)#7449"
> 
> A lot of transactions are inline and this is not a pessimistic optimization that happens rarely. In practice, this means that inline transactions are orthogonal to what you implemented. So maybe, a better solution than suppressing yields is actually figuring out if it's possible to not run those tx's during PollExecution since they are already running inline

I just noticed Roman wrote https://github.com/dragonflydb/dragonfly/issues/7450 which some of its context apply here. In particular what I wrote about `intent locks` + `OOO` and that it could be a better mechanism to forbit those tx in tx queue to execute 

And I would still like to understand why we run the tx while the inline preempts

mkaruza:
Running same bechmark as in PR #7391

**yield=500µs results are averaged over 2 runs.**

**1. ZSET: 100k members**

| Operation | Pipeline | QPS (yield=0) | QPS (yield=500µs) | ΔQPS | p99 ms (yield=0) | p99 ms (yield=500µs) | Δp99 |
|-----------|----------|---------------|-------------------|------|------------------|----------------------|------|
| ZRANGE | 1 | 186,985 | 180,096 | −4% | 9.57 | 11.82 | +24% |
| ZRANGE | 100 | 1,010,139 | 983,426 | −3% | 51.19 | 29.88 | **−42%** |
| ZRANGESTORE | 1 | 13,177 | 15,361 | +17% | 112.71 | 74.09 | **−34%** |
| ZRANGESTORE | 100 | 886,120 | 917,478 | +4% | 58.98 | 35.25 | **−40%** |
| ZDIFFSTORE | 1 | 19,532 | 25,778 | **+32%** | 104.34 | 70.08 | **−33%** |
| ZDIFFSTORE | 100 | 887,402 | 910,834 | +3% | 61.75 | 35.75 | **−42%** |

**2. ZSET: 1M members**

| Operation | Pipeline | QPS (yield=0) | QPS (yield=500µs) | ΔQPS | p99 ms (yield=0) | p99 ms (yield=500µs) | Δp99 |
|-----------|----------|---------------|-------------------|------|------------------|----------------------|------|
| ZRANGE | 1 | 184,337 | 176,539 | −4% | 3.03 | 3.79 | +25% |
| ZRANGE | 100 | 900,992 | 909,790 | +1% | 55.17 | 34.91 | **−37%** |
| ZRANGESTORE | 1 | 19,230 | 6,824 | −65% | 382.76 | 748.50 | +96% |
| ZRANGESTORE | 100 | 332,140 | 891,070 | **+168%** | 1,572.66 | 39.27 | **−98%** |
| ZDIFFSTORE | 1 | 15,236 | 14,726 | −3% | 549.05 | 604.91 | +10% |
| ZDIFFSTORE | 100 | 672,165 | 890,920 | **+33%** | 65.22 | 35.44 | **−46%** |

**3. Heavy operation iterations — ZSET: 100k members**

| Operation | Pipeline | Iterations (yield=0) | Iterations (yield=500µs) | Δ |
|-----------|----------|---------------------|--------------------------|---|
| ZRANGE | 1 | 931 | 917 | −2% |
| ZRANGE | 100 | 16 | 8 | **−50%** |
| ZRANGESTORE | 1 | 1,005 | 947 | −6% |
| ZRANGESTORE | 100 | 46 | 8 | **−83%** |
| ZDIFFSTORE | 1 | 726 | 679 | −6% |
| ZDIFFSTORE | 100 | 23 | 9 | **−61%** |

**4. Heavy operation iterations — ZSET: 1M members**

| Operation | Pipeline | Iterations (yield=0) | Iterations (yield=500µs) | Δ |
|-----------|----------|---------------------|--------------------------|---|
| ZRANGE | 1 | 74 | 24 | **−68%** |
| ZRANGE | 100 | 10 | 5 | −50% |
| ZRANGESTORE | 1 | 82 | 63 | −23% |
| ZRANGESTORE | 100 | 59 | 4 | **−93%** |
| ZDIFFSTORE | 1 | 61 | 43 | −30% |
| ZDIFFSTORE | 100 | 15 | 1 | **−93%** |


**Regressions with yield=500µs:**

- **ZRANGE pipeline=1** (both sizes): p99 worsens ~+24%. ZRANGE is read-only and the bottleneck is serializing results back to the client, not shard contention — yielding adds per-element overhead without relieving real contention.
- **ZRANGESTORE pipeline=1, 1M members**: the most significant regression — QPS drops −65% and p99 worsens +96%.
- **ZDIFFSTORE pipeline=1, 1M members**: slight regression (+10% p99, −3% QPS) — same underlying cause, less pronounced.

romange:
@mkaruza  Lets see if https://github.com/dragonflydb/dragonfly/pull/7459 solves the issue. I find that fix more powerful than here - if it works.

romange:
@mkaruza  please check if https://github.com/dragonflydb/dragonfly/pull/7459 solves the issue for preemptive callbacks.

mkaruza:
@romange  I have tested (locally, several times in debug/release) and it seems it fixes this problem. Update test to verify that LRANGE (with container yielding) returns consistent results. PR is rebased on top #7459.

romange:
please rebase and reenable the flag

romange (approved):

PR #7453, opened 2026-05-29
fix: HNSW index building bug on replica

### LINKED ISSUE #7244: Long HNSW index tests
We need to cover at least the next scenarios for the big HNSW index (100k docs):

- replication during index building (we don't replicate the index during building)
- whole index replication
- disconnect during big index replication
- several replicas 
- new index creation during replication
- dropping index during replication

BorysTheDev:
fixes: #7244

When `FT.CREATE` was journaled to a replica during the LOADING window of a full sync (i.e. between the snapshot cut and the replica reaching ACTIVE state), the new index was registered but never built: `ShardDocIndices::InitIndex` skipped `Rebuild` whenever `gstate != ACTIVE`, the existing `PerformPostLoad`/`RebuildAllIndices` pass had already run, and there was no later trigger. Subsequent `HSET`s replayed for that index returned `nullopt` in `AddDoc` because `indices_` stayed uninitialized — so the keys existed on the replica, but every `FT.SEARCH` returned zero hits. Silent wrong KNN results until the replica was restarted.

## Fix

`journal_emulated` was previously set by both RDB-aux replay and journal replay, so it couldn't distinguish them. De-conflate:

- `LoadSearchCommandFromAux` no longer sets `journal_emulated`. The LOADING state admission bypass in `main_service.cc` is shifted to `is_replicating`,   which every internal-dispatch path already sets — no behavior change for existing callers.
- `journal_emulated` now means exactly what its name says: the command came from `JournalExecutor` or `Replica::ConsumeRedisStream`.
- `ShardDocIndices::InitIndex` gains a `bool is_journal` parameter. The gate becomes `gstate == ACTIVE || is_journal`: direct dispatch rebuilds, journal replayed `FT.CREATE`/`FT.ALTER` rebuilds (bug fixed), RDB-aux replay defers as before.

## Test

`test_hnsw_multi_replica_with_concurrent_index_ops` (small + 100k variants):

1. On master, create `idx_keep` (HNSW, with traffic) and `idx_drop` (HNSW,  ≥1000 docs) before any replica connects.
2. Connect 3 replicas via `REPLICAOF`.
3. In parallel, 0.5–5 s after the replicas connect: drop `idx_drop`, and create+seed a new HNSW `idx_new` (≥500 docs). Both happen while replicas are still inside the LOADING window.
4. Wait for stable sync.
5. Assert every replica converges to `{idx_keep, idx_new}` with the exact same `num_docs` per index as the master.

### DIFF
--- a/src/server/main_service.cc
+++ b/src/server/main_service.cc
@@ -1319,7 +1319,7 @@ std::optional<ErrorReply> Service::VerifyCommandState(const CommandId& cid, CmdA
   const GlobalState gstate = etl.gstate();
   switch (gstate) {
     case GlobalState::LOADING:
-      allowed_by_state = dfly_cntx.journal_emulated || (cid.opt_mask() & CO::LOADING);
+      allowed_by_state = dfly_cntx.is_replicating || (cid.opt_mask() & CO::LOADING);
       break;
     case GlobalState::SHUTTING_DOWN:
       allowed_by_state = false;
--- a/src/server/rdb_load_context.cc
+++ b/src/server/rdb_load_context.cc
@@ -179,7 +179,6 @@ void LoadSearchCommandFromAux(Service* service, std::string&& def, std::string_v
 
   ConnectionContext cntx{nullptr, acl::UserCredentials{}};
   cntx.is_replicating = true;
-  cntx.journal_emulated = true;
   cntx.skip_acl_validation = true;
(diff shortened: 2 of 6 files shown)


BorysTheDev:
augment review

dranikpg (approved):

PR #7455, opened 2026-05-29
fix(search): accept phrases and affixes in parenthesized field conditions

### LINKED ISSUE #7431: `FT.SEARCH` rejects parenthesized quoted-phrase on `TEXT` field with `Query syntax error`
A query of the form `(@<text_field>:("<phrase>"))` - i.e. a parenthesized clause containing a `TEXT` field with a quoted-phrase argument - returns `ERR Query syntax error`. Both pieces work in isolation; only the combination is rejected.

Reproduction:
```
redis-cli -p 6380 FT.CREATE probe ON JSON PREFIX 1 probe: \
    SCHEMA '$.prefix' AS prefix TEXT
redis-cli -p 6380 JSON.SET probe:1 '$' '{"prefix":"hello"}'

# (1) plain quoted phrase - works
redis-cli -p 6380 FT.SEARCH probe '@prefix:"hello"' DIALECT 2
# -> 1) (integer) 1 ...

# (2) parenthesized but unquoted - works
redis-cli -p 6380 FT.SEARCH probe '(@prefix:hello)' DIALECT 2
# -> 1) (integer) 1 ...

# (3) combination - fails
redis-cli -p 6380 FT.SEARCH probe '(@prefix:("hello"))' DIALECT 2
# -> (error) Query syntax error
```

Actual:
`(error) Query syntax error`

Expected:
Same result as `@prefix:"hello"` (1 document).

Impact:
Discovered while running the LangGraph store API (langgraph-checkpoint-redis 0.4.1). `RedisStore._batch_get_ops` builds exactly this shape for every `store.get(ns, key)` call:

```
(@prefix:("u123\.documents") @key:{doc1})
```

The wrapping parens are emitted by `redisvl.FilterExpression` when combining two filter clauses with `&`. As a result, **every** `RedisStore.get()` call fails on Dragonfly.

The same client-side pattern is used by `langchain_redis.RedisVectorStore` for metadata filters, so other agent frameworks may be affected too.

Acceptance criteria:
- `FT.SEARCH probe '(@<text>:("<phrase>"))'` parses successfully and matches the same set of documents as `FT.SEARCH probe '@<text>:"<phrase>"'`.
- The behavior is consistent regardless of how many levels of parens wrap the clause.

vyavdoshenko:
A parenthesized field condition `@field:(...)` only accepted plain terms and integers inside the parentheses, so quoted phrases and prefix/suffix/infix patterns there were rejected with `Query syntax error` -- even though the unparenthesized forms (`@field:"phrase"`, `@field:value`) parse fine. This broke filter expressions that wrap a clause in parentheses, e.g. `(@prefix:("hello") @key:{doc1})`.

The change only adds new alternatives to the grammar (no shift/reduce or reduce/reduce conflicts), so existing queries parse unchanged.

Fixes #7431


dranikpg (dismissed):
There is a big overlap in search_unary_expr, field_cond and field_unary_expr, but I'm not an expert with bison to tell if it can be somehow regrouped and used as one block

vyavdoshenko:
> There is a big overlap in search_unary_expr, field_cond and field_unary_expr, but I'm not an expert with bison to tell if it can be somehow regrouped and used as one block

Good catch. They can't collapse into one rule because they sit at three different grammar positions: only search_unary_expr allows FIELD COLON ... (and a field condition must not nest another @field:), and only field_cond allows the field-only forms * / [range] / {tags}; field_unary_expr is just the boolean-group atom. So the structure has to stay distinct.

What can be deduplicated is the shared leaf set (TERM/PHRASE/PREFIX/SUFFIX/INFIX/WILDCARD/UINT32/DOUBLE) - factor it into a single term_atom nonterminal referenced by all three. Pure refactor, identical AST, no behavior change. It can be done as a small follow-up to keep this fix focused, but there is no speedup. Also, we will save only a few lines.

romange (approved):

PR #7456, opened 2026-05-29
feat(search): support w'...' glob wildcards on text and tag fields

vyavdoshenko:
Adds glob wildcard queries via the `w'...'` syntax in `FT.SEARCH` — on TEXT fields, globally, and inside tag braces (`@tag:{w'...'}`). In the pattern `*` matches any run of characters, `?` matches exactly one, and `\` escapes a metacharacter to a literal.

Examples:
```
w'hel*'            prefix
w'*llo'            suffix
w'*ell*'           infix
w'h?llo'           single-character wildcard
@title:w'mach*'    field-scoped
@tag:{w'doc?'}     inside a tag list
```

Details:
- **Lexer**: a lowercase `w` immediately before a quote starts a wildcard token; an uppercase `W` stays a plain term (the quoted part re-scans as a phrase). One layer of `\` escapes is stripped at lex time, so `w'a\*'` keeps `*` a metacharacter while `w'a\\*'` matches a literal `*`.
- **Grammar**: `WILDCARD` is accepted everywhere an affix (prefix/suffix/infix) is — top level, after `@field:`, inside parentheses, and in tag lists. No new shift/reduce or reduce/reduce conflicts.
- **Matching**: two-pointer glob matcher with single-`*` backtracking (`O(text x pattern)`, no catastrophic blowup). `?` and `*` advance over whole UTF-8 codepoints, so single-character matching works on multibyte text. The dictionary scan is bounded by the pattern's literal prefix and short-circuited through the suffix trie when present (same strategy as infix/suffix matching).
- Works for both TEXT (scored) and TAG fields; accepted under any DIALECT.

### DIFF
--- a/src/core/search/indices.cc
+++ b/src/core/search/indices.cc
@@ -184,6 +184,85 @@ void IterateAllSuffixes(const absl::flat_hash_set<string>& words,
   }
 }
 
+// Literal characters of `pat` up to the first unescaped `*` or `?`. Any term matching `pat` must
+// start with this, so it bounds the dictionary range that has to be scanned.
+string GlobLiteralPrefix(string_view pat) {
+  string prefix;
+  for (size_t i = 0; i < pat.size(); ++i) {
+    char c = pat[i];
+    if (c == '*' || c == '?')
+      break;
+    if (c == '\\' && i + 1 < pat.size())
+      c = pat[++i];
+    prefix.push_back(c);
+  }
+  return prefix;
+}
+
+// Longest run of literal characters in `pat` (bounded by unescaped `*`/`?`), with `\` escapes
+// resolved. Every term matching `pat` must contain this run as a contiguous substring, which the
+// suffix trie can test for a cheap early-exit.
+string GlobLongestLiteralSegment(string_view pat) {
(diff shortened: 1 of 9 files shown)


src/core/search/indices.cc
```diff
@@ -511,6 +590,39 @@ void BaseStringIndex<C>::MatchInfixWithTerm(
...
+  // segment as a substring, so if no term does, skip the dictionary scan. Mirrors
+  // MatchInfixWithTerm.
+  if (suffix_trie_) {
+    string segment = GlobLongestLiteralSegment(pattern);
+    if (!segment.empty()) {
+      auto it = suffix_trie_->lower_bound(segment);
+      if (it == suffix_trie_->end() || !(*it).first.starts_with(segment))
+        return;
+    }
+  }
```
dranikpg, line 610: I imagine that in some cases it's significanlty cheaper to iterate ofer the suffix trie if the longest literal segment is more selective than the entries, for example in cases when the first ? or * is at the start

dranikpg (approved):
LGTM, maybe there are some optimziations possible but I doubt we need them now

Issue #7458, opened 2026-05-29
Transactions: Lazy Locking


romange:
This phase optimizes the optimistic execution path so that intent locks are only acquired when a preemption hazard actually occurs.

### [ ] Task 2.1: Optimize Common Path in `ScheduleInShard`
- Modify `ScheduleInShard` to skip acquiring intent locks entirely when `running_tx_` is `nullptr`.
- The transaction should proceed to run its callback inline and complete without touching the lock table.
- *Files to modify:* `transaction.cc`.

### [ ] Task 2.2: Implement Lazy Locking on the Rare Path (`running_tx_ != nullptr`)
- When a new transaction (Tx B) enters `ScheduleInShard` and detects that another transaction (Tx A, pointed to by `running_tx_`) is in-flight:
  - Retrieve the lock arguments of the running transaction: `running_tx_->GetLockArgs()`.
  - Lazy-acquire the intent locks for Tx A's keys on its behalf (if not already locked).
  - Acquire the intent locks for Tx B's keys.

### [ ] Task 2.3: Conflict Checking and Queueing
- After acquiring the locks on the rare path, run conflict detection for Tx B against the lock table (which now correctly reflects Tx A's keys).
- If keys overlap: Tx B must be inserted into the transaction queue and wait (setting `OUT_OF_ORDER` as appropriate).
- If keys do not overlap: Tx B can proceed out-of-order (OOO) inline.

### [ ] Task 2.4: Ensure Correct Lock Release
- Ensure that if Tx A's keys were lazy-locked on its behalf during step 2.2, those locks are properly cleaned up and released when Tx A finally finishes its callback execution.

PR #7459, opened 2026-05-29
perf(tx): track running transactions per-shard and enable inline scheduling during replication

### LINKED ISSUE #7457: Improve Inline Transactions
This phase focuses on introducing the per-shard tracking of running transactions and using it to safely re-enable inline transaction scheduling during replication.

### [ ] Task 1.1: Extend `EngineShard` with `running_tx_`
- Add a new member variable `Transaction* running_tx_ = nullptr;` to the `EngineShard` class.
- Add accessors (same cooperative thread, no locking needed):
  - `Transaction* running_tx() const`
  - `void set_running_tx(Transaction* tx)`
- *Files to modify:* `engine_shard.h` (or the file defining `EngineShard`), `engine_shard.cc`.

### [ ] Task 1.2: Track Transaction Execution in Callback Paths
- Update `Transaction::RunCallback` to set `shard->running_tx_ = this` before invoking the command callback / journaling, and reset to `nullptr` after.
- Do the same in `Transaction::RunSquashedMultiCb`, which also invokes the callback and journal write paths and would otherwise be invisible to the gate.
- A `DCHECK(shard->running_tx() == nullptr)` at entry guards against accidental nesting.

### [ ] Task 1.3: Re-enable Inline Scheduling, Gated on `running_tx_`
- In `ServerState::AllowInlineScheduling`, remove the `journal::HasRegisteredCallbacks()` early-return. Keep the `LOADING` check (RdbLoader does not use transactions).
- In `Transaction::CanRunInlined`, remove the `DbSlice::HasRegisteredCallbacks()` check **and add** `es->running_tx() == nullptr` to the inline condition.
  - **This is the safety mechanism that replaces the two removed checks.** Without it, two transactions can interleave their journal / db_slice change-callback loops (the loop in `JournalSlice::CallOnChange` is not under `FiberAtomicGuard` — consumer callbacks may suspend), and consumers can observe out-of-order updates.

romange:
## Summary

Replace the broad bans on inline scheduling (any registered journal callback, any registered db_slice change callback) with a precise check: inline a transaction only when no other transaction is currently running on this shard.

## Why

The old guards in `ServerState::AllowInlineScheduling` and `Transaction::CanRunInlined` existed because `RunCallback` can preempt inside `LogAutoJournalOnShard` / journal `CallOnChange` and inside db_slice change callbacks. If a second transaction were scheduled inline on the same shard during that suspension, the two callback iterations would interleave and consumers could observe out-of-order updates.

Tracking the running transaction directly lets us keep inline scheduling enabled in the common case (no in-flight callback on this shard) and disable it only while it is actually unsafe — including during replication, where the old `journal::HasRegisteredCallbacks()` check disabled inlining for every command on the master.

## Changes

- `EngineShard` gains `running_tx_` with `running_tx()` / `set_running_tx()`.
- `Transaction::RunCallback` sets/clears `running_tx_` around the callback body.
- `Transaction::RunSquashedMultiCb` sets/clears `running_tx_` **only in `NON_ATOMIC` mode**. In atomic squashing (`LOCK_AHEAD`/`GLOBAL`) the parent transaction runs `SquashedHopCb` via `ScheduleSingleHop` → `RunCallback` and has already set `running_tx_` to itself; the stub must not overwrite it. In `NON_ATOMIC` mode `SquashedHopCb` is invoked directly without a wrapping `RunCallback`, so the stub owns the slot.
- `Transaction::CanRunInlined` now rejects when `es->running_tx() != nullptr` — this is the actual interleaving prevention.
- `ServerState::AllowInlineScheduling` no longer checks `journal::HasRegisteredCallbacks`; `CanRunInlined` no longer checks `DbSlice::HasRegisteredCallbacks` — both are subsumed by the `running_tx_` gate.

Fixes #7457

### DIFF
--- a/src/server/transaction.cc
+++ b/src/server/transaction.cc
@@ -644,6 +644,8 @@ bool Transaction::RunInShard(EngineShard* shard, bool allow_q_removal) {
 
 void Transaction::RunCallback(EngineShard* shard) {
   DCHECK_EQ(shard, EngineShard::tlocal());
+  DCHECK(shard->running_tx() == nullptr);
+  shard->set_running_tx(this);
 
   RunnableResult result;
   try {
@@ -686,6 +688,8 @@ void Transaction::RunCallback(EngineShard* shard) {
     LogAutoJournalOnShard(shard, result);
     MaybeInvokeTrackingCb();
   }
+
+  shard->set_running_tx(nullptr);
 }
 
 // TODO: For multi-transactions we should be able to deduce mode() at run-time based
@@ -727,6 +731,9 @@ void Transaction::ScheduleInternal() {
       // single shard schedule operation can't fail
       CHECK(ScheduleInShard(EngineShard::tlocal(), optimistic_exec));
       run_barrier_.Dec();
+
+      // Drain any work that was deferred while this inlined tx held running_tx_.
+      EngineShard::tlocal()->PollExecution("after_inline", nullptr);
       break;
     }
 
@@ -1005,12 +1012,13 @@ bool Transaction::CancelScheduledTx() {
   });
 
   // Issue PollExecutions to shards that were disarmed
-  auto poll_disarmed = [this, &disarmed_shards](const char* context) {
+  auto poll_disarmed = [this, &disarmed_shards](bool rearm) {
     IterateActiveShards([&](auto& sd, ShardId i) {
       if (disarmed_shards.test(i)) {
         use_count_.fetch_add(1, memory_order_relaxed);
-        shard_set->Add(i, [this, context] {
-          EngineShard::tlocal()->PollExecution(context, this);
+        shard_set->Add(i, [this, rearm] {
+          EngineShard::tlocal()->PollExecution(rearm ? "cancel_rearm" : "cancel_cancel",
+                                               rearm ? this : nullptr);
           intrusive_ptr_release(this);
         });
       }
@@ -1231,8 +1239,12 @@ bool Transaction::ScheduleInShard(EngineShard* shard, bool execute_optimistic) {
 
     DVLOG(3) << "Lock granted " << lock_granted << " for trans " << DebugId();
 
-    // Check if we can run immediately
-    if (lock_granted && execute_optimistic) {
+    // Check if we can run immediately. Skip if another transaction is already running on this shard
+    // (e.g., an inlined tx preempted). While lock_granted ensures no contention on keys, we must
+    // maintain the "running_tx == nullptr" invariant,
+    // since we only track one active transaction per shard.
+    bool immediate_run = execute_optimistic && lock_granted && shard->running_tx() == nullptr;
+    if (immediate_run) {
       sd.local_mask |= OPTIMISTIC_EXECUTION;
       shard->stats().tx_optimistic_total++;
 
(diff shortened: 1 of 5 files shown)


src/server/transaction.cc
```diff
@@ -1059,7 +1066,7 @@ bool Transaction::CancelScheduledTx() {
       if (disarmed_shards.test(i)) {
         use_count_.fetch_add(1, memory_order_relaxed);
         shard_set->Add(i, [this, context] {
-          EngineShard::tlocal()->PollExecution(context, this);
+          EngineShard::tlocal()->PollExecution(context, nullptr);
```
dranikpg: no, if its called from `poll_disarmed("cancel_rearm");` then it has to be this, otherwise it can be nullptr

dranikpg (commented):
LGTM other than the PollExecution call

src/server/transaction.cc
```diff
@@ -1280,8 +1287,9 @@ bool Transaction::ScheduleInShard(EngineShard* shard, bool execute_optimistic) {
 
     DVLOG(3) << "Lock granted " << lock_granted << " for trans " << DebugId();
 
-    // Check if we can run immediately
-    if (lock_granted && execute_optimistic) {
+    // Check if we can run immediately. Skip if another transaction is already running
+    // on this shard (e.g. an inlined tx preempted on the connection fiber).
+    if (lock_granted && execute_optimistic && shard->running_tx() == nullptr) {
       sd.local_mask |= OPTIMISTIC_EXECUTION;
       shard->stats().tx_optimistic_total++;
```
dranikpg: that's a little pessimistic because we do acquire locks here

src/server/transaction.cc
dranikpg: ok, i'll just fix it later 🙂

src/server/transaction.cc
```diff
@@ -1518,12 +1526,24 @@ OpStatus Transaction::RunSquashedMultiCb(RunnableType cb) {
...
   auto& db_slice = GetDbSlice(shard->shard_id());
 
+  // In NON_ATOMIC mode SquashedHopCb is invoked directly, without a wrapping RunCallback,
+  // so this stub owns the running_tx_ slot. In atomic modes (LOCK_AHEAD / GLOBAL) the
+  // parent transaction's RunCallback has already set running_tx_ to itself.
+  bool owns_running_tx = multi_->mode == NON_ATOMIC;
+  if (owns_running_tx) {
+    DCHECK(shard->running_tx() == nullptr);
+    shard->set_running_tx(this);
+  }
```
dranikpg, line 1452: I need to check, but I think in our codebase there is no such thing as a non-atomic squashed stub

src/server/transaction.cc
```diff
@@ -1518,12 +1526,24 @@ OpStatus Transaction::RunSquashedMultiCb(RunnableType cb) {
...
   db_slice.OnCbFinishBlocking();
 
   LogAutoJournalOnShard(shard, result);
   MaybeInvokeTrackingCb();
 
+  if (owns_running_tx)
+    shard->set_running_tx(nullptr);
+
   DCHECK_EQ(result.flags, 0);  // if it's sophisticated, we shouldn't squash it
   return result;
```
dranikpg, line 1464: if it was atomic, then the carrier transaction should follow up with pollexecution? also I though it will set itself to the running_tx as the hops are executed inside of it

src/server/transaction.cc
romange: Pessimistic as we will take slow path for executions when inline transaction preempt?
Given that I plan to follow up with https://github.com/dragonflydb/dragonfly/issues/7458 it will have to change anyways. I will add a TODO though.

src/server/transaction.cc
romange: I fixed it. PTAL

src/server/transaction.cc
romange: pipeline squashing is : `MultiCommandSquasher::ExecuteSquashed`.

src/server/transaction.cc
romange: The invariant is: there is at most one RunCallback at the same time. `SquashedHopCb` complicates the state a bit but at high level this is the invariant that I am trying to preserve for clarity.

dranikpg: > The invariant is: there is at most one RunCallback at the same time

But there are "nested" callback active in theory

src/server/transaction.cc
dranikpg: and you will see that local_tx is just a regular transaction

dranikpg: the way it is created `MultiCommandSquasher::PrepareShardInfo` here

dranikpg (approved):

PR #7460, opened 2026-05-30
fix(serializer): allow inline-scheduled tx on DflyConn fibers

dragonclaw-dragonflydb:
Fixes the serializer_base_test failure in #7459.

With inline scheduling enabled during replication, connection fibers (DflyConn_*) can now trigger OnChangeBlocking directly. Adds the DflyConn prefix to the fiber name allowlist. Safe because the running_tx_ gate prevents concurrent callback interleaving.

Signed-off-by: Dragon Claw <dragonclaw@dragonflydb.io>

### DIFF
--- a/src/server/engine_shard.h
+++ b/src/server/engine_shard.h
@@ -167,6 +167,14 @@ class EngineShard {
     return continuation_trans_;
   }
 
+  Transaction* running_tx() const {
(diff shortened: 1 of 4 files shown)


dragonclaw-dragonflydb:
Reopening with correct base branch

PR #7462, opened 2026-05-30
refactor(server): simplify DbSlice::ExpireParams

### LINKED ISSUE #3634: refactor DbSlice::ExpireParams
we have several fields in DbSlice::ExpireParams:
    int64_t value = INT64_MIN;  // undefined
    TimeUnit unit = TimeUnit::SEC;
    bool absolute = false;

All these fields can be exchanged with one int64_t ms_timestamp, that will be created in the constructor or method.

archief2910:
### Description
This PR simplifies `DbSlice::ExpireParams` by replacing three tightly-coupled fields (`value`, `unit`, `absolute`) with a single `int64_t ms_timestamp` that is always in milliseconds.

**Changes:**
* Removed `value`, `unit`, and `absolute` fields from `ExpireParams`.
* Added a single `ms_timestamp` field (always in milliseconds).
* Added factory methods (`Relative()`, `Absolute()`, `Persist()`) to enforce correct construction and handle unit conversions safely at the caller boundary.
* Added overflow checks to the factory methods, which properly screens out invalid huge expiration values, fixing bugs where huge `EX` values would bypass validation and wrap around or overflow.

Fixes #3634


src/server/db_slice.cc
```diff
@@ -1130,25 +1130,46 @@ OpResult<DbSlice::ItAndUpdater> DbSlice::AddNew(const Context& cntx, string_view
...
+
+optional<DbSlice::ExpireParams> DbSlice::ExpireParams::Absolute(int64_t value, TimeUnit unit) {
+  if (unit == TimeUnit::SEC && value > INT64_MAX / 1000) {
+    return nullopt;
+  }
+  ExpireParams p;
+  p.ms_timestamp = (unit == TimeUnit::SEC) ? value * 1000 : value;
+  p.absolute = true;
+  return p;
+}
```
BorysTheDev: code are the same except 1 line

src/server/db_slice.cc
```diff
@@ -1130,25 +1130,46 @@ OpResult<DbSlice::ItAndUpdater> DbSlice::AddNew(const Context& cntx, string_view
...
+
+DbSlice::ExpireParams DbSlice::ExpireParams::Persist() {
+  ExpireParams p;
+  p.persist = true;
+  return p;
```
BorysTheDev: looks redundant, you can have the same with the ExpireParams constructor

src/server/db_slice.cc
archief2910: ok will update try to resolve this asap.

src/server/db_slice.h
```diff
@@ -221,24 +221,25 @@ class DbSlice {
...
     bool IsExpired(uint64_t now_msec) const {
       return Calculate(now_msec, false).first < 0;
     }
 
-   public:
-    int64_t value = INT64_MIN;  // undefined
-    TimeUnit unit = TimeUnit::SEC;
-
+    int64_t ms_timestamp = INT64_MIN;  // undefined; always in milliseconds
     bool absolute = false;
```
BorysTheDev: the idea was to remove this field too, and always store absolute value

BorysTheDev:
@archief2910 I would like to see timestamp transforming encapsulation in one place, for example, in ExpireParams. to provide a cleaner interface and code reduction in places where we use expiration.

archief2910:
> @archief2910 I would like to see timestamp transforming encapsulation in one place, for example, in ExpireParams. to provide a cleaner interface and code reduction in places where we use expiration.

sure , i will try to do that and push them shortly .

archief2910:
@BorysTheDev , i have pushed the refactored approach . feel free to check . if there are any discrepancies i would be happy to resolve it .

src/server/db_slice.cc
```diff
@@ -1130,25 +1130,59 @@ OpResult<DbSlice::ItAndUpdater> DbSlice::AddNew(const Context& cntx, string_view
   return DbSlice::ItAndUpdater{.it = res.it, .post_updater = std::move(res.post_updater)};
 }
 
-int64_t DbSlice::ExpireParams::Cap(int64_t value, TimeUnit unit) {
-  return unit == TimeUnit::SEC ? min(value, kMaxExpireDeadlineSec)
-                               : min(value, kMaxExpireDeadlineMs);
+optional<DbSlice::ExpireParams> DbSlice::ExpireParams::Absolute(int64_t value, TimeUnit unit) {
+  if (unit == TimeUnit::SEC && (value > INT64_MAX / 1000 || value < INT64_MIN / 1000)) {
```
BorysTheDev: what does it mean if value is negative? I don't understand the check value < INT64_MIN / 1000

src/server/db_slice.cc
```diff
@@ -1130,25 +1130,59 @@ OpResult<DbSlice::ItAndUpdater> DbSlice::AddNew(const Context& cntx, string_view
...
+    return nullopt;
+  }
+
+  int64_t ms_timestamp = (unit == TimeUnit::SEC) ? value * 1000 : value;
+  if (ms_timestamp == kUndefined) {
```
BorysTheDev: how is this possible

src/server/db_slice.cc
```diff
@@ -1130,25 +1130,59 @@ OpResult<DbSlice::ItAndUpdater> DbSlice::AddNew(const Context& cntx, string_view
...
+optional<DbSlice::ExpireParams> DbSlice::ExpireParams::Relative(int64_t value, TimeUnit unit,
+                                                                uint64_t now_ms) {
+  auto params = Absolute(value, unit);
+  if (!params || now_ms > INT64_MAX ||
+      params->ms_timestamp > INT64_MAX - static_cast<int64_t>(now_ms)) {
+    return nullopt;
+  }
+
+  params->ms_timestamp += static_cast<int64_t>(now_ms);
+  if (params->ms_timestamp == kUndefined) {
```
BorysTheDev: too complex

src/server/db_slice.cc
```diff
@@ -1130,25 +1130,59 @@ OpResult<DbSlice::ItAndUpdater> DbSlice::AddNew(const Context& cntx, string_view
...
+    return nullopt;
+  }
+  return params;
+}
+
+DbSlice::ExpireParams DbSlice::ExpireParams::Persist() {
+  ExpireParams params;
+  params.persist = true;
+  return params;
```
BorysTheDev: still don't understand why do you need factory method

src/server/db_slice.h
```diff
@@ -221,27 +221,42 @@ class DbSlice {
...
 
   struct ExpireParams {
+    ExpireParams() = default;
+
+    // Factory methods for constructing ExpireParams.
+    // Accept raw value + unit; conversion to ms happens internally. Relative expiration also
+    // accepts a reference time so its timestamp stays consistent with the transaction context.
+    // Returns nullopt if the value overflows int64_t when converted to milliseconds.
+    static std::optional<ExpireParams> Relative(int64_t value, TimeUnit unit, uint64_t now_ms);
+    static std::optional<ExpireParams> Absolute(int64_t value, TimeUnit unit);
```
BorysTheDev: I think we don't need factory methods, we can use min(value, kMaxExpireDeadlineSec) and always get correct results and use constructors

src/server/generic_family.cc
```diff
@@ -1390,8 +1391,12 @@ void GenericFamily::ExpireAt(CmdArgList args, CommandContext* cmd_cntx) {
   if (!expire_options) {
     return cmd_cntx->SendError(expire_options.error());
   }
-  DbSlice::ExpireParams params{
-      .value = int_arg, .absolute = true, .expire_options = expire_options.value()};
+  auto opt_params = DbSlice::ExpireParams::Absolute(int_arg, TimeUnit::SEC);
+  if (!opt_params) {
+    return cmd_cntx->SendError(kExpiryOutOfRange);
+  }
```
BorysTheDev: it looks like you have changed behavior, but the task was only refactoring

src/server/generic_family.cc
```diff
@@ -1439,10 +1444,12 @@ void GenericFamily::PexpireAt(CmdArgList args, CommandContext* cmd_cntx) {
...
     return cmd_cntx->SendError(expire_options.error());
   }
-  DbSlice::ExpireParams params{.value = int_arg,
-                               .unit = TimeUnit::MSEC,
-                               .absolute = true,
-                               .expire_options = expire_options.value()};
+  auto opt_params = DbSlice::ExpireParams::Absolute(int_arg, TimeUnit::MSEC);
+  if (!opt_params) {
+    return cmd_cntx->SendError(kExpiryOutOfRange);
+  }
```
BorysTheDev: the same

src/server/string_family.cc
```diff
@@ -1514,8 +1527,10 @@ cmd::CmdR CmdGAT(CmdArgList args, CommandContext* cmd_cntx) {
     return cmd::kAborted;
   }
   int64_t expire_ts = cmd_cntx->mc_command()->expire_ts;
-  DbSlice::ExpireParams expire_params{
-      .value = expire_ts, .absolute = true, .persist = expire_ts == 0};
+  auto expire_params =
+      expire_ts == 0 ? DbSlice::ExpireParams::Persist()
+                     : DbSlice::ExpireParams::Absolute(expire_ts, TimeUnit::SEC)
+                           .value_or(DbSlice::ExpireParams::Persist());
```
BorysTheDev: The code becomes more complex

BorysTheDev:
I expect to see the constructor like ExpireParams(TimeUnit unit, int64_t value, int64_t now_ms = 0, bool persist = false)

if now_ms is 0 it is the absolute value already; otherwise relative. 
Maybe you need one more constructor, but in the most cases I think this will be main.

The result of your work should be the simpler code.

src/server/db_slice.cc
```diff
@@ -1130,25 +1130,19 @@ OpResult<DbSlice::ItAndUpdater> DbSlice::AddNew(const Context& cntx, string_view
...
   }
 
-  int64_t msec = (unit == TimeUnit::SEC) ? value * 1000 : value;
-  int64_t rel_msec = absolute ? msec - now_ms : msec;
-  if (cap)
-    rel_msec = Cap(rel_msec, TimeUnit::MSEC);
-  return make_pair(rel_msec, now_ms + rel_msec);
+  ms_timestamp = unit == TimeUnit::SEC ? value * 1000 : value;
+  if (ms_timestamp > INT64_MAX - now_ms) {
+    ms_timestamp = -1;
```
BorysTheDev: As I wrote previously you changed the behavior, the previous code was 
`return unit == TimeUnit::SEC ? min(value, kMaxExpireDeadlineSec)
                               : min(value, kMaxExpireDeadlineMs);`
why it is now -1?

src/server/db_slice.h
```diff
@@ -221,25 +221,20 @@ class DbSlice {
   using DocDeletionCallback = std::function<void(std::string_view, const Context&, PrimeValue& pv)>;
 
   struct ExpireParams {
+    ExpireParams() = default;
+    // now_ms is zero for absolute timestamps and the reference time for relative expiration.
+    ExpireParams(TimeUnit unit, int64_t value, int64_t now_ms = 0, bool persist = false);
+
     bool IsDefined() const {
-      return persist || value > INT64_MIN;
+      return persist || ms_timestamp > INT64_MIN;
```
BorysTheDev: I see it was previous check, but can it be more strict ms_timestamp > 0 ? please check

src/server/string_family.cc
```diff
@@ -1011,22 +1011,22 @@ std::variant<SetCmd::SetParams, facade::ErrorReply, NegativeExpire> ParseSetPara
...
-
       int64_t now_ms = GetCurrentTimeMs();
-      auto [rel_ms, abs_ms] = expiry.Calculate(now_ms, false);
+      bool absolute = *exp_type == ExpT::EXAT || *exp_type == ExpT::PXAT;
+      TimeUnit unit =
+          *exp_type == ExpT::PX || *exp_type == ExpT::PXAT ? TimeUnit::MSEC : TimeUnit::SEC;
+      DbSlice::ExpireParams expiry{unit, int_arg, absolute ? 0 : now_ms};
+      
+      int64_t abs_ms = expiry.ms_timestamp;
+      int64_t rel_ms = abs_ms - now_ms;
```
BorysTheDev: I prefer to see the code similar to the previous one `auto [rel_ms, abs_ms] = expiry.Calculate(now_ms);`

src/server/string_family.cc
```diff
@@ -1137,20 +1137,17 @@ cmd::CmdR CmdSetExGeneric(CmdArgList args, CommandContext* cmd_cntx) {
...
+  DbSlice::ExpireParams expiry{unit, exp_int, now_ms};
+  
+  int64_t abs_ms = expiry.ms_timestamp;
   if (abs_ms < 0)
     co_return facade::ErrorReply{InvalidExpireTime("set")};
 
   SetCmd::SetParams sparams;
   sparams.flags |= SetCmd::SET_EXPIRE_AFTER_MS;
-  sparams.expire_after_ms = expiry.Calculate(now_ms, true).first;
+  sparams.expire_after_ms = std::min(abs_ms - now_ms, kMaxExpireDeadlineMs);
```
BorysTheDev: abs_ms should already be adjust on kMaxExpireDeadlineMs

BorysTheDev:
@archief2910  looks better, but still has a lot of problems. Please read my comments carefully, as well as the previous ones and try to understand the task before implementation

archief2910:
> @archief2910 looks better, but still has a lot of problems. Please read my comments carefully, as well as the previous ones and try to understand the task before implementation

Sure

src/server/db_slice.h
```diff
@@ -221,26 +221,26 @@ class DbSlice {
...
-    int64_t value = INT64_MIN;  // undefined
-    TimeUnit unit = TimeUnit::SEC;
+    // Calculate relative and absolute timepoints.
+    std::pair<int64_t, int64_t> Calculate(int64_t now_ms) const {
+      return {ms_timestamp - now_ms, ms_timestamp};
+    }
 
-    bool absolute = false;
-    bool persist = false;        // persist means remove all expiry
+    int64_t ms_timestamp = 0;  // Undefined (0 means not set); always in milliseconds.
```
BorysTheDev: can we use uint64_t? check how it affects all the code and do we need any explicit casts before changes? If there are minimal changes, please use uint64_t

src/server/db_slice.h
```diff
@@ -221,26 +221,26 @@ class DbSlice {
...
     bool IsExpired(uint64_t now_msec) const {
-      return Calculate(now_msec, false).first < 0;
+      return ms_timestamp < now_msec;
     }
 
-   public:
-    int64_t value = INT64_MIN;  // undefined
-    TimeUnit unit = TimeUnit::SEC;
+    // Calculate relative and absolute timepoints.
+    std::pair<int64_t, int64_t> Calculate(int64_t now_ms) const {
```
BorysTheDev: We lost `bool cap` parameter, I think, because of my unclear previous comment. Please return it

archief2910: Yes that was what  i was thinking yesterday. So we want that cap Boolean as a parameter too right

src/server/db_slice.h
archief2910: Sure will do and update

src/server/db_slice.cc
```diff
@@ -1130,25 +1130,13 @@ OpResult<DbSlice::ItAndUpdater> DbSlice::AddNew(const Context& cntx, string_view
...
-
-  int64_t msec = (unit == TimeUnit::SEC) ? value * 1000 : value;
-  int64_t rel_msec = absolute ? msec - now_ms : msec;
-  if (cap)
-    rel_msec = Cap(rel_msec, TimeUnit::MSEC);
-  return make_pair(rel_msec, now_ms + rel_msec);
+DbSlice::ExpireParams::ExpireParams(TimeUnit unit, int64_t value, int64_t now_ms, bool persist)
+    : persist(persist) {
+  // Cap the value to avoid overflow, matching the original behavior.
+  int64_t capped = unit == TimeUnit::SEC ? min(value, kMaxExpireDeadlineSec)
```
BorysTheDev: kMaxExpireDeadlineSec and kMaxExpireDeadlineMs are applicable only to relative values, as I understand.
I suggest next:
```
uint64_t ms_value = value;
constexpr auto max_v = std::numeric_limits<decltype(ms_timestamp)>::max();
if (unit == TimeUnit::SEC) {
    ms_value = (ms_value > (max_v / 1000)) ? max_v : ms_value * 1000;
}
ms_timestamp = ms_value > (max_v - now_ms) ? max_v : ms_value += now_ms;
```
In this case, you will have the maximum value that allows you to support the cap in the Calculate function. you can define kMaxV as a private member if you want.
This code is just a suggestion that you need to check for correctness and apply the correct code style.

The main task is to have the same behavior as before, with cleaner code

BorysTheDev:
@archief2910 the code is clean now, but the behavior is a little different, let's fix it. please check all invariants.

src/server/db_slice.cc
```diff
@@ -1132,11 +1132,23 @@ OpResult<DbSlice::ItAndUpdater> DbSlice::AddNew(const Context& cntx, string_view
...
-  int64_t capped = unit == TimeUnit::SEC ? min(value, kMaxExpireDeadlineSec)
-                                         : min(value, kMaxExpireDeadlineMs);
-  ms_timestamp = unit == TimeUnit::SEC ? capped * 1000 : capped;
-  ms_timestamp += now_ms;
+  uint64_t ms_value = value;
+  constexpr auto kMaxV = std::numeric_limits<decltype(ms_timestamp)>::max();
+  if (unit == TimeUnit::SEC) {
+    ms_value = (ms_value > (kMaxV / 1000)) ? kMaxV : ms_value * 1000;
+  }
+  uint64_t uval_now_ms = now_ms;
```
BorysTheDev: Instead of creating a new var you can declare now_ms as uint64_t

src/server/string_family.cc
```diff
@@ -1011,22 +1011,20 @@ std::variant<SetCmd::SetParams, facade::ErrorReply, NegativeExpire> ParseSetPara
       if (int_arg <= 0)
         return facade::ErrorReply{InvalidExpireTime("set")};
 
-      DbSlice::ExpireParams expiry{
-          .value = int_arg,
-          .unit = *exp_type == ExpT::PX || *exp_type == ExpT::PXAT ? TimeUnit::MSEC : TimeUnit::SEC,
-          .absolute = *exp_type == ExpT::EXAT || *exp_type == ExpT::PXAT,
-      };
-
       int64_t now_ms = GetCurrentTimeMs();
```
BorysTheDev: I think now_ms can be uint64_t because we have next declaration `uint64_t GetCurrentTimeMs()` and it will be simpler for you to fix the interface, and use now_ms a uint64_t

src/server/db_slice.cc
```diff
@@ -1130,25 +1130,25 @@ OpResult<DbSlice::ItAndUpdater> DbSlice::AddNew(const Context& cntx, string_view
...
 
-pair<int64_t, int64_t> DbSlice::ExpireParams::Calculate(uint64_t now_ms, bool cap) const {
-  if (persist)
-    return {0, 0};
-
-  // return a negative absolute time if we overflow.
-  if (unit == TimeUnit::SEC && value > INT64_MAX / 1000) {
-    return {0, -1};
+std::pair<int64_t, int64_t> DbSlice::ExpireParams::Calculate(int64_t now_ms, bool cap) const {
+  int64_t rel = static_cast<int64_t>(ms_timestamp) - now_ms;
```
BorysTheDev: it's not a ptoblem, but you don't need static_cast for simple types

src/server/db_slice.cc
```diff
@@ -1130,25 +1130,25 @@ OpResult<DbSlice::ItAndUpdater> DbSlice::AddNew(const Context& cntx, string_view
...
 }
 
-pair<int64_t, int64_t> DbSlice::ExpireParams::Calculate(uint64_t now_ms, bool cap) const {
-  if (persist)
-    return {0, 0};
-
-  // return a negative absolute time if we overflow.
-  if (unit == TimeUnit::SEC && value > INT64_MAX / 1000) {
-    return {0, -1};
+std::pair<int64_t, int64_t> DbSlice::ExpireParams::Calculate(int64_t now_ms, bool cap) const {
```
BorysTheDev, line 1143: take into accout that previously Calculate returns  return {0, -1}; if overflow, in our case overflow is kMaxV

BorysTheDev:
@archief2910, please do rebase, it could be conflicts with your changes

BorysTheDev:
don't forget to use pre-commit hooks and sign your commits

archief2910:
@BorysTheDev should i now rebase my branch or there are some more changes that i need to do

BorysTheDev:
@archief2910 you haven't addressed all my comments, for example regarding  DbSlice::ExpireParams::Calculate().
I think after rebase it could be a build issue, so do rebase and check that everything works ok and build is correct. Let's do this round last, I can't spend so much time doing the same review

archief2910:
> ents, for example regarding DbSlice::ExpireParams::Calculate().
> I think after rebase it could be a build issue, so do rebase and check that everything works ok and build is correct. Let's do this round last, I can't spend so much time doing the same review

ok

archief2910:
i ran precommit hooks and ran ci-tests through github actions it worked.

BorysTheDev:
@archief2910 have you seen your diff? Sorry I can't proceed to spend so much time

archief2910:
> @archief2910 have you seen your diff? Sorry I can't proceed to spend so much time

Those changes are coming due to pre commit hook formating

BorysTheDev:
> > @archief2910 have you seen your diff? Sorry I can't proceed to spend so much time
> 
> Those changes are coming due to pre commit hook formating

we run it for every commit, the problem I think is because you have done a merge instead of a rebase, as I asked.

PR #7468, opened 2026-05-31
chore(server): Unify SerializeEntry calls in SerializerBase

dranikpg:
Create a single `SerializeEntryLocked` function that is used from both SliceSnapshot and RestoreStreamer to unify the logic of extracting tiered entries, acessing cooled values, extracing expire time and mc flags, etc...


dranikpg:
augment review

src/server/serializer_base_test.cc
```diff
@@ -202,13 +210,12 @@ void TestDriver::Loop() {
         ProcessBucket(snapshot_db_indx, it, false);
       });
 
-      util::ThisFiber::Yield();
+      // Simualte yield due to socket flushes
+      for (unsigned i = 0; i < 2; ++i)
+        util::ThisFiber::Yield();
```
abhijat, line 215: why are there exactly two yields here

abhijat (approved):

PR #7474, opened 2026-06-01
refactor: simplify ExpireParams by removing TimeUnit field

### LINKED ISSUE #3634: refactor DbSlice::ExpireParams

Priyanshuraturi28:
## Summary
Removes the `TimeUnit` enum from `ExpireParams` and renames `value` to `msec_val` (always in milliseconds).

## Problem
`ExpireParams` had three interrelated fields (`value`, `unit`, `absolute`) where `value` could represent either seconds or milliseconds depending on `unit`. This ambiguity made the API error-prone.

## Solution
- `value` (int64_t, ambiguous unit) → `msec_val` (int64_t, always milliseconds)
- `unit` (TimeUnit enum) → REMOVED
- `absolute` (bool) → KEPT (callers do not have `now_ms` at construction time)

## Changes
| File | Changes |
|------|---------|
| `db_slice.h` | Updated struct definition, IsDefined(), Cap() signature |
| `db_slice.cc` | Simplified Calculate() and Cap() — no more unit conversion |
| `generic_family.cc` | Updated 4 call sites to pre-convert seconds to ms |
| `string_family.cc` | Updated 4 call sites to pre-convert seconds to ms |

Fixes #3634

### DIFF
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -1130,24 +1130,17 @@ OpResult<DbSlice::ItAndUpdater> DbSlice::AddNew(const Context& cntx, string_view
   return DbSlice::ItAndUpdater{.it = res.it, .post_updater = std::move(res.post_updater)};
 }
 
-int64_t DbSlice::ExpireParams::Cap(int64_t value, TimeUnit unit) {
-  return unit == TimeUnit::SEC ? min(value, kMaxExpireDeadlineSec)
-                               : min(value, kMaxExpireDeadlineMs);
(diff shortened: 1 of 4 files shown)


BorysTheDev:
see others' PRs and comments on them; the task was solved incorrectly

PR #7475, opened 2026-06-01
fix: enable rdb_sbf_chunked by default and fix sbf serialization bug

romange:
## Summary
- Flip `rdb_sbf_chunked` flag default from `false` to `true` — the chunked SBF save format has been validated.
- Remove explicit flag overrides in `rdb_test.cc` and `replication_test.py` that are no longer needed.

#The bug

the chunked SBF path saved each filter chunk via SaveString, which integer-encodes short strings and LZF-compresses long ones in SINGLE_ENTRY mode. The loader reads chunks with a plain LoadLen + FetchBuf and can't decode that. A fresh filter is mostly zeros and compresses well, so COPY/DUMP/RESTORE (SINGLE_ENTRY) produced payloads the loader rejected, crashing in Renamer::FinalizeRename with deserialize_status=INVALID_VALUE. The on-disk RDB save (also SINGLE_ENTRY) had the same latent issue.

No, we don't break compatibility, and an old replica can still load SBF from a new master:

The fix only touches the save side. The loader is unchanged and always expected verbatim length-prefixed chunks.
Replication/snapshot run in MULTI_ENTRY mode, where SaveString already wrote chunks verbatim (LZF only triggers in SINGLE_ENTRY; whole-blob compression is separate and transparent to the loader). So the bytes on the wire for SBF2 chunks are identical before and after this change.
The only outputs that change are the SINGLE_ENTRY paths (DUMP/COPY/RESTORE, on-disk RDB), which previously produced payloads no version could load. There's no valid prior format to break there.
The flag still selects RDB_TYPE_SBF vs RDB_TYPE_SBF2; this only fixes the chunk encoding inside SBF2.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -35,6 +35,7 @@
    `ninja <unit_test> && ./unit_test`
 4. ✅ **Format Code** - `pre-commit run --files <files>`
 5. ✅ **Follow Architecture** - See [Architecture Patterns](#architecture-patterns) below
+6. ✅ **Never Push to Main** - Always create a feature branch and open a PR. Never run `git push origin main`.
 
 ### Pull Request Guidelines
 
--- a/src/server/bloom_family_test.cc
+++ b/src/server/bloom_family_test.cc
@@ -145,4 +145,16 @@ TEST_F(BloomFamilyTest, LoadChunkErrors) {
   EXPECT_THAT(Run({"bf.loadchunk", "b1", "-1", "data"}), ErrArg("not an integer"));
 }
 
+// COPY of an SBF must survive the chunked serialize/deserialize round-trip.
+TEST_F(BloomFamilyTest, CopyChunkedRoundTrip) {
+  Run({"bf.reserve", "b1", "0.01", "1000"});
+  for (int i = 0; i < 100; ++i)
+    Run({"bf.add", "b1", absl::StrCat("item", i)});
+
+  EXPECT_THAT(Run({"copy", "b1", "b2"}), IntArg(1));
+
+  for (int i = 0; i < 100; ++i)
+    EXPECT_THAT(Run({"bf.exists", "b2", absl::StrCat("item", i)}), IntArg(1));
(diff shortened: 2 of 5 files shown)


romange:
@vyavdoshenko  what is the bug? Do we break compatibility now with the previous format when the flag is on?
Meaning will old replica be able to load sbf from this master?

@abhijat  can you please review as well?

abhijat (approved):
the change looks good, the PR description is missing that the big change is in the saver format.

I think there should be no compatibility issues in replication, since in case of new master -> old replica : now new master sends the same format as replica always expected to load

vyavdoshenko:
The bug: the chunked SBF path saved each filter chunk via SaveString, which integer-encodes short strings and LZF-compresses long ones in SINGLE_ENTRY mode. The loader reads chunks with a plain LoadLen + FetchBuf and can't decode that. A fresh filter is mostly zeros and compresses well, so COPY/DUMP/RESTORE (SINGLE_ENTRY) produced payloads the loader rejected, crashing in Renamer::FinalizeRename with deserialize_status=INVALID_VALUE. The on-disk RDB save (also SINGLE_ENTRY) had the same latent issue.

No, we don't break compatibility, and an old replica can still load SBF from a new master:

- The fix only touches the save side. The loader is unchanged and always expected verbatim length-prefixed chunks.
- Replication/snapshot run in MULTI_ENTRY mode, where SaveString already wrote chunks verbatim (LZF only triggers in SINGLE_ENTRY; whole-blob compression is separate and transparent to the loader). So the bytes on the wire for SBF2 chunks are identical before and after this change.
- The only outputs that change are the SINGLE_ENTRY paths (DUMP/COPY/RESTORE, on-disk RDB), which previously produced payloads no version could load. There's no valid prior format to break there.
- The flag still selects RDB_TYPE_SBF vs RDB_TYPE_SBF2; this only fixes the chunk encoding inside SBF2.

PR #7477, opened 2026-06-01
fix(server): Exception based command cancelling

dranikpg:
With the V1 loop, we don't have an as nice mechanism as with V2 and we can only "kick" out the stack by cancelling the transaction and letting it _throw a runtime exception._ InvokeCmd catches it, it is also handed in asynchronous commands running in synchronous mode

Works quite well in tests and manual tests, need to check corner cases


src/server/conn_context.cc
```diff
@@ -204,8 +206,22 @@ size_t ConnectionContext::UsedMemory() const {
...
+  if (!transaction->IsScheduled())
+    return;
+
+  // Try cancelleding a scheduled transaction
+  // Dispatch because CancelScheduledTx might fiber-block.
+  boost::intrusive_ptr<Transaction> tx_alive = transaction;
+  util::ProactorBase::me()->Dispatch([this, tx_alive] {
+    if (tx_alive->IsScheduled() && !tx_alive->Blocker()->IsCompleted()) {
+      tx_alive->CancelScheduledTx();
+    }
```
dranikpg, line 229: I also would probably verify that this is the first hop, but in CancelScheduledTx, so it doesn't assume that the guarantees are true (I do now)

dranikpg: done

PR #7478, opened 2026-06-01
refactor: add more SIMD into OAHSet

BorysTheDev:
Summary: Refactors OAHSet to push more of the hot-path probing logic into SIMD-based scans for both the displacement window and the extension vector.

Changes:

- Moves `OAHSet` hot-path logic (`Add`, `AddMany`, `Find`, `Erase`) out of the header into a new implementation file: src/core/oah_set.cc
- Introduces SIMD helpers (ProbeLanes, lane masks) and consolidates shared lookup logic via FindMatch/FindInternal
- Adds refresh logic for lazy/stale extended-hash candidates during probing
- Updates the build to compile the new TU (src/core/CMakeLists.txt)
- Adds a stress test covering SIMD find/erase across window/vector paths and TTL expiry (SimdFindEraseStress)
- Tightens `SimdOp` APIs with more constexpr/noexcept qualifiers

Technical Notes: The new probe path uses SIMD masks to shortlist key-compare candidates (including lazy-zero hash lanes) and applies expiry during probe to keep lookups/erases consistent.

### DIFF
--- a/src/core/oah_set.cc
+++ b/src/core/oah_set.cc
@@ -1,0 +1,184 @@
+// Copyright 2024, DragonflyDB authors.  All rights reserved.
+// See LICENSE for licensing terms.
+//
+
+#include "core/oah_set.h"
+
+#include <bit>
+
+#include "base/logging.h"
+
+namespace dfly {
+
+// Several definitions below are `inline FORCE_INLINE`: the inline keyword makes
+// them COMDAT (non-interposable), which is what lets always_inline apply to an
+// out-of-line member, so they fold into their in-TU callers (notably AddImpl into
+// AddMany's bulk-insert loop, and FindMatch into both AddImpl and FindInternal).
+
+template <typename Wide>
+inline FORCE_INLINE OAHSet::LaneMasks OAHSet::ProbeLanes(const OAHEntry* base,
+                                                         uint64_t ext_hash) noexcept {
+  auto data_v = Wide::Load(reinterpret_cast<const uint64_t*>(base));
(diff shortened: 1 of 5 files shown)


vyavdoshenko (approved):
lgtm

dranikpg (approved):
Not that I understand much of the performance specifics and what takes time, is optimized, has memory latency etc in this domain, but the code "as logical code" looks fine

src/core/oah_set.cc
```diff
@@ -0,0 +1,184 @@
...
+template <typename Wide>
+inline FORCE_INLINE OAHSet::LaneMasks OAHSet::ProbeLanes(const OAHEntry* base,
+                                                         uint64_t ext_hash) noexcept {
+  auto data_v = Wide::Load(reinterpret_cast<const uint64_t*>(base));
+  auto hash_v = (data_v & Wide::Fill(OAHEntry::kExtHashShiftedMask)) >> OAHEntry::kExtHashShift;
+  // ~is_empty stops an empty lane's zero hash from aliasing a hash/lazy-zero match.
+  auto is_empty = data_v == uint64_t(0);
+  auto candidate = ((hash_v == ext_hash) | (hash_v == uint64_t(0))) & ~is_empty;
+  return {candidate.GetMSBs(), is_empty.GetMSBs()};
+}
```
dranikpg, line 27: I don't see any member access here, could've been just a function / static?

BorysTheDev: I will do it in next PR

PR #7480, opened 2026-06-01
chore(server): Small snapshot updates

dranikpg:
Cosmetic changes around code & comments


src/server/journal/streamer.cc
```diff
@@ -419,7 +419,6 @@ RestoreStreamer::RestoreStreamer(DbSlice* slice, cluster::SlotSet slots, Executi
   DCHECK(slice != nullptr);
   migration_buckets_serialization_threshold_cached =
       absl::GetFlag(FLAGS_migration_buckets_serialization_threshold);
-  db_array_ = slice->databases();  // Inc ref to make sure DB isn't deleted while we use it
```
abhijat, line 422: this is the only change that looks non trivial, but it looks safe because this is also done in `RestoreStreamer::Start` by registering the change listener

abhijat (approved):

src/server/journal/streamer.cc
dranikpg: Yes

Issue #7436, opened 2026-05-28
`dragonfly/connection_test.py::test_reply_count[df_factory1]` fails assertion


abhijat:
https://github.com/dragonflydb/dragonfly/actions/runs/26560858812/job/78243275077

```
>       assert pipe_flushes <= pipe_limit
E       assert 19 <= 12
```

seen on PR https://github.com/dragonflydb/dragonfly/pull/7416 not in CI, but does not seem related to PR. tried CI-repeat to verify and it is indeed flaky now

https://github.com/dragonflydb/dragonfly/actions/runs/26568346481/job/78268732182

```
======== 132 failed, 868 passed, 356000 deselected in 491.45s (0:08:11) ========
```

pattern of failures is very specific

```
FAILED 😰  dragonfly/connection_test.py::test_reply_count[df_factory1-3-500] - assert 14 <= 12
FAILED 😰  dragonfly/connection_test.py::test_reply_count[df_factory1-4-500] - assert 15 <= 12
FAILED 😰  dragonfly/connection_test.py::test_reply_count[df_factory1-9-500] - assert 16 <= 12
FAILED 😰  dragonfly/connection_test.py::test_reply_count[df_factory1-10-500] - assert 13 <= 12
FAILED 😰  dragonfly/connection_test.py::test_reply_count[df_factory1-15-500] - assert 16 <= 12
FAILED 😰  dragonfly/connection_test.py::test_reply_count[df_factory1-18-500] - assert 15 <= 12
FAILED 😰  dragonfly/connection_test.py::test_reply_count[df_factory1-22-500] - assert 19 <= 12
FAILED 😰  dragonfly/connection_test.py::test_reply_count[df_factory1-26-500] - assert 13 <= 12
FAILED 😰  dragonfly/connection_test.py::test_reply_count[df_factory1-31-500] - assert 14 <= 12
FAILED 😰  dragonfly/connection_test.py::test_reply_count[df_factory1-33-500] - assert 13 <= 12
FAILED 😰  dragonfly/connection_test.py::test_reply_count[df_factory1-37-500] - assert 17 <= 12
FAILED 😰  dragonfly/connection_test.py::test_reply_count[df_factory1-40-500] - assert 14 <= 12
FAILED 😰  dragonfly/connection_test.py::test_reply_count[df_factory1-41-500] - assert 16 <= 12
FAILED 😰  dragonfly/connection_test.py::test_reply_count[df_factory1-45-500] - assert 16 <= 12
[...]

PR #7484, opened 2026-06-02
server: Block for shards to finish before applying journal blob

### LINKED ISSUE #7476: failing `test_tagged_chunk_replication[df_factory0-NONE]`
https://github.com/dragonflydb/dragonfly/actions/runs/26749971387/job/78835146192

recently added test

abhijat:
In rdb load when handling journal blob RDB_OPCODE_JOURNAL_BLOB the expectation is that we flush all shards before consuming and applying journal change. 

The baseline (pending objects) must be created before applying a journal command for those objects, so that for example we do not end up with situation:

* Journal: DELETE key X, and then
* Baseline: Create key X

the above ordering resurrects the key X. 

Due to the way that the flush is issued, the above ordering is possible when the object is not on the shard running the loader.

There is a queue for these other shards which holds pending items to create in case of cross shard creation path (ReadAndDispatchObject). 

**But because the flush method does not wait for work to complete**, it only enqueues and continues, the journal can end up being applied first while the shard is still processing buffer items, the exact ordering issue that journal before baseline causes.

To fix this, a blocking counter is added similar to the FinishLoad method of the rdb loader, so that the work actually finishes before the journal blob is applied.

Note that this has performance implications, as now all shards must process their buffer items before executing every single journal command, which is the correct option.

FIXES https://github.com/dragonflydb/dragonfly/issues/7476

A new unit test is added, which fails without change and passes with it

### DIFF
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -2274,7 +2274,7 @@ error_code RdbLoader::Load(io::Source* src) {
       RETURN_ON_ERR(ConsumeInput(8));  // ignore 8 bytes
 
       if (full_sync_cut_cb) {
-        FlushAllShards();  // Flush as the handler awakes post load handlers
+        FlushAllShards();  // Drain queued loads before waking post-load handlers.
         full_sync_cut_cb();
       }
       continue;
@@ -2359,7 +2359,7 @@ error_code RdbLoader::Load(io::Source* src) {
     }
 
     if (type == RDB_OPCODE_JOURNAL_BLOB) {
-      FlushAllShards();  // Always flush before applying incremental on top
+      FlushAllShards();  // Always drain before applying incremental on top.
       RETURN_ON_ERR(HandleJournalBlob(service_));
       continue;
     }
@@ -2442,15 +2442,7 @@ error_code RdbLoader::Load(io::Source* src) {
 }
 
 void RdbLoader::FinishLoad(absl::Time start_time, size_t* keys_loaded) {
-  BlockingCounter bc(shard_set->size());
-  for (unsigned i = 0; i < shard_set->size(); ++i) {
-    // Flush the remaining items.
-    FlushShardAsync(i);
-
-    // Send sentinel callbacks to ensure that all previous messages have been processed.
-    shard_set->Add(i, [bc]() mutable { bc->Dec(); });
-  }
-  bc->Wait();  // wait for sentinels to report.
+  FlushAllShards();
   // Decrement local one if it exists
   if (EngineShard* es = EngineShard::tlocal(); es) {
     GetCurrentDbSlice().DecrLoadInProgress();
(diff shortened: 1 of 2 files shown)


abhijat:
augment review

abhijat:
~~test failure seems PR related, can be seen locally reliably~~

```
FAILED 😰  tests/dragonfly/connection_test.py::test_reply_count[0-df_factory1] - assert 17 <= 12
FAILED 😰  tests/dragonfly/connection_test.py::test_reply_count[1-df_factory1] - assert 14 <= 12
FAILED 😰  tests/dragonfly/connection_test.py::test_reply_count[2-df_factory1] - assert 16 <= 12
FAILED 😰  tests/dragonfly/connection_test.py::test_reply_count[3-df_factory1] - assert 22 <= 12
FAILED 😰  tests/dragonfly/connection_test.py::test_reply_count[4-df_factory1] - assert 14 <= 12
FAILED 😰  tests/dragonfly/connection_test.py::test_reply_count[5-df_factory1] - assert 16 <= 12
FAILED 😰  tests/dragonfly/connection_test.py::test_reply_count[6-df_factory1] - assert 20 <= 12
FAILED 😰  tests/dragonfly/connection_test.py::test_reply_count[7-df_factory1] - assert 21 <= 12
FAILED 😰  tests/dragonfly/connection_test.py::test_reply_count[8-df_factory1] - assert 14 <= 12
```

actually this is related to https://github.com/dragonflydb/dragonfly/issues/7436 seems flaky

src/server/rdb_load.cc
```diff
@@ -2826,8 +2818,14 @@ void RdbLoader::FlushShardAsync(ShardId sid) {
...
 void RdbLoader::FlushAllShards() {
-  for (ShardId i = 0; i < shard_set->size(); i++)
+  BlockingCounter bc(shard_set->size());
+  for (ShardId i = 0; i < shard_set->size(); i++) {
     FlushShardAsync(i);
+
+    // Send sentinel callbacks to ensure that all previous messages have been processed.
+    shard_set->Add(i, [bc]() mutable { bc->Dec(); });
+  }
+  bc->Wait();  // wait for sentinels to report.
```
dranikpg, line 2828: Hmm... asking myself how did it work before with the inlining optimizations that we have? 🤔

dranikpg (approved):

src/server/rdb_test.cc
```diff
@@ -1551,6 +1560,84 @@ TEST_F(RdbTest, InterleavedLoad) {
   EXPECT_EQ(Run({"SELECT", "0"}), "OK");
 }
 
+// https://github.com/dragonflydb/dragonfly/issues/7476
+// unlike the linked issue we test full sync cut vs flush, because it is easier to test with
+// callback which full sync cut works with rather than deletes. in the test we block the target
+// shard queue, and that may end up blocking the DEL command too
+TEST_F(RdbTest, FullSyncCutWaitsForTaggedShardLoads) {
```
romange, line 1567: @abhijat  so without the fix this test fails?

src/server/rdb_test.cc
```diff
@@ -1551,6 +1560,84 @@ TEST_F(RdbTest, InterleavedLoad) {
...
+    if (Shard(key, shard_set->size()) != 0)
+      break;
+  }
+  ASSERT_NE(Shard(key, shard_set->size()), 0u);
+  const ShardId sid = Shard(key, shard_set->size());
+
+  std::string entry;
+  entry.push_back(RDB_TYPE_STRING);
+  AppendString(&entry, key);
+  AppendString(&entry, "baseline");
```
romange, line 1583: It seems you try to recreate a very specific order but it's hard to understand from the code what the order is. Can you please add comments explaining that is being orchestrated?

abhijat: yes the test is a little awkward, as I couldn't use delete command, I will add a comment

src/server/rdb_test.cc
abhijat: yes if I remove the blocking behavior in the flush code then the test fails

abhijat:
closing as there is better solution already merged (disable batch)

PR #7486, opened 2026-06-02
fix(server): Propagate error on bigvalue cancel

### LINKED ISSUE #7187: test_big_huge_streaming_restart fails

dranikpg:
Better way to fix #7187 by actually returning an error from the FlushCallback to then propagate this error further in the serializer to abort big value streaming on an error

### DIFF
--- a/src/server/snapshot.cc
+++ b/src/server/snapshot.cc
@@ -318,14 +318,16 @@ void SliceSnapshot::HandleFlushData(std::string data) {
   VLOG(2) << "Pushed with Serialize() " << serialized;
 }
 
-void SliceSnapshot::ConsumeBigValueChunk(std::string data) {
-  if (!cntx_->IsRunning()) {
-    ThisFiber::Yield();
-    return;  // Short circuit flush on cancel or error to finish faster
-  }
+std::error_code SliceSnapshot::ConsumeBigValueChunk(std::string data) {
+  if (cntx_->IsError())
+    return cntx_->GetError();
+
+  if (cntx_->IsCancelled())
+    return std::make_error_code(std::errc::operation_canceled);
 
   HandleFlushData(std::move(data));
   ++ServerState::tlocal()->stats.big_value_preemptions;
+  return {};
 }
 
 size_t SliceSnapshot::FlushSerialized() {
--- a/src/server/snapshot.cc
+++ b/src/server/snapshot.cc
@@ -247,8 +247,7 @@ unsigned SliceSnapshot::SerializeBucketLocked(DbIndex db_index, PrimeTable::buck
 
 void SliceSnapshot::SerializeFetchedEntry(const TieredDelayedEntry& tde, const PrimeValue& pv) {
   std::lock_guard lk{stream_mu_};
-  auto res = serializer_->SaveEntry(tde.key, pv, tde.expire, tde.mc_flags, tde.dbid);
-  CHECK(res);
+  serializer_->SaveEntry(tde.key, pv, tde.expire, tde.mc_flags, tde.dbid);
 }
 
 void SliceSnapshot::SerializeEntry(BucketIdentity bucket, DbIndex db_indx, const PrimeKey& pk,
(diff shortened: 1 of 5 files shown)


src/server/snapshot.cc
```diff
@@ -318,14 +318,16 @@ void SliceSnapshot::HandleFlushData(std::string data) {
...
-  if (!cntx_->IsRunning()) {
-    ThisFiber::Yield();
-    return;  // Short circuit flush on cancel or error to finish faster
-  }
+std::error_code SliceSnapshot::ConsumeBigValueChunk(std::string data) {
+  if (cntx_->IsError())
+    return cntx_->GetError();
+
+  if (cntx_->IsCancelled())
+    return std::make_error_code(std::errc::operation_canceled);
```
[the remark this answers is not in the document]
dranikpg: True

src/server/snapshot.cc
```diff
@@ -247,8 +247,7 @@ unsigned SliceSnapshot::SerializeBucketLocked(DbIndex db_index, PrimeTable::buck
 
 void SliceSnapshot::SerializeFetchedEntry(const TieredDelayedEntry& tde, const PrimeValue& pv) {
   std::lock_guard lk{stream_mu_};
-  auto res = serializer_->SaveEntry(tde.key, pv, tde.expire, tde.mc_flags, tde.dbid);
-  CHECK(res);
+  serializer_->SaveEntry(tde.key, pv, tde.expire, tde.mc_flags, tde.dbid);
```
romange: why `res was removed?

dranikpg: returned error print, but we can't check fail on it now as it can return an error if the replication was cancelled while a value was written (pushed to consumer directly from serializer_)

romange:
@abhijat  please review it today.

romange:
LGTM from me

src/server/snapshot.cc
```diff
@@ -248,7 +248,7 @@ unsigned SliceSnapshot::SerializeBucketLocked(DbIndex db_index, PrimeTable::buck
 void SliceSnapshot::SerializeFetchedEntry(const TieredDelayedEntry& tde, const PrimeValue& pv) {
   std::lock_guard lk{stream_mu_};
   auto res = serializer_->SaveEntry(tde.key, pv, tde.expire, tde.mc_flags, tde.dbid);
-  CHECK(res);
+  LOG_IF(ERROR, !res.has_value()) << "Serialization error: " << res.error();
```
abhijat, line 251: will this fire at error level for things like cancelled operation? it might cause more alerts in that case for routine shutdown/disconnect et

src/server/snapshot.cc
```diff
@@ -265,9 +265,10 @@ void SliceSnapshot::SerializeEntry(BucketIdentity bucket, DbIndex db_indx, const
     ++type_freq_map_[RDB_TYPE_STRING];
   } else {
     std::lock_guard lk{stream_mu_};
-    io::Result<uint8_t> res = serializer_->SaveEntry(pk, pv, expire_time, mc_flags, db_indx);
-    CHECK(res);
-    ++type_freq_map_[*res];
+    if (auto res = serializer_->SaveEntry(pk, pv, expire_time, mc_flags, db_indx); res)
+      ++type_freq_map_[*res];
+    else
+      LOG(ERROR) << "Serialization error: " << res.error();
```
abhijat, line 271: same question as https://github.com/dragonflydb/dragonfly/pull/7486/changes#r3354215998

abhijat: also there is no `CHECK(res);` now, so we log error and continue, is that safe?

abhijat: I see the second q is answered in https://github.com/dragonflydb/dragonfly/pull/7486#discussion_r3340471807

abhijat (approved):
changes lgtm over all, may be good to calibrate log levels based on error type in future pr if possible

src/server/snapshot.cc
dranikpg: I thought about this and I don't want it to go unnoticed for now - maybe I could've done `res.error() != cancelled` for logging

PR #7490, opened 2026-06-02
chore: drop extra drakarys logs

BorysTheDev:
Summary: Lowers Drakarys-related logging from INFO to VLOG(1) to reduce default log noise.
Changes: Downgrades per-shard callback/decommit progress logs and simplifies the detached-mode completion message.

### DIFF
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -1015,15 +1015,15 @@ util::fb2::Fiber DbSlice::FlushDbIndexes(const std::vector<DbIndex>& indexes) {
 
   ShardId shard_id = owner_->shard_id();
   auto cb = [flush_db_arr = std::move(flush_db_arr), shard_id]() mutable {
-    LOG(INFO) << "Drakarys shard " << shard_id << " cb entered (pre-destructors)"
-              << " rss="
-              << strings::HumanReadableNumBytes(rss_mem_current.load(std::memory_order_relaxed));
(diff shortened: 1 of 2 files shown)


src/server/db_slice.cc
```diff
@@ -1015,15 +1015,15 @@ util::fb2::Fiber DbSlice::FlushDbIndexes(const std::vector<DbIndex>& indexes) {
 
   ShardId shard_id = owner_->shard_id();
   auto cb = [flush_db_arr = std::move(flush_db_arr), shard_id]() mutable {
-    LOG(INFO) << "Drakarys shard " << shard_id << " cb entered (pre-destructors)"
-              << " rss="
-              << strings::HumanReadableNumBytes(rss_mem_current.load(std::memory_order_relaxed));
+    VLOG(1) << "Drakarys shard " << shard_id << " cb entered (pre-destructors)"
```
kostasrim, line 1018: Could you plz double check that we don't run this on prod instances ? I don't remember if we use verbose logging or not so it's worth checking

kostasrim (approved):
LGTM

src/server/db_slice.cc
BorysTheDev: only if we turned them on

kostasrim: and we do have them turned on 😄

BorysTheDev: https://github.com/dragonflydb/dragonfly/pull/7497

PR #7491, opened 2026-06-02
fix(replication): reset partial sync index if dfly flow is received

kostasrim:
https://github.com/dragonflydb/dragonfly/pull/5580 allows skipping full sync entirely.

On rotating master scenarios where the rotating replica has a smaller version (older && < 1.34) than master dragonfly still sends DFLY SYNC command which forces the replication state machine to go over full sync. This is a bug because now START STABLE won't joint the snapshot fibers. When the replica disconnects, it crashes trigerring a fiber not joined check failure


tests/dragonfly/replication_test.py
```diff
@@ -4918,3 +4920,38 @@ async def link_gone():
...
+
+    # Both new_master and old_replica replicate from old_master and reach stable sync.
+    await start_replication(c_new_master, old_master.port)
+    await start_replication(c_old_replica, old_master.port)
+
+    await c_new_master.execute_command("REPLTAKEOVER 5")
+
+    await c_old_replica.execute_command(f"REPLICAOF localhost {new_master.port}")
+
+    await asyncio.sleep(1.0)
```
romange: Please never use asyncio.sleep to wait for transitions. This never works.

tests/dragonfly/replication_test.py
```diff
@@ -4916,3 +4914,46 @@ async def link_gone():
...
+    await start_replication(c_old_replica, old_master.port)
+
+    await c_new_master.execute_command("REPLTAKEOVER 5")
+
+    await c_old_replica.execute_command(f"REPLICAOF localhost {new_master.port}")
+
+    @assert_eventually(timeout=10)
+    async def replica_connected():
+        info = await c_new_master.info("replication")
+        assert info.get("connected_slaves", 0) == 1
```
kostasrim, line 4943: @romange The two protocols are incompatible.

1.38 -> Skips full sync if partial is requested 
1.34 -> Enters full sync even if partial is requested

We can't bridge them neither do we have a version (repl_version) for 1.34 (we added VERSION_5 on 1.36). 

There are two issues:

1. Version 1.38 would crash because we would start full sync (DFLY SYNC was sent by 1.34). Now we skip it and we return an error. `1.38` is now `safe`
2. 1.34 is trying to replicate from 1.38 but it really can't. There are two cases here: either the operator must cancel the replica, or wait for partial sync to expire, e.g. more data to be sent such that the LSN's go out of range (and both flows will agree to do a full sync flow).

This is the best compromise I could find that doesn't crash new versions or deadlocks 1.34.

romange (approved):

PR #7496, opened 2026-06-03
feat(server): Omittable DEL operation

dranikpg:
Enable omitting DEL calls by using journal omits + disabling autojounral to build the command itself

### DIFF
--- a/src/server/generic_family.cc
+++ b/src/server/generic_family.cc
@@ -1170,8 +1170,12 @@ OpResult<uint32_t> OpDelV2(const OpArgs& op_args, const ShardArgs& keys, bool as
       journal_args.push_back(key);
   }
 
-  if (journal_enabled && !journal_args.empty())
-    RecordJournal(op_args, "DEL", journal_args);
+  if (journal_enabled) {
+    if (!journal_args.empty())
+      RecordJournal(op_args, "DEL", journal_args);
+    else if (deleted_cnt > 0)  // operations that are not in lsn log
+      journal::ClearBuffer();
+  }
 
   return deleted_cnt;
 }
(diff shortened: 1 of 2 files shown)


src/server/generic_family.cc
```diff
@@ -2691,7 +2725,7 @@ void GenericFamily::Register(CommandRegistry* registry) {
   constexpr auto kSelectOpts = CO::LOADING | CO::FAST;
   registry->StartFamily();
   *registry
-      << CI{"DEL", CO::JOURNALED, -2, 1, -1, acl::kDel}.SetAsyncHandler(CmdDel)
+      << CI{"DEL", CO::JOURNALED | CO::NO_AUTOJOURNAL, -2, 1, -1, acl::kDel}.SetAsyncHandler(CmdDel)
```
dranikpg, line 2732: with all the omit logic we will sadly have more custom journaling code unless we invent something

dranikpg:
augment review

abhijat:
I got a bot comment when reviewing this locally and it looks like a valid risk to me, basically if journal write is skipped during `OpDelV2` then `ClearBuffer` should be called to prevent invalid partial sync, for example:

* DEL k is issued and omitted due to meeting all requirements of omitable command
* more commands are added to journal
* a replica reconnects with LSN before that DEL command, its partial sync req is accepted
* it replays the commands and finishes partial sync
* ends up missing the deletion of k because it wasn't in the journal

So perhaps like in `SetCmd::PostEdit` in string family the journal should be cleared.

dranikpg:
Oh, wow, you're right - I forgot about this

romange (approved):

PR #7499, opened 2026-06-03
chore: reproduce inline schedule during save

kostasrim:
* add coverage test for inline schedule during save
* add exception to DFATAL for eval fibers

### DIFF
--- a/src/server/serializer_base.cc
+++ b/src/server/serializer_base.cc
@@ -207,7 +207,10 @@ void SerializerBase::OnChangeBlocking(DbIndex db_index, PrimeTable::bucket_itera
   if (!absl::StartsWith(active_name, "shard_queue") &&  //
       !absl::StartsWith(active_name, "l2_queue") &&     // pipelining
       !absl::StartsWith(active_name, "SliceSnapshot") &&
-      !absl::StartsWith(active_name, "DflyConn_") &&
+      !absl::StartsWith(active_name, "DflyConn_") &&  // TOCTOU: connection checks
+                                                      // has_registered_callbacks before save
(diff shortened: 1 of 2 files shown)


src/server/serializer_base.cc
```diff
@@ -207,6 +207,7 @@ void SerializerBase::OnChangeBlocking(DbIndex db_index, PrimeTable::bucket_itera
   if (!absl::StartsWith(active_name, "shard_queue") &&  //
       !absl::StartsWith(active_name, "l2_queue") &&     // pipelining
       !absl::StartsWith(active_name, "SliceSnapshot") &&
+      !absl::StartsWith(active_name, "DflyConn_") &&
```
romange: please add an explanation how it happens exactly like below.

src/server/serializer_base.cc
```diff
@@ -207,6 +207,10 @@ void SerializerBase::OnChangeBlocking(DbIndex db_index, PrimeTable::bucket_itera
   if (!absl::StartsWith(active_name, "shard_queue") &&  //
       !absl::StartsWith(active_name, "l2_queue") &&     // pipelining
       !absl::StartsWith(active_name, "SliceSnapshot") &&
+      !absl::StartsWith(active_name, "DflyConn_") &&  // TOCTOU: connection checks
```
romange, line 210: wow so it's a race.

romange (approved):

src/server/serializer_base.cc
kostasrim: yes and the window is pretty tight!

PR #7500, opened 2026-06-03
feat(facade): add V2 idle-flush coalescing delay

glevkovich:
When IoLoopV2 finishes a pipeline and reaches the idle-await
block with no queued commands, waiting briefly (interruptible)
before flushing gives the next TCP segment time to arrive. If any
wake condition fires during the wait (new data, migration, admin
message), the fiber skips the flush and processes the new commands
with the previous replies still buffered — merging two TCP bursts
into one sendmsg call.

Two changes work together:
- ExecuteBatch keeps batch mode on the last command when
  pipeline_wait_batch_usec > 0 and the input buffer is empty,
  so the reply is not auto-flushed by FinishScope before the
  idle-await wait.
- IoLoopV2's idle-await path calls io_event_.await_until() for up
  to pipeline_wait_batch_usec microseconds. If should_wake() fires
  during the wait, the flush is skipped entirely.

Benchmarks (50 clients, 2KB SET, cross-machine, V2,
--pipeline_wait_batch_usec=50):
  pipeline=10: density 6.1→6.0, RPS unchanged (neutral)
  pipeline=100: density 53→75 (+42%), syscalls -29%, RPS +9%

ZADD overhead is negligible: the wait fires ~104 times per
connection over a 75s run (5ms total), no measurable impact
because ZADD throughput is transaction-bound, not flush-bound.

The wait is interruptible: migration requests, admin messages, or
new socket data all fire io_event_.notify() and wake the fiber
early, avoiding the blind-sleep latency penalty of SleepFor.

Controlled by --pipeline_wait_batch_usec (default 0, off).

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -2603,16 +2603,19 @@ bool Connection::ExecuteBatch() {
       continue;
     }
 
-    // We must continue with async execution if we already have executing commands
-    auto mode = is_head ? AsyncPreference::PREFER_ASYNC : AsyncPreference::ONLY_ASYNC;
-
-    if (!ioloop_v2_)  // only v2 loop supports any async commands so far
-      mode = AsyncPreference::ONLY_SYNC;
-
-    // V2: Batch the head command's reply when more commands are queued behind it.
-    // This prevents sync-only commands from triggering immediate flushes, keeping
-    // sendmsg syscalls to a minimum. IoLoopV2's idle-await block handles the final flush.
-    reply_builder_->SetBatchMode(ioloop_v2_ && is_head && (cmd->next != nullptr));
+    // V1 is always sync; V2 supports async dispatch with per-command mode selection.
+    auto mode = AsyncPreference::ONLY_SYNC;
+    if (ioloop_v2_) {
+      mode = is_head ? AsyncPreference::PREFER_ASYNC : AsyncPreference::ONLY_ASYNC;
+      // V2: Batch the head command's reply when more commands are queued behind it.
+      // This prevents sync-only commands from triggering immediate flushes.
+      // Also, keep batch mode for the last command if coalescing across TCP
+      // boundaries - replies stay in the buffer during the idle-await sleep so they can
+      // be merged with the next pipeline burst.
+      bool is_last = (cmd->next == nullptr);
+      bool coalesce_last = is_last && (pipeline_wait_batch_usec > 0) && (io_buf_.InputLen() == 0);
+      reply_builder_->SetBatchMode(is_head && (!is_last || coalesce_last));
+    }
 
     auto dispatch_res = (service_->*dispatch)(cmd, mode);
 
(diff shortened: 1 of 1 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -3047,6 +3050,18 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
...
 
+        // Coalescing delay - sleep briefly before flushing to let the next TCP
+        // segment arrive. During the sleep the proactor can fire RegisterOnRecv which
+        // reads and parses data (phase_ == READ_SOCKET enables the callback). If new
+        // commands land, skip the flush entirely and loop back to process them.
+        if (pipeline_wait_batch_usec > 0 && parsed_cmd_q_len_ == 0) {
+          ThisFiber::SleepFor(chrono::microseconds(pipeline_wait_batch_usec));
+          if (should_wake()) {
+            phase_ = PROCESS;
+            continue;
```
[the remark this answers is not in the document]
glevkovich: fixed

src/facade/dragonfly_connection.cc
```diff
@@ -2603,16 +2603,19 @@ bool Connection::ExecuteBatch() {
...
-    auto mode = is_head ? AsyncPreference::PREFER_ASYNC : AsyncPreference::ONLY_ASYNC;
-
-    if (!ioloop_v2_)  // only v2 loop supports any async commands so far
-      mode = AsyncPreference::ONLY_SYNC;
-
-    // V2: Batch the head command's reply when more commands are queued behind it.
-    // This prevents sync-only commands from triggering immediate flushes, keeping
-    // sendmsg syscalls to a minimum. IoLoopV2's idle-await block handles the final flush.
-    reply_builder_->SetBatchMode(ioloop_v2_ && is_head && (cmd->next != nullptr));
+    // V1 is always sync; V2 supports async dispatch with per-command mode selection.
```
romange, line 2606: there is no change here, right?

glevkovich: There is: till now last message (cmd->next is null) caused batch mode  to set to false. 
Now if coalesce_last is true, we will not flush yet (since batch mode set to true), hoping to get more data on next iteration. 
The flush is implicit/hidden, part of FinishScope.

glevkovich:
comment: `pipeline_wait_batch_usec` defaults to 0, so V2 gets no coalescing benefit out of the box. Benchmarks show negligible regression for p=1 and large gains for p≥10. Considering setting the V2 default to 50µs as part of this PR.

Will update soon or just push a fix.

src/facade/dragonfly_connection.cc
```diff
@@ -3047,6 +3056,30 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
       if (!should_wake()) {
         phase_ = READ_SOCKET;
 
+        // Coalescing delay - wait briefly before flushing to let the next TCP segment arrive.
+        // Uses an interruptible timed wait: if the recv callback (RegisterOnRecv) buffers new
+        // bytes, or any other wake condition triggers io_event_.notify() during the window,
+        // await_until returns immediately and we skip the flush to loop back and process.
+        // Parsing itself happens later in this loop (ReadPendingInput + ParseLoop), not in
+        // the recv callback.
+        if (pipeline_wait_batch_usec > 0 && parsed_cmd_q_len_ == 0) {
```
romange: I understand the fix but you can see - it's not trivial and complicates the flow.

1. can we just continue to line 3083 in case of `is_ready_to_migrate()`?
2. In V1 in AsyncFiber we Yield giving the opportunity to this fiber to read from the socket. Did you try Yielding here instead of await_until relying on helio completion handler to update `should_wake()` ?
3. Do you see difference with a single connection vs multiple? 
4. I did not understand `ZADD overhead ...` part, where ZADD comes to place?

src/facade/dragonfly_connection.cc
romange: `The flush is implicit/hidden, part of FinishScope.` So you are relying on that all commands have `FinishScope` ?

glevkovich: The commands are not the ones who has FinishScope, it's the reply builder (RAII in some of it's send functions from what I can recall).
I'm not relying in FinishScope, the opposite - I'm trying to suppress it not to flush eagrly, to reduce the number of flushes and increase the batch density. If a command is not flushed in the FinishScope part, it will eventually get flushed in the default flush in the ioloop_v2() at the top:
```
 reply_builder_->Flush();
        if (auto err = reply_builder_->GetError(); err) {
          return err;
        }

        io_event_.await(should_wake);
```

FinishScope flushing is an eager latency optimisation, this change deffer the flush intentionally.
In most workloads this proved to increase throughput (+10%) and reduce number of systems calls (-30%, kernel is less busy).
Comment: FinishScope will flush anyway of the accumulated buffer in reply_builder_ is too large.

glevkovich:
Closing this PR -  benchmarks shows significant regressions, especially on on synchronous commands.I didn't make accurate benchmarks before the PR.

PR #7502, opened 2026-06-03
fix: deep-validate listpack/intset payloads on RESTORE

vyavdoshenko:
A crafted RESTORE payload with a valid listpack header but a bogus interior entry length (e.g. a 32-bit string declaring 0x7fffffff bytes) passed the shallow integrity check and drove an out-of-bounds access while the collection was iterated, crashing the server. Since RESTORE is an unauthenticated keyspace command by default, a single ~24-byte payload could take down the whole process.

Root cause:
Loaders for listpack-encoded collections called `lpValidateIntegrity(..., deep=0, ...)`, which only checks the header (declared total bytes + trailing terminator) and never validates interior entry lengths. Iteration then trusted the attacker-controlled length. The only per-entry guard was an `assert` that is compiled out in release builds.

Changes:
- Add `LoadConfig::deep_integrity` and set it on the RESTORE path only.
- Thread the flag into the SET / HASH / ZSET / LIST listpack loaders; validate every entry (`deep=1`) for untrusted input.
- Extend stream listpack validation (`StreamValidateListpackIntegrity`) to fully validate entries, since `CreateStream` iterates the whole listpack via `streamGetEdgeID` at load time.
- Route intset loading through the same flag for consistency.
- Snapshot and replication loads are unchanged (trusted source, no extra pass).

Behavior:
Crafted payloads now return `ERR Bad data format` instead of crashing.

Tests:
- Add 5 regression tests in `generic_family_test` (SET, HASH, ZSET, LIST, STREAM), each of which crashed before the fix and passes after it.
- `generic_family_test`, `rdb_test`, `stream_family_test` all pass; valid RESTORE/reload of listpack-encoded data is unaffected.


src/server/collection_family_fallback.cc
```diff
@@ -40,12 +40,12 @@ uint32_t SetFamily::MaxIntsetEntries() {
   return 0;
 }
 
-LoadBlobResult SetFamily::LoadLPSetBlob(std::string_view blob, PrimeValue* pv) {
+LoadBlobResult SetFamily::LoadLPSetBlob(std::string_view blob, PrimeValue* pv, bool deep) {
```
romange: can you move `deep` to after `blob`

vyavdoshenko: done

romange (approved):

Issue #7414, opened 2026-05-26
INFO refactoring


BorysTheDev:
We need to improve INFO latency and I see we can try next approaches together:

- lock free
- reduce number of hops

First of all we need to check how many hops we do during INFO and if we have several we can try to refactoring to do Dispatch() step in INFO where we get data from all system per thread, and combine results after Dispatch().

PR #7503, opened 2026-06-03
refactor: drop extra dispatch for info replica section

BorysTheDev:
part of: #7414
Summary: Refactors master-side replication memory metrics collection to avoid an extra shard fan-out dispatch.

Changes:

- Changes DflyCmd::GetReplicationMemoryStats to return per-shard stats (called on the shard thread) instead of internally running RunBriefInParallel.
- Folds replication-memory collection into ServerFamily::GetMetrics fan-out and stores totals in Metrics::replication_stats.
- Updates Prometheus metrics and INFO MEMORY output to use the pre-collected replication_stats.
- Adds an opt-out flag (collect_replication_memory) to skip replication-memory collection when not needed.
- Adds unit tests to ensure replication memory appears only in INFO MEMORY/ALL (and defaults to 0 with no replicas).

Technical Notes: The new replication-memory path is lock-free w.r.t. replica-info snapshots and assumes a non-yielding, shard-thread execution context during metrics fan-out.

### DIFF
--- a/src/server/dflycmd.h
+++ b/src/server/dflycmd.h
@@ -252,8 +252,7 @@ class DflyCmd {
 
   // Must be called on the given shard's thread (e.g. from the GetMetrics
   // fan-out). Lock-free: reads thread-local replica infos.
-  ReplicationMemoryStats GetReplicationMemoryStats(EngineShard* shard) const
-      ABSL_NO_THREAD_SAFETY_ANALYSIS;
+  static ReplicationMemoryStats GetReplicationMemoryStats(EngineShard* shard);
 
   // Sets metadata.
   void SetDflyClientVersion(ConnectionState* state, DflyVersion version);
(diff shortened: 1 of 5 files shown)


src/server/dflycmd.h
```diff
@@ -249,7 +250,10 @@ class DflyCmd {
   // Master-side command. Provides Replica info.
   std::vector<ReplicaRoleInfo> GetReplicasRoleInfo() const ABSL_LOCKS_EXCLUDED(mu_);
 
-  void GetReplicationMemoryStats(ReplicationMemoryStats* out) const ABSL_NO_THREAD_SAFETY_ANALYSIS;
+  // Must be called on the given shard's thread (e.g. from the GetMetrics
+  // fan-out). Lock-free: reads thread-local replica infos.
+  ReplicationMemoryStats GetReplicationMemoryStats(EngineShard* shard) const
```
romange: does it use `this` ? can it be a static function?

romange (commented):
Nice!

romange (approved):

PR #3819, opened 2024-09-28
chore: schedule chains

romange:
Use intrusive queue that allows batching of scheduling calls instead of handling each call separately. This optimizations improves latency and throughput by 3-5%

PR #3794, opened 2024-09-25
chore: ScheduleInternal refactoring

romange:
A small refactoring to improve the flow of ScheduleInternal() as well as to prepare it for the next change that will reduce the CPU load from the shard queue.

PR #7507, opened 2026-06-04
fix(server): Disable single shard batch optimization

### LINKED ISSUE #7476: failing `test_tagged_chunk_replication[df_factory0-NONE]`

abhijat:
It was recently seen that there is a correctness issue with batch scheduling for single shard tx, which is seen in the unit test added here as below as well as in https://github.com/dragonflydb/dragonfly/issues/7476

test sends two commands to rdb loader:

- create key
- delete same key

at the end the key is still present and the test fails. the following preconditions are necessary

- the key is created on shard != 0 ie where the loader is not running
- before creating the key, a priming key is created which runs schedule batch in shard, that priming key pulls in the new delete command into its own batch without going through shard_set->Add
- the SET command for creating the key has to go via the shard set task queue
- end result is out of order

The root cause is because we have the following sequence

expected shard task queue order FIFO

A = create priming key
B = create actual key  (wrapped in CreateObjectOnShard etc)
C = DEL key

```
tx A -> task B -> tx C
```

what happens is the below

```
tx A -> task B
[tx C]
```

tx C gets latched on to the batch triggered by tx A as it skips shard set add altogether. whereas task B goes on to task set add via rdb load without any such short cuts

The batching is removed altogether to remove this correctness bug, the schedule queue and associated code and metrics are also removed.

FIXES https://github.com/dragonflydb/dragonfly/issues/7476

several past PRs are involved here, adding them as cross reference

https://github.com/dragonflydb/dragonfly/pull/3794
https://github.com/dragonflydb/dragonfly/pull/3819
https://github.com/dragonflydb/dragonfly/pull/4647
https://github.com/dragonflydb/dragonfly/pull/4925

### DIFF
--- a/src/server/transaction.cc
+++ b/src/server/transaction.cc
@@ -7,7 +7,6 @@
 #include <absl/strings/match.h>
 
 #include <memory>
-#include <new>
 
 #include "base/flags.h"
 #include "base/logging.h"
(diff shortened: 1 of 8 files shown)


romange:
@dranikpg  please review (seems fine to me)

src/server/transaction.cc
```diff
@@ -773,24 +742,17 @@ void Transaction::ScheduleInternal() {
 
     ScheduleContext schedule_ctx{this, optimistic_exec};
```
dranikpg, line 743: ScheduleContext was introduced with schedule_queues to group arguments, technically we don't need it anymore, but it doesn't hurt as well

dranikpg (approved):

src/server/transaction.cc
abhijat: removed here https://github.com/dragonflydb/dragonfly/pull/7516

PR #7510, opened 2026-06-04
feat: improve HSETEX compatibility with Redis

### LINKED ISSUE #5999: Fix incompatibility for HSETEX command
I think we can support our format and Redis simultaneously

BorysTheDev:
fixes: #5999

Summary: This PR improves HSETEX compatibility with Redis while preserving Dragonfly’s existing syntax.

Changes:

Extends OpSetParams with explicit set modes (Normal/NX/FNX/FXX) and a reply Format selector (Dragonfly vs Redis).

- Introduces SetReply() so OpSet can return either “created fields” (Dragonfly) or 1 (Redis-style “applied”).
- Reworks HSETEX parsing to support both syntaxes: Dragonfly’s ttl_sec form and Redis’s FIELDS numfields form with EX/PX/EXAT/PXAT/KEEPTTL.
- Adds collective conditional semantics (FNX/FXX) and a helper (CheckHSetExCondition) to enforce all-or-nothing application.
- Updates HSETNX to use the new mode-based API.
- Expands unit tests to cover Redis-format parsing, expiry options, condition behavior, and error cases.

Technical Notes: Redis-format HSETEX returns 1 when applied and 0 when an FNX/FXX condition fails; Dragonfly format keeps returning the number of newly created fields. 

Doesn't support tiering

### DIFF
--- a/src/server/hset_family.cc
+++ b/src/server/hset_family.cc
@@ -492,12 +492,30 @@ OpResult<vector<OptStr>> OpHMGet(const HMapWrap& hw, CmdArgList fields) {
 }
 
 struct OpSetParams {
-  bool skip_if_exists = false;
+  enum class Mode : uint8_t {
+    kNormal,  // overwrite every field
+    kNX,      // set each field only if it does not already exist
+    kFNX,     // set all fields only if none of them exist
+    kFXX,     // set all fields only if all of them exist
+  };
+
+  // Selects how OpSet reports its result: Dragonfly returns the number of created fields, Redis
+  // returns 1 when the fields were applied. (Tiering support is gated by ttl, not by format.)
+  enum class Format : uint8_t { kDragonfly, kRedis };
+
   uint32_t ttl = UINT32_MAX;
   bool keepttl = false;
+  Mode mode = Mode::kNormal;
+  Format format = Format::kDragonfly;
 
   optional<util::fb2::Future<bool>>* backpressure = nullptr;
 };
+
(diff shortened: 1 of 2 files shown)


src/server/hset_family.cc
```diff
@@ -655,61 +677,140 @@ OpResult<vector<long>> OpHExpire(const OpArgs& op_args, string_view key, uint32_
...
-  constexpr uint32_t kMaxTtl = (1UL << 26);
-  if (op_sp.ttl == 0 || op_sp.ttl > kMaxTtl) {
-    return cmd_cntx->SendError(kInvalidIntErr);
-  }
+// Parses HSETEX arguments after the key, reporting any error into `parser` (surfaced by the caller
+// via RETURN_ON_PARSE_ERROR). `cmd_name` is only used to format error messages.
+//
+// Dragonfly format: HSETEX key [NX | FNX | FXX] [KEEPTTL] ttl_sec field value [field value ...]
+// Redis format:     HSETEX key [FNX | FXX] [EX sec | PX ms | EXAT ts-sec | PXAT ts-ms | KEEPTTL]
+//                          FIELDS numfields field value [field value ...]
```
dranikpg, line 722: I see, so the formats are fully exclusive and you can distinguish them

dranikpg (approved):

src/server/hset_family.cc
BorysTheDev: absolutely

PR #7513, opened 2026-06-04
fix: bound bitmap bit offsets and deny BITFIELD/BITOP under memory pressure

### LINKED ISSUE #7512: sigabrt OOM in bit operations
[crash-000000.tar.gz](https://github.com/user-attachments/files/28606518/crash-000000.tar.gz)

[crash-000001.tar.gz](https://github.com/user-attachments/files/28606520/crash-000001.tar.gz)

[crash-000002.tar.gz](https://github.com/user-attachments/files/28606522/crash-000002.tar.gz)

```
mimalloc: error: thread 0xFFFFF7317A20: out of memory in 'new'*** SIGABRT received at time=1780591580 on cpu 0 ***
PC: @     0xfffff7537608  (unknown)  (unknown)
    @     0xaaaaacee8570         64  absl::lts_20250512::WriteFailureInfo()
    @     0xaaaaacee8888         96  absl::lts_20250512::AbslFailureSignalHandler()
    @     0xfffff7ffa7e8       4752  (unknown)
    @     0xfffff74ecb3c         64  raise
    @     0xfffff74d7e00        192  abort
    @     0xaaaaaccfe3fc         48  mi_heap_try_new
    @     0xaaaaabbcfd64         64  operator new()
    @     0xaaaaabbe7f98         48  std::__new_allocator<>::allocate()
    @     0xaaaaabbe4ae4        128  std::__cxx11::basic_string<>::_S_allocate()
    @     0xaaaaabbe49d8         16  std::__cxx11::basic_string<>::_M_create()
    @     0xaaaaabbea968        160  std::__cxx11::basic_string<>::_M_construct<>()
    @     0xaaaaabbe6028         96  std::__cxx11::basic_string<>::basic_string()
    @     0xaaaaac33b824         48  dfly::(anonymous namespace)::BitOpString<>()
    @     0xaaaaac3354c0        320  dfly::(anonymous namespace)::RunBitOperationOnValues()::{lambda()#1}::operator()()
    @     0xaaaaac335810        128  dfly::(anonymous namespace)::RunBitOperationOnValues()
    @     0xaaaaac335970        160  dfly::(anonymous namespace)::CombineResultOp()
    @     0xaaaaac339950        320  dfly::(anonymous namespace)::BitOp()
    @     0xaaaaac012220         32  fu2::abi_400::detail::invocation::invoke<>()
    @     0xaaaaac005868         64  fu2::abi_400::detail::type_erasure::invocation_table::function_trait<>::internal_invoker<>::invoke()
    @     0xaaaaac4433fc         96  fu2::abi_400::detail::type_erasure::tables::vtable<>::invoke<>()
    @     0xaaaaac443498        176  fu2::abi_400::detail::type_erasure::erasure<>::invoke<>()
    @     0xaaaaac443514         32  fu2::abi_400::detail::type_erasure::invocation_table::operator_impl<>::operator()()
    @     0xaaaaac440878         80  dfly::CommandId::Invoke()
    @     0xaaaaac3f07a0        512  dfly::Service::InvokeCmd()
    @     0xaaaaac61eb00       1248  dfly::MultiCommandSquasher::SquashedHopCb()
    @     0xaaaaac61f070        192  dfly::MultiCommandSquasher::ExecuteSquashed()::{lambda()#1}::operator()()
    @     0xaaaaac62349c         32  std::__invoke_impl<>()
    @     0xaaaaac6232dc         32  std::__invoke_r<>()
    @     0xaaaaac62300c         48  std::_Function_handler<>::_M_invoke()
    @     0xaaaaabc6d450         32  std::function<>::operator()()
    @     0xaaaaace14d54        256  util::fb2::FiberQueue::Run()
    @     0xaaaaac936bec         32  dfly::TaskQueue::Start()::{lambda()#1}::operator()()
    @ ... and at least 11 more frames
Aborted
```

vyavdoshenko:
`BITFIELD` and `SETBIT` grow a string to a byte index derived from the bit offset with no upper bound, and `BITFIELD`/`BITOP` were missing the OOM guard. A tiny request (e.g. `BITFIELD k SET u8 99999999999 1`) could therefore trigger an unbounded allocation and abort the process under a virtual-memory cap.

Changes:
- Reject bit offsets whose touched byte exceeds `max_bulk_len` in `SETBIT` and `BITFIELD` (with overflow guard for the `#` offset form).
- Mark `BITFIELD` and `BITOP` with `CO::DENYOOM`, matching `SETBIT`/`SETRANGE`, so they are rejected on OOM instead of allocating.

Tests:
- C++ unit tests: `SETBIT`/`BITFIELD` reject out-of-range offsets.
- Python tests (`opt_only`): under a memory cap, `BITFIELD`/`BITOP` return an OOM error and the server stays alive instead of crashing.

Fixes: #7512


src/server/bitops_family.cc
```diff
@@ -2,6 +2,8 @@
 // See LICENSE for licensing terms.
 //
 
+#include <absl/flags/declare.h>
+#include <absl/flags/flag.h>
```
BorysTheDev: do you really need both?

vyavdoshenko: fixed

vyavdoshenko: CI proved yes, I need both

src/server/bitops_family.cc
```diff
@@ -975,6 +979,19 @@ OpResult<vector<ResultType>> StateExecutor::Execute(const CommandList& commands)
...
     "invalid bitfield type. use something like i16 u8. note that u64 is not supported but i64 is.";
 
+const char kBitOffsetOutOfRange[] = "bit offset is not an integer or out of range";
+
+// Keeps the bit offset within the uint32 range used by the bit-index helpers and
+// bounds the touched byte by max_bulk_len, so a tiny request can neither wrap to a
+// different position nor allocate unboundedly.
+bool IsBitOffsetInRange(size_t bit_offset, size_t bit_size) {
+  if (bit_offset > std::numeric_limits<uint32_t>::max())
+    return false;
```
BorysTheDev: This check is redundant. Next check do the same and more strict `last_bit <= std::numeric_limits<uint32_t>::max()` because your types are size_t, but prefer uint64_t

src/server/bitops_family.cc
```diff
@@ -1024,8 +1041,15 @@ nonstd::expected<CommonAttributes, string> ParseCommonAttr(CmdArgParser* parser)
     return make_unexpected(kSyntaxErr);
   }
   if (is_proxy) {
+    // guard against overflow when scaling the field index to a bit offset
+    if (parsed.offset > std::numeric_limits<size_t>::max() / parsed.encoding_bit_size) {
```
BorysTheDev: are you sure that you should compare with size_t max?

src/server/bitops_family.cc
```diff
@@ -1024,8 +1041,15 @@ nonstd::expected<CommonAttributes, string> ParseCommonAttr(CmdArgParser* parser)
...
+      return make_unexpected(kBitOffsetOutOfRange);
+    }
     parsed.offset = parsed.offset * parsed.encoding_bit_size;
   }
+  if (!IsBitOffsetInRange(parsed.offset, parsed.encoding_bit_size)) {
```
BorysTheDev: quite the same check, I think you sshould define the rules and check the once, because now it looks awkward

src/server/bitops_family.cc
BorysTheDev: Improve this logic too, maybe we can add something into CmdArgParser.

src/server/bitops_family.cc
vyavdoshenko: fixed

src/server/bitops_family.cc
vyavdoshenko: fixed

src/server/bitops_family.cc
vyavdoshenko: fixed

src/server/bitops_family.cc
vyavdoshenko: added new helper to CmdArgParser

tests/dragonfly/memory_test.py
```diff
@@ -446,3 +446,42 @@ async def test_expiry_heartbeat_responsiveness(df_factory: DflyInstanceFactory):
...
+        await client.execute_command("BITOP", "AND", "dest", "filler", "filler")
+    assert await client.ping()
+
+
+@pytest.mark.opt_only
+async def test_bitfield_denyoom(df_factory: DflyInstanceFactory):
+    client = await _start_over_maxmemory(df_factory)
+    with pytest.raises(redis.exceptions.ResponseError, match="[Oo]ut of memory"):
+        await client.execute_command("BITFIELD", "bk", "SET", "u8", str(BIT_OOM_HUGE_OFFSET), "1")
+    assert await client.ping()
```
dranikpg: could've been united to avoid two instance startups/shutdowns

dranikpg (dismissed):

kostasrim (commented):
I don't see any comments about functionality so here they are:

1. ParseCommonAttr now rejects large offsets for all subcommands including GET. Redis allows BITFIELD k GET u8 4294967295 on a missing key and returns 0. So we just broke compatibility here.
2. IsBitOffsetInRange calls absl::GetFlag on every invocation
3.  BITOP writes max(len(src1), len(src2), ...) bytes. If the sources already fit in memory, the result does too. CO::DENYOOM will reject it even when no net allocation is happening (e.g., overwriting an equally-sized destination). In general any bit operation might not need to grow if the allocation is already there (overwrite).
4. I find the use of the flag max_bulk_len `confusing`. What does max protocol payload sz has to do with bitfields ? The flag is more of like "reject large wire payloads" vs now "reject bit offsets would cause large server-side allocation".  if someone bumps `max_bulk_len` to allow larger payloads they also `unknownigly" allow larger bit offsets....

I find the premise for (3):

```
 a SETBIT or BITFIELD SET/INCRBY that writes to bit offset N will grow the string to byte index N/8
```

Slightly off as this is not a guaranteed allocation on every path. We are slightly over restricting but it's ok for now.

kostasrim:
> I don't see any comments about functionality so here they are:
> 
> 1. ParseCommonAttr now rejects large offsets for all subcommands including GET. Redis allows BITFIELD k GET u8 4294967295 on a missing key and returns 0. So we just broke compatibility here.
> 2. IsBitOffsetInRange calls absl::GetFlag on every invocation
> 3. BITOP writes max(len(src1), len(src2), ...) bytes. If the sources already fit in memory, the result does too. CO::DENYOOM will reject it even when no net allocation is happening (e.g., overwriting an equally-sized destination). In general any bit operation might not need to grow if the allocation is already there (overwrite).
> 4. I find the use of the flag max_bulk_len `confusing`. What does max protocol payload sz has to do with bitfields ? The flag is more of like "reject large wire payloads" vs now "reject bit offsets would cause large server-side allocation".  if someone bumps `max_bulk_len` to allow larger payloads they also `unknownigly" allow larger bit offsets....
> 
> I find the premise for (3):
> 
> ```
>  a SETBIT or BITFIELD SET/INCRBY that writes to bit offset N will grow the string to byte index N/8
> ```
> 
> Slightly off as this is not a guaranteed allocation on every path. We are slightly over restricting but it's ok for now.

To summarize:

1. We should fix the compatibility we broke
2. Consider caching the flag
3. max_bulk_len is irrelevant in this context -> a new flag would be better. wire max payload size should not be related with bitfield operations

The rest don't matter as much (for example being over restrictive)

src/server/bitops_family.cc
```diff
@@ -975,11 +979,23 @@ OpResult<vector<ResultType>> StateExecutor::Execute(const CommandList& commands)
 const char kInvalidBitfieldTypeErr[] =
     "invalid bitfield type. use something like i16 u8. note that u64 is not supported but i64 is.";
 
+const char kBitOffsetOutOfRange[] = "bit offset is not an integer or out of range";
+
+// The last touched bit (bit_offset + bit_size - 1) must fit the uint32 the bit-index
+// helpers use and within max_bulk_len, so a tiny request can neither wrap to a
```
romange: I am sorry, how max_bulk_len is related to bits operations?!

romange:
I do not think that `max_bulk_len` is the right way to limit this. These are unrelated so not clear why would we do that. We currently have hardcoded limit on strings held if I remember correctly. We could use the same limit here.

tests/dragonfly/memory_test.py
vyavdoshenko: fixed

src/server/bitops_family.cc
vyavdoshenko: fixed

kostasrim (approved):
lgtm 🚢 🇮🇹

Issue #7471, opened 2026-06-01
Create test to assert memory differences created by GLOG/ABSL differences


abhijat:
https://github.com/dragonflydb/dragonfly/pull/7446#issuecomment-4590048254

test to expose memory usage change between LEGACY_GLOG (on vs off) 

In ON mode we use glog which uses less memory
In OFF mode we use absl which uses more memory

This is a runtime characteristic, influenced by the amount of log we produce

romange:
Also I would like to understand where the allocations come from.

abhijat:
Looking at absl vs glog source code, there is an allocation per message in absl which was not the behavior in glog

constructor of log message creates data struct
https://github.com/abseil/abseil-cpp/blob/20250512.1/absl/log/internal/log_message.cc#L276-L279

two buffers get created
https://github.com/abseil/abseil-cpp/blob/20250512.1/absl/log/internal/log_message.cc#L177
https://github.com/abseil/abseil-cpp/blob/20250512.1/absl/log/internal/log_message.cc#L199

both of size 15 kib
https://github.com/abseil/abseil-cpp/blob/20250512.1/absl/log/internal/log_message.h#L57

but this is short lived 30 kib per message. 
glog seems to reuse same memory area using placement new

but even with absl it should be churn not growing usage over time, looking at collected rss from the failing tests it seems absl increases very little but tips over the test into failure

```
  ┌─────────────────┬───────────────────────────────────────┬───────┬─────────────┬───────────┬──────────┬──────────────────────┐
  │     Run ID      │               Scenario                │ Fails │ used_memory │    rss    │ rss/used │ AnonHugePages (proc) │
  ├─────────────────┼───────────────────────────────────────┼───────┼─────────────┼───────────┼──────────┼──────────────────────┤
  │ 26941431711     │ absl, pair — baseline, uninstrumented │ 3/25  │ —           │ —         │ —        │ —                    │
  ├─────────────────┼───────────────────────────────────────┼───────┼─────────────┼───────────┼──────────┼──────────────────────┤
  │ 26942950298 (B) │ glog, pair, THP on                    │ 0/25  │ 266.3 MiB   │ 337.7 MiB │ 1.268    │ 298 MiB              │
  ├─────────────────┼───────────────────────────────────────┼───────┼─────────────┼───────────┼──────────┼──────────────────────┤
[...]

PR #7514, opened 2026-06-05
CI/build: switch from glog to absl

abhijat:
Recently we switched from glog -> absl and the test in https://github.com/dragonflydb/dragonfly/pull/7446 started failing. We briefly switched absl -> glog again. 

After investigation https://github.com/dragonflydb/dragonfly/issues/7471 it is seen that absl does use marginal, constant higher memory around 2 MiB. 

The test has been fixed separately in the linked PR, here we switch back to absl since a small constant overhead is acceptable.


abhijat:
test_reply_count failed which is a known issue but re-triggered to be sure

romange (approved):

PR #7516, opened 2026-06-05
server: Remove redundant ScheduleContext struct

abhijat:
with recent removal of scheduled batch in shard the context struct is no longer needed

ref https://github.com/dragonflydb/dragonfly/pull/7507


romange (approved):
I do not think this matters

PR #7517, opened 2026-06-06
fix: slave_repl_offset should be non zero for valkey master

kostasrim:
`slave_repl_offset` was improperly set to `0` when dragonfly replicated a redis/valkey instance. The correct value is the aggregate of bytes read from ConsumeRedisStream

* add test
* fix the metric to be non zero for redis/valkey master

### DIFF
--- a/tests/dragonfly/redis_replication_test.py
+++ b/tests/dragonfly/redis_replication_test.py
@@ -378,19 +378,25 @@ async def test_redis_replication_info_offset(df_factory, redis_server, port_pick
     # Give the ACK fiber time to fire (interval is 1000ms).
     await asyncio.sleep(2)
 
-    master_info = await c_master.info("replication")
     replica_info = await c_replica.info("replication")
-
-    slave_entry = next(
-        (v for k, v in master_info.items() if k.startswith("slave") and isinstance(v, dict)),
-        None,
-    )
-    assert slave_entry is not None, "Master sees no slaves"
-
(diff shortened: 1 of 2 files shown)


tests/dragonfly/redis_replication_test.py
```diff
@@ -355,3 +355,42 @@ async def test_disconnect_master(
...
+        None,
+    )
+    assert slave_entry is not None, "Master sees no slaves"
+
+    master_offset = slave_entry["offset"]
+    replica_offset = replica_info.get("slave_repl_offset", 0)
+
+    assert replica_offset > 0, f"slave_repl_offset={replica_offset}, expected non-zero"
+    assert (
+        replica_offset == master_offset
```
vyavdoshenko: IMO, it can be flaky. It depends on the timeout.

vyavdoshenko (dismissed):
lgtm

tests/dragonfly/redis_replication_test.py
kostasrim: it can because we call `info` in two separate clients so the in between the two nodes might not agree -- I will use assert eventually

kostasrim: @vyavdoshenko ptal!

vyavdoshenko (dismissed):

vyavdoshenko (approved):

PR #7523, opened 2026-06-07
refactor: reduce threads contention for INFO cmd

BorysTheDev:
related to: #7414
## Reduce thread contention for INFO

`ServerFamily::GetMetrics()` fanned every proactor's stats collection into a single
shared `Metrics result` guarded by a mutex. Because the lock was taken at the start
of the callback, all per-thread work (`GetStats`, `MergeCallStats`, etc.) ran
serialized — defeating the parallel fan-out.

### Change
- Drop the mutex. Each proactor writes its own slot in a `std::vector<Metrics>`
  (indexed by proactor id), fully in parallel and lock-free.
- Fold the per-thread partials into the aggregate on the calling thread via a new
  `MergeShardMetrics()` helper (sums, plus `max` for `tx_queue_len` and `min` for
  `oldest_pending_send_ts`).
- Uses AwaitBrief instead of AwaitFiberOnAll

# INFO command benchmark — main vs current

Measures the cost of the `INFO` command (full / all sections), which fans out across
every proactor to collect metrics, comparing **main** (mutex-guarded shared accumulator +
`AwaitFiberOnAll`) against the **current** branch (per-proactor partial-`Metrics` vector +
`AwaitBrief`).

## Setup

- **Build:** release (`-O3 -DNDEBUG`), the canonical `blaze.sh -release` configuration.
- **Hardware pinning** (cores isolated via `taskset`, distinct physical cores):
  - Dragonfly: 4 proactor threads on cores **0–3**.
  - Background traffic: cores **4–6**.
  - Single INFO client: core **7**.
- **Dataset:** 50,000 keys across 6 types (string/hash/list/set/zset/string-with-TTL).
- **Background load (the "other" commands):** full-tilt `redis-benchmark` GET (`-c 40`) and
  SET (`-c 20`) over the 50k keyspace; CPU is *not* capped — it settles at ~90%.
- **INFO measurement:** one closed-loop client issues `INFO` for a 6 s window
  (one outstanding call at a time), recording client-side round-trip latency.
- **Metrics:**
  - *INFO executions / rate* — count of INFO calls completed in the window.
  - *INFO p50/p90/p99* — client-side round-trip latency (includes loopback network).
  - *INFO server µs/call* — Dragonfly's own `cmdstat_info:usec_per_call` (server-side
    processing time only, excludes network).
  - *GET/SET throughput* — from Dragonfly's `cmdstat` counters over the window.
  - *CPU* — Dragonfly process `utime+stime` as a percentage of the 4 allocated cores.
- 3 interleaved rounds per build; values below are the averages.
- Harness: `tests/dragonfly/info_benchmark_test.py` (and `/tmp/info_pinned_bench.py` for
  the pinned/under-load variant).

## Results

| metric | main | current | Δ |
|---|---:|---:|---:|
| INFO executions (6 s window) | 35,555 | 41,528 | **+16.8%** |
| INFO rate | 5,926/s | 6,921/s | **+16.8%** |
| INFO p50 | 163 µs | 137 µs | −15.8% |
| INFO p90 | 181 µs | 159 µs | −12.3% |
| INFO p99 | 255 µs | 229 µs | −10.1% |
| INFO server µs/call (cmdstat) | 73.6 | 45.3 | **−38.4%** |
| GET+SET throughput | ~602k/s | ~592k/s | ~flat (noise) |
| CPU | 90% | 87% | −3 pts |

Per-round INFO server µs/call — main: `73.0 / 73.6 / 74.3`; current: `44.2 / 42.6 / 49.1`
(clean separation, no overlap).

### DIFF
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -3283,7 +3283,10 @@ Metrics ServerFamily::GetMetrics(Namespace* ns, bool collect_replication_memory)
     result.interned_string_stats = GetInternedStringStats();
   };  // cb
 
-  service_.proactor_pool().AwaitFiberOnAll(std::move(cb));
+  // AwaitBrief (not AwaitFiberOnAll): the callback only reads thread-local/atomic state into
+  // its own partials slot and never yields, so it can run directly in each IO loop, avoiding
+  // a per-proactor fiber spawn on every INFO call.
+  service_.proactor_pool().AwaitBrief(std::move(cb));
 
   // Fold per-thread partials into the aggregate result on this thread.
   for (const Metrics& partial : partials)
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -3120,15 +3120,59 @@ void ServerFamily::Save(CmdArgList args, CommandContext* cmd_cntx) {
   }
 }
 
-static void MergeDbSliceStats(const DbSlice::Stats& src, Metrics* dest) {
+// Folds a per-proactor partial Metrics (filled lock-free in the fan-out callback)
+// into the aggregate `dest`. Mirrors the combine semantics used while collecting:
+// almost everything sums, tx_queue_len takes the max and oldest_pending_send_ts the min.
+static void MergeShardMetrics(const Metrics& src, Metrics* dest) {
+  // Per-db stats / events / small_string_bytes are merged element-wise.
   if (src.db_stats.size() > dest->db_stats.size())
     dest->db_stats.resize(src.db_stats.size());
-
   for (size_t i = 0; i < src.db_stats.size(); ++i)
     dest->db_stats[i] += src.db_stats[i];
-
   dest->events += src.events;
   dest->small_string_bytes += src.small_string_bytes;
+
+  // Aggregate sub-structs.
+  dest->shard_stats += src.shard_stats;
+  dest->facade_stats += src.facade_stats;
+  dest->tiered_stats += src.tiered_stats;
+  dest->search_stats += src.search_stats;
+  dest->coordinator_stats.Add(src.coordinator_stats);
+  dest->qlist_stats += src.qlist_stats;
+  dest->replication_stats += src.replication_stats;
+  dest->lua_stats += src.lua_stats;
+  dest->interned_string_stats += src.interned_string_stats;
+
+  // Scalar sums.
+  dest->qps += src.qps;
+  dest->heap_used_bytes += src.heap_used_bytes;
+  dest->serialization_bytes += src.serialization_bytes;
+  dest->traverse_ttl_per_sec += src.traverse_ttl_per_sec;
+  dest->delete_ttl_per_sec += src.delete_ttl_per_sec;
+  dest->fiber_switch_cnt += src.fiber_switch_cnt;
+  dest->fiber_switch_delay_usec += src.fiber_switch_delay_usec;
+  dest->fiber_longrun_cnt += src.fiber_longrun_cnt;
+  dest->fiber_longrun_usec += src.fiber_longrun_usec;
+  dest->worker_fiber_stack_size += src.worker_fiber_stack_size;
+  dest->worker_fiber_count += src.worker_fiber_count;
+  dest->blocked_tasks += src.blocked_tasks;
+  dest->tls_bytes += src.tls_bytes;
+  dest->refused_conn_max_clients_reached_count += src.refused_conn_max_clients_reached_count;
+  dest->lsn_buffer_size += src.lsn_buffer_size;
+  dest->lsn_buffer_bytes += src.lsn_buffer_bytes;
+
+  // Non-sum reductions.
+  dest->tx_queue_len = std::max(dest->tx_queue_len, src.tx_queue_len);
+  dest->oldest_pending_send_ts = std::min(dest->oldest_pending_send_ts, src.oldest_pending_send_ts);
+
+  // Map merges.
+  for (const auto& [k, v] : src.connections_lib_name_ver_map)
+    dest->connections_lib_name_ver_map[k] += v;
+  for (const auto& [name, stat] : src.cmd_stats_map) {
+    auto& [calls, sum] = dest->cmd_stats_map[name];
+    calls += stat.first;
+    sum += stat.second;
+  }
 }
 
 void ServerFamily::ResetStat(Namespace* ns) {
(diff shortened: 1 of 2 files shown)


romange:
it complicates code. Is it something you can measure? I prefer to avoid complicating code without a measurable improvement

BorysTheDev:
@romange yes, I've made small benchmark for INFO cmd. this code faster on 6-17% for 8 threads. for 2 threads no difference. and if we have more threads the difference should be more

romange:
when you say "faster" what do you mean? Please provide a specific description of before and after.

See an example how I recently did it:
https://github.com/dragonflydb/dragonfly/pull/7412#issuecomment-4551614085

kostasrim:
> ## Reduce thread contention for INFO
> `ServerFamily::GetMetrics()` fanned every proactor's stats collection into a single shared `Metrics result` guarded by a mutex. Because the lock was taken at the start of the callback, all per-thread work (`GetStats`, `MergeCallStats`, etc.) ran serialized — defeating the parallel fan-out.
> 
> ### Change
> * Drop the mutex. Each proactor writes its own slot in a `std::vector<Metrics>`
>   (indexed by proactor id), fully in parallel and lock-free.
> * Fold the per-thread partials into the aggregate on the calling thread via a new
>   `MergeShardMetrics()` helper (sums, plus `max` for `tx_queue_len` and `min` for
>   `oldest_pending_send_ts`).

we don't spam info command and even your `vector` can still `fall short and cause core-2-core sync` and that 7% is `artifical`. See https://en.wikipedia.org/wiki/False_sharing

kostasrim:
https://github.com/dragonflydb/dragonfly/issues/7524 for the failing test -- seems irrelevant to the changes

BorysTheDev:
@romange I've added benchmark description and results

BorysTheDev:
> > ## Reduce thread contention for INFO
> > `ServerFamily::GetMetrics()` fanned every proactor's stats collection into a single shared `Metrics result` guarded by a mutex. Because the lock was taken at the start of the callback, all per-thread work (`GetStats`, `MergeCallStats`, etc.) ran serialized — defeating the parallel fan-out.
> > ### Change
> > 
> > * Drop the mutex. Each proactor writes its own slot in a `std::vector<Metrics>`
> >   (indexed by proactor id), fully in parallel and lock-free.
> > * Fold the per-thread partials into the aggregate on the calling thread via a new
> >   `MergeShardMetrics()` helper (sums, plus `max` for `tx_queue_len` and `min` for
> >   `oldest_pending_send_ts`).
> 
> we don't spam info command and even your `vector` can still `fall short and cause core-2-core sync` and that 7% is `artifical`. See https://en.wikipedia.org/wiki/False_sharing

This is not about spam, this is about INFO execution time. @romange told me that we have a problem with it.

kostasrim:
> > > ## Reduce thread contention for INFO
> > > `ServerFamily::GetMetrics()` fanned every proactor's stats collection into a single shared `Metrics result` guarded by a mutex. Because the lock was taken at the start of the callback, all per-thread work (`GetStats`, `MergeCallStats`, etc.) ran serialized — defeating the parallel fan-out.
> > > ### Change
> > > 
> > > * Drop the mutex. Each proactor writes its own slot in a `std::vector<Metrics>`
> > >   (indexed by proactor id), fully in parallel and lock-free.
> > > * Fold the per-thread partials into the aggregate on the calling thread via a new
> > >   `MergeShardMetrics()` helper (sums, plus `max` for `tx_queue_len` and `min` for
> > >   `oldest_pending_send_ts`).
> > 
> > 
> > we don't spam info command and even your `vector` can still `fall short and cause core-2-core sync` and that 7% is `artifical`. See https://en.wikipedia.org/wiki/False_sharing
> 
> This is not about spam, this is about INFO execution time. @romange told me that we have a problem with it.

Sure but it's also important to understand the context. We have a problem when `info blocks for seconds` not when its latency bumps up by `30microseconds`. We need `info` responsive (in ms) so we don't get health checks failing (because a mutex was held under IO). This is not the case here. 

Also:

GET+SET throughput | ~602k/s | ~592k/s | ~flat (noise)
-- | -- | -- | --

This might not be noise, you took CPU time by doing the reduction (e.g the merge on the vector after you collect the per thread results). 

The latency of INFO dropped but now the fiber has to do "extra work" (that's extra CPU slice). Yes now fibers don't race over the mutex. The difference before is you were "idle" and some other fiber would execute. Now your reduction step is not sleeping and it does "extra work" at the end.

p.s.1 I have no freaking clue if the reduction can penalize. I am trying to make a point here as a whole
p.s.2 How did those 500 extra info calls your optimization added help here in our case (you are spamming INFO and the measurement is wrong) ? 😄

BorysTheDev:
@kostasrim, it's one of the PRs that I'm going to provide. I'm going to remove all mutexes for INFO. Also, this mutex is a problem if we have 128 cores, for example. because every core struggles to get this mutex, and the latency increases.

kostasrim:
> , it's one of the PRs that I'm going to provide. I'm going to remove all mutexes for INFO

But what's the problem with "all the mutexes"? Why would you care for CPU-only cases ? I understand that if a mutex is kept locked under `IO` is a problem (lock under non CPU work) but I don't understand why is it a problem for CPU-only cases?

> problem if we have 128 cores, for example. because every core struggles to get this mutex, and the latency incr

Do you have a proof for this ? That would be true if *all fibers* under the same `core` raced over the same `mutex`. We have `1 fiber per CPU from the THOUSANDS` racing over that mutex. The actual congestion window is extremely thin. You need the same fiber to be scheduled and become active *on all* threads at the same time for them to congest. Under steady load, I highly doubt that what we schedule to the proactors ends up executing at the same time T.

BorysTheDev:
> e `1 fiber per CPU from the THOUSANDS` racing over that mutex. The actual congestion window is extremely thin. You need the same fiber to be scheduled and become active _on all_ threads at the same time for them to congest. Under steady load, I highly doubt that what we schedule to the proactors ends up executing

OK not all mutexes, only the ones that affect performance. I benchmark every change.

> problem if we have 128 cores, for example. because every core struggles to get this mutex, and the latency incr

> Do you have a proof for this ? 

It's logic. mutex prevents parallel work. so 128 cores should wait 127 * (calculations + cache invalidation (sometimes even TLB)). Also, my benchmark shows quite good latency improvement.

kostasrim:
> > e `1 fiber per CPU from the THOUSANDS` racing over that mutex. The actual congestion window is extremely thin. You need the same fiber to be scheduled and become active _on all_ threads at the same time for them to congest. Under steady load, I highly doubt that what we schedule to the proactors ends up executing
> 
> OK not all mutexes, only the ones that affect performance. I benchmark every change.
> 
> > problem if we have 128 cores, for example. because every core struggles to get this mutex, and the latency incr
> 
> > Do you have a proof for this ?
> 
> It's logic. mutex prevents parallel work. so 128 cores should wait 127 * (calculations + cache invalidation (sometimes even TLB)). Also, my benchmark shows quite good latency improvement.

The bottleneck was never the absence of parallel work. Or better the serialized execution imposed by the mutex is not an issue. Locks under non-cpu-only operations are the bottleneck or any other flow that can unboundedly hit the latency of the `INFO` command.

BorysTheDev:
> > > e `1 fiber per CPU from the THOUSANDS` racing over that mutex. The actual congestion window is extremely thin. You need the same fiber to be scheduled and become active _on all_ threads at the same time for them to congest. Under steady load, I highly doubt that what we schedule to the proactors ends up executing
> > 
> > 
> > OK not all mutexes, only the ones that affect performance. I benchmark every change.
> > > problem if we have 128 cores, for example. because every core struggles to get this mutex, and the latency incr
> > 
> > 
> > > Do you have a proof for this ?
> > 
> > 
> > It's logic. mutex prevents parallel work. so 128 cores should wait 127 * (calculations + cache invalidation (sometimes even TLB)). Also, my benchmark shows quite good latency improvement.
> 
> The bottleneck was never the absence of parallel work. Or better the serialized execution imposed by the mutex is not an issue. Locks under non-cpu-only operations are the bottleneck or any other flow that can unboundedly hit the latency of the `INFO` command.

But long execution time and inablility of parallel work make this worse. Execution time reduction should improve the situation anyway.

src/server/server_family.cc
```diff
@@ -3121,15 +3121,59 @@ void ServerFamily::Save(CmdArgList args, CommandContext* cmd_cntx) {
   }
 }
 
-static void MergeDbSliceStats(const DbSlice::Stats& src, Metrics* dest) {
+// Folds a per-proactor partial Metrics (filled lock-free in the fan-out callback)
+// into the aggregate `dest`. Mirrors the combine semantics used while collecting:
+// almost everything sums, tx_queue_len takes the max and oldest_pending_send_ts the min.
+static void MergeShardMetrics(const Metrics& src, Metrics* dest) {
```
romange: missing static_assert

src/server/server_family.cc
```diff
@@ -3152,93 +3196,102 @@ void ServerFamily::ResetStat(Namespace* ns) {
...
-    auto& [calls, sum] = dest[absl::AsciiStrToLower(name)];
-    calls += stat.first;
-    sum += stat.second;
-  };
+  // Each proactor accumulates into its own slot (indexed by proactor id), so the
+  // fan-out callback runs fully in parallel with no shared-state locking. The
+  // per-thread partials are folded into `result` on this thread afterwards.
+  std::vector<Metrics> partials(service_.proactor_pool().size());
 
   auto cb = [&](unsigned index, ProactorBase* pb) {
```
romange: what do we need to capture here besides `partials`?

src/server/server_family.cc
```diff
@@ -3152,93 +3196,102 @@ void ServerFamily::ResetStat(Namespace* ns) {
...
     EngineShard* shard = EngineShard::tlocal();
     ServerState* ss = ServerState::tlocal();
+    // Each proactor writes its own freshly-zeroed slot exactly once, so plain assignment
+    // is enough — cross-thread accumulation happens later in MergeShardMetrics. Named
+    // `result` (shadowing the outer aggregate) to keep this body close to the original.
+    Metrics& result = partials[index];
```
romange: if it's pure thread-local and stateless, can you factor out this code into a `Metrics::InitFromThread()` member function that fills all these fields.

src/server/server_family.cc
romange: Similarly to below - lets move it to Metrics function. Also add a todo - to move it into metrics.h/cc.

romange (commented):
lgtm

BorysTheDev:
augment review

src/server/server_family.cc
```diff
@@ -3120,15 +3120,73 @@ void ServerFamily::Save(CmdArgList args, CommandContext* cmd_cntx) {
...
+                             sizeof(EngineShard::Stats) + sizeof(facade::FacadeStats) +
+                             sizeof(TieredStats) + sizeof(SearchStats) +
+                             sizeof(ServerState::Stats) + sizeof(PeakStats) + sizeof(QList::Stats) +
+                             sizeof(ReplicationMemoryStats) + sizeof(InterpreterManager::Stats) +
+                             sizeof(std::map<std::string, std::pair<uint64_t, uint64_t>>) +
+                             sizeof(absl::flat_hash_map<std::string, uint64_t>) +
+                             sizeof(std::optional<Metrics::ReplicaInfo>) + sizeof(LoadingStats) +
+                             sizeof(absl::flat_hash_map<std::string, hdr_histogram*>) +
+                             sizeof(InternedStringStats) + sizeof(acl::UserRegistry::AclStats) +
+                             176,  // scalar fields (19 fields) + 4-byte alignment padding
```
BorysTheDev, line 3138: collections have different size on different platforms or for type of build, so we need to calculate them

romange (approved):

PR #7528, opened 2026-06-07
chore: introduce CmdRef and generator-based MultiCommandSquasher

romange:
## Summary

Avoid deep-copy of multi/exec arguments and remove reliance of MultiCommandSquasher on StoredCmd.

- Introduce `CmdRef` — a lightweight non-owning view (`{cid, ParsedArgs, ReplyMode}`) that both `StoredCmd` and `CommandContext` can produce
- Change `MultiCommandSquasher` from `Span<StoredCmd>` to a generator function (`std::function<CmdRef()>`) with invalid-CmdRef sentinel, eliminating intermediate vector materialization at EXEC and Lua call sites
- Add swap-based `StoredCmd` constructor for the EXEC path — moves args from `ParsedCommand` via `BackedArguments::SwapArgs` instead of deep-copying
- Change `DispatchManyCommands` generator to return `ParsedCommand*`, allowing direct use of `CommandContext` without a dummy instance

## Test plan
- [x] `multi_test` — 58 passed, 2 skipped
- [x] `generic_family_test` — 76 passed, 1 skipped
- [x] `string_family_test` — 39 passed

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -2574,9 +2574,6 @@ bool Connection::ExecuteBatch() {
     return parsed_head_;
   };
 
-  auto dispatch = protocol_ == Protocol::MEMCACHE ? &ServiceInterface::DispatchMC
-                                                  : &ServiceInterface::DispatchCommandSimple;
-
   // Execute sequentially all parsed commands.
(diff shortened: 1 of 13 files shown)


romange:
Superceded by #7551

PR #7534, opened 2026-06-08
fix(search): support parenthesized and filtered VECTOR_RANGE queries

### LINKED ISSUE #7521: `FT.SEARCH` rejects a parenthesized `VECTOR_RANGE` query (`Query syntax error`)
Dragonfly's query parser accepts a **bare** vector range query (`@field:[VECTOR_RANGE …]=>{…}`) but returns `Query syntax error` as soon as that range expression is **wrapped in parentheses** or **combined with a filter inside a group** - e.g. `(@vec:[VECTOR_RANGE …]=>{…} @tag:("x"))`.

This is exactly the shape that `redisvl` / `langchain-redis` generates for any range (distance-threshold) search, because they wrap the query in parentheses and AND-in an index-scoping filter. RediSearch (Redis 8.6) parses it fine.

---

Impact:

The official `langchain-redis` package (modern LangChain Redis integration, on `redisvl`) generates the failing form on these default code paths:

- `RedisVectorStore.similarity_search(query, distance_threshold=…)`
- `RedisVectorStore.similarity_search(query, filter=…)` combined with a range
- any `RedisSemanticCache` / vector lookup that carries a metadata filter

`langchain-redis` **always** injects `@_index_name:("<index>")` into the filter (`vectorstores.py`), and `redisvl`'s `VectorRangeQuery` wraps the whole expression in parentheses whenever a filter `!= *` is present - so the parens are unavoidable on these paths.

Plain KNN (`similarity_search(k=…)`) and the unfiltered `RedisSemanticCache.lookup` (a *bare* `VECTOR_RANGE`) are **not** affected and work correctly on Dragonfly.

---

Reproduction:

Minimal - isolated grammar boundary:

Run against Dragonfly and (for comparison) Redis 8.6:

```python
import redis, struct

def f32(xs):
    return struct.pack(f"{len(xs)}f", *xs)

r = redis.Redis(port=7399)            # Dragonfly build-dbg
r.flushall()
r.execute_command(
    "FT.CREATE", "ix", "ON", "HASH", "PREFIX", "1", "d:", "SCHEMA",
    "name", "TEXT",
    "embedding", "VECTOR", "FLAT", "6", "TYPE", "FLOAT32", "DIM", "2", "DISTANCE_METRIC", "COSINE",
)
r.hset("d:1", mapping={"name": "idxA", "embedding": f32([1.0, 0.0])})
q = f32([1.0, 0.0])

variants = {
    "bare VECTOR_RANGE":                 "@embedding:[VECTOR_RANGE $r $v]",
    "VR + YIELD":                        "@embedding:[VECTOR_RANGE $r $v]=>{$YIELD_DISTANCE_AS: vd}",
    "VR + YIELD in parens":              "(@embedding:[VECTOR_RANGE $r $v]=>{$YIELD_DISTANCE_AS: vd})",
    "VR + YIELD + filter in parens":     '(@embedding:[VECTOR_RANGE $r $v]=>{$YIELD_DISTANCE_AS: vd} @name:("idxA"))',
}
for label, qstr in variants.items():
    try:
        n = r.execute_command("FT.SEARCH", "ix", qstr, "PARAMS", "4", "r", "0.9", "v", q, "DIALECT", "2")[0]
        print(f"OK   | {label}: {n} hits")
    except Exception as e:
        print(f"FAIL | {label}: {e}")
```

**Observed on Dragonfly:**

```
OK   | bare VECTOR_RANGE: 1 hits
OK   | VR + YIELD: 1 hits
FAIL | VR + YIELD in parens: Query syntax error          <-- bug
FAIL | VR + YIELD + filter in parens: Query syntax error  <-- bug (langchain shape)
```

Expected: all four return `OK`.

Real-world - the exact query `langchain-redis` sends:

Captured via `MONITOR` while running `RedisVectorStore.similarity_search(distance_threshold=0.9)` (identical bytes sent to Dragonfly and Redis 8.6):

```
FT.SEARCH idx (@embedding:[VECTOR_RANGE $distance_threshold $vector]=>{$YIELD_DISTANCE_AS: vector_distance} @_index_name:("idx"))
  RETURN 4 text _index_name _metadata_json vector_distance
  SORTBY vector_distance ASC DIALECT 2 LIMIT 0 N PARAMS 4 distance_threshold 0.9 vector <blob>
```

- Dragonfly: `Error while searching: Query syntax error`
- Redis 8.6: returns results

---

Expected vs actual:

- **Expected:** a `VECTOR_RANGE` expression can be parenthesized and combined with a pre-filter, matching RediSearch behaviour, so `(@vec:[VECTOR_RANGE …]=>{…} <filter>)` is parsed and the filter is applied as a pre-filter to the range scan.
- **Actual:** any `VECTOR_RANGE` inside `( … )` (with or without an additional filter) -> `Query syntax error`.

---

Root cause:

`src/core/search/parser.y`:

```
final_query:
    filter                       (line 110)
  | filter ARROW knn_query       (line 112)   <-- KNN has a filtered form
  | vector_range_query           (line 114)   <-- VECTOR_RANGE: top-level ONLY

vector_range_query:              (lines 144-173)
    FIELD COLON LBRACKET VECTOR_RANGE … RBRACKET [ARROW LCURLBR YIELD_DISTANCE_AS …]

filter:        search_expr | star_expr            (line 180)
star_expr:     STAR | LPAREN star_expr RPAREN     (line 184)   <-- parens parsed here
```

`vector_range_query` is only reachable as a top-level `final_query` alternative
(`parser.y:114`). It is **not** a member of `search_expr` / `search_unary_expr` / `filter`,
so:

1. it cannot appear inside parentheses — `( … )` is parsed by the `filter` / `star_expr` grammar (`parser.y:180-186`), which has no `vector_range_query` member; and
2. there is no production combining a filter with a range query — unlike KNN, which has the dedicated `filter ARROW knn_query` rule (`parser.y:112`).

So the parenthesized / filtered range query has no matching production and fails.

---

Suggested fix:

Give `VECTOR_RANGE` a filtered/groupable form analogous to KNN. Options:

- add a `filter ARROW vector_range_query` (or `filter vector_range_query`) production so a pre-filter can be AND-ed with a range scan, **and/or**
- allow `vector_range_query` to be parenthesized (treat a parenthesized range expression as a `final_query`), so redisvl's `( … )` wrapping parses.

Either change makes `langchain-redis` `similarity_search(distance_threshold=…)` and filtered semantic-cache lookups work, closing the gap.

vyavdoshenko:
`FT.SEARCH` and `FT.AGGREGATE` rejected parenthesized and filtered `VECTOR_RANGE` queries with "Query syntax error" - exactly the shape langchain-redis / redisvl emit for `distance_threshold` and filtered range search. This makes `VECTOR_RANGE` a first-class query predicate that composes with filters on both FLAT and HNSW indexes.

Changes:
- Grammar: `VECTOR_RANGE` is now a regular query leaf, so it parses inside parentheses, AND-ed with a filter in any order, without parentheses, and when nested.
- FLAT: filtered ranges run natively; `knn_scores` is kept positionally aligned with the result ids (fixes an out-of-bounds read when a range is combined with a filter).
- HNSW: a bare range runs directly; a range AND-ed with a filter pre-filters and then intersects the range results (FT.SEARCH reuses loaded docs, FT.AGGREGATE loads only the survivors).
- HNSW guards: combining `VECTOR_RANGE` with OR/NOT, or using more than one `VECTOR_RANGE` clause, returns an explicit error instead of silently wrong results.
- Tests: parser combinations, a score-alignment regression, and FLAT/HNSW filtered-range and multi-range coverage for both FT.SEARCH and FT.AGGREGATE.

Fixes: #7521

### DIFF
--- a/src/core/search/parser.y
+++ b/src/core/search/parser.y
@@ -88,6 +88,12 @@ double toDouble(string_view src);
 %right NOT_OP TILDE
 %precedence LPAREN RPAREN
 
+// Dangling-else for the optional VECTOR_RANGE YIELD: at `]` with `=>` ahead, ARROW outranks the
+// no-yield rule (tagged %prec NO_YIELD), so bison shifts into the YIELD clause instead of reducing
+// and treating `=>` as a top-level KNN arrow.
+%precedence NO_YIELD
+%precedence ARROW
+
 %token <std::string> DOUBLE "double"
 %token <std::string> UINT32 "uint32"
 %nterm <AstExpr> final_query filter star_expr search_expr search_unary_expr search_or_expr search_and_expr bracket_filter_expr
@@ -111,8 +117,6 @@ final_query:
       { driver->Set(std::move($1)); }
   | filter ARROW knn_query
       { driver->Set(AstKnnNode(std::move($1), std::move($3))); }
-  | vector_range_query
-      { driver->Set(std::move($1)); }
 
 knn_query:
   LBRACKET KNN UINT32 FIELD TERM opt_ef_runtime opt_knn_alias RBRACKET
@@ -157,7 +161,7 @@ vector_range_query:
                                 std::move(alias));
       }
     }
-  | FIELD COLON LBRACKET VECTOR_RANGE vec_range_radius TERM RBRACKET
+  | FIELD COLON LBRACKET VECTOR_RANGE vec_range_radius TERM RBRACKET %prec NO_YIELD
     {
       double radius = $5;
       auto field = std::move($1);
@@ -215,6 +219,7 @@ search_unary_expr:
   | TILDE search_unary_expr           { $$ = AstOptionalNode(std::move($2)); }
   | term_atom                         { $$ = std::move($1);                  }
   | FIELD COLON field_cond            { $$ = AstFieldNode(std::move($1), std::move($3)); }
+  | vector_range_query                { $$ = std::move($1);                  }
 
 field_cond:
   term_atom                                             { $$ = std::move($1);                }
--- a/src/core/search/search.cc
+++ b/src/core/search/search.cc
@@ -564,7 +564,8 @@ struct BasicSearch {
     knn_distances_.resize(prefix_size);
   }
 
-  void SearchVectorRangeFlat(FlatVectorIndex* vec_index, const AstVectorRangeNode& node) {
+  void SearchVectorRangeFlat(FlatVectorIndex* vec_index, const AstVectorRangeNode& node,
+                             vector<DocId>* out) {
     const auto& all_docs = indices_->GetAllDocs();
     auto [dim, sim] = vec_index->Info();
     for (DocId doc : all_docs) {
@@ -573,7 +574,8 @@ struct BasicSearch {
         continue;
       float dist = VectorDistance(node.vec.first.get(), vec, dim, sim);
       if (dist <= static_cast<float>(node.radius)) {
-        knn_scores_.emplace_back(doc, dist);
+        knn_scores_[doc] = dist;
+        out->push_back(doc);
       }
     }
   }
@@ -606,12 +608,9 @@ struct BasicSearch {
     // HNSW fields are not stored in FieldIndices::indices_, so GetIndex<BaseVectorIndex> above
     // returns nullptr for HNSW before we reach this point.
     // HNSW range search support is planned separately (see hnsw_index.h).
+    vector<DocId> out;
     if (auto* flat_index = dynamic_cast<FlatVectorIndex*>(vec_index); flat_index)
-      SearchVectorRangeFlat(flat_index, node);
-
-    vector<DocId> out(knn_scores_.size());
-    for (size_t i = 0; i < knn_scores_.size(); i++)
-      out[i] = knn_scores_[i].first;
+      SearchVectorRangeFlat(flat_index, node, &out);
     return IndexResult{std::move(out)};
   }
 
@@ -643,8 +642,9 @@ struct BasicSearch {
     vector<DocId> out(knn_distances_.size());
     knn_scores_.reserve(knn_distances_.size());
 
+    // `out` carries KNN distance order; knn_scores_ is a by-id distance lookup.
     for (size_t i = 0; i < knn_distances_.size(); i++) {
-      knn_scores_.emplace_back(knn_distances_[i].second, knn_distances_[i].first);
+      knn_scores_[knn_distances_[i].second] = knn_distances_[i].first;
       out[i] = knn_distances_[i].second;
     }
 
@@ -684,23 +684,6 @@ struct BasicSearch {
     if (scorer_ && !matched_text_terms_.empty()) {
       // Score ALL matched docs and return top-K by score (not arbitrary cutoff).
       auto [out, total_size, text_scores] = TakeScoredTopK(std::move(result), cuttoff_limit);
-
-      // KNN populated knn_scores_ in distance order; TakeScoredTopK reordered
-      // `out` by text score, so realign knn_scores_ to the new id order.
-      // Consumers (e.g. search_family.cc) index ids and knn_scores by position.
-      if (!knn_scores_.empty()) {
-        absl::flat_hash_map<DocId, float> knn_by_id;
-        knn_by_id.reserve(knn_scores_.size());
-        for (auto& [doc, dist] : knn_scores_)
-          knn_by_id.emplace(doc, dist);
-        knn_scores_.clear();
-        knn_scores_.reserve(out.size());
-        for (DocId doc : out) {
-          if (auto it = knn_by_id.find(doc); it != knn_by_id.end())
-            knn_scores_.emplace_back(doc, it->second);
-        }
-      }
-
       return SearchResult{
           total_size,         std::move(out),   std::move(knn_scores_), std::move(text_scores),
           std::move(profile), std::move(error_)};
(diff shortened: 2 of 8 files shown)


src/core/search/parser.y
```diff
@@ -55,6 +55,9 @@ double toDouble(string_view src);
 %define parse.lac full
 %define api.token.prefix {TOK_}
 
+// One shift/reduce conflict: VECTOR_RANGE before `=>` shifts into its YIELD clause (correct).
+%expect 1
+
```
BorysTheDev: can we make the logic more strict?

src/core/search/search.cc
```diff
@@ -663,6 +663,24 @@ struct BasicSearch {
...
+  // Consumers index ids and knn_scores by position, so keep knn_scores_ 1:1 with `out`: a
+  // VECTOR_RANGE may be AND-ed/OR-ed with a filter, so `out` can differ from the scored set.
+  // Docs with no distance get 0.
+  void RealignKnnScores(const vector<DocId>& out) {
+    if (knn_scores_.empty())
+      return;
+    absl::flat_hash_map<DocId, float> by_id;
+    by_id.reserve(knn_scores_.size());
+    for (auto& [doc, dist] : knn_scores_)
+      by_id.emplace(doc, dist);
```
BorysTheDev: can we avoid such transfortmation and store it in flat_hash_map by default

src/core/search/search.cc
```diff
@@ -1225,8 +1229,66 @@ void SearchAlgorithm::SetScorer(ScorerFn scorer) {
...
+    return;
+  }
+  if (auto* neg = get_if<AstNegateNode>(node); neg)
+    CollectVectorRangeNodes(neg->node.get(), out);
+  else if (auto* opt = get_if<AstOptionalNode>(node); opt)
+    CollectVectorRangeNodes(opt->node.get(), out);
+  else if (auto* f = get_if<AstFieldNode>(node); f)
+    CollectVectorRangeNodes(f->node.get(), out);
+  else if (auto* knn = get_if<AstKnnNode>(node); knn)
+    CollectVectorRangeNodes(knn->filter.get(), out);
```
BorysTheDev: code is really similar, can we have visitor or some more clear pattern?

src/server/search/search_family.cc
```diff
@@ -1490,6 +1490,35 @@ vector<SearchResult> SearchGlobalHnswIndexRange(
...
+// matches, already loaded), reusing the loaded docs without an extra transaction hop. Exact for
+// a range query, unlike top-k KNN where dropping out-of-set docs after the scan loses results.
+vector<SearchResult> SearchGlobalHnswIndexRangePrefiltered(
+    const search::AstVectorRangeNode* range, const shared_ptr<search::HnswVectorIndex>& index,
+    std::vector<SearchResult>& sharded_prefilter_docs) {
+  absl::flat_hash_map<search::GlobalDocId, SerializedSearchDoc*> lookup;
+  for (size_t shard_id = 0; shard_id < sharded_prefilter_docs.size(); shard_id++) {
+    for (auto& doc : sharded_prefilter_docs[shard_id].docs)
+      lookup[search::CreateGlobalDocId(shard_id, doc.id)] = &doc;
+  }
```
BorysTheDev: you don't need this transformation. just do it vice versa, split global id on shard and doc_id

src/core/search/parser.y
vyavdoshenko: Resolved the ambiguity explicitly via precedence (dangling-else style): ARROW now outranks the no-yield vector_range_query rule (%prec NO_YIELD), so bison deterministically shifts into the YIELD clause. The grammar is now conflict-free (%expect 0), so any future conflict fails the build instead of silently keeping the count at 1. Generated parser is unchanged; parser tests stay green.

src/server/search/search_family.cc
```diff
@@ -3210,18 +3284,68 @@ void CmdFtAggregate(CmdArgList args, CommandContext* cmd_cntx) {
...
+
+        // Collect text scores per-shard from prefilter results for __score injection
+        if (params->add_scores) {
+          for (size_t shard_id = 0; shard_id < prefilter_docs.size(); shard_id++) {
+            for (const auto& doc : prefilter_docs[shard_id].docs)
+              prefilter_text_scores[shard_id][doc.id] = doc.text_score;
+          }
+        }
+      }
+
```
BorysTheDev: split CmdFtAggregate to several functions; it is impossible to review it.

src/server/search/search_family.cc
```diff
@@ -3210,18 +3284,68 @@ void CmdFtAggregate(CmdArgList args, CommandContext* cmd_cntx) {
...
+        auto ids = CollectPrefilterGlobalIds(prefilter_docs);
+        std::sort(ids.begin(), ids.end());
+        prefilter_ids = std::move(ids);
+
+        // Collect text scores per-shard from prefilter results for __score injection
+        if (params->add_scores) {
+          for (size_t shard_id = 0; shard_id < prefilter_docs.size(); shard_id++) {
+            for (const auto& doc : prefilter_docs[shard_id].docs)
+              prefilter_text_scores[shard_id][doc.id] = doc.text_score;
+          }
```
BorysTheDev: you do such transformations all the time, consider using plain sorted data structures and accessors. I don't understand memory complexity .

src/core/search/search.cc

src/core/search/search.cc

src/server/search/search_family.cc

src/server/search/search_family.cc

src/server/search/search_family.cc

src/core/search/parser.y
```diff
@@ -55,6 +55,9 @@ double toDouble(string_view src);
 %define parse.lac full
 %define api.token.prefix {TOK_}
 
+// No unresolved conflicts: the VECTOR_RANGE YIELD ambiguity is resolved by precedence below.
```
BorysTheDev: I think you don't need it anymore

src/core/search/search.cc
```diff
@@ -1225,8 +1208,65 @@ void SearchAlgorithm::SetScorer(ScorerFn scorer) {
...
+    for (const auto& child : l->nodes)
+      WalkAst(child, cb);
+  } else if (auto* neg = get_if<AstNegateNode>(&node)) {
+    WalkAst(*neg->node, cb);
+  } else if (auto* opt = get_if<AstOptionalNode>(&node)) {
+    WalkAst(*opt->node, cb);
+  } else if (auto* f = get_if<AstFieldNode>(&node)) {
+    WalkAst(*f->node, cb);
+  } else if (auto* knn = get_if<AstKnnNode>(&node)) {
+    WalkAst(*knn->filter, cb);
```
BorysTheDev: as I remember this is variant so you can use std::visit with several lambdas 

```
- [](AstLogicalNode*) { ... }
- [](AstKnnNode*) { WalkAst(*knn->filter, cb); }
- [](auto* n) { WalkAst(*n->node, cb);} // you have the same code so you can use template lambda
```

BorysTheDev (dismissed):

src/core/search/parser.y
vyavdoshenko: fixed

src/core/search/search.cc

BorysTheDev (approved):

PR #7536, opened 2026-06-08
chore(facade): replace io::Sink* with FiberSocketBase* in SinkReplyBuilder

romange:
## Summary
- Replace `io::Sink* sink_` with `util::FiberSocketBase* socket_` in `SinkReplyBuilder` and all subclasses, giving reply builders typed access to socket-specific features without downcasting
- Add `StringSocket` — a minimal `FiberSocketBase` subclass that captures writes to a `std::string` — replacing `io::StringSink` and `io::NullSink` at all construction sites

## Test plan
- [x] `reply_builder_test` passes (35/35)
- [x] `generic_family_test` passes
- [x] `dragonfly` binary builds cleanly
- [x] Pre-commit formatting passes

### DIFF
--- a/src/server/test_utils.cc
+++ b/src/server/test_utils.cc
@@ -21,12 +21,15 @@ extern "C" {
 #include "base/flags.h"
 #include "base/logging.h"
 #include "base/stl_util.h"
+#include "io/file_util.h"
+#include "util/fibers/pool.h"
+
+//
 #include "core/oah_set.h"
 #include "facade/dragonfly_connection.h"
 #include "facade/reply_builder.h"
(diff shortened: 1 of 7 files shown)


src/server/test_utils.cc
```diff
@@ -21,12 +21,15 @@ extern "C" {
 #include "base/flags.h"
 #include "base/logging.h"
 #include "base/stl_util.h"
+#include "io/file_util.h"
+#include "util/fibers/pool.h"
+
+//
```
BorysTheDev, line 27: delete

BorysTheDev (approved):

PR #7537, opened 2026-06-08
feat: use more SIMD ops in OAHSet

BorysTheDev:
Summary: This PR increases SIMD usage and adjusts growth policies in OAHSet/OAHEntry to improve probe efficiency.

Changes:

Refactors displacement-window probing to sweep the window in SIMD strides and pack candidate/empty masks via a new ProbeWindow helper.

- Changes kShiftLog (and thus kDisplacementSize) to a larger window and adds compile-time checks that the window fits in a 32-bit mask.
- Updates extension-vector probing assumptions from “power-of-two capacity” to “even capacity”, and changes PtrVector growth to +2 elements up to a linear threshold, then doubling.
- Replaces vector “defrag” via ResizeLog with a same-size Realloc path.
- Adjusts AddImpl growth trigger to a higher load threshold and updates related unit test expectations.
- Refactors GetRandomMember into SIMD-assisted scanning helpers (ScanRange/PickFromBucket) and removes Pop APIs from OAHEntry/OAHSet.
- Moves Rehash and Shrink definitions out-of-line into oah_set.cc and updates tests for the new behavior.

### DIFF
--- a/src/core/oah_entry.cc
+++ b/src/core/oah_entry.cc
@@ -98,13 +98,10 @@ ssize_t OAHEntry::ReallocIfNeeded(PageUsage* page_usage, bool* realloced) {
           *realloced = true;
       }
     }
-    // Then defrag the vector's array buffer if its page is underutilized. ResizeLog
-    // with the same log_size allocates a fresh buffer, move-constructs each element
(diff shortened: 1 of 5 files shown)


vyavdoshenko (approved):
LGTM.
Consider replacing asserts with DCHECKs.

PR #7543, opened 2026-06-08
fix(tiering): Avoid defrag spikes + fix negative memory budget bug

dranikpg:
1. Fix calculation bug where `db_size_->memory_budget()` returned **s**size_t, but it was cast to size_t and caused an overflow, stopping all offloading
2. When non-owned slots are flushed, a huge amount of DELs is issued which lead to a huge amount of fragmented bins, all of which are enqueued immediately for defragmentation, causing a memory spike. Limit defragmentation to a positive upload budget and at most 100 concurrent defrag operations. Otherwise, the offset is added to a delayed_defrag queue, which is processed later during offloading

### DIFF
--- a/src/server/tiered_storage.cc
+++ b/src/server/tiered_storage.cc
@@ -648,9 +648,9 @@ std::vector<std::string> TieredStorage::GetMutableFlagNames() {
 bool TieredStorage::ShouldOffload() const {
   // Cool values can be dropped, so count as free as well
   int64_t actual_free = op_manager_->db_slice_.memory_budget() + int64_t(CoolMemoryUsage());
-  int64_t expected_free = double(config_.offload_threshold) *
-                          max_memory_limit.load(memory_order_relaxed) / shard_set->size();
-  return actual_free < expected_free;
+  int64_t target_free = double(config_.offload_threshold) *
+                        max_memory_limit.load(memory_order_relaxed) / shard_set->size();
+  return actual_free < target_free;
 }
 
 int64_t TieredStorage::UploadBudget() const {
--- a/src/server/tiered_storage.cc
+++ b/src/server/tiered_storage.cc
@@ -441,11 +441,16 @@ bool TieredStorage::ShardOpManager::NotifyDelete(tiering::DiskSegment segment) {
 }
 
 void TieredStorage::ShardOpManager::EnqueueForDefrag(tiering::DiskSegment segment) {
-  // Trigger read to signal need for defragmentation. NotifyFetched will handle it.
+  // Trigger read to signal need for defragmentation. NotifyFetched will handle it on success.
   DVLOG(2) << "Enqueueing bin defragmentation for: " << segment.offset;
   stats_.pending_defrags++;
   Enqueue(
-      kFragmentedBin, segment, tiering::BareDecoder{}, [](auto) {}, true);
+      kFragmentedBin, segment, tiering::BareDecoder{},
+      [this](io::Result<tiering::Decoder*> res) {
+        if (!res)
+          stats_.pending_defrags--;
+      },
+      true);
 }
 
 void TieredStorage::ShardOpManager::RetireColdEntries(size_t additional_memory) {
--- a/src/server/tiered_storage.cc
+++ b/src/server/tiered_storage.cc
@@ -318,7 +318,7 @@ class TieredStorage::ShardOpManager : public tiering::OpManager {
   } stats_;
 
   // When we don't have memory to upload a page for defragmentation, we save it here to do it later
-  std::deque<uint32_t /* offset */> delayed_defrag_queue_;
+  std::deque<uint32_t /* page index (kPageSize) */> delayed_defrag_queue_;
 
   TieredStorage* ts_;
   DbSlice& db_slice_;
@@ -366,7 +366,6 @@ bool TieredStorage::ShardOpManager::NotifyFetched(const OwnedEntryId& id,
     if (*i == kFragmentedBin) {  // Generally we read whole bins only for defrag
       auto* bdecoder = static_cast<tiering::BareDecoder*>(decoder);
       Defragment(segment, bdecoder->slice);
-      stats_.pending_defrags--;
       return true;  // delete
     }
   }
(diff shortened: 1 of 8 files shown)


src/server/tiered_storage.cc
```diff
@@ -636,6 +665,9 @@ void TieredStorage::RunOffloading(DbIndex dbid) {
 
   const auto start_cycles = base::CycleClock::Now();
 
+  // Takes up a small fixed amount of time and is best done before offloading (to be picked up)
+  ProcessDelayedDeframents();
```
dranikpg, line 670: Maybe this should not be called from RunOffloading, but any other scheduled routine

dranikpg:
augment review

dranikpg:
augment review

dranikpg:
augment review

mkaruza (commented):
What is behavior if:

1. we enqueue DELETE for page X. 
2. Than we have DELETES (2-100 command ) for different pages (not page X as first delete)
3. We reach hard limit of 100 pending_fragments
4. DELETE arrives for same PAGE X.

Can we maybe batch it this if we see that page X read is already in flight ? 
Should we check for existing pending reads before enforcing the limit?

dranikpg:
> Can we maybe pass this if we see that page X read is already in flight ?

I thought about this as an optimization that we didn't have earlier and don't have here: there is no point in defrag if the page will be deleted soon. But I couldn't invent a simple and yet reliable metric for that. 

With small bins, we never get delete requests for the whole page. We either defrag if used% < 50 or delete if used% == 0. The defrag can happen (and make sense) after a few deletes, there can be multiple defrags piled up. What is more, if there are multipe pending operations for a page, the defrag is ordered to be last. 

So we must assume, that because 60% of the page was deleted now for example, another 40% are likely to be deleted in the future? And what if not?

dranikpg:
Either way defrag is free if there is already a read for the page pending, as we always read the full paeg

mkaruza (dismissed):

src/server/tiered_storage.cc
```diff
@@ -415,16 +425,30 @@ bool TieredStorage::ShardOpManager::NotifyDelete(tiering::DiskSegment segment) {
...
   }
 
+  // If we have memory, upload the page for defrag. It will be reshuffled and offloaded more packed.
+  // Otherwise, enqueue the bin to be defragmented later when memory is available
   if (bin.fragmented) {
-    // Trigger read to signal need for defragmentation. NotifyFetched will handle it.
-    DVLOG(2) << "Enqueueing bin defragmentation for: " << bin.segment.offset;
-    Enqueue(
-        kFragmentedBin, bin.segment, tiering::BareDecoder{}, [](auto res) {}, true);
+    constexpr size_t kMaxDelayedMem = 1_MB;
```
romange, line 431: I do not understand why try so hard.

1. The page is fragmented but not empty.
2. If we CAN defragment - then lets defragment
3. Otherwise - either we will try again on the next delete, or we have background defragment process that will kick-in. Why do we need `delayed_defrag_queue_` ?

src/server/tiered_storage.cc
```diff
@@ -81,6 +81,10 @@ bool OccupiesWholePages(size_t size) {
 // Stashed bins no longer have bin ids, so this sentinel is used to differentiate from regular reads
 constexpr auto kFragmentedBin = tiering::SmallBins::kInvalidBin - 1;
 
+// Memory budget (UploadBudget) does not account for buffers allocated for in-flight defrag reads,
+// so we cap the number of concurrent defragmentation operations to avoid unbounded memory growth.
+constexpr uint32_t kMaxPendingDefrags = 100;
```
romange, line 86: each page is 4kb so you allow 400KB? I think we can safely raise it to 1000.

src/server/tiered_storage.cc
```diff
@@ -617,16 +646,17 @@ std::vector<std::string> TieredStorage::GetMutableFlagNames() {
 }
 
 bool TieredStorage::ShouldOffload() const {
-  size_t free_memory = op_manager_->db_slice_.memory_budget();
-  size_t per_shard = max_memory_limit.load(memory_order_relaxed) / shard_set->size();
-  // Cool values are already offloadeded, so don't count them as used memory
-  return (free_memory + CoolMemoryUsage()) < config_.offload_threshold * per_shard;
+  // Cool values can be dropped, so count as free as well
+  int64_t actual_free = op_manager_->db_slice_.memory_budget() + int64_t(CoolMemoryUsage());
+  int64_t expected_free = double(config_.offload_threshold) *
```
romange: The current name expected_free is a bit misleading because "expected" implies a prediction. Maybe `target_free_bytes_per_shard` ?

romange (commented):
a few comments. Lets talk if I am missing something

src/server/tiered_storage.cc
dranikpg: We do not `have background defragment process that will kick-in` - there is no such thing - and this PR adds it.

I wanted to make one, but it is dificult to iterate a absl::flat_hash_map with yields and I did not want to add some other datastructure, so I decided to go with the queue instead

src/server/tiered_storage.cc
dranikpg: 400kb is actually already a lot, this is not kb/s, this is instantaneous kb, so with a latency of ~1ms this is around 400mb/s, but I can raise it more

romange: nm, it's not the point

src/server/tiered_storage.cc
romange: don't we have a background process that goes over all the bin-pages we have and does reverse lookups with Dash::FindFirst ?

dranikpg: No, it runs only when the page is deleted

romange: ok, please add an issue to ssd-tiering project, to make SmallBins traversable (we can use DashTable for that) so we could defragment properly

dranikpg: Added

romange (approved):

PR #7545, opened 2026-06-09
feat(facade): extend ok_main with sharded SET/GET and async dispatch

romange:
Extend the ok_backend binary to demonstrate Dragonfly's shard-per-thread architecture with SET/GET commands and non-blocking async dispatch.

**Changes:**
- Thread-local `absl::flat_hash_map` per proactor thread as shard storage
- Key sharding via XXH64 hash to determine the owning thread
- Async dispatch via `DispatchBrief` + `SuspendedCommand` coroutine path (non-blocking)
- Sync fallback via `AwaitBrief` when caller requires `ONLY_SYNC`
- `DispatchManyCommands` handling for pipeline squashing
- PING and COMMAND stubs for redis-cli compatibility

### DIFF
--- a/src/facade/ok_main.cc
+++ b/src/facade/ok_main.cc
@@ -43,7 +43,7 @@ namespace {
 thread_local absl::flat_hash_map<string, string> shard_db;
 
 // Determine the owning shard for a given key.
-const uint64_t kShardHashSeed = 120577240643ULL;
+constexpr uint64_t kShardHashSeed = 120577240643ULL;
 unsigned KeyShard(string_view key, unsigned num_shards) {
   XXH64_hash_t hash = XXH64(key.data(), key.size(), kShardHashSeed);
   return hash % num_shards;
@@ -57,10 +57,8 @@ struct AsyncCmd {
   // C++20 coroutine contract: every coroutine return type must contain a nested
   // "promise_type" that the compiler uses to control the coroutine lifecycle.
   struct promise_type {
-    // Called once to create the return object that the caller receives.
-    // We stash the coroutine_handle so we can later pass it to ParsedCommand::Resolve().
     AsyncCmd get_return_object() {
-      return AsyncCmd{std::coroutine_handle<promise_type>::from_promise(*this)};
+      return AsyncCmd{};
     }
 
     // Called before the coroutine body runs. suspend_never means "start executing
@@ -86,10 +84,6 @@ struct AsyncCmd {
       LOG(FATAL) << "Unhandled exception in AsyncCmd coroutine";
     }
   };
-
-  // The handle is no longer accessed externally since ResolveAwaiter
-  // registers it internally at the suspension point.
-  std::coroutine_handle<> handle;
 
   // Allow `return SetAsync(...)` in DispatchCommand.
   operator DispatchResult() const noexcept {
@@ -167,6 +161,7 @@ AsyncCmd GetAsync(CmdContext* ctx, ProactorPool* pool) {
 
   co_await ResolveAwaiter{ctx};
 
+  DCHECK(!ctx->mc_command());  // We do not support MC protocol.
   // Resumed by Connection::SendReply() — write directly to reply builder.
   auto* rb = static_cast<RedisReplyBuilder*>(ctx->rb());
   if (ctx->get_result) {
@@ -202,7 +197,7 @@ class OkService : public ServiceInterface {
   ProactorPool* pool_;
 };
 
-DispatchResult OkService::DispatchCommand(ParsedArgs args, ParsedCommand* cmd,
+DispatchResult OkService::DispatchCommand([[maybe_unused]] ParsedArgs args, ParsedCommand* cmd,
                                           AsyncPreference mode) {
   if (cmd->empty()) {
     cmd->rb()->SendError("ERR empty command");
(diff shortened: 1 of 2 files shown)


src/facade/ok_main.cc
```diff
@@ -21,35 +33,97 @@ namespace facade {
...
 
+// Async SET: dispatches write to shard, suspends, then sends OK on resume.
+AsyncCmd SetAsync(CmdContext* ctx, ProactorPool* pool) {
+  string key(ctx->at(1));
+  string value(ctx->at(2));
+  unsigned shard_id = KeyShard(key, pool->size());
+
+  ctx->blocker.Start(1);
+  pool->at(shard_id)->DispatchBrief([k = std::move(key), v = std::move(value), ctx] {
+    shard_db[k] = v;
```
glevkovich: I think you have here redundant copies in the hot path:
You did move key and value into the lambda.
But because k and v still have a name inside the lambda, the are still treated as lvalues.
`shard_db[k] = v;    `
if the key is not there: create on the left side a new copy of k and a default constructed empty string K. Then it allocates a new v and copy it to K . This is because they are lvalues and cant be "stolen". Aftert this expression the map has a key with value v, but k and v were n ever stolen.
Solution, replace with:
`shard_db.insert_or_assign(std::move(k), std::move(v));`

src/facade/ok_main.cc
```diff
@@ -21,52 +33,228 @@ namespace facade {
...
+      return {};
+    }
+
+    // The coroutine has no meaningful return value (replies go to the reply builder).
+    void return_void() {
+    }
+
+    // Required by the standard; we don't expect exceptions in this code.
+    void unhandled_exception() {
+    }
```
glevkovich, line 85: nit:
perhaps terminating here to be on the safe side is better. add:
`std::terminate();`
or
`LOG(FATAL) << "Unexpected exception in async command coroutine";`

src/facade/ok_main.cc
```diff
@@ -21,52 +33,228 @@ namespace facade {
 
 namespace {
 
+// Thread-local shard storage — each proactor thread has its own hash map.
+thread_local absl::flat_hash_map<string, string> shard_db;
+
+// Determine the owning shard for a given key.
+unsigned KeyShard(string_view key, unsigned num_shards) {
+  XXH64_hash_t hash = XXH64(key.data(), key.size(), 120577240643ULL);
```
glevkovich: nit: bare magic number.
consider:
`constexpr uint64_t kHashSeed = 120577240643ULL;`

src/facade/ok_main.cc
```diff
@@ -21,52 +33,228 @@ namespace facade {
...
+
+ private:
+  // Synchronous fallback handlers (used when async is not available).
+  DispatchResult HandleSetSync(ParsedCommand* cmd);
+  DispatchResult HandleGetSync(ParsedCommand* cmd);
+
+  ProactorPool* pool_;
 };
 
+DispatchResult OkService::DispatchCommand(ParsedArgs args, ParsedCommand* cmd,
```
glevkovich: nit: add [[maybe_unused]] to args. It is not used in the function.
What reminds me that I need to finialize adding strict flags from helio into all dragonfly code.

src/facade/ok_main.cc
romange: it is a toy example, not a prod code

src/facade/ok_main.cc
romange: yeah and I think lambda should be mutable too to allow this. will fix.

src/facade/ok_main.cc
romange: fixed

src/facade/ok_main.cc
```diff
@@ -21,34 +39,152 @@ namespace facade {
 
 namespace {
 
+// Thread-local shard storage — each proactor thread has its own hash map.
+thread_local absl::flat_hash_map<string, string> shard_db;
+
+// Determine the owning shard for a given key.
+const uint64_t kShardHashSeed = 120577240643ULL;
```
glevkovich: nit: constexpr

src/facade/ok_main.cc
```diff
@@ -21,34 +39,152 @@ namespace facade {
...
+
+    // Required by the standard; we don't expect exceptions in this code.
+    void unhandled_exception() {
+      LOG(FATAL) << "Unhandled exception in AsyncCmd coroutine";
+    }
+  };
+
+  // The handle is no longer accessed externally since ResolveAwaiter
+  // registers it internally at the suspension point.
+  std::coroutine_handle<> handle;
```
glevkovich: Seem like dead code, I dont see where you use this member. Can be removed.

src/facade/ok_main.cc
```diff
@@ -21,34 +39,152 @@ namespace facade {
...
+    auto it = shard_db.find(k);
+    if (it != shard_db.end())
+      ctx->get_result = it->second;
+    ctx->blocker.Dec();
+  });
+
+  co_await ResolveAwaiter{ctx};
+
+  // Resumed by Connection::SendReply() — write directly to reply builder.
+  auto* rb = static_cast<RedisReplyBuilder*>(ctx->rb());
```
glevkovich, line 166: nit: maybe for safety its better to DCHECK in some common place thatuser can only use RESP with ok_backend.. SendBulkString/SendNull are RESP only.  So no one will mistakenly think it supports MC.

src/facade/ok_main.cc
```diff
@@ -57,15 +193,239 @@ class OkService : public ServiceInterface {
...
+}
+
+DispatchManyResult OkService::DispatchManyCommands(ParsedCommand* head, unsigned count,
+                                                   SinkReplyBuilder* builder,
+                                                   ConnectionContext* cntx) {
+  for (unsigned i = 0; i < count; i++) {
+    ParsedCommand* cmd = head;
+    head = head->next;
+    cmd->Init(builder, cntx);
+    DispatchCommand(ParsedArgs{*cmd}, cmd, AsyncPreference::ONLY_SYNC);
```
glevkovich, line 260: I think it iswWorth commenting  here -  is it a development-stage? A TODO or a short comment would help explain why DispatchManyCommands dispatches all commands synchronously in current code.

glevkovich (dismissed):
Approved with 4 small comments.

src/facade/ok_main.cc
romange: done

kostasrim (approved):

PR #7546, opened 2026-06-09
experiment: celery zstd compression - DO NOT MERGE

abhijat:
sets up experiment to compare celery task memory usage with/without list compression

workload:
```
python churn.py --produce-burst 100000 --produce-pause 45 --payload-items 500 --drain-every 5 --drain-window 30
```

run config for dragonfly:

* without compression: `--list_experimental_zstd_dict_threshold=0`
* with compression: `--list_experimental_zstd_dict_threshold=16384`

observed `used_mem`

image

LHS is without compression
RHS (after 12:07) is with compression


abhijat:
Extending the experiment with sidekiq, which has a similar structure to celery but for ruby, the performance gains are quite similar

image

A ruby script similar to the python one is used

```
bundle exec ruby churn.rb --produce-burst 20000 --produce-pause 5 --payload-items 50 --drain-every 20 --drain-window 8
```

abhijat:
with similar workload as celery

image


```
bundle exec ruby churn.rb --produce-burst 100000 --produce-pause 45 --payload-items 500 --drain-every 5 --drain-window 30
```

abhijat:
image

```
bundle exec ruby churn.rb --produce-burst 100000 --produce-pause 45 --payload-items 500 --drain-every 30 --drain-window 15
```

PR #7548, opened 2026-06-09
server: Simplify Flush - PrepareFlush interaction

abhijat:
After recent changes for tagged chunk the interaction between PrepareFlush and Flush became a little convoluted, basically the two handled different levels of compression (tagged vs untagged). The code is simplified so that Flush handles compression differently based on the tagged chunk setting.

No functional changes.


dranikpg (approved):
👍🏻

PR #7550, opened 2026-06-09
server: Adds metric for connection memory bytes

abhijat:
Adds new metric connection_memory_bytes
Exposed as dragonfly_connection_memory_bytes

The metric is sourced from new thread local field `ConnectionStats::connection_memory_bytes`
Each connection is responsible for updating the tl field at appropriate points. Currently we update the metric on the following spots

* ParseRedis
* ParseMCBatch
* ExecuteBatch
* ReleaseParsedCommand
* DestroyParsedQueue

Some are updated via a new RAII helper and some explicitly.
Connection memory is only added after `IncreaseConnStats` call and before `DecreaseConnStats`
If a connection is used for replication (ie `conn_cntx->master_repl_flow != nullptr`) then it is not tracked in this metric and its current contributions are deducted. There is a separate metric for replication bytes

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -179,6 +179,18 @@ struct ReadBufTracker {
   size_t last_capacity_;
 };
 
+struct ConnectionMemoryTracker {
+  explicit ConnectionMemoryTracker(Connection* conn) : conn_(conn) {
+  }
+
+  ~ConnectionMemoryTracker() {
+    conn_->RefreshConnectionMemoryUsage();
+  }
+
+ private:
+  Connection* conn_;
+};
+
 size_t UsedMemoryInternal(const ParsedCommand& msg) {
   return msg.GetSize() + msg.HeapMemory();
 }
@@ -1430,6 +1442,7 @@ Connection::ParserStatus Connection::ParseRedis(base::IoBuf& io_buf, uint32_t ma
   // TODO(kostas): follow up on this
   size_t total_consumed = 0;
   do {
+    bool stop_parsing = false;
     DCHECK(parsed_cmd_);
     result = redis_parser_->Parse(read_buffer, &consumed, parsed_cmd_);
     request_consumed_bytes_ += consumed;
@@ -1455,7 +1468,7 @@ Connection::ParserStatus Connection::ParseRedis(base::IoBuf& io_buf, uint32_t ma
         if (GetQueueBackpressure().IsPipelineBufferOverLimit(
                 GetLocalConnStats().pipeline_queue_bytes, parsed_cmd_q_len_)) {
           DVLOG(2) << "Pipeline buffer over limit, breaking from parsing loop.";
-          break;
+          stop_parsing = true;
         }
       } else {
         DispatchSingle(has_more, dispatch_sync, dispatch_async);
@@ -1464,8 +1477,13 @@ Connection::ParserStatus Connection::ParseRedis(base::IoBuf& io_buf, uint32_t ma
     if (result != RespSrvParser::OK && result != RespSrvParser::INPUT_PENDING) {
       // We do not expect that a replica sends an invalid command so we log if it happens.
       LOG_IF(WARNING, cntx()->replica_conn)
-          << "Redis parser error: " << result << " during parse: " << io::View(read_buffer);
+          << "Redis parser error: " << static_cast<unsigned int>(result)
+          << " during parse: " << io::View(read_buffer);
     }
+    RefreshConnectionMemoryUsage();
+    if (stop_parsing)
+      break;
+
     read_buffer.remove_prefix(consumed);
 
     // We must yield from time to time to allow other fibers to run.
@@ -1475,7 +1493,7 @@ Connection::ParserStatus Connection::ParseRedis(base::IoBuf& io_buf, uint32_t ma
       GetLocalConnStats().num_read_yields++;
       ThisFiber::Yield();
     }
-  } while (RespSrvParser::OK == result && read_buffer.size() > 0 && !reply_builder_->GetError());
+  } while (RespSrvParser::OK == result && !read_buffer.empty() && !reply_builder_->GetError());
 
   io_buf.ConsumeInput(total_consumed);
 
@@ -2431,8 +2449,36 @@ size_t Connection::GetMemoryUsage() const {
   return mem;
 }
 
+void Connection::RefreshConnectionMemoryUsage() {
+  if (!conn_stats_registered_)
+    return;
+
+  size_t current = account_connection_memory_ ? GetMemoryUsage() : 0;
+  auto& conn_stats = tl_facade_stats->conn_stats;
+
+  if (current >= accounted_connection_memory_bytes_) {
+    conn_stats.connection_memory_bytes += current - accounted_connection_memory_bytes_;
+  } else {
+    const size_t delta = accounted_connection_memory_bytes_ - current;
+    DCHECK_GE(conn_stats.connection_memory_bytes, delta);
+    conn_stats.connection_memory_bytes -= delta;
+  }
+
+  accounted_connection_memory_bytes_ = current;
+}
+
+void Connection::SetConnectionMemoryAccounting(bool enabled) {
+  if (account_connection_memory_ == enabled)
+    return;
+
+  account_connection_memory_ = enabled;
+  // reduce memory to 0 and tl stat contribution to 0 on enabled=false
+  RefreshConnectionMemoryUsage();
+}
+
 void Connection::IncreaseConnStats() {
   DCHECK(tl_facade_stats);
+  DCHECK(!conn_stats_registered_);
   auto& conn_stats = tl_facade_stats->conn_stats;
   if (IsMainOrMemcache())
     ++conn_stats.num_conns_main;
@@ -2449,11 +2495,22 @@ void Connection::IncreaseConnStats() {
     conn_stats.dispatch_queue_subscriber_bytes += dispatch_q_subscriber_bytes_;
     qbp.subscriber_bytes.fetch_add(dispatch_q_subscriber_bytes_, std::memory_order_relaxed);
   }
+
+  conn_stats_registered_ = true;
+  accounted_connection_memory_bytes_ = account_connection_memory_ ? GetMemoryUsage() : 0;
+  conn_stats.connection_memory_bytes += accounted_connection_memory_bytes_;
 }
 
 void Connection::DecreaseConnStats() {
   DCHECK(tl_facade_stats);
+  DCHECK(conn_stats_registered_);
   auto& conn_stats = tl_facade_stats->conn_stats;
+  RefreshConnectionMemoryUsage();
+  DCHECK_GE(conn_stats.connection_memory_bytes, accounted_connection_memory_bytes_);
+  conn_stats.connection_memory_bytes -= accounted_connection_memory_bytes_;
+  accounted_connection_memory_bytes_ = 0;
+  conn_stats_registered_ = false;
+
   if (IsMainOrMemcache()) {
     DCHECK_GT(conn_stats.num_conns_main, 0u);
     --conn_stats.num_conns_main;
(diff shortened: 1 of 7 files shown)


abhijat:
augment review

src/server/server_family.cc
```diff
@@ -1795,6 +1795,9 @@ void PrintPrometheusMetrics(uint64_t uptime, const Metrics& m, DflyCmd* dfly_cmd
                             &resp->body());
   AppendMetricWithoutLabels("memory_used_peak_bytes", "", m.used_mem_peak, MetricType::GAUGE,
                             &resp->body());
+  AppendMetricWithoutLabels("connection_memory_bytes",
```
romange: please move it to `memory_by_class_bytes`

src/server/server_family.cc
```diff
@@ -3165,8 +3174,26 @@ Metrics ServerFamily::GetMetrics(Namespace* ns, bool collect_replication_memory)
     EngineShard* shard = EngineShard::tlocal();
     ServerState* ss = ServerState::tlocal();
 
+    size_t local_connection_memory = 0;
+    if (collect_connection_memory) {
```
romange: it's absolutely unaccepted. I explained several times we can not run `O(n)` computations inside this function.

abhijat: yeah it looks bad to me too, should it be more like some thread local counter style approach which is incremented/decremented in connection life cycle then?

abhijat:
converting to draft briefly to try out counter style approach, which is a bit more invasive but doesn't require doing anything in GetMetrics

abhijat:
augment review

src/server/dflycmd.cc
```diff
@@ -303,6 +303,7 @@ void DflyCmd::Flow(CmdArgList args, CommandContext* cmd_cntx) {
 
     auto& flow = replica_ptr->GetFlow(flow_id);
     conn_cntx->master_repl_flow = &flow;
+    conn_cntx->conn()->SetConnectionMemoryAccounting(false);
     flow.conn = cmd_cntx->conn();
```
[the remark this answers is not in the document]
abhijat: I will change the pr description but the idea is to mirror `memory stats` command which only counts direct bytes where master repl flow is unset.

src/server/dflycmd.cc
```diff
@@ -303,6 +303,7 @@ void DflyCmd::Flow(CmdArgList args, CommandContext* cmd_cntx) {
 
     auto& flow = replica_ptr->GetFlow(flow_id);
     conn_cntx->master_repl_flow = &flow;
+    conn_cntx->conn()->SetConnectionMemoryAccounting(false);
```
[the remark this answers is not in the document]
abhijat: see https://github.com/dragonflydb/dragonfly/pull/7550#discussion_r3395252093

src/facade/dragonfly_connection.cc
```diff
@@ -2416,8 +2434,36 @@ size_t Connection::GetMemoryUsage() const {
   return mem;
 }
 
+void Connection::RefreshConnectionMemoryUsage() {
+  if (!conn_stats_registered_)
+    return;
+
+  size_t current = account_connection_memory_ ? GetMemoryUsage() : 0;
+  auto& conn_stats = tl_facade_stats->conn_stats;
```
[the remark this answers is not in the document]
abhijat: fixed

src/server/dflycmd.cc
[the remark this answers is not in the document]
abhijat: pr description is changed

src/server/server_family.cc
abhijat: I changed and tried to follow similar approach as `client_read_buffer_bytes` is updated, the only tricky part is identifying places to update the counter.

abhijat:
@romange @glevkovich please take a look at the new approach. Now we update the bytes counter during connection lifecycle instead of walking.

src/facade/dragonfly_connection.cc
```diff
@@ -2592,6 +2652,7 @@ bool Connection::ExecuteBatch() {
 
   // Execute sequentially all parsed commands.
   for (auto& cmd = parsed_to_execute_; cmd != nullptr;) {
+    ConnectionMemoryTracker memory_tracker(this);
```
glevkovich: You have ConnectionMemoryTracker here and another one may be constructed in when you call ReleasePipelinedCommand in line 2647.
The 1st one is inside the loop, and the 2nd refresh again dtor.
This happen in multippe places, not only here.
This hurts performance, you refresh way too much for no reason.
2 Suggestions: 
1) Put this tracker outside the loop, you dont need to update memory stats so often, only per batch.
2)  Make the tracker re-entrant, add to connection:
`  unsigned mem_tracking_depth_ = 0;`


```
struct ConnectionMemoryTracker {
  explicit ConnectionMemoryTracker(Connection* conn) : conn_(conn) {
    conn_->BeginMemoryTrackingScope();
  }

  ~ConnectionMemoryTracker() {
    conn_->EndMemoryTrackingScope();
  }

 private:
  Connection* conn_;
};

```
and in connection:
```

void Connection::BeginMemoryTrackingScope() {
  ++mem_tracking_depth_;
}

void Connection::EndMemoryTrackingScope() {
  if (--mem_tracking_depth_ == 0)
    RefreshConnectionMemoryUsage();
}
```
That way when memory tracker is destroyed, it will refresh only for the most-outer one.

src/facade/dragonfly_connection.cc
```diff
@@ -1461,8 +1474,13 @@ Connection::ParserStatus Connection::ParseRedis(base::IoBuf& io_buf, uint32_t ma
     if (result != RespSrvParser::OK && result != RespSrvParser::INPUT_PENDING) {
       // We do not expect that a replica sends an invalid command so we log if it happens.
       LOG_IF(WARNING, cntx()->replica_conn)
-          << "Redis parser error: " << result << " during parse: " << io::View(read_buffer);
+          << "Redis parser error: " << static_cast<unsigned int>(result)
+          << " during parse: " << io::View(read_buffer);
     }
+    RefreshConnectionMemoryUsage();
```
glevkovich, line 1486: Why in the parse loop?
We dont need such resolution in my opinion.
Why not refreshing once per parsing batch?

src/facade/dragonfly_connection.cc
```diff
@@ -2419,8 +2437,36 @@ size_t Connection::GetMemoryUsage() const {
   return mem;
 }
 
+void Connection::RefreshConnectionMemoryUsage() {
+  if (!conn_stats_registered_)
+    return;
+
```
glevkovich, line 2460: consider adding here:
`DCHECK_EQ(socket()->proactor(), ProactorBase::me());
`
to make sure that the thread that currently executes this code also owns this specific thread_local stats.

src/facade/dragonfly_connection.cc
```diff
@@ -2419,8 +2437,36 @@ size_t Connection::GetMemoryUsage() const {
...
+  size_t current = account_connection_memory_ ? GetMemoryUsage() : 0;
+  ConnectionStats& conn_stats = GetLocalConnStats();
+
+  if (current >= accounted_connection_memory_bytes_) {
+    conn_stats.connection_memory_bytes += current - accounted_connection_memory_bytes_;
+  } else {
+    const size_t delta = accounted_connection_memory_bytes_ - current;
+    DCHECK_GE(conn_stats.connection_memory_bytes, delta);
+    conn_stats.connection_memory_bytes -= delta;
```
glevkovich: And what happens in production? you don't have the DCHECK.
Consider:
`conn_stats.connection_memory_bytes -= std::min(conn_stats.connection_memory_bytes, delta);
`
so if delta > conn_stats.connection_memory_bytes because of the min() we will get 0.
Else sinze it is a size_t you will get a huge number which will spike the metrics.

src/facade/dragonfly_connection.cc
```diff
@@ -178,6 +178,18 @@ struct ReadBufTracker {
   size_t last_capacity_;
 };
 
+struct ConnectionMemoryTracker {
```
glevkovich, line 182: I would add here deleters to prevent copies:

```
  ConnectionMemoryTracker(const ConnectionMemoryTracker&) = delete;
  ConnectionMemoryTracker& operator=(const ConnectionMemoryTracker&) = delete;
```

src/server/server_family.cc
```diff
@@ -1792,6 +1792,9 @@ void PrintPrometheusMetrics(uint64_t uptime, const Metrics& m, DflyCmd* dfly_cmd
                             &resp->body());
   AppendMetricWithoutLabels("memory_used_peak_bytes", "", m.used_mem_peak, MetricType::GAUGE,
                             &resp->body());
+  AppendMetricWithoutLabels("connection_memory_bytes",
+                            "Memory used directly by active client connections",
```
glevkovich: I think this is not accurate, because some thing are excluded, or calculated as part of other metrics. 
So I asked the AI (validated it partially, to make sure at least the comment "this isn't accurate" is correct. Up to you to check it yourself if this is valid) , so this should be probably:

```
AppendMetricWithoutLabels(
    "connection_memory_bytes",
    "Approximate memory used directly by each connection: the connection object, "
    "its parsers, reply builder, connection context, the command currently being "
    "parsed, and a fixed ~9KB estimate for the in-use fiber stack. "
    "Excludes the read buffer (io_buf), pipeline queue, and dispatch queue, which are "
    "tracked separately. Replication connections are excluded.",
    conn_stats.connection_memory_bytes, MetricType::GAUGE, &resp->body());
```

src/facade/dragonfly_connection.cc
glevkovich, line 182: I think it makes sense yo add above this class a short overview about it.

src/facade/dragonfly_connection.h
```diff
@@ -560,6 +567,13 @@ class Connection : public util::Connection {
   uint32_t id_;
   Protocol protocol_;
   Phase phase_ = SETUP;
+  // memory is not accounted before this flag is set in IncreaseConnStats
```
glevkovich: nit: I think the 3 new members in this block till line 576 have no clear explanation and use unclear abbreviations.
Each one should clearly explain what is the purpose of the var, for bool when it is set to true and false.

src/facade/dragonfly_connection.cc
abhijat: > This hurts performance, you refresh way too much for no reason.

The refresh is for resolution. I strongly believe the cost of `Connection::GetMemoryUsage` is trivial compared to the loop body itself. I tried to test this through benchmarks but could not get clean reading.

But the high resolution is not strictly required in this case for the metric so I don't have a problem with moving this out of the loop. I will move it

abhijat: moved

src/facade/dragonfly_connection.cc
abhijat: moved

src/facade/dragonfly_connection.cc
abhijat: added

src/facade/dragonfly_connection.cc
abhijat: changed

src/facade/dragonfly_connection.cc
abhijat: I think its pretty much self-explanatory and inline, adding a comment saying its an RAII object to track conn. memory doesn't add much.

src/server/server_family.cc
abhijat: I extended the description, but I didn't put all of those details in there

```
"Approximate direct memory used by active non-replication-flow connections, excluding separately tracked read buffers and queues."
```

src/facade/dragonfly_connection.h
abhijat: added more context

src/facade/dragonfly_connection.cc
abhijat: added

glevkovich (approved):
Everything is fixed, Thanks @abhijat .
approved.

PR #7557, opened 2026-06-09
fix(search): accept FT.AGGREGATE LOAD after SCORER/ADDSCORES

### LINKED ISSUE #7555: FT.AGGREGATE: `LOAD` rejected after `SCORER`/`ADDSCORES` (breaks redisvl hybrid search)
`FT.AGGREGATE` errors `LOAD cannot be applied after projectors or reducers` when `LOAD` comes after `SCORER`/`ADDSCORES`. RediSearch requires exactly that order, so redisvl's hybrid (BM25+vector) queries - e.g. MS Agent Framework `RedisContextProvider` - fail on Dragonfly but work on Redis 8.

Reproduce (redis-cli):
```
FT.CREATE idx ON HASH PREFIX 1 d: SCHEMA content TEXT
HSET d:1 content "python"
FT.AGGREGATE idx "@content:(python)" SCORER BM25STD ADDSCORES LOAD 1 content
```

| | Dragonfly | Redis 8.6 |
|---|---|---|
| `… SCORER ADDSCORES LOAD …` (redisvl/RediSearch order) | `ERR LOAD cannot be applied after projectors or reducers` | OK |
| `… LOAD … SCORER ADDSCORES` (LOAD first) | OK | `Unknown argument SCORER` |

The engines have **opposite** clause-ordering rules; redisvl emits the RediSearch one.

Root cause:
`src/server/search/search_family.cc:837` - once `SCORER`/`ADDSCORES` move parsing into the pipeline stage, a following `LOAD` is rejected. `LOAD` is only accepted before any projector/reducer.

Fix:
Accept `LOAD`/`LOAD_FROM` after `SCORER`/`ADDSCORES` (position-independent, RediSearch
parity), and confirm `ADDSCORES` populates `@__score` for the downstream `APPLY`.

vyavdoshenko:
redisvl hybrid (BM25+vector) queries emit `SCORER ... ADDSCORES ... LOAD ...`, which Dragonfly rejected with "LOAD cannot be applied after projectors or reducers". LOAD was only parsed in two loops at the start of FT.AGGREGATE; any LOAD reaching the main parser was unconditionally rejected.

Changes:
- Parse LOAD/LOAD_FROM in the main loop, gated by a has_pipeline_step flag set only by GROUPBY/SORTBY/LIMIT/FILTER/APPLY.
- LOAD/LOAD_FROM after query-level keywords (SCORER/ADDSCORES/WITHSCORES/PARAMS/DIALECT) is now accepted; LOAD after a real projector/reducer is still rejected.
- ADDSCORES already injects @__score, so a downstream APPLY on @__score works.
- Add regression test AggregateLoadAfterScorer.

Fixes #7555

### DIFF
--- a/src/server/search/search_family.cc
+++ b/src/server/search/search_family.cc
@@ -680,32 +680,41 @@ ParseResult<AggregateParams> ParseAggregatorParams(CmdArgParser* parser) {
   AggregateParams params;
   tie(params.index, params.query) = parser->Next<string_view, string_view>();
 
-  // Parse LOAD count field [field ...]
-  // LOAD options are at the beginning of the query, so we need to parse them first
-  while (parser->HasNext() && parser->Check("LOAD")) {
-    auto fields = ParseLoadOrReturnFields(parser, true);
-    if (!params.load_fields.has_value())
-      params.load_fields = std::move(fields);
-    else
-      params.load_fields->insert(params.load_fields->end(), make_move_iterator(fields.begin()),
-                                 make_move_iterator(fields.end()));
-  }
-
   // Used for join params
   absl::flat_hash_set<std::string> current_known_indexes;
   current_known_indexes.insert(std::string{params.index});
-  while (parser->HasNext() && parser->Check("LOAD_FROM")) {
-    auto join_params = ParseAggregatorJoinParams(parser, &current_known_indexes);
-    if (!join_params) {
-      return make_unexpected(join_params.error());
-    }
-    params.joins.emplace_back(std::move(join_params).value());
-  }
-  const bool joining_enabled = !params.joins.empty();
+
+  // LOAD/LOAD_FROM are rejected only after a projector or reducer step, not after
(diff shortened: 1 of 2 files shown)


src/server/search/search_family.cc
```diff
@@ -680,32 +680,41 @@ ParseResult<AggregateParams> ParseAggregatorParams(CmdArgParser* parser) {
...
+
+    if (parser->Check("LOAD_FROM")) {
+      if (has_pipeline_step)
+        return CreateSyntaxError("LOAD_FROM cannot be applied after projectors or reducers"sv);
+      auto join_params = ParseAggregatorJoinParams(parser, &current_known_indexes);
+      if (!join_params)
+        return make_unexpected(join_params.error());
+      params.joins.emplace_back(std::move(join_params).value());
+      continue;
+    }
```
BorysTheDev, line 713: maybe it's time to rewrite this code with better usage of CmdArgParser

BorysTheDev (approved):

PR #7562, opened 2026-06-10
fix: prevent integer overflow in max_busy_read_cycles_cached

romange:
Change `max_busy_read_cycles_cached` from `uint64_t` to `uint32_t` and initialize to `UINT32_MAX` instead of `1ULL << 32`, which overflows when narrowed to `uint32_t`. Also fix `ParseRedis` parameter type to `uint32_t` for consistency and add a DCHECK to guard against zero.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -515,7 +515,7 @@ ConnectionStats& __attribute__((noinline)) GetLocalConnStats() {
   return tl_facade_stats->conn_stats;
 }
 
-thread_local uint64_t max_busy_read_cycles_cached = 1ULL << 32;
+thread_local uint32_t max_busy_read_cycles_cached = UINT32_MAX;
 thread_local bool always_flush_pipeline_cached = absl::GetFlag(FLAGS_always_flush_pipeline);
 thread_local uint32_t pipeline_squash_limit_cached = absl::GetFlag(FLAGS_pipeline_squash_limit);
 
@@ -1401,8 +1401,10 @@ void Connection::DispatchSingle(bool has_more, absl::FunctionRef<void()> invoke_
   }
 }
 
-Connection::ParserStatus Connection::ParseRedis(base::IoBuf& io_buf, unsigned max_busy_cycles,
+Connection::ParserStatus Connection::ParseRedis(base::IoBuf& io_buf, uint32_t max_busy_cycles,
                                                 bool enqueue_only) {
+  DCHECK_GT(max_busy_cycles, 0u);
+
   uint32_t consumed = 0;
   RespSrvParser::Result result = RespSrvParser::OK;
 
--- a/src/facade/dragonfly_connection.h
+++ b/src/facade/dragonfly_connection.h
@@ -387,7 +387,7 @@ class Connection : public util::Connection {
   // If add is true, stats are incremented, otherwise decremented.
   void UpdateDispatchStats(const MessageHandle& msg, bool add);
 
-  ParserStatus ParseRedis(base::IoBuf& buf, unsigned max_busy_cycles, bool enqueue_only = false);
+  ParserStatus ParseRedis(base::IoBuf& buf, uint32_t max_busy_cycles, bool enqueue_only = false);
 
   void OnBreakCb(int32_t mask);
 


kostasrim (approved):

src/facade/dragonfly_connection.cc
```diff
@@ -1401,8 +1401,10 @@ void Connection::DispatchSingle(bool has_more, absl::FunctionRef<void()> invoke_
   }
 }
 
-Connection::ParserStatus Connection::ParseRedis(base::IoBuf& io_buf, unsigned max_busy_cycles,
+Connection::ParserStatus Connection::ParseRedis(base::IoBuf& io_buf, uint32_t max_busy_cycles,
                                                 bool enqueue_only) {
+  DCHECK_GT(max_busy_cycles, 0u);
```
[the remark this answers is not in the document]
romange: yes, but we should never pass huge values there, so it should be fine.

romange:
## ok_backend before

```
Type         Ops/sec    Avg. Latency     p50 Latency     p99 Latency   p99.9 Latency       KB/sec 
--------------------------------------------------------------------------------------------------
Pings     2341508.55         0.03848         0.03900         0.04700         0.07100     48019.22 
Totals    2341508.55         0.03848         0.03900         0.04700         0.07100     96038.44 
```

## ok_backend after

```
Type         Ops/sec    Avg. Latency     p50 Latency     p99 Latency   p99.9 Latency       KB/sec 
--------------------------------------------------------------------------------------------------
Pings     2784318.61         0.03059         0.03100         0.03900         0.05500     57100.28 
Totals    2784318.61         0.03059         0.03100         0.03900         0.05500    114200.57 
```

src/facade/dragonfly_connection.cc
```diff
@@ -515,7 +515,7 @@ ConnectionStats& __attribute__((noinline)) GetLocalConnStats() {
   return tl_facade_stats->conn_stats;
 }
 
-thread_local uint64_t max_busy_read_cycles_cached = 1ULL << 32;
+thread_local uint32_t max_busy_read_cycles_cached = UINT32_MAX;
```
glevkovich, line 518: I am not i understand  why you keep here the type as uint32_t.
This is the default value set, and will be correct if kept as default and the flag is not fed from the cli , but if user decides to give it from the outside, Then it is set in line  2794 to:
` max_busy_read_cycles_cached = base::CycleClock::FromUsec(GetFlag(FLAGS_max_busy_read_usec));
`
On a 5GHz CPU , if user gives more than 858000 for FLAGS_max_busy_read_usec -> 0.858sec FromUsec will wrap the value. 
Although this is a huge duration to spend in parsing, why not using here uint64_t, in PraseRedis too, and do a native comparison to  ThisFiber::GetRunningTimeCycles() , which also return uint64_t. BTW also FromUsec returns uint64_t, so uint64_t looks like the right type to use.

glevkovich: If I dont set max_busy_read_usec, default is 200 -> everything works fine.
Lets assume, user decids  to set --max_busy_read_usec=1,000,000 (doesnt make sense, but valid).

1) Assuming 5GHz cpu (freq):
FromUsec(1000000) returns  5,000,000,000  // fits inside uint64_t max_busy_read_cycles_cached
 
2) 
Calling ParseRedis (with uint32_t max_busy_cycles cause truncation):
4294967296 (UINT32_MAX) <   5,000,000,000 so:                  
5,000,000,000 % 2^32 = 705,032,704  // silent truncation (value is warped).
705,032,704÷5,000,000,000 -> instrad of waiting 1sec timeout for parsing, we get 141 millisec.

As an alternative if you dont want to change to uint64_4,  you can add another DCHECK in the top of ParseRedis:

```
// Warn in DEBUG if the cycle count was truncated due to uint32_t overflow
max_busy_read_cycles_cached = base::CycleClock::FromUsec(GetFlag(FLAGS_max_busy_read_usec));
DCHECK_EQ(max_busy_read_cycles_cached,
          base::CycleClock::FromUsec(GetFlag(FLAGS_max_busy_read_usec)))
    << "max_busy_read_usec value too large, cycle count truncated";
```

romange: I think most of all our flags if given unreasonable values the system won't work as expected.
This flag is not different from any other such flag.

romange: Here is the more formal description of my expectations:
1. Given reasonable flags the system should work reasonably well. 
2. It should not crash or behave incorrectly given the invalid input or invalid commands.

glevkovich: If that's the expectations, I guess you change is  correct.
I still think that having a warning in Connection::UpdateFromFlags(), won't hurt, to help detecting human errors:

```
  uint64_t raw_busy_cycles = base::CycleClock::FromUsec(GetFlag(FLAGS_max_busy_read_usec));
  LOG_IF(WARNING, raw_busy_cycles > UINT32_MAX)
      << "max_busy_read_usec value too large, cycle count will be truncated";
  max_busy_read_cycles_cached = static_cast<uint32_t>(raw_busy_cycles);
```

PR #7567, opened 2026-06-10
chore: pass CmdArgParser to async command handlers via SetAsyncHandler

romange:
## Summary
- Change `SetAsyncHandler` to accept `(CmdArgParser, CommandContext*)` instead of `(CmdArgList, CommandContext*)`
- The wrapper lambda constructs `CmdArgParser` from `tail_args()`, removing the need for handlers to access `tail_args` or create their own parser
- Migrate all async handlers in `string_family.cc` and `generic_family.cc` to use the passed-in parser directly
- Add `MakeParserFromContext()` helper to avoid incomplete-type issue in `command_registry.h`

### DIFF
--- a/src/server/string_family.cc
+++ b/src/server/string_family.cc
@@ -739,9 +739,8 @@ struct GetReplies {
   RedisReplyBuilder* rb;
 };
 
-cmd::CmdR ExtendGeneric(CmdArgList args, CommandContext* cmd_cntx) {
-  string_view key = ArgS(args, 0);
-  string_view value = ArgS(args, 1);
+cmd::CmdR ExtendGeneric(CmdArgParser parser, CommandContext* cmd_cntx) {
+  auto [key, value] = parser.Next<string_view, string_view>();
   bool prepend = cmd_cntx->cid()->name().starts_with('P');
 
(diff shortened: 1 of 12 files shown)


src/server/string_family.cc
```diff
@@ -1120,7 +1120,7 @@ std::variant<SetCmd::SetParams, facade::ErrorReply, NegativeExpire> ParseSetPara
 }
 
 cmd::CmdR CmdSet(CmdArgList args, CommandContext* cmd_cntx) {
-  facade::CmdArgParser parser{args};
+  facade::CmdArgParser parser{cmd_cntx->tail_args()};
```
dranikpg: I have a better suggestion - in the async commands, they are all called from a wrapper. Just make the signature to be `(CmdArgParser, CommandContext*)` and create the CmdArgParser instead of using custom accessors like tail_args and ignoring the argument, etc

dranikpg (dismissed):
LGTM, but let's do async command the clean way from the start

src/server/string_family.cc
romange: Great suggestion! Will do.

dranikpg (approved):

PR #7569, opened 2026-06-10
chore: retire --rename_command and --oom_deny_commands flags

romange:
## Summary
- Retire `--rename_command` and `--oom_deny_commands` flags via `ABSL_RETIRED_FLAG`
- Remove all associated runtime logic (command renaming, oom deny set, `RenamedOrOriginal`)
- Update fuzz scripts to use `--restricted_commands` instead

### DIFF
--- a/fuzz/run_fuzzer.sh
+++ b/fuzz/run_fuzzer.sh
@@ -120,10 +120,7 @@ write_repro_env() {
         echo "--proactor_threads=${AFL_PROACTOR_THREADS}"
         echo "--dbfilename=${DB_FILENAME}"
         echo "--omit_basic_usage"
-        echo "--restricted_commands=SHUTDOWN"
-        echo "--restricted_commands=DEBUG"
-        echo "--restricted_commands=FLUSHALL"
-        echo "--restricted_commands=FLUSHDB"
+        echo "--restricted_commands=SHUTDOWN,DEBUG,FLUSHALL,FLUSHDB"
         echo "--max_bulk_len=1048576"
         [[ "$TARGET" == "memcache" ]] && echo "--memcached_port=11211"
     } > "$out"
@@ -158,10 +155,7 @@ run_fuzzer() {
         --bind=::
         --dbfilename="${DB_FILENAME}"
         --omit_basic_usage
-        --restricted_commands=SHUTDOWN
-        --restricted_commands=DEBUG
-        --restricted_commands=FLUSHALL
-        --restricted_commands=FLUSHDB
+        --restricted_commands=SHUTDOWN,DEBUG,FLUSHALL,FLUSHDB
         --max_bulk_len=1048576
     )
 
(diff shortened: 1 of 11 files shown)


vyavdoshenko (dismissed):

fuzz/run_fuzzer.sh
```diff
@@ -120,10 +120,10 @@ write_repro_env() {
...
         echo "--dbfilename=${DB_FILENAME}"
         echo "--omit_basic_usage"
-        echo "--rename_command=SHUTDOWN="
-        echo "--rename_command=DEBUG="
-        echo "--rename_command=FLUSHALL="
-        echo "--rename_command=FLUSHDB="
+        echo "--restricted_commands=SHUTDOWN"
+        echo "--restricted_commands=DEBUG"
+        echo "--restricted_commands=FLUSHALL"
+        echo "--restricted_commands=FLUSHDB"
```
abhijat: I think this is incorrect, it only restricts `flushdb`, all of them have to be in the same arg as comma separated, eg

```
./build-dbg/dragonfly --proactor_threads=1 --logtostderr --restricted_commands=flushall --restricted_commands=flushdb
```

```
arc -3
127.0.0.1:6379> flushdb
(error) ERR Cannot execute restricted command (admin only)
127.0.0.1:6379> flushall
OK
127.0.0.1:6379>
```

so it has to be like

```
--restricted_commands=SHUTDOWN,DEBUG,FLUSHALL,FLUSHDB
```

src/server/command_registry.cc
```diff
@@ -21,14 +21,15 @@
 #include "server/acl/acl_commands_def.h"
 
 using namespace std;
-ABSL_FLAG(vector<string>, rename_command, {},
-          "Change the name of commands, format is: <cmd1_name>=<cmd1_new_name>, "
-          "<cmd2_name>=<cmd2_new_name>");
+
+ABSL_RETIRED_FLAG(std::vector<std::string>, rename_command, {},
+                  "DEPRECATED. Use ACL rules to restrict command access instead.");
+
```
abhijat: Is it safe to silently ignore renamed commands? It might have some surprising results on deployments, e.g. some command renamed to '' to blacklist/restrict is now accepted?

romange: we do not use them in prod and never did. it was done before we had restricted commands.

romange: but it's a good point, we can check if this flag is set and exit.

romange: actually the same "Exit" behavior will be achieved by removing the flag entirely.

romange: removed both flags. we implemented them but never used them.

fuzz/run_fuzzer.sh
vyavdoshenko: @romange 
Please fix it. @abhijat is right.

vyavdoshenko (changes requested):

abhijat (dismissed):
looks like fuzzer scripts are not fixed yet, maybe they can be fixed in follow up

vyavdoshenko (approved):

Issue #7566, opened 2026-06-10
Eliminate `arg_slice_backing` by making command dispatch work directly with `BackedArguments`


romange:
## Problem
Our interfaces are very tight. This requires lots of arguments translations and they sometimes are combersome and confusing.

`ArgSlice` is too rigid as a type even though it's correct as concept - we want a range arguments. 
Unfortunately we can not guarantee that the range of type string_view or slice<char> as sometimes we keep it implicitly.

Current invariant: BackingArguments is alive as long a command runs. 
Therefore, instead of slice, we need `BackingArguments` with an offset to the beginning argument. 
Approximate high-level plan:

Every command dispatch materializes a `vector<string_view>` from `BackedArguments` just to produce an `ArgSlice` (`Span<const string_view>`). This happens in three paths:

- **Normal dispatch**: `arg_slice_backing` in `CommandContext` (`conn_context.h`)
- **MULTI/EXEC replay**: local `CmdArgVec` scratch in `Exec()` and `MultiCommandSquasher`
- **`StoredCmd::Slice()`**: same materialization pattern

`BackedArguments` already owns the string data and supports `at(i)` and `size()`. The materialization exists only because the downstream API (`CommandId::Invoke`, `Transaction::InitByArgs`, `CmdArgParser`, etc.) requires `ArgSlice`.

Lua scripting is unaffected — it naturally produces `Span<string_view>` from the Lua stack and continues using `ArgSlice` directly.

### Phase 0: Decouple `CmdArgParser` from `ArgSlice` ✅

**PR**: #7565

Replaced `CmdArgParser::Tail()` (returned `ArgSlice`) with `UnparsedStart()` (returns index).
Removed `GetCurrentIndex()` (redundant). Callers that need a slice use
`args.subspan(parser.UnparsedStart())` instead — they already have `args` in scope.

This removes the last method on `CmdArgParser` that returned an `ArgSlice`, making it
possible to add a `BackedArguments`-based constructor without changing any callers.

### Phase 1: Make `CmdArgParser` work with `BackedArguments`

Add a second constructor to `CmdArgParser`:

```cpp
CmdArgParser(const BackedArguments& bargs, uint32_t offset = 0);
```

Internally the parser only uses two primitives: `SafeSV(i)` (random access) and `args_.size()` (bounds). Both map trivially to `BackedArguments::at(offset + i)` and `BackedArguments::size() - offset`.

[...]

PR #7571, opened 2026-06-11
chore: use BackedArguments in DEBUG POPULATE

romange:
## Summary
- Refactor `GeneratePopulateCommand` to populate a `BackedArguments` output parameter instead of returning `InlinedVector<string, 5>`
- Extract `invoke()` helper in `DoPopulateBatch` to reduce duplication between the main command and expire paths
- Eliminates the intermediate `args_view` copy loop

Prep work for #7566 — moves DEBUG POPULATE off the `InlinedVector<string> → string_view[]` pattern and onto `BackedArguments`.

## Test plan
- [x] All 76 `generic_family_test` pass (covers DEBUG POPULATE)
- [x] All 45 `string_family_test` pass

### DIFF
--- a/src/server/debugcmd.cc
+++ b/src/server/debugcmd.cc
@@ -152,11 +152,7 @@ const CommandId* GeneratePopulateCommand(string_view type, std::string key, size
     }
   }
 
-  absl::InlinedVector<string_view, 5> views;
-  for (const auto& arg : args) {
-    views.push_back(arg);
-  }
-  out->Assign(views.begin(), views.end(), views.size());
+  out->Assign(args.begin(), args.end(), args.size());
   return cid;
 }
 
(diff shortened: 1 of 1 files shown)


src/server/debugcmd.cc
```diff
@@ -151,7 +152,12 @@ tuple<const CommandId*, absl::InlinedVector<string, 5>> GeneratePopulateCommand(
     }
   }
 
-  return {cid, args};
+  absl::InlinedVector<string_view, 5> views;
+  for (const auto& arg : args) {
+    views.push_back(arg);
+  }
+  out->Assign(views.begin(), views.end(), views.size());
```
kostasrim: Why

```
 out->Assign(args.begin(), args.end(), args.size());  
```

Can't work ? 

I am asking because I thought you dangled here (assign is a deep copy so it's fine)

kostasrim (dismissed):

src/server/debugcmd.cc
kostasrim: views seems redundant: `args -> views -> assign` instead of `args -> assign`

romange: good catch!

kostasrim (approved):

PR #7572, opened 2026-06-11
chore: pass BackedArguments through Lua redis.call path

romange:
## Summary
- Store Lua command arguments in `Interpreter::backed_args_` instead of passing `SliceSpan` through `CallArgs`
- Update `CallFromScript` in `main_service.cc` to use pointer-based `BackedArguments` access

Part of the incremental migration away from `arg_slice_backing`. After this + #7571 (DEBUG POPULATE), the `Slice` variant in `ParsedArgs` can be removed.

### DIFF
--- a/src/server/main_service.cc
+++ b/src/server/main_service.cc
@@ -1289,8 +1289,6 @@ optional<ErrorReply> CheckKeysDeclared(const ConnectionState::ScriptInfo& eval_i
   if (!key_index_res)
     return ErrorReply{key_index_res.status()};
 
-  // TODO: Switch to transaction internal locked keys once single hop multi transactions are merged
-  // const auto& locked_keys = trans->GetMultiKeys();
   const auto& locked_tags = eval_info.lock_tags;
   for (string_view key : key_index_res->Range(args)) {
     if (!locked_tags.contains(LockTag{key})) {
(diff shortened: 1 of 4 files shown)


dranikpg (dismissed):

src/server/main_service.cc
```diff
@@ -2015,22 +2013,34 @@ void Service::CallFromScript(Interpreter::CallArgs& ca, CommandContext* cmd_cntx
...
+      saved_backing.swap(cmd_cntx->arg_slice_backing);
+
       auto* prev = cmd_cntx->SwapReplier(&replier);
-      DispatchCommand(ParsedArgs{ca.args}, cmd_cntx, AsyncPreference::ONLY_SYNC);
+      DispatchCommand(ParsedArgs{*ca.args}, cmd_cntx, AsyncPreference::ONLY_SYNC);
       cmd_cntx->SwapReplier(prev);
+
+      saved_backing.swap(cmd_cntx->arg_slice_backing);
+      cmd_cntx->SetTailArgs(saved_tail);
+    }
```
dranikpg, line 2043: I assume this works fine with async calls

dranikpg (approved):

src/server/main_service.cc
romange: well if we have tests. This should work. I hope we do :)

PR #7581, opened 2026-06-12
fix: eliminate false sharing in RecordLatency and fix data race in ok_main

romange:
## Summary
- Align per-thread `CmdCallStats` to 64-byte cache lines via `alignas(64) StatsCell` wrapper to eliminate false sharing in `RecordLatency`
- Protect `FacadeStats` aggregation in `ok_main` `HandleMetrics` with `fb2::Mutex` — `AwaitFiberOnAll` runs concurrently across proactor threads
- Disable mimalloc stats (`MI_STAT=0`) to reduce overhead

1. command_stats_[tid] is CmdCallStats = 16 bytes, so four thread slots share one 64-byte cacheline. Per-thread writes still invalidate each other.

2. `_mi_stat_counter_increase, stats.c:49` - Allocator stats counters are also contended/shared.  I enabled them a year ago but we did not find much help from them so disabling them by default. We can still enable them for custom builds if needed.

### DIFF
--- a/src/external_libs.cmake
+++ b/src/external_libs.cmake
@@ -61,7 +61,7 @@ add_third_party(
 set(MIMALLOC_ROOT_DIR ${THIRD_PARTY_LIB_DIR}/mimalloc2)
 set(MIMALLOC_INCLUDE_DIR ${MIMALLOC_ROOT_DIR}/include)
 set(MIMALLOC_PATCH_DIR ${CMAKE_CURRENT_LIST_DIR}/../patches/mimalloc-v2.2.4)
-set(MIMALLOC_C_FLAGS "-O3 -g -DMI_STAT=1 -DNDEBUG")
+set(MIMALLOC_C_FLAGS "-O3 -g -DMI_STAT=0 -DNDEBUG")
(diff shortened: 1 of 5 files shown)


src/external_libs.cmake
```diff
@@ -61,7 +61,7 @@ add_third_party(
 set(MIMALLOC_ROOT_DIR ${THIRD_PARTY_LIB_DIR}/mimalloc2)
 set(MIMALLOC_INCLUDE_DIR ${MIMALLOC_ROOT_DIR}/include)
 set(MIMALLOC_PATCH_DIR ${CMAKE_CURRENT_LIST_DIR}/../patches/mimalloc-v2.2.4)
-set(MIMALLOC_C_FLAGS "-O3 -g -DMI_STAT=1 -DNDEBUG")
+set(MIMALLOC_C_FLAGS "-O3 -g -DMI_STAT=0 -DNDEBUG")
```
kostasrim, line 64: This broke the tests @romange

kostasrim (approved):

abhijat (approved):

Issue #7585, opened 2026-06-12
Consolidate pipeline squashing onto DispatchSquashedBatch and remove DispatchManyCommands


romange:
## Background

The V1 pipeline-squashing path (`Connection::SquashPipeline`) currently goes through
`Service::DispatchManyCommands`, which relies on `MultiCommandSquasher`'s reply-reordering
machinery (`order_` vector + `CapturingReplyBuilder` replay on the coordinator). Since
`ParsedCommand` now carries its own deferred `reply_` field (and the generator-based
`MultiCommandSquasher` / `CmdRef` landed in d392b993), the pipeline path can route replies
directly into each `ParsedCommand` and let the connection send them in linked-list order —
removing the need for a second, parallel batch-dispatch entry point.

This issue tracks consolidating the two batch-dispatch paths onto a single
`DispatchSquashedBatch` and removing the now-redundant `DispatchManyCommands`.

The V2 dispatch flow (`enable_resp_io_loop_v2=true`) already has its own squashing and is
out of scope here.

## Goal

- Teach `MultiCommandSquasher` to optionally route a command's reply into its `ParsedCommand`
  (deferred reply) instead of replaying via `order_`.
- Add `DispatchSquashedBatch` and switch `Connection::SquashPipeline` to it.
- Remove `DispatchManyCommands` entirely once it is dead.

## Plan (3 incremental PRs)

Each PR builds and passes tests on its own; later PRs depend on earlier ones being merged.

### PR 1 — Squasher pipeline plumbing (no behavior change)
- Add `ParsedCommand::Resolve(payload::Payload&&)` overload.
- Fix `CapturingReplyBuilder::SendDirect` to `Capture()` into an open collection (strict
  superset; safe for existing callers).
- Add `CmdRef::cmd_cntx` field.
- Route replies in `MultiCommandSquasher` based on `cmd_cntx`: skip the `order_` replay in
  pipeline mode and capture deferred replies in `ExecuteStandalone`.
- No existing caller sets `cmd_cntx`, so all current flows are unchanged.
- Files: `parsed_command.{h,cc}`, `reply_capture.cc`, `conn_context.h`,
  `multi_command_squasher.cc`.

### PR 2 — Wire up V1 pipeline squashing
- Add `Service::DispatchSquashedBatch` (leaves `DispatchManyCommands` in place, now dead).
- Rewrite `Connection::SquashPipeline` to call it: deferred replies sent in linked-list order,
[...]

PR #7587, opened 2026-06-12
Add pipeline-reply plumbing to MultiCommandSquasher

romange:
First of three incremental PRs for #7585.

Adds the capability for `MultiCommandSquasher` to route a command's reply into its
`ParsedCommand` (deferred reply) instead of replaying via the `order_` vector. This is
pure plumbing — **no behavior change**: the capability is gated by an explicit
`Opts::pipeline_mode` flag that no caller enables on this branch (`DispatchSquashedBatch`
turns it on in PR 2), so every current flow is unchanged.

## Changes
- Add `ParsedCommand::Resolve(payload::Payload&&)` overload (guards against monostate).
- Fix `CapturingReplyBuilder::SendDirect` to `Capture()` into an open collection (strict
  superset; safe for existing callers).
- Add `CmdRef::cmd_cntx` field and an explicit `MultiCommandSquasher::Opts::pipeline_mode`
  flag. In pipeline mode, replies are deferred into each command's `ParsedCommand` and the
  `order_` replay is skipped; `cmd_cntx` is `DCHECK`-guaranteed present.
- Fix a pre-existing reply-size accounting bug: `ShardExecInfo::reply_size_delta` was not
  reset between flushes, so a squash spanning multiple flushes (a batch exceeding
  `max_squashed_cmd_num`) double-counted and underflowed `squashing_current_reply_size`.

## Tests
- `RedisReplyBuilderTest.CaptureApplyIntoOpenCollection` — covers the `SendDirect` fix
  (applying a captured payload into a builder mid-collection).
- `MultiTest.SquashReplySizeAccounting` — covers the multi-flush reply-size accounting fix.

The pipeline-squashing wire-up (PR 2) and `DispatchManyCommands` removal (PR 3) follow.

Part of #7585

### DIFF
--- a/src/server/multi_command_squasher.cc
+++ b/src/server/multi_command_squasher.cc
@@ -163,9 +163,26 @@ bool MultiCommandSquasher::ExecuteStandalone(RedisReplyBuilder* rb, CmdRef cmd) 
 
   auto args = cmd.Slice(&tmp_keylist_);
 
+  // In pipeline mode the reply is captured and deferred into the parsed command, preserving
+  // the reply order with squashed commands whose replies are sent later by the connection.
+  optional<CapturingReplyBuilder> crb;
+  if (opts_.pipeline_mode) {
+    DCHECK(cmd.cmd_cntx);
+    cmd.cmd_cntx->SetDeferredReply();
+    DCHECK(cmd.reply_mode == ReplyMode::FULL);
+    crb.emplace(ReplyMode::FULL, rb->GetRespVersion());
+    rb = &*crb;
+  }
+
+  auto resolve = [&] {
+    if (crb)
+      cmd.cmd_cntx->Resolve(crb->Take());
+  };
+
   if (opts_.verify_commands) {
     if (auto err = service_->VerifyCommandState(*cmd.cid, args, *cntx_); err) {
       rb->SendError(std::move(*err));
+      resolve();
       return !opts_.error_abort;
     }
   }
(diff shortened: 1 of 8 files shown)


romange:
augment review

src/server/multi_command_squasher.cc
```diff
@@ -163,9 +163,26 @@ bool MultiCommandSquasher::ExecuteStandalone(RedisReplyBuilder* rb, CmdRef cmd)
...
+  // In pipeline mode the reply is captured and deferred into the parsed command, preserving
+  // the reply order with squashed commands whose replies are sent later by the connection.
+  optional<CapturingReplyBuilder> crb;
+  if (opts_.pipeline_mode) {
+    DCHECK(cmd.cmd_cntx);
+    cmd.cmd_cntx->SetDeferredReply();
+    DCHECK(cmd.reply_mode == ReplyMode::FULL);
+    crb.emplace(ReplyMode::FULL, rb->GetRespVersion());
+    rb = &*crb;
+  }
```
dranikpg, line 175: wow, this is tricky and took me some minutes to understand that cmd.cmd_cntx and cmd_cntx (the one that is invoked) are _different_

because if you executed on cmd.cmd_cntx it wouldn't work as `SetDeferredReply` would trigger async execution

src/server/multi_command_squasher.cc
```diff
@@ -163,9 +163,26 @@ bool MultiCommandSquasher::ExecuteStandalone(RedisReplyBuilder* rb, CmdRef cmd)
...
+
+  auto resolve = [&] {
+    if (crb)
+      cmd.cmd_cntx->Resolve(crb->Take());
+  };
```
dranikpg, line 180: absl::Cleanup without resolve() calls

src/server/multi_command_squasher.cc
```diff
@@ -252,6 +285,16 @@ bool MultiCommandSquasher::ExecuteSquashed(facade::RedisReplyBuilder* rb) {
...
 
+  // Mark pipeline commands as deferred before dispatching to shards.
+  if (opts_.pipeline_mode) {
+    for (auto& sinfo : sharded_) {
+      for (auto& cmd : sinfo.dispatched) {
+        DCHECK(cmd.cmd.cmd_cntx);
+        cmd.cmd.cmd_cntx->SetDeferredReply();
+      }
+    }
+  }
```
dranikpg, line 296: and probably because of this - it doesn't do anything except setting a flag that is DCHECKd in resolve(payload)

dranikpg (approved):

PR #7588, opened 2026-06-12
Squash V1 pipelines via DispatchSquashedBatch

romange:
Second of three incremental PRs for #7585 (#7587 has merged; this now targets `main`).

Adds `Service::DispatchSquashedBatch` and routes `Connection::SquashPipeline` through it.
Consecutive single-shard pipeline commands are squashed and their replies deferred into each
`ParsedCommand`, then sent by the connection in linked-list order.

## Changes
- `Service::DispatchSquashedBatch` (`main_service.cc`, `main_service.h`, `service_interface.h`):
  accumulate squashable single-shard commands and **stop the batch at the first command that
  can't join it**, letting the connection's regular dispatch path handle it (and the rest of
  the pipeline) with the real reply builder, after the squashed replies are flushed in order.
  Such commands: MULTI/EXEC and the commands queued between them, EVAL, blocking commands,
  QUIT, subscribe/unsubscribe (multiple top-level replies), and unknown commands.
- `Connection::SquashPipeline` (`dragonfly_connection.cc`): send the deferred replies in
  linked-list order; advance the list pointers incrementally to stay consistent across
  `SendReply()` preemption points; fix the post-dispatch flush condition for released commands.
- `ParsedCommand::Resolve(payload)` (`parsed_command.cc`): marks the command deferred itself.
  `MultiCommandSquasher` (`multi_command_squasher.cc`) routes a captured reply to
  `cmd_cntx->Resolve()` when present, otherwise replays it via `order_`.
- `DispatchManyCommands` is left in place (now dead in the connection flow) and removed in the
  follow-up PR (#7589).

## Tests (`connection_test.py`)
- `test_squashed_pipeline_pubsub` — subscribe/unsubscribe in a squashed pipeline.
- `test_issue_6165_squash_invalid_syntax` — added an unknown-command case.
- `test_multi_exec_phantom_connections` — passes again (MULTI/EXEC streams via the real builder).
- `test_pipeline_cache_only_async_squashed_dispatches` — switched to a non-transactional
  pipeline (MULTI/EXEC is no longer squash-dispatched).
- C++ `multi_test`, `dragonfly_test`; full AFL fuzz corpus replay clean.

Part of #7585

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1740,20 +1740,47 @@ void Connection::SquashPipeline() {
 
   uint64_t start = CycleClock::Now();
 
-  // async_dispatch is a guard to prevent concurrent writes into reply_builder_, hence
-  // it must guard the Flush() as well.
+  // async_dispatch guards the whole batch: it gates sync dispatch and reply_builder_ writes
+  // (incl. Flush), and marks the connection as dispatching for DispatchTracker checkpoints.
+  // It must be set before DispatchSquashedBatch, which can preempt on shard hops.
   cc_->async_dispatch = true;
 
-  uint32_t num_dispatched_cmds = service_->DispatchManyCommands(parsed_to_execute_, pipeline_count,
-                                                                reply_builder_.get(), cc_.get());
+  uint32_t squashed =
+      service_->DispatchSquashedBatch(parsed_to_execute_, pipeline_count, cc_.get());
 
-  local_stats_.cmds += num_dispatched_cmds;
+  // Nothing was squashed (the head command can't join a batch, e.g. MULTI/EXEC, EVAL,
+  // subscribe, blocking, or unknown). Hand it off to regular dispatch without flushing or
+  // resetting batch mode here — doing so would break reply aggregation for the rest of the
+  // pipeline (e.g. a pipelined MULTI/EXEC would flush once per command).
+  if (squashed == 0) {
+    cc_->async_dispatch = false;
+    skip_next_squashing_ = true;
+    return;
+  }
+
+  // Send the deferred replies in linked-list order, then release the commands. SendReply() may
+  // preempt, allowing the producer to append new commands, so re-read the next pointer after it
+  // and advance the list incrementally to keep it consistent at any preemption point.
+  for (unsigned i = 0; i < squashed && parsed_head_; ++i) {
+    auto* current = parsed_head_;
+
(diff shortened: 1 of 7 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -1740,56 +1740,55 @@ void Connection::SquashPipeline() {
...
   // it must guard the Flush() as well.
   cc_->async_dispatch = true;
 
-  DispatchManyResult result = service_->DispatchManyCommands(parsed_to_execute_, pipeline_count,
-                                                             reply_builder_.get(), cc_.get());
+  // Send deferred replies in linked-list order, then release commands.
+  // SendReply() may preempt, allowing the producer to append new commands, so the list
+  // pointers are advanced incrementally to keep the list consistent at any preemption point.
+  for (unsigned i = 0; i < squashed && parsed_head_; ++i) {
+    auto* current = parsed_head_;
```
dranikpg, line 1765: DCHECK(current->CanReply())

dranikpg (dismissed):

dranikpg (dismissed):

dranikpg (dismissed):

src/server/main_service.cc
```diff
@@ -1796,6 +1796,97 @@ uint32_t Service::DispatchManyCommands(facade::ParsedCommand* head, unsigned cou
...
+    // far are flushed in order. Commands that stop the batch:
+    //  - unknown commands (cid == nullptr): dispatched standalone to produce their error reply;
+    //  - MULTI/EXEC and the commands queued between them: sequential, stored in ExecInfo;
+    //  - EVAL: scripts may require a stricter multi mode than the non-atomic squashing tx;
+    //  - blocking commands: prior replies must be flushed before the fiber blocks;
+    //  - QUIT: closes the reply builder, dropping any deferred replies not yet sent;
+    //  - subscribe/unsubscribe: emit one reply per channel (multiple top-level replies), which the
+    //    CapturingReplyBuilder backing deferred replies cannot represent.
+    if (cid == nullptr)
+      break;
```
dranikpg, line 1864: a little misleading as we don't know what commands will be actually squashed, we just know that those commands mess with the connection state, so can't be executed from the squasher

dranikpg (approved):

src/server/main_service.cc
romange: I will fix in the next PR

PR #7592, opened 2026-06-14
chore: retire squash_stats_latency_lower_limit and simplify dispatch stats

romange:
## Summary
Retire the `squash_stats_latency_lower_limit` flag in favor of the existing `pipeline_latency_seconds` histogram (which already exposes p95/p99). Squash and pipeline latency stats are now counted unconditionally, eliminating the confusing V1/V2 `account_in_stats` split, and the `ServiceInterface`↔facade boundary is simplified.

## Changes
- `squash_stats_latency_lower_limit` → `ABSL_RETIRED_FLAG`; remove the cached thread-local, the `squash_stats_ignored` counter, and the `cmd_squash_stats_ignored_total` metric.
- Drop `DispatchManyResult`; `DispatchManyCommands` now returns `uint32_t` (processed count).
- Squash stats accumulate unconditionally in `Service::DispatchManyCommands`.
- Split `ReleaseParsedCommand(cmd, is_pipelined)` into `ReleaseParsedCommand` (no stats) and `ReleasePipelinedCommand` (records pipeline stats), preserving each call site's exact predicate so accounting is unchanged.

## Notes
- Removes the `cmd_squash_stats_ignored_total` Prometheus metric — dashboards referencing it should be updated.
- Built `dragonfly`; `multi_test` passes (59 passed, 2 pre-existing skips).

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1749,24 +1749,23 @@ void Connection::SquashPipeline() {
 
   local_stats_.cmds += num_dispatched_cmds;
   last_interaction_ = time(nullptr);
-  uint64_t flush_start_cycle_cnt = CycleClock::Now();
-  //
+
   // TODO: to investigate if always flushing will improve P99 latency because otherwise we
   // wait for the next batch to finish before fully flushing the current response.
   if (parsed_cmd_q_len_ == pipeline_count ||
       always_flush_pipeline_cached) {  // Flush if no new commands appeared
+    uint64_t flush_start_cycle = CycleClock::Now();
     reply_builder_->Flush();
+    conn_stats.pipeline_dispatch_flush_count++;
     reply_builder_->SetBatchMode(false);  // in case the next dispatch is sync
-  } else {
-    conn_stats.skip_pipeline_flushing++;
+    conn_stats.pipeline_dispatch_flush_usec +=
+        CycleClock::ToUsec(CycleClock::Now() - flush_start_cycle);
   }
 
   cc_->async_dispatch = false;
 
   conn_stats.pipeline_dispatch_calls++;
   conn_stats.pipeline_dispatch_commands += num_dispatched_cmds;
-  conn_stats.pipeline_dispatch_flush_usec +=
-      CycleClock::ToUsec(CycleClock::Now() - flush_start_cycle_cnt);
 
   for (size_t i = 0; (i < num_dispatched_cmds) && parsed_head_; ++i) {
     auto* next = parsed_head_->next;
@@ -2554,12 +2553,20 @@ bool Connection::ExecuteBatch() {
   // Invariant: batched_ must be false on entry.
   // Both ReplyBatch() and ExecuteBatch() reset it via absl::Cleanup guards on all return paths.
   DCHECK(!reply_builder_->IsBatchMode());
+
+  if (parsed_to_execute_ == nullptr) {
+    return true;  // no errors.
+  }
+
   absl::Cleanup batch_guard = [this] { reply_builder_->SetBatchMode(false); };
   auto& conn_stats = tl_facade_stats->conn_stats;
-  auto advance_head = [this]() -> ParsedCommand* {
+
+  bool is_true_pipeline = (parsed_to_execute_->next) != nullptr;
+
+  auto advance_head = [&] {
     auto* cmd = parsed_head_;
-    parsed_head_ = cmd->next;
-    if (parsed_head_ != nullptr)
+    AdvanceParsedHead(parsed_head_->next);
+    if (is_true_pipeline)  // pipeline mode as we have
       ReleasePipelinedCommand(cmd);
     else
       ReleaseParsedCommand(cmd);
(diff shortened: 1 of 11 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -1764,27 +1763,18 @@ void Connection::SquashPipeline() {
...
 
-  if (result.account_in_stats) {
-    conn_stats.pipeline_dispatch_calls++;
-    conn_stats.pipeline_dispatch_commands += num_dispatched_cmds;
-    conn_stats.pipeline_dispatch_flush_usec +=
-        CycleClock::ToUsec(CycleClock::Now() - flush_start_cycle_cnt);
-  }
+  conn_stats.pipeline_dispatch_calls++;
+  conn_stats.pipeline_dispatch_commands += num_dispatched_cmds;
+  conn_stats.pipeline_dispatch_flush_usec +=
```
glevkovich: I think this one should moved into line 1759, since flush is conditional.

src/server/server_family.cc
```diff
@@ -1763,10 +1763,6 @@ void PrintPrometheusMetrics(uint64_t uptime, const Metrics& m, DflyCmd* dfly_cmd
                                conn_stats.pipelined_latency_hist, conn_stats.pipelined_cmd_cnt,
                                conn_stats.pipelined_cmd_latency, &resp->body());
 
-  AppendMetricWithoutLabels("cmd_squash_stats_ignored_total", "",
```
glevkovich, line 1766: You removed this, but in tools/local/monitoring/grafana/provisioning/dashboards/dragonfly.json there is still dragonfly_cmd_squash_stats_ignored_total.
Need to drop it or replace with a different metrics.

src/facade/dragonfly_connection.h
```diff
@@ -468,14 +468,23 @@ class Connection : public util::Connection {
...
 
-  // Releases the command memory back to the pool.
-  // - Set is_pipelined=true if the command was successfully executed and should be counted
-  // in latency/throughput stats.
-  // - Set is_pipelined=false if the command is being dropped/cleaned up without execution or should
-  // not be counted in stats.
-  void ReleaseParsedCommand(ParsedCommand* cmd, bool is_pipelined);
+  // Releases the command memory back to the pool, updating pipeline queue accounting
+  // (queue length/bytes) but NOT per-command latency/throughput stats.
+  // Use for commands that are dropped/cleaned up without execution. It is also the common
```
glevkovich: This is not accurate:
I can see that in ReplyBatch and ExecuteBatch, this is called also after successful execution.
I think you should add:
`Used both for cleanup/dropped commands and for executed commands where per-command pipeline stats are intentionally skipped`

src/server/server_family.cc
romange: I kept it deliberately. I still need to work with older versions to I need this dashboard to run them (100% prod versions have this metric)

src/facade/dragonfly_connection.cc
```diff
@@ -2572,7 +2559,10 @@ bool Connection::ExecuteBatch() {
   auto advance_head = [this]() -> ParsedCommand* {
     auto* cmd = parsed_head_;
     parsed_head_ = cmd->next;
```
glevkovich: This can be replaced by
`AdvanceParsedHead(cmd->next)` 
and dropping the code in line 2628. loop will break anyway so no inefficient.

src/facade/dragonfly_connection.cc
romange: good catch. I think the bug was always there. Fixing

src/facade/dragonfly_connection.cc
```diff
@@ -2654,7 +2644,10 @@ bool Connection::ReplyBatch() {
...
```
glevkovich: This can be replaced by
`AdvanceParsedHead(cmd->next)` 
and dropping the code in line 2655. loop will break anyway since HasInFlightCommands will be false so no inefficiencies.

dranikpg (dismissed):

src/facade/dragonfly_connection.cc
```diff
@@ -1764,27 +1763,18 @@ void Connection::SquashPipeline() {
...
+    conn_stats.pipelined_wait_latency += CycleClock::ToUsec(start - parsed_head_->parsed_cycle);
 
-    ReleaseParsedCommand(current, result.account_in_stats /* is_pipelined */);
-    current = next;
-  }
-  parsed_head_ = current;
-  if (!parsed_head_) {
-    parsed_tail_ = nullptr;
+    ReleasePipelinedCommand(parsed_head_);
+    AdvanceParsedHead(next);
```
glevkovich, line 1776: Nit: Consider traversing the list via a local pointer variable and calling AdvanceParsedHead exactly once after the loop exits. I think it's the more conventional way, but this way is also correct (true for all calls to AdvanceParsedHead inside a loop).

glevkovich (dismissed):
Approved with few comments.

romange:
@glevkovich I followed up on your suggestions into a rabbit hole. Please take a look at my second commit and also read its description. I changed (improved?) the semantics of how we track pipeline stats in Execute/Reply-Batch functions.

dranikpg (approved):

PR #7596, opened 2026-06-15
build: Make the release artifacts use GLOG instead of absl

abhijat:
`make release` command is fixed to use GLOG. The builds in CI and local dev etc use absl.

In https://github.com/dragonflydb/dragonfly/pull/7514 everything was switched to absl, here release is switched back to glog


abhijat:
@romange ptal

romange (approved):

PR #7597, opened 2026-06-15
feat(server): Concurrent squashing

dranikpg:
Implement the option to run commands concurrently in non-atomic squashing to allow parallel disk reads and writes. The blocking point is moved from the command handler to the reply phase by using the new async cmds infrastructure

So far this applies only to **single shard mget**

Pipeline load before:

Image

Pipeline load after:

image

One more update in the PR

image

-> parallelization shows that it is now less important how many keys are in an MGET request to reach higher load.

This optimization is only enabled when tiered storage is enabled (otherwise it doesn't make any sense). It is visible that it has noticable overhead (creating new transaction, preparing the coro, etc)

### DIFF
--- a/src/server/multi_command_squasher.cc
+++ b/src/server/multi_command_squasher.cc
@@ -6,6 +6,8 @@
 
 #include <absl/container/inlined_vector.h>
 
+#include <boost/smart_ptr/intrusive_ptr.hpp>
+
 #include "base/cycle_clock.h"
 #include "base/flag_utils.h"
 #include "base/flags.h"
@@ -151,7 +153,7 @@ MultiCommandSquasher::SquashResult MultiCommandSquasher::TrySquash(CmdRef cmd) {
 
   auto& sinfo = PrepareShardInfo(last_sid);
 
-  sinfo.dispatched.push_back({.cmd = cmd, .reply = {}});
+  sinfo.dispatched.push_back({cmd, {}});
   order_.push_back(last_sid);
 
   bool need_flush = sinfo.dispatched.size() >= opts_.max_squash_size;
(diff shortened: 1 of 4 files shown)


dranikpg:
augment review

src/server/multi_command_squasher.cc
```diff
@@ -210,53 +212,80 @@ OpStatus MultiCommandSquasher::SquashedHopCb(EngineShard* es, RespVersion resp_v
...
+    bool do_async = es->tiered_storage() && !IsAtomic() && dispatched.cid->SupportsAsync();
+    if (do_async) {
+      // Create new transaction for the command (high overhead, but worth for disk hits)
+      boost::intrusive_ptr<Transaction> stx = new Transaction{dispatched.cid};
+      dispatched.cmd_cntx->SetupTx(dispatched.cid, stx.get());
+
+      // Create entry to keep transaction and arguemnt backing, enable new contex + async mode
+      async_commands.push_back({&dispatched, stx, std::move(arg_vec)});
+      ctx = dispatched.cmd_cntx;
+      ctx->SetDeferredReply();
```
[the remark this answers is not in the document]
dranikpg: maybe I'm missing something but `move_reply` is literally guarded by `if (!async)`

src/server/multi_command_squasher.cc
```diff
@@ -210,53 +212,80 @@ OpStatus MultiCommandSquasher::SquashedHopCb(EngineShard* es, RespVersion resp_v
...
+
+    if (!do_async)
+      move_reply(&dispatched);
+  }
+
+  // Collect replies from async commands
+  for (auto& de : async_commands) {
+    auto* ctx = de.cmd->cmd_cntx;
+    if (!ctx->CanReply())
+      ctx->Blocker()->Wait();
```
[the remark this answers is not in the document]
dranikpg: pal...

dranikpg:
augment review

dranikpg:
augment review

src/server/multi_command_squasher.cc
```diff
@@ -210,53 +212,78 @@ OpStatus MultiCommandSquasher::SquashedHopCb(EngineShard* es, RespVersion resp_v
...
-    *dst = std::move(src);
-    size_t sz = Size(*dst);
+    cmd->reply = crb.Take();
+    size_t sz = Size(cmd->reply);
     sinfo.reply_size_delta += sz;
     sinfo.reply_size_total_ptr->fetch_add(sz, std::memory_order_relaxed);
   };
 
+  // Command dispatched asynchronously
+  struct AsyncDispatched {
```
romange: why not move everything under `if (do_async))` block including `for (auto& de : async_commands) {` loop?


Then you can factor out the entire code into a separate function

src/server/multi_command_squasher.cc
```diff
@@ -210,53 +212,78 @@ OpStatus MultiCommandSquasher::SquashedHopCb(EngineShard* es, RespVersion resp_v
...
+    // With tiered storage enabled, it makes sense to dispatch async commands concurrently
+    // to allow concurrent disk operations. Tiered futures are only blocked on during replies
+    bool do_async = es->tiered_storage() && !IsAtomic() && opts_.pipeline_mode &&
+                    dispatched.cid->SupportsAsync();
+    if (do_async) {
+      // Create new transaction for the command (high overhead, but worth for disk hits)
+      boost::intrusive_ptr<Transaction> stx = new Transaction{dispatched.cid};
+      dispatched.cmd_cntx->SetupTx(dispatched.cid, stx.get());
+
+      // Create entry to keep transaction and arguemnt backing, enable new contex + async mode
```
romange: arguemnt

src/server/multi_command_squasher.cc
```diff
@@ -210,53 +212,78 @@ OpStatus MultiCommandSquasher::SquashedHopCb(EngineShard* es, RespVersion resp_v
...
-    auto status = local_tx->InitByArgs(cntx_->ns, cntx_->conn_state.db_index, args);
-    if (status != OpStatus::OK) {
-      crb.SendError(status);
+    crb.SetReplyMode(dispatched.reply_mode);
+
+    // With tiered storage enabled, it makes sense to dispatch async commands concurrently
+    // to allow concurrent disk operations. Tiered futures are only blocked on during replies
+    bool do_async = es->tiered_storage() && !IsAtomic() && opts_.pipeline_mode &&
+                    dispatched.cid->SupportsAsync();
+    if (do_async) {
```
romange, line 246: ah, it's in the loop of ` for (auto& dispatched : sinfo.dispatched) {` ?

src/server/multi_command_squasher.cc
```diff
@@ -210,53 +212,78 @@ OpStatus MultiCommandSquasher::SquashedHopCb(EngineShard* es, RespVersion resp_v
...
-      dispatched.cmd.cmd_cntx->Resolve(std::move(payload));
+
+    auto status = ctx->tx()->InitByArgs(cntx_->ns, cntx_->conn_state.db_index, args);
+    if (status != OpStatus::OK) {
+      ctx->SendError(status);  // Calls Resolve() in async, routes to crb in non async
     } else {
-      move_reply(std::move(payload), &dispatched.reply);
+      ctx->UpdateCid(dispatched.cid);
+      ctx->SetTailArgs(dispatched.args);
+      service_->InvokeCmd(args, ctx);
```
romange, line 260: the core issue is that this command runs "synchronously" and resolves whatever tiering futures we have, correct?

As you pointed out, there is an overhead with this approach. You do not even know if the key will hit the disk but you pay the price.

Did you consider an option of passing the future further without resolving it?

So today you have `GetReplies` that blocks on a future in ` template <typename T> void Send(TResultOrT<T>&& res) const {`

but what it you pass the future as is similarly to how we do it with `cmn::BorrowedString` ? Conceptually it behaves exactly like a borrowed string - same concept: you want to delay realization to the actual send, so capture builder may pass it as is (future) and then you block and realize it in the loop that writes in the coordinator fiber. 

And no special code is needed in squasher - it's transparent to it.

dranikpg:
it got faster and now lands below 25 at the end

image

romange:
Why do you even load it with MGET? The change and the issue is around pipelining so 
having a pipeline depth of 100 with GET at a low QPS should demonstrate the latency problem, no?

src/server/multi_command_squasher.cc
```diff
@@ -210,53 +210,60 @@ OpStatus MultiCommandSquasher::SquashedHopCb(EngineShard* es, RespVersion resp_v
...
     }
+
+    if (!do_async)
+      move_reply(&dispatched);
+    else if (!ctx->CanReply())
```
romange: add curly braces and explain what does it mean `(!ctx->CanReply())` and why `Blocker()->Wait(); ` is needed. Since we have transactions and coroutines - it can be confusing to understand

romange (dismissed):

dranikpg:
there's an issue with arming async transctions, fixing now

dranikpg:
Fixed small transaction issue that actually had no correctness impact, but crashed debug build as it did not properly clean a flag (armed flag) after itself

mkaruza (approved):

PR #7602, opened 2026-06-15
Coroutine batch squasher for ok_main V2 dispatch

### LINKED ISSUE #7601: Coroutine batch squasher (V2 dispatch) in ok_main.cc
> Design proposal for `ok_main.cc`. Status: proposal, not yet implemented.

## 1. Problem

`ok_main.cc` is the minimal shard-per-thread demo (`SET`/`GET`). Each async command is a
coroutine (`SetAsync`/`GetAsync`) that, **per command**, performs two cross-thread operations:

```
SetAsync(ctx):
  ctx->blocker.Start(1);
  g_shard_queues[shard]->Add(lambda);     // (1) cross-thread enqueue  -> wakes consumer fiber
  co_await ResolveAwaiter{ctx};            //     registers (blocker, handle) on the command
                                           //     -- suspends here, returns to connection --
  // shard fiber runs lambda: db op, then ctx->blocker.Dec();   // (2) cross-thread decrement
  ctx->rb()->SendOk();                     // resumed by Connection::SendReply()
```

The connection side (`Connection::IoLoopV2` → `ReplyBatch`, `src/facade/dragonfly_connection.cc`)
waits on the head command's blocker via `EmbeddedBlockingCounter::OnCompletion(waiter)`
(`Connection::WaitEvent`), and resumes the coroutine with `SendReply()` when `CanReply()` turns true.

For a pipeline of **N** commands spanning **S** distinct shards (S ≤ thread count), the current cost is:

| Operation | Current | Notes |
|-----------|--------:|-------|
| `g_shard_queues[...]->Add()` (cross-thread MPSC push + EventCount wake) | **N** | one per command |
| `blocker.Dec()` (cross-thread atomic + EventCount wake of connection) | **N** | one per command |
| Connection `OnCompletion` waiter registrations | up to **N** | one per head as it advances |
| Coroutine frames | **N** | one per command |

The two genuinely expensive items are the cross-thread `Add` and `Dec`: each can wake a fiber on
another thread (consumer fiber on enqueue, connection fiber on completion). At pipeline depth they
dominate. The goal of this design is to make these **batch-granular (S) instead of
command-granular (N)** while keeping the coroutine-based reply model.

This is explicitly **not** a re-implementation of `MultiCommandSquasher` (the
`--enable_pipeline_squashing_v2` path). That squasher runs a non-atomic distributed transaction
(`StartMultiNonAtomic`, hops, `MultiSwitchCmd`) and captures each reply through a
`CapturingReplyBuilder`. Here we want something lighter and coroutine-native.

## 2. Entry point and how the connection drives it

`OkService` currently inherits the default `ServiceInterface::DispatchSquashedBatch`, which returns
`0` (no squashing). The V2 dispatch loop already calls it before the sequential loop
(`Connection::ExecuteBatch`, gated by `--enable_pipeline_squashing_v2`):

```cpp
// dragonfly_connection.cc, ExecuteBatch()
if (pipeline_squashing_v2_ && parsed_cmd_q_len_ > 1 && protocol_ == Protocol::REDIS) {
  unsigned squashed =
      service_->DispatchSquashedBatch(parsed_to_execute_, parsed_cmd_q_len_, cc_.get());
  for (unsigned i = 0; i < squashed && parsed_to_execute_; i++)
    parsed_to_execute_ = parsed_to_execute_->next;   // advance EXECUTE cursor past squashed run
  conn_stats.pipeline_dispatch_commands += squashed;
}
```

Contract the squasher must honour (per `service_interface.h`):

- It receives the run `first … first+count-1` (the parsed linked list via `->next`).
- For every command it claims, it must `SetDeferredReply()` and arrange that the command eventually
  reports `CanReply() == true` and that `SendReply()` writes that command's reply **in parse order**.
- It returns the number of commands it consumed. The connection advances the **execute** cursor by
  that many; the **reply** cursor (`parsed_head_`) is drained separately by `ReplyBatch`, one command
  at a time, each gated on `parsed_head_->CanReply()`.

Replies are therefore still emitted through the existing per-command path
(`reply_ = SuspendedCommand{blocker, coro}` → `CanReply()` → `SendReply()`). Nothing about this
contract forces each command to own a *separate* blocker — that is the lever this design pulls.

## 3. Core idea: one shared blocker per batch

`ParsedCommand::CanReply()` for a suspended command is exactly
`std::get<SuspendedCommand>(reply_).blocker->IsCompleted()`. If every command in the batch stores a
`SuspendedCommand` pointing at the **same** `EmbeddedBlockingCounter`, then:

- The whole batch flips to "ready" atomically when that one blocker reaches zero.
- The connection registers **one** `OnCompletion` waiter (on the head); when it fires, every command
  in the batch already satisfies `CanReply()`, so `ReplyBatch` drains them all without re-waiting.

We initialise the shared blocker to the number of **active shards**, and have each shard's callback
`Dec()` exactly once after processing all of its commands. That gives **S `Add` + S `Dec`**.

```
BatchState { fb2::EmbeddedBlockingCounter blocker; }     // heap, std::shared_ptr

DispatchSquashedBatch(first, count, cntx):
  # Pass 1 — classify & group. Stop at first non-squashable command.
  run = []
  per_shard[num_shards] = {}
  for cmd in first .. first+count-1:
      if not single-key SET/GET: break
      sid = KeyShard(key)
      per_shard[sid].push_back(cmd); run.push_back(cmd)
      cmd->SetDeferredReply()
  if run.empty(): return 0
  active = count of non-empty per_shard buckets

  batch = make_shared<BatchState>()
  batch->blocker.Start(active)              # (A) arm BEFORE anything can Dec or register

  # Pass 2 — register reply coroutines (local; each suspends on the shared blocker)
  for cmd in run: EmitReply(cmd, batch)     # (B) registers (blocker, handle); suspends

  # Pass 3 — one cross-thread Add per active shard
  for sid where per_shard[sid] not empty:
      g_shard_queues[sid]->Add([batch, cmds = move(per_shard[sid])]() mutable {
          for c in cmds: run op, store result in c
          batch->blocker.Dec();             # (C) ONE Dec per shard
      })
  return run.size()
```

`EmitReply` is the only coroutine, parameterised by command:

```cpp
struct BatchAwaiter {                 // like ResolveAwaiter, but with an explicit shared blocker
  CmdContext* ctx; fb2::EmbeddedBlockingCounter* blocker;
  bool await_ready()  const noexcept { return false; }
  void await_suspend(std::coroutine_handle<> h) const noexcept { ctx->Resolve(blocker, h); }
  void await_resume() const noexcept {}
};

AsyncCmd EmitReply(CmdContext* ctx, std::shared_ptr<BatchState> batch) {
  co_await BatchAwaiter{ctx, &batch->blocker};   // suspend until the WHOLE batch is fulfilled
  // resumed by Connection::SendReply() on the connection thread, in parse order:
  if (ctx->is_get) { auto* rb = ...; ctx->get_result ? rb->SendBulkString(*ctx->get_result)
                                                      : rb->SendNull(); }
  else             { ctx->rb()->SendOk(); }
  // `batch` shared_ptr is part of the frame; it drops when the frame is destroyed at final_suspend
}
```

### 3.1 What `EmitReply` is, and what pass 2 does

`EmitReply` is not a new mechanism — it is today's per-command coroutine (`SetAsync`/`GetAsync`)
**split in half**, with the cross-thread `Add` hoisted out and batched in pass 3:

```cpp
// today: one coroutine does cross-thread dispatch AND reply
AsyncCmd SetAsync(ctx, pool) {
  ctx->blocker.Start(1);
  g_shard_queues[shard]->Add([...]{ op; ctx->blocker.Dec(); });   // <-- hoisted to pass 3 (per shard)
  co_await ResolveAwaiter{ctx};                                   // register handle + suspend
  ctx->rb()->SendOk();                                            // write reply after resume
}

// EmitReply: only the back half remains
AsyncCmd EmitReply(ctx, batch) {
  co_await BatchAwaiter{ctx, &batch->blocker};   // register handle + suspend
  ctx->rb()->SendOk();   /* or GET reply */      // write reply after resume
}
```

A coroutine call does **not** run to completion. Because of the `co_await`, each
`EmitReply(cmd, batch)` runs eagerly only up to the await, then **suspends and returns**. So pass 2
quickly produces N coroutines, each parked on the shared blocker. The work that happens *before* the
suspend is the entire purpose of the pass — `BatchAwaiter::await_suspend(h)` calls
`ctx->Resolve(blocker, h)`, which records `reply_ = SuspendedCommand{blocker, h}` on the command. That
registration is what makes the connection's existing reply path function:

- `ParsedCommand::CanReply()` reads `reply_` → `blocker->IsCompleted()`
- `ParsedCommand::SendReply()` reads `reply_` → `h.resume()` → the coroutine continues past the
  `co_await` and writes its reply

Without pass 2 the commands are flagged `SetDeferredReply()` but their `reply_` is still empty
(`monostate`); the connection would call `Blocker()`/`SendReply()` on an unregistered command and
crash. In other words: **pass 2 = "for each command, record which blocker to wait on and which
coroutine to resume when it fires."** The cross-thread half of the old coroutine becomes pass 3's one
`Add` per shard; the reply half stays per-command here. See §5(a) for why pass 2 must sit strictly
between arming the blocker (A) and issuing the `Add`s (C).

## 4. Lifetime

`BatchState` (which owns the shared blocker) must stay alive until the last reply is written.
Ownership is the set of **coroutine frames**: each `EmitReply` frame captures a
`std::shared_ptr<BatchState>` by value. The blocker therefore lives exactly as long as any reply
coroutine is still suspended or running.

- Per-command **results** live in the `CmdContext` (e.g. `get_result`), which the connection owns
  until `ReleaseParsedCommand`. The shard callback writes into the command; the coroutine reads it on
  resume. No extra per-result allocation.
- The shard callbacks also capture the `shared_ptr` (cheap insurance so the blocker is valid at
  `Dec()` even under unusual scheduling). It is provably alive without this — see ordering below — but
  capturing it keeps the invariant local and obvious.

When the final shard `Dec()`s the blocker to zero, all coroutines are still suspended (they only
resume *after* zero), so `BatchState` is alive. The connection then resumes each frame in order; as
each completes (`final_suspend` is `suspend_never`) its frame and its `shared_ptr` are destroyed. The
last frame to finish frees `BatchState`.

## 5. Ordering / correctness hazards

**(a) Registration-before-dispatch race.** A shard `Dec()` (and thus blocker completion) must not be
observable by the connection before every command has registered its `(blocker, handle)`. The phases
above prevent it: blocker is `Start()`ed (A) and all handles are registered (B) **before** any `Add`
is issued (C). Because the connection only inspects `CanReply()` after `DispatchSquashedBatch`
returns, and all registration happened synchronously before the cross-thread `Add`s, no `Dec` can
race ahead of a registration.

Also note `Start()` must precede registration: a freshly constructed counter is `0`, which reads as
`IsCompleted() == true`; registering a handle before `Start(active)` would let the connection resume
a command before its op ran.

**(b) Reply order.** Replies must reach the socket in parse order. `ReplyBatch` walks `parsed_head_`
in list order and calls `SendReply()` (→ resume `EmitReply`) one command at a time, so each coroutine
writes its own reply in the correct position regardless of which shard produced it or when. Within a
shard, the callback processes that shard's commands in submission order, but cross-shard interleaving
is irrelevant because the connection, not the shards, decides emission order.

**(c) Partial / non-squashable runs.** A command that is not a single-key `SET`/`GET` (wrong arity,
`PING`, unknown, future multi-key) stops the batch; we return the count consumed so far, and the
connection's sequential loop handles the remainder with the normal (sync or per-command async) path.
This mirrors `MultiCommandSquasher`'s "stop at the first command that can't join".

**(d) Connection waiter.** `Connection::WaitEvent` registers one `OnCompletion` on the head's blocker.
With a shared blocker the registration count drops from up-to-N to one per batch — a side benefit on
the connection thread, not just the shard threads.

## 6. Performance model

For N commands across S active shards:

| Operation | Current (per-command) | Proposed (batched) |
|-----------|----------------------:|-------------------:|
| Cross-thread `Add` | N | **S** |
| Cross-thread `Dec` | N | **S** |
| Connection `OnCompletion` registrations | ≤ N | **1** |
| Coroutine frames | N | N *(Variant A)* / **1** *(Variant B)* |
| DB ops | N | N (unchanged) |
| `sendmsg` syscalls | unchanged (IoLoopV2 coalescing) | unchanged |

`S` is bounded by the thread/shard count and is typically far smaller than pipeline depth, so the
cross-thread traffic that previously scaled with N now scales with S. The wins are independent of the
reply-coroutine variant below; only the frame count differs.

> **This table assumes small replies, where per-command cross-thread coordination dominates.** When
> replies are large the picture inverts — see §8. The batching win (`S` vs `N` cross-thread ops) is
> only meaningful when that coordination is an appreciable fraction of per-command cost; for large
> payloads the data movement dwarfs it.

## 7. Variant A vs Variant B (reply emission)

Both variants are identical up to and including the cross-thread batching (§3 passes 1 and 3). They
differ only in who writes the replies.

### Variant A — per-command reply coroutines (service-only) — recommended first step

As written in §3: N thin `EmitReply` coroutines share one blocker. Each writes its own reply.

- **Pros:** zero changes outside `ok_main.cc`; reuses the connection's reply path verbatim; lowest
  risk. Achieves the full cross-thread reduction (S `Add` + S `Dec`, 1 waiter).
- **Cons:** still N coroutine frames. They are cheap and purely thread-local (no cross-thread work in
  them), so this is a frame-allocation cost, not a synchronization cost.

### Variant B — single batch coroutine (needs a small connection extension)

One coroutine awaits the shared blocker and writes **all** replies in order:

```cpp
AsyncCmd EmitBatch(std::vector<CmdContext*> cmds, std::shared_ptr<BatchState> batch) {
  co_await BatchAwaiter{cmds.front(), &batch->blocker};
  for (auto* c : cmds) c->WriteReply();   // every reply, in parse order
}
```

This collapses frames to one per batch, matching the "use coroutines, but not at single-command
granularity" intent most directly. But the connection's reply loop is strictly one-command-per-
`SendReply` today, and there is **no "ready but writes nothing" payload** to fake the other K−1
commands (`payload::make_simple_or_noreply("")` maps to `monostate`, i.e. `index()==0`, i.e.
`CanReply()==false`). So Variant B requires a small, general connection change:

- Give `ParsedCommand` a `reply_span` (default 1). The head of a batch carries `reply_span = K`.
- In `ReplyBatch` (and the deferred-reply branch of `ExecuteBatch`, and `DestroyParsedQueue`), when a
  command with `reply_span = K` is replied, advance/release **K** `parsed_head_` entries after the
  single `SendReply()` that emitted all K replies.

- **Pros:** one coroutine frame per batch; cleanest expression of the intent.
- **Cons:** touches shared `dragonfly_connection.cc` and `parsed_command.{h,cc}`; the K−1 "covered"
  commands still need their args kept alive until the batch coroutine reads them, so release must be
  deferred to the span boundary. Broader blast radius and more invariants to verify.

### Recommendation

Implement **Variant A** first — it captures the entire cross-thread win (the stated goal) with no
shared-code risk, and is independently shippable and benchmarkable. The large-reply simulation in §8
reinforces this: Variant B's one differentiator (a single `writev` for the whole batch) is marginal
exactly where it could apply, and costs higher peak-memory residency. If the residual N-frame cost
shows up in profiles for *small*-reply pipelines, layer **Variant B**'s `reply_span` extension on top;
the squasher's grouping and shard-callback code are unchanged between the two, so B is a strict,
additive follow-up.

## 8. Simulation: a 32 × 1 MB GET pipeline

The variants are easiest to separate under a worst case for memory: a pipeline of **N = 32** `GET`s
(the `kMaxSquashSize` cap), each returning a **1 MB** value — **32 MB** of reply data total. This
section traces both variants through the actual reply-builder mechanics and compares latency, peak
memory, and reply (syscall) efficiency.

### 8.1 Reply-builder mechanics that drive the result

From `src/facade/reply_builder.cc` (`RedisReplyBuilderBase::SendBulkString`, `FinishScope`):

- A `SendBulkString` of a value larger than `kMaxInlineSize` (32 B) does **not copy** the value. It
  writes the `$<len>\r\n` prefix into the small backing buffer, then `WriteRef(value)` — an **iovec
  pointing at the source** (zero-copy) — then `\r\n`. The whole call is wrapped in a `ReplyScope`.
- At scope end, `FinishScope` flushes when `!batched_ || total_size_ * 2 >= kMaxBufferSize` (8192 B).
  For a 1 MB value `total_size_` is ~1 MB, so **every large reply force-flushes its own `Send()`**,
  even in batch mode. `Send()` issues one `Sink::Write(vecs…)`, which loops `WriteSome` until all
  bytes drain (`helio/io/io.h`). Batch mode only coalesces *small* replies (< ~4 KB) into the 8 KB
  buffer.
- Because the 1 MB is referenced (not copied), its source — the command's `get_result` string — must
  stay alive until that `Send()` completes.

Two consequences for **both** variants: (1) the 32 MB lives in the per-command `get_result` slots, not
in the reply builder; (2) the shared batch blocker only fires after *every* shard has finished, so all
32 MB is materialized **before the first reply byte is written** (no reply can precede batch
completion).

### 8.2 Variant A — per-command coroutines, drained by `ReplyBatch`

`ReplyBatch` resumes one coroutine per loop iteration: `SendReply()` → `SendBulkString(*get_result)`
→ its `ReplyScope` force-flushes one `Write()` of ~1 MB → `ReleasePipelinedCommand` frees that
command (and its 1 MB `get_result`) → next.

```
heap (get_result):  32MB ─┐ reply#1 sent+freed → 31MB → … → 1MB → 0
                          │ (one ~1MB writev per reply, source freed right after)
writev calls:        32   (one per reply; 3 iovecs each: prefix, 1MB ref, CRLF)
reply-builder buffer: ~0  (value is referenced, never copied)
time-to-first-byte:  at batch-complete + first writev
```

Peak heap ≈ **32 MB**, reached at batch completion and **declining monotonically** as each reply is
sent and released. Small replies on this same path would instead coalesce in the 8 KB buffer and flush
in a handful of `writev`s — i.e. Variant A already gets good send-batching for the common case.

### 8.3 Variant B — single coroutine writes all replies

The interesting sub-variant (B2) wraps the whole emission loop in **one** `ReplyScope` and relies on
iovec accumulation: for each command `WritePieces(prefix); WriteRef(get_result); WritePieces(CRLF)`.
With 32 commands that is 96 iovecs (< `IOV_MAX`), so nothing flushes until the single `FinishScope` at
the end, which issues **one** `Write()` scattering 32 MB across 96 iovecs.

```
heap (get_result):  32MB ───────────────────────────── held ───────── 0
                         (all 32 sources pinned for the entire 32MB egress)
writev calls:        1    (96 iovecs; internally loops WriteSome over 32MB)
reply-builder buffer: ~0
time-to-first-byte:  at batch-complete + iovec assembly (negligible)
```

The commands are owned by the connection and released only at the batch's `reply_span` boundary —
*after* the coroutine returns — so all 32 MB stays pinned through the whole 32 MB `Write()`. (A naive
B1 that uses a per-reply `ReplyScope` degenerates to Variant A's 32 `writev`s with none of A's free
structural incremental release.)

### 8.4 Side-by-side

| Metric | Current (no squash) | Variant A | Variant B2 (single writev) |
|---|--:|--:|--:|
| cross-thread `Add` / `Dec` | 32 / 32 | S / S | S / S |
| `Write()`/`writev` calls (reply phase) | 32 | 32 | **1** |
| iovecs per `Write()` | 3 | 3 | 96 |
| bytes copied into reply buffer | ~0 | ~0 | ~0 |
| peak `get_result` heap | ≤ 32 MB¹ | 32 MB, then declines | 32 MB, **held through egress** |
| incremental free during egress | yes | yes | **no** |
| time-to-first-byte | head shard done | all shards done | all shards done |

¹ The current model can start replying when the *head* shard finishes; if completions arrive roughly
in parse order it frees early and peaks below 32 MB. The shared blocker removes that: A and B both
guarantee all 32 results are resident before any reply. For a *uniform* fetch (all ~equal latency) the
three peaks converge to ~32 MB.

### 8.5 What this says about latency, efficiency, and the variant choice

- **Latency.** Both batched variants gate the first reply on the *slowest* command in the batch (the
  shared blocker), whereas the current model can stream the head reply as soon as its shard is done.
  For uniform 1 MB fetches the shards finish at similar times, so the gap is small; for skewed
  latencies, batching adds head-of-line delay. The dominant term in all cases is the 32 MB egress,
  which is identical across variants.
- **Reply efficiency.** B2's single `writev` is the only place B beats A — and the win is marginal:
  `Sink::Write` loops `WriteSome` over the same 32 MB regardless, so it saves syscall *count*, not
  data movement, while the socket send buffer (≪ 32 MB) forces multiple kernel writes either way.
  Meanwhile A *already* coalesces small replies via the 8 KB buffer, so B's edge only materializes for
  large replies — exactly the regime where squashing barely helps.
- **Peak memory.** Dominated by `N × value` of `get_result` copies (32 MB), inherent to the
  shared-blocker batch. A releases incrementally during egress; **B2 pins the full 32 MB for the
  entire 32 MB `Write()`** (worst residency). The copies themselves are unavoidable here: a zero-copy
  reply straight from `shard_db` would let the connection thread `writev` from memory the shard thread
  may mutate/rehash — unsafe — so the cross-thread copy is the safety cost both variants pay.
- **Verdict.** For the realistic target (deep pipelines of *small* `GET`/`SET`), Variant A delivers the
  cross-thread reduction, free send-coalescing, incremental release, and zero shared-code risk.
  Variant B's distinguishing benefit appears only for large replies, where it both barely matters and
  worsens peak residency. **Recommendation unchanged: implement A.**
- **Mitigation, independent of variant.** `GET` reply size is unknown until the shard reads it, so the
  batch can only be bounded by *count* (`kMaxSquashSize = 32`), accepting a `32 × max_value` peak
  (32 MB here). If that peak is a concern, lower the cap or skip squashing for keys known to be large.
  Note also that `get_result`'s heap is **not** counted by `ParsedCommand::UsedMemory`, so the
  pipeline command-cache accounting under-counts a cached large-value command — worth fixing
  separately (see `Connection::ReleaseParsedCommand`).

## 9. Resolved decisions

1. **Squash gate.** Reuse the existing `--enable_pipeline_squashing_v2` flag — the V2 wiring already
   gates `DispatchSquashedBatch` behind it, so `OkService` only needs to override the method. No new
   flag.
2. **Batch size cap.** Hard-coded `kMaxSquashSize = 32` in `ok_main.cc` (mirrors
   `MultiCommandSquasher::Opts::max_squash_size`). A batch holds at most 32 commands; per §8.5 this
   bounds peak `get_result` residency to `32 × max_value`.
3. **Mixed runs.** Stop at the first non-squashable command (wrong arity, `PING`, `COMMAND`, unknown);
   return the count consumed so far and let the connection's sequential loop handle the remainder.

## 10. Open questions

1. **Variant.** A vs B — recommendation is **A** (§7, §8). Confirm before implementing.
2. **`UsedMemory` accounting.** Should `CmdContext` count `get_result`'s heap in `UsedMemory` so the
   pipeline command-cache doesn't under-account cached large-value commands (§8.5)? Orthogonal to the
   squasher, but surfaced by it.
```

romange:
Implements the coroutine batch squasher proposed in #7601 (Variant A) for `ok_main.cc`.

## Summary
Folds a leading run of single-key `SET`/`GET` pipeline commands into one batch that shares a **single blocker**, so per-pipeline cross-thread cost drops from one `FiberQueue::Add` + one `blocker.Dec` **per command** to one **per active shard**, while keeping the coroutine-based reply model (each command still has its own thread-local reply coroutine, all awaiting the shared blocker).

## Changes
- **`OkService::DispatchSquashedBatch`** (`ok_main.cc`): 3 passes — group a run by destination shard → arm the shared blocker → register one reply coroutine per command → one `Add` (+ one `Dec`) per active shard.
- **Per-shard cap**: `kMaxSquashSize = 32` limits commands **per destination shard** (not per incoming run); the run stops at the first command whose target shard is full or that isn't a single-key `SET`/`GET`.
- **`Shard` struct**: all per-shard state (`queue`, `consumer`, `db`) unified under a global `vector<Shard>`. The batch grouping scratch is a per-call local — `DispatchSquashedBatch` can preempt in `FiberQueue::Add` (full queue), so a shared/thread-local buffer would be corrupted by a re-entrant call.
- **V2 wiring** (`dragonfly_connection.cc`): the IoLoopV2 RESP path calls `DispatchSquashedBatch` behind the new `--enable_pipeline_squashing_v2` flag; counts one `pipeline_dispatch` call per squashed batch.
- Design doc added at `docs/coroutine-batch-squasher.md`.

## Testing
- `ok_backend` and `dragonfly` build clean; `reply_builder_test` passes (35/36, 1 pre-existing skip).
- Manual pipelines (`--enable_resp_io_loop_v2 --enable_pipeline_squashing_v2`): SET/GET/missing-key/PING replies all correct, in order; verified squashing actually runs via `pipeline_dispatch_calls`/`pipeline_dispatch_commands`; per-shard cap exercised with a single shard.

Fixes #7601.

### DIFF
--- a/src/facade/ok_main.cc
+++ b/src/facade/ok_main.cc
@@ -51,6 +51,10 @@ ABSL_FLAG(uint16_t, uring_recv_buffer_cnt, 0,
 // ok_backend always binds on all interfaces (AcceptServer default) and has no persistence.
 ABSL_FLAG(std::string, bind, "", "Bind address (no-op in ok_backend)");
 ABSL_FLAG(std::string, dbfilename, "", "DB filename (no-op in ok_backend)");
+ABSL_FLAG(bool, coro_squash, false,
+          "Squash pipelined SET/GET batches with the coroutine-based squasher (a shared blocker "
+          "plus a per-command reply coroutine, no connection-fiber blocking) instead of the "
+          "default blocking squasher (block the connection fiber, then store captured payloads).");
 
 using namespace util;
 using namespace std;
@@ -290,6 +294,50 @@ void RunShardOps(absl::Span<CmdContext* const> cmds) {
   }
 }
 
+// Shared state for one coroutine-squashed batch. Owns the single blocker that every command in the
+// batch awaits. Kept alive by the shared_ptr captured in each reply coroutine frame (and in the
+// shard callbacks), so it outlives the cross-thread work and all reply resumes.
+struct BatchState {
+  fb2::EmbeddedBlockingCounter blocker{0};
+};
+
+// Awaiter variant that suspends a command on an explicit (shared) blocker rather than the command's
+// own. Like ResolveAwaiter, await_suspend registers (blocker, handle) with the ParsedCommand so the
+// connection can resume it via SendReply().
+struct BatchAwaiter {
+  CmdContext* ctx;
+  fb2::EmbeddedBlockingCounter* blocker;
+
+  bool await_ready() const noexcept {
+    return false;  // always suspend.
+  }
+
+  void await_suspend(std::coroutine_handle<> h) const noexcept {
+    ctx->Resolve(blocker, h);
+  }
+
+  void await_resume() const noexcept {
+  }
+};
+
+// Reply coroutine for one coroutine-squashed command: the back half of SetAsync/GetAsync. The
+// cross-thread dispatch (the per-command Add) is hoisted into the squasher as one Add per shard;
+// here we only register on the shared blocker, suspend, and -- once the whole batch is fulfilled
+// and the connection resumes us in parse order -- write this command's reply from its result slot.
+AsyncCmd EmitBatchedReply(CmdContext* ctx, shared_ptr<BatchState> batch) {
+  co_await BatchAwaiter{ctx, &batch->blocker};
+
+  if (ctx->batch_is_get) {
+    auto* rb = static_cast<RedisReplyBuilder*>(ctx->rb());
+    if (ctx->get_result)
+      rb->SendBulkString(*ctx->get_result);
+    else
+      rb->SendNull();
+  } else {
+    ctx->rb()->SendOk();
+  }
+}
+
 class OkService : public ServiceInterface {
  public:
   explicit OkService(ProactorPool* pool) : pool_(pool) {
@@ -309,10 +357,13 @@ class OkService : public ServiceInterface {
   }
 
  private:
-  // Dispatches the run already grouped into per_shard (active = number of non-empty buckets): one
-  // task per active shard, blocks the connection fiber until they finish, then stores a captured
-  // payload per command for the connection to emit in parse order.
-  void EmitBlockingBatch(unsigned num_active_shards, ShardBuckets* per_shard);
+  // The two squashers selected by --coro_squash; both consume the run already grouped into
+  // per_shard (active = number of non-empty buckets) and both block until the shards finish (the
+  // connection's SquashPipeline needs every reply ready on return). They differ in how the reply is
+  // delivered: EmitBlockingBatch stores a captured payload per command; EmitCoroBatch has every
+  // command share one blocker and a per-command reply coroutine the connection resumes afterwards.
+  static void EmitBlockingBatch(unsigned num_active_shards, ShardBuckets* per_shard);
+  static void EmitCoroBatch(unsigned num_active_shards, ShardBuckets& per_shard);
 
   // Synchronous fallback handlers (used when async is not available).
   DispatchResult HandleSetSync(ParsedCommand* cmd);
@@ -385,7 +436,11 @@ uint32_t OkService::DispatchSquashedBatch(ParsedCommand* first, unsigned count,
   if (group.processed == 0)
     return 0;
 
-  EmitBlockingBatch(group.num_active, &per_shard);
+  if (absl::GetFlag(FLAGS_coro_squash))
+    EmitCoroBatch(group.num_active, per_shard);
+  else
+    EmitBlockingBatch(group.num_active, &per_shard);
+
   return group.processed;
 }
 
@@ -417,6 +472,50 @@ void OkService::EmitBlockingBatch(unsigned num_active_shards, ShardBuckets* per_
       }
     }
   }
+}
+
+// Coroutine squasher (Variant A, see docs/coroutine-batch-squasher.md): every command shares one
+// BatchState blocker and registers a thread-local reply coroutine that suspends on it. We arm the
+// blocker, register the coroutines, issue one cross-thread Add per active shard (each Decs once),
+// then wait for the shared blocker. The shard callbacks own their command vectors (the lambdas move
+// them out of per_shard). Replies are written by resuming the per-command coroutines -- the only
+// difference from the blocking squasher, which instead stores captured payloads. Because the
+// connection's SquashPipeline expects every reply ready on return, we still block here; the win is
+// the cross-thread reduction (one Add + one Dec per active shard) plus the single shared blocker.
+void OkService::EmitCoroBatch(unsigned num_active_shards, ShardBuckets& per_shard) {
+  auto batch = make_shared<BatchState>();
+  batch->blocker.Start(
+      num_active_shards);  // (A) arm before any registration or Dec can be observed
+
+  // (B) register a reply coroutine per command. Each runs eagerly to its co_await, registers
+  // (shared blocker, handle) on the command, and suspends -- so all handles are in place before any
+  // shard can Dec the blocker (docs §5a). Order doesn't matter (the connection replies in parse
+  // order), so we iterate the buckets directly.
+  for (auto& bucket : per_shard) {
+    for (auto* ctx : bucket) {
+      ctx->SetDeferredReply();
+      EmitBatchedReply(ctx, batch);
+    }
+  }
+
+  // (C) one cross-thread enqueue per active shard; each runs all of its ops then a single Dec.
+  for (unsigned sid = 0; sid < per_shard.size(); sid++) {
+    if (per_shard[sid].empty())
+      continue;
+    g_shards[sid].queue->Add([batch, cmds = std::move(per_shard[sid])]() mutable {
+      RunShardOps(cmds);
+      batch->blocker.Dec();  // one Dec per shard
+    });
+  }
+
+  // (D) block the connection fiber until every shard has decremented the shared blocker. The
+  // connection's SquashPipeline requires CanReply() (== blocker->IsCompleted()) to hold for every
+  // command the moment we return, so we must reach zero here -- exactly like the blocking
+  // squasher's bc->Wait(). The replies themselves are still emitted by the per-command coroutines,
+  // which the connection resumes via SendReply() afterwards. batch stays alive past this return
+  // because each coroutine frame holds its own shared_ptr until SendReply() drives it to
+  // completion.
+  batch->blocker.Wait();
 }
 
 DispatchResult OkService::HandleSetSync(ParsedCommand* cmd) {


dranikpg (approved):
Responded privately:

It seems like our `SquashPipeline` in dragonfly_connection is good enough because it doesn't use blockers at all - it just assumes that all commands at the end are ready to reply.

Having a fully asynchronous squashing process is difficult

PR #7604, opened 2026-06-15
fix: KNN score aliases in FT.SEARCH and FT.AGGREGATE

### LINKED ISSUE #7582: `FT.AGGREGATE` does not materialize the KNN distance alias (`=>[KNN ... AS <alias>]`) into the aggregation pipeline
When a KNN vector clause aliases its distance with `=>[KNN k @vec $v AS <alias>]` and the query is run via **`FT.AGGREGATE`**, DragonflyDB does **not** expose `<alias>` to the aggregation pipeline. Downstream `LOAD <alias>`, `APPLY ... @<alias>`, and `SORTBY @<alias>` therefore all evaluate to `NULL` / no-op, and the result rows come back **with the alias field absent, null derived scores, and in insertion order -- with no error**.

The same KNN-with-alias works correctly on DragonflyDB via **`FT.SEARCH`** (the alias is returned and is sortable), and the native **`FT.HYBRID`** command also works. The defect is isolated to **`FT.AGGREGATE`**.

This is **not** a Valkey-parity quirk: both Redis 8.6 and Valkey 9.1 return the KNN alias correctly inside `FT.AGGREGATE`. DragonflyDB is the only engine of the three that does not.

Behavior across the three search commands (verified):
Same HASH index (`content` TEXT, `embedding` VECTOR FLAT FLOAT32 DIM 4 COSINE), 3 docs, query vector = doc `a`'s vector. KNN aliases the distance as `vector_distance`.

| Command | What it does with `AS vector_distance` | DragonflyDB | Redis 8.6 | Valkey 9.1 |
| --- | --- | :---: | :---: | :---: |
| `FT.SEARCH` | `RETURN`/`SORTBY vector_distance` | **OK** (alias returned, sorted) | OK | FAIL (`Index field vector_distance does not exist`) |
| `FT.AGGREGATE` | `LOAD vector_distance` + `APPLY (2-@vector_distance)/2 AS sim` | **BROKEN** (alias absent, `sim=None`) | OK | OK |
| `FT.HYBRID` (native) | `COMBINE LINEAR` ranked hybrid score | **OK** (proper ranking) | OK (different arg grammar) | n/a (command not supported) |

Note the inversion on the middle row: `FT.AGGREGATE` is the one place DragonflyDB diverges, and it diverges from **both** other engines.

Minimal reproduction (raw commands, no framework):
```python
import numpy as np, redis

def probe(port):
    r = redis.Redis(host="127.0.0.1", port=port)
    r.execute_command("FT.CREATE", "tri", "ON", "HASH", "PREFIX", "1", "tri:", "SCHEMA",
                      "content", "TEXT",
                      "embedding", "VECTOR", "FLAT", "6", "TYPE", "FLOAT32", "DIM", "4", "DISTANCE_METRIC", "COSINE")
    for k, v in {"a": [1,0,0,0], "b": [0,1,0,0], "c": [0.7,0.7,0,0]}.items():
        r.hset("tri:"+k, mapping={"content": f"doc {k}", "embedding": np.array(v, dtype=np.float32).tobytes()})
    q = np.array([1,0,0,0], dtype=np.float32).tobytes()

    print("FT.SEARCH   :", r.execute_command(
        "FT.SEARCH", "tri", "*=>[KNN 3 @embedding $v AS vector_distance]",
        "RETURN", "2", "content", "vector_distance", "SORTBY", "vector_distance", "ASC",
        "PARAMS", "2", "v", q, "DIALECT", "2")[1:])

    print("FT.AGGREGATE:", r.execute_command(
        "FT.AGGREGATE", "tri", "*=>[KNN 3 @embedding $v AS vector_distance]",
        "LOAD", "2", "content", "vector_distance",
        "APPLY", "(2 - @vector_distance)/2", "AS", "sim",
        "PARAMS", "2", "v", q, "DIALECT", "2")[1:])

probe(6399)   # DragonflyDB
probe(6380)   # Redis 8.6
```

Observed output:
```
# DragonflyDB :6399
FT.SEARCH   : ['tri:a', ['content','doc a','vector_distance','0'],
               'tri:c', ['vector_distance','0.2928932309','content','doc c'],
               'tri:b', ['content','doc b','vector_distance','1']]          # OK: alias present + sorted
FT.AGGREGATE: [['sim', None, 'content', 'doc c'],
               ['sim', None, 'content', 'doc a'],
               ['sim', None, 'content', 'doc b']]                            # BROKEN: no vector_distance, sim=None, insertion order

# Redis 8.6 :6380
FT.AGGREGATE: [['vector_distance','0','content','doc a','sim','1'],
               ['vector_distance','0.292893230915','content','doc c','sim','0.853553384542'],
               ['vector_distance','1','content','doc b','sim','0.5']]        # OK
# Valkey 9.1 :6381 (FT.AGGREGATE) -> same as Redis: vector_distance present, sim computed
```

On DragonflyDB the `vector_distance` field never appears in the `FT.AGGREGATE` rows, so `APPLY (2 - @vector_distance)/2` yields `NULL` (`sim=None`) and `SORTBY @<derived>` is a no-op.

Expected vs actual:
- **Expected (Redis/Valkey-compatible):** inside `FT.AGGREGATE`, the KNN clause's `AS <alias>` distance is a first-class field available to `LOAD`, `APPLY`, `SORTBY`, `GROUPBY` -- exactly as it already is for `FT.SEARCH`.
- **Actual (DragonflyDB):** the alias is not added to the per-document field set fed into aggregation, so any expression/sort referencing it sees `NULL`.

Secondary issue: `SORTBY` over an absent/all-`NULL` field is a silent no-op:
Because the derived `hybrid_score` is `NULL` for every row, `SORTBY 2 @hybrid_score DESC` neither errors nor reorders -- rows return in insertion order. This **masks** the primary bug from callers (no exception is raised; the result merely looks unranked). Redis raises/handles the missing field; DragonflyDB should not silently ignore a `SORTBY` over a field it produced no values for.

Impact:
`redisvl`'s `AggregateHybridQuery` (BM25 + vector fusion) is built on `FT.AGGREGATE`:
```
(~@content:(...) AND <filters>)=>[KNN k @embedding $vector AS vector_distance]
  SCORER BM25STD ADDSCORES LOAD ... DIALECT 2
  APPLY (2 - @vector_distance)/2 AS vector_similarity
  APPLY @__score AS text_score
  APPLY alpha*@text_score + (1-alpha)*@vector_similarity AS hybrid_score
  SORTBY 2 @hybrid_score DESC
```
On DragonflyDB, `@vector_distance` is `NULL`, so `vector_similarity` and `hybrid_score` are `NULL` and `SORTBY` is a no-op. MS Agent Framework's `RedisContextProvider` hybrid path (used whenever a vectorizer is supplied) then returns memories in **insertion order with no relevance ranking and no error** -- a silent retrieval-quality regression. The BM25 text score itself is computed correctly (`ADDSCORES` populates `@__score`), so only the vector component is lost.

Suggested fix:
In `FT.AGGREGATE`, propagate the synthetic KNN distance alias from `=>[KNN k @vec $v AS <alias>]` into the document field set consumed by the aggregation pipeline (`aggregate::Process` in `src/server/search/search_family.cc`, `CmdFtAggregate`), so `LOAD`/`APPLY`/`SORTBY`/`GROUPBY` can reference `<alias>`. The machinery already exists elsewhere: `FT.SEARCH` exposes the alias, and the native `FT.HYBRID` path internally issues `=>[KNN ... AS __knn_dist_hybrid]` and consumes that alias (`src/server/search/search_family.cc` around the hybrid KNN query construction). The gap is specifically that the aggregation field map is built without the KNN alias. Separately, make `SORTBY` over a field with no produced values surface an error instead of silently no-op'ing.

### LINKED ISSUE #7584: `FT.SEARCH` KNN on a JSON index does not expose the implicit `__vector_score` distance field
For a vector index created `ON JSON`, a KNN query written **without an explicit alias** --

```
*=>[KNN k @vector $q]
```

implicitly produces a distance field named **`__vector_score`** on Redis and Valkey, which the caller can `RETURN`/`SORTBY`. **DragonflyDB does not produce this implicit `__vector_score` field for a JSON index**, so a `RETURN __vector_score` yields nothing and the similarity score is lost.

The KNN ranking itself is correct on DragonflyDB (documents come back in the right order); only the score field is missing. Adding an explicit alias (`=>[KNN k @vector $q AS myscore]`) **does** work on DragonflyDB -- so the gap is specifically the *implicit default* `__vector_score` name on JSON indexes.

This breaks `valkey-glide` based clients (e.g. deepset's `valkey-haystack`), whose result parser reads the `__vector_score` field directly.

Minimal reproduction (client-agnostic, plain `redis-py`):
```python
import json, numpy as np, redis
def f32(v): return np.array(v, dtype=np.float32).tobytes()

def probe(port):
    r = redis.Redis(host="127.0.0.1", port=port)
    r.execute_command("FT.CREATE", "jb", "ON", "JSON", "PREFIX", "1", "jb:", "SCHEMA",
                      "$.content", "AS", "content", "TEXT",
                      "$.vector",  "AS", "vector",  "VECTOR", "FLAT", "6",
                      "TYPE", "FLOAT32", "DIM", "4", "DISTANCE_METRIC", "COSINE")
    for k, v in {"0": [1,0,0,0], "1": [0,1,0,0], "2": [0.7,0.7,0,0]}.items():
        r.execute_command("JSON.SET", "jb:"+k, "$", json.dumps({"content": "doc "+k, "vector": v}))

    q = f32([1,0,0,0])
    # KNN WITHOUT an explicit AS -> relies on the implicit __vector_score
    res = r.execute_command("FT.SEARCH", "jb", "*=>[KNN 3 @vector $q]",
                            "RETURN", "2", "content", "__vector_score",
                            "PARAMS", "2", "q", q, "DIALECT", "2")
    print("port", port, "->", res)

probe(6399)   # DragonflyDB
probe(6380)   # Redis 8.6
probe(6381)   # Valkey 9.1
```

Observed output (distance field per returned doc):
```
DragonflyDB :6399  doc fields = ['content']                   __vector_score = <ABSENT>   <-- bug
Redis 8.6   :6380  doc fields = ['__vector_score','content']  __vector_score = 0 / 0.2928... / ...
Valkey 9.1  :6381  doc fields = ['content','__vector_score']  __vector_score = 0 / 0.2928... / ...
```

DragonflyDB returns only `content`; the `__vector_score` field is absent. Redis and Valkey both return the KNN distance under `__vector_score`.

Scope (verified):

- **JSON index + implicit `__vector_score`:** DragonflyDB = ABSENT; Redis & Valkey = present. **(the bug)**
- **JSON index + explicit `AS myscore`:** DragonflyDB = present (works); Redis = present; Valkey = present (but Valkey then rejects `SORTBY myscore`). So an explicit alias is a viable client-side workaround on DragonflyDB.
- **HASH index + implicit `__vector_score`:** absent on all three (Dragonfly, Redis, Valkey) -- this is consistent/expected; the implicit default applies only to JSON indexes. So the divergence is specific to the **JSON-index implicit default**.

Expected vs actual:
- **Expected (Redis/Valkey-compatible):** a KNN query `=>[KNN k @vector $q]` on an `ON JSON` index exposes the distance under the default field name `__vector_score`, available to `RETURN`/`SORTBY`.
- **Actual (DragonflyDB):** no `__vector_score` field is produced for a JSON index; the score is unavailable unless the query supplies an explicit `AS <alias>`.

Impact:
deepset's **`valkey-haystack`** (the official Haystack Valkey integration, built on the `valkey-glide` client) issues exactly `... =>[KNN k @vector $q]` (no explicit `AS`) and reads `doc[b"__vector_score"]` to populate `Document.score`. Against DragonflyDB:
- documents are retrieved in the **correct similarity order**, but
- every `Document.score` is `None`.

This breaks score thresholds, score-based re-ranking, and any consumer that reads the similarity score. (`valkey-glide` connects to DragonflyDB fine and all other operations -- `FT.CREATE ON JSON`, `JSON.SET`, ranking -- work; only the score field is missing.)

Suggested fix:
Make `FT.SEARCH` on an `ON JSON` index expose the KNN distance under the default field name **`__vector_score`** when the KNN clause has no explicit `AS`, matching RediSearch 8.6 and valkey-search 1.2. (DragonflyDB already exposes the distance for an explicit alias on JSON indexes, so the computation is present; only the implicit/default field name is missing.)

vyavdoshenko:
Fix two KNN score alias issues across search and aggregation.
 
 Changes:
 - Materialize explicit KNN distance aliases (`AS <alias>`) into the `FT.AGGREGATE` pipeline so `LOAD`, `APPLY`, and `SORTBY` can consume them.
 - Use `__<vector_field>_score` as the implicit KNN score alias when `AS` is omitted, consistently for HASH/JSON and FLAT/HNSW paths.
 - Keep explicit `AS` aliases unchanged.
 - Add regression coverage for explicit aggregate aliases, implicit score aliases, and sparse aggregate `SORTBY`.
 
 Fixes: #7582
 Fixes: #7584


src/server/search/search_family.cc
```diff
@@ -95,6 +96,27 @@ bool SendErrorIfOccurred(const ParseResult<T>& result, CmdArgParser* parser,
...
+}
+
+std::optional<search::KnnScoreSortOption> ApplyCapturedIndexType(
+    std::optional<search::KnnScoreSortOption> option, const atomic<int>& index_type) {
+  // Coordinator paths only learn the index type inside shard callbacks; capture it there and
+  // apply the JSON-only default alias after the existing hop, without adding a pre-hop.
+  auto type = LoadIndexType(index_type);
+  if (option && type)
+    option = ApplyDefaultKnnScoreAlias(*option, *type);
+  return option;
```
BorysTheDev: your first check is option, but you process type before, I suggest making it more sequential.

src/server/search/search_family.cc
```diff
@@ -1331,47 +1357,46 @@ vector<SearchResult> SearchGlobalHnswIndex(
...
 
-  // If we don't return all fields
-  if (return_fields) {
-    // We have sort_option and it's different than knn score
-    if (set_sort_score) {
+    StoreIndexType(index->base().type, index_type);
+    const auto& schema = index->base().schema;
+    auto effective_knn_score_option = knn_score_option;
+    if (effective_knn_score_option) {
+      effective_knn_score_option =
```
BorysTheDev: looks strange

src/server/search/search_family.cc
```diff
@@ -2857,6 +2882,7 @@ void CmdFtSearch(CmdArgList args, CommandContext* cmd_cntx) {
 
   // Because our coordinator thread may not have a shard, we can't check ahead if the index exists.
   atomic<bool> index_not_found{false};
+  atomic<int> index_type{kUnknownIndexType};
```
BorysTheDev: why do you need atomic

src/server/search/search_family_test.cc
```diff
@@ -40,6 +40,22 @@ auto vector_score = [](std::string_view score_name, const RespExpr::Vec& score_f
...
+auto map_score = [](std::string_view score_name, const RespExpr::Vec& fields) -> float {
+  for (size_t i = 0; i + 1 < fields.size(); i += 2) {
+    if (fields[i].GetString() == score_name) {
+      float score = 0.0f;
+      bool success = absl::SimpleAtof(fields[i + 1].GetView(), &score);
+      if (!success) {
+        ADD_FAILURE() << "Could not parse score field: " << score_name;
+        return 0.0f;
+      }
+      return score;
```
BorysTheDev: nit:
```
if (float score = 0; absl::SimpleAtof(fields[i + 1].GetView(), &score)) {
   return score;
} else {
    ADD_FAILURE() << "Could not parse score field: " << score_name;
    return 0;
}
```

src/server/search/search_family.cc
vyavdoshenko: refactored

src/server/search/search_family.cc
vyavdoshenko: changed to vector<int>

src/server/search/search_family_test.cc
vyavdoshenko: fixed

src/server/search/search_family.cc
```diff
@@ -1260,6 +1260,37 @@ void WarmupQueryParser() {
...
+  if (!options.return_fields || !options.set_sort_score)
+    return options;
+
+  bool found_sort_return_field = false;
+  for (const auto& return_field : *options.return_fields) {
+    if (params.sort_option->field.Name() == return_field.Name()) {
+      found_sort_return_field = true;
+      break;
+    }
+  }
```
BorysTheDev: It smells like find_if

src/server/search/doc_index.h
```diff
@@ -10,6 +10,7 @@
 #include <memory>
 #include <optional>
 #include <string>
+#include <string_view>
```
BorysTheDev: strange diff for file

src/server/search/search_family_test.cc
```diff
@@ -49,6 +73,15 @@ auto Vec3ToBytes = [](float x, float y, float z) -> string {
...
 
+auto Vec4ToBytes = [](float x, float y, float z, float w) -> string {
+  string result;
+  result.append(reinterpret_cast<const char*>(&x), sizeof(float));
+  result.append(reinterpret_cast<const char*>(&y), sizeof(float));
+  result.append(reinterpret_cast<const char*>(&z), sizeof(float));
+  result.append(reinterpret_cast<const char*>(&w), sizeof(float));
+  return result;
+};
+
```
BorysTheDev: try c++20, variadic floating point template lambda. and you can use, in this case, something like 
absl::StrCat(reinterpret_cast<const char*>(&args), ...);

in this case, it will be a universal converter

src/server/search/search_family_test.cc
```diff
@@ -5809,6 +5985,214 @@ TEST_F(SearchFamilyTest, FtAggregateFilterAfterLimit) {
...
+                   "SORTBY",
+                   "2",
+                   "@vector_distance",
+                   "ASC",
+                   "PARAMS",
+                   "2",
+                   "v",
+                   Vec4ToBytes(1.0f, 0.0f, 0.0f, 0.0f),
+                   "DIALECT",
+                   "2"});
```
BorysTheDev: I think we need to create a constructor or function that will convert a string into array of strings. it is too long.

BorysTheDev (dismissed):

BorysTheDev (approved):

PR #7608, opened 2026-06-15
perf(facade): Drain socket in recv callback

glevkovich:
V2's single fiber cannot read from the socket while executing
commands. When the fiber is suspended at a cross-shard barrier
or idle-await, the kernel TCP receive buffer accumulates data.
If the buffer fills, the kernel drops packets or advertises a
zero-window, causing client retransmits and latency spikes.

ReadPendingInput() is now called directly in the RegisterOnRecv
callback. The callback fires while the fiber is suspended,
draining the kernel buffer before the fiber resumes and reaches
the top of its loop.

ReadPendingInput() is made self-guarding: it returns immediately
if pending_input_ is false, so callers need no external guard.
This also replaces the guarded call at the IoLoopV2 top-of-loop
with an unconditional one.

Safe: cooperative scheduling guarantees the callback and the
connection fiber never execute concurrently - no locks needed.

Benchmarks (AWS Graviton, 4 proactors, cross-machine, SET 2KB,
50 connections, 3 runs, memtier_benchmark):

  pipeline=1: 150k -> 164k ops/s (+9.2%)
  pipeline=10: 402k -> 422k ops/s (+5.0%)
  pipeline=50: 461k -> 480k ops/s (+4.0%)
  pipeline=100: 433k -> 444k ops/s (+2.5%)

Largest gains at low pipeline depths where the callback fires
frequently during shard-hop suspensions. No regressions in
single-conn, pubsub, or high-pipeline runs.

The RPS improvement comes from having data pre-loaded in io_buf_
when the fiber resumes, shortening the read-to-execute round-
trip rather than reducing TCP retransmits.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -809,6 +809,7 @@ void Connection::OnPostMigrateThread() {
     MaybeEnableRecvMultishot();
     socket_->RegisterOnRecv([this](const FiberSocketBase::RecvNotification& n) {
       NotifyOnRecv(n);
+      ReadPendingInput();
       io_event_.notify();
     });
   }
@@ -2902,6 +2903,8 @@ void Connection::NotifyOnRecv(const util::FiberSocketBase::RecvNotification& n) 
 }
 
 void Connection::ReadPendingInput() {
+  if (!pending_input_)
+    return;
   // Drain available socket data into io_buf_.
   io::MutableBytes buf = io_buf_.AppendBuffer();
   // A recv call can return fewer bytes than requested even if the
@@ -3059,6 +3062,10 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
   peer->RegisterOnRecv([this](const FiberSocketBase::RecvNotification& n) {
     DVLOG(2) << "Calling DoReadOnRecv iobuf_len: " << io_buf_.InputLen();
     NotifyOnRecv(n);
+    // Eagerly drain the kernel TCP receive buffer while the connection fiber might be is
+    // blocked/busy. This prevents rbuf starvation (V2's single fiber can't read while executing).
+    // Safe: callback only fires when the fiber has yielded - no concurrent access to io_buf_.
+    ReadPendingInput();
     io_event_.notify();
   });
 
@@ -3088,9 +3095,7 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
       current_wait_.emplace(parsed_head_, &cmd_completion_waiter);
     }
 
-    if (pending_input_) {
-      ReadPendingInput();
-    }
+    ReadPendingInput();
 
     // Idle park: flush and sleep when there is nothing to do (ShouldWakeIdle is false only when
     // the read buffer is also empty).


src/facade/dragonfly_connection.cc
```diff
@@ -2648,6 +2648,20 @@ bool Connection::ExecuteBatch() {
...
+    // faster than we read (e.g. large pipelined 2KB SETs), the kernel receive
+    // buffer fills and triggers client-side retransmits (~200ms RTO stalls).
+    // Draining every kMidBatchDrainInterval dispatched commands keeps the rbuf from overflowing
+    // without adding a syscall per command.
+    // TODO: tune kMidBatchDrainInterval based on profiling; 8 is a conservative start which showed
+    // improvements on most workloads.
+    static constexpr uint32_t kMidBatchDrainInterval = 8;
+    if (ioloop_v2_ && pending_input_ &&
+        (conn_stats.pipeline_dispatch_commands % kMidBatchDrainInterval) == 0) {
+      ReadPendingInput();
```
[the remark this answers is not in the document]

src/facade/dragonfly_connection.cc
```diff
@@ -2995,6 +3009,12 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
   peer->RegisterOnRecv([this](const FiberSocketBase::RecvNotification& n) {
     DVLOG(2) << "Calling DoReadOnRecv iobuf_len: " << io_buf_.InputLen();
     NotifyOnRecv(n);
+    // Eagerly drain the kernel TCP receive buffer while the connection fiber might be is
+    // blocked/busy. This prevents rbuf starvation (V2's single fiber can't read while executing).
+    // Safe: callback only fires when the fiber has yielded - no concurrent access to io_buf_.
+    if (pending_input_) {
+      ReadPendingInput();
+    }
     io_event_.notify();
```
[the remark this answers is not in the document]
glevkovich: fixed

src/facade/dragonfly_connection.cc
```diff
@@ -809,6 +809,9 @@ void Connection::OnPostMigrateThread() {
     MaybeEnableRecvMultishot();
     socket_->RegisterOnRecv([this](const FiberSocketBase::RecvNotification& n) {
       NotifyOnRecv(n);
+      if (pending_input_) {
+        ReadPendingInput();
```
romange: what is the invariant for  `ReadPendingInput` call? 
can we `DCHECK(pending_input_)` at the beginning ?

src/facade/dragonfly_connection.cc
```diff
@@ -809,6 +809,9 @@ void Connection::OnPostMigrateThread() {
     MaybeEnableRecvMultishot();
     socket_->RegisterOnRecv([this](const FiberSocketBase::RecvNotification& n) {
       NotifyOnRecv(n);
```
romange, line 811: `NotifyOnRecv` sets `pending_input_=true` only in a single branch there. Would not it be cleaner to move the `ReadPendingInput` call into `NotifyOnRecv` where we actually set pending_input_?

src/facade/dragonfly_connection.cc
glevkovich: On it's most basic form, ReadPendingInput() doesn't absolutely require pending_input_ to be true. it will be just wasted if there is nothing pending (with EAGAIN).
pending_input_ is performance optimization, not for correctness, so I don't think DCHECK is the right thing.  It DCHECKS that we haven't called the function with pending_input_ set to false, but we can just put the return if its set to false - this is  cleaner code, and optimisation still works.

I checked the code and all calls to ReadPendingInput() are guarded by if (pending_input_ .. ) ...
So, I think that having :
```
void Connection::ReadPendingInput() {
  if (!pending_input_) return;
...
```
is the right thing to do.
I think I will add a change in this PR to do as above, it simplifies the code for both v1 and v2.

src/facade/dragonfly_connection.cc
glevkovich: In term of correctness it will work but i think it's not the right thing to do.
NotifyOnRecv should do one thing - map the notification into the connection state.  ReadPendingInput is an I/O action that drains the socket.
If we merge them together we create unreadable code and we create coupling here, it violates single-responsibility. You get a hidden IO operation happening inside notification handler. if all code is written like this it will be much harder to trace and understand.
The pattern is: notify → set state → caller decides action

By the way you will solve only one call site (removing the if), other call sites don't go via NotifyOnRecv, so anyway they will need that pattern from outside:

```
    if (pending_input_) {
      ReadPendingInput();
    }
```

romange:
While in general I agree with the direction I am not sure it's enough - I am still checking.


1. One thing that I found confusing  is that it's hard to test V2 vs V1 with pipelining and squashing on ok_backend.

#7610 solves that.

2. Another thing that I found (and related to this PR) is something I was telling from the start - our V1 consumer / producer io-pattern with AsyncFiber and io fiber  is close to perfection - while AsyncFiber is blocked on execution of the pipeline, the io fiber wakes up to read and parse more commands - which leads to **deep** pipeline depths that result in efficient squashing. If you run Dragonfly you will see that `pipeline_dispatch_commands_total/pipeline_dispatch_calls_total` is much larger for V1 vs V2 for large pipelines. V1 Yield/Sleep is boosting this behavior even more.  Btw this is why I encouraged you to run local monitoring dashboard - these stats are there. This PR will not solve this. So if V2 io fiber does not preempt, `NotifyOnRecv` won't have the opportunity to run and ExecuteBatch will process shallow pipelines with the socket having input data pending! This change won't make a differrence if completion callbacks do not run. 

I do not want to block this PR on (2) but lets run series of benchmarks showing we do not degrade anything with this change. Also can we add `DCHECK(!pending_input_)` at the beginning of ExecuteBatch? Does it hold?

src/facade/dragonfly_connection.cc
romange: I disagree. The other branch in `NotifyOnRecv` consumes data from provided buffers and it's not possible to avoid it. In addition, you literally add socket consumption in each lambda and break the pattern. Whether it's part of the function or not - it does not matter, the callback does both things now. The only difference is that there is code duplication in this variant.

glevkovich:
@romange 
Here are the benchmarks I've done to make sure no regression V2 before change vs V2 after change + to understand where is V1 vs V2 after the change:


## V2 vs V1 Benchmark Report (d9f04de1, DF 4t multi_conn, 3-run avg)

### RPS

| Pipeline | V1 | V2 | V2/V1 |
|---:|---:|---:|---:|
| 1 | 179,151 | 166,962 | **−6.8%** |
| 10 | 428,529 | 433,402 | **+1.1%** |
| 50 | 549,195 | 481,536 | **−12.3%** |
| 100 | 567,420 | 444,726 | **−21.6%** |

### Latency (ms)

| Pipeline | Metric | V1 | V2 | V2/V1 |
|---:|---|---:|---:|---:|
| 1 | avg | 0.279 | 0.299 | +7% |
| 1 | p99.9 | 0.551 | 0.655 | +19% |
| 10 | avg | 1.164 | 1.152 | −1% |
| 10 | p99 | 1.887 | 1.791 | **−5%** |
| 10 | p99.9 | 2.367 | 2.159 | **−9%** |
| 50 | p99 | 6.367 | 13.375 | **+110%** |
| 50 | p99.9 | 7.391 | 17.535 | **+137%** |
| 100 | p99 | 13.354 | 31.018 | **+132%** |
| 100 | p99.9 | 16.170 | 41.386 | **+156%** |

### Batch Density (cmds/send syscall)

| Pipeline | V1 | V2 | V1 syscalls | V2 syscalls |
|---:|---:|---:|---:|---:|
| 1 | 1.0 | 1.0 | 2,687,517 | 2,504,672 |
| 10 | 9.7 | 4.9 | 661,012 | 1,316,681 |
| 50 | 47.5 | 23.6 | 173,602 | 306,243 |
| 100 | 95.1 | 50.2 | 89,547 | 132,873 |

### Insights

1. **p=10 is V2's sweet spot** — matches V1 on RPS (+1%) and beats it on tail latency (−9% p99.9). The drain fix eliminated the idle→data gap at this depth.

2. **BATCH_DENSITY explains the high-pipeline gap.** V1 sends 95 cmds/syscall at p=100; V2 sends 50. V2 issues 132k send syscalls vs V1's 89k — 48% more syscalls for 21% fewer ops. Each intermediate flush interrupts batching.

3. **Latency diverges at p≥50 because V2 lacks squashing.** V2 dispatches each command with an individual cross-shard hop. More hops → more intermediate reply completions → more flushes → higher p99/p99.9.

4. **V2 sends 2× more syscalls at p=10** (1.3M vs 661k, density 4.9 vs 9.7) yet achieves the same RPS — its per-send payload is half the size. The drain fix compensates by reducing idle-await duration.

5. **p=1 is nearly parity** — the 7% RPS gap and 19% p99.9 gap are the minimum overhead of V2's event-driven wakeup path vs V1's blocking recv. Batch density is identical (1.0) at this depth.



### V2 Before vs After Drain Fix (RPS)

| Pipeline | V2 before | V2 after | Improvement |
|---:|---:|---:|---:|
| 1 | 151,200 | 166,962 | **+10.4%** |
| 10 | 405,200 | 433,402 | **+7.0%** |
| 50 | 463,100 | 481,536 | **+4.0%** |
| 100 | 445,200 | 444,726 | flat |

### V2 Before vs After Drain Fix (Latency, ms)

| Pipeline | Metric | V2 before | V2 after | Improvement |
|---:|---|---:|---:|---:|
| 10 | p99 | 2.255 | 1.791 | **−21%** |
| 10 | p99.9 | 2.847 | 2.159 | **−24%** |
| 100 | p99.9 | 42.068 | 41.386 | −2% (flat) |

The fix delivers measurable gains at low-to-medium pipeline depths (p=1 through p=10) where the idle→data gap was the bottleneck. At p=100 it's neutral as expected — the fiber is continuously busy and the bottleneck shifts to squashing.

glevkovich:
@romange 
` Also can we add DCHECK(!pending_input_) at the beginning of ExecuteBatch? Does it hold?`
See my reply, I made some changes to code:
https://github.com/dragonflydb/dragonfly/pull/7608#discussion_r3418469356

romange:
> ges to code:

I asked about ExecuteBatch

glevkovich:
`Also can we add DCHECK(!pending_input_) at the beginning of ExecuteBatch? Does it hold?`
> > ges to code:
> 
> I asked about ExecuteBatch

@romange 
DCHECK(!pending_input_)   won't hold, since in ParseLoop() -> ParseRedisBatch() -> ParseRedis()  there is:
```
  if (ThisFiber::GetRunningTimeCycles() > max_busy_cycles) {
      GetLocalConnStats().num_read_yields++;
      ThisFiber::Yield();
    }
```

glevkovich:
`Another thing that I found (and related to this PR) is something I was telling from the start - our V1 consumer / producer io-pattern with AsyncFiber and io fiber is close to perfection - while AsyncFiber is blocked on execution of the pipeline, the io fiber wakes up to read and parse more commands - which leads to deep pipeline depths that result in efficient squashing. If you run Dragonfly you will see that`

@romange 
You're right that V1 is hard to beat (and now you are improving it more) - the io fiber wakes exactly when AsyncFiber is parked on a real wait, so the scheduler does the classification for free.

Maybe I should retry parsing on the proactor thread (on a separate PR).  I think it didn't work before since it wasn't "smart enough", and unfortunately to work well it has to be more complex than just "parse everything I can".
It has to answer 2 questions:

- Should I parse at all? It needs to know if parsing in the proctor context is beneficial ("real-wait" - e,g cross shard dispatch or a blocking operation). If it parse on the idle await in the loop it just steal cycles from other connections and becomes we get an unfair scheduling. Also if it parse for a connection that is the tail of scheduler run queue, it inflates the commands latency -variance.
- If yes how many commands or how much time to parse or how many bytes to parse or a combination of them.

**The practical rule:** 
- if the fiber is about to call **bc->Wait(), Await(), or co_await** and waits for another operation that happens on another thread, there is dead time on this thread that a proactor-level parse could fill. 
- If the fiber is just spinning on **Yield() / await_idle** waiting for network bytes that haven't arrived yet, there is no dead time to fill - the connection is I/O-bound, not compute-bound, and parsing now just burns cycles that belong to other connections. And I think that the proactor needs to distinguish between these 2 before parsing.

To give better answers I probably need to do some measurements.

glevkovich:
`pipeline_dispatch_commands_total/pipeline_dispatch_calls_total is much larger for V1 vs V2 for large pipelines. V1 Yield/Sleep is boosting this behavior even more. Btw this is why I encouraged you to run local monitoring dashboard - these stats are there. This PR will not solve this. So if V2 io fiber does not preempt, NotifyOnRecv won't have the opportunity to run and ExecuteBatch will process shallow pipelines with the socket having input data pending! This change won't make a differrence if completion callbacks do not run.`

@romange 

- I will run local monitoring dashboard now.
- "V2 io fiber does not preempt" - PR doesn't try to solve this, agree. This is something I know, the majority of dispatching during the loop do not preempt and thread needs to go back  to the top of the loop. Small correction to what you said : on 4 shards, in rare cases, for local inline dispatch (sync) on the local shard there are rare cases were fiber preempt during execution, but they are less than 0.5% of the executions (I measured).

src/facade/dragonfly_connection.cc
```diff
@@ -2648,6 +2649,14 @@ bool Connection::ExecuteBatch() {
       DCHECK(is_head);       // only head can execute sync
       cmd = advance_head();  // advance it
     }
+
+    // Drain pending socket data mid-batch. In V2's single-fiber architecture, the fiber
+    // cannot read while executing. pending_input_ acts as a guard: if set, the kernel has
+    // notified us that data is available, so drain it now. The drain clears pending_input_
+    // (via EAGAIN), so subsequent iterations skip this until the next notification arrives.
+    if (ioloop_v2_) {
+      ReadPendingInput();
```
romange: why here? specifically what does it provide?

romange: I do not think it's a good addition.

romange:
Can you provide an exact memtier command that reproduces the retransmits?

glevkovich:
> Can you provide an exact memtier command that reproduces the retransmits?
I ran it with both 2k and 4k:
`memtier_benchmark -s 172.31.30.209 -p 6379 -t 2 -c 25 --pipeline=100 --ratio=1:0 -d [2048|4096] `

And also
```
Default server rbuf: net.ipv4.tcp_rmem = 4096 131072 33554432 (max 33MB)
Reduced server rbuf: net.ipv4.tcp_rmem = 4096 32768 131072 (max 128KB)
```

romange (approved):

PR #7609, opened 2026-06-15
fix(test): Make container yield check atomic in ConcurrentWritesDuringContainerYield

### LINKED ISSUE #7606: GenericFamilyTest.ConcurrentWritesDuringContainerYield failed
CI: https://github.com/dragonflydb/dragonfly/actions/runs/27548455060/job/81427986476

```
2026-06-15T13:34:02.8359026Z 71: [ RUN      ] GenericFamilyTest.ConcurrentWritesDuringContainerYield
2026-06-15T13:48:44.1047459Z 48/80 Test  #71: generic_family_test ..............Subprocess killed***Exception: 883.38 sec
```

mkaruza:
The previous ConcurrentWritesDuringContainerYield check compared LRANGE with
a separate LLEN call, making the test racy because writers could append
between the two LLEN and LRANGE.

Move the verification into a read-only transaction callback: iterate the QList,
assert that iteration yielded, and compare the list size before and after the
iteration while the key lock is still held. Writer fibers continue issuing
independent RPUSH commands to verify writes are blocked during the read and make
progress afterward.

Closes https://github.com/dragonflydb/dragonfly/issues/7606

### DIFF
--- a/src/server/generic_family_test.cc
+++ b/src/server/generic_family_test.cc
@@ -2093,8 +2093,8 @@ TEST_F(GenericFamilyTest, ConcurrentWritesDuringContainerYield) {
   absl::FlagSaver fs;
   absl::SetFlag(&FLAGS_container_iteration_yield_interval_usec, 1);
 
-  constexpr int kListSize = 100'000;
-  constexpr int kIterations = 100;
+  constexpr int kListSize = 75'000;
+  constexpr int kIterations = 50;
   constexpr int kPushIterations = 100'000;
   constexpr int kWriterFibersPerShard = 8;
   const char* key = "list1";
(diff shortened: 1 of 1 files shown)


src/server/generic_family_test.cc
```diff
@@ -2085,13 +2085,18 @@ TEST_F(GenericFamilyTest, ContainerIterationYields) {
...
+// While a read transaction iterates over a large list and yields, a concurrent
+// fiber performs RPUSH operations. The key must not be modified while the read
+// transaction is running.
 TEST_F(GenericFamilyTest, ConcurrentWritesDuringContainerYield) {
-  constexpr int kListSize = 50'000;
-  constexpr int kIterations = 25;
+  absl::FlagSaver fs;
+  absl::SetFlag(&FLAGS_container_iteration_yield_interval_usec, 1);
+
+  constexpr int kListSize = 100'000;
```
romange: the constants are huge. How long the test runs in debug?

src/server/generic_family_test.cc
```diff
@@ -2102,41 +2107,73 @@ TEST_F(GenericFamilyTest, ConcurrentWritesDuringContainerYield) {
...
-    while (!done.load()) {
-      Run("writer_same_shard", {"RPUSH", key, "B"});
+  vector<Fiber> writer_fibers;
+  writer_fibers.reserve(2 * kWriterFibersPerShard);
+
+  auto launch_writers = [&](unsigned proactor_index, string_view id_prefix, string_view value) {
+    for (int fiber_index = 0; fiber_index < kWriterFibersPerShard; fiber_index++) {
+      writer_fibers.push_back(
+          pp_->at(proactor_index)->LaunchFiber([&, fiber_index, id_prefix, value] {
+            string id = StrCat(id_prefix, fiber_index);
```
romange: nit: get rid of `id_prefix` argument and create an id as: 
`StrCat("shard_", proactor_index, "_", fiber_index); `

romange (commented):
lgtm with some comments

src/server/generic_family_test.cc
mkaruza: From current run around ~4s in debug. I could lower this to be on safe side.

romange: maybe reduce to 2s.

mkaruza: Reduced to 1.4s on build with debug/santizers.

romange (approved):

PR #7610, opened 2026-06-16
feat(facade): squash V1 pipelines and support io_uring provided buffers in ok_backend demo

romange:
## Summary

Two related improvements to the `ok_backend` demo, both confined to `src/facade/ok_main.cc`.

### 1. V1 pipeline squashing

Implement `ServiceInterface::DispatchSquashedBatch` in the demo so it supports V1 pipeline squashing. The connection's V1 `SquashPipeline` path already calls `DispatchSquashedBatch`; until now the demo used the default (no-op) implementation and fell back to per-command dispatch.

- Group a leading run of single-key `SET`/`GET` commands by destination shard.
- Dispatch one task per active shard and block the connection fiber until all shards finish.
- Store each reply as a deferred payload, applied by the connection in parse order.
- Net effect: cross-thread cost drops from one `Add`+`Dec` per command to one `Add`+`Dec` per active shard.

### 2. io_uring provided buffers

Add a `uring_recv_buffer_cnt` flag and `RegisterBufRings()` that registers an io_uring buffer ring per thread (requires the io_uring proactor and kernel >= 6.2), so the demo's connections can use provided buffers for receive. The connection already consumes the ring via `BufRingEntrySize(kRecvSockGid)`.

### DIFF
--- a/src/facade/ok_main.cc
+++ b/src/facade/ok_main.cc
@@ -2,9 +2,9 @@
 // See LICENSE for licensing terms.
 //
 // A minimal demo that implements SET/GET commands with Dragonfly's
-// shard-per-thread architecture: each proactor thread owns a thread-local
-// hash map, and commands are dispatched to the correct shard through a
-// per-shard fb2::FiberQueue (MPSC) using the SuspendedCommand async
+// shard-per-thread architecture: each proactor thread owns one shard (its entry in the global
+// g_shards vector, holding that shard's hash map), and commands are dispatched to the correct shard
+// through a per-shard fb2::FiberQueue (MPSC) using the SuspendedCommand async
 // mechanism. Each proactor runs a consumer fiber draining its own queue,
 // while connection fibers on any thread enqueue callbacks via Add().
 // The synchronous fallback uses FiberQueue::Await when async is not
@@ -20,22 +20,31 @@
 #include "absl/strings/str_cat.h"
 #include "base/cycle_clock.h"
 #include "base/init.h"
+#include "base/proc_util.h"
 #include "facade/conn_context.h"
 #include "facade/dragonfly_connection.h"
 #include "facade/dragonfly_listener.h"
 #include "facade/facade_stats.h"
+#include "facade/facade_types.h"
 #include "facade/reply_builder.h"
+#include "facade/reply_payload.h"
 #include "facade/service_interface.h"
 #include "util/accept_server.h"
 #include "util/fibers/fiberqueue_threadpool.h"
 #include "util/fibers/pool.h"
 #include "util/fibers/synchronization.h"
+#ifdef __linux__
+#include "util/fibers/uring_proactor.h"
+#endif
 #include "util/http/http_common.h"
 #include "util/http/http_handler.h"
 #include "util/http/http_server_utils.h"
 
 ABSL_FLAG(uint32_t, port, 6379, "server port");
 ABSL_FLAG(uint32_t, fq_size, 256, "per-shard FiberQueue capacity");
+ABSL_FLAG(uint16_t, uring_recv_buffer_cnt, 0,
+          "How many buffer ring entries to allocate per thread for io_uring receive operations. "
+          "Relevant only for modern kernels with io_uring enabled");
 
 using namespace util;
 using namespace std;
@@ -45,15 +54,27 @@ namespace facade {
 
 namespace {
 
-// Thread-local shard storage — each proactor thread has its own hash map.
-thread_local absl::flat_hash_map<string, string> shard_db;
+// All state for one shard. g_shards[sid] is owned by shard sid: in the shared-nothing model each
+// proactor thread owns exactly one shard and initializes its own slot (see RunEngine).
+//  - `queue`/`consumer`: the MPSC FiberQueue and the fiber draining it. Both live on the owning
+//    proactor thread, but callbacks are enqueued from connection fibers on ANY thread via
+//    Add()/Await() -- which is why g_shards is global, not thread-local.
+//  - `db`: this shard's slice of the keyspace (touched only by the owning thread).
+struct Shard {
+  unique_ptr<fb2::FiberQueue> queue;
+  fb2::Fiber consumer;
+  absl::flat_hash_map<string, string> db;
+};
 
-// One MPSC FiberQueue per shard (indexed by proactor index). Each queue is created and
-// drained by a consumer fiber running on its owning proactor thread (see RunEngine), but
-// callbacks are enqueued into it from connection fibers on any thread via Add()/Await().
-// g_shard_consumers holds the consumer fibers so they can be joined on shutdown.
-vector<unique_ptr<fb2::FiberQueue>> g_shard_queues;
-vector<fb2::Fiber> g_shard_consumers;
+// One entry per shard, indexed by proactor index. Resized once in RunEngine; each proactor thread
+// then initializes its own entry.
+vector<Shard> g_shards;
+
+// The shard owned by the calling proactor thread. Queue callbacks run on their shard's owning
+// thread, so inside a callback MyShard() is that shard.
+Shard& MyShard() {
+  return g_shards[fb2::ProactorBase::me()->GetPoolIndex()];
+}
 
 // Determine the owning shard for a given key.
 constexpr uint64_t kShardHashSeed = 120577240643ULL;
@@ -111,6 +132,9 @@ struct CmdContext : public facade::ParsedCommand {
 
   fb2::EmbeddedBlockingCounter blocker{0};
   optional<string> get_result;
+
+  // Set by DispatchSquashedBatch so the reply payload can be built after the shard work completes.
+  bool batch_is_get = false;
 };
 
 // Custom awaiter that registers the coroutine with the connection at the exact
@@ -148,8 +172,8 @@ AsyncCmd SetAsync(CmdContext* ctx, ProactorPool* pool) {
   unsigned shard_id = KeyShard(key, pool->size());
 
   ctx->blocker.Start(1);
-  g_shard_queues[shard_id]->Add([k = std::move(key), v = std::move(value), ctx]() mutable {
-    shard_db.insert_or_assign(std::move(k), std::move(v));
+  g_shards[shard_id].queue->Add([k = std::move(key), v = std::move(value), ctx]() mutable {
+    MyShard().db.insert_or_assign(std::move(k), std::move(v));
     ctx->blocker.Dec();
   });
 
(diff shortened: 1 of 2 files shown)


src/facade/ok_main.cc
glevkovich, line 130: nit:
 not part of PR but think you should also reset here:
`batch_is_get = false;`

src/facade/ok_main.cc
```diff
@@ -459,6 +568,40 @@ void RunEngine(ProactorPool* pool, AcceptServer* acceptor) {
...
+
+  base::sys::KernelVersion kver;
+  base::sys::GetKernelVersion(&kver);
+  unsigned ver = kver.kernel * 100 + kver.major;
+  if (ver < 602) {
+    LOG(WARNING) << "uring_recv_buffer_cnt requires kernel >= 6.2";
+    return;
+  }
+
+  bufcnt = absl::bit_ceil(bufcnt);
```
glevkovich, line 593: can be undefined behaviour.

- bufcnt is uint16_t (max value 65535).
- absl::bit_ceil(x) returns the smallest power of 2 which is >= x
- if bufcnt 32769 -> it returns 65536 which doesnt fit (UB)
- UringProactor::RegisterBufferRing()  has DCHECK at the beggining:
  CHECK_LT(nentries, 32768u);

so  you should DCHECK here:
`DCHECK_LE(bufcnt, 16384)`

This is because all the half closed range (16385, 32768]  will crash in RegisterBufferRing, 
since - absl::bit_ceil returns 32768 for all numbers in that range.

src/facade/ok_main.cc
```diff
@@ -257,13 +284,98 @@ DispatchResult OkService::DispatchCommand([[maybe_unused]] ParsedArgs args, Pars
...
+// reply as a captured payload via Resolve() (which also marks the command deferred). The connection
+// applies the payloads in parse order afterwards. Cross-thread cost drops from one Add + one Dec
+// per command to one Add + one Dec per active shard.
+uint32_t OkService::DispatchSquashedBatch(ParsedCommand* first, unsigned count,
+                                          [[maybe_unused]] ConnectionContext* cntx) {
+  const unsigned num_shards = pool_->size();
+
+  // Grouping scratch, private to this call (stays valid on the parked fiber's stack while pass 2
+  // tasks run on shard threads).
+  vector<absl::InlinedVector<CmdContext*, 4>> per_shard(num_shards);
```
glevkovich, line 303: If you define it like that think the outer vector allocates on the heap for each call to DispatchSquashedBatch.
I think you should change it to:

```
constexpr unsigned kPerShardInlineCmds = 4;
constexpr unsigned kInlineShardCount = 32;
absl::InlinedVector<absl::InlinedVector<CmdContext*, kPerShardInlineCmds>, kInlineShardCount> per_shard(num_shards);
```

32 covers up to 32 shards.

src/facade/ok_main.cc
```diff
@@ -257,13 +284,98 @@ DispatchResult OkService::DispatchCommand([[maybe_unused]] ParsedArgs args, Pars
...
+    if (per_shard[sid].empty())
+      continue;
+    g_shards[sid].queue->Add([cmds = &per_shard[sid], bc]() mutable {
+      for (auto* ctx : *cmds) {
+        if (ctx->batch_is_get) {
+          auto it = MyShard().db.find(ctx->at(1));
+          if (it != MyShard().db.end())
+            ctx->get_result = it->second;
+        } else {
+          MyShard().db.insert_or_assign(string(ctx->at(1)), string(ctx->at(2)));
```
glevkovich: - ctx->at(1) returns a string_view. 
- string(ctx->at(1)) constructs a heap-allocated std::string unconditionally, even if the key already exists in the map and the new string is immediately thrown away after the assignment. So this is the  waste.
- string(ctx->at(2)) -> same waste

consider replacing this line with:
```
auto [it, inserted] = MyShard().db.try_emplace(ctx->at(1), ctx->at(2));
          if (!inserted) {
            it->second.assign(ctx->at(2));
          }
```
So for insert of new key we will always have 2 allocs, but if only updating we will have in the current code 2 allocs, and in proposed code 0 or 1 allocs (0 if the new value is smaller size, assign will no allocate).

src/facade/ok_main.cc
```diff
@@ -257,13 +284,98 @@ DispatchResult OkService::DispatchCommand([[maybe_unused]] ParsedArgs args, Pars
...
+
+  // Pass 2: dispatch one task per active shard; each runs all of its ops then decrements the
+  // counter. Block here until every shard finishes (this parks the connection fiber, like
+  // dragonfly's squasher blocking on its transaction hops).
+  fb2::BlockingCounter bc(active);
+  for (unsigned sid = 0; sid < num_shards; sid++) {
+    if (per_shard[sid].empty())
+      continue;
+    g_shards[sid].queue->Add([cmds = &per_shard[sid], bc]() mutable {
+      for (auto* ctx : *cmds) {
```
glevkovich, line 343: nit: You call MyShard() 3 times in the lambda. It's cheap but accumulates.

```
Shard& MyShard() {
  return g_shards[fb2::ProactorBase::me()->GetPoolIndex()];
}
```
requires multiple operations until returns.
Consider setting
`Shard& shard = MyShard();`
at the top to reduce number of calls to MyShard in the lambda.

src/facade/ok_main.cc
```diff
@@ -442,11 +547,10 @@ void RunEngine(ProactorPool* pool, AcceptServer* acceptor) {
 
   // Stop each consumer fiber on its owning proactor thread and join it there.
   pool->AwaitFiberOnAll([](unsigned index, ProactorBase*) {
-    g_shard_queues[index]->Shutdown();
-    g_shard_consumers[index].Join();
+    g_shards[index].queue->Shutdown();
+    g_shards[index].consumer.Join();
```
glevkovich, line 555: i think you should add here:
`g_shards[index].db.clear();`
It will still work since  line 553 calls clear on the while vector which calls dtors, but its the right thing to do since each proactor should clean it's members.

src/facade/ok_main.cc
```diff
@@ -257,13 +284,98 @@ DispatchResult OkService::DispatchCommand([[maybe_unused]] ParsedArgs args, Pars
...
+    });
+  }
+  bc->Wait();
+
+  // Pass 3: store each command's reply as a captured payload. Resolve() also marks the command
+  // deferred; the connection emits them in parse order, so bucket order here does not matter.
+  for (auto& bucket : per_shard) {
+    for (auto* ctx : bucket) {
+      if (!ctx->batch_is_get) {
+        ctx->Resolve(payload::SimpleString{std::string("OK")});
```
glevkovich: nit:
I think it will work?
`ctx->Resolve(payload::SimpleString{"OK"});`

src/facade/ok_main.cc
```diff
@@ -257,13 +284,98 @@ DispatchResult OkService::DispatchCommand([[maybe_unused]] ParsedArgs args, Pars
...
+      } else if (ctx->get_result) {
+        ctx->Resolve(payload::BulkString{std::move(*ctx->get_result)});
+      } else {
+        ctx->Resolve(payload::Payload{payload::Null{}});
```
glevkovich: nit:
I think this will work?
`ctx->Resolve(payload::Null{})`

src/facade/ok_main.cc
romange: it's an optimization, I will consider it in the future.

src/facade/ok_main.cc
```diff
@@ -257,13 +284,98 @@ DispatchResult OkService::DispatchCommand([[maybe_unused]] ParsedArgs args, Pars
...
+  unsigned processed = 0;
+  unsigned active = 0;
+  auto* cmd = first;
+  for (unsigned i = 0; i < count && cmd != nullptr; i++, cmd = cmd->next) {
+    if (cmd->empty())
+      break;
+    string_view name = cmd->Front();
+    bool is_get = absl::EqualsIgnoreCase(name, "GET");
+    bool is_set = absl::EqualsIgnoreCase(name, "SET");
+    if (is_get ? cmd->size() < 2 : (is_set ? cmd->size() < 3 : true))
```
glevkovich: nit:
looks complicated and not-easy to read to me. I would refactor this to:
```
if (!is_get && !is_set) 
  break;
if (is_get && cmd->size()<2)
  break;
if (is_set && cmd->size() < 3)
  break;
```

src/facade/ok_main.cc
romange: Please note that it's not the goal at the moment to optimize shards. I will push it to the future improvements.

src/facade/ok_main.cc
```diff
@@ -257,13 +284,98 @@ DispatchResult OkService::DispatchCommand([[maybe_unused]] ParsedArgs args, Pars
...
+      break;
+
+    auto* ctx = static_cast<CmdContext*>(cmd);
+    auto& bucket = per_shard[KeyShard(ctx->at(1), num_shards)];
+    if (bucket.size() == kMaxSquashSize)  // this shard is full -> end the run here
+      break;
+    ctx->batch_is_get = is_get;
+    active += bucket.empty();  // first command for this shard -> one more active shard
```
glevkovich: nit (more readable):
```
if (bucket.empty()) 
   ++active;
```

src/facade/ok_main.cc
```diff
@@ -257,13 +284,98 @@ DispatchResult OkService::DispatchCommand([[maybe_unused]] ParsedArgs args, Pars
   return DispatchResult::OK;
 }
 
+// Maximum number of commands a single destination shard may take in one squashed batch; the run
+// stops at the first command whose target shard is already full.
+constexpr unsigned kMaxSquashSize = 32;
+
+// Pipeline squasher, mirroring dragonfly's --enable_pipeline_squashing_v2 path
```
glevkovich: is this a mistake in the config parameter name (enable_pipeline_squashing_v2) ?

src/facade/ok_main.cc
```diff
@@ -20,22 +20,31 @@
...
+#include "facade/facade_types.h"
 #include "facade/reply_builder.h"
+#include "facade/reply_payload.h"
 #include "facade/service_interface.h"
 #include "util/accept_server.h"
 #include "util/fibers/fiberqueue_threadpool.h"
 #include "util/fibers/pool.h"
 #include "util/fibers/synchronization.h"
+#ifdef __linux__
+#include "util/fibers/uring_proactor.h"
```
glevkovich, line 37: I think you are missing here 2 headers, currently it works by calling other headers which include them.

```
absl/container/inlined_vector.h
absl/numeric/bits.h
```

glevkovich (dismissed):
Approved with some comments/nits

src/facade/ok_main.cc
romange: removed.

src/facade/ok_main.cc
romange: We do not have strict policy of including everything directly. I find it hard to manage.

src/facade/ok_main.cc
romange: it's subjective :)

glevkovich (approved):

PR #7615, opened 2026-06-16
fix(facade): propagate parse status from batch parsers

romange:
## Summary

`ParseRedisBatch` / `ParseMCBatch` returned a `bool` that collapsed four distinct outcomes — clean command boundary (`OK`), needs more input (`NEED_MORE`), backpressure throttle, and protocol error (`ERROR`) — into `true`/`false`. Two problems followed:

- **Swallowed RESP protocol errors.** `ParseLoop` ended with `return commands_parsed ? OK : NEED_MORE` and never produced `ERROR`. In the event-driven loop (`enable_resp_io_loop_v2`, default **off** — so this is latent), a malformed RESP command was therefore treated as "need more data": the connection never sent a protocol-error reply and never closed. Unlike Memcache, the RESP `enqueue_only` path does not bake errors into a deferred reply, so the error had nowhere to go. (The default RESP path calls `ParseRedis` directly and was unaffected.)
- **Misleading semantics.** `true` meant "the parser stopped at a clean boundary," not "a command was parsed." A buffer ending in a partial command yields `NEED_MORE` → `false` even though complete commands ahead of it were parsed and enqueued.

## Fix

- Both batch parsers now return the 3-valued `ParserStatus`:
  - `ParseRedisBatch` forwards `ParseRedis`'s status verbatim; backpressure throttle → `NEED_MORE`.
  - `ParseMCBatch` returns `OK` (it converts parse errors into deferred replies) and `NEED_MORE` on `INPUT_PENDING`.
- `ParseLoop` executes/replies the commands parsed so far (preserving in-order replies), then returns `ERROR` on a protocol error so `HandleRequests`/`IoLoopV2` send the protocol-error reply (via `parser_error_`) and close.

## Verification

- `dragonfly` builds.
- Pipeline + Memcache error/pipeline integration tests pass (`connection_test::test_pipeline_support`, `pymemcached_test::TestMemcached::{test_error_in_pipeline,test_noreply_pipeline}`).
- Manual raw-socket test with `--enable_resp_io_loop_v2=true`: a valid `PING` returns `+PONG`, and a malformed `*x\r\n` now returns `-ERR Protocol error: ...` and closes the connection (previously it hung waiting for more input).

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1496,18 +1496,25 @@ auto Connection::ParseLoop() -> ParserStatus {
   auto parse_func =
       protocol_ == Protocol::MEMCACHE ? &Connection::ParseMCBatch : &Connection::ParseRedisBatch;
 
-  bool commands_parsed = false;
+  ParserStatus parse_status = NEED_MORE;
   do {
-    commands_parsed = (this->*parse_func)(io_buf_);
+    parse_status = (this->*parse_func)(io_buf_);
 
+    // Execute/reply the commands parsed so far first, so a trailing protocol error still flushes
+    // earlier replies in order before we report it.
     if (!ExecuteBatch())
       return ERROR;
 
     if (!ReplyBatch())
       return ERROR;
-  } while (commands_parsed && io_buf_.InputLen() > 0);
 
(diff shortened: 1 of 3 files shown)


kostasrim (approved):
clean!

PR #7616, opened 2026-06-16
fix(search): ensure index-level options persist across restart

yy782:
Update DocIndexInfo::BuildRestoreCommand() to include:
- NOOFFSETS option
- LANGUAGE option
- LANGUAGE_FIELD option
- NOSTEM option for text fields

This fixes the issue where index configuration was lost after server restart or during replication, because these options were not serialized in the restore command.

Add test_index_level_options_persist_after_restart to verify the fix.

### DIFF
--- a/src/server/search/doc_index.cc
+++ b/src/server/search/doc_index.cc
@@ -204,6 +204,18 @@ string DocIndexInfo::BuildRestoreCommand() const {
       absl::StrAppend(&out, " ", sw);
   }
 
+  if (base_index.options.no_offsets) {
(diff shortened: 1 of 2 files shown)


yy782:
I won't fix the hardcoded "english" issue in this PR. It's a pre-existing pattern used in multiple places, and I'd prefer to keep this PR focused on the persistence fix.

yy782:
I see that #7620 has been merged and fixes this issue. I'm closing my PR. Thanks for the work on this!

PR #7617, opened 2026-06-16
refactor(facade): extract IoLoopV2 helpers to cut duplication

romange:
## Summary
Pure, no-behavior-change extraction of repeated patterns in `Connection::IoLoopV2` into named helpers. The loop body shrinks and each duplicated condition now has a single source of truth.

## Changes
- `FlushReplies()` now returns `std::error_code`, collapsing four `Flush()`+`GetError()` sites (idle park, backpressure park, `io_ec_`, migration).
- `IsOverPipelineLimit()` — single source of truth for the parse-throttle predicate, reused in `ParseRedisBatch` and both data-path checks.
- `NotifyIfMemReleased()` replaces the duplicated mem-before/notify pattern (kept as a call, not a scope guard, to preserve notify-before-park ordering).
- `ShouldUnpark()` factors the shared tail of the two `await` predicates.
- `IsReadyToMigrate()` replaces the local `is_ready_to_migrate` lambda.

Comments were rewritten to explain intent rather than restate the boolean terms.

## Testing
- `ninja dragonfly` builds clean
- `dragonfly_test`: 45 passed, 1 skipped (unrelated)
- `pre-commit` clang-format passes

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1005,9 +1005,10 @@ unsigned Connection::GetSendWaitTimeSec() const {
   return 0;
 }
 
-void Connection::FlushReplies() {  // NOLINT must not be const due to flush side effect
+std::error_code Connection::FlushReplies() {  // NOLINT must not be const due to flush side effect
   DCHECK(reply_builder_);
   reply_builder_->Flush();
+  return reply_builder_->GetError();
 }
 
 std::string FormatClientInfo(const ClientInfo& ci) {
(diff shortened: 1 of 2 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -2984,12 +2979,36 @@ void Connection::MaybeEnableRecvMultishot() {
...
+  // administrative commands (CONFIG SET, etc.) can execute and relieve backpressure.
+  return parsed_cmd_q_len_ > 0 && GetQueueBackpressure().IsPipelineBufferOverLimit(
+                                      GetLocalConnStats().pipeline_queue_bytes, parsed_cmd_q_len_);
+}
+
+void Connection::NotifyIfMemReleased(size_t bytes_before) {
+  // Executing and replying to commands frees up memory. Because those internal functions only
+  // wake up this specific connection, we manually notify other connections on this thread that
+  // there is now room to resume.
+  if (GetLocalConnStats().pipeline_queue_bytes < bytes_before)
```
vyavdoshenko, line 2993: This re-reads thread-local stats/backpressure after `bytes_before` was captured by the caller. `ParseLoop()` / `ExecuteBatch()` can synchronously migrate the connection, so after migration this may compare source-thread `bytes_before` against destination-thread stats and notify destination waiters while source-thread backpressure waiters remain parked. Consider keeping the captured `ConnectionStats` / `QueueBackpressure` pair with the before value, or explicitly notify the source thread when migration moves queued pipeline memory off it.

vyavdoshenko (dismissed):
lgtm

src/facade/dragonfly_connection.cc
romange: while in theory it's true, in practice we never issue migration command in pipeline mode. It's a control command issued by a Dragonfly replica during handshaking and there is no pipelining involved. I agree that in general case it is broken (always was btw)

vyavdoshenko (approved):

PR #7618, opened 2026-06-16
refactor(facade): extract IoLoopV2 data-path into drain/parse helpers

romange:
## Summary
Pull the inline data-path blocks of `Connection::IoLoopV2` into two helpers and collapse the parse/backpressure logic into two clear branches. No behavior change.

Stacked on #7623 (`refactor/ioloopv2-extract-methods`) — that PR should merge first.

## New helpers
| Helper | Contract |
|---|---|
| `DrainQueuedCommands()` | execute ready commands + reply to free pipeline memory without growing the queue (no new input, or backpressure) |
| `ParkOnBackpressure()` | park while still over the pipeline memory limit, subscribing to global memory-relief notifications |

The base (#7623) already provides `ShouldWakeIdle()`, `DrainControlPath()`, and `RunParsePath()`.

## Data path — two branches
- **No input or over limit** → `DrainQueuedCommands()`, then `ParkOnBackpressure()` only when over the limit
- **Input, under budget** → `RunParsePath()`

Gating the park on `over_limit` makes the empty-input case explicit. It stays equivalent to the old code, where `ParkOnBackpressure()` early-returns when under the limit.

## Read socket ASAP
After the idle park wakes on a recv notification (`pending_input_` set but data not yet in `io_buf_`), the loop restarts so `ReadPendingInput()` pulls the data before the data path runs on an empty buffer. Guarded by an empty `io_buf_` so buffered input is never skipped and a full buffer never busy-spins.

## Testing
- `dragonfly_test`: 45 passed, 1 skipped
- `connection_test.py --df enable_resp_io_loop_v2=true`: pipelining, pubsub control path, squashing, migration, and `test_pipeline_overlimit` / `test_pipeline_backpressure_v2_correctness` all pass
- clang-format clean

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -3016,12 +3016,13 @@ bool Connection::ShouldWakeIdle() const {
 }
 
 bool Connection::DrainControlPath(uint32_t quota) {
+  DCHECK(!dispatch_q_.empty());
+
+  // TODO: ProcessControlMessages should be fused into here.
+
   // Bounded quota prevents the control path from starving the data path: under a PubSub flood
   // dispatch_q_ can accumulate thousands of messages, and without a quota the fiber would drain
   // them all before parsing any socket data. Mirrors V1's async_dispatch_quota mechanism.
-  if (dispatch_q_.empty())
-    return false;
-
   bool quota_reached = ProcessControlMessages(quota);
   GetQueueBackpressure().pubsub_ec.notifyAll();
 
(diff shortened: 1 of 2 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -3144,74 +3163,27 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
...
-      // commands (execute + reply) to free memory, then park until pressure is relieved.
+    bool over_limit = IsOverPipelineLimit();
+    if (io_buf_.InputLen() == 0 || over_limit) {
+      // Drain-only path: either there is no new input to parse, or we are over the pipeline memory
+      // limit (parsing would only grow the queue further). Either way, drain queued commands to
+      // free memory instead of parsing. When over the limit, also park until another connection
+      // relieves the pressure. (Empty queues are never over the limit, so admin commands can still
+      // parse.)
+      DrainQueuedCommands();
+      if (over_limit)
```
glevkovich, line 3181: This is actually better code in my opinion - but please be aware it's not a "pure refactor"  - it changes behaviour. The old code always checked IsOverPipelineLimit() unconditionally after draining (regardless of why the branch was entered, see line 3151 in old code). The new code only calls ParkOnBackpressure (which does its own post-drain recheck) when over_limit is true at entry, intentionally skipping the park for connections that entered due to an empty input buffer.

src/facade/dragonfly_connection.h
```diff
@@ -484,15 +484,23 @@ class Connection : public util::Connection {
   // ready to run.
   bool ShouldWakeIdle() const;
 
-  // IoLoopV2 control path: drains dispatch_q_ up to `quota`. Returns true if the caller should
-  // restart the loop (so freshly arrived socket data is read before the data path runs), false to
-  // fall through to the data path.
+  // IoLoopV2 control path: drains dispatch_q_, processing up to `quota` control messages.
+  // Expects dispatch_q_ to be non-empty. Returns true if dispatch_q_ was depleted, false if the
+  // quota was reached first.
```
glevkovich: This is inaccurate since ProcessControlMessages returns false if  MigrationRequestMessage arrives  and it needs to stop.
In that case DrainControlPath will return true as well. This should be added to documentation.
Or maybe you should change the documentation: "return true means quota was not reached"

src/facade/dragonfly_connection.cc
```diff
@@ -3069,6 +3041,57 @@ Connection::ParserStatus Connection::RunParsePath() {
...
+  if (parsed_head_) {
+    if (HasCommandToExecute())
+      ExecuteBatch();
+    ReplyBatch();
+  }
+
+  NotifyIfMemReleased(mem_before);
+}
+
+void Connection::ParkOnBackpressure(util::fb2::detail::Waiter* bp_waiter) {
```
glevkovich: nit: consider changing `bp_waiter` to exact name on header for better clarity and making it easier to search:
`backpressure_waiter`

glevkovich (dismissed):
LGTM with few minor comments.

src/facade/dragonfly_connection.h
romange: ok, I did not know this. I must say I spent unreasonable amount of time to make the contract clear and it's still confusing.

src/facade/dragonfly_connection.cc
romange: but if it was not over_limit before, can it be over_limit now?

Actually, it can  since it's thread-local, but not because this connection contributed additional load.
In any case, I am not sure if it matters and if we find I am wrong we will have to add a test for this.

glevkovich (approved):

PR #7623, opened 2026-06-17
refactor(facade): extract IoLoopV2 phases into per-concern methods

romange:
## Summary
Pull the inline blocks of the `Connection::IoLoopV2` loop into named, single-purpose helpers so the loop body reads as one step per concern. No behavior change.

## Changes
- `ShouldWakeIdle()` — idle-park wake predicate; collapses the nested `if (io_buf empty) { if (!should_wake) ... }` into one guard.
- `DrainControlPath(quota)` — drains `dispatch_q_`; returns `true` if the caller should restart the loop (so freshly arrived socket data is read before the data path runs).
- `RunParsePath()` — the under-budget parse / execute / reply branch of the data path.

First of a stacked series; #7622 builds on top with the data-path restructure and the read-ASAP optimization.

## Testing
- `ninja dragonfly` builds clean
- `dragonfly_test`: 45 passed, 1 skipped
- clang-format clean

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -3119,9 +3119,6 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
       continue;
     }
 
-    // Handle Parsed Commands Queue (Data Path)
-    auto& conn_stats = GetLocalConnStats();
-
     // Only parse data if we are under the memory limit (backpressure).
     // Exception: If the queue is empty, we always parse to allow admin commands
     // (like CONFIG SET) to run so they can fix the memory limits if needed.
@@ -3132,6 +3129,9 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
       // Do NOT parse - that would grow the queue further. Instead, drain already-queued
       // commands (execute + reply) to free memory, then park until pressure is relieved.
       parse_status = NEED_MORE;
+
+      // Handle Parsed Commands Queue (Data Path)
+      auto& conn_stats = GetLocalConnStats();
 
       size_t mem_before = conn_stats.pipeline_queue_bytes;
 
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -3006,6 +3006,40 @@ bool Connection::HasControlEvent() const {
          IsReadyToMigrate();
 }
 
+bool Connection::ShouldWakeIdle() const {
+  // TODO: optimize CanReply with looking up waiter key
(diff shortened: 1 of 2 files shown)


src/facade/dragonfly_connection.cc
glevkovich: This  should be probably moved to the else under line 3130 to narrow the scope, since you've changed the code at 3129 and conn_stats is now not being used.

romange: fixed

src/facade/dragonfly_connection.h
```diff
@@ -480,6 +480,19 @@ class Connection : public util::Connection {
...
   bool HasControlEvent() const;
 
+  // Wake condition for the idle park: HasControlEvent() plus incoming data and a head command
+  // ready to run.
+  bool ShouldWakeIdle() const;
+
+  // IoLoopV2 control path: drains dispatch_q_ up to `quota`. Returns true if the caller should
+  // restart the loop (so freshly arrived socket data is read before the data path runs), false to
+  // fall through to the data path.
+  bool DrainControlPath(uint32_t quota);
```
glevkovich, line 490: nit: I think the name is misleading. drain means "to remove all liquid". If it returned true, means not fully drained. so I think  a better name would be ProcessControlPath.

src/facade/dragonfly_connection.cc
```diff
@@ -3058,64 +3092,31 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
...
-               HasControlEvent();
-      };
-
-      // Only flush and park if the fiber is truly idle. When synchronous commands
-      // (e.g. PUBLISH) are pipelined, ExecuteBatch processes one at a time and loops
-      // back here with HasCommandToExecute() == true. Skipping the flush lets the
-      // entire pipeline execute in-memory before a single sendmsg at the end.
-      if (!should_wake()) {
-        phase_ = READ_SOCKET;
+    // Idle park: flush and sleep when there is nothing to do (ShouldWakeIdle is false only when
```
glevkovich, line 3095: remove or fix the comment in braces:
ShouldWakeIdle is false if this is false:
`io_buf_.InputLen() > 0 || pending_input_ || HasCommandToExecute() || HasControlEvent();`
So it's not only when the read buffer is empty, more conditions need to be false for this function to return false.

glevkovich (approved):
Approved with few small comments.

romange:
I will fix these in the next PR.

src/facade/dragonfly_connection.h
romange: I will check

PR #7624, opened 2026-06-17
feat(tiering): Skip offloading keys with TTL below minimum threshold

mkaruza:
Avoid wasting disk I/O by offloading keys that are about to expire soon. Introduces
a new flag `tiered_min_ttl_to_offload_ms` (default 5000ms) that sets the minimum
remaining TTL a key must have to be eligible for tiered storage offloading.

### DIFF
--- a/src/server/tiered_storage.cc
+++ b/src/server/tiered_storage.cc
@@ -64,6 +64,9 @@ ABSL_FLAG(bool, tiered_experimental_hash_support, false, "Experimental hash data
 
 ABSL_FLAG(bool, tiered_experimental_list_support, false, "Experimental list node offloading");
 
+ABSL_FLAG(uint32, tiered_min_ttl_to_offload_ms, 5000,
+          "Min remaining TTL in ms for a value to be eligible for offloading");
+
 namespace dfly {
 
 using namespace std;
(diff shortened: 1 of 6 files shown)


dranikpg (commented):
This breaks in a way the premise that we try to not overload the memory. I would suggest somethign different - let's skip offloading of soon-to-expire entries UNLESS we are low on memory and have to provide backpressure (500ms is still a lot)

src/server/tiered_storage.cc
```diff
@@ -64,6 +64,9 @@ ABSL_FLAG(bool, tiered_experimental_hash_support, false, "Experimental hash data
 
 ABSL_FLAG(bool, tiered_experimental_list_support, false, "Experimental list node offloading");
 
+ABSL_FLAG(uint32, tiered_min_ttl_to_offload_ms, 5000,
+          "Min remaining TTL in ms for a value to be eligible for offloading");
```
dranikpg, line 68: thats a 5s default value

src/server/tiered_storage.cc
```diff
@@ -838,17 +852,19 @@ TieredCoolRecord* TieredStorage::PopCool() {
   return &res;
 }
 
-void StashPrimeValue(DbIndex dbid, std::string_view key, PrimeValue* pv, TieredStorage* ts,
-                     BackPressureFuture* backpressure) {
-  if (auto blobs = ts->ShouldStash(*pv); blobs) {
+void StashPrimeValue(DbIndex dbid, std::string_view key, PrimeKey* pk, PrimeValue* pv,
```
dranikpg: PrimeKey could've been a constref

dranikpg (dismissed):
small nits

dranikpg:
@copilot+claude-opus-4.7 

can you make PrimeKey* a constref where I commented (where it was added) as it's only read from

dranikpg (approved):

PR #7625, opened 2026-06-17
Track dispatch_waiting_count_ and rename ok_backend metrics

romange:
## Summary
Groundwork for V2 pipeline dispatch, with no behavioral squashing change.

## Changes
- **Connection**: add `dispatch_waiting_count_` — the number of parsed-but-not-yet-dispatched commands in the run `[parsed_to_execute_, parsed_tail_]`, maintained incrementally (incremented on enqueue, decremented as `parsed_to_execute_` advances). Add the `AdvanceToExecute()` helper and route `parsed_to_execute_` advances through it.
- **Connection**: track `pipelined_wait_latency` (the parse→dispatch wait) per command on the V2 path.
- **ok_backend**: rename the exported Prometheus metrics with a `dragonfly_` prefix to be compatible with the local monitoring stack.

## Testing
`dragonfly` and `ok_backend` build clean; clang-format passes.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1782,6 +1782,8 @@ void Connection::SquashPipeline() {
     ReleasePipelinedCommand(current);
     AdvanceParsedHead(next);
   }
+  DCHECK_GE(dispatch_waiting_count_, squashed);
+  dispatch_waiting_count_ -= squashed;  // the squashed run was waiting; it is now fully handled
   parsed_to_execute_ = parsed_head_;
 
   local_stats_.cmds += squashed;
@@ -1840,6 +1842,7 @@ void Connection::ClearPipelinedMessages() {
   DCHECK_EQ(parsed_cmd_q_bytes_, 0u);
   parsed_tail_ = nullptr;
   parsed_to_execute_ = nullptr;
+  dispatch_waiting_count_ = 0;
 
   QueueBackpressure& qbp = GetQueueBackpressure();
   qbp.NotifyPipelineWaiters();
@@ -1914,8 +1917,7 @@ bool Connection::ProcessAdminMessage(MessageHandle* msg, AsyncOperations* async_
 void Connection::ProcessPipelineCommandV1() {
   DCHECK(parsed_head_ && parsed_to_execute_) << DebugInfo();
   auto* cmd = parsed_to_execute_;
-  parsed_to_execute_ = cmd->next;
-  AdvanceParsedHead(parsed_to_execute_);
+  AdvanceParsedHead(AdvanceToExecute());
 
   tl_facade_stats->conn_stats.pipelined_wait_latency +=
       CycleClock::ToUsec(CycleClock::Now() - cmd->parsed_cycle);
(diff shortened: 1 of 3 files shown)


src/facade/dragonfly_connection.h
```diff
@@ -521,6 +521,14 @@ class Connection : public util::Connection {
     }
   }
 
+  // Advance parsed_to_execute_ past the command it currently points at (now dispatched), keeping
+  // dispatch_waiting_count_ in sync. Returns the new parsed_to_execute_.
+  ParsedCommand* AdvanceToExecute() {
+    --dispatch_waiting_count_;
```
glevkovich: Add guard:
`DCHECK_GT(dispatch_waiting_count_, 0u);`

src/facade/dragonfly_connection.cc
```diff
@@ -1782,6 +1782,8 @@ void Connection::SquashPipeline() {
     ReleasePipelinedCommand(current);
     AdvanceParsedHead(next);
   }
+  DCHECK_GE(dispatch_waiting_count_, squashed);
+  dispatch_waiting_count_ -= squashed;  // the squashed run was waiting; it is now fully handled
```
glevkovich, line 1786: Isn't it dangerous and might corrupt size_t dispatch_waiting_count_ if we run this on V2 release (move to squashing)? Not sure how the DCHECK_GE will help us in that case.

In release builds, DCHECK is stripped, so the subtraction will (silently) underflow the unsigned size_t.

SquashPipeline() relies on an invariant (that is never DCHECKed when entering to this function): parsed_to_execute_ == parsed_head_ (since it is sync, it has zero in-flight async commands). V2 breaks this because parsed_to_execute_ moves ahead during dispatch.

So might have real cases where squashed > dispatch_waiting_count_. And also if we run this with V2 parse_to_Execute should be updated.

As long as the logic doesn't change I think at least you should add at the top of the function:

`DCHECK_EQ(parsed_to_execute_, parsed_head_);`

To make sure this runs only in V1.

src/facade/ok_main.cc
```diff
@@ -425,92 +425,96 @@ void HandleMetrics(ProactorPool* pool, const util::http::QueryArgs&, util::HttpC
...
-  APPEND_BODY("pipeline_dispatch_flush_seconds ", conn.pipeline_dispatch_flush_usec * 1e-6, "\n");
+  APPEND_BODY(
+      "# HELP dragonfly_pipeline_dispatch_flush_seconds Pipeline dispatch flush duration\n");
+  APPEND_BODY("# TYPE dragonfly_pipeline_dispatch_flush_seconds counter\n");
+  APPEND_BODY("dragonfly_pipeline_dispatch_flush_seconds ",
+              conn.pipeline_dispatch_flush_usec * 1e-6, "\n");
 
-  APPEND_BODY("# TYPE pipeline_dispatch_flush_total counter\n");
-  APPEND_BODY("pipeline_dispatch_flush_total ", conn.pipeline_dispatch_flush_count, "\n");
+  APPEND_BODY("# TYPE dragonfly_pipeline_dispatch_flush_total counter\n");
```
glevkovich, line 486: you forgot here "# HELP" line.

src/facade/dragonfly_connection.h
glevkovich, line 527: nit: 
Yes, it saves code and is documented in the header, but consider taking this out of the function:
`--dispatch_waiting_count_;`
when someone reads the code, they are unaware of this side effect, reader expect pure pointer manipulation helper.

src/facade/dragonfly_connection.cc
romange: no, it should support V2, that's the whole point. You can not squash more than number of commands waiting to be squashed. If there is a bug - we should fix it. So it's great that we have this DCHECK. We cover v2 in our tests so we will see if it fails.

src/facade/dragonfly_connection.cc
```diff
@@ -2802,12 +2817,23 @@ void Connection::DestroyParsedQueue() {
...
   CHECK_EQ(parsed_cmd_q_bytes_, 0u);
   delete parsed_cmd_;
   parsed_cmd_ = nullptr;
 }
 
+ParsedCommand* Connection::AdvanceToExecute() {
+  DCHECK(parsed_to_execute_ != nullptr);
+  DCHECK_GT(dispatch_waiting_count_, 0u);
+
+  --dispatch_waiting_count_;
```
glevkovich, line 2832: nit: 
consider taking dispatch_waiting_count_ out. It's a side effect in a pure pointer manipulation helper.
Reader might not be aware of that and it makes harder to understand code.

src/facade/dragonfly_connection.h
romange: I do not understand the point of duplicating the logic. The whole point of this function is to wrap this inside.

src/facade/ok_main.cc
romange: I will remove all the help statements in the next commit.

src/facade/dragonfly_connection.cc
romange: i disagree.

src/facade/dragonfly_connection.cc
glevkovich, line 1747: With your change wouldn't it be more accurate to calculate:
` unsigned pipeline_count = std::min<uint32_t>(dispatch_waiting_count_, pipeline_squash_limit_cached);`
?

`parsed_cmd_q_len_` is total number of commands in the queue, in V2 some of them completed and some are in-flights.
You want to squash maximum `dispatch_waiting_count_` ( what left between `parse_to_execute_` -> to tail).
it will still work like this since `DispatchSquashedBatch` walks the list, but semantically its wrong.

romange: What change you are referring to? 
Please note that `SquashPipeline` is only relevant for V1 and I have not implemented squashing for V2 yet. This will come in the next PR

romange: `dispatch_waiting_count_ == parsed_cmd_q_len_` in V1

glevkovich: OK I think misunderstanding on my side then...I'll drop it.
I thought some of this code logic was ready for V2 but this is not the PR purpose.

glevkovich (approved):
LGTM

src/facade/dragonfly_connection.cc
romange: no, it's only to count `dispatch_waiting_count_`. The next PR is the one that adds squashing.

PR #7630, opened 2026-06-17
feat: Add ReplyScope to SquashPipeline

dranikpg:
Send replies from `SquashPipeline` under a ReplyScope to avoid copies and use the zero-copy vectorized IO features of the sink reply builder. Deallocate commands only after all replies were sent

Performance comparision against the main branch for GET benchmark. With 10+ connections and long pipelines (50) the difference in throughput becomes 15% and more

image

image

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1796,21 +1796,22 @@ void Connection::SquashPipeline() {
     return;
   }
 
-  // Send the deferred replies in linked-list order, then release the commands.
-  // Wrap all replies under a single ReplyScope so the reply builder can use vectorized IO
-  // (writev) with zero intermediate copies — string refs stay valid until we release commands.
+  // Send all replies under a ReplyScope before releasing the commands.
+  // This allows the reply builder to flush without copies
   {
     SinkReplyBuilder::ReplyScope scope(reply_builder_.get());
     auto* cmd = parsed_head_;
-    for (unsigned i = 0; i < squashed && cmd; ++i) {
+    for (unsigned i = 0; i < squashed && cmd; i++, cmd = cmd->next) {
       DCHECK(cmd->CanReply());
+      // TODO: Coroutine replies don't preserve lifetimes after running, so pause the scope
+      std::optional<SinkReplyBuilder::ScopePause> pause;
+      if (cmd->IsSuspendedReply())
+        pause.emplace(reply_builder_.get());
       cmd->SendReply();
       conn_stats.pipelined_wait_latency += CycleClock::ToUsec(start - cmd->parsed_cycle);
-      cmd = cmd->next;
     }
   }
 
-  // Release all commands after the scope has flushed — refs into payloads are no longer needed.
   for (unsigned i = 0; i < squashed && parsed_head_; ++i) {
     auto* current = parsed_head_;
     auto* next = current->next;
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1796,17 +1796,23 @@ void Connection::SquashPipeline() {
     return;
   }
 
-  // Send the deferred replies in linked-list order, then release the commands. SendReply() may
-  // preempt, allowing the producer to append new commands, so re-read the next pointer after it
-  // and advance the list incrementally to keep it consistent at any preemption point.
+  // Send the deferred replies in linked-list order, then release the commands.
+  // Wrap all replies under a single ReplyScope so the reply builder can use vectorized IO
+  // (writev) with zero intermediate copies — string refs stay valid until we release commands.
+  {
+    SinkReplyBuilder::ReplyScope scope(reply_builder_.get());
+    auto* cmd = parsed_head_;
+    for (unsigned i = 0; i < squashed && cmd; ++i) {
+      DCHECK(cmd->CanReply());
+      cmd->SendReply();
+      conn_stats.pipelined_wait_latency += CycleClock::ToUsec(start - cmd->parsed_cycle);
+      cmd = cmd->next;
+    }
+  }
+
+  // Release all commands after the scope has flushed — refs into payloads are no longer needed.
   for (unsigned i = 0; i < squashed && parsed_head_; ++i) {
     auto* current = parsed_head_;
-
-    DCHECK(current->CanReply());
-    current->SendReply();
-
-    conn_stats.pipelined_wait_latency += CycleClock::ToUsec(start - current->parsed_cycle);
-
     auto* next = current->next;
     ReleasePipelinedCommand(current);
     AdvanceParsedHead(next);
--- a/src/facade/reply_builder.cc
+++ b/src/facade/reply_builder.cc
@@ -15,6 +15,7 @@
 #include "absl/types/span.h"
 #include "base/cycle_clock.h"
 #include "base/logging.h"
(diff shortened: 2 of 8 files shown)


src/facade/reply_builder.cc
```diff
@@ -344,26 +345,25 @@ void RedisReplyBuilderBase::SendBulkString(std::string_view str) {
...
   } else {
     WriteRef(bs.view());
   }
   WritePieces(kCRLF);
+}
 
-  // Drain all iovecs (including any WriteRef into bs.view()) before this
-  // function returns. ~BorrowedString releases the pin afterward — by then
-  // the source bytes are already in the kernel.
-  Flush();
```
dranikpg, line 366: This was totally misleading and tried to make a normal use-case look like the exception. 

First, all Send functions assume that the lifetime of the passed reference (or reference object like string_view) does not extend beyond the function. This is why there is a `ReplyScope` at the top of the function - and when it drops it calls `FinishScope` that decides on it's own whether to flush or to copy based on it's allowed copy size (all borrowed refs exceeed it, so it would just flush)

dranikpg: I haven't had a deep though when reviewing the PRs, but now we temporarily end up with two functions (const lvalue, rvalue). I think there would have been more elegant solutions

romange: I do not understand. `ReplyScope` is not aware of BorrowedString semantics.

Suppose `BorrowedString` is passed, reply scope says references can be kept, but then BorrowedString goes out of scope and pin is raised.

when ReplyScope finishes, it will reference possible invalid string.
Maybe I broke ReplyScope contract but why removing Flush is valid?

FYI, I assumed that flushing a BorrowedString does not significantly impact performance compared to sending strings themselves.

src/facade/dragonfly_connection.cc
```diff
@@ -1790,17 +1790,24 @@ void Connection::SquashPipeline() {
...
-  // and advance the list incrementally to keep it consistent at any preemption point.
+  // Send all replies under a ReplyScope before releasing the commands.
+  // This allows the reply builder to flush without copies
+  {
+    SinkReplyBuilder::ReplyScope scope(reply_builder_.get());
+    auto* cmd = parsed_head_;
+    for (unsigned i = 0; i < squashed && cmd; i++, cmd = cmd->next) {
+      DCHECK(cmd->CanReply());
+      // TODO: Coroutine replies don't preserve lifetimes after running, so pause the scope
+      std::optional<SinkReplyBuilder::ScopePause> pause;
```
romange, line 1807: Instead of `ScopePause` you can add ReplyScope::Pause() / UnPause().

src/facade/reply_capture.h
```diff
@@ -56,6 +57,10 @@ class CapturingReplyBuilder : public RedisReplyBuilder {
   // Send payload to builder.
   static void Apply(Payload&& pl, SinkReplyBuilder* builder);
```
romange, line 58: There are too many semantics now. Do we still need "r-value" interface?

src/facade/reply_builder.cc
dranikpg: Because you are in full control of the lifetime. I've added a comment at the top that the reply builder accepts only reference types and assumes each of them goes out if scope immediately - it never says "references can be kept" if you never tell it

dranikpg: If you want to break that assumption, use ReplyScopes

dranikpg: > Maybe I broke ReplyScope contract but why removing Flush is valid?

Because there is a ReplyScope inside the function and when ut goes out of scope it will either flush or copy the data

romange: But if there is another replyscope above this function? The scope will persist, right?

romange: I think I can only exend the scoping, I can not disable it.

romange: Maybe replacing `ReplyScope` in this function with `ScopePause` - would work.

src/facade/reply_builder.h
```diff
@@ -197,15 +215,8 @@ class RedisReplyBuilderBase : public SinkReplyBuilder {
...
-  // whose bytes are borrowed from a stable in-memory source (e.g. a CompactObj
-  // LargeString under the read-only invariant). The default impl handles both
-  // raw (iovec ref) and packed (chunked decode into scratch) encodings and
-  // Flushes before returning, so ~BorrowedString releases the pin only after
-  // the source bytes are in the kernel. CapturingReplyBuilder overrides this
-  // to move `bs` into the captured payload, extending the pin's lifetime
-  // through replay.
-  virtual void SendBulkStringBorrowed(cmn::BorrowedString bs);
+  void SendBulkStringBorrowed(const cmn::BorrowedString& bs);
+  virtual void SendBulkStringBorrowed(cmn::BorrowedString&& bs);
```
romange, line 222: @dranikpg  why do we need in this case `virtual void SendBulkStringBorrowed(cmn::BorrowedString&& bs);` with temporary value semantics?


do you agree that this function can not run under any scope?

src/facade/reply_capture.cc
```diff
@@ -56,7 +56,7 @@ void CapturingReplyBuilder::SendBulkString(std::string_view str) {
 
 // Capture the borrow into the payload, extending the pin's lifetime until
 // replay moves it into the real sink where it is parked across the writev.
-void CapturingReplyBuilder::SendBulkStringBorrowed(cmn::BorrowedString bs) {
+void CapturingReplyBuilder::SendBulkStringBorrowed(cmn::BorrowedString&& bs) {
```
romange, line 59: string_family.cc is the only user of this function. I think this signature should be removed 
and the code below should be adjusted to use const ref

```
void Send(StringResult&& res) const {
    if (holds_alternative<std::string>(res))
      return Send(get<std::string>(res));
    if (holds_alternative<cmn::BorrowedString>(res)) {
      // Move the BorrowedString into SendBulkStringBorrowed; the reply builder takes ownership
      // of the borrow (and associated pin) and parks the pin until all the writes complete.
      rb->SendBulkStringBorrowed(std::move(get<cmn::BorrowedString>(res)));
      return;
    }
    auto fut = get<TieredStorage::TResult<std::string>>(std::move(res));
    io::Result<std::string> iores = fut.Get();
    if (iores.has_value())
      Send(*iores);
    else
      Send(iores.error().message());
  }
```

src/facade/reply_builder.h
```diff
@@ -197,15 +215,11 @@ class RedisReplyBuilderBase : public SinkReplyBuilder {
...
-  // raw (iovec ref) and packed (chunked decode into scratch) encodings and
-  // Flushes before returning, so ~BorrowedString releases the pin only after
-  // the source bytes are in the kernel. CapturingReplyBuilder overrides this
-  // to move `bs` into the captured payload, extending the pin's lifetime
-  // through replay.
-  virtual void SendBulkStringBorrowed(cmn::BorrowedString bs);
+  void SendBulkStringBorrowed(const cmn::BorrowedString& bs);
+
+  // The interface exposes only the rvalue function to allow squashing to "steal" the value,
+  // the real builder implmenetation forwards it to the constref version
```
romange:
```suggestion
  // the real builder implementation forwards it to the constref version
```

romange (approved):

PR #7633, opened 2026-06-17
refactor(facade): extract ok_main squasher into EmitBlockingBatch

romange:
## Summary
Behavior-preserving refactor of `OkService::DispatchSquashedBatch` in `ok_main.cc`, splitting the monolithic squasher into reusable pieces so an alternative strategy can be slotted in later.

## Changes
- `GroupSquashableRun()` — pass 1: classify and group a leading run of single-key `SET`/`GET` commands into per-shard buckets (returns processed/active counts).
- `RunShardOps()` — the per-shard `GET`/`SET` op body, run on the owning proactor thread.
- `EmitBlockingBatch()` — pass 2/3: one task per active shard, block the connection fiber until they finish, then store a captured payload per command.
- `DispatchSquashedBatch()` now just groups then calls `EmitBlockingBatch()`.

No functional change — same commands squashed, same replies, same cross-thread cost (one Add + one Dec per active shard).

### DIFF
--- a/src/facade/ok_main.cc
+++ b/src/facade/ok_main.cc
@@ -16,8 +16,10 @@
 #include <coroutine>
 
 #include "absl/container/flat_hash_map.h"
+#include "absl/container/inlined_vector.h"
 #include "absl/strings/match.h"
 #include "absl/strings/str_cat.h"
+#include "absl/types/span.h"
 #include "base/cycle_clock.h"
 #include "base/init.h"
 #include "base/proc_util.h"
@@ -65,6 +67,17 @@ namespace {
 //    Add()/Await() -- which is why g_shards is global, not thread-local.
 //  - `db`: this shard's slice of the keyspace (touched only by the owning thread).
 struct Shard {
+  void Set(string_view key, string_view value) {
+    db.insert_or_assign(string(key), string(value));
+  }
+
+  optional<string> Get(string_view key) const {
+    auto it = db.find(key);
+    if (it == db.end())
+      return nullopt;
+    return it->second;
+  }
+
   unique_ptr<fb2::FiberQueue> queue;
   fb2::Fiber consumer;
   absl::flat_hash_map<string, string> db;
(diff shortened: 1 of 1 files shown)


src/facade/ok_main.cc
```diff
@@ -65,6 +67,17 @@ namespace {
 //    Add()/Await() -- which is why g_shards is global, not thread-local.
 //  - `db`: this shard's slice of the keyspace (touched only by the owning thread).
 struct Shard {
+  void Set(string_view key, string_view value) {
+    db.insert_or_assign(string(key), string(value));
+  }
+
+  optional<string> Get(string_view key) const {
```
kostasrim, line 74: An empty string has the same semantics as an empty optional (just more explicit + the wrapper). I guess we don't allow empty key/values so the optional is kinda redundant here

kostasrim (approved):

src/facade/ok_main.cc
```diff
@@ -214,6 +225,71 @@ AsyncCmd GetAsync(CmdContext* ctx, ProactorPool* pool) {
...
+    unsigned sid = KeyShard(ctx->at(1), num_shards);
+    DCHECK_LT(sid, num_shards);
+    auto& bucket = (*per_shard)[sid];
+
+    if (bucket.size() == kMaxSquashSize)  // this shard is full -> end the run here
+      break;
+    ctx->batch_is_get = is_get;
+
+    // first command for this shard -> one more active shard
+    res.num_active += unsigned(bucket.empty());
```
abhijat, line 272:
```suggestion
    if (bucket.empty())
      res.num_active++;
```

src/facade/ok_main.cc
```diff
@@ -289,79 +370,43 @@ DispatchResult OkService::DispatchCommand([[maybe_unused]] ParsedArgs args, Pars
...
+  EmitBlockingBatch(group.num_active, &per_shard);
+  return group.processed;
+}
+
+// Blocking squasher: dispatch one task per active shard, block the connection fiber until every
+// shard finishes (like dragonfly's squasher blocking on its transaction hops), then store each
+// reply as a captured payload. Resolve() also marks the command deferred; the connection emits the
+// payloads in parse order afterwards, so bucket order here does not matter. per_shard stays valid
+// on this parked fiber's stack while the tasks run, so the shard callbacks reference it by pointer.
+void OkService::EmitBlockingBatch(unsigned num_active_shards, ShardBuckets* per_shard) {
```
abhijat, line 397: can be static

abhijat (approved):
lgtm

small nit, there are many `auto` type inferences in new code here. for consistency with our normal rules the types should be explicit unless nested etc.

src/facade/ok_main.cc
romange: we allow empty values

romange:
> nested

ok, I will look at it at the next PR.

PR #7636, opened 2026-06-18
fix(tiering): Fix QList node cleanup after clearing offloaded state

mkaruza:
Clear the node entry pointer when restoring an offloaded QList node, and guard 
node deletion so cleanup does not free a null or stale entry.

### DIFF
--- a/src/core/qlist.cc
+++ b/src/core/qlist.cc
@@ -1220,7 +1220,8 @@ void QList::DelNode(Node* node) {
     }
   }
 
-  if (!node->offloaded) {
+  if (!node->offloaded && node->entry) {
+    DCHECK(node->entry != nullptr);
     zfree(node->entry);
   }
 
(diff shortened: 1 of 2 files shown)


mkaruza:
@dranikpg 

There is no simple/deterministic reproduction so i'll explain what is this PR fixing:

1. Node is offloaded  (`node->offloaded=1`) and we delete node
2. We call `tiering_params_->cleanup(this, node);` that executed `TieredStorage::Delete` where we set `node->offloaded=0`
3.  Execution of delete continues where it check `if(!node->offloaded) { free(node->entry) }`
4. Since we never invalidated `node->entry \ node->ext_offset` (it's represented as union of these two variables). We call `zfree(node->entry)` that frees pointer of whatever is stored as `ext_offset` and can lead to UB.

dranikpg (approved):
I see you added the same to your other PR

mkaruza:
@dranikpg yes i continued from this branch and created new with another fix. I'll merge in order.

PR #7638, opened 2026-06-18
fix(tiering): Keep tiered list endpoints materialized

mkaruza:
QList endpoint operations assume head and tail nodes are in memory and access
their listpack entries directly. Deleting a node can promote an offloaded or
pending interior node to head or tail, breaking that invariant.

Materialize the current head and tail after node deletion and add a regression
for RPUSH after LTRIM promotes an offloaded node to the tail.


mkaruza:
Should be merged after #7636

dranikpg (approved):

PR #7642, opened 2026-06-18
feat: add /benchmark Claude Code skill

romange:
## Summary

- Adds a `/benchmark` Claude Code skill for driving end-to-end Dragonfly benchmarks against Valkey/Redis on remote cloud instances
- Includes reusable shell scripts (`bench_server.sh`, `monitor_fill.sh`, `capture_result.sh`) and Python scripts (`collect_metrics.py`, `make_charts.py`, `plot_fill.py`, `bench_util.py`)
- Bundled operational reference docs (`running-benchmarks.md`, `report-template.md`) and eval fixtures

## Key operational rules encoded in the skill

- Always use private/VPC-internal IPs for cross-instance traffic
- `--pipeline 30` required for all prefill/write phases (client bottlenecks without it)
- Zero-error enforcement: any eviction/OOM error invalidates the run
- Port 6380 standard for all servers (Dragonfly, Valkey, Redis)
- Valkey/Redis Docker: `--io-threads min(nproc, 8)` always set
- OS-level host memory tracked alongside `used_memory_rss` (they can diverge significantly with mimalloc large pages)

### DIFF
--- a/.claude/skills/benchmark/SKILL.md
+++ b/.claude/skills/benchmark/SKILL.md
@@ -1,0 +1,264 @@
+---
+name: benchmark
+description: >
+  Benchmark Dragonfly (and compare against Valkey/Redis) on local or remote
+  cloud instances, then produce a performance + memory report with charts and a
+  raw-data appendix. Use this whenever the user runs /benchmark, or asks to
+  benchmark / load-test / measure throughput, QPS, latency, or memory efficiency
+  of Dragonfly or another Redis-compatible server, to compare Dragonfly vs
+  Valkey or Redis, or to generate a benchmark report. Trigger it even when the
+  user just describes the goal ("see how fast Dragonfly is on this box",
+  "compare memory per key vs Valkey") without saying the word "benchmark".
+argument-hint: (optional) free-form spec, e.g. "server=10.0.0.1 client=10.0.0.2 strings vs-valkey"
+allowed-tools: Bash, Read, Write, Edit, AskUserQuestion
+---
+
+# Benchmark Dragonfly
+
+Drive a benchmark of Dragonfly (optionally side-by-side with Valkey or Redis) on
+local or remote instances, collect throughput / latency / memory numbers, and
+assemble a report that mirrors the team's published format: a narrative TLDR,
+per-instance metric tables, charts, and an appendix with the exact commands and
+raw output so anyone can reproduce it.
+
+The whole point of the report is **trust**: a reader should be able to see the
+headline numbers, understand the conditions they were measured under, and
+re-run the exact commands themselves. Everything below serves that goal.
+
+## When you start
+
+If the user passed a free-form spec in `$ARGUMENTS`, parse what you can from it
+(IPs, use-cases, comparison targets) and only ask about what's missing. The work
+splits into clear phases — walk through them in order, but stay flexible: the
+user may already have servers running, or may only want the report half (handed
+raw output to format). Read what they're actually asking for.
+
+Phases: (1) gather spec → (2) prepare instances → (3) network tuning →
+(4) start monitoring → (5) run workloads → (6) extract metrics, charts, report.
+
+The two reference files carry the operational detail — read them before acting,
+not from memory:
+
+- `references/running-benchmarks.md` — how to prep instances, start servers,
+  drive `dfly_bench` and `memtier_benchmark`, run Valkey/Redis for comparison,
+  apply network tuning, and measure memory. **Read this before Phase 2.**
+- `references/report-template.md` — the exact report structure, with an example.
+  **Read this before Phase 5.**
+
+Bundled `scripts/` (reuse these instead of re-improvising the same loops — they
+encode the safety discipline learned the hard way):
+
+- `bench_server.sh start|stop|status <ssh> <binary> [maxmem] [port]` — launch a
+  server tracking its PID; **stop kills by PID and verifies the process is gone +
+  port free** (tmux kill-session does not reap it).
+- `monitor_fill.sh <server_ssh> <client_ssh> [session] [abort_pct]` — the live
+  RSS guard: polls `used_memory_rss`, prints a line per sample, aborts the write
+  if RSS crosses the threshold (default 92%). Its output is also the input to
+  `plot_fill.py`.
+- `capture_result.sh <server_ssh> <client_ssh> <log> <outdir> [label]` — saves
+  the run summary + INFO memory + **OS-level `free` snapshot** (`host_memory.txt`)
+  and exits non-zero if there were any eviction errors (enforces the zero-error
+  rule). The OS snapshot matters: `used_memory_rss` and host free can diverge
+  significantly — e.g. with `MIMALLOC_ALLOW_LARGE_OS_PAGES=1` (default), mimalloc
+  holds large-page arenas that the OS counts as used but INFO doesn't fully reflect,
+  making the machine appear near-OOM while RSS looks fine.
+- `collect_metrics.py` / `make_charts.py` — metrics + charts from dfly_bench
+  JSON (GET/SET workloads).
+- `plot_fill.py` — memory-fill chart from `monitor_fill.sh` output, for
+  `--command` workloads (e.g. SETEX) that produce no JSON time series.
+
+Every benchmark output directory must include a `commands_audit.log`
+artifact with the exact command lines used for setup, server launch, writes,
+reads, captures, diagnostics, and chart/report generation. Raw data without the
+commands that produced it is not reproducible.
+
+**Run-control scripts are mandatory for benchmark phases.** For Phase 5 runs,
+use the bundled scripts as the default API:
+
+- start/stop/status servers with `scripts/bench_server.sh`;
+- monitor long write fills with `scripts/monitor_fill.sh`;
+- capture write results and enforce the zero-error rule with
+  `scripts/capture_result.sh`;
+- generate metrics/charts with `scripts/collect_metrics.py`,
+  `scripts/make_charts.py`, and `scripts/plot_fill.py`.
+
+Inline SSH is fine for inspection, package setup, one-off diagnostics, and
+bounded foreground probes, but not as a replacement for the run-control scripts
+in the main benchmark path. If a script is not usable for a run, document the
+reason in `commands_audit.log` and in the report appendix before using a manual
+equivalent.
+
+## Phase 1 — Gather the benchmark spec (interactive)
+
+Use `AskUserQuestion` to pin down anything not already given. You need:
+
+1. **Where it runs.** Local, or remote instances? For remote, get the **server**
+   host/IP and the **client** host/IP (the load generator should run on a
+   separate machine from the server — co-locating them steals CPU and corrupts
+   the numbers). Get the SSH user / key if not obvious. On cloud instances,
+   **always use the private/internal IP** (VPC-internal) for both SSH and as the
+   load generator's `-h` target — the public IP routes through the internet
+   gateway, adding latency and egress cost, and the result measures the gateway
+   rather than the server.
+2. **Use-cases.** Default menu: simple strings (SETEX writes + GET reads),
+   pipelined / peak-throughput, memory efficiency. Always also ask whether there
+   is a **specific focus** (a particular command, data type, key distribution,
+   value size, or scenario) — the canned use-cases are a starting point, not a
+   cage.
+3. **Comparison targets.** Dragonfly only, or also Valkey / Redis? Which
+   versions?
+4. **Instance specs** (for the report's conditions section): instance type, CPU
+   count, OS/kernel. Read these directly from the box at the start of Phase 2
+   rather than asking the user.
+5. **Load parameters** if the user has opinions: value size (`-d`), key space
+   (`--key_maximum`), connections (`-c`), pipeline depth, test duration. Pick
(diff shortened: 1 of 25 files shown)


romange:
augment review

.claude/skills/benchmark/evals/fixtures/RUN_NOTES.txt
```diff
@@ -0,0 +1,24 @@
+Benchmark run notes (raw artifacts in this directory)
+
+Hardware:
+  Server: m7g.2xlarge (8 vCPU) and m7g.4xlarge (16 vCPU), Ubuntu 24.04.2, kernel 6.8.0
+  Client: c7gn.4xlarge (16 vCPU) running dfly_bench
+  Network tuning: IRQ SMP affinity spread across CPUs (applied)
+
+Servers launched on port 6380:
+  Dragonfly: ./dragonfly --conn_use_incoming_cpu --dbfilename= --logtostderr --port 6380
+  Valkey 9.0: docker run --network=host --rm valkey/valkey:9.0 --save "" --appendonly no --io-threads 8 --protected-mode no --port 6380
```
vyavdoshenko, line 10: I think it's time to use Valkey 9.1.

Is it OK to reuse the same port?

.claude/skills/benchmark/scripts/capture_result.sh
```diff
@@ -0,0 +1,44 @@
...
+dbsize=$($SSH "$srv" "redis-cli -p $port dbsize")
+# OS-level host memory — used_memory_rss alone misses allocator arena waste
+# (visible as large discrepancy between INFO rss and OS `free`, e.g. with huge pages).
+read -r host_free host_total < <($SSH "$srv" 'free -b | awk "/^Mem:/{print \$4, \$2}"')
+$SSH "$srv" 'free -h' > "$out/host_memory.txt"
+host_used_pct=$(awk "BEGIN{printf \"%.1f\", ($host_total-$host_free)/$host_total*100}")
+
+summary=$(grep -E 'Total time' "$out/write.log" | tail -1 || true)
+errline=$(grep -iE 'error responses' "$out/write.log" | tail -1 || true)
+errs=$(echo "$errline" | grep -oE '[0-9]+' | head -1)
```
vyavdoshenko, line 30: With set -euo pipefail, clean runs with no "error responses" line will exit here because grep -oE returns non-zero before errs=${errs:-0} runs. Default errs=0 before parsing or guard this extraction with || true / a bash regex conditional.

.claude/skills/benchmark/scripts/bench_server.sh
```diff
@@ -0,0 +1,77 @@
...
+# Shared: build --maxmemory flag (called per-subcommand after args are parsed)
+_mm() { [ -n "${1:-}" ] && echo "--maxmemory $1" || echo ""; }
+
+case "$cmd" in
+  start)
+    bin=${3:?binary path}; maxmem=${4:-}; port=${5:-6380}; env_prefix=${6:-}
+    $SSH "$tgt" "$env_prefix nohup $bin --conn_use_incoming_cpu --dbfilename= $(_mm "$maxmem") \
+        --logtostderr --port $port >/tmp/bench-server.log 2>&1 & echo \$! > /tmp/bench-server.pid; sleep 3; \
+        echo PID=\$(cat /tmp/bench-server.pid); redis-cli -p $port ping; \
+        redis-cli -p $port info server | grep -iE 'dragonfly_version|redis_version' | head -2"
```
vyavdoshenko, line 46: This can report a successful start even if the new process failed to bind and an old server is already serving this port. Please verify the recorded PID/container is alive and that the port is served by the launched process, otherwise benchmarks can run against a stale state.

.claude/skills/benchmark/scripts/bench_server.sh
```diff
@@ -0,0 +1,77 @@
...
+        redis-cli -p $port ping; \
+        redis-cli -p $port info server | grep -iE 'redis_version|valkey_version' | head -2; \
+        redis-cli -p $port config get io-threads"
+    ;;
+  stop)
+    port=${3:-6380}
+    $SSH "$tgt" "kill -9 \$(cat /tmp/bench-server.pid 2>/dev/null) 2>/dev/null || true; sleep 3; \
+        pgrep -af 'dragonfly|valkey-server|redis-server' | grep -v 'pgrep\|bash -c' || echo 'stopped'; \
+        redis-cli -p $port ping 2>&1 || true; free -h | head -2"
+    ;;
```
vyavdoshenko, line 82: Stop currently always exits successfully even if the benchmark port remains occupied. Please fail after stop when redis-cli can still reach the port, so the next run does not silently benchmark a stale server.

.claude/skills/benchmark/scripts/make_charts.py
```diff
@@ -0,0 +1,118 @@
...
+    stats = find_first(data, "ALL STATS") or data
+
+    op_block = None
+    if op_hint:
+        op_block = find_first(stats, op_hint)
+    if op_block is None:
+        for name, blk in stats.items():
+            if isinstance(blk, dict) and find_first(blk, "Time-Serie") is not None:
+                op_block = blk
+                break
```
vyavdoshenko: When --op is omitted, this picks the first operation with a Time-Serie. dfly_bench writes Sets before Gets, so read-only runs can plot the zero-count Sets series instead of Gets; choose the first non-zero operation or infer it from configuration.ratio.

.claude/skills/benchmark/SKILL.md
```diff
@@ -0,0 +1,264 @@
...
+# when benchmarking Valkey/Redis, also bring up the redis exporter:
+ssh <server> 'cd ~/projects/dragonfly/tools/local/monitoring && docker compose --profile redis up -d'
+```
+
+Prometheus scrapes at 1s. **Always run servers on port 6380** (Dragonfly,
+Valkey, and Redis alike). The bundled `prometheus.yml` ships pointing at `6379`,
+so before bringing the stack up, fix the scrape targets:
+
+```bash
+sed -i 's/:6379/:6380/g' tools/local/monitoring/prometheus/prometheus.yml
```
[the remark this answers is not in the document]
vyavdoshenko: This sed command runs locally, but the monitoring stack is started on the remote server. As written, Prometheus on the server keeps scraping 6379; run the edit and Prometheus restart inside the server SSH command.

.claude/skills/benchmark/references/running-benchmarks.md
```diff
@@ -0,0 +1,552 @@
...
+**Write phase** (load the keyspace):
+
+```bash
+./dfly_bench -h <server> -p 6380 --qps=0 -d 64 --key_maximum=$KMAX -c $CONN \
+    --pipeline 30 -n $N --json_out_file df_write.json --command "setex __key__ 10000 __data__"
+```
+
+Note: with `--command`, `--json_out_file` is ignored — for the canonical strings
+test use the built-in SET path (`--ratio 1:0`) instead if you need the JSON/time
+series, and reserve `--command "setex ..."` for when TTLs matter. For SETEX,
```
vyavdoshenko: Should be:

The canonical strings test uses the built-in SET path (`--ratio 1:0`) so it
produces the JSON/time series the charts need. If TTLs matter, swap in
`--command "setex __key__ 10000 __data__"` — but note `--json_out_file` is then
ignored, so for SETEX

.claude/skills/benchmark/evals/fixtures/RUN_NOTES.txt
```diff
@@ -0,0 +1,24 @@
...
+
+Workload: simple strings, 64-byte values.
+  Write: ./dfly_bench -p 6380 --qps=0 -d 64 --key_maximum=$KMAX -c 50 --ratio 1:0 --json_out_file <f>
+  Read:  ./dfly_bench -p 6380 --qps=0 -d 64 --key_maximum=$KMAX -c 50 --ratio 0:1 --json_out_file <f>
```
vyavdoshenko, line 14: Should be: -c 25

```
  Write: ./dfly_bench -p 6380 --qps=0 -d 64 --key_maximum=$KMAX -c 25 --ratio 1:0 --json_out_file <f>
  Read:  ./dfly_bench -p 6380 --qps=0 -d 64 --key_maximum=$KMAX -c 25 --ratio 0:1 --json_out_file <f>
```

vyavdoshenko (dismissed):

.claude/skills/benchmark/evals/fixtures/RUN_NOTES.txt
romange: We run it one after another.

.claude/skills/benchmark/SKILL.md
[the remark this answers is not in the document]
romange: No, it runs sed there. Basically the assumption is that hosts involved have dragonfly repo checked out.

The agent checks if it's the case and if not it clones it.

.claude/skills/benchmark/scripts/capture_result.sh
romange: it worked. I will check again empirically.

.claude/skills/benchmark/scripts/bench_server.sh
romange: Fixed in 00b11dfc: `start` now checks `kill -0 \$pid` after launch (and `start-docker` checks the container is running) and exits non-zero if the process died, so a stale server holding the port no longer reads as a successful start.

.claude/skills/benchmark/scripts/bench_server.sh
romange: Fixed in 00b11dfc: `stop`/`stop-docker` now `redis-cli ping` the port after stopping and exit non-zero if it still responds. `stop` also only kills the recorded PID when it is actually a server process, to guard against a stale/reused PID file.

.claude/skills/benchmark/scripts/make_charts.py
romange: Fixed in 00b11dfc: when `--op` is omitted, `make_charts.py` now picks the op with the highest total Count instead of the first one, so read-only runs plot Gets rather than the zero-count Sets series. Verified against the read fixture.

.claude/skills/benchmark/references/running-benchmarks.md
romange: Adopted in 00b11dfc — running-benchmarks.md now leads with the canonical `--ratio 1:0` SET path (JSON/time-series producing) and presents `--command "setex ..."` as the TTL caveat, using your suggested wording. report-template.md updated to match.

.claude/skills/benchmark/scripts/plot_fill.py
```diff
@@ -0,0 +1,82 @@
...
+        t, db, rss = parse(path, args.interval)
+        if not t:
+            print(f"warning: no poll lines in {path}", file=sys.stderr)
+            continue
+        m = markers[i % len(markers)]
+        ax1.plot(t, rss, m + "-", label=label, lw=2)
+        ax2.plot(db, rss, m + "-", label=label, lw=2)
+
+    if args.ram_gib:
+        ax1.axhline(args.ram_gib * 0.92, ls="--", color="r", alpha=0.6, label="92% RSS guard")
```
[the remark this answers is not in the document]
romange: Fixed in 00b11dfc: `plot_fill.py` now takes `--abort-pct` (default 92) so the guard line matches the threshold actually used by `monitor_fill.sh`.

.claude/skills/benchmark/scripts/monitor_fill.sh
```diff
@@ -0,0 +1,49 @@
...
+  host_free_g=$(awk "BEGIN{printf \"%.2f\", $host_free/2^30}")
+  echo "poll $i: $state dbsize=$db rss=${rssg}GiB (${pct}%) host_used=${host_used_g}GiB host_free=${host_free_g}GiB"
+  if [ "$pct" -ge "$abort" ]; then
+    echo "!! ABORT: RSS ${pct}% >= ${abort}% — killing write"
+    $SSH "$cli" "tmux kill-session -t $sess 2>/dev/null" || true
+    exit 3
+  fi
+  [ "$state" = "DONE" ] && { echo "write finished"; exit 0; }
+done
+echo "monitor: max polls reached, write still running"
```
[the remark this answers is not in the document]
romange: Addressed: `monitor_fill.sh` already returns `exit 2` on max polls, and 00b11dfc adds numeric validation of the RSS/total reads so a dropped SSH/redis-cli read fails closed (exit 4) instead of silently skipping the abort comparison.

.claude/skills/benchmark/evals/fixtures/RUN_NOTES.txt
romange: These notes document the actual runs that produced the bundled fixtures, which used `-c 50`. @vyavdoshenko do you mean the runs should be re-done at `-c 25`, or that the notes mis-record the connection count? Holding off on editing the recorded value until confirmed.

romange: It's fixtures of past runs and the numbers there are used for tests and evaluations.

vyavdoshenko: I mentioned it only because it differs from other settings.

vyavdoshenko (approved):

Issue #7579, opened 2026-06-12
Cuckoo family


kostasrim:
Base issue for Cuckoo Family

- [x] Add cuckoo core data structure
- [x] Add RDB serialization
- [x] Integrate CuckooFilter into CompactObj 
- [x] cuckoo family #5446
- [x] update docs

Follow ups:

- [ ] do we need defragmentation

PR #7646, opened 2026-06-18
feat: add cuckoo_family.cc and cf.reserve  

kostasrim:
* add cf.reserve command
* add tests

First steps of the family for #7579

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -1150,17 +1150,11 @@ TOPK* CompactObj::GetTOPK() const {
   return u_.topk;
 }
 
-void CompactObj::SetCuckooFilter(uint64_t capacity, uint8_t slots_per_bucket,
-                                 uint16_t max_iterations, uint16_t expansion) {
+void CompactObj::SetCuckooFilter(const CuckooFilterOptions& options) {
   // Allocate before touching taglen_/u_: if constructor throws (e.g. bad_alloc)
   // this object must stay in its previous, valid state.
-  CuckooFilter* cf =
-      AllocateMR<CuckooFilter>(capacity, tl.local_mr, slots_per_bucket, max_iterations, expansion);
-  if (taglen_ == CUCKOO_FILTER_TAG) {
-    DeleteMR<CuckooFilter>(u_.cuckoo_filter);
-  } else {
-    SetMeta(CUCKOO_FILTER_TAG);
-  }
+  CuckooFilter* cf = AllocateMR<CuckooFilter>(options, tl.local_mr);
+  SetMeta(CUCKOO_FILTER_TAG);  // frees the previous value, whatever tag it had
   u_.cuckoo_filter = cf;
 }
 
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -1152,14 +1152,16 @@ TOPK* CompactObj::GetTOPK() const {
 
 void CompactObj::SetCuckooFilter(uint64_t capacity, uint8_t slots_per_bucket,
                                  uint16_t max_iterations, uint16_t expansion) {
+  // Allocate before touching taglen_/u_: if constructor throws (e.g. bad_alloc)
+  // this object must stay in its previous, valid state.
+  CuckooFilter* cf =
+      AllocateMR<CuckooFilter>(capacity, tl.local_mr, slots_per_bucket, max_iterations, expansion);
   if (taglen_ == CUCKOO_FILTER_TAG) {
-    *u_.cuckoo_filter =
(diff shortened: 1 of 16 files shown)


src/core/compact_object_test.cc
```diff
@@ -599,6 +600,57 @@ TEST_F(CompactObjectTest, CuckooFilter) {
...
+  uint64_t hash = CuckooFilter::Hash("still-works");
+  EXPECT_TRUE(cobj_.GetCuckooFilter()->Insert(hash));
+  EXPECT_TRUE(cobj_.GetCuckooFilter()->Exists(hash));
+}
+
+TEST_F(CompactObjectTest, AllocateMRDoesNotLeakOnConstructorThrow) {
+  // AllocateMR<T> does: ptr = allocate(sizeof(T)); return new (ptr) T(args). If T's
+  // constructor throws, ptr must be freed before the exception propagates.
+  // For example: bloom filter constructor can throw if its std::vector allocation fails.
+  // DENY_OOM won't help there.
```
kostasrim: @dranikpg PTAL. Do not review the PR yet but plz review what I write here:

The general issue: some data types (sbf, cuckoo) etc can throw in the constructor. For those cases we don't catch the exception and `AllocateMr` leaves the scope and the allocated block: ` void* ptr = memory_resource()->allocate(sizeof(T), alignof(T));` sadly leaks because by the time we catch the exception we already left the function scope.

This test uncovered it + the fix is a try/catch block within AllocateMr. Straightforward.

Now a `second` issue is the following:

A command creates a new key which is inserted via `AddOrFind()` (is_new=true) yet `AllocateMR` throws. We catch it but now we are also `out of command scope` which means that we have a `zombie` compact object in the table.

This is not something we intended right ?

kostasrim:
augment review

src/server/cuckoo_filter_family.cc
```diff
@@ -0,0 +1,94 @@
...
+
+namespace dfly {
+
+using namespace facade;
+using namespace std;
+
+namespace {
+
+OpStatus OpReserve(const OpArgs& op_args, string_view key, uint64_t capacity, uint8_t bucket_size,
+                   uint16_t max_iterations, uint16_t expansion) {
```
abhijat: capacity, bucket size, max iterations and expansion appear together multiple times, it might be good to have a params struct to group them together.

src/core/compact_object.cc
```diff
@@ -1152,14 +1152,16 @@ TOPK* CompactObj::GetTOPK() const {
...
 void CompactObj::SetCuckooFilter(uint64_t capacity, uint8_t slots_per_bucket,
                                  uint16_t max_iterations, uint16_t expansion) {
+  // Allocate before touching taglen_/u_: if constructor throws (e.g. bad_alloc)
+  // this object must stay in its previous, valid state.
+  CuckooFilter* cf =
+      AllocateMR<CuckooFilter>(capacity, tl.local_mr, slots_per_bucket, max_iterations, expansion);
   if (taglen_ == CUCKOO_FILTER_TAG) {
-    *u_.cuckoo_filter =
-        CuckooFilter(capacity, tl.local_mr, slots_per_bucket, max_iterations, expansion);
+    DeleteMR<CuckooFilter>(u_.cuckoo_filter);
```
abhijat: why not let `SetMeta(...)` alone do this without the condition check, it also calls `DeleteMR<CuckooFilter>(u_.cuckoo_filter)` if `taglen_ == CUCKOO_FILTER_TAG`, so the end result should be the same?

src/server/rdb_extensions.h
```diff
@@ -16,12 +16,13 @@ constexpr uint8_t RDB_TYPE_SBF = 33;
...
 constexpr uint8_t RDB_TYPE_CMS = 35;
 constexpr uint8_t RDB_TYPE_TOPK = 36;
+constexpr uint8_t RDB_TYPE_CUCKOO = 37;
 
 constexpr bool rdbIsObjectTypeDF(uint8_t type) {
   return __rdbIsObjectType(type) || (type == RDB_TYPE_JSON) ||
          (type == RDB_TYPE_HASH_WITH_EXPIRY) || (type == RDB_TYPE_SET_WITH_EXPIRY) ||
          (type == RDB_TYPE_SBF) || (type == RDB_TYPE_SBF2) || (type == RDB_TYPE_CMS) ||
-         (type == RDB_TYPE_TOPK);
+         (type == RDB_TYPE_TOPK) || (type == RDB_TYPE_CUCKOO);
```
kostasrim, line 25: I need this cause fuzzer will crash it 😄

// to be continued 😄

abhijat (approved):

PR #7648, opened 2026-06-18
feat: Inline buffer for captured replies

dranikpg:
Results: +8% in best case

image

1. Reduce size of CommandContext to get more free space
2. Allow using storage_ from backed arguments to save reply strings to reduce allocations

### DIFF
--- a/src/common/backed_args.h
+++ b/src/common/backed_args.h
@@ -14,7 +14,7 @@ namespace cmn {
 
 class BackedArguments {
   constexpr static size_t kLenCap = 5;
-  constexpr static size_t kStorageCap = 88;
+  constexpr static size_t kStorageCap = 128;
 
   constexpr static size_t kShrinkFloor = 64 << 10;  // 64 KiB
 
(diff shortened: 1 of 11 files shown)


src/server/transaction.h
```diff
@@ -608,7 +608,7 @@ class Transaction {
   absl::InlinedVector<LockFp, 4> kv_fp_;
 
   // Stores the full undivided command.
-  CmdArgList full_args_;
+  absl::InlinedVector<std::string_view, 4> full_args_;
```
dranikpg: This is a big tradeoff - transaction still uses `ArgSlice` instead of `ParsedArgs`, so it needs an contiguous array - I moved it here from CommandContext to save 24 bytes there (tail_args_backing_).

This might slow down single commant path

src/common/backed_args.h
```diff
@@ -14,7 +14,7 @@ namespace cmn {
 
 class BackedArguments {
   constexpr static size_t kLenCap = 5;
-  constexpr static size_t kStorageCap = 88;
+  constexpr static size_t kStorageCap = 128;
```
dranikpg, line 17: we got this by removing two fields + growing to hit mi_good_size of 320

src/server/multi_command_squasher.cc
```diff
@@ -239,6 +239,13 @@ OpStatus MultiCommandSquasher::SquashedHopCb(EngineShard* es, RespVersion resp_v
 
     crb.SetReplyMode(dispatched.reply_mode);
 
+    // Allow captured replies to be stored in argument storage.
+    // Some commands might include arguments in replies, so we have a limited set
+    if (dispatched.cmd_cntx && dispatched.cid->SupportsAsync())
+      crb.ProvideInlineBuffer(dispatched.cmd_cntx->GetInlineBuffer());
+    else
+      crb.ProvideInlineBuffer({});  // reset buffer
+
```
dranikpg, line 228: or add some flag to CommandId instead?

src/server/transaction.h
romange: so lets solve it first.

romange: I want to get rid of `tail_args_backing_` as well. Lets do it right

romange (approved):

Issue #5268, opened 2025-06-10
proposal: simplify memory accounting by type


romange:
Currently our memory accounting by type relies on method `CompactObject::MallocUsed` and assumes that every memory change is tracked and properly reflected in per-type stats.

It seems as very fragile and error prone approach where each data type is responsible to track its own usage.

This sometimes even not possible because malloc library can allocate block sizes greater than were requested. In addition, it is very hard to track memory reserved in complicated data types like json where various internal data structures may reserve more than they use and what they account for.

Our goal is to provide insights into per type memory usage but we do not really need byte precision usage.

The critical code is in `AccountObjectMemory` function that uses `CompactObject::MallocUsed` to count per object deltas.

## Proposal

Instead of relying on MallocUsed we can track changes in `EngineShard::UsedMemory()` that  increases or decreases as the underlying objects mutate. If we know the context of the object type that is being added/mutated we could attribute the change to the correct type. There are few caveats that needs to be handled: 
  * possible preemptions due to replication or snapshotting when we measure  "before" and "after" of `EngineShard::UsedMemory()`  
  * cases when we override the existing objects (MOVE, COPY, SET etc)
  * Overflows (negative numbers) should be trimmed to 0 at least in release mode.
  * `EngineShard::UsedMemory()` also include dash table table memory. it should not be accounted for when tracking object memory usage.

## Further enhancements
Could be nice if we track keys related memory usage (i.e strings that are used for keys) separately as `type_used_memory_keys`

romange:
## Design goals - to have a central approach to track memory without writing a complicated code in each data structure.

romange:
but what if we have 

```
cb1();                  // preempts
                                                                        //         cb2();  
cb1();                 // finishes
```

solved by https://github.com/dragonflydb/dragonfly/pull/7459

romange:
```
before = ObjMemoryUsed();
cb();  // LIST? HASH ? JSON?
after = ObjMemoryUsed();
```
where `ObjMemoryUsed := UsedMemory - TableMemory`.

romange:
Key Memory used must still rely on MallocUsed.

abhijat:
One issue with accounting by family is that commands like delete belong to generic family. So for example one family adds the memory and another deducts.

Maybe it is acceptable as a tolerance, or maybe we can track the generic DEL separately (it will be only deductions).

It does not seem possible to account per type at the transaction level, as the transaction would only have the opaque ptr, not what keys / type we touch.

PR #7649, opened 2026-06-19
fix(tiering): Fix tiered list memory accounting on materialized nodes

mkaruza:
When a QList node is offloaded, uploading it back into memory allocates the node 
payload before DbSlice::AutoUpdater recalculates object memory usage. Both mutable
and non-mutable list commands can materialize nodes through the same tiering load
path, so accounting has to be handled at the tiering/QList layer instead of only in individual 
command handlers.

Track bytes accounted by QList uploads while tiering is enabled, then have list mutation 
paths pass that amount into AutoUpdater so it is subtracted from the final object-memory 
delta. This keeps OBJ_LIST metrics aligned with the list object's actual malloc usage across
read-only fetches and mutating commands that materialize tiered nodes.

Also add a tiered list regression test that compares tracked OBJ_LIST memory against the 
owning list object's MallocUsed value.

### DIFF
--- a/src/core/qlist.cc
+++ b/src/core/qlist.cc
@@ -430,7 +430,7 @@ void QList::Node::Upload(QList* ql, std::string_view val) {
   entry = static_cast<unsigned char*>(zmalloc(val.size()));
   memcpy(entry, val.data(), val.size());
   ql->AdjustMallocSize(val.size());
-  ql->AdjustAccountedObjectSize(val.size());
+  ql->AddReportedMemorySizeDelta(val.size());
   ql->AdjustOffloadNodeCount(-1);
   offloaded = 0;
 }
--- a/src/core/qlist.cc
+++ b/src/core/qlist.cc
@@ -430,7 +430,7 @@ void QList::Node::Upload(QList* ql, std::string_view val) {
   entry = static_cast<unsigned char*>(zmalloc(val.size()));
   memcpy(entry, val.data(), val.size());
   ql->AdjustMallocSize(val.size());
-  ql->AddReportedMemorySizeDelta(val.size());
+  ql->AddReportedMemorySizeDelta(static_cast<int32_t>(val.size()));
   ql->AdjustOffloadNodeCount(-1);
   offloaded = 0;
 }
--- a/src/core/qlist.cc
+++ b/src/core/qlist.cc
@@ -430,6 +430,7 @@ void QList::Node::Upload(QList* ql, std::string_view val) {
   entry = static_cast<unsigned char*>(zmalloc(val.size()));
   memcpy(entry, val.data(), val.size());
   ql->AdjustMallocSize(val.size());
+  ql->AdjustAccountedObjectSize(val.size());
   ql->AdjustOffloadNodeCount(-1);
   offloaded = 0;
 }
--- a/src/core/qlist.h
+++ b/src/core/qlist.h
@@ -123,7 +123,7 @@ class QList {
     malloc_size_ += delta;
   }
 
-  void AddReportedMemorySizeDelta(size_t delta) {
+  void AddReportedMemorySizeDelta(int32_t delta) {
     if (tiering_enabled_) {
       tiering_params_->reported_memory_size_delta += delta;
     }
--- a/src/core/qlist.h
+++ b/src/core/qlist.h
@@ -119,13 +119,13 @@ class QList {
   using IterateFunc = absl::FunctionRef<bool(Entry)>;
   enum InsertOpt : uint8_t { BEFORE, AFTER };
 
-  void AdjustMallocSize(size_t delta) {
+  void AdjustMallocSize(ssize_t delta) {
     malloc_size_ += delta;
   }
 
-  void AdjustAccountedObjectSize(size_t delta) {
+  void AddReportedMemorySizeDelta(size_t delta) {
     if (tiering_enabled_) {
-      tiering_params_->accounted_object_size += delta;
+      tiering_params_->reported_memory_size_delta += delta;
     }
   }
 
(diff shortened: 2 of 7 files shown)


src/core/qlist.h
```diff
@@ -119,10 +119,16 @@ class QList {
   using IterateFunc = absl::FunctionRef<bool(Entry)>;
   enum InsertOpt : uint8_t { BEFORE, AFTER };
 
-  void AdjustMallocSize(ssize_t delta) {
+  void AdjustMallocSize(size_t delta) {
```
romange: `AdjustMallocSize` was added by you and you used it in two places.  In a second place in tiered_storage.cc you call it with a negative size.

src/core/qlist.h
```diff
@@ -119,10 +119,16 @@ class QList {
...
     malloc_size_ += delta;
   }
 
+  void AdjustAccountedObjectSize(size_t delta) {
```
romange: But not it is even more confusing:
When `AdjustMallocSize` should be used and when `AdjustAccountedObjectSize` ?

what are the different semantics between both tracked values?

src/server/db_slice.cc
```diff
@@ -531,9 +531,11 @@ void DbSlice::AutoUpdater::Run() {
     // from a counter that never had the original bytes added to it.
     AccountObjectMemory(fields_.key, fields_.orig_obj_type,
                         -static_cast<int64_t>(fields_.orig_value_heap_size), table);
-    AccountObjectMemory(fields_.key, current_type, current_size, table);
+    AccountObjectMemory(fields_.key, current_type,
+                        current_size - fields_.already_accounted_object_size, table);
```
romange: it looks like a hack. Lets describe the exact scenario that causes misculculation and think of a more elegant solution

src/core/qlist.h
mkaruza: Agreed, the current function and variable names don't clearly convey their purpose. I'll update them to use more concise and descriptive names in this change.

src/server/db_slice.cc
mkaruza: It's a bit hacky, but it makes sense.

The problem is that tiering adjusts the object's malloc_size_ and updates the stats variable.

For read-only commands, this works fine:

We try to fetch a node that is tiered.
When we load the value back into memory, we update malloc_size_ and adjust stats accordingly.

However, mutable commands use Autoupdater, which updates the stats after the command finishes executing. As a result, the same memory usage is accounted for twice: once when the tiered value is loaded back into memory and again when Autoupdater performs its update.

src/server/list_family.cc
```diff
@@ -527,6 +542,8 @@ OpResult<uint32_t> OpPush(const OpArgs& op_args, std::string_view key, ListDir d
   lw.Launder(&res.it->second);
   len = lw.Size();
 
+  res.post_updater.AddManuallyReportedMemorySizeDelta(lw.ReportedMemorySizeDelta());
```
romange, line 545: Can you please explain here what is the use-case we are trying to fix?

src/core/qlist.cc
```diff
@@ -430,6 +430,7 @@ void QList::Node::Upload(QList* ql, std::string_view val) {
   entry = static_cast<unsigned char*>(zmalloc(val.size()));
   memcpy(entry, val.data(), val.size());
   ql->AdjustMallocSize(val.size());
+  ql->AddReportedMemorySizeDelta(static_cast<int32_t>(val.size()));
```
romange, line 433: I am trying to understand this.
if we call `ql->AdjustMallocSize(val.size());`  why do we need 

`ql->AddReportedMemorySizeDelta` ?  We already update malloc size. What am I missing here?

mkaruza: @romange 

When a QList has tiering enabled, nodes can be offloaded to disk and uploaded back to memory as part of normal command processing. 

In both cases, the tiering path immediately updates the QList's `malloc_size_` to reflect the node's new in-memory footprint and also directly calls `AddTypeMemoryUsage` to keep `obj_memory_usage` stats in sync with the actual in-memory size of the object.

1. For the READ path this works correctly. 

The node gets uploaded, `malloc_size_`increases, and `obj_memory_usage` is updated right there in the tiering callback. 

**No command-level accounting follows, so there is no double-count.**

2. The problem arises on the MUTABLE path. 

Commands uses `Autoupdater` to track object memory.

At the start of a command it records `malloc_before = MallocUsed()`, and at the end it reads `malloc_after = MallocUsed()` and reports `diff = malloc_after - malloc_before` to `obj_memory_usage`.

Flow when a node gets uploaded during a mutable command

```
malloc_before = size without the node (node was on disk, not counted with malloc_size_)
tiering fires: AdjustMallocSize(+N) and AddTypeMemoryUsage(+N)
malloc_after = size with the node (now counted with malloc_size_)
AutoUpdater computes diff = +N and calls AccountObjectMemory(+N)
```
That +N was already reported by tiering, so it gets counted twice.

**For non-tiered lists AddReportedMemorySizeDelta is a no-op and nothing changes.**

romange: @mkaruza  where tiering fires  `AddTypeMemoryUsage(+N)` during uploading? in `RecordDeleted` ?

mkaruza: Yes, when value is loaded back to memory.

romange: `RecordDeleted` is a confusing name, imho.

Is it possible to fix double accounting there? i.e. avoid calling `AddTypeMemoryUsage` when we upload for mutable commands? or we do not have this context?


I am looking for simpler ways to solve it as you can see how this PR makes things more complicated by introducing another variable we need to track now in several layers.

mkaruza: @romange

I agree that adding another variable is adding complexity.

How i see it we have two "components". First one is command execution that uses `Autoupdater` and assumes if `object size` changed it should update also `stats`. Second is tiering part that does update `object size` and `stats` directly.

Another solution to avoid updating `stats` from tiering is that tiering somehow know that stats will be updated outside. Each call to qlist methods is done through `ListWrapper` that is constructed on each command. We can have pass extra argument to this wrapper that can indicate if we should update stats or not (it should be done for each command).

romange: And another approach is https://github.com/dragonflydb/dragonfly/issues/5268
where we do not rely on specific object to track per type memory consumption

mkaruza: That would work as general solution to this and other situations.

mkaruza:
Should be done/covered with #5268

PR #7653, opened 2026-06-19
chore: migrate server_family/dflycmd/debug/cluster/search to ParsedArgs

romange:
## Summary

Final batch of the `CmdArgParser` migration away from raw `CmdArgList` args.

## Changes

- `cluster_family`, `debugcmd`, `dflycmd`, `search_family`: replace `CmdArgParser parser(args)` with `cmd_cntx->tail_args()` (or `.Tail()` for sub-dispatched helpers that skip the command name)
- `server_family`: same migration, plus convert `ReplicaOfArgs::FromCmdArgs` and `ReplicaOfInternal` to accept `facade::ParsedArgs` instead of `CmdArgList`
- `Replicate()`: replace manual `StringVec`/`CmdArgVec`/`CmdArgList` construction with `BackedArguments`


dranikpg:
The next pain point is Transaction itself which requires a contigious array to reference it with its index slices

PR #7654, opened 2026-06-19
chore: migrate generic functions to use ParsedArgs

romange:
## Summary
Moves the remaining `ArgSlice`-based `CmdArgParser` construction in the core dispatch and generic command paths onto `cmn::BackedArguments`, accessed via `cmd_cntx->tail_args()`.

## Changes
- `ScanOpts::TryFrom` / `RestoreArgs::TryFrom` now take `facade::ParsedArgs`; SCAN/HSCAN/SSCAN/ZSCAN and RESTORE callers pass `tail_args()` (with `.Tail()` to skip leading key/cursor args).
- `Expire`/`ExpireAt`/`Pexpire`/`PexpireAt`, `FieldExpire`, `Move`, `Copy`, `Select` build their parser from `tail_args()`; `DELEX` delegates to `DEL` via a `tail_args()`-backed parser.
- `StoredCmd` ctor drops the dead `ArgSlice`-based `ParsedArgs` initializer.
- `EvalValidator` parses its `CmdArgList` directly (the validator path only ever has an unbacked slice).

## Note
Builds on the prerequisite commit `65d71c96` ("pass BackedArguments through Lua redis.call path"), which rides along in this PR until that change lands on `main`.

### DIFF
--- a/src/server/generic_family.cc
+++ b/src/server/generic_family.cc
@@ -169,7 +169,7 @@ class RestoreArgs {
 
   [[nodiscard]] bool UpdateExpiration(int64_t now_msec);
 
-  static OpResult<RestoreArgs> TryFrom(const CmdArgList& args);
+  static OpResult<RestoreArgs> TryFrom(facade::ParsedArgs args);
 };
 
 class RdbRestoreValue : protected RdbLoaderBase {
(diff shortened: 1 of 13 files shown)


src/server/generic_family.cc
```diff
@@ -2334,7 +2335,7 @@ void GenericFamily::FieldTtl(CmdArgList args, CommandContext* cmd_cntx) {
 }
 
 void GenericFamily::Move(CmdArgList args, CommandContext* cmd_cntx) {
-  facade::CmdArgParser parser{args};
+  facade::CmdArgParser parser = MakeParserFromContext(cmd_cntx);
```
dranikpg, line 2338: this is a classic duality that is introduced that will be then more complicated to unify back

dranikpg: what's the point of passing down arguments now if cmd_cntx has them now?

dranikpg (dismissed):

src/server/generic_family.cc
romange: The point is to stop relying on `CmdArgList`, and use ParsedArgs.
Once we will transition everywhere to `ParsedArgs`, it will be degenerated to a suffix slice over BackedArguments:
`BackedArguments&` and an `offset` field so we could pass "tail" of the arguments.

variant will be removed and the class will become small. 

Once everything is complete, we can remove `CmdArgList` from handler interface entirely and do additional cleanups in  `CommandContext`

dranikpg (approved):

PR #7657, opened 2026-06-20
chore: migrate more families to tail_args()

romange:

### DIFF
--- a/src/server/memory_cmd.cc
+++ b/src/server/memory_cmd.cc
@@ -184,7 +184,7 @@ MemoryCmd::MemoryCmd(ServerFamily* owner, CommandContext* cmd_cntx)
 }
 
 void MemoryCmd::Run(CmdArgList args) {
-  CmdArgParser parser(args);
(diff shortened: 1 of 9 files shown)


src/server/memory_cmd.cc
```diff
@@ -495,7 +495,7 @@ void MemoryCmd::Track(CmdArgList args) {
   return cmd_cntx_->SendError("MEMORY TRACK must be enabled at build time.");
 #endif
 
-  CmdArgParser parser(args);
+  CmdArgParser parser(cmd_cntx_->tail_args().Tail());
```
abhijat: Why does this need an extra `Tail()`, isn't `tail_args()` already usable?

abhijat (dismissed):

src/server/memory_cmd.cc
romange: You are right, fixed.

abhijat (approved):

PR #7662, opened 2026-06-21
chore(server): pass CmdArgParser to bitops/json handlers

romange:
## Summary
Adds a parser-based `CommandId::SetHandler` overload and migrates bitops/JSON command
handlers to receive `CmdArgParser` from dispatch.

## Changes
- Add a non-async parser handler overload and make `SetAsyncHandler` reuse it
- Convert bitops and JSON handlers away from local `CmdArgParser` construction
- Keep arity and parsing checks on the shared dispatch parser where needed

### DIFF
--- a/src/server/json_family.cc
+++ b/src/server/json_family.cc
@@ -1709,8 +1709,7 @@ OpStatus OpMerge(const OpArgs& op_args, string_view key, string_view path,
   return OpStatus::SYNTAX_ERR;
 }
 
-void CmdSet(CmdArgList args, CommandContext* cmd_cntx) {
-  CmdArgParser parser{cmd_cntx->tail_args()};
+void CmdSet(CmdArgParser parser, CommandContext* cmd_cntx) {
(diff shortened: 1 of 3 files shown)


src/server/json_family.cc
```diff
@@ -1871,10 +1865,9 @@ void CmdDebug(CmdArgList args, CommandContext* cmd_cntx) {
   builder->SendError(facade::UnknownSubCmd(command, "JSON.DEBUG"), facade::kSyntaxErrType);
 }
 
-void CmdMGet(CmdArgList args, CommandContext* cmd_cntx) {
-  DCHECK_GE(args.size(), 1U);
```
abhijat, line 1875: why is the DCHECK removed?

abhijat: There are already arity checks in the validation steps so maybe these checks were redundant

romange: yes

abhijat (approved):

PR #7663, opened 2026-06-21
refactor(server): move VerifyCommandState out of squasher, migrate dispatch path to ParsedArgs

romange:
## Summary

- Removes `verify_commands` from `MultiCommandSquasher::Opts` and eliminates `VerifyCommandState` calls from `ExecuteStandalone` and `SquashedHopCb`, simplifying the squasher to a pure execution loop.
- Moves verification to the two call sites that set `verify_commands = true`:
  - `DispatchSquashedBatch`: verifies each command in the pre-squash loop; on failure, captures the error into the command's deferred reply slot and skips it.
  - `FlushEvalAsyncCmds`: verifies all async commands before executing any; on first failure, clears the queue and returns the error.

## ParsedArgs migration

Migrates `VerifyCommandState` and its entire helper chain from `CmdArgList`/`ArgSlice` to `const facade::ParsedArgs&`, following #7656:
- `VerifyCommandState`, `CheckKeysOwnership`, `TakenOverSlotError`, `CheckKeysDeclared`, `VerifyConnectionAclStatus`, `FindKeys`
- `DetermineKeys`, `Transaction::InitByArgs`, `Transaction::InitBase`
- `acl::IsUserAllowedToInvokeCommand`, `IsUserAllowedToInvokeCommandGeneric`, `IsPubSubCommandAuthorized`

`ParsedArgs` has an implicit `ArgSlice` constructor so existing raw-span call sites compile unchanged.

## Testing

- `ninja dragonfly` builds clean.
- `multi_test`: 59 passed, 2 skipped (pre-existing).

### DIFF
--- a/src/server/acl/validator.cc
+++ b/src/server/acl/validator.cc
@@ -55,11 +55,10 @@ bool ValidateCommand(const std::vector<uint64_t>& acl_commands, const CommandId&
   if (!pub_sub.all_channels) {
     std::string_view name = id.name();
     if (name == "PUBLISH" || name == "SPUBLISH") {
-      auto channel = tail_args[0];
-      allowed &= iterate_globs(facade::ToSV(channel));
+      allowed &= iterate_globs(tail_args[0]);
     } else {
       for (auto channel : tail_args) {
-        allowed &= iterate_globs(facade::ToSV(channel));
+        allowed &= iterate_globs(channel);
       }
     }
   }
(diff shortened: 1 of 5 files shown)


kostasrim (approved):
This is *not* a purely refactor PR because it breaks existing behaviour with ACL's. 

Before: 
We checked from within the shard. verification and execution was essentially back to back. If, first check succeeded the second could fail because by the time the shard callback is dispatched the ACL's of that user can be revoked/changed.

After:
this PR does widen the verify-then-execute gap effectively allowing an authorized user to execute a given command.

Since we are an eventually consistent system then the semantics are operationally acceptable. We do break behaviour here so maybe this is something worth documenting.

kostasrim:
And as far as I remember I have an acl test that tests that behaviour specifically (revoking after we squash before we execute)

kostasrim:
> And as far as I remember I have an acl test that tests that behaviour specifically (revoking after we squash before we execute)

ok I am wrong, I did not have a test for this -- the pr should be fine :)

PR #7670, opened 2026-06-22
fix(search): serialize only KNN winners for filtered vector search

### LINKED ISSUE #7632: significant performance degradation for wide filtered KNN search
I've written small script to reproduce the problem.
output example: 

Redis
```
engine: 8.6.2
loading 50000 docs...
  loaded in 27.1s

5s search per scenario:
  narrow ~3%       3524.1 ops/s
  wide   ~50%      6834.8 ops/s
```

Dragonfly
```
engine: df-v1.39.0
loading 50000 docs...
  loaded in 8.9s

5s search per scenario:
  narrow ~3%       4653.8 ops/s
  wide   ~50%       607.5 ops/s
```

```
#!/usr/bin/env python3
"""Reproduce Dragonfly v1.39 HNSW + wide-filter regression.

No args, no download. Connects to 127.0.0.1:6379. Generates 50 000
random 384-d FLOAT32 vectors with a uniform 0..100 `score` field,
builds HNSW, then runs narrow vs wide filtered KNN for 5 s each:
    narrow  @score:[0 2]      ~3 %   (control)
    wide    @score:[50 100]   ~50 %  (the regression)

On Dragonfly v1.39 wide is ~9× slower than narrow.

Requires: pip install "redis[hiredis]"
"""
import random, struct, time, redis

N, DIM, INDEX, DUR = 50_000, 384, "idx", 5.0
ROW = DIM * 4

r = redis.Redis(host="127.0.0.1", port=6379)
srv = r.info("server")
print(f"engine: {srv.get('dragonfly_version') or srv.get('redis_version')}")

try: r.execute_command("FT.DROPINDEX", INDEX)
except redis.ResponseError: pass
r.execute_command(
    "FT.CREATE", INDEX, "SCHEMA",
    "score", "NUMERIC",
    "v", "VECTOR", "HNSW", "10",
    "TYPE", "FLOAT32", "DIM", DIM,
    "DISTANCE_METRIC", "COSINE", "M", "16", "EF_CONSTRUCTION", "200",
)

print(f"loading {N} docs...")
rng = random.Random(42)
vecs = struct.pack(f"<{N*DIM}f", *(rng.random() for _ in range(N * DIM)))
t0 = time.perf_counter()
pipe = r.pipeline(transaction=False)
for i in range(N):
    pipe.hset(str(i), mapping={"score": rng.randint(0, 100), "v": vecs[i * ROW:(i + 1) * ROW]})
    if i % 1000 == 999:
        pipe.execute()
pipe.execute()
time.sleep(2)  # Dragonfly is sync (no-op); Redis 8 with --search-workers is async
print(f"  loaded in {time.perf_counter() - t0:.1f}s")

q = vecs[:ROW]


def measure(filt: str, label: str):
    n, t0 = 0, time.perf_counter()
    while time.perf_counter() - t0 < DUR:
        r.execute_command(
            "FT.SEARCH", INDEX,
            f"({filt})=>[KNN 10 @v $q]", "PARAMS", "2", "q", q,
            "NOCONTENT", "DIALECT", "2",
        )
        n += 1
    elapsed = time.perf_counter() - t0
    print(f"  {label:<14}{n/elapsed:>9.1f} ops/s")


print(f"\n{DUR:g}s search per scenario:")
measure("@score:[0 2]",    "narrow ~3%")
measure("@score:[50 100]", "wide   ~50%")
```

vyavdoshenko:
Filtered KNN queries (`(@filter)=>[KNN k @v $q]`) previously serialized **every** filter-matched document on each shard before running the global HNSW search, so a wide filter (e.g. matching ~50% of the index) paid a huge serialization cost to return only `k` results. This now fetches only matched **ids** for the prefilter, runs KNN, and serializes **only the `k` winners**. Also adds `FT.PROFILE` support for HNSW queries, hardens the new multi-hop transaction flow, and fixes a `SORTBY` result-drop bug found while verifying the change.

Fixes #7632.

Wide filtered KNN (the perf fix):
- New `ShardDocIndex::SearchIds()` / `SearchIdResult`: the prefilter pass returns matched doc ids (+ text scores when scoring) instead of fully serialized docs.
- `FT.SEARCH` KNN-prefilter is now two hops: collect ids -> global HNSW KNN on the coordinator → serialize only the winners (`LoadHnswSearchDocs`, gated by a `finalize` flag; `NOCONTENT` uses a key-index fast path that skips doc loads).
- `SearchHnswWithPrefilter()` sorts the allow-list before the membership-checked `Knn(..., allowed)` path.
- `PrepareHnswLoadOptions()` resolves the sort field through `RETURN ... AS` aliases.

FT.PROFILE for HNSW:
- Adds dedicated branches for KNN prefilter, bare `VECTOR_RANGE`, range prefilter, and FLAT range alias, via a shared `SendFtProfileSearchResponse` helper.

Transaction safety:
- The KNN-prefilter / range-prefilter flows leave the transaction open between hops. Added an `absl::Cleanup` guard in `CmdFtSearch`/`CmdFtProfile` that concludes the transaction if an exception unwinds out of the command (e.g. `bad_alloc` inside the global KNN), preventing a leaked scheduled global tx from stalling the shards. `Conclude()` is idempotent, so normal paths are unaffected.

SORTBY on a missing non-sortable field:
- A `SORTBY` on a non-`SORTABLE` field used to **drop** docs that lack the field from the result while still counting them in `total` (`count` disagreed with the returned rows). Such docs are now kept and ranked **last** in both directions, and never occupy a top-K slot - consistent with the reference engine behavior.
- `KeepTopKSorted` keeps missing values (as null) with null-last selection;
  `PartialSortBySortScore` orders the global merge null-last.

Tests:
- `FilteredKnnHnswWithScores` — `WITHSCORES` on filtered KNN (multi-shard scoring path).
- `FilteredKnnHnswKeepsDocMissingSortField` — field-less nearest doc stays a KNN candidate.
- `SortByKeepsDocsWithoutSortFieldLast` — replaces the old drop-asserting crash test; still guards the original out-of-bounds crash.
- `SortByNonSortableFieldNullsLast` — null-last ordering (ASC/DESC) + top-K exclusion.

### DIFF
--- a/src/server/search/doc_index.cc
+++ b/src/server/search/doc_index.cc
@@ -827,8 +827,13 @@ vector<search::SortableValue> ShardDocIndex::KeepTopKSorted(vector<DocId>* ids, 
   DCHECK_GT(limit, 0u) << "Limit=0 still has O(ids->size()) complexity";
 
   using QPair = std::pair<search::SortableValue, DocId>;
-  // Docs missing the sort field are kept but rank worst (sorted last) in both directions;
-  // returns true when `a` should outrank `b` so q.top() is always the worst kept value.
+  // priority_queue comparator: q.top() is the worst kept entry, evicted first when a better
+  // candidate arrives. better(a, b) is true when a should rank before b. Docs missing the sort
+  // field hold a monostate value and always rank last, in both directions. Cases (a, b -> result):
+  //   present, present -> order by (value, doc id): a < b for ASC, b < a for DESC
+  //   present, missing -> true   (present ranks before missing)
+  //   missing, present -> false  (missing ranks after present)
+  //   missing, missing -> order by (value, doc id); values are equal, so deterministic by doc id
   auto better = [order = sort.order](const QPair& a, const QPair& b) {
     bool a_null = std::holds_alternative<std::monostate>(a.first);
     bool b_null = std::holds_alternative<std::monostate>(b.first);
--- a/src/server/search/doc_index.cc
+++ b/src/server/search/doc_index.cc
@@ -6,6 +6,7 @@
 
 #include <absl/strings/str_join.h>
 
+#include <functional>
 #include <memory>
 #include <queue>
 #include <ranges>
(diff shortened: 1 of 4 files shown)


vyavdoshenko:
augment review

src/server/search/doc_index.cc
```diff
@@ -826,28 +826,34 @@ vector<search::SortableValue> ShardDocIndex::KeepTopKSorted(vector<DocId>* ids,
...
   using QPair = std::pair<search::SortableValue, DocId>;
-  std::priority_queue<QPair, std::vector<QPair>, decltype(comp)> q(comp);
+  // Docs missing the sort field are kept but rank worst (sorted last) in both directions;
+  // returns true when `a` should outrank `b` so q.top() is always the worst kept value.
+  auto better = [order = sort.order](const QPair& a, const QPair& b) {
+    bool a_null = std::holds_alternative<std::monostate>(a.first);
+    bool b_null = std::holds_alternative<std::monostate>(b.first);
+    if (a_null != b_null)
+      return b_null;
+    return order == SortOrder::ASC ? a < b : b < a;
```
BorysTheDev: I'm not sure about the correctness of this code. Please write in the comment above all cases and comparison results for them.

src/server/search/doc_index.cc
```diff
@@ -826,28 +826,34 @@ vector<search::SortableValue> ShardDocIndex::KeepTopKSorted(vector<DocId>* ids,
...
 
-  auto comp = [order = sort.order](const auto& lhs, const auto& rhs) {
-    return order == SortOrder::ASC ? lhs < rhs : lhs > rhs;
-  };
-  // Priority queue keeps top-k values in reverse order (to compare against top - worst value)
   using QPair = std::pair<search::SortableValue, DocId>;
-  std::priority_queue<QPair, std::vector<QPair>, decltype(comp)> q(comp);
+  // Docs missing the sort field are kept but rank worst (sorted last) in both directions;
+  // returns true when `a` should outrank `b` so q.top() is always the worst kept value.
+  auto better = [order = sort.order](const QPair& a, const QPair& b) {
```
BorysTheDev: in the code you want to choose the comparator, so rewrite the code to show it. create compile-time comparator and use std::less or std::greater to select behavior. by the way you can try to write <=> with it you can describe the null case

src/server/search/doc_index.cc
vyavdoshenko: Done - added a full case table above the comparator (and the same one to PartialSortBySortScore, which uses the identical rule):

present, present -> by (value, doc id): a<b for ASC, b<a for DESC
present, missing -> a first (missing sorts last)
missing, present -> a last
missing, missing -> by (value, doc id); values equal

src/server/search/doc_index.cc
vyavdoshenko: The per-direction part is just less/greater, but "missing always last" is direction-independent, so it stays an explicit branch regardless. <=> doesn't help here: the default operator<=> for variant<monostate, double, string> orders monostate first (opposite of what we want), and double's <=> is partial_ordering, so a three-way would be more code, not less. Kept it explicit and documented. Correctness is covered by SortByNonSortableFieldNullsLast and SortByKeepsDocsWithoutSortFieldLast.

BorysTheDev (approved):

PR #7671, opened 2026-06-22
fix: SetCMS/SetSBF/SetTOPK/SetCuckooFilter orphan and leak on bad_alloc   

### LINKED ISSUE #7668: AllocateMR<> can leak when kernel is configured *without* overcommit=1 and T constructor throws
We use overcommit=1 so allocations generally succeed. However, this is not true for other modes and if T's (let's say SBF here) constructor throws then we get a `memory leak`.

```                                                                                                                                                                                                                                                  
    std::vector<SBF*> allocations;                                                                                                                                                                                                                                            
    bool threw = false;                                                                                                                
    try {                                                                                                                                                                                                                                                                     
      for (int i = 0; i < 64; ++i) {  // safety cap: ~64 GiB ceiling even if it never throws
        allocations.push_back(CompactObj::AllocateMR<SBF>(kCapacityFor1GiB, 0.001, 2.0, mr));
      }                                                             
    } catch (const std::bad_alloc&) {                               
      throw std::bac_alloc;
    }  
```

And this will leak because `AllocateMR<>`:

```
return new (ptr) T{std::forward<Args>(args)...}; 
```

will throw `std::bad_alloc` in `T's` constructor and the allocated block:

```
void* ptr = memory_resource()->allocate(sizeof(T), alignof(T)); 
```

`ptr` just leaked here.

The fix is straightforward but is this something we care ? 🤔

kostasrim:
When vm.overcommit_memory=0, the kernel enforces memory limits and the allocator correctly throws std::bad_alloc. This exposed two bugs in SetCMS, SetSBF, SetTOPK, and SetCuckooFilter:

* Orphan/crash: SetMeta(TAG) was called before AllocateMR, leaving CompactObj with a valid tag but u_.ptr == nullptr. Any subsequent destruction (expiry, FLUSHDB, etc.) would trigger a SIGSEGV.
* Leak: AllocateMR allocated sizeof(T) bytes for the shell object, then placement new threw. The shell allocation was never freed.

Also, `std::bad_alloc` can be thrown by the constructor if an rdb file is injected with improper dim parameters as demonstrated in the test. This is *not* a very realistic scenario.

### The fix

* Make constructors non-throwing
* Add an Init() function that can-throw

fixes https://github.com/dragonflydb/dragonfly/issues/7668

### DIFF
--- a/src/core/bloom.cc
+++ b/src/core/bloom.cc
@@ -165,6 +165,9 @@ inline bool Bloom::Set(size_t bit_idx) {
 ///////////////////////////////////////////////////////////////////////////////
 void SBF::Init(uint64_t initial_capacity, double fp_prob, double grow_factor,
                PMR_NS::memory_resource* mr) {
+  // Called exactly once, right after construction: no previous filters to free.
+  DCHECK(filters_.empty());
+
   double new_fp_prob = fp_prob * kSBFErrorFactor;
 
   // Reserve storage for the new filter before initializing it. Bloom::~Bloom() CHECKs that
@@ -172,17 +175,13 @@ void SBF::Init(uint64_t initial_capacity, double fp_prob, double grow_factor,
   // filters_ — otherwise unwinding would destroy a Bloom that still owns its bit array and
   // abort instead of cleanly propagating bad_alloc. Reserving first guarantees the push_back
   // below cannot reallocate (and Bloom's move ctor is noexcept).
-  decltype(filters_) new_filters(filters_.get_allocator());
-  new_filters.reserve(1);
+  filters_.reserve(1);
 
   Bloom b;
   b.Init(initial_capacity, new_fp_prob, mr);
-  new_filters.push_back(std::move(b));
+  filters_.push_back(std::move(b));
 
   // Nothing below can throw.
-  for (auto& f : filters_)
-    f.Destroy(mr);
-  filters_.swap(new_filters);
   grow_factor_ = grow_factor;
   fp_prob_ = new_fp_prob;
   max_capacity_ = filters_.front().Capacity(fp_prob_);
--- a/src/core/bloom.cc
+++ b/src/core/bloom.cc
@@ -165,17 +165,24 @@ inline bool Bloom::Set(size_t bit_idx) {
 ///////////////////////////////////////////////////////////////////////////////
 void SBF::Init(uint64_t initial_capacity, double fp_prob, double grow_factor,
                PMR_NS::memory_resource* mr) {
-  // Build and initialize the new filter before touching any existing state: if Bloom::Init
-  // throws, *this is left completely unchanged.
   double new_fp_prob = fp_prob * kSBFErrorFactor;
+
+  // Reserve storage for the new filter before initializing it. Bloom::~Bloom() CHECKs that
+  // bf_ is null, so once b.Init() below succeeds, b must never throw on its way into
+  // filters_ — otherwise unwinding would destroy a Bloom that still owns its bit array and
+  // abort instead of cleanly propagating bad_alloc. Reserving first guarantees the push_back
+  // below cannot reallocate (and Bloom's move ctor is noexcept).
+  decltype(filters_) new_filters(filters_.get_allocator());
+  new_filters.reserve(1);
+
   Bloom b;
   b.Init(initial_capacity, new_fp_prob, mr);
+  new_filters.push_back(std::move(b));
 
   // Nothing below can throw.
   for (auto& f : filters_)
     f.Destroy(mr);
-  filters_.clear();
-  filters_.push_back(std::move(b));
+  filters_.swap(new_filters);
   grow_factor_ = grow_factor;
   fp_prob_ = new_fp_prob;
   max_capacity_ = filters_.front().Capacity(fp_prob_);
--- a/src/core/bloom.cc
+++ b/src/core/bloom.cc
@@ -163,10 +163,24 @@ inline bool Bloom::Set(size_t bit_idx) {
 ///////////////////////////////////////////////////////////////////////////////
 // SBF implementation
 ///////////////////////////////////////////////////////////////////////////////
-SBF::SBF(uint64_t initial_capacity, double fp_prob, double grow_factor, PMR_NS::memory_resource* mr)
-    : filters_(1, mr), grow_factor_(grow_factor), fp_prob_(fp_prob * kSBFErrorFactor) {
-  filters_.front().Init(initial_capacity, fp_prob_, mr);
+void SBF::Init(uint64_t initial_capacity, double fp_prob, double grow_factor,
+               PMR_NS::memory_resource* mr) {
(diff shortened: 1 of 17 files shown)


src/core/compact_object.cc
```diff
@@ -1109,11 +1109,12 @@ void CompactObj::SetJson(const uint8_t* buf, size_t len) {
 }
 
 void CompactObj::SetSBF(uint64_t initial_capacity, double fp_prob, double grow_factor) {
-  if (taglen_ == SBF_TAG) {  // already json
+  if (taglen_ == SBF_TAG) {
     *u_.sbf = SBF(initial_capacity, fp_prob, grow_factor, tl.local_mr);
   } else {
+    SBF* sbf = AllocateMR<SBF>(initial_capacity, fp_prob, grow_factor, tl.local_mr);
```
romange: why does it change the semantics when AllocateMR throws?

src/core/compact_object.h
```diff
@@ -506,10 +506,15 @@ class CompactObj {
 
   template <typename T, typename... Args> static T* AllocateMR(Args&&... args) {
     void* ptr = memory_resource()->allocate(sizeof(T), alignof(T));
-    if constexpr (std::is_constructible_v<T, decltype(memory_resource())> && sizeof...(args) == 0)
-      return new (ptr) T{memory_resource()};
-    else
-      return new (ptr) T{std::forward<Args>(args)...};
+    try {
```
romange: My problem is that now we arbitratry try catch on a hot path, though the rest of our code is not exception safe. 

I agree we have problem here but I prefer we solve it by removing dangerous allocations from c'tors of these classes and instead introduce a separate Init function that will be called in those families.  
c'tors should finish safely in our codebase. They should not throw.

kostasrim: > My problem is that now we arbitratry try catch on a hot path,

This should be fine in principle. You don't pay what you don't use and there shall be no overhead unless it actually throws. When the stack unwinds and propagate the exception this is where you pay the price. That case is not the hot path anymore but a rather "exceptional state". 

> I agree we have problem here but I prefer we solve it by removing dangerous allocations from c'tors of these classes and instead introduce a separate Init function that will be called in those families.

Happy to do that, I don't have a strong argument on this

src/core/compact_object.cc
kostasrim: Because when AllocateMr<> throws, sbf=nullptr and SetMeta == SBF_TAG. You just left a compact object with nullptr thinking its an SBF

romange: yeah, that's why I do not like throwing c'tors. In general C++ should have prohibited exceptions in c'tors imho.

kostasrim:
augment review

src/core/compact_object.cc
```diff
@@ -1123,12 +1123,11 @@ SBF* CompactObj::GetSBF() const {
...
 void CompactObj::SetCMS(uint32_t width, uint32_t depth) {
-  if (taglen_ == CMS_TAG) {
-    *u_.cms = CMS(width, depth, tl.local_mr);
-  } else {
+  if (taglen_ != CMS_TAG) {
     SetMeta(CMS_TAG);
-    u_.cms = AllocateMR<CMS>(width, depth, tl.local_mr);
+    u_.cms = AllocateMR<CMS>(tl.local_mr);  // trivial ctor, never throws
   }
+  u_.cms->Init(width, depth);  // may throw; u_.cms is already safely owned either way
```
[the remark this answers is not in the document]
kostasrim: no, we leave orphans but those won't crash the data store at least. Handling orphans is a little bit more complicated because now all paths must `Delete` the newly added compacted object

src/core/bloom.cc
```diff
@@ -163,10 +163,24 @@ inline bool Bloom::Set(size_t bit_idx) {
...
+  // throws, *this is left completely unchanged.
+  double new_fp_prob = fp_prob * kSBFErrorFactor;
+  Bloom b;
+  b.Init(initial_capacity, new_fp_prob, mr);
+
+  // Nothing below can throw.
+  for (auto& f : filters_)
+    f.Destroy(mr);
+  filters_.clear();
+  filters_.push_back(std::move(b));
```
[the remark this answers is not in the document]
kostasrim: fixed

kostasrim:
@romange Only `Init()` now throws but the scope of changes became larger

src/core/bloom.cc
```diff
@@ -163,10 +163,31 @@ inline bool Bloom::Set(size_t bit_idx) {
...
+               PMR_NS::memory_resource* mr) {
+  double new_fp_prob = fp_prob * kSBFErrorFactor;
+
+  // Reserve storage for the new filter before initializing it. Bloom::~Bloom() CHECKs that
+  // bf_ is null, so once b.Init() below succeeds, b must never throw on its way into
+  // filters_ — otherwise unwinding would destroy a Bloom that still owns its bit array and
+  // abort instead of cleanly propagating bad_alloc. Reserving first guarantees the push_back
+  // below cannot reallocate (and Bloom's move ctor is noexcept).
+  decltype(filters_) new_filters(filters_.get_allocator());
+  new_filters.reserve(1);
```
romange:
```suggestion
```

src/core/bloom.cc
```diff
@@ -163,10 +163,31 @@ inline bool Bloom::Set(size_t bit_idx) {
...
+void SBF::Init(uint64_t initial_capacity, double fp_prob, double grow_factor,
+               PMR_NS::memory_resource* mr) {
+  double new_fp_prob = fp_prob * kSBFErrorFactor;
+
+  // Reserve storage for the new filter before initializing it. Bloom::~Bloom() CHECKs that
+  // bf_ is null, so once b.Init() below succeeds, b must never throw on its way into
+  // filters_ — otherwise unwinding would destroy a Bloom that still owns its bit array and
+  // abort instead of cleanly propagating bad_alloc. Reserving first guarantees the push_back
+  // below cannot reallocate (and Bloom's move ctor is noexcept).
+  decltype(filters_) new_filters(filters_.get_allocator());
```
romange: I do not understand. Do we call Init on the already initialized `SBF` ? 
can we assume that `filters_` are empty?

romange (commented):
please rebase

src/core/bloom.cc
kostasrim: > Do we call Init on the already initialized SBF ?

No we won't call init on an already initialized sbf. I was being defensive here. I removed it, added a comment about the contract and a DCHECK in case we ever violate it

src/core/bloom.cc
kostasrim: This one is more tricky.

Bloom::Init allocates and if push_back throws, before we unwind the stack we call b's destructor which has a check: `CHECK(bf_ == nullptr)` and we crash. I left a comment for clarity

kostasrim:
augment review

kostasrim:
closing this not worth it.

PR #7677, opened 2026-06-23
feat(string_family): add CL.PROBE read-only throttle inspection command

### LINKED ISSUE #7128: Add a true read-only probe mode for CL.THROTTLE
## Motivation

CL.THROTTLE is documented as Dragonfly's native GCRA/leaky-bucket rate-limiting command and returns useful limiter state (limited, limit, remaining, retry-after, reset). In the current implementation, OpThrottle uses mutable access and persists state on the non-limited path, so there does not appear to be a side-effect-free way to ask "would this request be allowed right now?" without updating limiter state or creating state for a missing key. ([dragonflydb.io](https://www.dragonflydb.io/blog/introducing-a-rate-limiting-api-by-dragonfly))

## Use case

We are building a global rate-limiter service on top of Dragonfly and would like to rely on native CL.THROTTLE for GCRA-style throttling, but we also need probe semantics for:

- dark launches / observe-only mode
- candidate selection across multiple keys before consuming quota
- preflight/debug endpoints that show remaining, retry-after, and reset without mutating state
- safer migration/canary workflows where "would throttle" must not burn quota

## Requested behavior

Please add one of the following:

- a new CL.PROBE command with the same arguments and response shape as CL.THROTTLE
- or a READONLY / NOAPPLY mode on CL.THROTTLE

Desired semantics:

- no TAT update
- no TTL refresh
- no key creation for a missing key
- same response shape as CL.THROTTLE
- should work for arbitrary quantity, not only via a special-case workaround

## Why this would help

This would make Dragonfly's native throttling much easier to use in higher-level control planes and shared rate-limiter services, where "check" and "consume" are separate operations. It would also make CL.THROTTLE easier to use for observability, dry runs, and staged rollouts.

## Backward compatibility

This could be a purely additive feature. Existing CL.THROTTLE behavior would stay unchanged.

mvanhorn:
## Summary
Adds `CL.PROBE`, a read-only companion to `CL.THROTTLE`. It returns the same 5-element rate-limit response but inspects the throttle state without consuming quota, mutating the stored TAT, or creating the key.

## Why this matters
`CL.THROTTLE` always writes (it stores the new theoretical arrival time and bumps the key TTL), so there is no way to ask "would this be limited?" without actually consuming a request. #7128 asks for a non-consuming probe. `CL.PROBE` takes the same `<key> <max_burst> <count per period> <period> [<quantity>]` arguments and computes the identical decision, but takes the read path: it reads through `FindReadOnly` and skips the store/TTL update, so it is registered `CO::READONLY` and can run on replicas.

## Changes
- `OpThrottle` gains a `probe` flag; on the probe path it uses `FindReadOnly` and returns the computed decision without persisting the new TAT or TTL.
- `CmdClThrottle` and the new `CmdClProbe` share one `CmdClThrottleOrProbe` body.
- Registered `CL.PROBE` with `CO::READONLY | CO::FAST` under the existing `acl::THROTTLE` category.

## Testing
Added `StringFamilyTest.ClProbe`: verifies `CL.PROBE` never creates a missing key (`EXISTS` stays 0), never advances the TTL of an existing throttle key (repeated probes leave `PTTL` unchanged), and that `CL.THROTTLE` interleaved with probes still advances state as before. Built and ran the C++ suite locally is not included here; the test exercises the read-only invariants the change introduces.

Fixes #7128

AI was used for assistance.

### DIFF
--- a/src/server/string_family.cc
+++ b/src/server/string_family.cc
@@ -10,6 +10,7 @@
 #include <array>
 #include <chrono>
 #include <cstdint>
+#include <optional>
 #include <variant>
 
 #include "base/flags.h"
@@ -405,7 +406,7 @@ bool IsValueWithinBounds(const int64_t value, const int64_t bound) {
 // limit is assumed to be positive
 OpResult<array<int64_t, 5>> OpThrottle(const OpArgs& op_args, const string_view key,
                                        const int64_t limit, const int64_t emission_interval_ns,
-                                       const uint64_t quantity) {
+                                       const uint64_t quantity, bool probe = false) {
   constexpr uint64_t kSecondToMilliSecond = 1000;
   constexpr uint64_t kMilliSecondToNanoSecond = 1000000;
   auto& db_slice = op_args.GetDbSlice();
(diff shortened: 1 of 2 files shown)


romange:
@mvanhorn thanks! I assume you are contributing this because you need it?

mvanhorn:
Honest answer: not from a production deployment of my own - I came to it from the issue tracker rather than from needing it myself. The gap felt real though: CL.THROTTLE always writes (stores the new TAT, bumps the TTL), so there's no way for dashboards or debugging sessions to ask 'would this request be limited right now?' without consuming quota. If the use case is too speculative without a user attached to it, I'm fine with parking this until someone shows up needing it - your call.

PR #7678, opened 2026-06-23
feat(search): support WITHSCORES/SCORER in cluster search (CSS) mode

### LINKED ISSUE #7182: Support WITHSCORES/SCORER in cluster search (CSS) mode
Problem:

FT.SEARCH with WITHSCORES/SCORER is currently blocked in cluster search (CSS) mode with an error:

`WITHSCORES/SCORER is not yet supported in cluster search mode`

The coordinator-side reply parser in `FtSearchCSS` expects each hit as `[key, [sortkey], [fields...]]`, but WITHSCORES adds an extra score element `[key, score, [sortkey], [fields...]]` which misaligns parsing and produces errors.

Blocked in #7181.

Required changes:
1. **CSS reply parser**: when `params.with_scores` is set, consume the score string after the key and populate `SerializedSearchDoc::text_score`
2. **Score merging**: when collecting results from multiple shards, scores from different shards use per-shard statistics. Consider whether cross-shard score normalization is needed or if per-shard scores are acceptable for ranking
3. **SCORER forwarding**: ensure the SCORER parameter is forwarded to remote shards in the CSS command string
4. **Remove the guard** added in #7181 once the above is implemented

mvanhorn:
## Summary
Cluster single-shard search (CSS) errored on `WITHSCORES` / `SCORER`. This parses the score and sortkey fields out of each shard's sub-search response so cluster search returns scores the same way non-cluster search does.

## Changes
Replace the inline CSS response handler with `ParseFtSearchCSSResponse`, which reads `total_hits`, then per-doc reads the key plus (conditionally) the score and sortkey depending on `with_scores`/`scorer`/`sort_option`/`with_sortkeys`.

## Testing
Added cluster-search WITHSCORES coverage in `search_family_test.cc` and `tests/dragonfly/cluster_test.py`.

Fixes #7182

AI was used for assistance.

### DIFF
--- a/src/server/search/search_family.cc
+++ b/src/server/search/search_family.cc
@@ -1148,7 +1148,7 @@ void SendSerializedDoc(const SerializedSearchDoc& doc, SinkReplyBuilder* builder
 
 template <typename T>
 void PartialSort(absl::Span<SerializedSearchDoc*> docs, size_t limit, SortOrder order,
-                 T SerializedSearchDoc::*field) {
+                 T SerializedSearchDoc::* field) {
   auto cb = [order, field](SerializedSearchDoc* l, SerializedSearchDoc* r) {
     return order == SortOrder::ASC ? l->*field < r->*field : r->*field < l->*field;
   };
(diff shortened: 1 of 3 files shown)


vyavdoshenko:
not planned

PR #7679, opened 2026-06-23
refactor(facade): add OnRecvNotification helper & suppress spurious wakeups

glevkovich:
- Merge the two identical RegisterOnRecv lambdas in IoLoopV2 and OnPostMigrateThread by extracting their body into a new private OnRecvNotification() method.

- Adds HasNetworkEvent() to guard io_event_.notify(): the callback now skips the wakeup when the completion harvested nothing (no error, no pending data, no newly buffered bytes), avoiding a spurious wake that would just re-park on ShouldWakeIdle().

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -825,11 +825,8 @@ void Connection::OnPostMigrateThread() {
 
   if (ioloop_v2_ && socket_ && socket_->IsOpen() && migration_allowed_to_register_) {
     MaybeEnableRecvMultishot();
-    socket_->RegisterOnRecv([this](const FiberSocketBase::RecvNotification& n) {
-      NotifyOnRecv(n);
-      ReadPendingInput();
-      io_event_.notify();
-    });
+    socket_->RegisterOnRecv(
+        [this](const FiberSocketBase::RecvNotification& n) { OnRecvNotification(n); });
   }
 
   migration_in_process_ = false;
@@ -3031,6 +3028,21 @@ bool ConnectionRef::operator==(const ConnectionRef& other) const {
   return client_id_ == other.client_id_;
 }
 
+void Connection::OnRecvNotification(const util::FiberSocketBase::RecvNotification& n) {
+  DVLOG(2) << "OnRecvNotification iobuf_len: " << io_buf_.InputLen();
+  size_t input_before = io_buf_.InputLen();
+  NotifyOnRecv(n);
+  // Eagerly drain the socket while the fiber is suspended to prevent receive buffer starvation.
+  ReadPendingInput();
+  // Both epoll and io_uring poll can deliver spurious POLLIN completions where the subsequent
+  // recv() returns EAGAIN (no actual data). HasNetworkEvent() filters these out so the spurious
+  // proactor wakeup is not propagated as a spurious fiber wakeup: if nothing changed (no error,
+  // no new bytes, no pending input remaining) waking the fiber is wasteful — it would only
+  // re-evaluate ShouldWakeIdle(), find it false, and immediately re-park.
+  if (HasNetworkEvent(input_before))
+    io_event_.notify();
+}
+
 void Connection::NotifyOnRecv(const util::FiberSocketBase::RecvNotification& n) {
   if (std::holds_alternative<std::error_code>(n.read_result)) {
     io_ec_ = std::get<std::error_code>(n.read_result);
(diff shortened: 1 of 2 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -3031,6 +3028,17 @@ bool ConnectionRef::operator==(const ConnectionRef& other) const {
...
 }
 
+void Connection::OnRecvNotification(const util::FiberSocketBase::RecvNotification& n) {
+  DVLOG(2) << "OnRecvNotification iobuf_len: " << io_buf_.InputLen();
+  size_t input_before = io_buf_.InputLen();
+  NotifyOnRecv(n);
+  // Eagerly drain the socket while the fiber is suspended to prevent receive buffer starvation.
+  ReadPendingInput();
+  // Suppress spurious fiber wakeups from completions that harvested nothing.
+  if (HasNetworkEvent(input_before))
```
romange, line 3042: what is the scenario you are trying to prevent?
In other words in what case `OnRecvNotification` is being called but `io_event_.notify();` should not be called?

glevkovich: The case is spurious wakeup of the connection fiber. Since the connection fiber already read, and might have set the pending_input_ as false, In some cases there is no need to wake up the connection fiber by the proactor:

-  no IO error
- no new data was written to the input buffer
- no pending data in the socket (pending_input_ was cleared by TryRecv returning EAGAIN)

If the proactor notifies without this guard, the connection fiber wakes up, re-evaluates ShouldWakeIdle() (which checks all of the same conditions), finds nothing actionable, and immediately re-parks in io_event_.await().
There is always a chance that other conditions (which are not in not in HasNetworkEvent() but io_event_ waits on have changed, but we rely on notifications here - not opportunistic polls - for all the conditions io_event_ waits on.

romange: I still do not understand. `Since the connection fiber already read, and might have set the pending_input_ as false`. If it set `pending_input_ as false` it means new data was written to the input buffer, no?

I would like to understand better how it is possible that we have spurious notifications and how much it affects in your tests?

glevkovich: I'm going to push a clarification above this above the code itself.

Both epoll and io_uring polls can deliver a completion where the recv() that comes after immediately returns EAGAIN - that is a spurious wakeup from the socket.

The idea here is to mask this wakeup from propagating to the fiber.  The proactor received a POLLIN notification from the socket ("I have something for you to read"),  but when it actually tried to read, the socket returned EAGAIN - nothing was there. That is the definition of a spurious socket wakeup. Waking the connection fiber in that case would be wasteful: 
it would only re-evaluate ShouldWakeIdle(), find nothing actionable on the network events (again, it might find an event from other sources, but that's random), and immediately re-park.

romange: Does this code improve performance? Do you see more than 1 in 1000 spurious wakeups? I have never seen these, so that's what I am asking. If I run memtier will I observe spurious wake-ups ?

glevkovich: Because of the share-nothing architecture I'm not sure you will see any, since threads to don't contend on the same socket fd.
This is is a micro optimisation, I don't expect it to have major impact on performance.

- In epoll, spurious wakeups are rare, depends on the network pattern and thread activity, but in general  around 0.001% of wakeups are spurious wakeups.
- In iouring modetn use , it should not happen at all in, but it can happen in some cases (for example when using IORING_OP_POLL_ADD to mimic epoll behaviour, they way we do).

One part of the PR is actually refactoring which I think should be merged (having a single OnRecvNotification).

glevkovich: @romange 

So I made a check , and it actually happens in some workloads much more than expected.

for a single connection:

memtier_benchmark -s 172.31.30.209 -p 6379 --pipeline=100 -t 1 -c 1 --ratio=1:0 -d 256 --key-prefix=bench --test-time=30 --hide-histogram
OnRecvNotification called 136372, notifications skipped 55375
OnRecvNotification called 134966, notifications skipped 56807 (another attempt)

Same memtier command with -d=1024:
OnRecvNotification called 105319, notifications skipped 1
OnRecvNotification called 90645, notifications skipped 0 (another attempt)

Same memtier command with -d=4096:
OnRecvNotification called 163745, notifications skipped 16
OnRecvNotification called 115890 , notifications skipped 0 (another attempt)

============

for multiple connections:
memtier_benchmark -s 172.31.30.209 -p 6379 --pipeline=100 -t 2 -c 25 --ratio=1:0 -d 256 --key-prefix=bench --test-time=30 --hide-histogram
sum of OnRecvNotification on all connections 554,209: sum of notification skipped 37,871

the same with -d 1024:
sum of OnRecvNotification on all connections 1,007,476 sum of notification skipped 95,644

the same with -d 4096:
sum of OnRecvNotification on all connections 1,133,416 sum of notification skipped 101,444

I'm going to post performance benchmark on the another comment soon.

glevkovich:
@romange 
I've decided to minimize this PR to a pure refactoring and open a separate issue to research the spurious-wakeup avoidance properly.

The short story: I ran many configurations (single vs. multiple connections, varying message sizes) and the results were very ambiguous. Since I wasn't planning to invest in a longer investigation right now, I'd prefer to minimize this PR's scope . I'll open an issue to follow up on the optimization later.

I'm closing this and opening a new one.

Issue #5446, opened 2025-07-09
Implement CF (cuckoo filter)


kostasrim:
- [ ] CF.ADD
- [ ] CF.ADDNX
- [ ] CF.COUNT
- [ ] CF.DEL
- [ ] CF.EXISTS
- [ ] CF.INFO
- [ ] CF.INSERT
- [ ] CF.INSERTNX
- [ ] CF.LOADCHUNK
- [ ] CF.MEXISTS
- [ ] CF.RESERVE
- [ ] CF.SCANDUMP

kostasrim:
https://github.com/dragonflydb/dragonfly/pull/7646

PR #7681, opened 2026-06-23
chore: add cf.add and cf.addnx

kostasrim:
* add cf.add
* add cf.addnx
* add tests

#5446

### DIFF
--- a/src/server/cuckoo_filter_family.cc
+++ b/src/server/cuckoo_filter_family.cc
@@ -19,6 +19,42 @@ using namespace std;
 
 namespace {
 
+constexpr uint64_t kDefaultCapacity = 1024;
+
+OpResult<CuckooFilter*> FindOrCreate(const OpArgs& op_args, string_view key) {
+  auto& db_slice = op_args.GetDbSlice();
+  auto op_res = db_slice.AddOrFind(op_args.db_cntx, key, OBJ_CUCKOOFILTER);
+  RETURN_ON_BAD_STATUS(op_res);
+
+  PrimeValue& pv = op_res->it->second;
+  if (op_res->is_new) {
+    pv.SetCuckooFilter(CuckooFilterOptions{.capacity = kDefaultCapacity});
+  }
+  return pv.GetCuckooFilter();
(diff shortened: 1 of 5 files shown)


kostasrim:
augment review

src/server/cuckoo_filter_family.cc
```diff
@@ -82,6 +118,34 @@ void CmdReserve(CmdArgList args, CommandContext* cmd_cntx) {
...
 
+void CmdAdd(CmdArgList args, CommandContext* cmd_cntx) {
+  string_view key = ArgS(args, 0);
+  string_view item = ArgS(args, 1);
+
+  const auto cb = [&](Transaction* t, EngineShard* shard) {
+    return OpAdd(t->GetOpArgs(shard), key, item);
+  };
+
+  OpResult<bool> res = cmd_cntx->tx()->ScheduleSingleHopT(std::move(cb));
```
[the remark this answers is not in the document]
kostasrim: this is a generic thing around the codebase...

kostasrim:
@abhijat PTAL

src/server/cuckoo_filter_family.cc
```diff
@@ -82,6 +118,34 @@ void CmdReserve(CmdArgList args, CommandContext* cmd_cntx) {
...
+  if (!res)
+    return cmd_cntx->SendError(res.status());
+  cmd_cntx->SendLong(*res);
+}
+
+void CmdAddNx(CmdArgList args, CommandContext* cmd_cntx) {
```
abhijat, line 135: not something to change here, just fyi, but since we are slowly migrating to parser based handlers which take `facade::CmdArgParser` instead of args, and these are new commands, it might be good to use the parser based signature and corresponding `SetHandler` to reduce future migration work.

abhijat (approved):

src/server/cuckoo_filter_family.cc
kostasrim: I agree. This is a good principle! 

Will do.

PR #7682, opened 2026-06-23
chore: add cf.exists and cf.mexists

kostasrim:
* add cf.exists
* add cf.mexists

#5446

### DIFF
--- a/src/server/cuckoo_filter_family.cc
+++ b/src/server/cuckoo_filter_family.cc
@@ -55,6 +55,20 @@ OpResult<bool> OpAddNx(const OpArgs& op_args, string_view key, string_view item)
   return true;
 }
 
+OpResult<vector<bool>> OpExists(const OpArgs& op_args, string_view key, ParsedArgs items) {
+  auto& db_slice = op_args.GetDbSlice();
+  auto op_res = db_slice.FindReadOnly(op_args.db_cntx, key, OBJ_CUCKOOFILTER);
+  if (!op_res)
+    return op_res.status();
+
+  const CuckooFilter* cf = op_res.value()->second.GetCuckooFilter();
+  vector<bool> result(items.size());
+  for (size_t i = 0; i < items.size(); ++i) {
+    result[i] = cf->Exists(CuckooFilter::Hash(items[i]));
+  }
+  return result;
+}
+
 OpStatus OpReserve(const OpArgs& op_args, string_view key, const CuckooFilterOptions& options) {
   auto& db_slice = op_args.GetDbSlice();
   auto op_res = db_slice.AddOrFind(op_args.db_cntx, key, OBJ_CUCKOOFILTER);
@@ -67,8 +81,7 @@ OpStatus OpReserve(const OpArgs& op_args, string_view key, const CuckooFilterOpt
   return OpStatus::OK;
 }
 
-void CmdReserve(CmdArgList args, CommandContext* cmd_cntx) {
-  CmdArgParser parser(cmd_cntx->tail_args());
+void CmdReserve(CmdArgParser parser, CommandContext* cmd_cntx) {
   string_view key = parser.Next();
   uint64_t capacity = parser.Next<uint64_t>();
   auto* rb = static_cast<RedisReplyBuilder*>(cmd_cntx->rb());
@@ -118,9 +131,9 @@ void CmdReserve(CmdArgList args, CommandContext* cmd_cntx) {
   return rb->SendError(res);
 }
 
-void CmdAdd(CmdArgList args, CommandContext* cmd_cntx) {
-  string_view key = ArgS(args, 0);
-  string_view item = ArgS(args, 1);
+void CmdAdd(CmdArgParser parser, CommandContext* cmd_cntx) {
+  string_view key = parser.Next();
+  string_view item = parser.Next();
 
   const auto cb = [&](Transaction* t, EngineShard* shard) {
     return OpAdd(t->GetOpArgs(shard), key, item);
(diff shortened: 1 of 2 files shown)


src/server/cuckoo_filter_family_test.cc
```diff
@@ -92,4 +92,47 @@ TEST_F(CuckooFilterFamilyTest, AddFilterFull) {
...
+  EXPECT_THAT(Run({"cf.mexists", "f1", "foo", "bar", "baz"}),
+              RespArray(ElementsAre(IntArg(1), IntArg(1), IntArg(1))));
+  EXPECT_THAT(Run({"cf.mexists", "f1", "foo", "nope"}),
+              RespArray(ElementsAre(IntArg(1), IntArg(0))));
+
+  // Missing key returns an all-zero array, not an error.
+  EXPECT_THAT(Run({"cf.mexists", "nonexist-key", "blah"}), RespArray(ElementsAre(IntArg(0))));
+}
+
+TEST_F(CuckooFilterFamilyTest, MExistsWrongArity) {
```
kostasrim, line 128: Kinda redundant tbh, but oh well more coverage 🤷

kostasrim:
augment review

kostasrim:
> ### Code Review by Qodo
> `🐞 Bugs (1)` `📘 Rule violations (0)` `📎 Requirement gaps (0)` `📜 Skill insights (0)`
> 
> **Context used**
> ✅ Tickets: [`🎫 Implement CF (cuckoo filter) family`](https://github.com/dragonflydb/dragonfly/issues/5446)
> Grey Divider
> 
> 
> Remediation recommended
> 
> 1. Unnecessary OpExists copy `🐞 Bug` `➹ Performance`
> 
> > Description
> > 
> > OpExists returns a local vector<bool> via return result;, which constructs
> > OpResult<vector<bool>> through the const V& constructor and copies the whole result buffer. For
> > CF.MEXISTS with many items this adds avoidable per-call CPU/memory overhead.
> 
> > Code
> > 
> > `[src/server/cuckoo_filter_family.cc[R65-69]](https://github.com/dragonflydb/dragonfly/pull/7682/files#diff-7ea63baba4ebdddf92ae5d1de7aab28c4d1ac847ea33e3f297da3f2d64080939R65-R69)`
> > ```diff
> > +  vector<bool> result(items.size());
> > +  for (size_t i = 0; i < items.size(); ++i) {
> > +    result[i] = cf->Exists(CuckooFilter::Hash(items[i]));
> > +  }
> > +  return result;
> > ```
> 
> > Evidence
> > 
> > OpExists returns an lvalue result (no move), and facade::OpResult<V> has both rvalue and
> > const-lvalue constructors; returning an lvalue in this converting-return context selects the `const
> > V& overload, copying into v_`.
> > 
> > `[src/server/cuckoo_filter_family.cc[58-70]](https://github.com/dragonflydb/dragonfly/blob/ab688ccad533c800f060f5c391e322d6d46f4097/src/server/cuckoo_filter_family.cc/#L58-L70)`
> > `[src/facade/op_status.h[71-80]](https://github.com/dragonflydb/dragonfly/blob/ab688ccad533c800f060f5c391e322d6d46f4097/src/facade/op_status.h/#L71-L80)`
> 
> > Agent prompt
> > 
> > ```
> > The issue below was found during a code review. Follow the provided context and guidance below and implement a solution
> > 
> > ## Issue description
> > `OpExists` builds a local `vector<bool> result` and returns it with `return result;`. Because the function returns `facade::OpResult<vector<bool>>`, this return statement constructs the wrapper from an lvalue, selecting `OpResult(const V&)` and copying the vector.
> > 
> > ## Issue Context
> > This can be avoided by moving the local vector into the `OpResult` to prevent an extra copy, especially beneficial for large `CF.MEXISTS` argument lists.
> > 
> > ## Fix Focus Areas
> > - src/server/cuckoo_filter_family.cc[58-70]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `ⓘ Copy this prompt and use it to remediate the issue with your preferred AI generation tools`
> 
> Grey Divider
> 
> [Qodo Logo](https://www.qodo.ai)

Nope, return result; is treated as an `rvalue` which is then fed to the right constructor. This is not a deep copy

src/server/cuckoo_filter_family.cc
```diff
@@ -146,6 +159,38 @@ void CmdAddNx(CmdArgList args, CommandContext* cmd_cntx) {
...
+}
+
+void CmdExists(CmdArgParser parser, CommandContext* cmd_cntx) {
+  string_view key = parser.Next();
+  ParsedArgs items = parser.UnparsedArgs();
+
+  OpResult<vector<bool>> res = RunExists(cmd_cntx, key, items);
+  if (!res && res.status() != OpStatus::KEY_NOTFOUND && res.status() != OpStatus::WRONG_TYPE)
+    return cmd_cntx->SendError(res.status());
+  cmd_cntx->SendLong(res ? res->front() : 0);
```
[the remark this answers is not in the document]
kostasrim: that's why we validate and we have arity

src/server/cuckoo_filter_family.cc
```diff
@@ -55,6 +55,20 @@ OpResult<bool> OpAddNx(const OpArgs& op_args, string_view key, string_view item)
...
+OpResult<vector<bool>> OpExists(const OpArgs& op_args, string_view key, ParsedArgs items) {
+  auto& db_slice = op_args.GetDbSlice();
+  auto op_res = db_slice.FindReadOnly(op_args.db_cntx, key, OBJ_CUCKOOFILTER);
+  if (!op_res)
+    return op_res.status();
+
+  const CuckooFilter* cf = op_res.value()->second.GetCuckooFilter();
+  vector<bool> result(items.size());
+  for (size_t i = 0; i < items.size(); ++i) {
+    result[i] = cf->Exists(CuckooFilter::Hash(items[i]));
```
abhijat, line 67: this is not a nit or anything but more a question but perhaps there should be an API for cuckoo filters that directly takes a string view and adds/checks etc, so the caller doesn't have to hash or know about hashing at all? 

unless its not possible due to internals or something. 

its fine as it is though, its not like it leaks internal implementation details.

abhijat (approved):

src/server/cuckoo_filter_family.cc
kostasrim: I actually did that 2 commits ago and reverted it because it had not tangible benefits

Certain family members call the cuckoo API multiple times. Externalizing the hash avoids rehashing on its operation in the command flow. 

Adding an overload that accepts a string_view + hashes to win 2 lines was not really a big saving so I avoided it

PR #7686, opened 2026-06-23
Migrate hset family handlers to CmdArgParser

romange:
Summary: Migrate hset family command handlers to CmdArgParser on top of the probabilistic-family parser changes.

Changes:
- Use CmdArgParser handlers for hash commands.
- Propagate ParsedArgs field/value tails to callbacks without argument-vector materialization.

### DIFF
--- a/src/server/hset_family.cc
+++ b/src/server/hset_family.cc
@@ -50,12 +50,12 @@ using OptStr = std::optional<std::string>;
 enum GetAllMode : uint8_t { FIELDS = 1, VALUES = 2 };
 
 // TODO: replace all the listpack code with our detail::Listpack wrapper.
-bool IsGoodForListpack(CmdArgList args, const uint8_t* lp) {
+bool IsGoodForListpack(const ParsedArgs& args, const uint8_t* lp) {
   DCHECK_GE(args.size(), 2u);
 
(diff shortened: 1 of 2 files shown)


src/server/hset_family.cc
```diff
@@ -1165,12 +1171,12 @@ void CmdHSet(CmdArgList args, CommandContext* cmd_cntx) {
...
-void CmdHSetNx(CmdArgList args, CommandContext* cmd_cntx) {
-  string_view key = ArgS(args, 0);
+void CmdHSetNx(CmdArgParser parser, CommandContext* cmd_cntx) {
+  string_view key = parser.Next();
 
-  auto cb = [&](Transaction* t, EngineShard* shard) {
-    return OpSet(t->GetOpArgs(shard), key, args.subspan(1),
-                 OpSetParams{.mode = OpSetParams::Mode::kNX});
+  ParsedArgs values = parser.UnparsedArgs();
+  auto cb = [&, values](Transaction* t, EngineShard* shard) {
```
abhijat, line 1178: I suppose this is fine as we're just copying the `values` handle and not the wrapped string values

abhijat (approved):

src/server/hset_family.cc
romange: yes, indeed

Issue #7199, opened 2026-04-21
Add remaining scorers (BM25STD.NORM, BM25STD.TANH, DISMAX, DOCSCORE)


vyavdoshenko:
Dragonfly search currently supports three scorers: BM25STD (default), TFIDF and TFIDF.DOCNORM are in progress. For full feature parity with competitors, the following scorers still need to be implemented.

Scorers to add:

**BM25STD.NORM**
Min-max normalized BM25STD. Scores are rescaled to [0, 1] across the result set:
`normalized_score = (score - min_score) / (max_score - min_score)`

Requires a two-pass approach in `TakeScoredTopK`:
1. First pass: compute raw BM25STD scores for all matched docs, track min/max
2. Second pass: rescale before returning top-K

Alternatively, compute raw scores, then normalize just the top-K before output (simpler but min/max are over the top-K subset, not the full result set - matches Redis Stack's behavior).

See: https://redis.io/docs/latest/develop/ai/search-and-query/advanced-concepts/scoring/

**BM25STD.TANH**
BM25STD with tanh-based normalization: `tanh(score * alpha)` where alpha is a configurable sharpness parameter. Scores are bounded in (0, 1) without requiring min/max. Single-pass, but requires an API way to pass the alpha parameter.

**DISMAX**
Disjunction Max: takes the **maximum** per-term score instead of the sum. Changes the aggregation in `ScoreDocument` from `+=` to `std::max`. Requires either a new path in the switch or a template parameter/callback for the aggregation operator.

Useful for OR-heavy queries where summing multiple term scores over-inflates documents that match many loosely-related terms.

**DOCSCORE**
Returns the document's static score set at index time (currently `DocIndex.default_score = 1`). Does not use term frequencies at all.

Requires:
- Per-document score storage (if FT.CREATE ever gains a per-doc WEIGHT parameter)
- Or just returns the index-wide `default_score` for all matches (simpler, matches current Redis behavior when no per-doc score is set)
- Plumbing `default_score` from `DocIndex` through to the scoring context

**HAMMING**
Hamming distance between query payload and document payload. Dragonfly does not support document payloads, so this scorer is out of scope until/unless payload support is added.

- Scoring documentation: https://redis.io/docs/latest/develop/ai/search-and-query/advanced-concepts/scoring/
- Current implementation: #7093 (TF storage), #7101 (BM25STD engine), #7181 (API integration)

vyavdoshenko:
[...]

PR #7692, opened 2026-06-24
feat: Add BM25STD.NORM and BM25STD.TANH scorers

vyavdoshenko:
Add `BM25STD.NORM` and `BM25STD.TANH` as normalized variants of the `BM25STD` scorer, for `FT.SEARCH`, `FT.AGGREGATE`, and `FT.HYBRID`.

What's added:
- **`SCORER BM25STD.NORM`** - divides each score by the max score over the full matched set (`score / max_score`), so the best matched document gets `1.0`. The max is global across shards (per-shard local max merged at the coordinator); `LIMIT`/`OFFSET` does not change a document's score.
- **`SCORER BM25STD.TANH`** - `tanh(score/factor)`, single-pass, no max needed. `BM25STD_TANH_FACTOR` is a query-level positive integer (default `4`); `0`/negatives/ non-integers are rejected with `BM25STD_TANH_FACTOR must be between 1 and 10000 inclusive`.
- Internally, `ScorerFn` is replaced by a `ScorerSpec` descriptor (kind + tanh factor); the raw `BM25Std` formula is unchanged.

Tests:
- Core scoring + `FT.SEARCH`/`FT.AGGREGATE`/`FT.HYBRID` unit tests, including factor validation, vector+text aggregate (post-vector-selection max; vector-only -> `0`), KNN+prefilter, `FT.PROFILE`, and a multi-shard single-global-winner test.
- Python integration: cross-shard consistency (1-shard vs 4-shard) for all scorers and a multi-shard `BM25STD.NORM` single-winner test.

Part of #7199 (`DISMAX` and `DOCSCORE` are out of scope here).

### DIFF
--- a/src/core/search/search.cc
+++ b/src/core/search/search.cc
@@ -749,6 +749,9 @@ struct BasicSearch {
     scored.reserve(all_docs.size());
     vector<ScoringTermInfo> term_infos(cursors.size());
 
+    // Track the max score over the full matched set (before top-K trimming) so the command
+    // layer can normalize BM25STD.NORM by the global max. Cheap byproduct of scoring.
+    float max_text_score = 0.0f;
     for (DocId doc : all_docs) {
       for (size_t t = 0; t < cursors.size(); t++) {
         term_infos[t].term_docs = cursors[t].term_docs;
@@ -758,10 +761,10 @@ struct BasicSearch {
           term_infos[t].field_avg_doc_len = cursors[t].field_avg_doc_len;
         }
       }
-      scored.emplace_back(static_cast<float>(ScoreDocument(*scorer_, ctx, term_infos)), doc);
+      float score = static_cast<float>(ScoreDocument(*scorer_, ctx, term_infos));
+      max_text_score = max(max_text_score, score);
+      scored.emplace_back(score, doc);
     }
-    auto max_it = max_element(scored.begin(), scored.end());
-    float max_text_score = max_it == scored.end() ? 0.0f : max_it->first;
 
     // Top-K by score (skip sort when no actual cutoff, e.g. FT.AGGREGATE)
     size_t k = min(limit, scored.size());
@@ -1233,17 +1236,6 @@ void SearchAlgorithm::EnableProfiling() {
 
 void SearchAlgorithm::SetScorer(ScorerSpec scorer) {
   scorer_ = scorer;
-}
-
-void SearchAlgorithm::SetScorer(ScorerFn scorer) {
-  if (scorer == &BM25Std)
-    scorer_ = ScorerSpec{ScorerKind::BM25STD};
-  else if (scorer == &TfIdf)
-    scorer_ = ScorerSpec{ScorerKind::TFIDF};
-  else if (scorer == &TfIdfDocNorm)
-    scorer_ = ScorerSpec{ScorerKind::TFIDF_DOCNORM};
-  else
-    scorer_ = nullopt;
 }
 
 // Visits `node` and recurses into its sub-expressions, invoking `cb` on every node in DFS order.
--- a/src/core/search/search.cc
+++ b/src/core/search/search.cc
@@ -119,7 +119,8 @@ struct ProfileBuilder {
 struct BasicSearch {
   using LogicOp = AstLogicalNode::LogicOp;
 
-  BasicSearch(const FieldIndices* indices, ScorerFn scorer, const GlobalScoringStats* global_stats)
+  BasicSearch(const FieldIndices* indices, optional<ScorerSpec> scorer,
+              const GlobalScoringStats* global_stats)
       : indices_{indices}, scorer_{scorer}, global_stats_{global_stats} {
   }
 
@@ -755,7 +756,7 @@ struct BasicSearch {
           term_infos[t].field_avg_doc_len = cursors[t].field_avg_doc_len;
         }
       }
-      scored.emplace_back(static_cast<float>(ScoreDocument(scorer_, ctx, term_infos)), doc);
+      scored.emplace_back(static_cast<float>(ScoreDocument(*scorer_, ctx, term_infos)), doc);
     }
 
     // Top-K by score (skip sort when no actual cutoff, e.g. FT.AGGREGATE)
@@ -784,7 +785,7 @@ struct BasicSearch {
   }
 
   const FieldIndices* indices_;
-  ScorerFn scorer_ = nullptr;
+  optional<ScorerSpec> scorer_;
   const GlobalScoringStats* global_stats_ = nullptr;
 
   string error_;
@@ -1226,8 +1227,19 @@ void SearchAlgorithm::EnableProfiling() {
   profiling_enabled_ = true;
 }
 
-void SearchAlgorithm::SetScorer(ScorerFn scorer) {
+void SearchAlgorithm::SetScorer(ScorerSpec scorer) {
   scorer_ = scorer;
+}
+
+void SearchAlgorithm::SetScorer(ScorerFn scorer) {
+  if (scorer == &BM25Std)
+    scorer_ = ScorerSpec{ScorerKind::BM25STD};
+  else if (scorer == &TfIdf)
+    scorer_ = ScorerSpec{ScorerKind::TFIDF};
+  else if (scorer == &TfIdfDocNorm)
+    scorer_ = ScorerSpec{ScorerKind::TFIDF_DOCNORM};
+  else
+    scorer_ = nullopt;
 }
 
 // Visits `node` and recurses into its sub-expressions, invoking `cb` on every node in DFS order.
--- a/src/core/search/search.cc
+++ b/src/core/search/search.cc
@@ -684,15 +684,16 @@ struct BasicSearch {
 
     if (scorer_ && !matched_text_terms_.empty()) {
       // Score ALL matched docs and return top-K by score (not arbitrary cutoff).
-      auto [out, total_size, text_scores] = TakeScoredTopK(std::move(result), cuttoff_limit);
+      auto [out, total_size, text_scores, max_text_score] =
+          TakeScoredTopK(std::move(result), cuttoff_limit);
       return SearchResult{
-          total_size,         std::move(out),   std::move(knn_scores_), std::move(text_scores),
-          std::move(profile), std::move(error_)};
+          total_size,     std::move(out),     std::move(knn_scores_), std::move(text_scores),
+          max_text_score, std::move(profile), std::move(error_)};
     }
 
     auto [out, total_size] = result.Take(cuttoff_limit);
-    return SearchResult{total_size, std::move(out),     std::move(knn_scores_),
-                        {},         std::move(profile), std::move(error_)};
+    return SearchResult{total_size, std::move(out),     std::move(knn_scores_), {},
+                        0.0f,       std::move(profile), std::move(error_)};
   }
 
  private:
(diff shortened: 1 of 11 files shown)


src/core/search/search.cc
```diff
@@ -755,7 +761,9 @@ struct BasicSearch {
           term_infos[t].field_avg_doc_len = cursors[t].field_avg_doc_len;
         }
       }
-      scored.emplace_back(static_cast<float>(ScoreDocument(scorer_, ctx, term_infos)), doc);
+      float score = static_cast<float>(ScoreDocument(*scorer_, ctx, term_infos));
```
BorysTheDev, line 764: why do we use double to cast to float, maybe we should use float everywhere instead?

src/server/search/doc_index.cc
```diff
@@ -1014,7 +1014,14 @@ SearchResult ShardDocIndex::Search(const OpArgs& op_args, const SearchParams& pa
         {result.ids[i], string{key}, std::move(fields), knn_score, text_score, sort_score});
   }
 
-  return {result.total - expired_count, std::move(out), std::move(result.profile)};
+  // Recompute the max over docs that survived loading. Expired/missing docs were dropped above,
+  // so the pre-filter result.max_text_score could otherwise normalize BM25STD.NORM by a document
+  // that is not returned, pushing the best returned document below 1.0.
+  float max_text_score = 0.0f;
+  for (const auto& doc : out)
+    max_text_score = std::max(max_text_score, doc.text_score);
```
BorysTheDev: smell like std::max_element or ranges::max_element

src/server/search/doc_index.cc
```diff
@@ -1040,14 +1047,20 @@ SearchIdResult ShardDocIndex::SearchIds(const OpArgs& op_args, const SearchParam
         live_text_scores[id] = it->second;
     }
 
+    // Recompute the max over live docs only, mirroring ShardDocIndex::Search: a dropped
+    // top-scoring doc must not inflate the BM25STD.NORM denominator.
+    float max_text_score = 0.0f;
+    for (const auto& [id, score] : live_text_scores)
+      max_text_score = std::max(max_text_score, score);
```
BorysTheDev: the same

src/server/search/doc_index.cc
```diff
@@ -1131,9 +1144,11 @@ vector<SearchDocData> ShardDocIndex::LoadDocEntriesWithScores(
         out.back()[string{score_alias}] = static_cast<double>(it->second);
     }
 
-    if (!text_score_map.empty()) {
+    if (params.add_scores) {
+      double text_score = 0;
       if (auto it = text_score_map.find(doc); it != text_score_map.end())
-        out.back()["__score"] = static_cast<double>(it->second);
+        text_score = it->second;
```
BorysTheDev: nit: 
auto it = text_score_map.find(doc);
out.back()["__score"] = it != text_score_map.end() ? it->second : 0.0;

src/server/search/search_family.cc
```diff
@@ -511,14 +512,39 @@ search::QueryParams ParseQueryParams(CmdArgParser* parser) {
...
+
+constexpr string_view kBM25StdTanhFactorErr =
+    "BM25STD_TANH_FACTOR must be between 1 and 10000 inclusive";
+
+std::optional<uint64_t> TryParseBM25StdTanhFactor(CmdArgParser* parser) {
+  if (!parser->HasNext())
+    return nullopt;
+
+  uint64_t factor = 0;
+  if (!absl::SimpleAtoi(parser->Next(), &factor) || factor == 0)
```
BorysTheDev: use parser and FInt and you will get error

BorysTheDev: if you need custom error, you can extend parser functionality if we don't have such

src/server/search/search_family.cc
```diff
@@ -561,6 +587,11 @@ ParseResult<SearchParams> ParseSearchParams(CmdArgParser* parser) {
       if (!scorer)
         return CreateSyntaxError(absl::StrCat("No such scorer: ", parser->Peek()));
       params.scorer = *scorer;
+    } else if (parser->Check("BM25STD_TANH_FACTOR")) {
+      auto factor = ParseBM25StdTanhFactor(parser);
+      if (!factor)
+        return make_unexpected(factor.error());
+      tanh_factor = *factor;
```
BorysTheDev: it feels that this code should be simplified with new parser abilities

src/server/search/search_family.cc
```diff
@@ -1179,6 +1224,19 @@ void PartialSortBySortScore(absl::Span<SerializedSearchDoc*> docs, size_t limit,
...
+  return scorer && scorer->kind == search::ScorerKind::BM25STD_NORM;
+}
+
+// Global max raw text score across per-shard results, used to normalize BM25STD.NORM by the
+// max over the full matched set. Returns 0 when there are no scored docs.
+float GlobalMaxTextScore(absl::Span<const SearchResult> results) {
+  float max_text_score = 0.0f;
+  for (const auto& shard : results)
+    max_text_score = std::max(max_text_score, shard.max_text_score);
+  return max_text_score;
```
BorysTheDev: max_element

src/server/search/search_family.cc
```diff
@@ -1745,10 +1813,20 @@ ParseResult<HybridSearchParams> ParseHybridParams(CmdArgParser* parser) {
     else
       p.scorer = *scorer_fn;
   };
+  auto parse_tanh_factor = [&](CmdArgParser* sub) {
+    auto factor = TryParseBM25StdTanhFactor(sub);
+    if (!factor)
+      sub->ReportCustom(std::string{kBM25StdTanhFactorErr});
+    else
```
BorysTheDev: we should do it using parser

src/server/search/search_family.cc
```diff
@@ -1826,29 +1904,14 @@ ParseResult<HybridSearchParams> ParseHybridParams(CmdArgParser* parser) {
...
-
   for (auto& [_, e] : docs) {
-    float norm_text = (e.has_text && max_text > 0.f) ? e.text_score / max_text : 0.f;
-    float norm_knn = 0.f;
+    float text_score = e.has_text ? e.text_score : 0.f;
+    float knn_score = 0.f;
     if (e.has_knn)
-      norm_knn = (dist_range > 0.f) ? 1.f - (e.knn_dist - min_dist) / dist_range : 1.f;
-    e.combined = alpha * norm_text + beta * norm_knn;  // ALPHA=text, BETA=vector
+      knn_score = search::DistanceToSimilarity(e.knn_dist, vsim_metric);
```
BorysTheDev: float knn_score = e.has_knn ? search::DistanceToSimilarity(e.knn_dist, vsim_metric) : 0.f;

src/server/search/search_family.cc
```diff
@@ -2112,6 +2175,40 @@ HybridDocMap MergeHybridResults(const vector<SearchResult>& text_docs,
...
+void NormalizeAggregateScores(absl::Span<aggregate::DocValues> values) {
+  static constexpr std::string_view kScoreField = "__score";
+
+  double max_score = 0;
+  for (const auto& row : values) {
+    auto it = row.find(kScoreField);
+    if (it == row.end())
+      continue;
+    if (const auto* score = std::get_if<double>(&it->second))
+      max_score = std::max(max_score, *score);
```
BorysTheDev:
```
auto get_score = [] (const auto& row) {
    if (auto it = row.find(kScoreField); it != row.end())
        if (const auto* score = std::get_if<double>(&it->second))
          return *score;
     return 0;
}

max_element(values, [&get_score] (auto row1, auto row2) {return get_score(row1) < et_score(row2); })
```

src/server/search/search_family.cc
```diff
@@ -2112,6 +2175,40 @@ HybridDocMap MergeHybridResults(const vector<SearchResult>& text_docs,
...
+  if (max_score == 0)
+    return;
+
+  for (auto& row : values) {
+    auto it = row.find(kScoreField);
+    if (it == row.end())
+      continue;
+    if (auto* score = std::get_if<double>(&it->second))
+      *score /= max_score;
+  }
```
BorysTheDev: consider using ranges::for_each

src/core/search/search.cc
vyavdoshenko: ScoreDocument genuinely computes in double (IDF log, TF divisions, per-term sum), and everything downstream is already float, so this cast is a real narrowing, not a relabel. Going float-everywhere would shift low-order bits (rare tie flips) and break the EXPECT_DOUBLE_EQ scorer tests. It also predates this PR, so I'd keep it as-is.

src/server/search/doc_index.cc
vyavdoshenko: fixed

src/server/search/doc_index.cc
vyavdoshenko: fixed

src/server/search/doc_index.cc
vyavdoshenko: fixed

src/server/search/search_family.cc

src/server/search/search_family.cc

src/server/search/search_family.cc

src/server/search/search_family.cc

src/server/search/search_family.cc

src/server/search/search_family.cc

src/server/search/search_family.cc

src/facade/cmd_arg_parser.h
```diff
@@ -184,6 +186,25 @@ struct CmdArgParser {
...
+      return {};
+    }
+    size_t idx = cur_i_++;
+    T out{};
+    if (!absl::SimpleAtoi(SafeSV(idx), &out) || out < min || out > max) {
+      Report(CUSTOM_ERROR, idx, std::string{err_msg});
+      return {};
+    }
+    return out;
+  }
```
BorysTheDev: we have fint type. Consider update Next() method with custom error

src/server/search/doc_index.cc
```diff
@@ -1018,8 +1018,10 @@ SearchResult ShardDocIndex::Search(const OpArgs& op_args, const SearchParams& pa
   // so the pre-filter result.max_text_score could otherwise normalize BM25STD.NORM by a document
   // that is not returned, pushing the best returned document below 1.0.
   float max_text_score = 0.0f;
-  for (const auto& doc : out)
-    max_text_score = std::max(max_text_score, doc.text_score);
+  if (!out.empty())
+    max_text_score = std::max_element(out.begin(), out.end(), [](const auto& a, const auto& b) {
+                       return a.text_score < b.text_score;
+                     })->text_score;
```
BorysTheDev: nit:
```
auto max_it = ranges::max_element(out, [](const auto& a, const auto& b) {
                       return a.text_score < b.text_score;
                     })
float max_text_score = max_it == out.end() ? 0.f : max_it->text_score;
```

src/server/search/search_family.cc
```diff
@@ -2180,33 +2158,36 @@ void NormalizeHybridTextScores(absl::Span<SearchResult> text_docs) {
...
     return;
 
-  for (auto& shard : text_docs) {
-    for (auto& doc : shard.docs)
-      doc.text_score /= max_text_score;
-  }
+  std::ranges::for_each(text_docs, [max_text_score](SearchResult& shard) {
+    std::ranges::for_each(shard.docs,
+                          [max_text_score](auto& doc) { doc.text_score /= max_text_score; });
+  });
```
BorysTheDev: I liked the previous implementation more.

src/server/search/search_family.cc
```diff
@@ -2180,33 +2158,36 @@ void NormalizeHybridTextScores(absl::Span<SearchResult> text_docs) {
...
+        return *score;
+    return 0.0;
+  };
+
+  if (values.empty())
+    return;
+
+  double max_score = get_score(*std::max_element(
+      values.begin(), values.end(),
+      [&get_score](const auto& a, const auto& b) { return get_score(a) < get_score(b); }));
```
BorysTheDev: to avoid extra get_score use std::views::transform() to iterate score instead of value. you can use ranges::max

src/facade/cmd_arg_parser.h
vyavdoshenko: fixed

src/server/search/doc_index.cc
vyavdoshenko: fixed

src/server/search/search_family.cc

src/server/search/search_family.cc

BorysTheDev (approved):

Issue #7676, opened 2026-06-22
consolidate hashing computations


romange:
We need to track XXH computations for a single GET command.
This is what comes to my mind:
1. Shard id (hash % modulo)
2. transaction hash-tag fingerprints
3. Dash-table Find

Maybe more?
Consider ditching XXHash in favor of RapidHash_Micro - https://github.com/Nicoshev/rapidhash or wyhash.

1. shard modulo can benefit from Lemire trick: https://lemire.me/blog/2016/06/27/a-fast-alternative-to-the-modulo-reduction/  - should we fold the hash into 32bit for this?
2. Dash tables also have modulo. Need to check
3. Can we reuse key hashes? Maybe not for hashtags (which are not keys in general case) but for ShardId/Dash. Does it make sense?

PR #7694, opened 2026-06-24
chore: Bring in rapidhash

romange:
Currently only in ok_main.cc, mainly to demo it.
The CPU usage of KeyShard is 1% at most but in Dragonfly the CPU usage of XXH64 is higher.

The goal of this change is to have rapidhash part of the codebase and investigate "reduce" function as a replacement to modulo operation. Partially addresses #7676

### DIFF
--- a/src/facade/ok_main.cc
+++ b/src/facade/ok_main.cc
@@ -11,7 +11,6 @@
 // supported by the caller.
 
 #include <mimalloc-new-delete.h>  // Routes global operator new/delete through mimalloc.
-#include <xxhash.h>
 
 #include <coroutine>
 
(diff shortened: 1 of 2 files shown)


src/facade/ok_main.cc
```diff
@@ -97,11 +97,16 @@ Shard& MyShard() {
...
+  return (uint64_t(h) * uint64_t(n)) >> 32;
+}
+
 // Determine the owning shard for a given key.
-constexpr uint64_t kShardHashSeed = 120577240643ULL;
 unsigned KeyShard(string_view key, unsigned num_shards) {
-  XXH64_hash_t hash = XXH64(key.data(), key.size(), kShardHashSeed);
-  return hash % num_shards;
+  uint64_t hash = rapidhashMicro(key.data(), key.size());
+  return reduce(hash, num_shards);
```
kostasrim, line 109: it's sneaky to allow an implicit narrowing conversion to enable semantics. Ain't production code so it doesn't really matter though tbh

kostasrim (approved):

PR #7695, opened 2026-06-24
refactor(server): factor out metrics code into metrics.{h,cc}

romange:
## Summary
Factors the metrics aggregation and Prometheus rendering out of the oversized `server_family.{h,cc}` into a dedicated `metrics.{h,cc}` translation unit. Pure code movement — no behavior change.

## Changes
- Move `Metrics`, `ReplicationMemoryStats`, `LoadingStats`, `PeakStats` to `metrics.h`
- Convert the free function `PrintPrometheusMetrics()` into a `Metrics::Print()` method; move `Merge()` and `InitFromThread()` to `metrics.cc`
- Promote `GetDelayMs`/`ReadProcStats` to header-declared free functions so both `server_family.cc` (INFO command) and `metrics.cc` can share them
- Register `metrics.cc` in `dragonfly_lib`


src/server/metrics.cc
```diff
@@ -0,0 +1,858 @@
...
+#include "facade/dragonfly_listener.h"
+#include "facade/reply_builder.h"
+#include "io/proc_reader.h"
+#include "server/cluster_support.h"
+#include "server/command_registry.h"
+#include "server/dflycmd.h"
+#include "server/journal/journal.h"
+#include "server/namespaces.h"
+#include "server/search/doc_index.h"
+#include "server/server_family.h"
```
BorysTheDev, line 23: do we really need it? it looks like we can move 
```
bool ReadProcStats(io::StatusData* sdata);
uint64_t GetDelayMs(uint64_t cycles_ts);
```
to another header or even here and remove this huge include

BorysTheDev (approved):

src/server/metrics.cc
romange: you mean just declare it here?
we can but I do not think it will save you a lot.

BorysTheDev: we use them only in this file, you can move it completely, including definition. it is not critical I've already approved

romange: no, we also use them in server_family.cc as we also render "info" output.

PR #7696, opened 2026-06-24
feat(search): support EF_RUNTIME for HNSW vector search

vyavdoshenko:
`EF_RUNTIME` was parsed but ignored for HNSW vector indexes, and KNN searches always used a hardcoded search breadth of 10. This makes `EF_RUNTIME` functional for HNSW index defaults and per-query KNN overrides, while avoiding shared mutable HNSW state during search.
 
 Changes:
 - Store `EF_RUNTIME` on HNSW vector field params, defaulting to 10.
 - Use the index-level `EF_RUNTIME` as the default search breadth.
 - Pass per-query `EF_RUNTIME` into HNSW search calls instead of mutating shared `world_.ef_`.
 - Report `ef_runtime` in `FT.INFO`.
 - Round-trip `EF_RUNTIME` through `BuildRestoreCommand` (`HNSW 12` -> `HNSW 14`).
 - Add coverage for `FT.INFO`, restore command generation, and a deterministic recall test showing that wider `EF_RUNTIME` affects HNSW search results.
 
 Backward compatibility:
 Indexes that do not specify `EF_RUNTIME` keep the previous behavior with search breadth 10.
 
 Note:
 KNN query-attributes syntax is intentionally left for a follow-up PR.

### DIFF
--- a/src/core/search/ast_expr.h
+++ b/src/core/search/ast_expr.h
@@ -149,7 +149,7 @@ struct AstTagsNode {
 struct AstKnnNode {
   AstKnnNode() = default;
   AstKnnNode(uint32_t limit, std::string_view field, OwnedFtVector vec,
-             std::string_view score_alias, std::optional<size_t> ef_runtime);
+             std::string_view score_alias, std::optional<uint32_t> ef_runtime);
 
   AstKnnNode(AstNode&& sub, AstKnnNode&& self);
 
@@ -168,7 +168,7 @@ struct AstKnnNode {
   std::string field;
   OwnedFtVector vec;
   std::string score_alias;
-  std::optional<size_t> ef_runtime;
+  std::optional<uint32_t> ef_runtime;
 
   bool HasPreFilter() const;
 };
@@ -224,6 +224,6 @@ using AstExpr = AstNode;
 }  // namespace dfly
 
 namespace std {
-ostream& operator<<(ostream& os, optional<size_t> o);
+ostream& operator<<(ostream& os, optional<uint32_t> o);
 ostream& operator<<(ostream& os, dfly::search::AstTagsNode::TagValueProxy o);
 }  // namespace std
--- a/src/core/search/ast_expr.h
+++ b/src/core/search/ast_expr.h
@@ -7,6 +7,7 @@
 #include <algorithm>
 #include <iosfwd>
 #include <memory>
+#include <optional>
 #include <variant>
 #include <vector>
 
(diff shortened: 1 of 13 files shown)


vyavdoshenko:
augment review

src/core/search/ast_expr.h
```diff
@@ -167,7 +168,7 @@ struct AstKnnNode {
   std::string field;
   OwnedFtVector vec;
   std::string score_alias;
-  std::optional<float> ef_runtime;
+  std::optional<size_t> ef_runtime;
```
romange: nit: it's a very small number so `uint32_t ` or even `uint16_t` should suffice

src/core/search/hnsw_alg.h
```diff
@@ -1380,7 +1376,14 @@ template <typename dist_t> class HierarchicalNSW : public hnswlib::AlgorithmInte
...
   std::priority_queue<std::pair<dist_t, labeltype>> searchKnn(
-      const void* query_data, size_t k, BaseFilterFunctor* isIdAllowed = nullptr) const {
+      const void* query_data, size_t k, BaseFilterFunctor* isIdAllowed = nullptr) const override {
+    return searchKnnWithEf(query_data, k, isIdAllowed, ef_);
+  }
+
+  std::priority_queue<std::pair<dist_t, labeltype>> searchKnnWithEf(const void* query_data,
+                                                                    size_t k,
+                                                                    BaseFilterFunctor* isIdAllowed,
+                                                                    size_t ef_runtime) const {
```
romange: same

src/core/search/parser.y
```diff
@@ -119,13 +128,18 @@ final_query:
       { driver->Set(AstKnnNode(std::move($1), std::move($3))); }
 
 knn_query:
-  LBRACKET KNN UINT32 FIELD TERM opt_ef_runtime opt_knn_alias RBRACKET
+  LBRACKET KNN UINT32 FIELD TERM opt_ef_runtime opt_knn_alias RBRACKET opt_knn_attrs
```
romange: suggestion: to split it into two PRs: the parsing part and the rest.

vyavdoshenko: The parser part will be in the following PR

src/core/search/ast_expr.h
vyavdoshenko: fixed

src/core/search/hnsw_alg.h
vyavdoshenko: fixed

romange (approved):

PR #7699, opened 2026-06-25
tests: ensure that proxy chooses random port and closes on teardown

abhijat:
previously proxy chose fixed port did not close on failing path.
this causes cascading errors on CI sometimes:

* test 1 fails, does not stop proxy cleanly
* test 2 cannot use same port, fails
* test 3 cannot reuse same port, fails
* ...

The test 1 still has to be fixed, but we should not leak processes and also try to use non-fixed ports to avoid such cascades. proxy already supports random port with `port=0` so we choose that.

### DIFF
--- a/tests/dragonfly/proxy.py
+++ b/tests/dragonfly/proxy.py
@@ -11,6 +11,7 @@ class Proxy:
         self.stop_connections = []
         self.server = None
         self._handler_tasks = set()
+        self._serve_task = None
 
     async def handle(self, reader, writer):
         task = asyncio.current_task()
@@ -65,9 +66,23 @@ class Proxy:
             _, port = self.server.sockets[0].getsockname()[:2]
             self.port = port
 
-    async def serve(self):
-        async with self.server:
-            await self.server.serve_forever()
+    async def start_serving(self):
+        await self.start()
+        self._serve_task = asyncio.create_task(self.serve(self.server))
+
+    async def serve(self, server=None):
+        server = server or self.server
+        if server is None:
+            return
+        async with server:
+            await server.serve_forever()
+
+    async def __aenter__(self):
+        await self.start_serving()
+        return self
+
+    async def __aexit__(self, exc_type, exc, tb):
+        await self.close()
 
     def drop_connection(self):
         """
@@ -79,6 +94,11 @@ class Proxy:
             cb()
 
     async def close(self, task=None):
+        if task is None:
+            task = self._serve_task
+        if task is self._serve_task:
+            self._serve_task = None
+
         if self.server is not None:
             self.server.close()
             self.server = None
--- a/tests/dragonfly/redis_replication_test.py
+++ b/tests/dragonfly/redis_replication_test.py
@@ -330,60 +330,53 @@ async def test_disconnect_master(
     c_master = aioredis.Redis(port=master.port)
     assert await c_master.ping()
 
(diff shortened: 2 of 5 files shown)


abhijat:
example of issue https://github.com/dragonflydb/dragonfly/actions/runs/28008791181/job/82896808782

tests/dragonfly/redis_replication_test.py
```diff
@@ -1,11 +1,14 @@
...
-import pytest
 import asyncio
-from redis import asyncio as aioredis
-import subprocess
-from .utility import *
+import logging
+import random
+import string
+
+import aioredis
```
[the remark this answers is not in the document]
abhijat: fixed to be same as other tests

tests/dragonfly/redis_replication_test.py
```diff
@@ -1,11 +1,14 @@
...
+import pytest
+
```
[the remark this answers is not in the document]
abhijat: changed to make it into

```
import redis.asyncio as aioredis
```

abhijat:
augment review

tests/dragonfly/redis_replication_test.py
```diff
@@ -266,29 +276,35 @@ async def test_redis_master_restart(
 
     # Start data stream
     stream_task = asyncio.create_task(seeder.run())
-    await asyncio.sleep(0.0)
+    try:
```
kostasrim, line 278: @abhijat that would require a `try/catch` wherever we do resource management right ? This is not a very scalable solution.

Two alternatives:

1. Make proxy/streamer a fixture. That way cleanup is done `unconditionally at teardown` 
2. Make Proxy/seeder-stream async context managers:

```
@asynccontextmanager
async def streaming(seeder):
  task = asyncio.create_task(seeder.run())
  try:
    yield task
  finally:
   ...
```

That way we get self contained changes that do not require try/catching the whole block

abhijat: yeah I thought of a context manager too, I'll look into it, I suppose there are more places this pattern leaks into

kostasrim: yes and it seems that a Context maanger will do the same as what the try/catch would only with a very small line number of changes. then just ask the ai to replace all occurencies in pytests and we have a fix that works for all tests 😄

kostasrim: if I missed anything and it's not trivial just let me know and I will accept it as is

abhijat: @kostasrim please take a look, I tried to address it

tests/dragonfly/proxy.py
```diff
@@ -79,6 +94,11 @@ def drop_connection(self):
             cb()
 
     async def close(self, task=None):
```
kostasrim, line 96: We should not need close or server (for serve function) parameters at all since their scope/fixture manages that. 

I guess we left some of the old cases around -- if you wish you can follow up in another small PR 😄

kostasrim (approved):
It actually removed most of the try/catch blocks with 2 exceptions 😄

tests/dragonfly/proxy.py
abhijat: yeah Im planning to do a follow up once the tests are stable, I think there are more places to clean up

PR #7704, opened 2026-06-25
Fix memtier latency plot for zero-count operations

romange:
## Summary
- skip memtier operations that do not include latency time-series samples
- tolerate missing per-interval latency fields when extracting chart data

## Validation
- BROWSER=true ./tools/plot_memtier_latency.py /tmp/ssd_tiering_50GB/fill.json /tmp/fill_latency_chart.html
- BROWSER=true ./tools/plot_memtier_latency.py /tmp/ssd_tiering_50GB/fill.json /tmp/fill_latency_chart.svg
- pre-commit run --files tools/plot_memtier_latency.py

### DIFF
--- a/tools/plot_memtier_latency.py
+++ b/tools/plot_memtier_latency.py
@@ -82,15 +82,15 @@ def extract_latency_timeseries(data, operation, ignore_last_seconds=3):
     for time_point in sorted_times:
         interval_data = time_serie[time_point]
         times.append(int(time_point))
-        avg_latencies.append(interval_data["Average Latency"])
-        p50_latencies.append(interval_data.get("p50.00", 0))
-        p99_latencies.append(interval_data.get("p99.00", 0))
-        p99_9_latencies.append(interval_data.get("p99.90", 0))
-        min_latencies.append(interval_data["Min Latency"])
-        max_latencies.append(interval_data["Max Latency"])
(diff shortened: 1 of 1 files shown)


tools/plot_memtier_latency.py
```diff
@@ -104,6 +104,20 @@ def extract_latency_timeseries(data, operation, ignore_last_seconds=3):
...
 
+def get_latency_operations(all_stats):
+    """Return operations with latency time-series samples."""
+    operations = []
+    for key, op_stats in all_stats.items():
+        if key == "Runtime" or not isinstance(op_stats, dict) or "Time-Serie" not in op_stats:
+            continue
+
+        time_serie = op_stats["Time-Serie"]
+        if any("Average Latency" in interval for interval in time_serie.values()):
```
abhijat, line 115: might be safer and simpler to write something like this ( [EAFP](https://docs.python.org/3/glossary.html#term-EAFP) )

```python
try:
     if key != "Runtime" and any("Average Latency" in i for i in op_stats["Time-Serie"].values()):
         operations.append(key)
except:
      pass
```

abhijat (approved):

tools/plot_memtier_latency.py
romange: Yeah, both approaches are fine I guess. I personally do not like using exceptions as part of the normal code logic.

PR #7706, opened 2026-06-25
Add benchmark-report skill

romange:
## Summary

- Adds `.claude/skills/benchmark-report/` skill for generating technical markdown reports from raw memtier benchmark data
- Skill interviews the user for context (instance types, scenario, storage config) before writing, since that information isn't in the raw data
- Parses memtier JSON output and log files to extract throughput and latency percentiles (p50/p99/p99.9)
- Generates interactive latency charts via `tools/plot_memtier_latency.py`
- Produces a structured report with test environment, dataset, benchmark config, results tables, and factual observations

## Writing style enforced

No marketing language ("blazing fast", "industry-leading"), no narrative hooks. Data and percentages only. Missing context gets `[TODO]` stubs rather than guesses.


vyavdoshenko (approved):
lgmt

PR #7712, opened 2026-06-26
chore(server): migrate zset_family handlers to CmdArgParser

romange:
## Summary
First of two stacked changes migrating `zset_family` to `CmdArgParser`. This one covers the **self-contained zset handler migration**; a follow-up converts the shared `ZGetMembers`/`OpMScore` helper.

## Notes
- All registered `zset_family` handlers now take `facade::CmdArgParser`; args are read via the parser, and where the remaining args are needed they use `parser.UnparsedArgs()` (a `facade::ParsedArgs` view) rather than `cmd_cntx->tail_args()`.
- Option / computed-index handlers (`ZADD`, `ZINTERCARD`, `ZDIFF[STORE]`, the `ZRANGE` family, `ZRANGESTORE`) grab `UnparsedArgs()` before consuming so positional indices and error ordering are preserved.
- The set-op parser (`ZUNION`/`ZINTER[STORE]`) and `ZMPOP`/`BZMPOP` validation, plus the `BZPOP`/`ZPOP`/`ZREM` helpers, were converted to consume the parser directly — no `CmdArgVec` materialization. Removed the now-trivial `ParseKeyCount`/`ParseWithScores` helpers.
- The shared `ZGetMembers`/`OpMScore` (also used by geo) are intentionally left as `CmdArgList` here and converted in the follow-up. Registration is unchanged.

## Verification
- Full `dragonfly` binary builds clean off `main`.
- `zset_family_test` passes 48/48.

### DIFF
--- a/src/server/zset_family.cc
+++ b/src/server/zset_family.cc
@@ -1025,89 +1025,64 @@ OpResult<void> FillAggType(string_view agg, SetOpArgs* op_args) {
   return OpStatus::OK;
 }
 
-// Parse functions return the number of arguments read from CmdArgList
-OpResult<unsigned> ParseAggregate(CmdArgList args, bool store, SetOpArgs* op_args) {
-  if (args.size() <= 1) {
+// AGGREGATE <SUM|MIN|MAX>: consumes the aggregation type from the parser.
+OpResult<void> ParseAggregate(CmdArgParser* parser, SetOpArgs* op_args) {
+  if (!parser->HasNext()) {
     return OpStatus::SYNTAX_ERR;
   }
-
-  string agg_type = absl::AsciiStrToUpper(ArgS(args, 1));
-  auto filled = FillAggType(agg_type, op_args);
-  if (!filled) {
-    return filled.status();
-  }
-  return 1;
+  return FillAggType(absl::AsciiStrToUpper(parser->Next()), op_args);
 }
 
-OpResult<unsigned> ParseWeights(CmdArgList args, SetOpArgs* op_args) {
-  if (args.size() <= op_args->num_keys) {
+// WEIGHTS <w1> ... <wN>: consumes num_keys weights from the parser.
+OpResult<void> ParseWeights(CmdArgParser* parser, SetOpArgs* op_args) {
+  if (!parser->HasAtLeast(op_args->num_keys)) {
     return OpStatus::SYNTAX_ERR;
   }
 
   op_args->weights.resize(op_args->num_keys, 1);
   for (unsigned i = 0; i < op_args->num_keys; ++i) {
-    string_view weight = ArgS(args, i + 1);
-    if (!absl::SimpleAtod(weight, &op_args->weights[i])) {
-      return OpStatus::INVALID_FLOAT;
-    }
+    op_args->weights[i] = parser->Next<double>();
   }
 
-  return op_args->num_keys;
-}
-
-OpResult<void> ParseKeyCount(string_view arg_num_keys, SetOpArgs* op_args) {
-  // we parsed the structure before, when transaction has been initialized.
-  if (!absl::SimpleAtoi(arg_num_keys, &op_args->num_keys)) {
-    return OpStatus::SYNTAX_ERR;
+  // Consume (don't just peek) any parse error so the parser doesn't trip its destructor DCHECK.
+  if (parser->TakeError()) {
+    return OpStatus::INVALID_FLOAT;
   }
   return OpStatus::OK;
 }
 
-OpResult<unsigned> ParseWithScores(CmdArgList args, SetOpArgs* op_args) {
-  op_args->with_scores = true;
-  return 0;
-}
-
-OpResult<SetOpArgs> ParseSetOpArgs(CmdArgList args, bool store) {
-  string_view num_keys_str = store ? ArgS(args, 1) : ArgS(args, 0);
+OpResult<SetOpArgs> ParseSetOpArgs(CmdArgParser parser, bool store) {
   SetOpArgs op_args;
 
-  auto parsed = ParseKeyCount(num_keys_str, &op_args);
-  if (!parsed) {
-    return parsed.status();
+  // numkeys was already validated against the arg count during key resolution.
+  op_args.num_keys = parser.Next<unsigned>();
+  if (parser.TakeError() || op_args.num_keys == 0) {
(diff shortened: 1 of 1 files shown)


src/server/zset_family.cc
```diff
@@ -2546,25 +2523,26 @@ void CmdZIncrBy(CmdArgList args, CommandContext* cmd_cntx) {
...
 }
 
-void CmdZInterCard(CmdArgList args, CommandContext* cmd_cntx) {
+void CmdZInterCard(CmdArgParser parser, CommandContext* cmd_cntx) {
+  facade::ParsedArgs args = parser.UnparsedArgs();
   unsigned num_keys;
   auto* builder = cmd_cntx->rb();
 
-  if (!absl::SimpleAtoi(ArgS(args, 0), &num_keys)) {
+  if (!absl::SimpleAtoi(args[0], &num_keys)) {
```
abhijat: I think `parser.Next` can do this directly

src/server/zset_family.cc
```diff
@@ -2627,10 +2606,13 @@ void ZMPopGeneric(CmdArgList args, CommandContext* cmd_cntx, bool is_blocking) {
...
-  ArgRange arg_keys(args.subspan(1 + is_blocking, zmpop_args.num_keys));
-  // Find the first arg_key which exists in any shard and is not empty.
-  for (std::string_view key : arg_keys) {
+  // Skip the leading [timeout] (BZMPOP only) and numkeys args to reach the key list, then find the
+  // first key (in command order) that exists and is non-empty on some shard. The string_views point
+  // into the command's backing args, which outlive this call.
+  parser.Skip(1 + is_blocking);
+  facade::ParsedArgs arg_keys = parser.UnparsedArgs();
+  for (size_t i = 0; i < zmpop_args.num_keys; ++i) {
+    std::string_view key = arg_keys[i];
```
abhijat: why not keep calling `parser.Next()` like

```c++
  parser.Skip(1 + is_blocking);
  for (size_t i = 0; i < zmpop_args.num_keys; ++i) {
    std::string_view key = parser.Next();
    if (first_found_keys_for_shard.contains(key)) {
      key_to_pop = key;
      break;
    }
  }
```

src/server/zset_family.cc
```diff
@@ -1025,89 +1025,64 @@ OpResult<void> FillAggType(string_view agg, SetOpArgs* op_args) {
...
 
-  for (size_t i = opt_args_start; i < args.size(); ++i) {
-    string arg = absl::AsciiStrToUpper(ArgS(args, i));
-    if (arg == "WEIGHTS") {
-      auto parsed_cnt = ParseWeights(args.subspan(i), &op_args);
-      if (!parsed_cnt) {
-        return parsed_cnt.status();
+  while (parser.HasNext()) {
+    if (parser.Check("WEIGHTS")) {
+      if (auto res = ParseWeights(&parser, &op_args); !res) {
```
BorysTheDev: using parser.Apply() this while loop can be rewritten better

src/server/zset_family.cc
```diff
@@ -1732,12 +1710,12 @@ OpResult<ScoredArray> ZPopMinMaxInternal(std::string_view key, FilterShards shou
...
-void ZPopMinMaxFromArgs(CmdArgList args, bool reverse, CommandContext* cmd_cntx) {
-  string_view key = ArgS(args, 0);
+void ZPopMinMaxFromArgs(CmdArgParser parser, bool reverse, CommandContext* cmd_cntx) {
+  string_view key = parser.Next();
   uint32 count = 1;
-  if (args.size() > 1) {
-    string_view count_str = ArgS(args, 1);
-    if (!SimpleAtoi(count_str, &count)) {
+  if (parser.HasNext()) {
+    count = parser.Next<uint32>();
```
BorysTheDev: you can use Check()

abhijat (dismissed):

src/server/zset_family.cc
```diff
@@ -2546,25 +2523,26 @@ void CmdZIncrBy(CmdArgList args, CommandContext* cmd_cntx) {
...
 
-void CmdZInterStore(CmdArgList args, CommandContext* cmd_cntx) {
-  ZBooleanOperation(args, "zinterstore", false, true, cmd_cntx);
+void CmdZInterStore(CmdArgParser parser, CommandContext* cmd_cntx) {
+  ZBooleanOperation(std::move(parser), "zinterstore", false, true, cmd_cntx);
 }
 
-void CmdZInterCard(CmdArgList args, CommandContext* cmd_cntx) {
+void CmdZInterCard(CmdArgParser parser, CommandContext* cmd_cntx) {
+  facade::ParsedArgs args = parser.UnparsedArgs();
```
BorysTheDev: you can do it later when you parse num_keys

BorysTheDev (dismissed):

src/server/zset_family.cc
romange: Done in a01264e1 — `CmdZInterCard` now parses numkeys with `parser.Next<unsigned>()`.

src/server/zset_family.cc
romange: Done — moved the `UnparsedArgs()` snapshot to after numkeys is parsed, so the keys/LIMIT indices no longer carry the leading `1 +` offset.

src/server/zset_family.cc
romange: Done in a01264e1 — the key loop now uses `parser.Next()` after `Skip(1 + is_blocking)`, as suggested.

src/server/zset_family.cc
romange: Done — rewrote the loop with `parser.Apply(Tag("WEIGHTS", ...), Tag("AGGREGATE", Map(...)), If(!store, Exist("WITHSCORES", ...)))` + `Finalize()`, and dropped the now-dead `FillAggType`/`ParseAggregate`/`ParseWeights` helpers.

src/server/zset_family.cc
romange: Replaced the `if (HasNext())` block with `parser.NextOrDefault<uint32>(1)`. `Check()` matches a keyword tag, but the ZPOPMIN `count` is a bare positional arg with no tag, so `NextOrDefault` is the closest fit here — let me know if you meant something else.

BorysTheDev (approved):

PR #7715, opened 2026-06-26
chore(tiering): Make disk storage initial size configurable and add grow policy

mkaruza:
Replace hardcoded kInitialSize with FLAGS_tiering_disk_storage_initial_size (default 256MB),
allowing configuration of initial disk storage size.

Introduce ComputeGrowSize in RequestGrow to replace fixed grow size logic. Growth is now 
computed as 5% of current capacity, rounded up to 256MB alignment.

This makes disk storage sizing fully policy-driven instead of constant-based.

### DIFF
--- a/src/server/tiering/disk_storage.cc
+++ b/src/server/tiering/disk_storage.cc
@@ -237,14 +237,13 @@ DiskStorage::Stats DiskStorage::GetStats() const {
 
 error_code DiskStorage::RequestGrow(off_t min_size) {
   // Request the file to grow by (default ExternalAllocator::kExtAlignment = 256_MB):
-  // - take max(5% of capacity, 256 MB)
+  // - take 5% of capacity
   // - round up to 256 MB boundary
   // - ensure result is at least min_size
   auto ComputeGrowSize = [](size_t capacity, size_t min_size) -> size_t {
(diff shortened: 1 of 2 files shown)


src/server/tiering/disk_storage.cc
```diff
@@ -216,7 +220,21 @@ DiskStorage::Stats DiskStorage::GetStats() const {
...
-error_code DiskStorage::RequestGrow(off_t grow_size) {
+error_code DiskStorage::RequestGrow(off_t min_size) {
+  // Request the file to grow by (default ExternalAllocator::kExtAlignment = 256_MB):
+  // - take max(5% of capacity, 256 MB)
+  // - round up to 256 MB boundary
+  // - ensure result is at least min_size
+  auto ComputeGrowSize = [](size_t capacity, size_t min_size) -> size_t {
+    const size_t kGrowPercent = 5;
+    const size_t kAlign = ExternalAllocator::kExtAlignment;
+    size_t grow = std::max((capacity * kGrowPercent) / 100, kAlign);
```
romange: Seems redundant, no? You apply max later, and round up too, so why do you need max here?

romange (dismissed):

romange (approved):

PR #7716, opened 2026-06-26
refactor: OAHSet - split OAHEntry on 3 classes

BorysTheDev:
# refactor(OAHSet): split OAHEntry into OAHEntry, PtrVector, and OAHPtr

## Summary
The experimental `OAHSet` packed each bucket into a `uint64_t` and `reinterpret_cast` it into
`OAHEntry&` / `PtrVector<OAHEntry>&`, materializing typed references over a type-punned word —
object-lifetime and strict-aliasing UB that can miscompile under optimization/LTO or trip UBSan.
This splits that monolith into three single-responsibility, **non-owning** accessors over the raw
tagged word, so all mutation happens on `uint64_t` bits with no type-punned references.

## Changes
- **OAHEntry** (`oah_entry.h/.cc`): non-owning accessor over a `TaggedPtr` (a tagged `uint64_t`)
  pointing at a `[expiry, key, cached-hash]` blob; `static Create()/Destroy()` own allocation.
- **PtrVector<T>** (new `ptr_vector.h`): non-owning accessor over a `TaggedPtr` encoding
  `(array ptr, count)`; `static Create()/Destroy()`; grows +2 then doubles.
- **OAHPtr** (new `oah_ptr.h`, header-only): bucket-slot manager that discriminates entry vs
  vector (bit 0) and owns the lifetime logic (promote-to-vector, grow, clear).
- Replace all `reinterpret_cast` type-punning with operations on the raw `TaggedPtr`.
- Always store each entry's current ext-hash; rehash recomputes it (drops the lazy-hash paths).
- Swap `OAHSet::Hash` from XXH3 to `rapidhashMicro` (faster for short keys).
- Guard the `PtrVector` log-mode size invariant (power-of-two) with an assert.
- `server_family.cc`: bucket sizing uses `sizeof(uint64_t)` (OAHEntry is now a thin accessor).
- Tests: add PtrVector linear-then-double growth and destructor-coverage cases.

### DIFF
--- a/src/core/oah_entry.h
+++ b/src/core/oah_entry.h
@@ -4,12 +4,13 @@
 
 #pragma once
 
-#include <bit>
-#include <cassert>
-#include <cstring>
-#include <string_view>
+#include <sys/types.h>
 
-#include "base/hash.h"
+#include <cassert>
+#include <cstdint>
+#include <cstring>
+#include <limits>
+#include <string_view>
 
 extern "C" {
 #include "redis/zmalloc.h"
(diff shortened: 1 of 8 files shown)


BorysTheDev:
augment review

src/core/oah_ptr.h
```diff
@@ -0,0 +1,116 @@
...
+      return;
+    if (IsVector())
+      AsVector().~PtrVector<OAHEntry>();
+    else
+      AsEntry().Clear();
+    slot_ = 0;
+  }
+
+ private:
+  uint64_t& slot_;
```
romange: Why is it a reference?  Seems to me as dangerous. Can we avoid it?

src/core/oah_ptr.h
```diff
@@ -0,0 +1,116 @@
...
+#pragma once
+
+#include <cstdint>
+
+#include "core/oah_entry.h"
+#include "core/ptr_vector.h"
+
+namespace dfly {
+
+// A non-owning view over one tagged uint64_t slot that is EITHER a single OAHEntry or
```
romange: once it's a mutable reference it's not a "view".
I wish it were a view

src/core/oah_ptr.h
BorysTheDev: we can use a pointer. we should have the ability to modify the original field.

src/core/oah_entry.h
```diff
@@ -186,64 +50,47 @@ class OAHEntry {
...
   static constexpr size_t kTagMask = (4095ULL << 52) | 7;  // we reserve 12 high bits and 3 low.
 
-  OAHEntry() = default;
+  // Allocates a [expiry, key_size, key] blob and returns the tagged control word (no ext
+  // hash set). The returned uint64_t carries ownership; free it with Destroy().
+  static uint64_t Create(std::string_view key, uint32_t expiry = UINT32_MAX);
 
-  OAHEntry(std::string_view key, uint32_t expiry = UINT32_MAX);
+  // Frees the blob referenced by `bits` (no-op for an empty/zero control word).
+  static void Destroy(uint64_t bits);
```
romange: naming: what `bits` means?

src/core/oah_entry.h
```diff
@@ -186,64 +50,47 @@ class OAHEntry {
...
   }
 
   std::string_view Key() const {
-    assert(!IsVector());
     return {GetKeyData(), GetKeySize()};
   }
 
   bool HasExpiry() const {
-    return (data_ & kExpiryBit) != 0;
+    return (Bits() & kExpiryBit) != 0;
```
romange: I think `Bits` is not a good name too

src/core/oah_entry.h
```diff
@@ -22,158 +23,21 @@ class PageUsage;
 #define PREFETCH_READ(x) __builtin_prefetch(x, 0, 1)
 #define FORCE_INLINE __attribute__((always_inline))
 
-// TODO add allocator support
+// A single set member: a non-owning accessor over one tagged uint64_t control word that
```
romange: you use `uint64_t control word` or slot or bits

and I am trying to understand the jargon. Is not it a tagged pointer?

Also, I am trying to understand the design and responsibilities between this class and OAHPtr

BorysTheDev:
augment review

BorysTheDev:
augment review

src/core/oah_set.h
```diff
@@ -34,17 +41,17 @@ class OAHSet {  // Open Addressing Hash Set
     using iterator_category = std::forward_iterator_tag;
     using difference_type = std::ptrdiff_t;
     using value_type = OAHEntry;
-    using pointer = OAHEntry*;
-    using reference = OAHEntry&;
+    using pointer = OAHEntry;
+    using reference = OAHEntry;
```
vyavdoshenko, line 45: Why do you need these two usings? It looks bad.

src/server/server_family.cc
```diff
@@ -2317,7 +2317,7 @@ void ServerFamily::Shrink(facade::CmdArgParser parser, CommandContext* cmd_cntx)
       // OAHEntry-sized and the entries vector includes displacement slots.
       auto bucket_bytes = [](Set* s) -> size_t {
         if constexpr (std::is_same_v<Set, OAHSet>)
-          return size_t{s->Capacity()} * sizeof(OAHEntry);
+          return size_t{s->Capacity()} * sizeof(uint64_t);
```
vyavdoshenko, line 2320: It looks sucpicious.

src/core/oah_set.h
BorysTheDev: for iterator

src/server/server_family.cc
BorysTheDev: previously OAHentry encapsulated PtrVector. And we store OAHEntry in the OAHSet. Now design more data oriented and to avoid type punning I store uint64_t directly in OAHSet. OAHptr can interpret it as OAHEntry or PtrVector and use corresponding wrapper.

src/core/oah_set.h
vyavdoshenko: But they are neither a reference nor a pointer. This is a bad change because it creates a wrong understanding.

BorysTheDev: OAHentry is reference like entity so I don't see problem with it

BorysTheDev: What is your suggestion?

vyavdoshenko (approved):

kostasrim:
This is *not* a refactor PR. It `silently` changed `eager vs lazy` fingerprint rehashing :smile:

PR #7718, opened 2026-06-26
feat(search): support EPSILON for HNSW vector range search

vyavdoshenko:
Adds `EPSILON` support across the vector search surface. EPSILON is the HNSW range-search overscan factor: the graph traversal frontier is widened to `radius * (1 + epsilon)` to improve recall near the boundary, while results stay filtered by `distance <= radius`. Previously, `EPSILON` was parsed and ignored.

Changes:
- **FT.CREATE**: `VECTOR HNSW ... EPSILON <x>` is now stored and validated - `0` maps to the default `0.01`, negative/non-finite values are rejected, and `EPSILON` is rejected for `FLAT` fields.
- **Runtime override**: `@v:[VECTOR_RANGE <r> $b]=>{$EPSILON: <x>}` for both FT.SEARCH and FT.AGGREGATE. Valid only for HNSW range queries, must be `> 0`, and is rejected on KNN and FLAT.
- **FT.HYBRID**: the `VSIM ... RANGE` clause now applies `EPSILON` and accepts both the shorthand `RANGE <radius> [EPSILON <e>]` and the counted form `RANGE <nargs> RADIUS <r> [EPSILON <e>]` (RADIUS/EPSILON in any order).
- **Engine**: a configurable epsilon (schema default + runtime override) is threaded into `HnswVectorIndex::RangeQuery` and `searchRange`; the dynamic-range boundary update was aligned with the reference algorithm for consistent recall.
- **Persistence & introspection**: `EPSILON` is included in index restore, and FT.INFO now reports `epsilon` and `initial_cap` for vector fields.


src/core/search/hnsw_alg.h
```diff
@@ -1581,7 +1585,15 @@ template <typename dist_t> class HierarchicalNSW : public hnswlib::AlgorithmInte
     // because the entry point is farther than radius.
     dist_t ep_dist = curdist;
     dist_t dynamic_range = std::max(ep_dist, radius);
-    dist_t dyn_boundary = static_cast<dist_t>(dynamic_range * (1.0 + epsilon));
+    auto boundary = [](dist_t range, double eps) {
+      double bound = static_cast<double>(range) * (1.0 + eps);
+      if (!std::isfinite(bound))
+        return std::numeric_limits<dist_t>::max();
+      return static_cast<dist_t>(
+          std::min(bound, static_cast<double>(std::numeric_limits<dist_t>::max())));
```
BorysTheDev: constexpr auto dist_t_max = std::numeric_limits<dist_t>::max()
return !std::isfinite(bound) ? dist_t_max : std::min(static_cast<dist_t>(bound), dist_t_max);

src/facade/cmd_arg_parser.h
```diff
@@ -163,6 +167,20 @@ struct CmdArgParser {
...
+      ReportCustom(std::string{err_msg});
+    return val;
+  }
+
+  double NextNonNegativeDouble(std::string_view err_msg) {
+    double val = Next<double>(err_msg);
+    if (!HasError() && (!(val >= 0) || !std::isfinite(val)))
+      ReportCustom(std::string{err_msg});
+    return val;
+  }
```
BorysTheDev: The idea is good. I would prefer a more generic approach like FInt. We cannot use double as a template parameter, but we can create a constexpr method that can validate the result and use it with CRTP class.

src/facade/cmd_arg_parser.h
```diff
@@ -229,6 +247,59 @@ struct CmdArgParser {
     return res;
   }
 
+  template <class Func>
+  bool TryParseCountedKeyValueList(std::initializer_list<std::string_view> count_forcing_tags,
+                                   std::initializer_list<std::string_view> tags,
+                                   std::string_view context, Func func) {
```
BorysTheDev: I didn't get what it is, and why do we need it

src/server/search/search_family.cc
```diff
@@ -1796,6 +1811,51 @@ ParseResult<HybridSearchParams> ParseHybridParams(CmdArgParser* parser) {
...
+        }
+        double epsilon = opt_parser->NextPositiveDouble("Invalid EPSILON value");
+        if (!opt_parser->HasError() &&
+            !search::SchemaField::VectorParams::IsValidRuntimeHnswEpsilon(epsilon)) {
+          opt_parser->ReportCustom("Invalid EPSILON value");
+          return;
+        }
+        p.range_epsilon = epsilon;
+        has_epsilon = !opt_parser->HasError();
+      }
```
BorysTheDev: looks overcomplicated or incorrect parser usage

src/facade/cmd_arg_parser.h
BorysTheDev: Simpler solution:

```
template <class T>
concept FNum = requires(T val) { 
   static_cast<T::underlying_t>(val);
   { T::validate(std::declval<T::underlying_t>()) } -> std::same_as<bool>;
}
```

we create Next with this concept. Rewrite FInt to use it too. And we can define classes even out of CmdArgParser like 
```
struct PositiveDouble {
   using underlying_t = double;
   bool validate(double val) { return val > 0; }
}; 
```

src/core/search/hnsw_alg.h
vyavdoshenko: fixed

src/facade/cmd_arg_parser.h
vyavdoshenko: fixed

src/facade/cmd_arg_parser.h
vyavdoshenko: removed

src/server/search/search_family.cc

src/facade/cmd_arg_parser.h
```diff
@@ -72,21 +69,52 @@ namespace facade {
...
 };
 
-template <class T> constexpr bool is_fint = false;
+// Rejects non-positive and non-finite doubles. Use with Next<PositiveDouble>().
+struct PositiveDouble {
+  using underlying_t = double;
+  underlying_t value = {};
+  operator underlying_t() const {
+    return value;
+  }
```
BorysTheDev: create base template struct with value and operator underlying_t to avoid code duplication. in thi case your class will be
`struct PositiveDouble : VNum<double>`

src/facade/cmd_arg_parser.h
```diff
@@ -72,21 +69,52 @@ namespace facade {
...
 //   // or: if (parser.HasError()) ...
 
+// Validated numeric type for Next<T>(): exposes an underlying_t, an implicit conversion to it, and
+// a static validate() predicate. FInt and the double validators below are instances. See
+// Convert<T>().
+template <class T>
+concept FNum = requires(T t, typename T::underlying_t v) {
+  static_cast<typename T::underlying_t>(t);
+  { T::validate(v) } -> std::same_as<bool>;
+};
```
BorysTheDev, line 83: maybe it is better to rename concept to as_vnum, to be more logical

FInt was - fixed int, now it is Validated number, so let's rename it to VNum or create a better name

src/facade/cmd_arg_parser.h
```diff
@@ -72,21 +69,52 @@ namespace facade {
...
-  static constexpr auto kMin = min;
-  static constexpr auto kMax = max;
+  static constexpr bool validate(underlying_t v) {
+    return v >= min && v <= max;
+  }
 };
 
-template <class T> constexpr bool is_fint = false;
+// Rejects non-positive and non-finite doubles. Use with Next<PositiveDouble>().
+struct PositiveDouble {
```
BorysTheDev: consider move your doubles to search_family file because I haven't seen double limitation in other files.

src/server/search/search_family.cc
```diff
@@ -1811,50 +1811,54 @@ ParseResult<HybridSearchParams> ParseHybridParams(CmdArgParser* parser) {
       p.scorer = *scorer_fn;
   };
   auto parse_tanh_factor = [&](CmdArgParser* sub) { tanh_factor = ParseBM25StdTanhFactor(sub); };
+  auto read_range_epsilon = [&](CmdArgParser* sub) {
+    double epsilon = sub->Next<facade::PositiveDouble>("Invalid EPSILON value");
+    if (sub->HasError())
+      return;
+    if (!search::SchemaField::VectorParams::IsValidRuntimeHnswEpsilon(epsilon))
```
BorysTheDev: you can validate it directly in class, you've created a concept for to avoid such code

src/server/search/search_family.cc
```diff
@@ -1796,6 +1811,55 @@ ParseResult<HybridSearchParams> ParseHybridParams(CmdArgParser* parser) {
...
+    p.use_range = true;
+
+    // Two accepted forms after RANGE: the counted form "<nargs> RADIUS <r> [EPSILON <e>]" (keywords
+    // in any order) and the shorthand "<radius> [EPSILON <e>]". The counted form starts with an
+    // integer count immediately followed by a RADIUS/EPSILON keyword; the shorthand starts with the
+    // bare radius value.
+    size_t nargs = 0;
+    bool keyword_form =
+        absl::SimpleAtoi(sub->Peek(0), &nargs) && (absl::EqualsIgnoreCase(sub->Peek(1), "RADIUS") ||
+                                                   absl::EqualsIgnoreCase(sub->Peek(1), "EPSILON"));
```
BorysTheDev, line 1842: interesting case, I think it should be possible improve Apply(OneOf(...)) approach to parse such logic. It you want you can try to extend Apply functionality; if not I will do it later.

src/server/search/search_family.cc
```diff
@@ -1796,6 +1811,55 @@ ParseResult<HybridSearchParams> ParseHybridParams(CmdArgParser* parser) {
...
+          p.range_radius = sub->Next<facade::NonNegativeDouble>("Invalid RADIUS value");
+          has_radius = !sub->HasError();
+        } else if (sub->Check("EPSILON")) {
+          if (p.range_epsilon) {
+            sub->ReportCustom("Duplicate EPSILON argument");
+            return;
+          }
+          read_range_epsilon(sub);
+        }
+      }
```
BorysTheDev, line 1862: I'm not sure regarding custom errors, but this logic can be implemented using Apply and OneOf.

src/facade/cmd_arg_parser.h
vyavdoshenko: fixed

src/facade/cmd_arg_parser.h
vyavdoshenko: fixed

src/facade/cmd_arg_parser.h
vyavdoshenko: fixed

src/server/search/search_family.cc

src/server/search/search_family.cc
vyavdoshenko: I leave it for you

src/server/search/search_family.cc
vyavdoshenko: I leave it for you as well

BorysTheDev (approved):

PR #7719, opened 2026-06-26
chore: add cf.insert and cf.insertnx

kostasrim:
* add cf.insert 
* add cf.insertnx
* add tests

#5446

### DIFF
--- a/src/server/cuckoo_filter_family.cc
+++ b/src/server/cuckoo_filter_family.cc
@@ -125,6 +125,47 @@ OpStatus OpCompact(const OpArgs& op_args, string_view key) {
   return OpStatus::OK;
 }
 
+struct InsertOptions {
+  uint64_t capacity = kDefaultCapacity;
+  bool nocreate = false;
+};
+
+// Shared op for CF.INSERT and CF.INSERTNX. Returns one integer per item:
+//   1  — item inserted
+//   0  — item already exists (nx only) or filter is full
+// Returns KEY_NOTFOUND if nocreate is set and the key does not exist.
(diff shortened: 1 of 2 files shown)


kostasrim:
augment review

kostasrim:
augment review

kostasrim:
augment review

src/server/cuckoo_filter_family.cc
```diff
@@ -307,6 +353,48 @@ void CmdDel(CmdArgParser parser, CommandContext* cmd_cntx) {
...
+  string_view key = parser.Next();
+  auto* rb = static_cast<RedisReplyBuilder*>(cmd_cntx->rb());
+  RETURN_ON_PARSE_ERROR(parser, rb);
+
+  InsertOptions opts;
+  parser.Apply(Tag("CAPACITY", &opts.capacity), Exist("NOCREATE", &opts.nocreate));
+  RETURN_ON_PARSE_ERROR(parser, rb);
+
+  if (!opts.nocreate && opts.capacity == 0)
+    return rb->SendError("CF: capacity must be greater than 0");
```
abhijat, line 366: not a nit, more of a question, should capacity be non zero in all cases, or just `nocreate`?

abhijat (approved):

src/server/cuckoo_filter_family.cc
kostasrim: when nocreate is true there is already a CF with capacity X. But when nocreate is set, we can't create a CF with 0 capacity. 

So yes capacity should be non zero for cases it affects the CF

PR #7722, opened 2026-06-28
feat(facade): add parse-ahead and parse-in-proactor for IoLoopV2

glevkovich:
feat(facade): grow V2 pipeline batches pre-squash

V1 runs two fibers per connection: one does read+parse, the other
does execute+reply. Because they run concurrently, fresh messages
keep arriving into the pipeline while commands execute, so by the
time a batch is dispatched it has naturally grown larger. This
overlap is a side effect of the two-fiber design, not something V1
asks for explicitly. The V2 consumer takes advantage of this overlap.

V2 is a single fiber and until recently ran a strictly serialized
read -> parse -> execute -> reply cycle: no reads between parses, and
no second attempt to parse more. It therefore could not grow a batch
the way V1 does.

The recent synchronous squash dispatch (SquashPipelineV2, where the
fiber blocks at bc->Wait for remote shards) opened a real suspension
window. These two features break V2's serialized cycle to use it.

1. pipeline_parse_in_proactor: while the connection fiber is blocked
   at the squash hop, the proactor OnRecv callback now reads new
   socket bytes AND parses them into queued commands. When the fiber
   resumes, the next batch is already built. This is the V2 analogue
   of V1's producer fiber running alongside execution.
2. pipeline_prioritize_large_batches: in ParseLoop, once a pipeline
   is forming (>1 queued) and more input is expected (pending_input_),
   return without executing so IoLoopV2's read path refills io_buf_
   and re-enters ParseLoop. Commands accumulate into one larger squash
   batch instead of many small ones.

Supporting changes:

- Two new feature flags: pipeline_prioritize_large_batches and
  pipeline_parse_in_proactor, both enabled by default.
- FiberParkSpot enum tracks where the V2 fiber is parked at any
  moment (kNone/kIdleAwait/kSquashHop/kParseYield). Used as a safety
  gate so parse-in-proactor only fires when the parser is idle.
- recv_multishot_active_ flag added and used to gate the new
  ParseLoop deferral path.
- ParseRedis gains an explicit allow_yield argument (false from the
  proactor callback, which must never suspend); enqueue_only is now
  a required argument at all call sites.
- OnRecvNotification drains the socket and runs parse-in-proactor
  only while under the pipeline limit, and counts proactor_reads.
- When pipeline_parse_in_proactor is on, ExecuteBatch skips the
  post-squash break (which would detour through the io-loop to grow
  the batch from raw bytes) and re-squashes in place instead. The
  proactor already parsed those bytes during bc->Wait, so the detour
  is redundant.
- proactor_parse_error_ flag lets the proactor callback (which can't
  return a status) signal a protocol error from its parse; IoLoopV2
  observes it and surfaces ParserStatus::ERROR to close the
  connection and send the protocol-error reply.
- squash_batch_size_hist records commands-per-squash distribution;
  exposed as pipeline_squash_batch_size Prometheus summary for both
  V1 and V2 squash paths.
- proactor_reads counter tracks how often OnRecvNotification drained
  new bytes; exposed as proactor_reads_total Prometheus counter.
- Both stats are cleared in ResetStats() and merged in operator+=.
- CONN_ID macro replaces the inline "[" << id_ << "] " logging
  prefix across the connection code.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -3015,15 +3015,12 @@ void Connection::UpdateFromFlags() {
   always_flush_pipeline_cached = GetFlag(FLAGS_always_flush_pipeline);
   pipeline_squash_limit_cached = GetFlag(FLAGS_pipeline_squash_limit);
   pipeline_wait_batch_usec = GetFlag(FLAGS_pipeline_wait_batch_usec);
-  pipeline_prioritize_large_batches_cached = GetFlag(FLAGS_pipeline_prioritize_large_batches);
-  pipeline_parse_in_proactor_cached = GetFlag(FLAGS_pipeline_parse_in_proactor);
 }
 
 std::vector<std::string> Connection::GetMutableFlagNames() {
-  return base::GetFlagNames(
-      FLAGS_pipeline_queue_limit, FLAGS_pipeline_buffer_limit, FLAGS_max_busy_read_usec,
-      FLAGS_always_flush_pipeline, FLAGS_pipeline_squash_limit, FLAGS_pipeline_wait_batch_usec,
-      FLAGS_pipeline_prioritize_large_batches, FLAGS_pipeline_parse_in_proactor);
+  return base::GetFlagNames(FLAGS_pipeline_queue_limit, FLAGS_pipeline_buffer_limit,
+                            FLAGS_max_busy_read_usec, FLAGS_always_flush_pipeline,
+                            FLAGS_pipeline_squash_limit, FLAGS_pipeline_wait_batch_usec);
 }
 
 void Connection::GetRequestSizeHistogramThreadLocal(std::string* hist) {
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -126,6 +126,16 @@ ABSL_FLAG(uint32_t, pipeline_wait_batch_usec, 0,
           "If non-zero, waits for this time for more I/O "
           " events to come for the connection in case there is only one command in the pipeline. ");
 
+ABSL_FLAG(
+    bool, pipeline_prioritize_large_batches, true,
+    "V2 only: in ParseLoop, defer executing a parsed pipeline batch and return to the read loop "
+    "to accumulate more already-available input, so the squasher sees one large batch instead of "
+    "many small ones.");
+
+ABSL_FLAG(bool, pipeline_parse_in_proactor, true,
+          "V2 only: parse newly-arrived bytes from the proactor OnRecv callback while the fiber "
+          "is blocked in a squash hop, so the next batch is already grown when execution resumes.");
+
 ABSL_FLAG(bool, enable_memcache_io_loop_v2, true,
           "Enable the event-driven IoLoopV2 for non-TLS Memcache connections.");
 ABSL_FLAG(bool, enable_resp_io_loop_v2, false,
@@ -140,6 +150,8 @@ using namespace std;
 using absl::GetFlag;
 using base::CycleClock;
 using nonstd::make_unexpected;
+
+#define CONN_ID "[" << id_ << "] "
 
 namespace facade {
 
@@ -537,6 +549,10 @@ ConnectionStats& __attribute__((noinline)) GetLocalConnStats() {
 thread_local uint32_t max_busy_read_cycles_cached = UINT32_MAX;
 thread_local bool always_flush_pipeline_cached = absl::GetFlag(FLAGS_always_flush_pipeline);
 thread_local uint32_t pipeline_squash_limit_cached = absl::GetFlag(FLAGS_pipeline_squash_limit);
+thread_local bool pipeline_prioritize_large_batches_cached =
+    absl::GetFlag(FLAGS_pipeline_prioritize_large_batches);
+thread_local bool pipeline_parse_in_proactor_cached =
+    absl::GetFlag(FLAGS_pipeline_parse_in_proactor);
 
 }  // namespace
 
@@ -816,7 +832,7 @@ void Connection::OnPreMigrateThread() {
 }
 
 void Connection::OnPostMigrateThread() {
-  DVLOG(1) << "[" << id_ << "] OnPostMigrateThread";
+  DVLOG(1) << CONN_ID << "OnPostMigrateThread";
 
   // Once we migrated, we should rearm OnBreakCb callback.
   if (socket()->IsOpen()) {
@@ -865,7 +881,7 @@ void Connection::OnConnectionStart() {
 }
 
 void Connection::HandleRequests() {
-  VLOG(1) << "[" << id_ << "] HandleRequests";
+  VLOG(1) << CONN_ID << "HandleRequests";
   DCHECK(tl_facade_stats);
   auto& conn_stats = tl_facade_stats->conn_stats;
 
@@ -1260,7 +1276,7 @@ void Connection::ConnectionFlow() {
   if (io_buf_.InputLen() > 0) {
     phase_ = PROCESS;
     if (redis_parser_ && !ioloop_v2_) {
-      parse_status = ParseRedis(io_buf_, 10000);
+      parse_status = ParseRedis(io_buf_, 10000, /*enqueue_only=*/false);
     } else {
       parse_status = ParseLoop();
     }
@@ -1423,8 +1439,6 @@ void Connection::DispatchSingle(bool has_more, absl::FunctionRef<void()> invoke_
 
 Connection::ParserStatus Connection::ParseRedis(base::IoBuf& io_buf, uint32_t max_busy_cycles,
                                                 bool enqueue_only) {
-  DCHECK_GT(max_busy_cycles, 0u);
-
   uint32_t consumed = 0;
   RespSrvParser::Result result = RespSrvParser::OK;
 
@@ -1494,9 +1508,21 @@ Connection::ParserStatus Connection::ParseRedis(base::IoBuf& io_buf, uint32_t ma
     // We must yield from time to time to allow other fibers to run.
     // Specifically, if a client sends a huge chunk of data resulting in a very long pipeline,
     // we want to yield to allow AsyncFiber to actually execute on the pending pipeline.
-    if (ThisFiber::GetRunningTimeCycles() > max_busy_cycles) {
+    //
+    // If max_busy_cycles == 0, never yield. We rely on io_buf_ to bound the work.
+    if ((max_busy_cycles > 0) && ThisFiber::GetRunningTimeCycles() > max_busy_cycles) {
       GetLocalConnStats().num_read_yields++;
+
+      fiber_park_spot_ = FiberParkSpot::kParseYield;
       ThisFiber::Yield();
+      fiber_park_spot_ = FiberParkSpot::kNone;
+
+      // Note:
+      // - read_buffer stays valid across this yield since proactor only calls ReadPendingInput ->
+      // AppendBuffer + CommitWrite. which writes the free tail and bumps size_ without touching
+      // buf_/offs_.
+      // - If provided buffers are ever enabled, we should revisit this code again when
+      // parse-in-proactor enabled.
     }
   } while (RespSrvParser::OK == result && !read_buffer.empty() && !reply_builder_->GetError());
 
@@ -1527,6 +1553,16 @@ auto Connection::ParseLoop() -> ParserStatus {
     DCHECK_GT(io_buf_.InputLen(), 0u);
     parse_status = (this->*parse_func)(io_buf_);
 
+    // V2 large-batch prioritization (pipeline_prioritize_large_batches):
+    // - When a real pipeline is forming (>1 queued) and more socket data is expected
+    // (pending_input_), return without executing so IoLoopV2's read path fill the buffer and
+    // re-enters ParseLoop.
+    // - This helps accumulating commands into the same queue for one larger squash batch instead of
+    // many small ones.
+    if (ioloop_v2_ && pipeline_prioritize_large_batches_cached && (parse_status != ERROR) &&
+        pending_input_ && (parsed_cmd_q_len_ > 1) && !IsOverPipelineLimit())
+      return parse_status;
+
     // Execute/reply the commands parsed so far first, so a trailing protocol error still flushes
     // earlier replies in order before we report it.
     if (!ExecuteBatch())
@@ -1553,9 +1589,9 @@ void Connection::OnBreakCb(int32_t mask) {
     return;
   }
 
-  DCHECK(reply_builder_) << "[" << id_ << "] " << unsigned(phase_) << " " << migration_in_process_;
+  DCHECK(reply_builder_) << CONN_ID << unsigned(phase_) << " " << migration_in_process_;
 
-  VLOG(1) << "[" << id_ << "] Got event " << mask << " " << unsigned(phase_) << " "
+  VLOG(1) << CONN_ID << "Got event " << mask << " " << unsigned(phase_) << " "
           << reply_builder_->IsSendActive() << " " << reply_builder_->GetError();
 
   cc_->conn_closing = true;
@@ -1622,7 +1658,7 @@ bool Connection::ProcessControlMessages(uint32_t quota) {
   while (!dispatch_q_.empty()) {
     // If Quota reached: stop draining the dispatch queue and fall through to the data path.
     if ((quota > 0) && (dispatched >= quota)) {
-      LOG_EVERY_T(INFO, 1) << "[" << id_ << "] V2 dispatch_q_ quota reached (" << dispatched << "/"
+      LOG_EVERY_T(INFO, 1) << CONN_ID << "V2 dispatch_q_ quota reached (" << dispatched << "/"
                            << quota << "), falling through to data path";
       return true;
     }
@@ -1692,7 +1728,7 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoop() {
     bool reached_capacity = io_buf_.AppendLen() == 0;
 
     if (redis_parser_) {
-      parse_status = ParseRedis(io_buf_, max_busy_read_cycles_cached);
+      parse_status = ParseRedis(io_buf_, max_busy_read_cycles_cached, /*enqueue_only=*/false);
     } else {
       DCHECK(memcache_parser_);
       parse_status = ParseLoop();
@@ -2091,7 +2127,7 @@ void Connection::AsyncFiber() {
       }
       if (dispatch_q_.empty() || prefer_pipeline_execution) {  // 2. Process pipeline Queue
         VLOG_IF(1, prefer_pipeline_execution)
-            << "[" << id_ << "] Preferring pipeline execution over admin queue. "
+            << CONN_ID << "Preferring pipeline execution over admin queue. "
             << "Migration requested: " << is_migration_req
             << ", dispatch quota reached: " << quota_reached
             << ", async_dispatch_quota: " << async_dispatch_quota
@@ -2244,7 +2280,7 @@ void Connection::SendInvalidationMessageAsync(InvalidationMessage msg) {
 void Connection::LaunchAsyncFiberIfNeeded() {
   DCHECK(!ioloop_v2_);
   if (!async_fb_.IsJoinable() && !migration_in_process_) {
-    VLOG(1) << "[" << id_ << "] LaunchAsyncFiberIfNeeded ";
+    VLOG(1) << CONN_ID << "LaunchAsyncFiberIfNeeded ";
     async_fb_ = fb2::Fiber(fb2::Launch::post, "connection_dispatch", [this]() { AsyncFiber(); });
   }
 }
@@ -2596,7 +2632,7 @@ Connection::ParserStatus Connection::ParseRedisBatch(base::IoBuf& buf) {
   }
   // Forward ParseRedis's status verbatim (OK / NEED_MORE / ERROR) so a protocol error
   // propagates to ParseLoop instead of being flattened into "no commands parsed".
-  return ParseRedis(buf, max_busy_read_cycles_cached, true);
+  return ParseRedis(buf, max_busy_read_cycles_cached, /*enqueue_only=*/true);
 }
 
 Connection::ParserStatus Connection::ParseMCBatch(base::IoBuf& io_buf) {
@@ -2669,8 +2705,10 @@ bool Connection::SquashPipelineV2() {
   auto& conn_stats = tl_facade_stats->conn_stats;
 
   uint64_t dispatch_start = CycleClock::Now();
+  fiber_park_spot_ = FiberParkSpot::kSquashHop;
   unsigned squashed =
       service_->DispatchSquashedBatch(parsed_to_execute_, dispatch_waiting_count_, cc_.get());
+  fiber_park_spot_ = FiberParkSpot::kNone;
 
   if (squashed == 0)
     return false;
@@ -2737,10 +2775,12 @@ bool Connection::ExecuteBatch() {
                << " " << io_buf_.InputLen();
 
       if (SquashPipelineV2()) {
-        // This helps with throughput. Explanation:
-        // when we suspend the thread calls io-callbacks that fill up the input buffer.
-        // By breaking now we give the io-loop a chance to add more commands to the pipeline.
-        if (pending_input_ || io_buf_.InputLen() > 0)
+        // - This helps with throughput. Explanation:
+        //   when we suspend the thread calls io-callbacks that fill up the input buffer.
+        //   By breaking now we give the io-loop a chance to add more commands to the pipeline.
+        // - Skip the break when parse-in-proactor is on: the proactor already parsed those bytes
+        //   into the queue during the squash wait, so keep squashing in place instead.
+        if (!pipeline_parse_in_proactor_cached && (pending_input_ || io_buf_.InputLen() > 0))
           break;
         continue;
       }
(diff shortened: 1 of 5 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -3029,10 +3087,32 @@ bool ConnectionRef::operator==(const ConnectionRef& other) const {
...
+    ++GetLocalConnStats().proactor_reads;
+
+  // Parse In Proactor : parse newly-read bytes now, while the fiber is parked in a squash hop, so
+  // the next batch is already larger on resume.
+  // This is safe since parser state is idle. allow_yield=false because callbacks must not suspend
+  // the proactor.
+  if (pipeline_parse_in_proactor_cached && (fiber_park_spot_ == FiberParkSpot::kSquashHop) &&
+      redis_parser_ && (io_buf_.InputLen() > 0) && !IsOverPipelineLimit()) {
+    size_t cmds_before = parsed_cmd_q_len_;
+    ParseRedis(io_buf_, max_busy_read_cycles_cached, /*enqueue_only=*/true, /*allow_yield=*/false);
```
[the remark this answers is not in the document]
glevkovich: will fix.

src/facade/dragonfly_connection.cc
```diff
@@ -3029,10 +3087,32 @@ bool ConnectionRef::operator==(const ConnectionRef& other) const {
...
 
 void Connection::OnRecvNotification(const util::FiberSocketBase::RecvNotification& n) {
-  DVLOG(2) << "OnRecvNotification iobuf_len: " << io_buf_.InputLen();
+  DVLOG(2) << "OnRecvNotification: [c" << id_ << "] io_buf_ input_len=" << io_buf_.InputLen()
+           << " pending_input=" << pending_input_;
   ProcessRecvNotification(n);
-  // Eagerly drain the socket while the fiber is suspended to prevent receive buffer starvation.
+
+  // Compact if full so ReadPendingInput has room to append.
+  if (io_buf_.AppendLen() == 0)
```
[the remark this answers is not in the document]
glevkovich: I'm going to fix the comment.

src/server/metrics.cc
```diff
@@ -163,6 +163,24 @@ void Metrics::Print(uint64_t uptime, DflyCmd* dfly_cmd, util::http::StringRespon
...
+    if (conn_stats.squash_batch_size_hist.count() > 0) {
+      auto pctls = conn_stats.squash_batch_size_hist.Percentiles(50, 95, 99);
+      AppendMetricValue("pipeline_squash_batch_size", pctls[0], {"quantile"}, {"0.5"},
+                        &resp->body());
+      AppendMetricValue("pipeline_squash_batch_size", pctls[1], {"quantile"}, {"0.95"},
+                        &resp->body());
+      AppendMetricValue("pipeline_squash_batch_size", pctls[2], {"quantile"}, {"0.99"},
+                        &resp->body());
+    }
+    absl::StrAppend(&resp->body(), sbs_full, "_count ", conn_stats.squash_batch_size_hist.count(),
```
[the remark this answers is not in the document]
glevkovich: will fix.

src/facade/dragonfly_connection.cc
```diff
@@ -3029,10 +3087,32 @@ bool ConnectionRef::operator==(const ConnectionRef& other) const {
...
+
+  // Parse In Proactor : parse newly-read bytes now, while the fiber is parked in a squash hop, so
+  // the next batch is already larger on resume.
+  // This is safe since parser state is idle. allow_yield=false because callbacks must not suspend
+  // the proactor.
+  if (pipeline_parse_in_proactor_cached && (fiber_park_spot_ == FiberParkSpot::kSquashHop) &&
+      redis_parser_ && (io_buf_.InputLen() > 0) && !IsOverPipelineLimit()) {
+    size_t cmds_before = parsed_cmd_q_len_;
+    ParseRedis(io_buf_, max_busy_read_cycles_cached, /*enqueue_only=*/true, /*allow_yield=*/false);
+    DVLOG(1) << "[c" << id_ << "] Parse-in-proactor added " << (parsed_cmd_q_len_ - cmds_before)
```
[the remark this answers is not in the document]
glevkovich: duplicated. fixed.

src/server/metrics.cc
```diff
@@ -163,6 +163,24 @@ void Metrics::Print(uint64_t uptime, DflyCmd* dfly_cmd, util::http::StringRespon
...
+                    "\n");
+  }
```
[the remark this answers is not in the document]
glevkovich: duplicated. fixed.

src/facade/dragonfly_connection.cc
```diff
@@ -1494,9 +1506,20 @@ Connection::ParserStatus Connection::ParseRedis(base::IoBuf& io_buf, uint32_t ma
...
       ThisFiber::Yield();
+      fiber_park_spot_ = FiberParkSpot::kNone;
+
+      // A proactor recv callback may have run during the yield and moved io_buf_'s data
+      // (Compact: memmove to front) or grown it (Reserve: realloc). Either way, the pointer
+      // inside read_buffer may be stale. Rebuild it from scratch.
+      read_buffer = io_buf.InputBuffer();
+      read_buffer.remove_prefix(total_consumed);
+      read_buffer = read_buffer.first(remaining);
     }
```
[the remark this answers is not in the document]
glevkovich: fixed, added bound by number of commands.

src/facade/dragonfly_connection.cc
```diff
@@ -126,6 +126,15 @@ ABSL_FLAG(uint32_t, pipeline_wait_batch_usec, 0,
...
           " events to come for the connection in case there is only one command in the pipeline. ");
 
+ABSL_FLAG(
+    bool, pipeline_parse_ahead, false,
+    "V2 only: before executing a parsed pipeline batch, opportunistically read and parse "
+    "more already-available input so the squash sees one large batch instead of many small ones.");
+
+ABSL_FLAG(bool, pipeline_parse_in_proactor, false,
+          "V2 only: parse newly-arrived bytes from the proactor OnRecv callback while the fiber "
+          "is blocked in a squash hop, so the next batch is already grown when execution resumes.");
```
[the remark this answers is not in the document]
glevkovich: both flags are enabled by default. dropping.

src/facade/dragonfly_connection.cc
```diff
@@ -3029,10 +3107,43 @@ bool ConnectionRef::operator==(const ConnectionRef& other) const {
...
+
+  // Compact the buffer when it is full so that ReadPendingInput() has room to append new bytes.
+  // Compact() moves the unconsumed data to the front of the buffer (memmove). This is safe even
+  // when a parser is suspended and holds a pointer into io_buf_:
+  //   - FiberParkSpot::kSquashHop: the squash batch was already fixed before the hop, so the
+  //     dispatch is not reading io_buf_ at this point.
+  //   - FiberParkSpot::kParseYield: ParseRedis refreshes its local buffer pointer from
+  //     io_buf_.InputBuffer() right after waking up, so the move happens before it reads again.
+  if (io_buf_.AppendLen() == 0)
+    io_buf_.Compact();
```
romange: As I said before, I am not a fan of it as it's a symptom of poor state machine. 
This may create redundant memory copies. 

We should either avoid reading if io_buf_ append space is malls  and create backpressure, or alternatively, we can unload `io_buf_` to parse queue before reading into it.

src/facade/dragonfly_connection.h
```diff
@@ -406,7 +406,15 @@ class Connection : public util::Connection {
...
 
-  ParserStatus ParseRedis(base::IoBuf& buf, uint32_t max_busy_cycles, bool enqueue_only = false);
+  // Parse RESP commands from buf until the buffer is exhausted, a protocol error is hit, or
+  // max_busy_cycles CPU time elapses (triggering a yield to avoid starving other fibers).
+  //
+  // enqueue_only=true (V2): all parsed commands are enqueued without inline dispatch.
+  // allow_yield=false: disables the periodic yield, MUST be false when called from a proactor
+  // callback (OnRecvNotification) which must never suspend the connection fiber.
+  // max_parse: caps how many commands one call parses (0 = unlimited).
+  ParserStatus ParseRedis(base::IoBuf& buf, uint32_t max_busy_cycles, bool enqueue_only = false,
```
romange: We do not use default arguments - they are not very readable.

src/facade/dragonfly_connection.cc
```diff
@@ -1517,16 +1554,43 @@ Connection::ParserStatus Connection::ParseRedis(base::IoBuf& io_buf, uint32_t ma
...
 
-  do {
+  // Why while(true)+continue: a parse-ahead hit must (unconditionally) re-enter parse without
+  // executing.
+  while (true) {
     DCHECK_GT(io_buf_.InputLen(), 0u);
     parse_status = (this->*parse_func)(io_buf_);
 
+    // V2 parse-ahead: top-up the buffer and re-parse if new bytes arrived.
+    if (ioloop_v2_ && pipeline_parse_ahead_cached && (parse_status != ERROR) && MaybeParseAhead())
```
romange: I understand the idea at high - level but this complicates the flow, imho.
Why not exit here before calling ExecuteBatch and continue to the head of IoLoopV2?

This way you will rely on the usual flow to read and parse later

src/facade/dragonfly_connection.cc
```diff
@@ -3029,10 +3107,43 @@ bool ConnectionRef::operator==(const ConnectionRef& other) const {
...
+    ++GetLocalConnStats().proactor_reads;
+
+  // Parse In Proactor : parse newly-read bytes now, while the fiber is parked in a squash hop, so
+  // the next batch is already larger on resume.
+  // This is safe since parser state is idle. allow_yield=false because callbacks must not suspend
+  // the proactor.
+  if (pipeline_parse_in_proactor_cached && (fiber_park_spot_ == FiberParkSpot::kSquashHop) &&
+      redis_parser_ && (io_buf_.InputLen() > 0) && !IsOverPipelineLimit()) {
+    size_t cmds_before = parsed_cmd_q_len_;
+    ParserStatus st = ParseRedis(io_buf_, max_busy_read_cycles_cached, /*enqueue_only=*/true,
```
romange: I think this can be simplified if you allow soft constraints.

1. io_buf_ is limited so whatever is being parsed here - is limited by definition
2. IsOverPipelineLimit can be checked earlier - before the read. There is no point reading from socket if we want o apply back-pressure
3. `ParseRedis` - do we really need both `max_busy_read_cycles_cached` and `max_parse` if `io_buf_` is limited? It seems as over-engineering.  We can just limit the input buffer in the worst case but do you have a specific use-case where those limits improved?

src/facade/dragonfly_connection.cc
```diff
@@ -3029,10 +3107,43 @@ bool ConnectionRef::operator==(const ConnectionRef& other) const {
...
+                                 /*allow_yield=*/false, /*max_parse=*/kParseMaxBatch);
+    // The recv callback cannot return a status. If parsing hit a protocol error, flag it so
+    // IoLoopV2 surfaces ParserStatus::ERROR and sends the protocol-error reply.
+    if (st == ERROR)
+      proactor_parse_error_ = true;
+    DVLOG(1) << "[c" << id_ << "] Parse-in-proactor added " << (parsed_cmd_q_len_ - cmds_before)
```
romange: nit - we use `"[" << id_ << "]` everywhere. if you change it - make it a macro and change everywhere.

src/facade/dragonfly_connection.cc
```diff
@@ -1517,16 +1554,43 @@ Connection::ParserStatus Connection::ParseRedis(base::IoBuf& io_buf, uint32_t ma
   return ERROR;
 }
 
+bool Connection::MaybeParseAhead() {
```
romange: the function is called 
MaybeParseAhead but inside it calls `ReadPendingInput` and has multiple conditions copied from other places. I already gave other comments, will stop reviewing now but try to think if whatever you need from this function can be achieved by redirecting the main loop to do what you need.

src/facade/dragonfly_connection.cc
glevkovich: I'll make it a macro then.

src/facade/dragonfly_connection.h
glevkovich: I'll fix this (remove all defaults).

src/facade/dragonfly_connection.cc
glevkovich: - max_busy_read_cycles_cached is irrelevant for parse-in-proactor since it marks allow_yield=false. Any non-zero value I send will be ignored.
- I will remove max_parse. Rethinking, it was wrongly added last minute with a bot PRs comment. I don't have any specific case where I saw them improve.
- IsOverPipelineLimit should be a condition for the read and for the parse together. I will move them under that same condition.

ps -  you said "There is no point reading from socket if we want o apply back-pressure" ->  we are also reading in the main IO loop unconditionally, without checking back-pressure. Perhaps (in another PR) this should be under back-pressure condition as well?

```
  do {
    HandleMigrateRequest();

    // Register completion for current head if its pending and we don't wait on current_wait_.
    if (HasInFlightCommands() && !current_wait_.has_value()) {
      current_wait_.emplace(parsed_head_, &cmd_completion_waiter);
    }

    ReadPendingInput();
    ...
```

src/facade/dragonfly_connection.cc
glevkovich: I chose option 1 (avoid the read and create back-pressure).
Dropped the Compact().

src/facade/dragonfly_connection.cc
glevkovich: I had to do some benchmark to check if I get similar results, since there are some differences. I decided to drop MaybeParseAhead() and go back to the main loop to read more under specific conditions.

src/facade/dragonfly_connection.cc
glevkovich: Dropping this comment, since code is obsolete, i removed MaybeParseAhead.
See  https://github.com/dragonflydb/dragonfly/pull/7722#discussion_r3487822666

romange:
Please resolve

glevkovich:
> Please resolve
@romange 
Resolved all, else than this comment :+1: https://github.com/dragonflydb/dragonfly/pull/7722#discussion_r3488001111
regarding the  ReadPendingInput in the main loop.

src/facade/dragonfly_connection.cc
```diff
@@ -126,6 +126,16 @@ ABSL_FLAG(uint32_t, pipeline_wait_batch_usec, 0,
...
 
+ABSL_FLAG(
+    bool, pipeline_prioritize_large_batches, true,
+    "V2 only: in ParseLoop, defer executing a parsed pipeline batch and return to the read loop "
+    "to accumulate more already-available input, so the squasher sees one large batch instead of "
+    "many small ones.");
+
+ABSL_FLAG(bool, pipeline_parse_in_proactor, true,
+          "V2 only: parse newly-arrived bytes from the proactor OnRecv callback while the fiber "
+          "is blocked in a squash hop, so the next batch is already grown when execution resumes.");
```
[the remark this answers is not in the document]
glevkovich: Description was changed already. dropping.

src/facade/dragonfly_connection.cc
```diff
@@ -1494,9 +1510,22 @@ Connection::ParserStatus Connection::ParseRedis(base::IoBuf& io_buf, uint32_t ma
     // We must yield from time to time to allow other fibers to run.
     // Specifically, if a client sends a huge chunk of data resulting in a very long pipeline,
     // we want to yield to allow AsyncFiber to actually execute on the pending pipeline.
-    if (ThisFiber::GetRunningTimeCycles() > max_busy_cycles) {
+    //
+    // If allow_yield == false, current context can't yield and won't respect max_busy_cycles.
+    // That is fine - parse-in-proactor is the only such caller, and io_buf_'s size bounds its work.
+    if (allow_yield && ThisFiber::GetRunningTimeCycles() > max_busy_cycles) {
```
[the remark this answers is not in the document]
glevkovich: We already decided that the buffer itself is the budget, we removed any upper bound on parsing for now. Dropping.

src/facade/dragonfly_connection.cc
```diff
@@ -1422,7 +1438,7 @@ void Connection::DispatchSingle(bool has_more, absl::FunctionRef<void()> invoke_
 }
 
 Connection::ParserStatus Connection::ParseRedis(base::IoBuf& io_buf, uint32_t max_busy_cycles,
-                                                bool enqueue_only) {
+                                                bool enqueue_only, bool allow_yield) {
```
romange: Instead of adding an additional argument, have a special value to max_busy_cycles, either 0 or uint32_max

src/facade/dragonfly_connection.cc
```diff
@@ -1494,9 +1510,22 @@ Connection::ParserStatus Connection::ParseRedis(base::IoBuf& io_buf, uint32_t ma
...
+    // That is fine - parse-in-proactor is the only such caller, and io_buf_'s size bounds its work.
+    if (allow_yield && ThisFiber::GetRunningTimeCycles() > max_busy_cycles) {
       GetLocalConnStats().num_read_yields++;
+
+      fiber_park_spot_ = FiberParkSpot::kParseYield;
       ThisFiber::Yield();
+      fiber_park_spot_ = FiberParkSpot::kNone;
+
+      // Note:
+      // 1) read_buffer stays valid across this yield: here the recv callback only appends to
```
romange: Why it is important that there are no reallocs, compactions?

src/facade/dragonfly_connection.cc
```diff
@@ -1539,7 +1579,11 @@ auto Connection::ParseLoop() -> ParserStatus {
     // protocol error reply (using parser_error_) and close the connection.
     if (parse_status == ERROR)
       return ERROR;
-  } while (parse_status == OK && io_buf_.InputLen() > 0);
+
+    // Loop again only if more input remains after a clean parse, otherwise return OK/NEED_MORE.
```
romange: Why this code is better than while?

src/facade/dragonfly_connection.cc
```diff
@@ -1523,10 +1552,21 @@ auto Connection::ParseLoop() -> ParserStatus {
...
     parse_status = (this->*parse_func)(io_buf_);
 
+    // V2 large-batch prioritization (pipeline_prioritize_large_batches):
+    // - When a real pipeline is forming (>1 queued) and more socket data is expected
+    // (pending_input_), return without executing so IoLoopV2's read path fill the buffer and
+    // re-enters ParseLoop.
+    // - This helps accumulating commands into the same queue for one larger squash batch instead of
+    // many small ones.
+    if (ioloop_v2_ && pipeline_prioritize_large_batches_cached && (parse_status != ERROR) &&
+        pending_input_ && (parsed_cmd_q_len_ > 1) && !recv_multishot_active_ &&
```
romange: Can pending_input and recv multishot be true at the same time?

src/facade/dragonfly_connection.cc
```diff
@@ -1834,6 +1879,8 @@ void Connection::SquashPipeline() {
   conn_stats.pipeline_dispatch_calls++;
   conn_stats.pipeline_dispatch_commands += squashed;
 
+  conn_stats.squash_batch_size_hist.Add(squashed);
```
romange: I do not want to add a histogram. can you show where histogram gave you additional value over averages?

romange (changes requested):

src/facade/dragonfly_connection.cc
glevkovich: I'm going to revert it to do/while (leftover from previous logic).

src/facade/dragonfly_connection.cc
glevkovich: removing recv_multishot_active_

src/facade/dragonfly_connection.cc
glevkovich: It did while developing,  but this can also be inferred from simple logs during development.
So I'll drop it.

src/facade/dragonfly_connection.cc
glevkovich: This is clarification since now another party (proactor) is accessing the io_buff_ during parking of connection fiber and with future code changes it may help to better understand and avoid pitfalls. I'm going to change the comment (which is valid) to make it clear:

- During this yield, the only thing the recv callback can do to io_buf_ is ReadPendingInput() , which calls AppendBuffer() + CommitWrite(). These 2 functions don't touch  buf_ nor offs_, so the existing read_buffer span stays valid.

- Proactor may also call ParseRedis() when connection fiber parks, but not here, since it won't be beneficial, so it does it only during execution, but it is completely safe there.

src/facade/dragonfly_connection.cc
romange: Please make sure to add as little code as possible and remove whatever is not needed when sending the PR

src/facade/dragonfly_connection.cc
```diff
@@ -1527,6 +1553,16 @@ auto Connection::ParseLoop() -> ParserStatus {
     DCHECK_GT(io_buf_.InputLen(), 0u);
     parse_status = (this->*parse_func)(io_buf_);
 
+    // V2 large-batch prioritization (pipeline_prioritize_large_batches):
+    // - When a real pipeline is forming (>1 queued) and more socket data is expected
+    // (pending_input_), return without executing so IoLoopV2's read path fill the buffer and
+    // re-enters ParseLoop.
+    // - This helps accumulating commands into the same queue for one larger squash batch instead of
+    // many small ones.
+    if (ioloop_v2_ && pipeline_prioritize_large_batches_cached && (parse_status != ERROR) &&
```
romange, line 1562: you can also get rid of  ioloop_v2_ by moving it into `pipeline_prioritize_large_batches_cached` definition. 

In any case, I would like to learn as well the use-case that demands this logic - does it reproduce with a single connection?

src/facade/dragonfly_connection.cc
```diff
@@ -2975,12 +3015,15 @@ void Connection::UpdateFromFlags() {
...
 }
 
 std::vector<std::string> Connection::GetMutableFlagNames() {
-  return base::GetFlagNames(FLAGS_pipeline_queue_limit, FLAGS_pipeline_buffer_limit,
-                            FLAGS_max_busy_read_usec, FLAGS_always_flush_pipeline,
-                            FLAGS_pipeline_squash_limit, FLAGS_pipeline_wait_batch_usec);
+  return base::GetFlagNames(
+      FLAGS_pipeline_queue_limit, FLAGS_pipeline_buffer_limit, FLAGS_max_busy_read_usec,
+      FLAGS_always_flush_pipeline, FLAGS_pipeline_squash_limit, FLAGS_pipeline_wait_batch_usec,
+      FLAGS_pipeline_prioritize_large_batches, FLAGS_pipeline_parse_in_proactor);
```
romange: Are you changing these settings via "CONFIG SET"  when load-testing? 

`GetMutableFlagNames` is intended for production, it is not required adding each flag automatically here.

src/facade/dragonfly_connection.cc
glevkovich: `you can also get rid of ioloop_v2_ by moving it into pipeline_prioritize_large_batches_cached definition.`

At this stage I think it's better not to merge flags and keep them specific so we can still benchmark with/without some of the features, so I prefer not to tie them together. These flag should be probably dropped at some point soon (enable always).

---
`In any case, I would like to learn as well the use-case that demands this logic - does it reproduce with a single connection?`

If I mark:
```
A=pipeline_prioritize_large_batches_cached
B=parse-in-proactor
```
At some point i did lots of benchmarks of v2 base vs v2-A vs v2-B bs v2-A+B.
V2-A+B
So my only conclusion both should be enabled was based on these bencharks. I can say that this feture is less important than proactor parsing, and the change in execution also contributed to performance:

```
 if (SquashPipelineV2()) {
        // - This helps with throughput. Explanation:
        //   when we suspend the thread calls io-callbacks that fill up the input buffer.
        //   By breaking now we give the io-loop a chance to add more commands to the pipeline.
        // - Skip the break when parse-in-proactor is on: the proactor already parsed those bytes
        //   into the queue during the squash wait, so keep squashing in place instead.
        if (!pipeline_parse_in_proactor_cached && (pending_input_ || io_buf_.InputLen() > 0))
          break;
        continue;
      }
```

src/facade/dragonfly_connection.cc
glevkovich: I probably misunderstood the intended use case.
I will remove this.

src/facade/dragonfly_connection.cc
glevkovich: Regarding your question about how to check it's impact:

Run twice with:

- X=false Y=true -> feature enabled
- X=false Y=false -> V2 base (both featured disabled)

`taskset -c 1,2,3 ./dragonfly --proactor_threads=3 --enable_resp_io_loop_v2=true --bind=0.0.0.0 --port=6379 --admin_port=8099 --dbfilename="" --maxmemory 22173mb --pipeline_parse_in_proactor=X --pipeline_prioritize_large_batches=Y`
and
`memtier_benchmark -s 172.31.30.209 -p 6379 -t 1 -c 1 --pipeline=100 --ratio=1:0 -d 4096 --key-pattern=R:R --key-maximum=1000000 --key-prefix=bench --test-time=15 --hide-histogram`

romange: ok, I will check it out.

src/facade/dragonfly_connection.cc
```diff
@@ -2737,10 +2775,12 @@ bool Connection::ExecuteBatch() {
...
-        // This helps with throughput. Explanation:
-        // when we suspend the thread calls io-callbacks that fill up the input buffer.
-        // By breaking now we give the io-loop a chance to add more commands to the pipeline.
-        if (pending_input_ || io_buf_.InputLen() > 0)
+        // - This helps with throughput. Explanation:
+        //   when we suspend the thread calls io-callbacks that fill up the input buffer.
+        //   By breaking now we give the io-loop a chance to add more commands to the pipeline.
+        // - Skip the break when parse-in-proactor is on: the proactor already parsed those bytes
+        //   into the queue during the squash wait, so keep squashing in place instead.
+        if (!pipeline_parse_in_proactor_cached && (pending_input_ || io_buf_.InputLen() > 0))
```
romange, line 2783: Are you sure that this change is needed? 
If `pipeline_parse_in_proactor_cached` is true and its path was executed, then I expect both `pending_input_` and `io_buf_.InputLen() ` be empty.

romange (approved):

src/facade/dragonfly_connection.cc
glevkovich: I think the main thing is to ask is whether we need this break at all, and whether checking those two members still makes sense when` pipeline_parse_in_proactor_cached=true.`

Usually they are both empty - but not always. Two cases leave them non-empty:

1. The proactor's call to `ReadPendingInput` filled `io_buf_` -> `pending_input_` stays true.
2. The proactor's call to `ParseRedis` stopped on backpressure  or left a trailing partial command -> `io_buf_.InputLen() > 0`

Even so, in my opinion it doesn't matter: breaking is the wrong thing to do when the feature is on. If we want to improve things, we should do it in the proactor, not by breaking. Reasons:

- Breaking burns CPU. It runs the fiber through `ReplyBatch` and then usually breaks/exits anyway (parse_status == NEED_MORE is the common case) only to go back and read.
- When the proactor ran, the common case (under heavy load) is that `dispatch_waiting_count_` already grew during the hop (preemption) from commands the proactor parsed in - exactly the work the break would go fetch.
- Under backpressure (case 2) you don't want to read (and parse) more.
- That leaves only `pending_input_ == true` (case 1) as something that COULD justify the break.

In my opinion the real fix belongs in the proactor (and I didn't want to do it in this PR): we need to run the read+parse work in a loop per notification, since the buffer is limited. Once the proactor loops, this concern is completely void.

romange: I guess what bothers me is the complexity of maintaining all these conditions. Lets make sure we get rid of most of them (incuding what I added) before 1.40 is released.

PR #7729, opened 2026-06-29
chore: cf chunked serialization and seeder support

kostasrim:
* add chunking to rdb save for CF
*  add CF support to seeder
* enable CF for seeder v2
* fix ACL's of CF family

### DIFF
--- a/src/server/cuckoo_filter_family.cc
+++ b/src/server/cuckoo_filter_family.cc
@@ -418,15 +418,17 @@ void RegisterCuckooFilterFamily(CommandRegistry* registry) {
   registry->StartFamily(acl::CUCKOO_FILTER);
 
   *registry << CI{"CF.RESERVE", CO::JOURNALED | CO::DENYOOM | CO::FAST, -3, 1, 1}.HFUNC(Reserve)
-            << CI{"CF.ADD", CO::JOURNALED | CO::DENYOOM | CO::FAST, 3, 1, 1}.HFUNC(Add)
-            << CI{"CF.ADDNX", CO::JOURNALED | CO::DENYOOM | CO::FAST, 3, 1, 1}.HFUNC(AddNx)
+            // It should really be fast but I am keeping compat.
+            << CI{"CF.ADD", CO::JOURNALED | CO::DENYOOM, 3, 1, 1}.HFUNC(Add)
+            << CI{"CF.ADDNX", CO::JOURNALED | CO::DENYOOM, 3, 1, 1}.HFUNC(AddNx)
             << CI{"CF.EXISTS", CO::READONLY | CO::FAST, 3, 1, 1}.HFUNC(Exists)
             << CI{"CF.MEXISTS", CO::READONLY | CO::FAST, -3, 1, 1}.HFUNC(MExists)
             << CI{"CF.INFO", CO::READONLY | CO::FAST, 2, 1, 1}.HFUNC(Info)
             << CI{"CF.COUNT", CO::READONLY | CO::FAST, 3, 1, 1}.HFUNC(Count)
-            << CI{"CF.DEL", CO::JOURNALED | CO::DENYOOM | CO::FAST, 3, 1, 1}.HFUNC(Del)
-            << CI{"CF.INSERT", CO::JOURNALED | CO::DENYOOM | CO::FAST, -4, 1, 1}.HFUNC(Insert)
-            << CI{"CF.INSERTNX", CO::JOURNALED | CO::DENYOOM | CO::FAST, -4, 1, 1}.HFUNC(InsertNx)
+            << CI{"CF.DEL", CO::JOURNALED | CO::FAST, 3, 1, 1}.HFUNC(Del)
+            << CI{"CF.INSERT", CO::JOURNALED | CO::DENYOOM, -4, 1, 1}.HFUNC(Insert)
+            << CI{"CF.INSERTNX", CO::JOURNALED | CO::DENYOOM, -4, 1, 1}.HFUNC(InsertNx)
+            // They mark it fast and read only. It's neither.
             << CI{"CF.COMPACT", CO::JOURNALED, 2, 1, 1}.HFUNC(Compact);
 }
 
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -350,6 +350,13 @@ void RdbLoaderBase::OpaqueObjLoader::operator()(const RdbCMS& src) {
 }
 
 void RdbLoaderBase::OpaqueObjLoader::operator()(const RdbCuckoo& src) {
+  if (config_.append) {
+    CuckooFilter* cf = pv_->GetCuckooFilter();
+    DCHECK(cf);
+    for (const std::string& blob : src.filters)
+      cf->AppendFilter(blob);
+    return;
+  }
   CuckooFilter* cf =
       CompactObj::AllocateMR<CuckooFilter>(CuckooFilterOptions{}, CompactObj::memory_resource());
   cf->Deserialize({src.slots_per_bucket, src.max_iterations, src.expansion, src.num_buckets,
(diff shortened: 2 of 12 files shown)


kostasrim:
augment review

src/server/rdb_save.cc
```diff
@@ -792,7 +792,19 @@ std::error_code RdbSerializer::SaveCuckooFilterObject(const PrimeValue& pv) {
   size_t num_filters = cf->NumFilters();
   RETURN_ON_ERR(SaveLen(num_filters));
   for (size_t i = 0; i < num_filters; ++i) {
-    RETURN_ON_ERR(SaveString(cf->FilterBytes(i)));
+    string_view blob = cf->FilterBytes(i);
+    bool is_last_filter = (i + 1 == num_filters);
+    RETURN_ON_ERR(SaveLen(blob.size()));
```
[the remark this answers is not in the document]
kostasrim: that won't happen, we haven't even released 😄

src/server/rdb_load.cc
```diff
@@ -2086,49 +2093,92 @@ io::Result<RdbLoaderBase::OpaqueObj> RdbLoaderBase::ReadCMS() {
 
 io::Result<RdbLoaderBase::OpaqueObj> RdbLoaderBase::ReadCuckoo() {
   RdbCuckoo res;
+  uint64_t num_filters = 0;
+
+  if (!pending_read_.cf_filter.has_value() && pending_read_.remaining == 0) {
+    uint64_t slots_per_bucket, max_iterations, expansion;
```
kostasrim, line 2099: When we don't have yet a pending partial cf_filter. Read slots, iteration, expansion (metadata) and create the pending read.

src/server/rdb_load.cc
```diff
@@ -2086,49 +2093,92 @@ io::Result<RdbLoaderBase::OpaqueObj> RdbLoaderBase::ReadCMS() {
...
+      return Unexpected(errc::rdb_file_corrupted);
+    pending_read_.cf_slots_per_bucket = res.slots_per_bucket;
+  } else {
+    num_filters = pending_read_.remaining;
+    pending_read_.remaining = 0;
+  }
 
-  uint64_t num_filters;
-  SET_OR_UNEXPECT(LoadLen(nullptr), num_filters);
+  auto read_cf_chunks = [&](std::string& data, size_t start_offset) -> io::Result<size_t> {
```
kostasrim, line 2131: As long as the budget allows drain into the data

src/server/rdb_load.cc
```diff
@@ -2086,49 +2093,92 @@ io::Result<RdbLoaderBase::OpaqueObj> RdbLoaderBase::ReadCMS() {
...
+      if (auto ec = FetchBuf(chunk_size, data.data() + curr_offset))
+        return make_unexpected(ec);
+      curr_offset += chunk_size;
+    }
+    return curr_offset;
+  };
 
-  if (num_filters == 0)
-    return Unexpected(errc::rdb_file_corrupted);
+  if (pending_read_.cf_filter) {
```
kostasrim, line 2146: This is when the previous run finished but we did not fully finished reading. The if above will be false and we will start filling the partial read filter until we exhaust the budget or read it fully. 

* if the partial read filter is not fully read we return early since the budget is drained
* partial filter filled we must continue and drain the budget

src/server/rdb_load.cc
```diff
@@ -2086,49 +2093,92 @@ io::Result<RdbLoaderBase::OpaqueObj> RdbLoaderBase::ReadCMS() {
...
+  // If the previous filter is complete and there are more start the next one
+  const uint8_t spb = pending_read_.cf_slots_per_bucket;
   res.filters.reserve(num_filters);
-  for (uint64_t i = 0; i < num_filters; ++i) {
-    std::string blob;
-    SET_OR_UNEXPECT(FetchGenericString(), blob);
-    // Empty blobs pass the modulo check (0 % n == 0) but would cause division-by-zero
-    // in BucketIndices (NumBuckets returns 0 → p.h1 % 0).
-    if (blob.empty() || blob.size() % res.slots_per_bucket != 0)
+  for (; num_filters > 0 && !ChunkBudgetExhausted(); --num_filters) {
```
kostasrim, line 2165: That's the drain phase of the above comment. The code above fully consumed the filter but the budget was not drain. Continue here until we exhaust the budget

src/server/rdb_load.cc
```diff
@@ -2086,49 +2093,92 @@ io::Result<RdbLoaderBase::OpaqueObj> RdbLoaderBase::ReadCMS() {
 
 io::Result<RdbLoaderBase::OpaqueObj> RdbLoaderBase::ReadCuckoo() {
```
kostasrim, line 2094: @abhijat this is the most important function in the changes. I left 3 comment on each section to exlain what the intent is. 

1. I am open if we can simplify it
2. If you have any questions around the logic shoot. I am happy to do this offline as well if the code ends up being too complicated

src/server/rdb_load.cc
```diff
@@ -2086,49 +2093,92 @@ io::Result<RdbLoaderBase::OpaqueObj> RdbLoaderBase::ReadCMS() {
...
+
+    std::string blob(total_size, '\0');
+    size_t offset = 0;
+    SET_OR_UNEXPECT(read_cf_chunks(blob, 0), offset);
+
+    if (offset < total_size) {
+      pending_read_.cf_filter = {std::move(blob), offset, spb};
+      pending_read_.remaining = num_filters;
+      return OpaqueObj{std::move(res), RDB_TYPE_CUCKOO};
+    }
```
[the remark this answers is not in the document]
kostasrim: we haven't release the code yet and we don;t need versioning.

src/server/cuckoo_filter_family.cc
```diff
@@ -418,15 +418,17 @@ void RegisterCuckooFilterFamily(CommandRegistry* registry) {
...
             << CI{"CF.MEXISTS", CO::READONLY | CO::FAST, -3, 1, 1}.HFUNC(MExists)
             << CI{"CF.INFO", CO::READONLY | CO::FAST, 2, 1, 1}.HFUNC(Info)
             << CI{"CF.COUNT", CO::READONLY | CO::FAST, 3, 1, 1}.HFUNC(Count)
-            << CI{"CF.DEL", CO::JOURNALED | CO::DENYOOM | CO::FAST, 3, 1, 1}.HFUNC(Del)
-            << CI{"CF.INSERT", CO::JOURNALED | CO::DENYOOM | CO::FAST, -4, 1, 1}.HFUNC(Insert)
-            << CI{"CF.INSERTNX", CO::JOURNALED | CO::DENYOOM | CO::FAST, -4, 1, 1}.HFUNC(InsertNx)
+            << CI{"CF.DEL", CO::JOURNALED | CO::FAST, 3, 1, 1}.HFUNC(Del)
+            << CI{"CF.INSERT", CO::JOURNALED | CO::DENYOOM, -4, 1, 1}.HFUNC(Insert)
+            << CI{"CF.INSERTNX", CO::JOURNALED | CO::DENYOOM, -4, 1, 1}.HFUNC(InsertNx)
+            // They mark it fast and read only. It's neither.
```
kostasrim, line 431: I decided to diverge.

tests/dragonfly/replication_test.py
```diff
@@ -5061,3 +5076,52 @@ async def bgsave_loop():
...
+        await wait_for_replicas_state(c_replica)
+
+    replica_results = await asyncio.gather(
+        *(c_replica.execute_command("CF.EXISTS", "cf", item) for item in random_items)
+    )
+    missing = sum(1 for r in replica_results if not r)
+    assert missing == 0, f"{missing} of {num_items} items missing on replica"
+
+    master.stop()
+    replica.stop()
```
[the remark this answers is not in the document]
kostasrim: nah

tests/dragonfly/seeder/__init__.py
```diff
@@ -138,6 +138,8 @@ async def _run_unit(self, client: aioredis.Redis, dtype: str, prefix: str):
 
 
 class Seeder(SeederBase):
+    DEFAULT_TYPES = SeederBase.DEFAULT_TYPES + ["CF"]
```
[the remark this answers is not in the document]
kostasrim: gold catch

src/server/rdb_load.cc
abhijat: its fine, its not too complicated to understand as it seems to follow the `ReadSBFImpl` logic which I updated recently for chunking 😄

abhijat: I thought of helpers to reduce duplication in this code and SBF but I think it would only make things less clear, this current structure is fine

tests/dragonfly/replication_test.py
```diff
@@ -5061,3 +5076,52 @@ async def bgsave_loop():
...
+
+    # kFilterChunkSize = 1 << 26 (64MB); peak should stay under 2x that.
+    MAX_CF_CHUNK_SIZE = 2**27
```
abhijat: nit: please make lower case to follow pep8

abhijat:
the chunking related changes look good, however I would suggest to add a unit test to focus on the split filter case to assert that it can be loaded correctly. there is a similar case in `SplitSBF` in `rdb_test.cc` for SBF

abhijat (approved):

PR #7732, opened 2026-06-29
feat: add range processing to CmdArgParser

BorysTheDev:
<h3>PR Summary by Qodo</h3>

Add counted and remaining argument ranges to CmdArgParser
<code>✨ Enhancement</code> <code>🧪 Tests</code> <code>🕐 40+ Minutes</code>

Grey Divider

<details open>
<summary>Description</summary>

<dl>
<dd>
<br/>

><pre>
>• Add iterable Range views for counted lists and remaining arguments in CmdArgParser.
>• Refactor CMS/HSET/HEXPIRE parsing to use ranges and tighten num-args validation.
>• Expand unit tests to cover range behavior and numfields mismatch cases.
></pre>

</dd>
</dl>

</details>

<details>
<summary>Diagram</summary>

<dl>
<dd>

<br/>

```mermaid
graph TD
  A["Command handlers"] --> B["CmdArgParser"] --> C["NextRange()"] --> D["Range view"] --> E["Business ops"]
  B --> F["RemainingRange()"] --> D
  E --> G[("Error replies")]

  subgraph Legend
    direction LR
    _h["Handler"] ~~~ _p["Parser"] ~~~ _v["View"] ~~~ _o["Ops"] ~~~ _e[("Error")]
  end
```

</dd>
</dl>

</details>

<details>
<summary>High-Level Assessment</summary>

<dl>
<dd>

<br/>

>The following are alternative approaches to this PR:

<details>
<summary><b>1. Return absl::Span instead of a custom Range</b></summary>

<dl>
<dd>

- ➕ Uses a well-known view type with clearer iterator semantics
- ➕ Can interop easily with algorithms expecting spans
- ➖ Harder to model “counted list consumes args” semantics without extra bookkeeping
- ➖ May require exposing/depending on underlying contiguous storage and lifetimes more directly

</dd>
</dl>

</details>

<details>
<summary><b>2. Eagerly materialize vectors for counted lists</b></summary>

<dl>
<dd>

- ➕ Simpler lifetime model (no parser-backed view)
- ➕ Avoids custom iterator implementation
- ➖ Allocates/copies for every command parse path
- ➖ Pushes reserve/validation logic into each caller unless centralized

</dd>
</dl>

</details>

>**Recommendation:** The PR’s Range view approach is a good fit: it centralizes counted-list validation (including “0 count” and short lists) inside CmdArgParser and avoids allocations while still supporting iteration and size() for reserving. Keep an eye on lifetime expectations (Range is only valid while the originating CmdArgParser is alive) and consider documenting that explicitly near Range/NextRange.

</dd>
</dl>

</details>

<details>
<summary> Files changed (7) <code> +186 / -71 </code> </summary>

<dl>
<dd>

<br/>

<details>
<summary>Enhancement (2) <code> +131 / -34 </code></summary>

<dl>
<dd>

<details>
<summary>cmd_arg_parser.h<code>Introduce CmdArgParser::Range with NextRange/RemainingRange helpers</code> <code>+100/-4</code></summary>

<br/>

>Introduce CmdArgParser::Range with NextRange/RemainingRange helpers
>
><pre>
>• Adds a lightweight iterable Range view over a parser-backed argument subrange, plus NextRange(group) for counted lists and RemainingRange() for consuming all remaining args. Adjusts Next()/Next(err_msg) return typing for conversion-aware defaults and fixes FInt conversion return style.
></pre>
>
><a href='https://github.com/dragonflydb/dragonfly/pull/7732/files#diff-d6013252da7fc24d360752d2bd33d4107b9482648ba0bfada090cf44c24273b4'>src/facade/cmd_arg_parser.h</a>

</details>

<details>
<summary>hset_family.cc<code>Switch HSET/HSETEX/HEXPIRE parsing and helpers to Range-based APIs</code> <code>+31/-30</code></summary>

<br/>

>Switch HSET/HSETEX/HEXPIRE parsing and helpers to Range-based APIs
>
><pre>
>• Introduces a shared kNumFieldsMismatch message and updates multiple helpers (listpack checks, sizing, OpSet, expiry checks) to accept CmdArgParser::Range. Reworks HSETEX and HEXPIRE parsing to use NextRange for numfields validation (including detecting extra trailing args) and templatizes SetFieldsExpireTime to accept either ParsedArgs or Range with explicit instantiations.
></pre>
>
><a href='https://github.com/dragonflydb/dragonfly/pull/7732/files#diff-4eaf710d62df976d913fde0544e4a9445457eea1b1b2fd7ad7d0db0b0fad054b'>src/server/hset_family.cc</a>

</details>

</dd>
</dl>

</details>

<details>
<summary>Refactor (3) <code> +18 / -37 </code></summary>

<dl>
<dd>

<details>
<summary>cms_family.cc<code>Use NextRange/RemainingRange for CMS MERGE key/weight parsing</code> <code>+13/-35</code></summary>

<br/>

>Use NextRange/RemainingRange for CMS MERGE key/weight parsing
>
><pre>
>• Refactors CMS merge argument parsing to use NextRange for the counted key list and RemainingRange for WEIGHTS. Tightens behavior for mismatch cases and simplifies weight parsing by iterating over the range.
></pre>
>
><a href='https://github.com/dragonflydb/dragonfly/pull/7732/files#diff-81a046fe35fc9a40014b8fff3ebab675f59e92d9fc9188ad7638bcede5041332'>src/server/cms_family.cc</a>

</details>

<details>
<summary>generic_family.cc<code>Pass RemainingRange into hash field-expire implementation</code> <code>+2/-1</code></summary>

<br/>

>Pass RemainingRange into hash field-expire implementation
>
><pre>
>• Wraps the provided values slice in CmdArgParser and forwards RemainingRange() to HSetFamily::SetFieldsExpireTime, enabling Range-based consumption without copying.
></pre>
>
><a href='https://github.com/dragonflydb/dragonfly/pull/7732/files#diff-f8f9387edb04c03ce7333ad3b69778d7b379fcf8c8765543f62b34d883e187c5'>src/server/generic_family.cc</a>

</details>

<details>
<summary>hset_family.h<code>Generalize SetFieldsExpireTime signature to accept Range-like field lists</code> <code>+3/-1</code></summary>

<br/>

>Generalize SetFieldsExpireTime signature to accept Range-like field lists
>
><pre>
>• Adds the CmdArgParser header dependency and updates SetFieldsExpireTime to be a template over the field-list type, supporting both ParsedArgs and CmdArgParser::Range call sites.
></pre>
>
><a href='https://github.com/dragonflydb/dragonfly/pull/7732/files#diff-934cabf32cee80b7fa718bf804097223e58cbaca0cf95979d0638234fae82463'>src/server/hset_family.h</a>

</details>

</dd>
</dl>

</details>

<details>
<summary>Tests (2) <code> +37 / -0 </code></summary>

<dl>
<dd>

<details>
<summary>cmd_arg_parser_test.cc<code>Add unit coverage for counted and remaining ranges</code> <code>+36/-0</code></summary>

<br/>

>Add unit coverage for counted and remaining ranges
>
><pre>
>• Adds RangeList tests validating iteration, size(), leftover preservation, count==0 handling, and short-list INVALID_CASES behavior.
></pre>
>
><a href='https://github.com/dragonflydb/dragonfly/pull/7732/files#diff-fa7c4926266553bb845c49e4be581ae35f7bc50431edca19bdf64775174b1314'>src/facade/cmd_arg_parser_test.cc</a>

</details>

<details>
<summary>hset_family_test.cc<code>Add test for extra-arg numfields mismatch in HTTL</code> <code>+1/-0</code></summary>

<br/>

>Add test for extra-arg numfields mismatch in HTTL
>
><pre>
>• Extends syntax error coverage to ensure extra trailing args after the declared numfields are rejected with a numfields-related error.
></pre>
>
><a href='https://github.com/dragonflydb/dragonfly/pull/7732/files#diff-20bc3fc9c9ebb4521b7a3498e8992925288afc762d6ed214736fc00cc014399e'>src/server/hset_family_test.cc</a>

</details>

</dd>
</dl>

</details>

</dd>
</dl>

</details>

### DIFF
--- a/src/facade/cmd_arg_parser.h
+++ b/src/facade/cmd_arg_parser.h
@@ -29,6 +29,8 @@ namespace facade {
 //   auto f  = parser.Next<FInt<1, 99>>("bad f");                // FInt with a custom out-of-range
 //                                                               // / non-integer error message
 //   auto count = parser.NextOrDefault<size_t>(10);              // read optional with default
+//   for (auto fld : parser.NextRange())  ...                    // [N, e1..eN] counted list
+//   for (auto a : parser.RemainingRange())  ...                 // all remaining args, no count
 //
 // Tag matching:
 //   parser.ExpectTag("LOAD");                                   // required literal keyword
@@ -112,6 +114,66 @@ struct CmdArgParser {
     ErrorReply MakeReply() const;
   };
 
+  // View over a counted arg list, from NextRange(). Yields string_view; size() for reserve().
+  class Range {
+   public:
+    Range() = default;
+
+    class iterator {
+     public:
+      using iterator_category = std::input_iterator_tag;
+      using value_type = std::string_view;
+      using difference_type = ptrdiff_t;
+      using pointer = const std::string_view*;
+      using reference = std::string_view;
+
+      iterator(const CmdArgParser* parser, size_t index) : parser_{parser}, index_{index} {
+      }
+      std::string_view operator*() const {
+        return parser_->SafeSV(index_);
+      }
+      iterator& operator++() {
+        ++index_;
+        return *this;
+      }
+      bool operator==(const iterator& o) const {
+        return index_ == o.index_;
+      }
+      bool operator!=(const iterator& o) const {
+        return index_ != o.index_;
+      }
+
+     private:
+      const CmdArgParser* parser_;
+      size_t index_;
+    };
+
+    iterator begin() const {
+      return {parser_, base_};
+    }
+    iterator end() const {
+      return {parser_, base_ + count_};
+    }
+    std::string_view operator[](size_t i) const {
+      return parser_->SafeSV(base_ + i);
+    }
+    size_t size() const {
+      return count_;
+    }
+    bool empty() const {
+      return count_ == 0;
+    }
+
+   private:
+    friend struct CmdArgParser;
+    Range(const CmdArgParser* parser, size_t base, size_t count)
+        : parser_{parser}, base_{base}, count_{count} {
+    }
+    const CmdArgParser* parser_ = nullptr;
+    size_t base_ = 0;
+    size_t count_ = 0;
+  };
+
  public:
   explicit CmdArgParser(ArgSlice args) : args_{args} {
   }
@@ -136,7 +198,8 @@ struct CmdArgParser {
   template <class T = std::string_view, class... Ts> auto Next() {
     if (cur_i_ + sizeof...(Ts) >= args_.size()) {
       Report(OUT_OF_BOUNDS, cur_i_);
-      return std::conditional_t<sizeof...(Ts) == 0, T, std::tuple<T, Ts...>>();
+      return std::conditional_t<sizeof...(Ts) == 0, decltype(Convert<T>(0)),
+                                std::tuple<T, Ts...>>();
     }
 
     if constexpr (sizeof...(Ts) == 0) {
@@ -153,14 +216,47 @@ struct CmdArgParser {
   // Like Next<T>(), but on a failed read (non-numeric, out-of-range FInt, or missing arg) replaces
   // the generic error with a caller-supplied CUSTOM_ERROR message. Pair with FInt<lo,hi> to attach
   // a custom out-of-range / non-integer message: parser.Next<FInt<1u, 99u>>("bad f").
-  template <class T = std::string_view> T Next(std::string_view err_msg) {
+  template <class T = std::string_view> auto Next(std::string_view err_msg) {
     bool prior = bool(error_);
-    T val = Next<T>();
+    auto val = Next<T>();
     if (!prior && error_ && !err_msg.empty()) {
       error_.type = CUSTOM_ERROR;
       error_.custom_msg = std::string{err_msg};
     }
     return val;
+  }
+
+  // Reads [count, e1..e(count*group)]; group=2 reads count pairs. Errors if count is 0 or short.
+  Range NextRange(size_t group = 1) {
+    uint32_t count = Next<uint32_t>();
+    size_t base = cur_i_;
+    // count==0 or fewer than count*group trailing args is a count/args mismatch, not a truncated
+    // command: report INVALID_CASES (syntax) at the count arg.
+    if (!error_ && (count == 0 || !HasAtLeast(count * group)))
+      Report(INVALID_CASES, base > 0 ? base - 1 : 0);
+    if (error_)
+      return {};
+    cur_i_ += count * group;
+    return Range{this, base, count * group};
+  }
+
+  // NextRange() with a custom error message on failure.
+  Range NextRange(size_t group, std::string_view err_msg) {
+    bool prior = bool(error_);
+    Range r = NextRange(group);
+    if (!prior && error_ && !err_msg.empty()) {
+      error_.type = CUSTOM_ERROR;
+      error_.custom_msg = std::string{err_msg};
+    }
+    return r;
+  }
+
+  // Consumes all remaining args as a Range (no leading count).
+  Range RemainingRange() {
+    size_t base = cur_i_;
+    size_t count = args_.size() - cur_i_;
+    cur_i_ = args_.size();
+    return Range{this, base, count};
   }
 
   template <class T = std::string_view> auto NextOrDefault(T default_value = {}) {
@@ -350,7 +446,7 @@ struct CmdArgParser {
     } else if constexpr (std::is_constructible_v<T, std::string_view>) {
       return static_cast<T>(SafeSV(idx));
     } else if constexpr (is_fint<T>) {
-      return {ConvertFInt<T::kMin, T::kMax>(idx)};
+      return ConvertFInt<T::kMin, T::kMax>(idx);
     }
   }
 
--- a/src/facade/cmd_arg_parser.h
+++ b/src/facade/cmd_arg_parser.h
@@ -7,6 +7,7 @@
 #include <absl/strings/match.h>
 #include <absl/strings/numbers.h>
 
+#include <cassert>
 #include <concepts>
 #include <optional>
 #include <string_view>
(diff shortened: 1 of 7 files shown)


src/server/hset_family.cc
```diff
@@ -1675,4 +1670,10 @@ vector<long> HSetFamily::SetFieldsExpireTime(const OpArgs& op_args, uint32_t ttl
   return res;
 }
 
+template vector<long> HSetFamily::SetFieldsExpireTime(const OpArgs&, uint32_t, ExpireFlags,
+                                                      string_view, facade::ParsedArgs, PrimeValue*);
```
kostasrim: Why do you need an explicit template instantiation ?! 

plz remove them

BorysTheDev: to avoid linkage error. it is template that is defined in header

kostasrim: I don't understand. 

If it's in the header then any source file can explicitly instantiate it right ? And templates are implicitly inline, so multiple definitions are folded by the linker among different translation units. 

p.s. declaring a header and defining it in a translation unit is an `anti-pattern`.

p.s.2 you do not need a template: `absl::Span<const string_view>` as a parameter and you materilize in one line within the function.

BorysTheDev: declaration in header, definition in .cc so to avoid a linkage error, we need to explicitly instantiate it

BorysTheDev: it is not anti-pattern it is standard approach

kostasrim: No it's not a standard approach. The definition belongs to the header and you don't need a template.

kostasrim (changes requested):
The template should be:

1. declared and defined in the header

It should not be declared in the header and defined in the source. This is wrong. The template is dead for other TU's like that because the compiler can only see the declaration and not the definition.

A better solution is to ditch the template. You don't need it and you can use `absl::span` instead.

P.s. I did not review the full PR. But templates split among declaration and definition in the header and the translation unit is a big no (unless you want to use extern + explicit instantiation and in the PR's context it is NOT the design pattern that should be applied)

src/server/hset_family.cc
kostasrim: And instead of arguing I would ask the AI what's wrong with the current approach. I am pretty sure it can explain as cleanly as what I write here

src/server/hset_family.h
```diff
@@ -36,7 +37,7 @@ class HSetFamily {
 
   static std::vector<long> SetFieldsExpireTime(const OpArgs& op_args, uint32_t ttl_sec,
                                                ExpireFlags flags, std::string_view key,
-                                               const facade::ParsedArgs& fields, PrimeValue* pv);
+                                               facade::CmdArgParser::Range fields, PrimeValue* pv);
 };
```
kostasrim, line 40: `SetFieldsExpireTime` should not take `CmdArgParser::Range` in the public header. `Range` is a lazy view that holds a raw `const CmdArgParser*` — it's an internal parser implementation detail leaking into a public class API, and it forces `#include "facade/cmd_arg_parser.h"` on every consumer of `hset_family.h`.

The parameter should be `absl::Span<const std::string_view>`. Both `ParsedArgs` and `CmdArgParser::Range` iterate over `string_view`; call sites can materialize the span in one line:
```cpp
absl::InlinedVector<std::string_view, 8> flds(r.begin(), r.end());
SetFieldsExpireTime(..., absl::MakeSpan(flds), pv);
```
The definition stays in the `.cc`, `UpdateTTL` stays in the `.cc`, nothing else changes.

src/server/generic_family.cc
```diff
@@ -858,8 +858,9 @@ OpResult<vector<long>> OpFieldExpire(const OpArgs& op_args, string_view key, uin
     SetFamily::DeleteSetIfEmpty(db_slice, op_args.db_cntx, key, *pv);
     return result;
   } else {
+    CmdArgParser fields_parser{values};
     auto result = HSetFamily::SetFieldsExpireTime(op_args, ttl_sec, ExpireFlags::EXPIRE_ALWAYS, key,
```
kostasrim, line 861: This is a direct consequence of the wrong parameter type in `SetFieldsExpireTime`. Creating a throwaway `CmdArgParser` just to produce a `Range` that holds a pointer back to it is a footgun — safe only because it's a single expression. If anyone separates it into two lines and stores the `Range`, it's a dangling pointer. Fix the API to `absl::Span<const std::string_view>` and this workaround disappears entirely.

src/facade/cmd_arg_parser.h
```diff
@@ -112,6 +114,66 @@ struct CmdArgParser {
...
+        ++index_;
+        return *this;
+      }
+      bool operator==(const iterator& o) const {
+        return index_ == o.index_;
+      }
+      bool operator!=(const iterator& o) const {
+        return index_ != o.index_;
+      }
+
```
kostasrim, line 155: Missing postfix `operator++(int)`. `std::input_iterator_tag` requires it; some stdlib paths (libc++, certain range algorithms) will fail without it.

kostasrim (commented):
The `Range`/`NextRange`/`RemainingRange` addition is good and the `cms_family.cc` refactor is a clear win. Two real issues below.

src/server/hset_family.h
BorysTheDev: Incorrect, Range is not private view

kostasrim (commented):
Ok after spending a little bit time on this:

The idea of using ranges to parse/check for errors is nice. 

However, my main issue with this PR is that it leaks `Range` type to functions (OpSet, etc) and creates a duality: some parts of the code end up accepting `ParsedArgs` while some others `Range`.  To avoid this (and the large number of changes the come with it) is to create a bridge. This way `parser logic` stays where it should and helpers accept universally `ParsedArgs`.

So for `CmdArgParser` is bridged via `UnparsedArgs()` and you get `ParsedArgs`

you need the same to convert a `Range` to `UnparsedArgs()`.

And `unless I missed something`:

Range adds error-checking on the count, but after that validation it's just the `tail` really ;)

src/server/hset_family.h
kostasrim: no it leaks in the helpers. Plz see my latest comment and how to clean this.

kostasrim (approved):
Good work! The Range will be useful :)

PR #7748, opened 2026-06-30
Adds `enable_resp_io_loop_v2` as a matrix dimension to the regression tests

glevkovich:
Adds `enable_resp_io_loop_v2` as a matrix dimension to the regression workflow so the full suite runs against both the legacy (v1) and new (v2) RESP IO loop, on both x86 and ARM.

- **`.github/workflows/regression-tests.yml`**: add `enable_resp_io_loop_v2: [false, true]` to the matrix and forward it to the regression action via the new `df-runtime-flags` input.
- **`.github/actions/regression-tests/action.yml`**: add an optional `df-runtime-flags` input that is translated into `--df <flag>` args and appended to the pytest invocation (both iouring and epoll paths).
- **`tests/dragonfly/instance.py`**: when a test spins up a binary older than 1.39, strip `enable_resp_io_loop_v2` / `enable_memcache_io_loop_v2` from both `args` and `params.args`. `DflyInstance` merges `params.args` on top of `args`, so the globally-injected flag would otherwise be re-added and crash old binaries used in replication/compat tests.

`build-type` (Debug, Release) × `runner` (x86, ARM64) × `enable_resp_io_loop_v2` (false, true) = 8 jobs.

- The `--df` flags are appended after `force_epoll=true` on the epoll path and as the sole runtime flags on the iouring path.

- Testing: when applying all changes under PRs: #7744, #7733, #7747 i all 4 jobs passed (x86) on my fork. For aarch64 we will have to check on the main repo.

### DIFF
--- a/.github/actions/regression-tests/action.yml
+++ b/.github/actions/regression-tests/action.yml
@@ -138,12 +138,17 @@ runs:
         export FILTER="${{inputs.filter}}"
         JUNIT_DIR="${REGRESSION_JUNIT_DIR}"
 
-        # Runtime flags forwarded to the dragonfly process via --df
-        DF_RUNTIME_FLAGS=""
-        if [[ -n "${{inputs.df-runtime-flags}}" ]]; then
-          for flag in ${{inputs.df-runtime-flags}}; do
-            DF_RUNTIME_FLAGS="$DF_RUNTIME_FLAGS --df $flag"
+        # Runtime flags forwarded to the dragonfly process via --df.
+        # Globbing is disabled while splitting so values like vmodule=*=1 are preserved,
+        # and the flags are collected into an array to avoid re-splitting/globbing
+        # when expanded on the pytest command line.
+        DF_RUNTIME_ARGS=()
+        if [[ -n "${DF_RUNTIME_FLAGS_INPUT}" ]]; then
+          set -f # disables filename expansion (globbing)
+          for flag in ${DF_RUNTIME_FLAGS_INPUT}; do
+            DF_RUNTIME_ARGS+=(--df "$flag")
           done
+          set +f
         fi
 
         # Exclude large tests unless explicitly requested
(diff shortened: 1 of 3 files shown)


.github/actions/regression-tests/action.yml
```diff
@@ -134,6 +138,14 @@ runs:
         export FILTER="${{inputs.filter}}"
         JUNIT_DIR="${REGRESSION_JUNIT_DIR}"
 
+        # Runtime flags forwarded to the dragonfly process via --df
+        DF_RUNTIME_FLAGS=""
+        if [[ -n "${{inputs.df-runtime-flags}}" ]]; then
+          for flag in ${{inputs.df-runtime-flags}}; do
```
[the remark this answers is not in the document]
glevkovich: fixed

.github/actions/regression-tests/action.yml
```diff
@@ -134,6 +138,14 @@ runs:
...
+            DF_RUNTIME_FLAGS="$DF_RUNTIME_FLAGS --df $flag"
+          done
+        fi
```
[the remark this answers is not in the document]
glevkovich: fixed

.github/actions/regression-tests/action.yml
```diff
@@ -134,6 +138,19 @@ runs:
         export FILTER="${{inputs.filter}}"
         JUNIT_DIR="${REGRESSION_JUNIT_DIR}"
 
+        # Runtime flags forwarded to the dragonfly process via --df.
```
kostasrim, line 141: Why not simply like filter above ? 

something like  `export RUNTIME_FLAG="${{inputs.runtime_flags_input}}"`

The caller can then just:

df-runtime-flags:"--df some_flag --df some_other_flag --df etc" 

Is there any added value for doing it like that below ?

glevkovich: - The real fix here is the use of bash array + set -f for **correct flag splitting**. 
- Passing the value via env: is just the GitHub-recommended hygiene (don't inline ${{ }} into run).  I'm not claiming it's a security feature, it is just the recommended way and I fixed it as a results of 2 bots Pr comments. For more info see : https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable .Quoted:
`For inline scripts, the preferred approach to handling untrusted input is to set the value of the expression to an intermediate environment variable.`

- To add on the array question: It helps to preserve the tokens through expansion. set -f and the array handle two different steps. set -f stops globbing while we split the input into individual flags (so for example vmodule=*=1 survives). The array then handles putting them back on the pytest command line. A plain string can't do this, it wo'tr work. So here I'm coming to solve edge issues which may be bugs when sending DF flags.

kostasrim: I dont have a strong opinion on this. My thinking around it is unless something is needed just support the basic use case. Additionally, from my understanding the set -f is required because we split and inject the `--df` to each flag. A string "--df flag1 --df flag2` would not require any special parsing/tokenizing which would make the set -f redundant at a small expense on the caller side. Imo inputs.df-runtime-flags is definitely trustworthy so I am not sure if the gh comment really applies here

kostasrim (approved):

.github/actions/regression-tests/action.yml
glevkovich: > I dont have a strong opinion on this. My thinking around it is unless something is needed just support the basic use case. Additionally, from my understanding the set -f is required because we split and inject the `--df` to each flag. A string "--df flag1 --df flag2` would not require any special parsing/tokenizing which would make the set -f redundant at a small expense on the caller side. Imo inputs.df-runtime-flags is definitely trustworthy so I am not sure if the gh comment really applies here


@kostasrim I get your point, but the pattern i'm trying to tackle isn't "exotic". As I said, I'm not trying to make the system "more secure" - I agree the gh injection note is about untrusted input and doesn't strictly apply here.

The thing to consider (and i'm taking --vmodule as an example,  but it can apply to any flag with */?):
With the simplified string form these can silently break depending on the working directory contents. The use of an array tohether with set -f just makes that more robust at no cost to the caller and with minimal added complexity here,  it also goe's  with gh's recommended pattern. I also think it might prevent extra work for future developers - so I'll leave it as is.

PR #7750, opened 2026-06-30
feat: Write connection throttling and metrics exposed

mkaruza:
Throttle pipeline batches containing write commands by introducing a configurable sleep interval.

The mutable --write_connection_throttling_sleep_usec option specifies the sleep duration in microseconds.
By default it is 0 which indicates that throttling is disabled.

The throttling policy is intentionally simple: if any command in a pipeline batch is a write command,
the entire batch is delayed by the configured sleep interval.

Expose the following metrics in INFO STATS and DEBUG IOSTAT:

- batch_read_commands_total
- batch_write_commands_total
- batch_read_bytes_total
- batch_write_bytes_total
- rw_throttle_batches_total

### DIFF
--- a/src/server/debugcmd.cc
+++ b/src/server/debugcmd.cc
@@ -522,24 +522,34 @@ struct IOStat {
   uint64_t cmd_total = 0, pipelined_cmd_total = 0;
   size_t io_read_bytes = 0;
   uint64_t io_reads_total = 0;
+  uint64_t rw_throttle_batch_read_commands = 0;
+  uint64_t rw_throttle_batch_write_commands = 0;
+  uint64_t rw_throttle_batch_read_bytes = 0;
+  uint64_t rw_throttle_batch_write_bytes = 0;
+  uint64_t rw_throttle_batches_total = 0;
 
-  void From(const facade::FacadeStats& fs);
+  void From(const facade::FacadeStats& fs, const ServerState::Stats& ss);
   void Print(RedisReplyBuilder* rb) const;
 
   IOStat& operator-=(const IOStat& other);
 };
 
-void IOStat::From(const facade::FacadeStats& fs) {
+void IOStat::From(const facade::FacadeStats& fs, const ServerState::Stats& ss) {
   conn_received = fs.conn_stats.conn_received_cnt;
   curr_conn_count = fs.conn_stats.num_conns_main;
   cmd_total = fs.conn_stats.command_cnt_main;
   pipelined_cmd_total = fs.conn_stats.pipelined_cmd_cnt;
   io_read_bytes = fs.conn_stats.io_read_bytes;
   io_reads_total = fs.conn_stats.io_read_cnt;
+  rw_throttle_batch_read_commands = ss.rw_throttle_batch_read_commands;
+  rw_throttle_batch_write_commands = ss.rw_throttle_batch_write_commands;
+  rw_throttle_batch_read_bytes = ss.rw_throttle_batch_read_bytes;
+  rw_throttle_batch_write_bytes = ss.rw_throttle_batch_write_bytes;
+  rw_throttle_batches_total = ss.rw_throttle_batches_total;
 }
 
 void IOStat::Print(RedisReplyBuilder* rb) const {
-  rb->StartCollection(6, CollectionType::MAP);
+  rb->StartCollection(11, CollectionType::MAP);
   rb->SendSimpleString("connections_received");
   rb->SendLong(conn_received);
   rb->SendSimpleString("current_conn_count");
@@ -552,6 +562,16 @@ void IOStat::Print(RedisReplyBuilder* rb) const {
   rb->SendLong(io_read_bytes);
   rb->SendSimpleString("io_reads_total");
   rb->SendLong(io_reads_total);
+  rb->SendSimpleString("rw_throttle_batch_read_commands");
+  rb->SendLong(rw_throttle_batch_read_commands);
+  rb->SendSimpleString("rw_throttle_batch_write_commands");
+  rb->SendLong(rw_throttle_batch_write_commands);
+  rb->SendSimpleString("rw_throttle_batch_read_bytes");
+  rb->SendLong(rw_throttle_batch_read_bytes);
+  rb->SendSimpleString("rw_throttle_batch_write_bytes");
+  rb->SendLong(rw_throttle_batch_write_bytes);
+  rb->SendSimpleString("rw_throttle_batches_total");
+  rb->SendLong(rw_throttle_batches_total);
 }
 
 IOStat& IOStat::operator-=(const IOStat& other) {
@@ -561,6 +581,11 @@ IOStat& IOStat::operator-=(const IOStat& other) {
   pipelined_cmd_total -= other.pipelined_cmd_total;
   io_read_bytes -= other.io_read_bytes;
   io_reads_total -= other.io_reads_total;
+  rw_throttle_batch_read_commands -= other.rw_throttle_batch_read_commands;
+  rw_throttle_batch_write_commands -= other.rw_throttle_batch_write_commands;
+  rw_throttle_batch_read_bytes -= other.rw_throttle_batch_read_bytes;
+  rw_throttle_batch_write_bytes -= other.rw_throttle_batch_write_bytes;
+  rw_throttle_batches_total -= other.rw_throttle_batches_total;
 
   return *this;
 }
@@ -1673,14 +1698,16 @@ void DebugCmd::IOStats(facade::CmdArgParser parser, CommandContext* cmd_cntx) {
   bool per_second = parser.HasNext() && absl::EqualsIgnoreCase(parser.Peek(), "PS");
   vector<IOStat> stats(shard_set->pool()->size());
 
-  shard_set->pool()->AwaitBrief(
-      [&](unsigned index, ProactorBase*) { stats[index].From(*facade::tl_facade_stats); });
+  shard_set->pool()->AwaitBrief([&](unsigned index, ProactorBase*) {
+    stats[index].From(*facade::tl_facade_stats, ServerState::tlocal()->stats);
+  });
 
   if (per_second) {
     ThisFiber::SleepFor(1s);
     vector<IOStat> stats2(shard_set->pool()->size());
-    shard_set->pool()->AwaitBrief(
-        [&](unsigned index, ProactorBase*) { stats2[index].From(*facade::tl_facade_stats); });
+    shard_set->pool()->AwaitBrief([&](unsigned index, ProactorBase*) {
+      stats2[index].From(*facade::tl_facade_stats, ServerState::tlocal()->stats);
+    });
 
     for (size_t i = 0; i < stats.size(); ++i) {
       stats2[i] -= stats[i];
--- a/src/server/debugcmd.cc
+++ b/src/server/debugcmd.cc
@@ -522,10 +522,10 @@ struct IOStat {
   uint64_t cmd_total = 0, pipelined_cmd_total = 0;
   size_t io_read_bytes = 0;
   uint64_t io_reads_total = 0;
-  uint64_t rw_throttle_batch_read_commands = 0;
-  uint64_t rw_throttle_batch_write_commands = 0;
-  uint64_t rw_throttle_batch_read_bytes = 0;
-  uint64_t rw_throttle_batch_write_bytes = 0;
+  uint64_t batch_read_commands_total = 0;
+  uint64_t batch_write_commands_total = 0;
+  uint64_t batch_read_commands_bytes = 0;
+  uint64_t batch_write_commands_bytes = 0;
   uint64_t rw_throttle_batches_total = 0;
 
   void From(const facade::FacadeStats& fs, const ServerState::Stats& ss);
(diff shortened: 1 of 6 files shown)


mkaruza:
### Benchmark - [rw_throttle_bench.py](https://github.com/user-attachments/files/29551482/rw_throttle_bench.py)

This benchmark measures the effect of `--write_connection_throttling` on read and write
latency/throughput under a noisy-neighbor workload: many concurrent readers and writers
pipelining commands against a shared, contended keyspace. The goal is to show whether delaying
write-containing batches (`write_connection_throttling_sleep_usec`) protects read latency, and
at what cost to writes.

Dragonfly server was run on EC2 instance with **8 CPU cores**
Benchmark was running on instance with **16 CPU cores**

**rw_throttle_bench.py** tool was run with arguments: ```--pipeline 64 --readers 128 --writers 128 --client-processes 16```

## Results

**Average latency vs. sleep duration** — read (blue) and write (red) latency diverge as the
sleep value increases, while each role's own baseline (dashed) stays flat:

latency_avg

**p99 latency vs. sleep duration** — the same divergence, more pronounced at the tail:

latency_p99

**Command throughput vs. sleep duration** — read throughput rises and write throughput falls as
sleep increases, with combined throughput roughly flat:

qps

src/server/main_service.cc
```diff
@@ -135,6 +136,10 @@ ABSL_RETIRED_FLAG(uint32_t, squash_stats_latency_lower_limit, 0,
                   "Deprecated. Squash latency stats are now tracked unconditionally; "
                   "use the pipeline_latency_seconds histogram for percentiles instead.");
 
+ABSL_FLAG(bool, write_connection_throttling, false, "Throttle write connections.");
```
romange: No need for two options. Combine them into 1 with default 0 - no throttling.

src/server/main_service.cc
```diff
@@ -1787,10 +1795,26 @@ uint32_t Service::DispatchSquashedBatch(facade::ParsedCommand* first, unsigned c
...
+      ss->stats.rw_throttle_batch_write_commands++;
+      ss->stats.rw_throttle_batch_write_bytes += cmd_cntx->EnqueuedBytes();
+    }
+
     cmd_refs.push_back(CmdRef{cid, tail_args, ReplyMode::FULL, cmd_cntx});
     cmd = cmd->next;
   }
 
+  // Simple policy that throttles batches that contain write commands by sleeping before dispatch.
+  if (absl::GetFlag(FLAGS_write_connection_throttling) && batch_has_write_command) {
```
romange: reverse these conditions

romange:
While I appreciate the effort to load-test, these results are not very credible. You can not benchmark Dragonfly with a python loadtest  like this. Was your python runner at 100% CPU during the test? How loaded Dragonfly was during the test?

What you CAN do is to use a slim python wrapper to orchestrate a run of two dfly_bench processes - one to inflict write-only traffic, another one to send read only - both with pipelined traffic. your P99 should stay at single ms levels, otherwise you either overload the server of the client or both.

romange:
`--pipeline 64 --readers 128 --writers 128 --client-processes 16` 

reduce number of readers and writers. reduce number of `client-processes` . It's very easy to load Dragonfly with `--pipeline 64 ` so tune down these parameters for your base as long as the throughput stays the same. You need a "stable" base, not overloaded. I would even say - your goal is to have Dragonfly at 80% CPU utilization on average. This is what a healthy, heavily-loaded production looks like.

src/server/main_service.cc
```diff
@@ -135,6 +136,9 @@ ABSL_RETIRED_FLAG(uint32_t, squash_stats_latency_lower_limit, 0,
                   "Deprecated. Squash latency stats are now tracked unconditionally; "
                   "use the pipeline_latency_seconds histogram for percentiles instead.");
 
+ABSL_FLAG(uint32_t, write_connection_throttling_sleep_usec, 500,
```
romange: this must be 0 by default. you do not want to do this to all the workloads

mkaruza: @copilot apply changes from comment

mkaruza:
## Benchmark  [rw_dfly_throttle_bench.py](https://github.com/user-attachments/files/29627915/rw_dfly_throttle_bench.py)

Benchmark tool is a slim wrapper that orchestrates two dfly_bench processes - one sending write-only traffic
(--ratio 1:0), another sending read-only traffic (--ratio 0:1) - concurrently against the same key range, both pipelined. 

### `write_connection_throttling_sleep_usec` benchmark results

Ran `rw_dfly_throttle_bench.py` on an 8-shard server (8 read + 8 write conns, pipeline 16, 15s/phase) sweeping the sleep value: 100 / 500 / 1000 / 5000 µs. Baseline (throttling off) is `stable across all runs (~1.0M ops/s reads and writes, ~1ms avg latency, ~92% CPU)`.

| sleep_usec | total qps | read ops/s (Δ) | write ops/s (Δ) | read.avg | read.p99 | write.avg | write.p99 |
|---|---|---|---|---|---|---|---|
| baseline | ~2.00M | ~1.01M | ~1.00M | ~1.00ms | ~1.85ms | ~1.00ms | ~1.85ms |
| 100  | 2.12M | 1.37M (**+36%**) | 748K (**−25%**) | 737us | 1.45ms | 1.36ms | 2.94ms |
| 500  | 2.14M | 1.52M (**+51%**) | 617K (**−38%**) | 661us | 1.20ms | 1.65ms | 3.25ms |
| 1000 | 2.08M | 1.57M (**+56%**) | 516K (**−48%**) | 643us | 1.17ms | 1.98ms | 4.15ms |
| 5000 | 2.25M | 2.07M (**+106%**)| 175K (**−83%**) | 486us | 844us | 5.85ms | 8.04ms |

**Takeaways:**
- **The throttle works as intended**: read throughput up, write throughput down, roughly monotonic with the sleep value. Total qps stays roughly flat (~2.0–2.25M) — throttling redistributes throughput between reads and writes rather than adding capacity.
- **100µs gives the smallest write latency/throughput hit** for a solid read gain (+36% reads, write p99 2.94ms vs ~1.85ms baseline).
- **500→1000µs is a weak marginal step**: read gain slows (+51%→+56%) while write cost keeps climbing (−38%→−48%, write p99 3.25ms→4.15ms).
- **5000µs is disproportionately costly**: write throughput collapses (−83%), write p99 hits 8ms, and CPU utilization drops (93.2%→87.5%), suggesting wasted idle time rather than useful work.

romange:
results good in general
using `--json_out_file=` and `./tools/plot_memtier_latency.py` you can show latency/throughput graphs instead of just final numbers

mkaruza:
Attaching file that contain:

1. analysis
2. script
3. results of benchmark (json+svg)

[rw-throttle-results.zip](https://github.com/user-attachments/files/29693883/rw-throttle-results.zip)

romange:
> pasting the analysis here

# Write Connection Throttling Benchmark Analysis

This benchmark compares Dragonfly read/write behavior with `write_connection_throttling_sleep_usec` set to 100, 500, 1000, and 5000 microseconds. Each run contains a baseline phase with throttling disabled and a throttled phase with the selected sleep value enabled.

The workload used two concurrent `dfly_bench` clients against the same key range:

| Role | Ratio | Connections | Threads | Pipeline |
|---|---:|---:|---:|---:|
| Read client | `0:1` | 8 | 8 | 16 |
| Write client | `1:0` | 8 | 8 | 16 |

## Summary

The throttle shifts capacity from writes to reads while keeping total throughput roughly flat at about 2.0M ops/sec. Larger sleep values continue improving read latency, but the write-side cost grows quickly. In this dataset, 100 usec looks like the best conservative tradeoff: it gives a meaningful read latency and read throughput improvement while preserving more write throughput than the larger settings.

| Sleep usec | Read ops/sec | Read p99 | Write ops/sec | Write p99 | Total ops/sec |
|---:|---:|---:|---:|---:|---:|
| 100 | 1,265,569 | 1.466 ms | 785,045 | 2.600 ms | 2,050,614 |
| 500 | 1,423,247 | 1.206 ms | 651,519 | 2.900 ms | 2,074,766 |
| 1000 | 1,543,794 | 1.171 ms | 497,705 | 3.640 ms | 2,041,499 |
| 5000 | 1,850,834 | 0.854 ms | 177,455 | 6.936 ms | 2,028,290 |

## Delta vs Baseline

Each row compares the throttled phase to the baseline phase from the same run.

| Sleep usec | Read ops delta | Read p99 delta | Write ops delta | Write p99 delta |
|---:|---:|---:|---:|---:|
| 100 | +23.9% | -17.4% | -21.5% | +44.8% |
| 500 | +38.9% | -32.5% | -35.7% | +62.5% |
| 1000 | +50.5% | -36.4% | -50.2% | +90.9% |
| 5000 | +77.5% | -51.1% | -81.9% | +288.7% |

## Baseline Consistency

The baseline phases are stable across the four runs, so the throttled deltas are comparable.

| Sleep run | Baseline read ops/sec | Baseline read p99 | Baseline write ops/sec | Baseline write p99 |
|---:|---:|---:|---:|---:|
| 100 | 1,021,759 | 1.775 ms | 999,594 | 1.795 ms |
| 500 | 1,024,468 | 1.786 ms | 1,012,483 | 1.785 ms |
| 1000 | 1,025,650 | 1.840 ms | 999,522 | 1.906 ms |
| 5000 | 1,042,435 | 1.745 ms | 980,567 | 1.784 ms |

## Interpretation

`100 usec` provides the best conservative tradeoff in these runs. It improves read throughput by 23.9% and reduces read p99 by 17.4%, while write throughput drops by 21.5%. This is the least disruptive setting and still shows a clear benefit for reads.

`500 usec` provides stronger read protection, reducing read p99 by 32.5%, but write throughput drops by 35.7%. This may be reasonable if the system should bias more heavily toward read latency, but it is a noticeably more expensive tradeoff than 100 usec.

`1000 usec` appears to be past the knee of the curve for this workload. Compared with 500 usec, read p99 only improves from 1.206 ms to 1.171 ms, while write throughput falls from 651k ops/sec to 498k ops/sec and write p99 rises from 2.900 ms to 3.640 ms.

`5000 usec` behaves more like a read-priority mode than a balanced throttle. It gives the best read numbers, but write throughput falls by 81.9% and write p99 increases by 288.7%.

## Recommendation

Based on these results, `100 usec` is the best candidate for a balanced default or conservative setting. It captures a meaningful portion of the read-latency benefit without the steeper write penalties seen at 500 usec and above.

If the goal is stronger read protection and reduced read p99 is worth a larger write-side cost, `500 usec` is the next reasonable candidate. The 1000 and 5000 usec settings do not look attractive for a balanced default under this workload.

## Artifact Privacy Note

This analysis intentionally omits the original benchmark endpoint. The raw `dfly_bench` JSON artifacts were sanitized so `configuration.server` is `127.0.0.1` and `configuration.port` is `6379`. The generated SVG charts did not appear to include the original endpoint in the metadata scan.

src/server/debugcmd.cc
```diff
@@ -552,6 +562,16 @@ void IOStat::Print(RedisReplyBuilder* rb) const {
   rb->SendLong(io_read_bytes);
   rb->SendSimpleString("io_reads_total");
   rb->SendLong(io_reads_total);
+  rb->SendSimpleString("rw_throttle_batch_read_commands");
```
romange: did you use it in your experiments? in other words why do we need per i/o thread stats?

src/server/server_state.h
```diff
@@ -137,6 +137,13 @@ class ServerState {  // public struct - to allow initialization.
     uint64_t psync_requests_total = 0;
     std::valarray<uint64_t> tx_width_freq_arr, squash_width_freq_arr;
 
+    // Throttling metrics
+    uint64_t rw_throttle_batch_write_commands = 0;
+    uint64_t rw_throttle_batch_read_commands = 0;
```
romange: I think the naming is misleading here.
`rw_throttle_batch_read_commands`  - we do not throttle reads.

Moreover, `rw_throttle_batch_write_commands` - it's unrelated to throttling.

I think a better name should be `batch_write_commands_total`, `batch_read_commands_total`, `batch_write_commands_bytes`, `batch_read_commands_bytes`

src/server/server_state.h
```diff
@@ -137,6 +137,13 @@ class ServerState {  // public struct - to allow initialization.
...
+    uint64_t rw_throttle_batch_read_bytes = 0;
+    uint64_t rw_throttle_batch_write_bytes = 0;
+    uint64_t rw_throttle_batches_total = 0;
```
romange, line 145: this is the only one that is related to throttling.

src/server/debugcmd.cc
mkaruza: With this simple throttling this is not used, but part of this PR was to expose all metrics that can be collected from batch (command count, bytes).

romange: i am not against the metrics, i was asking about per i/o thread collection. nm, we can keep it.

src/server/main_service.cc
romange: @mkaruza  it was not changed

romange (changes requested):
pls update the flag value

src/server/main_service.cc
mkaruza: @copilot apply again

mkaruza:
> pls update the flag value

strange, maybe i override history, copilot added this commit previously to disable throttling.

romange (dismissed):

src/server/main_service.cc
mkaruza: @copilot apply again

romange (approved):

PR #7753, opened 2026-06-30
Fix quadratic zstd list load compression

romange:
Summary:
Fix chunked RDB list loading with experimental zstd dictionaries so post-load compression scans only newly appended nodes after the initial pass.

Changes:
- Add incremental post-load compression for appended list chunks
- Preserve QList move state needed by chunked load handoff
- Add a regression test for chunked append after move

Validation:
- pre-commit run --files src/core/qlist.cc src/core/qlist.h src/core/qlist_test.cc
- ninja -C build-opt qlist_test && ./build-opt/qlist_test --gtest_filter=QListZstdTest.*
- ninja -C build-opt dragonfly
- Timed DFLY LOAD on the provided DFS snapshot with and without the zstd flag

### DIFF
--- a/src/core/qlist.cc
+++ b/src/core/qlist.cc
@@ -1626,13 +1626,12 @@ void QList::BackfillCompressWithZstdDict() {
     }
   }
 
-  // Only mark failure if we actually tried to compress nodes and all failed.
-  if (any_attempted) {
-    if (any_compressed)
-      dict_bulk_finished_ = 1;
-    else
-      dict_bulk_failed_ = 1;
-  }
+  // Avoid full scan again by setting either of flags.
+  // In case none of the nodes succeeded to compress, we will not attempt to compress again.
+  if (any_attempted && !any_compressed)
+    dict_bulk_failed_ = 1;
+  else
+    dict_bulk_finished_ = 1;
 }
 
 void QList::CompressAfterLoad() {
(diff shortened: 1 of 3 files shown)


src/core/qlist.cc
```diff
@@ -1620,16 +1627,19 @@ void QList::BackfillCompressWithZstdDict() {
   }
 
   // Only mark failure if we actually tried to compress nodes and all failed.
-  if (any_attempted && !any_compressed) {
-    dict_bulk_failed_ = 1;
+  if (any_attempted) {
+    if (any_compressed)
+      dict_bulk_finished_ = 1;
```
abhijat: lets say that all nodes we see are too small, so none are compressed, then this flag is not set. then we need to walk the full list every time, when called from `CoolOff`? because `dict_bulk_finished_` is not 1.

Wouldn't it be better to mark the list finished once we are able to walk its length, even if we didnt compress some nodes? What would walking over it a second time (and each time) accomplish, as we can't compress in the second/third/Nth pass anyway? 

The only false returns I see in `CompressNodeWithDict` is when the node is not compressible anyway (too small/not enough gain on compression) which are going to be permanent.

romange: hmm, you have a point.

romange: I will make it as conservative as possible.

abhijat (approved):

PR #7754, opened 2026-06-30
fix(transaction): Fix polling after running_tx conflict

### LINKED ISSUE #7698: crash seen during assert in `test_disconnect_replica`
seen twice recently

https://github.com/dragonflydb/dragonfly/actions/runs/28075026366/job/83117367582
https://github.com/dragonflydb/dragonfly/actions/runs/28090798722/job/83168025554

both times the same signature of the crash

```
dragonfly.unknown-user.ERROR.20260624-045334.22628.log
1:F0624 04:53:34.151309   22630 engine_shard.cc:622] Check failed: trans == nullptr || !trans->DEBUG_IsArmedInShard(sid) || trans->DEBUG_GetTxqPosInShard(sid) != TxQueue::kEnd exec_cb ZADD@4126/1:54 {id=2852 {cb_ptr=1}
2:F0624 04:53:34.151309   22630 engine_shard.cc:622] Check failed: trans == nullptr || !trans->DEBUG_IsArmedInShard(sid) || trans->DEBUG_GetTxqPosInShard(sid) != TxQueue::kEnd exec_cb ZADD@4126/1:54 {id=2852 {cb_ptr=1}

dragonfly.unknown-user.WARNING.20260624-045231.22628.log
9:F0624 04:53:34.151309   22630 engine_shard.cc:622] Check failed: trans == nullptr || !trans->DEBUG_IsArmedInShard(sid) || trans->DEBUG_GetTxqPosInShard(sid) != TxQueue::kEnd exec_cb ZADD@4126/1:54 {id=2852 {cb_ptr=1}
10:F0624 04:53:34.151309   22630 engine_shard.cc:622] Check failed: trans == nullptr || !trans->DEBUG_IsArmedInShard(sid) || trans->DEBUG_GetTxqPosInShard(sid) != TxQueue::kEnd exec_cb ZADD@4126/1:54 {id=2852 {cb_ptr=1}

dragonfly.unknown-user.INFO.20260624-045231.22628.log
610:F0624 04:53:34.151309   22630 engine_shard.cc:622] Check failed: trans == nullptr || !trans->DEBUG_IsArmedInShard(sid) || trans->DEBUG_GetTxqPosInShard(sid) != TxQueue::kEnd exec_cb ZADD@4126/1:54 {id=2852 {cb_ptr=1}
611:F0624 04:53:34.151309   22630 engine_shard.cc:622] Check failed: trans == nullptr || !trans->DEBUG_IsArmedInShard(sid) || trans->DEBUG_GetTxqPosInShard(sid) != TxQueue::kEnd exec_cb ZADD@4126/1:54 {id=2852 {cb_ptr=1}
```

the check is relatively new added in f87492e7f4f5244110bd63c78da96a395cecd447 
seen both times in epoll + debug mode

dranikpg:
The core problem is that running inline via `DispatchHop` does not follow up with PollExecution like running optimstic in `ScheduleInternal`. This issue never occurs because:
* running inline inside `DispatchHop` implies that the command could've ran inline inside `ScheduleInternal`, so it implies that this is a second hop
* We almost never suspend on a second hop

However I assume it must be possible with a good selection of multi/move/etc calls

The theoretical case to reproduce is quite simple
* multi-hop transaction A schedules on shards 0, 1 and runs the first hop on them, becoming the continuation
* two-hop transaction B runs on shard 0 inline its first and second hops - it suspends during the second hop
* transaction A submits its second hop and its PollExecution aborts as it sees transaction B blocked on shard 0
* because B ran with `DispatchHop`, it has no follow up and if no more transactions are submitted, A is never picked up

I even assume that there are more tricky cases like this - the simplest solution is just to cover all paths. However to not increase the overhead (as PollExecution reads atomics), we can do this re-poll conditionally only if the shard was interrupted

Fixes #7698

### DIFF
--- a/src/server/engine_shard.cc
+++ b/src/server/engine_shard.cc
@@ -606,24 +606,30 @@ void EngineShard::DestroyThreadLocal() {
   VLOG(1) << "Shard reset " << shard_id;
 }
 
-// Is called by Transaction::ExecuteAsync in order to run transaction tasks.
-// Only runs in its own thread.
 void EngineShard::PollExecution(const char* context, Transaction* trans) {
+  // Only one transaction can run at a time, so we have to abort the Poll while running_tx is
+  // suspended. Running tx issues a follow up Poll is needs_repoll_ was set
+  if (running_tx_) {
+    needs_repoll_ = true;
+    return;
+  }
+
+  // Call PollExecutionInternal and repeat the call if we interrupted someone elses Poll
+  do {
+    needs_repoll_ = false;
+    PollExecutionInternal(context, trans);
+    trans = nullptr;  // we repeat with null transaction
+    context = "repoll";
+  } while (needs_repoll_);
+}
+
+void EngineShard::PollExecutionInternal(const char* context, Transaction* trans) {
   DVLOG(2) << "PollExecution " << context << " " << (trans ? trans->DebugId() : "") << " "
            << txq_.size() << " " << (continuation_trans_ ? continuation_trans_->DebugId() : "");
 
+  DCHECK(running_tx_ == nullptr) << running_tx_->DebugId();
   ShardId sid = shard_id();
   stats_.poll_execution_total++;
-
-  // If another transaction is currently running on this shard (e.g. an inlined tx preempted
-  // on the connection fiber), we must not run any callbacks to avoid interleaving.
-  if (running_tx_) {
-    // The transaction (if any) if armed, must be in the txq so a future PollExecution picks it up.
-    DCHECK(trans == nullptr || !trans->DEBUG_IsArmedInShard(sid) ||
-           trans->DEBUG_GetTxqPosInShard(sid) != TxQueue::kEnd)
-        << context << " " << trans->DebugId();
-    return;
-  }
 
   // If any of the following flags are present, we are guaranteed to run in this function:
   // 1. AWAKED_Q -> Blocking transactions are executed immediately after waking up, they don't
--- a/src/server/engine_shard.cc
+++ b/src/server/engine_shard.cc
@@ -618,13 +618,18 @@ void EngineShard::PollExecution(const char* context, Transaction* trans) {
     return;
   }
 
-  // Call PollExecutionInternal and repeat the call if we interrupted someone elses Poll
-  do {
-    needs_repoll_ = false;
-    PollExecutionInternal(context, trans);
-    trans = nullptr;  // we repeat with null transaction
-    context = "repoll";
-  } while (needs_repoll_);
+  needs_repoll_ = false;
+  PollExecutionInternal(context, trans);
+
+  // Executing `trans` out of order above would not follow up with queue processing, so
+  // we need to issue another poll if we interruped another poll request.
+  // Because it runs with trans=nullptr, it progresses only on the loop and does not need follow ups
+  PollExecutionIfDeferred();
+}
+
+void EngineShard::PollExecutionIfDeferred() {
+  if (std::exchange(needs_repoll_, false))
+    PollExecutionInternal("repoll", nullptr);
 }
 
 void EngineShard::PollExecutionInternal(const char* context, Transaction* trans) {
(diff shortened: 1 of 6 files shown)


src/server/transaction_test.cc
```diff
@@ -0,0 +1,162 @@
+// Copyright 2026, DragonflyDB authors.  All rights reserved.
```
romange, line 1: nice idea to have a low level unit test. Would never think of this direction.

Did you wrote it yourself or asked AI to reproduce the scenario?

src/server/engine_shard.cc
```diff
@@ -606,25 +606,35 @@ void EngineShard::DestroyThreadLocal() {
...
-// Is called by Transaction::ExecuteAsync in order to run transaction tasks.
-// Only runs in its own thread.
 void EngineShard::PollExecution(const char* context, Transaction* trans) {
+  // Only one transaction can run at a time, so we have to abort the Poll while running_tx is
+  // suspended. Running tx issues a follow up Poll is needs_repoll_ was set
+  if (running_tx_) {
+    // Check that trans doesn't need processing or is guaranteed to be found via txq/continuation
+    DCHECK(trans == nullptr || !trans->DEBUG_IsArmedInShard(shard_id())  // tx was already disarmed
+           || trans->DEBUG_GetTxqPosInShard(shard_id()) != TxQueue::kEnd ||
+           continuation_trans_ == trans);
```
romange, line 616: nit: move `continuation_trans_ == trans` to be a second term before DEBUG_ checks

src/server/engine_shard.cc
```diff
@@ -606,25 +606,35 @@ void EngineShard::DestroyThreadLocal() {
...
+    needs_repoll_ = true;
+    return;
+  }
+
+  // Call PollExecutionInternal and repeat the call if we interrupted someone elses Poll
+  do {
+    needs_repoll_ = false;
```
romange: Why do we need this wrapper in PollExecution ? 

In other words, why PollExecutionIfDeferred is not enough to trigger omitted polls.

Or why do we need  PollExecutionIfDeferred if we have this loop?

src/server/transaction_test.cc
dranikpg: I had the idea and then asked AI to code it, you can see by the code being quite verbose

src/server/engine_shard.cc
dranikpg: Because in some places like `DispatchHop` we use only `PollExecution`. And the loop is needed because theoretically this can happen not once, but in a loop

dranikpg:
augment review

src/server/engine_shard.cc
dranikpg: Ok, it looks like it can't happen more than once

dranikpg:
augment review

src/server/transaction.cc
```diff
@@ -1620,18 +1621,32 @@ bool Transaction::CanRunInlined() const {
...
-    return true;
-  }
-  return false;
+  if (es->running_tx() != nullptr)
+    return false;
+
+  // Lastly: we refuse with blocking transactions. They run fully apart in PollExecution without
+  // being tracked in the tx queue, so a conflict with a suspended running tx can't be resolved.
+  if (auto* bc = namespace_->GetBlockingController(es->shard_id());
+      bc != nullptr && bc->HasBlockedTransactions())
```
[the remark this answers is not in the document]
dranikpg: Hm... Yes, the namespace thingy is ambigious, good comment

src/server/engine_shard.cc
```diff
@@ -606,25 +606,40 @@ void EngineShard::DestroyThreadLocal() {
...
+    return;
+  }
+
+  needs_repoll_ = false;
+  PollExecutionInternal(context, trans);
+
+  // Executing `trans` out of order above would not follow up with queue processing, so
+  // we need to issue another poll if we interruped another poll request.
+  // Because it runs with trans=nullptr, it progresses only on the loop and does not need follow ups
+  PollExecutionIfDeferred();
```
romange, line 627: it is the only call, imho.

Just remove the function and add the conditional call here.

romange (commented):
LGTM

src/server/engine_shard.cc
dranikpg: There's also one in transaction.cc, so not sure 🤔

romange (approved):

PR #7755, opened 2026-06-30
refactor: extract QList move helper and ZSTD dict predicates

romange:
## Summary
- extract a shared `MoveFrom()` path for `QList` move construction and move assignment
- rename the ZSTD bulk-compression failure state and helper entry point for clearer intent
- replace repeated inline ZSTD guard logic with small helper predicates without changing behavior

## Changes
- update `src/core/qlist.cc` to reuse `MoveFrom()` from the move constructor and assignment operator
- rename `dict_compress_failed_` to `dict_bulk_failed_` and `CompressWithZstdDict()` to `BackfillCompressWithZstdDict()`
- add `IsZstdDictMode()`, `IsInterior()`, and `CanCompressWithZstdDict()` in `src/core/qlist.h` and use them in the ZSTD compression paths

## Test plan
- Not run; refactor-only split of an existing branch commit.

### DIFF
--- a/src/core/qlist.cc
+++ b/src/core/qlist.cc
@@ -470,27 +470,15 @@ size_t QList::DefragIfNeeded(PageUsage* page_usage) {
 QList::QList(int fill, int compress)
     : fill_(fill),
       dict_learning_failed_(0),
-      dict_compress_failed_(0),
+      dict_bulk_failed_(0),
       dict_bulk_finished_(0),
       tiering_enabled_(0),
       compress_(compress),
       bookmark_count_(0) {
 }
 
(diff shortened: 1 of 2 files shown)


mkaruza (approved):
lgtm

kostasrim:
@romange before you merge this plz double check bot's comment. I understand you just moved the code as it was but it's write we don't copy/omit some fields

romange:
> @romange before you merge this plz double check bot's comment. I understand you just moved the code as it was but it's write we don't copy/omit some fields

you are correct, the next PR fixes this, it was too confusing to do all the changes in the same commit.

PR #7774, opened 2026-07-02
feat(server): Egress throttling for snapshot

dranikpg:
## Egress Throttling

image

_Problem_: Bandwidth for cloud instances can be limited. Fast snapshots to S3 storage, replication, migration, etc. can exceed the limit, causing heavy penalties to all network traffic, including regular client operations. Self sabotage by being too fast.

_Desired behavior_:  
* The snapshot is aware of the bandwidth limit and tries to slow down if needed to avoid exceeding it. 
* Out of order writes caused by client writes should either be not throttled at all or have direct priority over the traversal loop
* Progress guarantees should exist in either case to avoid stalling the traversal loop indefinitely

### Tracking egress

Done with GCA

### Limiting

The iteration loop will look at the current egress after each bucket or every few. If it is more than the limit, it will suspend its fiber until the next cell is cleared in the tracker (it should provide this information). It will then retry the condition.

However to ensure progress, the loop should also measure egress that comes directly from itself. It should contribute to at least 10% of the total volume allowed to guarantee stable progress. Loop egress is tracked with a separate SlidingTracker. It is okay if those 10% together overfill the limit.

### Choosing a time unit

| Time source | Cost | Accuracy / caveat |
|---|---|---|
| `base::CycleClock::Now()` (rdtsc) | ~cheap | Needs cycles→ns calibration; drifts on VMs / non-invariant TSC; not wall-clock |
| Proactor cached monotonic (`GetMonotonicTimeNs()`) | ~free (cached read) | Refreshed only once per scheduler round-trip → lags real time by other fibers' runtime (ms-scale under load) |
| `absl::GetCurrentTimeNanos()` | multiple of rdtsc| Accurate clock |

I've chosen the `GetCurrentTimeNanos` because the other two can be very unaccurate. Without any limit the timestamp is not queried

### DIFF
--- a/src/server/detail/egress_throttle.cc
+++ b/src/server/detail/egress_throttle.cc
@@ -6,40 +6,69 @@
 
 #include <absl/time/clock.h>
 
+#include <algorithm>
 #include <chrono>
 
+#include "base/logging.h"
 #include "util/fibers/fibers.h"
 
 namespace dfly::detail {
 
 using namespace std;
 
-void EgressThrottler::Record(uint64_t bytes, bool out_of_order) {
-  if (!enabled())
-    return;
+namespace {
 
-  socket_.IncBy(bytes);
-  if (!out_of_order)
-    loop_.IncBy(bytes);
+// CycleClock may drift in real time units, cached monotonic time is updated only once per scheduler
+// round, so use real time with higher query cost and possible backwards looping
+inline uint64_t NowUs() {
+  return static_cast<uint64_t>(absl::GetCurrentTimeNanos()) / 1000;
 }
 
-void EgressThrottler::Throttle() {
+}  // namespace
+
+void EgressThrottler::RecordAt(uint64_t bytes, bool out_of_order, uint64_t now_us) {
   if (!enabled())
     return;
 
-  const uint64_t reserve = static_cast<uint64_t>(limit_ * kGuaranteedRatio);
+  // bytes * kMicrosPerSec would overflow only at ~18TB
+  DCHECK_GT(std::numeric_limits<uint64_t>::max() / kMicrosPerSec, bytes);
+
+  socket_tat_ = std::max(now_us, socket_tat_) + bytes * kMicrosPerSec / limit_;
+  if (!out_of_order) {
+    uint64_t reserve = std::max<uint64_t>(1, limit_ / kLoopReserveDivisor);
+    loop_tat_ = std::max(now_us, loop_tat_) + bytes * kMicrosPerSec / reserve;
+  }
+}
+
+uint64_t EgressThrottler::WakeTime(uint64_t now_us) const {
+  if (!enabled())
+    return 0;
+
+  if (socket_tat_ <= now_us + kBurstToleranceUs)  // socket under budget
+    return 0;
+
+  if (loop_tat_ <= now_us + kBurstToleranceUs)  // loops share
+    return 0;
+
+  return std::min(socket_tat_, loop_tat_) - kBurstToleranceUs;
+}
+
+void EgressThrottler::Record(uint64_t bytes, bool out_of_order) {
+  RecordAt(bytes, out_of_order, NowUs());
+}
+
+void EgressThrottler::Throttle() const {
+  if (!enabled())
+    return;
+
   while (true) {
-    if (socket_.Sum() <= limit_)  // under budget
+    uint64_t now = NowUs();
+    uint64_t wake = WakeTime(now);
+    if (wake == 0)
       break;
 
-    if (loop_.Sum() < reserve)  // not reached guaranteed throughput
-      break;
-
-    // Suspend until the current slice ends and stale bytes roll off - the earliest point
-    // Sum() can decrease. The wait is the real time until the next slot boundary.
-    uint64_t now_ns = static_cast<uint64_t>(absl::GetCurrentTimeNanos());
-    uint64_t wait_ns = kEgressBucketNs - (now_ns % kEgressBucketNs);
-    util::ThisFiber::SleepFor(chrono::microseconds(wait_ns / 1000 + 1));
+    DCHECK_GT(wake, now);
+    util::ThisFiber::SleepFor(chrono::microseconds(wake - now + 1));
   }
 }
 
--- a/src/server/detail/egress_throttle.cc
+++ b/src/server/detail/egress_throttle.cc
@@ -1,0 +1,46 @@
+// Copyright 2024, DragonflyDB authors.  All rights reserved.
+// See LICENSE for licensing terms.
+//
+
+#include "server/detail/egress_throttle.h"
+
+#include <absl/time/clock.h>
+
+#include <chrono>
+
+#include "util/fibers/fibers.h"
+
+namespace dfly::detail {
+
+using namespace std;
+
+void EgressThrottler::Record(uint64_t bytes, bool out_of_order) {
+  if (!enabled())
+    return;
+
+  socket_.IncBy(bytes);
+  if (!out_of_order)
+    loop_.IncBy(bytes);
+}
+
+void EgressThrottler::Throttle() {
+  if (!enabled())
+    return;
+
+  const uint64_t reserve = static_cast<uint64_t>(limit_ * kGuaranteedRatio);
+  while (true) {
+    if (socket_.Sum() <= limit_)  // under budget
+      break;
+
+    if (loop_.Sum() < reserve)  // not reached guaranteed throughput
+      break;
+
+    // Suspend until the current slice ends and stale bytes roll off - the earliest point
+    // Sum() can decrease. The wait is the real time until the next slot boundary.
+    uint64_t now_ns = static_cast<uint64_t>(absl::GetCurrentTimeNanos());
+    uint64_t wait_ns = kEgressBucketNs - (now_ns % kEgressBucketNs);
+    util::ThisFiber::SleepFor(chrono::microseconds(wait_ns / 1000 + 1));
+  }
+}
+
+}  // namespace dfly::detail
--- a/src/server/detail/egress_throttle.cc
+++ b/src/server/detail/egress_throttle.cc
@@ -24,7 +24,7 @@ constexpr uint64_t kLowPrioReserveDivisor = 10;  // 1/10 = 10%
 
 // Burst tolerance (GCRA tau): how far ahead of schedule a stream may run. Absorbs chunk
 // granularity and avoids tiny frequent sleeps without affecting the long-run average rate.
-constexpr uint64_t kBurstToleranceUs = 100'000;  // 100ms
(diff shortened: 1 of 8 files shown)


src/server/snapshot.cc
```diff
@@ -198,6 +212,11 @@ void SliceSnapshot::IterateBucketsFb(bool send_full_sync_cut) {
           }
         }
       }
+
+      // Suspend the traversal loop if we are exceeding the egress budget, letting
+      // out-of-order writes drain first. Guarantees the loop its reserved share
+      throttler_.SetLimit(CurrentEgressLimitBytes());  // Re-read to pick up CONFIG SET
```
dranikpg: the other option is to add a callback to flag updates and update it only when the flag changes

romange:
1. btw helio has `util::SlidingCounter` 
2. GetMonotonicTimeNs being deprecated. Too confusing to use, to expensive to update

src/server/snapshot.cc
```diff
@@ -22,11 +22,16 @@
 #include "server/search/serialization_utils.h"
 #include "server/server_state.h"
 #include "server/tiered_storage.h"
+#include "strings/human_readable.h"
 #include "util/fibers/fibers.h"
 #include "util/fibers/stacktrace.h"
 #include "util/fibers/synchronization.h"
 
 ABSL_FLAG(bool, background_snapshotting, false, "Whether to run snapshot as a background fiber");
+ABSL_FLAG(strings::MemoryBytesFlag, snapshot_egress_limit_bytes, 0,
```
romange, line 31: I prefer having per thread limits.
1. It is consistent with most our limits.
2. It is consistent with how major clouds limit their instances (linearly with number of cpus). 

So the limit will be more or less the same no matter where we run.

dranikpg:
Relies on https://github.com/romange/helio/pull/608 now

dranikpg:
augment review

romange:
I had an idea how to simplify this, so I tinkered with AI to write it up. The motivation - to simplify the measurement state. See below:

## Recommendation

Use GCRA with byte-cost events. A byte chunk of size `n` at configured rate
`R` advances the theoretical-arrival time (TAT) by `n / R` seconds. This is
still a bandwidth limiter: `R` is bytes per second and chunks may be any size.
Its explicit burst tolerance must be selected deliberately (for example,
one bounded write chunk, rather than the implicit 50 ms/window behavior of
the sliding counter).

GCRA simplifies the limiter state and wait calculation: **one TAT and a
time-to-admit calculation replace a ring buffer, rolling window, bucket
expiration, and retry-to-next-boundary loop.**

Preserve the selected policy: out-of-order writes are never delayed and
traversal receives a guaranteed 10% of the configured rate, even if this
allows aggregate egress to exceed the configured limit. Model this with:

1. A global GCRA schedule at `limit` that accounts for every admitted byte.
2. A traversal-reservation GCRA schedule at exactly `limit / 10` that
   accounts only for traversal bytes.

Traversal normally uses the global schedule. When out-of-order traffic has
exhausted the global schedule, traversal can use its reservation schedule
instead. This preserves the PR's behavior: traversal can use all available
bandwidth in the absence of out-of-order traffic, but retains a 10% share
when it competes with it. The global schedule is still charged for reserved
traversal bytes, but cannot reject them; it may therefore accumulate debt.
This means the configured value remains a target for regular traffic, not an
absolute aggregate ceiling under sustained out-of-order traffic.

## GCRA design

`GcraRateLimiter` represents the scheduled finish time of admitted work, not
a count over a retrospective window. It has three values:

- `rate`: bytes per second; zero disables the limiter.
- `burst`: the maximum immediately available byte credit.
- `tat`: the theoretical arrival/finish time (TAT) of previously admitted
  traffic.

For a byte chunk `n`, its virtual duration is `n / rate` seconds. At monotonic
time `now`, the outstanding virtual debt is
`max(tat - now, 0)`. The chunk is admissible only when its debt plus its
virtual duration fits within the effective burst duration:

```text
cost       = n / rate
burst_time = burst / rate
effective_burst_time = max(burst_time, cost)  // allow one oversized record
allowed    = max(tat - now, 0) + cost <= effective_burst_time
next_tat   = max(tat, now) + cost
retry_at   = tat - (effective_burst_time - cost)
```

On admission, set `tat = next_tat`. On rejection, do not mutate `tat`; sleep
until `retry_at` and re-evaluate because another producer may have added
debt. This is a byte-cost version of GCRA/token-bucket behavior: it sustains
`rate` over time and permits at most `burst` bytes of accumulated credit.

Avoid floating point and per-record rounding drift. Internally represent the
comparison in scaled integer units, using a 128-bit intermediate:

```text
now_units   = now_ns * rate
cost_units  = n * 1_000_000_000
burst_units = burst * 1_000_000_000
```

`tat` is stored in these units. This makes the admission test exact for any
integer byte rate and chunk size; saturate rather than wrap on extreme input.
Changing the rate must either reset both schedules or convert their remaining
byte debt before the next admission. Preserve debt rather than silently
granting a new burst after `CONFIG SET`.

Serialized data remains atomic at the existing `HandleFlushData` record
granularity. A record larger than the configured byte credit is admitted as a
single burst, then advances TAT by its full virtual cost. This can overshoot
the instantaneous target, but is acceptable because the feature is not a
strict cap; it avoids data splitting, preserves the current serialization
pipeline, and converges back to the configured average rate afterwards.

The combined policy is:

```text
RecordOutOfOrder(n):
  global.Record(n, now)                 // never waits

AdmitTraversal(n):
  if global.CanAdmit(n, now):
    global.Record(n, now)
    reservation.Record(n, now)          // charge future reserved capacity
    return immediately
  if reservation.CanAdmit(n, now):
    global.Record(n, now)                // accounting only; may accumulate debt
    reservation.Record(n, now)
    return immediately
  sleep(min(global.RetryAfter(n, now), reservation.RetryAfter(n, now)))
  retry
```

Use the ratio `1/10` in integer arithmetic rather than a floating-point
`0.1`; this avoids the current small-limit truncation bug. Both state machines
are accessed on the owning shard's cooperative fibers, so their admission and
recording operations are atomic with respect to fiber scheduling and need no
thread mutex.

romange:
And I know you did all the work already but please revisit this unbiased

romange (commented):
reviewed

dranikpg:
Looks valid actually, let me simplify

dranikpg:
Done. Switched to GCA, ran it on a setup to see how it works in real use case: it had around 2% accuracy and peaked at 102mb. It can be achieved by reducing tau and lowering overshoot, though it might introduce jitter when its at limits

image

dranikpg:
Will also undo the helio changes

romange:
`it had around 2% accuracy` - wdym?  sounds pretty bad.

dranikpg:
> it had around 2% accuracy - wdym? sounds pretty bad.

Peak overshoot. GCRA allows overshoot

romange:
so 2% inaccuracy ? :)

src/server/detail/egress_throttle.cc
```diff
@@ -0,0 +1,75 @@
...
+// CycleClock may drift in real time units, cached monotonic time is updated only once per scheduler
+// round, so use real time with higher query cost and possible backwards looping
+inline uint64_t NowUs() {
+  return static_cast<uint64_t>(absl::GetCurrentTimeNanos()) / 1000;
+}
+
+}  // namespace
+
+void EgressThrottler::RecordAt(uint64_t bytes, bool out_of_order, uint64_t now_us) {
+  if (!enabled())
```
romange: nit: I prefer not fusing enabled/disabled state into such classes and handle this at the caller level.

src/server/detail/egress_throttle.h
```diff
@@ -0,0 +1,59 @@
...
+  bool enabled() const {
+    return limit_ > 0;
+  }
+
+  // Updates the byte/second limit, 0 disables all throttling
+  void SetLimit(uint64_t limit_bytes) {
+    limit_ = limit_bytes;
+  }
+
+  // Record egress and whether it occured from an out of order write.
```
romange: I do not know if `out of order` is the right term in this context. 
maybe `bool high_prio`

romange: On a second thought, we need limit regardless so it does not simplify

src/server/detail/egress_throttle.h
```diff
@@ -0,0 +1,59 @@
...
+// EgressThrottler allows to record bytes and throttle accordingly to keep the throughput
+// (bytes/second) under a predefined limit by suspending the fiber.
+// Differentiates between regular and out of order writes. Does not throttle if regular writes
+// did not reach a baseline limit - this guarantees stable progress for the operation even if
+// out of order writes eats all the limit (will exceed the limit).
+// Uses GCRA for the throttling calculations
+class EgressThrottler {
+ public:
+  // The loop is guaranteed at least limit/kLoopReserveDivisor of egress (reserved share).
+  static constexpr uint64_t kLoopReserveDivisor = 10;  // 1/10 = 10%
```
romange: move implementation constants to cc file

src/server/detail/egress_throttle.h
```diff
@@ -0,0 +1,59 @@
...
+
+  // Burst tolerance (GCRA tau): how far ahead of schedule a stream may run. Absorbs chunk
+  // granularity and avoids tiny frequent sleeps without affecting the long-run average rate.
+  static constexpr uint64_t kBurstToleranceUs = 100'000;  // 100ms
+
+  static constexpr uint64_t kMicrosPerSec = 1'000'000;
```
romange: also this one.

src/server/detail/egress_throttle.cc
```diff
@@ -0,0 +1,75 @@
...
+    uint64_t reserve = std::max<uint64_t>(1, limit_ / kLoopReserveDivisor);
+    loop_tat_ = std::max(now_us, loop_tat_) + bytes * kMicrosPerSec / reserve;
+  }
+}
+
+uint64_t EgressThrottler::WakeTime(uint64_t now_us) const {
+  if (!enabled())
+    return 0;
+
+  if (socket_tat_ <= now_us + kBurstToleranceUs)  // socket under budget
```
romange: I would align nomenclature: use `bool high_prio` for argument, and then `total_tat_ ` and `low_prio_tat_`

romange (commented):
Much simpler, some style comments.

src/server/detail/egress_throttle.h
dranikpg: Hm... Ok. Even though it seems like high_prio somehow shadows and decoupler a lot from the real logic

dranikpg:
Made the following changes:
1. Scaled down the "tau" to 50ms which is 5% theoretical overshoot
2. Scaled down the limit by 5%, so the "theoretical peak" is 5% less yet it remains responsive (due to the tau) and almost never overshoots the real limit

romange (approved):

PR #7779, opened 2026-07-03
fix(facade): Use ReplyScope in loop2

### LINKED ISSUE #7759: [test] connection_test.py::test_squashed_reply_count fails
https://github.com/dragonflydb/dragonfly/actions/runs/28510483277/job/84509314364

dranikpg:
Fixes #7759

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -2902,27 +2902,43 @@ bool Connection::ExecuteBatch() {
 }
 
 bool Connection::ReplyBatch() {
-  if (!HasInFlightCommands())
-    return true;
-
   ConnectionMemoryTracker memory_tracker(this);
   reply_builder_->SetBatchMode(true);
   absl::Cleanup batch_guard = [this] { reply_builder_->SetBatchMode(false); };
-  while (parsed_head_->CanReply()) {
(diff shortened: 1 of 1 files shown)


dranikpg:
augment review

src/facade/dragonfly_connection.cc
```diff
@@ -2902,28 +2902,44 @@ bool Connection::ExecuteBatch() {
 }
 
 bool Connection::ReplyBatch() {
-  if (!HasInFlightCommands())
```
romange, line 2905: why did you remove this condition?

romange (commented):
How to validate the improvement?
is there a regression test that covers this?

src/facade/dragonfly_connection.cc
dranikpg: I made it part of the loop

dranikpg:
> How to validate the improvement?
is there a regression test that covers this?

Yes, #7759

romange:
Ok, but for the reference, did you run any dfly_bench commands that show before and after?

romange (approved):

dranikpg:
> Ok, but for the reference, did you run any dfly_bench commands that show before and after?

Yes, it shows improvement

PR #7783, opened 2026-07-04
Optimize pipelined single-key command hot path

romange:
## Summary
Optimize the deeply pipelined single-key command path by removing redundant key analysis and reducing transaction setup overhead. Also disable memory tracking by default so regular builds avoid that debug-only overhead.

## Changes
- Carry the `KeyIndex` computed by `MultiCommandSquasher::TrySquash` into the squashed hop, letting `Transaction::InitByArgs` reuse it instead of running `DetermineKeys` again.
- Add `FIXED_SINGLE_KEY` command metadata and a `DetermineKeys` fast path for commands with one key at a fixed position.
- Replace `StoreKeysInArgs`' generic `KeyIndex::Range` iteration with plain indexed loops and reserve fingerprint storage up front.
- Default `DF_ENABLE_MEMORY_TRACKING` to `OFF`; custom debug builds can still opt in.

### DIFF
--- a/src/server/CMakeLists.txt
+++ b/src/server/CMakeLists.txt
@@ -1,4 +1,4 @@
-option(DF_ENABLE_MEMORY_TRACKING "Adds memory tracking debugging via MEMORY TRACK command" ON)
+option(DF_ENABLE_MEMORY_TRACKING "Adds memory tracking debugging via MEMORY TRACK command" OFF)
 option(PRINT_STACKTRACES_ON_SIGNAL "Enables DF to print all fiber stacktraces on SIGUSR1" OFF)
 
 option(WITH_COLLECTION_CMDS "Compile SET/HASH/ZSET/STREAM commands" ON)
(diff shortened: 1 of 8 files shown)


dranikpg (approved):
I'm really skeptic it helps at all. You skip `DetemineKeys` as if there is more than just a few if and else blocks. What shows up on the profile is computing Shard() multiple times, but we already optimized one away with `if (is_stub)` in InitByKeys

romange:
Based on the CPU profile it clearly helps, I actually had `before` and `after` results but they were probably overwritten in one of my iterations.  `DetermineKeys` was in top 3 in CPU profile and it went away entirely.   The improvement is around 3-4%  - not huge but still visible.

Note: I profiled an I/O thread that was processing pipelines at 100% CPU - shards threads were  running on other CPUs. This is a nice trick to be able to analyse each thread separately.

It did NOT help with throughput - the variance of socket performance is higher than such CPU optimizations.

romange:
One more thing - I have been running empty GET traffic at 8-10 million QPS on 16 CPU machine.  it's 500K+ qps per thread or less than 2usec CPU time on average per request. That includes reading from socket, parsing, request allocations, transaction allocations etc. As you can imagine with such small budget calling a function with 20 conditions can also slow down.

PR #2458, opened 2024-01-23
fix(transaction): Improve ACTIVE flags management

dranikpg:
Fix for #2457 

1. Previously we set the ACTIVE flag on the wrong shard for non-atomic multi transactions, however we determined our activeness by the `shard_id == unique_shard_id` rule, so the code was working correctly. Added a DCHECK that asserts this (yes it failed tests)

2. I'm now explicitly clearing local_mask for atomic multi in `MultiSwitchCmd` and DHCECK that it's cleared during the initialization

PR #7784, opened 2026-07-04
fix(server): add squasher-local transaction role

romange:
## Summary
Adds `MultiRole::SQUASHED_LOCAL` for non-atomic local squash transactions pinned to one shard. The role lets those transactions use single-cell `shard_data_` sizing while keeping normal scheduling and locking distinct from stub transactions.

## Changes
- Add `SQUASHED_LOCAL` to `MultiRole` and check it in `IsShardLocalMulti()`.
- Size `shard_data_` through `NeedsFullShardData()` for single-shard local transaction paths.
- Require `StartMultiNonAtomic()` call sites to pass `DEFAULT` or `SQUASHED_LOCAL` explicitly.
- Keep small transaction enum types compact and annotate implicit `RunnableResult` conversions.

### DIFF
--- a/src/server/transaction.h
+++ b/src/server/transaction.h
@@ -576,12 +576,17 @@ class Transaction {
   }
 
   // A non-atomic transaction pinned to a single shard for its whole lifetime (squasher-local).
+  // The pinning is enforced by MultiCommandSquasher, which only routes single-shard commands to
+  // such a tx and creates a separate one per shard, so it never fans out or migrates shards.
   bool IsShardLocalMulti() const {
     return multi_ && multi_->role == SQUASHED_LOCAL && multi_->mode == NON_ATOMIC;
   }
 
-  // Full per-shard shard_data_ is needed only for multi transactions that may span/migrate
-  // across shards. Stubs (SQUASHED_STUB) and shard-local non-atomic squash txs use a single cell.
+  // Full per-shard shard_data_ is needed only for multi transactions that may span/migrate across
+  // shards, whose leftover callbacks on other shards might still read shard_data_ after we moved on
+  // to the next command (shrinking it would be a cross-thread use-after-free). Stubs
+  // (SQUASHED_STUB) and shard-local non-atomic squash txs are single-shard for their whole
+  // lifetime, have no such leftover callbacks, and use a single cell.
   bool NeedsFullShardData() const {
     return IsActiveMulti() && !IsShardLocalMulti();
   }
--- a/src/server/transaction.h
+++ b/src/server/transaction.h
@@ -11,7 +11,6 @@
 #include <absl/functional/function_ref.h>
 
 #include <atomic>
-// #include <boost/smart_ptr/intrusive_ptr.hpp>
 #include <string_view>
 #include <variant>
 #include <vector>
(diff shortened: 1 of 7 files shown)


dranikpg (commented):
There was some kind of reason why non atomic multi transactions could not use single-cell array and I suppose it was because of unlocking or stale callbacks, but maybe the issue is resolved (we should document it at least). It might be that we don't need it at all, also this PR includes your other one

romange:
I do not know what was the reason. How do you suggest we find it?

romange:
I found this PR: https://github.com/dragonflydb/dragonfly/pull/4074

this is what you meant?

romange:
Actually, I think you are talking about this one: 
https://github.com/dragonflydb/dragonfly/pull/2458

which is without a test. I will try to see if I can reverse engineer the bug you fixed and reproduce it using AI.
If I succeed, I will add the test to this PR as well.

src/server/transaction.h
```diff
@@ -146,16 +145,19 @@ class Transaction {
...
-    SQUASHED_STUB = 2,  // Stub transaction
+  // Squashed parallel execution uses per-shard local transactions. Atomic squashing uses "stubs"
+  // that perform no scheduling or real hops and execute handlers directly inline. Non-atomic
+  // squashing uses local transactions that still schedule/acquire locks normally but are pinned to
+  // a single shard for their whole lifetime.
+  enum MultiRole : uint8_t {
+    DEFAULT = 0,         // Regular multi transaction
+    SQUASHER = 1,        // Owner of stub transactions
+    SQUASHED_STUB = 2,   // Stub transaction
+    SQUASHED_LOCAL = 3,  // Non-atomic squasher-local transaction
```
dranikpg: technically what you do is not squashing at all in transaction terms, we just dispatch on the target shard and hope it runs inline

dranikpg (dismissed):
It feels wrong because it cuts through many parts of code with a single optimization and doesn't ask any questions about the surroundings. 

First, this is technically not squashing from the transaction point of view. Second, there is this comment in `InitByKeys`

```c++
 // Multi transactions that execute commands on their own (not stubs) can't shrink the backing
 // array, as it still might be read by leftover callbacks.
```

what if we just never shrink it? We can resize to 1 the first time. So the whole optimization works implicitly

src/server/transaction.h
romange: ok, I can rename

romange:
> It feels wrong because it cuts through many parts of code with a single optimization and doesn't ask any questions about the surroundings.
> 
> First, this is technically not squashing from the transaction point of view. Second, there is this comment in `InitByKeys`
> 
> ```c++
>  // Multi transactions that execute commands on their own (not stubs) can't shrink the backing
>  // array, as it still might be read by leftover callbacks.
> ```
> 
> what if we just never shrink it? We can resize to 1 the first time. So the whole optimization works implicitly

Maybe it feels wrong because the original SidToId design is cumbersome and fragile.
Here I opened: https://github.com/dragonflydb/dragonfly/issues/7791
which simplifies the access and explicitly adopts single shard / multi-shard complexity. Let me know what you think there but I would not want to block this PR on #7791

dranikpg (approved):

PR #7787, opened 2026-07-05
fix(helm): remove .Release.Name to resolve name collisions

### LINKED ISSUE #7381: Modify Helm-Chart to better allow being used / bundled as a dependency
When using the current chart as a dependency, there seems to be no practical way to use a good name for all the ressources, since `Release.Name` ist used a lot of the time.

For example the volumes will be named "myapplication-data" which is confusing to say the least.

VincentNguyenDuc:
## Summary
StatefulSet's `serviceName` and PVC/volumeMount names used raw `.Release.Name` instead of the `dragonfly.fullname` helper, causing name collisions when the chart is used as a subchart dependency. Fixed as opt-in via `storage.useFullnameForVolumes` so existing storage-enabled releases render unchanged on upgrade (Kubernetes forbids changing `serviceName`/renaming `volumeClaimTemplates` in place). Also fixes truncation order so the combined `<fullname>-data` string respects the 63 char limit.

Fixes #7381

### DIFF
--- a/contrib/charts/dragonfly/templates/_helpers.tpl
+++ b/contrib/charts/dragonfly/templates/_helpers.tpl
@@ -24,6 +24,39 @@ If release name contains chart name it will be used as a full name.
 {{- end }}
 
 {{/*
+StatefulSet serviceName. Defaults to the release name, matching this chart's
+historical behavior so existing storage-enabled releases are unaffected on
+upgrade (Kubernetes does not allow changing serviceName in place). Set
+storage.useFullnameForVolumes to base it on the fullname instead, which
+avoids name collisions when this chart is installed as a subchart dependency.
+*/}}
+{{- define "dragonfly.serviceName" -}}
+{{- if .Values.storage.useFullnameForVolumes }}
+{{- include "dragonfly.fullname" . }}
+{{- else }}
+{{- .Release.Name }}
+{{- end }}
+{{- end }}
+
+{{/*
+Data volume / PVC name. Defaults to "<release-name>-data", matching this
+chart's historical behavior so existing storage-enabled releases are
+unaffected on upgrade (Kubernetes does not allow renaming
+volumeClaimTemplates in place). Set storage.useFullnameForVolumes to base it
+on the fullname instead, which avoids name collisions when this chart is
+installed as a subchart dependency. Truncates the combined string (rather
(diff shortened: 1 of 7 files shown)


romange:
Hello @VincentNguyenDuc . Thank you for your contributions! Are you helping your company with their Dragonfly prod deployments?

VincentNguyenDuc:
@romange Oh no 😄 I'm a student with some free time recently. I was actually trying to build something similar to DragonflyDB (but could not), so I'm learning from you guys here.

contrib/charts/dragonfly/templates/_helpers.tpl
```diff
@@ -23,6 +23,39 @@ If release name contains chart name it will be used as a full name.
...
+storage.useFullnameForVolumes to base it on the fullname instead, which
+avoids name collisions when this chart is installed as a subchart dependency.
+*/}}
+{{- define "dragonfly.serviceName" -}}
+{{- if .Values.storage.useFullnameForVolumes }}
+{{- include "dragonfly.fullname" . }}
+{{- else }}
+{{- .Release.Name }}
+{{- end }}
+{{- end }}
```
miledxz, line 39: Nice work,

worth calling out that this helper also fixes a latent bug when the flag is enabled. 

The Helm chart's only `Service` is named via `dragonfly.fullname` (e.g. `test-dragonfly`), 
but the legacy `serviceName` here resolves to `.Release.Name (e.g. test`), 

which points at a `Service` that doesn't exist. 

So with `useFullnameForVolumes: true`, `serviceName` becomes the fullname and finally matches the real Service name.


No change required ! 👍

miledxz (approved):
LGTM ! 👍

Abhra303 (approved):

VincentNguyenDuc:
Hi @romange, I believe we can go ahead and merge this PR?

vyavdoshenko:
@VincentNguyenDuc 
Please follow the instructions: https://github.com/dragonflydb/dragonfly/blob/main/CONTRIBUTING.md
Your commit should be signed before merging.

VincentNguyenDuc:
@vyavdoshenko
Done, rebased and signed. Thanks!

VincentNguyenDuc:
@vyavdoshenko, seem like it still need your approving review for the PR to be merged.

vyavdoshenko (approved):

Issue #7450, opened 2026-05-29
Transactions: allow per-shard callback preemption


romange:
## Problem

The transaction framework pays two system-wide costs to guard against a rare event (callback preemption during journal writes):

### 1. Inline scheduling disabled during replication

`AllowInlineScheduling()` returns false whenever journal callbacks are registered, and `CanRunInlined()` also checks `DbSlice::HasRegisteredCallbacks()`. This forces ALL single-shard transactions to be posted to the shard queue instead of running inline on the coordinator fiber — adding message dispatch overhead for every command during replication, even though actual preemptions are rare.

### 2. Intent locks acquired for optimistic single-shard transactions

In `ScheduleInShard`, intent locks are **always** acquired — even on the optimistic fast path where the transaction runs its callback inline and concludes immediately without ever entering the tx-queue. The code comment (`transaction.cc`, line 1207-1208) explains why:

```cpp
// We need to acquire the fp locks because the executing callback
// within RunCallback below might preempt.
const bool keys_unlocked = GetDbSlice(shard->shard_id()).Acquire(mode, lock_args);
```

The concern: if an optimistic transaction runs its callback without holding intent locks and that callback preempts, another transaction scheduling on the same shard would check the intent locks, find the keys uncontended, set `OUT_OF_ORDER`, and start executing its own callback — before the preempted one finishes, on the same keys.

The chain of events without locks:

```
1. Tx A: ScheduleInShard (optimistic, no locks acquired)
         RunCallback → preempts during journal write

2. Tx B: ScheduleInShard (on a different fiber, same thread)
         Check intent locks → keys appear free (Tx A holds no locks)
         Set OUT_OF_ORDER
         RunCallback → starts executing before Tx A finishes!

3. Tx A: resumes and finishes → operations on same keys interleaved
```

With intent locks held (current behavior), step 2 would see the keys as contended, Tx B would NOT get `OUT_OF_ORDER`, and it would be inserted into the tx-queue to wait. This is correct but expensive: every optimistic single-shard transaction pays the cost of `Acquire()` + `Release()` on the lock table, even though the vast majority never preempt.

## Goal

Track callback execution at the shard level so the system can handle the rare preemption case precisely, rather than pessimistically disabling inlining and acquiring intent locks for all transactions.

## Background: `continuation_trans_` vs `running_tx_`

[...]

PR #7788, opened 2026-07-05
perf(server): Lazy locking for optimistic single-shard transactions

romange:
## Summary

**Before:** for `execute_optimistic` cases we locked transactions because they could preempt . For all the cases where they do not preempt we did not need to lock them, just check that locks can be acquired. In fact - this is how we run several years ago - we only checked locks, and never grabbed them when execute_optimistic=true.
But today we punish the happy path for 99.9 - 100% cases

** This PR**: we check the locks but do not lock. But if we recognize preemption, we lock on  behalf of the preempted transaction in ScheduleInShard.

- Completes issue #7458 (follow-up to #7450): `ScheduleInShard` no longer registers intent locks up front for uncontended optimistic single-shard transactions.
- Common path (`running_tx_ == nullptr`): peeks the lock table via a new `DbSlice::IsLockFree` instead of calling `Acquire`/`Release`, so an uncontended, non-preempting callback never touches `trans_locks`.
- Rare path (`running_tx_ != nullptr`, i.e. a callback is mid-flight/preempted on this shard): lazily registers the running transaction's lock on its behalf before doing the normal conflict check, so OOO/queueing decisions stay correct. The running transaction releases the lazily-registered lock itself once it concludes.

## Changes
- `db_slice.{h,cc}`: add `DbSlice::IsLockFree(mode, lock_args)`, a non-mutating conflict check built on the existing `CheckLock`.
- `transaction.h`: add `Transaction::HasKeyLock`/`MarkKeyLockAcquired` accessors so scheduling can act on another in-flight transaction's per-shard lock state.
- `transaction.cc` (`ScheduleInShard`): restructure the optimistic-execution branch per the above; slow/queued path is unchanged.
- `transaction_test.cc`: new tests `OptimisticFastPathSkipsLockTable`, `LazyLockConflictingKey`, `LazyLockDisjointKeyMarkedOutOfOrder` (built on the existing `ParkHoldingRunningTx` harness); verified they fail against the pre-fix code.

## Test plan
- [x] `transaction_test` (8/8, including 3 new tests) passes
- [x] `dragonfly_test`, `multi_test`, `generic_family_test`, `string_family_test`, `blocking_controller_test`, `list_family_test`, `server_family_test` pass
- [x] `ctest -L DFLY` shows no new failures vs. unmodified `main` (remaining failures are pre-existing/unrelated, e.g. `acl_family_test`, and unbuilt test binaries)
- [x] Manual: `redis-benchmark`/`memtier_benchmark` under high contention (small keyspace) and concurrent `BGSAVE` (real snapshot-triggered preemption) with no crashes/DCHECKs

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -77,7 +77,6 @@ Checks: >
   readability-string-compare,
   readability-suspicious-call-argument,
   readability-uniqueptr-delete-release,
-  readability-use-anyofallof
 
 
 # Disabled because they're currently too disruptive, but one day might be nice to have:
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -1310,6 +1310,10 @@ bool DbSlice::IsLockFree(IntentLock::Mode mode, const KeyLockArgs& lock_args) co
 
 bool DbSlice::CheckLock(IntentLock::Mode mode, DbIndex dbid, uint64_t fp) const {
   const auto& lt = db_arr_[dbid]->trans_locks;
+  if (lt.Size() == 0) {
+    return true;
+  }
+
   auto lock = lt.Find(fp);
   if (lock) {
     return lock->Check(mode);
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -1300,6 +1300,14 @@ void DbSlice::Release(IntentLock::Mode mode, const KeyLockArgs& lock_args) {
   uniq_fps_.clear();
 }
 
+bool DbSlice::IsLockFree(IntentLock::Mode mode, const KeyLockArgs& lock_args) const {
+  for (LockFp fp : lock_args.fps) {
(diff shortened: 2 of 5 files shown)


dranikpg (dismissed):
Can you explain the optimizational idea? it's not less cpu work I assume, so it is to have more real ooo transactions?

romange:
it is to save CPU for common case. So for `execute_optimistic` cases we currently lock transactions because they may preempt . For all the cases where they do not preempt we do not need to lock them if locks can be acquired. In fact - this is how we run before - we only checked locks, and never grabbed them when  `execute_optimistic=true`. 

So we punished the happy path for 99.9 - 100% cases

romange:
> Can you explain the optimizational idea? it's not less cpu work I assume, so it is to have more real ooo transactions?

Updated the PR description

dranikpg (approved):
acquiring a lock in terms of cpu cost is basically zero

PR #7794, opened 2026-07-07
feat(server): Track family memory usage through transactions

abhijat:
WIP

### DIFF
--- a/src/server/command_registry.cc
+++ b/src/server/command_registry.cc
@@ -393,7 +393,7 @@ int TypeForFamily(size_t family) {
       kNoCommandMemoryType,  // cluster
       kNoCommandMemoryType,  // acl
   };
-  DCHECK_LT(kFamilyToType.size(), acl::NumberOfFamilies());
+  DCHECK_LT(family, kFamilyToType.size());
   return kFamilyToType[family];
 }
 
--- a/src/server/command_registry.cc
+++ b/src/server/command_registry.cc
@@ -214,6 +214,8 @@ CommandId CommandId::Clone(const std::string_view name) const {
     hdr_close(cloned.latency_histogram_);  // Free the histogram in the cloned object.
     cloned.latency_histogram_ = static_cast<hdr_histogram*>(latency_histogram_);
   }
+  cloned.SetFamily(GetFamily());
+  cloned.SetBitIndex(GetBitIndex());
   return cloned;
 }
 
--- a/src/server/command_registry.cc
+++ b/src/server/command_registry.cc
@@ -18,6 +18,7 @@
 #include "base/stl_util.h"
 #include "facade/dragonfly_connection.h"
 #include "facade/error.h"
+#include "redis/redis_aux.h"
 #include "server/acl/acl_commands_def.h"
 #include "server/conn_context.h"
 
@@ -360,6 +361,40 @@ absl::flat_hash_map<std::string, hdr_histogram*> CommandRegistry::LatencyMap() c
     cmd_latencies.insert({absl::AsciiStrToLower(cmd_name), cmd.GetLatencyHist()});
   }
   return cmd_latencies;
+}
+
+int TypeForFamily(size_t family) {
+  static constexpr int kNoCommandMemoryType = -1;
+  static const std::vector<int> kFamilyToType = {
+      kNoCommandMemoryType,  // core
+      kNoCommandMemoryType,  // server
+      kNoCommandMemoryType,  // generic
+      OBJ_LIST,              // list
+      OBJ_STRING,            // string
+#ifdef WITH_COLLECTION_CMDS
+      OBJ_SET,     // set
+      OBJ_HASH,    // hash
+      OBJ_ZSET,    // sorted_set
+      OBJ_STREAM,  // stream
+#endif
+#ifdef WITH_EXTENSION_CMDS
+      OBJ_ZSET,          // geo
+      OBJ_STRING,        // bitmap
+      OBJ_STRING,        // hyperloglog
+      OBJ_SBF,           // bloom
+      OBJ_CMS,           // cms
+      OBJ_TOPK,          // topk
+      OBJ_CUCKOOFILTER,  // cuckoo_filter
+      OBJ_JSON,          // json
+#endif
+#ifdef WITH_SEARCH
+      kNoCommandMemoryType,  // search
+#endif
+      kNoCommandMemoryType,  // cluster
+      kNoCommandMemoryType,  // acl
+  };
+  DCHECK_LT(kFamilyToType.size(), acl::NumberOfFamilies());
+  return kFamilyToType[family];
 }
 
 absl::flat_hash_map<std::string, CmdCallStats> CommandRegistry::NamedCallStats(
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -23,6 +23,7 @@ extern "C" {
 #include "search/doc_index.h"
 #include "server/channel_store.h"
 #include "server/cluster/slot_set.h"
+#include "server/cmd_memory_scope.h"
 #include "server/conn_context.h"
 #include "server/engine_shard_set.h"
 #include "server/error.h"
@@ -1912,6 +1913,9 @@ unique_ptr<base::Histogram> DbSlice::StopSampleValues(DbIndex db_ind) {
 void DbSlice::PerformDeletionAtomic(const Iterator& del_it, DbTable* table, bool async) {
   FiberAtomicGuard guard;
   size_t table_before = table->table_memory();
+
+  const int obj_type = del_it->second.ObjType();
+  CmdMemoryScope scope{obj_type};
 
   if (del_it->second.HasFlag()) {
     if (!SetMCFlag(table->index, del_it->first, 0)) {
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -81,6 +81,8 @@ void AccountObjectMemory(string_view key, unsigned type, int64_t size, DbTable* 
   DbTableStats& stats = db->stats;
 
   stats.AddTypeMemoryUsage(type, size);
+  if (type == OBJ_KEY)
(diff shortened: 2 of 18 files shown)


src/server/main_service.cc
```diff
@@ -3040,6 +3033,20 @@ void Service::RegisterCommands() {
...
+      "set",     "hash",   "sorted_set",  "stream",
+#endif
+#ifdef WITH_EXTENSION_CMDS
+      "geo",     "bitmap", "hyperloglog", "bloom",  "cms",    "topk", "cuckoo_filter", "json",
+#endif
+#ifdef WITH_SEARCH
+      "search",
+#endif
+      "cluster", "acl",
+  });
```
abhijat: this is pretty bad, maybe there is a better way

abhijat:
seems to cause CI failures (the DCHECK in the scope dtor firing)

ETA: fixed

src/server/transaction.cc
```diff
@@ -39,6 +39,53 @@ atomic_uint64_t op_seq{1};
...
+      unattributed_fiber_epoch_(IsUnattributed() ? util::fb2::FiberSwitchEpoch() : 0),
+      parent_(current_mem_accounting_scope) {
+  current_mem_accounting_scope = this;
+}
+
+CmdMemAccountScope::~CmdMemAccountScope() {
+  // An unattributed scope wraps non-yielding memory mutations outside a command family.
+  if (IsUnattributed())
+    DCHECK_EQ(unattributed_fiber_epoch_, util::fb2::FiberSwitchEpoch())
+        << "CmdMemAccountScope::Unattributed must not yield";
```
abhijat: maybe instead of DCHECK this should abort the current scope

abhijat: changed

src/server/main_service.cc
abhijat: it would require registering a name per family perhaps, but for now it might be okay, it is asserted in pytest

src/server/transaction.cc
```diff
@@ -681,7 +727,10 @@ void Transaction::RunCallback(EngineShard* shard) {
 
   RunnableResult result;
   try {
-    result = (*cb_ptr_)(this, shard);
+    {
+      CmdMemAccountScope mem_accounting{cid_->GetFamily()};
```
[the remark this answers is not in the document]
abhijat: added dcheck

src/server/db_slice.cc
```diff
@@ -327,7 +328,12 @@ int32_t AsyncDeleter::IdleCb() {
 
   auto* current = head_;
   DVLOG(2) << "IdleCb " << current->cursor;
-  if (current->step(current)) {
+  bool done = false;
+  {
+    auto mem_accounting = CmdMemAccountScope::Unattributed();
+    done = current->step(current);
+  }
+  if (done) {
```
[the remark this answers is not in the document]
abhijat: changed

src/server/engine_shard.cc
```diff
@@ -1097,6 +1100,20 @@ size_t EngineShard::UsedMemory() const {
...
+int64_t EngineShard::UsedMemoryForCommandAccounting() const {
+  return static_cast<int64_t>(mi_resource_.used()) + zmalloc_used_memory_tl;
+}
+
+void EngineShard::AddCommandFamilyMemDelta(size_t family, int64_t delta) {
+  if (delta == 0)
+    return;
+
+  if (family >= command_family_mem_delta_.size())
+    command_family_mem_delta_.resize(family + 1);
```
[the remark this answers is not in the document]
abhijat: added range check

abhijat:
augment review

abhijat:
augment review

src/server/engine_shard.cc
```diff
@@ -1097,6 +1101,26 @@ size_t EngineShard::UsedMemory() const {
...
+}
+
+void EngineShard::AddCommandFamilyMemDelta(size_t family, int64_t delta) {
+  if (delta == 0)
+    return;
+
+  constexpr size_t kMaxCommandFamilies = std::numeric_limits<uint64_t>::digits;
+  if (family >= kMaxCommandFamilies) {
+    LOG(DFATAL) << "Invalid command family index " << family;
+    return;
```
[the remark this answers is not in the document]
abhijat: I think this is overkill

src/server/transaction.cc
```diff
@@ -667,7 +729,11 @@ void Transaction::RunCallback(EngineShard* shard) {
 
   RunnableResult result;
   try {
-    result = (*cb_ptr_)(this, shard);
+    {
+      DCHECK(cid_);
+      CmdMemAccountScope mem_accounting{cid_->GetFamily()};
```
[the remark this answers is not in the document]
abhijat: fixed

abhijat:
augment review

src/server/command_registry.cc
```diff
@@ -378,4 +378,15 @@ absl::flat_hash_map<std::string, CmdCallStats> CommandRegistry::NamedCallStats(
   return res;
 }
 
+void CommandRegistry::SetFamilyNames(vector<string> names) {
+  CHECK_EQ(names.size(), acl::NumberOfFamilies());
+  family_names_ = std::move(names);
+}
+
+optional<string_view> CommandRegistry::FamilyName(size_t family) const {
+  if (family >= family_names_.size())
```
romange: Is it a viable path? why would someone pass an invalid index?

src/server/engine_shard.cc
```diff
@@ -1097,6 +1101,26 @@ size_t EngineShard::UsedMemory() const {
          search_indices()->GetUsedMemory();
 }
 
+int64_t EngineShard::UsedMemoryForCommandAccounting() const {
```
romange: why not `UsedMemory()` ?

src/server/transaction.cc
```diff
@@ -39,6 +39,59 @@ atomic_uint64_t op_seq{1};
...
+  generation_ = mem_accounting_generation;
+  unattributed_fiber_epoch_ = IsUnattributed() ? fb2::FiberSwitchEpoch() : 0;
+  parent_ = current_mem_accounting_scope;
+  current_mem_accounting_scope = this;
+}
+
+CmdMemAccountScope::~CmdMemAccountScope() {
+  // An unattributed scope wraps memory mutations outside a command family. If it yields, abort the
+  // active accounting chain instead of risking misattribution.
+  if (IsUnattributed() && unattributed_fiber_epoch_ != fb2::FiberSwitchEpoch()) {
```
romange: I do not understand the details but my intuition says that trying to have the same class/code for family and  "Unattributed" case, leads to unnecessary complexity, imho. This mergure is unnecessary as in EVERY place you know exactly whether it's attributed or unattributed - you call `Unattributed()`; method explicitly.

Therefore, you could have a clean code ( like CmdMemDeletionsScope class) or something else - that is dedicated for this specific use-case.

abhijat: I will look into this, btw I originally thought of adding a metric label for this kind of action, we don't know the command but we know what the action is, I didn't add it as I could not think of the metric being useful in terms of family 

But maybe it is useful, its not really a family but more like a category (bytes delta for defrag/expired/evicted etc).

romange: Wait, the high level goal is the same though. We need to measure LIST/SET/ZSET etc memory usage correctly - with  deletions taken into account. This is the high level goal. Are we aligned about this?

romange: The product goal - have the same "INFO MEMORY" output like today but with a new system. Intermediate step: have both metrics approaches exposed at the same time to be able to compare and ensure correctness of the new approach

abhijat: yes, the goal is the same, but I was talking about the following situation, for example:

* defrag task runs 
* it results in memory usage drop
* we mark it as gap / unattributed, because no command caused it - its a side effect which is not tied to a family/command/transaction

I was proposing that we can also have a metric for this side effect. ie make a label "defrag" and it will show us the delta caused by that. ie basically instead of saying unattributed, we can have metric for these side effects also, if it is useful.

This is different from the case covered by this PR, where we create an object via a family related command and transaction, and then delete via the family related command (eg HDEL). Then the usage is correctly taken into account because it flows through the transaction.

The only problematic part is generic DEL command etc because the delta goes into generic family, which does not tie back to the original family that created the key.

src/server/command_registry.cc
abhijat: I removed it, it was overly defensive

src/server/engine_shard.cc
abhijat: originally I added the new helper to exclude the search index memory, I made an error, the small string memory should also be part of the new helper.

But the search index memory is tricky. Its not part of object memories (whatever we return from MallocUsed() right now for keys). But memory changes in that are caused by command families. So 

* from object perspective it doesnt make sense to include it
* from family perspective, we should include it in the delta because increase/decrease in that number is caused by families

The other problem is that `search_indices()->GetUsedMemory()` is not constant time like the other three parts of `UsedMemory()`, as it loops over all existing indices to get non pmr memory usage. It will be very expensive to add it around each transaction callback. Its `local_mr_` usage is read in constant time.

abhijat: I added `SmallString::UsedThreadLocal()` to the new helper, maybe for search indices it can be split into the `local_mr_` (which is O(1) readable) and non pmr (which is not O(1) and should be skipped or cached)

abhijat:
fuzzer error
https://github.com/dragonflydb/dragonfly/actions/runs/29157982254/job/86558184616
[fuzz-smoke-resp-crashes-2938.zip](https://github.com/user-attachments/files/29960636/fuzz-smoke-resp-crashes-2938.zip)

abhijat:
https://github.com/dragonflydb/dragonfly/actions/runs/30004935259/job/89198594187?pr=7794

latest change broke many tests

ETA: does not reproduce, but I had introduced a bug which was causing many tests to crash (bad assert) it seems because of that fuzzing failed, will confirm

PR #7798, opened 2026-07-07
test: more benchmarks for Set

BorysTheDev:
Summary: This PR expands the benchmark coverage for Set implementations and adjusts some OAHSet internals.

Changes:

- Remove the local FORCE_INLINE macro and drop forced-inlining annotations from several OAHSet hot-path helpers.
- Change OAHSet probe-window configuration via kShiftLog (affecting kDisplacementSize) and adjust the grow threshold in AddImpl.
- Update OAHSet unit test expectations to reflect the new grow/load-factor behavior.
- Parameterize existing OAHSet and StringSet benchmarks by key size for more realistic workloads.
- Add new benchmarks for GetRandomMember, Scan, and Shrink for both set implementations.
- Normalize benchmark argument naming (KeySize).

Technical Notes: New benchmark matrices cover element counts up to 100k and key sizes up to 1000 bytes; OAHSet behavior changes (probe window + grow threshold) may affect both memory usage and runtime characteristics beyond tests/benchmarks.

### DIFF
--- a/src/core/oah_set.h
+++ b/src/core/oah_set.h
@@ -32,7 +32,7 @@ class OAHSet {  // Open Addressing Hash Set
   using Buckets = std::vector<TaggedPtr, StatelessAllocator<TaggedPtr>>;
 
  public:
-  static constexpr std::uint32_t kShiftLog = 5;                         // TODO make template
+  static constexpr std::uint32_t kShiftLog = 2;                         // TODO make template
   static constexpr std::uint32_t kMinCapacityLog = kShiftLog;           // should be >= ShiftLog
   static constexpr std::uint32_t kDisplacementSize = (1 << kShiftLog);  // TODO check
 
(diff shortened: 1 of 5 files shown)


BorysTheDev:
augment review

vyavdoshenko:
augment review

src/core/oah_set.h
```diff
@@ -32,7 +32,7 @@ class OAHSet {  // Open Addressing Hash Set
   using Buckets = std::vector<TaggedPtr, StatelessAllocator<TaggedPtr>>;
 
  public:
-  static constexpr std::uint32_t kShiftLog = 5;                         // TODO make template
+  static constexpr std::uint32_t kShiftLog = 2;                         // TODO make template
```
kostasrim, line 35: would it make sense to have these as abseil flags so we don't have to recompile again (read in the constructor so we can still have them as const)? That way you can play around with different capacities easily and you can test memory vs performance ? (doesn't ahve to be in this PR -- just throwing ideas)

kostasrim (approved):
lgtm!

src/core/oah_set.h
BorysTheDev: not now

PR #7802, opened 2026-07-07
feat(tiered): add --tiered_writes_cooling_bypass to keep write bursts out of the cool pool

shahyash2609:
## Summary

Adds an opt-in flag `--tiered_writes_cooling_bypass` (default `false`) that keeps **write/overwrite-driven** offloads out of the in-RAM cooling pool, so a burst of writes can no longer evict genuinely read-hot values down to disk.

## Motivation

When the tiered cooling layer (`--tiered_experimental_cooling`) is enabled, **every** offloaded value is retained in the in-RAM cooling pool (`cool_queue_`) — including values offloaded on the write/overwrite path. The cool pool is a fixed-size FIFO read cache; when it fills, `ReclaimMemory`/`PopCool` evicts from the back.

This hurts a common tiered-storage workload: **a large periodic write/refresh job mixed with latency-sensitive reads** (e.g. a catalog/geo dataset bulk-refreshed on a schedule while served read-heavy the rest of the time). During the refresh, the overwrite traffic streams freshly-written (cold, not-yet-read) values into the cooling pool, pushing genuinely read-hot entries out of RAM and onto disk — a read-latency regression that tracks the write job.

The write values don't belong in a *read* cache: nothing has read them yet. Retaining them only pays off if they're read back soon, which is not the case for a refresh/ingest job.

## What this changes

With `--tiered_writes_cooling_bypass=true`:

- Values offloaded via the **write path** are externalized **straight to disk**, bypassing the cooling pool.
- **Reads are unchanged** — a read still uploads the value back into RAM, so the read-hot working set is (re)admitted by actual reads, not by writes.
- **Background/scan-driven offload** (`RunOffloading`) is unaffected and keeps the normal cooling behavior.
- Only takes effect when `--tiered_experimental_cooling` is on. Default `false` → **zero behavior change unless explicitly enabled.**

## Implementation

The bypass intent is decided on the write path (`ShouldStash`) but consumed at stash completion (`SetExternal`), which is separated from the write by the async stash. It rides on the stash itself — **no side table**:

- A `bool bypass_cooling` on `StashDescriptor` (the "how to stash" option), set on the write path.
- It is carried across the async gap through `OpManager::Stash`/`PrepareAndStash` as an opaque `bypass_cooling` argument that is echoed back at completion via `NotifyStashed`/`ProcessStashed`. For an own-page (single-`PrimeValue`) stash the bit flows straight into `SetExternal(key, bypass_cooling, segment)`.
- For small-bin stashes the flag is recorded **per-key inside `SmallBins`** (the bin yields `[dbid, key, bypass, segment]` tuples), so batching preserves granularity when a bin mixes write-path and background entries; the bin-level flag is unused. List-node stashes never enter the cooling pool.
- At completion `SetExternal` calls `CoolDown` only when `experimental_cooling && !bypass_cooling`; otherwise it externalizes straight to disk (`PrimeValue::SetExternal`).

No `CompactObj`/`PrimeValue` state bit and no per-shard side set — the intent is passed as a stash parameter (per review; the earlier `cooling_bypass_pending_` map was dropped in `eb8a27a`).

## Benefit to users

- Makes the "write-heavy refresh + read-hot serving" pattern viable on tiered storage — the cool pool stays populated with values that are actually read, so read latency no longer degrades during write bursts.
- Fully opt-in and orthogonal to existing tiering behavior; off by default.

## Testing

- Added `TieredStorageTest.WritesCoolingBypass`: with cooling on, a stashed write grows `cold_storage_bytes` when the flag is off, and leaves it unchanged (disk-only) when the flag is on, while the value remains readable (uploaded back on access).
- CI is green across all build variants (including both `-Werror` g++ builds and the ubuntu-24 sanitizer build) and the test suite.

## Follow-up (not in this PR)

This addresses the **insertion** side (keep write-driven values out of the cool pool). A complementary **eviction**-side change — SIEVE/second-chance on cool eviction (`PopCool`), reusing the existing `touched` bit — would protect read-hot cooled values regardless of how they entered, and would make this insertion-side flag less necessary. Happy to follow up separately if the direction is agreeable.

### DIFF
--- a/src/server/tiered_storage.cc
+++ b/src/server/tiered_storage.cc
@@ -13,6 +13,7 @@
 #include <variant>
 
 #include "absl/cleanup/cleanup.h"
+#include "absl/container/flat_hash_set.h"
 #include "absl/flags/internal/flag.h"
 #include "absl/functional/bind_front.h"
 #include "absl/functional/overload.h"
@@ -63,6 +64,12 @@ ABSL_FLAG(float, tiered_upload_threshold, 0.1,
 ABSL_FLAG(bool, tiered_experimental_hash_support, false, "Experimental hash datatype offloading");
 
 ABSL_FLAG(bool, tiered_experimental_list_support, false, "Experimental list node offloading");
+
+ABSL_FLAG(bool, tiered_writes_cooling_bypass, false,
+          "If true, values offloaded as a result of a write/overwrite are externalized straight "
+          "to disk instead of being retained in the in-RAM cooling pool. Prevents write-heavy "
+          "bursts from evicting genuinely read-hot entries out of the cool pool; reads still "
+          "upload values back into RAM. Only takes effect when tiered_experimental_cooling is on");
 
 ABSL_FLAG(uint32, tiered_min_ttl_to_offload_ms, 5000,
           "Min remaining TTL in ms for a value to be eligible for offloading");
@@ -154,9 +161,18 @@ class TieredStorage::ShardOpManager : public tiering::OpManager {
       : tiering::OpManager{max_size}, ts_{ts}, db_slice_{*db_slice} {
   }
 
+  // Records that the stash for (dbid, key) should bypass the cooling pool and go disk-only. The
+  // intent must survive the async stash gap (write path -> async stash -> SetExternal), so it is
+  // kept in a per-shard set instead of a PrimeValue state bit. Only populated when bypass is set,
+  // so it stays empty in the common case.
+  void MarkCoolingBypass(DbIndex dbid, std::string_view key) {
+    cooling_bypass_pending_.emplace(dbid, key);
+  }
+
   // Clear Stash pending flag for entry
   void ClearStashPending(OpManager::KeyRef key) {
     UnblockBackpressure(key, false);
+    cooling_bypass_pending_.erase(key);
     if (auto pv = Find(key.first, key.second); pv) {
       pv->SetStashPending(false);
       stats_.total_cancels++;
@@ -181,6 +197,10 @@ class TieredStorage::ShardOpManager : public tiering::OpManager {
 
   void CancelStash(tiering::KeyRef id, size_t size) {
     UnblockBackpressure(id, false);
+    // A cancelled stash never reaches SetExternal/ClearStashPending, so drop any bypass marker
+    // here to avoid a stale entry (which could wrongly bypass cooling on a later stash of the
+    // same key, or leak). Covers both the own-page and small-bin sub-cases below.
+    cooling_bypass_pending_.erase(id);
     // TODO: Don't recompute size estimate, try-delete bin first
     if (OccupiesWholePages(size)) {
       CancelPending(id);
@@ -269,7 +289,13 @@ class TieredStorage::ShardOpManager : public tiering::OpManager {
       stats_.total_stashes++;
 
       StashDescriptor blobs{FragmentRef{*pv}.GetSerializationDescr()};
-      if (ts_->config_.experimental_cooling) {
+
+      // A write/overwrite driven stash bypasses the cooling pool (see
+      // --tiered_writes_cooling_bypass) and goes disk-only, so write bursts don't evict read-hot
+      // entries. The intent is carried in cooling_bypass_pending_ across the async stash gap.
+      bool bypass = cooling_bypass_pending_.erase(key) > 0;
+
+      if (ts_->config_.experimental_cooling && !bypass) {
         RetireColdEntries(pv->MallocUsed());
         ts_->CoolDown(key.first, key.second, segment, blobs.rep, pv);
       } else {
@@ -322,6 +348,12 @@ class TieredStorage::ShardOpManager : public tiering::OpManager {
 
   // When we don't have memory to upload a page for defragmentation, we save it here to do it later
   std::deque<uint32_t /* page index (kPageSize) */> delayed_defrag_queue_;
+
+  // Keys whose in-flight stash should bypass the cooling pool (see --tiered_writes_cooling_bypass).
+  // Stored as owning DbKeyId with heterogeneous lookup, so we can probe with an OpManager::KeyRef
+  // (which holds a string_view) without allocating. Populated only when bypass is requested.
+  absl::flat_hash_set<tiering::DbKeyId, tiering::detail::Hasher, tiering::detail::Eq>
+      cooling_bypass_pending_;
 
   TieredStorage* ts_;
   DbSlice& db_slice_;
@@ -523,6 +555,11 @@ void TieredStorage::StashPrimeValue(DbIndex dbid, string_view key, const StashDe
   tiering::OpManager::PendingId id;
   error_code ec;
 
+  // Carry the cooling-bypass intent per-key across the async stash gap. Done per-key (not per-bin)
+  // so small-bin batching keeps per-key granularity. Cleaned up in ClearStashPending on abort.
+  if (blobs.bypass_cooling)
+    op_manager_->MarkCoolingBypass(dbid, key);
+
   if (OccupiesWholePages(est_size)) {  // large enough for own page
     id = KeyRef(dbid, key);
     auto serialize = absl::bind_front(&StashDescriptor::Serialize, &blobs);
@@ -638,6 +675,7 @@ void TieredStorage::UpdateFromFlags() {
       .upload_threshold = absl::GetFlag(FLAGS_tiered_upload_threshold),
       .experimental_hash_offload = absl::GetFlag(FLAGS_tiered_experimental_hash_support),
       .experimental_list_offload = absl::GetFlag(FLAGS_tiered_experimental_list_support),
+      .writes_cooling_bypass = absl::GetFlag(FLAGS_tiered_writes_cooling_bypass),
       .min_ttl_to_offload_ms = absl::GetFlag(FLAGS_tiered_min_ttl_to_offload_ms),
   };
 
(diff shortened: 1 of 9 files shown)


src/server/tiered_storage.cc
```diff
@@ -64,6 +64,12 @@ ABSL_FLAG(bool, tiered_experimental_hash_support, false, "Experimental hash data
 
 ABSL_FLAG(bool, tiered_experimental_list_support, false, "Experimental list node offloading");
 
+ABSL_FLAG(bool, tiered_cold_overwrites, false,
```
romange: `tiered_writes_direct` (similarly to linux io direct mode) or `tiered_writes_cooling_bypass` ?

src/server/tiered_storage.cc
```diff
@@ -858,6 +872,11 @@ void StashPrimeValue(DbIndex dbid, std::string_view key, const PrimeKey& pk, Pri
           ts->ShouldStash(*pv, TieredStorage::StashContext{.key_expire_ms = pk.GetExpireTime()});
       blobs) {
     pv->SetStashPending(true);
+    // This is the write path (proactive offload driven by an edit). Mark the value cold so it
+    // skips the cooling pool and is externalized straight to disk. Background/scan-driven offload
+    // (RunOffloading) does not go through here and keeps the normal cooling behavior.
+    if (ts->ColdStashWrites())
```
romange: If I understand correctly, you use state bits in PrimeValue to pass parameters down the call chain.

I would rather we do not do it even if the change will become larger.
Instead I suggest you introduce a boolean variable in `StashDescriptor`  blobs.
I think we can extend its scope to serve the purpose of passing options on how to stash.

romange (commented):
Good feature, thanks for this contribution @shahyash2609

src/server/tiered_storage.cc
dranikpg: direct would be if we wouldn't even store it in a prime value

dranikpg (commented):
The broader problem of cooling is in general that it conflicts a little with "caching" - values that are pure in-memory even though it tries to go out of the way by retiring whenever possible. 

For example, there is no real reason to cool down a value that was offloaded. If offloading determined that it is not frequently used and it needs to free up ram, then why keep it any longer? It did not come from a recent write

There is also the catch that if the main share of frequently accessed keys came after their writes, they will be stuck in the cold queue. They should not stay there and be pure in-ram values, the more safe option. The cold queue is now only a potential threat to them as they will be first in the eviction queue. So we should think about moving from cooled to cold

src/server/tiered_storage.cc
shahyash2609: Renamed to `--tiered_writes_cooling_bypass`. Went with this over `tiered_writes_direct` per @dranikpg's point below — the value is still stored in a PrimeValue, we only skip the cool queue on stash completion.

src/server/tiered_storage.cc
shahyash2609: Done — dropped the `CompactObj` `stash_cold` bit entirely (the file is now untouched by this PR). The bypass intent is a `bool bypass_cooling` on `StashDescriptor`, set on the write path from `ShouldStash`. Since the descriptor doesn't survive to the async completion, the member `StashPrimeValue` records it per-key in a small per-shard set (`cooling_bypass_pending_`, populated only when bypass is requested); `SetExternal` consumes/erases it and `ClearStashPending` cleans up on abort. No PrimeValue state bit, and per-key granularity is preserved through small-bin batching.

src/server/tiered_storage.cc
shahyash2609: Agreed — that's exactly why I avoided `direct`; the value still lives in a PrimeValue, we only bypass the cool queue at stash completion. Named it `tiered_writes_cooling_bypass`.

shahyash2609:
Thanks for the review @romange @dranikpg — pushed a rework:

1. Renamed the flag to `--tiered_writes_cooling_bypass`.
2. Removed the `CompactObj` state bit; the intent now rides on `StashDescriptor` and is carried across the async stash gap in a per-shard set (details in the inline thread).

On @dranikpg's broader point — fully agree cooling conflicts with cache semantics. Worth noting this flag targets only the **write path** (proactive offload on edit), where the value hasn't been read at all, so there's even less reason to cool it than a scan-offloaded value. Background/scan offload (`RunOffloading`) is deliberately left on the normal cooling path.

Your second point — read-hot keys that arrived via writes getting stuck in the cool queue as first-to-evict — is the complementary **eviction-side** problem this PR doesn't touch. A SIEVE/second-chance pass on cool eviction (`PopCool`), reusing the existing `touched` bit, would protect read-hot cooled values regardless of how they entered (and would make this insertion-side flag less necessary). Happy to follow up with that as a separate PR if you think it's the right direction.

src/server/tiered_storage.cc
```diff
@@ -523,6 +551,11 @@ void TieredStorage::StashPrimeValue(DbIndex dbid, string_view key, const StashDe
   tiering::OpManager::PendingId id;
   error_code ec;
 
+  // Carry the cooling-bypass intent per-key across the async stash gap. Done per-key (not per-bin)
+  // so small-bin batching keeps per-key granularity. Cleaned up in ClearStashPending on abort.
+  if (blobs.bypass_cooling)
+    op_manager_->MarkCoolingBypass(dbid, key);
+
```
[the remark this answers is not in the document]
shahyash2609: Good catch — fixed. `CancelStash(KeyRef)` now erases the `cooling_bypass_pending_` entry, so a stash cancelled on read/delete no longer leaves a stale bypass marker (which could have wrongly bypassed cooling on a later stash of the same key, or leaked). Covers both the own-page and small-bin sub-cases since both flow through that per-key path.

src/server/tiered_storage.cc
```diff
@@ -269,7 +285,13 @@ class TieredStorage::ShardOpManager : public tiering::OpManager {
       stats_.total_stashes++;
 
       StashDescriptor blobs{FragmentRef{*pv}.GetSerializationDescr()};
-      if (ts_->config_.experimental_cooling) {
+
+      // A write/overwrite driven stash bypasses the cooling pool (see
+      // --tiered_writes_cooling_bypass) and goes disk-only, so write bursts don't evict read-hot
+      // entries. The intent is carried in cooling_bypass_pending_ across the async stash gap.
+      bool bypass = cooling_bypass_pending_.erase(key) > 0;
```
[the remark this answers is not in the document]
shahyash2609: Fixed — added `cooling_bypass_pending_.erase(id)` in `CancelStash(KeyRef)` (the path `CancelPending`/`bins_->Delete` take), so cancelled/removed stashes no longer leave a stale entry or leak. The list-node `CancelStash` variant never carries bypass (only string writes set it), so it needs no change.

src/server/tiered_storage.cc
```diff
@@ -64,6 +65,12 @@ ABSL_FLAG(bool, tiered_experimental_hash_support, false, "Experimental hash data
 
 ABSL_FLAG(bool, tiered_experimental_list_support, false, "Experimental list node offloading");
 
+ABSL_FLAG(bool, tiered_writes_cooling_bypass, false,
```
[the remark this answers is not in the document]
shahyash2609: Fixed — updated both the PR title and description to `--tiered_writes_cooling_bypass`.

romange:
qodo review

src/server/tiered_storage.cc
```diff
@@ -323,6 +349,12 @@ class TieredStorage::ShardOpManager : public tiering::OpManager {
   // When we don't have memory to upload a page for defragmentation, we save it here to do it later
   std::deque<uint32_t /* page index (kPageSize) */> delayed_defrag_queue_;
 
+  // Keys whose in-flight stash should bypass the cooling pool (see --tiered_writes_cooling_bypass).
+  // Stored as owning DbKeyId with heterogeneous lookup, so we can probe with an OpManager::KeyRef
+  // (which holds a string_view) without allocating. Populated only when bypass is requested.
+  absl::flat_hash_set<tiering::DbKeyId, tiering::detail::Hasher, tiering::detail::Eq>
```
romange: that's not what I would like to see. 
I understand that my suggestion to just pass this bit through the call chain is problematic because the state must persist across async calls. Having said that I still want to check myself whether it can be done.

romange:
@dranikpg  I now pass the bypass flag via the call stack of tiering module. PTAL

shahyash2609:
@dranikpg gentle reminder on this one when you get a chance to review 🙏

shahyash2609:
@dranikpg gentle reminder on this one when you get a chance to review 🙏

shahyash2609:
@romange @dranikpg Can you please help by checking this?

romange:
@shahyash2609 I am sorry I wrote the patch, I can not review it. When @dranikpg  will be able to review he will do it. Thank you! 🙏🏼

dranikpg:
Hi @shahyash2609. Sorry for the reply delay 😞 , I was out of office for a week.

I still have trouble understanding what specifically happens with your usecase

> overwrite traffic streams into the cooling pool, pushing genuinely read-hot entries out of RAM

Two clarifications:
1. Genuinely read-hot entries _are not_ in the cooling queue, they are regular values. In fact, values in the cooling queue have never been read as they are promoted on the first read
2. The cooling queue has the lowest "memory priority" so it should (in theory) never cause regular values to be evicted - in all our caculations we prefer to remove values from it first

Picture this:

The cool queue was specifically designed for client writes, offload writes slipped in more or less unintentionally. If you disable insertion to this queue for client writes, it just acts as a "second" layer for offloading. A kind of slow down queue in exchange for memory. 

_What active function does it perform?_ I don't see any. So why not _disable_ it at all then? 🤔 

However, I might assume that you found problems connected with a recent bug I found just now (#7982), which caused cool queue entries after writes to stay there longer than intended and not make room for uploading entries, keeping the cache hit rate low. The solution to it is to empty the cool queue more agressively instead of avoiding to fill it.

So at the end I assume that you will see improvements with
1. Disabling experimental cooling at all if you never actually rely on it
2. Using #7982 

But I still fail to see what case this flag solves. It effectively neutralizes the cool queues purpose, but keeps it... enabled. If you observe that pure offloading works better with the cool queue, let's discuss this specifically

shahyash2609:
@dranikpg Thanks for the update.
Will try the suggested approach and will come back soon on this.

PR #7806, opened 2026-07-07
fix: bitops on tiered string values

vyavdoshenko:
Fix bit operations on string values that were offloaded to tiered storage. These commands could crash because they accessed the `CompactObj` string data directly, even when the value was external.

Changes:
- Add a shared helper to materialize string values from tiered storage before bitops access them.
- Update `GETBIT`, `SETBIT`, `BITCOUNT`, `BITPOS`, `BITOP`, `BITFIELD`, and `BITFIELD_RO` to handle external string values.
- Safely overwrite external destination values by deleting the old tiered segment before storing the new in-memory string.

### DIFF
--- a/src/server/bitops_family.cc
+++ b/src/server/bitops_family.cc
@@ -20,6 +20,7 @@
 #include "server/engine_shard_set.h"
 #include "server/error.h"
 #include "server/namespaces.h"
+#include "server/tiered_storage.h"
 #include "server/transaction.h"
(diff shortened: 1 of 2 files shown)


vyavdoshenko:
augment review

dranikpg (approved):
Fine from an implementation point of view. With strings, all tiering operations are asynchronous and use callbacks, the synchronous approach is much more difficult to parallelize and achieve good performance with

PR #7807, opened 2026-07-07
perf: shrink squash callback and avoid regex stack overflow

romange:
## Summary

Shrink the per-shard squash callback capture while preserving inline storage, and remove `std::regex` from Lua script flag parsing to avoid recursive stack usage on sanitizer fiber stacks.

## Changes

- `multi_command_squasher.cc`: move squash callback state into `CbCntx`, use `FiberBlockingCounter`, and keep the callback within the inline buffer.
- `script_mgr.cc`: replace script flag `std::regex_search` with a bounded string parser.
- `AGENTS.md`: document avoiding `std::regex` in fiber/server paths.
- `compact_object.cc`: handle `ptr == nullptr` in `LargeString::SetString`.
- `small_string.{cc,h}`: move `SmallString::kMaxSize` into the `.cc` anonymous namespace.
- `helio`: submodule bump.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -118,6 +118,7 @@ git log --oneline -10
 - Global mutable state
 - Edit without reading
 - Skip tests
+- Use `std::regex` in fiber/server paths (recursive implementation can overflow small fiber stacks)
 - Use `./tools/docker/build.sh` for local development (use `ninja` instead)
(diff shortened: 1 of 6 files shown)


romange:
qodo what's the wheather today?

dranikpg (approved):

PR #7810, opened 2026-07-07
Handle EPERM in io_uring support check

kant2002:
WSL return EPERM to indicate that iouring is not supported

### DIFF
--- a/src/server/dfly_main.cc
+++ b/src/server/dfly_main.cc
@@ -532,7 +532,7 @@ bool ShouldUseEpollAPI(const base::sys::KernelVersion& kver) {
 
   iouring_res = -iouring_res;
 
-  if (iouring_res == ENOSYS) {
+  if (iouring_res == ENOSYS || iouring_res == EPERM) {
     LOG(WARNING) << "iouring API is not supported. switching to epoll.";
   } else if (iouring_res == ENOMEM) {
     LOG(WARNING) << "io_uring does not have enough memory. That can happen when your "
--- a/src/server/dfly_main.cc
+++ b/src/server/dfly_main.cc
@@ -532,15 +532,17 @@ bool ShouldUseEpollAPI(const base::sys::KernelVersion& kver) {
 
   iouring_res = -iouring_res;
 
-  if (iouring_res == ENOSYS || iouring_res == EPERM) {
-    LOG(WARNING) << "iouring API is not supported. switching to epoll.";
+  if (iouring_res == ENOSYS) {
+    LOG(WARNING) << "iouring API is not supported. Switching to epoll.";
   } else if (iouring_res == ENOMEM) {
     LOG(WARNING) << "io_uring does not have enough memory. That can happen when your "
                     "max locked memory is too limited. If you run via docker, "
                     "try adding '--ulimit memlock=-1' to \"docker run\" command."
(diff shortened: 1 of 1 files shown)


kant2002:
Log of startup 

```
I0707 20:26:23.671198       7 dfly_main.cc:1096] Starting dragonfly df-v1.39.0-699862e5da7cb29bf5642a1da422c4c78cefae38
W0707 20:26:23.672029       7 dfly_main.cc:1144] SWAP is enabled. Consider disabling it when running Dragonfly.
I0707 20:26:23.672709       7 dfly_main.cc:1150] maxmemory has not been specified. Deciding myself....
I0707 20:26:23.672730       7 dfly_main.cc:1159] Found 7.29GiB available memory. Setting maxmemory to 5.83GiB
W0707 20:26:23.672875       7 dfly_main.cc:543] Weird error 1 switching to epoll
I0707 20:26:23.677993       7 proactor_pool.cc:149] Running 4 io threads
I0707 20:26:23.680526       7 dragonfly_listener.cc:152] SSL version: OpenSSL 3.4.5 7 Apr 2026
I0707 20:26:23.684226       7 server_family.cc:1221] Host OS: Linux 6.18.35.2-microsoft-standard-WSL2 x86_64 with 4 threads
```

kant2002:
I can add message like this `Check if io_uring is disabled via /proc/sys/kernel/io_uring_disabled` specifically for EPERM error path.

src/server/dfly_main.cc
```diff
@@ -532,7 +532,7 @@ bool ShouldUseEpollAPI(const base::sys::KernelVersion& kver) {
 
   iouring_res = -iouring_res;
 
-  if (iouring_res == ENOSYS) {
+  if (iouring_res == ENOSYS || iouring_res == EPERM) {
     LOG(WARNING) << "iouring API is not supported. switching to epoll.";
```
romange: use `strerror(iouring_res)`  to pretty pring the actual error in the warning

src/server/dfly_main.cc
```diff
@@ -532,7 +532,7 @@ bool ShouldUseEpollAPI(const base::sys::KernelVersion& kver) {
 
   iouring_res = -iouring_res;
 
-  if (iouring_res == ENOSYS) {
+  if (iouring_res == ENOSYS || iouring_res == EPERM) {
```
vyavdoshenko: @kant2002 
Thanks for your contribution, but this fix is incorrect. I suppose it should be a separate `else if` branch because it may be disabled by system policy.  So I suggest adding something like this:
```
} else if (iouring_res == EPERM) {
  LOG(WARNING) << "Detailed error description.";
```

vyavdoshenko (changes requested):

src/server/dfly_main.cc
kant2002: Done

src/server/dfly_main.cc
kant2002: Make sense.

vyavdoshenko:
@kant2002 
Please follow the instructions from https://github.com/dragonflydb/dragonfly/blob/main/CONTRIBUTING.md
You have to fix the clang-format issue to pass the CI checks.

vyavdoshenko (approved):

PR #7813, opened 2026-07-08
refactor: use ADL for CmdArgParser

BorysTheDev:
Summary: This PR refactors facade::CmdArgParser to support ADL-based parsing for custom types and to standardize parser-function usage via compile-time callables.

Changes:

- Added ArgParsable + ADL hook (ParseArg(CmdArgParser*, T*)) so Next<T>() can parse user-defined types.
- Reworked parser-function support to use parser.Next<Fn>() (NTTP) and removed the runtime Next(fn)/NextOrDefault(fn, ...) APIs.
- Introduced reusable helpers: Upper (token parser), MinMax (two-value pair with ordering validation), and Flag (bit-OR option).
- Enhanced OneOf with .Repeat() to allow idempotent repeats while still rejecting conflicts.
- Updated multiple command families (ACL/Cluster/Debug/Generic/Geo/Main/Script/Server/String/ZSet) to use the new parsing primitives.
- Adjusted DEBUG POPULATE expiry range handling and TTL selection for equal endpoints.
- Expanded unit tests to cover ADL custom types, Upper, MinMax, Flag, and OneOf().Repeat() behavior.

Technical Notes: Error semantics remain centralized through CmdArgParser::ErrorInfo::MakeReply(); most new parsing paths still surface invalid input as syntax error unless a custom message is provided.

PR #7825, opened 2026-07-09
Revert "refactor: use ADL for CmdArgParser (#7813)"

romange:
This reverts commit b47ca6af06b5d8b88f342edb269fe825a62c444a.


kostasrim (approved):
This can happen on any commit by any team member without `observability` and we will never know. We should have guards against regressions on main. Even a simple memtier on every commit can actually offer some insights. As far as I remember Vlad was doing something around this ?

PR #7827, opened 2026-07-09
chore: enforce fuzzer coverage for newly registered commands

kostasrim:
Until now, keeping the RESP fuzzer's command table in sync with newly added commands relied on reviewers remembering to ask for it. This adds a CI check that makes the sync mandatory.

Changes
- fuzz/check_fuzz_coverage.py: extracts all CI{"..."} registrations from src/server and verifies each command is either in resp_mutator.COMMANDS or explicitly excluded with a justification (admin/replication/cluster/memcache-only). Also flags stale mutator entries and dead exclusions.
- .github/workflows/ci.yml: runs the check in the pre-commit job (pure Python, sub-second, no build required).

A PR adding a new command now fails CI with instructions to add the command to the mutator (plus seed + dict entry) or exclude it consciously.


kostasrim:
augment review

kostasrim:
augment review

fuzz/check_fuzz_coverage.py
```diff
@@ -0,0 +1,200 @@
...
+        for name, arity in CI_RE.findall(path.read_text(encoding="utf-8", errors="replace")):
+            arities.setdefault(name, int(arity))
+    return arities
+
+
+def mutator_commands():
+    sys.path.insert(0, str(FUZZ_DIR))
+    import resp_mutator
+
+    return {name.decode(): (lo, hi) for name, lo, hi in resp_mutator.COMMANDS}
```
vyavdoshenko, line 87: This is a nice idea.

fuzz/check_fuzz_coverage.py
```diff
@@ -0,0 +1,200 @@
...
+    "BF.INFO",
+    "RM",
+    "ROLE",
+    "LASTSAVE",
+}
+
+# Matches `CI{"NAME", <mask expr, no top-level commas>, <arity>, ...}`. The mask
+# expression is a `CO::FOO | CO::BAR`-style OR chain, so it never contains a comma,
+# which lets `[^,]+` skip over it to reach the arity field.
+CI_RE = re.compile(r'CI\{"([A-Z0-9._]+)",\s*[^,]+,\s*(-?\d+),')
```
vyavdoshenko, line 71: Does it support multiline registration?

vyavdoshenko (approved):
I really like this change!

fuzz/check_fuzz_coverage.py
kostasrim: yes it should!

PR #7828, opened 2026-07-10
feat: add OAHMap

BorysTheDev:
Summary: This PR adds an open-addressing hash map container (OAHMap) on top of Dragonfly’s OAH (open-addressing hash) infrastructure, and refactors shared OAH logic into a reusable base.

Changes:

- Introduced OAHPair (key+value entry with expiry + cached fingerprint) and OwnedOAHPair RAII owner for extracted/exchanged entries.
- Added OAHTable<Entry> template to centralize shared probing, lookup, erase, scan, rehash, iteration, and expiry collection logic.
- Refactored OAHSet to inherit from OAHTable<OAHEntry> and keep only the set-specific insertion hot path (AddImpl).
- Templatized OAHPtr so buckets/vectors can manage either OAHEntry or OAHPair.
- Implemented OAHMap with add/update/skip/exchange semantics, TTL/KEEPTTL handling, extraction, and unique random-pair sampling.
- Added oah_map_test (unit tests + microbenchmarks) and updated core CMake targets to build oah_pair.cc and the new test.

Technical Notes: The map and set now share the same SIMD probe window and extension-vector collision strategy; TTL expiry is handled lazily during iteration/find/erase and via SizeSlow() when requested.


BorysTheDev:
augment review

BorysTheDev:
augment review

src/core/oah_map_test.cc
```diff
@@ -0,0 +1,543 @@
+// Copyright 2024, DragonflyDB authors.  All rights reserved.
```
vyavdoshenko, line 1: nit: I think you can use 2026 year for new files

src/core/oah_set.h
```diff
@@ -260,272 +53,69 @@ class OAHSet {  // Open Addressing Hash Set
...
-  // Frees the blob/vector of every non-empty bucket. Used by ~OAHSet and Clear().
-  void FreeAllSlots() {
-    for (size_t i = 0, n = entries_.size(); i < n; ++i) {
-      if (entries_[i])
-        At(i).Clear();
+  bool AddImpl(std::string_view str, uint32_t ttl_sec) {
+    if (size_ >= entries_.size()) [[unlikely]] {
+      Reserve(BucketCount() * 2);
+    }
+    assert(Capacity() >= kDisplacementSize);
```
vyavdoshenko, line 61: Is it intentional? Is assert better than the project's macro checks?

vyavdoshenko (approved):

src/core/oah_set.h
BorysTheDev: we shouldn't use DCHECK in headers

src/core/oah_map_test.cc
BorysTheDev: I will update it in next PR

PR #7834, opened 2026-07-10
feat(tiering): Prefetch neighboring tiered list nodes on load

mkaruza:
Add list_tiering_prefetch_depth flag to asynchronously load offloaded list nodes near the
one being accessed, so sequential list access can find them already resident once needed. 
Also disambiguate io_pending into IsStashPending()/IsLoadPending() on QList::Node and
add CancelLoad() to abort an in-flight prefetch when its node is deleted.

### DIFF
--- a/src/server/tiering/op_manager.cc
+++ b/src/server/tiering/op_manager.cc
@@ -84,6 +84,9 @@ void OpManager::CancelPendingLoad(DiskSegment segment) {
   DCHECK(it != pending_reads_.end());
   auto* ops = it->second.Find(segment);
   DCHECK(ops);
+  for (auto& read_cb : ops->read_cbs) {
+    read_cb(nonstd::make_unexpected(make_error_code(errc::operation_canceled)));
+  }
   ops->read_cbs.clear();
 }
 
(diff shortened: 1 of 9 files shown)


src/server/tiering/op_manager.cc
```diff
@@ -79,6 +79,14 @@ void OpManager::CancelPending(PendingId id) {
   pending_stash_ver_.erase(ToOwned(id));
 }
 
+void OpManager::CancelPendingLoad(DiskSegment segment) {
+  auto it = pending_reads_.find(segment.ContainingPages().offset);
+  DCHECK(it != pending_reads_.end());
+  auto* ops = it->second.Find(segment);
+  DCHECK(ops);
+  ops->read_cbs.clear();
```
dranikpg, line 90: This is always tricky because what if the read_cbs is larger than one? You assume that with your sequential model there are no conflicts, so you can delete a node only after all reads finished and it is preloading at most?

My question generally is: what if read_cbs is larger than one and you never resolve some callbacks

dranikpg (dismissed):

src/server/tiering/op_manager.cc
mkaruza: Good catch. I think that we should notify all readers before clearing read_cbs vector.

dranikpg (approved):

PR #7839, opened 2026-07-11
feat(search): add vector data-type foundation (types, kernels, validation)

vyavdoshenko:
Introduces the type infrastructure for native-width vector data types in the search module, without changing storage or query behavior yet. Existing FLOAT32 indexes are unaffected; follow-up PRs will wire this into FLAT/HNSW/JSON storage.

Changes:
- Add `VectorDataType` enum {FLOAT32, FLOAT64, FLOAT16, BFLOAT16, INT8, UINT8} + `ElementSize()`; `VectorParams::data_type` becomes this enum (was a free-form string).
- Add typed distance kernels: each element is widened to `float` (`double` accumulator for FLOAT64) before L2/IP/COSINE; scalar `HalfToFloat`/`Bf16ToFloat` decoders, no SIMD dep.
- Byte-safe (`memcpy`) element reads, so unaligned native-width blobs are not an alignment/UB risk on any target.
- Validate `TYPE` on `FT.CREATE` (reject unknown names). Coerce unknown -> FLOAT32 only during RDB/replica load (`LOADING` state), so legacy snapshots with a free-form TYPE still load instead of losing the index.
- `FT.INFO` and index restore serialize the canonical dtype token via `VectorDataTypeToString`.

Tests:
- `VectorDistanceTypedDtypes` - every dtype x {L2, IP, COSINE} matches the FLOAT32 reference.
- `VectorUnknownTypeRejected` - `FT.CREATE ... TYPE <bogus>` returns an error.

Notes:
- No functional change for non-float dtypes yet: they parse/validate but are not stored until the FLAT-storage PR. FLOAT32 behavior is unchanged.
- Part 1 of 4 (**foundation** -> FLAT -> HNSW -> JSON).

### DIFF
--- a/src/core/search/base.h
+++ b/src/core/search/base.h
@@ -40,6 +40,24 @@ inline std::pair<ShardId, DocId> DecomposeGlobalDocId(GlobalDocId id) {
 
 enum class VectorSimilarity { L2, IP, COSINE };
 
+enum class VectorDataType { FLOAT32, FLOAT64, FLOAT16, BFLOAT16, INT8, UINT8 };
+
+constexpr size_t ElementSize(VectorDataType dt) {
+  switch (dt) {
+    case VectorDataType::FLOAT32:
+      return 4;
+    case VectorDataType::FLOAT64:
+      return 8;
+    case VectorDataType::FLOAT16:
+    case VectorDataType::BFLOAT16:
+      return 2;
+    case VectorDataType::INT8:
+    case VectorDataType::UINT8:
+      return 1;
+  }
+  return 4;
+}
+
 using OwnedFtVector = std::pair<std::unique_ptr<float[]>, size_t /* dimension (size) */>;
 using BorrowedFtVector = const char*;
 
--- a/src/core/search/vector_utils.cc
+++ b/src/core/search/vector_utils.cc
@@ -4,6 +4,7 @@
 
 #include "core/search/vector_utils.h"
 
+#include <bit>
 #include <cmath>
 #include <cstdint>
 #include <cstring>
@@ -123,12 +124,11 @@ float HalfToFloat(uint16_t h) {
     if (mant == 0) {
       bits = sign;  // +/- zero
     } else {
-      exp = 127 - 15 + 1;  // normalize the subnormal
-      while ((mant & 0x400) == 0) {
-        mant <<= 1;
-        exp--;
-      }
-      mant &= 0x3FF;
+      // Normalize the subnormal: shift the mantissa left until bit 10 is set. `mant` is in
+      // [1, 0x3FF], so countl_zero is in [22, 31] and the shift in [1, 10].
+      int shift = std::countl_zero(mant) - 21;
+      exp = (127 - 15 + 1) - shift;
+      mant = (mant << shift) & 0x3FF;
       bits = sign | (exp << 23) | (mant << 13);
     }
   } else if (exp == 0x1F) {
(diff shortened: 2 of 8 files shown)


src/core/search/base.h
```diff
@@ -40,6 +40,24 @@ inline std::pair<ShardId, DocId> DecomposeGlobalDocId(GlobalDocId id) {
...
+    case VectorDataType::FLOAT64:
+      return 8;
+    case VectorDataType::FLOAT16:
+    case VectorDataType::BFLOAT16:
+      return 2;
+    case VectorDataType::INT8:
+    case VectorDataType::UINT8:
+      return 1;
+  }
+  return 4;
```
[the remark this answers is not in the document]
vyavdoshenko: It will be fixed in follow-up PRs.

src/core/search/vector_utils.h
```diff
@@ -22,8 +22,22 @@ float IPDistance(const float* u, const float* v, size_t dims);
 float CosineDistance(const float* u, const float* v, size_t dims);
 float VectorDistance(const float* u, const float* v, size_t dims, VectorSimilarity sim);
 
+// Distance between two native-width vector blobs of the given dtype. Elements are widened to
+// float (double for FLOAT64) before the metric is applied, matching the reference engine.
+float VectorDistance(const void* u, const void* v, size_t dims, VectorSimilarity sim,
+                     VectorDataType dt);
+
+// Widen a single half-precision element to float. Byte-safe for unaligned input.
```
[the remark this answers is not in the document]
vyavdoshenko: fixed

src/core/search/vector_utils.cc
```diff
@@ -124,6 +288,41 @@ std::string_view VectorSimilarityToString(VectorSimilarity sim) {
...
+    case VectorDataType::INT8:
+      return "INT8";
+    case VectorDataType::UINT8:
+      return "UINT8";
+  }
+  DCHECK(false) << "Unhandled VectorDataType enum value: " << static_cast<int>(dt);
+  return "FLOAT32";
+}
+
+std::optional<VectorDataType> ParseVectorDataType(std::string_view name) {
```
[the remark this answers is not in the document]
vyavdoshenko: fixed

src/core/search/vector_utils.cc
```diff
@@ -111,6 +114,167 @@ float VectorDistance(const float* u, const float* v, size_t dims, VectorSimilari
...
+      bits = sign | (exp << 23) | (mant << 13);
+    }
+  } else if (exp == 0x1F) {
+    bits = sign | 0x7F800000u | (mant << 13);  // inf / nan
+  } else {
+    bits = sign | ((exp + 112) << 23) | (mant << 13);  // rebias 127 - 15 = 112
+  }
+  float out;
+  memcpy(&out, &bits, sizeof(out));
+  return out;
```
BorysTheDev, line 141: can we use bitmask, it should be shorter and more clear

src/core/search/vector_utils.cc
```diff
@@ -111,6 +114,167 @@ float VectorDistance(const float* u, const float* v, size_t dims, VectorSimilari
...
+}
+
+float Bf16ToFloat(uint16_t b) {
+  uint32_t bits = static_cast<uint32_t>(b) << 16;
+  float out;
+  memcpy(&out, &bits, sizeof(out));
```
BorysTheDev, line 147: can we do 
```

  float out = 0;
  memcpy(&out, &b, sizeof(b));
```

src/core/search/vector_utils.cc
```diff
@@ -111,6 +114,167 @@ float VectorDistance(const float* u, const float* v, size_t dims, VectorSimilari
...
+namespace {
+
+uint16_t LoadU16(const void* base, size_t i) {
+  uint16_t v;
+  memcpy(&v, static_cast<const char*>(base) + i * sizeof(uint16_t), sizeof(v));
+  return v;
+}
+
+// Byte-safe reads: native-width blobs may be unaligned (borrowed keyspace storage), so a
+// typed deref would be UB. memcpy lowers to a single load on the target ISAs.
```
BorysTheDev: why do you think it can be unalligned

src/core/search/vector_utils.cc
```diff
@@ -111,6 +114,167 @@ float VectorDistance(const float* u, const float* v, size_t dims, VectorSimilari
...
+  float v;
+  memcpy(&v, static_cast<const char*>(base) + i * sizeof(float), sizeof(v));
+  return v;
+}
+
+double LoadF64(const void* base, size_t i) {
+  double v;
+  memcpy(&v, static_cast<const char*>(base) + i * sizeof(double), sizeof(v));
+  return v;
+}
```
BorysTheDev: it feels like template needed

src/core/search/vector_utils.cc
```diff
@@ -111,6 +114,167 @@ float VectorDistance(const float* u, const float* v, size_t dims, VectorSimilari
...
+  static float Get(const void* p, size_t i) {
+    return static_cast<const int8_t*>(p)[i];
+  }
+};
+template <> struct Reader<VectorDataType::UINT8> {
+  using Acc = float;
+  static float Get(const void* p, size_t i) {
+    return static_cast<const uint8_t*>(p)[i];
+  }
+};
```
BorysTheDev: consider create types for UINT8, BFLOAT16 etc. in this case you can write cleaner code without specialization

src/core/search/vector_utils.cc
```diff
@@ -111,6 +114,167 @@ float VectorDistance(const float* u, const float* v, size_t dims, VectorSimilari
...
+      return DistByMetric<VectorDataType::FLOAT16>(u, v, dims, sim);
+    case VectorDataType::BFLOAT16:
+      return DistByMetric<VectorDataType::BFLOAT16>(u, v, dims, sim);
+    case VectorDataType::INT8:
+      return DistByMetric<VectorDataType::INT8>(u, v, dims, sim);
+    case VectorDataType::UINT8:
+      return DistByMetric<VectorDataType::UINT8>(u, v, dims, sim);
+  }
+  return 0.0f;
+}
```
BorysTheDev, line 249: I think if we have types instead of enum this function should be deleted

src/server/search/search_family.cc
```diff
@@ -121,7 +122,13 @@ search::SchemaField::VectorParams ParseVectorParams(CmdArgParser* parser) {
     } else if (parser->Check("M", &params.hnsw_m)) {
     } else if (parser->Check("EF_CONSTRUCTION", &params.hnsw_ef_construction)) {
     } else if (parser->Check("TYPE")) {
-      params.data_type = absl::AsciiStrToUpper(parser->Next<string_view>());
+      std::string type_name = absl::AsciiStrToUpper(parser->Next<string_view>());
+      if (auto dt = search::ParseVectorDataType(type_name); dt)
```
BorysTheDev, line 126: why don't you use parser->MapNext

BorysTheDev (dismissed):

src/core/search/vector_utils.cc
vyavdoshenko: bfloat16 is the high 16 bits of a float32. memcpy(&out, &b, sizeof(b)) copies the two bytes into the low half of out on little-endian, so the result would be wrong. The <<16 places the bits in the high half and is endian-agnostic, which is why it's needed here.

src/core/search/vector_utils.cc
vyavdoshenko: These readers consume borrowed keyspace storage in the HNSW path (later PR in this series), where the vector data starts at an sds payload offset (header size 3/5/9/17 bytes) that is never 4-/8-aligned. A typed deref there is alignment UB; memcpy avoids it and lowers to a single load on our target ISAs, so there's no cost.

src/server/search/search_family.cc
vyavdoshenko: I looked at that: MapNext errors on an unknown token, and TryMapNext doesn't consume the token when nothing matches. Neither fits what's needed here - consume the TYPE token, then either map it, coerce unknown -> FLOAT32 during RDB/replica load (LOADING, for backward-compat with legacy snapshots), or reject it. The explicit parse handles all three. If we drop the legacy-coerce path, MapNext would be the natural fit.

src/core/search/vector_utils.cc
vyavdoshenko: fixed

src/core/search/vector_utils.cc
vyavdoshenko: fixed

src/core/search/vector_utils.cc
vyavdoshenko: fixed

src/core/search/vector_utils.cc
vyavdoshenko: The enum stays because it's needed for further changes

BorysTheDev (approved):

PR #7844, opened 2026-07-12
feat(server): include TLS certificate info in INFO output

### LINKED ISSUE #3292: Include TLS information into INFO command output

akhilp19:
Adds TLS certificate metadata to the `INFO SERVER` section.

When TLS is enabled and the certificate file can be parsed successfully, the `INFO` output now includes:

- `tls_cert_subject` — certificate subject DN
- `tls_cert_issuer` — certificate issuer DN
- `tls_cert_not_before` — validity start date
- `tls_cert_not_after` — validity end date

The fields are emitted only when:
- Dragonfly is built with SSL support (`DFLY_USE_SSL`)
- TLS is enabled at runtime (`--tls`)
- The certificate file configured via `--tls_cert_file` can be parsed

If the certificate file is missing or invalid, the fields are silently omitted.

Fixes #3292

### DIFF
--- a/src/facade/dragonfly_connection.h
+++ b/src/facade/dragonfly_connection.h
@@ -15,6 +15,7 @@
 #include <variant>
 
 #include "facade/connection_ref.h"
+#include "facade/dragonfly_listener.h"
 #include "facade/facade_types.h"
 #include "facade/parsed_command.h"
 #include "io/io_buf.h"
@@ -264,6 +265,10 @@ class Connection : public util::Connection {
   size_t GetMemoryUsage() const;
 
   ConnectionContext* cntx();
+
+  // Returns the TLS certificate metadata for this connection's listener, or nullptr if the listener
+  // has no TLS configured or the certificate could not be parsed.
+  std::shared_ptr<const TlsCertInfo> GetTlsCertInfo() const;
 
   // For non replication connections refresh memory usage field as well as update tl stats if conn.
   // is still live.
--- a/src/facade/dragonfly_listener.h
+++ b/src/facade/dragonfly_listener.h
@@ -13,8 +13,10 @@
 #include <vector>
 
 #include "facade/facade_types.h"
+#include "facade/tls_helpers.h"
 #include "util/fiber_socket_base.h"
 #include "util/fibers/proactor_base.h"
+#include "util/fibers/synchronization.h"
 #include "util/http/http_handler.h"
 #include "util/listener_interface.h"
 
@@ -42,6 +44,12 @@ class Listener : public util::ListenerInterface {
 
   // ReconfigureTLS MUST be called from the same proactor as the listener.
   bool ReconfigureTLS();
+
+#ifdef DFLY_USE_SSL
+  // Returns the TLS certificate metadata currently used by this listener.
+  // nullptr if TLS is not configured or the certificate could not be parsed.
+  std::shared_ptr<const TlsCertInfo> GetTlsCertInfo() const;
+#endif
 
   // Returns thread-local dynamic memory usage by TLS.
   static size_t TLSUsedMemoryThreadLocal();
@@ -78,6 +86,10 @@ class Listener : public util::ListenerInterface {
 
   Protocol protocol_;
   SSL_CTX* ctx_ = nullptr;
+  // Updated in ReconfigureTLS (listener proactor) and read from any proactor
+  // via INFO, so all accesses are guarded by tls_cert_info_mu_.
+  mutable util::fb2::Mutex tls_cert_info_mu_;
+  std::shared_ptr<const TlsCertInfo> tls_cert_info_ ABSL_GUARDED_BY(tls_cert_info_mu_);
 };
 
 // Dispatch tracker allows tracking the dispatch state of connections and blocking until all
--- a/src/facade/dragonfly_listener.h
+++ b/src/facade/dragonfly_listener.h
@@ -45,11 +45,9 @@ class Listener : public util::ListenerInterface {
   // ReconfigureTLS MUST be called from the same proactor as the listener.
   bool ReconfigureTLS();
 
-#ifdef DFLY_USE_SSL
   // Returns the TLS certificate metadata currently used by this listener.
   // nullptr if TLS is not configured or the certificate could not be parsed.
   std::shared_ptr<const TlsCertInfo> GetTlsCertInfo() const;
-#endif
 
   // Returns thread-local dynamic memory usage by TLS.
   static size_t TLSUsedMemoryThreadLocal();
--- a/src/facade/tls_helpers.cc
+++ b/src/facade/tls_helpers.cc
@@ -6,11 +6,15 @@
 #include <openssl/err.h>
 
 #ifdef DFLY_USE_SSL
+#include <openssl/pem.h>
 #include <openssl/ssl.h>
+#include <openssl/x509.h>
 #endif
 
+#include <absl/cleanup/cleanup.h>
 #include <absl/functional/bind_front.h>
 
+#include <optional>
 #include <string>
 
 #include "base/flags.h"
@@ -34,7 +38,43 @@ ABSL_FLAG(size_t, tls_session_cache_timeout, 300, "Timeout for each session/tick
 
 namespace facade {
 
+using absl::GetFlag;
+
 #ifdef DFLY_USE_SSL
+
+namespace {
+
+std::string Asn1TimeToString(const ASN1_TIME* asn1_time) {
+  if (!asn1_time)
+    return {};
+
+  BIO* bio = BIO_new(BIO_s_mem());
+  if (!bio)
+    return {};
+  absl::Cleanup bio_cleanup = [bio] { BIO_free(bio); };
+
+  ASN1_TIME_print(bio, asn1_time);
+
+  BUF_MEM* buf_mem = nullptr;
+  BIO_get_mem_ptr(bio, &buf_mem);
+  std::string result(buf_mem ? buf_mem->data : "", buf_mem ? buf_mem->length : 0);
+  return result;
+}
+
+std::string X509NameToString(X509_NAME* name) {
+  if (!name)
+    return {};
+
+  char* buf = X509_NAME_oneline(name, nullptr, 0);
+  if (!buf)
+    return {};
+  absl::Cleanup buf_cleanup = [buf] { OPENSSL_free(buf); };
+
+  std::string result(buf);
+  return result;
+}
+
+}  // namespace
 
 // Creates the TLS context. Returns nullptr if the TLS configuration is invalid.
 // To connect: openssl s_client -state -crlf -connect 127.0.0.1:6380
@@ -62,8 +102,8 @@ SSL_CTX* CreateSslCntx(TlsContextRole role) {
   const auto& tls_cert_file = GetFlag(FLAGS_tls_cert_file);
 
   if (!tls_cert_file.empty()) {
-    // TO connect with redis-cli you need both tls-key-file and tls-cert-file
-    // loaded. Use `redis-cli --tls -p 6380 --insecure  PING` to test
+    // To connect with redis-cli you need both tls-key-file and tls-cert-file
+    // loaded. Use `redis-cli --tls -p 6380 --insecure PING` to test
     if (SSL_CTX_use_certificate_chain_file(ctx, tls_cert_file.c_str()) != 1) {
       LOG(ERROR) << "Failed to load TLS certificate";
       return nullptr;
@@ -131,6 +171,19 @@ void PrintSSLError() {
         return 1;
       },
       nullptr);
+}
+
+std::optional<TlsCertInfo> ParseTlsCertInfo(const X509* cert) {
+  if (!cert)
+    return std::nullopt;
+
+  TlsCertInfo info;
+  info.subject = X509NameToString(X509_get_subject_name(cert));
+  info.issuer = X509NameToString(X509_get_issuer_name(cert));
+  info.not_before = Asn1TimeToString(X509_get_notBefore(cert));
+  info.not_after = Asn1TimeToString(X509_get_notAfter(cert));
+
+  return info;
 }
 
 #endif
--- a/src/server/server_family.cc
+++ b/src/server/server_family.cc
@@ -47,6 +47,7 @@ extern "C" {
 #include "facade/dragonfly_connection.h"
 #include "facade/dragonfly_listener.h"
 #include "facade/reply_builder.h"
+#include "facade/tls_helpers.h"
 #include "io/file_util.h"
 #include "io/proc_reader.h"
 #include "search/doc_index.h"
@@ -2581,8 +2582,9 @@ Metrics ServerFamily::GetMetrics(Namespace* ns, const MetricsCollectOpts& opts) 
   return result;
 }
 
-string ServerFamily::FormatInfoMetrics(const Metrics& m, std::string_view section,
-                                       bool priveleged) const {
+string ServerFamily::FormatInfoMetrics(
+    const Metrics& m, std::string_view section, bool priveleged,
+    const std::shared_ptr<const facade::TlsCertInfo>& cert_info) const {
   string info;
   DbStats total;
 
@@ -2637,6 +2639,15 @@ string ServerFamily::FormatInfoMetrics(const Metrics& m, std::string_view sectio
     }
     append("multiplexing_api", multiplex_api);
     append("tcp_port", GetFlag(FLAGS_port));
+
+#ifdef DFLY_USE_SSL
+    if (cert_info) {
+      append("tls_cert_subject", cert_info->subject);
+      append("tls_cert_issuer", cert_info->issuer);
+      append("tls_cert_not_before", cert_info->not_before);
+      append("tls_cert_not_after", cert_info->not_after);
+    }
+#endif
 
     // Add availability_zone if it's not empty
     const auto& az = GetFlag(FLAGS_availability_zone);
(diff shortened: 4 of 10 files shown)


vyavdoshenko:
@akhilp19 Thank you for your contribution.

`g_tls_cert_info` as a single mutable global has two problems - both go away if we drop it.

1. The admin listener deterministically wipes the cert info. Every listener writes the same global (last-writer-wins), and `ReconfigureTLS`'s `else` branch calls `ClearTlsCertInfo()`. With `--no_tls_on_admin_port=true`, the PRIVILEGED listener takes that branch and clears the info that `main` just set. Listeners are written in construction order `[main, uds, admin, memcache]`, so when `admin` is the last writer (i.e. no `--memcached_port`), the info is wiped even though TLS is fully active on the main port:

    --tls --tls_key_file=key.pem --tls_cert_file=cert.pem \
    --admin_port=16380 --no_tls_on_admin_port=true   # no --memcached_port

`INFO SERVER` shows no `tls_cert_*` fields. Adding `--memcached_port` accidentally masks it.

2. (Possible solution). The mutex isn't needed. `INFO` is a cold path - instead of caching in a global guarded by a lock, just read the cert on demand in the INFO handler (from the SSL_CTX via `SSL_CTX_get0_certificate`, or by parsing `--tls_cert_file`). That removes the global, the `Set/Clear/Get` plumbing, and the mutex entirely - and fixes (1) as a side effect. (As-is the lock is also `std::mutex`, which shouldn't be held on a proactor thread.)

vyavdoshenko:
@akhilp19 
Please follow the instructions and sing your commit. https://github.com/dragonflydb/dragonfly/blob/main/CONTRIBUTING.md

akhilp19:
Hi @vyavdoshenko,

Thank you for the detailed review.

I have reverted the global caching approach entirely. The latest commit reads the TLS certificate metadata on demand directly from `--tls_cert_file` inside the `INFO` handler. This removes:

- The mutable global `g_tls_cert_info`
- The `std::mutex`
- The `SetTlsCertInfo()` / `ClearTlsCertInfo()` plumbing
- The last-writer-wins issue across listeners

Changes:

- `src/facade/tls_helpers.{h,cc}`: only `GetTlsCertInfo()` remains; it parses the cert file on demand.
- `src/server/server_family.cc`: unchanged — still emits the four fields when `FLAGS_tls` is true and metadata is available.
- `src/facade/dragonfly_listener.cc`: reverted the `ClearTlsCertInfo()` call.
- `tests/dragonfly/connection_test.py`: unchanged — `test_tls_info` continues to pass.

Local verification:
- `ninja dragonfly` ✅
- `pytest connection_test.py::test_tls_info` ✅
- Full `connection_test.py` has 2 pre-existing failures (`test_client_pause_v2_*`) unrelated to this change.

I have also corrected the commit author and sign-off email to `akhilrpawar@gmail.com` and re-signed the commit with the GPG key.

Please let me know if anything else needs adjustment.

vyavdoshenko:
@akhilp19 
Your commit is still `Unverified`. Please revise your signing procedure.

The error message is `The email in this signature doesn’t match the committer email.`

akhilp19:
Hi @vyavdoshenko,

Fixed the signature verification issue.

vyavdoshenko:
@akhilp19 
Please follow the instructions and fix the pre-commit step.
https://github.com/dragonflydb/dragonfly/blob/main/CONTRIBUTING.md

src/facade/tls_helpers.cc
```diff
@@ -133,5 +172,29 @@ void PrintSSLError() {
...
+  const auto& tls_cert_file = GetFlag(FLAGS_tls_cert_file);
+  if (tls_cert_file.empty())
+    return std::nullopt;
+
+  FILE* fp = fopen(tls_cert_file.c_str(), "r");
+  if (!fp)
+    return std::nullopt;
+
+  X509* cert = PEM_read_X509(fp, nullptr, nullptr, nullptr);
+  fclose(fp);
```
BorysTheDev: we shouldn't read from disk for every INFO cmd. please use thread_local shared_ptr variable to cache this data

akhilp19: I have updated GetTlsCertInfo() to cache the parsed TLS certificate metadata in a thread_local std::shared_ptr<const TlsCertInfo>. This avoids reading the certificate file from disk on every INFO request while keeping the data immutable and thread-local. The first call on a thread parses the cert file once; subsequent calls return the cached value.

akhilp19:
Hi @vyavdoshenko,

Thank you for pointing that out.

I have now installed pre-commit, clang-format, and black via pipx as described in CONTRIBUTING.md, and ran the hooks on all modified files. All pre-commit checks pass.

I also updated GetTlsCertInfo() to use a thread_local std::shared_ptr<const TlsCertInfo> cache, so the certificate file is not read from disk on every INFO request.

vyavdoshenko:
@akhilp19 
This change looks good, but there is an issue. INFO will return old TLS data after `ReconfigureTLS`.

src/facade/tls_helpers.cc
```diff
@@ -133,5 +176,34 @@ void PrintSSLError() {
...
+
+  TlsCertInfo info;
+  info.subject = X509NameToString(X509_get_subject_name(cert));
+  info.issuer = X509NameToString(X509_get_issuer_name(cert));
+  info.not_before = Asn1TimeToString(X509_get_notBefore(cert));
+  info.not_after = Asn1TimeToString(X509_get_notAfter(cert));
+
+  X509_free(cert);
+
+  tl_tls_cert_info = std::make_shared<const TlsCertInfo>(std::move(info));
```
BorysTheDev: It is a problem, you set it only for one thread. you should use DispatchBrief to set it for all threads

akhilp19: That thread-local cache has been removed. The implementation now uses a per-listener immutable std::shared_ptr<const TlsCertInfo> cache instead:

The cache lives inside facade::Listener as tls_cert_info_.
It is populated once when Listener::ReconfigureTLS() creates the new SSL_CTX, so it is tied to the active TLS context.
INFO reads from listener->GetTlsCertInfo() via the connection's listener reference.
This avoids the per-thread state problem entirely — no DispatchBrief is needed, and there is no global mutable cache or mutex on proactor threads.
This also ensures INFO never returns stale data after ReconfigureTLS and never reads the certificate file on every request.

akhilp19:
Hi @vyavdoshenko,

This has been addressed. The current implementation no longer uses a thread-local cache. Instead, Listener holds an immutable std::shared_ptr<const TlsCertInfo> that is repopulated inside Listener::ReconfigureTLS() when CreateSslCntx() succeeds. This means the cache is always tied to the active SSL_CTX, so INFO cannot return stale TLS data after ReconfigureTLS.

src/server/server_family.cc
```diff
@@ -2596,6 +2601,27 @@ string ServerFamily::FormatInfoMetrics(const Metrics& m, std::string_view sectio
...
+    } else if (GetFlag(FLAGS_tls)) {
+      // Fallback for test contexts where no listener is attached.
+      auto cert_info = facade::ParseTlsCertInfo();
+      if (cert_info) {
+        append("tls_cert_subject", cert_info->subject);
+        append("tls_cert_issuer", cert_info->issuer);
+        append("tls_cert_not_before", cert_info->not_before);
+        append("tls_cert_not_after", cert_info->not_after);
+      }
+    }
```
vyavdoshenko: We don't add test logic in production code.

tests/dragonfly/connection_test.py
```diff
@@ -1319,6 +1091,19 @@ async def test_tls_reject(
         await client.ping()
 
 
+async def test_tls_info(with_ca_tls_server_args, with_ca_tls_client_args, df_factory):
```
vyavdoshenko, line 1504: I suggest upgrading this test using a test certificate.

src/server/server_family.h
```diff
@@ -161,8 +153,8 @@ class ServerFamily {
   // Collects server metrics. See MetricsCollectOpts; a default-constructed value collects all.
   Metrics GetMetrics(Namespace* ns, const MetricsCollectOpts& opts) const;
 
-  std::string FormatInfoMetrics(const Metrics& metrics, std::string_view section,
-                                bool priveleged) const;
+  std::string FormatInfoMetrics(const Metrics& metrics, std::string_view section, bool priveleged,
+                                facade::Listener* listener = nullptr) const;
```
vyavdoshenko: We avoid default parameters.

src/facade/dragonfly_listener.h
```diff
@@ -78,6 +83,7 @@ class Listener : public util::ListenerInterface {
 
   Protocol protocol_;
   SSL_CTX* ctx_ = nullptr;
+  std::shared_ptr<const TlsCertInfo> tls_cert_info_;
```
vyavdoshenko: Unfortunately, this is not enough.
shared_ptr has separated data inside. tls_cert_info_ will be written in a proactor fiber on one thread, but also can be read on another thread. It can be race.

src/server/server_family.cc
akhilp19: Removed the test-context fallback from FormatInfoMetrics. ServerFamily::FormatInfoMetrics now only renders TLS cert fields when a listener is provided; test_utils.cc passes nullptr and simply omits those fields in that test path.

tests/dragonfly/connection_test.py
akhilp19: test_tls_info now reads the actual test certificate via openssl x509 and compares the extracted subject, issuer, notBefore, and notAfter against the raw INFO SERVER output.

src/server/server_family.h
akhilp19: Removed the default = nullptr from the listener parameter in FormatInfoMetrics. All callers now pass the listener explicitly.

src/facade/dragonfly_listener.h
akhilp19: std::atomic_load/std::atomic_store with a per-listener util::fb2::Mutex. ReconfigureTLS and GetTlsCertInfo both lock tls_cert_info_mu_, so cross-proactor reads/w are serialized.

src/server/server_family.cc
```diff
@@ -2596,6 +2597,18 @@ string ServerFamily::FormatInfoMetrics(const Metrics& m, std::string_view sectio
     append("multiplexing_api", multiplex_api);
     append("tcp_port", GetFlag(FLAGS_port));
 
+#ifdef DFLY_USE_SSL
+    if (listener) {
+      auto cert_info = listener->GetTlsCertInfo();
+      if (cert_info && !cert_info->subject.empty()) {
```
vyavdoshenko: `if (cert_info) {`

vyavdoshenko (commented):
lgtm

src/server/server_family.cc
akhilp19: I removed !cert_info->subject.empty() to keep the condition simple: if the listener has parsed cert info, we emit all fields.

src/facade/dragonfly_connection.h
```diff
@@ -265,6 +265,11 @@ class Connection : public util::Connection {
 
   ConnectionContext* cntx();
 
+  // Returns the listener that accepted this connection, or nullptr if unknown.
+  util::ListenerInterface* GetListener() const {
+    return listener();
+  }
```
vyavdoshenko: This is a way to expose a protected method to the public, but this is needed only for passing it into FormatInfoMetrics().
It leads to 3 static_casts. I suggest considering avoiding this.

src/facade/tls_helpers.cc
```diff
@@ -34,8 +37,44 @@ ABSL_FLAG(size_t, tls_session_cache_timeout, 300, "Timeout for each session/tick
...
+  BIO* bio = BIO_new(BIO_s_mem());
+  if (!bio)
+    return {};
+
+  ASN1_TIME_print(bio, asn1_time);
+
+  BUF_MEM* buf_mem = nullptr;
+  BIO_get_mem_ptr(bio, &buf_mem);
+  std::string result(buf_mem ? buf_mem->data : "", buf_mem ? buf_mem->length : 0);
+  BIO_free(bio);
```
BorysTheDev: nit: use absl::Cleanup to avoid leakage

src/facade/tls_helpers.cc
```diff
@@ -34,8 +37,44 @@ ABSL_FLAG(size_t, tls_session_cache_timeout, 300, "Timeout for each session/tick
...
+std::string X509NameToString(X509_NAME* name) {
+  if (!name)
+    return {};
+
+  char* buf = X509_NAME_oneline(name, nullptr, 0);
+  if (!buf)
+    return {};
+
+  std::string result(buf);
+  OPENSSL_free(buf);
```
BorysTheDev: nit: absl::Cleanup

src/facade/dragonfly_listener.h
```diff
@@ -78,6 +84,10 @@ class Listener : public util::ListenerInterface {
 
   Protocol protocol_;
   SSL_CTX* ctx_ = nullptr;
+  // Updated in ReconfigureTLS (listener proactor) and read from any proactor
+  // via INFO, so all accesses are guarded by tls_cert_info_mu_.
+  mutable util::fb2::Mutex tls_cert_info_mu_;
+  std::shared_ptr<const TlsCertInfo> tls_cert_info_;
```
BorysTheDev: if you want to use mutex, use ABSL_GUARDED_BY(tls_cert_info_mu_)

src/facade/dragonfly_connection.h
akhilp19: I have refactored it so that `FormatInfoMetrics()` takes `std::shared_ptr<const TlsCertInfo>` (or a raw pointer) instead of the listener, and the caller pulls the cert info from the listener. That removes the `GetListener()` accessor and the static_casts.

src/facade/tls_helpers.cc
akhilp19: The current `ParseTlsCertInfo(const X509*)` only reads from the borrowed `X509*` returned by `SSL_CTX_get0_certificate()`; it does not own the certificate or allocate any OpenSSL resources that need freeing, so `absl::Cleanup` is not needed here.

src/facade/dragonfly_listener.h
akhilp19: `tls_cert_info_` is now annotated with `ABSL_GUARDED_BY(tls_cert_info_mu_)`.

src/facade/dragonfly_connection.h
vyavdoshenko: You still use GetListener() and expose a protected method via GetListener().
You need it only for the GetTlsCertInfo() call. 
Consider adding another method for getting TLS info instead of GetListener().

BorysTheDev (dismissed):

BorysTheDev:
augment review

vyavdoshenko:
augment review

vyavdoshenko (changes requested):
GetListener() issue should be resolved

src/facade/dragonfly_connection.h
akhilp19: I removed the public `Connection::GetListener()` accessor and added `Connection::GetTlsCertInfo()` instead. `FormatInfoMetrics()` now receives `std::shared_ptr<const TlsCertInfo>` directly, so the listener object is no longer exposed through the connection API.

src/facade/tls_helpers.cc
```diff
@@ -133,5 +173,18 @@ void PrintSSLError() {
       nullptr);
 }
 
+std::optional<TlsCertInfo> ParseTlsCertInfo(const X509* cert) {
```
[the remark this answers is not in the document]
akhilp19: Updated the comment to match the actual behavior: `ParseTlsCertInfo()` returns `std::nullopt` only when the certificate pointer is null. If the certificate parses but has empty subject/issuer fields, the empty strings are emitted as-is, which is consistent with the certificate data.

src/facade/dragonfly_listener.h
```diff
@@ -43,6 +45,10 @@ class Listener : public util::ListenerInterface {
   // ReconfigureTLS MUST be called from the same proactor as the listener.
   bool ReconfigureTLS();
 
+  // Returns the TLS certificate metadata currently used by this listener.
+  // nullptr if TLS is not configured or the certificate could not be parsed.
+  std::shared_ptr<const TlsCertInfo> GetTlsCertInfo() const;
```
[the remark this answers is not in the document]
akhilp19: `Listener::GetTlsCertInfo()` is now declared under `#ifdef DFLY_USE_SSL`, matching its definition and preventing a potential link-time failure in non-SSL builds.

tests/dragonfly/connection_test.py
```diff
@@ -1479,6 +1480,64 @@ async def test_tls_reject(
...
+    def cert_field(flag):
+        out = subprocess.check_output(["openssl", "x509", "-in", cert_file, "-noout", flag])
+        return out.decode().strip().split("=", 1)[1]
+
+    expected_subject = _parse_dn(cert_field("-subject"))
+    expected_issuer = _parse_dn(cert_field("-issuer"))
+    expected_not_before = cert_field("-startdate")
+    expected_not_after = cert_field("-enddate")
+
+    client = server.client(**with_ca_tls_client_args)
```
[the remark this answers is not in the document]
akhilp19: `await client.aclose()` is now inside the `finally` block, so the client is always closed even if an assertion fails earlier in the test.

vyavdoshenko (approved):

PR #7846, opened 2026-07-13
perf: snapshot regression background_snapshotting=false

kostasrim:
P99 dropped from 25ms down to 7ms


tests/dragonfly/snapshot_test.py
```diff
@@ -1106,3 +1107,116 @@ async def test_tagged_chunk_replication(df_factory, compression_mode: str):
...
+        "proactor_threads": 2,  # matches the CI runner's core count
+        "dbfilename": "snap-time-parity",
+        # Pin the real production default (src/server/server_state.cc) - the test harness
+        # otherwise auto-injects 300000 (instance.py), which is 4.7x bigger and hides the
+        # per-chunk cost this test is meant to catch.
+        "serialization_max_chunk_size": 65536,
+        "latency_tracking": True,
+    }
+)
+async def test_snapshot_time_parity(async_client: aioredis.Redis):
```
kostasrim, line 1239: My idea is to have a test that compares all 4 ways to snapshot:

1. rdb + background_snapshotting=false
2. rdb + background_snapshotting=true
3. dfs + background_snapshotting=false
4. dfs + background_snapshotting=true

And latency stats for each case.

When one regresses too much we assert fail. 

This is not yet a stable test and it's just a prototype. Waiting for the CI to confirm its behaviour and I will see how I will follow up

kostasrim: I originally tried with a variance of 10% but that didn't fly (rdb with and without background snapshotting is >10% but defintely not 2000x as it was for 3.)

romange: The url only interesting path to cover is dfs. Rdb is mostly for exports, if someone needs to migrate back to redis

kostasrim: so just remove rdb (the two cases). 

Shall I follow up with a fix/direction or ?

romange: Yes please, I will try running this later as well

kostasrim: We have two directions which I thought of but I want your opinion before I actually move on with any of them:

1. Minimal: We can replace sleep with yield. My *concern* here is obvious, if we don't sleep based on a budget the opposite of what we see here can happen (we prioritize the bucket serialization over command execution so CPU time favours serialization). This will work for cases where the bucket latch is held cumulatively (what we have here) but might backfire because we now loose the CPU budget mechanism.

2. We move the SleepFor decision *out of HandleFLushData*. That way we can control the budget && avoid CPU hogging by sleeping *only per bucket batch*. 

I am leaning towards 2. because it keeps control

romange: yes, we tried yield() and it was not good enough.

The goal is to throttle snapshotting fiber to maintain reads and writes responsive. But currently we throttle everything with write commands included . Sleep will eventually evolve into a better algorithm but in any case I think any algorithm will have to throttle only the snapshotting fiber and not the writes. Hence I think we should sleep only if `snapshot_fb_.IsActive()`.

(A better algorithm  - https://github.com/dragonflydb/dragonfly/pull/7774)

kostasrim: indeed but this also solves half of the issue. A write command on a key that belongs to the bucket being serialized by the snapshot fiber will still observe a latency spike. IsActive() will only help the command when it tries to serialize the bucket but it won't help when the prolonged sleep is coming from the snapshot fiber and the write command kicks in (and it can't write because the bucket is being serialized). 

IMO moving the budget boundary as explained in (2) covers both in principle without any negative consequences (like cpu hoging unless I am missing anything)

It's a very small change which I took the liberty to apply. 

If you still think we should only stop the sleep for write commands (i.e. the IsActive) that needs to serialize the bucket I will drop the commit and just do the single line change required for this.

Sorry for pushing back here -- I don't want to be a bottleneck -- but it seems there is an added value for doing (2)

please take a look on the PR and I will revert accordingly if needed

src/server/snapshot.cc
```diff
@@ -193,10 +193,21 @@ void SliceSnapshot::IterateBucketsFb(bool send_full_sync_cut) {
...
-          if (!use_background_mode_ && ThisFiber::GetRunningTimeCycles() > kCyclesPerJiffy) {
+          if (ThisFiber::GetRunningTimeCycles() > kCyclesPerJiffy) {
             ThisFiber::Yield();
           }
         }
+
+        // Pay down CPU-time debt accrued by HandleFlushData() (including any big-value
+        // chunk flushes) since the last batch. This is the only place the snapshot fiber's
+        // backpressure sleep runs - always between TraverseBuckets() batches, so it never
+        // happens while a bucket's BucketDependencies latch is held.
```
kostasrim, line 226: @romange and this is the key difference. Otherwise a write command would still pay the full sleep penalty here even though it's not the one that serializes the bucket. In other words, the penalty is paid on both paths, explicitly (because the command must serialize the bucket) or implicitly (because the command must block to wait the sleep imposed by snapshot fb)

src/server/snapshot.cc
```diff
@@ -193,10 +193,21 @@ void SliceSnapshot::IterateBucketsFb(bool send_full_sync_cut) {
         PushSerialized(false);
       } else {
         if (!PushSerialized(false)) {
-          if (!use_background_mode_ && ThisFiber::GetRunningTimeCycles() > kCyclesPerJiffy) {
+          if (ThisFiber::GetRunningTimeCycles() > kCyclesPerJiffy) {
```
romange, line 218: what is the reason behind this change?

tests/dragonfly/snapshot_test.py
romange: are you saying that we sleep inder lock? what lock prevents from the concurrent write command to progress?

src/server/snapshot.cc
romange: are you saying that we sleep inder lock? what lock prevents from the concurrent write command to progress?

kostasrim: lock is the wrong term. The bucket `latch` is the precise term.

What happens is:

1. we increment bucket dependency 
2. we preempt because we call HandleFlushData SleepFor (latch is on -- big value is being serialized)
3. after the bucket is serialized we release the latch and wake the awaiters. 

It's the same pathogenic case through a different execution path (the snapshot fiber). The write command will trigger OnChange and block on `BucketDependecies::Wait()`.

p.s. It only affects writes on keys whose bucket latch is already taken.

src/server/snapshot.cc
kostasrim: tidy. we take the else branch if `use_background_mode_=false` so the `!use_background_mode` is redundant

src/server/snapshot.cc
kostasrim: We still budget the CPU but at a more appropriate time basically --> after a bucket is serialized but not throughout

kostasrim:
P99 dropped from `25ms down to 7ms`. Test fails because of total running time is more than 1s but I guess that's fine here.

src/server/snapshot.cc
```diff
@@ -287,12 +298,12 @@ void SliceSnapshot::HandleFlushData(std::string data) {
...
-    // serializer_->Flush can be quite slow for large values or due to compression, therefore
-    // we counter-balance CPU over-usage by sleeping.
-    // We measure running_cycles before the preemption points, because they reset the counter.
-    uint64_t sleep_usec = (running_cycles * 1000'000 / base::CycleClock::Frequency()) / 2;
-    ThisFiber::SleepFor(chrono::microseconds(std::min<uint64_t>(sleep_usec, 2000ul)));
+  // Accrue CPU-time debt instead of sleeping here - see accrued_run_cycles_ comment in the
+  // header. Only for the snapshot fiber's own traversal work: a write command reaching this
+  // via inline catch-up serialization (SerializeBucketLocked called from OnChange) must not be
+  // throttled - it needs to complete its own write, not pay for the snapshot's backpressure.
+  if (!use_background_mode_ && snapshot_fb_.IsActive()) {
```
romange: One doubt I have is around `snapshot_fb_.IsActive()` condition here.
I actually think that `accrued_run_cycles_` should increase regardless in this case. I know I mentioned `snapshot_fb_.IsActive`,  but the intuition behind this is that `accrued_run_cycles_` represents the cost of serialization and the only one that pay this cost is snapshot fiber. "serialization budget" is a limited resource and we prefer to spend it on mutations . So if a huge value was serialized from OnChange,  the cost will be high and the snapshot fiber will not progress. if writes do not trigger `SerializeBucketLocked` anymore, then accrued_run_cycles_ will go down and snapshot_fb_ will throttle less.

kostasrim: done. ptal (I also removed the rdb test cases as suggested)

kostasrim:
augment review

src/server/snapshot.cc
```diff
@@ -193,10 +193,21 @@ void SliceSnapshot::IterateBucketsFb(bool send_full_sync_cut) {
...
+
+        // Pay down CPU-time debt accrued by HandleFlushData() (including any big-value
+        // chunk flushes) since the last batch. This is the only place the snapshot fiber's
+        // backpressure sleep runs - always between TraverseBuckets() batches, so it never
+        // happens while a bucket's BucketDependencies latch is held.
+        if (accrued_run_cycles_ > 0) {
+          uint64_t sleep_usec =
+              (accrued_run_cycles_ * 1000'000 / base::CycleClock::Frequency()) / 2;
+          ThisFiber::SleepFor(chrono::microseconds(std::min<uint64_t>(sleep_usec, 2000ul)));
+          accrued_run_cycles_ = 0;
```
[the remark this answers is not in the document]
kostasrim: exceptional catch.

PR #7847, opened 2026-07-13
feat: new format for OAHEntry size

BorysTheDev:
Summary: This PR changes the on-heap layout/encoding of OAHEntry key sizes to use a compact control byte format while preserving optional expiry and cached-hash tagging.

Changes:

- Replaces the old [expiry, key_size(1|4), key] blob format with [expiry?, control, key].
- Encodes key length in a single control byte for keys < 64B, and uses 1 or 3 extra size bytes for larger keys (up to ~0.5GB).
- Removes the tagged-pointer SSO bit and associated size decoding helpers.
- Updates OAHEntry::Create() to emit the new control/extra-bytes layout and set the expiry flag directly.
- Updates OAHEntry::Key() to decode the new layout, with a KeyBig() slow path for large keys.
- Adds unit tests covering boundary sizes and full Add/Find/Erase behavior for large keys.

Technical Notes: The new size format reduces per-entry overhead for small keys while still supporting large keys via a variable-length size field.

### DIFF
--- a/src/core/oah_entry.cc
+++ b/src/core/oah_entry.cc
@@ -1,4 +1,4 @@
-// Copyright 2024, DragonflyDB authors.  All rights reserved.
+// Copyright 2026, DragonflyDB authors.  All rights reserved.
 // See LICENSE for licensing terms.
 //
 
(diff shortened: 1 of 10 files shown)


src/core/oah_entry.cc
```diff
@@ -9,28 +9,35 @@
 namespace dfly {
 
 OAHEntry::TaggedPtr OAHEntry::Create(std::string_view key, uint32_t expiry) {
-  uint32_t key_size = key.size();
+  const uint32_t key_size = key.size();
+  assert(key_size <= kMaxKeySize);
```
vyavdoshenko, line 13: This limit needs to be enforced at runtime, or the encoding needs to cover the allowed max_bulk_len. In release builds this assert is compiled out, but OAHSet can receive SET members up to the default max_bulk_len (2 GiB), which is larger than kMaxKeySize. For a member length of 1 << 29, the encoder truncates the size to 29 bits, Key() decodes it as 0, and lookups/duplicate detection can return incorrect results.

BorysTheDev: As I remember, the maximum string size is 512MB

vyavdoshenko: Double-check it. It can be accepted via max_bulk_len.

vyavdoshenko: I suggest adding a runtime check.

BorysTheDev: https://www.dragonflydb.io/docs/managing-dragonfly/known-limitations

vyavdoshenko (approved):

PR #7849, opened 2026-07-13
feat(search): store & search vectors at native width (FLAT + HNSW, all dtypes)

vyavdoshenko:
Builds on #7839. Stores and queries vector fields at their declared native width (FLOAT32/FLOAT64/FLOAT16/BFLOAT16/INT8/UINT8) end-to-end across FLAT and HNSW, on both HASH and JSON storage.

Changes:
- Query vector is now a raw byte blob on the AST, decoded per-dtype at search time (the parser is schema-agnostic and cannot know the element width).
- OwnedFtVector holds native-width bytes; GetVector takes the field dtype.
- FLAT index stores native-width vectors with a presence bitmap.
- HNSW space uses one typed distance kernel; data size and the deleted-node stub (native-encoded 1.0) follow the dtype.
- HASH borrows client bytes as-is (width-validated); JSON quantizes each number to the native width (round-to-nearest-even; int types reject out-of-range).
- All KNN/RANGE/HYBRID/AGGREGATE paths validate dim via blob_size / width.

Tests: +13 dtype integration tests (HASH/HNSW/JSON x INT8/FP16/FP64, wrong-dim, out-of-range, vector-range, hybrid) and +2 kernel tests. FLOAT32 path unchanged.

### DIFF
--- a/src/core/search/indices.cc
+++ b/src/core/search/indices.cc
@@ -867,76 +867,77 @@ void TagIndex::Tokenize(std::string_view value, uint32_t* /*pos_counter*/,
   NormalizeTags(value, case_sensitive_, separator_, out);
 }
 
-BaseVectorIndex::BaseVectorIndex(size_t dim, VectorSimilarity sim) : dim_{dim}, sim_{sim} {
+BaseVectorIndex::BaseVectorIndex(size_t dim, VectorSimilarity sim, VectorDataType data_type)
+    : dim_{dim}, sim_{sim}, data_type_{data_type} {
 }
 
-std::pair<size_t /*dim*/, VectorSimilarity> BaseVectorIndex::Info() const {
-  return {dim_, sim_};
+VectorIndexInfo BaseVectorIndex::Info() const {
+  return {dim_, sim_, data_type_};
 }
 
 bool BaseVectorIndex::Add(DocId id, const DocumentAccessor& doc, std::string_view field) {
-  auto vector = doc.GetVector(field, dim_);
+  auto vector = doc.GetVector(field, dim_, data_type_);
 
   if (!vector)
     return false;
 
-  if (std::holds_alternative<OwnedFtVector>(*vector)) {
-    const auto& owned_vector = std::get<OwnedFtVector>(*vector);
-    AddVector(id, owned_vector.first.get());
-  } else {
-    const auto& borrowed_vector = std::get<BorrowedFtVector>(*vector);
-    AddVector(id, borrowed_vector);
-  }
+  if (std::holds_alternative<OwnedFtVector>(*vector))
+    AddVector(id, std::get<OwnedFtVector>(*vector).first.get());
+  else
+    AddVector(id, std::get<BorrowedFtVector>(*vector));
 
   return true;
 }
 
-// Each document occupies (dim_ + 1) floats in entries_: dim_ floats for the vector data,
-// followed by one float as a presence marker (1.0 = present, 0.0 = absent/removed).
-// This avoids the previous heuristic of treating all-zero vectors as null.
-static constexpr float kPresent = 1.0f;
-static constexpr float kAbsent = 0.0f;
-
 FlatVectorIndex::FlatVectorIndex(const SchemaField::VectorParams& params,
                                  PMR_NS::memory_resource* mr)
-    : BaseVectorIndex{params.dim, params.sim}, entries_{mr} {
+    : BaseVectorIndex{params.dim, params.sim, params.data_type},
+      stride_bytes_{params.dim * ElementSize(params.data_type)},
+      entries_{mr},
+      present_{mr} {
   DCHECK(!params.use_hnsw);
-  entries_.reserve(params.capacity * (params.dim + 1));
+  entries_.reserve(params.capacity * stride_bytes_);
+  present_.reserve((params.capacity + 63) / 64);
 }
 
 void FlatVectorIndex::AddVector(DocId id, const void* vector) {
-  const size_t stride = dim_ + 1;
-  DCHECK_LE(id * stride, entries_.size());
-  if (id * stride == entries_.size())
-    entries_.resize((id + 1) * stride, 0.0f);
+  const size_t need_bytes = (static_cast<size_t>(id) + 1) * stride_bytes_;
+  if (entries_.size() < need_bytes)
+    entries_.resize(need_bytes);
+
+  const size_t word = id / 64;
+  if (present_.size() <= word)
+    present_.resize(word + 1, 0);
 
   if (vector) {
-    memcpy(&entries_[id * stride], vector, dim_ * sizeof(float));
-    entries_[id * stride + dim_] = kPresent;
+    memcpy(entries_.data() + static_cast<size_t>(id) * stride_bytes_, vector, stride_bytes_);
+    present_[word] |= uint64_t{1} << (id % 64);
+  } else {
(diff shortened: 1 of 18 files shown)


src/server/search/doc_accessors.cc
```diff
@@ -305,17 +308,65 @@ std::optional<BaseAccessor::VectorInfo> JsonAccessor::GetVector(string_view acti
...
+        memcpy(dst, &x, sizeof(x));
+        break;
+      }
+      case search::VectorDataType::INT8: {
+        double d = v.as<double>();
+        // Bound the magnitude before lround (keeps it in-range for long); round to nearest and
+        // reject anything that rounds outside int8 rather than wrapping.
+        if (!std::isfinite(d) || d < -129.0 || d > 128.0)
+          return std::nullopt;
+        long r = std::lround(d);
```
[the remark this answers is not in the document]
vyavdoshenko: fixed

src/core/search/indices.cc
```diff
@@ -867,76 +867,77 @@ void TagIndex::Tokenize(std::string_view value, uint32_t* /*pos_counter*/,
...
                                  PMR_NS::memory_resource* mr)
-    : BaseVectorIndex{params.dim, params.sim}, entries_{mr} {
+    : BaseVectorIndex{params.dim, params.sim, params.data_type},
+      stride_bytes_{params.dim * ElementSize(params.data_type)},
+      entries_{mr},
+      present_{mr} {
   DCHECK(!params.use_hnsw);
-  entries_.reserve(params.capacity * (params.dim + 1));
+  entries_.reserve(params.capacity * stride_bytes_);
+  present_.reserve((params.capacity + 63) / 64);
```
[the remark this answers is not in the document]
vyavdoshenko: fixed

vyavdoshenko:
augment review

src/core/search/indices.cc
```diff
@@ -867,76 +867,77 @@ void TagIndex::Tokenize(std::string_view value, uint32_t* /*pos_counter*/,
...
-  DCHECK_LE(id * stride, entries_.size());
-  if (id * stride == entries_.size())
-    entries_.resize((id + 1) * stride, 0.0f);
+  const size_t need_bytes = (static_cast<size_t>(id) + 1) * stride_bytes_;
+  if (entries_.size() < need_bytes)
+    entries_.resize(need_bytes);
+
+  const size_t word = id / 64;
+  if (present_.size() <= word)
+    present_.resize(word + 1, 0);
```
BorysTheDev: why you can not call resize every time without checking the size?

src/core/search/indices.h
```diff
@@ -298,7 +308,9 @@ struct FlatVectorIndex : public BaseVectorIndex {
   void AddVector(DocId id, const void* vector) override;
 
  private:
-  PMR_NS::vector<float> entries_;
+  size_t stride_bytes_;                // dim_ * ElementSize(data_type_)
+  PMR_NS::vector<std::byte> entries_;  // contiguous native-width vectors
+  PMR_NS::vector<uint64_t> present_;   // presence bitmap, one bit per doc
```
BorysTheDev: `maybe vector<bool> `? if you want to have bitmap

src/core/search/indices.cc
```diff
@@ -867,76 +867,77 @@ void TagIndex::Tokenize(std::string_view value, uint32_t* /*pos_counter*/,
...
   }
 }
 
 void FlatVectorIndex::Remove(DocId id, const DocumentAccessor& doc, string_view field) {
-  const size_t stride = dim_ + 1;
-  if (id * stride + dim_ < entries_.size())
-    entries_[id * stride + dim_] = kAbsent;
+  const size_t word = id / 64;
+  if (word < present_.size())
+    present_[word] &= ~(uint64_t{1} << (id % 64));
```
BorysTheDev: consider to make small function SetDoc/RemDoc because it is hard to read bit magic

src/core/search/indices.cc
```diff
@@ -867,76 +867,77 @@ void TagIndex::Tokenize(std::string_view value, uint32_t* /*pos_counter*/,
...
-  size_t num_slots = entries_.size() / stride;
   std::vector<DocId> result;
-  result.reserve(num_slots);
-  for (DocId id = 0; id < num_slots; ++id) {
-    if (entries_[id * stride + dim_] == kPresent)
-      result.push_back(id);
+  for (size_t word = 0; word < present_.size(); ++word) {
+    uint64_t bits = present_[word];
+    while (bits) {
+      result.push_back(static_cast<DocId>(word * 64 + __builtin_ctzll(bits)));
```
BorysTheDev: prefer standard library instead of compiler specific function. I think you need std::countr_zero

src/core/search/vector_utils.cc
```diff
@@ -148,8 +146,120 @@ float Bf16ToFloat(uint16_t b) {
...
+}
+
+uint16_t FloatToBf16(float f) {
+  uint32_t bits;
+  memcpy(&bits, &f, sizeof(bits));
+  if ((bits & 0x7FFFFFFF) > 0x7F800000)  // nan -> quiet nan
+    return static_cast<uint16_t>((bits >> 16) | 0x0040);
+  bits += 0x7FFF + ((bits >> 16) & 1);  // round to nearest even
+  return static_cast<uint16_t>(bits >> 16);
+}
```
BorysTheDev, line 130: check gcc and clang type supports I think they support native __bf16 type gcc. it will be cool if we avoid all these calculations. or maybe we can use some lib.

src/server/search/doc_accessors.cc
```diff
@@ -305,17 +308,66 @@ std::optional<BaseAccessor::VectorInfo> JsonAccessor::GetVector(string_view acti
...
   for (const auto& v : res[0].array_range()) {
-    if (!v.is_number()) {
+    if (!v.is_number())
       return std::nullopt;
+    std::byte* dst = bytes.get() + i * width;
+    switch (dtype) {
+      case search::VectorDataType::FLOAT32: {
+        float x = v.as<float>();
+        memcpy(dst, &x, sizeof(x));
+        break;
```
BorysTheDev: I don't like we do it every time. it should be possible to solve this with templates

BorysTheDev (dismissed):

src/core/search/indices.cc
vyavdoshenko: Kept the guard - DocIds are recycled via ShardDocIndex::free_ids_, so AddVector can be called with an id smaller than the current max; an unconditional resize would shrink entries_ and drop other docs' vectors. Added a comment making the never-shrink intent explicit.

src/core/search/indices.h
vyavdoshenko: agree. fixed

src/core/search/indices.cc
vyavdoshenko: fixed

src/core/search/indices.cc
vyavdoshenko: fixed

src/core/search/vector_utils.cc
vyavdoshenko: fixed

src/server/search/doc_accessors.cc
vyavdoshenko: fixed

BorysTheDev (approved):

PR #7851, opened 2026-07-14
fix(core): Fix segment dangle

### LINKED ISSUE #7795: serializer_base_test fails: DCHECK it.GetVersion() == snapshot_version_  in SerializerBaseTest.IncreasingLists
### Summary
`serializer_base_test` aborted (`SIGABRT`) on a daily CI run. In `SerializerBaseTest.IncreasingLists`, the debug assertion at `src/server/serializer_base_test.cc:255` failed:

`Check failed: it.GetVersion() == snapshot_version_ (6660 vs. 5001)`

The physical DashTable bucket's version advanced ~1659 versions past `snapshot_version_` while the traversal fiber was still iterating that bucket's slots inside `SerializeBucketLocked`.

### Failing run
https://github.com/dragonflydb/dragonfly/actions/runs/28848798546/job/85558849066 — ctest #99 `serializer_base_test`, `Child aborted`, 99% (1/83) failed.

### Environment
- Debug build (DCHECK active), ctest label `DFLY`.
- SIGABRT only killed test #99's child; ctest continued to #100.

### Assertion
```text
F0707 07:49:04.907188   23068 serializer_base_test.cc:255] Check failed: it.GetVersion() == snapshot_version_ (6660 vs. 5001)
*** SIGABRT received at time=1783410545 on cpu 0 ***
```

Failing loop (`TestDriver::SerializeBucketLocked`):
```cpp
for (it.AdvanceIfNotOccupied(); !it.is_done(); ++it) {
  DCHECK_EQ(it.GetVersion(), snapshot_version_);   // line 255
  Serialize(it.bucket_address(), it->first.ToString(), it->second.ObjType());
  ++serialized;
}
```

### Stack trace
```text
*** Check failure stack trace: ***
    @          0x199e014  absl::lts_20250512::log_internal::LogMessage::PrepareToDie()
    @          0x199e083  absl::lts_20250512::log_internal::LogMessage::SendToLog()
    @          0x199db35  absl::lts_20250512::log_internal::LogMessage::Flush()
    @           0x446398  absl::lts_20250512::log_internal::Voidify::operator&&<>()
    @           0x42444c  dfly::TestDriver::SerializeBucketLocked()
    @           0x89ff06  dfly::SerializerBase::ProcessBucket()
    @           0x42686d  dfly::TestDriver::Loop()::{lambda()#1}::operator()<>()
    @           0x4269b8  dfly::DashTable<>::TraverseBuckets<>()
    @           0x423cbc  dfly::TestDriver::Loop()
    @           0x44202b  dfly::TestDriver::Start()::{lambda()#1}::operator()()
    @           0x4640e4  std::__invoke_impl<>()
    @           0x4622e3  std::__invoke<>()
    @           0x46037e  std::__apply_impl<>()
    @           0x4603b8  std::apply<>()
    @           0x460426  util::fb2::detail::WorkerFiberImpl<>::run_()
    @           0x45d854  util::fb2::detail::WorkerFiberImpl<>::WorkerFiberImpl<>()::{lambda()#1}::operator()()
    @           0x46763a  std::__invoke_impl<>()
    @           0x4667f7  std::__invoke<>()
    @           0x465d28  std::invoke<>()
    @           0x46516d  boost::context::detail::fiber_record<>::run()
    @           0x464216  boost::context::detail::fiber_entry<>()
    @          0x1b21d5f  make_fcontext

*** SIGABRT received at time=1783410545 on cpu 0 ***
PC: @     0x7fe345599e35  (unknown)  raise
    @          0x1a267a8         64  absl::lts_20250512::WriteFailureInfo()
    @          0x1a269b9         96  absl::lts_20250512::AbslFailureSignalHandler()
    @     0x7fe345b2ac70       2048  (unknown)
    @          0x199e056         32  absl::lts_20250512::log_internal::LogMessage::Die()
    @          0x199e0fd         64  absl::lts_20250512::log_internal::LogMessage::SendToLog()
    @          0x199db35        128  absl::lts_20250512::log_internal::LogMessage::Flush()
    @           0x446398         32  absl::lts_20250512::log_internal::Voidify::operator&&<>()
    @           0x42444c        192  dfly::TestDriver::SerializeBucketLocked()
    @           0x89ff06        144  dfly::SerializerBase::ProcessBucket()
    @           0x42686d         48  dfly::TestDriver::Loop()::{lambda()#1}::operator()<>()
    @           0x4269b8        112  dfly::DashTable<>::TraverseBuckets<>()
    @           0x423cbc         80  dfly::TestDriver::Loop()
    @           0x44202b         32  dfly::TestDriver::Start()::{lambda()#1}::operator()()
```

### Analysis / observations
- Failing frame is the **traversal path** (`on_update == false`): `TraverseBuckets → TestDriver::Loop → SerializerBase::ProcessBucket → SerializeBucketLocked`.
- `SerializerBase::ProcessBucket` stamps `it.SetVersion(snapshot_version_)` immediately before `SerializeBucketLocked`; the per-slot loop asserts the stamp holds for every slot. The bucket read back `6660` while `snapshot_version_` was `5001` — the physical bucket was re-versioned mid-iteration while the serializer fiber was preempted inside the loop.
- The identical assertion guards **production** serialization at `src/server/snapshot.cc:234` (`SliceSnapshot::SerializeBucketLocked`, comment "Version is already stamped by SerializerBase::ProcessBucket"). This is a shared PIT-serialization invariant, not a test-only check.
- Intermittent/race: `IncreasingLists` runs `DEBUG POPULATE 5000 ... TYPE LIST` with 3 concurrent `LPUSH` writer fibers; the harness intentionally yields inside serialization to widen the preemption window. Abort only manifests in debug builds (DCHECK).
- Suite already carries a known concurrency-fragile case (`DelayedAllDeleted` is `GTEST_SKIP() << "Deadlocks"`).
- No artifacts: the `SIGABRT` aborted the ctest child; the unit-test step does not collect core dumps.

### Reproduction
- Test: `SerializerBaseTest.IncreasingLists` (`src/server/serializer_base_test.cc`)
- Target: `serializer_base_test` (`src/server/CMakeLists.txt:170`, `helio_cxx_test ... LABELS DFLY`)
- Not deterministic; appears to be a rare race. No existing upstream issue found for `serializer_base_test`.

dranikpg:
With out changes in serialization code, bucket modifications can now happen concurrently. Previously, the single mutex introduced a thread-global linear ordering. When modifications happen concurrently, the dash table can grow, "invalidating" all suspended bucket iterators held by the serialization process. 

The solution is to use segment pointers instead of the segment ids, so when the table grows we still point to the last segment

Fixes #7795

### DIFF
--- a/src/core/dash.h
+++ b/src/core/dash.h
@@ -462,14 +462,40 @@ class DashTable<_Key, _Value, Policy>::Iterator {
   uint8_t slot_id_;
   bool done_;
 
+  // For single bucket iterators we pin the segment to its pointer. Because the iterators are held
+  // across suspension points during concurrent modifications, growing the dashtable will displace
+  // the seg_id_ pointing it to a different bucket.
+  using PinnedType = std::conditional_t<IsSingleBucket, SegmentType*, std::nullptr_t>;
+  PinnedType pinned_seg_ = {};
+
   friend class DashTable;
+
+  // Resolves the segment backing this iterator. Single-bucket iterators use the pinned pointer;
+  // multi-bucket iterators always read the live directory entry.
+  SegmentType* GetSegment() const {
+    if constexpr (IsSingleBucket) {
+      return pinned_seg_;
+    } else {
+      return owner_->segment_[seg_id_];
+    }
+  }
+
+  // Pins the segment pointer from the current directory index (single-bucket iterators only).
+  void PinSegment() {
+    if constexpr (IsSingleBucket) {
+      pinned_seg_ =
+          (owner_ && seg_id_ < owner_->segment_.size()) ? owner_->segment_[seg_id_] : nullptr;
+    }
+  }
 
   Iterator(Owner* me, uint32_t seg_id, detail::PhysicalBid bid, uint8_t sid)
       : owner_(me), seg_id_(seg_id), bucket_id_(bid), slot_id_(sid), done_(false) {
+    PinSegment();
   }
 
   Iterator(Owner* me, uint32_t seg_id, detail::PhysicalBid bid)
       : owner_(me), seg_id_(seg_id), bucket_id_(bid), slot_id_(0), done_(false) {
+    PinSegment();
     Seek2Occupied();
   }
 
@@ -489,6 +515,7 @@ class DashTable<_Key, _Value, Policy>::Iterator {
         bucket_id_(other.bucket_id_),
         slot_id_(other.slot_id_),
         done_(other.done_) {
+    PinSegment();
   }
 
   // Copy constructor from iterator to bucket_iterator and vice versa.
@@ -499,6 +526,7 @@ class DashTable<_Key, _Value, Policy>::Iterator {
         bucket_id_(other.bucket_id_),
         slot_id_(IsSingleBucket ? 0 : other.slot_id_),
         done_(other.done_) {
+    PinSegment();
     // if this - is a bucket_iterator - we reset slot_id to the first occupied space.
     if constexpr (IsSingleBucket) {
       Seek2Occupied();
@@ -515,20 +543,21 @@ class DashTable<_Key, _Value, Policy>::Iterator {
   Iterator& operator=(const Iterator& other) = default;
   Iterator& operator=(Iterator&& other) = default;
 
-  // pre
-  Iterator& operator++() {
+  // Advancing is restricted to single-bucket iterators. A multi-bucket iterator may easily be
+  // invalidated across suspension points, so the cursors and the Traverse function should be used
+  Iterator& operator++() requires IsSingleBucket {
     ++slot_id_;
     Seek2Occupied();
     return *this;
   }
 
-  Iterator& operator+=(int delta) {
+  Iterator& operator+=(int delta) requires IsSingleBucket {
     slot_id_ += delta;
     Seek2Occupied();
     return *this;
   }
 
-  Iterator& AdvanceIfNotOccupied() {
+  Iterator& AdvanceIfNotOccupied() requires IsSingleBucket {
     if (!IsOccupied()) {
       this->operator++();
     }
@@ -536,7 +565,7 @@ class DashTable<_Key, _Value, Policy>::Iterator {
   }
 
   IteratorPairType operator->() const {
-    auto* seg = owner_->segment_[seg_id_];
+    auto* seg = GetSegment();
     return {seg->Key(bucket_id_, slot_id_), seg->Value(bucket_id_, slot_id_)};
   }
 
@@ -546,8 +575,12 @@ class DashTable<_Key, _Value, Policy>::Iterator {
   }
 
   bool IsOccupied() const {
-    return (seg_id_ < owner_->segment_.size()) &&
-           ((owner_->segment_[seg_id_]->IsBusy(bucket_id_, slot_id_)));
+    if constexpr (IsSingleBucket) {
+      return pinned_seg_ && pinned_seg_->IsBusy(bucket_id_, slot_id_);
+    } else {
+      return (seg_id_ < owner_->segment_.size()) &&
+             ((owner_->segment_[seg_id_]->IsBusy(bucket_id_, slot_id_)));
+    }
   }
 
   Owner& owner() const {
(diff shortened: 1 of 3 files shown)


src/core/dash.h
```diff
@@ -515,28 +543,29 @@ class DashTable<_Key, _Value, Policy>::Iterator {
...
 
-  // pre
-  Iterator& operator++() {
+  // Advancing is restricted to single-bucket iterators. A multi-bucket iterator may easily be
+  // invalidated across suspension points, so the cursors and the Traverse function should be used
+  Iterator& operator++() requires IsSingleBucket {
     ++slot_id_;
     Seek2Occupied();
     return *this;
   }
```
dranikpg, line 552: For now i have restricted this functionality with compile time guards, but in the future functions like Traverse can supply the callbacks with somethign more primitive than an "iterator" - as it no iterator anymore

dranikpg:
augment review

kostasrim:
maybe it would worth having vlogs for dash splits such that we can have some sort of observability for those cases shall we wish to. The issue is obvious though, dash is a header and we don't usually put logs there.

Basically what you found could be proven with LOGS. A policy hook could be nice but oh well that's orthogonal on what is being done here

dranikpg:
> maybe it would worth having vlogs for dash splits such that we can have some sort of observability for those cases shall we wish to. 

I added them manually when I investigated this, but there is not much value for main-branch code in it I assume as long as the test is reproducable. I think disabling the iterator operations had the greatest hardening effect

src/core/dash.h
```diff
@@ -462,14 +462,40 @@ class DashTable<_Key, _Value, Policy>::Iterator {
...
+  }
+
+  // Pins the segment pointer from the current directory index (single-bucket iterators only).
+  void PinSegment() {
+    if constexpr (IsSingleBucket) {
+      pinned_seg_ =
+          (owner_ && seg_id_ < owner_->segment_.size()) ? owner_->segment_[seg_id_] : nullptr;
+    }
+  }
+
```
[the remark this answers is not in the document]
romange: @dranikpg  do you need to change anything in Seek2Occupied?

src/server/serializer_base_test.cc
```diff
@@ -348,7 +348,7 @@ TEST_F(SerializerBaseTest, IncreasingLists) {
   // Select keys in range [0, 2 * kKeys] to have a balance of new and existing keys
   std::atomic_bool running = true;
   std::vector<util::fb2::Fiber> workers;
-  for (size_t w = 0; w < 3; w++) {
+  for (size_t w = 0; w < 5; w++) {
```
romange, line 351: curious,  why the change?

src/core/dash_test.cc
```diff
@@ -1445,12 +1445,15 @@ TEST_F(DashTest, Version) {
   }
 
   unsigned items = 0;
-  for (auto it = dt.begin(); it != dt.end(); ++it) {
-    ASSERT_FALSE(it.is_done());
-    ASSERT_GE(it.GetVersion(), it->first + 65000)
-        << it.segment_id() << " " << it.bucket_id() << " " << it.slot_id();
-    ++items;
-  }
+  VersionDT::Cursor cursor;
```
romange, line 1448: does it catch the bug we had before?

src/server/serializer_base_test.cc
kostasrim: to increase the likehood of splitting (by increasing the concurrency) I suspect

src/core/dash.h
```diff
@@ -462,14 +462,40 @@ class DashTable<_Key, _Value, Policy>::Iterator {
...
+  // across suspension points during concurrent modifications, growing the dashtable will displace
+  // the seg_id_ pointing it to a different bucket.
+  using PinnedType = std::conditional_t<IsSingleBucket, SegmentType*, std::nullptr_t>;
+  PinnedType pinned_seg_ = {};
+
   friend class DashTable;
 
+  // Resolves the segment backing this iterator. Single-bucket iterators use the pinned pointer;
+  // multi-bucket iterators always read the live directory entry.
+  SegmentType* GetSegment() const {
```
kostasrim, line 475: We could `DCHECK_EQ(construction_epoch_, FiberSwitchEpoch()) for !IsSingleBucket` had we captured the epoch on pin construction. That way we could enforce the contract on debug builds. 

But that requires to pull fiber header which I don't think it's a great idea :(

kostasrim (commented):
lgtm, ptal Roman's comment -- I do not approve so we don't accidentally miss it

kostasrim:
Also, thinking about this `segment merging` is also safe. We have a check that we are not `snapshotting` so no issues there as well unless I missed anything

kostasrim:
> Also, thinking about this `segment merging` is also safe. We have a check that we are not `snapshotting` so no issues there as well unless I missed anything

uh I think I am wrong because of how it interacts with RestoreStreamer. Ignore it for now though, I never even thought about migrations when I wrote the algorithm

src/core/dash.h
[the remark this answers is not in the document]
dranikpg: Yes, I fixed that already, it's just at a different place

src/core/dash_test.cc
dranikpg: No, it hasn't anything to do with it, I just removed the ++ operator as it was the only place where its used

dranikpg:
> Also, thinking about this segment merging is also safe. We have a check that we are not snapshotting so no issues there as well unless I missed anything

Yes, I remember we added that feature

romange (approved):

PR #7853, opened 2026-07-14
feat: add new size format for OAHPair

BorysTheDev:
## Summary

OAHPair (the OAHMap member) now uses the same variable-width size encoding as OAHEntry and always
stores its value in a separate, 8-byte-aligned allocation so values can be reinterpreted as typed
arrays (e.g. by vector search). The shared size codec and pointer-tag layout are extracted into a
new `oah_base.h`, so OAHEntry and OAHPair no longer duplicate them.

## Changes

- Add `src/core/oah_base.h` with a shared `dfly::oah` namespace: `TaggedPtr`, the pointer
  tag/flag constants, `PrefetchRead`, and the `oah::size` variable-width size codec (1/2/4-byte
  fields, sizes up to ~0.5 GB).
- Reencode OAHPair key and value sizes with `oah::size`; the blob layout is now
  `[expiry?, key_size, value_size, value_ptr:8B, key]`.
- Always store the OAHPair value in its own zmalloc buffer (>= 8-byte aligned); an empty value
  stores a null pointer and allocates no buffer.
- Relocate the blob and the external value buffer independently in `ReallocIfNeeded`, so defrag
  moves only the under-utilized allocation while accounting for both.
- Refactor OAHEntry to consume the shared `oah` / `oah::size` definitions instead of its own
  copies; replace the `PREFETCH_READ` macro with `oah::PrefetchRead` and drop `<sys/types.h>`.
- Give each accessor a class-scope `using TaggedPtr = oah::TaggedPtr;` to keep names local
  without polluting the enclosing namespace.
- Extend `oah_map_test` and `oah_set_test` to cover the new encoding, external-value ownership,
  expiry rebuilds, and selective realloc/defrag accounting.

### DIFF
--- a/src/core/oah_base.h
+++ b/src/core/oah_base.h
@@ -67,12 +67,6 @@ inline uint32_t FieldSize(size_t size) {
                                     : kLargeFieldSize;
 }
 
-inline uint32_t FieldSizeFromControl(uint8_t control) {
-  return (control & kBigSizeBit) == 0 ? kInlineFieldSize
-         : control & kThreeBytesBit   ? kLargeFieldSize
-                                      : kMediumFieldSize;
-}
-
 inline void Write(size_t size, uint32_t field_size, char* dest) {
   assert(field_size == FieldSize(size));
   uint8_t control;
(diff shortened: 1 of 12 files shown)


BorysTheDev:
augment review

BorysTheDev:
augment review

BorysTheDev:
augment review

src/core/oah_base.h
```diff
@@ -0,0 +1,120 @@
...
+  uint32_t extra = static_cast<uint8_t>(src[1]);
+  if (three_bytes) {
+    extra |= static_cast<uint32_t>(static_cast<uint8_t>(src[2])) << 8;
+    extra |= static_cast<uint32_t>(static_cast<uint8_t>(src[3])) << 16;
+  }
+  const uint32_t size = (control & kLowSizeMask) | (extra << kLowSizeBits);
+  return {size, three_bytes ? kLargeFieldSize : kMediumFieldSize};
+}
+
+inline uint32_t ReadLarge(const char* src) {
```
vyavdoshenko: dead code

src/core/oah_base.h
```diff
@@ -0,0 +1,120 @@
...
+};
+
+inline uint32_t FieldSize(size_t size) {
+  assert(size <= kMaxSize);
+  return size <= kInlineSizeMax     ? kInlineFieldSize
+         : size <= kOneExtraByteMax ? kMediumFieldSize
+                                    : kLargeFieldSize;
+}
+
+inline uint32_t FieldSizeFromControl(uint8_t control) {
```
vyavdoshenko: dead code
executed from ReadLarge only

vyavdoshenko (dismissed):
lgtm

vyavdoshenko (approved):
lgtm

PR #7857, opened 2026-07-14
fix(tiering): Fix offloading cooling

dranikpg:
Currently, we cool down each offloaded value, including cases where it was offloaded by background offloading. If it was offloaded by background offloading, it is already considered to be cool and better stored disk only - so we should avoid cooling it at all and polluting the cold queue with it

### DIFF
--- a/src/server/tiered_storage.h
+++ b/src/server/tiered_storage.h
@@ -66,12 +66,6 @@ class TieredStorage : public TieredStorageBase {
   std::error_code Open(std::string_view path);
   void Close();
 
-  // Stash flags below:
-  //
-  // Set if the stash was issued by a client write (and not background offloading).
-  // Such writes can be "cooled down" as we don't yet know the access frequency of the value
-  static constexpr uint8_t kWasClientWrite = 1;
-
   // Enqueue read external value with generic decoder.
   template <typename D, typename F>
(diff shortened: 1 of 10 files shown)


dranikpg:
The big logical difference is that we won't now use all the available RAM and only as much as the offloading threshold has to offer - so maybe instead of avoiding cooling at all, we just need different priority queues?

src/server/tiered_storage.h
```diff
@@ -66,6 +66,12 @@ class TieredStorage : public TieredStorageBase {
   std::error_code Open(std::string_view path);
   void Close();
 
+  // Stash flags below:
+  //
+  // Set if the stash was issued by a client write (and not background offloading).
+  // Such writes can be "cooled down" as we don't yet know the access frequency of the value
+  static constexpr uint8_t kWasClientWrite = 1;
```
romange: there is not need to name based on the context.
Why not have direct `kBypassCool` or similar?

romange:
I think there are cons and pros for every policy. For example if background offloading stashed a "warm" value, it will be fetched faster and the mistake won't affect latency.  Therefore, if you extend this behavior, it's better to introduce a policy flag to allow both options.

dranikpg:
I changed it: first it is guarded by a flag, second the offloaded values are still cooled, but only if there is enough memory to insert them without cooled evictions and they are inserted at the front (lowest priority)

PR #7862, opened 2026-07-15
fix(facade): decrement dropped checkpoint bc on connection close

### LINKED ISSUE #7782: replication_test.py::test_take_over_seeder fails
https://github.com/dragonflydb/dragonfly/actions/runs/28661035575/job/85001424223

glevkovich:
A checkpoint sent to a closing connection was dropped in SendAsync without decrementing its BlockingCounter, while SendCheckpoint had already Add()-ed it. This leaked the DispatchTracker counter, so REPLTAKEOVER / CLIENT PAUSE / migration / shutdown waiting on it blocked until timeout - surfacing as "Couldn't execute takeover: Bad message" in test_take_over_seeder.

- SendAsync now Dec()s a dropped checkpoint's bc at the drop site.
- Harden V2: never read async_fb_ (V1-only) in SendAsync/Migrate; a V2 checkpoint on a closing connection is always dropped+healed.
- Add DVLOG diagnostics for the checkpoint lifecycle.

Fixes #7782 (failed test_take_over_seeder)

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -1948,6 +1948,8 @@ void Connection::ReleaseDeferredCheckpoints() {
   if (deferred_checkpoints_.empty())
     return;
 
+  DVLOG(1) << CONN_ID << " Releasing " << deferred_checkpoints_.size() << " deferred checkpoint(s) "
+           << DebugInfo();
   for (auto& bc : deferred_checkpoints_) {
     bc->Dec();
   }
@@ -2263,7 +2265,7 @@ bool Connection::Migrate(util::fb2::ProactorBase* dest) {
   // Migrate is only used by DFLY Thread and Flow command which both check against
   // the result of Migration and handle it explicitly in their flows so this can act
   // as a weak if condition instead of a crash prone CHECK.
-  if (async_fb_.IsJoinable() || cc_->conn_closing) {
+  if ((!ioloop_v2_ && async_fb_.IsJoinable()) || cc_->conn_closing) {
     return false;
   }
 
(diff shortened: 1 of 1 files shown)


dranikpg:
What about `ClearPipelinedMessages` that closes all the checkpoints?

glevkovich:
> What about `ClearPipelinedMessages` that closes all the checkpoints?

@dranikpg  , good question
- ClearPipelinedMessages only closes checkpoints that are in the dispatch_q_  BEFORE the connection started closing. When cc_->conn_closing is true, any checkpoint triggered  won't reach dispatch_q_.
- The issue in #7782 is for a checkpoint that arrives at SendAsync AFTER cc_->conn_closing is set, so it will never reach disaptch queue -> it is dropped.
- That dropped path is exactly where the Add(1) was "orphan", and it's what the Dec() added fixes.
- Enqueuing it instead so ClearPipelinedMessages could catch it isn't safe  in my opnion.

dranikpg:
> ClearPipelinedMessages only closes checkpoints that are in the dispatch_q_ BEFORE the connection started closing. When cc_->conn_closing is true, any checkpoint triggered won't reach dispatch_q_.

No, this is why there is the `IsJoinable` check - the last step of the async fiber is to process all pending messages without suspensions. `conn_closing` marks that the connection wants to close or is in the closing process, yet it might still be running some command

dranikpg (approved):
I still think it is possible to solve it with proper cleanup. Please see my comment above about the `conn_closing` and `IsJoinable` duality

glevkovich:
> I still think it is possible to solve it with proper cleanup. Please see my comment above about the `conn_closing` and `IsJoinable` duality

@dranikpg I understand what you mean, but the current code will just not work, so we would need a fix in `ClearPipelinedMessages()` to support this for V2.

`ClearPipelinedMessages() `clears `dispatch_q_` first (no suspend), then suspends on `curr->Blocker()->Wait() ` for the in-flight messages waiting on an in-flight tiered command. When this connection fiber is suspended, phase_ == PRECLOSE (see ConnectionFlow) and HasInFlightCommands() can still be true - which allows SendCheckpoint to send commands. So DispatchTracker can call SendAsync at` phase_ == PRECLOSE`, after dispatch_q_ was already drained, **so we'd need to change ClearPipelinedMessages() too.**

My opinion is **there's no reason to send these commands on a closing connection in the first place.** Maybe the things done in ClearPipelinedMessages() are there to deal with an already-existing situation, but that situation should never have happened to begin with -  the message should be dropped at the caller site, not at the queue site.

Anyhow, I don't see how changing ClearPipelinedMessages() makes the solution more elegant or simpler -> it just makes it more complex. And by design, sending a message on a closing connection is the wrong thing to do, in my opinion - the waiter waits longer, and the closing connection does more work for nothing. The best thing is to avoid it all completely.

PR #7866, opened 2026-07-15
feat(server): Cascaded psync

dranikpg:
Phase one of cascaded PSYNC support: pass down "lineage" id to replicas so they recognize "grand" masters

Changes here:
1. Adds cascaded psync flag
2. When set, replica uses same LSN numbering as master
3. `CAPA DRAGONFLY` now passes down master "lineage" id
4. replication now detects own descendant and allows PSYNC
5. Adds two tests for this behaviour

The caveat is that this approach does not work if in M -> R1 -> R2 -> R3 for example R1 becomes standalone and executes a few more commands and we then try to connect R3 to M (it will think it has all the data). Basically when the state of two instances diverges that think of themselves as the master. For this case we need to introduce speical lsn "breaks" in a follow up PR

### DIFF
--- a/src/server/dflycmd.cc
+++ b/src/server/dflycmd.cc
@@ -40,6 +40,7 @@ using namespace std;
 ABSL_DECLARE_FLAG(bool, info_replication_valkey_compatible);
 ABSL_DECLARE_FLAG(uint32_t, replication_timeout);
 ABSL_DECLARE_FLAG(uint32_t, shard_repl_backlog_len);
+ABSL_DECLARE_FLAG(bool, experimental_cascaded_partial_sync);
 
 namespace dfly {
 
@@ -348,6 +349,27 @@ void DflyCmd::Flow(CmdArgParser parser, CommandContext* cmd_cntx) {
 
       // The LSN vector of the same master must be the same size on all replicas.
       if (lsns.size() != my_last_master->last_journal_LSNs.size()) {
+        return cmd_cntx->SendError(facade::kSyntaxErr);
+      }
+
+      DCHECK_LT(flow_id, lsns.size());
+      if (flow_id >= lsns.size()) {
+        LOG(ERROR) << "Invalid flow_id: " << flow_id << " exceeds LSN vector size: " << lsns.size()
+                   << ". Disabling partial sync.";
+        return;  // Fall back to full sync without an error reply.
+      }
+      flow_lsn = lsns[flow_id];
+    } else if (replica_last_master && replica_last_master->id == sf_->GetLineageId() &&
+               absl::GetFlag(FLAGS_experimental_cascaded_partial_sync)) {
+      // Cascaded replication: the connecting replica shares our lineage root, and every node in
+      // the chain keeps its journal in that root's LSN space. So its LSN vector is expressed in
+      // our own LSN space and we can partial-sync it straight from our replication buffer -
+      // whether we are the lineage root itself or an intermediate replica in the chain.
+      ++ServerState::tlocal()->stats.psync_requests_total;
+      const std::vector<LSN>& lsns = replica_last_master->last_journal_LSNs;
+
+      // The LSN vector must carry one entry per shard.
+      if (lsns.size() != shard_set->size()) {
         return cmd_cntx->SendError(facade::kSyntaxErr);
       }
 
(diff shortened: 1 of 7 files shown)


dranikpg:
augment review

src/server/replica.cc
```diff
@@ -412,10 +416,20 @@ std::error_code Replica::HandleCapaDflyResp() {
     PC_RETURN_ON_BAD_RESPONSE(LastResponseArgs()[3].type == RespExpr::INT64);
     master_context_.version = DflyVersion(get<int64_t>(LastResponseArgs()[3].u));
   }
+
+  // If our master is itself a replica (cascaded), parse lineage id (grandparent id)
+  if (LastResponseArgs().size() >= 5) {
+    PC_RETURN_ON_BAD_RESPONSE(LastResponseArgs()[4].type == RespExpr::STRING);
+    master_context_.lineage_id = ToSV(LastResponseArgs()[4].GetBuf());
```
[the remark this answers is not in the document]
dranikpg: True

src/server/replica.cc
```diff
@@ -889,6 +903,13 @@ error_code Replica::ConsumeDflyStream() {
   RETURN_ON_ERR(exec_st_.SwitchErrorHandler(std::move(err_handler)));
 
   LOG(INFO) << "Transitioned into stable sync";
+
+  // Continue the master's LSN numbering so cascaded sub-replicas share the lineage root's LSN
+  // space and can negotiate partial sync when reconnecting up the chain.
+  if (absl::GetFlag(FLAGS_experimental_cascaded_partial_sync)) {
+    StartJournalAtOwnLSN();
```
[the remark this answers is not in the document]
dranikpg: True...

dranikpg:
This is something beyond WIP but still not fully polished

tests/dragonfly/replication_test.py
```diff
@@ -3565,7 +3546,134 @@ async def test_partial_replication_on_same_source_master_with_replica_lsn_inc(df
     assert len(lines) == 1
 
 
-async def test_replicate_hset_with_expiry(df_factory: DflyInstanceFactory):
```
kostasrim: @dranikpg you removed this test accidentally 😄

tests/dragonfly/replication_test.py
```diff
@@ -3565,7 +3546,134 @@ async def test_partial_replication_on_same_source_master_with_replica_lsn_inc(df
...
+    assert (await c_r2.get("diverge")) == "base_R1"
+
+    # r2 reconnects directly to master. It must converge to master's authoritative value.
+    await c_r2.execute_command(f"REPLICAOF localhost {master.port}")
+    await check_all_replicas_finished([c_r2], c_master)
+
+    master_val, r2_val = await asyncio.gather(c_master.get("diverge"), c_r2.get("diverge"))
+    assert r2_val == master_val, f"split-brain divergence: master={master_val!r} r2={r2_val!r}"
+
     master = df_factory.create(proactor_threads=2)
```
kostasrim: And here you continue it 😄

kostasrim:
Is this ready for review ? (It's a draft!)

tests/dragonfly/replication_test.py
dranikpg: true, good catch

src/server/dflycmd.cc
```diff
@@ -331,6 +332,22 @@ void DflyCmd::Flow(CmdArgParser parser, CommandContext* cmd_cntx) {
         lsn_to_start_partial.emplace(*flow_lsn);
       }
 
+    } else if (last_master_id && *last_master_id == sf_->GetLineageId() &&
+               absl::GetFlag(FLAGS_experimental_cascaded_partial_sync)) {
```
kostasrim: The block below is 99.99% the same with the block above with one parameter difference 😄

kostasrim (dismissed):
On takeovers we do not update linage id. So a grandchild connecting on the newly promoted node will fail to psync right? (I guess you already know and you want to follow up just writing here -- also same aplies for replicaof no one)

small comments lgtm

src/server/dflycmd.cc
dranikpg: Refactored, united - looks much cleaner now

BorysTheDev (approved):

PR #7867, opened 2026-07-15
Fix mutable cluster announce flags

romange:
## Summary

Update the active cluster topology when `cluster_announce_ip` or `announce_port` changes through `CONFIG SET`. Preserve configured endpoints when flags are unset, and reapply active announce flags after subsequent `DFLYCLUSTER CONFIG` updates.

Fixes faulty #7860

### DIFF
--- a/src/server/cluster/cluster_defs.cc
+++ b/src/server/cluster/cluster_defs.cc
@@ -64,6 +64,20 @@ ClusterShardInfos::ClusterShardInfos(std::vector<ClusterShardInfo> infos)
   rng::sort(infos_, std::less<>{});
 }
 
+ClusterExtendedNodeInfo* ClusterShardInfos::Find(std::string_view id) noexcept {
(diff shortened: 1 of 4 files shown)


kostasrim:
I will take a look on Monday in detail

BorysTheDev:
@romange, these changes look dangerous. Our design was that all information regarding the cluster we get from the config, and now the config is changed by other commands. It creates inconsistency between different node configurations. I can accept that we do it for emulated mode, but for a real cluster, let's discuss it. I would like to better understand the problem

romange:
I agree but we need it temporary. We will remove them later

BorysTheDev (dismissed):

ashotland:
@romange lets talk before you merge - I think we actually don't need it

kostasrim:
> @romange lets talk before you merge - I think we actually don't need it

are we closing it or ?

PR #7871, opened 2026-07-16
fix(acl): support arguments in DRYRUN

### LINKED ISSUE #5437: ACL DRYRUN does not support multiple arguments
Expected: `ACL DRYRUN SET FOO BAR` checks the permissions for `SET FOO BAR`. Actual:
` ERR wrong number of arguments for 'acl dryrun' command`

danfeideliaoji:
Fixes #5437

## Summary

Allow `ACL DRYRUN` to accept simulated command arguments and check the target user's key patterns.

## Root cause

`DRYRUN` discarded simulated arguments and used unrestricted key permissions, so it could not apply key ACL rules.

## Changes

- Make `ACL DRYRUN` accept a variadic simulated command tail.
- Pass valid simulated arguments to the ACL validator with the target user's key permissions.
- Preserve the existing command-only behavior for incomplete simulated commands.
- Add allow/deny coverage for key-pattern ACL rules.

## Validation

- `ninja -C build-dbg acl_family_test`
- `build-dbg/acl_family_test`
- `pre-commit run --files src/server/acl/acl_family.cc src/server/acl/acl_family_test.cc`

### DIFF
--- a/src/server/acl/acl_family.cc
+++ b/src/server/acl/acl_family.cc
@@ -594,6 +594,7 @@ void AclFamily::DryRun(CmdArgParser parser, CommandContext* cmd_cntx) {
     rb->SendError(error);
     return;
   }
+  const facade::ParsedArgs simulated_args = parser.UnparsedArgs();
 
   const auto& user = registry.find(username)->second;
   // Stub, used to mimic connection context for a user.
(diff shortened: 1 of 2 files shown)


kostasrim:
Hi @danfeideliaoji 

Thank you for your contribution. plz let me know when to review (it's a draft atm) 😄

kostasrim (commented):
A few issues:

* For pubsub it won't work. 
* For acl subcommands won't work. 
* For commands that are parsed with error (DetermineKeys) returning an OpResult() with an error will DCHECK.

I will sent a quick patch for this on a commit from your branch 😄 

Thank you for your contribution 🚀

danfeideliaoji:
> A few issues:
> 
> * For pubsub it won't work.
> * For acl subcommands won't work.
> * For commands that are parsed with error (DetermineKeys) returning an OpResult() with an error will DCHECK.
> 
> I will sent a quick patch for this on a commit from your branch 😄
> 
> Thank you for your contribution 🚀

Thanks, please go ahead with the patch 😄

I also noticed a more general Redis compatibility issue with invalid command
arity. It is not specific to `SET`: commands invoked with the wrong number of
arguments, such as `ACL DRYRUN <user> SET` or `ACL DRYRUN <user> GET`, should
return a wrong-number-of-arguments error.

The current implementation may instead return `OK` when the user has permission
to run the command, because any error returned by `Validate()` causes key checks
to be skipped rather than returning the arity error.

I had an unpushed local draft that separated arity validation, but I'll leave
this to your patch to avoid conflicts. Thanks again!

kostasrim:
@danfeideliaoji I thank you for the contribution! 🚀

kostasrim (approved):

Issue #7666, opened 2026-06-22
S3 snapshots are not working in 1.39


bobelev:
**Describe the bug**

I have clusters that were running version 1.38. After node maintenance, the operator automatically updated the version to 1.39, and the pods could no longer start.

They fail with the following error:

```
E0619 15:42:20.365695      10 utils.cc:255] Unexpected response HTTP/1.1 400 Bad Request
x-amz-bucket-region: eu-south-1
x-amz-request-id: E847D…
x-amz-id-2: GF0QIfLbrpC7FhJ8…
Content-Type: application/xml
Transfer-Encoding: chunked
Date: Fri, 19 Jun 2026 15:42:19 GMT
Connection: close
Server: AmazonS3

47a
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>InvalidToken</Code><Message>The provided token is malformed or otherwise invalid.</Message><Token-0>FwoGZXI…</HostId></Error>
0


<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>InvalidToken</Code><Message>The provided token is malformed or otherwise invalid.</Message><Token-0>FwoGZXIvYX…</HostId></Error>

E0619 15:42:20.383545       7 server_family.cc:1326] Failed to load snapshot with error: end of stream: Failed list objects in S3 bucket
```

**To Reproduce**
1. Create a ServiceAccount with an IRSA role that has access to S3.
1. Deploy a cluster via the operator.
1. Upgrade or allow the operator to upgrade the cluster from 1.38 to 1.39.
1. …
1. Observe that the pods fail to start.

**Expected behavior**

Pods should start successfully after the upgrade and should be able to access the S3 bucket using the IRSA-provided credentials.

Rollback to 1.38.1 works as expected.

romange:
@bobelev  Does it work for you in other regions? Could you try in eu-west-1 or eu-central-1? 

could you confirm that these are the relevant environment variables injected by EKS?
```
 AWS_STS_REGIONAL_ENDPOINTS=regional
 AWS_DEFAULT_REGION=eu-south-1
 AWS_REGION=eu-south-1
 AWS_ROLE_ARN=....
 AWS_WEB_IDENTITY_TOKEN_FILE=/var/run/secrets/eks.amazonaws.com/serviceaccount/token
```

romange:
FYI, I can not reproduce it,  I deployed in `eu-south-1` on fargate I used  an IRSA role and it used it correctly. 

[...]

PR #7872, opened 2026-07-16
fix(stream): honor COUNT when a blocked XREAD/XREADGROUP is woken

jacovdbergh:
## Problem

A blocked `XREADGROUP ... COUNT 1 BLOCK ...` consumer that is woken by a MULTI/EXEC transaction containing multiple `XADD`s receives **all of the transaction's entries in one reply** — `COUNT` is ignored on the wake path. Redis 7 delivers exactly `COUNT` entries to each woken consumer for the same command sequence.

Reproduction against `dragonfly:v1.38.1` / `v1.39.0` (both affected):

```bash
redis-cli XGROUP CREATE q workers '$' MKSTREAM

# Park consumers (e.g. in 4 shells):
redis-cli XREADGROUP GROUP workers c1 COUNT 1 BLOCK 15000 STREAMS q '>'

# Burst entries in one transaction:
(echo MULTI; for i in $(seq 50); do echo "XADD q * f $i"; done; echo EXEC) | redis-cli
```

One parked consumer receives all 50 entries in a single reply (all 50 land in its PEL, per `XPENDING q workers`); the others receive nothing. On Redis 7 each parked consumer receives exactly 1 entry and 46 remain undelivered.

We hit this in production with a job queue built on streams + consumer groups: a batch of ~500 jobs enqueued in one MULTI/EXEC was delivered whole to a single worker's PEL, so one process ran the entire batch serially while the rest of the fleet saw an empty queue. (Entries delivered but unacknowledged are invisible to other consumers' `>` reads, so the work cannot be picked up elsewhere until idle-reclaim.)

## Root cause

`XReadBlock()`'s resume-after-wake callback builds its `RangeOpts` without setting `count`, which therefore stays at the struct default `kuint32max`. The non-blocking path (`OpRead()`) already sets `range_opts.count = opts.count`.

## Fix

Set `range_opts.count = opts->count` in the wake path, mirroring `OpRead()`. Behaviour without an explicit `COUNT` is unchanged (`ReadOpts::count` defaults to `kuint32max` as well).

### DIFF
--- a/src/server/stream_family.cc
+++ b/src/server/stream_family.cc
@@ -3068,6 +3068,10 @@ void XReadBlock(ReadOpts* opts, Transaction* tx, SinkReplyBuilder* builder,
   auto range_cb = [&](Transaction* t, EngineShard* shard) {
     if (auto wake_key = t->GetWakeKey(shard->shard_id()); wake_key) {
       RangeOpts range_opts;
+      // Ensure the woken read honors COUNT like the non-blocking path
+      // (OpRead) does: without it, a consumer woken by a transaction that
+      // added multiple entries receives all of them in one reply.
+      range_opts.count = opts->count;
       range_opts.end = ParsedStreamId{.val = streamID{
                                           .ms = UINT64_MAX,
                                           .seq = UINT64_MAX,
--- a/src/server/stream_family_test.cc
+++ b/src/server/stream_family_test.cc
@@ -448,7 +448,10 @@ TEST_F(StreamFamilyTest, XReadGroupBlockHonorsCount) {
     resp0 = Run({"xreadgroup", "group", "group", "alice", "count", "1", "block", "0", "streams",
                  "foo", ">"});
   });
-  ThisFiber::SleepFor(50us);
+  // Wait until the reader is parked in the blocking path (WaitOnWatch flips
+  // the connection's `blocked` flag) so the transaction below exercises the
+  // wake-up read, not the immediate one. fb0 runs on proactor 0 -> "IO0".
+  ASSERT_TRUE(WaitUntilCondition([&] { return IsConnBlocked("IO0"); }, 500ms));
 
   // Wake it with a transaction adding multiple entries at once. The woken
   // read must honor COUNT like the non-blocking path does, instead of
--- a/src/server/stream_family_test.cc
+++ b/src/server/stream_family_test.cc
@@ -439,6 +439,39 @@ TEST_F(StreamFamilyTest, XReadGroupBlockDelconsumer) {
   EXPECT_THAT(resp_del_consumer, IntArg(0));
 }
 
+TEST_F(StreamFamilyTest, XReadGroupBlockHonorsCount) {
+  Run({"xgroup", "create", "foo", "group", "0", "MKSTREAM"});
+
+  // Block a consumer with COUNT 1.
+  RespExpr resp0;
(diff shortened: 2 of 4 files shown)


jacovdbergh:
Addressed the review point about the fixed `SleepFor(50us)` race in 88cb481: the test now waits deterministically for the reader to park (new `BaseFamilyTest::IsConnBlocked` helper observing the connection context's `blocked` flag, which `Transaction::WaitOnWatch` toggles while parked) before dispatching the waking transaction — so it provably exercises the wake path. Verified both ways in the ubuntu-dev:24 container: passes 5/5 repeats with the fix, fails with "Actual length 3, expected: 1" against the pre-fix source.

romange:
@vyavdoshenko  see another example of behavioral discrepancy with streams. 

@jacovdbergh  how did you find it?

kostasrim:
@romange if we have many of those why don't we simply pull valkey behavioural tests (which they have) and see where we fail and where don't ? We don't need to adopt them in our repo but we can just use them to check against our own implementation.  As far as I remember they have tcl tests and nowdays with the AI we can actually move pretty fast with those.

jacovdbergh:
> We hit this in production with a job queue built on streams + consumer groups: a batch of ~500 jobs enqueued in one MULTI/EXEC was delivered whole to a single worker's PEL, so one process ran the entire batch serially while the rest of the fleet saw an empty queue.

Real world scenario where we batch enqueued many jobs, and throughput seemed very slow. Saw that only 1 queue worker was doing anything, the other 19 stayed idle, until the 30min idle-reclaim passed, and then still pending jobs were picked up by all workers.

romange:
> @romange if we have many of those why don't we simply pull valkey behavioural tests (which they have) and see where we fail and where don't ? We don't need to adopt them in our repo but we can just use them to check against our own implementation.  As far as I remember they have tcl tests and nowdays with the AI we can actually move pretty fast with those.

I asked yesterday @vyavdoshenko to do exactly that

kostasrim:
> > @romange if we have many of those why don't we simply pull valkey behavioural tests (which they have) and see where we fail and where don't ? We don't need to adopt them in our repo but we can just use them to check against our own implementation.  As far as I remember they have tcl tests and nowdays with the AI we can actually move pretty fast with those.
> 
> I asked yesterday @vyavdoshenko to do exactly that

I literally had the prompt written 🤣

kostasrim (approved):
@jacovdbergh very nice catch/fix and thank you for your contribution! 🚀 

Could you plz sign your commits ? (Currently is unverified) See: https://github.com/dragonflydb/dragonfly/blob/main/CONTRIBUTING.md#signing-commits-with-gpg

jacovdbergh:
Thanks! The commits were SSH-signed all along, but the key wasn't registered as a *signing* key on my GitHub account, so they showed as unverified. Fixed — both commits now show Verified.

romange:
One last question it was It was fixed with ai assistance?

jacovdbergh:
Yes indeed, I use Claude Code extensively, but review and tweak all PRs by hand to make sure I don't waste people's time

Would not have spent the time to find the issue and submit a PR without AI tbh

romange:
Yes, 2026 is the year where our community users without extensive knowledge of the project internals become contributors like you or submit high quality issues like this one https://github.com/dragonflydb/dragonfly/issues/7666#issuecomment-4987099037

cheers

PR #7873, opened 2026-07-16
fix(server): emit db= once in CLIENT INFO

vyavdoshenko:
`CLIENT INFO` appended a second `db=<index>` (and a stray trailing `\r\n`) on top of the `db=` that `GetClientInfo()` already produces, so the reply carried the field twice and ended with an extra newline. Valkey returns a single line with `db=` once and no trailing newline.

Changes:
- Remove the redundant `db=`/`\r\n` append in the `CLIENT INFO` handler; `GetClientInfo()` already includes `db=` and the dummy counter fields.
- Add regression test `ServerFamilyTest.ClientInfoSingleDbField` asserting `db=` appears exactly once and the reply has no trailing CRLF.


kostasrim (approved):
lgtm!

PR #7874, opened 2026-07-16
fix(stream): avoid use-after-free garbling long stream ids in replies

vyavdoshenko:
A stream id whose text form exceeds 32 bytes (a large `ms` together with a large `seq`) was garbled when read back - `XRANGE`/`XREAD`/`XINFO`/`XPENDING`/`XAUTOCLAIM` returned an id with its first bytes overwritten. `StreamIdRepr()` returns a temporary `std::string`; passed to `SendBulkString` under an enclosing `ReplyScope`, a >32-byte string is enqueued by reference and copied only when the outer scope ends - after the temporary is destroyed (use-after-free).

Changes,
- Add a `SendStreamId(rb, id)` helper that sends the id under a `ScopePause`, forcing it to be copied while still alive, and route all stream id replies through it (14 sites).
- Add regression test `StreamFamilyTest.XRangeLongIdRoundTrip`.

### DIFF
--- a/src/facade/reply_builder.h
+++ b/src/facade/reply_builder.h
@@ -217,6 +217,14 @@ class RedisReplyBuilderBase : public SinkReplyBuilder {
   void SendSimpleString(std::string_view str) override;
   virtual void SendBulkString(std::string_view str);  // RESP: Blob String
 
+  // Reject a temporary std::string: under a ReplyScope it would be enqueued by reference and read
+  // after destruction (use-after-free). Materialize into a named std::string first. Constrained to
+  // std::string rvalues, so string literals / string_view / lvalue strings are unaffected (a plain
+  // `std::string&&` overload would make `SendBulkString("literal")` ambiguous).
+  template <typename T>
+  requires std::is_same_v<T, std::string>
+  void SendBulkString(T&&) = delete;
+
   void SendBulkStringBorrowed(const cmn::BorrowedString& bs);
 
   // The interface exposes only the rvalue function to allow squashing to "steal" the value,
--- a/src/facade/reply_builder.h
+++ b/src/facade/reply_builder.h
@@ -9,6 +9,7 @@
 #include <optional>
 #include <string_view>
 
+#include "base/logging.h"
 #include "common/borrowed_string.h"
 #include "facade/facade_stats.h"
 #include "facade/facade_types.h"
@@ -115,6 +116,10 @@ class SinkReplyBuilder {
     return batched_;
   }
 
+  bool IsScoped() const {
+    return scoped_;
+  }
+
   void CloseConnection();
 
   static const ReplyStats& GetThreadLocalStats() {
@@ -217,13 +222,17 @@ class RedisReplyBuilderBase : public SinkReplyBuilder {
   void SendSimpleString(std::string_view str) override;
   virtual void SendBulkString(std::string_view str);  // RESP: Blob String
 
-  // Reject a temporary std::string: under a ReplyScope it would be enqueued by reference and read
-  // after destruction (use-after-free). Materialize into a named std::string first. Constrained to
-  // std::string rvalues, so string literals / string_view / lvalue strings are unaffected (a plain
-  // `std::string&&` overload would make `SendBulkString("literal")` ambiguous).
+  // Forward a temporary std::string to the string_view overload. Sending a temporary under a
+  // ReplyScope would enqueue it by reference and read it after destruction (use-after-free), so the
+  // DCHECK guards that the builder is not scoped (when unscoped the value is copied immediately).
+  // Constrained to std::string rvalues, so string literals / string_view / lvalue strings are
+  // unaffected (a plain `std::string&&` overload would make `SendBulkString("literal")` ambiguous).
   template <typename T>
   requires std::is_same_v<T, std::string>
-  void SendBulkString(T&&) = delete;
+  void SendBulkString(T&& str) {
+    DCHECK(!IsScoped());
+    SendBulkString(std::string_view{str});
+  }
 
   void SendBulkStringBorrowed(const cmn::BorrowedString& bs);
 
--- a/src/facade/reply_builder.h
+++ b/src/facade/reply_builder.h
@@ -7,9 +7,10 @@
 
 #include <boost/intrusive/list.hpp>
 #include <optional>
+#include <string>
 #include <string_view>
+#include <type_traits>
 
-#include "base/logging.h"
 #include "common/borrowed_string.h"
 #include "facade/facade_stats.h"
 #include "facade/facade_types.h"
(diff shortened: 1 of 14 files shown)


src/server/stream_family.cc
dranikpg: I'm not sure, why not move the scope instead

```c++
rb->StartArray(2);
SendStreamId(rb, record.id);

ReplyScope scope();
rb->StartArray(....)
for (...)
  Send()
```

dranikpg: hm... and we would have to add a scope for batching now so that the single `SendBulkString(StreamIdRepr` doesn't make a send call

src/server/stream_family.cc
```diff
@@ -3857,14 +3869,16 @@ void CmdXPending(CmdArgParser parser, CommandContext* cmd_cntx) {
...
+  SinkReplyBuilder::ReplyAggregator agg{rb};
   SinkReplyBuilder::ReplyScope scope{rb};
   if (std::holds_alternative<PendingReducedResult>(result)) {
     const auto& res = std::get<PendingReducedResult>(result);
     rb->StartArray(4);
     rb->SendLong(res.count);
     if (res.count) {
-      rb->SendBulkString(StreamIdRepr(res.start));
-      rb->SendBulkString(StreamIdRepr(res.end));
+      SendStreamId(rb, res.start);
```
dranikpg: ok here it would be more difficult I agree

dranikpg (dismissed):
Feels a little weird to have an exception to a rule where we control the rule, but other approaches might not be simpler

src/server/stream_family_test.cc
```diff
@@ -1289,6 +1289,19 @@ TEST_F(StreamFamilyTest, XAddMaxSeq) {
   EXPECT_THAT(resp, ErrArg("The ID specified in XADD is equal or smaller"));
 }
 
+TEST_F(StreamFamilyTest, XRangeIdRoundTrip) {
```
romange, line 1292: @kostasrim  this could be the bug we have been chasing with use-after free and corrupted listpacks for a year

src/server/stream_family.cc
```diff
@@ -709,6 +709,20 @@ string StreamIdRepr(const streamID& id) {
   return absl::StrCat(id.ms, "-", id.seq);
 };
 
+// Sends a stream id as a bulk string. A bulk string longer than kMaxInlineSize is enqueued by
+// reference and copied only when the enclosing ReplyScope ends — after this temporary is gone
+// (use-after-free that garbles long ids). Pause the scope for long ids so they are materialized
+// while still alive; short ids are inlined immediately and need no pause.
+void SendStreamId(RedisReplyBuilder* rb, const streamID& id) {
+  string s = StreamIdRepr(id);
+  if (s.size() > SinkReplyBuilder::kMaxInlineSize) {
```
romange: no, it's not a good fix, it leaks implementation details of SinkReplyBuilder and looks fragile.

src/server/stream_family.cc
```diff
@@ -2716,7 +2730,7 @@ struct StreamReplies {
 
   void SendRecord(const Record& record) const {
     RedisReplyBuilder::ArrayScope scope{rb, 2};
-    rb->SendBulkString(StreamIdRepr(record.id));
```
romange, line 2719: @dranikpg  is it possible to prevent passing rvalues to SendBulkString ? Will it be a good idea?

src/server/stream_family.cc
```diff
@@ -2716,7 +2730,7 @@ struct StreamReplies {
...
+    SendStreamId(rb, record.id);
```
romange: The fix should be - to not use ArrayScope in this case. Just send ids without the reference scope. RedisReplyBuilder should copy them internally to its batching buffer. @dranikpg  is it correct?

romange: we should use `ReplyAggregator` imho

src/server/stream_family.cc
```diff
@@ -2727,7 +2741,7 @@ struct StreamReplies {
   void SendIDs(absl::Span<const streamID> ids) const {
     RedisReplyBuilder::ArrayScope scope{rb, ids.size()};
```
romange: same - use `ReplyAggregator`

romange (commented):
really good catch! 💯

src/server/stream_family.cc
dranikpg: We can use the scope just below

src/server/stream_family.cc
dranikpg: SendBulkString(string&&) = delete, but we use that method without scopes in places

vyavdoshenko:
@romange @dranikpg reworked to `ReplyAggregator` (dropped the `ScopePause` helper) and added the guard `template<class T> requires same_as<T,std::string> SendBulkString(T&&) = delete` (plain `string&&` -> `SendBulkString("literal")` ambiguous; `CapturingReplyBuilder` needs a `using`).
All 36 rvalue sites materialized. Also fixes the same UAF in `FT.SEARCH WITHSORTKEYS`.

src/server/acl/acl_family.cc
```diff
@@ -219,7 +219,10 @@ void AclFamily::DelUser(CmdArgParser parser, CommandContext* cmd_cntx) {
 
 void AclFamily::WhoAmI(CmdArgParser parser, CommandContext* cmd_cntx) {
   auto* rb = static_cast<facade::RedisReplyBuilder*>(cmd_cntx->rb());
-  rb->SendBulkString(absl::StrCat("User is ", cmd_cntx->server_conn_cntx()->authed_username));
+  {
+    std::string sbs_reply = absl::StrCat("User is ", cmd_cntx->server_conn_cntx()->authed_username);
```
romange: So as @dranikpg said sometimes it's a valid case. I think it's a great that we found another bug but it's a drag to use these named variables everywhere. Let's revert the deletion of rvalue parameter. Instead we should add a runtime check - so rvalue overload should dcheck that scoped flag is not set and delegate to the regular SendBulkString. This way we will be able to send tempoaries AND catch these bugs

romange (changes requested):

src/server/acl/acl_family.cc
romange: `DCHECK(!scoped_)` for rvalue overload.

src/server/stream_family_test.cc
kostasrim: I do not want to be the party pooper here but I highly doubt they are related :( 

Here we have a bug while replying to a client which is related to the reply builder. We have a dangling temporary that is irrelevant to the actual state of the listpack.

As far as I remember, the crash we had on the corrupted listpack was within the `data structure state itself`. And (maybe I am wrong here) it always happened on the replica side while executing the command. Those two don't share any code path. One is on the client level while the other is on the data structure level.

The only thing it seems those two share is the nature of the bug: use-after-free. 

Also this `StreamIdRepr` is bounded to two ints ? So the allocation class is a small bucket. For listpack it's not (it's 4kb).

If my mental model around is wrong or I am missing something plz correct me 🤷

kostasrim: @romange ptal :)

src/server/acl/acl_family.cc
vyavdoshenko: fixed

src/facade/reply_builder.h
```diff
@@ -9,6 +9,7 @@
 #include <optional>
 #include <string_view>
 
+#include "base/logging.h"
```
romange: do not add `"base/logging.h"` in include files. move the implementation to cc file

src/facade/reply_builder.h
```diff
@@ -217,6 +222,18 @@ class RedisReplyBuilderBase : public SinkReplyBuilder {
...
   virtual void SendBulkString(std::string_view str);  // RESP: Blob String
 
+  // Forward a temporary std::string to the string_view overload. Sending a temporary under a
+  // ReplyScope would enqueue it by reference and read it after destruction (use-after-free), so the
+  // DCHECK guards that the builder is not scoped (when unscoped the value is copied immediately).
+  // Constrained to std::string rvalues, so string literals / string_view / lvalue strings are
+  // unaffected (a plain `std::string&&` overload would make `SendBulkString("literal")` ambiguous).
+  template <typename T>
+  requires std::is_same_v<T, std::string>
+  void SendBulkString(T&& str) {
```
romange: why do you need template for this to work? what happens if you directly declare `std::string&&`

romange (commented):
almost there

src/facade/reply_builder.h
vyavdoshenko: A plain std::string&& makes every SendBulkString("literal") call ambiguous - const char* converts equally to string_view and std::string (~100 sites break).

src/facade/reply_builder.h
vyavdoshenko: fixed

romange (approved):

PR #7875, opened 2026-07-16
fix(search): avoid use-after-free garbling long WITHSORTKEYS values

vyavdoshenko:
`FT.SEARCH … WITHSORTKEYS` with a string sort value whose `$`-prefixed form exceeds 32 bytes returned a garbled sortkey. The `"$" + s` temporary was enqueued into the reply by reference under the result `ArrayScope` and copied only when the scope ended - after the temporary was destroyed (use-after-free). Same class as the stream-id UAF (#7874).

Changes,
- Materialize the sortkey and force a copy (via `ScopePause`) for values > `kMaxInlineSize`.
- Add regression test `SearchFamilyTest.WithSortKeysLongValue`.


vyavdoshenko:
Should be fixed in #7874

PR #7876, opened 2026-07-16
fix(tiering): Defrag uses DashTable + iteration cycles

dranikpg:
1. Lower max current defrags limit
2. Use DashTable instead of absl::flat_hash_map in small bins
3. Add background defrag iteration
4. Make it adaptive to raise the max cycles quota if it occured

### DIFF
--- a/src/server/tiered_storage.cc
+++ b/src/server/tiered_storage.cc
@@ -86,7 +86,7 @@ constexpr auto kFragmentedBin = tiering::SmallBins::kInvalidBin - 1;
 
 // Memory budget (UploadBudget) does not account for buffers allocated for in-flight defrag reads,
 // so we cap the number of concurrent defragmentation operations to avoid unbounded memory growth.
-constexpr uint32_t kMaxPendingDefrags = 100;
+constexpr uint32_t kMaxPendingDefrags = 50;
 
 // Called after setting new value in place of previous segment
 void RecordDeleted(const FragmentRef& fragment_ref, size_t tiered_len, DbTableStats* stats) {
@@ -225,7 +225,7 @@ class TieredStorage::ShardOpManager : public tiering::OpManager {
   bool NotifyFetched(const OwnedEntryId& id, tiering::DiskSegment segment,
                      tiering::Decoder* decoder) override;
 
-  bool NotifyDelete(tiering::DiskSegment segment) override;
+  bool NotifyDelete(tiering::DiskSegment segment, bool was_read) override;
 
   void EnqueueForDefrag(tiering::DiskSegment segment);
 
@@ -319,9 +319,6 @@ class TieredStorage::ShardOpManager : public tiering::OpManager {
     uint64_t total_uploads = 0;
     uint32_t pending_defrags = 0;
   } stats_;
-
-  // When we don't have memory to upload a page for defragmentation, we save it here to do it later
-  std::deque<uint32_t /* page index (kPageSize) */> delayed_defrag_queue_;
 
   TieredStorage* ts_;
   DbSlice& db_slice_;
@@ -417,7 +414,7 @@ bool TieredStorage::ShardOpManager::NotifyFetched(const OwnedEntryId& id,
   return false;
 }
 
-bool TieredStorage::ShardOpManager::NotifyDelete(tiering::DiskSegment segment) {
+bool TieredStorage::ShardOpManager::NotifyDelete(tiering::DiskSegment segment, bool was_read) {
   DVLOG(2) << "NotifyDelete [" << segment.offset << "," << segment.length << "]";
 
   if (OccupiesWholePages(segment.length))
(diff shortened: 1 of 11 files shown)


dranikpg:
augment review

src/server/tiered_storage.cc
```diff
@@ -717,19 +710,44 @@ void TieredStorage::RunOffloading(DbIndex dbid) {
...
+    }
+  };
+
+  // Scale the cpu time-slice by how much work the previous scan found: sleepy on a stale table,
+  // more aggressive while there's a backlog to clear.
+  const uint64_t time_budget_us = std::min(kBaseUs + last_defrag_scan_hits_ * kUsPerHit, kMaxUs);
+
+  uint64_t cycles = 0;
+  do {
+    if (UploadBudget() < 0)
```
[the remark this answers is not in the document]
dranikpg: +4kb but fixed

src/server/tiered_storage.cc
```diff
@@ -717,19 +710,44 @@ void TieredStorage::RunOffloading(DbIndex dbid) {
...
+      break;
+
+    if (op_manager_->stats_.pending_defrags > kMaxPendingDefrags)
```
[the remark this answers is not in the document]
dranikpg: +1 but fixed

src/server/tiered_storage.cc
```diff
@@ -717,19 +710,44 @@ void TieredStorage::RunOffloading(DbIndex dbid) {
...
 
-    if (!bins_->IsFragmented(offset))
-      continue;
-    op_manager_->EnqueueForDefrag({offset, tiering::kPageSize});
-  }
+  unsigned hits = 0;
+  auto cb = [this, &hits](size_t offset) {
+    tiering::DiskSegment segment{offset, tiering::kPageSize};
+    if (!op_manager_->HasReadPending(kFragmentedBin, segment)) {
+      op_manager_->EnqueueForDefrag({offset, tiering::kPageSize});
```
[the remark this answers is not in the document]
dranikpg: is ok

src/server/tiering/op_manager.h
```diff
@@ -92,7 +94,8 @@ class OpManager {
   virtual bool NotifyFetched(const OwnedEntryId& id, DiskSegment segment, Decoder*) = 0;
 
   // Notify delete. Return true if the filled segment needs to be marked as free.
-  virtual bool NotifyDelete(DiskSegment segment) = 0;
+  // If was_read is set, the item was just read and can still be reached with a read
```
romange: I do not understand the comment. What does it mean?

dranikpg: It means NotiftyDelete can either be called after a read or just so arbitrarily when a value is deleted - in the first case we have access to the value that was deleted

src/server/tiering/op_manager.h
```diff
@@ -92,7 +94,9 @@ class OpManager {
   virtual bool NotifyFetched(const OwnedEntryId& id, DiskSegment segment, Decoder*) = 0;
 
   // Notify delete. Return true if the filled segment needs to be marked as free.
-  virtual bool NotifyDelete(DiskSegment segment) = 0;
+  // Under was_read this function was called after a read completion as the follow-up delete
```
romange: I saw this style several times so I want to comment here about naming and contracts. 
`was_read` and the comment above explains how you got here (the caller context) and not what it does for the function. The contract is not clear. While the logic is sound I think a better style, imho - to describe the variables (and their meaning) in terms of a contract without bringing the caller context into it if possible. For example,

```
// Notify delete. Return true if the filled segment needs to be marked as free.
// page_in_memory is true when the delete follows a completed read, so the page's data is
// still in the pending-reads cache and defrag can reuse it without an extra disk read.
bool NotifyDelete(DiskSegment segment, bool page_in_memory) = 0;
```

if you are changing the name, please sync it with every `NotifyDelete` declaration/implementation

src/server/tiering/small_bins.cc
```diff
@@ -102,7 +102,8 @@ SmallBins::KeySegmentList SmallBins::ReportStashed(BinId id, DiskSegment segment
   }
 
   stats_.stashed_entries_cnt += list.size();
-  stashed_bins_[segment.offset] = {uint8_t(list.size()), bytes};
+  stashed_bins_.InsertNew(segment.offset, StashInfo{uint8_t(list.size()), bytes});
```
romange, line 111: how are you sure the key is unique? 
where in the flow before we check for it?

src/server/tiered_storage.cc
```diff
@@ -676,7 +669,7 @@ void TieredStorage::RunOffloading(DbIndex dbid) {
   const auto start_cycles = base::CycleClock::Now();
 
   // Takes up a small fixed amount of time and is best done before offloading (to be picked up)
```
romange: nit: fixed -> bounded, but also revisit the comment if it still applies as is.

src/server/tiering/small_bins.h
```diff
@@ -102,8 +108,37 @@ class SmallBins {
...
+    template <typename U> static void DestroyValue(const U&) {
+    }
+    template <typename U> static void DestroyKey(const U&) {
+    }
+
+    template <typename U, typename V> static bool Equal(U&& u, V&& v) {
+      return u == v;
+    }
+
+    static uint64_t HashFn(uint64_t v) {
```
romange: nice! 👏🏼

romange (commented):
LGTM

src/server/tiering/op_manager.h
dranikpg: Done

src/server/tiering/small_bins.h
dranikpg: actually that was an overkill

dranikpg: no, no overkill - dividing by kPageSize breaks the dashtable as it requires high bits to be set as well

dranikpg:
@romange @mkaruza PTAL

romange (approved):

PR #7877, opened 2026-07-17
[WIP/DNM] server: Clear now_chunked_ items on home shards

abhijat:
Previously the items in now_chunked_ were destroyed on shard other than the home shard potentially, this is okay in mimalloc itself but we have checks on top to assert that the shard matches.

On finishing load we now submit the task to home shards to clear the items.

Root cause identified and fix by by https://github.com/avy2 ref https://github.com/dragonflydb/dragonfly/pull/7780

The entry is first removed during `RdbLoader::CreateObjectOnShard` if it has expired. The cleanup at finish load treats pending entries as error log but still cleans them up.

### DIFF
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -2582,6 +2582,7 @@ void RdbLoader::DiscardChunkedValuesOnFinish() {
     if (values.empty())
       continue;
 
+    LOG(ERROR) << "unexpected " << values.size() << " values found in chunk map on RDB load finish";
     bc->Add(1);
     shard_set->Add(sid, [values = std::move(values), bc]() mutable {
       values.clear();
@@ -3013,6 +3014,15 @@ void RdbLoader::CreateObjectOnShard(const DbContext& db_cntx, const Item* item, 
   PrimeValue* pv_ptr = &pv;
   DbIndex db_ind = db_cntx.db_index;
 
+  bool failed_should_cleanup = false;
+  ChunkedKey key{db_ind, item->key};
+  absl::Cleanup maybe_remove_chunk_entry = [&] {
+    if (!failed_should_cleanup)
+      return;
+    std::unique_lock l{now_chunked_mu_};
+    now_chunked_.erase(key);
+  };
+
   auto error_msg = [](const auto* item, auto db_ind) {
     return absl::StrCat("Found empty key: ", item->key, " in DB ", db_ind, " rdb_type ",
                         item->val.rdb_type);
@@ -3039,6 +3049,7 @@ void RdbLoader::CreateObjectOnShard(const DbContext& db_cntx, const Item* item, 
   }
 
   if (auto ec = FromOpaque(item->val, config_copy, pv_ptr); ec) {
+    failed_should_cleanup = item->load_config.chunked;
     if (ec.value() == errc::value_expired) {
       // hmap and sset values can expire and we ok with it,
       // so we don't set ec_ in this case
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -2566,6 +2566,32 @@ error_code RdbLoader::Load(io::Source* src) {
   return kOk;
 }
 
+void RdbLoader::DiscardChunkedValuesOnFinish() {
+  using Values = std::vector<std::unique_ptr<PrimeValue>>;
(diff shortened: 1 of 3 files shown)


src/server/rdb_load.cc
```diff
@@ -2581,7 +2607,7 @@ void RdbLoader::FinishLoad(absl::Time start_time, size_t* keys_loaded) {
     GetCurrentDbSlice().DecrLoadInProgress();
   }
 
-  now_chunked_.clear();
+  DiscardChunkedValuesOnFinish();
```
romange, line 2611: @abhijat  what is scenario where we have non empty chunked items?

abhijat: basically it can happen inside `RdbLoader::CreateObjectOnShard`. That method does not return error, it only sets `stop_early_` conditionally and returns void, there are two types of error return there:

* some will set `stop_early_` and return, eg oom while adding key, non recoverable errors
* some will just log error and return but not set `stop_early_` and return, eg value found expired when loading `RdbLoaderBase::FromOpaque`

In the second case we do not get to the point where the value is extracted from `now_chunked_` to add to the db, so it remains around in the map at EOF.

I guess another (better?) approach could be to add an `absl::Cleanup` inside `RdbLoader::CreateObjectOnShard` and act on some local error flag, and the cleanup will remove the entry from map if the flag is set. 

It would cover all the error paths and make sure that the map is empty on EOF.

abhijat: actually `stop_early_` doesnt really matter, even if that is set, on failed create attempt, entries will be left behind in `now_chunked_`, and `FinishLoad` will need to clean them up. 

> I guess another (better?) approach could be to add an absl::Cleanup inside RdbLoader::CreateObjectOnShard and act on some local error flag, and the cleanup will remove the entry from map if the flag is set.

The good thing about this approach is it (`CreateObjectOnShard`) runs on the keys home shard any way so it doesn't need any shard set add etc. It can directly destroy the map entry.

abhijat: I added this cleanup for expired keys but I still think we need to cleanup up on `FinishLoad` any way

avy252004:
> Previously the items in now_chunked_ were destroyed on shard other than the home shard potentially, this is okay in mimalloc itself but we have checks on top to assert that the shard matches.
> 
> On finishing load we now submit the task to home shards to clear the items.
> 
> Root cause identified and fix by by https://github.com/avy2 ref #7780
> 
> The entry is first removed during `RdbLoader::CreateObjectOnShard` if it has expired. The cleanup at finish load treats pending entries as error log but still cleans them up.

I worked on this issue independently as well and compared my implementation with this draft. Apart from a small readability tweak(aggregating the log message) and a comment in the regression test, I couldn't find any meaningful changes to make. If you'd still like me to open a separate PR for this issue, I'd be happy to do so.

I also considered a regression test covering multiple remote shard pending chunked values, but it ends up being more of a complex stress-style test. I don't think it's worth adding here since the current regression already covers the reported bug.

abhijat:
> I worked on this issue independently as well and compared my implementation with this draft. Apart from a small readability tweak(aggregating the log message) and a comment in the regression test, I couldn't find any meaningful changes to make. If you'd still like me to open a separate PR for this issue, I'd be happy to do so.

The code has diverged a little bit, early cleanup during shard local work when creating objects seems better to me. The cleanup during `FinishLoad` is still needed though.

abhijat:
@avy252004 since it is mostly your original code shape plus the test, please feel free to create a PR if you want, if not let us know and I will put this up for review etc.

avy252004:
> @avy252004 since it is mostly your original code shape plus the test, please feel free to create a PR if you want, if not let us know and I will put this up for review etc.

ok i will open the PR and try to address this complete issue.

abhijat:
@avy252004 do you think you can raise a PR some time this week? Since this is a bug we would like to merge a fix sooner rather than later, if you can't get to it soon we can go ahead with this current PR

avy252004:
> @avy252004 do you think you can raise a PR some time this week? Since this is a bug we would like to merge a fix sooner rather than later, if you can't get to it soon we can go ahead with this current PR

sorry for delay, i can try by this weekend if  i am unable to, please go ahead.

PR #7878, opened 2026-07-17
Fix Azure snapshot URI account resolution

### LINKED ISSUE #7863: Support standard azure storage url in `dfly load <url>` command
Dragonfly requires an ENV variable `AZURE_STORAGE_ACCOUNT ` to get the storage account and the url format it supports is - `az://<container>/<key>`. However, this url format is not a valid azure storage url. The correct azure storage URL format is `https://<account>.blob.core.windows.net/<container>/<key>`.

Support the azure storage URL format and use the `<account>` from the URL as storage account.

romange:
Support standard Azure Blob Storage snapshot URLs and derive the storage account from the URI rather than `AZURE_STORAGE_ACCOUNT`.

- Accept `https://<account>.blob.core.windows.net/<container>/<key>` for Azure SAVE and LOAD paths.
- Validate and normalize Azure URL components, including percent-encoded blob keys.
- Add real Azure and self-starting Azurite SAVE → FLUSH → LOAD coverage.
- Update the Helio submodule for URI account handling and request-target escaping (romange/helio#617), plus custom endpoint overrides that preserve URI account selection (romange/helio#618).

Fixes #7863

### DIFF
--- a/src/server/detail/snapshot_storage.cc
+++ b/src/server/detail/snapshot_storage.cc
@@ -5,8 +5,12 @@
 
 #include <absl/base/optimization.h>
 #include <absl/flags/flag.h>
+#include <absl/strings/ascii.h>
 #include <absl/strings/str_replace.h>
+#include <absl/strings/str_split.h>
 #include <absl/strings/strip.h>
+
+#include <algorithm>
 
 #ifdef WITH_AWS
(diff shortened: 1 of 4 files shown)


dranikpg (dismissed):

Abhra303 (approved):
Thanks 👍

dranikpg (approved):

PR #7880, opened 2026-07-18
fix(zset): clamp NaN once after full aggregation in ZUNIONSTORE

avy252004:
`ZUNIONSTORE` with `SUM` aggregate could return `inf` instead of `0` when a member's score across 3+ sources included
both `inf` and `-inf` that should cancel out (e.g. `inf + -inf + inf`).

`Aggregate()` clamped `NaN` to `0` on every pairwise merge instead of once at the end. The first merge (`inf + -inf = NaN`)
got clamped to `0` immediately, erasing the fact a cancellation happened. A later merge then saw a clean `0 + inf` and
returned `inf` instead of `0`.

**Fix:** removed the per-merge clamp from `Aggregate()`. Added a single clamp in `ZBooleanOperation`, applied once after
all shard-local and cross-shard merges are done.

Added `ZUnionStoreSumNanCancelsToZero` covering both cases (all sources on one shard, and sources split across shards),
each verified to fail on the old code and pass on the fix.

### DIFF
--- a/src/server/zset_family_test.cc
+++ b/src/server/zset_family_test.cc
@@ -753,6 +753,42 @@ TEST_F(ZSetFamilyTest, ZUnionStoreOpts) {
   EXPECT_THAT(resp, DoubleArg(0));
 }
 
+// Regression test: inf + -inf during SUM aggregation must clamp to 0, even when
+// an early clamp inside a single shard's merge would otherwise get overwritten
+// by a later merge and produce inf instead. See Aggregate()/ZBooleanOperation()..
+TEST_F(ZSetFamilyTest, ZUnionStoreSumNanCancelsToZero) {
+  // Case 1: all sources on the SAME shard — the corrupting merge sequence happens
+  // entirely inside UnionShardKeysWithScore's local merge loop.
+  const char kKeySameShardA[] = "x";
+  const char kKeySameShardB[] = "m";
+  ASSERT_EQ(Shard(kKeySameShardA, shard_set->size()), Shard(kKeySameShardB, shard_set->size()))
+      << "test requires these two keys on the same shard";
+
+  Run({"ZADD", kKeySameShardA, "inf", "e1"});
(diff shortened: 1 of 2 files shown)


src/server/zset_family_test.cc
```diff
@@ -753,6 +753,42 @@ TEST_F(ZSetFamilyTest, ZUnionStoreOpts) {
...
+  const char kKeySameShardA[] = "x";
+  const char kKeySameShardB[] = "m";
+  ASSERT_EQ(Shard(kKeySameShardA, shard_set->size()), Shard(kKeySameShardB, shard_set->size()))
+      << "test requires these two keys on the same shard";
+
+  Run({"ZADD", kKeySameShardA, "inf", "e1"});
+  Run({"ZADD", kKeySameShardB, "-inf", "e1"});
+  Run({"ZUNIONSTORE", "dest_same_shard", "3", kKeySameShardA, kKeySameShardB, kKeySameShardA});
+  auto resp = Run({"ZSCORE", "dest_same_shard", "e1"});
+  EXPECT_THAT(resp, DoubleArg(0)) << "same-shard union: inf + -inf + inf must clamp to 0";
```
romange, line 771: I run on valkey:
```
"ZADD" a  "inf" "e1"
"ZADD" b  "-inf" "e1"
"ZUNIONSTORE" c 3 a b a
zscore c e1
```

and got "inf".

why do you think this should be clamped?

avy252004: thank you for that pointing out, i was thinking mathematically because (INF-INF +INF = NaN) so it should be 0 but in   valkey this runs and check sequentially i think. That is why the original code was doing the same for compatibility.

i only noticed this as bug because in parallel the original code was more dependent on shards and the order they are fetched and union, so answers would be different when we do per union clamp.

but if this is the case than this pr has no point, But only highlighting shard-topology can influence sequential-clamp  outcomes based .

PR #7881, opened 2026-07-18
feat: add ASCII packing for OAH

BorysTheDev:
Summary: Adds an ASCII 7-bit packing layer for OAH (open-addressing hash) keys to reduce in-memory key footprint while keeping hashing/comparison on stored bytes and decoding only when returning logical keys.

Changes:

- Introduced ASCIIStr (7-bit pack/unpack + fused “try-pack”) in src/core/str_codec.h.
- Added oah::key header/codec in src/core/oah_base.h, reusing size header bit 7 as an “encoded” flag without affecting size decoding.
- Updated OAHEntry/OAHPair blobs to store a key header + packed/raw content; added StoredKey, KeyContent, and KeyMatches.
- Adjusted OAHTable to hash/compare on stored content, provide DecodeKey, and support both ephemeral Scan and retainable ScanDecoded.
- Modified OAHSet/OAHMap insertion/find/erase paths to encode query keys consistently via ASCIIStr.
- Updated set-related server code paths (iteration, SCAN, RDB save, randmember) to use Key() + GetKeyView().

Technical Notes: Encoded keys are limited to 8..128 bytes of pure 7-bit ASCII; the encoding flag is stored in the key header’s control byte (bit 7), and decoding happens only when producing logical key views.

### DIFF
--- a/src/core/oah_base.h
+++ b/src/core/oah_base.h
@@ -15,20 +15,19 @@
 namespace dfly {
 
 // A key reduced once to the bytes used for hashing and comparison: 7-bit ASCII-packed into buf_
-// when worth it (ascii, length 8..128), else a view of the original bytes. Uses the shared
-// detail::bitpacking codec. Non-copyable: content() may alias buf_.
+// when worth it (ascii, length 8..128) via the shared detail::ascii_try_pack codec, else a view
+// of the original bytes. Non-copyable: content() may alias buf_.
 class ASCIIStr {
  public:
   static constexpr uint32_t kMinLen = 8;    // shorter strings pack to the same size, so stay raw
   static constexpr uint32_t kMaxLen = 128;  // largest length we ascii-encode
 
-  explicit ASCIIStr(std::string_view s) : len_(static_cast<uint32_t>(s.size())) {
-    if (s.size() >= kMinLen && s.size() <= kMaxLen &&
-        detail::validate_ascii_fast(s.data(), s.size())) {
-      detail::ascii_pack_simd2(s.data(), s.size(), reinterpret_cast<uint8_t*>(buf_));
-      content_ = {buf_, detail::binpacked_len(s.size())};
-    } else {
-      content_ = s;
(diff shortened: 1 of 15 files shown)


BorysTheDev:
augment review

BorysTheDev:
augment review

src/core/oah_base.h
```diff
@@ -7,9 +7,50 @@
...
+// detail::bitpacking codec. Non-copyable: content() may alias buf_.
+class ASCIIStr {
+ public:
+  static constexpr uint32_t kMinLen = 8;    // shorter strings pack to the same size, so stay raw
+  static constexpr uint32_t kMaxLen = 128;  // largest length we ascii-encode
+
+  explicit ASCIIStr(std::string_view s) : len_(static_cast<uint32_t>(s.size())) {
+    if (s.size() >= kMinLen && s.size() <= kMaxLen &&
+        detail::validate_ascii_fast(s.data(), s.size())) {
+      detail::ascii_pack_simd2(s.data(), s.size(), reinterpret_cast<uint8_t*>(buf_));
```
vyavdoshenko: This enables packing on big-endian s390x, but ascii_pack_simd2 falls back to ascii_pack, which performs a little-endian load and then copies seven bytes from the native-endian uint64_t. Consequently, encoded keys can be found using the same representation but decode to different bytes during SMEMBERS, SSCAN, SPOP, and RDB serialization. Please use an endian-stable store or disable packing on big-endian targets.

vyavdoshenko: ascii_pack_simd2 is not safe for inputs shorter than 32 bytes: it computes ascii + len - 32, which forms a pointer before the beginning of the source object. That is undefined behavior even if the pointer is not dereferenced. Since ASCIIStr enables packing starting at 8 bytes, all 8–31 byte ASCII keys hit this path. Please use the scalar implementation for len < 32 or fix the helper to use remaining-length arithmetic.

vyavdoshenko (approved):

PR #7882, opened 2026-07-18
refactor: drop runtime parser

BorysTheDev:
Summary: This PR refactors command argument parsing by removing the legacy runtime “Apply” combinators and consolidating parsing on the compile-time (consteval) grammar API.

Changes:

- Extends CmdArgParser’s compile-time grammar with new primitives: Flags, Switch, TagValue, Into, Skip, and parsed-field support (Field<Parsed>).
- Tightens parsing/type constraints via ParsedArg, TokenParser, and numeric helper aliases like Positive/NonNegativeInt.
- Removes runtime parsing helpers (Apply/ApplyOrSkip and related option factories) and updates unit tests accordingly.
- Migrates multiple commands to the compile-time grammar style (e.g., DEBUG RELOAD, TOPK LIST, RESTORE, EXPIRE*, SORT, GEOSEARCH/GEORADIUS*, HSETEX/HGETEX, GETEX, ZRANGE options, ZADD, and hybrid search parsing).
- Adds/updates tests and micro-bench parsing coverage to exercise the new grammar primitives.

Technical Notes: The new approach keeps parsing rules static/constexpr, reduces per-call construction, and centralizes option conflict handling (e.g. via OneOf/Switch).

### DIFF
--- a/src/facade/cmd_arg_parser.h
+++ b/src/facade/cmd_arg_parser.h
@@ -11,6 +11,7 @@
 #include <cassert>
 #include <cmath>
 #include <concepts>
+#include <limits>
 #include <optional>
 #include <string_view>
 #include <tuple>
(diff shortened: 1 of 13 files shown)


BorysTheDev:
augment review

BorysTheDev:
augment review

src/server/zset_family.cc
```diff
@@ -1780,23 +1780,18 @@ void ZRangeGeneric(facade::ParsedArgs args, ZSetFamily::RangeParams range_params
...
-  // TODO: remove runtime parsing (migrate to cap grammar).
-  parser.Apply(Tag("BYSCORE", set_interval(RP::SCORE)), Tag("BYLEX", set_interval(RP::LEX)),
-               Exist("REV", &range_params.reverse), Exist("WITHSCORES", &range_params.with_scores),
-               Tag("LIMIT", [&](CmdArgParser* p) {
-                 auto [offset, limit] = p->Next<int32_t, int32_t>();
-                 range_params.limit = limit < 0 ? UINT32_MAX : static_cast<uint32_t>(limit);
-                 range_params.offset = offset < 0 ? UINT32_MAX : static_cast<uint32_t>(offset);
-               }));
+  static constexpr auto kGrammar =
+      Compile(Options(OneOf("BYSCORE and BYLEX options are not compatible",
```
vyavdoshenko, line 1788: range_params.interval_type may already be SCORE or LEX for the legacy handlers, but OneOf starts with an unmatched state. Therefore, an opposite option overwrites the preselected type instead of reporting a conflict. For example, ZRANGEBYLEX z 0 10 BYSCORE now executes a score range, while the previous implementation rejected it. Please preserve the initial interval type or disallow BYLEX/BYSCORE for fixed-type handlers.

src/server/hset_family.cc
```diff
@@ -718,9 +718,22 @@ OpResult<bool> CheckHSetExCondition(const OpArgs& op_args, string_view key,
...
 };
 
+optional<DbSlice::ExpireParams> MakeFieldExpireParams(ExpT type, int64_t value, uint64_t now_ms,
+                                                      bool allow_expired) {
+  if (value < 0)
+    return nullopt;
+
+  DbSlice::ExpireParams params{type, value, now_ms};
+  auto [ttl_ms, expire_at_ms] = params.Calculate(now_ms, false);
+  if (expire_at_ms < 0 || ttl_ms > kMaxExpireDeadlineMs || (!allow_expired && ttl_ms <= 0))
```
vyavdoshenko, line 732: This changes the existing member-TTL ceiling from 1 << 26 to kMaxExpireDeadlineSec ((1 << 28) - 1). SADDEX still enforces 1 << 26, and the HGETEX tests document the same limit for hash-field expiration. The same widening occurs at hset_family.cc:783,855 and generic_family.cc:2239. Please retain the existing limit in this refactor, or migrate all entry points and update the documented behavior/tests separately.

src/facade/cmd_arg_parser.h
```diff
@@ -1007,10 +1016,12 @@ template <class T, class... Alts> struct OneOf {
...
   using Target = T;
   using State = std::tuple<typename Rules::State...>;
-  consteval explicit Options(Rules... rules) : rules_(rules...) {
+  consteval explicit Options(Rules... rules) : Options(0, rules...) {
+  }
+  consteval Options(size_t tail, Rules... rules) : tail_(tail), rules_(rules...) {
   }
   void Consume(CmdArgParser* p, T* o, State& st) const {
-    while (p->InBounds()) {
+    while (p->HasAtLeast(tail_ + 1)) {
```
vyavdoshenko: tail_ + 1 only checks the remaining argument count before applying a rule. A multi-token rule such as Field or TagValue can then consume the reserved tail, contradicting the documented guarantee that Options(n, ...) leaves at least n arguments. Please add rule-aware lookahead, or narrow the documented contract to single-token rules, and add a regression test. A post-match check alone cannot restore already consumed arguments.

src/server/hset_family.cc
BorysTheDev: I did it explicitly, I want to have the same behavior in all code. 1 << 26 was added by me previously to support full Redis compatibility, but I don't see a problem to use 1 << 28, because we already use this constant in other part of the code and we don't need to be 100% compatible regarding ttl with Redis

vyavdoshenko (approved):
lgtm

Intentional widening makes sense, but SADDEX still uses 1 << 26, so the member-TTL entry points are not uniform yet. Could you switch it to the shared constant and update the stale HGETEX comment/add a boundary test?

set_family.cc:1605
hset_family_test.cc:1029

BorysTheDev:
> SADDEX still uses 1 << 26, so the member-TTL entry points are not uniform yet.

I will add it to the separate PR
https://github.com/dragonflydb/dragonfly/issues/7893

PR #7885, opened 2026-07-19
CI: minor QOL improvements to repeat-tests

abhijat:
* Fix title of repeat-tests - it always showed DEBUG before which was confusing/wrong
* Allow yes/no/true/false etc for the booleans instead of forcing specific words

### DIFF
--- a/.github/actions/repeat/action.yml
+++ b/.github/actions/repeat/action.yml
@@ -52,15 +52,14 @@ runs:
         # used by PyTests
         export DRAGONFLY_PATH="${GITHUB_WORKSPACE}/${{inputs.build-folder-name}}/${{inputs.dfly-executable}}"
         export UBSAN_OPTIONS=print_stacktrace=1:halt_on_error=1 # to crash on errors
-        if [[ "${{ inputs.epoll }}" == "epoll" ]]; then
+        FORCE_EPOLL=""
+        if ${{ contains(fromJSON('["epoll","yes","true"]'), inputs.epoll) }}; then
           FORCE_EPOLL="--df force_epoll=true"
-        else
-          FORCE_EPOLL=""
         fi
-        if [[ $"{{ inputs.vmodule_expression }}" != "" ]]; then
+
+        VMOD=""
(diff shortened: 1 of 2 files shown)


.github/workflows/repeat-tests.yml
```diff
@@ -93,7 +94,7 @@ jobs:
 
       - name: Fetch release
         shell: bash
-        if: ${{ inputs.use_release == 'yes' }}
+        if: ${{ contains(fromJSON('["yes","true","Yes","True","YES","TRUE"]'), inputs.use_release) }}
```
kostasrim: contains is `case insensitive` https://docs.github.com/en/actions/reference/workflows-and-actions/expressions 

note: 

* GitHub ignores case when comparing strings. 

Plz test this manually after the change so we don't get any surprises 😄

.github/actions/repeat/action.yml
```diff
@@ -52,7 +52,9 @@ runs:
         # used by PyTests
         export DRAGONFLY_PATH="${GITHUB_WORKSPACE}/${{inputs.build-folder-name}}/${{inputs.dfly-executable}}"
         export UBSAN_OPTIONS=print_stacktrace=1:halt_on_error=1 # to crash on errors
-        if [[ "${{ inputs.epoll }}" == "epoll" ]]; then
+        EPOLL_INPUT="${{ inputs.epoll }}"
+        EPOLL_INPUT="${EPOLL_INPUT,,}"  # lowercase
```
kostasrim: you can use `jq`

`if jq -e --arg v "${{ inputs.epoll }}" `

@romange taught me this trick a year ago on some internal thing we worked on 😄

kostasrim (commented):
lgtm small comment

.github/actions/repeat/action.yml
abhijat: I changed it to just use `contains`

.github/workflows/repeat-tests.yml
abhijat: changed to use just the two strings for comparing

kostasrim (approved):
much cleaner now 🚀

PR #7889, opened 2026-07-19
feat(pubsub): close stuck subscribers to avoid publisher throttling

### LINKED ISSUE #7888: Improve back-pressure behavior around PUBLISH operation
# Avoid publisher throttling by closing stuck Pub/Sub subscribers

## Problem

In the RESP V1 path, a subscriber that stops reading can leave its connection
blocked in socket output. Published messages accumulate in its dispatch queue
until the per-I/O-thread `publish_buffer_limit` is exceeded. Publishers
targeting that thread then park in `EnsureMemoryBudget`, even when the
subscriber is permanently stuck. `EnsureMemoryBudget` can enforce a memory
red line, but it runs from the publishing path and cannot safely identify or
interrupt an individual connection.

The current implementation exposes the queue size through
`dispatch_queue_subscriber_bytes`, but it does not:

- count how often the Pub/Sub budget is crossed;
- identify that a particular subscriber is blocking socket output;
- close a subscriber that is stuck long enough to cause publisher throttling;
- count connections closed by Dragonfly for this reason; or
- expose how much queued Pub/Sub data was discarded when such a connection
  was closed.

See [Pub/Sub publisher back-pressure](pub-sub.md#backpressure-resp-v1) for the
current V1 behavior.

## Proposed behavior

Use two thresholds:

- **Soft limit:** the existing `publish_buffer_limit`. Crossing it starts a
  back-pressure episode and enables slow-subscriber protection, but does not
  immediately park publishers.
- **Hard limit:** a higher red-line derived from the soft limit, for example
  `publish_buffer_limit * pubsub_hard_limit_multiplier` with a default
  multiplier of `4`. `EnsureMemoryBudget` may park publishers only at this
  limit, preserving a final memory safety mechanism.

Add a slow-subscriber protection policy for RESP V1:

1. Detect a transition where an I/O thread's queued Pub/Sub bytes cross the
   soft limit.
2. Continue accepting and dispatching messages while the thread remains below
   the hard limit.
3. Track the active Pub/Sub socket write at the subscriber connection level.
4. If the soft-limit pressure is active and that write has been blocked
   continuously for at least `subscriber_stalled_millis`, request closure of the
   subscriber.
5. Interrupt the active write through a safe asynchronous mechanism, then use
   the normal connection shutdown path.
6. Drain and account for all pending dispatch messages during connection
   cleanup. This releases the per-thread subscriber budget and wakes blocked
   publishers.

The two conditions should be conjunctive. A slow send below the configured
Pub/Sub budget should not cause an otherwise healthy subscriber to be closed,
and a full Pub/Sub budget should not close a subscriber whose output is still
making progress.

The close should be treated as a server-side overload protection action, not
as a protocol error delivered to the subscriber. The connection must be closed
using the normal shutdown path so all queued-message accounting, waiters, and
connection statistics remain consistent.

### Detecting and closing a stuck asynchronous operation

The connection should track a general `async_op_in_progress` state around
control-path execution, rather than adding Pub/Sub-specific state to the
send implementation. In the RESP V1 path, `ProcessAdminMessage` should set
the state immediately before `std::visit` and clear it immediately afterward.
The state should include a monotonic start time. This also
keeps the mechanism useful for other asynchronous control operations.

Note that `SinkReplyBuilder::Send()` already measures send duration and tracks pending
sends in a thread-local `pending_list`, but the existing
`oldest_pending_send_ts` metric has aggregate scope. This policy should use the
connection-local operation state to determine whether the same socket write
has been blocked continuously for at least `subscriber_stalled_millis`; therefore `pending_list`
does not help here.

When `SendPubMessageAsync` observes soft-limit pressure and an overdue
Pub/Sub-producing operation, it must not wait, join, or call
`EnsureMemoryBudget`. It may:

1. discard the new Pub/Sub message;
2. remove already queued `PubMessage` items from `dispatch_q_`, releasing their
   accounting immediately; and
3. call `MarkForClose` to set the reply-builder error and
   `request_shutdown_`.

`MarkForClose` must remain non-blocking: it may only mark the reply builder and
set the shutdown flag. `ProcessAdminMessage` should check
`request_shutdown_` immediately after `std::visit` returns. If it is set, it
should stop normal processing and let the connection's existing shutdown path
close the connection after the pending asynchronous operation returns.

Once `request_shutdown_` is set, `SendPubMessageAsync` should drop new
`PubMessage` items instead of repopulating the queue. Queue eviction and normal
connection cleanup remain responsible for releasing all accounting and waking
publishers. the post-`std::visit` check only runs once that pending operation returns.

`subscriber_stalled_millis` should be a configurable server flag with a
conservative default, and `0` should disable automatic slow-subscriber
closure. The issue should initially keep it startup-only, like
`publish_buffer_limit`, to avoid
inconsistent per-thread deadlines. Runtime mutability can be added later only
if the value is propagated safely to every I/O thread.

### Required counters

Use one counter `pubsub_backpressure_events_total` with a bounded `event` label:

| Metric | `event` value | Meaning |
| --- | --- | --- |
| `pubsub_backpressure_events_total` | `soft_limit_crossing` | Number of transitions where an I/O thread's queued Pub/Sub bytes crossed from at/below the soft limit to above it. |
| `pubsub_backpressure_events_total` | `hard_limit_throttled` | Number of publisher wait episodes caused by reaching the hard limit. |
| `pubsub_backpressure_events_total` | `close_request` | Number of non-blocking close requests issued after the soft-limit and stuck-send conditions were met. |
| `pubsub_backpressure_events_total` | `forced_disconnect` | Number of subscriber connections actually closed by Dragonfly for this policy. |
| `pubsub_backpressure_events_total` | `messages_discarded` | Total queued Pub/Sub messages discarded during policy-driven disconnects. |

## Logging and diagnostics

Emit a rate-limited structured log when Dragonfly closes a subscriber, including
only bounded diagnostic fields:

- connection identifier in logs, not metric labels;
- I/O thread index;
- configured `publish_buffer_limit`;
- thread's current `subscriber_bytes`;
- connection's queued Pub/Sub bytes and entries;
- blocked send duration;
- number of discarded Pub/Sub messages and bytes.

Do not log once per queued message or once per blocked publisher. The
budget-crossing and forced-disconnect counters are the durable signals.

## Configuration and safety

- Add a flag such as `pubsub_slow_subscriber_timeout_ms` for `subscriber_stalled_millis`.
- Add a constand such as `kPubSubHardMultiplier`, defaulting to `4`.
- Default `pubsub_slow_subscriber_timeout_ms` to disabled initially, or choose
  a value only after measuring normal large-response send durations in
  production.

## Testing

Add or extend tests to cover:

1. A stuck subscriber below the budget is not closed.
2. Crossing the soft limit does not park publishers before the hard limit.
3. A soft-limit crossing with a subscriber that continues making send
   progress is not closed.
4. A soft-limit crossing plus an asynchronous operation blocked for at least
   `subscriber_stalled_millis` milliseconds closes the subscriber without blocking or preempting
   `SendPubMessageAsync`.
5. Existing `test_publish_stuck` behavior remains correct, and a dedicated
   test verifies automatic closure without requiring the client to disconnect.
6. Pending Pub/Sub messages are evicted and accounted exactly once while the
   current operation unwinds.
7. Repeated close requests result in one shutdown and one accounting cleanup.
8. Forced close releases queued Pub/Sub bytes and unblocks all publishers.
9. The soft-limit and hard-limit counters increment once per episode, not
   once per publish.
10. The forced-disconnect counter and discarded-byte/message metrics match the
   connection's queue at close.
11. `subscriber_stalled_millis=0` disables forced closure.

romange:
## Summary

**The problem** 
A slow (stuck or zombie) subscriber - blocks any publisher indefinitely until it either unblocks or closed. Publisher back-pressure does not pass reality test as such cases occur in production. 

**Solution**
This Pr splits the per-IO-thread Pub/Sub back-pressure budget into a **soft** limit (`publish_buffer_limit`) and a **hard** limit (`soft * kPubSubHardMultiplier`, hard-coded to 4), and adds a RESP IoLoop-V1 slow-subscriber protection policy that closes a subscriber whose socket write has been stuck long enough to throttle publishers. 

We still use throttling mechanism but its thershold grows to `publish_buffer_limit * 4`. In addition we force-close subscribers that are stuck for more than `pubsub_slow_subscriber_timeout_ms` AND the subsriber memory crossed the soft limit of `publish_buffer_limit`

And what's even more important - we add metrics covering pubsub behavior around both soft and hard limit states.

Fixes #7888.

## Changes
- `EnsureMemoryBudget` parks publishers only at the **hard** limit; crossing the **soft** limit starts a back-pressure episode (counted).
- New startup-only flag `--pubsub_slow_subscriber_timeout_ms` (default `0` = disabled). When a thread is above the soft limit **and** a connection's active Pub/Sub send has been blocked continuously past the timeout, `SendPubMessageAsync` discards the message, evicts the connection's queued `PubMessage`s (releasing the per-thread budget and waking parked publishers), and calls the non-blocking `MarkForClose`.
- `ProcessAdminMessage` tracks the async-op start time around `std::visit` and stops normal processing once `request_shutdown_` is set; the connection closes through the normal shutdown path when the parked send returns. No socket-write interruption.
- Metric `pubsub_backpressure_events_total` with a bounded `event` label: `soft_limit`, `hard_limit`, `forced_disconnect`, `messages_discarded`.
- remove slow and quite useless test `test_pubsub_busy_connections` that did not have any assertions and
did not provide any real coverage. Now with AI we should add better tests than that.

## Notes
- The soft/hard split is always active: with the policy disabled, publishers now park at `4 * publish_buffer_limit` instead of `publish_buffer_limit`.

## Testing
- `test_pubsub_slow_subscriber_closed` — stuck subscriber is force-closed, publishers unblock, counters/discards accounted.
- `test_pubsub_slow_subscriber_disabled` — with the timeout at 0, a stuck subscriber is never closed.
- Existing `test_publish_stuck` and other pubsub tests still pass.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -82,8 +82,16 @@ ABSL_FLAG(uint32_t, pipeline_queue_limit, 10000,
           "may require increasing this limit to prevent the risk of deadlocking."
           "See https://github.com/dragonflydb/dragonfly/discussions/3997 for details");
 
-ABSL_FLAG(strings::MemoryBytesFlag, publish_buffer_limit, 128_MB,
-          "Amount of memory to use for storing pub commands in bytes - per IO thread");
+ABSL_FLAG(strings::MemoryBytesFlag, publish_buffer_limit, 196_MB,
+          "Amount of memory to use for storing pub commands in bytes - per IO thread. This is the "
+          "soft Pub/Sub back-pressure limit; publishers are only parked once the per-thread "
+          "subscriber memory reaches this value times the internal hard-limit multiplier.");
+
+ABSL_FLAG(uint32_t, pubsub_slow_subscriber_timeout_ms, 0,
+          "If a subscriber connection keeps a Pub/Sub socket write blocked for at least this many "
+          "milliseconds while the per-IO-thread subscriber memory is above the soft "
+          "publish_buffer_limit, Dragonfly closes that subscriber to release the back-pressure. "
+          "0 disables this slow-subscriber protection. Startup-only, like publish_buffer_limit.");
 
 ABSL_FLAG(uint32_t, pipeline_squash, 1,
           "Number of queued pipelined commands above which squashing is enabled, 0 means disabled");
@@ -156,6 +164,11 @@ using nonstd::make_unexpected;
 namespace facade {
 
 namespace {
+
+// Multiplier applied to the soft publish_buffer_limit to derive the hard Pub/Sub back-pressure
+// limit. Publishers are only parked in EnsureMemoryBudget once the per-thread subscriber memory
+// reaches soft_limit * kPubSubHardMultiplier.
+constexpr uint32_t kPubSubHardMultiplier = 4;
 
 void SendProtocolError(RespSrvParser::Result pres, SinkReplyBuilder* builder) {
   constexpr string_view res = "-ERR Protocol error: "sv;
@@ -468,8 +481,27 @@ struct QueueBackpressure {
   QueueBackpressure() {
   }
 
-  // Block until subscriber memory usage is below limit, can be called from any thread.
+  // Block until subscriber memory usage is below the hard limit, can be called from any thread.
   void EnsureBelowLimit();
+
+  // Hard Pub/Sub back-pressure limit (soft publish_buffer_limit times kPubSubHardMultiplier).
+  // EnsureBelowLimit only parks publishers once subscriber memory reaches this value.
+  size_t PubSubHardLimit() const {
+    return publish_buffer_limit * kPubSubHardMultiplier;
+  }
+
+  // True while the per-thread subscriber memory is above the soft publish_buffer_limit, i.e. a
+  // Pub/Sub back-pressure episode is in progress. Read on the owning I/O thread.
+  bool IsSubscriberSoftLimited() const {
+    return subscriber_bytes.load(memory_order_relaxed) > publish_buffer_limit;
+  }
+
+  // Accounts `mem` subscriber bytes on this thread and, on an at/below-soft -> above-soft
+  // transition, counts one soft_limit_crossing event. Called only on the owning I/O thread.
+  void AddSubscriberBytes(size_t mem);
+
+  // Releases `mem` subscriber bytes on this thread. Called only on the owning I/O thread.
+  void SubSubscriberBytes(size_t mem);
 
   // Checks if backpressure should be applied.
   // 'size' should be the total bytes currently consumed by all connections on this thread.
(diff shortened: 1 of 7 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -2303,7 +2375,74 @@ bool Connection::IsCurrentlyDispatching() const {
...
+  for (auto it = dispatch_q_.begin(); it != dispatch_q_.end();) {
+    if (it->IsPubMsg()) {
+      ++discarded_msgs;
+      discarded_bytes += it->UsedMemory();
+      UpdateDispatchStats(*it, false /* subtract */);
+      it = dispatch_q_.erase(it);
+    } else {
+      ++it;
+    }
+  }
```
dranikpg, line 2400: I think that regular closure would've done esentially the same, so this is just for logs?

dranikpg (approved):

src/facade/dragonfly_connection.cc
romange: no, not for logs.

Suppose you have a zombie connection that will close in 60s due to linux write timeout.
Now all the messages stacked up increase memory usage until this connection is closed.

Here we identify the connection we will close and remove these messages immediately - and reduce the memory pressure for this thread.

PR #7890, opened 2026-07-20
feat(tls): enable V2 io loop for TLS connections

glevkovich:
Enable the event-driven V2 io loop for TLS connections by removing the !is_tls_ gate in HandleRequests. ReadPendingInput now follows the new TlsSocket::TryRecv contract: a positive read may be short (keep looping), EAGAIN means a wake is coming (clear pending_input_ and park), and EBUSY (TLS-only) means another fiber holds the socket (keep the latch, retry later, don't surface as an error).

On thread migration, OnPostMigrateThread now forces a re-read (pending_input_ = true) for TLS: a fresh RegisterOnRecv only fires on new socket activity, so bytes already buffered in the TLS engine at hop time must be drained explicitly. TLS and multishot are mutually exclusive, so this only applies to TLS.

IoLoopV2 re-checks ShouldWakeIdle() after FlushReplies before parking, since the flush suspends the fiber and new input may arrive meanwhile.

Add tests: a TLS connection surviving cross-thread migration, and a full-duplex large-pipeline TLS stream that would deadlock the V2 loop if a TryRecv deferred during WRITE_IN_PROGRESS left buffered data unseen. Both run under V1 and V2.

bump helio to support the TlsSocket::TryRecv contract.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -3698,14 +3698,9 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
         return ec;
       }
 
-      // FlushReplies (above) suspends the fiber. While the fiber is suspended, pending_input may
-      // change and/or new bytes might arrive in io_buf_, so re-check ShouldWakeIdle() before
-      // parking. This re-check is an optimization, and not a must for correctness.
-      if (!ShouldWakeIdle()) {
-        fiber_park_spot_ = FiberParkSpot::kIdleAwait;
-        io_event_.await([this] { return ShouldWakeIdle(); });
-        fiber_park_spot_ = FiberParkSpot::kNone;
-      }
+      fiber_park_spot_ = FiberParkSpot::kIdleAwait;
+      io_event_.await([this] { return ShouldWakeIdle(); });
+      fiber_park_spot_ = FiberParkSpot::kNone;
     }
 
     phase_ = PROCESS;
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -922,6 +922,14 @@ void Connection::OnPostMigrateThread() {
     MaybeEnableRecvMultishot();
     socket_->RegisterOnRecv(
         [this](const FiberSocketBase::RecvNotification& n) { OnRecvNotification(n); });
+
+    // TLS and multishot are mutually exclusive: MaybeEnableRecvMultishot() only enables multishot
+    // when !is_tls_, so for a TLS connection it is a no-op above. We might still have bytes in the
+    // TLS engine, and a fresh RegisterOnRecv only fires on new socket activity, so force a read to
+    // drain those already-buffered bytes at hop time.
+    if (is_tls_) {
+      pending_input_ = true;
+    }
   }
 
   migration_in_process_ = false;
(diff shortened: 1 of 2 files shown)


glevkovich:
Note for Reviewers:
Helio is not bumped yet, but I uploaded this PR so I can merge the both back to back later. This PR should include Helio bump. Until Helio is not bumped,  CI runs would fail.

src/facade/dragonfly_connection.cc
```diff
@@ -3290,17 +3296,23 @@ void Connection::ReadPendingInput() {
 
   // Drain available socket data into io_buf_.
   io::MutableBytes buf = io_buf_.AppendBuffer();
-  // A recv call can return fewer bytes than requested even if the
-  // socket buffer actually contains enough data to satisfy the full request.
+  // pending_input_ lifecycle in this drain loop (see TlsSocket::TryRecv contract in tls_socket.h):
```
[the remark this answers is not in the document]
glevkovich: i need to bump helio. drop.

src/facade/dragonfly_connection.cc
```diff
@@ -3290,17 +3296,23 @@ void Connection::ReadPendingInput() {
...
+  // - Positive read: possibly short, not a "drained" signal - keep looping without clearing it. If
+  //   the loop exits because io_buf_ is full, it stays set for the next pass.
+  // - EAGAIN/EWOULDBLOCK: a wake is coming (drained kernel, or a TLS read deferred behind a
+  //   background write) - clear pending_input_ and park.
+  // - EBUSY (device_or_resource_busy, TLS only): another fiber holds the socket, no wake - keep it
+  //   latched and retry on a later pass (neither clear it nor set io_ec_).
+  // - Any other error: a real I/O failure - surface via io_ec_.
   while (!buf.empty()) {
     io::Result<size_t> res = socket_->TryRecv(buf);
```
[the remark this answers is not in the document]
glevkovich: known, differ for next task

src/facade/dragonfly_connection.cc
```diff
@@ -3290,17 +3296,23 @@ void Connection::ReadPendingInput() {
...
     if (!res) {
       auto ec = res.error();
-      // TryRecv is non-blocking: it returns EAGAIN/EWOULDBLOCK when nothing is ready.
-      if (ec == errc::resource_unavailable_try_again || ec == errc::operation_would_block)
+      if (ec == errc::resource_unavailable_try_again || ec == errc::operation_would_block) {
         pending_input_ = false;
-      else
+      } else if (ec != errc::device_or_resource_busy) {
```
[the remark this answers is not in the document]
glevkovich: known issue, differ for a separate task.

glevkovich:
Depends on https://github.com/romange/helio/pull/619

src/facade/dragonfly_connection.cc
```diff
@@ -3682,15 +3694,18 @@ variant<error_code, Connection::ParserStatus> Connection::IoLoopV2() {
...
       if (auto ec = FlushReplies(); ec) {
         return ec;
       }
 
-      fiber_park_spot_ = FiberParkSpot::kIdleAwait;
-      io_event_.await([this] { return ShouldWakeIdle(); });
-      fiber_park_spot_ = FiberParkSpot::kNone;
+      // FlushReplies (above) suspends the fiber. While the fiber is suspended, pending_input may
+      // change and/or new bytes might arrive in io_buf_, so re-check ShouldWakeIdle() before
+      // parking. This re-check is an optimization, and not a must for correctness.
```
romange: it's not really an optimization as `await` checks the condition before parking.

tests/dragonfly/connection_test.py
```diff
@@ -2291,6 +2292,53 @@ async def test_client_migrate(df_server: DflyInstance):
     assert resp == 1  # migrated successfully
 
 
+@dfly_multi_test_args(
+    {"proactor_threads": 4, "enable_resp_io_loop_v2": "false"},
```
romange, line 2296: should we have both running in matrix mode in regression runs at some point ?

src/facade/dragonfly_connection.cc
glevkovich: agree, redundent.

tests/dragonfly/connection_test.py
glevkovich: We already do, we have regression tests running true and false. There is a redundant run because of these 2 lines - each test case will run in regression tests twice (x2 false, x2 true). The above setting is to check the test in PR with both false and true. We do not have support for that and need explicit setting like this to run both loops in Pr level for specific tests.

romange (approved):

PR #7899, opened 2026-07-21
refactor: unify TTL limitation for different commands

### LINKED ISSUE #7893: make SADDEX support 1 << 26 member-TTL
SADDEX still uses 1 << 26, so the member-TTL entry points are not uniform yet. Switch it to the shared constant 1<< 28 and update the stale HGETEX comment/add a boundary test.

BorysTheDev:
fixes: #7893

Summary: Unifies SADDEX member TTL validation to use the shared kMaxExpireDeadlineSec cap (8.5 years), matching the hash-field TTL commands.

Tests: Adds a boundary unit test verifying TTL at the cap is accepted and cap+1 is rejected.


BorysTheDev:
augment review

vyavdoshenko (approved):

PR #7907, opened 2026-07-21
fix(search): unescape quoted tag values

### LINKED ISSUE #7906: Quoted TAG values with escaped characters no longer match since v1.39.0 (FT.SEARCH)
**Describe the bug**

Since v1.39.0, `FT.SEARCH` on a `TAG` field using the quoted dialect-2 syntax (`@field:{"..."}`) fails to match values containing escaped characters (`\`, `"`). The quoted value is looked up verbatim, so the backslash added to escape the query is never stripped, and the stored value is never found. Plain quoted values (no escapes) still match. This diverges from RediSearch, where a quoted tag query unescapes one layer before lookup.

I've traced the regression to a specific commit; details below.

**To Reproduce**

Index a value containing a backslash — e.g. a Windows/AD-style account name — and search for it with the quoted dialect-2 syntax:

```
docker run -d --name df -p 6379:6379 docker.dragonflydb.io/dragonflydb/dragonfly:v1.39.0
redis-cli FT.CREATE idx ON HASH PREFIX 1 doc: SCHEMA account TAG
redis-cli HSET doc:1 account 'ACME\jdoe'
redis-cli FT.SEARCH idx '@account:{"ACME\\jdoe"}' DIALECT 2
```

| Backend | Result |
|---|---|
| redis/redis-stack-server:latest | 1 match |
| dragonfly v1.38.1 | 1 match |
| dragonfly v1.39.0 | **0 matches** |

The unquoted form still works on v1.39.0, isolating the bug to the quoted-tag path:

```
redis-cli FT.SEARCH idx '@account:{ACME\\jdoe}'   DIALECT 2   → 1   (unquoted, works)
redis-cli FT.SEARCH idx '@account:{"ACME\\jdoe"}' DIALECT 2   → 0   (quoted, broken)
```

**Expected behavior**

`@account:{"ACME\\jdoe"}` should match the stored value `ACME\jdoe` — one layer of `\X` escapes stripped from the query before lookup — consistent with RediSearch and with the unquoted form `@account:{ACME\\jdoe}`, which still works.

**Environment**

- OS: any (reproduced in the official Docker image)
- Containerized?: Docker
- Dragonfly Version: v1.39.0 (regressed); v1.38.1 last good

**Root cause**

Bisected to #7438 (`fix(search): honor backslash escapes in TEXT tokenization`, commit `1181400f`; first bad release v1.39.0, v1.38.1 good). That change made `make_PhraseTok` pass quoted content verbatim and added compensating unescape logic on the TEXT path only. Quoted strings inside `{...}` reach the tag grammar as `PHRASE` tokens, and the rule hands the raw value straight through with no unescaping:

```cpp
// src/core/search/parser.y
tag_list_element:
  | PHRASE { ... $$ = AstTermNode(std::move(p.raw)); }   // p.raw is verbatim since #7438
```

So the tag lookup searches for the literal bytes `ACME\\jdoe` (two backslashes) instead of `ACME\jdoe`.

Byte-level trace (indexed `ACME\jdoe`, query `@account:{"ACME\\jdoe"}`):

| Version | Lookup bytes | Match |
|---|---|---|
| v1.38.1 | `ACME\jdoe` (CUnescape applied) | yes |
| v1.39.0 (post-#7438) | `ACME\\jdoe` (verbatim) | no |

The predecessor #7319 (which first routed quoted strings through `PHRASE`) did not break it — it still ran `absl::CUnescape` at that point; #7438 removing that unescape changed the lookup bytes. Tag *indexing* is unchanged — the value is stored identically on both versions. (This also inverts the single-backslash form: `@account:{"ACME\jdoe"}` matches on v1.39.0 but not v1.38.1, consistent with the verbatim-vs-unescape change.)

**Proposed fix**

Opened #7907 with a candidate fix — it runs the tag-path phrase value through the existing `UnescapeTerm` helper, mirroring the unquoted `make_Tag` path, with parser- and server-level tests. It's my personal take and throwaway; please feel free to close it and resolve the root cause however you see fit. Happy to carry the PR forward if you find it useful.

alexluong:
Fixes #7906.

**Problem**

Since v1.39.0, `FT.SEARCH` on a `TAG` field with the quoted dialect-2 syntax (`@field:{"..."}`) fails to match values containing escaped characters (`\`, `"`). The quoted value is looked up verbatim, so the query's escaping backslash is never stripped and the stored value can't be found:

```
FT.CREATE idx ON HASH PREFIX 1 doc: SCHEMA account TAG
HSET doc:1 account 'ACME\jdoe'
FT.SEARCH idx '@account:{"ACME\\jdoe"}' DIALECT 2   → 0 results   (should be 1)
```

The unquoted form `@account:{ACME\\jdoe}` still matches, which isolates it to the quoted-tag path.

**Cause**

#7438 made `make_PhraseTok` pass quoted content verbatim (moving `\X` handling into the TEXT phrase tokenizer) and added compensating unescape on the TEXT path only. Quoted strings inside `{...}` reach the tag grammar as `PHRASE` tokens, and `tag_list_element: PHRASE` handed `p.raw` straight to `AstTermNode` with no unescaping — so the lookup used the literal bytes `ACME\\jdoe` instead of `ACME\jdoe`. Tag indexing is unchanged; only the query-side lookup string diverged.

**Fix**

Run the tag-path phrase value through the existing `UnescapeTerm` helper before building the term node, mirroring the unquoted tag path (`make_Tag`), which already strips one layer of `\X` escapes. One behavioral line in `src/core/search/parser.y`.

I also removed the `DCHECK(!escaped)` at the end of `UnescapeTerm`. That assertion assumed input always came from the `{term_part}+` pattern, which never ends in a lone `\`. Quoted tag and wildcard values can, though — e.g. a Windows path `@t:{"C:\"}` or `w"a\"` — and on such input the helper hit `escaped == true` at the end and aborted in debug/ASAN builds (and silently dropped the backslash in release). It now drops the dangling backslash, matching what `make_Tag` and `SplitWithEscapes` already do. This also closes the same latent abort in `make_Wildcard`.

**Behavior change**

`UnescapeTerm` strips one layer of `\X` for *any* `X`, so a quoted query with an unrecognized escape now drops the backslash: `@t:{"ACME\jdoe"}` resolves to `ACMEjdoe` rather than passing `\j` through verbatim as v1.39 did. This is intentional — it matches Dragonfly's own unquoted `make_Tag` behavior, keeping the quoted and unquoted tag paths consistent. It differs from RediSearch, which passes unknown escapes through. A characterization test pins the `\X → X` rule so the semantics stay explicit. Happy to change the policy if you'd rather emulate RediSearch here.

**Tests**

- `SearchParserTest.QuotedTagEscapes` — AST-level: `\\`/`\"` resolve to their single char, an unrecognized `\X` drops the backslash, and a trailing lone `\` (`{"C:\"}` → `C:`, doubled `{"C:\\"}` → `C:\`) parses instead of aborting.
- `SearchParserTest.WildcardLex` — extended with `w"a\"` / `w'a\'` (trailing backslash → `a`), the wildcard side of the same abort.
- `SearchFamilyTest.QuotedTagEscapes` — end-to-end HSET/FT.SEARCH round-trip: backslash, quote, plain, value-with-space, an OR-list of quoted members, and a `C:\` value (the doubled query matches; the lone-backslash query returns nothing without crashing).
- `SearchFamilyTest.QuotedTagEscapesJson` — the same escaped query on an `ON JSON` tag index.
- `SearchFamilyTest.QuotedTagEscapesCaseSensitive` — exact-case escaped query matches, wrong case does not.

Everything fails before the change (or aborts, for the trailing-backslash cases) and passes after; the full `search_parser_test` and `search_family_test` suites remain green.

**Notes**

- `UnescapeTerm` is defined in `lexer.lex`; I added a forward declaration in `parser.y` to call it. If you'd prefer it promoted to a shared header instead of a cross-unit forward declaration, happy to adjust.

This PR is just my personal take on a fix — treat it as throwaway. Please feel free to close it and resolve the issue however you see fit. If you find it useful, I'm glad to carry it forward and address review feedback.

### DIFF
--- a/src/core/search/lexer.lex
+++ b/src/core/search/lexer.lex
@@ -154,9 +154,8 @@ string UnescapeTerm(string_view src) {
       res.push_back(c);
     }
   }
-  // The {term_part}+ pattern always pairs `\\` with a following char, so a
-  // trailing lone backslash is unreachable.
-  DCHECK(!escaped);
+  // A trailing lone backslash (reachable from quoted tag/wildcard values, e.g. a
+  // Windows path `C:\`) is dropped, matching make_Tag and SplitWithEscapes.
   return res;
 }
 
--- a/src/core/search/parser.y
+++ b/src/core/search/parser.y
@@ -68,6 +68,8 @@ using namespace std;
 
 uint32_t toUint32(string_view src);
 double toDouble(string_view src);
+// Defined in lexer.lex: strips backslashes from \X sequences.
+string UnescapeTerm(string_view src);
 
 }
 
@@ -460,12 +462,13 @@ tag_list:
 tag_list_element:
   TERM        { $$ = AstTermNode(std::move($1));   }
   | PHRASE {
-      /* Inside {..}, quoted strings are literal tag values. ~N slop is only meaningful for
-         phrases, so reject it here rather than silently dropping it. */
+      /* Inside {..}, quoted strings are literal tag values with one layer of `\X` escapes,
+         matching the unquoted tag path (make_Tag). ~N slop is only meaningful for phrases,
+         so reject it here rather than silently dropping it. */
       auto p = std::move($1);
       if (p.slop != 0)
         throw Parser::syntax_error(@$, "slop is not allowed in tag values");
-      $$ = AstTermNode(std::move(p.raw));
+      $$ = AstTermNode(UnescapeTerm(p.raw));
     }
   | PREFIX    { $$ = AstPrefixNode(std::move($1)); }
   | SUFFIX    { $$ = AstSuffixNode(std::move($1)); }
--- a/src/core/search/search_parser_test.cc
+++ b/src/core/search/search_parser_test.cc
@@ -622,6 +622,37 @@ TEST_F(SearchParserTest, PhraseSlopRejectedInTagContext) {
   EXPECT_NE(0, Parse("@t:{\"foo\"~2}"));
 }
 
+// Quoted tag values must process one layer of \X escapes, mirroring the unquoted tag path,
+// so values containing backslashes or quotes round-trip between HSET and FT.SEARCH.
+TEST_F(SearchParserTest, QuotedTagEscapes) {
+  auto tag_affix = [this](const string& query) -> string {
+    EXPECT_EQ(0, Parse(query));
+    AstExpr e = query_driver_.Take();
+    EXPECT_TRUE(std::holds_alternative<AstFieldNode>(e));
+    const AstNode& tags = *std::get<AstFieldNode>(e).node;
+    EXPECT_TRUE(std::holds_alternative<AstTagsNode>(tags));
+    const auto& tn = std::get<AstTagsNode>(tags);
+    EXPECT_EQ(tn.tags.size(), 1u);
+    EXPECT_TRUE(std::holds_alternative<AstTermNode>(tn.tags[0]));
+    return std::get<AstTermNode>(tn.tags[0]).affix;
+  };
+
+  // Recognized escapes \\ and \" resolve to their single-character values.
+  EXPECT_EQ(tag_affix(R"(@t:{"tnt\\backslash"})"), R"(tnt\backslash)");
+  EXPECT_EQ(tag_affix(R"(@t:{"tnt\"quote"})"), "tnt\"quote");
+
+  // Characterization: one layer of \X is stripped for any X, mirroring the unquoted make_Tag
+  // path, so an unrecognized escape drops its backslash rather than passing through verbatim.
+  // This is an intentional policy change from v1.39's verbatim behavior; pinning it keeps a
+  // future reader from "restoring" the old semantics.
+  EXPECT_EQ(tag_affix(R"(@t:{"ACME\jdoe"})"), "ACMEjdoe");
+
+  // A trailing lone backslash (e.g. a Windows path) must not abort — it previously tripped the
+  // DCHECK in UnescapeTerm — and drops the dangling backslash; the doubled form is a literal `\`.
+  EXPECT_EQ(tag_affix(R"(@t:{"C:\"})"), "C:");
+  EXPECT_EQ(tag_affix(R"(@t:{"C:\\"})"), R"(C:\)");
+}
+
 TEST_F(SearchParserTest, PhraseParse) {
   // Top-level phrase.
   ASSERT_EQ(0, Parse("\"machine learning\""));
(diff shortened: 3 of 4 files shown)


vyavdoshenko (changes requested):
@alexluong 
Thanks for your contribution, but there are a few points to fix.

**Blocker**:

The PR's `AstTermNode(UnescapeTerm(p.raw))` is the first caller to feed quoted-tag content into `UnescapeTerm`, whose `DCHECK(!escaped)` only holds for `{term_part}+` inputs. A quoted tag ending in a lone backslash (`@tag:{"C:\"}`, a Windows path, exactly the #7906 domain) reaches it with `escaped == true` and aborts in debug/ASAN builds (silently drops the backslash in release, so the lookup is wrong). The fix is to remove the `DCHECK` and drop the dangling backslash, matching what `make_Tag` and `SplitWithEscapes` already do; this also closes the same latent abort in `make_Wildcard` (`w"a\"`). Without it, PR CI stays green (its tests use only well-formed inputs) while the abort ships.

**Behavior change to call out**:

The fix also makes single-backslash quoted queries strip the backslash (`@t:{"ACME\jdoe"}` becomes `ACMEjdoe`), which is correct because it matches Dragonfly's own unquoted `make_Tag` path, but it is an observable change from v1.39's verbatim behavior and must be mentioned in the PR description. Because it is an intentional policy rather than an accident, it needs a characterization test that pins `\X` becomes `X`, so a future reader does not "restore" the old semantics.

**Tests to add (which scenarios)**:

- Trailing lone backslash (guards the blocker): `@t:{"C:\"}` must not abort and resolves to `C:`, while the doubled `@t:{"C:\\"}` resolves to a literal `C:\`.
- Wildcard trailing backslash: `w"a\"` and `w'a\'` in the existing `WildcardLex` test (shared helper).
- Single-backslash characterization: `@t:{"ACME\jdoe"}` resolves to `ACMEjdoe`.
- OR-list with quoted members: `@tags:{"a\\b"|"c"}` matches both.
- Value with a space: `@tags:{"foo bar"}` matches a tag stored as `foo bar`.
- JSON-backed tag field: same escaped query on an `ON JSON` index.
- Case-sensitive tag: exact-case escaped query matches; wrong case does not.

alexluong:
Thanks for the careful review — good catch on the abort. Pushed an update addressing all of it:

**Blocker (DCHECK):** Removed `DCHECK(!escaped)` from `UnescapeTerm` and made it drop a trailing lone backslash, matching `make_Tag` and `SplitWithEscapes`. So `@t:{"C:\"}` (and `w"a\"`) no longer abort in debug/ASAN — they unescape to `C:` / `a`. This also closes the pre-existing latent abort on the `make_Wildcard` path.

**Behavior change:** Called it out explicitly in the PR description now, and added a characterization test pinning `\X → X` (`@t:{"ACME\jdoe"}` → `ACMEjdoe`) so the intentional divergence from RediSearch's pass-through isn't "fixed" back later.

**Tests added:** trailing lone backslash (`{"C:\"}`→`C:`, `{"C:\\"}`→`C:\`); wildcard trailing backslash (`w"a\"`, `w'a\'`) in `WildcardLex`; single-backslash characterization; OR-list of quoted members (`{"a\\b"|"c"}`); value with a space (`{"foo bar"}`); JSON-backed tag index; and a case-sensitive tag (exact case matches, wrong case doesn't).

Built in `ubuntu-dev:24` — full `search_parser_test` (28) and `search_family_test` (328) suites green, and the trailing-backslash cases pass without aborting.

Still very much your call on how to resolve #7906 — happy to keep iterating if this is useful.

vyavdoshenko:
@alexluong 
Please follow the instructions: https://github.com/dragonflydb/dragonfly/blob/main/CONTRIBUTING.md
Your commit must be signed for merge.

alexluong:
@vyavdoshenko done, the commit should be signed now. Let me know if there's anything else!

vyavdoshenko (approved):
Nice work!

PR #7909, opened 2026-07-22
fix(facade): preserve pub/sub order on unsubscribe (v1 IO loop)

glevkovich:
Since v1.38 the async_dispatch_quota anti-starvation logic in AsyncFiber
could run a pipelined command such as UNSUBSCRIBE ahead of pub/sub
messages that were queued before it, once the quota was reached. Running
UNSUBSCRIBE early sets subscriptions to 0, so the still-queued messages
are discarded and the subscriber loses messages published before it
unsubscribed.

Gate the quota-based pipeline preference on FIFO order: only prefer the
pipeline when its head command is strictly older than the oldest queued
control message, so it never jumps ahead of pub/sub messages queued
before it.

dispatch_q_ is FIFO by dispatch_cycle except for checkpoints, which
SendAsync inserts at the front. A newer leading checkpoint would hide
older queued PubMessages, so when the head is a checkpoint we defer the
pipeline preference for one iteration: the checkpoint is processed first
and the decision is re-evaluated against the real oldest message.

Add regression test test_pubsub_unsubscribe_message_loss.

The V2 io loop (IoLoopV2) fix will be done in a separate commit; the
test is skipped under V2 until then.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -2234,16 +2234,23 @@ void Connection::AsyncFiber() {
         }
       }
 
-      // We prioritize pipeline execution over the admin queue in two distinct cases (Pipeline queue
-      // must be non-empty for both cases):
-      // 1. A migration is requested (Redis only), but we must drain the existing
-      // pipeline first.
-      // 2.  The dispatch quota was reached, forcing a pipeline execution to prevent
-      // starvation.
+      // Prefer the pipeline over the admin queue when case 1 OR case 2 holds (pipeline non-empty):
+      // 1. Migration requested (Redis only): drain the pipeline first.
+      // 2. Anti-starvation, requires 2a AND 2b AND 2c to be all true:
+      //    2a. the dispatch quota was reached, and
+      //    2b. the pipeline head is STRICTLY older than the dispatch queue head (so UNSUBSCRIBE
+      //        never jumps ahead of pub/sub messages queued before it), and
+      //    2c. the head is not a checkpoint. Checkpoints are front-inserted so they can hide older
+      //        PubMessages; process the (cheap) checkpoint first and re-evaluate next round.
       bool prefer_pipeline_execution = false;
       if (parsed_head_ != nullptr) {
-        prefer_pipeline_execution =
-            quota_reached || (is_migration_req && (protocol_ == Protocol::REDIS));
+        const bool drain_for_migration = is_migration_req && (protocol_ == Protocol::REDIS);
+        const bool head_preserves_order =
+            dispatch_q_.empty() ||
+            (!dispatch_q_.front().IsCheckPoint() &&
+             parsed_head_->parsed_cycle < dispatch_q_.front().dispatch_cycle);
+        const bool quota_reached_while_preserves_order = quota_reached && head_preserves_order;
+        prefer_pipeline_execution = drain_for_migration || quota_reached_while_preserves_order;
       }
       if (dispatch_q_.empty() || prefer_pipeline_execution) {  // 2. Process pipeline Queue
         VLOG_IF(2, prefer_pipeline_execution)
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -2236,12 +2236,12 @@ void Connection::AsyncFiber() {
 
       // Prefer the pipeline over the admin queue when case 1 OR case 2 holds (pipeline non-empty):
       // 1. Migration requested (Redis only): drain the pipeline first.
-      // 2. Anti-starvation, requires 2a AND 2b AND 2c to be all true:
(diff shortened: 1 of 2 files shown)


src/facade/dragonfly_connection.cc
```diff
@@ -2234,16 +2234,20 @@ void Connection::AsyncFiber() {
...
+      // 2.  Quota fired to prevent starvation - requires BOTH 2a and 2b to be true:
+      //     2a. the dispatch quota  was reached (quota_reached=true), and
+      //     2b. the head command is STRICTLY older than the control-queue head, so (for example) it
+      //     never jumps ahead of pub/sub messages queued before it (UNSUBSCRIBE would drop them).
       bool prefer_pipeline_execution = false;
       if (parsed_head_ != nullptr) {
-        prefer_pipeline_execution =
-            quota_reached || (is_migration_req && (protocol_ == Protocol::REDIS));
+        const bool drain_for_migration = is_migration_req && (protocol_ == Protocol::REDIS);
+        const bool quota_reached_while_preserves_order =
```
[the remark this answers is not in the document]
glevkovich: agree, and fixing.

src/facade/dragonfly_connection.cc
```diff
@@ -2234,16 +2234,23 @@ void Connection::AsyncFiber() {
...
       }
 
-      // We prioritize pipeline execution over the admin queue in two distinct cases (Pipeline queue
-      // must be non-empty for both cases):
-      // 1. A migration is requested (Redis only), but we must drain the existing
-      // pipeline first.
-      // 2.  The dispatch quota was reached, forcing a pipeline execution to prevent
-      // starvation.
+      // Prefer the pipeline over the admin queue when case 1 OR case 2 holds (pipeline non-empty):
+      // 1. Migration requested (Redis only): drain the pipeline first.
```
romange, line 2238: what does it mean "redis only" ?
```suggestion
      // 1. Migration requested (RESP only): drain the pipeline first.
```

src/facade/dragonfly_connection.cc
```diff
@@ -2234,16 +2234,23 @@ void Connection::AsyncFiber() {
...
+      // 2. Anti-starvation, requires 2a AND 2b AND 2c to be all true:
```
romange:
```suggestion
      // 2. Anti-starvation, requires all conditions below to be true:
```

src/facade/dragonfly_connection.cc
glevkovich: That was also the original condition:
`(is_migration_req && (protocol_ == Protocol::REDIS))`

src/facade/dragonfly_connection.cc

src/facade/dragonfly_connection.cc
```diff
@@ -2234,16 +2234,23 @@ void Connection::AsyncFiber() {
...
+      //    2a. the dispatch quota was reached, and
+      //    2b. the pipeline head is STRICTLY older than the dispatch queue head (so UNSUBSCRIBE
+      //        never jumps ahead of pub/sub messages queued before it), and
+      //    2c. the head is not a checkpoint. Checkpoints are front-inserted so they can hide older
+      //        PubMessages; process the (cheap) checkpoint first and re-evaluate next round.
```
romange:
```suggestion
      //    2c. the head is not a checkpoint. Checkpoints are front-inserted, taking precedence over older PubMessages.
```

tests/dragonfly/connection_test.py
```diff
@@ -677,6 +677,89 @@ async def publisher():
...
+    publisher = async_client
+    channel = "ordering-channel"
+    # Well above async_dispatch_quota (and the socket buffer) so the surplus piles up
+    # in the server-side dispatch queue and the anti-starvation path triggers.
+    message_count = 60
+    payload = b"x" * 100_000  # large enough to fill the tiny receive window quickly
+    message_token = b"$7\r\nmessage\r\n"
+    unsubscribe_token = b"$11\r\nunsubscribe\r\n"
+
+    loop = asyncio.get_running_loop()
```
romange, line 706: can we avoid code duplication? there is an utility function `_open_stuck_subscriber` - why not use it?

src/facade/dragonfly_connection.cc

tests/dragonfly/connection_test.py
glevkovich: Agree, I didn't pay attention to this one. I'll use use it.

glevkovich: Update:
`_open_stuck_subscriber` cant be used here. It's used in a case where we open a subscriber that is "getting stuck" and doesn't read till the end of the test. But we need to continue reading at some point. Changing it to match the use case doesn't make sense for a single test.

romange (approved):

PR #7912, opened 2026-07-22
feat(rdb): Skip keys with unsupported module types instead of failing

### LINKED ISSUE #7848: RDB load aborts on unsupported Redis module types (TopK, etc.) instead of skipping
## Summary

When replicating from a Redis Stack instance (via `REPLICAOF`), Dragonfly aborts the entire RDB load upon encountering an unsupported Redis module type. This prevents migration from any Redis Stack instance that contains even a single unsupported module key.

## Reproduction

1. Run Redis Stack (e.g. `redis-stack-server`) with TopK keys:
   ```
   TOPK.RESERVE mykey 50 2000 7 0.925
   TOPK.ADD mykey item1 item2 item3
   ```
2. Run Dragonfly and replicate:
   ```
   REPLICAOF <redis-host> <redis-port>
   ```
3. Dragonfly logs show:
   ```
   E0713 14:22:14.106398  rdb_load.cc:1809] Unsupported module: TopK-TYPE
   W0713 14:22:14.107629  replica.cc:291] Error syncing [...] Internal error when loading RDB file
   ```
4. All keys serialized after the TopK key in the RDB stream are lost.

## Root Cause

In `rdb_load.cc`, all `RDB_TYPE_MODULE_2` entries are routed through `ReadRedisJson()` (line 1274). This function only handles `ReJSON-RL` — any other module name hits the `Unsupported module` error and returns `errc::unsupported_operation`, which aborts the entire RDB load.

Dragonfly *does* have native support for TopK (`RDB_TYPE_TOPK` / `ReadTOPK()`), SBF, and CMS via their own RDB type codes. However, Redis Stack serializes these as generic `RDB_TYPE_MODULE_2` with module-specific encoding, not the native type codes that Dragonfly uses in its own snapshots.

Meanwhile, `RDB_OPCODE_MODULE_AUX` entries (module auxiliary data) are already gracefully skipped via `SkipModuleData()` (line 2346). The same approach could be applied to unsupported `RDB_TYPE_MODULE_2` key data.

## Expected Behavior

When encountering an unsupported module type during RDB load:
1. Log a warning (e.g. `WARNING: Skipping unsupported module key: TopK-TYPE`)
2. Skip the module key data using the existing `SkipModuleData()` infrastructure
3. Continue loading the rest of the RDB stream

This would allow replication from Redis Stack instances, with unsupported module keys gracefully excluded.

## Bonus

Since Dragonfly already has native TopK/SBF/CMS readers, it may also be possible to deserialize from the Redis module encoding into Dragonfly's native format — but graceful skipping is the minimum viable fix.

## Environment

- Dragonfly: df-v1.39.0
- Redis Stack: redis_version 7.4.7 (Redis Stack)
- OS: Linux aarch64 (Fedora 43)
- Container runtime: Podman

mkaruza:
RDB_TYPE_MODULE_2 values for module types we don't support were aborting the whole load. 
Detect unsupported module types, skip their value bytes via the generic module opcode format, 
and skip just that key instead.

- Split SkipModuleData into SkipModuleAuxData (aux opcode, has a leading timestamp field) 
  and SkipModuleKeyData (per-key module values).
- Rename ReadRedisJson to ReadRedisModule2 since it handles RDB_TYPE_MODULE_2 type.

Closes #7848

### DIFF
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -3234,7 +3234,7 @@ io::Result<bool> RdbLoader::ReadAndDispatchObject(int object_type, std::string& 
   // Read a part of the object. Updates remaining items
   if (auto ec = ReadObj(object_type, &item->val); ec) {
     if (ec == RdbError(errc::feature_not_supported)) {
-      LOG(WARNING) << "Skipping unsupported key '" << key;
+      LOG(WARNING) << "Skipping unsupported key: " << key;
       pending_read_ = {};
       return true;
     }
--- a/src/server/rdb_load.cc
+++ b/src/server/rdb_load.cc
@@ -1301,7 +1301,7 @@ error_code RdbLoaderBase::ReadObj(int rdbtype, OpaqueObj* dest) {
       }
       break;
     case RDB_TYPE_MODULE_2:
-      iores = ReadRedisJson();
+      iores = ReadRedisModule2();
       break;
     case RDB_TYPE_SBF:
       iores = ReadSBF();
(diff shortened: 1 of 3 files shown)


mkaruza:
Tested manually by loading RedisBloom module and created TOPK/TDIGEST/CMS keys than replicated

```
W20260722 11:44:18.727154 62032 rdb_load.cc:2928] Unrecognized RDB AUX field: 'aof-base'
W20260722 11:44:18.727380 62032 rdb_load.cc:1842] Skipping key with unsupported module type: CMSk-TYPE
W20260722 11:44:18.727393 62032 rdb_load.cc:3237] Skipping unsupported key 'bikes:profit
W20260722 11:44:18.727399 62032 rdb_load.cc:1842] Skipping key with unsupported module type: TopK-TYPE
W20260722 11:44:18.727411 62032 rdb_load.cc:3237] Skipping unsupported key 'bikes:keywords
W20260722 11:44:18.727416 62032 rdb_load.cc:1842] Skipping key with unsupported module type: TDIS-TYPE
W20260722 11:44:18.727423 62032 rdb_load.cc:3237] Skipping unsupported key 'bikes:sales
```

abhijat:
The changes themselves look ok but can you add a unit test in rdb_test.cc?  It should be possible to build the representation using the helpers there. If its not possible then its fine.

mkaruza:
> The changes themselves look ok but can you add a unit test in rdb_test.cc? It should be possible to build the representation using the helpers there. If its not possible then its fine.

@abhijat added

abhijat (approved):

PR #7915, opened 2026-07-22
fix(server): emit expired keyspace events when the new expiration is already in the past

vyavdoshenko:
Deleting a key because its newly-set expiration is already in the past emitted no `__keyevent@<db>__:expired` notification, so subscribers waited forever (found via the upstream Valkey pubsub TCL suite: "expired event (Expiration time is already expired)" hangs). There are exactly two silent-delete paths: `SET` with a past `EXAT/PXAT` (incl. memcached SET/ADD/REPLACE) and `DbSlice::UpdateExpire` (EXPIRE family, GETEX, GAT). Both now emit the event after journaling (mutation -> journal -> publish). Fixing this surfaced several adjacent bugs in the same paths and in the config plumbing that gates the events - all fixed here to keep the fix correct.

Changes:

**Core fix**
- New `DbSlice::SendExpiredKeyEvent()`, called by all past-expiry delete paths after their journal record; `UpdateExpire` delete branch also counts `expired_keys` stats.

**Correctness bugs exposed in the same paths**
- `SET` with a past expiry short-circuited parsing: NX/XX/GET were ignored (`SET k v NX PXAT <past>` deleted the key and replied OK), trailing garbage was accepted, `PXAT == now` made the key persistent, and mc REPLACE-on-missing reported STORED. The early-return is gone; conditions go through the regular SetCmd machinery; the delete reuses `UpdateExpire`.
- `SET` is NO_AUTOJOURNAL and the past-expiry delete was not journaled -> replicas kept the key forever. It now journals `DEL` (same for GETEX, which journaled a past `PEXPIREAT`).
- Pre-existing: memcached `GAT 0` persists the key on master but journaled `PEXPIREAT key 0`, so replicas deleted it. It now journals `PERSIST`.

**Config plumbing (gates the events, `notify_keyspace_events`)**
- `ConfigRegistry::Set` rolls back the flag value when the validation callback rejects it; previously the rejected value stayed visible via CONFIG GET and raw flag reads.
- New DbSlices inherit a validated process-global default instead of re-reading the raw flag, closing a race where a namespace created during CONFIG SET could read an unvalidated value and kill the process (the ctor also demanded exact "Ex" while the validator accepts "EX").
- `CONFIG SET notify_keyspace_events` now updates every namespace, not only the default one (new `Namespaces::ForEachDbSliceOnShard`).
- Startup validation moved to dfly_main before the pidfile/listeners exist; an invalid flag previously exited mid-init with code 0, leaving a stale pidfile and unix socket.

**TLS (follow-ups from the registry rollback making callback side effects observable)**
- The `tls` callback builds the SSL_CTX once before touching listeners (new infallible `Listener::ApplyTlsCtx`), so a bad cert can no longer leave listeners split between old and new TLS state.
- TLS memory accounting is now a global atomic: contexts are allocated on one thread and freed on another, so the summed per-thread counters wrapped `tls_bytes` to ~UINT64_MAX after reload.

Tests:

- Unit: expired events for EXPIRE/SET/GETEX/GAT/mc SET/REPLACE; SET past-expiry condition semantics; config rollback, all-namespaces propagation, hermetic restores of prior config.
- Pytest: `GAT 0` persist replication, past-expiry SET DEL replication, `tls_bytes` regression with main+admin listeners; TLS suite passes end-to-end.


src/facade/dragonfly_listener.cc
```diff
@@ -283,8 +291,33 @@ std::shared_ptr<const TlsCertInfo> Listener::GetTlsCertInfo() const {
...
+  } else {
+    ctx_ = nullptr;
+  }
+  {
+    util::fb2::LockGuard lk(tls_cert_info_mu_);
+    tls_cert_info_ = std::move(info);
+  }
+  if (prev_ctx) {
+    SSL_CTX_free(prev_ctx);
+  }
```
BorysTheDev, line 315: you do it anycase why do you postpone it

src/facade/dragonfly_listener.cc
```diff
@@ -283,8 +291,33 @@ std::shared_ptr<const TlsCertInfo> Listener::GetTlsCertInfo() const {
...
+  if (ctx && (!IsPrivilegedInterface() || tls_on_privileged_port)) {
+    SSL_CTX_up_ref(ctx);
+    ctx_ = ctx;
+    auto parsed = ParseTlsCertInfo(SSL_CTX_get0_certificate(ctx));
+    if (parsed) {
+      info = std::make_shared<const TlsCertInfo>(*parsed);
+    }
+  } else {
+    ctx_ = nullptr;
+  }
```
BorysTheDev, line 308: you can do it unconditionally

src/server/config_registry.cc
```diff
@@ -53,13 +53,22 @@ auto ConfigRegistry::Set(string_view config_name, string_view value) -> SetResul
...
   CHECK(flag) << config_name;
+  const string prev_value = flag->CurrentValue();
   if (string error; !flag->ParseFrom(value, &error)) {
     LOG(WARNING) << error;
     return SetResult::INVALID;
   }
 
   bool success = !cb || cb(*flag);
-  return success ? SetResult::OK : SetResult::INVALID;
+  if (!success) {
```
BorysTheDev, line 63: you don't need success variable

src/server/db_slice.cc
```diff
@@ -1211,16 +1210,28 @@ OpResult<int64_t> DbSlice::UpdateExpire(const Context& cntx, Iterator prime_it,
   if (!satisfied)
     return OpStatus::SKIPPED;
 
-  // If we update and the new value is already expired, delete the key
+  // If we update and the new value is already expired, delete the key. The caller is responsible
+  // for emitting the expired keyspace event (via SendExpiredKeyEvent) AFTER journaling the
+  // deletion, so a publish that suspends or fails cannot separate the mutation from its journal
+  // record.
```
BorysTheDev, line 1216: comment looks redundant

src/server/main_service.cc
```diff
@@ -1091,19 +1092,36 @@ void Service::Init(util::AcceptServer* acceptor, std::vector<facade::Listener*>
   config_registry.RegisterMutable("search_query_string_bytes");
 #endif
 
+  // The default inherited by DbSlices is set here, before any of them exists: slices use the
+  // validated state instead of re-reading the raw flag, which transiently holds unvalidated
+  // input while CONFIG SET runs (a racing namespace creation must never observe a rejected
+  // value). dfly_main validates the flag before the pidfile/listeners exist; this exit is a
+  // backstop for other embeddings.
```
BorysTheDev, line 1099: I'm not sure that we need this description

src/server/metrics.cc
```diff
@@ -852,7 +852,11 @@ void Metrics::InitFromThread(Namespace* ns, const CommandRegistry* registry,
       replication_stats = dfly_cmd->GetReplicationMemoryStats(shard);
   }
 
-  tls_bytes = Listener::TLSUsedMemoryThreadLocal();
+  // The TLS counter is process-wide (contexts are allocated and freed on different threads),
+  // so read it once instead of summing it per thread.
```
BorysTheDev, line 856: if you need such a comment, rename the method

src/server/server_family.cc
```diff
@@ -1228,12 +1238,25 @@ void ServerFamily::Init(util::AcceptServer* acceptor, std::vector<facade::Listen
...
         return false;
       }
     }
+    for (facade::Listener* l : listeners_) {
+      // Must reconfigure in the listener proactor to avoid a race.
+      l->socket()->proactor()->Await([l, ctx] { l->ApplyTlsCtx(ctx); });
+    }
+    if (ctx) {
+      SSL_CTX_free(ctx);
+    }
```
BorysTheDev, line 1258: absl::Cleanup

BorysTheDev: also you don't need if to call SSL_CTX_free

src/server/string_family.cc
```diff
@@ -112,6 +112,9 @@ class SetCmd {
     uint16_t flags = SET_ALWAYS;
     uint32_t memcache_flags = 0;
     uint64_t expire_after_ms = 0;  // Relative value based on now. 0 means no expiration.
+    // EXAT/PXAT resolved to a time that already elapsed: delete instead of store (after the
+    // NX/XX/GET conditions are honored).
+    bool expire_in_past = false;
```
BorysTheDev, line 117: can you use ExpireParams instead of time related fields?

src/server/string_family.cc
```diff
@@ -922,8 +935,21 @@ OpStatus SetCmd::Set(const SetParams& params, string_view key, string_view value
...
     }
+  } else if (params.expire_in_past) {
+    // Unconditional set whose expiration already elapsed: delete the existing key (if any)
+    // instead of storing the value.
+    auto find_res = db_slice.FindMutable(op_args_.db_cntx, key);
+    if (auto status = CachePrevIfNeeded(params, find_res.it); status != OpStatus::OK)
+      return status;
+    if (!IsValid(find_res.it))
+      return OpStatus::OK;
+    return DeleteExpiredKey(key, &find_res);
```
BorysTheDev, line 952: expire_in_past logic is a corner case and always do early return, I suggest create a separate method with expire_in_past logic and do this check before other logic to reduce complexity

src/server/string_family.cc
```diff
@@ -1064,6 +1090,24 @@ void SetCmd::RecordJournal(const SetParams& params, string_view key, string_view
...
+  auto& db_slice = op_args_.GetDbSlice();
+  it_upd->post_updater.Run();  // finalize memory accounting before the delete
+
+  DbSlice::ExpireParams past{TimeUnit::MSEC, -1, op_args_.db_cntx.time_now_ms, /*cap=*/true};
+  auto res = db_slice.UpdateExpire(op_args_.db_cntx, it_upd->it, past);
+  RETURN_ON_BAD_STATUS(res);
+  DCHECK_EQ(*res, -1);
+
+  // SET is NO_AUTOJOURNAL: without an explicit record the replica never learns about this
+  // deletion and keeps the old value forever.
```
BorysTheDev, line 1103: you don't need this comment

src/server/string_family.cc
```diff
@@ -1107,11 +1150,13 @@ std::variant<SetCmd::SetParams, facade::ErrorReply, NegativeExpire> ParseSetPara
...
-        return NegativeExpire{};
-
-      tie(sparams.expire_after_ms, ignore) = expiry.Calculate(now_ms, true);
+      // An expiration that already elapsed deletes instead of storing. Keep parsing so the
+      // remaining options (NX/XX/GET, syntax validation) are honored.
+      if (rel_ms <= 0) {
+        sparams.expire_in_past = true;
+      } else {
+        tie(sparams.expire_after_ms, ignore) = expiry.Calculate(now_ms, true);
+      }
```
BorysTheDev, line 1159: if expiry becomes part of sparams, code will be cleaner

src/facade/dragonfly_listener.cc
```diff
@@ -109,29 +109,37 @@ bool ConfigureKeepAlive(int fd) {
...
   uint64_t refused_conn_maxclients_reached_cnt = 0;
 };
 
 thread_local ListenerStats listener_tl_stats;
 atomic_int ssl_init_refcount = 0;
 
+// Global: TLS contexts are built once and shared across listeners, so OpenSSL objects are
+// routinely allocated on one thread and freed on another. Per-thread counters would go
+// negative on the freeing threads and wrap the summed metric.
+atomic_size_t tls_allocated_bytes{0};
```
romange, line 121: did you actually encountered this?

src/server/metrics.cc
```diff
@@ -852,7 +852,11 @@ void Metrics::InitFromThread(Namespace* ns, const CommandRegistry* registry,
...
+  if (proactor_index == 0) {
```
romange, line 857: it should move to GetMetrics

romange (changes requested):
This PR is bug and covers several unrelated bugs. Lets start with everything that is not keyspace events in a separate PR.

PR #7917, opened 2026-07-23
fix: name idle tasks and evict stalled Pub/Sub subscribers

romange:
Name recurring idle tasks in diagnostics. Evict slow Pub/Sub subscribers at a per-connection queue threshold and ensure the regression test cleans up on failure.

### DIFF
--- a/src/facade/dragonfly_connection.cc
+++ b/src/facade/dragonfly_connection.cc
@@ -89,8 +89,9 @@ ABSL_FLAG(strings::MemoryBytesFlag, publish_buffer_limit, 196_MB,
 
 ABSL_FLAG(uint32_t, pubsub_slow_subscriber_timeout_ms, 0,
           "If a subscriber connection keeps a Pub/Sub socket write blocked for at least this many "
-          "milliseconds while the per-IO-thread subscriber memory is above the soft "
-          "publish_buffer_limit, Dragonfly closes that subscriber to release the back-pressure. "
+          "milliseconds, Dragonfly closes it when either its queued Pub/Sub memory reaches "
+          "one-sixteenth of publish_buffer_limit or the per-IO-thread subscriber memory is "
+          "above publish_buffer_limit. "
           "0 disables this slow-subscriber protection. Startup-only, like publish_buffer_limit.");
 
 ABSL_FLAG(uint32_t, pipeline_squash, 1,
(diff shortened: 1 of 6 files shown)


romange:
Motivation -  saw warnings like below, showing that a single zombie subscriber was closed after a long time (blocked_send_usec=1506207) even the threshold is 100ms. The reason for that is that it probably took lots of time to fill in its queue with 192MB of messages.

```
W20260722 23:01:32.782955  1349 dragonfly_connection.cc:2421] Closing slow Pub/Sub subscriber {id=2569, phase=1, dispatch(s/a)=0 1, closing=0, df:joinable=1, dq:size=0, pq:parsed_cmd_q_len=4, pq:is_empty=0, state= age=14308 idle=6} tid=
0 publish_buffer_limit=205520896 thread_subscriber_bytes=27345 blocked_send_usec=1506207 discarded_entries=113829 discarded_bytes=205496750                                                                                                 
```

src/facade/dragonfly_connection.cc
```diff
@@ -2448,14 +2450,20 @@ void Connection::SendPubMessageAsync(PubMessage msg) {
...
-  uint64_t timeout_cycles = pubsub_slow_subscriber_timeout_cycles_cached;
-  if (timeout_cycles > 0 && GetQueueBackpressure().IsSubscriberSoftLimited() &&
-      IsAsyncOpOverdue(timeout_cycles)) {
-    RequestPubsubClose();
-    return;  // discard the message that triggered the close
+  // Close a subscriber only after its active Pub/Sub send has been blocked continuously past the
+  // configured deadline. Either its own queued Pub/Sub data has reached the per-connection limit,
+  // or the thread is in a soft-limit back-pressure episode. The local limit prevents one blocked
+  // connection from consuming the whole thread-wide budget before it becomes eligible for eviction.
+  QueueBackpressure& qbp = GetQueueBackpressure();
```
glevkovich: nit: better putt this line inside the condition.

src/facade/dragonfly_connection.cc
```diff
@@ -2448,14 +2450,20 @@ void Connection::SendPubMessageAsync(PubMessage msg) {
...
+  if (IsAsyncOpOverdue()) {
+    // A single blocked subscriber must not consume the whole per-thread Pub/Sub budget before the
+    // slow-subscriber policy can evict it.
+    constexpr uint32_t kPubSubConnectionLimitDivisor = 16;
+    size_t conn_soft_limit = qbp.publish_buffer_limit / kPubSubConnectionLimitDivisor;
```
glevkovich, line 2462: nit:  would assert here, if someone do some testing and set small value for publish_buffer_limit < 16 - or if in the future kPubSubConnectionLimitDivisor becomes larger -> conn_soft_limit will be zero and the condition on the next line will always be true. 

or make it:

`size_t conn_soft_limit = std::max<size_t>(1, qbp.publish_buffer_limit / kPubSubConnectionLimitDivisor);`

glevkovich (approved):
LGTM, 2 nits.

PR #7918, opened 2026-07-23
fix(pytest): Drop flaky soft_limit assert and skip test for IoLoopV2

mkaruza:
* The test times out on IoLoopV2.
* soft_limit assert is racy and depends on scheduling. The test can produce
  more than 32MB of data within the 100ms timeout.


mkaruza:
Closing. Fixed.

PR #7927, opened 2026-07-23
fix(squash): avoid mutable cross-thread access on shared reply builder

dranikpg:
1. `reply_builder->last_error_` is a mutable variable that is reset and read in InvokeCmd
2. In squashing, Regular commands use their own local context and a capturing reply builder
3. In squashing, Coroutine commands do not use the reply builder at all "in theory" as they keep the reply inside themselves - so we never exchange it and they keep the connection ones in the context
4. As a result, any access to the connection reply builder is unsafe as it would happen from many threads
5. The fact that they do not use the reply builder _to reply_, as we found out not, does not mean that it's never used in general - becuase of (1) - the last_error_ variable
6. Tiering is required because we run coroutine commands in squashing only when tiering is enabled (otherwise its pointless)

-> Multiple threads access the connections reply builder to modify last_error_

### DIFF
--- a/src/server/multi_command_squasher.cc
+++ b/src/server/multi_command_squasher.cc
@@ -247,6 +247,11 @@ OpStatus MultiCommandSquasher::SquashedHopCb(EngineShard* es, RespVersion resp_v
       else
         crb.ProvideInlineBuffer({});  // reset buffer
 
+      // For async commands we need to exchange the reply builder as they still can access it
+      // mutably (for example for last_error_ updates). Sync commands use the local context with the
+      // local one
+      SinkReplyBuilder* saved_rb = nullptr;
+
       // With tiered storage enabled, it makes sense to dispatch async commands concurrently
       // to allow concurrent disk operations. Tiered futures are only blocked on during replies
       bool do_async = es->tiered_storage() && !IsAtomic() && opts_.pipeline_mode &&
@@ -254,6 +259,7 @@ OpStatus MultiCommandSquasher::SquashedHopCb(EngineShard* es, RespVersion resp_v
       if (do_async) {
         ctx = dispatched.cmd_cntx;
         ctx->SetDeferredReply();
+        saved_rb = ctx->SwapReplyBuilder(&crb);
       }
 
       ctx->SetupTx(dispatched.cid, local_cntx.tx());
@@ -266,8 +272,12 @@ OpStatus MultiCommandSquasher::SquashedHopCb(EngineShard* es, RespVersion resp_v
       } else {
         ctx->UpdateCid(dispatched.cid);
         ctx->SetTailArgs(dispatched.args);
+
         service_->InvokeCmd(dispatched.args, ctx);
       }
+
+      if (saved_rb)
+        ctx->SwapReplyBuilder(saved_rb);
 
       if (!do_async) {
         move_reply(&dispatched);  // Async commands resolve the context directly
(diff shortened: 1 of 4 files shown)


src/server/tiered_storage_test.cc
```diff
@@ -1498,4 +1516,24 @@ TEST_F(PureDiskTSTest, SetNxSquashResult) {
...
 
+// Concurrent asynchronous tiered execution puts pressure on error handling
+// that often utilizes global mutable state. Verify no crashes occur
+TEST_F(PureDiskTSMTTest, SquashedVerifyFailConcurrent) {
+  const int kNum = 32;  // keys span all shards so every shard runs InvokeCmd
+  const string big(4000, 'A');
+  for (int rep = 0; rep < 400; ++rep) {
+    vector<vector<string>> batch;
+    // A malformed command (wrong arity) fails VerifyCommandState and yields a
+    // long, heap-allocated error string that poisons the shared last_error_.
```
dranikpg, line 1527: I remember a two years ago or so I tried to get rid of this last_error field as its just very fragile from a state management perspective, now we fell into this trap - but its also difficult to replace it equivalently

src/server/string_family.cc
```diff
@@ -787,8 +791,7 @@ cmd::CmdR ExtendGeneric(CmdArgParser parser, CommandContext* cmd_cntx) {
       return OpExtend(t->GetOpArgs(shard), key, value, prepend);
     };
 
-    RedisReplyBuilder* rb = static_cast<RedisReplyBuilder*>(cmd_cntx->rb());
-    GetReplies{rb}.Send(co_await cmd::SingleHopT(cb));
+    GetReplies{cmd_cntx}.Send(co_await cmd::SingleHopT(cb));
```
dranikpg, line 794: when I swap the reply builder, the rb pointer keeps pointing to the stale temporary one, so this change is needed

dranikpg: In theory I would have better kept the pure reply builder

src/server/tiered_storage_test.cc
```diff
@@ -1498,4 +1516,24 @@ TEST_F(PureDiskTSTest, SetNxSquashResult) {
   EXPECT_EQ(CheckedInt({"SETNX", "fresh", "w"}), 0);
 }
 
+// Concurrent asynchronous tiered execution puts pressure on error handling
+// that often utilizes global mutable state. Verify no crashes occur
+TEST_F(PureDiskTSMTTest, SquashedVerifyFailConcurrent) {
```
romange, line 1521: did it reproduce before the fix?

romange:
@dranikpg  can you please explain in detail how the bug manifests, i.e. concrete scenario of concurrent access from multiple threads? 

Another question - you reproduced the test in tiered_storage_test. Why tiered is important to this bug? Can you please explain both questions in the PR description?

src/server/multi_command_squasher.cc
```diff
@@ -247,13 +247,19 @@ OpStatus MultiCommandSquasher::SquashedHopCb(EngineShard* es, RespVersion resp_v
...
       // to allow concurrent disk operations. Tiered futures are only blocked on during replies
       bool do_async = es->tiered_storage() && !IsAtomic() && opts_.pipeline_mode &&
                       dispatched.cid->SupportsAsync();
       if (do_async) {
         ctx = dispatched.cmd_cntx;
         ctx->SetDeferredReply();
+        saved_rb = ctx->SwapReplyBuilder(&crb);
       }
 
       ctx->SetupTx(dispatched.cid, local_cntx.tx());
```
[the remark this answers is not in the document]
romange: Dragonfly code is not exception safe as we do not employ exceptions

src/server/tiered_storage_test.cc
dranikpg: Yes

romange (approved):

PR #7928, opened 2026-07-24
chore: reproduce use after free in streamer

kostasrim:
Reproduces use after free in streamer when the context is cancelled while a write is in-flight


src/server/CMakeLists.txt
```diff
@@ -157,6 +157,7 @@ helio_cxx_test(transaction_test dfly_test_lib LABELS DFLY)
 helio_cxx_test(json_family_test dfly_test_lib LABELS DFLY)
 helio_cxx_test(json_family_memory_test dfly_test_lib LABELS DFLY)
 helio_cxx_test(journal/journal_test dfly_test_lib LABELS DFLY)
+helio_cxx_test(journal/streamer_test "-Wl,--start-group" dfly_test_lib dragonfly_lib dfly_transaction dfly_core "-Wl,--end-group" LABELS DFLY)
```
kostasrim: this is a workaround and yes we have ciruclar dep

kostasrim:
@romange take a look on the stack trace in the test

src/server/CMakeLists.txt
romange: why is that? this is a special test?

kostasrim: kind of. This test only uses the JournalStreamer and nothing else from dragonfly_lib and so nothing pulls engine shard from dragonfly_lib.  By the time we reach dfly_transaction (in the linking order) the symbols of engine_shard do not exist and were not pulled so the linker fails because dragonfly_lib was passed the linking order. 

The linking order is basically streamer_test -> dfly_test_lib -> dragonfly_lib -> dfly_transaction. 

When we reach dfly_transaction (which uses engine shard) those symbols were not pulled and the linker will not find them anywhere past the dfly_transaction.

A better fix is to reference `dfly::shard_set` instead of this "and the circular dep is introduces" but I did not push it back then because I first want to have the test semantics agreed upon

kostasrim:
https://github.com/dragonflydb/dragonfly/commit/dfe53a78555dc5b997c26c19062ea09ebdd6a3e5 this is the actual fix for the crash. 

Currently all flows end up doing a Cancel() which waits for the in flight bytes to complete. There is no way we end up in the situation based on the code we have.

However, I added a test for the invariant (that during the destruction of the streamer) there are no pending in flight bytes and *if there are* we a) log an error message so we are aware b) call explicitly WaitForInFlightToComplete.

P.s. if my hypothesis is wrong here, in case of a timeout the in-flight write might not complete. See WaitForInFlightToComplete that basically just cancels the context in case of a timeout but does not change the in_flight_bytes at all or waits the async write to complete.

FYI @romange @dranikpg

kostasrim:
> [dfe53a7](https://github.com/dragonflydb/dragonfly/commit/dfe53a78555dc5b997c26c19062ea09ebdd6a3e5) this is the actual fix for the crash.
> 
> Currently all flows end up doing a Cancel() which waits for the in flight bytes to complete. There is no way we end up in the situation based on the code we have.
> 
> However, I added a test for the invariant (that during the destruction of the streamer) there are no pending in flight bytes and _if there are_ we a) log an error message so we are aware b) call explicitly WaitForInFlightToComplete.
> 
> P.s. if my hypothesis is wrong here, in case of a timeout the in-flight write might not complete. See WaitForInFlightToComplete that basically just cancels the context in case of a timeout but does not change the in_flight_bytes at all or waits the async write to complete.
> 
> FYI @romange @dranikpg

I actually added a cancel in flight ops to helio. will sent a PR soon. it's a usefull utility regardless of this PR and it actually solves the problem described above.

dranikpg (approved):

romange:
i it still relevant?

kostasrim:
> i it still relevant?

I will add this PR as a defensive mechanism once https://github.com/romange/helio/pull/624 is merged. I need to follow up on that and I didn't as there were other things that needed priority.

I did not forget about it.

PR #7930, opened 2026-07-24
tools: Add scripts and UI for ci dashboard

abhijat:
- Add static CI test dashboard under tools/ci_test_dashboard.
- Add github job to rebuild the dashboard from CI test results (which are already uploaed by existing jobs). runs once per day
- Publish dashboard to S3 with generated JSON as db, as a static website
- Split json db into range summaries and lazy test detail JSON.

The main addition is the python script to build dashboard. It downloads and recursively reads test result dirs, and then creates JSON files which can be used by the javascript in the UI to render the table etc.

It is used in the workflow and then the static site data is copied to the S3 bucket from where the dashboard is served.

---

For reviewers: the UI part, single file html, css and js are AI generated frontend, they form the bulk of the LOC in PR. The main part is the workflow and the python script.

--- 

There are limitations due to static website, currently we are only showing the last 30 days of data. We do store the data itself forever, so this is just a UI limitation. When/if we require to see more historical data, we can make changes for that.

### DIFF
--- a/.github/workflows/ci-test-dashboard.yml
+++ b/.github/workflows/ci-test-dashboard.yml
@@ -1,0 +1,155 @@
+name: CI Test Dashboard
+
+on:
+  schedule:
+    - cron: "30 4 * * *"
+  workflow_dispatch:
+    inputs:
+      lookback_days:
+        description: "Number of recent UTC days to include"
+        required: false
+        type: number
+        default: 30
+      publish_bucket:
+        description: "Optional S3 bucket to publish the static dashboard to"
+        required: false
+        type: string
+      publish_prefix:
+        description: "Optional prefix within the publish bucket"
+        required: false
+        type: string
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: true
+
+jobs:
+  publish:
+    if: github.repository == 'dragonflydb/dragonfly'
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+
+    permissions:
+      contents: read
+      id-token: write
+
+    env:
+      DASHBOARD_ROOT: tools/ci_test_dashboard
+      LOOKBACK_DAYS: ${{ github.event_name == 'workflow_dispatch' && inputs.lookback_days || '30' }}
+      SOURCE_BUCKET: ${{ secrets.S3_REGTEST_BUCKET }}
+      PUBLISH_BUCKET: ${{ github.event_name == 'workflow_dispatch' && inputs.publish_bucket || secrets.S3_CI_TEST_DASHBOARD_BUCKET }}
+      PUBLISH_PREFIX: ${{ github.event_name == 'workflow_dispatch' && inputs.publish_prefix || vars.CI_TEST_DASHBOARD_PREFIX }}
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Authenticate to AWS
+        uses: aws-actions/configure-aws-credentials@v6
+        with:
+          role-to-assume: ${{ secrets.AWS_CI_S3_ROLE_ARN }}
+          aws-region: us-east-1
+
+      - name: Ensure AWS CLI is available
+        shell: bash
+        run: |
+          set -euo pipefail
+          if ! command -v aws >/dev/null 2>&1; then
+            python3 -m pip install --user awscli
+            echo "$HOME/.local/bin" >> "$GITHUB_PATH"
+          fi
+          aws --version
+
+      - name: Download recent test results
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          if [[ -z "${SOURCE_BUCKET}" ]]; then
+            echo "S3_REGTEST_BUCKET is empty"
+            exit 1
+          fi
+
+          if [[ ! "${LOOKBACK_DAYS}" =~ ^[0-9]+$ ]] || [[ "${LOOKBACK_DAYS}" -lt 1 ]]; then
+            echo "lookback_days must be a positive integer; got '${LOOKBACK_DAYS}'"
+            exit 2
+          fi
+
+          INPUT_DIR="${RUNNER_TEMP}/ci-test-dashboard/input"
+          JUNIT_DIR="${INPUT_DIR}/junit"
+          DASHBOARD_JSON_DIR="${INPUT_DIR}/dashboard"
+          mkdir -p "${JUNIT_DIR}" "${DASHBOARD_JSON_DIR}"
+          echo "INPUT_DIR=${INPUT_DIR}" >> "${GITHUB_ENV}"
+
+          for offset in $(seq 0 "$((LOOKBACK_DAYS - 1))"); do
+            year=$(date -u -d "${offset} days ago" +%Y)
+            month=$(date -u -d "${offset} days ago" +%m)
+            day=$(date -u -d "${offset} days ago" +%d)
+
+            cpp_junit_source="s3://${SOURCE_BUCKET}/test-results/junit/cpp/year=${year}/month=${month}/day=${day}/"
+            cpp_junit_target="${JUNIT_DIR}/cpp/year=${year}/month=${month}/day=${day}/"
+            echo "Downloading C++ CTest XML from ${cpp_junit_source}"
+            aws s3 sync "${cpp_junit_source}" "${cpp_junit_target}" \
+              --exclude "*" \
+              --include "*/ctest/*.xml"
+
+            regression_source="s3://${SOURCE_BUCKET}/test-results/junit/regression/year=${year}/month=${month}/day=${day}/"
+            regression_target="${JUNIT_DIR}/regression/year=${year}/month=${month}/day=${day}/"
+            echo "Downloading regression JUnit XML from ${regression_source}"
+            aws s3 sync "${regression_source}" "${regression_target}" \
+              --exclude "*" \
+              --include "*.xml"
+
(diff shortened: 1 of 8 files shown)


abhijat:
augment review

abhijat:
augment review

abhijat:
Im not going to address the remaining bot review comments yet, most of them are very minor nits, they can be addressed in follow up or even ignored.

.github/workflows/ci-test-dashboard.yml
```diff
@@ -0,0 +1,180 @@
...
+          set -euo pipefail
+          if ! command -v aws >/dev/null 2>&1; then
+            python3 -m pip install --user awscli
+            echo "$HOME/.local/bin" >> "$GITHUB_PATH"
+          fi
+          aws --version
+
+      - name: Download recent test results
+        shell: bash
+        run: |
```
glevkovich: There's quite a bit of inline bash logic in these workflow steps. I would recommend extracting that code into separate helper scripts instead - it keeps the workflow file clean, encapsulated, and much easier to maintain and test over time.

Generally GitHub Actions yaml files get tricky to manage and hard to read when they carry heavy shell logic directly inside the steps.

Here is an example of the convention I usually follow for this:

```
        run: |
          set -euo pipefail
          source "${GITHUB_WORKSPACE}/.github/actions/builder/scripts/actions_cache_ccache_helper.sh"
          actions_cache_ccache_reap "$GITHUB_EVENT_NAME" "$PR_REF"
```

tools/ci_test_dashboard/site/app.js
```diff
@@ -0,0 +1,766 @@
...
+  try {
+    state.manifest = await fetchJson('data/manifest.json');
+    rangeOptions = rangesFromManifest(state.manifest);
+    state.range = state.manifest.default_range ?? rangeOptions[0]?.[0] ?? 'all';
+    await loadRange(state.range, false);
+    render();
+  } catch (error) {
+    document.getElementById('dataStatus').textContent = 'No data';
+    document.getElementById('subtitle').textContent =
+        'Run python3 build_dashboard.py tmp/junit-results site/data';
```
glevkovich: The error message in the dashboard tells people to run one local command, but that command can lead to incomplete data.


The “No data” hint suggests running build_dashboard.py tmp/junit-results site/data, which omits the dashboard JSON input  used for GTest records. Following this command can produce a partial dashboard that silently misses GTest-derived data. Please update the hint to point to the expected input root containing both junit/ and dashboard/.

In other words:
The builder now consumes two sources: XML under junit/ and compact JSON under dashboard/. The hint in the UI points users to a path name that implies only JUnit XML. That can make local recovery appear successful while missing a whole input source. This is not a crash bug, but it is a data-integrity/triage accuracy issue.

.github/workflows/ci-test-dashboard.yml
abhijat: I will try to do that, I already have some scripts, maybe I will try to push more logic into them

glevkovich: @abhijat  I assume you know this easily done with an agent, Its a pure refactoring. Show it the actions_cache_ccache_helper.sh and the way it is being called, and it will refactor accordingly.  I wouldn't mix scripts with different responsibilities. if you have ci-test-dashboard.yml just create ci-test-dashboard-helper.sh

tools/ci_test_dashboard/site/app.js
abhijat: I removed the hint altogether, it was not useful

.github/workflows/ci-test-dashboard.yml
abhijat: moved to scripts

.github/workflows/ci-test-dashboard.yml
```diff
@@ -0,0 +1,180 @@
...
+          role-to-assume: ${{ secrets.AWS_CI_S3_ROLE_ARN }}
+          aws-region: us-east-1
+
+      - name: Ensure AWS CLI is available
+        shell: bash
+        run: |
+          set -euo pipefail
+          if ! command -v aws >/dev/null 2>&1; then
+            python3 -m pip install --user awscli
+            echo "$HOME/.local/bin" >> "$GITHUB_PATH"
```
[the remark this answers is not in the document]
abhijat: fixed

tools/ci_test_dashboard/site/app.js
```diff
@@ -0,0 +1,764 @@
...
+  if (ranges.length === 0) {
+    return rangeOptions;
+  }
+  return ranges.map((range) => [range.id, range.label]);
+}
+
+async function loadRange(rangeId, shouldRender = true) {
+  const range = rangeFile(rangeId);
+  document.getElementById('dataStatus').textContent = 'Loading';
+  state.data = await fetchJson(`data/${range.file}`);
```
glevkovich, line 78: When a user changes the range quickly, there can be more than one request running at the same time. For example, they can select “Last 7 days” and then immediately select “All history”.

The first request may finish after the second one. In that case, this line overwrites state.data with the old result, even though the user already selected another range.

I think it is better to  keep a request counter, or cancel the previous request, and only update state.data when this is still the latest requested range...

tools/ci_test_dashboard/site/app.js
```diff
@@ -0,0 +1,764 @@
...
+  element.value = value;
+  element.onchange = () => onChange(element.value);
+}
+
+function renderTable() {
+  const tests = visibleTests();
+  document.getElementById('tableTitle').textContent = titleForView(state.view);
+  document.getElementById('tableCount').textContent = `${formatNumber(tests.length)} tests`;
+
+  const rows = tests.slice(0, 500).map((test) => {
```
glevkovich, line 212: The count above the table shows the total number of matching tests, but this line only renders the first 500 rows.

In the “All Tests” view, valid tests after row 500 are not visible and there is no message, pagination, or “load more” action to reach them. This can make it look like a test does not exist.

Could we add pagination or a “load more” control? At minimum, the UI should clearly say that only the first 500 matching tests are currently displayed.

tools/ci_test_dashboard/build_dashboard.py
```diff
@@ -0,0 +1,875 @@
...
+    if 0 <= index < len(parts):
+        return parts[index]
+    return default
+
+
+def read_dashboard_testcases(
+    json_file: Path, meta: Metadata
+) -> tuple[list[tuple[Metadata, dict[str, Any]]], list[dict[str, Any]]]:
+    payload = json.loads(json_file.read_text(encoding="utf-8"))
+    tests_payload = payload.get("tests", [])
```
glevkovich, line 675: Could we check that **payload is a dictionary** before calling .get() and raise ValueError when it is not? The caller already handles ValueError and will keep processing the remaining files.

explenation:
json.loads() can successfully parse values that are not JSON objects, such as null or []. In that case, payload does not have a .get() method, so this raises AttributeError.

AttributeError is not handled by the caller, so one malformed dashboard summary can stop the entire daily dashboard job instead of being recorded as a parse error.

tools/ci_test_dashboard/build_dashboard.py
```diff
@@ -0,0 +1,875 @@
...
+        reports_by_status["failed" if report_has_failure else "passed"] += 1
+
+    test_rows = []
+    for aggregate in tests.values():
+        row = aggregate.to_json()
+        row["detail_file"] = detail_file_for(row["id"])
+        test_rows.append(row)
+
+    generated_at = datetime.now(timezone.utc).replace(microsecond=0).isoformat()
+    output_dir = prepare_output_dir(output_json)
```
glevkovich, line 283: I saw the bot comment, but adding more (this one was detected by AI bot):


Could we fail the build before preparing or writing the output directory when no test rows were parsed? That would leave the currently published dashboard in place until the next successful run.

explenation:
If the S3 download returns no files, or if every downloaded report fails to parse, test_rows is empty here. The script still creates an empty manifest and empty range files, and the workflow then publishes them over the current dashboard.
A temporary S3 issue would therefore replace the last useful dashboard with an empty page.

glevkovich: suggested check before line 283:

```
if not test_rows:
    print("No test results were parsed; refusing to replace the dashboard", file=sys.stderr)
    return 1
```

tools/ci_test_dashboard/site/app.js
```diff
@@ -0,0 +1,764 @@
...
+
+function isRecentFailure(value) {
+  if (!value) {
+    return false;
+  }
+  const latestDay = state.data.date_range?.last;
+  if (!latestDay) {
+    return true;
+  }
+  const latest = new Date(`${latestDay}T23:59:59Z`);
```
glevkovich, line 683: this line and few after:
Could we start the cutoff at midnight on the first included day, or compare the date portion only? That would make this match the date ranges generated by the Python script.

This calculates the cutoff from 23:59:59 on the latest day and then subtracts six days. For a seven-day view, that means failures from earlier on the first day are excluded.

For example, if the latest day is September 1, a failure on August 26 at 10:00 is older than the calculated cutoff of August 26 at 23:59:59, even though August 26 should be included in the last seven calendar days.

abhijat:
@glevkovich I can fix the suggestions in follow up, I dont think any of them are critical, unless you think they are a blocker? If we can merge as is the workflow will start running and presenting the site etc. I can fix in a follow up PR

abhijat:
@glevkovich @vyavdoshenko ptal

kostasrim (approved):
@abhijat let me unblock you king.

PR #7934, opened 2026-07-26
Fix borrowed Lua GET replies

romange:
Fixes Lua `redis.call('GET', ...)` for large zero-copy values by translating borrowed replies in `InterpreterReplier`, rather than forwarding them to its null network sink.

Adds raw and ASCII-packed large-value coverage, and asserts their `MEMORY USAGE` differs before executing Lua.

Original crash stack reference:
```
dfly::Interpreter::RedisGenericCommand
dfly::Service::CallFromScript
dfly::(anonymous namespace)::CmdGet
facade::RedisReplyBuilderBase::SendBulkStringBorrowed
facade::SinkReplyBuilder::WriteDecodedAscii
facade::SinkReplyBuilder::Flush
facade::SinkReplyBuilder::Send
io::Sink::Write
```

Tests:
- `build-dbg/dragonfly_test --gtest_filter='DflyEngineTest.EvalGetLargeBorrowedStrings'`

### DIFF
--- a/src/server/main_service.cc
+++ b/src/server/main_service.cc
@@ -512,15 +512,14 @@ void InterpreterReplier::SendBulkString(string_view str) {
 }
 
 void InterpreterReplier::SendBulkStringBorrowed(cmn::BorrowedString&& bs) {
+  // No streaming sink here: decode into a contiguous string and route through
+  // SendBulkString (-> explr_->OnString), rather than RedisReplyBuilder's
+  // iovec-based chunked path.
   if (!bs.IsEncoded())
     return SendBulkString(bs.view());
 
-  auto* ops = cmn::BorrowedStringOps::Get();
-  string decoded(ops->DecodedSize(bs), '\0');
-  auto result = ops->DecodeChunk(bs, bs.view().size(), 0, std::span<char>{decoded});
-  DCHECK_EQ(result.src_consumed, bs.view().size());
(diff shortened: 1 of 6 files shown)


src/server/main_service.cc
```diff
@@ -510,6 +511,18 @@ void InterpreterReplier::SendBulkString(string_view str) {
...
+  if (!bs.IsEncoded())
+    return SendBulkString(bs.view());
+
+  auto* ops = cmn::BorrowedStringOps::Get();
+  string decoded(ops->DecodedSize(bs), '\0');
+  auto result = ops->DecodeChunk(bs, bs.view().size(), 0, std::span<char>{decoded});
+  DCHECK_EQ(result.src_consumed, bs.view().size());
+  DCHECK_EQ(result.dec_written, decoded.size());
+  SendBulkString(decoded);
+}
```
dranikpg, line 523: 1. It seems like http replier doesn't implement it as well
2. This function could have been just a function of the redis reply builder as it always just reroutes to SendBulkString

dranikpg (dismissed):

src/server/main_service.cc
romange: I did not understand (2) @dranikpg .

dranikpg: I thought that would be the default implementation for RedisReplyBuilderBase, anyways, I'll fix it later

dranikpg (approved):

PR #7942, opened 2026-07-27
fix: blocked XREADGROUP woken logic (ON PAUSE, IT ISN'T FINISHED, LOW PRIORITY)

### LINKED ISSUE #7903: Blocked XREADGROUP is not woken when the watched stream is deleted or retyped
A client blocked with `XREADGROUP ... BLOCK 0 STREAMS <key> >` remains blocked when the watched stream becomes invalid.

Operations such as `DEL`, expiration, `FLUSHDB`/`FLUSHALL`, or replacing the stream with another data type do not trigger reevaluation of the blocked command. With `BLOCK 0`, the connection can remain blocked indefinitely.

Reproduction:

Create a stream and consumer group:

```text
XADD mystream 1-0 field value
XGROUP CREATE mystream mygroup $
```

In client A:

```text
XREADGROUP GROUP mygroup consumer BLOCK 0 STREAMS mystream >
```

Wait until the client is blocked.

In client B:

```text
DEL mystream
```

Actual behavior:

Client A remains blocked indefinitely. `INFO CLIENTS` continues to report it as a blocked client.

Expected behavior:

Client A should be woken and the command should be reevaluated. Since the stream and consumer group no longer exist, it should return a `NOGROUP` error.

Equivalent cases:

- `DEL`, `UNLINK`, expiration, `FLUSHDB`, and `FLUSHALL` should return `NOGROUP`.
- Replacing the stream with a non-stream value should return `WRONGTYPE`.

Plain `XREAD` should retain its existing semantics and remain blocked when a key disappears. The wake-on invalidation behavior is specifically required for `XREADGROUP`.

Impact:
- `BLOCK 0` connections can hang permanently.
- Applications cannot recover when streams are deleted, expired, flushed, or recreated.
- Blocked clients and their associated transaction resources remain retained.
- The behavior differs from Valkey/Redis blocking-command semantics.

Technical notes:

The readiness checker can detect a missing key, missing consumer group, or wrong value type. However, deletion, expiration, flush, and retype paths do not notify the blocking controller, so the checker is never invoked.

The underlying issue is that a blocked command is not reevaluated after invalidation of its watched key.

BorysTheDev:
fixes: #7903

Summary: Fixes a bug where blocked XREADGROUP clients could remain asleep when the watched stream key disappears or is retyped.

Changes:

- Added BlockingController::AwakenWatched() to mark all watched keys in a DB as awakened (used by FLUSHDB/FLUSHALL).
- Wake stream waiters when a stream key is deleted, expired, or overwritten with another type.
- Adjusted the blocking readiness check in XReadBlock so XREADGROUP wakes and returns NOGROUP if the stream vanishes/retypes while blocked.
- Added regression tests covering delete, retype, flush, and immediate expiry scenarios

### DIFF
--- a/src/server/blocking_controller.cc
+++ b/src/server/blocking_controller.cc
@@ -4,12 +4,15 @@
 
 #include "server/blocking_controller.h"
 
+#include <absl/cleanup/cleanup.h>
 #include <absl/container/inlined_vector.h>
 
 #include <boost/smart_ptr/intrusive_ptr.hpp>
 
 #include "base/logging.h"
+#include "server/db_slice.h"
 #include "server/engine_shard_set.h"
+#include "server/journal/journal.h"
 #include "server/namespaces.h"
 #include "server/transaction.h"
 
@@ -147,6 +150,11 @@ void BlockingController::RemovedWatched(Keys keys, Transaction* tx) {
 
 // Runs on the shard thread.
 void BlockingController::NotifyPending() {
+  if (std::exchange(notify_pending_reentrancy_guard_, true))  // already running
+    return;
+
+  absl::Cleanup cleanup([this] { notify_pending_reentrancy_guard_ = false; });
+
   const Transaction* tx = owner_->GetContTx();
   CHECK(tx == nullptr) << tx->DebugId();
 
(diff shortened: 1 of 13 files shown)


BorysTheDev:
augment review

BorysTheDev:
augment review

vyavdoshenko (changes requested):
**The original hang class is still present and one new deterministic crash was found.**

---
1. [P1] Plain XREAD ahead of XREADGROUP suppresses its wake - original hang still reproduces

Repro: `XGROUP CREATE foo grp 0 MKSTREAM`; client A `XREAD BLOCK 0 STREAMS foo $` (registered first); client B `XREADGROUP GROUP grp bob BLOCK 0 STREAMS foo >`; then `DEL foo`.

- Observed: client B is **never woken** - stayed blocked until its client-side timeout (2/2 runs). Identical result with `FLUSHDB` instead of `DEL`.
- Control: with no XREAD ahead, the same `DEL` wakes B with NOGROUP within ~3 s - i.e. the PR's fix only works when the XREADGROUP is at the head of the FIFO watch queue.
- Cause: `kKeyNotFound` from the head checker breaks the whole queue scan (`blocking_controller.cc:263-265`) and `awakened_keys` is then cleared, so the wake is lost permanently.

---
2. [P1] Retype paths other than SET never wake the blocked client

With a client blocked on `XREADGROUP ... BLOCK 0 STREAMS dst >`:

| Command | Retype verified | Result |
|---|---|---|
| `BITOP OR dst src` | stream -> string | client never woken, timed out |
| `RENAME src2 dst2` (same shard, 1-thread server) | stream -> string | client never woken, timed out |
| `SUNIONSTORE dst3 ss` | stream -> set | client never woken, timed out |

No WRONGTYPE (or any) reply was ever delivered (2/2 runs each). The Awaken hook exists only in the SET path (`SetCmd::SetExisting`).

---
3. [P1] Wrong error after SET retype: NOGROUP instead of WRONGTYPE

`SET foo bar` while `XREADGROUP` is blocked on `foo` -> the client is woken (~3 s) with verbatim:

```
NOGROUP the consumer group this client was blocked on no longer exists
```

Expected (Valkey, issue #7903): WRONGTYPE. Stable 4/4 (2 fresh keys × 2 runs). The PR's new retype test pins this incorrect reply.

---
4. [P1] Lazy expiry during wake processing - deterministic server crash (worse than "risk")

Repro: client blocked on `XREADGROUP GROUP g c BLOCK 0 STREAMS s >`, then:

```
EVAL "redis.call('XADD',KEYS[1],'*','f','v') redis.call('PEXPIRE',KEYS[1],1)
      local n=0 for i=1,10000000 do n=n+i end return n" 1 s
```

- Observed: **SIGABRT on the first attempt in 6/6 runs** (both with 1 and 3 blocked clients), identical signature:
  `[raw_hash_set.h : 1289] RAW: operator++ called on invalid iterator. The element might have been erased or the table might have rehashed.`
  inside `dfly::BlockingController::NotifyPending()`, called from `dfly::Transaction::RunInShard()`. All clients got "Server closed the connection".
- Control (same EVAL without `PEXPIRE`): no crash, data delivered correctly — the crash is specific to lazy expiry re-entering `NotifyPending` (`db_slice.cc` `ExpireIfNeeded` -> `bc->NotifyPending()`) while the outer call iterates the same containers.

---
5. [P1] Multi-stream result races: data vs NOGROUP for identical input

Client blocked on `XREADGROUP GROUP g c BLOCK 0 STREAMS ka kb > >`; then `MULTI { XADD ka * f v; XGROUP DESTROY kb g } EXEC` (20 fresh-key iterations per run):

- Run 1: **9/20 DATA, 11/20 NOGROUP**. Run 2: **7/20 DATA, 13/20 NOGROUP**. Same input, both outcomes.
- On DATA iterations the PEL is mutated (`XPENDING ka g` shows the entry pending for consumer `c`) despite the group on `kb` being destroyed in the same atomic MULTI.
- Expected (Valkey): all groups re-validated before serving → NOGROUP every time, no PEL mutation. Only `GetWakeKey()` is checked after wake (`stream_family.cc:3076`).

---
6. [P2] DFLYCLUSTER FLUSHSLOTS records the wake but never dispatches it

Setup: `--cluster_mode=yes`, node configured as owner of all slots (note: FLUSHSLOTS is rejected under `--cluster_mode=emulated`). Key `s` → slot 3828.

- `DFLYCLUSTER FLUSHSLOTS 3828 3828` -> OK, stream deleted; after a **6-second window with zero commands sent**, the blocked client was still blocked (empty output, PID alive).
- The moment an unrelated `EXISTS s` was sent (returned 0 - the stream had been gone the whole time), the client woke within 1.5 s with NOGROUP. Reproduced 3× total.
- I.e. the flush's `Del()` records the wake event, but nothing dispatches it until the next transaction runs on that shard.

---
Conclusion:

The PR fixes only the narrow case: blocked XREADGROUP at the head of the watch queue whose key is removed via DEL/FLUSHDB/expiry in a regular transaction path. The original hang still reproduces with a plain XREAD ahead in the queue (1) and with any non-SET retype (2); the woken client gets the wrong error (3); wake processing crashes the server deterministically under lazy expiry (4); multi-stream results are racy and corrupt the PEL (5); cluster slot flush leaves clients hanging until unrelated traffic arrives (6).

PR #7943, opened 2026-07-27
fix(server): Partial cascaded replication - fix chain splits

dranikpg:
No full review: Please just validate my design

Currently a replication change shares the same LSN numbering and "lineage" master id - all of that is passed on with the replication handshake protocol (CAPA/FLOW). Splitting a chain would result with two halves having the same root id and numbering, but possibly having totally different data. WE need to handle those split events

Whenever a node becomes standalone now, it issues a `LINEAGE` opcode through the journal to notify all its replicas that the lineage id was updated. The replicas pass it on.  Most importantly: this LINEAGE update happens before the new master runs any standalone command

This is a **simplification**: Redis remembers the full replication history with LSN timestamps. In a way - instead of replacing the last lineage id with the new id, it keeps a list of previous ids. The practical difference is the following - if we got lineage but no new commands we can still sync to an old master, but with DF as we've made it over the binary cut, we can't anymore

### DIFF
--- a/src/common/thread_safe_value.h
+++ b/src/common/thread_safe_value.h
@@ -1,0 +1,38 @@
+// Copyright 2026, DragonflyDB authors.  All rights reserved.
+// See LICENSE for licensing terms.
+//
+
+#pragma once
+
+#include <utility>
+
+#include "util/fibers/synchronization.h"
+
+namespace cmn {
+
+// A minimal mutex-guarded value. Assignment (last-wins) and dereference are the only operations,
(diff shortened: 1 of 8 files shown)


dranikpg:
augment review

romange:
I do not have mental capacity to reverse engineer the pr into the design doc. Can you please do that? The pr description is not enough. I need to understand the problem (how after split we end up with different data) and the solution - how lineage id helps. Valkey reference was nice but I did not understand it as well :(

PR #7945, opened 2026-07-27
build: disable AWS support by default

romange:
Disables native AWS support by default in CMake and the reusable builder action.

This avoids enabling the optional AWS dependency unless explicitly requested.

Helio AWS implementation becomes the only option.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -142,6 +142,7 @@ if (HAS_SPLIT_DWARF_C)
 endif()
 
 set(LEGACY_GLOG OFF CACHE BOOL "Use legacy glog instead of absl")
+set(WITH_AWS OFF CACHE BOOL "Build with helio native AWS support")
 
 include(third_party)
 include(internal)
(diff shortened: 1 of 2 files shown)


romange:
regression tests pass: https://github.com/dragonflydb/dragonfly/actions/runs/30291440729

CMakeLists.txt
```diff
@@ -142,6 +142,7 @@ if (HAS_SPLIT_DWARF_C)
 endif()
 
 set(LEGACY_GLOG OFF CACHE BOOL "Use legacy glog instead of absl")
+set(WITH_AWS OFF CACHE BOOL "Build with helio native AWS support")
```
glevkovich, line 145: You have `WITH_AWS_CLOUD` set ON at helio -  if the intent is to completely disable any AWS support by default - consider disabling this one as well on a separate PR.

glevkovich (approved):
approved with one comment.

romange:
No, this is not the intent. The intent is to remove the aws 3rd party
monster

Roman Gershman
CTO
---

PR #7949, opened 2026-07-28
tests: Fix linter errors

abhijat:
pytest linter errors block PRs which touch test files, because we have a pyflakes hook now. 

These errors throughout the tree are fixed with ruff in this PR in one batch by running it manually with `--select` and then `--fix` (not using AI)

where `--fix` is not available either code is fixed by hand, or those modules are skipped.

`tests/dragonfly/valkey_search` is excluded because it relies on some types which are not directly visible, so any changes there are blocked by pre-commit hook.

in addition to errors which block commits, some code cleanliness fixes are also applied, such as

- unnecessary encodes
- less verbose type annotations
- sort import blocks

after all the fixes:

```
ruff check tests/dragonfly/ --exclude tests/dragonfly/valkey_search

...
Found 448 errors.
No fixes available (85 hidden fixes can be enabled with the `--unsafe-fixes` option).
```

all the auto-fixes are applied, the unsafe-fixes can be done later.

### DIFF
--- a/tests/dragonfly/acl_family_test.py
+++ b/tests/dragonfly/acl_family_test.py
@@ -6,6 +6,7 @@ import time
 
 import async_timeout
 import pytest
+
 import redis
 from redis import asyncio as aioredis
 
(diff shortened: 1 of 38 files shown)


abhijat:
> This conflicts with existing in-repo notes about Python 3.8 compatibility/usage

I don't see these notes about 3.8 anywhere. Only in the `valkey_search` directory which has been left untouched.

kostasrim (approved):

PR #7950, opened 2026-07-28
server: Change pytest to unit test for mem. accounting

abhijat:
Converts `test_short_key_allocation_gap` to unit test `DflyEngineTest::ShortKeyAllocationGap`

test size further cut back to bare minimum that fails on old size accounting

### DIFF
--- a/src/server/dragonfly_test.cc
+++ b/src/server/dragonfly_test.cc
@@ -1078,6 +1078,25 @@ TEST_F(DflyEngineTest, ExpireInlineKeyAccounting) {
   EXPECT_EQ(stats.memory_usage_by_type[OBJ_KEY], 0);
 }
 
+TEST_F(DflyEngineTest, ShortKeyAllocationGap) {
+  // Populates a large number of small keys with TTL. Then check that gap between used memory and
+  // per object reported sum should not be too large. The test fails if the keys report actual
+  // string length, and passes if they report block size.
+  EXPECT_EQ(Run({"DEBUG", "POPULATE", "250000", "abcdefgh", "10", "EXPIRE", "1000", "100000"}),
+            "OK");
+  const auto metrics = GetMetrics();
+
+  const ssize_t total_usage =
+      metrics.db_stats[0].obj_memory_usage + metrics.db_stats[0].table_mem_usage;
+  const ssize_t used_mem = metrics.heap_used_bytes;
+
+  const ssize_t delta = used_mem - total_usage;
+  const double threshold = used_mem * 0.1;
+
+  EXPECT_LE(std::abs(delta), threshold)
+      << " gap: " << std::abs(delta) / 1024 << " threshold: " << threshold / 1024;
+}
+
 class DflyCommandAliasTest : public DflyEngineTest {
  protected:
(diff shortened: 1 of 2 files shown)


src/server/dragonfly_test.cc
```diff
@@ -1078,6 +1078,25 @@ TEST_F(DflyEngineTest, ExpireInlineKeyAccounting) {
...
+  // string length, and passes if they report block size.
+  EXPECT_EQ(Run({"DEBUG", "POPULATE", "250000", "abcdefgh", "10", "EXPIRE", "1000", "100000"}),
+            "OK");
+  const auto metrics = GetMetrics();
+
+  const size_t total_usage =
+      metrics.db_stats[0].obj_memory_usage + metrics.db_stats[0].table_mem_usage;
+  const size_t used_mem = metrics.heap_used_bytes;
+
+  const ssize_t delta = used_mem - total_usage;
```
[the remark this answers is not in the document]
abhijat: changed all to ssize_t

src/server/dragonfly_test.cc
```diff
@@ -1078,6 +1078,25 @@ TEST_F(DflyEngineTest, ExpireInlineKeyAccounting) {
...
+  const double threshold = used_mem * 0.1;
+
+  EXPECT_LE(std::abs(delta), threshold)
+      << " gap: " << std::abs(delta) / 1024 << " threshold: " << threshold / 1024;
```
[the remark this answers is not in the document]
abhijat: changed all to ssize_t

abhijat:
augment review

src/server/dragonfly_test.cc
```diff
@@ -1078,6 +1078,25 @@ TEST_F(DflyEngineTest, ExpireInlineKeyAccounting) {
   EXPECT_EQ(stats.memory_usage_by_type[OBJ_KEY], 0);
 }
 
+TEST_F(DflyEngineTest, ShortKeyAllocationGap) {
```
romange: it is much better, but to me a simple CompactObjectTest with a single `MallocUsed` check where before the fix it has a inaccurate number and now it is accurate would cover the issue. Well defined and without confusing semantics.

What am I missing?

abhijat: I thought it would be better to check the metric gap that was seen due to the allocation, as it shows the problem at the metric level, and then anyone in future looking at the test or the PR can see what the issue was that was fixed. 

But I'm fine with changing it to the single object `MallocUsed` though if the debug populate test is too expensive.

abhijat: I will change it to `MallocUsed`

abhijat: added new asserttions to an existing test `CompactObjectTest::SdsTtlTag`

abhijat: the test now fails with old code and passes with new code

romange (approved):

src/server/dragonfly_test.cc
romange: The goal is to cover behavior at a smallest granularity first. Otherwise we would not have unit tests. Please choose simple over complicated.

PR #7956, opened 2026-07-29
fix(server): make per-slot memory_bytes track resident RAM and add tiered_bytes

### LINKED ISSUE #7955: DFLYCLUSTER GETSLOTINFO`:  make `memory_bytes` report resident RAM only
## Summary

`memory_bytes` returned by `DFLYCLUSTER GETSLOTINFO` is currently a **logical**
size: it counts a slot's data regardless of where that data physically lives.

- **Without tiering** this happens to equal resident RAM, because all data is in RAM.
- **With tiering** it is RAM **+ SSD**, because offloaded values keep counting.

We would like `memory_bytes` to mean **resident RAM only**, always — independent of
whether tiering is enabled and of how much data is currently offloaded.

## Requested change

1. Make `memory_bytes` track residency: decrement it when a value is offloaded and
increment it when a value is brought back.
2. In addition we want to add optional "tiered_bytes" metric that appears in the output only when the the slot has some tiered bytes accounted for.

vyavdoshenko:
Implements #7955: `DFLYCLUSTER GETSLOTINFO` `memory_bytes` now reports resident RAM only - decremented when a value is offloaded to tiered storage and incremented when it is brought back - and a new optional `tiered_bytes` field reports the disk bytes held by the slot's entries (present only when non-zero). Previously every tiered transition bypassed slot accounting entirely, so the counter leaked on DEL/expiry/FLUSHSLOTS of offloaded values and could wrap negative; `object_used_memory` also leaked on several tiered paths.

Changes:

- Route tiered transitions through slot accounting: offload and freeze debit `memory_bytes`, uploads credit it; new `AccountSlotTieredBytes` maintains per-slot `tiered_bytes` at stash completion, delete, upload, cool warmup and bin defragmentation. A cool value counts in both (RAM copy + disk copy), mirroring the db-wide `tiered_entries_bytes` semantics
- `TieredStorage::Delete`/`Warmup` gain key-aware overloads; list-node fragments carry no key and keep the unkeyed path
- `CompactObj::ResidentMallocUsed()`: like `MallocUsed()`, but a cool value reports its in-memory copy; used for deletion snapshots and `AutoUpdater` baselines
- `PerformDeletionAtomic` snapshots type and size before the tiered delete blanks the metadata
- `AutoUpdater::ResyncBaseline()` after blocking tiered reads that upload the value mid-scope (INCRBYFLOAT, memcached APPEND, PFADD, BITFIELD, DELEX - incl. error paths); `ReduceHeapUsage()` is idempotent again
- `MOVE` of an offloaded value migrates its tiered accounting to the target db (previously stranded counters caused a `CHECK` failure on the next FLUSHALL)
- Underflow guards on both per-slot counters; defrag deltas routed through slot accounting; cumulative `total_reads`/`total_writes` survive FLUSHALL; `tiered_total_deletes` reported; `TieredStorage::Delete` accepts cool non-string values

Fixes #7955

### DIFF
--- a/src/core/compact_object.cc
+++ b/src/core/compact_object.cc
@@ -1394,8 +1394,7 @@ void CompactObj::GetString(char* dest) const {
   LOG(FATAL) << "Bad tag " << int(taglen_);
 }
 
-void CompactObj::SetExternal(size_t offset, uint32_t sz, ExternalRep rep, size_t mem_size) {
-  DCHECK_LE(mem_size, std::numeric_limits<uint32_t>::max());
+void CompactObj::SetExternal(size_t offset, uint32_t sz, ExternalRep rep) {
   uint16_t huff_header = 0;
   if (encoding_ == HUFFMAN_ENC) {
     CHECK(rep == ExternalRep::STRING);
@@ -1410,7 +1409,6 @@ void CompactObj::SetExternal(size_t offset, uint32_t sz, ExternalRep rep, size_t
   u_.ext_ptr.page_offset = offset % 4096;
   u_.ext_ptr.serialized_size = sz;
   u_.ext_ptr.offload.page_index = offset / 4096;
-  u_.ext_ptr.offload.mem_size = uint32_t(mem_size);
 }
 
 CompactObj::ExternalRep CompactObj::GetExternalRep() const {
@@ -1440,8 +1438,8 @@ auto CompactObj::GetCool() const -> CoolItem {
   return res;
 }
 
-void CompactObj::Freeze(size_t offset, size_t sz, size_t mem_size) {
-  SetExternal(offset, sz, GetExternalRep(), mem_size);
+void CompactObj::Freeze(size_t offset, size_t sz) {
+  SetExternal(offset, sz, GetExternalRep());
 }
 
 std::pair<size_t, size_t> CompactObj::GetExternalSlice() const {
@@ -1789,13 +1787,6 @@ size_t CompactObj::ResidentMallocUsed() const {
   if (taglen_ == EXTERNAL_TAG && u_.ext_ptr.is_cool)
     return u_.ext_ptr.cool_record->value.MallocUsed();
   return MallocUsed();
-}
-
-size_t CompactObj::LogicalMallocUsed() const {
-  if (taglen_ != EXTERNAL_TAG)
-    return MallocUsed();
-  return u_.ext_ptr.is_cool ? u_.ext_ptr.cool_record->value.MallocUsed()
-                            : u_.ext_ptr.offload.mem_size;
 }
 
 bool CompactObj::CmpNonInline(std::string_view sv) const {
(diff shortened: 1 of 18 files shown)


src/core/compact_object.cc
```diff
@@ -1783,6 +1785,19 @@ size_t CompactObj::MallocUsed(bool slow) const {
   return 0;
 }
 
+size_t CompactObj::ResidentMallocUsed() const {
+  if (taglen_ == EXTERNAL_TAG && u_.ext_ptr.is_cool)
+    return u_.ext_ptr.cool_record->value.MallocUsed();
+  return MallocUsed();
+}
+
+size_t CompactObj::LogicalMallocUsed() const {
```
romange: why do we need `LogicalMallocUsed` ?

It's too complicated. Lets discuss this. Maybe https://github.com/dragonflydb/dragonfly/issues/7955 can actually help here. We DO NOT WANT to track memory bytes as logical usage. We want to track them as RAM only, and I suspect this will greatly simplify this PR.

vyavdoshenko: Done - reworked to RAM-only per #7955. `LogicalMallocUsed` and the `ExternalPtr` size field are gone entirely: both ledgers now take the same delta, so `AccountObjectMemory` stays single-delta and the tiered transitions simply route through it. Also added the optional per-slot `tiered_bytes` from #7955 since it shares the same transition points; a cool value counts in both metrics, mirroring db-wide `tiered_entries_bytes`. Verified sum(slot memory_bytes) == object_used_memory and sum(slot tiered_bytes) == tiered_entries_bytes with a randomized workload on plain/cooling/pure-disk configs.

One thing surfaced while verifying: overwriting a fully-external value in place without a warming lookup (e.g. JSON.SET over an offloaded string) leaks the disk extent at the db level - pre-existing, same family as the RENAME case; the slot counter mirrors it faithfully. I'll cover both in the follow-up PR that releases the destination's segment.

src/server/db_slice.cc
```diff
@@ -384,6 +370,41 @@ template <typename F> struct CallbackConsumer : public DbSlice::ChangeConsumerIn
...
+  uint64_t& slot_memory = db->slots_stats[sid].memory_bytes;
+  if (delta < 0 && slot_memory < uint64_t(-delta)) {
+    LOG_EVERY_T(DFATAL, 1) << "Encountered underflow of per-slot memory usage: " << slot_memory
+                           << " + " << delta << ", slot: " << sid;
+    delta = -int64_t(slot_memory);
+  }
+  slot_memory += delta;
+}
+
+void AccountSlotTieredBytes(string_view key, int64_t delta, DbTable* db) {
```
romange, line 397: @vyavdoshenko  did you look at this code before sending to me?

romange:
@vyavdoshenko  lets split this PR to ram only first, and separately  to account for tiered bytes. RAM only is what we really need to slots migration.

PR #7960, opened 2026-07-29
fix(cluster): enforce db0-only in cluster mode and fix two slot-ops crashes

vyavdoshenko:
Cluster mode is db0-only by design, but only `SELECT` enforced it: `MOVE` had no cluster-mode check, and an ACL user created with `$db` is switched to that db at AUTH time. Keys reaching db > 0 through these routes are invisible to `GETSLOTINFO`, survive `FLUSHSLOTS`, and a db > 0 write during an in-flight flush aborts debug builds. Two more crashes lived in the same code.

Changes:

- Reject `MOVE` in cluster mode, matching `SELECT`.
- Reject `ACL SETUSER ... $db` with `db != 0` in cluster mode (`$0` and `$ALL` stay allowed).
- The flush's `on_change` callback fires for every db index but deleted through db 0's table, underflowing its memory accounting (FATAL in debug) and erasing through the wrong prime table. It now ignores db != 0.
- `GetSlotStats` dereferenced the null `slots_stats` of non-cluster tables - SIGSEGV reachable via `GETSLOTINFO` replayed on an emulated-mode replica. Now returns empty stats.

### DIFF
--- a/src/server/acl/acl_family.cc
+++ b/src/server/acl/acl_family.cc
@@ -35,6 +35,7 @@
 #include "server/acl/acl_commands_def.h"
 #include "server/acl/acl_log.h"
 #include "server/acl/validator.h"
+#include "server/cluster_support.h"
 #include "server/command_registry.h"
 #include "server/common.h"
 #include "server/config_registry.h"
@@ -1116,6 +1117,10 @@ std::variant<User::UpdateRequest, ErrorReply> AclFamily::ParseAclSetUser(
     if (auto res = MaybeParseAclDflySelect(facade::ToSV(arg), dbnum_); res) {
       if (req.select_db) {
         return ErrorReply("ERR Error, select db $ was used twice");
+      }
+      // AUTH switches the connection to this db, bypassing the SELECT restriction.
+      if (IsClusterEnabled() && *res != 0 && *res != std::numeric_limits<size_t>::max()) {
+        return ErrorReply("ERR Error, select db $ is not allowed in cluster mode");
       }
       req.select_db = res;
       continue;
--- a/src/server/cluster/cluster_family.cc
+++ b/src/server/cluster/cluster_family.cc
@@ -501,9 +501,13 @@ void DeleteSlots(Transaction* trans, const SlotRanges& slots_ranges) {
     return;
   }
 
+  // Namespaced keys must not survive the slot hand-off.
+  auto all_namespaces = namespaces->GetAllNamespaces();
   trans->Execute(
-      [&slots_ranges](Transaction* t, EngineShard* shard) {
-        namespaces->GetDefaultNamespace().GetDbSlice(shard->shard_id()).FlushSlots(slots_ranges);
+      [&](Transaction* t, EngineShard* shard) {
+        for (Namespace* ns : all_namespaces) {
+          ns->GetDbSlice(shard->shard_id()).FlushSlots(slots_ranges);
+        }
         return OpStatus::OK;
       },
       true);
@@ -696,6 +700,7 @@ void ClusterFamily::DflyClusterGetSlotInfo(CmdArgParser parser, CommandContext* 
   }
 
   fb2::Mutex mu;
+  auto all_namespaces = namespaces->GetAllNamespaces();
 
   auto cb = [&](auto*) ABSL_LOCKS_EXCLUDED(mu) {
     EngineShard* shard = EngineShard::tlocal();
@@ -703,8 +708,11 @@ void ClusterFamily::DflyClusterGetSlotInfo(CmdArgParser parser, CommandContext* 
       return;
 
     util::fb2::LockGuard lk(mu);
-    for (auto& [slot, data] : slots_stats) {
-      data += namespaces->GetDefaultNamespace().GetDbSlice(shard->shard_id()).GetSlotStats(slot);
+    for (Namespace* ns : all_namespaces) {
+      auto& db_slice = ns->GetDbSlice(shard->shard_id());
+      for (auto& [slot, data] : slots_stats) {
+        data += db_slice.GetSlotStats(slot);
+      }
     }
   };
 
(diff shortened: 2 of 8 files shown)


romange:
It can not and it's redundant PR. Cluster mode does not allow "SELECT" and multiple databases

vyavdoshenko:
> It can not and it's redundant PR. Cluster mode does not allow "SELECT" and multiple databases

Agreed that cluster mode is meant to be db0-only - but only `SELECT` enforces it. Two routes bypass it on `main`:

1. `MOVE` has no cluster-mode check (this PR adds one):
```
SET {mv}foo bar
MOVE {mv}foo 1                      -> (integer) 1
DFLYCLUSTER GETSLOTINFO SLOTS 8999  -> key_count 0   # invisible
DFLYCLUSTER FLUSHSLOTS 8999 8999    -> OK            # key survives in db 1, served again after the slot returns
```

2. `ACL SETUSER u1 on >pw ~* +@all $1` pins the connection to db 1 at AUTH time, no `SELECT` involved. A db-1 write during an in-flight FLUSHSLOTS hits the flush's `on_change`, which deleted through db 0's table: debug aborts, release erases through the wrong prime table.

The PR also fixes a SIGSEGV that needs no multiple databases (`GETSLOTINFO` replayed on an emulated-mode replica dereferences null `slots_stats`). With only db 0 present the new loops degenerate to the previous single-table behavior.

Happy to reduce this to strict enforcement instead (reject `$db != 0` in cluster mode + MOVE check + crash fixes, slot ops stay db0-only) - the trade-off is that keys already in db > 0 stay invisible and survive migrations.

kostasrim:
> > It can not and it's redundant PR. Cluster mode does not allow "SELECT" and multiple databases
> 
> Agreed that cluster mode is meant to be db0-only - but only `SELECT` enforces it. Two routes bypass it on `main`:
> 
> 1. `MOVE` has no cluster-mode check (this PR adds one):
> 
> ```
> SET {mv}foo bar
> MOVE {mv}foo 1                      -> (integer) 1
> DFLYCLUSTER GETSLOTINFO SLOTS 8999  -> key_count 0   # invisible
> DFLYCLUSTER FLUSHSLOTS 8999 8999    -> OK            # key survives in db 1, served again after the slot returns
> ```
> 
> 2. `ACL SETUSER u1 on >pw ~* +@all $1` pins the connection to db 1 at AUTH time, no `SELECT` involved. A db-1 write during an in-flight FLUSHSLOTS hits the flush's `on_change`, which deleted through db 0's table: debug aborts, release erases through the wrong prime table.
> 
> The PR also fixes a SIGSEGV that needs no multiple databases (`GETSLOTINFO` replayed on an emulated-mode replica dereferences null `slots_stats`). With only db 0 present the new loops degenerate to the previous single-table behavior.
> 
> Happy to reduce this to strict enforcement instead (reject `$db != 0` in cluster mode + MOVE check + crash fixes, slot ops stay db0-only) - the trade-off is that keys already in db > 0 stay invisible and survive migrations.

I haven't followed the context but as far as I remember for (2) it's not true. The $1 is our extension and `pins the connection to db 1` is a wrong statement (plz double check me). What happens is it `restricts the connection to db1`. A new connection still uses db 0 and has its commands rejected until it explicitly switches via the select command. 

It could be that my memory "cheats me" so plz double check my words/behaviour here 🙏

vyavdoshenko:
> > > It can not and it's redundant PR. Cluster mode does not allow "SELECT" and multiple databases
> > 
> > 
> > Agreed that cluster mode is meant to be db0-only - but only `SELECT` enforces it. Two routes bypass it on `main`:
> > 
> > 1. `MOVE` has no cluster-mode check (this PR adds one):
> > 
> > ```
> > SET {mv}foo bar
> > MOVE {mv}foo 1                      -> (integer) 1
> > DFLYCLUSTER GETSLOTINFO SLOTS 8999  -> key_count 0   # invisible
> > DFLYCLUSTER FLUSHSLOTS 8999 8999    -> OK            # key survives in db 1, served again after the slot returns
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 2. `ACL SETUSER u1 on >pw ~* +@all $1` pins the connection to db 1 at AUTH time, no `SELECT` involved. A db-1 write during an in-flight FLUSHSLOTS hits the flush's `on_change`, which deleted through db 0's table: debug aborts, release erases through the wrong prime table.
> > 
> > The PR also fixes a SIGSEGV that needs no multiple databases (`GETSLOTINFO` replayed on an emulated-mode replica dereferences null `slots_stats`). With only db 0 present the new loops degenerate to the previous single-table behavior.
> > Happy to reduce this to strict enforcement instead (reject `$db != 0` in cluster mode + MOVE check + crash fixes, slot ops stay db0-only) - the trade-off is that keys already in db > 0 stay invisible and survive migrations.
> 
> I haven't followed the context but as far as I remember for (2) it's not true. The $1 is our extension and `pins the connection to db 1` is a wrong statement (plz double check me). What happens is it `restricts the connection to db1`. A new connection still uses db 0 and has its commands rejected until it explicitly switches via the select command.
> 
> It could be that my memory "cheats me" so plz double check my words/behaviour here 🙏

Double-checked - it does both. The validator side indeed *restricts* (rejects MOVE and cross-db transactional commands for `$db` users), but AUTH also *switches*: `DoAuth` sets `cntx->conn_state.db_index = cred.db` and activates the db on all shards. No SELECT involved.

Verified live on `--cluster_mode=yes` with slots configured:
```
ACL SETUSER u1 on >pw ~* +@all $1   -> OK
(as u1) SELECT 1                    -> ERR SELECT is not allowed in cluster mode
(as u1) SET db1key hello            -> OK        # no SELECT needed
INFO keyspace                       -> db0:keys=0  db1:keys=1
```

kostasrim:
> but AUTH also _switches_: `DoAuth` s

you got me first! I just verified this is indeed the case. once the user auth's we autoselect!

romange (approved):

PR #7961, opened 2026-07-29
chore(hll): sync hyperloglog.c with valkey upstream

romange:
## Summary

Re-syncs `src/redis/hyperloglog.c`, forked from valkey `56ec1ff1c` two years ago, with upstream `06d30f293`. The file now keeps upstream's own formatting (`src/redis/**` is excluded from pre-commit clang-format, so this is free), which is what makes the remaining diff readable.

**Divergence from upstream: 2139 → 767 differing lines**, and the remainder is accounted for:

| | lines | |
|---|---|---|
| shared library half (lines 1-1620) | **72** | all load-bearing — see below |
| upstream command/reply layer | 695 | `pfaddCommand`/`pfcountCommand`/`pfmergeCommand`/`pfselftestCommand`/`pfdebugCommand`/`isHLLObjectOrReply` vs the `HllBufferPtr` API that replaces them |

The 72 lines in the shared half are: the include block, `HLL_SPARSE_MAX_BYTES` replacing `server.hll_sparse_max_bytes`, dropping the unused `invalid_hll_err`, the `robj`/`objectGetVal` → `sds *`/`hll` conversion in the three sparse helpers, the `promoted` out-param, and five `#if 0`/`#endif` markers. No comment- or formatting-only differences are left. Upstream functions that have a Dragonfly counterpart (`hllAdd`, `hllMerge`, `createHLLObject`) sit under `#if 0` as its reference, which is both cheaper than deleting them and useful documentation; the command/reply layer is genuinely irrelevant here and stays deleted. The provenance note (sync point + the list of deliberate deltas) lives in `hyperloglog.h`, so it costs nothing against upstream.

## Imported from upstream

- **CVE-2025-32023** (`20f5199d9`) — run-length overflow guards on every sparse opcode, not just `VAL`. A sparse payload whose `XZERO` runs sum past `INT_MAX` wrapped the `idx` cursor negative, so the trailing `VAL` slipped past the guard and wrote a register through a wild pointer. Reachable from `PFCOUNT`, `PFMERGE` and `PFADD` on any `SET` string carrying a `HYLL` + sparse header.
- **`__builtin_ctzll()` in `hllPatLen()`** (`744b13e30`) in place of the trailing-zero loop.
- **AVX2 (`3df609ef0`) and ARM NEON (`dd772c490`) `hllMergeDense()`/`hllDenseCompress()` kernels** and their runtime dispatchers. `PFMERGE` and multi-key `PFCOUNT` now go through them. `src/redis/config.h` gains upstream's `HAVE_X86_SIMD`, `HAVE_ARM_NEON`, `ATTRIBUTE_TARGET_*` and `VALKEY_NO_SANITIZE` macros.
- `VALKEY_NO_SANITIZE("alignment")` on `MurmurHash64A`. Including `config.h` also makes `BYTE_ORDER` actually defined there — it previously wasn't, and the little-endian branch was taken only because `0 == 0`.

`06d30f293` itself (an `hllSparseSet` return-type regression from upstream's small-strings refactor) never applied here, but its behaviour — `PFADD` on a corrupted sparse HLL must report `INVALIDOBJ` — is now covered.

## Dragonfly-local bugs the sync exposed

- **`pfcountMulti()` merged into the raw register array `HLL_HDR_SIZE` bytes too low.** `hllCount(HLL_RAW)` reads `hdr->registers` (`max + 16`) but the merge wrote from `max`, so every multi-key `PFCOUNT` dropped registers 0-15 and counted 16 zeroed ones instead. It did not crash only because `encoding = HLL_RAW = 255` is larger than any register value, so the merge never overwrote it.
- **`goto promote` in `hllSparseSet()` jumped over the `sds hll` initialization**, leaving the promote path dereferencing an uninitialized pointer. It needs a hash with 32+ zero bits above the 14-bit register index, so random data reaches it about once in 2³² elements — but `MurmurHash64A` is a bijection on 8-byte inputs, so an element that lands there can be constructed directly (see Tests).
- **`pfadd_sparse()` invalidated the cached cardinality through a stale header pointer** captured before `hllSparseAdd()`, which may `sdsResize` (moving `s_realloc`) or `sdsfree` on promotion. ASAN misses it because sds allocates through mimalloc.

`pfmerge()`'s dense fast path overwrites the destination registers rather than raising them, matching upstream; the header doc is updated to say so. The only caller always passes a freshly created dense HLL, so behaviour is unchanged.

## Tests

5 new cases in `hll_family_test.cc`. Each was checked against the unfixed code rather than trusted on a green pass:

| Test | Unfixed behaviour |
|---|---|
| `CorruptedSparseRunLengthOverflow` | `SIGSEGV` in `convertSparseToDenseHll` via `PFCOUNT` |
| `CountMultipleAgreesWithMerge` | `40207` vs `40332` |
| `SparseSetPromotesOnLargeCount` | no promotion, uninitialized `sds` on the promote path |
| `SimdMatchesScalar` | mutation-tested: 4 injected AVX2 faults (mask, shift, shuffle in both kernels) all caught |
| `CorruptedSparseTruncatedRun` | passes before and after — coverage for the ZERO/XZERO arm, not a regression test |

`SparseSetPromotesOnLargeCount` drives the promote branch through a real `PFADD` rather than calling `hllSparseSet()` directly. Every step of `MurmurHash64A` — multiply by the odd constant `m`, `x ^= x >> 47`, the xors with the seed — is invertible, so for an 8-byte element you can pick the hash you want and solve for the input. The element in the test hashes to 35 zero bits above the register index (count 36); the same inversion reaches any count up to the `HLL_Q + 1 = 51` ceiling.

`SimdMatchesScalar` builds dense HLLs spanning all 64 register values; `PFADD` data only ever fills the low bits, and an earlier version of the test using `PFADD` failed to catch a real fault.

## Benchmarks

5 `--bench` benchmarks in the same file, calling `hyperloglog.c` directly so the numbers are not diluted by command parsing, the transaction framework or key lookups. Each takes a `simd:0`/`simd:1` argument driven by `hllEnableSimd()`; the `simd:1` rows skip themselves on a CPU with neither AVX2 nor NEON, so you never get two rows silently timing the same scalar code.

Build a release binary — at `-O0` the scalar C loops are penalized far more than they are in production:

```
./helio/blaze.sh -release -DWITH_AWS=OFF -DWITH_GCP=OFF
cd build-opt && ninja hll_family_test
```

Then run it pinned to one core. The gtest filter matters: without it the 21 tests run first (`--gtest_filter=-` does *not* skip them — it excludes nothing).

```
$ taskset -c 4 build-opt/hll_family_test --bench --gtest_filter=-'*' --benchmark_min_time=0.5s

Note: Google Test filter = -*
[==========] Running 0 tests from 0 test suites.
[==========] 0 tests from 0 test suites ran. (0 ms total)
[  PASSED  ] 0 tests.
WARNING: filter "-*" did not match any test; no tests were run
2026-07-30T09:38:41+03:00
Running ./hll_family_test
Run on (24 X 5482 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x12)
  L1 Instruction 32 KiB (x12)
  L2 Unified 1024 KiB (x12)
  L3 Unified 32768 KiB (x2)
Load Average: 1.02, 1.10, 1.73
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
***WARNING*** ASLR is enabled, the results may have unreproducible noise in them.
-------------------------------------------------------------------------------------------
Benchmark                                 Time             CPU   Iterations UserCounters...
-------------------------------------------------------------------------------------------
BM_HllMergeDense/simd:0               12582 ns        12582 ns        41273 items_per_second=1.30219G/s
BM_HllMergeDense/simd:1                 383 ns          383 ns      1821945 items_per_second=42.7621G/s
BM_HllDenseCompress/simd:0            66762 ns        66761 ns        10226 items_per_second=245.411M/s
BM_HllDenseCompress/simd:1              466 ns          466 ns      1502399 items_per_second=35.1731G/s
BM_HllDenseCompressUnchained           3676 ns         3676 ns       187047 items_per_second=4.45658G/s
BM_PfMerge/simd:0/sources:1           77648 ns        77647 ns         8997 items_per_second=211.006M/s
BM_PfMerge/simd:1/sources:1             944 ns          944 ns       744688 items_per_second=17.3546G/s
BM_PfMerge/simd:0/sources:8          154197 ns       154198 ns         4529 items_per_second=850.022M/s
BM_PfMerge/simd:1/sources:8            3625 ns         3624 ns       193324 items_per_second=36.1629G/s
BM_PfCountMulti/simd:0/sources:1      15164 ns        15164 ns        46004 items_per_second=1.08048G/s
BM_PfCountMulti/simd:1/sources:1       3760 ns         3760 ns       186322 items_per_second=4.35774G/s
BM_PfCountMulti/simd:0/sources:8      92929 ns        92930 ns         7535 items_per_second=1.41044G/s
BM_PfCountMulti/simd:1/sources:8       6446 ns         6446 ns       108804 items_per_second=20.3323G/s
```

Both buffers stay L1-resident across iterations, so these are compute-bound figures; a `PFMERGE` over cold keys will look different. CPU scaling is on, as the run's own warning says.

**`BM_HllDenseCompress/simd:0` is not a fair vectorization baseline.** `HLL_DENSE_SET_REGISTER` does four byte-wide read-modify-writes per register and consecutive registers share bytes, so every load waits on the previous store to forward: the loop is one dependency chain at ~20 cycles per register (4.06 ns/register, cv 0.01% over 5 repetitions). `BM_HllDenseCompressUnchained` is a scalar reference — not production code, and `CHECK`ed byte-identical against `hllDenseCompress()` — that builds each output byte once instead of updating it twice:

| | time | vs AVX2 |
|---|---|---|
| `BM_HllDenseCompress/simd:0` (in-tree scalar) | 66762 ns | 143× |
| `BM_HllDenseCompressUnchained` | 3676 ns | **7.9×** |
| `BM_HllDenseCompress/simd:1` (AVX2) | 466 ns | — |

So the 143× is ~18× store-forwarding stall × ~8× vectorization. `BM_HllMergeDense` has no such problem — `HLL_DENSE_GET_REGISTER` is read-only, so its 33× is a genuine width-bound result against 32 registers per AVX2 iteration.

Side note, not addressed here: that leaves ~18× on the table for `hllDenseCompress()` on any build without AVX2 or NEON. It is upstream valkey code, so it belongs upstream rather than as a local divergence in a file this PR is trying to bring *closer* to upstream.

## Also run green

`hll_family_test` (21), `string_family_test` (45), `rdb_test` HLL cases, `tiered_storage_test` PF cases. Release build compiles clean. The NEON kernels are syntax-checked for aarch64 with clang but not executed locally (x86 host) — the ARM CI runner covers them.

### DIFF
--- a/src/redis/hyperloglog.c
+++ b/src/redis/hyperloglog.c
@@ -1,7 +1,7 @@
-/* hyperloglog.c - Redis HyperLogLog probabilistic cardinality approximation.
- * This file implements the algorithm and the exported Redis commands.
+/* hyperloglog.c - HyperLogLog probabilistic cardinality approximation.
+ * This file implements the algorithm and the exported commands.
  *
- * Copyright (c) 2014, Salvatore Sanfilippo <antirez at gmail dot com>
+ * Copyright (c) 2014, Redis Ltd.
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -28,25 +28,44 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  * POSSIBILITY OF SUCH DAMAGE.
  */
+/*
+ * Copyright (c) Valkey Contributors
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ * Last synced with valkey 06d30f293
+ */
 
 #include "redis/hyperloglog.h"
 
+#include <stdint.h>
 #include <math.h>
 #include <string.h>
 
+#include "redis/config.h"
 #include "redis/redis_aux.h"
 #include "redis/util.h"
 
 #define min(a, b) ((a) < (b) ? (a) : (b))
 
-/* The Redis HyperLogLog implementation is based on the following ideas:
+#if HAVE_X86_SIMD
+/* Define __MM_MALLOC_H to prevent importing the memory aligned
+ * allocation functions, which we don't use. */
+#define __MM_MALLOC_H
+#include <immintrin.h>
+#endif
+
+#if HAVE_ARM_NEON
+#include <arm_neon.h>
+#endif
+
+/* The HyperLogLog implementation is based on the following ideas:
  *
  * * The use of a 64 bit hash function as proposed in [1], in order to estimate
  *   cardinalities larger than 10^9, at the cost of just 1 additional bit per
  *   register.
  * * The use of 16384 6-bit registers for a great level of accuracy, using
  *   a total of 12k per key.
- * * The use of the Redis string data type. No new type is introduced.
+ * * The use of the string data type. No new type is introduced.
  * * No attempt is made to compress the data structure as in [1]. Also the
  *   algorithm used is the original HyperLogLog Algorithm as in [2], with
  *   the only difference that a 64 bit hash function is used, so no correction
@@ -58,7 +77,7 @@
  * [2] P. Flajolet, Éric Fusy, O. Gandouet, and F. Meunier. Hyperloglog: The
  *     analysis of a near-optimal cardinality estimation algorithm.
  *
- * Redis uses two representations:
+ * We use two representations:
  *
  * 1) A "dense" representation where every entry is represented by
  *    a 6-bit integer.
@@ -93,7 +112,7 @@
  * Dense representation
  * ===
  *
- * The dense representation used by Redis is the following:
+ * The dense representation is the following:
  *
  * +--------+--------+--------+------//      //--+
  * |11000000|22221111|33333322|55444444 ....     |
@@ -183,23 +202,23 @@
  * when this implementation switches to the dense representation is
  * configured via the define HLL_SPARSE_MAX_BYTES.
  */
+
 #define HLL_SPARSE_MAX_BYTES 3000
 
 struct hllhdr {
-  char magic[4];       /* "HYLL" */
-  uint8_t encoding;    /* HLL_DENSE or HLL_SPARSE. */
-  uint8_t notused[3];  /* Reserved for future use, must be zero. */
-  uint8_t card[8];     /* Cached cardinality, little endian. */
-  uint8_t registers[]; /* Data bytes. */
+    char magic[4];       /* "HYLL" */
+    uint8_t encoding;    /* HLL_DENSE or HLL_SPARSE. */
+    uint8_t notused[3];  /* Reserved for future use, must be zero. */
+    uint8_t card[8];     /* Cached cardinality, little endian. */
+    uint8_t registers[]; /* Data bytes. */
 };
 
 /* The cached cardinality MSB is used to signal validity of the cached value. */
 #define HLL_INVALIDATE_CACHE(hdr) (hdr)->card[7] |= (1 << 7)
 #define HLL_VALID_CACHE(hdr) (((hdr)->card[7] & (1 << 7)) == 0)
 
-#define HLL_P 14 /* The greater is P, the smaller the error. */
-#define HLL_Q                                                                           \
-  (64 - HLL_P)                         /* The number of bits of the hash value used for \
+#define HLL_P 14                       /* The greater is P, the smaller the error. */
+#define HLL_Q (64 - HLL_P)             /* The number of bits of the hash value used for \
                                           determining the number of leading zeros. */
 #define HLL_REGISTERS (1 << HLL_P)     /* With P=14, 16384 registers. */
 #define HLL_P_MASK (HLL_REGISTERS - 1) /* Mask to index register. */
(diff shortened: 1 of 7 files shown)


romange:
[murmur_invert.py](https://github.com/user-attachments/files/30533854/murmur_invert.py)

creates an input with arbitrary long 0-tail hashes.

src/server/hll_family_test.cc
```diff
@@ -1,6 +1,18 @@
 // Copyright 2023, DragonflyDB authors.  All rights reserved.
 // See LICENSE for licensing terms.
 //
+extern "C" {
```
kostasrim, line 4: Don't we compile it as C++ anyway ?

kostasrim (approved):
seems the tests cover all the cases/bugs 👍

src/redis/hyperloglog.c
```diff
@@ -341,31 +391,31 @@ struct hllhdr {
...
-    target = ((b0 >> _fb) | (b1 << _fb8)) & HLL_REGISTER_MAX; \
-  } while (0)
+#define HLL_DENSE_GET_REGISTER(target, p, regnum)                 \
+    do {                                                          \
+        uint8_t *_p = (uint8_t *)p;                               \
+        unsigned long _byte = regnum * HLL_BITS / 8;              \
+        unsigned long _fb = regnum * HLL_BITS & 7;                \
+        unsigned long _fb8 = 8 - _fb;                             \
+        unsigned long b0 = _p[_byte];                             \
+        unsigned long b1 = _p[_byte + 1];                         \
```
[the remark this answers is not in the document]
romange: fixed

src/server/hll_family_test.cc
romange: no

PR #7964, opened 2026-07-30
refactor: tune OAH for memory consumption

BorysTheDev:
Tune OAH allocation and compaction around its 2× overload factor, reducing bucket memory while ensuring SHRINK never selects an invalid capacity.
Updates reserve/growth logic and OAH-enabled regression, benchmark, and sizing expectations for the new capacity behavior.

### DIFF
--- a/src/core/oah_set.h
+++ b/src/core/oah_set.h
@@ -62,8 +62,8 @@ class OAHSet : public OAHTable<OAHEntry> {
 
  private:
   bool AddImpl(std::string_view content, uint32_t len, uint32_t ttl_sec) {
-    if (size_ >= entries_.size()) [[unlikely]] {
-      Reserve(BucketCount() * 2);
+    if (size_ >= entries_.size() * kOverloadFactor) [[unlikely]] {
+      GrowCapacity(BucketCount() * 2);
     }
     assert(Capacity() >= kDisplacementSize);
 
--- a/src/core/oah_table.h
+++ b/src/core/oah_table.h
@@ -42,9 +42,13 @@ template <typename Entry> class OAHTable {  // Open Addressing Hash table
   using Buckets = std::vector<TaggedPtr, StatelessAllocator<TaggedPtr>>;
 
  public:
-  static constexpr std::uint32_t kShiftLog = 2;                         // TODO make template
+  static constexpr std::uint32_t kShiftLog = 4;                         // TODO make template
   static constexpr std::uint32_t kMinCapacityLog = kShiftLog;           // should be >= ShiftLog
   static constexpr std::uint32_t kDisplacementSize = (1 << kShiftLog);  // TODO check
+
+  // Table grows once live entries reach capacity * kOverloadFactor (200% load, by default).
+  static constexpr std::uint32_t kOverloadFactor = 2;
+  static constexpr size_t kMinBucketCount = 1u << kMinCapacityLog;
 
   class iterator {
    public:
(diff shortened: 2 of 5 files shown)


BorysTheDev:
augment review

src/core/oah_set.h
```diff
@@ -62,8 +62,8 @@ class OAHSet : public OAHTable<OAHEntry> {
 
  private:
   bool AddImpl(std::string_view content, uint32_t len, uint32_t ttl_sec) {
-    if (size_ >= entries_.size()) [[unlikely]] {
-      Reserve(BucketCount() * 2);
+    if (size_ >= entries_.size() * kOverloadFactor) [[unlikely]] {
```
vyavdoshenko, line 65: Allowing size_ to grow up to 2 * entries_.size() makes vector-backed buckets unavoidable once the number of members exceeds the number of physical slots. However, GetRandomMember() randomizes only the physical bucket, and PickFromBucket() always returns the first live entry in a PtrVector. Consequently, all later entries in that vector have zero probability of being returned by SRANDMEMBER or SPOP. Please make random selection vector-aware before allowing load factors above 1.

src/core/oah_table.h
```diff
@@ -194,27 +198,21 @@ template <typename Entry> class OAHTable {  // Open Addressing Hash table
...
-  //
-  // Shrinks the table to new_size (power of 2, >= 1 << kMinCapacityLog and >= element count).
+  // Shrinks toward `new_size` buckets, never below what's needed for live entries at
+  // kOverloadFactor. No-op if not smaller than the current bucket count.
   void Shrink(size_t new_size) {
-    assert(absl::has_single_bit(new_size));
-    assert(new_size >= (1u << kMinCapacityLog));
-    assert(new_size < entries_.size());
+    size_t required = size_ / kOverloadFactor + (size_ % kOverloadFactor != 0);
+    size_t target = absl::bit_ceil(std::max({new_size, kMinBucketCount, required}));
```
vyavdoshenko, line 210: ServerFamily::Shrink still passes bit_ceil(UpperBoundSize()) as new_size. For normal OAH sets this value always dominates required = ceil(size / 2), so SHRINK continues compacting to approximately 100% load instead of the intended 200% load. There is also a no-op case for sets with up to eight members: the server enters FindMutable(), but this clamp keeps the existing 16 buckets, causing version/WATCH/client-tracking invalidation despite returning zero. Please use the same OAH-aware target in both the server precheck and mutation path.

src/core/oah_table.h
```diff
@@ -42,10 +42,14 @@ template <typename Entry> class OAHTable {  // Open Addressing Hash table
   using Buckets = std::vector<TaggedPtr, StatelessAllocator<TaggedPtr>>;
 
  public:
-  static constexpr std::uint32_t kShiftLog = 2;                         // TODO make template
+  static constexpr std::uint32_t kShiftLog = 4;                         // TODO make template
```
vyavdoshenko, line 45: Increasing kShiftLog from 2 to 4 also increases the minimum allocation from 7 to 31 TaggedPtr slots, or from 56 to 248 bytes, even for a single-member OAH set. Could we add a small-cardinality memory benchmark or document this trade-off? The existing benchmarks start at 1,000 elements and do not capture this regression.

src/core/oah_set.h
BorysTheDev: it's ok for now, I will consider this improvement later

src/core/oah_table.h
BorysTheDev: it's ok

src/core/oah_table.h
BorysTheDev: it's ok

vyavdoshenko (approved):

PR #7965, opened 2026-07-30
chore: cuckoo compatibility fixes and missing flags

kostasrim:
* replace default flags with absl flags and allow their modification via config set (similar to valkey)
* fix some corner cases on command parsing/check for cuckoo
* add tests that cover all the corner cases fixed"
* rollback to old flag value when config set command cb fails to validate it (parsing succeeded but cb did not)


kostasrim:
augment review

src/server/cuckoo_filter_family.cc
```diff
@@ -44,10 +87,14 @@ OpResult<bool> OpAdd(const OpArgs& op_args, string_view key, string_view item) {
...
-    pv.SetCuckooFilter(CuckooFilterOptions{.capacity = kDefaultCapacity});
+    pv.SetCuckooFilter(NewFilterOptions(tl_cf_initial_size));
   }
 
-  if (!pv.GetCuckooFilter()->Insert(CuckooFilter::Hash(item)))
+  CuckooFilter* cf = pv.GetCuckooFilter();
+  if (cf->NumFilters() >= tl_cf_max_expansions)
+    return OpStatus::CUCKOO_FILTER_MAX_EXPANSIONS;
+
+  if (!cf->Insert(CuckooFilter::Hash(item)))
```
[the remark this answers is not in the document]
kostasrim: it's compatibility :D

kostasrim:
augment review

src/facade/error.h
```diff
@@ -58,5 +58,8 @@ inline constexpr char kNoAuthErrType[] = "no_auth";
 
 inline constexpr char kBloomFilterLoadInProgress[] = "bloom filter load in progress";
 inline constexpr char kCuckooFilterFull[] = "Filter is full";
+inline constexpr char kCuckooFilterMaxExpansions[] = "Maximum expansions reached";
+inline constexpr char kCuckooFilterInvalidCapacity[] =
```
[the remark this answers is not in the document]
kostasrim: nah it's fine

src/server/config_registry.cc
```diff
@@ -53,12 +53,20 @@ auto ConfigRegistry::Set(string_view config_name, string_view value) -> SetResul
...
   CHECK(flag) << config_name;
+  string old_value = flag->CurrentValue();
   if (string error; !flag->ParseFrom(value, &error)) {
     LOG(WARNING) << error;
     return SetResult::INVALID;
   }
 
   bool success = !cb || cb(*flag);
+  if (!success) {
+    // rollback to the old value in case cb validation did not work. Otherwise, we end up with
```
kostasrim, line 64: @abhijat this is the most important change as it adds the ability to rollback a flag value if we fail to parse it.

src/server/cuckoo_filter_family.cc
```diff
@@ -408,13 +479,61 @@ void CmdCompact(CmdArgParser parser, CommandContext* cmd_cntx) {
...
 
+void RegisterCuckooFilterConfig() {
+  auto* pool = shard_set->pool();
+
+  config_registry.RegisterMutable("cf_bucket_size", [pool](const absl::CommandLineFlag& flag) {
+    auto val = flag.TryGet<uint32_t>();
+    if (!val || *val < 1 || *val > 255)
+      return false;
+    pool->AwaitBrief([v = static_cast<uint8_t>(*val)](unsigned, auto*) { tl_cf_bucket_size = v; });
+    return true;
```
abhijat, line 490: this logic is repeated a few times, but I suppose it is not easy to extract a common helper because of the different int types unless you use template parameters

kostasrim: no need to a template parameter. We can pass is as an Arg via a struct RangeCheck { left, righ };

kostasrim: nah, honestly it's the same lines just a little lest boilerplate -- not really worth it IMO

abhijat (approved):

PR #7968, opened 2026-07-31
fix(tiered): release the disk extent when a value is overwritten in place

vyavdoshenko:
An offloaded value reports no heap allocation, so overwriting it in place silently dropped the only reference to its disk extent. The extent leaked, and the next `FLUSHALL` aborted on `CHECK_EQ(db_ptr->stats.tiered_entries, 0u)` (db_slice.cc) - a `CHECK`, so it aborts in release builds too.

For the paths that also retype the key the state was **unrecoverable**: after the overwrite the value is no longer external, so `PerformDeletionAtomic` cannot release it and `RemoveOffloadedEntriesFromTieredStorage` finds nothing to delete — every subsequent `FLUSHALL`/`FLUSHDB` aborted, even on an empty keyspace.

Root cause:

Seven sites overwrite an existing `PrimeValue` in place. Most reach it through `AddOrFind(key, std::nullopt)` — deliberately untyped so the command may *retype* the key - and then call `InitRobj` / `InitSet` / `Reset` / `SetSBF`. All of those go through `CompactObj::SetMeta`, which frees only `if (HasAllocated())`, and `HasAllocated()` is false for `EXTERNAL_TAG`.

Changes:

- add `DbSlice::ReleaseOffloadedValue()` and call it before every in-place overwrite:
  - `OpRen` - `RENAME` / `RENAMENX`
  - `DbSlice::AddOrUpdateInternal` - `SORT ... STORE`, `RESTORE`, rdb load
  - `set_family` `OpAdd` - `SINTERSTORE` / `SUNIONSTORE` / `SDIFFSTORE`
  - `zset_family` `PrepareZEntry` - `ZUNIONSTORE` / `ZINTERSTORE` / `ZDIFFSTORE` / `ZRANGESTORE` /
    `GEORADIUS ... STORE`
  - `json_family` `SetFullJson` - `JSON.SET`
  - `bloom_family` `OpLoadChunk` - `BF.LOADCHUNK`
- in `set_family` and `zset_family` the replacement is now built *before* the release: freeing the extent is irreversible, while a `bad_alloc` there is caught in `Transaction::RunCallback` and returned as `OUT_OF_MEMORY`, so the old value has to stay readable. The other five sites need no reordering - they receive an already-constructed value, or allocate nothing after the release.
- 7 regression tests in `tiered_storage_test.cc`, each confirmed failing before the fix.

Verified clean and left unchanged: `hset_family` (both sites already check `IsExternal()`), `string_family` / `bitops_family` / `hll_family` (already release), and the typed `AddOrFind` sites for types that never offload.

### DIFF
--- a/src/server/bloom_family.cc
+++ b/src/server/bloom_family.cc
@@ -130,6 +130,7 @@ OpStatus OpLoadChunk(const OpArgs& op_args, std::string_view blob, std::string_v
       // existing key might not necessarily be SBF, it could be HASH/JSON, and indexed
       RemoveKeyFromIndexesIfNeeded(key, op_args.db_cntx, op_res->it->second, op_args.shard);
       db_slice.RemoveExpire(op_args.db_cntx.db_index, op_res->it);
+      db_slice.ReleaseOffloadedValue(op_args.db_cntx.db_index, &op_res->it->second);
     }
 
     op_res->it->second.SetSBF(load_result.value());
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -1187,6 +1187,11 @@ pair<int64_t, int64_t> DbSlice::ExpireParams::Calculate(uint64_t now_ms, bool ca
   return {rel_ms, ms_timestamp};
 }
 
+void DbSlice::ReleaseOffloadedValue(DbIndex db_ind, PrimeValue* pv) {
+  if (pv->IsExternal())
+    shard_owner()->tiered_storage()->Delete(db_ind, pv);
+}
+
 OpResult<int64_t> DbSlice::UpdateExpire(const Context& cntx, Iterator prime_it,
                                         const ExpireParams& params) {
   constexpr uint64_t kPersistValue = 0;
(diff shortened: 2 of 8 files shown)


vyavdoshenko:
augment review

dranikpg (approved):
Its ok as a fix, but I remember that there were similar exceptions with keeping the expiry, forgetting to remove from search indices, etc. At that point every modification code in different command families is just really complicated and they manage lots of other uncorrelated state, maybe it would have been simpler to just delete a value with the generic code path if we know that overwriting is allowed

vyavdoshenko:
> Its ok as a fix, but I remember that there were similar exceptions with keeping the expiry, forgetting to remove from search indices, etc. At that point every modification code in different command families is just really complicated and they manage lots of other uncorrelated state, maybe it would have been simpler to just delete a value with the generic code path if we know that overwriting is allowed

Agreed, and the review of this PR produced evidence for it - I found three more pre-existing bugs of exactly that shape: an empty SINTERSTORE/ZUNIONSTORE result fails to delete a destination of a different type (typed FindMutable returns WRONG_TYPE, the delete is skipped), and the search accessor treats an offloaded HASH as resident, which aborts the server on DEL/RENAME/FT.SEARCH. Filing them separately; the generic-delete-on-overwrite refactor is worth its own issue rather than riding on a crash fix.

PR #7970, opened 2026-07-31
fix(tiered): stop RENAME and MOVE from corrupting values in shared disk pages

vyavdoshenko:
A value smaller than `kMinOccupancySize` shares a 4KB disk page with others through `SmallBins`.
The page records the *key hash* each value was stashed under, and defragmentation uses that hash to find the owner - so a value must not change keys while it sits in a bin. `RENAME` and `MOVE` did exactly that, and the result was silent data corruption.

Root cause:

`SmallBins::SerializeBin` writes `CompactObj::HashCode(key)` (and the dbid) into the page.
`ShardOpManager::Defragment` recovers each entry with `prime.FindFirst(hash, predicate)`. After the key changes, that lookup no longer resolves, the entry is silently skipped (`continue`), and the page is freed while the destination `PrimeValue` still points into it. Once the freed page is handed out again, `GET` on the renamed key returns **another key's value** - no error, no crash.

Observed on a live server (40 x 200 byte values, `--tiered_min_value_size=64`): after the defrag the instance reports `tiered_entries=1` with `tiered_allocated_bytes=0` - a dangling external reference - and after the page is reused the renamed key returns a neighbour's data.

`MOVE` has the identical defect, since the page records the dbid too.

Fix:

The value is read back into memory before it leaves its key, so the stale binding never exists:

- `TieredStorage::UnstashForKeyChange`, called from `OpRen` and `OpMove` through `DbSlice::UnstashValueForKeyChange`
- only bin-backed values pay for it: a cool value already has its copy in memory, and a value owning whole pages is recovered by its segment alone
- `RENAME` / `MOVE` now return `IO_ERROR` if that read fails
- blocking on tiered storage is not new to `RENAME`: the cross-shard path already reads through `Renamer::SerializeSrc` -> `DumpToString`

### DIFF
--- a/src/server/db_slice.cc
+++ b/src/server/db_slice.cc
@@ -1192,6 +1192,12 @@ void DbSlice::ReleaseOffloadedValue(DbIndex db_ind, PrimeValue* pv) {
     shard_owner()->tiered_storage()->Delete(db_ind, pv);
 }
 
+bool DbSlice::UnstashValueForKeyChange(DbIndex db_ind, string_view key, PrimeValue* pv) {
+  if (!pv->IsExternal())
+    return true;
+  return shard_owner()->tiered_storage()->UnstashForKeyChange(db_ind, key, pv);
+}
+
 OpResult<int64_t> DbSlice::UpdateExpire(const Context& cntx, Iterator prime_it,
                                         const ExpireParams& params) {
   constexpr uint64_t kPersistValue = 0;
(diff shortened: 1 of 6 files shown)


dranikpg (commented):
I have a better suggestion, just unstash this value before moving. Those are small values, we don't loose anything on them

vyavdoshenko:
> I have a better suggestion, just unstash this value before moving. Those are small values, we don't loose anything on them

Rewrote it your way

